PR #715, opened 2015-12-03
test(CI): fix pr done on develop script

vvo:
bash is no game

hotfix


pixelastic:
I recommend http://www.shellcheck.net/ as a shell linter. Helped be catch a lot of edge-cases and made me code better :)

Issue #665, opened 2015-11-18
[numericSelector] Issue when selecting undefined


Jerskouille:
When selecting an option with the undefined value, we get an error back from Algolia.
We should probably just clearRefinements in that case.

vvo:
How can this happen that we display a value that will refine to undefined?

Jerskouille:
See the dev/app.js, popularity select.

vvo:
I am sorry the easiest way to report this bug or any UI bug is to quickly drop a GIF explaining it. So all devs will get the issue in no time VS having to try to understand the bug.

Can you do that?

Jerskouille:
![Repro](http://g.recordit.co/jQmytkDv2h.gif)

Don't mind the console logs and warnings, the issue is just with the last error displayed.

vvo:
Now I immediately get it, in particular the "undefined value" which I did not understand at all, thx

vvo:
> We should probably just clearRefinements in that case.

Seems reasonable

pixelastic:
Not sure we should `clearRefinements` in that case, this seems a bit overkill (will reset all faceting defined). I would rather simply fallback to the default index.

vvo:
> I would rather simply fallback to the default index.

Indeed seems like a better idea.

bobylito:
Will it be obvious for the user that an index is selected and which is the one selected?

bobylito:
Also I'm not sure that having 2 index selector is such a good idea. One should change the other right?

bobylito:
Also if you change the index, the previously selected filters might not be relevant. How should we deal with that?

pixelastic:
> Will it be obvious for the user that an index is selected and which is the one selected?

We will always advocate for not having an "empty" value in the list, and always select the default one. But I'm ready to bet that some users will want to add an empty element (maybe because this is how their CMS forces them to do), so we need to handle that gracefully. Going back to default is the one that will raise the least surprise I guess.

> Also I'm not sure that having 2 index selector is such a good idea. One should change the other, right?

Yes... I do not understand the UI either. Seems weird to me

> Also if you change the index, the previously selected filters might not be relevant. How should we deal with that?

That's also true, and I think we discussed it with @Jerskouille and @redox in the past already. Not sure how to handle it, but that should be a different issue.

bobylito:
> That's also true, and I think we discussed it with @Jerskouille and @redox in the past already. Not sure how to handle it, but that should be a different issue.

[...]

PR #716, opened 2015-12-03
chore(dev): fix dev app popularity selector

vvo:
Was not working with `undefined` first value (linked with #665)

Did not have the right operator nor the right values for the dataset


redox:
The idea of the `undefined` first value is to be able to have a value which is selected by default and that doesn't refine: I think we want to keep that behavior, no?

vvo:
Yes but it is currently bugged and so breaks the app thats why. But the behavior I added is the same, showing 10000 products

PR #719, opened 2015-12-04
refactor(PoweredBy): Use "Search by Algolia"

Jerskouille:
Suggestion of Guillaume D that @redox approved. Makes a bit more sense, but definitely open to discussion.
If we chose to accept that, we could maybe also rename the `PoweredBy` component and the `searchBox` parameter `poweredBy`, but this would be a breaking change.

### DIFF
--- a/src/components/PoweredBy/PoweredBy.js
+++ b/src/components/PoweredBy/PoweredBy.js
@@ -4,7 +4,7 @@ class PoweredBy extends React.Component {
   render() {
     return (
       <div className={this.props.cssClasses.root}>
-        Powered by
+        Search by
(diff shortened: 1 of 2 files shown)


redox:
@lae25c What do you think about that?

> If we chose to accept that, we could maybe also rename the PoweredBy component and the searchBox parameter poweredBy, but this would be a breaking change.

Not sure it worth the renaming.

vvo:
LGTM

pixelastic:
Could you elaborate on why "Search by Algolia" is better than "Powered by Algolia"?

redox:
> Could you elaborate on why "Search by Algolia" is better than "Powered by Algolia"?

So we understand that we're providing the search technology, not the actual content.

lae25c:
^^
Search powered by Algolia is the most clear.

PR #720, opened 2015-12-07
refactor(import): force usage of import and export ES6 keywords (+extension)

Jerskouille:
These are three separate commits that should probably be reviewed in order because of the way it has been done (mostly scripted, a few manual modifications when needed).
I didn't do three PRs because of the potential conflicts that would have followed. I think squashing them will be the easiest solution.

### style(require): Force extension for local requires

If `require` content starts with a dot, add `.js` to the `require`.
```sh
ag -l "require\([\"']\." | xargs -I {} sed -i -r "s#require\([\"'](\.[^\"']*/[^\.\"']*)[\"']\)#require('\1.js')#g" {} 
```
Few manual modifications afterwards.

### style(import): Use import over require at the root level

```sh
# For `let xxx = require('xxx');`:
ag -l "^let [a-zA-Z]+\s*=.*require\(['\"].*['\"]\);$" src/ | xargs -I {} sed -i -r "s/^let ([a-zA-Z]+)\s*=.*require\(['\"]([^'\"]*)['\"]\);$/import \1 from '\2';/g" {}

# For `require('xxx');`
ag -l "^require\(['\"].*['\"]\);$" src/ | xargs -I {} sed -i -r "s/^require\(['\"]([^'\"]*)['\"]\);$/import '\1';/g" {}
```

### style(export): Use export default instead of module.exports …

```sh
ag -l 'module.exports =' index.js src | xargs -I {} sed -i -r 's/^module.exports =/export default/g' {}
```

Special treatment for `src/lib/version.js` which is generated and
used in `scripts`, hence the update of
`scripts/bump-package-version.js`.

### Notes

As you might have noticed, the linter is now only passed on `index.js` and `src`. Scripts shouldn't rely on `node` understanding the ES6 syntax yet.
This has one drawback though : your editor will be screaming while editing js files in `scripts`. Not sure how to handle that. Maybe exclude them using `.eslintrc` ?

Also, I'm still unable to run `npm shrinkwrap`, hence the failing tests.


vvo:
> Also, I'm still unable to run npm shrinkwrap, hence the failing tests.

npm install -g npm@2 to ensure you have latest 2
Go to the branch,
do npm install
do npm install package-to-update@version --save-dev
do npm install package-to-update@version --save-dev
do npm run shrinkwrap

should work

vvo:
> Scripts shouldn't rely on node understanding the ES6 syntax yet.

I don't really get the change, was it needed? The scripts were linted previously so I wonder why the change?

We use babel-node internally for every .js script we need to run so we should be good.

Let me know

vvo:
Very good work, lots of small and big chores here!

I think the right tag would be refactor since we are using a new token, style being only "Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)" (https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit)

But: no big deal having 3 commits etc.. WFM

Jerskouille:
> npm install -g npm@2 to ensure you have latest 2
Go to the branch,
do npm install
do npm install package-to-update@version --save-dev
do npm install package-to-update@version --save-dev
do npm run shrinkwrap

Thanks, worked great. I need to run a `shrinkwrap` on `develop` too for `babel-istanbul` that I added for coverage.

vvo:
> I need to run a shrinkwrap on develop too for babel-istanbul that I added for coverage.

Ok you can maybe submit a PR, I wanted to fail testing when those were not added to shrinkwrap but it's currently not doable (works for direct dependencies not in package.json but in shrinkwrap)

Jerskouille:
> We use babel-node internally for every .js script we need to run so we should be good.

Hadn't noticed that, I'm doing the modification.

Jerskouille:
Quid of conf files, like `scripts/karma.conf.js` ?

Jerskouille:
Need review of the last commits. Then I'll squash all of them (and do the refactor renaming), because the last modifications I've done could go in each commit.

vvo:
:guitar:

PR #727, opened 2015-12-09
fix(url-sync): adds indexName in the helper configuration

bobylito:
This solves a bug found using safari and that may or may not
break an implementation based on InstantSearch.

Safari triggers a pop state event on load, which triggers a new
search. But at this point the index is not in the initial
configuration, so this search will fail with a 400, indexName not
valid.

With this fix the value for the index name is taken from the parameters
and its value is added to the initial search configuration.

### DIFF
--- a/src/components/Hits.js
+++ b/src/components/Hits.js
@@ -21,7 +21,7 @@ class Hits extends React.Component {
   }
 
   renderNoResults() {
-    let className = `${this.props.cssClasses.root} ${this.props.cssClasses.empty}`;
+    let className = this.props.cssClasses.root + ' ' + this.props.cssClasses.empty;
     return (
       <Template
         cssClass={className}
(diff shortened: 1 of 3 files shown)


redox:
But we're instantiating the helper with (https://github.com/algolia/instantsearch.js/blob/master/src/lib/InstantSearch.js#L103):

```js
    let helper = algoliasearchHelper(
      this.client,
      this.searchParameters.index || this.indexName,
      this.searchParameters
    );
```

So why would it solve the undefined index?

bobylito:
Well the url sync stores an initial configuration that is based [on the configuration of all the widgets](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/InstantSearch.js#L99). It is the [last one to be called](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/InstantSearch.js#L96) and this configuration will not contain the index name. 

This is usually not a problem. But with Safari, the [popstate event](https://developer.mozilla.org/en-US/docs/Web/Events/popstate) is triggered before we render the url with the index name (from the line you showed). When the url is read the index name is empty and the initial configuration does not contain it either. And therefore when the popstate search is done, it leads to the error.

bobylito:
Will fix the tests tomorrow

PR #728, opened 2015-12-09
fix(search-box): update value when state changes from the outside

vvo:
we were previously using the previous state I believe.

This bug can be seen by hitting the back button while not being
focused.


pixelastic:
Nice catch, LGTM.

But I'll let @bobylito have a look before merging.

bobylito:
LGTM!

PR #734, opened 2015-12-11
chore(deps): update all deps

vvo:
we were in a dependency hell I don't wanna deal with that was preventing
shrinkwrap to work

solved by upgrading what could be upgraded


src/lib/utils.js
```diff
@@ -181,6 +181,7 @@ function getRefinements(results, state) {
   forEach(state.numericRefinements, (operators, attributeName) => {
     forEach(operators, (values, operator) => {
       forEach(values, (name) => {
+        name = name + '';
```
vvo: cc @Jerskouille Tests were failing somewhere because we were comparing name: 1 to name: '1'. Then I thought: name should be a string, no matter what since it's called "name". If at some point we want the value we can still add it.

What do you think?

Jerskouille: We chose `name` instead of `value`, I don't know why, but numeric refinements are supposed to be `Number`s, don't you think ?

vvo: I do not know either, I thought you coded the getRefinements util for the current refined values widget that's why I though you knew more maybe?

Since it's currently only exposed through the "name" attribute in the current refined value template, I would go for a string since it is stringified by default.

You could pass a transformData and expect name to be a number but I think most of the cases this is not what you want. Even if people want the original value we can add it later on in a "value" prop

vvo:
We can solve this number/string issue in another PR if we need to

PR #741, opened 2015-12-15
Removed active classes on widgets with custom templates

reinaris:
Removed active classes on widgets with custom templates.
And added some consistency in js file (use of quotes and line breaks).


reinaris:
Ah should be done on develop branch, closing.

PR #742, opened 2015-12-15
Removed active classes on widgets with custom templates

reinaris:
Removed active classes on widgets with custom templates.
And added some consistency in js file (quotes and line breaks).


vvo:
Thanks!

pixelastic:
Thanks @reinaris!

reinaris:
Glad I could do something back, instantsearch.js just rocks!

Issue #743, opened 2015-12-15
`npm run docs:jsdoc` fails


pixelastic:
I cannot run `npm run docs:jsdoc`, this fails on a weird error.

```
$ npm run docs:jsdoc


> instantsearch.js@1.2.0-beta.0 docs:jsdoc /home/tim/local/www/algolia/javascript/instantsearch.js
> babel-node ./scripts/docs/gen-widget-doc.js

events.js:141
      throw er; // Unhandled 'error' event
      ^

Error: /home/tim/local/www/algolia/javascript/instantsearch.js/node_modules/jsdoc-parse/node_modules/jsdoc-75lb/lib/jsdoc/src/astnode.js:214
                var key = prop.key.name;
                                  ^

TypeError: Cannot read property 'name' of undefined
```

If, in `src/lib/InstantSearch.js`, I remove line 68 (`this.searchParameters = {...searchParameters, index: indexName};`), then it works again... I have no idea what is going on here.

vvo:
cc maybe @bobylito knows

pixelastic:
Oh yes, makes sense. I'll update.

vvo:
Handling this one

pixelastic:
Thanks. I'm curious :)

vvo:
So the real fix is here: https://github.com/jsdoc3/jsdoc/pull/1141

And waiting for https://github.com/jsdoc2md/jsdoc-to-markdown/issues/61#issuecomment-169064903 answer

PR #744, opened 2015-12-15
docs(searchParameters): Add more explicit example

pixelastic:
**do not merge yet**

I had a question on support about the usage of `searchParameters`.
I updated the doc to make it more explicit.

But I cannot regenerate the doc website from the jsdoc, `npm run docs:jsdoc` fails (see #743)


vvo:
Most of the time we try to link to https://community.algolia.com/algoliasearch-helper-js/docs/SearchParameters.html, what do you think?

pixelastic:
Oh yes, makes sense. Will update.

pixelastic:
I'm a little afraid of accepting this PR knowing that #743 is not fixed, but we have been having support questions about `searchParameters` that could have been avoided with the proper doc.

Does anyone know why #743 is failing?

vvo:
@pixelastic Can you just comment the {...} from lib/InstantSearch.js, check that the doc is well generated and then we merge this until I fix the jsdoc thing (which is a bug un jsdoc)

pixelastic:
Nice workaround. I've just pushed again with the new generated HTML, but I also have a new `currentRefinedValues.md` added to the log, but not present in the HTML doc. Is this widget officially released?

vvo:
> Is this widget officially released?

Since it's not in the html docs no but I believe the widget is done and can be used. We need to add it thought

vvo:
Ok let's merge this and mode on

pixelastic:
@Jerskouille is working on the "official" release of the currentRefinedValues I think.

PR #762, opened 2016-01-05
fix(pagination): remove default value of maxPages. Fixes #761

### LINKED ISSUE #761: [Pagination] when maxPage is not specified it's set to 20
```
search.addWidget(
  instantsearch.widgets.pagination({
    container: '#pagination-container',
    scrollTo: false
  })
);
```

Will limit the number of page to 20, this is really weird

I think it is ok to document it like this
```
search.addWidget(
  instantsearch.widgets.pagination({
    container: '#pagination-container',
    maxPages: 20,
    scrollTo: false
  })
);
```

but removing the param should not let the 20 pages limit

maxiloc:


src/widgets/pagination/__tests__/pagination-test.js
```diff
@@ -136,3 +136,71 @@ describe('pagination()', () => {
...
+      cssClasses: {active: 'ais-pagination--item__active active', disabled: 'ais-pagination--item__disabled disabled', first: 'ais-pagination--item__first first', item: 'ais-pagination--item item', last: 'ais-pagination--item__last last', link: 'ais-pagination--link link', next: 'ais-pagination--item__next next', page: 'ais-pagination--item__page page', previous: 'ais-pagination--item__previous previous', root: 'ais-pagination root'},
+      labels: {first: '«', last: '»', next: '›', previous: '‹'},
+      nbHits: 300,
+      nbPages: 30,
+      padding: 3,
+      setCurrentPage: function noRefCheck() {},
+      shouldAutoHideContainer: false,
+      showFirstLast: true
+    };
+  });
```
pixelastic: All this boilerplate (css classes, labels, spies) makes the whole tests really hard to read and understand. Any change in the way we handle CSS will break the `maxPages` tests. I know this is not especially related to this issue, but we need to find a better way.

What about:

- Movign the simple if/else/min logic into its own `getMaxPage(maxPages, result.nbPages)` method, so we can test it separatly, with all the edge cases
- Mock this method (and making it return something like `42`), then check that `Pagination` is really called with this props.

Of course, it would be better if we could simply test that a component is called with a specific value in its props.

maxiloc:
@pixelastic I changed it. What do you think ?

redox:
> @pixelastic I changed it. What do you think ?

As @vvo [said](https://github.com/algolia/instantsearch.js/issues/761#issuecomment-168971692), we cannot change it without breaking the behavior (so major update only).

vvo:
> @pixelastic I changed it. What do you think ?
> As @vvo said, we cannot change it without breaking the behavior (so major update only).

Sorry, in the end not such of a big issue: https://github.com/algolia/instantsearch.js/issues/761#issuecomment-169008737

vvo:
So we can merge this once you have re-reviewed @pixelastic

src/widgets/pagination/__tests__/pagination-test.js
```diff
@@ -136,3 +136,49 @@ describe('pagination()', () => {
...
+      first: 'first',
+      last: 'last',
+      active: 'active',
+      disabled: 'disabled'
+    };
+    results = {hits: [{first: 'hit', second: 'hit'}], nbHits: 300, hitsPerPage: 10, nbPages: 30};
+  });
+
+  it('does to have any default', () => {
+    widget = pagination({container, scrollTo: false, cssClasses});
```
pixelastic: Why do we need to put a `scrollTo: false`?

maxiloc: We don't it was a copy paste from previous test

src/widgets/pagination/__tests__/pagination-test.js
```diff
@@ -136,3 +136,49 @@ describe('pagination()', () => {
...
+    expect(widget.getMaxPage(results)).toEqual(30);
+  });
+
+  it('does reduce the number of page if lower than nbPages', () => {
+    widget = pagination({container, scrollTo: false, cssClasses, maxPages: 20});
+    expect(widget.getMaxPage(results)).toEqual(20);
+  });
```
pixelastic: This is much more readable than before, thanks! 

I still find tests are even more readable when all that is needed for it is explicitly set in the test itself. Here, part of the knowledge is in the `beforeEach` (the fact that we set `nbPages: 30`).

What about putting a `paginationOptions = {container, scrollTo: false, cssClasses}` in the `beforeEach`, and then writing tests like:

```
// Given
paginationOptions.maxPages = 20;
results.nbPages = 30;

// When
var actual = pagination(paginationOptions).getMaxPage(results);

// Then
expect(actual).toEqual(20);
```

I wouldn't say this is blocking though, I just find it easier to understand the tests that way. We split it in the initial test state, the action, and the expectation. Any thoughts @vvo?

maxiloc:
@pixelastic updated the commit

src/widgets/pagination/__tests__/pagination-test.js
vvo: Yes having only the relevant information in every test case is easier to read

pixelastic:
Perfect, LGTM, let's go

PR #764, opened 2016-01-05
chore(jsdoc): update script and fix {...} handling

vvo:


vvo:
You may ask why is the full shrinkwrap and dependencies updated? Well why is the sky blue, why rabbits are white are questions I'd like to have an answer too.

On a serious note, I was unable to only update the relevant part and also took the shameless opportunity to update things that could be updated.

pixelastic:
LGTM

PR #772, opened 2016-01-08
chore(beta): cannot release beta

vvo:
Since jsDelivr do not care about dist-tags like the @beta tag we were
using on npm:
jsdelivr/jsdelivr#8841 (comment)

We cannot provide a release-beta script because then we cannot release
any patch fixes while working on a beta.

At some point if npmcdn.com gains traction, it might be a better
choice for us: https://npmcdn.com/


bobylito:
@vvo So you remove the ability to release beta version, why? :)

vvo:
Argh sorry I had put the reasoning in the git message, now pasted in description

bobylito:
Looks great! :D

PR #777, opened 2016-01-12
feat(refinementlist): lets configure showmore feature FIX #401

### LINKED ISSUE #401: RefinementList improvement: "show XX more"
Below the list of X refinements, it would be cool to have a "Show Y more" toggling hidden values.

image

### Need

Let the user display more choices in a single `refinementList` widget. Those will be actually fetched each time but they are hidden. We would need the max number of items to display when not hidden, and also the max number of items to fetch at each search result to avoid loading too much items. We also need two templates for the 'show more' link/button so that we have the display for each of its state.

### Proposal

Here are the following new options for the `refinementList` widget : 
  - options.templates.showmore-active : template used when we displayed more facets values
  - options.templates.showmore-inactive : template used when the more facets are hidden
  - options.showmore
  - options.showmore.maxDefaultDisplayItems : number of items displayed when the state is inactive
  - options.showmore.maxFetchedFacetValues : the number of facet values fetched at each request

### Behaviour specifics

This feature relies on `maxValuesPerFacet` which is common to all the facets retrieved. If two `refinementList` use showmore, the value used in configuration should be the max() of the two.

bobylito:
The API changed slightly from the original proposition; It still adds a new object in the configuration of the refinementlist, but the attributes are : 
 - limit : for the max number of items when showing more
 - templates : 
  - active: when showing more
  - inactive: when showing less
and it reuses the currently implemented limit parameter .

### DIFF
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -5,6 +5,13 @@ import {isSpecialClick} from '../../lib/utils.js';
 import Template from '../Template.js';
 
 class RefinementList extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isShowMoreOpen: false
+    };
+  }
+
   refine(value) {
     this.props.toggleRefinement(value);
   }
@@ -45,7 +52,7 @@ class RefinementList extends React.Component {
       <div
         className={cssClassItem}
         key={key}
-        onClick={this.handleClick.bind(this, facetValue[this.props.attributeNameKey])}
+        onClick={this.handleItemClick.bind(this, facetValue[this.props.attributeNameKey])}
       >
         <Template data={templateData} templateKey="item" {...this.props.templateProps} />
         {subList}
@@ -69,7 +76,7 @@ class RefinementList extends React.Component {
   //
   // Finally, we always stop propagation of the event to avoid multiple levels RefinementLists to fail: click
   // on child would click on parent also
-  handleClick(value, e) {
+  handleItemClick(value, e) {
     if (isSpecialClick(e)) {
       // do not alter the default browser behavior
       // if one special key is down
@@ -101,6 +108,11 @@ class RefinementList extends React.Component {
     this.refine(value);
   }
 
+  handleClickShowMore() {
+    const isShowMoreOpen = !this.state.isShowMoreOpen;
+    this.setState({isShowMoreOpen});
+  }
+
   render() {
     // Adding `-lvl0` classes
     let cssClassList = [this.props.cssClasses.list];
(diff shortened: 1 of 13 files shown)


src/components/RefinementList/RefinementList.js
```diff
@@ -108,9 +120,20 @@ class RefinementList extends React.Component {
...
 
+    const limit = this.state.isShowMoreOpen ? this.props.limitMax : this.props.limitMin;
+    const showmoreBtn =
+      this.props.showMore ?
+        <Template
+          onClick={() => this.handleClickShowMore()}
+          templateKey={'showmore-' + (this.state.isShowMoreOpen ? 'active' : 'inactive')}
+          {...this.props.templateProps}
+        /> :
+        undefined;
```
vvo: Always wondering, should it be `null` or `undefined` here?

vvo: (tested and both seems to trigger the same behavior when rendered, no extra `<div>`)

src/components/RefinementList/__tests__/RefinementList-test.js
```diff
@@ -56,8 +56,8 @@ describe('RefinementList', () => {
         </div>
       </div>
     );
-    expect(out.props.children[0].key).toEqual('facet1');
-    expect(out.props.children[1].key).toEqual('facet2');
+    expect(out.props.children[0][0].key).toEqual('facet1');
```
vvo: I fail at understanding why this test changed, the underlying DOM tree is different than before I guess but looking at the changes of RefinementList I do not see it. Maybe because of changes inside Template?

src/widgets/refinement-list/refinement-list.js
```diff
@@ -44,24 +50,31 @@ refinementList({
...
     operator = 'or',
     sortBy = ['count:desc'],
-    limit = 1000,
+    limit = 10,
     cssClasses: userCssClasses = {},
     templates = defaultTemplates,
     transformData,
-    autoHideContainer = true
+    autoHideContainer = true,
+    showMore
```
vvo: Can we add a test for the new option in refinement-list-test.js?

bobylito: Yes sure :)

src/components/RefinementList/__tests__/RefinementList-test.js
bobylito: It's because we have 2 children, therefore the children is not a single element but an array of elements. https://facebook.github.io/react/tips/children-props-type.html

src/components/RefinementList/RefinementList.js
bobylito: Cool :)

src/components/RefinementList/__tests__/RefinementList-test.js
vvo: ok!

vvo:
@bobylito I believe you only miss a test for the new option in the widget test

bobylito:
Updated with some tests, and the few comments to be taken care :)

vvo:
need rebase (maybe because of urlsync merge?)

bobylito:
Erf done! :)

vvo:
LGTM

PR #782, opened 2016-01-13
docs(all): cssClasses is always {string|string[]}

vvo:


bobylito:
LGTM

pixelastic:
Are you sure about that one? I don't think we have any test that checks that an array is actually supported and I don't remember ever using that syntax myself. It _should_ work, but was never tested.

PR #783, opened 2016-01-15
fix(url-sync): Makes url sync more reliable

### LINKED ISSUE #729: search-box is not updated when input has focus and back button pressed
bug

This is not a new bug since 1.0.0, it was already here. I guess it depends on the browser/OS but for me clicking on back button does not focus out the input. Thus it is not updated.

I think we need to have a way to know when the query was updated from the outside rather than from the input itself.

Right now we use [helper.on('change')](https://github.com/algolia/instantsearch.js/blob/fix/update-search-box-on-helper-state/src/widgets/search-box/search-box.js#L163-L166) and then see if the input is focused.

But there are cases where the input is focused and the query might change still:
- custom widget modifying the query based on a timer for example (weird but ok)
- back button while the input is focused and stay focused
- back navigation shortcut while focused (fails)

I wonder how could we change the searchbox so that we really update it and distinguish when the event comes from the outside.

I feel this is something REACT solved for instance :D

### LINKED ISSUE #730: render calls order is bad when urlSync is used
I managed to break the render calls result order.

Using the standard dev app on develop, by typing "mp3" (fast), you can manage to have results for the query "mp".

This is linked to the url-sync feature since without it it works well.

What I did is add a "render" method to the search box widget and console.log inside it. Render calls and the query are bad given the current search box value.

It seems the url sync is re-triggering a search somewhere?

Any idea welcomed, this is linked with #729 because a naive approach with flags to solve #729 can't work if the render call order is not good.

bug2

bobylito:
2 main changes here :
 - the url-sync internal widget now relies on the helper directly and
   not on the instantsearch lifecycle anymore.
 - there is dedicated channel of update for the changes coming from the
   history (private API for the moment)

FIX #730
FIX #729

### DIFF
--- a/dev/app.js
+++ b/dev/app.js
@@ -5,9 +5,7 @@ var search = instantsearch({
   appId: 'latency',
   apiKey: '6be0576ff61c053d5f9a3225e2a90f76',
   indexName: 'instant_search',
-  urlSync: {
-    useHash: true
-  }
+  urlSync: true
 });
 
 search.addWidget(
@@ -281,3 +279,4 @@ search.addWidget(
 );
 
 search.start();
+
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -5,7 +5,7 @@ import forEach from 'lodash/collection/forEach';
 import merge from 'lodash/object/merge';
 import union from 'lodash/array/union';
 
-let EventEmitter = require('events').EventEmitter;
+import {EventEmitter} from 'events';
 
 import urlSyncWidget from './url-sync.js';
 import version from './version.js';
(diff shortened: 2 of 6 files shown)


dev/app.js
```diff
@@ -6,7 +6,6 @@ var search = instantsearch({
   apiKey: '6be0576ff61c053d5f9a3225e2a90f76',
   indexName: 'instant_search',
   urlSync: {
-    useHash: true
```
vvo: maybe we can explicitly send urlSync: true instead of an empty object?

src/lib/InstantSearch.js
```diff
@@ -136,10 +138,11 @@ Usage: instantsearch({
   }
 
   _init(state, helper) {
+    const onHistoryChange = this._onHistoryChange || function() {};
```
vvo: Why no test the default value in the upper else (like we do for _createURL), would look more consisten I feel.

vvo:
Other than the two comments I made: LGTM

dev/app.js
pixelastic: Yep, this would seem more readable to me

bobylito: Yep makes sense! Will make the modification to be sure that true is the value we want :)

bobylito:
Tests are passing and all the remarks are taken care of, merging!

PR #786, opened 2016-01-20
functional tests

vvo:
This is a preview of the functional tests stack. I opened this PR to **get some feedback** on how the tests are written and what is the stack that we are using.

The goal of the PR in is to be able to write functional tests that are ensuring our widgets are actually working inside a real browser on a real application.

This is different test level from our unit tests, we are not going to test every behavior of the widget. We only want to check that once everything is plugged together, the main feature of every widget is here.

## Selenium
To start real browsers and do clicks we have only one real solution: **selenium**.

Basically we will run a local selenium when hacking those tests and we will leverage saucelabs on CI to run on multiple browsers.

I use https://github.com/elgalu/docker-selenium to develop my tests: it's awesome. It has everything we need to run tests locally, inside a docker image.

## webdriver.io
I choosed http://webdriver.io to write functionnal tests. I have already participated in the project, it's really well done and maintained. It's the go-to.

## how

So webdriver.io test runner kicks off, we start a web server that will serve a complete search interface with all the widgets. Much like our npm run dev example but separated to avoid changes in the dev example triggering tests failures.

Then we use the webdriver.io API to launch commands.

Here's a sample test of the clearAll widget:

```js
import expect from 'expect';
import {clearAll, getCurrentRefinements} from '../utils.js';

describe('clearAll', () => {
  it('clears refinements', () => {
    const firstRefinement = '#hierarchical-categories .item:nth-child(2)';
    const secondRefinement = '#brands .item:nth-child(1)';

    return getCurrentRefinements()
      .then(refinements => expect(refinements.length).toBe(0))
      .click(firstRefinement)
      .click(secondRefinement)
      .then(getCurrentRefinements)
      .then(refinements => expect(refinements.length).toBe(2))
      .then(clearAll)
      .then(getCurrentRefinements)
      .then(refinements => expect(refinements.length).toBe(0));
  });
});
```

So here we just check that:
- the initial state is good, there's no refinement
- once we click on two refinements, we find some refinements
- after clicking on clearAll, we check that we have no more refinements

And that's it for a start on the clearAll widget.

As you can see I started developing some utils that will need to be tested also.

Tell me what you think and in the meantime I will continue adding tests.


functional-tests/app/index.html
```diff
@@ -0,0 +1,96 @@
...
+<head>
+  <meta charset="UTF-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=Edge">
+  <title>Instant search demo built with instantsearch.js</title>
+  <link rel="stylesheet" href="//cdn.jsdelivr.net/bootstrap/3.3.5/css/bootstrap.min.css">
+  <link rel="stylesheet" href="//cdn.jsdelivr.net/bootstrap/3.3.5/css/bootstrap-theme.min.css">
+  <!--[if lte IE 9]>
+  <script src="https://cdn.polyfill.io/v2/polyfill.min.js"></script>
+  <![endif]-->
+  <link rel="stylesheet" href="./instantsearch.min.css">
```
vvo: This is important, we will be testing using a real build that will be done in CI, we are not using a full webpack thing like the npm ru dev

bobylito:
I'm already loving it :) I'm going to have some fun with it this afternoon (CET :))

bobylito:
Why does the `getCurrentRefinements` returns something that looks like a promise? How come the promise have click? :)

How does `getCurrentRefinements` get the state?

vvo:
> Why does the getCurrentRefinements returns something that looks like a promise?

So all the utils are just returning the global "browser" variable created by webdriver.io, it represents the currently opened browser and you access all the API methods from http://webdriver.io/api.html

But it's also a sort of magic promise chain so that after each command you can use .then() to check the results.


> How does getCurrentRefinements get the state?

Here's the implementation:

```js
export function getCurrentRefinements() {
  return browser
    .element('#current-refined-values .item')
    .then(() => browser
      .getText('#current-refined-values .item')
      .then(res => {
        return res.map(refinement => {
          const data = refinement.split('\n');
          const out = {
            name: data[0]
          };

          if (data[1] !== undefined) {
            out.count = parseInt(data[1], 10);
          }

          return out;
        });
      })
    )
    .catch(() => Promise.resolve([]));
}
```

So it's pretty raw but does the job.

> I'm already loving it :) I'm going to have some fun with it this afternoon (CET :))

If you already want to add new tests: AWESOME. Here's what you can do:

- install the https://github.com/elgalu/docker-selenium thing
- run it like this:

```sh
docker run --net="host" --rm --name=grid -e VNC_PASSWORD=fun -e NOVNC=true elgalu/selenium:2.48.2k
```

- try to access in a browser in osx: http://localhost:6080/vnc.html and use password fun, you should see the underlying linux desktop

- once you are here, take this current branch, npm install, npm run test:functional

vvo:
This is ready for re-review.

Added:
- minimal functional test stack
- minimal set of tests
- ability to run tests locally (README)
- tests are ran on travis on chrome, firefox, IE11, safari (other browsers are too broken in selenium to be tested for now)

vvo:
The thing that is bothering me right now is that I do have some tests that are not stable and can fail 1/4 time. booooh

vvo:
Seems by adding the right pause(s) at the right place those are now stable, ready..to...merge :)

PR #789, opened 2016-01-21
feat(hits): adds allItems template as an alternative to item

### LINKED ISSUE #771: hits widget should support defining the whole hits template, not just per item
Following #707

We want the user to be able to iterate on all the hits through a mustache.js template. Not just per item.

Solution: create a new template entry for the hits widget called `hits` that will receive the all the hits and allow rendering the whole data

bobylito:
Adds a way to let the user specify how to do the complete rendering of the hits. FIX #771

### DIFF
--- a/dev/templates/all-items.html
+++ b/dev/templates/all-items.html
@@ -1,0 +1,10 @@
+<table class="table table-striped table-condensed">
+  <thead>
+    <tr><th>name</th><th>brand</th><th>price</th></tr>
+  </thead>
+  <tbody>
+  {{#hits}}
+    <tr><td><a href="#hit-{{objectID}}">{{name}}</a></td><td>{{brand}}</td><td>{{price}}</td></tr>
+  {{/hits}}
+  </tbody>
+</table>
--- a/dev/templates/item.html
+++ b/dev/templates/item.html
@@ -1,4 +1,4 @@
-<div class="hit">
+<div class="hit" id="hit-{{objectID}}">
   <div class="media">
     <a class="pull-left" href="{{url}}">
       <img class="media-object" src="{{image}}">
@@ -10,4 +10,5 @@
       {{#free_shipping}}<span class="badge pull-right">Free Shipping</span>{{/free_shipping}}
     </div>
   </div>
+  <a href="#">Go back to top</a>
 </div>
--- a/src/components/Hits.js
+++ b/src/components/Hits.js
@@ -20,6 +20,17 @@ class Hits extends React.Component {
     return <div className={this.props.cssClasses.root}>{renderedHits}</div>;
   }
 
+  renderAllResults() {
+    return (
+      <Template
+        cssClass={this.props.cssClasses.allItems}
+        data={this.props.results}
+        templateKey="allItems"
+        {...this.props.templateProps}
+      />
+    );
+  }
+
   renderNoResults() {
     let className = this.props.cssClasses.root + ' ' + this.props.cssClasses.empty;
     return (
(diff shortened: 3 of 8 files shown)


src/widgets/hits/hits.js
```diff
@@ -41,7 +44,11 @@ function hits({
     hitsPerPage = 20
   } = {}) {
   if (!container) {
-    throw new Error(usage);
+    throw new Error('Must provide a container.' + usage);
+  }
+
+  if (templates.item && templates.items) {
+    throw new Error('Must contain only items OR item template.' + usage);
```
vvo: Its allItems I believe but that made me think, what's the best API entry to express this new option? allItems? items? full?

dev/templates/item.html
```diff
@@ -1,4 +1,4 @@
-<div class="hit">
+<div class="hit" id="hit-{{objectID}}">
```
vvo: is this mandatory or only here for testing?

dev/templates/all-items.html
```diff
@@ -0,0 +1,10 @@
+<table class="table table-striped table-condensed">
+  <thead>
+    <tr><th>name</th><th>brand</th><th>price</th></tr>
+  </thead>
+  <tbody>
+  {{#hits}}
+    <tr><td><a href="#hit-{{objectID}}">{{name}}</a></td><td>{{brand}}</td><td>{{price}}</td></tr>
```
Jerskouille: Why not use the `item` template here ?

src/widgets/hits/hits.js
bobylito: Good question. I first tried items, but it was too visually similar and I though it would be prone to errors. I didn't tried full, but it's an adjective.

dev/templates/item.html
bobylito: Testing purpose, I use a second hits widget in our kitchen sink to make a summary of the results, and I wanted to make links :)

src/components/Hits.js
```diff
@@ -20,6 +20,17 @@ class Hits extends React.Component {
     return <div className={this.props.cssClasses.root}>{renderedHits}</div>;
   }
 
+  renderAllResults() {
+    return (
+      <Template
+        cssClass={this.props.cssClasses.item}
```
Jerskouille: Shouldn't it use a new `cssClass`? `body` for instance?

dev/templates/all-items.html
bobylito: Can we already do that?

Jerskouille: Well, it is doable with a small `Hogan` helper I believe. I can help you with that if you want.

Jerskouille:
I like `items`, but maybe that's because Rails uses a lot singular/plural forms.

dev/templates/all-items.html
bobylito: Well I'd rather not add this ability then.

src/components/Hits.js
bobylito: True!

bobylito:
@Jerskouille I tried `items` but given that it's the same level of usage (both parameters of templates and both are valid attributes) it's likely to generate mistakes which are hard to find and debug.

dev/templates/all-items.html
Jerskouille: That doesn't seem like the right reason to refuse this change.

src/widgets/hits/hits.js
vvo: I am good with allItems thanks for the explanation. We just need to fix the small places where you said "items" and we are good (usage, throw)

pixelastic:
> @Jerskouille I tried items but given that it's the same level of usage (both parameters of templates and both are valid attributes) it's likely to generate mistakes which are hard to find and debug.

I don't get why `allItems` would be better than `items` then?

bobylito:
It's more possible to not see a missing `s` than `all...s`

pixelastic:
Oh ok. I, also, am used to the Rails conventions so quite used to spot that :)

bobylito:
I am also used to spot that. But I got bitten the other day, and I don't like that :P

vvo:
LGTM

PR #792, opened 2016-01-25
docs(tourism): fixes handles in tourism sample (fix #642)

### LINKED ISSUE #642: tourism example: Unable to get results if price range is narrowed before query
I first do a narrow search on the price range, then do a search query. After that the map is centered somewhere (?) and I cannot move the price range anymore.

2015-11-17_16h00m42

Not sure if this is an example issue or a price range one.

https://community.algolia.com/instantsearch.js/examples/tourism/

bobylito:
Could be backported to default stylesheet, no?

### DIFF
--- a/docs/examples/tourism/_style.scss
+++ b/docs/examples/tourism/_style.scss
@@ -177,6 +177,14 @@ body {
       border-color: $aisdemo-primary-color;
     }
 
+    .ais-range-slider--handle-lower {
+      transform: translateX(-50%);
+    }
+
+    .ais-range-slider--handle-upper {
+      transform: translateX(50%);
+    }
+
     .ais-range-slider--tooltip {
       &:last-child {
         right: 0;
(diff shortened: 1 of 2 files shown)


pixelastic:
I think the `transform` properties will use the GPU (which can consume more battery on mobile). Isn't there a way to do it with a more classical CSS property (`top`, `margin-top`, etc).

I haven't looked at the code, just at the PR so feel free to ignore my comment.

vvo:
How does the fix works? Does it changes the UI?

I thought there was a JS bug here.

bobylito:
No there is no other bug than the graphic bug. The fix moves the handles so they don't overlap.

> I think the transform properties will use the GPU (which can consume more battery on mobile). Isn't there a way to do it with a more classical CSS property (top, margin-top, etc).

This is the best way to achieve what we're doing. 2D transform operations are not done on the GPU (it seems hard to be final here though as it is dependent on the browser implementation). The main problem when the GPU is triggered with 3D transform operations is memory consumption. The browser creates a layer dedicated to the transformed element which can create crazy memory issues.

PR #797, opened 2016-01-26
chore(dev app): fix IE9 working for dev app

vvo:


bobylito:
Nice fix. Even though I'm not sure we care in the kitchen sink.

vvo:
This fix is here so that we can ensure we are able to test if IE9 is working with instantsearch

And theres no polyfill for the history API, only weird workarounds or libs that will try to use the hash, that we already know how to do

dev/app.js
```diff
@@ -5,7 +5,7 @@ var search = instantsearch({
   appId: 'latency',
   apiKey: '6be0576ff61c053d5f9a3225e2a90f76',
   indexName: 'instant_search',
-  urlSync: true
+  urlSync: window.hasHistoryAPI === false ? {useHash: true} : true
```
hkdobrev: Why not using a normal feature detection which does not rely on conditional comments for IE9 only.
E.g. Modernizr does it like that: https://github.com/Modernizr/Modernizr/blob/fc6e5f47a0df837d568007f165b5331acdf1e26a/feature-detects/history.js#L40
`(window.history && 'pushState' in window.history)`

vvo:
Why not yes it would also work

bobylito:
Thanks for the explanation @vvo! :)
LGTM!

PR #798, opened 2016-01-27
test(unit): add chrome unit test run

vvo:


src/widgets/current-refined-values/__tests__/current-refined-values-test.js
```diff
@@ -44,9 +44,9 @@ describe('currentRefinedValues()', () => {
     context('options.container', () => {
       it('doesn\'t throw usage with a string', () => {
         let element = document.createElement('div');
-        element.id = 'testid';
+        element.id = 'testid2';
         document.body.appendChild(element);
-        parameters.container = '#testid';
+        parameters.container = '#testid2';
```
vvo: tests, when ran in a real browser, were stepping on each others. we did not see it with jsdom since jsdom is re-creating a new dom on every it('').

bobylito:
LGTM!

PR #800, opened 2016-01-28
fix(currentRefinedValues): allow array of strings for cssClasses.*

vvo:


Jerskouille:
I've forced push a new test instead of the hack of the default values.
Default values are only here so that each test can just check what it's testing and not have to modify anything else.

vvo:
thx

PR #801, opened 2016-01-28
test(cssClasses): test at least once in every widget that we support array of strings

vvo:


bobylito:
Looks great! Travis is green so good to go!

PR #822, opened 2016-02-02
Fix/searchbox watch form reset event

Shipow:


src/widgets/search-box/search-box.js
```diff
@@ -124,7 +124,24 @@ function searchBox({
       // Add all the needed attributes and listeners to the input
       this.addDefaultAttributesToInput(input, state.query);
 
-      // Keep keyup to handle searchOnEnterKeyPressOnly
+      // like jquery closest()
+      function closest(el, selector) {
+        let matches = el.webkitMatchesSelector ? 'webkitMatchesSelector' : (el.msMatchesSelector ? 'msMatchesSelector' : 'matches');
+        while (el.parentElement) {
```
vvo: why not just smg like while(parent !== FORM) parent=parentNode?

dunno msMatchesSelector will have to look

vvo:
Is it feasible to handle such case by listening on events directly on the input? Don't pushing the reset button triggers a 'change' event on the input?

Shipow:
"change" event is triggered for many reasons, I didn't find a way to to link it to the reset event.

vvo:
> "change" event is triggered for many reasons, I didn't find a way to to link it to the reset event.
> No specific event are triggered on the input when reseting the form.

What about checking "change" event but also that the input value is now empty? That would normally mean that it was a reset event?

Shipow:
Could work but you have to handle the default value elsewhere.
Also there is a pattern of event that you could detect: blur -> change

vvo:
I will have another take at this by allowing having a wrapping `<form>` element which is the natural way to handle cases like a submit or reset button. works for you?

Shipow:
can we work on that from this branch ?

vvo:
Will be easier in another branch, see https://github.com/algolia/instantsearch.js/issues/859 for the underlying issue

PR #824, opened 2016-02-02
feat(menu): add showMore option

vvo:


vvo:
cc @bobylito what do you think (also cc @olance)

Shipow:
\o/

src/widgets/hierarchical-menu/hierarchical-menu.js
```diff
@@ -46,7 +46,7 @@ hierarchicalMenu({
   [ separator=' > ' ],
   [ rootPath ],
   [ showParentLevel=true ],
-  [ limit=1000 ],
+  [ limit=30 ],
```
bobylito: 30 is a lot. I think 10 should be enough especially since we have the show more config.

vvo: I thought 30 was a good "this is a default value that will show you a lot but you should configure it"

bobylito: It goes against the idea of having sensible defaults, which you don't have to configure. I think it's better if you have less to learn, and if you could the best results with the least effort.

vvo: > It goes against the idea of having sensible defaults, which you don't have to configure

I agree on having sensible defaults but we do not know if either 10 or 30 are sensible defaults given the user POV: it might be too low (some items are missing) or too high (UX issue).

bobylito:
LGTM

Jerskouille:
> 30 is a lot. I think 10 should be enough especially since we have the show more config.

A "Show More" button on a hierarchical is a bit weird, isn't it?

vvo:
> A "Show More" button on a hierarchical is a bit weird, isn't it?

Yes, we did not implement it on hierarchicalMenu

slimtimbodin:
I have a great need for the show more button on the hierarchicalMenu.   I have over 50 first level categories  and each one of these has 3 to 5 sub categories.  Ideally I would like to have show more button similar to the hierarchicalMenu in the  React InstantSearch.

PR #828, opened 2016-02-08
docs(hits): Update documentation for the `allItems` property

pixelastic:
Seems like the `allItems` property for both template and transformData
was a copy-paste of the `item` one.


pixelastic:
Oh wait, actually I'm not sure of what I wrote. Will the `transformData.allItems` only be used for the `templates.allItems` template, or will it also work for `templates.item` individually?

pixelastic:
Just tested it. What I wrote is correct, the `item` and `allItems` templates are mutually exclusive, and the `item` and `allItems` `transformData` only apply to their respective templates.

So this can be merged.

vvo:
thx :)

Issue #45, opened 2015-08-06
Template as React component?


bobylito:
What if we could directly pass React component as templates? It could make sense for people doing React already, no?

vvo:
Yes I do think this would make sense if you are in a react environement.

But I am also wondering, when in a react environment, would you use our .widgets.* wrappers?

As a react user, maybe you would prefer to access our react components and then do whatever you want. But they would then need to do some good coding.

As I did not build any full react website, I am really not sure.

pixelastic:
I have not enough experience with React to see if this would be valuable or not

redox:
We did that while upgrading to React 0.14 (d880ea258363a536d6ab334353675e08acd377b3).

bobylito:
@redox humm that's not  what this issue is about. It's about being able to use a react component as it was a template. Conceptually they are the same so it would make sense to be able to do so.

vvo:
Do we investigate more this? How would passing a react template work in details?

bobylito:
```js
class MyTemplateComponent extends React.Component {};
```

pixelastic:
I really like this idea

vvo:
:horse: CLOSING BOT SAYS:
> This is very old and not done, reopen if you have the need and want to fix it

Willianvdv:
@bobylito did you get this working? I'd love to Reactify our widget templates.

bobylito:
@Willianvdv well I would love to have this working but I can't figure out a simple way to detect if a value is a really a react component : 
 - if it's a function it collides with the custom function that we can already use ( (data) => (string) vs (data) => (react stuff) )
 - if it's an React Class, I don't know how to figure out if it's a constructor of a React Component or something else
 - if it's an ES6 class, we might be more lucky because it will extend React.Component and we can use instanceof

What do you think? Am I missing something?

vvo:
We could have a isJSX: true somewhere but even that I wonder what would be feasible and if we can even make it work

VonD:
another option would be a function returning a React element instead of a string : 

```es6
instantsearch.widgets.hits({
    templates : { hit: hit => <MyHitComponent hit={ hit } />  }
})
```
We can then use `React.isValidElement` to check the template function return value : if its a string, use `dangerouslySetInnerHTML`, if it's a valid element, just render it.

What do you think?

bobylito:
[...]

PR #831, opened 2016-02-09
feat(Template): allow template functions to return a React element

VonD:
about [#45](https://github.com/algolia/instantsearch.js/issues/45)

### DIFF
--- a/src/components/Template.js
+++ b/src/components/Template.js
@@ -27,16 +27,22 @@ function Template(props) {
   }
 
   const otherProps = omit(props, keys(Template.propTypes));
-  const isElement = React.isValidElement(content);
+
+  if (React.isValidElement(content)) {
+    return (
+      <div
+        {...otherProps}
+        className={props.cssClass}
+      >{content}</div>
+    );
+  }
 
   return (
     <div
       {...otherProps}
       className={props.cssClass}
-      dangerouslySetInnerHTML={isElement ? null : {__html: content}}
-    >
-      {isElement ? content : null}
-    </div>
+      dangerouslySetInnerHTML={{__html: content}}
+    />
   );
 }
 
--- a/src/components/Template.js
+++ b/src/components/Template.js
@@ -27,13 +27,16 @@ function Template(props) {
   }
 
   const otherProps = omit(props, keys(Template.propTypes));
+  const isElement = React.isValidElement(content);
 
   return (
     <div
       {...otherProps}
       className={props.cssClass}
-      dangerouslySetInnerHTML={{__html: content}}
-    />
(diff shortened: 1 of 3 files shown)


src/components/Template.js
```diff
@@ -27,13 +27,16 @@ function Template(props) {
...
   return (
     <div
       {...otherProps}
       className={props.cssClass}
-      dangerouslySetInnerHTML={{__html: content}}
-    />
+      dangerouslySetInnerHTML={isElement ? null : {__html: content}}
+    >
+      {isElement ? content : null}
+    </div>
```
bobylito: Maybe we should not systematically use dangerouslySetInnerHTML. What about:

```js
if(isEelement) {
  return <div...>{content}</div>
}
return <div...  dangerouslySetInnerHTML={{__html: content}}/>
```

VonD: @bobylito ok, makes sense, i've pushed the change.

vvo:
Somehow the browsers cannot start, this is a test build issue. The PR is nice

vvo:
Thanks @VonD, a lot. Awesome contribution, where are you using/planning to use instantsearch.js? Any feedback on how in integrates with an existing react applications is very nice to have, open issues!

VonD:
@vvo thanks! Not much feedback for the moment as the project is still at a very early stage and as i'm just discovering the lib.

I was wondering, though, if there might be a more straightforward integration within a React app. Excuse-me if it's already the case. I was thinking about an api like, for example : 

```es6
import React, { Component } from 'react' 
import { InstantSearch, SearchBox, Hits } from 'instantsearch.js'

class MySearchComponent extends Component {
  render() {
    return (
      <InstantSearch
        appId={ appId }
        apiKey={ apiKey }
        indexName={ indexName }
      >
        <SearchBox
          placeholder={ 'Custom placeholder' }
        />
        <Hits
          hitsPerPage={ 6 }
          renderHit={ hit => <MyItemComponent item={ hit } /> }
        />
      </InstantSearch>
    );
  }
}
```

Some of the benefits i see in such an approach would be : 
* the whole tree gets re-rendered when the parent component's `props` or `state` are updated
* the `context` does not get lost : useful with libs relying on it such as `react-router`
* we get server-side rendering for free
* no risk of conflicts between different React versions
* maybe modularize dependencies to avoid bundling `hogan` or other libs that react apps won't use

Would that be a big refactoring ? I understand the lib is not mainly targeting React apps, but as it is all React, it doesn't seem too difficult. The reference to the search helper could be passed to the widgets via `context`

Sorry if all that's already possible or off-topic.

vvo:
@VonD can you open an new issue about this subject (just copy paste it)?

VonD:
@vvo done

PR #834, opened 2016-02-10
Perf improvements: shouldComponentUpdate, less bind, Number.prototype.toLocaleString

vvo:
This is a pretty big PR splitted in 3 main improvements:
- use babel specific [react optimizations](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#compiler-optimizations) 95f940c078fba362592245c4c23422f51d317c63
- stop forcing the locale to Number.prototype.formatNumber (see https://jsperf.com/number-prototype-tolocalestring) b05655444d69786a534cb6a0ab714cd65e1119ef
- use shouldComponent update and compare props before asking to re-render b91603fe0d2b59e0911809656baabd99879ff76c

To measure the results, I used the [instant search demo](https://github.com/algolia/examples/tree/master/instant-search/instantsearch.js).

Warming up the interface with some queries, I then did 10 (manual) JavaScript profile measurements. I would:
- hit "start profiling"
- put "c" letter in searchbox
- hit "stop profiling"

Doing so 10 times, taking the fastest render time.

In the JS flame chart, the main instantsearch.js render() function highlighted. Then you have the profiling table with the slowest calls.

**BEFORE**
2016-02-10-181541_1946x616_scrot
2016-02-10-181726_985x773_scrot

Render is ~33ms, createElement and formatNumber are the two most called functions.

**AFTER**

2016-02-10-181559_2125x633_scrot
2016-02-10-181701_975x608_scrot

Render is ~23ms

So that's 10ms better, or 30% faster.

We still have room for improvements, it seems that now a good time is spent in the helper (parseNumbers) and in the slider (adding pips).

### DIFF
--- a/src/components/CurrentRefinedValues/CurrentRefinedValues.js
+++ b/src/components/CurrentRefinedValues/CurrentRefinedValues.js
@@ -5,8 +5,13 @@ import Template from '../Template.js';
 import {isSpecialClick} from '../../lib/utils.js';
 import map from 'lodash/collection/map';
 import cloneDeep from 'lodash/lang/cloneDeep';
+import {isEqual} from 'lodash';
 
 class CurrentRefinedValues extends React.Component {
+  shouldComponentUpdate(nextProps) {
+    return !isEqual(this.props.refinements, nextProps.refinements);
+  }
+
   _clearAllElement(position, requestedPosition) {
     if (requestedPosition !== position) {
       return undefined;
--- a/src/components/Pagination/PaginationLink.js
+++ b/src/components/Pagination/PaginationLink.js
@@ -1,8 +1,22 @@
 import React from 'react';
 
+import {isEqual} from 'lodash';
+
 class PaginationLink extends React.Component {
+  componentWillMount() {
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  shouldComponentUpdate(nextProps) {
+    return !isEqual(this.props, nextProps);
+  }
+
+  handleClick(e) {
+    this.props.handleClick(this.props.pageNumber, e);
+  }
+
   render() {
-    let {cssClasses, label, ariaLabel, handleClick, url} = this.props;
+    let {cssClasses, label, ariaLabel, url} = this.props;
 
     return (
       <li className={cssClasses.item}>
@@ -11,7 +25,7 @@ class PaginationLink extends React.Component {
           className={cssClasses.link}
           dangerouslySetInnerHTML={{__html: label}}
           href={url}
-          onClick={handleClick}
+          onClick={this.handleClick}
         ></a>
       </li>
     );
@@ -32,6 +46,7 @@ PaginationLink.propTypes = {
     React.PropTypes.string,
     React.PropTypes.number
   ]).isRequired,
+  pageNumber: React.PropTypes.number,
   url: React.PropTypes.string
 };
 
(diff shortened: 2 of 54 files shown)


Jerskouille:
I can't believe that < 100 `formatNumber` calls would take 3ms, just wow. Great profiling :+1:

src/components/CurrentRefinedValues/CurrentRefinedValues.js
```diff
@@ -5,8 +5,13 @@ import Template from '../Template.js';
 import {isSpecialClick} from '../../lib/utils.js';
 import map from 'lodash/collection/map';
 import cloneDeep from 'lodash/lang/cloneDeep';
+import {isEqual} from 'lodash';
 
 class CurrentRefinedValues extends React.Component {
+  shouldComponentUpdate(nextProps) {
+    return !isEqual(this.props.refinements, nextProps.refinements);
```
redox: Shouldn't we ensure we only manipulate immutable props so we could replace the deep `isEqual` by a regular `===`?

src/components/Pagination/PaginationLink.js
```diff
@@ -1,8 +1,22 @@
...
 
+import {isEqual} from 'lodash';
+
 class PaginationLink extends React.Component {
+  componentWillMount() {
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  shouldComponentUpdate(nextProps) {
+    return !isEqual(this.props, nextProps);
```
redox: What about comparing only `props.pageNumber` with a simple `!==`?

src/components/RefinementList/RefinementList.js
```diff
@@ -10,6 +12,11 @@ class RefinementList extends React.Component {
     this.state = {
       isShowMoreOpen: false
     };
+    this.handleItemClick = this.handleItemClick.bind(this);
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return !isEqual(nextState, this.state) || !isEqual(this.props.facetValues, nextProps.facetValues);
```
redox: Here the `state`/`nextState` is immutable by definition, right?

redox:
I would also unfold a few minified functions (`r`) to see what code is behind, a bunch of them are still consuming 2-5% of the time.

src/components/Pagination/PaginationLink.js
vvo: This is where shouldComponentUpdate is tricky, multiple cases to take into account:
- url changes? (new search = different url)
- active page (click = item class is "active")

So you would end up with a pretty long if .. condition while just comparing all the props deeply is ok (small objects).

But yes, my first iteration was to just do this.props.pageNumber !== nextProps.pageNumber, then it failed

src/components/RefinementList/RefinementList.js
vvo: Yes indeed so we can use a simple === compare

vvo:
> I would also unfold a few minified functions (r or just run the code without minification ) to see what code is behind, a bunch of them are still consuming 2-5% of the time.

We can do so by running the dev example, here it was using a production build to precisely measure the results. But if we want to go further, we can, come at my performance office when you want :D

vvo:
> I can't believe that < 100 formatNumber calls would take 3ms, just wow. Great profiling

Yes, I was like "COME.ON!", thanks :)

src/components/CurrentRefinedValues/CurrentRefinedValues.js
vvo: That would be awesome to only have to do so, currently not feasible without refactoring a lot I guess. I need to better understand how we could do so.

vvo:
So we stll have room for improvements but we can iterate on it later on.

PR #836, opened 2016-02-11
perf(nouislider): upgrade nouislider, shaves some more ms

vvo:


src/widgets/range-slider/range-slider.js
```diff
@@ -127,6 +127,10 @@ function rangeSlider({
         };
       }
 
+      if (tooltips.format !== undefined) {
+        tooltips = [{to: tooltips.format}, {to: tooltips.format}];
```
vvo: I rebased our nouislider fork to use the new one, which change a lil his tooltip format option

redox:
:+1:

PR #847, opened 2016-02-15
fix(currentRefinedValues): clear numeric refinements using original value

vvo:


src/widgets/current-refined-values/current-refined-values.js
```diff
@@ -233,7 +233,7 @@ function clearRefinementFromState(state, refinement) {
   case 'exclude':
     return state.removeExcludeRefinement(refinement.attributeName, refinement.name);
   case 'numeric':
-    return state.removeNumericRefinement(refinement.attributeName, refinement.operator, refinement.name);
```
Jerskouille: What about a +`refinement.name` here?

vvo: I do prefer fixing the issue at the root level than parsing a string as an int that was previously an int.

Jerskouille:
I think I know what bugs me about this PR. While I do agree with you we should fix it at the source, I think this should be at the helper level.
Indeed, in the helper:
- `addNumericRefinement(attribute, operator, value)` type-checks value and convert it to a float if it's a string
- but `removeNumericRefinement(attribute, operator, value)` just does a simple `===` comparison

Which means that:
```js
var val = /* ... */;
helper.addNumericRefinement('attribute', '=', val);
helper.removeNumericRefinement('attribute', '=', val);
```
is actually not guaranteed to work, which is, to me, weird.

By fixing this, we would actually fix the need for this PR.

vvo:
@Jerskouille That's another discussion but I do think we should not try to convert a string in the helper if we expect a numeric. It's magical but also error prone (re, this issue).

Just throw and that's it (breaking change in the helper).

cc @bobylito what do you think?

vvo:
That being said, the current issue underlined by @Jerskouille should be fixed in the helper because that's no consistent with the behavior we put.

src/widgets/current-refined-values/current-refined-values.js
bobylito: Why is it transformed into a string in the first place?

bobylito:
I think that what is expected to be numeric should be parsed. The reason is that we have a serialized form that may cause the number to be transformed to string (url stuff) but for the sake of convenience we should provide numbers when we can (who wants to parse a value coming from a library?)

I'm fixing the bug you found @Jerskouille. I couldn't find the ticket on the helper, if there is one, can you reference it here?

Jerskouille:
@bobylito I didn't file one, I was waiting to see the end of this discussion. Now submitted here: https://github.com/algolia/algoliasearch-helper-js/issues/285 .

bobylito:
https://github.com/algolia/algoliasearch-helper-js/pull/284

src/widgets/current-refined-values/current-refined-values.js
Jerskouille: To be consistent, I think `name` should be a string value.I would have used a numeric value if the attribute name was `value`.

bobylito: Well it's badly named for sure. I don't remember why though. Would it be possible to change the name of the property to `value` and leave the value as a number?

vvo: > Well it's badly named for sure. I don't remember why though. Would it be possible to change the name of the property to value and leave the value as a number?

The thing is that in every template we pass to * refinement list (menus, hierarchical ..), we always use "name" that's why it was set as name, for consistency and ease of use.

vvo: > Would it be possible to change the name of the property to value

Would be a breaking change when passed to the template data

vvo:
This is currently an issue affecting users so gonna merge it, release and then once https://github.com/algolia/algoliasearch-helper-js/pull/284 is merged and helper is released I will revert this commit.

Issue #835, opened 2016-02-10
Collapsable facets


vvo:
In some UI with a lot of facets, it seems people would want to be able to collapse/expand them on click.

We should provide this functionnality built-in, it would allow us to avoid rendering while the widget is collapsed (++ perf).

vvo:
Was done, to be released as

```js
collapsible: { // it it collapsible
  collapsed: true // is it collapsed at first
}
```

For all widgets having a header option

PR #854, opened 2016-02-17
perf(autoHideContainer): stop re-creating React components when going from hide to show

vvo:
The previous strategy for autoHideContainer was to either return an
empty div or the wrapped element.

So when the widget cycle was like this: init => hide (no results) =>
show (results) then we would re-create instances of React components because of a
different render() output in autoHideContainer.

This I discovered it while going to do #835: when the widgets would
hide => show we would lose the wrapped React component state resulting
in lost collaspable state.

Not only this commit will permit doing #835 easily but it also somehow
enhance performance since I also added a shouldComponentUpdate
strategy.

Added more precise tests on this decorator/HOC.

The slider needed an update because we are now trying to render it (once as soon as we are transitioning from hidden) with ranges being the same, we could also use shouldComponentUpdate inside the Slider that would work but not as specific than here.

### DIFF
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -22,6 +22,12 @@ class Slider extends React.Component {
   }
 
   render() {
+    if (this.props.range.min === this.props.range.max) {
+      // There's no need to try to render the Slider, it will not be usable
+      // and will throw
+      return null;
+    }
+
     // setup pips
     let pips;
     if (this.props.pips === false) {
--- a/functional-tests/app/app.js
+++ b/functional-tests/app/app.js
@@ -82,6 +82,7 @@ search.addWidget(
 
 search.addWidget(
   instantsearch.widgets.currentRefinedValues({
+    autoHideContainer: false,
     container: '#current-refined-values',
     cssClasses: {
       header: 'facet-title',
--- a/src/components/Slider/__tests__/Slider-test.js
+++ b/src/components/Slider/__tests__/Slider-test.js
@@ -15,6 +15,7 @@ describe('Slider', () => {
   let renderer;
   let Slider;
   let Nouislider;
+  let props;
 
   beforeEach(() => {
     let {createRenderer} = TestUtils;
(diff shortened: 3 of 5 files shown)


src/components/Slider/Slider.js
```diff
@@ -22,6 +22,11 @@ class Slider extends React.Component {
   }
 
   render() {
+    // There's no need to try to render the Slider, it will not be usable
+    if (this.props.range.min === this.props.range.max) {
```
Jerskouille: Shouldn't this be in AutoHideContainer?

bobylito: It's specific to this component and we want to keep its state.

vvo:
I added the full reasoning in the PR description @Jerskouille if you wanna have more infos on why we need this, or why we do it this way

Jerskouille:
Thank you @vvo .

But I have trouble understanding one thing:

> So when the widget cycle was like this: init => hide (no results) =>
show (results) then we would re-create instances of React components because of a
different render() output in autoHideContainer.

I still don't understand why we need to modify the component file.

When it goes from 'shown -> hidden', then `nextProps.shouldAutoHideContainer === true`, then `shouldUpdateComponent` is false, then `Slider.render` shouldn't be called.

vvo:
> nextProps.shouldAutoHideContainer === true, then shouldUpdateComponent is false, then Slider.render shouldn't be called.

Yes indeed there's something not right here, the shouldAutoHideContainer from the range-slider call should already be false earlier maybe. Will have to double-check. Thx

src/components/Slider/Slider.js
vvo: So this can still happen in cases where you are not using autoHideContainer for the slider

Jerskouille: OK, it makes sense.

vvo: Yes, a bit weird, I am thinking on providing templates.noRefinements, smg like that :)

Jerskouille: Yes, the small issue I have here is that basically we could extend this to all the other components, always with the `shouldAutoHideContainer` condition. What about renaming it, since it's internal, to a `containerEmpty` condition, create an `EmptyContainer` decorator that would return `null` or `<div/>` as `AutoHide` but only for the content, and make both of them use the same condition?

Jerskouille: Forget about it, if it goes to empty -> with content, we'll end up recreating some DOM nodes.

PR #855, opened 2016-02-17
feat(collapsable widgets): add collapsable and collapsed option

vvo:
Demo:

demo

All `header` capable widgets are now collapsable aware.

New options:
- collapsable: Hide the widget body and footer when clicking on header
- collapsed: Initialize the widget in collapsed state

This also adds:
- ais-root generic css class
- ais-body generic css class

To be in par with already existing ais-header, ais-footer.

This also adds:
- ais-root__collapsable css class
- ais-root__collapsed css class

### DIFF
--- a/dev/app.js
+++ b/dev/app.js
@@ -122,6 +122,9 @@ search.addWidget(
 
 search.addWidget(
   instantsearch.widgets.refinementList({
+    collapsible: {
+      collapsed: true
+    },
     container: '#brands',
     attributeName: 'brand',
     operator: 'or',
@@ -133,7 +136,7 @@ search.addWidget(
       active: 'facet-active'
     },
     templates: {
-      header: 'Brands'
+      header: 'Brands with collapsible <span class="collapse-arrow"></span>'
     },
     showMore: {
       templates: {
--- a/dev/style.css
+++ b/dev/style.css
@@ -122,3 +122,21 @@ body {
 .ais-show-more__active, .ais-show-more__inactive {
   cursor: pointer;
 }
+
+.ais-header {
+  position: relative;
+}
+
+.collapse-arrow {
+  position: absolute;
+  right: 0;
+}
+
+.collapse-arrow:after {
+  content: "[-]";
+  font-family: monospace;
+}
+
+.ais-root__collapsed .collapse-arrow:after {
+  content: "[+]";
+}
--- a/src/decorators/headerFooter.js
+++ b/src/decorators/headerFooter.js
@@ -9,16 +9,30 @@ import Template from '../components/Template.js';
 
 function headerFooter(ComposedComponent) {
   class HeaderFooter extends React.Component {
-    componentWillMount() {
-      // Only add header/footer if a template is defined
-      this._header = this.getTemplate('header');
-      this._footer = this.getTemplate('footer');
-      this._classNames = {
-        root: cx(this.props.cssClasses.root),
-        body: cx(this.props.cssClasses.body)
+    constructor(props) {
+      super(props);
+      this.handleHeaderClick = this.handleHeaderClick.bind(this);
+      this.state = {
+        collapsed: props.collapsible && props.collapsible.collapsed
       };
+
+      this._headerElement = this._getElement({
+        type: 'header',
+        handleClick: props.collapsible ? this.handleHeaderClick : null
+      });
+
+      this._cssClasses = {
+        root: cx('ais-root', this.props.cssClasses.root),
+        body: cx('ais-body', this.props.cssClasses.body)
+      };
+
+      this._footerElement = this._getElement({type: 'footer'});
     }
-    getTemplate(type) {
+    shouldComponentUpdate(nextProps, nextState) {
+      return nextState.collapsed === false ||
+        nextState !== this.state;
+    }
+    _getElement({type, handleClick = null}) {
       let templates = this.props.templateProps.templates;
       if (!templates || !templates[type]) {
         return null;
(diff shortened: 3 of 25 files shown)


vvo:
This is ready for review

vvo:
@algolia/javascript Anyone for review?

dev/app.js
```diff
@@ -122,6 +122,8 @@ search.addWidget(
 
 search.addWidget(
   instantsearch.widgets.refinementList({
+    collapsable: true,
+    collapsed: true,
```
bobylito: Is it only for the first rendering?

dev/style.css
```diff
@@ -122,3 +122,20 @@ body {
...
+  position: relative;
+}
+
+.collapse-arrow {
+  position: absolute;
+  right: 0;
+}
+
+.collapse-arrow:after {
+  content: "⬆";
```
bobylito: Aren't + and - more universally recognized?

vvo: Oh yeah, crazy dev came up with crazy solution here :D.

This is only the style.css from the dev app.

There's currently no default styling nor default template modification when you use "collapsable" option.

I am wondering if we should right away set +/- in the default template, seems to make sense. What do you think?

dev/app.js
vvo: Yes collapsed: true is only to give the initial state of the widget

bobylito: What about we change the name of the option to `initiallyCollapsed` or something that translate the fact that it is for the initial state of the widget?

dev/style.css
bobylito: Yes it makes sense to have a default style for that :+1:

src/decorators/headerFooter.js
```diff
@@ -9,16 +9,30 @@ import Template from '../components/Template.js';
...
+        collapsed: props.collapsed
       };
+
+      this._headerElement = this._getElement({
+        type: 'header',
+        handleClick: props.collapsable === true ? this.handleHeaderClick : null
+      });
+
+      this._cssClasses = {
+        root: cx('ais-root', this.props.cssClasses.root),
```
bobylito: What are `ais-root` and `ais-body`?

src/decorators/headerFooter.js
```diff
@@ -9,16 +9,30 @@ import Template from '../components/Template.js';
...
+      };
+
+      this._footerElement = this._getElement({type: 'footer'});
     }
-    getTemplate(type) {
+    shouldComponentUpdate(nextProps, nextState) {
+      return nextState.collapsed === false ||
+        nextState !== this.state;
+    }
+    _getElement({type, handleClick = null}) {
```
bobylito: getComponent?

bobylito:
On Google, you get more results with `collapsible` than with `collapsable` (and my spellchecking reports a typo :))

dev/app.js
vvo: Ultimately I would have wanted to have only a single configuration entry for this option, like we have for showMore: {limit..}.

But could not think of any good one.

Do you have any idea?

dev/style.css
vvo: After thinking about it, there's no default "header" template, you always submit yours. In comparison with the item template where we do have a default template.

So I will just pass collapsable and collapsed data to the header template when you use collapsable option.

Along with the "collapsed" CSS class that already exists.

src/decorators/headerFooter.js
vvo: Those were added to be consistent with already existing "ais-footer" "ais-header".

vvo:
> On Google, you get more results with collapsible than with collapsable (and my spellchecking reports a typo :))

Indeed! We may need to find better semantics here because it's not very clean/easy to say "collapsible"

src/decorators/headerFooter.js
vvo: > What about _getComponent since we get a react component?

Oh yeah, this fine grained difference is insane.. https://facebook.github.io/react/blog/2015/12/18/react-components-elements-and-instances.html

dev/app.js
bobylito: Oh that makes sense. I don't really like what we did for the URL sync but that could be consistent with what you want: 
 - type for collapsable `boolean|object`
 - having an object (even an empty one) is equivalent to having `true`
 - the object can contain an attribute `defaultState:boolean`

src/decorators/headerFooter.js
bobylito: Feels like root is too technical and abstract. What about `component`? Just a feeling  and not very important ;)

src/decorators/headerFooter.js
bobylito: Ok it makes perfect sense. When it's an instance it's an element. WFM

src/decorators/headerFooter.js
vvo: ah yes but at the same time we already have cssClasses.root (this is not a new semantic)

hkdobrev:
In Bootstrap documentation and other similar plugins they say `collapse,` `collapsing`, `collapsed` and `collapsible`. http://getbootstrap.com/javascript/#collapse-example-accordion

I think `collapse` is pretty good verb as it really shows what it is for compared to `hide`, `slide up`, `accordion` or something else.

vvo:
Ok so `collapsible: {
  collapsed: true
}` works for you?

hkdobrev:
@vvo :+1:

vvo:
implemented `collapsible: { collapsed: true }` mapped on showMore, thanks all for review!

PR #856, opened 2016-02-17
fix(searchBox): stop updating at every keystroke

vvo:
WIP but ok for review


vvo:
One drawback to this approach is that the url is updated after you stop typing. That's the goal of the PR but it can be disturbing if it takes too long and "too long" is user dependent. cc @bobylito 

Still, the refactor around search-box is valid and I will extract it and submit a PR

vvo:
Not quite good/ready yet. Refactor still legit and moved

PR #863, opened 2016-02-18
fix(Template): now render() when templateKey changes

vvo:
templateKey can be updated in some cases while the data stayed the
same, example: showMore => showLess


bobylito:
LGTM!

PR #864, opened 2016-02-18
fix(toggle): pass isRefined to toggleRefinement

vvo:
Was broken since perf improvements: we did not wanted to bind the
function with isRefined in render() to avoid generating new function
references each time.

But we still need it, so passed all the way down from RefinementListItem

### DIFF
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -19,8 +19,8 @@ class RefinementList extends React.Component {
     return nextState !== this.state || !isEqual(this.props.facetValues, nextProps.facetValues);
   }
 
-  refine(value) {
-    this.props.toggleRefinement(value);
+  refine(facetValueToRefine, isRefined) {
+    this.props.toggleRefinement(facetValueToRefine, isRefined);
   }
 
   _generateFacetItem(facetValue) {
@@ -53,8 +53,9 @@ class RefinementList extends React.Component {
 
     return (
       <RefinementListItem
-        facetValue={facetValue[this.props.attributeNameKey]}
+        facetValueToRefine={facetValue[this.props.attributeNameKey]}
         handleClick={this.handleItemClick}
+        isRefined={facetValue.isRefined}
         itemClassName={cssClassItem}
         key={key}
         subItems={subItems}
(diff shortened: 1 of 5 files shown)


src/components/RefinementList/RefinementListItem.js
```diff
@@ -34,11 +38,12 @@ class RefinementListItem extends React.Component {
 }
 
 RefinementListItem.propTypes = {
-  facetValue: React.PropTypes.oneOfType([
+  facetValueName: React.PropTypes.oneOfType([
```
bobylito: Why do we add name in the prop name? It can be either a string or a number.

vvo: Yes indeed, I guess it was to avoid mixing facetValue the object (with the name property) and facetValue here which is in fact, facetValue.name.

Yes it can contain either a string or number but it is always sent as "name" to the templates. Wonder how we can make this better?

src/components/RefinementList/RefinementList.js
```diff
@@ -80,36 +81,36 @@ class RefinementList extends React.Component {
   //
   // Finally, we always stop propagation of the event to avoid multiple levels RefinementLists to fail: click
   // on child would click on parent also
-  handleItemClick(value, e) {
-    if (isSpecialClick(e)) {
+  handleItemClick({facetValueName, originalEvent, isRefined}) {
```
bobylito: Destructuring is nice but why using it here? This is not a public API, is it?

src/components/RefinementList/RefinementListItem.js
bobylito: **spinner** ;)

bobylito:
Overall it's great! I like the renaming of e to originalevent. We should now go through all the code base and make it a convention :)

src/components/RefinementList/RefinementList.js
vvo: Nope, not a public API, still found it easier than having to know argument positions.
What do you think?

vvo:
inlined the commit message to the PR description!

src/components/RefinementList/RefinementList.js
bobylito: I think it is very convenient for user facing API, but it still add an overhead (even small). Your call :)

vvo:
updated the facetValueName => facetValueToRefine

PR #865, opened 2016-02-19
fix(searchBox): stop setting the query twice

vvo:
We were previously setting the query on every keystroke twice, because
of the keyup + input events.

Now fixed.

This was not triggering any bad behavior but extra work and maybe bad
performance because the way we set queries in the helper (to be worked
on)

+ refactor
+ update tests


vvo:
This is ready for review

vvo:
@algolia/javascript Anyone for review?

bobylito:
on it

bobylito:
LGTM

vvo:
thx @bobylito

PR #873, opened 2016-02-25
feat(instantsearch): allow overriding the helper.search function

vvo:
add a new option to the instantsearch: searchFunction. It allows user
to override the default call to helper.search() by anything they want.

With this, they will be able to implement:
- do not search at first render
- do not search when query < 3 chars
- any conditional we cannot think of

Example usages:

```js
var search = instantsearch({
  searchFunction(helper) {
    if (initialRender) {
      initialRender = false;
      return;
    }

    if (helper.state.query.length < 3) {
      return;
    }

    helper.search();
  }
})
```

### DIFF
--- a/dev/app.js
+++ b/dev/app.js
@@ -302,10 +302,6 @@ search.addWidget(
   })
 );
 
-search.once('render', function() {
-  document.querySelector('.search').className = 'row search search--visible';
-});
-
 search.addWidget(
   instantsearch.widgets.priceRanges({
     container: '#price-ranges',
@@ -337,6 +333,11 @@ search.addWidget(
     ]
   })
 );
+
+search.once('render', function() {
+  [...document.querySelectorAll('.smooth-search--hidden')]
+    .forEach(element => element.classList.remove('smooth-search--hidden'));
+});
 
 search.start();
 
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -4,11 +4,13 @@ import algoliasearchHelper from 'algoliasearch-helper';
 import forEach from 'lodash/collection/forEach';
 import merge from 'lodash/object/merge';
 import union from 'lodash/array/union';
+import clone from 'lodash/lang/clone';
 
 import {EventEmitter} from 'events';
 
 import urlSyncWidget from './url-sync.js';
 import version from './version.js';
+import helpers from './helpers.js';
 
 function defaultCreateURL() { return '#'; }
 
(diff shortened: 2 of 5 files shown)


vvo:
cc @seafoox this is what you are looking for

src/lib/InstantSearch.js
```diff
@@ -109,14 +120,25 @@ Usage: instantsearch({
       this.searchParameters
     );
 
+    if (this._searchFunction) {
+      this._originalHelperSearch = helper.search.bind(helper);
+      helper.search = this._wrappedSearch.bind(this);
```
vvo: We might want to clone the helper first, to leave the original one "untouched". What do you think? cc @bobylito

vvo:
Other possible "implementations" of this:
- create a instantsearch.search = function() {} that widget will need to call

But it would mean having to re-do our widgets and also might surprise our users.

src/lib/InstantSearch.js
vvo: But should be ok, since this helper belongs to instantsearch instance, its not created by the user and then injected in the lib.

We do expose instantsearch.helper but its pretty clear thats the one from the helper, and it this case .search will behave differently.

Shipow:
up :)

Jerskouille:
I really like this.

dev/app.js
```diff
@@ -338,5 +334,10 @@ search.addWidget(
   })
 );
 
+search.once('render', function() {
+  [...document.querySelectorAll('.smooth-search--hidden')]
```
vvo: I modified the dev example to show the searchbox first and then render the whole interface when ready.

vvo:
This PR means that when you are using searchFunction, we will do a shallow clone of the helper. Not a deep clone.

timkelty:
> This PR means that when you are using searchFunction, we will do a shallow clone of the helper at each keystroke (or search). Not a deep clone.

@vvo Got thrown by this as I was trying to use `AlgoliaSearchHelper` and they don't exist. Docs just say `options.searchFunction`'s...["with the helper as a parameter"](https://community.algolia.com/instantsearch.js/documentation/#initialization)

At any rate, if I want to call `AlgoliaSearchHelper` methods from within `searchFunction` (e.g. `.on`, `.setPage`), what's the best way to do this?

vvo:
@timkelty Sorry this is a documentation mistake, the helper you are being given in searchFunction only has the  "helper.search()" method on it.

Here's how to use it:

```js
var search = instantsearch({
    search.helper.anyHelperMethod();
    search.helper.anyHelperMethod();
    helper.search(); // trigger a new search
  }
})
```

PR #876, opened 2016-02-26
Do not update URL with searchOnEnterKeyPressOnly while typing

### LINKED ISSUE #875: When searchOnEnterKeyPressOnly is true, URL is still updated on each keystroke
Related to #837, but this is a separate issue only when `searchOnEnterKeyPressOnly` option is set to `true`.

I would expect when search is triggered only on Enter, the URL to also be updated only when Enter is pressed and search is performed. However the URL is actually updated on each keystroke.

hkdobrev:
Fixes #875

### DIFF
--- a/src/widgets/search-box/__tests__/search-box-test.js
+++ b/src/widgets/search-box/__tests__/search-box-test.js
@@ -245,9 +245,9 @@ describe('searchBox()', () => {
         widget.getInput = sinon.stub().returns(input);
       });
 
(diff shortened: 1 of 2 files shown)


vvo:
As yes @hkdobrev we always use develop as the main develop branch, sorry about that.

I did a few fixes on the searchbox lately: https://github.com/algolia/instantsearch.js/pull/865/files

vvo:
see #880

PR #878, opened 2016-02-29
chore(babel6): move to babel6

vvo:
A lot of the work was about removing all calls to require() on code
that is using es6 export.

See this good blog post for more information:
https://medium.com/@kentcdodds/misunderstanding-es6-modules-upgrading-
babel-tears-and-a-solution-ad2d5ab93ce0


Jerskouille:
:100:

PR #880, opened 2016-02-29
fix(searchBox): stop updating query at eachkeystroke with searchOnEnterKeyPressOnly

vvo:


redox:
LGTM :+1:

hkdobrev:
@vvo Thanks! Would this something which could be released as `1.2.x` or it'd need to be released in `1.3.x` because of the other changes in `develop`?

vvo:
@hkdobrev It will be released in 1.3.x indeed since the changes to searchBox. But we are very close now to 1.3.x release

PR #884, opened 2016-03-01
feat(searchBox): allow to pass a queryHook

vvo:
The queryHook can be used to implement debouncing for example.

Usage:

```js
searchBox({
  queryHook: (originalQuery, search) => // either call search() or search(newQuery);
});
```

### DIFF
--- a/src/widgets/search-box/search-box.js
+++ b/src/widgets/search-box/search-box.js
@@ -27,6 +27,9 @@ const KEY_SUPPRESS = 8;
  * wrapping div (if `wrapInput` set to `true`)
  * @param  {string|string[]} [options.cssClasses.input] CSS class to add to the input
  * @param  {string|string[]} [options.cssClasses.poweredBy] CSS class to add to the poweredBy element
+ * @param  {function} [queryHook] A function that will be called everytime a new search would be done. You
+ * will get the query as first parameter and a search(query) function to call as the second parameter.
+ * This queryHook can be used to debounce the number of searches done from the searchBox.
  * @return {Object}
  */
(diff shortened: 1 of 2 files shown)


src/widgets/search-box/search-box.js
```diff
@@ -137,26 +147,33 @@ function searchBox({
...
       }
 
-      function setQuery(e) {
-        helper.setQuery(getValue(e));
+
+      function maybeSearch(originalQuery) {
+        if (queryHook) {
+          queryHook(
+            originalQuery,
+            newQuery => newQuery !== undefined ? search(newQuery) : search(originalQuery)
```
bobylito: Less magic if we were to let the user pass the original query himself.

vvo: Agreed, changing that

src/widgets/search-box/search-box.js
```diff
@@ -168,6 +185,10 @@ function searchBox({
         return actualEvent => actualEvent.keyCode === expectedKeyCode && fnToExecute(actualEvent);
       }
 
+      function getQueryAndCall(fnToExecute) {
```
bobylito: fnToExecute could be reduced to `f` or `fn`. Also those functional helpers :100: could be moved somewhere else.

vvo: I renamed to "func" and moved them at the bottom of the file for now to indicate that there's no shared state

vvo: If we reuse them somewhere we can move them

bobylito:
Nice! LGTM!

PR #886, opened 2016-03-01
fix(searchBox): handle external updates of the query

### LINKED ISSUE #803: Programmatically updating the query will not update the searchbox
Should we provide an option or handle it by default?

There have been usecases report lately about this and the workaround was not only to change the helper query but also update the DOM element manually (which is hacky and may result in multiple searches?)

vvo:
fixes #803


bobylito:
Nice LGTM

PR #895, opened 2016-03-02
docs(toggle): refresh the widget description

Jerskouille:
When we updated the toggle widget to provide on/off values, we didn't properly document the update.
Related to http://stackoverflow.com/questions/35750620/setting-initial-facet-refinement-values-when-using-instantsearch-js .


vvo:
thx

PR #896, opened 2016-03-03
chore(docs): switch to jekyll 3, digest all assets

vvo:
It was a bit of a nightmare: I only wanted to digest all assets and thought it would be easy.

Turns ou I had to move to jekyll 3, then handle a lot of weird sh## (expected).

In the end, everything is working but I was unable to make the dev docs task mount the website on the virtual /instantsearch.js/ path.

I guess it was working before, but it's no more. Submitted some issues to relevant (I hope) projects.

In the meantime, this should be merged

- digest all assets to avoid caching issues
- concatenate scripts that can be contenated
- minify css and js


docs/_config.yml
```diff
@@ -1,7 +1,5 @@
-# Site settings
 title: instantsearch.js
 description: instantsearch.js is a library of widgets designed to help you build high performance instant search experiences with Algolia
-baseurl: "/instantsearch.js" # the subpath of your site, e.g. /blog/
```
redox: @vvo why removing the `baseurl`? Is that the reason why don't manage to get it work inside the subdir?

vvo:
Yep basically its no more feasible to serve the dev app in a virtual directory /instantsearch.js/.

Building works well and generates the correct uris and assets.

But dev fails with a not root baseurl.

I think I have another solution to still serve it on the virtual directory, will add it this morning.

docs/_production.yml
```diff
@@ -0,0 +1,2 @@
+# waiting for https://github.com/imathis/guard-jekyll-plus/issues/49
+baseurl: "/instantsearch.js"
```
redox: ^ alright :)

docs/_config.yml
pixelastic: We have to keep the baseurl if we're going to push on github pages. It has to match the name of the repo (the part after `algolia.github.io`) and be used as a prefix in every link in our templates.

docs/_layouts/default.html
```diff
@@ -37,9 +36,7 @@
     <link rel="stylesheet" media="screen" href="//fonts.googleapis.com/css?family=Raleway:300,800" />
     <link rel="stylesheet" media="screen" href="//fonts.googleapis.com/css?family=Open+Sans:300,400,600" />
 
-    <!--[if IE]><link rel="shortcut icon" href="{{ "/img/favicon.png" | prepend: site.baseurl }}"><![endif]-->
-    <link rel="icon" href="{{ "/img/favicon.png" | prepend: site.baseurl }}">
-
+    <link rel="icon" href="{% asset_path favicon.png %}">
```
pixelastic: Do we revv the favicon as well? Don't browsers automatically do a call to `favicon.ico`?

vvo: Indeed, they do both I believe => they use the provided one but also always call /favicon.ico

docs/_config.yml
vvo: Its kept, just only provided when building for production, see https://github.com/algolia/instantsearch.js/pull/896#discussion_r54847698

PR #913, opened 2016-03-07
fix(Template): stop leaking `data="[object Object]"` attributes in production

### LINKED ISSUE #899: `<div data="[object Object]">` in 1.3.0
I just moved from 1.2 to 1.3, and I realize that my markup now have a lot of `data="[object Object]"` while inspecting with the Chrome inspector. This has no side effect that I can currently see, but this seems like something that shouldn't be visible in the markup.

## 1.3
image
image

## 1.2.5
image
image

vvo:
Avoid `<div data="[object Object]">` being inserted in the DOM in
production builds.

The bug was introduced by
https://github.com/algolia/instantsearch.js/commit/fcfb990c9d059f59389b753b5abdecfba0dcd0f8
where we started using transform-react-remove-prop-types.

We cannot rely on PropTypes in production env because they are
removed.

We ask the dev to be explicit in what he's doing: use rootProps versus
forwarding everything a bit magically.

fixes #899

### DIFF
--- a/src/components/Template.js
+++ b/src/components/Template.js
@@ -2,8 +2,6 @@ import React from 'react';
 
 import curry from 'lodash/function/curry';
 import cloneDeep from 'lodash/lang/cloneDeep';
-import keys from 'lodash/object/keys';
-import omit from 'lodash/object/omit';
 import mapValues from 'lodash/object/mapValues';
 
 import hogan from 'hogan.js';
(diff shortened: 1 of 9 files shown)


src/components/Template.js
```diff
@@ -37,12 +35,10 @@ class Template extends React.Component {
       return null;
     }
 
-    const otherProps = omit(this.props, keys(Template.propTypes));
-
     if (React.isValidElement(content)) {
       return (
         <div
-          {...otherProps}
```
vvo: This is where data={} was sent. In dev it was ok because we did omit using the propTypes, but in production it was not

src/components/Template.js
```diff
@@ -50,7 +46,7 @@ class Template extends React.Component {
 
     return (
       <div
-        {...otherProps}
```
vvo: This is where data={} was sent. In dev it was ok because we did omit using the propTypes, but in production it was not because propTypes are removed from the production builds. Thus doing an omit() with an empty object

redox: Good catch :+1:

PR #915, opened 2016-03-07
perf(linters): Greatly improve the `npm run lint` task speed

pixelastic:
The `npm run lint` task now takes about ~3s on my machine (was
taking XXX).

Previously:
```
$ time npm run lint
npm run lint  10,84s user 1,22s system 107% cpu 11,187 total
```
Now:
```
$ time npm run lint
npm run lint  3,73s user 0,28s system 116% cpu 3,445 total
```

Basically all I did was specify the `./**/*.js` or `./**/*.scss`
pattern on both linters and that makes it much faster.

For scss, I also cleaned up a bit the list of ignore patterns (as we
now use Jekyll 3, some files do not have the front-matter anymore and
so we do not need to exclude them).


scripts/lint-js
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+eslint ./**/*.js --quiet
```
vvo: Always wonder if the index.js is linted with this?

vvo: Might not be a big deal, we could even run another command if needed

pixelastic:
@vvo You were right that it wasn't checking at the root. I updated it

vvo:
Thx :)

PR #922, opened 2016-03-10
feat(url-sync): use the new mapping option

### LINKED ISSUE #838: Ability to change the shortened parameters
On some integrations, we want to integrate instantsearch without changing the current behaviour (SEO purposes for instance). So we integrate on top of them.
For the URL, most search pages use the `q` query parameter to handle search, which makes it trivial to integrate in them.
However, some others use `query`. For these, it would be really neat to change the shortened version of the `query` parameter in order to have the parameter synchronization directly.

My example:
- Zendesk's `/search` page accepts a `query` parameter.
- I want to keep the default back-end search (which uses `query`) to have decent SEO on a query
- However, by activating instantsearch, I end up with two parameters in the url `/search?query=a&q=a`
- When I change the state the url becomes `/search?query=a&q=abc`.
  If someone shares this URL, the back-end search will return results for `a` and the front-end results for `abc`, which is pretty uncool.

If I could change the "shortened" version of `query` from `q` to `query`, everything would be ultra simple.
I post it here because I believe it could be a `urlSync` parameter.

In order to do so, it would require us to expose the shortener on the helper, which is what https://github.com/algolia/algoliasearch-helper-js/pull/280 aims to do.

Jerskouille:
No test added to this commit because the `url-sync-test.js` does not yet exist.
Will close #838 .
Uses https://github.com/algolia/algoliasearch-helper-js/pull/281 .

### DIFF
--- a/dev/app.js
+++ b/dev/app.js
@@ -5,7 +5,14 @@ var search = instantsearch({
   appId: 'latency',
   apiKey: '6be0576ff61c053d5f9a3225e2a90f76',
   indexName: 'instant_search',
-  urlSync: window.history && 'pushState' in window.history ? true : {useHash: true}
+  urlSync: {
+    useHash: !(window.history && 'pushState' in window.history),
+    mapping: {
+      q: 'query',
+      hPP: 'hits',
+      hFR: 'hierarchical'
+    }
+  }
 });
 
 search.addWidget(
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -18,11 +18,19 @@ function defaultCreateURL() { return '#'; }
  * @param  {string} options.indexName The name of the main index
  * @param  {string} [options.numberLocale] The locale used to display numbers. This will be passed
  * to Number.prototype.toLocaleString()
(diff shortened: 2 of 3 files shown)


dev/app.js
```diff
@@ -5,7 +5,14 @@ var search = instantsearch({
...
   indexName: 'instant_search',
-  urlSync: window.history && 'pushState' in window.history ? true : {useHash: true}
+  urlSync: {
+    useHash: !(window.history && 'pushState' in window.history),
+    mapping: {
+      q: 'query',
+      hPP: 'hits',
+      hFR: 'hierarchical'
+    }
+  }
```
Jerskouille: Changes here only for demo purposes. We can remove them as soon as it's reviewed.

vvo: It's ok to keep them I believe, the demo should be a frankenstein setup of all the features. So that if one break, we can see it :)

bobylito: This is the kitchen sink of instantsearch.js :)

Jerskouille: As you wish :)

vvo:
Thanks!

PR #934, opened 2016-03-17
perf(refinementList): Stop creating URL for hidden refinements.

pixelastic:
This actually fixes all my perf issues on the SteamDB demo.

### DIFF
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -37,7 +37,8 @@ class RefinementList extends React.Component {
       );
     }
 
-    let templateData = {...facetValue, cssClasses: this.props.cssClasses};
+    let url = this.props.createURL(facetValue.name);
+    let templateData = {...facetValue, url, cssClasses: this.props.cssClasses};
 
     let cssClassItem = cx(this.props.cssClasses.item, {
       [this.props.cssClasses.active]: facetValue.isRefined
@@ -127,6 +128,8 @@ class RefinementList extends React.Component {
     }
 
     const limit = this.state.isShowMoreOpen ? this.props.limitMax : this.props.limitMin;
+    let displayedFacetValues = this.props.facetValues.slice(0, limit);
+
     const showMoreBtn =
       this.props.showMore ?
         <Template
@@ -138,7 +141,7 @@ class RefinementList extends React.Component {
 
     return (
       <div className={cx(cssClassList)}>
-        {this.props.facetValues.map(this._generateFacetItem, this).slice(0, limit)}
+        {displayedFacetValues.map(this._generateFacetItem, this)}
         {showMoreBtn}
       </div>
     );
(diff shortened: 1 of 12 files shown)


vvo:
props are frozen so you cannot change them inside render, otherwise: legit change.

src/components/RefinementList/RefinementList.js
```diff
@@ -127,6 +128,8 @@ class RefinementList extends React.Component {
     }
 
     const limit = this.state.isShowMoreOpen ? this.props.limitMax : this.props.limitMin;
+    this.props.facetValues = this.props.facetValues.slice(0, limit);
```
vvo: this is not doable in React, you need to get a new reference instead of modifying props

vvo:
@pixelastic You need some modifications for this PR to be merged

src/components/RefinementList/RefinementList.js
pixelastic: What do you mean by not doable? The above code is currently working. 

Is that a React best practice?

pixelastic: Ok my bad, just saw the Travis tests. I'll fix it.

Can't we have an eslint rule to detect that somehow?

vvo: I guess it must exists yes

vvo: Otherwise running `npm test` on your machine should have made them fail?

pixelastic: Nope, tests went through without issues (I have an `npm test` on `pre-push`)

vvo: So its weird if they failed on travis, maybe you need to npm install

pixelastic:
@vvo  I've updated the PR to not update the `this.props` directly, but it still fails on Travis, while it's seems to be working locally.

vvo:
Hmm the functional tests seems broken, I would say this PR breaks something because the functional tests are looking for [href=Appliances something] and they cannot find it anymore.

See the readme for how to launch the functional tests

pixelastic:
I fixed the functional test. I actually had to make the change (passing a method to generate the url instead of the generated url) to each widget that was using the RefinementList component.

vvo:
looks very nice and clean.

Should help a lot performance wise.

My mistake for generating the urls upfront in a desperate try to avoid render() calls by doing deep comparisons while since you activate the urlSync, every facet value is different at every user interaction (query is different, url is different, render)

src/widgets/hierarchical-menu/hierarchical-menu.js
```diff
@@ -137,19 +135,23 @@ function hierarchicalMenu({
...
+      // Bind createURL to this specific attribute
+      function _createURL(facetValue) {
+        return createURL(state.toggleRefinement(hierarchicalFacetName, facetValue));
+      }
+
       ReactDOM.render(
         <RefinementList
           attributeNameKey="path"
           collapsible={collapsible}
+          createURL={_createURL}
```
vvo: The only "better" optimisation we could do is know upfront that urlSync is deactivated (thus, no different urls per facet values) but I do not see an easy way to do so.

And it may be an edge case we do not need to support.

vvo:
LGTM, I had only one comment, let me know what you think @pixelastic

pixelastic:
I say we wait for it to become a bottleneck before fixing it ;)

PR #942, opened 2016-03-21
fix(pr): Testing PR on master

### LINKED ISSUE #931: PR sent to master are accepted by default
See https://github.com/algolia/instantsearch.js/pull/930

cc @pixelastic?

pixelastic:
The script to validate PRs was really confusing about what being a PR
meant or not. This version should work better.

Basically it will succeed when not on Travis, reject PRs not on
develop, except if they are on master with either a `docs()` or
a `hotfix` mention.

So this very own PR, even if done on master, should just be valid and fix #931


vvo:
Thx

PR #968, opened 2016-04-04
docs(codebox): the options tab is now the default

### LINKED ISSUE #966: [Docs] When I search for an option, I want the "All options" tab to be displayed
Otherwise, we don't understand the result:

gif

redox:
- Renamed "All Options" to "Usage"
 - Renamed "Snippet" to "Example"
 - Moved "Requirements" at the 2nd position
 - do not use display:none in the raw HTML code

Fix #966 

screen shot 2016-04-04 at 12 01 37

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -117,7 +117,7 @@ var search = instantsearch({
 });
 {% endhighlight %}
   </div>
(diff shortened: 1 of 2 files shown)


vvo:
will be live in a bit

PR #969, opened 2016-04-04
Fix/967

redox:


vvo:
This fails when building from production: asset digests are no more generated. Arg

vvo:
Only je jekyll-assets functions (js, css, asset_path) will generate digests

redox:
> Only je jekyll-assets functions (js, css, asset_path) will generate digests

Arg...

vvo:
The whole jekyll + livereload + jekyll assets field is a complete mess. Not because of our requirements/usage, it's just a mess solved by either doing code (gulp, grunt) or using extensions that are failing like we see right now.

I suggest we move to middleman very soon because I already spent one hour trying to figure out what's not working.

redox:
@vvo I finally made it work downgrading to jekyll 2.5 (livereload OK, update CSS/JS triggers rebuild, asset digests in production).

PR #972, opened 2016-04-11
feat(serve): Adding `npm run serve` command

pixelastic:
`npm run serve` will build the current instantsearch and expose the
file on http://127.0.0.1:8080/instantsearch.js. It will get
automatically updated in each change.

This is useful when you want to test your branch in a specific
project. Just run `npm run serve` in the instantsearch repo and link
to the file in your project.

I've regenerated the shrinkwrap but had to run a `npm prune` before
shrinkwrap allowed me to do it, so I updated the doc accordingly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,6 +14,7 @@ Hi (future) collaborator!
 - [New API proposal](#new-api-proposal)
 - [How to write issues](#how-to-write-issues)
 - [Development workflow](#development-workflow)
+- [Serving the build](#serving-the-build)
 - [Adding/Updating a package](#addingupdating-a-package)
 - [Removing a package](#removing-a-package)
 - [Commit message guidelines](#commit-message-guidelines)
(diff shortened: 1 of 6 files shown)


vvo:
Can you only commit the relevant npm-shrinkwrap part about http-server because React was updated to v15 here also.

bobylito:
Can't you do that with 'npm run dev'?

vvo:
Indeed it seems you can just access http://0.0.0.0:8080/instantsearch.js when running `npm run dev`

bobylito:
So maybe it could be a recipe instead of yet another command? I mean, we need to document the workflow and our tips and tricks anyway so why not start by this one (using a command generate a custom build of instantsearch.js to use in our projects) ?

vvo:
This can be added in CONTRIBUTING.md

vvo:
I am ok with npm run serve (we already have it in other repos), can you just add a note in CONTRIBUTING.md?

pixelastic:
Ah yes, that's right it is available in `npm run dev` as well. I thought it would come packaged with the custom `app.js`. My bad.

Not sure if you want to keep using `dev` and document that we can access the generated `instantsearch.js` from here, or use a new command that only expose the `instantsearch.js` (without the playground).

Another thing to note is that by using `http-server` to serve the directory instead of webpack, it now uses available ports (starting from `8080`). This helps a lot when you have multiple projects, all using the `8080` port.

vvo:
> This helps a lot when you have multiple projects, all using the 8080 port.

Agreed this is what I thought also buuut you inlined "go to localhost:8080" while it may be 8081 8082 .. :D

Can you just add a note in CONTRIBUTING.md and fix the conflict, then we merge

vvo:
updated and force pushed (sorry, was a bit tricky with shrinkwrap changes)

PR #975, opened 2016-04-11
fix(pagination): Disabled pagination link can no longer be clicked

### LINKED ISSUE #974: [pagination]: Disabled links can still be clicked
The way we _disable_ pagination links currently is by adding a CSS class to them, and make them look greyed out. But in the DOM, they are still links and can be followed, leading the user to a page with no results.

image
image

I suggest we replace the `a` with `span` when they are disabled.

pixelastic:
Fixes #974.

`PaginationLinks` with the `isDisabled` props now render as `span` and
not as `a`. We also remove the now unwanted `href`, `aria-label` and
`onClick` props.

There were no tests files for PaginationLink, and I did not added one.
I'd like #948 to be merged before adding those kind of fine-grained
tests.

### DIFF
--- a/src/components/Pagination/Pagination.js
+++ b/src/components/Pagination/Pagination.js
@@ -32,6 +32,7 @@ class Pagination extends React.Component {
         ariaLabel={ariaLabel}
         cssClasses={cssClasses}
         handleClick={this.handleClick}
+        isDisabled={isDisabled}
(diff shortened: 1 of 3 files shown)


vvo:
Only worried this could lead to rendering issues if some users were using "selector a" but should be ok since BEM

PR #976, opened 2016-04-14
fix(showMore): hide showMore when no more facet values to show

vvo:
If number of facet values is 5 and limit is 5 then we were still
displaying the showMore button.

Now fixed, no added tests, waiting for
#948


bobylito:
👍 Looks good!

PR #988, opened 2016-04-25
Add `pushStateObject` to get Turbolinks compatibility

### LINKED ISSUE #985: urlSync + Turbolinks back button not working
I am using the using instantsearch with turbolinks and am experiencing an issue where the back button is not working correctly.

My search coffeescript

```
search = instantsearch(
      appId: 'id'
      apiKey: 'key'
      indexName: 'index'
      urlSync: {
        threshold: 300
      })
```

When I search products, the url is updated as expected, search works.  If I click on a product, than click the `back` button, the url changes to the previously generated search url, but I stay on the product#show page for the instance I was viewing.  Clicking `back` button a second time takes me to the root /products page.  

This issue doesn't occur when setting `useHash: true`, unfortunately analytics team needs this to be query params.

kle-roy:
This PR aims to fix #985 and get Turbolinks compatibility by setting `pushStateObject: { turbolinks: true }`.

You may reproduce the problem described by @ThomasBush on http://www.welcometothejungle.co/companies : 

* Activate some filters/facets
* Click on whatever company
* Go back => NOK (it doesn't work until you reach back `/companies` in the history)

---

see:

* https://github.com/algolia/instantsearch.js/issues/985
* https://github.com/turbolinks/turbolinks-classic/issues/363#issuecomment-85626145

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,9 @@ In order to activate this feature, you need to add the urlSync object. It accept
       it'll use the query parameters using the modern history API.
    - threshold:number time in ms after which a new state is created in the browser
       history. The default value is 700.
+   - pushStateObject:object allows you to give a custom state object to `pushState`
+      (for example, to make this library compatible with Turbolinks, you will
+      need to set your state object to `{ turbolinks: true }`)
 
 All those parameters are optional and a minimal configuration looks like:
 
--- a/src/lib/url-sync.js
+++ b/src/lib/url-sync.js
@@ -58,11 +58,11 @@ let modernUrlUtils = {
   onpopstate: function(cb) {
     window.addEventListener('popstate', cb);
   },
-  pushState: function(qs) {
-    window.history.pushState(null, '', getFullURL(this.createURL(qs)));
+  pushState: function(qs, options = {}) {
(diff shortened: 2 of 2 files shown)


vvo:
Hi! Thanks for the contribution.

I think I get the fix: instead of pushing "null" as the state of the history API, you want to push some default value needed by your app to work?

If that's it, then I will make it a proper generic non-turbolinks related option that other people could use (like instantsearch({getHistoryState: function() { return {turboLinks: true} }}).

What do you think?

kle-roy:
> If that's it, then I will make it a proper generic non-turbolinks related option that other people could use (like instantsearch({getHistoryState: function() { return {turboLinks: true} }}).

Yep, I just need to specify `{turbolinks: true}` on the state object.

The idea of `pushStateObject` was to be generic, but if you do have a proper solution, I'll take it :)

vvo:
Indeed! You already made it generic. I just misread your PR. Merging :)

vvo:
Can you have a look at https://github.com/algolia/instantsearch.js/pull/992?

vvo:
@kle-roy I did a new PR that allows passing a function rather a single object so that you can update it easily based on your own app state

PR #990, opened 2016-04-26
fix(lifecycle): save configuration done in widget.init

vvo:
Before this commit, every configuration done in widget.init would not
be remembered when going back in the browser history.

Scenario: You built a custom widget using helper.setQueryParameter in
widget.init. Then you load "/", widget.init() sets some helper params,
you go to page 2, you click on <= back. Before this commit we would
forget what was done in widget.init(). Now it works.

This was needed to fix the "rootPath" option of the hierarchicalMenu.
The rootPath option triggers a state change from inside the helper
(which may have been a bad idea after all) => it was not remembered
when going back to empty url.

Thanks to @maxiloc for the bug report and big help solving this!

### DIFF
--- a/dev/app.js
+++ b/dev/app.js
@@ -305,6 +305,7 @@ search.addWidget(
       link: 'facet-value',
       count: 'facet-count pull-right'
     },
+    rootPath: 'Cameras & Camcorders > Digital Camera Accessories',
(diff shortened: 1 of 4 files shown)


maxiloc:
👍

PR #992, opened 2016-04-26
feat(urlSync): allow overriding replaceState(state)/pushState(state)

vvo:


kle-roy:
@vvo I did test this branch on WTTJ with the following parameters : 

```
urlSync:
  useHash: false
  getHistoryState: ->
    { turbolinks: true }
```

=> everything is working as expected w/ Turbolinks, thanks 👍

vvo:
Great! This will be released today

vvo:
We have a regression on IE11 so release will be a bit delayed, sorry

ThomasBush:
@vvo this was closed, does that mean the IE11 issue is fixed?  I am getting a lot of pressure to switch from hashed params to query params to solve issues we're experiencing in Adwords.

vvo:
This is definitely merged @ThomasBush, if you have a specific need or question around urlSync let us know in a new issue and we will have a look.

Switching from hash based to history API (basically removing useHash: true) should work by default.

PR #998, opened 2016-04-28
fix(rootpath): remember rootpath option on 'back' button

vvo:
Before this commit, when using the rootpath option and clicking on
backbutton after switching a page would result in the rootpath option
seeming "forgotten"

This was triggered because rootpath is a "magical" (BAD!!) option that
is triggering a refinement internally in the helper's SearchParameters
constructor.

Thus, in the urlSync widget, to determine the "original configuration"
(= all widgets getConfiguration before url sync) we need to go through
the SearchParameters constructor.

Also note that we originally wanted to understand the url sync
"original config" as being widget's getConfiguration + widget's init
but that's no doable because the way we implemented url sync. Indeed
there's no way to differentiate the state of the url from the state of
the widget's init action after all widget.init are called.

Thus when clicking back, we are not able to tell what are the
parameters we want to go back to.

In a V2 of instantsearch.js we should be able to have init({helper,
state, urlState}).

Or maybe there's another way. I will need to discuss this with
@bobylito :)

### DIFF
--- a/src/lib/url-sync.js
+++ b/src/lib/url-sync.js
@@ -98,7 +98,10 @@ class URLSync {
   }
 
   getConfiguration(currentConfiguration) {
-    this.originalConfig = currentConfiguration;
+    // we need to create a REAL helper to then get its state. Because some parameters
+    // like hierarchicalFacet.rootPath are then triggering a default refinement that would
+    // be not present if it was not going trough the SearchParameters constructor
+    this.originalConfig = algoliasearchHelper({}, currentConfiguration.index, currentConfiguration).state;
     let queryString = this.urlUtils.readUrl();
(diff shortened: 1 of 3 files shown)


src/lib/url-sync.js
```diff
@@ -98,7 +98,10 @@ class URLSync {
   }
 
   getConfiguration(currentConfiguration) {
-    this.originalConfig = currentConfiguration;
+    // we need to create a REAL helper to then get its state. Because some parameters
+    // like hierarchicalFacet.rootPath are then triggering a default refinement that would
+    // be not present if it was not going trough the SearchParameters constructor
+    this.originalConfig = algoliasearchHelper({}, currentConfiguration.index, currentConfiguration).state;
```
vvo: that's the only fix we needed in the end

maxiloc: nice

PR #1008, opened 2016-05-09
docs(transformData): Only document explicit objects for `transformData`

pixelastic:
We can currently pass for `transformData` either an object or
a function. If passed a function, it will be applied on every
template, while if passing an object, each key function will only be
applied to the template with a matching name.

We think the "top level" call to a function creates more issues than
an explicit definition. It usually creates weird effects when the
method is applied to a "no result" set, resulting in errors. To avoid
that, we will stop documenting the feature, and remove it completely
in a next major version.

I've updated the documentation to reflect this change, ie always
documenting `transformData.item` instead of just `transformData` as
the place to put the custom method. I've updated the `usage` errors
accordingly as well. It should still silently accept the top level function, but at least it is no longer documented and will be easier to deprecate in the future.

I also fixed a few errors in the docs, like removing empty default
values (it was only adding noise to the docs, and was not consistent).

I've also fixed an inconsistency between the documentation and the
code in the `currentRefinedValues` widget. It was previously not
possible to pass a `transformData.item` (this resulted in an error),
while passing a `transformData` directly was accepted, but didn't
produce any effect. We can now correctly pass `transformData.item`.

### DIFF
--- a/src/widgets/clear-all/clear-all.js
+++ b/src/widgets/clear-all/clear-all.js
@@ -21,9 +21,9 @@ let bem = bemHelper('ais-clear-all');
  * @function clearAll
  * @param  {string|DOMElement} options.container CSS Selector or DOMElement to insert the widget
  * @param  {Object} [options.templates] Templates to use for the widget
- * @param  {string|Function} [options.templates.header=''] Header template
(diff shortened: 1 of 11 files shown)


vvo:
Thanks!

PR #1018, opened 2016-05-16
fix(numericRefinementList): Correctly apply active class

### LINKED ISSUE #1010: numericRefinementList is not updating the "active" class when clicking items

pixelastic:
Fixes #1010

The `shouldComponentUpdate` of `RefinementList` was checking if the
`facetValues` from the props where actually changed before issuing
a new rendering. It turns out that in `numericRefinementList`, the
new `facetValues` was always equal to the already set one. More
precisely, the filters seemed to already be set in the props before
reaching `shouldComponentUpdate`.

Turns out that we were editing a variable in place instead of working
on a copy so the internal props where changing without the component
really noticing.

I updated the `render` method of the `numericRefinementList` to pass
a new array instead of working on the same copy over and over again.

I refactored the `shouldComponentUpdate` and `_toggleRefinement`
methods by splitting in several variables in my quest to find the
issue. Thought I might keep it that way.

I added an explicit regression test to make sure we are really not
touching existing variables.

Now the current and next props are really different, and the full
render is propagated and the correct classes applied.

### DIFF
--- a/src/widgets/numeric-refinement-list/numeric-refinement-list.js
+++ b/src/widgets/numeric-refinement-list/numeric-refinement-list.js
@@ -93,15 +93,18 @@ function numericRefinementList({
         templates
       });
 
-      this._toggleRefinement = facetValue => helper
-        .setState(refine(helper.state, attributeName, options, facetValue))
-        .search();
+      this._toggleRefinement = (facetValue) => {
+        let refinedState = refine(helper.state, attributeName, options, facetValue);
+        helper.setState(refinedState).search();
+      };
     },
     render: function({results, state, createURL}) {
-      let facetValues = options.map(facetValue => {
-        facetValue.isRefined = isRefined(state, attributeName, facetValue);
-        facetValue.attributeName = attributeName;
-        return facetValue;
+      let facetValues = options.map((facetValue) => {
+        return {
+          ...facetValue,
+          isRefined: isRefined(state, attributeName, facetValue),
+          attributeName: attributeName
+        };
       });
 
       // Bind createURL to this specific attribute
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -17,7 +17,10 @@ class RefinementList extends React.Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
-    return nextState !== this.state || !isEqual(this.props.facetValues, nextProps.facetValues);
+    let isStateDifferent = nextState !== this.state;
+    let isFacetValuesDifferent = !isEqual(this.props.facetValues, nextProps.facetValues);
+    let shouldUpdate = isStateDifferent || isFacetValuesDifferent;
+    return shouldUpdate;
   }
 
   refine(facetValueToRefine, isRefined) {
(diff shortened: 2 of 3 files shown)


pixelastic:
Do we build the Jekyll website on each PR? Seems a bit excessive (and slow)

src/widgets/numeric-refinement-list/numeric-refinement-list.js
```diff
@@ -93,15 +94,18 @@ function numericRefinementList({
         templates
       });
 
-      this._toggleRefinement = facetValue => helper
-        .setState(refine(helper.state, attributeName, options, facetValue))
-        .search();
+      this._toggleRefinement = (facetValue) => {
```
bobylito: What's the difference in this change?

src/widgets/numeric-refinement-list/numeric-refinement-list.js
```diff
@@ -93,15 +94,18 @@ function numericRefinementList({
...
+        let refinedState = refine(helper.state, attributeName, options, facetValue);
+        helper.setState(refinedState).search();
+      };
     },
     render: function({results, state, createURL}) {
-      let facetValues = options.map(facetValue => {
-        facetValue.isRefined = isRefined(state, attributeName, facetValue);
-        facetValue.attributeName = attributeName;
-        return facetValue;
```
bobylito: What about keeping the map and return a new object, like that:
```js
let facetValues = options.map(facetValue => {
  return {
    isRefined: isRefined(state, attributeName, facetValue),
    attributeName: attributeName
  };
})

**Comment by bobylito** [comment]
Very nice catch @pixelastic. I did some suggestions to improve the proposed fix :)

File: src/widgets/numeric-refinement-list/numeric-refinement-list.js (hunk 1 above)
Comment (pixelastic) [inline src/widgets/numeric-refinement-list/numeric-refinement-list.js]: Conceptually, there are no difference.

I refactored it while debugging to more easily inspect the new state. In the end I decided to keep the change, but it does not change anything, it's just (maybe) more readable.

File: src/widgets/numeric-refinement-list/numeric-refinement-list.js (hunk 2 above)
Comment (pixelastic) [inline src/widgets/numeric-refinement-list/numeric-refinement-list.js]: We still need the initial keys from the `facetValue`, so maybe something like this instead:

```
let facetValues = options.map(facetValue => {
  return {
    ...facetValue,
    isRefined: isRefined(state, attributeName, facetValue),
    attributeName: attributeName
  };
})
```

Yes, this could work. This would avoid including the `cloneDeep`.

bobylito: 👍

src/widgets/numeric-refinement-list/numeric-refinement-list.js
bobylito: OK 👍

vvo:
> Do we build the Jekyll website on each PR? Seems a bit excessive (and slow)

Yes, this is to ensure we are still able to build it. Otherwise one could break the releasing process or the website more easily.

pixelastic:
Could we maybe run all this tests in parallel if we subscribe to a paid Travis access?

vvo:
On travis the way to do so is to use build matrixes with different env variables like SCRIPT_TO_RUN. doable.

Current build time for PRS is 8 minutes, I think we can wait a bit to optimize it?

pixelastic:
This feels at least 4 times too long to me :)

If we could split the "build website" and "run webdriver.io tests" in separate Travis builds I think this will help. I'll try to give it a try

PR #1026, opened 2016-05-23
docs(faq): Add "hide results on init"

pixelastic:
I suggest adding a FAQ section to the documentation where we could put
some of the most common use cases of instantsearch that are either not
yet into the core, or are not suitable to be put into the core.

I just had a user asking me how to make instantsearch not display the
first results, and only display them on the first keystroke.

I wrote a small explanation, with code example and added it to the
docs.

Let me know what you think

image

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -1095,7 +1095,44 @@ in your index configuration.
 
 </div>
 
+### Hide results on init
 
+<div class="codebox-combo">
+<div class="code-box">
+  <div class="code-sample-snippet ignore">
+{% highlight javascript %}
+var search = instantsearch({
+  [...],
+  searchFunction: function(helper) {
+    var searchResults = $('.search-results');
+    if (helper.state.query === '') {
+      searchResults.hide();
+      return;
+    }
+    helper.search();
+    searchResults.show();
+  }
+}
+{% endhighlight %}
+  </div>
+</div>
+
+By default, the library will do an empty query (`''`) on startup and display the
+results for this query. If you'd like to display results only when users
+actually start typing, you can use the `searchFunction` hook.
+
+This method takes only one argument, `helper`, on which you can call `.search()`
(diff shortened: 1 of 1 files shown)


olance:
👍

bobylito:
Right now, the documentation page is quite long so what about having a dedicated page for the FAQ?

pixelastic:
I'm not against it, but with only two entries this might look weird (this could even got down to one if we create the `helper configurator` option).

@Shipow any thoughts on a new page?

bobylito:
The configuration widget, needs its space in the FAQ. I think we should answer there all the tricks we've learned so far. Thanks for starting that :)

Shipow:
yes worth a dedicated page I guess. This part can be long let's think when we will have 10 more entries. Right now i would put the content to the bottom of the page.

vvo:
This is not even a FAQ what you wrote, it's a tutorial/guide and the first of its kind. Let's write more of them in the future and release a guides page.

Also this PR has conflicts for now

olance:
@vvo I would say it's a long answer to a frequently asked question :P

ThibautGery:
Hi I am the user asking for help on the matters. The explanation is good enough for me.
You already have 3 different document pages about `instantsearch.js`:
 * the landing
 * the documentation
 * the tutorial  

Please don't add one more

vvo:
@ThibautGery This would be on the same website than instantsearch.js, not a new website.

It's hard to combine on the same page both the API documentation and advanced guides. And we do not want to put advanced guides on Algolia.com even if that's currently the case.

At some point instantsearch.js has became a subproduct and thus we need to put most of the documentation on the same website.

Having the instantsearch website with:
- homepage
- documentation
- guides

is a common pattern for tech projects so it should not be an issue? Or would you really prefer to have a single documentation page with everything inlined?

pixelastic:
I rebased the branch, but GitHub is slow on catching up.

PR #1027, opened 2016-05-23
docs(faq): Add explanation on default filters

pixelastic:
Following #1026 I also added an entry on how to set default filters on
the first loading of instantsearch. This question often pop ups in
support and we were never really explaining how to do it.

I tried to keep it as simple as possible for most cases when you
already have a widget in place on that specific attribute, and only
need to set default values.

I also explained how to do it on attributes on which you do not have
a widget, with a bit more config.

Let me know what you think

image

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -1031,6 +1031,72 @@ instantsearch.widgets.stats(options);
 
 <div id="stats-container" class="widget-container"></div>
 
+## FAQ
+
+### Default filters
+
+<div class="codebox-combo">
+<div class="code-box">
+  <div class="code-sample-snippet ignore">
+{% highlight javascript %}
+var search = instantsearch({
+  [...],
+  searchParameters: {
+    facetsRefinements: {
+      categories: ['Cell Phones'],
+      isPremium: [true]
+    },
+    disjunctiveFacetsRefinements: {
+      brand: ['Samsung', 'Apple']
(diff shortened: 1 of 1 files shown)


olance:
👍 👍

bobylito:
Looks good to me too :) Thanks @pixelastic :)

redox:
Sorry guys, I'm late on that.

I really think it's not a good idea to expose the `facetRefinements` and `disjunctiveRefinements` in the `searchParameters` options. Those are super internal values of the underlying AlgoliaSearch Helper and it's way too confusing for our users to deal with them mixed with the actual search parameters. (I'm not even speaking about hierarchical refinements & numerical refinements.)

The simple way to add initial refinements is to build a custom widget:

```javascript
      var search = instantsearch({
        appId: 'YourApplicationID',
        apiKey: 'YourSearchOnlyAPIKey',
        indexName: 'YourIndexName'
      });

      search.addWidget({
        init: function(args) {
          args.helper. addFacetRefinement('category', 'phone');
        }
      });
```

It's documented, I don't need to think about internal naming & I can manipulate the state super easily.

What do you think?

bobylito:
I agree @redox and we should probably add a core widget for that so it's easier to reference. Created an issue for that #1031

olance:
It's true that it removes complexity from the user side.

PR #1029, opened 2016-05-24
docs(readme): Adding animated demo on the readme

pixelastic:
I've added an animated gif of what exactly is a full page "instant search". An image is worth a thousand words, so I feel it will be more compelling.

image


Shipow:
It's a bit confusing to have it next to the website link.
Maybe it just needs a title.

redox:
> It's a bit confusing to have it next to the website link.
> Maybe it just needs a title.

What about putting it below the **Quick Start** section?

pixelastic:
I've added a title but I'd really like to keep it at the top of the readme, so people can see and understand what it does really quickly.

image

vvo:
Why not show a demo of the examples part of the website (like the furnitures one) and link to that example page.

Also add a small text like: "Here's an example of what can be built easily using instantsearch.js and Algolia:" then after the image "Find this example on our website at http://...examples/..." so that we are transparent

pixelastic:
Good point. I wanted to use the examples of the website at first, but I was struggling with doing an animated gif with nice enough colors (not dithered), and then I realized there is this gif we already use in several tutorials and slides we show at events. So I reused the same to stay consistent.

So instead of using an animated gif from the example page, why don't we add this example to the `/examples` page. We even have the full guide available on [instant search](https://www.algolia.com/doc/search/instant-search) demo. 

That's another point, but I would also like to add more external examples/demo to that page like [marvel](http://pixelastic.github.io/marvel/), [steamdb](https://steamdb.info/instantsearch/) or [searchstone](http://searchstone.io/).

pixelastic:
And I've added some context as per your suggestion:

image

vvo:
WFM

PR #1034, opened 2016-05-25
chore(deps): upgrade all deps, regenerate shrinkwrap

vvo:


pixelastic:
I think you removed the fixed header in that commit as well. Was it on purpose?

I re-added it (and extended it to all pages) in another PR.

pixelastic:
I think you removed the fixed header in that commit as well. Was it on purpose?

I re-added it (and extended it to all pages) in another PR.

vvo:
Indeed, "url_documentation_" seems to be added by something else eheh

PR #1036, opened 2016-05-25
docs(header): Fix the header when scrolling

pixelastic:
I've had feedbacks that when discovering instantsearch.js for the
first time, people scroll down on the home page and once they are at
the bottom they don't know where to go because the menu disappeared.
They thought it was just a promo page, with no more content, so they
closed it.

I've set the header as `position: fixed` like on the documentation
page. Seems like a recent commit did remove the fixed header by
mistake.

| Before | After |
|---------- | --------- |
| 2016-05-25_15h33m39 | 2016-05-25_15h36m18 |

cc @Shipow

### DIFF
--- a/docs/_assets/stylesheets/_header.scss
+++ b/docs/_assets/stylesheets/_header.scss
@@ -3,6 +3,8 @@
 
 .site-header {
   width: 100%;
+  position: fixed;
+  z-index: 999;
 
   .navbar {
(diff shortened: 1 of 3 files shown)


vvo:
Oh wow it was fixed for sure at some point

vvo:
will just re-add what was removed in my previous PR

PR #1037, opened 2016-05-26
fix(lite): use lite algoliasearch build (js client)

### LINKED ISSUE #1024: Usage in meteor
I am having this error (Chrome 50.0.2661.102 / Meteor 1.3). Any idea ?

screen shot 2016-05-22 at 11 33 46 pm

vvo:
fixes #1024 by forcing to load /lite and avoid bad resolve of the
browser field.

This could break if people were using search.client to do things like
indexing. But we do document this as an API entry. We only have an
infinite scroll tutorial using the client passed to the helper. Thus I
had to make browse and browseFrom part of the lite build.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -5,8 +5,8 @@
   "node-version": "v5.11.0",
   "dependencies": {
     "algoliasearch": {
-      "version": "3.14.4",
-      "resolved": "https://registry.npmjs.org/algoliasearch/-/algoliasearch-3.14.4.tgz",
+      "version": "3.14.5",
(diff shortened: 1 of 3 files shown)


pixelastic:
I think we're ok for the indexing part, people should not be using instantsearch to do any indexing. Ok for the browse, I did not know we could do browser with a search-only key. Maybe only with a key with specific ACL?

LGTM

vvo:
> I did not know we could do browser with a search-only key

No you cannot but still it's a "search" operation available with specific keys.

vvo:
As for the indexing part, people can use instantsearch.js (example: back office application). Truth is that we do not expose the client anywhere in an official way. So people willing to do so should include it.

PR #1046, opened 2016-05-31
Docs: makes parameter descriptions collapsible

bobylito:
Proposition to enhance the readability of the parameters in the documentation. All the descriptions are hidden by default, this let the user have a better overview of the parameters, makes the list of parameters less daunting.

Updated based on @Shipow feedback :) (thanks man :))

Demo:

collapsibledesc-demo-3

### DIFF
--- a/docs/_assets/javascripts/doc.js
+++ b/docs/_assets/javascripts/doc.js
@@ -181,6 +181,15 @@
     });
   }
 
+  function foldableDetails() {
+    $('.attr-name').on('click', function() {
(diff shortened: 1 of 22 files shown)


pixelastic:
Love it. I would still allow to open several items at once, though.

vvo:
Let's try this way and if it's not so easy to use it, allow multiple opens

Shipow:
yes! looks nice
 +1 multi-open and you can now get ride of the help cursor I guess

Issue #1047, opened 2016-05-31
Ability to change the Powered By template


Jerskouille:
I'm handling internationalization of the search in Algolia for Zendesk right now and the string "Search by Algolia" is the only one I'm not able to change depending of the language.

I can hack it by passing `poweredBy: false` to the widget and manually adding it myself, but I feel like we could just provide a template to replace the `PoweredBy` component on the SearchBox widget.

pixelastic:
Yep, agreed. Care to do a PR?

pixelastic:
Actually, I'm taking care of that.

pixelastic:
Suggested a PR here: https://github.com/algolia/instantsearch.js/issues/1048

PR #1048, opened 2016-05-31
fix(poweredBy): Let users define their own poweredBy template

pixelastic:
Fixes: #1047

You can now pass an object to `poweredBy` in the `searchBox` widget like this:

```javascript
poweredBy: {
  template: 'Link to {{data.url}}', // Also accepts function
  cssClasses: {
    root: 'applied to wrapper',
    link: 'applied to link'
  }
}
```

Everything is handled directly in the widget, without the need for
a React component. This was overkill because we never update the
render of the `poweredBy` so it was easier to just generate the
template at init time without going through all the process of
creating a template.

I've also added more thorough tests. Users will be able to
independently change CSS classes or template.


pixelastic:
I've used `poweredBy.template`. I wonder if `poweredBy.templates.item` is better?

Jerskouille:
Wow, a PR so fast? :100: 
About the way you defined it, I'm just wondering: wouldn't this be the first time we wrap a template inside a subobject of a widget?

I couldn't find any other case of this, correct me if I'm wrong, but for consistency, it might be better to go with `templates.poweredBy`, and then `cssClasses.poweredBy{Root,Link}` or `cssClasses.poweredBy.{root,link}`, no?

pixelastic:
We did it for the `showMore.templates` as well. And here we even had several templates (`active` / `inactive`).

I remember discussing it with @vvo at some point, and we more or less agreed that using a subobject was better, but we never wrote any standard for that. I'm open for discussing it

Jerskouille:
I have no issues with it, if we're already using this elsewhere, I'm happy with these options. I'm just an advocate of consistency. :)

vvo:
LGTM, even the template part is ok, no need for "item" here

PR #1049, opened 2016-06-01
feat(pagination): Add CSS class denoting the number of pages

### LINKED ISSUE #1039: I can't hide the pagination when there is only one result page.
image

pixelastic:
Fixes #1039

The pagination widget wrapper now has a CSS class of
`ais-pagination__pagecount-42`, if there are 42 possibles pages of
results.

image

I've used enzyme for the tests, and refactored a bit an unrelated test
file (just moved the duplicate `beforeEach`/`afterEach` one level
higher).

Two issues:

- The class name is not configurable. Do you think it should be? and
  if so, what name would you suggest for the `cssClasses` option.
- It uses the `nbPages`, not the actual number of pages displayed in
  the UI. Do you think it's an issue?

### DIFF
--- a/src/components/Pagination/Pagination.js
+++ b/src/components/Pagination/Pagination.js
@@ -1,12 +1,16 @@
 import React from 'react';
 import forEach from 'lodash/collection/forEach';
 import defaultsDeep from 'lodash/object/defaultsDeep';
-import {isSpecialClick} from '../../lib/utils.js';
-
+import {
+  bemHelper,
+  isSpecialClick
+} from '../../lib/utils.js';
 import Paginator from './Paginator.js';
 import PaginationLink from './PaginationLink.js';
 
 import cx from 'classnames';
+
+let bem = bemHelper('ais-pagination');
 
 class Pagination extends React.Component {
   constructor(props) {
@@ -123,8 +127,14 @@ class Pagination extends React.Component {
 
     let createURL = this.props.createURL;
 
+    // Mark the wrapper with a class denoting the number of pages
+    let cssClassRoot = cx(
+      this.props.cssClasses.root,
+      bem(null, `pagecount-${this.props.nbPages}`)
+    );
+
     return (
-      <ul className={this.props.cssClasses.root}>
+      <ul className={cssClassRoot}>
         {this.props.showFirstLast ? this.firstPageLink(pager, createURL) : null}
         {this.previousPageLink(pager, createURL)}
         {this.pages(pager, createURL)}
--- a/src/components/Pagination/__tests__/Pagination-test.js
+++ b/src/components/Pagination/__tests__/Pagination-test.js
@@ -2,6 +2,7 @@
 
 import React from 'react';
 import expect from 'expect';
+import {shallow} from 'enzyme';
 import sinon from 'sinon';
 import TestUtils from 'react-addons-test-utils';
 import Pagination from '../Pagination';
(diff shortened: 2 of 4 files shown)


Shipow:
I would only go with one wrapping class  **useless**
I think there is no need to adapt the pagination regarding the number of pages, but if there is, it should use the UI number.

vvo:
One way to solve this while not creating another API entry for a case that is a UX choice to be made by people:

```js
var hidePaginationWhenNoResults = {
  render: function(data) {
    if (data.results.nbPages === 1) {
      hidePaginationContainer();
    }
  }
}
```

Shipow:
interesting, but a bit complex regarding the solution where you would just have to display:none on the state.

vvo:
> interesting, but a bit complex regarding the solution where you would just have to display:none on the state.

Yes but we should aim at reducing our API entries not adding more, because when you have the number of pages as a CSS property on the pagination widget, why not having it also on the refinementList widgets?

Shipow:
you don't need an api entry for that. just add a class when only 1 page and the rest can be done in css.

pixelastic:
Actually, we have another open issue with regard to the CSS classes of the refinement list and the number of active elements. The fact that we might need it on other widgets does not change anything about the fact that it is needed here.

I don't want to add a `.has-only-one-page` class, it's too specific. I'll stick with the `pagecount`, but using the current number of displayed pages instead.

vvo:
I really don't think we need that today, the small snippet I put solves this without us having to add more API. The need to "hide" the pagination here is only valid for our use case. As a user I prefer to know when there's only one page, while if you hide it I will just think "Maybe there are more pages, but I cannot tell"

pixelastic:
Having to create a whole new widget (even a simple one) seems way more complex than simply styling a CSS class. I don't see adding this new CSS as a new API for instantsearch, it in the same vein as all the other `__active` classes we are adding to all our components. Just denoting the current state through CSS classes so users can style their widgets accordingly.

vvo:
Everything we add to the library, be it JavaScript option or a cssClass in the default templates is an API addition. Once it's there, we cannot remove it.

Since the need was only expressed on one demo, I don't see the value of adding another API entry to hide the pagination, really no trying to dismiss the work you did. Just it doesn't seem to be that of a need that we need a new API entry.

Shipow:
I don't agree Vincent, not being able to hide a useless widget is a weakness, we really don't need a user to tell us that.

vvo:
I gave 7 lines of JS that solves this (hiding the widget).

If we keep adding options like that this has no end.

Problem: when there's a single page I want to hide the pagination widget
Solution: add a new css class with the pages count

Problem: when the user selects a specific facet value I want to hide the refinement list
Solution: add a new css class the the facet value currently selected

This do not scale. The particular use case we faced does not need to be adressed with a new API entry, we need to stop adding options like that.

What we are doing here is take data from the results.* and put it as API entries inside the markup through html structure (here, css classes). There's no added value from taking all the results.* data and putting it everywhere in the API markup level.

I won't add that just because it's easy to add and we can.

Shipow:
I won't fight here pagination is dead anyway

Shipow:
But maybe you could imagine somehow a generic behavior on such components that creates few semantic class regarding their state.

pixelastic:
> Problem: when the user selects a specific facet value I want to hide the refinement list

This does not feel like a real world example. The need of hiding the pagination is real. Even if it was only requested by us internally, we never waited for users to tell us something was broken to fix it if we spotted it.

The CSS classes we added to our widgets already helped us a lot in adding UI features. I really don't see how this is different from the `active` classes we add to our refinement list. We could have stayed with a checkbox that is ticked or not, but adding a wrapping CSS class let users do much more than this. CSS classes are the only way currently that users can interact with the internal state of the components.

I think that the problem is more generic than this issue only. We need to have a way for users to change the display based on the widget/component state. Either we expose it through `render` events, or through CSS classes or any other methods is up to us. But I feel like whatever the method we chose, we will still increase the API surface for any new part of the state we expose, and have to document it.

And as we already started exposing state through CSS in this v1, we should keep doing it and think about a better way for a v2.

Shipow:
yes problem is generic and could maybe summarize by: "we should have enough granularity to target any elements and states with css selectors"

vvo:
I am closing this for now, there's an easy solution to solve it in https://github.com/algolia/instantsearch.js/pull/1049#issuecomment-222927757

timkelty:
Interestingly, with @vvo's [solution](https://github.com/algolia/instantsearch.js/pull/1049#issuecomment-222927757), if you try and `display: none` the container (in `hidePaginationContainer()`), it wont work. Presumably because something else (`autoHideContainer`?) is resetting the display prop after.

PR #1051, opened 2016-06-03
feat(header): Pass count of current refined filters in header

### LINKED ISSUE #1013: Add active refinements info for the menu, hierarchicalMenu and the refinementList widgets
Would be great to add more info in the markup about the active. That would help to build UI/UX where the list of those possible filter items are hidden.

We could for instance imagine to have:
- add a class at the top level that there is 1 or more items selected
- add a count value in the `ais-menu--header` to indicate how many items have been selected (+ extra attribute, to get that value easily accessible programmatically)

pasted image at 2016_05_05 23_26

```html
<div data-reactroot="" class="ais-root ais-menu">
  <div class="ais-menu--header ais-header ais-menu__active" data-active-items="1">
    Cuisine <span class="ais-header-active--count">1</span>
  </div>
  <div class="ais-body ais-menu--body">
     ...
```

What do you guys think?

### LINKED ISSUE #1041: The collapsible options on a refinement list breaks the refresh of the widget
bug-collaspe

pixelastic:
Partially fixes #1013

The goal of the PR was to make the `count` variable accessible to the
header template of the `refinementList` widget. This `count` is the
number of active filters applied to the facet.

2016-06-03_15h47m14
_(I click on the header in this demo to show that it works with `collapsible` as well)_

```javascript
        templates: {
          header: (data) => {
            if (data.count === 0) {
              return `<h3 class="facet--title">Teams</h3>`;
            }
            return `<h3 class="facet--title">Teams (${data.count})</h3>`;
          }
        },
```

To do so the first step was to make the `header` template dynamic.
Before the PR, it was static, meaning that it was defined at the
widget init and never updated. I changed that part so the header
template is re-rendered on every render of its component.

It also meant that I had to remove the custom `shouldComponentUpdate`
method. The method was checking if the component was collapsed or not
before triggering the render. As we now always need to update the
render of the header, being collapsed or not, I removed the check.
I think it might even fix #1041 as well.

To pass template data to the header, you now need to pass the
`headerFooterData` prop to the component that is wrapped by
`headerFooterHOC`. It's an object with a key `header` containing the
data you want to pass. It could also potentially have a key
`footer`, but the footer part is not implemented yet because we had no
need for it.

Improvements and discussions:

- The `footer` template is still static. It would be trivial to set it
  as dynamic, but I left it like this until we have a use-case.
- Only the `refinementList` widget takes advantage of this feature. It
  would not be too hard to pass dynamic data to headers in other
  widgets, but I left that to other PRs.
- I will tackle the other issue of adding a CSS class on the root on
  another PR.

### DIFF
--- a/src/decorators/headerFooter.js
+++ b/src/decorators/headerFooter.js
@@ -4,6 +4,7 @@
 import React from 'react';
 
 import cx from 'classnames';
+import getKey from 'lodash/object/get';
 
 import Template from '../components/Template.js';
 
@@ -16,11 +17,6 @@ function headerFooter(ComposedComponent) {
         collapsed: props.collapsible && props.collapsible.collapsed
       };
 
-      this._headerElement = this._getElement({
-        type: 'header',
-        handleClick: props.collapsible ? this.handleHeaderClick : null
-      });
-
       this._cssClasses = {
         root: cx('ais-root', this.props.cssClasses.root),
         body: cx('ais-body', this.props.cssClasses.body)
@@ -28,18 +24,18 @@ function headerFooter(ComposedComponent) {
 
       this._footerElement = this._getElement({type: 'footer'});
     }
-    shouldComponentUpdate(nextProps, nextState) {
-      return nextState.collapsed === false ||
-        nextState !== this.state;
-    }
     _getElement({type, handleClick = null}) {
       let templates = this.props.templateProps.templates;
       if (!templates || !templates[type]) {
         return null;
       }
       let className = cx(this.props.cssClasses[type], `ais-${type}`);
+
+      let templateData = getKey(this.props, `headerFooterData.${type}`);
+
       return (
         <Template {...this.props.templateProps}
+          data={templateData}
           rootProps={{className, onClick: handleClick}}
           templateKey={type}
           transformData={null}
@@ -67,9 +63,14 @@ function headerFooter(ComposedComponent) {
         root: cx(rootCssClasses)
       };
 
+      let headerElement = this._getElement({
+        type: 'header',
+        handleClick: this.props.collapsible ? this.handleHeaderClick : null
+      });
+
       return (
         <div className={cssClasses.root}>
-          {this._headerElement}
+          {headerElement}
           <div
             className={cssClasses.body}
           >
--- a/src/decorators/__tests__/headerFooter-test.js
+++ b/src/decorators/__tests__/headerFooter-test.js
@@ -2,6 +2,7 @@
 
 import React from 'react';
 import expect from 'expect';
+import {shallow} from 'enzyme';
 import TestUtils from 'react-addons-test-utils';
 import TestComponent from './TestComponent';
 import headerFooter from '../headerFooter';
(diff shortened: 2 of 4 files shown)


vvo:
> The method was checking if the component was collapsed or not
> before triggering the render

Some websites are using collapsing widgets to be able to have a lot of widgets without messing up the display. Thus always rendering widgets was not efficient since a lot of them were collapsed.

> As we now always need to update the
> render of the header, being collapsed or not

We only need to update it when the active count has changed right? We should find a way to not render the widget if no values were updated. Unsure that's doable.

vvo:
Do you have other naming ideas for "count"? Maybe "activeFacetsValues" or smg adding context to what is the count about.

src/decorators/headerFooter.js
```diff
@@ -16,30 +17,25 @@ function headerFooter(ComposedComponent) {
...
-
       this._cssClasses = {
         root: cx('ais-root', this.props.cssClasses.root),
         body: cx('ais-body', this.props.cssClasses.body)
       };
 
       this._footerElement = this._getElement({type: 'footer'});
     }
-    shouldComponentUpdate(nextProps, nextState) {
-      return nextState.collapsed === false ||
```
vvo: We can keep that shouldComponentUpdate but we should do a deep comparison on the passed data to avoid re-rendering

vvo: This is maybe overkill because a lot of data can be passed to the widget.

pixelastic: Yes, I wonder if we are not over-engineering the `shouldComponentUpdate` of our components. I'm not really familiar with it but I would have thought that React was already doing the deep comaprison check?

pixelastic:
> Some websites are using collapsing widgets to be able to have a lot of widgets without messing up the display. Thus always rendering widgets was not efficient since a lot of them were collapsed.

Maybe we should not use the `shouldComponentUpdate` for that then. Maybe just not render the `ComposedComponent` when it is collapsed.

pixelastic:
> Do you have other naming ideas for "count"? Maybe "activeFacetsValues" or smg adding context to what is the count about.

I had the same question, but I settled for `count` because this is what we use in the `RefinementListItem`, but I'm ok changing. Especially considering this is the number of currently selected facets, not the count of hits for thoses facets. So maybe `refinedFacetsCount`? We already use `isRefined` in the `RefinementList` template.

src/decorators/headerFooter.js
pixelastic: Maybe we need a `Header` and `Footer` components, with their own `shouldComponentUpdate` methods?

I feel like our `shouldComponentUpdate`s are too weak at the moment and are not worth the complexity cost. Does it really bring any perf improvement? I don't really know that part but from reading [this doc](https://facebook.github.io/react/docs/advanced-performance.html) it seems that it only really make sense with immutability in mind (which we only have in the helper).

vvo: > React was already doing the deep comparison check of the passed data?

No React does not do a deep props comparison out of the box, neither a shallow one (===).

vvo:
> So maybe refinedFacetsCount?

This seems better yes!

vvo:
@pixelastic small changes needed before merging

vvo:
Merging, will do the small changes afterwards

PR #1052, opened 2016-06-04
feat(hits): Add a `__position` attribute to data passed to hit items

### LINKED ISSUE #903: Passing the item position to `transformData`?
For the latest integration I did, I had to apply a specific markup to the 4 first displayed hits. In my case it was _not_ applying lazyloading of images to those elements.

So I needed to know, from inside the `transformData` method, the current position of the item in the returned set.

There is no easy way to do that at the moment, so I resorted to a hack involving setting a global counter (reseted on every `on('render')`), and increasing it on each call to my `hits.transformData`. It worked, but was far from optimal (code is separated in several parts, I have no clear guarantee that the `transformData` are actually executed in order), etc).

So I'm wondering if we shouldn't change the `transformData` signature to `transformData(item, position)` instead of just `transformData(item)`.

WDYT?

pixelastic:
Fixes #903

The change itself was easy to do, but it broke all the tests because
we were testing a full JSX output and we now had one more key passed
to the `data`.

It took this as an opportunity to refactor the Hits tests uzing enzyme
to make them easier to read. It allow me to spot one or two
inconsistencies in the way the `root` class was added or not depending
on the kind of template used.

### DIFF
--- a/src/components/Hits.js
+++ b/src/components/Hits.js
@@ -3,6 +3,7 @@ import map from 'lodash/collection/map';
 
 import Template from './Template.js';
 
+import hasKey from 'lodash/object/has';
 import isEqual from 'lodash/lang/isEqual';
 import cx from 'classnames';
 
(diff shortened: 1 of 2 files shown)


pixelastic:
Seems like the Travis build never started... I'll close the PR and start a new one.

Issue #848, opened 2016-02-16
Encoding / Decoding issue


NicoD:
The bracket characters (unsafe) and equal character (reserved) are directly set in the url query string or fragment without encoding.

example:
https://demo.algolia.com/instant-search-demo/?q=&hPP=10&idx=instant_search&p=0&nR[price][>=][0]=2676&is_v=1
should be:
https://demo.algolia.com/instant-search-demo/?q=&hPP=10&idx=instant_search&p=0&nR%5Bprice%5D%5B%3E%3D%5D%5B0%5D=2676&is_v=1
  
(same thing observed with useHash = true)

vvo:
Hi @NicoD this should be handled by the browsers in a natural way: the url looks good , did you observe any bad behavior?

NicoD:
Hi @vvo ,

here some examples with a failure:
- here in github, in my previous comment. When I copied the first url (which is the instantsearch url), github had itself encoded the unsafe characters (brackets) but not the =. The link does not work when following it. (however the second link - which i encoded manually - works fine)
- i had the same issue when reading an instantsearch url through a slack client (not the web client)
 
=> i suppose this error may happened on others clients that encode the query string: the brackets are normaly url urlencoded, but not the = which is considered as a separator.

- js url parsing lib (like urljs and browserify url) fail parsing the query string or fragment because of the = and the non encoded chars.

I think it would be safer to urlencode all the reserved and unsafe chars in the query string and the fragment (https://tools.ietf.org/html/rfc3986)

vvo:
http://localhost:8080/?q=&hPP=24&idx=instant_search&p=0&nR[popularity][%3E=][0]=0&is_v=1

vvo:
Test: http://127.0.0.1:8080/?q=&hPP=24&idx=instant_search&p=0&nR[popularity][%3E=][0]=0&is_v=1

vvo:
@NicoD As for copy pasting the url from the urlbar or right click a link and copy from the context menu then paste it anywhere => it's the responsibility of the client to do things well. Github does, slack also it seems

On this issue, I was able to fix it by decoding the uri every time its needed (loading the page, back button): decodeURIComponent.

Seems to work well, let me know. The fix is incoming

vvo:
This is not as easy as I thought, decoding when receiving has his owns quirks: values (facet refinements) with & or = are decoded and then mis interpreted as separators.

vvo:
[...]

PR #1104, opened 2016-06-22
fix(urlSync): urls should be safe by default

### LINKED ISSUE #982: Urls should be safe by default
Following https://github.com/algolia/instantsearch.js/issues/848#issuecomment-191795411 and a support request we had also recently, I believe we should fix our urls.

We will get more and more issues about unsafe urls in the future.

## Problem :trollface: 

we implemented urls so that they do not look too bad:
`?query=&hits=24&idx=instant_search&p=0&nR[price][<=][0]=320&is_v=1`

But this is not really standard because most parameters should be encoded so that our urls are compatible with every use case (copy pasting from/to email/app/slack, server side parsing, library parsing..)

Thus it would mean having urls like:

`?query=&hits=24&idx=instant_search&p=0&nR%5Bprice%5D%5B%3C%3D%5D%5B0%5D=380&is_v=1`

Not so clean and readable but would work everywhere.

## Possible solutions :horse: 

- allow the helper to always encode. It's currently [disabled](https://github.com/algolia/algoliasearch-helper-js/blob/4357cf95d7efb0c1d017a0c897093289bf31c266/src/url.js#L157) and [we specifically asked](https://github.com/ljharb/qs/issues/116) for this feature in the `qs` project (lol)

This even seem to be backward compatible. It would be a WTF moment for developers but safety first!

- create a brand new stringifier for complex objects => url which would not be based on unsafe characters like [ ] > =

Here we should be careful about URL versions if we want to stay backward compatible (is_v= will help)

- use https://github.com/vjeux/URLON

Help and feedback appreciated.

vvo:
tl;dr; Urls stored in the browser bar were made to look not too bad by omitting to encode some characters. Turns out it was a bad idea, copy pasting the url from browser bar to slack or github failed.

Fixes #982 #848 by using the new safe: true option from the helper when asking for a state as a url (https://github.com/algolia/algoliasearch-helper-js/pull/317).

This seems to be backward compatible (instantsearch wise):
- for current implementations, urls in the browser bar will move from `&hRF[color]` to `&hRF%05Bcolor%05D`
- urls sent to the server (if it happen people were reading them) will move from `&hRF[color]` to `&hRF%05Bcolor%05D` but I believe most server side code is already using decodeURI
- no change to current urls, they will be properly decoded (no change was needed on the decoding part)

If you have any idea how this might break people's website, please add a comment.

cc @olance

### DIFF
--- a/src/lib/url-sync.js
+++ b/src/lib/url-sync.js
@@ -134,7 +134,8 @@ class URLSync {
       state.filter(this.trackedParameters),
       {
         moreAttributes: foreignConfig,
-        mapping: this.mapping
+        mapping: this.mapping,
+        safe: true
       }
     );
 


bobylito:
No config for this behaviour?

vvo:
> No config for this behaviour?

I first started doing it but then thought "well, the current urls are broken, do we want to allow developers to have `broken` urls?"

bobylito:
Lgtm :) :+1:

PR #1146, opened 2016-07-22
feat(toggle): Provide a better default widget

### LINKED ISSUE #919: Provide both toggle states count
At the moment, we only provide the count for the current state of the toggle. A good idea from @davidpike7 is to also provide the count for the other state of the toggle.

So that you can show things like "Only show free shipping (20)" 20 being the number of free shipping items. It's easier to understand than "Only show free shpping (340)" Where 340 is the actual number of items not having free shipping.

### LINKED ISSUE #1096: Toggle widget displaying 0 for count when there are results
We are using toggle widget to check a numeric boolean value (can either be 1 or 0). It is setup like so:

```js
this.instantsearch.addWidget(
    instantsearch.widgets.toggle({
        container: '#journeys-specials-widget',
        attributeName: 'hasSpecialOffer',
        label: 'Only show specials',
        values: {
            on: 1,
        },
        templates: {
            item: refinementItem
        },
    })
);
```

The filter is working as intended but is displaying an incorrect value of 0 next to the toggle as there are 40 results displayed when toggled.

I feel like this has something to do with the off value not working correctly. If I set it to `0` the refinement is activated by default which isn't what we want.

The toggle was working previously when the values were `true` or `false` but our data changed and it stopped working.

vvo:
This update brings:
- default to showing the next user action items count instead of the
current number of displayed hits

which means:
if checkbox is not checked, show: [ ] free shipping (countWhenChecked)
if checkbox is checked, show: [x] free shipping (countWhenNotChecked)
- fix numeric based on and off values (fixes #1096)
- provide `onFacetValue` and `offFacetValue` template data, fixes #919

### DIFF
--- a/src/widgets/toggle/__tests__/toggle-test.js
+++ b/src/widgets/toggle/__tests__/toggle-test.js
@@ -64,7 +64,7 @@ describe('toggle()', () => {
     });
 
     it('configures hitsPerPage', () => {
-      expect(widget.getConfiguration()).toEqual({facets: ['world!']});
+      expect(widget.getConfiguration()).toEqual({disjunctiveFacets: ['world!']});
     });
 
     it('uses autoHideContainer() and headerFooter()', () => {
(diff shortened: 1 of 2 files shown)


vvo:
Hi @davidpike7, I just did some more fixes on the toggle widget, can you this build on your website: https://gist.githubusercontent.com/vvo/19d16de5d738ab55eaa3a9e1e3c7b67b/raw/71ff01360b6591186142b1457d5add5f40f746a3/instantsearch.min.js

Here's what's new:

    feat(toggle): Provide a better default widget

    This update brings:
    - default to showing the next user action items count instead of the
    current number of displayed hits

    which means:
    if checkbox is not checked, show: [ ] free shipping (countWhenChecked)
    if checkbox is checked, show: [x] free shipping (countWhenNotChecked)
    - fix numeric based on and off values (fixes #1096)
    - provide `onFacetValue` and `offFacetValue` template data, fixes #919

PR #1148, opened 2016-07-25
fix(searchParameters): avoid mutating provided objects

### LINKED ISSUE #1130: CurrentRefinedValues widget bug?
When you visit [URL1] first and afterwards [URL2] the current refined values aren't shown anymore and the sort by widget has an issue as well.

I get this error in the console:
index.js:959Uncaught Error: merkFull is not defined in the facets attribute of the helper configuration

Which is true, I haven't defined it in searchParameters.facets, but when I do I can't select other values in this refinementList. Shouldn't the same code that gathers the values from the url define the facets in the helper or so? Or am I doing something wrong?

Both URL's are non public test sites. You can find the url's at http://codepen.io/Maaark123/pen/GqvBqb

vvo:
Fixes #1130

Thanks a lot @morhaus for your help when debugging

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -6,6 +6,7 @@ import forEach from 'lodash/collection/forEach';
 import merge from 'lodash/object/merge';
 import union from 'lodash/array/union';
 import clone from 'lodash/lang/clone';
+import isObject from 'lodash/lang/isObject';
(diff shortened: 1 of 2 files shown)


vvo:
91aa70ea5d393aec34f8aad3982598b3

bobylito:
The hardest bug have always the smallest solution :) LGTM 👍

PR #1154, opened 2016-07-28
fix(toggle): add backward compatibility for previous toggle implem

vvo:
Context: I upgrade the default toggle behavior in #1146.
To do this, I switched from using a facet to a disjunctive facet. Which
led to inconsistencies with the already public instantsearch websites urls using
facetRefinements for toggle.

So now, if there's a matching facetRefinement for the attributeName used
in the toggle, I will switch to a legacy implementation.

Otherwise, I will switch to the current implementation.

I wrote some tests around this but this is very messy. Our whole test
stack has evolved in a monster, mostly because of the way we are writing
our own code.

This should get better with instantsearch-react and so I did not spend
my whole day on this. Tested this multiple times: it works.

### DIFF
--- a/scripts/mocha.opts
+++ b/scripts/mocha.opts
@@ -1,3 +1,2 @@
 --compilers js:babel-register
---require jsdom-global/register
-@(src)/**/__tests__/**/*-test.js
+@(src)/**/__tests__/**/*-test.js scripts/jsdom.js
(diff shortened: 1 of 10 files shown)


scripts/mocha.opts
```diff
@@ -1,3 +1,2 @@
 --compilers js:babel-register
---require jsdom-global/register
```
vvo: Discovered that this led to not having the html page of jsdom cleaned between tests. Ugh

PR #1165, opened 2016-08-02
Big refactoring/simplification pass

Morhaus:
Refactor RefinementList/Menu/Hierarchical/etc. to use a single List component
and move the show more logic there.

Add themeable and translatable HOCs to reduce boilerplate and prepare for
future work.

Remove PaginationLink, prefer single component rendering.

Add defaultProps option to createHOC's output in order to further parameterize
custom HOCs and move default props configuration to widgets.


packages/instantsearch-react/src/translatable.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+
+export default function translatable(defaultTranslations) {
+  return Composed => {
+    function Translatable(props) {
+      const {translations, ...otherProps} = props;
+      const translate = (key, ...params) => {
+        const translation =
+          translations && {}.hasOwnProperty.call(translations, key) ?
```
bobylito: why not `translations.hasOwnProperty(key)`?

Morhaus: https://github.com/eslint/eslint/issues/2693

Javascript ✨

packages/instantsearch-react/src/translatable.js
```diff
@@ -0,0 +1,40 @@
...
+            translations[key] :
+            defaultTranslations[key];
+        if (typeof translation === 'function') {
+          return translation(...params);
```
bobylito: This should probably be moved up before `const translation...`

Morhaus: Translations are defined as either a simple string `Search your website` or a function `pageNo => 'Page  ' + pageNo`. Here, we're testing whether the translation we've just retrieved on the above line is a function in order to apply it additional parameters.

packages/instantsearch-react/src/translatable.js
bobylito: Well for all the stupid stuff anyone can do with javascript, I wouldn't write that this way. What is the use case of `Object.create(null)`? Will it apply in our case? If it's not likely, I wouldn't use that... Sanity of the code first because we'll have to maintain it.

packages/instantsearch-react/src/translatable.js
bobylito: If only I had read that correctly in the first place :/

packages/instantsearch-react/src/translatable.js
Morhaus: I don't mind either way, but if we do choose to use object prototypes builtins directly in our projects, we should update the [algolia eslint config](https://github.com/algolia/eslint-config-algolia/blob/master/index.js#L50).

packages/instantsearch-react/src/impl/List.js
```diff
@@ -0,0 +1,115 @@
...
+          item.value,
+          'item',
+          selected && 'itemSelected',
+          children && 'itemParent',
+          hasSelectedChild(item, selectedItems) && 'itemSelectedParent'
+        )}
+      >
+        {this.props.renderItem(item, selected)}
+        {children &&
+          <div {...applyTheme('itemChildren', 'itemChildren')}>
```
bobylito: What moving this piece of logic out in a variable and just add the variable here?

packages/instantsearch-react/src/themeable.js
```diff
@@ -0,0 +1,32 @@
...
+    function Themeable(props) {
+      const {theme = defaultTheme, ...otherProps} = props;
+      const applyTheme = reactThemeable(theme);
+
+      return <Composed {...otherProps} applyTheme={applyTheme} />;
+    }
+
+    Themeable.propTypes = {
+      theme(props, propName, componentName) {
+        const prop = props[propName];
```
bobylito: What about moving that prop type checker to another file? Or at least another function to be able to keep the same proptypes structure?

packages/instantsearch-react/src/createHOC.js
```diff
@@ -1,27 +1,26 @@
 import React, {Component} from 'react';
 import {connect} from 'react-algoliasearch-helper';
 
-import {applyDefaultProps} from './utils';
 import {stateManagerPropType, configManagerPropType} from './propTypes';
 
 const hasOwnProperty = (...args) => ({}.hasOwnProperty.call(...args));
```
bobylito: Ah mais ça ça vaut peut être le coup de le mettre dans un fichier et d'utiliser ça. Quitte à utiliser un pattern moche, on peut peut être le cacher :)

packages/instantsearch-react/src/createHOC.js
```diff
@@ -1,27 +1,26 @@
...
 
 const ORIGINAL_PROPS_KEY = '__original';
 
 export default function createHOC(desc) {
```
bobylito: `abstractConnectorDesc`?

packages/instantsearch-react/src/createHOC.js
```diff
@@ -1,27 +1,26 @@
...
-  return (mapPropsToConfig = a => a) => {
+  return (desc2 = {}) => {
```
bobylito: `concreteConnectorDesc`?

packages/instantsearch-react/src/translatable.js
bobylito: Oh I wasn't aware of that... Nevermind then :)

packages/instantsearch-react/src/themeable.js
Morhaus: Do you mean something like `theme: themePropType(defaultTheme)`?

packages/instantsearch-react/src/createHOC.js
Morhaus: That, or use lodash `has`.

packages/instantsearch-react/src/createHOC.js
Morhaus: Or maybe `adapterDesc`, since the primary function of this parameter is to adapt the connector to a particular component's interface?

packages/instantsearch-react/src/createHOC.js
bobylito: Oh that was the name of the function... Couldn't find it earlier :) 👍

packages/instantsearch-react/src/themeable.js
bobylito: Yep!

bobylito:
LGTM! (there is still one comment open ;))

bobylito:
Boom let's merge this bad boy!! 💃

PR #1169, opened 2016-08-03
Range implementation

Morhaus:
The connector should be feature complete, however the implementation is subject to change. Notably, rc-slider doesn't support our theming options, so we might end up having to create our own slider.

### DIFF
--- a/packages/instantsearch-react/src/impl/Range.js
+++ b/packages/instantsearch-react/src/impl/Range.js
@@ -1,0 +1,50 @@
+import React, {PropTypes, Component} from 'react';
+import Slider from 'rc-slider';
+
+// @TODO: rc-slider doesn't support our customization options relative to
+// theming. We'll need to implement our own slider down the road.
+export default class Range extends Component {
+  static propTypes = {
+    min: PropTypes.number.isRequired,
+    max: PropTypes.number.isRequired,
+    value: PropTypes.arrayOf(PropTypes.number).isRequired,
+    refine: PropTypes.func.isRequired,
+  };
+
+  constructor() {
+    super();
+
+    this.state = {
+      controlled: false,
+      value: null,
+    };
+  }
+
+  onChange = value => {
+    this.setState({
+      controlled: true,
+      value,
+    });
+  };
+
+  onAfterChange = () => {
+    this.props.refine(this.state.value);
+    this.setState({
+      controlled: false,
+      value: null,
+    });
+  };
+
+  render() {
+    return (
+      <Slider
+        min={this.props.min}
+        max={this.props.max}
+        range
+        value={this.state.controlled ? this.state.value : this.props.value}
+        onChange={this.onChange}
+        onAfterChange={this.onAfterChange}
+      />
+    );
+  }
+}
--- a/packages/instantsearch-react/src/utils.js
+++ b/packages/instantsearch-react/src/utils.js
@@ -12,3 +12,20 @@ export function isSpecialClick(event) {
 export function capitalize(key) {
   return key.length === 0 ? '' : `${key[0].toUpperCase()}${key.slice(1)}`;
 }
+
+export function assertFacetDefined(searchParameters, searchResults, facet) {
+  const wasRequested =
+    searchParameters.isConjunctiveFacet(facet) ||
+    searchParameters.isDisjunctiveFacet(facet);
+  const wasReceived =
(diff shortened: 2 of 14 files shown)


bobylito:
> The connector should be feature complete

What do you mean feature-complete?

Morhaus:
@bobylito It works \o/ It retrieves the correct data and feeds it to the underlying component implementation.

packages/instantsearch-react/src/utils.js
```diff
@@ -12,3 +12,20 @@ export function isSpecialClick(event) {
 export function capitalize(key) {
   return key.length === 0 ? '' : `${key[0].toUpperCase()}${key.slice(1)}`;
 }
+
+export function assertFacetDefined(searchParameters, searchResults, facet) {
+  const wasRequested =
+    searchParameters.isConjunctiveFacet(facet) ||
+    searchParameters.isDisjunctiveFacet(facet);
```
bobylito: What about hierarchical facets?

packages/instantsearch-react/src/utils.js
```diff
@@ -12,3 +12,20 @@ export function isSpecialClick(event) {
...
+  const wasReceived =
+    Boolean(searchResults.getFacetByName(facet));
+  if (searchResults.nbHits > 0 && wasRequested && !wasReceived) {
```
bobylito: Why only when nbHits > 0?

packages/instantsearch-react/src/impl/Range.js
```diff
@@ -0,0 +1,50 @@
+import React, {PropTypes, Component} from 'react';
+import Slider from 'rc-slider';
+
+// @TODO: rc-slider doesn't support our customization options relative to
+// theming. We'll need to implement our own slider down the road.
+export default class Range extends Component {
+  static propTypes = {
+    min: PropTypes.number.isRequired,
+    max: PropTypes.number.isRequired,
+    value: PropTypes.arrayOf(PropTypes.number).isRequired,
```
bobylito: `values` instead of `value`?

Morhaus: I like the singular `value`. `values` implies that all items inside have the same role, while in reality the format of the prop is `[min, max]`.

packages/instantsearch-react/src/utils.js
Morhaus: I'm not checking that hierarchical facets are defined yet. I don't think it would work quite the same way since the name of a hierarchical facet does not map to a facet.

packages/instantsearch-react/src/utils.js
Morhaus: When `nbHits == 0`, the API won't send any facets, so `wasReceived` is always false.

packages/instantsearch-react/src/utils.js
bobylito: Well the name of the facet used when defining a hierarchical facet is the one used throughout the code. We usually don't use the real facet name. And it seems that's what you would be checking here.

packages/instantsearch-react/src/utils.js
bobylito: Yep makes sense! Thanks :)

packages/instantsearch-react/src/impl/Range.js
bobylito: Ok so it's the values of the range? The actual values, not the boundaries (which are min and max)?

Morhaus: Yep!
min====value[0]========value[1]====max

packages/instantsearch-react/src/utils.js
Morhaus: I'm not using this function in connectHierarchicalMenu yet, so I haven't updated it for that purpose. I think if we wanted to check that a hierarchical facet is correctly defined on the server side, we'd need to test every of its `attributes` properties on the response?

packages/instantsearch-react/src/impl/Range.js
bobylito: Ok from the explanation I had this morning, it makes sense to call it value since it is sort of standard, right?

Morhaus: Yep.

bobylito:
👍 LGTM

Issue #1170, opened 2016-08-03
Individual widget state


Morhaus:
Right now, both instantsearch.js and instantsearch-react keep a shared helper state, that all components can connect to and refine.

However, this means that we cannot retrieve which user actions led to a given state. This is an issue for implementing the "Current Filters" widget (previously "currentRefinedValues"), for it won't display user friendly filter messages. Take a look at [this demo](http://codepen.io/team/algolia/pen/akaRyJ) and refine a few parameters. The filters are displayed one by one, as they are defined in the helper's state, and we loose track of which component defined which filter.

![image](https://cloud.githubusercontent.com/assets/1621758/17363068/b0f8c432-5978-11e6-951c-abd2eb1e61c8.png)
 
On the other hand, it looks like SearchKit has per-component state. Take a look at [their demo](http://demo.searchkit.co/imdb?categories[0][0]=Movie&categories[1][0]=Drama&metascore[min]=27&metascore[max]=79&countries[0]=USA&runtimeMinutes[0]=61_1000): filters are linked to their widgets. The length filter even displays the label that was defined on the component.

![image](https://cloud.githubusercontent.com/assets/1621758/17363264/a28ff2de-5979-11e6-88b4-ac44aafec35b.png)

If you look at the URL, you'll see that it is not the serialization of a shared state, but the serialization of each component's state separately.

In order for us to have a similar UX, we'd need to drop the shared helper's state and instead keep track of each component's state separately.

I believe that the current architecture can be modified to achieve this: the refine method of a connector's descriptor already takes in a "component state" argument and applies it to the shared state.

Morhaus:
I made a POC of this in [instantsearch-react-individual-states](https://github.com/algolia/instantsearch.js/tree/instantsearch-react-individual-states).

Morhaus:
I ended up implementing this approach, and it incidentally brought a few nice features. Among those, the ability to easily define default state (values) for components, which has been requested for a while. We also have much better state management, which gives our users an easy way to hook into the state of their instantsearch components. This is very useful for routing and/or when using redux to manage your application's state.

Also, readable current filters 🎉
![image](https://cloud.githubusercontent.com/assets/1621758/17478760/544c6384-5d6e-11e6-9d86-25cab319ef6d.png)

PR #1179, opened 2016-08-08
Individual states for widgets

Morhaus:
See #1170.

I also added a few widgets and moved things around a bit more 😇


packages/instantsearch-react/examples/history.js
```diff
@@ -2,7 +2,17 @@ import {useRouterHistory} from 'react-router';
 import {createHistory} from 'history';
 import qs from 'qs';
 
+function alphabeticalSort(a, b) {
```
bobylito: Can [localeCompare](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) do the trick?

packages/instantsearch-react/package.json
```diff
@@ -26,8 +26,10 @@
   },
   "dependencies": {
     "algoliasearch": "^3.16.0",
-    "algoliasearch-helper": "^2.11.1",
+    "algoliasearch-helper": "algolia/algoliasearch-helper-js#develop",
```
bobylito: YOLO :)

packages/instantsearch-react/examples/history.js
Morhaus: Oh wow. Are there cases where we might need the specific behavior of localeCompare? (sorting strings based on locale)

packages/instantsearch-react/package.json
Morhaus: Living on the edge!

packages/instantsearch-react/src/InstantSearch.js
```diff
@@ -1,141 +1,329 @@
...
+    };
+  }
+  return {
+    ...location,
+    search: query ? `?${qs.stringify(query, {sort: alphabeticalSort})}` : '',
+  };
+}
+
+function formatProps(props) {
+  return props.map(prop => `\`${prop}\``).join(', ');
```
bobylito: Is it the same as?
```js
return '`' + props.join('`, `') + '`';
```

Probably less readable :)

packages/instantsearch-react/src/InstantSearch.js
```diff
@@ -1,141 +1,329 @@
...
+      missingProps.push(prop);
+    }
+  }
+  if (presentProps.length !== 0 && missingProps.length !== 0) {
+    const plural = missingProps.length > 1;
+    const missingPropsStr = formatProps(missingProps);
+    const presentPropsStr = formatProps(presentProps);
+    const propName = `prop${plural ? 's' : ''}`;
+    throw new Error(
+      `You must provide ${plural ? '' : 'a '} ${missingPropsStr} ${propName} ` +
```
bobylito: Nice and fancy error message :)

packages/instantsearch-react/src/InstantSearch.js
```diff
@@ -1,141 +1,329 @@
...
+    apiKey: PropTypes.string.isRequired,
+    indexName: PropTypes.string.isRequired,
+
     history: PropTypes.object,
-    createURL: PropTypes.func,
-    treshold: PropTypes.number,
-    configureState: PropTypes.func,
     urlSync: PropTypes.bool,
-    trackedParameters: PropTypes.arrayOf(PropTypes.string),
+    treshold: PropTypes.number,
```
bobylito: Typo: threshold

bobylito:
Good thing I had a preview of what you were doing :D LGTM (check comments though ;) )

packages/instantsearch-react/examples/history.js
bobylito: I don't think so :)

packages/instantsearch-react/src/InstantSearch.js
Morhaus: Yep, it is, although ESLint would want us to do
```
return `\`${props.join('`, `')}\``;
```
Although it's a bit longer, I prefer the map/join version, it has clearer intent imho.

packages/instantsearch-react/src/InstantSearch.js
Morhaus: I think I have an extra space in here!

packages/instantsearch-react/src/InstantSearch.js
Morhaus: TIL, thanks!

packages/instantsearch-react/examples/history.js
Morhaus: I'll keep `alphabeticalSort` for now.

PR #1183, opened 2016-08-09
Stats component and fixes

Morhaus:
Depends on #1182


packages/instantsearch-react/examples/Search.js
```diff
@@ -70,7 +75,12 @@ class Search extends Component {
             defaultSelectedIndex="instant_search"
           />
           <HitsPerPage items={[10, 20, 30]} defaultHitsPerPage={20} />
-          <Range attributeName="price" />
+          <Range
+            attributeName="price"
+            translations={{
+              value: v => `$${v.toLocaleString()}`,
```
bobylito: What is the double $ for? :)

Morhaus: I'm trying to display the price as `$16`. The first `$` is the actual dollar sign, while the second one is there for  string interpolation.

packages/instantsearch-react/src/InstantSearch.js
```diff
@@ -259,24 +264,35 @@ class InstantSearch extends Component {
...
+        ...this.store.getState(),
+        results: content,
+        searching: false,
+      });
+    }).catch(error => {
+      // Since setState is synchronous, any error that occurs in the render of a
+      // component will be swallowed by this promise.
+      // This is a trick to make the error show up correctly in the console.
+      // See http://stackoverflow.com/a/30741722/969302
+      setTimeout(() => {
```
bobylito: isn't `console.error(error)` enough?

Morhaus: `console.error()` creates two stack traces.

image

 This trick allows for the error to appear as if it was thrown outside of a promise.

packages/instantsearch-react/src/InstantSearch.js
```diff
@@ -259,24 +264,35 @@ class InstantSearch extends Component {
...
+        throw error;
+      });
+    });
+
+    promise.catch(error => {
```
bobylito: this one is not super obvious in terms of timing in the promise flow. Why not using the second callback of the `then` a few lines before?

Morhaus: I always forget this form exists, will use it :+1:

Morhaus: That's even better, since it also catches errors in the second callback!

packages/instantsearch-react/src/InstantSearch.js
bobylito: Ok. That's kind of a hack :P but it's for the greater good :)

bobylito:
✨ shiny 👍 :D

PR #1184, opened 2016-08-09
Custom range component

Morhaus:
Depends on #1183


packages/instantsearch-react/src/impl/Slider.js
```diff
@@ -0,0 +1,169 @@
...
+    const {applyTheme, translate, min, max, value} = this.props;
+    const {activeIdx} = this.state;
+
+    const offsets = value.map(v => inverseLerp(min, max, v) * 100);
+
+    const handles = offsets.map((offset, i) =>
+      <div
+        {...applyTheme(i, 'handle', i === activeIdx && 'handleActive')}
+        style={{
+          left: `${offset}%`,
```
bobylito: I haven't seen this new slider live, but `transform: translate` usually works better when positioning stuff.

packages/instantsearch-react/examples/Search.js
```diff
@@ -75,7 +75,12 @@ class Search extends Component {
             defaultSelectedIndex="instant_search"
           />
           <HitsPerPage items={[10, 20, 30]} defaultHitsPerPage={20} />
-          <Range attributeName="price" />
+          <Range
+            attributeName="price"
+            translations={{
```
bobylito: Isn't the proptype `translate` and is a simple function?

Morhaus: The [`translatable`](https://github.com/algolia/instantsearch.js/blob/instantsearch-react/packages/instantsearch-react/src/translatable.js) HOC consumes a `translations` prop that is a map of translation key to translation strings/functions, and forwards a `translate` prop that is a function.

bobylito:
LGTM

We'll see if it needs optimisation with real life tests.

PR #1188, opened 2016-08-12
Reset

Morhaus:
Depends on #1187

### DIFF
--- a/packages/instantsearch-react/src/connectors/connectHierarchicalMenu.js
+++ b/packages/instantsearch-react/src/connectors/connectHierarchicalMenu.js
@@ -9,6 +9,9 @@ function getId(props) {
 function getSelectedItem(props, state) {
   const id = getId(props);
   if (typeof state[id] !== 'undefined') {
+    if (state[id] === '') {
+      return null;
+    }
     return state[id];
   }
   if (props.defaultSelectedItem) {
@@ -68,7 +71,7 @@ export default createConnector({
     const id = getId(props);
     return {
       ...state,
-      [id]: nextSelectedItem,
+      [id]: nextSelectedItem || '',
     };
   },
 
@@ -116,12 +119,12 @@ export default createConnector({
     const selectedItem = getSelectedItem(props, state);
     return {
       id,
-      filters: !selectedItem ? [] : [{
+      filters: selectedItem === null ? [] : [{
         key: `${id}.${selectedItem}`,
         label: `${props.name}: ${selectedItem}`,
         clear: nextState => ({
           ...nextState,
-          [id]: null,
+          [id]: '',
         }),
       }],
     };
--- a/packages/instantsearch-react/src/connectors/connectHierarchicalMenu.js
+++ b/packages/instantsearch-react/src/connectors/connectHierarchicalMenu.js
@@ -116,7 +116,7 @@ export default createConnector({
     const selectedItem = getSelectedItem(props, state);
     return {
       id,
-      filters: selectedItem === null ? [] : [{
+      filters: !selectedItem ? [] : [{
         key: `${id}.${selectedItem}`,
         label: `${props.name}: ${selectedItem}`,
         clear: nextState => ({
--- a/packages/instantsearch-react/src/connectors/connectSearchBox.js
+++ b/packages/instantsearch-react/src/connectors/connectSearchBox.js
@@ -37,9 +37,16 @@ export default createConnector({
     return searchParameters.setQuery(getQuery(props, state));
   },
 
-  getMetadata(props) {
+  getMetadata(props, state) {
+    const query = getQuery(props, state);
     return {
       id: props.id,
+      filters: !query ? [] : [{
+        key: props.id,
+        label: query,
+        hide: true,
+        clear: nextState => ({...nextState, [props.id]: ''}),
+      }],
     };
   },
 });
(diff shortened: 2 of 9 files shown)


packages/instantsearch-react/src/connectors/connectHierarchicalMenu.js
```diff
@@ -116,7 +116,7 @@ export default createConnector({
     const selectedItem = getSelectedItem(props, state);
     return {
       id,
-      filters: selectedItem === null ? [] : [{
+      filters: !selectedItem ? [] : [{
```
bobylito: Why isn't this test consistent with l.104?
What is the type of selectedItem?

packages/instantsearch-react/src/connectors/connectSearchBox.js
```diff
@@ -37,9 +37,16 @@ export default createConnector({
...
 
-  getMetadata(props) {
+  getMetadata(props, state) {
+    const query = getQuery(props, state);
     return {
       id: props.id,
+      filters: !query ? [] : [{
+        key: props.id,
+        label: query,
+        hide: true,
```
bobylito: What is the semantic of `hide`?

packages/instantsearch-react/src/connectors/connectHierarchicalMenu.js
Morhaus: The test on L104 should be the same as this one. Indeed, selectedItem can be either a non-empty string, in which case we consider that there is a selected item whose value is `selectedItem`, or an empty string or null, and in both cases we consider that there is no selected item. I'm not sure why I kept null as a possible value, as the default value should always be `''`.

The reason why the explicit absence of value must be represented as an empty string in state is because the state gets serialized to an URL:

* `{q: 'hello', category: null}` serializes to `?q=hello`, which deserializes to `{q: 'hello'}`: we lose information.
* `{q: 'hello', category: ''}` serializes to `?category=&q=hello`, which deserializes to the original `{q: 'hello', category: ''}`

So in the case where we set `category` to `null`, we lose the knowledge that the state of the `category` widget was explicitly reset. This only has consequences when the user of the library provides a default state for the `category` widget.

0. `const defaultState = {category: 'defaultCategory'};`
1. The state is initialized from the url.  
    `?q=hello` -> `{...defaultState, q: 'hello'}` -> `{q: 'hello', category: 'defaultCategory'}`
2. The user explicitly deselects the `defaultCategory` category.  
    `{q: 'hello', category: ''}` -> `'?category=&q=hello` -> `{...defaultState, q: 'hello', category: ''}` -> `{q: 'hello', category: ''}`

Does that make sense?

From the looks of it, this behaviour might be broken in the current version. I've updated this widget in a future PR (for tests), so maybe I should try to fix it there?

packages/instantsearch-react/src/connectors/connectSearchBox.js
Morhaus: We have two different widgets that consume the metadata declared by other widgets: `Reset` and `CurrentFilters`.

* `CurrentFilters` displays filters from `RefinementList`, `Menu`, `Toggle`, etc. However, it doesn't display the current query nor the current page. And when you click on its `Clear All` button, it doesn't clear them either.
* `Reset` clears *everything*.

As such, I've made a distinction between two types of filters: visible filters (that are displayed in `CurrentFilters`) and invisible filters, that aren't displayed nor cleared by `CurrentFilters` but that are correctly cleared by `Reset`. Invisible filters are declared by setting the `hide` property to `true` on a filter object.

packages/instantsearch-react/src/connectors/connectHierarchicalMenu.js
bobylito: Thanks for the thorough explanation :)

This is definitely tricky. Maybe best is not to slow this down, do something for consistency and keep that in mind as you are moving forward with the tests. WDYT?

packages/instantsearch-react/src/connectors/connectSearchBox.js
bobylito: Shouldn't the visibility be configured or configurable by the user?

Morhaus: It could be, it's just a matter of taking `props` into account here. We could also decide on the `CurrentFilters` side by passing it the `ids` of the filters we want to show. I went with this as a default because I wasn't sure of what we wanted to achieve with this widget, but this is a possibility for sure.

packages/instantsearch-react/src/connectors/connectHierarchicalMenu.js
Morhaus: https://github.com/algolia/instantsearch.js/pull/1188/commits/0609244f5a883922e4e863e9acba8179e0b6148c

Should fix this, by having selectedItem always be either `null` or a non-empty string. On the other hand, the actual value in state should always be a string, and the empty string represents the absence of a value.

packages/instantsearch-react/src/connectors/connectSearchBox.js
bobylito: Either solution sound valid 👍 . Let's move forward, and figure this out when we have a complete overview of the API?

Morhaus: Works for me :+1:

PR #1190, opened 2016-08-12
List Refactor

Morhaus:
Depends on #1189.

### DIFF
--- a/packages/instantsearch-react/src/impl/HitsPerPage.js
+++ b/packages/instantsearch-react/src/impl/HitsPerPage.js
@@ -4,48 +4,43 @@ import themeable from '../themeable';
 
 import LinkList from './LinkList';
 
-class HitsPerPage extends Component {
+const ThemedLinkList = themeable({
+  root: 'HitsPerPage',
+  item: 'HitsPerPage__item',
+  itemSelected: 'HitsPerPage__item--selected',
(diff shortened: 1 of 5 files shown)


packages/instantsearch-react/src/impl/HitsPerPage.js
```diff
@@ -4,48 +4,43 @@ import themeable from '../themeable';
...
-    applyTheme: PropTypes.func.isRequired,
     hitsPerPage: PropTypes.number.isRequired,
-    createURL: PropTypes.func.isRequired,
-    refine: PropTypes.func.isRequired,
     items: PropTypes.oneOfType([
-      PropTypes.arrayOf(PropTypes.shape({
-        label: PropTypes.string.isRequired,
-        value: PropTypes.number.isRequired,
-      })),
+      PropTypes.arrayOf(PropTypes.object),
```
bobylito: What is the new expected shape of the objects?

bobylito:
LGTM

packages/instantsearch-react/src/impl/HitsPerPage.js
Morhaus: This component doesn't care about the shape of an item, since they get passed directly to the `listComponent`.

Issue #1064, opened 2016-06-09
It should be possible to avoid clearing some filters


vvo:
When using clearAll or currentRefinedValues's clear all button, it should be possible to avoid clearing some filters.

The use case is to be able to preselect some filters #1061 while not displaying them in the interface and avoid having them cleared by the user.

Also see #1063 and #740

rogerroelofs:
If we add an excludeAttributes list option to clearAll and currentRefinedValues, what would be the expected UI for currentRefinedValues?  Should the excluded attributes be hidden or visible but not clickable?  Or it could be another option :-)

timkelty:
@rogerroelofs I would say hidden.

With `currentRefinedValues`, we'd also have to consider how this interacts with `options.attributes` and `options.onlyListedAttributes`.

olance:
@vvo this has been closed but only half resolved: the currentRefinedValues widget doesn't offer anything similar to `excludeAttributes` but it really is needed.

It's a real pain to have to list everything manually in the `attributes` option and use `onlyListedAttributes: true`.

Issue #1063, opened 2016-06-09
Merge clearAll and currentRefinedValues in a single widget


vvo:
Those two widgets are really close.

One of them is a single "clear all" button, the other one is a list of the active filters plus a "clear all" button.

We need to review the API and use cases of those widgets to see if and how could we merge them.

As a user I would be a bit lost here, I would prefer to have a "clearFilters" widget that could do both with a straightforward API.

timkelty:
@vvo what was the resolution here, are the filters going to merged or remain separate?

vvo:
Those will remain separate, we are working on react instantsearch whose API will drive IS v2 API.

We will have two separate components but both will be able to show and filter only a subset on attributes if tou want to

PR #1194, opened 2016-08-15
feat(clearAll): Add optional excludeAttributes to list protected filters

rogerroelofs:
Implemented by getting refinements in the widget and removing the protected ones before passing the attribute list to clearRefinementsAndSearch

### DIFF
--- a/src/widgets/clear-all/clear-all.js
+++ b/src/widgets/clear-all/clear-all.js
@@ -41,14 +41,16 @@ clearAll({
   [ cssClasses.{root,header,body,footer,link}={} ],
   [ templates.{header,link,footer}={link: 'Clear all'} ],
   [ autoHideContainer=true ],
-  [ collapsible=false ]
+  [ collapsible=false ],
+  [ excludeAttributes=[] ]
 })`;
 function clearAll({
     container,
     templates = defaultTemplates,
     cssClasses: userCssClasses = {},
     collapsible = false,
-    autoHideContainer = true
+    autoHideContainer = true,
+    excludeAttributes = []
   } = {}) {
   if (!container) {
     throw new Error(usage);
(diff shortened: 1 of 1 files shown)


rogerroelofs:
I feel like there should be a test for the new feature but I'm not sure how to write one given that the widget doesn't have enough state to know whether the correct behavior has occurred.

timkelty:
@rogerroelofs's solution is working great for me, and addresses #1064.

@vvo Would be great to get this merged in in the interim (until #1063 potentially changes the API for the widgets).

:shipit: !

src/widgets/clear-all/clear-all.js
```diff
@@ -90,6 +95,12 @@ function clearAll({
         />,
         containerNode
       );
+    },
+
+    clearAll() {
+      if (this.clearAttributes.length > 0) {
```
bobylito: Why should we change the behaviour? Previously it could have been called and trigger a new search, but the widget sets autohideContainer to true by default, which removes the ability to call `clearRefinementsAndSearch`.

src/widgets/clear-all/clear-all.js
```diff
@@ -70,17 +72,20 @@ function clearAll({
...
+      this.clearAttributes = getRefinements(results, state)
+        .map(one => one.attributeName)
+        .filter(one => excludeAttributes.indexOf(one) === -1);
+      let hasRefinements = this.clearAttributes.length !== 0;
       let url = createURL(clearRefinementsFromState(state));
 
       ReactDOM.render(
         <ClearAll
-          clearAll={this._clearRefinementsAndSearch}
+          clearAll={this.clearAll.bind(this)}
```
bobylito: `bind` should not be called in render. It's best to store the bound function on `this`, and do that in the `init` method.

src/widgets/clear-all/clear-all.js
```diff
@@ -70,17 +72,20 @@ function clearAll({
 
   return {
     init({helper, templatesConfig}) {
-      this._clearRefinementsAndSearch = clearRefinementsAndSearch.bind(null, helper);
+      this.helper = helper;
```
bobylito: Why not binding the helper on the `clearAll` method directly?

src/widgets/clear-all/clear-all.js
rogerroelofs: The issue is that in order to protect certain attributes from clearing, an attribute list must be passed to `clearRefinementsAndSearch`.  If there are no clearable attributes then nothing should happen  even if the caller overrides autohideContainer

rogerroelofs:
@bobylito  I have made the changes you requested.  Thanks for the feedback!

bobylito:
Thanks a lot for your contribution @rogerroelofs. This is greatly appreciated :)

timkelty:
Would love a release with this!

PR #1195, opened 2016-08-16
Docs

Morhaus:
Depends on #1190


bobylito:
W00t I will check this out tomorrow :)

Morhaus:
@bobylito Please make as many comments as you can, be it on the form of the doc or its content :) This is a rough first draft and a lot is missing!

bobylito:
This looks good! I like the layout you used.

packages/instantsearch-react/docs/package.json
```diff
@@ -6,6 +6,7 @@
   },
   "devDependencies": {
     "babel-cli": "^6.11.4",
+    "babel-plugin-transform-object-rest-spread": "^6.8.0",
```
Morhaus: In the end, this isn't needed as it's already present one level above (unless we repeat all babel deps here too).

packages/instantsearch-react/docs/layouts/api.ejs
```diff
@@ -23,11 +23,11 @@
 
       <div class="headings">
         <ul>
-          <% for (const heading of headings) { %>
+          <% headings.forEach( heading => { %>
```
Morhaus: Did you figure out why `for...of` wasn't working correctly in this case?

bobylito: No :) Is it worth looking it?

packages/instantsearch-react/docs/package.json
bobylito: erf sorry for that one :(

bobylito: It feels weird to have dependencies spread over two packages. :)

packages/instantsearch-react/docs/layouts/api.ejs
Morhaus: It would be nice to understand why it fails, but no biggie, we might not even use this template later.

packages/instantsearch-react/docs/package.json
Morhaus: I think we're already depending on the parent .eslintrc, so depending on the parent .babelrc makes sense.

Issue #1185, opened 2016-08-09
toggle widget: autoHideContainer option not respected


timkelty:
See example: https://jsfiddle.net/timkelty/gov4cwhb/1/

I've explicitly set `autoHideContainer` to `true` (even though it is supposed to be the default).
There are 0 matched items, but the widget container still shows.

The `autoHideContainer` option seems to work fine in other widgets (e.g. refinementList)

![screen shot 2016-08-09 at 4 45 14 pm](https://cloud.githubusercontent.com/assets/18329/17533072/439a2510-5e51-11e6-8fc7-ad1ebc43daf4.png)

rogerroelofs:
The issue here is that autoHide is calculated based on the total number of hits for both on and off states.  Switching to using facetValue.count seems reasonable to me but I'm not convinced that behavior will fit every use case.  What do you all think?

PR #1197, opened 2016-08-17
fix(toggle) make autoHide check facetValue.count

rogerroelofs:
Activating a toggle with 0 hits puts the user in a state where they can't get back to the previous state.  The toggle and many other widgets autoHide when no results are found.  This change will autoHide the toggle if changing it's state would result in 0 hits.

Issue #1185

### DIFF
--- a/src/widgets/toggle/implementations/currentToggle.js
+++ b/src/widgets/toggle/implementations/currentToggle.js
@@ -76,11 +76,11 @@ const currentToggle = ({
       isRefined: onData !== undefined ? onData.isRefined : false,
       count: onData === undefined ? null : onData.count
     };
-    const offData = find(allFacetValues, {name: offValue.toString()});
+    const offData = hasAnOffValue ? find(allFacetValues, {name: offValue.toString()}) : undefined;
     const offFacetValue = {
       name: label,
       isRefined: offData !== undefined ? offData.isRefined : false,
-      count: offData === undefined ? null : offData.count
+      count: offData === undefined ? results.nbHits : offData.count
     };
 
         // what will we show by default,
--- a/src/widgets/toggle/implementations/currentToggle.js
+++ b/src/widgets/toggle/implementations/currentToggle.js
@@ -111,7 +111,7 @@ const currentToggle = ({
             createURL={_createURL}
             cssClasses={cssClasses}
             facetValues={[facetValue]}
-            shouldAutoHideContainer={results.nbHits === 0}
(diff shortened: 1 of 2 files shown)


rogerroelofs:
In the case where the off state doesn't have a value (i.e. means don't use the attribute in the search) the count should be the number of hits for 'don't use' rather than the number of hits for 'false'   I'd like someone from the project to validate that before I go changing the tests to match my assumption.

vvo:
Hi @rogerroelofs, thanks for reporting this issue and for the associated PR! Here are my comments:

> This change will autoHide the toggle if changing it's state would result in 0 hits.

This would be in-line with what is done today for other widgets indeed. But we could maybe also do something a bit different: what if as soon as we know toggling would result in no hits we show the toggle as disabled? As for the default rendering it would mean adding `disabled` to the toggle checkbox.

Then we would be able to still allow the user to untoggle if the current state is 0 hits and untoggling would mean showing some hits.

What do you think?

> In the case where the off state doesn't have a value (i.e. means don't use the attribute in the search) the count should be the number of hits for 'don't use' rather than the number of hits for 'false'

That's right and this we can change it.

I am blow away by the fact that you understood so much from our codebase and were able to hack in the toggle code which is not that easy to understand (read: complex, not smart :p).

If you are down to continue this PR based on my first comment then go for it, add tests if you feel like it otherwise we will do it.

Thanks a lot

timkelty:
@vvo @rogerroelofs I'm actually working with @rogerroelofs on this issue.

As the PR stands, this is exactly how I want it to behave in my case. I see the value of the `disabled` suggestion, but feel like that should be a different option (`autoHideContainer` is pretty explicit).

Could be `disableOnEmpty` or something.

vvo:
@timkelty Ok this works for me, in the next instantsearch.js iteration we might default to the disabled solution.

vvo:
I will update the tests, thanks

timkelty:
@vvo Yep, I think defaulting to disabled would be good (since you never want to toggle to see nothing). Then the container hiding feature can still be controlled via `autoHideContainer`

vvo:
> Yep, I think defaulting to disabled would be good (since you never want to toggle to see nothing). Then the container hiding feature can still be controlled via autoHideContainer

Exactly

vvo:
Fixed with #1213, sorry I messed up in merging and did not include commits. I will be careful next time.

timkelty:
Great work everyone, hope to see it in a release soon!

PR #1198, opened 2016-08-17
fix(nouislider): fix the slider for nouislider 8.5.1

Morhaus:
NoUiSlider added a `cssClasses` option in 8.5.0. Since we also have a `cssClasses` prop on our Slider component that is consumed by the higher order components (but still forwarded), we were passing an invalid option to the underlying NoUiSlider.


bobylito:
👍 GG @Morhaus

PR #1207, opened 2016-08-23
InstantSearch refactor + Tests

Morhaus:
Rewrote InstantSearch.js to make it easier to test. Added tests for all connectors.

Still missing tests for implementations and InstantSearch.js.


redox:
> Rewrote InstantSearch.js

😎

PR #1212, opened 2016-08-24
fix(searchBox): handle BFCache browsers

vvo:
When the page is coming from BFCache
(https://developer.mozilla.org/en-US/docs/Working_with_BFCache)
then we force the input value to be the current query.

Otherwise, this happens:
- `<input>` autocomplete = off (default)
- search $query
- navigate away
- use back button
- input query is empty (because `<input>` autocomplete = off)

Firefox and Safari are using BFCache for instant back button


bobylito:
👍

PR #1217, opened 2016-08-24
fix(lodash): use lodash v4, reduce build size

vvo:


bobylito:
LGTM 👍 

How did you reduce the build size?

vvo:
> How did you reduce the build size?

Since we were using lodash v3 and helper used lodash v4, we had two different versions living in the build. Now there's only lodash v4

bobylito:
👍 let's merge!

PR #1220, opened 2016-08-25
Add an POC for ISv2

Morhaus:
Example implementation of an instantsearch.js-like API, but using the architecture from IS-react under the hood.


vvo:
thx this will be very handy for us :)

PR #1223, opened 2016-08-26
fix(poweredBy): fixed Algolia logo version

### LINKED ISSUE #1222: Better PoweredBy logo
The current powered by logo is not very nice: it is blurred, not retina ready and doesn't scale well.

image

redox:
This blue-ish one is the very latest version.

algolia-light- 1

Fix #1222


vvo:
Does It Scales?:tm:

bobylito:
And it is smaller! :D

redox:
> Does It Scales?™️

The image is twice the size specified in the CSS: it's the basic/simple way to support retina screen.

vvo:
I am ok updating the logo to the latest design, just worried that we need to maybe go a bit further to really solve #1222 in case it's still blurry (that's the initial request)

redox:
Updated to use the SVG version, bye bye IE < 9.

rayrutjes:
Looks so much better:
image

vvo:
That's ok for IE8, we do not support it.

PR #1224, opened 2016-08-26
perf(filesize): use preact in production build

vvo:
instantsearch.min.js **React build** => 113kb (gzipped)
instantsearch.min.js **preact build** => 77kb (gzipped)


bobylito:
Love that! Is it possible to offer two builds?

vvo:
> Love that! Is it possible to offer two builds?

Yes feasible, but how would we guide users towards them? As a user of instantsearch.js I am not aware of either React or preact so not sure how I could make a choice.

But it's a good idea to be able start testing this build yes. We could also just push it and see what happens. Our heavy demo is working with it so if we have glitches they should not happen on that many websites.

redox:
That's awesome 👍

bobylito:
> Yes feasible, but how would we guide users towards them? As a user of instantsearch.js I am not aware of either React or preact so not sure how I could make a choice.

True, mainly wondering on the possibility... Would you like to test that in the wild?

vvo:
> Would you like to test that in the wild?

Will test it with magento and wordpress extensions, if it works then :shipit:

vvo:
:white_check_mark: tested in magento
:white_check_mark: tested in wordpress

vvo:
- [ ] investigate functional tests
- [ ] test on mobile

developit:
😍

vvo:
Tests are :white_check_mark:, let's release this

vvo:
Since we were exposing that we accepted React components as templates and it was accessible in the production build, I had to rollback the default build and create a specific preact build. This should not be an issue in ISV2 where we will completely hide the fact that we use react/preact for non react users.

developit:
@vvo Good catch, sorry I didn't have the context to spot that before hand

vvo:
@developit its entirely our responsibility no worriess ahah. In V2 this need will disappear and we will use preact by default

developit:
Are you guys switching to allowing templates instead, or something else? (just curious!)

PR #1255, opened 2016-09-02
fix(rangeSlider): round pips numbers when step is integer

vvo:


pixelastic:
Oh thanks. I had to work around that lately with ```tooltips: {
        format: _.parseInt
      },``` :)

PR #1265, opened 2016-09-07
chore(dev): enhance the dev experience

vvo:
It's now a lot faster to get updates on the page.

Still some issues (that were already present before): some menu items
are disappearing when you change layouts or markdown files.

Most probably because the navigation building is based on the
processed files.

We will have to fix that.


README.md
```diff
@@ -12,7 +12,6 @@ We use our documentation as the main way to develop instantsearch.js.
 
 ```
 npm install
-(cd packages/react-instantsearch && npm install)
```
bobylito: 👍

bobylito:
LGTM

docgen/.eslintrc.js
```diff
@@ -0,0 +1,12 @@
+var join = require('path').join;
+
+module.exports = {
+  "extends": join(__dirname, "../.eslintrc.js"),
+  "settings": {
+    "import/resolver": {
```
vvo: This lint module is very helpful, it already warned us when:
- the module/file you want to import does not exists
- the module is not present in package.json

Here we configure it to use the webpack resolver so that the webpack alias in the docs for "react-instantsearch" points to the local code

PR #1268, opened 2016-09-08
chore(dev): only use browser-sync, webpack HMR

vvo:


docgen/webpack.config.babel.js
```diff
@@ -3,13 +3,13 @@ import {join} from 'path';
 
 export default {
   entry: {
-    api: join(__dirname, 'assets/js/api.js'),
-    media: join(__dirname, 'assets/js/examples/media.js'),
+    'assets/js/api': join(__dirname, 'assets/js/api.js'),
```
vvo: This is done to be able to have the same directory structure once generated in the build output.

Issue #1266, opened 2016-09-08
How to easily conditionally display components based on the search state


mthuret:
A typical use case is:
- when there're no search results, I want to display a specific component
- when there's no query I want to display a completely different homepage (jadopado like)

Currently in `react-instantsearch` we have a <NoResults/> component that allows us to display specific behavior when there're no results. 

The <NoResults/> current implementation is simple: it returns the child if there're results, and `null` if not. 

It means we can write:

```jsx
<NoResults>
	<div> No results</div>
<NoResults/>
<div>
//rest of the page
</div>
```

Now the issue is that `// rest of the page` will still be displayed when there're no results: **we need to have an if/else behavior**. 

With the current implementation, we could use an HOC and write something like: 

```js
function MyResults(props) {
  if (props.noResults) {
    return (
      <div className="results-wrapper">
        <div className="no-results">
          No results found matching <span className="query">{props.query}</span>
        </div>
      </div>
    );
  } else {
    return (
      <div className="results-wrapper">
      </div>
    );
  }
}

const ConnectedResults = NoResults.connect(MyResults);
```

As you can see, having a NoResults component is not that useful because the only thing we need is the ability to use the connect function. Also, it's not very simple in a user point of view, regarding that it's a common pattern. 

Now, what we could also do is to give the opportunity to the user to write directly something like this : 

```jsx
<NoResults>
        <div className="no-results">
          No results found matching <span className="query">{this.props.query}</span>
        </div>
<NoResults/>
<HasResults>
	<stuff>
<HasResults/>
```

And if we want to access parent props from the child, we can change the <NoResults/> implementation to this : 

```js
render() {
    if (!this.props.noResults) {
      return null;
    }

    return Children.map(this.props.children, child => {
      return React.cloneElement(child, this.props); //perform a merge between parent and child props
    })[0];
  }
 ```

That's one way to have a simpler api. 

With that in mind, opening the conservation: 

* Are HOC sufficient? Or are those patterns (NoQuery, NoResults, HasQuery ..) too frequent in UIs?
* Is the pattern with NoResults/HasResults is a better way to go (with some more documentation to know the props that are available) 
* Does you have any other ideas about the api we should use?

bobylito:
I don't know if it's relevant to our discussion but I have two use cases that we often see:
 - when first displaying the page, display a content page instant of the empty query (a query with no parameters set)
 - when there are no results, display a list of all the refinements so that the end-user can discard some of them to get back to a state with results

[...]

PR #1269, opened 2016-09-09
E commerce example

mthuret:


redox:
Boom 💥 first commits \o/ @mthuret 👍

docgen/layouts/example.ejs
```diff
@@ -5,6 +5,8 @@
     <meta charset="utf-8">
     <meta content="width=device-width,initial-scale=1.0,minimum-scale=1.0,maximum-scale=1.0,user-scalable=no" name="viewport">
     <link rel="icon" href="assets/favicon.png">
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/bootstrap/3.3.5/css/bootstrap.min.css" />
```
vvo: @bobylito You are working on making the templates use pug, It would be also good to be able from an example to inject things in the `<head>` (style, js).

vvo: examples.js could just be a base layout that api reuse and examples are directly using (while injecting some css/js).

bobylito: That's exactly what I've done :)

bobylito: But for now it's good :)

packages/react-instantsearch/src/widgets/NoResults/connect.js
```diff
@@ -9,6 +9,7 @@ export default createConnector({
     }
     return {
       noResults: search.results.nbHits === 0,
+      query: search.results.query,
```
vvo: This is where it gets tricky (and to link to #1266), we don't want to limit ourselves to query, we might want to display more things. So it means the `<NoResults>` widget cannot have query here.

But if you need the query inside your example, we need to be able to provide an easy way to get that. And that's where the widget states and custom widget creation comes into play.

Here you would need to be able to easily create a custom widget that is able to access any InstantSearch state param. Right now this is documented but may be too hard to do.

Help :)

packages/react-instantsearch/src/widgets/SortBy/SortBy.js
```diff
@@ -34,12 +34,11 @@ class SortBy extends Component {
 
   render() {
     const {applyTheme, refine, items, selectedIndex} = this.props;
-
     return (
       <Select
-        {...applyTheme('root', 'root')}
+        applyTheme={applyTheme}
```
vvo: The sortBy was not working?

mthuret: When I first try it, no. Only the SortByLinks. 
Got this : uncaught TypeError: applyTheme is not a function

vvo: Yep indeed I guess it was something not finished. Thanks

packages/react-instantsearch/src/widgets/NoResults/connect.js
mthuret: Rollback done. The example uses now a custom widget to display the results.

docgen/assets/js/examples/ecommerce.js
```diff
@@ -184,7 +185,14 @@ function CustomHits(props) {
   );
 }
 
-function CustomResults(props) {
+const CustomResults = createConnector({
+  displayName: 'CustomResults',
+
+  getProps(props, state, search) {
```
vvo: Nice in the end even if the naming/way to do this is not obvious right now, you still managed to do something very small

docgen/assets/css/examples/ecommerce.css
```diff
@@ -0,0 +1,652 @@
...
+
+aside .facet-color[data-facet-value="Pink"] {
+    background-color: #C0116D;
+}
+
+aside .facet-color[data-facet-value="Gray"] {
+    background-color: #6F6E6C;
+}
+
+aside .facet-color[data-facet-value="Blue"] {
```
vvo: 2 spaces for the indentation (all the JavaScript projects at Algolia)

vvo:
Let's merge this, we will add some linting to avoid 4 spaces.

PR #1270, opened 2016-09-09
chore(deps): upgrade all deps

vvo:


vvo:
Also fixes dependencies because that's the safest

PR #1277, opened 2016-09-13
docs(materialui): add example based on material-ui react library

mthuret:


docgen/src/3rd-party-librairies.md
```diff
@@ -0,0 +1,62 @@
...
+layout: api.ejs
+nav_groups:
+  - core
+nav_sort: 1
+---
+
+# Integration with other React UI Libraries
+
+Even if react-instantsearch provides widgets out of the box you are free to use it with
+any other React UI library (such as Material-UI, React Toolbox, React-Bootstrap and so
```
vvo: "you are free to use it with
 +any other React UI"

Feels good to say so :)

docgen/webpack.config.babel.js
```diff
@@ -24,6 +25,8 @@ export default {
   resolve: {
     alias: {
       'react-instantsearch': join(__dirname, '../packages/react-instantsearch/'),
+      'material-ui': join(__dirname, '../node_modules/material-ui/'),
```
vvo: Is this mandatory? I would say there should be no need to do so. We do it for react-instantsearch only to link a local library

mthuret: Yes you are right. This is useless. I removed it.

PR #1281, opened 2016-09-14
docs(nav): small reorg

vvo:
added an examples section,
commented out the headers#, I feel like it's not useful for now and
hard to understand.

Preview:
2016-09-14-230336_592x1249_scrot


vvo:
It's only a small step, I will work on drafting a proposal for the documentation tree today including what we need to add, group, change

PR #1289, opened 2016-09-16
fix(pagination): always transition state

### LINKED ISSUE #1287: When clicking on the current page, search should not go back to the first one (no toggle)
This is an example of this behaviour on the tourism example.

pagination-revert-bug

bobylito:
fix #1287

### DIFF
--- a/packages/react-instantsearch/src/widgets/Pagination/connect.js
+++ b/packages/react-instantsearch/src/widgets/Pagination/connect.js
@@ -1,5 +1,4 @@
 import {PropTypes} from 'react';
-import {omit} from 'lodash';
 
 import createConnector from '../../core/createConnector';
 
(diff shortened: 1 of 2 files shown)


Morhaus:
This is the behavior that clears the page when the state changes from somewhere else than the pagination widget. For instance, to go back to page 0 when we add a refinement.

Morhaus:
If you remove this behavior and implement it differently, you can clear all code related to transitionState since that is the only place where it's being used.

vvo:
Owww @bobylito Maybe we need to revert and find a different approach?

Morhaus:
What issue did you meet with this implementation of transitionState.

vvo:
> What issue did you meet with this implementation of transitionState.

When page 2 is selected, if you select it again it removes the current page and goes to one. Maybe we need some special handling on it and that's it.

But it should be at the implementation side perhaps, not the connector

vvo:
cc @bobylito did you have a look?

bobylito:
I will have a look :) Sorry for the lag :)

Issue #1296, opened 2016-09-19
Synchronizing state of widgets that share the same attribute name 


mthuret:
Currently widgets that share the same attribute name doesn't share the same state. It means that if a user has two widgets for the same refinement, there will not be synchronized. 

Example : on the e-commerce demo we have two widgets to refine on prices. If the user selects a range on the prices list, the input range will not be automatically synchronized. Same thing for the url.

One solution would be that when the store is updated, widgets that share the same attribute name as the one that is responsible for the update, would have their state automatically changed as well. 
But this synchronization is currently not that simple because widgets doesn't necessarily share the same internal state. 

ex: taking the e-commerce demo again, for displaying prices, we use NumericRefinementList and InputRange widgets. The first one uses a "min:max" string as its internal, the seconde one uses an object with min/max values. 

@Morhaus  : when developing react-instantsearch, have you thought about this problematic and already have some ideas about it ?

Morhaus:
Since the state is shared between all children of the InstantSearch component, different components *can* read and update the same state key. For instance, you can have two Pagination components at the same time. However, the format of their state should be normalized. In the case of NumericRefinementList and InputRange, I think it makes sense to pick one format for both.

However, for widgets that control the same refinement but have incompatible state, then maybe only one of those widgets should have a state at any point in time. I don't know of any such widget, so I'll use NumericRefinementList and InputRange as an example: when you select a value in a NumericRefinementList, the InputRange's state would get cleared, and vice versa. That would also mean using different state keys for both widgets.

rayrutjes:
IMHO there should be only one source of truth which is the search state. Every widget should use that store for its own state. Widget internal state should only be some sugar to compute values but should always be directly bound to the search state.

Morhaus:
See https://github.com/algolia/instantsearch.js/issues/1170

rayrutjes:
Just took a look @Morhaus thx for pointing that out.

I think the solution implemented in #1170 brings more problems than it actually solves. 
It makes reasoning about widget states difficult.

Regarding the `currentRefined` widget, if it depends on other facets, then it should reference those facets and react on their changes.

It could even just use the attribute name for the display and let users provide an array of `attribute => label` for the display, and extract the refined values from the helper's state. IMHO that is the way to go about it.

Not sure if I'm missing something here @Morhaus

Morhaus:
[...]

PR #1297, opened 2016-09-19
feat(widget): add an input range

mthuret:
This PR add a new widget : InputRange.

It also updates the e-commerce examples with refinements on prices. 

Be careful : there's no sync between the two widgets that refines on price. I open a discussion about it here #1296

### DIFF
--- a/docgen/src/examples/e-commerce/App.js
+++ b/docgen/src/examples/e-commerce/App.js
@@ -6,12 +6,13 @@ import {
   SearchBox,
   HierarchicalMenu,
   RefinementList,
-  Range,
   Hits,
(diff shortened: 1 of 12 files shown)


vvo:
Awesome, I would say that to be consistant with other widget where we provide different implementation, this should be in src/widgets/Range/RangeInput like we have for src/widgets/Menu/MenuSelect.

What do you think?

vvo:
Also, can you add tests, the whole testing infrastructure is ready. I know we do not have tests for everything right now, but having tests for new widgets should be mandatory now

vvo:
Awesome work GG

Issue #1306, opened 2016-09-19
NumericRefinementList and Range distinction


vvo:
Our NumericRefinementList has it's own namespace while this is a Range with multiple start/end values and a possible multiple/single element selected behavior.

I don't know yet how to make this easier. I find it weird that we have both NumericRefinementList and Range. A NumericRefinementList is a Range dealing with multiple ranges. So we either have a naming issue or we need to merge them.

Maybe we can just rename NumericRefinementList to MultiRange?

vvo:
Also add `multiselect={false}` to allow selecting only one value. By default you should be able to select multiple of them.

Also visually indicate non-relevant ranges given the current faceting stats.

mthuret:
The multiselect option is no longer a priority for the first react-instantsearch release.

PR #1317, opened 2016-09-20
feat(widget): add a new rating range

mthuret:
ratings range

### DIFF
--- a/docgen/src/examples/e-commerce/App.js
+++ b/docgen/src/examples/e-commerce/App.js
@@ -89,6 +89,11 @@ const Facets = () =>
             item={<ConnectedColorRefinementList attributeName="colors" operator="or"/>}
           />,
           <RefinementListWithTitle
+            title="Rating"
+            key="rating"
+            item={<Range.Rating attributeName="rating" max={5}/>}
+          />,
+          <RefinementListWithTitle
             title="Price"
             key="Price"
             item={ <CustomPriceRanges
@@ -105,7 +110,7 @@ const Facets = () =>
               ]}
             />}
           />,
-          <Range.Input attributeName="price" id="price_input"/>,
+          <Range.Input key="price_input" attributeName="price" id="price_input"/>,
         ]}
       />
       <div className="thank-you">Data courtesy of <a href="http://www.ikea.com/">ikea.com</a></div>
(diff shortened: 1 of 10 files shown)


bobylito:
Looks good to me. I'm not sure about the naming and that it is related to range (not a big deal though)

mthuret:
About the naming `RangeRatings`:
Currently without default theming, this widget only provides span as stars. Also, I was thinking that maybe for a rating, a user will want to use some other icons than stars. (maybe heart, or whatever :))

Also, be aware that there is a wierd behavior that you can see with the e-commerce example : if you clear all filters or click again on a selected rating, the largest one would become the new selected one. That's because I use the Range connector which when no refinement is set, put its value to the largest range possible. 
To avoid that we can compute an extra props that will say if there's refinements or not. Do you have any other ideas ?

mthuret:
And to complete my previous comment: we had a discussion with @vvo on how to implement this widget. At first, I used the NumericRefinementList. But this widget needs a static list of ranges. So it would have required some changes on the connector side, to handle both statics and dynamics range computation. (There's also this issue about the naming of NumericRefinementList #1306)

But considering the new issue with default values for Range, maybe it would have been better to modify NumericRefinementList. Anyway, need your thoughts about that guys :)

vvo:
> (maybe heart, or whatever :))

good idea! :heart:

vvo:
> Also, be aware that there is a wierd behavior that you can see with the e-commerce example : if you clear all filters or click again on a selected rating, the largest one would become the new selected one. That's because I use the Range connector which when no refinement is set, put its value to the largest range possible. 
> To avoid that we can compute an extra props that will say if there's refinements or not. Do you have any other ideas ?

The fact that the largest range is shown selected when there's no range selected is something we should do. Because it does represent the current search state. The user is free to handle this largest range as he wants, maybe with a "Clear" button or "All".

bobylito:
I must say that I'm not sure I grasp the wholeness of the problem, so bear with me. I feel that if we need to work around a behaviour implemented as a connector using a new variable that will used only in one widget, we need a new connector.

In any case, the extension of a connector should be consistent with its role. If it's the case that's fine with me. Do you have an example of the API that you want to implement?

PS : http://explosm.net/comics/3902/

vvo:
> I must say that I'm not sure I grasp the wholeness of the problem

I will sum it up: the current StarRating widget @mthuret implemented has a different behavior than what's available in IS V1.

In the instant_search dataset, ratings are ranged from 1 (min) to 5 (max). When the page loads, like https://community.algolia.com/instantsearch.js/examples/media/. The starRating widget is instantiated and no ratings refinement are selected. BUT it's lying, there's a default refinement being shown, which is "no refinement on ratings" => show all movies from 1 to 5.

So if you click on ★☆☆☆☆ & up then there's no effect on the result set. This is not good since a button should have some effect in this situation.

On amazon.com for example, ratings are ranged from 0 to 5, but they never show ☆☆☆☆☆ & up; because that's what's already being shown by default.

While in react-instantsearch, now the behavior when the page loads or when you deselect ★★★☆☆ is that ★☆☆☆☆ will be selected. Which is coherent with the search state but may be misleading.

So what we could do is always hide the largest range if not relevant (= already what's being shown), or gray it (as we want).

For now, let's do the obvious and let show the default range as being selected (☆☆☆☆★). Then we will iterate quickly once we decide that maybe, like on amazon.com, the largest range, if already applied, should not be shown. That's a design decision.

vvo:
Also, one thing we are not doing good in IS V1 is that we do have movies that have five stars, but we are unable to refine only on them.

I would add this: if the Algolia stats reveals that there are some objects reaching the maximum rating, then show the refinement for ★★★★★ because it's useful for users.

And we can go even further, by denoting the available/unavailable ranges like how it's done today:

2016-09-21-154338_244x136_scrot

mthuret:
Alright I make the following change : 

- largest range is by default always selected 
- display a line for the [max:max] range

vvo:
Are we good here @mthuret? Do we also only display relevant ranges, or are we able to distinguish non relevant ranges via a specific theme prop?

mthuret:
@vvo If you don't specify a min/max you got a rating line from lowerBound to upperBound. 
However if the count for a line is 0, I currently didn't add a specific classname for it. Do you want this to be added ?

vvo:
> However if the count for a line is 0, I currently didn't add a specific classname for it. Do you want this to be added ?

I would say yes, since the current starRating widget has this ability (see https://github.com/algolia/instantsearch.js/pull/1317#issuecomment-248616488), and it is very valuable. If the range is not relevant then we should hint the display/user.

What do you think

mthuret:
Yes that makes sens 👍

mthuret:
Classnames were added. 

capture d ecran 2016-09-22 a 12 09 15

PR #1322, opened 2016-09-21
docs(example): add Media example

vvo:
This PR also includes a change to expose Component.defaultTheme so you
can easily extend it. I am not sure this is the right approach, only
did it so I could move on.

Also small tweaks around JS file ignoring, eslint and update all
packages


packages/react-instantsearch/src/core/createConnector.js
```diff
@@ -28,6 +28,7 @@ export default function createConnector(connectorDesc) {
 
   return Composed => class Connector extends Component {
     static displayName = `${connectorDesc.displayName}(${getDisplayName(Composed)})`;
+    static defaultTheme = Composed.defaultTheme;
```
vvo: @mthuret You may be interested in this change (and the next one in themeable.js). This is what I needed to be able to do:

```js
var theme = {
  ...RangeRatings.defaultTheme,
  class: 'override'
};
```

Since you already explored the CSS theming of react-instantsearch, do you think that's an OK way to expose this defaultTheme?

vvo:
Missing some key={}. Gonna try if eslint-plugin-react could detect that or if travis-CI could fail on this because it's painful to see it afterwards.

vvo:
I am good here, added some eslint rule on key={} and added the new Ratings widget.

PR #1333, opened 2016-09-26
feat(theme): add default themes using css-modules

mthuret:
This PR lets you have a first look at what using css-modules for default theming could look like. I modify the examples to use them. 

You also can run storybook (npm run storybook), to see how the widgets looks like. 
The widgets that has default theme are SearchBox, Pagination, RangeRatings, Range. 

Warning : 
- tests are not working yet. 
- props docs generation was commented because of some issues with css-modules loader.


docgen/src/examples/tourism/App.js
```diff
@@ -234,11 +239,15 @@ const RoomType = RefinementList.connect(({items, refine, selectedItems}) => {
 });
 
 function Price() {
+  const theme = _.entries(Range.defaultTheme).reduce((acc, i) => {
```
mthuret: This is one way to do a more smart merge letting you have your own css redefining stuff and the default theme.

docgen/plugins/inlineProps/extractMetadata.js
```diff
@@ -28,6 +28,10 @@ const getCompiler = memoize(entriesArray => {
             join(__dirname, 'transformPropsLoader.js'),
           ],
         },
+        {
+          test: /\.css$/,
```
vvo: Here we might just want to return an empty object every time we require some css. This should be sufficient for props/docs generation.

docgen/src/examples/media/App.js
```diff
@@ -126,5 +128,5 @@ const Results = () =>
     <div id="stats" className="text-right text-muted"><Stats/></div>
     <hr />
     <div id="hits"><Hits itemComponent={Hit} hitsPerPage={10} /></div>
-    <div id="pagination" className="text-center"><Pagination/></div>
+    <div id="pagination" className="text-center"><Pagination theme={paginationTheme}/></div>
```
vvo: nice

docgen/src/examples/tourism/App.js
vvo: Yep I wonder if we should provide it or not as an utility. For now maybe just explain how to do it in the theming guide and let's revisit it after.

package.json
```diff
@@ -12,10 +12,13 @@
     "postinstall": "cd packages/react-instantsearch && npm install",
     "doctoc": "doctoc --maxlevel 1 README.md CONTRIBUTING.md",
     "test": "(cd packages/react-instantsearch && npm test) && npm run lint && NODE_ENV=production DOCS_MOUNT_POINT=/instantsearch.js/react/ DOCS_DIST=docs/react/ npm run docs:build && NODE_ENV=production npm run build",
-    "dev": "cd packages/react-instantsearch && npm run dev"
+    "dev": "cd packages/react-instantsearch && npm run dev",
+    "storybook": "start-storybook -p 6006",
```
vvo: Do you use storybook heavily while developing here? (I believe yes). If so, what are we missing in our documentation website to not have to use storybook at all while developing? Maybe widgets previews with easy save/update?

I am not sure we want to have to ways to develop/try widgets on react-instantsearch. The current state of the doc dev flow is not ready compared to storybook but ultimately I would only keep one (the dev website).

What's your POV on this subject? Do you think it would be great to use storybook aside? I am really not into it so I may be missing a lot of possibilities that storybook offers that would complement well our own dev flow and library.

Maybe we could even integrate it in the dev website flow and documentation website output? Do  you know any library doing so for exhaustivity maybe along with ease of try/see?

packages/react-instantsearch/src/widgets/SearchBox/SearchBox.js
```diff
@@ -2,6 +2,21 @@ import React, {Component, PropTypes} from 'react';
 
 import themeable from '../../core/themeable';
 import translatable from '../../core/translatable';
+import theme from './SearchBox.css';
+
+const SubmitIcon = () =>
+  <svg xmlns="http://www.w3.org/2000/svg" id="ris-search-icon" viewBox="0 0 40 40">
```
vvo: Ultimately having those as svg files and using some react-svg-loader (or smg) would be easier. It adds noise to the code currently. Let's keep that in mind for later. I am not sure this is easy/compatible with babel webpack loader plugin that we need to use.

docgen/src/examples/tourism/App.js
mthuret: Yes, we could provide two utility method 👍 
* an erase merge 
* a combined merge 

In the meantime, updating the docs is fine.

package.json
mthuret: I use storybook mainly because of issues in the doc side conf and I wanted to move on. Also it guaranteed me that it was a complete isolated env and it's quite fast when you want to do a lot of little experiments/variations. 

But I believe once those issues will be solved, we will have the same things with our widgets demo pages. I push it because I wanted to have a look at the default styling but It was not intended to last. 

As for storybook, it's still under development so they add some nice stuff regularly like snapshoting for free this week. But for now I don't believe it could feet nicely within our doc. For example Airbnb add it as a complete external website (http://airbnb.io/react-dates)

vvo: I think I would be ok with having react/story endpoint.

Being in dev a live storybook env and in prod our full blown storybook for people to get inspired/play (well it might just be the same actually).

What do you think?

If it really feels like a very good fast env for you, then let's do it

mthuret: Well we can have a try, but I don’t want to add something because it fits my workflow but it somehow duplicates stuff :slightly_smiling_face: Maybe I can fill it while working for the default theme stories, and you guys tell me what you think of it.

vvo: Yes

docgen/src/examples/tourism/App.js
bobylito: What are the semantics of this two merge?

package.json
bobylito: Imho storybook seems to be a nice addition and will help us or any contributor. I'm wondering if it could work with jest snapshot to check visually the regressions / changes.

packages/react-instantsearch/src/widgets/Range/__snapshots__/RangeRatings.test.js.snap
```diff
@@ -3,6 +3,7 @@ exports[`Range applies translations 1`] = `
   className="RangeRatings">
   <a
     className="RangeRatings__link"
+    href="#"
```
bobylito: We should be able to generate a url for that, no?

packages/react-instantsearch/src/widgets/SearchBox/SearchBox.css
```diff
@@ -0,0 +1,121 @@
...
+  width: 13px;
+  height: 13px;
+}
+.input:valid ~ .reset {
+  display: block;
+  animation-name: sbx-reset-in;
+  animation-duration: .15s;
+}
+
+@keyframes sbx-reset-in {
```
bobylito: Isn't that possible with a simple transition?

packages/react-instantsearch/src/widgets/SearchBox/SearchBox.js
bobylito: :+1:

docgen/src/examples/tourism/App.js
mthuret: Saying that the default theme as a key root bind to classname A and the user theme as a key root bind to classname B.
If you perform an erase merge root will be bind to classname B. 
If you perform a combined merge, root will be bind to classname A and B. 

Maybe not the better name though :)

packages/react-instantsearch/src/widgets/Range/__snapshots__/RangeRatings.test.js.snap
mthuret: You mean in the test file having a createURL functions a little be smarter than this : createURL={() => '#'} ?

packages/react-instantsearch/src/widgets/SearchBox/SearchBox.css
mthuret: I don't know :) Most of the code is from the @Shipow searchbox. But i can try to simplify a little bit more.

mthuret:
Next steps are : 

* resolve tests issues
* add default empty theme for every widgets
* change pattern for generated css-modules classnames with webpack
* configure the build 
* investigate issue with the Pagination widget and defaultTheme (Pagination as currently no default theme that you can retrieve in the user side because the theme is applied to a LinkList).
* configure and add widgets/use case to storybook

package.json
mthuret: Yes you can, and since this week, you are able to do it without a lot of configuration. 
It's their own cli though: https://voice.kadira.io/snapshot-testing-in-react-storybook-43b3b71cec4f#.jjp5n6ivb

vvo:
> The current state looks good to me. I like it! Good work @mthuret What do we do with the open points, @vvo?

Marie did a nice summary of the next steps, let's do that!

vvo:
I would add: document how to run Storybook (README.md) and what is the current purpose of it (When to use it).

We will most probably make it part of the normal `npm run dev` command afterwards

mthuret:
Alright, I made some changes to the PR (refer to next steps comment to see them). 

I modify a little bit the storybook to show you what can be done (like modifying props live). I also tried to use snapshots, that works, but we have issues when snapshoting components from the public api: some are null. I Didn't take the time to investigate yet.
If you like what's already there, I will add all the widgets. 

Also, I didn't do the build part, because I think we need to discuss a little bit about it before. 

I modify the styling docs, it's not satisfying yet though. I think it would be better with a complete example and with us providing some utility functions to help perform merge/override of styles. 

Finally about the tests warnings. I believe it happens because now default theme are css files, and mocked with jest. So does we ignore them or maybe we can put the __DOC__ variable to true when testing ? Any ideas ?

vvo:
> Also, I didn't do the build part, because I think we need to discuss a little bit about it before.

We can discuss it right now, we cannot merge the PR without the build part though, that would break the build ultimately

> Finally about the tests warnings. I believe it happens because now default theme are css files, and mocked with jest. So does we ignore them or maybe we can put the DOC variable to true when testing ? Any ideas ?

For now this is good we can still merge without having to worry to much how to solve it. Tests are passing, we have only two warnings. I don't know yet how to solve it.

packages/react-instantsearch/src/widgets/SearchBox/SearchBox.js
proudlygeek: We're using `react-svg-loader` on our dashboard react app:

https://github.com/algolia/AlgoliaWeb/blob/master/package.json#L68

Works flawlessly! 👍

mthuret:
I added almost all widgets to storybook. So I guess we had our examples :) 

I didn't apply any style for basic element, so the raw is kinda ugly. Maybe it's better if I style a little bit those things no ?

bobylito:
> I didn't apply any style for basic element, so the raw is kinda ugly. Maybe it's better if I style a little bit those things no ?

It's a great idea! However this PR starts to aggregate everything and best would to merge it and do that in a new one, WDYT?

vvo:
> It's a great idea! However this PR starts to aggregate everything and best would to merge it and do that in a new one, WDYT?

Yes, I am adding the last part (babel build)

vvo:
giphy

Let's ship this and iterate

PR #1336, opened 2016-09-26
fix(core): InstantSearch should accept 0 children

bobylito:
This way it doesn't return an error while building the initial structure
of a new search. It still throws when there is more than one child because
there is no generic container that would work without react-dom.


vvo:
Great, indeed returning an error here was not so smart since it's expected nothing is rendered if there's nothing inside it.

PR #1346, opened 2016-09-27
feat(highlight): provide a function to use highligth in react

bobylito:
Usage in a custom hits widget:

```javascript
const ProductHits = Hits.connect(({hits}) => {
  const hitComponents = hits.map( hit =>
    <div key={hit.objectID}>
      <span className="hit-name">{hightlight('name', hit)}</span>
    </div>
  );

  return <div className="hits">{hitComponents}</div>;
});
```

### DIFF
--- a/packages/react-instantsearch/src/core/highlight.js
+++ b/packages/react-instantsearch/src/core/highlight.js
@@ -1,0 +1,67 @@
+import React from 'react';
+
+function extractTagName(highlightedValue) {
+  const tagParse = (/<(.+?)>/).exec(highlightedValue);
+  return tagParse === null ? null : tagParse[1];
+}
+
+/**
+ * This function transforms an highlighted response
+ * to an array of React elements. This should be
+ * used in place of dangerouslySetInnerHTML when
+ * using the highlightReasult provided by Algolia.
+ *
+ * The function autodetects which tag is used in
+ * results to highlight, and creates the corresponding
+ * React element.
+ *
+ * It also creates keys on React elements to avoid warnings.
+ *
+ * The genegrated highlighted Elements use the `ais-hightlightedValue`
+ * class name attached.
+ *
+ * @param {string} attribute the name of the attribute for which we want the highlighted value
+ * @param {object} hit the hit from the algolia response
+ * @returns {Array.<string|ReactElement>} the array of react element
+ * @example
+ * const ProductHits = Hits.connect(({hits}) => {
+ *   const hitComponents = hits.map( hit =>
+ *     <div key={hit.objectID}>
+ *       <span className="hit-name">{hightlight('name', hit)}</span>
+ *     </div>
+ *   );
+ *
+ *   return <div className="hits">{hitComponents}</div>;
+ * });
+ */
+export default function hightlight(attribute, hit) {
+  if (!hit) throw new Error('The hit containing the attribute should be provided');
+  if (!hit.hasOwnProperty(attribute)) throw new Error(`${attribute} should be a retrievable attribute.`);
+  if (!hit._highlightResult ||
+      !hit._highlightResult.hasOwnProperty(attribute)){
+    throw new Error(`${attribute} should be an highlighted attribute`);
+  }
+
+  const highlightedValue = hit._highlightResult[attribute].value;
+  const tag = extractTagName(highlightedValue);
+  const firstPassSplit = highlightedValue.split(`<${tag}>`);
+  const firstValue = firstPassSplit.shift();
+  const elements = firstValue === '' ? [] : [firstValue];
+
+  firstPassSplit.forEach((split, i) => {
+    const s = split.split(`</${tag}>`);
+    elements.push(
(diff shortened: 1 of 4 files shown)


packages/react-instantsearch/src/core/highlight.js
```diff
@@ -0,0 +1,64 @@
...
+ *       <span className="hit-name">{hightlight('name', hit)}</span>
+ *     </div>
+ *   );
+ *
+ *   return <div className="hits">{hitComponents}</div>;
+ * });
+ */
+export default function hightlight(attribute, hit) {
+  if (!hit) throw new Error('The hit containing the attribute should be provided');
+  if (!{}.hasOwnProperty.call(hit, attribute)) throw new Error(`${attribute} should be a retrievable attribute.`);
```
vvo: Can we use `hit.hasOwnProperty(attribute)` instead of `{}.hasOwnProperty`?

packages/react-instantsearch/src/core/highlight.test.js
```diff
@@ -0,0 +1,104 @@
...
+  it('throws when the attribute is not returned in the hit', () => {
+    expect(highlight.bind(null, 'unknownAttribute', {}))
+      .toThrowError('unknownAttribute should be a retrievable attribute');
+  });
+
+  it('throws when the attribute is not highlighted in the hit', () => {
+    expect(highlight.bind(null, 'notHighlightedAttribute', {notHighlightedAttribute: ''}))
+      .toThrowError('notHighlightedAttribute should be an highlighted attribute');
+  });
+});
```
vvo: missing it('throws.. when there's no hit

packages/react-instantsearch/src/core/highlight.js
```diff
@@ -0,0 +1,64 @@
...
+  const elements = firstValue === '' ? [] : [firstValue];
+
+  firstPassSplit.forEach((split, i) => {
+    const s = split.split(`</${tag}>`);
+    elements.push(
+      React.createElement(
+        tag,
+        {
+          key: `split-${i}-${s[0]}`,
+          className: 'highlighted-value',
```
vvo: Do you think the user need this className?
If yes, could we go for ais-highlightedValue (since we went for camelCase in a recent discussion)
Otherwise, maybe we can remove it, see if someone needs it and then make it an option with a default class name.

packages/react-instantsearch/src/core/highlight.js
```diff
@@ -0,0 +1,64 @@
...
+  const firstValue = firstPassSplit.shift();
+  const elements = firstValue === '' ? [] : [firstValue];
+
+  firstPassSplit.forEach((split, i) => {
+    const s = split.split(`</${tag}>`);
+    elements.push(
+      React.createElement(
+        tag,
+        {
+          key: `split-${i}-${s[0]}`,
```
vvo: I have trouble getting my head around why do we need to add keys?

packages/react-instantsearch/src/core/highlight.js
bobylito: I wished but we have an eslint rules for that. Should we disable this rule?

vvo: Just read the documentation on eslint, we are find disabling this rule as a whole in the .eslintrc.js. We will then backport it in eslint-config-algolia. The reasoning for the rule is not worth the weird syntax they provide as an alternative to me.

packages/react-instantsearch/src/core/highlight.js
bobylito: Because React. Each time you use an array variable as a child of an Element, React will warn you about the keys. Then why? It's because the key identifies the element in the array (not it's position) which gives React the ability to change the DOM in an optimized way (move instead of destroy for example)

packages/react-instantsearch/src/core/highlight.js
bobylito: I think I want this class name and that it's nice to be able to style by class name instead of by tag name :)

Will make the change.

packages/react-instantsearch/src/core/highlight.test.js
bobylito: Ho shoot!

packages/react-instantsearch/src/core/highlight.js
vvo: Understood, I think I was tricked by the fact that I never really did use keys outside some JSX. In JSX it's easier to see you need keys (not saying we need to use JSX here, I am fine).

Also tricked by the keys samples in the tests that includes spaces in it, never used keys with spaces. I guess it's fine :)

vvo:
Argh I forgot @bobylito, we also need a small documentation about it in the Hits doc. And it needs to be exposed in index.js

bobylito:
I added the exposition at the last minute, so might have missed (or I did something wrong). About the doc, when I'll work on the JSDoc stuff, I'll do something about that.

PR #1347, opened 2016-09-28
docs(getting-started): add a getting started guide to react-is doc

bobylito:
This PR is mainly done for providing the getting started. Along the way I had change / tweak a few things (that's why there are other files modified). Amongst the different other changes:
 - The sidebar contains the title of the page
 - The sidebar is smarter and adapts on the content of the navgroup or the document (if either of them contains only one element, this level is discarded)
 - The sidebar is a tad wider

While the content of the getting started is mostly done, I need you:
 - can you review the [content itself](https://github.com/algolia/instantsearch.js/blob/d26ab3986c9c3b0421291c9db33f7e36cf12746d/docgen/src/gettingstarted.md), make sure I talked every important and I didn't go into too much details for others?
 - there is no live examples yet. The getting started of the helper have its samples on codepen, but I'm not sure we want to go this. What do you suggest?


vvo:
> there is no live examples yet. The getting started of the helper have its samples on codepen, but I'm not sure we want to go this. What do you suggest?

As for the live examples, we need to have them also for the widget examples.

One possible solution is to use https://github.com/react-toolbox/react-toolbox/blob/dea14144b8969eca748d4899fb05bc8c52734a31/docs/app/components/preview/index.js (babel-standalone).

Basically we would need to create a small Preview component that would take the code sample (that can be obtained via .innerText on a `<code>` block in DOM) then transform that with babel-standalone, injecting the right context (basically every InstantSearch export).

What do you think?

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
 
-This is the home of `react-instantsearch`, a set of [React](https://facebook.github.io/react/) components to create unique instant-search experiences using [Algolia](https://www.algolia.com/) search engine.
+## Welcome to react-instantsearch
 
-**This is an alpha release, everything may be broken**.
+React-instantsearch is the ultimate toolbox for creating instant search
+experience using React.js and Algolia.
 
-Links:
-- [Chat with us](https://gitter.im/algolia/instantsearch.js) on gitter, we are active and waiting for feedback
```
vvo: Can we keep those links visible somewhere obvious while we are still in alpha release? People need to understand that it's not released, and they need to come to the channel to give/get feedback/answers

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
-- Code is at <https://github.com/algolia/instantsearch.js/tree/v2>, jump in
+In this tutorial, you'll learn how to:
 
+ - add react-instantsearch in your react project
```
vvo: React

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+ - display results from Algolia
+ - add components to filter the results
+ - connect your own component to the search
 
-## Example
+## Before we start
+
+As mentionned earlier react-instantsearch is meant to be used with Algolia.
+
+Therefore, you'll need the credentials to an Algolia index. There are two
```
vvo: Since this getting started is also a getting started tutorial, I would just not leave the choice to the user and force him to use the demo credentials.

While still insisting that he can also create his own account later on.

The goal being, in the getting started, I should not have to make any choices, I just want to try the thing, not to take decisions (at this point)

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+ - `searchKey`, the search key to your application
+ - `indexName`, the name of the index in which you uploaded the data
+
+In case you just want to try the library here are the credentials for
+our sample dataset:
+ - `appId`: `latency` 
+ - `searchKey`: `3d9875e51fbd20c7754e65422f7ce5e1`
+ - `indexName`: `bestbuy`
+
+This guide expects you to have a working react.js project. If
```
vvo: We should stick to "React" as for the React project naming

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+you want to start from scratch, we suggest you use
+[create-react-app](https://github.com/facebookincubator/create-react-app#getting-started).
```
vvo: :+1:

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+
+## Install react-instantsearch
+
+react-instantsearch is available on NPM. To install it you just need
```
vvo: It's npm. I know it seems as being picky but it's a bit like JavaScript javascript Javascript :D

vvo: I would also always link to the according website when that's the first time we encounter "npm" or "React". Then afterwards we are good not linking them.

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
 ```sh
-npm install react-instantsearch --save
+npm install --save react-instantsearch
 ```
 
-app.js
-```js
+## Add the InstantSearch wrapper
+
+The instantsearch wrapper is the component that will connect to Algolia
```
vvo: :+1: simple, to the point

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
 
-import {
-  InstantSearch,
-  SearchBox,
-  Hits,
-} from 'react-instantsearch';
+import {InstantSearch} from 'react-instantsearch';
 
-const App = () =>
+var App = () =>
```
vvo: Can you detail the motivation for switching to `var`?

I would have thought that on a lib like react-instantsearch, we could be fine writing examples/tutorials the way we write code inside the library. Since it's a bit more expected people will: use const already, have a look at our code.

The constraints on code samples may not be as they were on ISV1, where the public skillset and habits was broader.

This would also not force us to context switch our syntax.

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+// Needed only if you're js app doesn't do it already.
+// Create-react-app does it for you ;)
+React.render(<App />, document.querySelector('#app'));
+```
+
+As you can see, there are a few mandatory props. Those props are the
+credentials of your application in Algolia. They can be found in your Algolia
+dasboard.
+
+Congratulations, your application is now connected to Algolia. This wrapper
```
vvo: We feel no excitement, reading "Congratulations, ..". Maybe a small "!" somewhere?

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
-      <SearchBox />
+   // The instantsearch components will go here
+  </InstantSearch>
+
+// Needed only if you're js app doesn't do it already.
+// Create-react-app does it for you ;)
+React.render(<App />, document.querySelector('#app'));
+```
+
+As you can see, there are a few mandatory props. Those props are the
```
vvo: `appId, apiKey and indexName are mandatory props. Those props..` instead of `As you can see` (yes I can :D)

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+ - how to connect a part of a react application to Algolia
+ - configure your credentials with react-instantsearch
+
+ > To get more *under the hood* information about the `InstantSearch` wrapper
+ > component, we have [full documentation](InstantSearch.html) in the widget
+ > section.
+
+## Display the results
+
+The core of a search is to display results. By default, react-instantsearch
```
vvo: of a search experience?

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+will do a query at the start of the page, because the relevance from the
```
vvo: I am not sure there's a link between "We will do a query at the start of the page" and "Because the relevance..". I thought I would read a reason why there's a query at the start of the page.

vvo: I would stop at we will do a query at the start of the page.

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+Algolia engine will always sort your records even though they are not
+filtered yet.
+
+Therefore the next widget we want to use is the Hits widget. This widget will
```
vvo: We will need to provide Links to the documentation of widgets too

vvo: Do we use "widgets" or "components"? In the previous part of the document we used Components.

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+display all the results returned by Algolia, and it will update when there
+are new results.
+
+And while we're doing that, let's create a new component for the search:
+
+```javascript
+// First, we need to add the Hits component to our import
+import {InstantSearch, Hits, SearchBox} from 'react-instantsearch';
+
+// ... We skip the rest of the code here
```
vvo: [...] is enough to convey that we re'use the previous code.

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+  <InstantSearch
+    appId={appId}
+    apiKey={apiKey}
+    indexName={indexName}
+  >
+   <Search/>
+  </InstantSearch>
+```
+
+You should now be able to see the results... in a very raw way. This
```
vvo: I would not use "..." but be direct and say "Now the results are displayed without any styling".

vvo: Ellipsis _can_ be seen as an hesitation from the reader, it's easier to not use them most of the time.

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+custom search related logic, as well as providing them an API to read
+and write the search state. The connectors will have the same name as
+their connected component counterpart.
+
+Here we're going to use the Hits connector, since we want to implement
+a custom rendering for a Hits widget.
+
+ > *TODO*: provide a link to the Hits connector documentation
+
+```javascript
```
vvo: Here, there's no practical need (but for teaching connectors) to talk about connectors, you can use `<Hits itemComponent={} />`

Should we speak about connectors in the getting started? I feel this is too detailed, the need for connectors arise as soon as you want to control the full rendering. But in the getting started I would stick to just discovering the big picture and get a running example in 5 minutes, no more.

While still for example adding "This RefinementList renders a list of links, you could also completely control the rendering and the binding of DOM events for refinements, have a look at our connectors guide".

docgen/src/gettingstarted.md
bobylito: Ok I will add that on top of the tutorial as well as on top of the index.

docgen/src/gettingstarted.md
bobylito: Maybe in the same way, we force our credentials, we should soft-force the use of create-react-app?

docgen/src/gettingstarted.md
bobylito: NP

docgen/src/gettingstarted.md
bobylito: No because it shouldn't :P

docgen/src/gettingstarted.md
bobylito: Sure!

docgen/src/gettingstarted.md
bobylito: That's a tricky one. Our components are components, but they are also connected which, I feel, makes them widget. What about an explanation about that? Introducing the vocabulary?

docgen/src/gettingstarted.md
bobylito: Was wondering. WFM!

docgen/src/gettingstarted.md
bobylito: Yep it's too spoken english for this context.

docgen/src/gettingstarted.md
vvo: Yes but I feel like what you already wrote is conveying this almost exactly. "Either you have an already working app, or use create-react-app.

Detailing create-react-app install more would not help more I feel. What do you meant by "soft-force"?

docgen/src/gettingstarted.md
bobylito: Well like others who will read this documentation I had no clue that this prop existed :P Hence the cumbersome introduction so early :) I will use that, so much more practical :)

docgen/src/gettingstarted.md
bobylito: Linking all the things!

docgen/src/gettingstarted.md
bobylito: 👍

docgen/src/gettingstarted.md
bobylito: Do we need emojis for the millenials?

vvo: :dizzy:

docgen/src/gettingstarted.md
```diff
@@ -6,49 +6,323 @@ nav_groups:
...
+we can only access the hits, using the `hits` prop.
+
+Now that we made our fist custom component, our Search component should be updated
+to use it:
+
+```javascript
+function Search() {
+  return (
+    <div className='container'>
+      <ProductHits /> // << Here :)
```
mthuret: If copy & paste, << will make it fail

mthuret:
I've did the tutorial. I find it easy to follow except that the connector part may come too soon, but this will change. 

Results with the e-commerce data :) 

image

mthuret:
Regarding the live example, it would have been nice to use those we are going to add in storybook. 
We can do that by publishing the storybook and then have en embed iframe using the direct link to a story ?

docgen/src/gettingstarted.md
bobylito: 👍

docgen/src/gettingstarted.md
vvo: I think we can call them components, even if they are connected and sleep well

docgen/src/gettingstarted.md
bobylito: I just slightly modified the text. :)

vvo:
> We can do that by publishing the storybook and then have en embed iframe using the direct link to a story ?

Yep if it works well then why not and is invisible for the user that we are tricking him a bit ;)

But you will first have to be able to create a working /react/storybook to do so.

bobylito:
That would be very cool. And it seems less complex than using the stuff from react-toolbox.

PR #1360, opened 2016-09-29
fix(pagination): reset on other filter + no toggle

bobylito:
The bug is caused by the use of two conflicting semantics
with the same set of changes. Previously we had "if the value for the
page, reset the page number", which led to reseting the pager to 0
if we were selecting the same page (the toggle effect). I previously
removed this behaviour, but it introduced the impossibility to reset
the pager to 0 when changing any other filter (setting the value was
the trigger).

In order to fix this issue I introduce a new type of value to
express the intent of using the same value without changing the page.
This way we have two distinctly defined intents and values:
 - same value as numbers -> reset to 0
 - value with isSamePage flag -> change to page value

It relies on `valuesOf`, which is not that nice and introduces the use
of the dreaded `==`. At least, I can say that I found a legit use case
for it. Bottom line, I'm open to suggestion here :)

### DIFF
--- a/packages/react-instantsearch/src/components/LinkList.js
+++ b/packages/react-instantsearch/src/components/LinkList.js
@@ -24,6 +24,7 @@ export default class LinkList extends Component {
         value: PropTypes.oneOfType([
           PropTypes.string,
           PropTypes.number,
+          PropTypes.object,
         ]).isRequired,
 
         /**
@@ -50,6 +51,7 @@ export default class LinkList extends Component {
     selectedItem: PropTypes.oneOfType([
       PropTypes.string,
       PropTypes.number,
+      PropTypes.object,
     ]),
     onSelect: PropTypes.func.isRequired,
   };
@@ -63,7 +65,8 @@ export default class LinkList extends Component {
             {...applyTheme(
               has(item, 'key') ? item.key : item.value,
               'item',
-              item.value === selectedItem && 'itemSelected',
+              // eslint-disable-next-line
+              item.value == selectedItem && 'itemSelected',
               item.disabled && 'itemDisabled',
               item.modifier
             )}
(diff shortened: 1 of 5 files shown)


Morhaus:
Another way to do this would be to have [`Pagination.connect.refine`](https://github.com/algolia/instantsearch.js/blob/v2/packages/react-instantsearch/src/widgets/Pagination/connect.js#L41) be the identity function when the page hasn't changed.

Then have [`Pagination.connect.transitionState`](https://github.com/algolia/instantsearch.js/blob/v2/packages/react-instantsearch/src/widgets/Pagination/connect.js#L51) also be the identity function in the case where `prevState === nextState`.

bobylito:
@Morhaus I will try that, that's obviously less convoluted :)

bobylito:
> Then have Pagination.connect.transitionState also return prevState unchanged in the case where prevState === nextState.

Humm but then when do we reset the page?

Morhaus:
When `prevState[id] === nextState[id]` and `prevState !== nextState`.

This might mean that we need to have other widget's refine method return the same state object when it would previously yield an equivalent state.

bobylito:
> When prevState[id] === nextState[id] and prevState !== nextState.

Ok I see, that's indeed some big change all around the code base :)

Morhaus:
Yep, but I think it makes sense.

We could also replace the triple equality check with a deep comparison, which would remain cheap provided we don't have too many complicated states. That shouldn't be a problem with the widgets we have right now.

vvo:
reading the PR and the previous implementation I wonder if we are going the right direction.

Basically the need is to reset page to 0 on most refines. So why would the pagination widget be responsible for that? This circles back to how the helper does it to me.

What if the various widgets were made responsible for the page to show and not the pagination connector itself.

Wouldn't it make things obvious and remove the need for "transitionState" API (which only exists because of that page: 0 need). Not very DRY but explicit and simple, before we find a good API to express that (if there's the need, I am not sure).

Morhaus:
So would every widget reset the page state on refine?

This is something I wanted to avoid, because it makes changing the pagination state key difficult.

bobylito:
Or it could be handled by the core like in instantsearch.js / helper.

For now, I would rather have something that does the job (this PR) without having to rewrite the core or go against the current principles of the library. We could rethink that later.

vvo:
> So would every widget reset the page state on refine?

Yes

vvo:
> For now, I would rather have something that does the job (this PR) without having to rewrite the core or go against the current principles of the library. We could rethink that later.

WFM

PR #1375, opened 2016-10-03
docs(styling): add tutorial about styling + renaiming

mthuret:


docgen/src/Styling.md
```diff
@@ -10,58 +10,133 @@ All widgets that render one or more DOM nodes accept a `theme` prop. This prop i
 The different theme keys supported by components are described on their respective documentation page. You can also provide themes using css-modules, Aphrodite, JSS and others if you want.
```
bobylito: It feels odd to me that a tutorial on theming would start with that. What about something more general and more indicative that it is a tutorial? For example:

```markdown
#Welcome to the react-instantsearch theming tutorial

react-instant aims to provide the best to make the best search experience your way! Theming 
is a big part of customising the search experience. That's why react-instantsearch uses
[themeable](link to themeable website) to provides easily customizable widgets. In this tutorial,
we'll cover only css modules theming options with themeable, but it supports much more
options that might fit your current stack better (see their [website]())
```

docgen/src/Styling.md
```diff
@@ -10,58 +10,133 @@ All widgets that render one or more DOM nodes accept a `theme` prop. This prop i
...
+In order to get you started with styling, we are going to use the example you made in the [Getting Started Tutorial](gettingstarted.html) and add some style to it. 
+If you haven't follow the [Getting Started Tutorial](gettingstarted.html) yet, please do and comeback here right after. 
 
-```
-const searchBoxDefaultTheme = SearchBox.defaultTheme;
-```
+In this tutorial, you'll learn how to: 
 
-### Example: Styling a SearchBox
+* Customize a default provided theme.
```
bobylito: Maybe we should introduce the theme concept in react-instantsearch first?

docgen/src/Styling.md
```diff
@@ -10,58 +10,133 @@ All widgets that render one or more DOM nodes accept a `theme` prop. This prop i
...
+* Create a totally new theme for a widget.  
 
-#### Redefining class names 
+Before we start, you need to know that this tutorial will use css-modules as a way to style components. Be aware that you are totally free to use any other way supported by [`react-themeable`](https://github.com/markdalgleish/react-themeable). 
+Please also note that every `react-instantsearch` widgets already have class names and if you want you can use them directly. 
 
-If you want to use pure css on your side, either you can use the generated class names we provide for each widgets, or you can redefine your own. You can also directly use some css. 
+## Modify the SearchBox colors 
 
+The `SearchBox` is the most opinionated widget that you will find in `react-instantsearch`. It's already nice out-of-the-box but the submit button background color may not fit your own theme.
```
bobylito: s/nice/styled/ Nice is a matter of taste :)

docgen/src/Styling.md
```diff
@@ -10,58 +10,133 @@ All widgets that render one or more DOM nodes accept a `theme` prop. This prop i
...
+To change it, let's create a new css file called `searchBox.css`. 
+
+```css
+.submit {
```
bobylito: Maybe we can explain how did we get the info that we needed to change `.submit`? Can I do that using the css files in the project? Or should I do that from the chrome dev tools?

docgen/src/Styling.md
```diff
@@ -10,58 +10,133 @@ All widgets that render one or more DOM nodes accept a `theme` prop. This prop i
...
-export default function ThemedSearchBox(props) {
+Now that we have redefine the background color of the submit button to green, we need to import it. But because we don't want to redefine completely the `SearchBox` theme, we are going to
+use the `mergeClassNames` function provided by `react-instantsearch`. 
+
+```js
+//We need to add mergeClassNames to our import
+import {InstantSearch, Hits, SearchBox, hightlight, RefinementList, Pagination, CurrentFilters, mergeClassNames} from 'react-instantsearch';
+
+//we need to import our custom SearchBox style
+import searchBoxStyle from './searchbox.css';
```
bobylito: We need a loader for CSS files using webpack, right? If so, maybe we could give some information about that? What happens if I use browserify or rollup?

docgen/src/Styling.md
```diff
@@ -10,58 +10,133 @@ All widgets that render one or more DOM nodes accept a `theme` prop. This prop i
...
+      <Pagination />
+    </div>
   );
 }
 ```
 
-If you want to use css-modules, aphrodite and other libraries supported by react-themeable you will also need to use the `theme` prop. Please refer to the [`react-themeable documentation`](https://github.com/markdalgleish/react-themeable)
-to see the correct syntax. 
+You can retrieve the default class names used by a widget by accessing the `defaultClassNames` attribute on a component. 
+The `mergeClassNames` function will create a new object, containing both default and new class names.
```
bobylito: Maybe we can have a link to the function documentation? (even if it's empty for the moment)

docgen/src/Styling.md
```diff
@@ -10,58 +10,133 @@ All widgets that render one or more DOM nodes accept a `theme` prop. This prop i
...
+ 
+Now the searchBox has a green submit button and it has kept the default theme provided out-of-the-box. 
+
+In this part, we've seen the following:
+
+* How to make minor changes to a default theme
```
bobylito: "minor" is a qualifier of the changes which diminishes the value of the task. Why not go for something more neutral like "how to extend / override a default theme"?

docgen/src/Styling.md
```diff
@@ -10,58 +10,133 @@ All widgets that render one or more DOM nodes accept a `theme` prop. This prop i
...
+* How to use the `mergeClassNames` function
+
+## Create a new theme for the RefinementList widget 
+
+The default refinementList theme is pretty raw. Let's recreate a whole theme for it.
```
bobylito: "pretty raw" does not seem very factual. What about just "unstyled"?

docgen/src/Styling.md
```diff
@@ -10,58 +10,133 @@ All widgets that render one or more DOM nodes accept a `theme` prop. This prop i
...
+Here are some suggested guides and reference that you might be interested in:
+ - [Advanced examples using react-instantsearch](examples.html)
+ - [The API for widgets and components](InstantSearch.html)
+ - [All the other guides](guides.html)
 
+## Under the hood
 
+Default styling is done using css-modules. However, you have nothing to import other than the component itself to enjoy our default themes. 
+When building `react-instantsearch` we are using [babel-plugin-transform-inline-localize-css-import](https://github.com/algolia/babel-plugin-transform-inline-localize-css-import).
+We've made this babel plugin in order to be able to embed both class names and styles inside our components.
```
bobylito: Have you tried to integrate this part into the tutorial itself?

docgen/src/Styling.md
```diff
@@ -6,62 +6,174 @@ nav_groups:
...
-The different theme keys supported by components are described on their respective documentation page. You can also provide themes using css-modules, Aphrodite, JSS and others if you want. 
-See [`react-themeable`](https://github.com/markdalgleish/react-themeable) for the underlying implementation and options.
 
-By default, we provide `defaultTheme` for each widgets using css-modules under the hood. Most widgets have empty default theme, but the `SearchBox`, `RangeRatings`, `Range` and `Pagination` widgets 
-are delivered with a built-in one and are ready to use. 
+The different theme keys supported by components are described on their respective documentation page. 
+
+By default, we provide `defaultClassNames` for each widgets but they are mostly not styled. However, the `SearchBox`, `RangeRatings`, `Range` and `Pagination` widgets are delivered with a built-in style and are ready to use without any configuration on your side. 
+
+>Note: default styling in `react-instantsearch` is done using css-modules. However, you have nothing to import other than the component itself to enjoy our default themes.
```
mthuret: I was thinking that maybe for some more advanced topic, we can add them as quote style. It doesn't render nicely yet on the documentation, let me know what you think about it.

mthuret:
New version shipped. 
Let me know what you think. Especially also on the mergeClassNames part. Until we add full inline style support, providing your own class (either with css-module or using a plain css file) will be the only way to extend a theme (otherwise, you need to redefine it completely).

docgen/src/Styling.md
```diff
@@ -6,62 +6,174 @@ nav_groups:
...
+Welcome to the react-instantsearch theming tutorial!
+
+`react-instantsearch` aims to provide the best search experience your way! We believe that theming is a big part when comes the customization of the search experience. That's why react-instantsearch uses [`react-themeable`](https://github.com/markdalgleish/react-themeable) to provides easily customizable widgets. 
+
+In this tutorial, we'll cover only css modules as theming option with `react-themeable`. However, it supports much more options and you may find one that might fit your current stack better ([Check their website](https://github.com/markdalgleish/react-themeable)). 
+
+On top of that, each widgets is shipped with class names that you can use directly if you want. 
+
+In order to get you started with styling, we are going to use the example you made in the [Getting Started Tutorial](gettingstarted.html) and add some style to it. 
+If you haven't follow the [Getting Started Tutorial](gettingstarted.html) yet, please do and comeback here right after.
```
vvo: I am unsure we should force people to follow a certain learning path. My intent here is probably just to learn about the different styling features react-instantsearch provides.

I would remove that part.

docgen/src/Styling.md
```diff
@@ -6,62 +6,174 @@ nav_groups:
 nav_sort: 1
 ---
 
+Welcome to the react-instantsearch theming tutorial!
```
vvo: I am sorry to say it only now, but should all our guide pages be tutorials? Can't they just be simple guide pages with different use cases and explanations?

Tutorial requires a lot more focus and will to follow than a simpler documentation on styling.

If all I want is "Learn more about styling" then it's unlikely I want to follow a complete tutorial?

vvo: It's maybe only a naming issue, tutorial => guide

docgen/src/Styling.md
```diff
@@ -6,62 +6,174 @@ nav_groups:
...
+
+`react-instantsearch` aims to provide the best search experience your way! We believe that theming is a big part when comes the customization of the search experience. That's why react-instantsearch uses [`react-themeable`](https://github.com/markdalgleish/react-themeable) to provides easily customizable widgets.
```
vvo: We do not need to try to sale the library on every page, at some point we need to be a bit more down to earth, direct in our approach so that the user do not feel like "Okai, yeah, I know you are awesome"

Recommendation: drop the salesy speach, go directly to the core of what the user is searching for: styling use cases and solutions.

docgen/src/Styling.md
vvo: css-modules is not part of the public API, I think there's no need to talk about it. Can confuse the user.

vvo:
I read the tutorial, I must say as a user I would not find that the information is easy to grasp. What I would be looking for in a styling guide are answers to: "What are the ways to style or extend the default styles of react-instantsearch". Here I will need to read the whole text and decipher the information that it contains.

I would move from a tutorial style to a more focused guide/documentation on the styling API features we provide. Less text, more code. The getting started guide we have is sufficient as for a guided tutorial, then we can go for a more direct style.

Proposal:

```
# Styling

Default widgets in `react-instantsearch` comes with some styling already applied and loaded.

When styling components, you can decide to either extend or completely replace our default styles, using CSS classes or inline style.

All widget accepts a theme prop and the theme keys are defined in the widget docs.

# CSS classes styling

## Reusing our class names

Show how to style by adding a style.css and redefining our BEM classes

## Providing your own classes

theme={yourClasses}

## Extending the default theme

theme=extendTheme(defaultTheme, yourClasses)

# Inline styling

## Providing inline style

theme={inlineStyles}

## Extending our default theme

theme=extendTheme(defaultTheme, inlineStyles)

# react-themeable

The theme={} props accepts any parameter that react-themeable support...

# CSS modules with webpack

You can use webpack css modules on your side and still extend our themes.

Small example.

# Localized CSS with babel

Default themes of react-instantsearch are built with babel.. You can use it with your own code, and load css with insert-css.
```

I know this is completely different from what you proposed. But I really believe our guides should follow a very down to earth simple use case => solution rather than "follow this tutorial line by line and understand what is the available API from experience".

mthuret:
@vvo Long comment here :) 

Well, I understand your point. It's true that if you don't want to follow a tutorial or you already done it, there's no references to go look after. 

However, when discussing with @bobylito, I really like the idea to have a tutorial for styling that will follow the one of the getting started. It may not be for everyone, but It can be a way to show how you can style our widgets (there's a lot of ways to do it...) based on a simple example. Of course, for now without inline style to perform small customizations, I believe it's not that powerful. But it's definitely a feature that I think we need to add. 

In the end, maybe we can have both: a quick tutorial and a reference guide? 
Do you think it will perform the same kind of jobs twice and this will be useless in the end?

vvo:
Closing while you rework it/rework the styling API. Let's get this right right now

Issue #1339, opened 2016-09-26
Change props auto documentation to simple jsDoc annotations


vvo:
The current props auto documentation tries to infer most of the props documentation from PropTypes definitions by running the code into webpack/babel and then using somehow complex code from a custom loader.

Unfortunately what started as a good idea is a pain to understand, maintain and update.
The current state is that the prop generation is buggy and misses important informations, as seen in #1302.

We could try to fix it but given what's coming next #1303 and #1338, let's make it simple right now.

**I propose** to move everything to very simple jsDoc that will be parseable, don't fear not being 100% DRY. Not being DRY is good if it helps a lot our own documentation flow and understanding of our own documentation generation.

Let's start with something boring but exhaustive and effectiveness that will then give us some time to think about how to really make it better/DRY.

@bobylito I know you worked heavily with jsDoc, if you are up to trying this and removing most of the "plugins/inlineProps" files, go

The plan:
 - [x] POC the doc on Hits
 - [x] Improve the metalsmith jsDoc plugin to be able to read mutliple files
 - [x] Integrate the plugin to generate pages for the Hits component and connect HOC
 - [ ] Reproduce for all the other widgets
 - [ ] Add to every widgets the live example

bobylito:
Challenge accepted (OMG what am I doing??? :) )

vvo:
Awesome, our current documentation is in the end not exhaustive. I dunno if there's a quick win. I was not able to grasp the entire workflow of it while browsing the code.

mthuret:
Once the storybook will be published (when publishing the doc), you can add each widget example by adding an iframe with the following url:

https://community.algolia.com/instantsearch.js/react/storybook/iframe.html?selectedKind=WidgetName&selectedStory=default&dataId=0

vvo:
First step is done, let's iterate in other issues

PR #1376, opened 2016-10-03
docs(storybook): add default style + add storybook to docs build

mthuret:
storybook


bobylito:
The changes are mostly about deleting some content, is it normal?

mthuret:
Nop. It's me forgetting to commit stuff.

It should be ok now

vvo:
How does this integrates with the dev environement right now? I guess it does not, maybe I can help you here. let me know

mthuret:
@vvo What do you mean with "integrates with the dev". Launching storybook with npm start ? 
Currently it does not. If you want to use it, you need to do "npm run storybook".
What's your expectation ?

vvo:
> What's your expectation ?

If I understood well, we are gonna show examples that are iframes linking to /storybook examples? If so, at some point (maybe not now?) we will need to tackle that.

Is that correct?

mthuret:
Include storybook examples within the doc will be done with #1339 👍

PR #1382, opened 2016-10-05
docs(homepage): fix whitespace by using markdown-it

vvo:


vvo:
> Can we move from marked (and the metalsmith plugin) to that? In order to keep just one library per usage (markdown parsing).

Yes I can try to use also marked in metalsmith, won't that trigger any conflicts with what you are doing?

PR #1398, opened 2016-10-07
fix: having hot-reload in dev mode when modifying css files

mthuret:


vvo:
Some context: https://github.com/algolia/babel-plugin-transform-inline-localize-css-import is only good for production builds, while in dev, style-loader and css-loader are the best (hot reload).

So I will need to add my side to this PR: babel for build.

PR #1405, opened 2016-10-09
Better close icon

Kikobeats:
Related: http://stackoverflow.com/questions/5353461/which-font-for-css-x-close-button

Looks equals but is different 😄


vvo:
Thx

Issue #644, opened 2015-11-17
documentation: Add a bower release


pixelastic:
As discussed in https://github.com/algolia/instantsearch.js/issues/593, we should also release the library on bower (even if we do not push it as the main way to get the code).

vvo:
Closing, if you feel like it's absolutely needed for your use case to have a bower release then please reopen.

nicekiwi:
+1

PR #1411, opened 2016-10-12
[WIP] Bower

brianjd:
Decided to take a stab at this for #644.


vvo:
Hi, thanks a lot for the PR. I should have updated the documentation prior to it though because now you can do:

```sh
bower install https://cdn.jsdelivr.net/instantsearch.js/1/instantsearch.js
```

Thanks! (updating the docs now)

PR #1416, opened 2016-10-14
feat(inline styles): add extra keys to allow inline styles styling

mthuret:
After a lot of back and forth this PR include: 

* New theme keys to allow a real usage of inline styles/Aphrodite/And co.
* New method to extend a theme by using only inline style
* A new styling guide 

image

will result with: 

image


vvo:
As a user, can I use CSS modules with webpack (css/style loaders) and use the theme property to provide that theme?

If so in tooling we could just say that yes this is doable, with a simple example code (no webpack, just import theme from css).

mthuret:
Yes It's possible. But for me it belongs to the react-themeable function. And then it's a C/C of theirs docs no ? And why not do that for other ways ?

mthuret:
And I added the fix for the doc in prod.

vvo:
> Yes It's possible. But for me it belongs to the react-themeable function. And then it's a C/C of theirs docs no ? And why not do that for other ways ?

Just after reading the paragraph on babel inline thing, I thought "Why is this page not talking about webpack css modules if it's talking about this?".

Let's merge this and iterate afterwards

PR #1417, opened 2016-10-15
Associate a key with each map element

Kikobeats:
I'm trying to do this:

```jsx
import React from 'react'
import {RefinementList} from 'react-instantsearch'

function CustomRefinementList (props) {
  return (
    {props.items.length &&
      <section className='pa3 pa4-ns'>
        <div className='f6 fw6 ttu tracked pb2 bb b--black-10 black-70'>{props.title}</div>
        <RefinementList {...props} />
      </section>
    }
  )
}

export default RefinementList.connect(CustomRefinementLis
```

It doesn't work because it's in a infinite rendering loop.

Because `RefinementList` and `List` element are generated dynamically, I think that the problem is because is necessary associate a `key` (related: https://facebook.github.io/react/docs/multiple-components.html#dynamic-children)


packages/react-instantsearch/src/components/List.js
```diff
@@ -47,15 +47,15 @@ class List extends Component {
     return extended ? limitMax : limitMin;
   };
 
-  renderItem = (item, parent = null) => {
+  renderItem = (item, key, parent = null) => {
```
Kikobeats: `parent` is used? the function is called with `item, key` all time at  line 58.

vvo:
Hi, thanks a lot!

We did not think of that use case indeed, what you would need to use here is the unconnected component (RefinementList) I guess. We do not expose it today.

I am not comfortable with your patch right now, I feel like we are patching something that is not meant to be used this way. I guess we could just expose the non-connected components for you to reuse like you just did.

What do you think?

Kikobeats:
My current custom implementation of this is based in an example:

```jsx
import React from 'react'
import {Range, RefinementList} from 'react-instantsearch'
import './style.scss'

function SidebarSection ({title, item}) {
  return (
    <article className='pa3 pa4-ns'>
      <div className='f6 fw6 ttu tracked pb2 bb b--black-10 black-70'>{title}</div>
      {item}
    </article>
  )
}

function Facets () {
  return (
    <div data-app='facets-wrapper'>
      <SidebarSection
        title='category'
        item={
          <RefinementList
            attributeName='category'
            id='c'
            operator='or'
            limitMin={10}
          />
        }
      />
```

But this not enough. The point is, when `{item}` render `null`, the wrap around the element is rendered. In this case the output is:

screen shot 2016-10-16 at 01 08 09

This PR was a work around. I suppose that if you expose a mechanism to know if the item will be rendered for be possible associate a correct  wrapper to be rendered with the element will be a better solution.

maybe something like `itemWrapper`? I saw `itemComponent` for a other element in docs.

vvo:
@Kikobeats We are not yet sure on the best way on how to do that and so will hold on this feature for now. You can definitely reimplement your own RefinementList instead of tryin to reuse the `<RefinementList/>` here by using the connector API.

Sorry this is not the best solution we think right now but we don't want to go too fast and keep some space for API additions by getting more feedback and ideas. Thanks a lot for contribution though.

Kikobeats:
Sure, I understand. Just need a fancy way to access to https://github.com/algolia/instantsearch.js/blob/v2/packages/react-instantsearch/src/components/RefinementList.js#L4-L9 for do the same but out of the library. 

I know that styles are exposed under `defaultClassNames`, but exists a way to import the rest?

Kikobeats:
Hmm, I fixed the problem with the imports doing:

```jsx
import React, {PropTypes, Component} from 'react';
import pick from 'lodash/pick';
import themeable from 'react-instantsearch/src/core/themeable';
import translatable from 'react-instantsearch/src/core/translatable';
import List from 'react-instantsearch/src/components/List';

const theme = {
  root:  {},
  item:  {},
  itemSelected:  {},
  itemLabel:  {},
  itemCount:  {},
  showMore:  {},
  showMoreDisabled:  {},
  itemLabelSelected:  {},
  itemCountSelected:  {},
  itemCheckbox: {},
  itemCheckboxSelected: {}
}
```

new problem:

```
warning.js:36 Warning: Failed prop type: Required prop `refine` was not specified in `RefinementList`.
```

Now I don't know what do 🎃

Kikobeats:
oh, I need to connect it!

```
const CustomRefinementList = themeable(theme)(
  translatable({
    showMore: extended => extended ? 'Show less' : 'Show more',
    count: count => count.toLocaleString(),
  })(RefinementListFactory)
);

export default RefinementList.connect(CustomRefinementList);
```

Looks horrible but it's something

vvo:
That's very hacky eheh indeed since none of those are official API entries and might just be removed.

I would go for reimplementing your own small refinement list, since it should be less code than what you did.

I understand that we are not very explicit on the default widgets goals. Basically as soon as you hit a feature wall (can't do X), you need to use connectors.

The Panel question is currently a pain we need to solve

Issue #1305, opened 2016-09-19
Cannot share custom component within Menu.connect and RefinementList.connect


vvo:
We cannot share custom components within Menu.connect and RefinementList.connect. This was a feedback from @GregoryPotdevin here: https://gitter.im/algolia/instantsearch.js?at=57da5121df21a75730fbd208

Menu connectors:

- selectedItem
- items
- refine, expects single value
- createURL expects single value

RefinementList connector:

- selectedItems
- items
- refine, expects array of values
- createURL expects array of values

Every connector rendering a list of items should provide the exact same API for consistency and ease of use.

Something like:

- items[{value, isRefined, name, toggleState}]
- refine(item.toggleState)
- createURL(item.toggleState)
- defaultRefinement

Linked to #1304, what do you think?

GregoryPotdevin:
Easy part : menu connector could provide both `selectedItem` (for ease of use) and `selectedItems` (for compatibility).

What's `toggleState` in your example ?

vvo:
> Easy part : menu connector could provide both selectedItem (for ease of use) and selectedItems (for compatibility).

Easy but already including legacy code :D

`toggleState` would be the precomputed state for every item, see #1304

GregoryPotdevin:
It feels kind of weird to have on each item a field with the state of the other items. Something that feels a bit cleaner (but is probably too verbose) would be to provide a "toggle" function : 
```js
refine(toggle(item.value))
createURL(toggle(item.value))
```

The main advantage of this API is that it's possible to implement "clear" buttons (`createURL({})`, `refine({})`). The "business" logic of how to toggle items would be contained in the `toggle` function. For multi-select, it's a classic toggle, while for Menu toggling an item unselects the previously selected item.

vvo:
@GregoryPotdevin indeed that would work, but what are the obvious differences between

```js
refine(toggle(item.value))
```

and

```js
refine(item.state);
```

?

vvo:
> The main advantage of this API is that it's possible to implement "clear" buttons (createURL({}), refine({}))

This is already the case, you can call refine(null), but we will still provide a clear() method because that's the intent of the user.

GregoryPotdevin:
I'm not 100% confortable with the idea of storing in each item a computed value that depends on other (selected) items but I don't like `refine(toggle(item.value))` either ^^

vvo:
Moved to #1304

PR #1423, opened 2016-10-17
feat(connectors API): consistent default/current refinement naming

vvo:
Goal: provide a more consistent and smaller connector API surface by defining only two way to: define default refinement, access current refinement.

Move from:
- Accessing current refined value: selectedItem, selectedItems, selectedIndex, hitsPerPage (forwarded props)
- Providing default refined value: defaultSelectedIndex, defaultRefinement, defaultSelectedItems, defaultHitsPerPage (connector props)

=> __8 API entries__

To:
- Accessing current refined value: currentRefinement (forwarded prop)
- Providing default refined value: defaultRefinement (connector prop)

=> __2 API entries__

Linked to #1305

Still missing:
- forwarded consistent `items` (+ isRefined)
- forwarded `toggleRefinement` method to easily toggle one refinement
- forwarded `clearRefinement` method to easily remove all refinements

Questions:
I think I will remove the "currentRefinement" forwarded prop once we have proper `items` (with isRefined) and toggle method. Then we will see if people need it, we will add it.

About the "refine" method, it allows to refine multiple values at once,  should we make toggleRefinement works with multiple values just like the helper allows it (if I remember well?) or should we just remove it?

Should we name `items` `refinements`? To be in line with the other API entries? (toggleRefinement, clearRefinement, defaultRefinement)

### DIFF
--- a/docgen/src/examples/e-commerce/App.js
+++ b/docgen/src/examples/e-commerce/App.js
@@ -163,13 +163,13 @@ const ColorItem = ({item, selectedItems, createURL, refine}) => {
   );
 };
 
-const CustomColorRefinementList = ({items, selectedItems, refine, createURL}) =>
+const CustomColorRefinementList = ({items, currentRefinement, refine, createURL}) =>
     <div>
       {items.map(item =>
         <ColorItem
           key={item.value}
           item={item}
-          selectedItems={selectedItems}
+          selectedItems={currentRefinement}
           refine={refine}
           createURL={createURL}
         />
@@ -237,7 +237,7 @@ const CustomResults = createConnector({
                 {value: 'ikea_price_asc', label: 'Price asc.'},
                 {value: 'ikea_price_desc', label: 'Price desc.'},
               ]}
-              defaultSelectedIndex="ikea"
+              defaultRefinement="ikea"
             />
           </div>
           <Stats />
--- a/docgen/src/examples/material-ui/App.js
+++ b/docgen/src/examples/material-ui/App.js
@@ -78,7 +78,7 @@ const Content = React.createClass({
         <div className="Header">
           <AppBar
             title="AMAZING"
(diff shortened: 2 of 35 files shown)


vvo:
cc @kikobeats @harshmaur what do you think of this, is this moving to the right direction? We still have important API changes like this to do to make the API easier to use.

harshmaur:
@vvo I am not sure if I understand it completely. 

When I talk about Refinement, I can only relate to RefinementList(dFR) and Menu(fR). 

Even if we keep the same name, there is a confusion in the way it accepts the props. 

NumericRefinement( string maybe ?), DisjunctiveFacetRefinement(array of items), FacetRefinement (string)

Also I would be pretty confused if hitsPerPage is renamed to default or current Refinement.

vvo:
> When I talk about Refinement, I can only relate to RefinementList(dFR) and Menu(fR).

Yep, still, ultimately all connectors are doing refinements, in a way or another (numeric, faceting, sort). And I would rather not have the user to learn all the different APIS those connectors expose.

> Even if we keep the same name, there is a confusion in the way it accepts the props.

Yes and that will be the only thing you need to know, not the parameter or method naming.

> Also I would be pretty confused if hitsPerPage is renamed to default or current Refinement.

I hesitated on this one indeed. If you come from a previous Algolia experience with instantsearch.js it will be seen as weird. If you are new to Algolia and only learning the connectors API, this should be a consistent experience with all the other connectors.

harshmaur:
> I hesitated on this one indeed. If you come from a previous Algolia experience with instantsearch.js it will be seen as weird. If you are new to Algolia and only learning the connectors API, this should be a consistent experience with all the other connectors.

Makes sense then. But then you have to change this all across plugins. I use python, java and javascript as well as instantsearch.

Kikobeats:
Have sense for reduce namespaces and keywords for refer the same!

PR #1434, opened 2016-10-18
docs: add a guide to explain how to conditionally display content

mthuret:
About #1266

NoResults/EmptyQuery were'nt enough to achieve simple use case like NoResults/HasResults (like in the e-commerce example). You need to quickly go to a custom widgets. So this is a guide to explain how to do it for now.

May be replace by a better solution when our state management will be easier.


bobylito:
Otherwise I like it, it's very hands on :)

vvo:
Awesome, no API addition, great feature

vvo:
FFF I had the window opened for too long and did not see @bobylito comment.

PR #1435, opened 2016-10-18
doc(jsdoc): jsDoc for Component & connector documentation

bobylito:
First step for the way we manage the documentation of the components and the connectors. I'm sure we'll find lots of stuff to add :) (I need to rebase or something :) )


vvo:
Some conflicts

docgen/plugins/navigation.js
```diff
@@ -0,0 +1,32 @@
...
+      const category = data.category || 'other';
+      categories[category] = categories[category] || [];
+      categories[category].push({
+        path,
+        title: data.title,
+        navWeight: data.navWeight,
+      });
+    });
+
+    forEach(files, (data, path) => {
```
vvo: Maybe add a small comment on why do we need to loop twice?

docgen/plugins/onlyChanged.js
```diff
@@ -16,8 +16,13 @@ const layoutFiles = join(__dirname, '../layouts/**/*');
 const cssFiles = join(__dirname, '../src/stylesheets/**/*');
 const CSSEntryPoints = ['index.css'];
 
-const hasChanged = file => Date.parse(file.stats.ctime) > lastRunTime ||
-  Date.parse(file.stats.mtime) > lastRunTime;
+const hasChanged = file => {
+  if (file.stats && file.stats.ctime) {
+    return Date.parse(file.stats.ctime) > lastRunTime || Date.parse(file.stats.mtime) > lastRunTime;
+  } else {
+    return true;
```
vvo: Maybe later in the file can we reuse hasChanged since it now handles the case where there's no stats?

docgen/src/Styling.md
```diff
@@ -1,9 +1,7 @@
 ---
 title: Styling
 layout: guide.pug
-nav_groups:
-  - core
-nav_sort: 1
+category: core
```
vvo: So category: core means it's a guide? Maybe we could express this by using category: guide

docgen/start.js
```diff
@@ -17,6 +17,7 @@ builder({middlewares}, err => {
 
 // then we watch and rebuild
 watch([
+  rootPath('packages/react-instantsearch/src/**/*.js'),
```
vvo: Do you remember why you had to add this line? The devServer is supposed to handle this part (watch webpack build, rebuild, reload)

package.json
```diff
@@ -27,19 +27,18 @@
...
-    "ejs": "2.5.2",
     "escape-html": "1.0.3",
     "eslint-config-algolia": "5.5.1",
     "gh-pages": "0.11.0",
     "google-map-react": "0.19.1",
+    "jsdoc-parse": "1.2.7",
     "lodash": "4.16.2",
     "marked": "0.3.6",
     "material-ui": "0.15.4",
-    "memoizee": "0.4.1",
```
vvo: I guess "react-docgen" is no more needed also

packages/react-instantsearch/src/core/InstantSearch.js
```diff
@@ -182,4 +191,31 @@ class InstantSearch extends Component {
   }
 }
 
+InstantSearch.propTypes = {
```
vvo: Is this needed because of jsdoc parsing limitation? (static vs .)

bobylito:
Yes I know, I'm working on them :P

docgen/start.js
bobylito: Now there is content in the js files, that's why we should trigger the build of the jsdoc (which is in the metalsmith build)

vvo: Ah yes the jsDoc needs to be re-read, got it.

vvo: Previously this was handled by the markdown files updates

vvo: I guess we can remove the line watching markdown files then

docgen/plugins/onlyChanged.js
bobylito: Ok I fixed what went wrong here in the code that checks the availability of stats. I also kept a check on the value used anyway (boyscout rule cc @pixelastic)

packages/react-instantsearch/src/core/InstantSearch.js
```diff
@@ -150,13 +157,13 @@ class InstantSearch extends Component {
     } else {
       return '#';
     }
-  };
+  }
 
-  onHistoryInternalStateUpdate = state => {
+  onHistoryInternalStateUpdate(state) {
     this.aisManager.onExternalStateUpdate(state);
```
Morhaus: How does that work? The `state =>` notation automatically binds to the current instance. In that case, since the `onHistoryInternalStateUpdate` method is passed by reference, it looses its context, so we need to have it bound to the `InstantSearch` instance somehow.

packages/react-instantsearch/src/core/InstantSearch.js
bobylito: yep!

packages/react-instantsearch/src/core/InstantSearch.js
bobylito: It does not. :(

Fixed

docgen/start.js
bobylito: done

packages/react-instantsearch/src/core/InstantSearch.js
```diff
@@ -137,12 +138,22 @@ class InstantSearch extends Component {
   }
 
   getChildContext() {
+    // If not already cached, cache the bound methods so that we can forward them as part
+    // of the context.
+    if (!this._aisContextCache) {
+      this._aisContextCache = {
+        ais: {
+          onInternalStateUpdate: this.onWidgetsInternalStateUpdate.bind(this),
+          createHrefForState: this.createHrefForState.bind(this),
```
Morhaus: Usually you'd just bind/reassign them in the constructor.

```js
constructor(props) {
  /* ... */
  this.onWidgetsInternalStateUpdate = this.onWidgetsInternalStateUpdate.bind(this);
  // etc.
}
```

Is there a particular reason we want to avoid the `method = () =>` construct?

bobylito: Yes but bind is only mandatory because we export those. So this way makes the bind more explicit and localised. I didn't choose to do it the way you propose because I thought that it wasn't needed.

We avoid the assignment construct in class because jsdoc is as of now unable to parse it :(

vvo: Is there an issue on the jsdoc project speaking about this? I am curious.

package.json
```diff
@@ -56,6 +56,7 @@
     "google-map-react": "^0.20.0",
     "insert-css": "^1.0.0",
     "jest-cli": "^16.0.0",
+    "jsdoc-parse": "^1.2.7",
```
vvo: Could we use the latest jsdoc-parse version? (now 2.) Or is there any blocker?

vvo:
This looks good, only worried about having to change our coding style given limitations of the jsdoc parsing. Would be awesome to fix that.

75lb:
hi @vvo .. this is a big PR with a lot going on - what do you need me to look at? What is the question?

vvo:
> hi @vvo .. this is a big PR with a lot going on - what do you need me to look at? What is the question?

Parsing things like https://github.com/algolia/instantsearch.js/pull/1435/files#diff-872aa81dff71f8151ef16821d4e665f4L22 seems not OK with jsdoc-parse@1. @bobylito will provide more context tomorrow if that's not sufficient. Thanks for jumping in!

75lb:
does the source code parse correctly with [jsdoc](https://github.com/jsdoc3/jsdoc)? If not, your issue is with that tool as jsdoc-parse sits downstream..

packages/react-instantsearch/src/widgets/CurrentFilters/index.js
```diff
@@ -1,8 +1,25 @@
+/**
+ * @module widget/CurrentFilters
+ */
 import connect from './connect';
-import CurrentFilters from './CurrentFilters';
+import CurrentFiltersComponent from './CurrentFilters';
```
vvo: Could we say that any capitalized token is a Component otherwise it's just a function/HOC?

I am not sure suffixing Component everywhere brings more context/value to the code?

For sure we need to understand that some component are connected, some are not but do we actually need to express it everywhere?

vvo:
> does the source code parse correctly with jsdoc?

Not sure, still I am curious, from reading the jsdoc-parse I cannot see anywhere we are actually requiring/using the jsdoc module. How does that works? Did you reimplement part of it.

What I was proposing is that since we do not infere anything (it seems) from the code itself but we are just parsing jsDoc comments then there should be no need to actually try to parse the code. Just being able to extract a jsDoc block comment should be sufficient (in our case). And would allow us to keep our coding style.

75lb:
i wrote a simplistic description of how jsdoc2md works [here](https://github.com/jsdoc2md/jsdoc-to-markdown/wiki/How-jsdoc2md-works). The raw jsdoc data is passed into jsdoc-parse, which transforms it into something more useful for passing into handlebars. 

jsdoc2md works downstream of jsdoc. If you only want to parse jsdoc comments (and not source code, as jsdoc does) then your best option is [doctrine](https://github.com/eslint/doctrine).

packages/react-instantsearch/src/widgets/CurrentFilters/index.js
bobylito: Interesting point! Previously we were constructing the widgets in an anonymous way. In every file we had `Connected` as the variable name. Instead now, we are writing down the name of the widget / component as it is documented. So if you look in the code for Hits you will find matches in the index.js as well, the code is more explicit about the intention and, the @name is now inferred from the code.

bobylito:
> Technically we could just do a regexp that extract a jsDoc comment then work from there, if it requires us to just add @name InstantSearch, that's ok I guess. There should not even be a need to entirely parse the JavaScript file.

There is some inference from the code: names and types.

> The parts we had to reformat (.bind(this)) are not used as for the documentation so I hope we can workaround this.

This is part of stage-2, and basically we have this preset only to support this syntax. Can we work with that and move forward? (the only file being impacted is InstantSearch.js).

We can also workaround that by using babel to process the files prior to using jsdoc-parse. Since this is only syntax sugar, that could work.

> If you only want to parse jsdoc comments (and not source code, as jsdoc does) then your best option is doctrine.

doctrine only works on extracted comments which adds yet another step...

package.json
bobylito: Documentation on v2 is very scarce, and I don't get the intent is for this next version as it mentions right away the use of jsdoc2md, but we don't want the markdown transformation https://github.com/jsdoc2md/jsdoc-parse

vvo:
> There is some inference from the code: names and types.

I cannot find where do we infer types from the code, can you detail?

> Since this is only syntax sugar, that could work.

If this allows us to keep the same syntax (Static, bindings) then that would be better yes. Mostly so that the PR is easier to read and so that we know we do not have to code in a particular way for the documentation generation to work.

PR #1439, opened 2016-10-19
docs: change guide about conditional display

mthuret:


docgen/src/conditional-display.md
```diff
@@ -6,15 +6,24 @@ nav_groups:
...
-Let's see some examples:
+Below are some typical use cases when using conditional display makes sense but don't hesitate to read the 
+[making your own widget documentation](Customization.html#creating-your-own-connectors), it will give you more details 
+about the api. 
 
 ## Conditionally displaying content based on query state
 
+The first example where conditional display may apply is: 
+
+**What if you want to display a particular content when the user has not enter a query yet?**
```
vvo: "Displaying content when the query is empty"

vvo: You could just change the title to this ^ then remove the introduction sentences

docgen/src/conditional-display.md
```diff
@@ -6,15 +6,24 @@ nav_groups:
 nav_sort: 1
 ---
 
-Sometimes, you will need to render some widgets instead of another if the query hits no results or if there is currently
-no query used. 
+Sometimes, you will need to render some widgets or content instead of another. To do that you can use the `createConnector`
```
vvo: "Using our connector and createConnector (link) approach, you can conditionnaly display content
based on the search state." should be enough

Direct style.

docgen/src/conditional-display.md
```diff
@@ -31,6 +40,16 @@ const Content = createConnector({
 
 ## Conditionally displaying content based on hits results 
 
+Another example where conditional display may apply is: 
+
+**What if you want to display a particular content when the search made by the user don't return any results?**
```
vvo: Same, replace the title with "Displaying content when there's no results" and remove introduction sentences

docgen/src/conditional-display.md
```diff
@@ -31,6 +40,16 @@ const Content = createConnector({
...
+ 
+By using the [`createConnector`](Customization.html#creating-your-own-connectors) function you may also access the `search` object 
+that holds the search results, search errors and search loading state. By doing so you're able to get the number of hits 
+returned and decide what to do according to this number. 
+
+Let's see an example:
```
vvo: Minor: "Here's an example" seems more common, or just `Example:`.

(Let's see an example google result: About 222,000 results)
(Here's an example: 9M)

:D

vvo:
LGTM

PR #1445, opened 2016-10-19
feat(CurrentFilters): remove `key`, remove `hide`, rename filters to `items`

vvo:
Before:
1. we computed a `key` property that was pretty similar to the `label`
one but different.
2. we injected a `filters: [{label, key}]` with the connector
3. we used hide: true in the searchBox metadata to not display it in
our default component

After:
1. Cannot see the reasoning for this, it works well without it, maybe
in edge cases where different ids are manipulating the same
attributeName. But then, I can't imagine any situation where it would
make sense to display: Category: Movies, Type: Movies since it's the
same attributeName
2. Now injecting `items: [{label}]`, on my way to provide a consistent
connector API (#1423, #1405)
3. hide: true was only used to hide and avoid refining the query but
still give the possibility to users to display and clear the query in
their CurrentFilters connected custom components. Let's wait for that
to happen.

### DIFF
--- a/packages/react-instantsearch/src/widgets/CurrentFilters/CurrentFilters.css
+++ b/packages/react-instantsearch/src/widgets/CurrentFilters/CurrentFilters.css
@@ -1,16 +1,16 @@
 .root {
 }
 
-.filters {
+.items {
 }
 
-.filter {
+.item {
 }
 
-.filterLabel {
+.itemLabel {
 }
 
-.filterClear {
+.itemClear {
 }
 
 .clearAll {
--- a/packages/react-instantsearch/src/widgets/CurrentFilters/CurrentFilters.js
+++ b/packages/react-instantsearch/src/widgets/CurrentFilters/CurrentFilters.js
@@ -9,42 +9,41 @@ class CurrentFilters extends Component {
   static propTypes = {
     translate: PropTypes.func.isRequired,
     applyTheme: PropTypes.func.isRequired,
-    filters: PropTypes.arrayOf(PropTypes.shape({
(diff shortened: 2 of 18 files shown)


Morhaus:
The idea was that `label` could be a React node, while `key` would necessarily be a string, in cases where you want your label to contain more complex markup.

I added a filter to search so that the `ClearAll` component could also clear the search, while the `CurrentFilters` clear all button would clear everything but the search. I believe this behavior is consistent with instantsearch v1?

vvo:
Thanks @Morhaus for clarification.

I think I will go with something that will not just provide a way to create a key but also contextualize the data. So that you can know what type of filter this is (faceting, searchParameter..):

```js
{
  type: 'facet',
  attributeName: '..',
  value: '..', // refinementValue
  label: '..'
}
```

So that connected Component will have more power to decide how to show them. Otherwise for query for instance you had no way to know what the query was.

> I added a filter to search so that the ClearAll component could also clear the search

Not sure I got this right, there's no ClearAll component right now.

Here's my plan:
- remove clear button from currentFilters
- create a ClearAll component and connector
- default ClearAll component will display filters the right way (no query clear)
- connected ClearAll will be able to clear only some filters
- connected CurrentFilters will be able to clear only some filters

> I believe this behavior is consistent with instantsearch v1?

I #failed on this one while you were here, because the goal of react-instantsearch was to get feedback on ISV1 then make it so that the API is better, even if it means not being consistent with ISV1. There's no need to be 100% consistent with ISV1, we just need to be better and keep the overall experience.

Issue #1421, opened 2016-10-16
Improve Range Slider on Mobile


Kikobeats:
desktop:

![](https://d17oy1vhnax1f7.cloudfront.net/items/1w2Y451x020i3w320117/Screen%20Recording%202016-10-16%20at%2011.26%20am.gif)

mobile:

![](https://d17oy1vhnax1f7.cloudfront.net/items/3g3X2z351D3Y2F3U1N0o/Screen%20Recording%202016-10-16%20at%2011.27%20am.gif)

vvo:
Thanks, ultimately we will either bundle an existing slider OR only provide an easy way to connect any slider.

vvo:
moved to #1407

PR #1447, opened 2016-10-20
docs(storybook): add example connecting a Range with Airbnb Rheostat

mthuret:
rheostat-example

Is that enough to "fix" #1421 for the release?


bobylito:
Does it work on mobile @mthuret ?

mthuret:
I can try. Currently the styling of the slider is done by an external css file (but there's a [PR indicating that they will switch to react-with-styles](https://github.com/airbnb/rheostat/pull/35) (their own wrapper for css-in-js 💥 ). I don't know if it can work easily with the babel-plugin of @vvo.

Also, do we want our default impl to be this one? For the release I'll stayed with showing how to connect the Range to an external implementation.

bobylito:
> Also, do we want our default impl to be this one?

I don't know and I guess that's sort of unrelated to the current. Doing the review :)

PR #1452, opened 2016-10-21
fix(searchbox): poweredBy Algolia logo weren't visible in firefox

mthuret:


vvo:
This can be released @mthuret if you want to do a release (npm run release in instantsearch v1 repo, requirement: yarn)

Issue #1286, opened 2016-09-15
When there is no refinement the name of the attribute stays in the URL


bobylito:
Where there is no refinement the attribute stays in the URL.

![attr-url](https://cloud.githubusercontent.com/assets/393765/18548383/adc13c80-7b48-11e6-9650-ddacf9e0cc8b.gif)

mthuret:
While looking to correct this bug, I saw this comment on the RefinementList connector: 

 // Setting the value to an empty string ensures that it is persisted in
      // the URL as an empty value.
      // This is necessary in the case where `defaultRefinement` contains one
      // item and we try to deselect it. `nextSelected` would be an empty array,
      // which would not be persisted to the URL.
      // {foo: ['bar']} => "foo[0]=bar"
      // {foo: []} => ""

@vvo @Morhaus : do you remember the reason for that choice to persist empty value in the URL?

Morhaus:
Let's say you set the default value of a `RefinementList` on the `"animals"` facet to `["cats"]`. When you first load the page, this default value won't be shown in the URL. The only way to deselect the `"cats"` value in this case is to explicitly set the `"animals"` facet value to an empty array. Just removing the entire facet value won't do, because then the default value will be selected again.

mthuret:
Ok got it, thanks @Morhaus. So one thing we could do, is actually building the query without the facets that have an empty string value.

vvo:
> The only way to deselect the "cats" value in this case is to explicitly set the "animals" facet value to an empty array

@mthuret In any way, we just need to keep this behavior, since it's feasible for people to do so, it's a feature (not just a weird edge case)

PR #1453, opened 2016-10-21
fix(url): removed facet were still present in the url with empty value

mthuret:
query-bug

link to #1286


vvo:
To give some more context: while defining the API, we made a lot of hypothesis (bad). One of them was that MAYBE some users would want to set defaultRefinements while letting the user unselect them.

This is an anti pattern. If you want some default refinements while letting the user unselect them then you just link to color=White, yes the default refinement will be present in the url but no big deal.

The main usage (never seen any other usage) of default Refinement being: "I want to preselect this value, and not show the underlying widget, and not have the url being updated with the default refinement".

If a user still wants to show the underlying widget, then it's his responsibility to not allow unselecting default refinements. Maybe we could enforce that ourself in our default widgets.

terranboy:
Hi @mthuret
 why the example here:
https://community.algolia.com/instantsearch.js/react/examples/tourism/index.html
still present in the url with empty value

mthuret:
Hello @terranboy, 

Between the 21 of October and today, we changed our way to provide url synchronisation. Today, if you want to have it, you need to build it yourself because we figured it out that we will never be able to cover all the possible use cases. You can find more information about this on our [dedicated url routing guide](https://community.algolia.com/instantsearch.js/react/guide/Routing.html).

So the reason why empty values are still there on the tourism example is that because the implementation we made for the url synchronisation of our examples after removing it from react-instantsearch doesn't handle them. 

Hope this help :)

PR #1465, opened 2016-10-24
feat(API): new export strategy

### LINKED ISSUE #1454: Export naming strategy
Change our export strategy to:

``` js
import {connectRange, connectSearchBox} from 'react-instantsearch/connectors';
import {RangeInput, InstantSearch} from 'react-instantsearch/dom';
```

As you can see it's now RangeInput and not Range.Input, we should not attach widgets to other widgets namespaces.

This issue might require or lead to file reorganization since we will no more require to have a single index.js exporting the whole connector + different widgets. It could be named exports or better, only connect in our main index.js.

Previous issues: #1303

vvo:
feat(API): new export strategy

BREAKING CHANGE: The way to access connectors (NameOfWidget.connect)
and default widgets (NameOfWidget) has changed. You need to update
your imports.

Now:

```js
import {InstantSearch, SearchBox, RangeRatings} from 'react-instantsearch/dom';
import {connectSearchBox} from 'react-instantsearch/connectors';
import {createConnector} from 'react-instantsearch';
```

Before:

```js
import {InstantSearch, SearchBox, Range} from 'react-instantsearch/dom';
// use SearchBox.connect
// use Range.Ratings
```

fixes #1454

Ultimately we might have different packages. This will also ease usage
in react-native moving forward.


vvo:
Here's what I did:

- moving /src/widgets/Name/Name.js to /src/components/Name/Name.js
- moving /src/widgets/Name/connect.js to /src/connectors/connectName.js
- removing /src/widgets/Name/index.js
- removing /src/widgets/Name/README.md
- creating /connectors.js
- creating /dom.js
- creating /src/widgets/Name.js, where we take components and connect them
- updating examples, doc, stories
- moving jsDoc from /src/widgets/Name/index.js to /src/widgets/Name.js

packages/react-instantsearch/src/widgets/Hits.js
```diff
@@ -1,11 +1,9 @@
-/**
- * @module widgets/Hits
```
bobylito: Good call removing those. I expected to have a use for them, but in the end it's doesn't work.

bobylito: This one is obviously not a request for change :)

docgen/src/Customization.md
```diff
@@ -10,10 +10,10 @@ While react-instantsearch already provides widgets out of the box, there are cas
 
 All default widgets have a corresponding higher-order component that acts as a connector, providing the required props to the widget.
 
-Those connectors are available as the `connect` property on the widgets they connect. For instance, the `SearchBox` widget uses the `SearchBox.connect` higher-order component under the hood in order to retrieve and refine the current query.
+Those connectors are available via `import connectName from from 'react-instantsearch/connectors`. For instance, the `SearchBox` widget uses the `connectSearchBox` higher-order component under the hood in order to retrieve and refine the current query.
```
vvo: I will rephrase it and do an example wit SearchBox, this should make it obvious. Thanks

vvo: Fixed

PR #1469, opened 2016-10-25
docs(storybook): integrate storybook example to our docs

mthuret:
This PR contains the inclusion of the storybook examples inside our doc. 

* I voluntarily let localhost url because our online docs is very old. I thought it might be better for designing to see the actual stuff. Unfortunately, you'll need to run `npm run storybook` on top of `npm run start` for now. Do you think it worth to try to have two url (dev & prod) ? 

This PR also includes: 

* Fix a bug on the Pagination widget.
* Removing duplicate widgets such as RefinementListLinks and so on.
* Some fix about new props name (items, currentRefinement and stuff).

storybook-inside-docs


mthuret:
I just add the switch for url. It should be ok now :)

vvo:
lint error, lets go and fix it in v2 branch (Ill do it)

Issue #1315, opened 2016-09-20
Write the guide content


vvo:
We need a very good guide for react-instantsearch that will explain basic and advanced concepts.

Parts that need to be written:
- [ ] Introduction (what's react instantsearch)
- [ ] Widgets (connected widgets to Algolia)
- [ ] Connectors (HOC providing simple mechanism for controlling rendering)
- [ ] Preselecting values (VirtualWidgets)
- [ ] Configuring Algolia parameters (distinct use case)
- [ ] Wrapping 3rd party libraries
- [ ] Styling (CSS Class names only)
- [ ] Creating a custom menu with manual items (#1062) 
- [ ] Conditionally displaying widgets
- [ ] Url synchronisation (Only show it can be activated/deactivated)

We should leave out parts like creating your own widget, injecting state, passing a history etc. for now because those APIs are unstable and may well change in the future.

bobylito:
What about making a concepts guide to explain the overall architecture? This group several aspects like: 
 - what is instantsearch?
 - widgets
 - connectors

vvo:
All in for an architecture overview item. As long as we are able to keep it simple. Even if we have multiple concepts in the library, once explained those should remain simple to understand.

Then how do we introduce that guide/link it? Just wondering because most of the time I tend to skip any "Architecture" link in a guide. Because I think "Well, that's your job to walk me through the API in a effective way, I don't want to read your architecture details.

In the proposal, we have " What is react-instantsearch" then "Widgets" and "Connectors" Maybe just filling those items without the need to create an - architecture link would better guide people to clicking/learning them?

vvo:
updated the issue with better description on the next steps to be done

mthuret:
What do you think of a structure like this ? 

1. Introduction
   - what's react-instantsearch
   - purposes of widgets
   - purposes of connectors
2. Styling
3. Connectors
   - How to use connectors (half the part that already exists in making your own widget)
4. Wrapping 3rd party libraries
5. Create your own widget
   - create connector api (used for conditional display)
6. Conditionally displaying widgets
7. FAQ
  - How to preselect values (virtual widgets)
  - How to add manual values to a Menu/RefinementList...
  - How to synchronize the url (this link being referenced within the InstantSearch component doc)

Also, do we want to keep the intro page for connectors/components?

vvo:
It's taking shape! Some suggestions:
~~purposes of~~ widgets
[...]

PR #1471, opened 2016-10-25
core(instantsearch-manager): provide a fn for setting SearchParameters

### LINKED ISSUE #1342: How to pass default SearchParameters
With our focus on widget states, we missed one important usecase: being able to define simple SearchParameters that will always be sent.

This goes back to the idea of @bobylito of having a `<Configure/>` widget. For instance we could imagine having such simple `<Configure/>` widget that we pass a state to search parameters function that will allow users to set distinct=true for instance or any other Search parameter.

What do you think?

bobylito:
FIX #1342

I haven't been able to do the tests yet :/ Working on it.


vvo:
Should we also write a guide section on configuring Algolia search parameters like distinct? #1315

packages/react-instantsearch/src/core/InstantSearch.js
```diff
@@ -57,6 +57,9 @@ function validateNextProps(props, nextProps) {
  * @propType {string} appId - The Algolia application id.
  * @propType {string} apiKey - Your Algolia Search-Only API key.
  * @propType {string} indexName - The index in which to search.
+ * @propType {function=identity} configureSearchParameters - function to tweak the parameters sent to Algolia.
```
vvo: What we want to convey is that by default there's nothing being done. By saying the default value is "identity", any user not knowing what identity functions are (like I was at some point) may be confused.

Should we make sure everyone understands what "identity" means here?

packages/react-instantsearch/src/core/createConnector.js
```diff
@@ -4,6 +4,19 @@ import {omit, has} from 'lodash';
 import {shallowEqual, getDisplayName} from './utils';
 
 /**
+ * @typedef {object} ConnectorDescription
```
bobylito: I started exploring the possibility of writing the <Configure /> widget, and didn't have a proper documentation for the object that could be passed to `createConnector`. This is its documentation, and I thought maybe we could keep it since it was already done.

vvo: Unsure, if it currently does not have any output in our documentation website I would say we do not need it for now.

packages/react-instantsearch/src/core/InstantSearch.js
bobylito: That's a pretty good question, I have no idea if this concept is well known or if we should make it more approachable. The nice thing of using `identity` is that it convey perfectly what it means, there is no approximation about its meaning, on the contrary of using words. On the other hand, if you don't know this vocab,  you don't know it...

What do you think @mthuret? :)

bobylito: I added some information about the behaviour of identity. Hope it's enough :D

mthuret: Definitely we should explain what identity means because I'm sure some people will not know, and that can be frustrating. Your explanation is fine, and associated with an example (on the guide you mention @vvo ?), it would be imho nice :)

packages/react-instantsearch/src/core/InstantSearch.js
```diff
@@ -57,6 +57,10 @@ function validateNextProps(props, nextProps) {
  * @propType {string} appId - The Algolia application id.
  * @propType {string} apiKey - Your Algolia Search-Only API key.
  * @propType {string} indexName - The index in which to search.
+ * @propType {object} [searchParameters] - Object containing query parameters to be sent to Algolia.
+ * It will be overriden by the search parameters resolved via the widgets. Typical use case:
```
vvo: I like the "Typical use case"

PR #1473, opened 2016-10-26
feat(api): fix consistency between CurrentFilters and Reset widgets

mthuret:


packages/react-instantsearch/src/widgets/ClearAll.js
```diff
@@ -0,0 +1,4 @@
+import connectCurrentRefinements from '../connectors/connectCurrentRefinements.js';
+import ResetComponent from '../components/ClearAll.js';
```
vvo: ClearAllComponent ?

vvo:
LGTM

PR #1475, opened 2016-10-26
feat(api): remove the range slider implementation

mthuret:


packages/react-instantsearch/src/widgets/Range.js
```diff
@@ -1,4 +1,66 @@
...
+          </li>
+        </ol>
+      </div>
+    );
+  }
+}
+
+ const AirbnbRheostatConnected = connectRange(AirbnbRheostat);
+ */
+export default connectRange(null);
```
vvo: Should we have a console.error() that says "We do not provide a Slider, see documentation for how to do it easily" something like that?

mthuret: yes good idea

mthuret: also, I removed the widget from the namespace.

vvo:
LGTM

PR #1482, opened 2016-10-28
fix(List): disable show more button if no more item to display

mthuret:
showmore-when-needed

### DIFF
--- a/packages/react-instantsearch/src/components/List.js
+++ b/packages/react-instantsearch/src/components/List.js
@@ -80,14 +80,14 @@ class List extends Component {
   renderShowMore() {
     const {showMore, translate, applyTheme} = this.props;
     const {extended} = this.state;
-
+    const disabled = this.props.limitMin >= this.props.items.length;
     if (!showMore) {
       return null;
     }
 
     return (
-      <button
-        {...applyTheme('showMore', 'showMore')}
+      <button disabled={disabled}
+        {...applyTheme('showMore', 'showMore', disabled && 'showMoreDisabled')}
         onClick={this.onShowMoreClick}
       >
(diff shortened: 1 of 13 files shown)


packages/react-instantsearch/src/components/List.js
```diff
@@ -80,14 +80,14 @@ class List extends Component {
...
-
+    const disabled = this.props.limitMin >= this.props.items.length;
     if (!showMore) {
       return null;
     }
 
     return (
       <button
-        {...applyTheme('showMore', 'showMore')}
+        {...applyTheme('showMore', 'showMore', disabled && 'showMoreDisabled')}
```
bobylito: I would compute that in an outer variable that I would then use :) I think that it creates a better readability. Also using && will propagate the value `false` which is probably not what we want.

```jsx
    if (!showMore) {
        return null;
    }

    const disabled = this.props.limitMin >= this.props.items.length;
    const theme = disabled ?
        applyTheme('showMore', 'showMore', 'showMoreDisabled'):
        applyTheme('showMore', 'showMore');
      
    return (
        <button {...theme} >
        </button>
    );
```

mthuret: I tried to stay consistent with our current way to apply conditional theme key. You'll get this in the dom: 
```html
<button class="ais-RefinementList__showMore ais-RefinementList__showMoreDisabled" disabled>Show more</button>
<!-- or -->
<button class="ais-RefinementList__showMore">Show more</button>
```
If we change the way to compute theme here, It would be better to do it everywhere no? And since I believe that we will remove soon the react-themeable part, we would need a new way to compute classnames so maybe it's better to wait?

PR #1483, opened 2016-10-28
feat(InfiniteHits): add an infinite hits widgets with load more

### LINKED ISSUE #1344: Create a "Load more" widget
As a starting point of an infinite scroll documentation/idea, at least having a "Load More" widget seems mandatory.

I am not yet sure on the best way on how to do that in react-instantsearch.

Should we also provide a connector?

bobylito:
☠️ There are some missing stuff and important changes in this PR:
- it lets the connector description function `getProps` access the wrapping component to be able to store the previous results hits
- there are no tests (but I did test the widget by hand)
- it introduce a simple example but it is not meant to be kept
- the load more button is disabled is it is the last page (not in the original proposal)
- since the header in the widget is broken, I commented it in the layout.pug

Next steps: 
- [x] add some tests
- [ ] integrate this widget in another (TBD)
- [x] uncomment the header in layout.pug

FIX #1344


packages/react-instantsearch/src/core/createConnector.js
```diff
@@ -139,7 +139,7 @@ export default function createConnector(connectorDesc) {
         metadata,
       } = store.getState();
       const searchState = {results, searching, error};
-      return connectorDesc.getProps(props, widgets, searchState, metadata);
+      return connectorDesc.getProps.call(this, props, widgets, searchState, metadata);
```
vvo: There's only a single instance of every connectorDesc (connectSearchBox({connectorDesc})) at runtime. Is it a good idea to use context on it and allow storing state on them while it will be shared ultimately? In the connectInfiniteHits that should be OK since we do not expect more than one instance of it.

I am ok with the change, we just have to keep this in mind while we create connectors and allow people creating them.

bobylito: The `this` used in this code should refer to the Element itself. This is the one we want here because we're using an arrow function to bind this method with the instance and it is safe to use it because this is where the state lives as well. So it's like storing data in the `this` of a component.

vvo: :+1:

bobylito:
Tests are done. Unleash the 🐲

vvo:
I have one small comment in that I wonder if "isLastPage" is the right naming, since there's no real pagination here (no 1.2.3..). Maybe:
- noMoreHits
- end

What do you think? (Not maybe the proposals but the isLastPage name)

bobylito:
What about `hasMoreResults` or simply `hasNext`? The `hasNext` is something I have seen in other iterator API (like java).

vvo:
> What about hasMoreResults or simply hasNext? The hasNext is something I have seen in other iterator API (like java). Other propositions:
> 
> hasMoreResults
> hasMoreHits

We settled on hasMore IRL

yarn.lock
```diff
@@ -6697,7 +6697,7 @@ react-inspector@^1.1.0:
     is-dom "^1.0.5"
 
 "react-instantsearch@file:./packages/react-instantsearch/":
-  version "2.0.0-beta.16"
+  version "2.0.0-beta.17"
```
vvo: FWIW: Keep this change. It was triggered by "yarn". I then fixed that in the release process so that it's always updated.

packages/react-instantsearch/src/connectors/connectInfiniteHits.js
```diff
@@ -0,0 +1,104 @@
...
+ * @category connector
+ * @propType {number} hitsPerPage - How many hits should be displayed for every page.
+ *   Ignored when a `HitsPerPage` component is also present.
+ * @propType {Component} itemComponent - Component used for rendering each hit from
+ *   the results. If it is not provided the rendering defaults to displaying the
+ *   hit in its JSON form. The component will be called with a `hit` prop.
+ * @providedPropType {array.<object>} hits - the records that matched the search state
+ */
+export default createConnector({
+  displayName: 'Algolia1000HitsDragon',
```
vvo: => AlgoliaInfinitHits

vvo:
Out of curiosity, to be sure: given that we use the "page" parameter, what happens:
- if you sync the parameter in the URL and hit "reload" or copy paste the url?
- if you do not sync the parameter and hit "reload" or copy paste the url?

yarn.lock
bobylito: I don't get your comment here @vvo

bobylito:
Curiosity killed the cat @vvo 

I expect it will use the page in the url as the starting point and only display a single page. I will integrate this in an example so that we can see what's going on.

vvo:
> I expect it will use the page in the url as the starting point and only display a single page. I will integrate this in an example so that we can see what's going on.

I guess as for infiniteHits example we could force the example to NOT use the page parameter when syncing or creating URLS. So that the behavior is always the same. Stateful (scroll) pagination + infinite hits is something we might not want to try to solve, yet. (Twitter fails at this for example)

bobylito:
@vvo the url is not integrated in the examples. @mthuret is working on it, I'll figure this out (the problems that might arise when reloading the page) when integrated.

Example is done, and changes are done.

I think we're good to merge.

yarn.lock
vvo: This was only to warn if you did not get why this was added.

PR #1484, opened 2016-10-29
Fix minor documentation typos

venyii:
Just read through some of the the new docs in anticipation and found some minor things


vvo:
Its awesome, thanks a lot @venyii

PR #1486, opened 2016-10-31
feat(api): remove usage of theme/extendTheme in our examples

mthuret:
This makes sure we do not provide examples using our old styling api.


storybook/webpack.config.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     loaders: [
       {
         test: /\.css$/,
-        loader: 'style',
+        loader: 'style?insertAt=top',
```
vvo: Is this required? What was the behavior without this change?

vvo:
That's good being consistent

storybook/webpack.config.js
mthuret: In dev, you'll have the css-modules style loaded at the end, which is not good for overriding them by class names. I'll make the change for the docs because I need it for the examples. I did the same with the storybook to stay consistent.

PR #1504, opened 2016-11-03
docs(connectors): add jsdoc to connectors

mthuret:
add props/providedProps to connectors.
Also fixes some typos on widgets docs.

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectMultiRange.js
+++ b/packages/react-instantsearch/src/connectors/connectMultiRange.js
@@ -36,46 +36,31 @@ function getCurrentRefinement(props, state) {
   return '';
 }
 
+/**
+ * MultiRange connector provides the logic to build a widget that will
+ * give the user the ability to select a range value for a numeric attribute.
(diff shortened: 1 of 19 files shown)


packages/react-instantsearch/src/connectors/connectMultiRange.js
```diff
@@ -36,46 +36,31 @@ function getCurrentRefinement(props, state) {
   return '';
 }
 
+/**
+ * MultiRange connector provides the logic to build a widget that will
+ * give the user the ability to select a range value for a numeric attribute.
+ * Ranges are defined statically.
+ * @name MultiRange
+ * @kind HOC
+ * @category connector
```
vvo: Why do we have to say its both a connector and a HOC? Has this any impact on the documentation/website?

mthuret: You need because one is used to create html files, the other for the navigation on the header. Both are probably needed because at some point, we were not very mature on how to organize the doc.

PR #1516, opened 2016-11-07
docs(guide): reorganising the guide section

mthuret:
Contains the reorganization of our guide section following #1315 proposition.

Also includes:
1. Usage of navWeight to make intro page first in their TOC
2. Remove Examples section from connectors API
3. Resizing of the storybook iframe
4. Correcting dead links

I'll let you watch the doc but they are currently some duplications between the Guides Intro Page and the API Intro Page, as well as the intro for the Using Connectors guide. Your feedback on that is more than welcome :)

### DIFF
--- a/docgen/layouts/HOC/connector.pug
+++ b/docgen/layouts/HOC/connector.pug
@@ -52,8 +52,3 @@ block content
             td(colspan=3)=type.description
   else
     p This connector does not provided props to its wrapped Component.
-  h2#example Example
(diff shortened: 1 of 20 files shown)


vvo:
I read it all, LGTM.

I think we will get a better feedback once we can eploy the whole website rather than reading only the code.

PS: Navweight does work in the end, I think @LukyVj tried it at some point

PR #1520, opened 2016-11-08
docs(storybook): use knob array for RefinementList

mthuret:
knobs-array


vvo:
What is knob?

mthuret:
The storybook term for playing dynamically with components

PR #1522, opened 2016-11-08
docs(design): community trigger click, tagline

vvo:
- update the tagline
- little CSS tweaks
- only display and hide the community menu on click

The community menu was triggered with enter/leave and it's painful
because it's very easy to mismove your mouse and get the menu
displayed while you did not want to.

Plus in weird situations it's tricky to handle the case where the
mouse move out the trigger button to the rigth side of the screen and
leave the menu opened.


vvo:
cc @LukyVj

vvo:
Need to add back shortcuts

PR #1524, opened 2016-11-08
docs(sort function): more info on expected function param

olance:
This is a small update of our docs to indicate that the function option for `sortBy` in the menu, hierarchicalMenu and refinementList widgets is the same as the `compareFunction` param for the standard Javascript `Array.prototype.sort`.


vvo:
THX

vvo:
will be live in the coming minutes

PR #1526, opened 2016-11-08
chore(release): separate release from docs update

vvo:
There's now two separate commands:
- npm run release
- npm run docs:update

So that we can update the website without having to do a new release
of the libraries.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6697,7 +6697,7 @@ react-inspector@^1.1.0:
     is-dom "^1.0.5"
 
 "react-instantsearch@file:./packages/react-instantsearch/":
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -6697,7 +6697,7 @@ react-inspector@^1.1.0:
     is-dom "^1.0.5"
 
 "react-instantsearch@file:./packages/react-instantsearch/":
-  version "2.0.0-beta.15"
+  version "2.0.0-beta.16"
```
mthuret: isn't 17 now?

vvo: Yep I think its just a matter of running yarn again, should be fixed next release.

PR #1527, opened 2016-11-09
docs(guides): add a guide explaining how to use react-router with ris

mthuret:
Contains the react-router example within a guide and re-document props on InstantSearch.

history stays undocumented, and the idea is to never expose it (and remove it from the props in the end)

### DIFF
--- a/docgen/src/guides/advanced-topics.md
+++ b/docgen/src/guides/advanced-topics.md
@@ -73,31 +73,75 @@ const ConnectedMenu = connectMenu(props => {
 <ConnectedMenu attributeName="category"/>
 ```
 
-## How to synchronize the url with the search
+## URL Routing
 
-The InstantSearch component features a complete URL synchronization solution. Whenever a widget's state changes, the URL will be updated to reflect the new state of the search UI. This has two main benefits:
+One thing that could be nice to have is whenever a widget's state changes, the URL would be updated to reflect the new state of the search UI. 
+Doing so has two main benefits:
 
 * the user can use the browser's back and forward buttons to navigate back and forth between the different states of the search.
 * the user can bookmark, copy and share a custom search URL.
 
-To activate this feature, you need to pass the `urlSync` props when instantiating the InstantSearch component.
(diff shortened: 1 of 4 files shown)


vvo:
What about we remove the `urlSync` option right away (we are in beta, that's the right time)? In the react-world there is a high probability people will want to use a custom routing solution rather than ours.

The example you created provides a nicer interface and allow more customization easily (like doing `/search/some+products?filters=..` or renaming any parameter).

docgen/src/guides/advanced-topics.md
```diff
@@ -97,7 +97,77 @@ Here's an example
 **Location Debouncing**
  
 Since UI updates can happen in quick succession, for instance when the user types in a query, 
-the new locations are debounced. The treshold prop controls how long the component should wait between 
+the new locations are debounced. The threshold prop controls how long the component should wait between 
 state changes before pushing a new location instead of replacing the old one.
 
+## How to synchronize the url and the state with your own store and routing solution
```
vvo: Would remove the previous topic and call this one "URL Routing" or just "Routing"

mthuret:
Yes we can remove the urlSync props and readd it later if needed. I'll remove the threshold one as well.

docgen/src/guides/advanced-topics.md
mthuret: what about external store? I mentioned them in the guide.

mthuret:
So we're fine with the api contrary to what I said earlier. I updated the example accordingly. Feel free to try it/make your suggestion if you think of an easier way. 

On another PR, what I want to do, is changing our props validation. Basically if a user provides a state we should make sure the onStateChange is their, the createURL remaining optional. But a user must be able to provide an onStateChange function without an associated state. Do we agree on this ?

vvo:
> But a user must be able to provide an onStateChange function without an associated state. Do we agree on this ?

Yes!

vvo:
I like this a lot! We are going the right way in doing this and externalizing state synchronization in the url.

PR #1540, opened 2016-11-14
refactor(history): remove our mode when  we handle history object from the history lib

mthuret:
IS works with only two mode now: controlled or uncontrolled.
-> uncontrolled means that the state is handled by IS
-> controlled means that you need to handle the state on your side.

onStateChange can be now used individually.


packages/react-instantsearch/src/core/InstantSearch.test.js
```diff
@@ -76,7 +70,7 @@ describe('InstantSearch', () => {
           <div />
         </InstantSearch>
       );
-    }).toThrowError('You must provide a `onStateChange` prop alongside the `state`, `createURL` props on <InstantSearch>');
```
vvo: Side note: I do like to always denote any API entry with backticks to ease reading anywhere :p

vvo:
2016-11-14-113927_262x99_scrot

soo good :D

PR #1542, opened 2016-11-14
docs(examples): add url sync

mthuret:
uses browser history and a 700 ms threshold. 

the url sync on our examples does not cover:
* removing empty refinements
* ordering query params by alphabetical order

I can add them if you think it worth it.

### DIFF
--- a/docgen/src/examples/urlSync.js
+++ b/docgen/src/examples/urlSync.js
@@ -1,0 +1,30 @@
+import React, {Component} from 'react';
+import qs from 'qs';
+
+export const withUrlSync = App => class extends Component {
+  constructor() {
+    super();
+    this.state = {searchState: qs.parse(window.location.search.slice(1))};
+  }
+
+  onStateChange = nextState => {
+    const THRESHOLD = 700;
+    const newPush = Date.now();
+    this.setState({lastPush: newPush, searchState: nextState});
+    const search = nextState ? `?${qs.stringify({...qs.parse(window.location.search.slice(1)), ...nextState})}` : '';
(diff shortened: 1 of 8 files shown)


docgen/src/examples/urlSync.js
```diff
@@ -0,0 +1,32 @@
+import React, {Component} from 'react';
+import qs from 'qs';
+
+export const withUrlSync = App => class extends Component {
+  constructor() {
+    super();
+    const history = window.history;
+    this.state = {history, state: {...qs.parse(window.location.search.slice(1))}};
```
vvo: Why do you need to store window.history in the state? For convenience?

mthuret: could totally be replaced with window.history each time yeah (like location)

docgen/src/examples/urlSync.js
```diff
@@ -0,0 +1,32 @@
...
+    } else {
+      this.state.history.pushState(null, null, search);
+    }
+  };
+
+  createURL = state => `?${qs.stringify(state)}`;
+
+  render() {
+    return <App {...this.props}
+                state={this.state.state}
```
vvo: Maybe we could name it `this.state.searchState`?

vvo:
LGTM, ultimately empty refinements entries could just be removed from the state I guess?

mthuret:
Yes I was thinking the same. a new refactor to do ? :]

vvo:
> Yes I was thinking the same. a new refactor to do ? :]

Yep in another step :)

PR #1544, opened 2016-11-14
fix(priceRanges): avoid displaying solo ranges

### LINKED ISSUE #1536: When only one result is returned, priceRanges widget displays bad values instead of none
*bug*

When only one result is returned from Algolia the price ranges widget displays:
pasted_image_at_2016_11_10_02_15_pm

**If the current behavior is a bug, please provide all the steps to reproduce and if possible a minimal [JSFiddle](https://jsfiddle.net/) example or a repository on GitHub that we can `npm install` and `npm start`.**
https://jsfiddle.net/JanPetr/w6vwmer9/1/ - When you search for "a" it keeps the price ranges displayed. When you search for "s", it hides the widget.
There seems to be issue with higher numbers as when "Apple" had price 80, everything worked.

Completely hidden `priceRanges` widget

vvo:
fixes #1536


JanPetr:
🎉

Issue #1455, opened 2016-10-21
Add `disabled` CSS class name when showMore not relevant


vvo:
Following #1428

PR #1545, opened 2016-11-14
fix(ClearAll): always display clearAll button, disabled when no filter

### LINKED ISSUE #1530: ClearAll should be shown deactivated when not relevant, instead of hidden
See https://github.com/algolia/instantsearch.js/issues/1455

bobylito:
**Summary**

ClearAll is hidden when there are no filters. Following on #1455, it can be confusing to have part of the UI appear / disappear, making it disabled is more subtle and is a very common pattern (disabled buttons)

Fix #1530


vvo:
Should we add a CSS disabled class and style it or use the disabled attribute? I have no preference, opening the discussion

packages/react-instantsearch/src/components/ClearAll.enzyme.test.js
```diff
@@ -0,0 +1,36 @@
...
+describe('ClearAll', () => {
+  it('is disabled when there is no filters', () => {
+    const refine = jest.fn();
+    const wrapper = mount(
+      <ClearAll refine={refine} items={[]}/>
+    );
+
+    const btn = wrapper.find('button');
+    expect(refine.mock.calls.length).toBe(0);
+    btn.simulate('click');
```
vvo: BOOM TESTS!

vvo:
As for tests, what about we only test the optimistic case with enzyme (there are items, we can click on the button) and then just use snapshots for other tests? (items, no items)

bobylito:
Previously approved :)

PR #1549, opened 2016-11-15
feat(sortBy): implement default sortBy, remove option

bobylito:
**Summary**

We have chosen to make the widget API simple and let the customisation of behavior for components that will be wrapped in connectors. This sortBy option is now removed.


vvo:
LGTM, will add the BREAKING CHANGE note while merging.

PR #1550, opened 2016-11-15
feat(api): add data to CurrentRefinements connector

mthuret:
final items structure: 

```js
items: [
 {
    label: string (for display),
    attributeName: string, 
    currentRefinement: string, {min/max}, or array
    value: to pass to refine
    items: array optional [{label, value}]
 }
]
```

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectCurrentRefinements.js
+++ b/packages/react-instantsearch/src/connectors/connectCurrentRefinements.js
@@ -8,7 +8,7 @@ import createConnector from '../core/createConnector';
  * @kind connector
  * @category connector
  * @providedPropType {function} refine - a function to remove a single filter
- * @providedPropType {array.<{key: string, label: string}>} items - all the filters, the key for calling the refine prop function, label is for the display.
+ * @providedPropType {array.<{label: string, attributeName: string, currentRefinement: string || object, items: array, value: function}>} items - all the filters, the `value` is to pass to the `refine` function for removing all currentrefinements, `label` is for the display. When existing several refinements for the same atribute name, then you get a nested `items` object that contains a `label` and a `value` function to use to remove a single filter. `attributeName` and `currentRefinement` are metadata containing row values.
  */
 export default createConnector({
   displayName: 'AlgoliaCurrentRefinements',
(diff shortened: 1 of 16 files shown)


packages/react-instantsearch/src/connectors/connectCurrentRefinements.js
```diff
@@ -8,20 +8,22 @@ import createConnector from '../core/createConnector';
...
+          typeof meta.items !== 'undefined' ? res.concat(meta.items) : res
+        , []),
     };
   },
 
-  refine(props, state, filters) {
-    return filters.reduce((res, filter) => filter.clear(res), state);
+  refine(props, state, items) {
+    // `value` corresponds to our internal clear function computed in each connector metadata.
+    const clears = items instanceof Array ? items.map(item => item.value) : [items];
```
mthuret: I'm very not satisfied of this naming. If someone has a better idea, i'll take it!
The tricky part is that either the user pass us an array of items (to remove all of them), or directly a clear function..

mthuret:
I voluntarly didn't spent much time to improve our CurrentRefinements widget. We first need to determine what would be a kick-ass CurrentRefinements widget.

packages/react-instantsearch/src/connectors/connectCurrentRefinements.js
vvo: I do not understand your comment, can you detail a bit more?

mthuret: refine here has one purposes: calling the clear function to remove refinements. 
We hide this to the end user by passing a clear function as value, then he need to use it like this: props.refine(item.value).

However we have two use cases: either we directly have the clear function if it's a single clear, or we get an array of items, and then the idea is to call each clear function to remove all filters at the same time (clearAll behavior).

I'm just unhappy with the naming I found to express this in the code.

packages/react-instantsearch/src/connectors/connectHierarchicalMenu.js
```diff
@@ -80,7 +80,7 @@ function transformValue(value, limit, props, state) {
  * @propType {string} [separator='>'] -  Specifies the level separator used in the data.
  * @propType {string[]} [rootPath=null] - The already selected and hidden path.
  * @propType {boolean} [showParentLevel=true] - Flag to set if the parent level should be displayed.
- * @providedPropType {function} refine - a function to remove a single filter
+ * @providedPropType {function} refine - a function to select a refinement
```
bobylito: Does it toggle the value or just select it?

mthuret: toggle is more accurate

vvo:
Are we good here?

mthuret:
There's still the naming issue I pointed. It might not be so easy to understand the code. But currently I don't have a better idea for it.

packages/react-instantsearch/src/connectors/connectCurrentRefinements.js
vvo: > There's still the naming issue I pointed. It might not be so easy to understand the code. But currently I don't have a better idea for it.

I do not see any BIG issue with it right now, is your naming issue only about "const clears"? Then we can call it "const refinementsToClear"?

PR #1551, opened 2016-11-15
fix(toggle): support negative numeric values for on/off

### LINKED ISSUE #1537: Toggle widget does not refine properly when numeric value is passed to values
*Bug*

When you pass numeric value to values, eg.
```js
values: {            
  on: -2        
}
```
the toggle widget stops working properly. It sends the filter in `facetFilters` parameter with double dot instead of sending it in `numericFilters` parameter with equals.
screenshot 2016-11-10 17 01 53

**If the current behavior is a bug, please provide all the steps to reproduce and if possible a minimal [JSFiddle](https://jsfiddle.net/) example or a repository on GitHub that we can `npm install` and `npm start`.**
https://jsfiddle.net/JanPetr/4rq2zs3h/2/

Proper filtering of results which have the exact value in given attribute. In the fiddle - only products which are on stock.

vvo:
fixes #1537

Summary:

When the user passed values={on: -2, off: 5}, we would send to Algolia as for facet refinements: stock:-2 which would not be what we wanted to express.

Using stock:-2 as a facetRefinement is supported only if you escape
the minus sign, otherwise it's considered as negative facet refinement
(remove "-2" from facets).

Now we escape this sign when it makes sense.

### DIFF
--- a/src/widgets/toggle/implementations/currentToggle.js
+++ b/src/widgets/toggle/implementations/currentToggle.js
@@ -19,106 +19,120 @@ const currentToggle = ({
   containerNode,
   RefinementList,
   cssClasses,
-} = {}) => ({
-  getConfiguration() {
-    return {
-      disjunctiveFacets: [attributeName],
-    };
-  },
-  toggleRefinement(helper, facetValue, isRefined) {
-    const on = userValues.on;
-    const off = userValues.off;
+} = {}) => {
+  const on = userValues ? escapeRefinement(userValues.on) : undefined;
+  const off = userValues ? escapeRefinement(userValues.off) : undefined;
 
-    // Checking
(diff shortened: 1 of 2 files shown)


src/widgets/toggle/implementations/currentToggle.js
```diff
@@ -19,106 +19,120 @@ const currentToggle = ({
...
+          toggleRefinement={this.toggleRefinement}
+        />,
+        containerNode
       );
-    }
+    },
+  };
+};
+
+function escapeRefinement(value) {
```
vvo: This is the interesting part, otherwise it's just reindenting :/

bobylito:
That's a nasty bug :)

Issue #1100, opened 2016-06-21
Hits widget: transformData not always called when refining


timkelty:
When using a `refinementList` (or presumably other sorts of filtering), if a hit exists before and after refinement, `transformData` is NOT called the subsequent time(s).

Original post: http://stackoverflow.com/questions/37924976/instantsearch-js-hits-widget-transforming-data-on-refinement?noredirect=1#comment63313526_37924976

Demo: https://jsfiddle.net/timkelty/LyuL5eze/7/

vvo:
@timkelty Revisiting this issue so basically the hits widget works like that:
- if the hits list is 0
- if the hits list length is different from the previous hits length
- or if some hits are different from the previous list (deep comparison)

Then we do render the hits again and call transformData.

So yes, if the hits list is absolutely the same between two clicks then we do not re-render it.

Can I ask you what you are doing in this formatData that would require the hits to still be re-rendered? Thanks

timkelty:
In this case `formatData` is changing the display of the data based on `refinementList` criteria.
Specifically, choosing which product image is displayed (e.g. user filters "green", and an existing hit still needs to be re-rendered to display the image tagged as "green").

vvo:
Ok so I fixed this issue by removing any check we had on the rendering of hits, it's too tricky to try to optimize that for now.

timkelty:
Thanks, will test.

vvo:
Not yet released, test in one hour

PR #1555, opened 2016-11-15
fix(transformData): always call transformData

### LINKED ISSUE #1538: Bug: Hit not re-rendered when you modify the data in transformData
When making two queries that return the same hit, but this hit is handled differently in transformData between the first and second query, the hit component won't be re-rendered.

Apparently, that comes from the fact that `shouldComponentUpdate` is based on the data retrieved by the API instead of the data sent after `transformData`.

An easy way to reproduce this is to add a timestamp to the hit object in `transformData`, and display it in the component.

vvo:
Summary:
Even when data does not changes, we now call transformData. So that
if the transformed data changes, there will be a new render call.

fixes #1538
linked to #1100

### DIFF
--- a/src/components/Template.js
+++ b/src/components/Template.js
@@ -8,7 +8,7 @@ import hogan from 'hogan.js';
 
 import isEqual from 'lodash/isEqual';
 
-class Template extends React.Component {
+export class Template extends React.Component {
   shouldComponentUpdate(nextProps) {
     return !isEqual(this.props.data, nextProps.data) || this.props.templateKey !== nextProps.templateKey;
   }
@@ -22,7 +22,7 @@ class Template extends React.Component {
       templateKey: this.props.templateKey,
       compileOptions,
       helpers: this.props.templatesConfig.helpers,
-      data: transformData(this.props.transformData, this.props.templateKey, this.props.data),
+      data: this.props.data,
     });
 
     if (content === null) {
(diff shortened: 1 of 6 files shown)


src/components/Template.js
```diff
@@ -137,4 +137,15 @@ function transformHelpersToHogan(helpers, compileOptions, data) {
   );
 }
 
-export default Template;
+// Resolve transformData before Template, so transformData is always called
+// even if the data is the same. Allowing you to dynamically inject conditions in
+// transformData that will force re-rendering
+const withTransformData =
```
vvo: This is the interesting part

src/widgets/refinement-list/__tests__/refinement-list-test.js
```diff
@@ -187,17 +179,6 @@ describe('refinementList()', () => {
       createURL = () => '#';
     });
 
-    it('formats counts', () => {
```
vvo: This is not the place to test this, applied helpers are already tested in Template-test

src/components/Template.js
bobylito: Have you considered using alternative implementations? What about use transformData in willReceiveProps and store the result in the state?

vvo:
> Just wondering if we could have not used a HOC for that.

Indeed this complexifies a bit the code, I will have a look at your idea using the state, wonder if any drawbacks.

PR #1557, opened 2016-11-15
fix(connectors): remove itemComponent from hits connectors

### LINKED ISSUE #1554: connectHits should not talk about itemComponent
When using a connector we are already providing the mean to provide something like itemComponent. Removing it will reduce the API surface and make it less awkward to use.

bobylito:
**Summary**

Removes the itemComponent that doesn't make sense on the connector (only on the widget).

fix #1554


vvo:
LGTM

PR #1564, opened 2016-11-16
fix(id): remove id props

### LINKED ISSUE #1556: Remove id props
The id props was used to let a user choose a custom key for the internal map that holds the values of filters of each widgets. This was used in the URL sync system that we no longer have. Let's remove it!

bobylito:
> The id props was used to let a user choose a custom key for the internal map that holds the values of filters of each widgets. This was used in the URL sync system that we no longer have. Let's remove it!

fix #1556

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectHierarchicalMenu.js
+++ b/packages/react-instantsearch/src/connectors/connectHierarchicalMenu.js
@@ -3,8 +3,10 @@ import {PropTypes} from 'react';
 import createConnector from '../core/createConnector';
 import {SearchParameters} from 'algoliasearch-helper';
 
+export const getId = props => props.attributes[0];
+
 function getCurrentRefinement(props, state) {
-  const id = props.id;
+  const id = getId(props);
   if (typeof state[id] !== 'undefined') {
     if (state[id] === '') {
       return null;
@@ -72,7 +74,6 @@ const sortBy = ['name:asc'];
  * @name connectHierarchicalMenu
  * @kind connector
  * @category connector
- * @propType {string} id - URL state serialization key. The state of this widget takes the shape of a `string`, which corresponds to the full path of the current selected refinement.
  * @propType {string} attributes - List of attributes to use to generate the hierarchy of the menu. See the example for the convention to follow.
  * @propType {string} defaultRefinement - the item value selected by default
  * @propType {boolean} [showMore=false] - Flag to activate the show more button, for toggling the number of items between limitMin and limitMax.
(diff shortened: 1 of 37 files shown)


docgen/webpack.config.start.babel.js
```diff
@@ -12,7 +12,7 @@ const {
 
 export default {
   ...webpackConfig,
-  devtool: 'eval-source-map', // a lot faster than 'source-map', not ok for production though
+  devtool: 'source-map', // 'eval-source-map', // a lot faster than 'source-map', not ok for production though
```
vvo: Did you change it for ease of debugging locally? I had the struggle too.

packages/react-instantsearch/src/connectors/connectHierarchicalMenu.js
```diff
@@ -133,7 +133,12 @@ export default createConnector({
   },
 
   getProps(props, state, search) {
-    const {id, showMore, limitMin, limitMax} = props;
+    const {attributes, showMore, limitMin, limitMax} = props;
+
+    if (typeof attributes === 'undefined' || !attributes.length)
+      throw new Error('attributes props should at least have one element');
```
vvo: Do we need to do this check since in dev mode propTypes would catch it?

packages/react-instantsearch/src/connectors/connectInfiniteHits.js
```diff
@@ -5,6 +5,10 @@ import {PropTypes} from 'react';
 
 import createConnector from '../core/createConnector';
 
+function getId() {
+  return 'pager';
```
vvo: Why pager instead of p? Do you want to handle the case where the user might use infiniteHits along with a normal pagination widget?

mthuret:
id prop should also be removed from hierarchicalMenu.stories.js

docgen/webpack.config.start.babel.js
bobylito: yes I forgot to remove it :(

packages/react-instantsearch/src/connectors/connectHierarchicalMenu.js
bobylito: It won't catch the cardinality. I need at least one element, unless we can express that with propTypes.

packages/react-instantsearch/src/connectors/connectInfiniteHits.js
bobylito: I didn't think that much on that topic. My thinking here is that those ids are for internal use so I didn't think it would matter...

bobylito:
I didn't check the stories :( On it!

packages/react-instantsearch/src/connectors/connectInfiniteHits.js
vvo: The id here is "public" API, it's available for url synchronisation via onStateChange() that's why I asked. Ultimately we might want to go for easy to understand IDS since now those are not directly synced in the URL (in another iteration)

bobylito: True, will update that to page. I will make that consistent with the pagination then.

packages/react-instantsearch/src/connectors/connectHierarchicalMenu.js
bobylito: It doesn't check the number of items, but it still better! I will move to checking with proptypes.

bobylito:
Waiting for the test to pass on Travis CI. I think we should be good otherwise.

PR #1570, opened 2016-11-16
fix(transformData): default data is an object when not provided

vvo:
This fixes the nasty regression bug we have seen and reimplement the
transformData fix

### DIFF
--- a/src/components/Template.js
+++ b/src/components/Template.js
@@ -8,7 +8,7 @@ import hogan from 'hogan.js';
 
 import isEqual from 'lodash/isEqual';
 
-class Template extends React.Component {
+export class Template extends React.Component {
   shouldComponentUpdate(nextProps) {
     return !isEqual(this.props.data, nextProps.data) || this.props.templateKey !== nextProps.templateKey;
   }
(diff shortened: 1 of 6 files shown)


src/components/Template.js
```diff
@@ -137,4 +137,17 @@ function transformHelpersToHogan(helpers, compileOptions, data) {
...
 }
 
-export default Template;
+// Resolve transformData before Template, so transformData is always called
+// even if the data is the same. Allowing you to dynamically inject conditions in
+// transformData that will force re-rendering
+const withTransformData =
+  TemplateToWrap =>
+    props => {
+      const data = props.data === undefined ? {} : props.data; // eslint-disable-line react/prop-types
```
bobylito: Well it fixes the problem and is similar to what we were doing already.

vvo: I think it happens in currentRefinedValues where you can easily `create` items via attributes and those might not yet be refined.

vvo:
Added a test to ensure we default to {}

bobylito:
The test looks fine :)

PR #1573, opened 2016-11-17
feat(react-native): make react-instantsearch compatible for native dev

mthuret:
Contains:

- new native namespace
- guide for react-native with the source code of an example showing searchbox+infinitHits (the api of the connector is really simple to use, thanks @bobylito !)

react-native-example

### DIFF
--- a/docgen/src/guides/advanced-topics.md
+++ b/docgen/src/guides/advanced-topics.md
@@ -51,7 +51,7 @@ To do that, you can use the appropriate connector and manually add your values.
 
 Here's an example to force category values for a Menu.
 
-```javascript
+```jsx
 const ConnectedMenu = connectMenu(props => {
     const items = [];
     
@@ -90,7 +90,7 @@ This function while provided to every widgets and connectors is only useful if y
 
 Here's an example showing you how to use [react-router](https://github.com/ReactTraining/react-router) with react-instantsearch. 
 
-```javascript
+```jsx
 import React, {Component} from 'react';
 import {
     InstantSearch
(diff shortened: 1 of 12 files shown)


docgen/src/guides/advanced-topics.md
```diff
@@ -153,3 +153,127 @@ class App extends Component {
 export default withRouter(App);
 ```
 
+## React Native
```
vvo: Given the length of the topic, I propose (after this PR) that we rename advanced section into tutorials and give them their own pages. What do you think?

packages/react-instantsearch/dom.js
```diff
@@ -1,4 +1,8 @@
-export {default as InstantSearch} from './src/core/InstantSearch.js';
+import InstantSearch from './src/core/InstantSearch';
+import forBrowser from './src/core/InstantSearchDom';
```
vvo: Would name the file forBrowser.js so that we understand this is a HOC

packages/react-instantsearch/native.js
```diff
@@ -0,0 +1,5 @@
+import InstantSearch from './src/core/InstantSearch';
```
vvo: Don't know *exactly* how but I would say we should be able to do something "better" in term of DRY (YAY!):

```js
import createInstantSearch from './src/core/createInstantSearch';
import algoliaClient from 'algoliasearch/native';

export const InstantSearch = createInstantSearch({algoliaClient});
```

And reuse all the code around appId/client creation.

What do you think? This is feasible technically

docgen/src/guides/advanced-topics.md
mthuret: Yes I agree but then the How-To topics are smaller so it might be a little too much for those no?

packages/react-instantsearch/native.js
mthuret: Yes you're right, I can pass the algoliasearch as a HOC props.

In the meantime, do we need to keep this dual mode between algoliaClient with appid/appkey already provided, and the one that we instantiate ourself.
Can we just ask for the algoliasearch dependency and then do the algoliasearch(appId, appKey) all the time?

docgen/src/guides/advanced-topics.md
vvo: We will see but as soon as you have 15 lines of code you are good for a single page

packages/react-instantsearch/native.js
vvo: Not sure to understand what you propose as for changes

packages/react-instantsearch/src/core/InstantSearchDom.js
```diff
@@ -0,0 +1,22 @@
+import React, {Component, PropTypes} from 'react';
+import algoliasearch from 'algoliasearch';
+
+export default function forNative(InstantSearch) {
```
mthuret: No more forNative/forBrowser just a createInstantSearch :)

packages/react-instantsearch/src/core/InstantSearch.js
```diff
@@ -136,11 +133,9 @@ class InstantSearch extends Component {
 
 InstantSearch.propTypes = {
   // @TODO: These props are currently constant.
-  appId: PropTypes.string,
-  apiKey: PropTypes.string,
   indexName: PropTypes.string.isRequired,
 
-  algoliaClient: PropTypes.object,
+  algoliaClient: PropTypes.object.isRequired,
```
mthuret: Right now no yet :) But appId/appKey are now passed to the upper component :)

vvo:
LGTM!

PR #1584, opened 2016-11-20
chore(dev): tune dev env for 200ms updates

vvo:
Before this commit, editting a single CSS file or markdown guide took
8s on my computer.

Now it takes 200ms.

It's ugly as a fix, it would need to be baked in. Having a good static
site builder AND a very nice dev experience on it is key (but hard) to
ease the contribution/iteration.


packages/react-instantsearch/src/components/InfiniteHits.js
```diff
@@ -1,6 +1,3 @@
-/**
- * @namespace widgets.Hits
```
bobylito: I'm not sure you wanted to delete this one.

vvo:
For now I will remove all @namespaces addition because they are not used and not up to date in the docgen.

Still, I will open an issue about making it obvious in the docs as for where to import things (our export strategy basically)

vvo:
did not find any other @namespace, either I had already removed them :blush: or I don't know

PR #1588, opened 2016-11-22
feat(unmount): clean state if a widget is unmounted except if it is persistent

mthuret:
Clean the state when a widget is unmounted except if the widget is marked has persistent (via a props). 

Needs review about the way to achieve the clean:
-> I did it forcing the user to define an id inside the connector (instead of having it outside). 
It avoids the repetition of the clean up code and in a way clarify the need of an id (when there's a refinement). But it also complexify the user api and some widgets code (you need to pass the id a lot of time). Also it hides from the user how the widget cleanup is done. 

An alternative from Vincent would be to let the user do the cleanup using the same interface as the refine method (you got the whole state, and return the whole state). The cleanup function would be called when the widget unmount. 

What do you think?

### DIFF
--- a/docgen/src/examples/urlSync.js
+++ b/docgen/src/examples/urlSync.js
@@ -11,13 +11,13 @@ export const withUrlSync = App => class extends Component {
   onStateChange = nextState => {
     const THRESHOLD = 700;
     const newPush = Date.now();
-    this.setState({lastPush: newPush, searchState: nextState});
-    const search = nextState ? `?${qs.stringify({...qs.parse(window.location.search.slice(1)), ...nextState})}` : '';
+    const search = nextState ? `?${qs.stringify(nextState)}` : '';
     if (this.state.lastPush && newPush - this.state.lastPush <= THRESHOLD) {
       window.history.replaceState(null, null, search);
     } else {
       window.history.pushState(null, null, search);
     }
+    this.setState({lastPush: newPush, searchState: nextState});
   };
 
   createURL = state => `?${qs.stringify(state)}`;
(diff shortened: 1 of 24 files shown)


vvo:
I am down for using a cleanup method instead rather than introducing IDs and making the state manager aware of those IDS. We may need this in the future but right now I don't think this is needed.

The reasoning behind using a cleanup method like this:

```js
cleanup: function(state) {
  return omit(state, attributeName);
}
```

rather than having to specify an ID in a connector is that it's more in line with the way connectors need to handle the state today. They need to handle it fully (in refine methods)

I see IDs and automatic cleaning of IDs a bit magical and doing things in the background (in state manager) that are sometime being done in the foreground (in the connector, in refine).

vvo:
Something we did not discuss: should a new search be triggered when unmounting the widget? I would say yes: the cleanup must be complete, state and search parameters.

If you agree we will have to change the way it's done here, you may need two  methods.

bobylito:
I think that having a cleanup function is giving too much power for something that should be "automatic". Even though propagating the concept is not ideal, I think it's still a better solution.

vvo:
Before we take "shortcuts" because we foresee it will be easier and should be automatic. I would like us to first experiment with it and see how it goes, how users are using it. I am not even sure people will actually ADD to the state if we do enough connectors.

Just not confident about doing anything "automatic" for now. I would rather force us to be explicit for things like state manipulation.

mthuret:
=> use a cleanUp function called when unmounting a widget. 

Removed the isPersistent behavior, hard to do it properly for now and not sure if its needed yet.

vvo:
> Removed the isPersistent behavior, hard to do it properly for now and not sure if its needed yet.

:+1: Ultimately the usecase "I have a full state I want to be applied" can be solved today with VirtualRefinements (https://community.algolia.com/instantsearch.js/react/guides/advanced-topics.html#how-to-preselect-values-using-virtual-widgets) and later maybe with an easier API entry

PR #1589, opened 2016-11-22
docs: add react-router small demo app

mthuret:
I decided to remove the code sample from the guide and just add the link pointing to the demo app. 
Let me know if you think we need both.

Also, can someone just perform a yarn install in the demo directory and tell me if the react-instantsearch directory under node_modules is correct? (without any /dist, /.storybook...etc)?

### DIFF
--- a/docgen/src/guides/advanced-topics.md
+++ b/docgen/src/guides/advanced-topics.md
@@ -88,77 +88,11 @@ To be able to do that, you will need to provide three props to the [InstantSearc
 * createURL(state): this function is needed for every widgets that will render a link. It expects a string in return.
 This function while provided to every widgets and connectors is only useful if you are in a browser context. 
 
-Here's an example showing you how to use [react-router](https://github.com/ReactTraining/react-router) with react-instantsearch. 
-
-```jsx
-import React, {Component} from 'react';
-import {
-    InstantSearch
-} from 'react-instantsearch/dom';
-import {withRouter} from 'react-router';
-import qs from 'qs';
-
-class App extends Component {
-   constructor (props) {
-       super(props);
-       // we initialize the state by parsing the url
-       this.state = {state: {...qs.parse(this.props.router.location.query)}}; 
-    }
-    
-    /*
-    Update the state when the back/forward button is used
-     */
-    componentWillReceiveProps(){
-        this.setState({state: qs.parse(this.props.router.location.query)});
-    }
-    
-    /*
-    Push the new state to the react-router history. The threshold is there 
-    to specify how long we should wait between state changes before pushing 
-    a new location instead of replacing the old one. This is a very basic 
-    implementation you might want to perform advanced behaviors like removing 
-    empty values from the url or being able to keep others query params. 
-    */
-    onStateChange = (nextState) => {
-        const THRESHOLD = 700;
-        const newPush = Date.now();
-        /*
-        the current state is saved to be given as the 
-        state props of InstantSearch root component
-        */
-        this.setState({lastPush: newPush, state: nextState});
-        if (this.state.lastPush && newPush - this.state.lastPush  <= THRESHOLD) {
-            this.props.router.replace(nextState ? `?${qs.stringify(nextState)}` : '');
-        } else {
-            this.props.router.push(nextState ? `?${qs.stringify(nextState)}` : '');
-        }
-    };
-    
-    /*
-    This is the function that will be provided to every widgets and connectors. 
-    It allows you to be able to create a link. 
-    */
-    createURL = (state) => {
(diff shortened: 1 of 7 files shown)


packages/react-instantsearch/examples/react-router/.gitignore
```diff
@@ -0,0 +1,15 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
```
vvo: This should be already handled by the root gitignore no?

packages/react-instantsearch/examples/react-router/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "react-router-with-is",
```
vvo: I think you can remove name and version

packages/react-instantsearch/examples/react-router/public/index.html
```diff
@@ -0,0 +1,30 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <!--
```
vvo: Would remove all the big comments etc

packages/react-instantsearch/examples/react-router/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "lodash": "^4.16.6",
+    "qs": "^6.3.0",
+    "react": "^15.3.2",
+    "react-autosuggest": "^7.0.1",
+    "react-dom": "^15.3.2",
+    "react-instantsearch": "^2.0.0-beta.17",
+    "react-router": "^3.0.0"
+  },
+  "scripts": {
+    "start": "react-scripts start",
```
vvo: Would only leave the necessary parts

docgen/src/guides/advanced-topics.md
```diff
@@ -88,77 +88,11 @@ To be able to do that, you will need to provide three props to the [InstantSearc
...
-            </InstantSearch>
-        );
-    }
-}
-
-export default withRouter(App);
-```
+It exists many ways to manipulate the browser history. Here are some you could use:
+* Using [the browser history api](https://developer.mozilla.org/en-US/docs/Web/API/History_API)
+* Using a javascript history library
```
bobylito: Maybe provide an example here?

vvo: `JavaScript`

packages/react-instantsearch/examples/react-router/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "private": true,
+  "devDependencies": {
+    "react-scripts": "0.7.0"
+  },
+  "dependencies": {
+    "lodash": "^4.16.6",
+    "qs": "^6.3.0",
+    "react": "^15.3.2",
+    "react-autosuggest": "^7.0.1",
```
vvo: How is autosuggest used? Cannot find it

packages/react-instantsearch/examples/react-router/src/App.js
```diff
@@ -0,0 +1,99 @@
+import React, {Component} from 'react';
+import {
+    InstantSearch, HierarchicalMenu,
+    Hits, Menu, Pagination, PoweredBy, RangeRatings,
+    RefinementList, SearchBox, ClearAll,
+} from 'react-instantsearch/dom';
+/* eslint-disable import/no-unresolved */
```
vvo: Seems like we have some eslint issues here, we could solve them I'll have a look afterwards

packages/react-instantsearch/examples/react-router/src/App.js
```diff
@@ -0,0 +1,99 @@
...
+
+  render() {
+    return (
+            <InstantSearch
+                appId="latency"
+                apiKey="6be0576ff61c053d5f9a3225e2a90f76"
+                indexName="ikea"
+                state={this.state.state}
+                onStateChange={this.onStateChange.bind(this)}
+                createURL={this.createURL.bind(this)}
```
vvo: Indentation should be two spaces, weird there's no error

docgen/src/guides/advanced-topics.md
mthuret: You mean adding an real browser history api example (like pointing to the file used in our docs examples) or re-adding a code sample in the guide?

packages/react-instantsearch/examples/react-router/src/App.js
mthuret: Yes I added that because it was failing on travis saying that he couldn't find react-router module (but everything went perfectly fine locally).

PR #1598, opened 2016-11-24
docs: multiple instantsearch instances

mthuret:


docgen/src/guides/synchronize-multiple-instantsearch.md
```diff
@@ -0,0 +1,89 @@
...
+
+class App extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {searchState: {}};
+  }
+
+  onStateChange = (nextState) => {
+    //we merge both InstantSearch search state in one
+    this.setState({searchState: {...this.state.searchState, ...nextState}});
```
vvo: I think the main usecase is to show hits from a different index, so there's no need to READ the state of the second instance since there's no clear usecase on why we would like to read the changes of an instance that we use only to display, not allowing refinements.

I would take the opportunity to remove the constraint of passing both `state=` and `onStateChange`.

What do you think?

mthuret: wfm. let's do that.

mthuret:
The example works but i'm currently after fixing the fact that the usage of state/onChangeState ends up by performing two queries (same behavior on our url routing examples).

packages/react-instantsearch/src/core/createConnector.js
```diff
@@ -105,7 +105,7 @@ export default function createConnector(connectorDesc) {
         props: this.getProps(nextProps),
       });
 
-      if (isWidget) {
+      if (isWidget && !isEqual(nextProps, this.props)) {
```
mthuret: This is the optimization to avoid making a new search if new props are the same as the old props.

docgen/src/guides/using-multiple-instantsearch.md
```diff
@@ -0,0 +1,87 @@
...
+title: Using multiple InstantSearch instances
+layout: guide.pug
+category: guide
+navWeight: 0
+---
+
+Having several instances of `InstantSearch` can be useful when: 
+
+* You want to make a search on multiple indices 
+* You want to synchronize multiple `InstantSearch` instances
```
vvo: "Having several instances of `InstantSearch` can be useful" => "You want to synchronize multiple `InstantSearch` instances" is not easy to understand.

I would rather say: "
You can use multiple `<InstantSearch/>` instances for cases like:
- displaying hits from different indices,
- while sharing a single SearchBox,
- or any use case involving synchronizing widgets between different `<InstantSearch>` instances
"

docgen/src/guides/using-multiple-instantsearch.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Using multiple InstantSearch instances
```
vvo: Multiple InstantSearch could be sufficient as a title

docgen/src/guides/using-multiple-instantsearch.md
```diff
@@ -0,0 +1,87 @@
...
+* You want to synchronize multiple `InstantSearch` instances
+
+To be able to do that, we are going to take advantage of two props on the [InstantSearch root component](/component/InstantSearch.html).
+
+* onStateChange(nextState): a function being called every time the `InstantSearch` state is updated. 
+* state: an object that is the current state of InstantSearch
+
+The idea is to have a main component that will receive every new state of the first instance and then pass it back to each `InstantSearch` instances.  
+
+Be aware that just having a refinement inside the `InstantSearch` state doesn't apply it automatically to the search.
```
vvo: Would not use "be aware" since the user does not know it first.

"Refinements and parameters of an `InstantSearch` state needs to have corresponding widgets, or VirtualWidgets, added to be effectively applied".

docgen/src/guides/using-multiple-instantsearch.md
```diff
@@ -0,0 +1,87 @@
...
+
+* onStateChange(nextState): a function being called every time the `InstantSearch` state is updated. 
+* state: an object that is the current state of InstantSearch
+
+The idea is to have a main component that will receive every new state of the first instance and then pass it back to each `InstantSearch` instances.  
+
+Be aware that just having a refinement inside the `InstantSearch` state doesn't apply it automatically to the search. 
+To apply the refinement, you will need to use a [virtual widget](/guides/advanced-topics.html#how-to-preselect-values-using-virtual-widgets). 
+
+Here's an example showing you how to make the same search on two different indices:
```
vvo: Here's an example displaying hits from two different indices

docgen/src/guides/using-multiple-instantsearch.md
```diff
@@ -0,0 +1,87 @@
...
+category: guide
+navWeight: 0
+---
+
+Having several instances of `InstantSearch` can be useful when: 
+
+* You want to make a search on multiple indices 
+* You want to synchronize multiple `InstantSearch` instances
+
+To be able to do that, we are going to take advantage of two props on the [InstantSearch root component](/component/InstantSearch.html).
```
vvo: Two props can be used to inject state or be notified of state changes:

docgen/src/guides/using-multiple-instantsearch.md
```diff
@@ -0,0 +1,87 @@
...
+import {connectSearchBox} from 'react-instantsearch/connectors'
+
+class App extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {searchState: {}};
+  }
+
+  onStateChange = searchState => {
+    // we merge both InstantSearch search state in one
```
vvo: comment no more needed

docgen/src/guides/using-multiple-instantsearch.md
```diff
@@ -0,0 +1,87 @@
...
+    this.setState({searchState});
+  };
+
+  render() {
+    return (
+    <div>
+      {/* Those two `InstantSearch` instances could also have been nested in each other,
```
vvo: You can also nest `<InstantSearch>` components, as long..

packages/react-instantsearch/src/core/createConnector.js
vvo: So we are fixing a bug somehow here? Do you think we can add a test that checks the number of times search is called? I would rather try to fix that in another PR also because this one can be focused only on InstantSearch multi guide + removing the constraint

vvo:
LGTM

PR #1607, opened 2016-11-25
Fix cast string into number

Kikobeats:


vvo:
Nice @Kikobeats! What was the underlying bug here (and the effects?), can you provide a way to reproduce it so we can write a bug fix?

Kikobeats:
Yes, basically, when it try to calculate the new state, because the number is serialized the sum is wrong:

```
p=0
p=01
p=011
p=0111
```

expected (and the thing that this issue resolve)

```
p=0
p=1
p=2
p=3
```

vvo:
I think I get it, since the state can be serialized in the URL, when you read it back p is now a string even if it contains a number?

Kikobeats:
exactly

vvo:
Are you willing to add a test for this? Otherwise I'll do it.

Kikobeats:
Better if you do it and I learn for the next time

vvo:
moved to #1619

PR #1612, opened 2016-11-28
fix(search): don't update widgets if props hasn't changed

mthuret:


vvo:
Can you remove the commit about docs on multiple instances?

mthuret:
@vvo: I was on it ;)

vvo:
Could you walk us briefly around this bugfix? When what this happening?

mthuret:
Bug: when managing the state of InstantSearch externally, several identical queries were sent to Algolia. 

Root cause: The componentWillReceiveProps hook of React gets called for both the InstantSearch root component and the widgets. But since we don't check for the props equality we're doing a new search even if their props are still the same.

packages/react-instantsearch/src/core/createConnector.js
```diff
@@ -105,7 +105,7 @@ export default function createConnector(connectorDesc) {
         props: this.getProps(nextProps),
       });
 
-      if (isWidget) {
+      if (isWidget && !shallowEqual(this.props, nextProps)) {
```
vvo: Right before we update the state, do we need to update the state if props did not change?

PR #1615, opened 2016-11-28
fix(Pagination): when page === 1, first should not be selected

vvo:
Before this commit, when on first page, the "First" link was selected
AND disabled.

Ultimately it's overkill to have LinkList.js since Pagination is the
only one using it, should we remove this layer of abstraction?


mthuret:
We should remove this layer of abstraction if it simplifies the readability of the Pagination source code. 

Otherwise LGTM.

vvo:
> We should remove this layer of abstraction if it simplifies the readability of the Pagination source code.

It should since we compute "disabled" and "selected" using different semantics, not very nice

PR #1616, opened 2016-11-28
[WIP][POC] feat(ga-widget): new GA widget which allows users to plug instant sea…

JanPetr:
**Summary**

Hey guys, we are working with @Shipow on easy way how to plug instant search pages to GA as currently there is no easy way and a lot of people are struggling with it.

So I create a simple widget for IS which easily allows people to plug IS to their Google Analytics and analyze their search data there in relations with conversion and other stuff...

**Result**

New widget with easy API: 
```js
search.addWidget(
  instantsearch.widgets.ga()
);
```

Two optional configuration are present: 
- `delay` - time between last search key stroke and push to GA
- `triggerOnUIInteraction` - if the push to GA should happen after any click or reload of website even before `delay` is out 

This is a **POC** and my first widget, so I'm definitely not rock solid about it. It's missing tests and few checks right now. I'm pushing this PR to get your feedback on it about if it's the right way to go, how to structure the code in better way and how to inject the `click` and `beforeunload` events correctly.

Thank you guys for any help. Cheers.

### DIFF
--- a/dev/app.js
+++ b/dev/app.js
@@ -29,6 +29,10 @@ search.addWidget(
 );
 
 search.addWidget(
+  instantsearch.widgets.ga()
+);
+
+search.addWidget(
   instantsearch.widgets.stats({
     container: '#stats',
   })
--- a/src/lib/main.js
+++ b/src/lib/main.js
@@ -24,6 +24,7 @@ import sortBySelector from '../widgets/sort-by-selector/sort-by-selector.js';
 import starRating from '../widgets/star-rating/star-rating.js';
 import stats from '../widgets/stats/stats.js';
 import toggle from '../widgets/toggle/toggle.js';
+import ga from '../widgets/ga/ga.js';
 import version from './version.js';
 
 const instantsearch = toFactory(InstantSearch);
@@ -45,6 +46,7 @@ instantsearch.widgets = {
   starRating,
   stats,
   toggle,
+  ga,
 };
 instantsearch.version = version;
 instantsearch.createQueryString = algoliasearchHelper.url.getQueryStringFromState;
--- a/src/widgets/ga/ga.js
+++ b/src/widgets/ga/ga.js
@@ -1,0 +1,115 @@
+/**
+ * Pushes analytics data to GoogleAnalytics
+ * @function ga
+ * @param  {int} [options.delay=3000] Number of milliseconds between last search key stroke and push data to GA
+ * @param  {boolean} [options.triggerOnUIInteraction=true] Trigger push to GA after click on page or redirecting the page
+ * @return {Object}
+ */
+function ga({
+  delay = 3000,
+  triggerOnUIInteraction = true,
+} = {}) {
+
+  let lastSentGa = '';
+  let sendAnalytics = function(state, results) {
+    let params = [];
+
+    params.push(serializeRefinements(Object.assign({}, state.disjunctiveFacetsRefinements, state.facetsRefinements)));
+    params.push(serializeNumericRefinements(state.numericRefinements));
+
+    params = params.filter(function(n) {
+      return n != '';
+    }).join('&');
+
+    let paramsToSend = 'Query: ' + state.query + ', ' + params;
+
+    if(lastSentGa !== paramsToSend) {
+      if(typeof window.ga === 'undefined') {
+        throw new Error('Google Analytics are not present in the website');
+      }
+
+      // console.log({'event': 'search', 'Search Query': state.query, 'Facet Parameters': params, 'Number of Hits': results.nbHits});
+      window.ga('set', 'page', '/search/query/?query = ' + state.query + '&params= ' + params + '&numberOfHits=' + results.nbHits);
+      window.ga('send', 'pageView');
+
+      lastSentGa = paramsToSend;
+    }
+  };
+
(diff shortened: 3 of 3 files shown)


vvo:
Nice! Couple of questions:
- what is the actual behavior/willing behavior of the widget: what will it do in the end, browser wise, GA wise. What does it provides? I am a complete beginner with GA and I do not know the intents/goals/use case. This will help us identify the potential users.
- why do you use a recurring setTimeout loop? What do you want to monitor?

JanPetr:
Out of the box GA tracks visited pages on it's load / refresh. So it has a hard time measure IS pages because there are no refreshes and page is re-rendered on each key stroke with JavaScript.
GA (or any other analytics tool)  provides a way how to track those re-renders by sending pageViews programatically from JavaScript. But a lot of our customers struggle on how to plug it to IS, what queries and what data send to GA.

And this is issue what this widget solves. It pushes only relevant queries and relevant data to GA and make it easy for our customers to integrate GA to their IS pages. So they can analyse searches in their GA dashboard, connect searches with their conversions, analyse how many users use search on their website, what are the conversion rates for each search query etc. 

Does it make sense? :)

About the timeout - we don't want to send each keystroke to GA to reduce the noise of data. So that's why there is a 3s (by default) timeout and the push to GA is made with latest rendered data when the search is probably finished. Is there a better way to do it?

-------------------

Today I got a great help from @bobylito and we refactored the widget a lot. And in the end it won't be `ga` widget but `analytics` widget and it'll support more analytics services then just GA. Currently proposed API is:

```js
search.addWidget(
  instantsearch.widgets.analytics({
    services: ['ga', 'segment.io', 'gtm'],
    configuration = {
      'ga': {...},
      'segment.io': {...}
    }
  })
);
```

I'd like to push it this afternoon.

WDYT @vvo?

vvo:
Do we need to actually support multiple services (and possibly miss somes) or should we make it so that the widget exposes the necessary events/endpoint and it's easy to plug any analytics service into it?

APIs like "services: ['ga', 'segment.io', 'gtm']" are tricky to handle on the long run because we make ourselves dependant on their own APIs/configuration APIs.

If it makes sense, I would prefer having something that exposes the necessary API entries making it really easy to plug analytics services to search state. The rest being only documentation/guide for our users.

JanPetr:
You're right @vvo, great point!

How about?

```js
search.addWidget(
  instantsearch.widgets.analytics({
    pushFunction: function(query, parameters, nbHits) {
      // Do whatever you want - use API of your analytics services 
      // and we'll make sure they are pushed in the right time    
      
      // For example GA:

      window.ga('set', 'page', '/search/query/?query = ' + query + '&params= ' + params + '&numberOfHits=' + nbHits);
      window.ga('send', 'pageView');
    }, 
    // other configs like delay etc...
  })
);
```

So we won't be responsible for maintaining 3rd parties APIs and we just make sure the data are pushed in the right time with correct data.

Looks better?

cc @bobylito @Shipow

Shipow:
LGFM

bobylito:
I don't know the power of GA so it was attractive to have support of it out of the box without having prior knowledge. But I guess that those are no tools for amateurs, so the benefit is pretty thin to have it packaged. On the other hand being able to support services by default is great 👍 

Is the way the refinements are currently serialized a standard way? Might we encounter people with opinions on that matter? (like we do with URL's)

About the time out scheme, is a delay of 3 seconds could be an issue for the tracking (which is now shifted by 3 seconds)? If a user do actions that last for less than 3 seconds, we will receive no tracking. Is it an issue? Should we have something more aggressive? Do something smarter that forces some updates?

JanPetr:
Having services already integrated would be great until certain time. I tend to agree on this with @vvo, as in long term it'd be pain to maintain.
And we have to remember power-users who already have their GA configured and when we provide them prepared service, it might have some limitations. For example now we push GA pageView with URL `/search/query/?query=' + query...` but a lot of customers can have configured GA to handle search with URL `/?q=' + query...`. So using widget with already integrated services might cause them issues and additional effort in the implementation unless we give them all options as configurable (and I'm afraid we cannot handle all).

For me the best way is to expose them `pushFuction` as proposed and handle different services in guide/tutorial/documentation. And we won't be limited by integrated services and when somebody wants to use other one - feel free :)

Right now we serialise parameters as we think is good. Not the best option :D But what we can do is to expose whole `helper` inside the `pushFunction` so power-users can get any data they want. Would it be a good idea?

Timeout 3 seconds is default for now and is configurable. In the end, we can always reduce default value to 2 seconds or less?
The case when user do action before 3 seconds are handled by `click` and `beforeunload` event listeners. Do you think it's not enough?

bobylito:
> Having services already integrated would be great until certain time. I tend to agree on this with @vvo, as in long term it'd be pain to maintain.

Erf sorry for the misleading comment earlier. I agree with both of you :) The push function gives the ability to handle any service, and people who want to use it know what they are doing.

The timeout issue is not much about the 3 seconds but the image you get on the GA side. Do you want to have an idea of the kind of actions the user has taken before actually going to a product? I don't think the timeout brings much. It might never send any information before the actual trigger of beforeunload or click. Maybe the part that has the timeout is no even necessary. I'm wondering because I'm not a GA user :D

Issue #1581, opened 2016-11-18
Define and implement final shape of the state


vvo:
- [x] separate facet names from the root of instantsearch state to avoid ID collisions
- [x] document instantsearch state in a GUIDE
- [x] explain where to find state of a widget in instantsearch state in every widget doc
- [x] stop using minified tokens (p, q), this is the responsibility of state synchronisation

```js
{
  page: 0, // any widget handling a single parameter that is a direct mapping to an Algolia REST parameter should put the parameter at root level
  hitsPerPage: 10, // same here
  query: 'a', // same here, we use `q` I believe today, this needs to be changed, there's no reason to use minified parameters in the full state shape
  menu: {
    attributeName: 'refinement'
  },
  hierarchicalMenu: {
    attributeName: 'refinement'
  },
}
```

mthuret:
@vvo, @bobylito: Here is an example of state with all of our connectors used. Is everything ok for you? 

```json
{
  "range": {
    "price": {
      "min": "200",
      "max": "500"
    }
  },
  "refinementList": {
    "colors": [
      "White",
      "High gloss gray",
      "Black-brown",
      "High gloss white",
      "Walnut effect"
    ]
  },
  "hierarchicalMenu": {
    "category": "Bathroom"
  },
  "hitsPerPage": "2",
  "menu": {
    "type": "High cabinet with mirror door, black-brown"
  },
  "multiRange": {
    "price": "100:500"
  },
  "toggle": {
    "materials": "on"
  },
  "sortBy": "index_name",
  "query": "chair",
  "page": 2
}
```

vvo:
For a first iteration, this looks good. We will iterate if needed on it. go!

mthuret:
It seems that only `toggle` as a state that differs from the format asked for `defaultRefinement`. 
state is `on` or `off` and `defaultRefinement` is a simple boolean. 
Do we want to change that?

vvo:
> It seems that only toggle as a state that differs from the format asked for defaultRefinement.
> state is on or off and defaultRefinement is a simple boolean.
> Do we want to change that?

Yes this is not consistent right?

vvo:
State for toggle should be 1 or 0 too, not "on" "off", "on" and "off" could be non relevant token givent the attribute name manipulated.

PR #1627, opened 2016-11-30
feat(api): add namespace when storing widgets state

mthuret:
Does not contain the doc yet, only the namespacing.

```json
{
  "range": {
    "price": {
      "min": "200",
      "max": "500"
    }
  },
  "refinementList": {
    "colors": [
      "White",
      "High gloss gray",
      "Black-brown",
      "High gloss white",
      "Walnut effect"
    ]
  },
  "hierarchicalMenu": {
    "category": "Bathroom"
  },
  "hitsPerPage": "2",
  "menu": {
    "type": "High cabinet with mirror door, black-brown"
  },
  "multiRange": {
    "price": "100:500"
  },
  "toggle": {
    "materials": "on"
  },
  "sortBy": "index_name",
  "query": "chair",
  "page": 2
}
```

### DIFF
--- a/docgen/src/examples/e-commerce-infinite/App.js
+++ b/docgen/src/examples/e-commerce-infinite/App.js
@@ -206,7 +206,7 @@ const CustomResults = createConnector({
 
   getProps(props, state, search) {
     const noResults = search.results ? search.results.nbHits === 0 : false;
-    return {query: state.q, noResults};
+    return {query: state.query, noResults};
   },
 })(({noResults, query}) => {
   if (noResults) {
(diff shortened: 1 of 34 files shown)


mthuret:
Added the first documentation attempt.

Also see this https://github.com/algolia/instantsearch.js/issues/1581#issuecomment-263883196.

vvo:
Do we need to update our examples synchronizing the state into the url and any example reading the state?

vvo:
In the future (noted as a task to discuss), we will have to think about an easy way/API to sync/parse state in the url so that urls are not too long and easily shareable.

Issue #1614, opened 2016-11-28
Separate examples styles from docs header styles. 


mthuret:
Since we added our doc header in all of our examples page, styles are collapsing.

We need to find a proper way to isolate both styles, otherwise, each time someone will change the design of our docs it could break our examples.

vvo:
How do you want to do this @mthuret? Do you mean styles are being merged because there's no proper namespacing?

mthuret:
Styles are being merged mainly because we are using bootstrap on both our docs website and our examples. 

We only need the header on our example, so one solution would be to split our documentation css file and only use the needed one on our example page.

PR #1628, opened 2016-11-30
docs: isolate header from other documentation style

mthuret:
Our style examples should be safe now.


vvo:
Can you point or explain the part that effectively avoid merging styles from header in examples as described in #1614? Thanks

mthuret:
I created an header.scss file to separate the header styles from the rest of the documentation styles. Then our examples are now only using this file to avoid conflict. 

It works fine, except for the material-ui one (there is conflict with z-index/webkit- properties) but for the others, it seems to be enough.

PR #1632, opened 2016-12-01
feat(theme): move to CSS file and CSS class names only theming

vvo:
Remove any reference to react-themeable, theme={} props. Load the
Algolia theme when relevant.

2016-12-01-152218_1116x744_scrot

BREAKING CHANGE:
- CSS is no more injected by default, read our styling guide to know how
to load it
- react-themeable, theme={} prop have been removed from the codebase,
the only way to style widgets is now to use CSS class names


docgen/src/examples/tourism/style.css
```diff
@@ -50,7 +50,7 @@ body {
   color: #DCE0E0;
 }
 
-.ais-SearchBox__root {
+body.example .ais-SearchBox__root {
```
vvo: removing this, should not be needed

LukyVj:
:boom:

vvo:
Tests ..
...are
 OK...

:shipit:

PR #1645, opened 2016-12-02
refactoring(highlight): refactor highlighting into a widget and connector

bobylito:
**Summary**

This PR adds a new Highlight widget for the react-dom users and a connector for other flavours of react. The underlying parsing function now transforms the highlighted attribute into a  intermediary agnostic structure: `[{value: string, isHighlighted}, {value: string, isHighlighted}, ...]`

**Result**

The highlighter can now be used as a component which is more convenient and we can easily use it for react-native. The widget and the connector are documented, and the getting started and examples are up to date.

Side note: I also updated the connector layout to conditionally add examples if some exist.


vvo:
Since the implementation is not using:
- fullyHighlighted
- matchedWords
- matchLevel

Can we remove those tokens from the tests so that the reader will not wonder if those have any impact in the code?

packages/react-instantsearch/src/core/highlight.test.js
```diff
@@ -1,101 +1,112 @@
 /* eslint-env jest, jasmine */
-import React from 'react';
+import {parseAlgoliaHit, parseHighlightedAttribute} from './highlight.js';
```
vvo: Since only parseAlgoliaHit is actually used externally by other files, should we export it from the file and test it in isolation? Making it an export can induce the reader in error by thinking this is actually used by other modules while it's not.

I would go for:
- export default function highlight
- only test highlight()

What do you think?

bobylito: > export default function highlight

There is a benefit in having both function namely exported: it explicitly state what are each functions. Also there is not real hierarchy between them. So I would stay like that, especially because this is not an API to be used outside of the library.

> only test highlight()

Having two functions allows for finer grained tests and makes responsibility of each functions  more explicit. If tests on `parseHighlightedAttribute` do not fail but tests on `parseAlgoliaHit` fail, we know that the error is on `parseAlgoliaHit` for sure. Otherwise it means that we have to analyze both functions to know the source of the error.

bobylito:
> Can we remove those tokens from the tests so that the reader will not wonder if those have any impact in the code?

The tests should not make any assumptions concerning the implementation. As far as the reader is concerned, the function takes an Algolia response which includes those parameters.

mthuret:
If some data declared under the "given" part of a test are not relevant for the "when" part, I think it's better to remove them. It's just noise that sometimes can lead to a futur misunderstanding.

bobylito:
Removed the extra data.

packages/react-instantsearch/src/core/highlight.test.js
vvo: There should be few cases where testing a "private" method (here parseHighlightedAttribute) should be really necessary. By private I mean a function only used in one file. I would prefer not having to expose or test this function in isolation and rather only test the wrapping method which is parseAlgoliaHit.

parseHighlightedAttribute is not used outside of highlight.js. parseAlgoliaHit is checking for bad usage of parseHighlightedAttribute.

Exporting this function and testing it makes it harder to refactor the code in highlight.js: you cannot tell easily if parseHighlightedAttribute is part of the API or used by other files, you will have to search in all files to understand that it's exported to be able to test it.

vvo:
We have conflicts since merged #1652, I am responsible for that I can resolve them in the PR if you want.

packages/react-instantsearch/src/core/highlight.js
```diff
@@ -1,67 +1,75 @@
...
- *       <span className="hit-name">{hightlight('name', hit)}</span>
- *     </div>
- *   );
+ * @param {string} preTag - string used to identify the start of an highlighted value
+ * @param {string} postTag - string used to identify the end of an highlighted value
+ * @param {string} pathToAttribute - path to the highlighted attribute in the results
+ * @param {object} hit - the actual hit returned by Algolia.
+ * @return {object[]} - An array of {value: string, isDefined: boolean}.
+ */
+export default function parseAlgoliaHit({
```
vvo: If file is highlight.js we should rename export to `export default function highlight`. We do this in other modules I believe.

packages/react-instantsearch/src/components/Highlight.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+export default function Highlight({hit, attributeName, highlight}) {
+  const parsedHighlightedValue = highlight({hit, attributeName});
+  const reactHighlighted = parsedHighlightedValue.map((v, i) => {
+    if (!v.isHighlighted) return v.value;
+    const key = `split-${i}-${v.value}`;
+    return <em key={key} className="ais-highlighted__value">{v.value}</em>;
+  });
+  return <span className="ais-highlighted">{reactHighlighted}</span>;
```
vvo: Small changes here. Since:
- users might want to style both the highlighted and non-highlighted values (like the google instant autocomplete), without having to go for complex CSS selectors (thus, add a class name)
- the widget is called Highlight

I propose we go for JSX that will output:

```html
<span class="ais-Highlight">
  <em class="ais-Highlight__highlighted>co</em>
  <span class="ais-Highlight__nonHighlighted">ca</span>
</span>
```

This way we are both consistent and we should answer most use cases of styling/targeting both highlighted and non-highlighted values. Even if the google instant example is a bit different (it's actually suggestions), I believe the use case should still be addressed because doable with Algolia.

packages/react-instantsearch/src/connectors/connectHighlight.js
```diff
@@ -0,0 +1,45 @@
+import createConnector from '../core/createConnector';
+import parseAlgoliaHit from '../core/highlight';
+
+const config = {
+  highlightPreTag: '<ais-highlight>',
```
vvo: Can we add some randomness to pre and post tags? I know we did it here: autocomplete.js#132. This would provide something slighty maybe more secure :copyright: security selfmade consultant

bobylito:
Updated with the request for changes :)

vvo:
:ok_hand:

PR #1652, opened 2016-12-03
docs(review): complete docs review

vvo:
- /guide/ and /examples/ instead of /guide.html /examples.html so that
if you are on /guide/Installation.html and try to go on /guide/, it
works
- moved advanced topics at the root of the guide
- reorganized the guide so it follows a progressive learning path
- read the guide entirely and made adjustement as if I was a new
user/reader
- made small adjustements on the rendering
- removed the need for @kind + @category, only kept @kind
(InstantSearch component is OK as a widget, it's not really a widget
but 90% of our components are widgets so it's easier to only talk
about widgets for users given the current shape of the API navigation)
- removed headings left navigation BUT from the getting started, pages are too small
to have headings (and sometime repetitive headings). Did not feel the
pain not having them, reduces noise

I have other ideas how to make the docs better, will bring them to
discuss once we are live. Right now we should be safe for a first
release **content wise** (we still have small adjustements/fine tuning to do).

new-docs


docgen/devServer.js
```diff
@@ -14,6 +14,7 @@ export default function() {
   const compiler = webpack(webpackConfig);
   const bs = browserSync.create();
   bs.init({
+    ghostMode: false,
```
vvo: This disable the mobile/desktop scroll sync, was triggering bad behavior for now

vvo: seems ok now :)

bobylito:
That's a great PR. Two questions:
 - is the centered page title a new feature? I feel like the left aligned was better :) (just some personal preference here, not blocking)
 - do we plan to have a review by the anglophones at Algolia? I am not sure about some reformulation, but I'm no native :)

If you don't plan any change, I would merge this PR :+1:

vvo:
> is the centered page title a new feature? I feel like the left aligned was better :) (just some personal preference here, not blocking)

Its only something I wanted to test. I have no strong preference. Not even convinced by the centered one, will try both and see.

> do we plan to have a review by the anglophones at Algolia? I am not sure about some reformulation, but I'm no native :)

Indeed this is required, I will ask for it.

vvo:
> is the centered page title a new feature? I feel like the left aligned was better :) (just some personal preference here, not blocking)

Re-tested, it's better left aligned clearly, updated

PR #1653, opened 2016-12-04
feat(hitsPerPage): hitsPerPage is now only configured by HitsPerPage

vvo:
Before this commit, Hits, connectHits, InfinitHits,
connectInfiniteHits, HitsPerPage, connectHitsPerPage were responsible
for dealing with the hitsPerPage parameter.

This encouraged usages like:

```jsx
Hits hitsPerPage=20
HitsPerPage items={[{value: 20, value:30}]}
```

Leading to confusions and conflict handling if both Hits's hitsPerPage
and items of HitsPerPage are different.

Now only HitsPerPage or the searchParameters prop (InstantSearch prop)
are able to control it.

Our goal being to always emphasis on the best UX, providing a way to
choose the number of hits to be shown by the user should be something
we push for. Therefore this PR removing hitsPerPage parameter on Hits
(and InfiniteHits).

This also removes inconsistency because utlimately <Hits
hitsPerPage={20}> should have been <Hits defaultRefinement={20}> if
they followed our connector rules/API.

### DIFF
--- a/docgen/src/examples/e-commerce-infinite/App.js
+++ b/docgen/src/examples/e-commerce-infinite/App.js
@@ -26,6 +26,7 @@ export default function App() {
       appId="latency"
       apiKey="6be0576ff61c053d5f9a3225e2a90f76"
       indexName="ikea"
+      searchParameters={{hitsPerPage: 16}}
     >
       <div>
         <Header />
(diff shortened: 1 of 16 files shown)


mthuret:
Merging previous PR made conflicts here.

PR #1654, opened 2016-12-04
fix(connectSearchBox): provide currentRefinement instead of query

vvo:
Connectors are following the currentRefinement, refine pattern.
Somehow connectSearchBox was missed.

BREAKING CHANGE:
- connectSearchBox now forward the query as props.currentRefinement
like any other connector


mthuret:
Merging previous PR made a conflict here.

PR #1659, opened 2016-12-05
feat(api): make hitsPerPage/SortBy connector consistent

mthuret:
* connectHitsPerPage must provide isRefined in items data
* connectHitsPerPage must disallow configuring with only numbers (HitsPerPage items=[10,20,30]). We should always force using the object form. For consistency and single way to do one thing.

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectHitsPerPage.js
+++ b/packages/react-instantsearch/src/connectors/connectHitsPerPage.js
@@ -1,3 +1,4 @@
+import {PropTypes} from 'react';
 import createConnector from '../core/createConnector';
 import {omit} from 'lodash';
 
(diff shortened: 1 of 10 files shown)


packages/react-instantsearch/src/connectors/connectHitsPerPage.js
```diff
@@ -22,17 +23,30 @@ function getCurrentRefinement(props, state) {
  * @name connectHitsPerPage
  * @kind connector
  * @propType {number} defaultRefinement - The number of items selected by default
- * @propType {{value, label}[]|number[]} items - List of hits per page options. Passing a list of numbers [n] is a shorthand for [{value: n, label: n}].
+ * @propType {{value: number, label: string||number}[]} items - List of hits per page options.
```
vvo: Could we only allow string labels? I feel like it's unecessary to try to handle both.
Also we can say that the default label is just the value when not provided. In the docs.

What do you think?

packages/react-instantsearch/src/connectors/connectHitsPerPage.js
```diff
@@ -22,17 +23,30 @@ function getCurrentRefinement(props, state) {
...
+    })).isRequired,
+  },
+
   getProps(props, state) {
+    const currentRefinement = getCurrentRefinement(props, state);
+    const items = props.items.map(i => i.value === currentRefinement
+      ? {...i, isRefined: true} : {...i, isRefined: false});
     return {
-      currentRefinement: getCurrentRefinement(props, state),
+      items,
```
vvo: Was items direclty passed down before this commit? If so I think I just discovered this implementation "detail" :)

packages/react-instantsearch/src/connectors/connectHitsPerPage.js
```diff
@@ -22,17 +23,30 @@ function getCurrentRefinement(props, state) {
...
+    items: PropTypes.arrayOf(PropTypes.shape({
+      label: PropTypes.string.isRequired,
+      value: PropTypes.number.isRequired,
+    })).isRequired,
+  },
+
   getProps(props, state) {
+    const currentRefinement = getCurrentRefinement(props, state);
+    const items = props.items.map(i => i.value === currentRefinement
+      ? {...i, isRefined: true} : {...i, isRefined: false});
```
vvo: Can we use `item` instead of `i`? So that we avoid single letter variables

packages/react-instantsearch/src/connectors/connectHitsPerPage.js
mthuret: well, items was a props from the user so no need to redefine it in the getProps (or getProvidedProps in the future;)).

Now that I enrich the structure, I need to pass it down through getProvidedProps.

PR #1661, opened 2016-12-05
fix(example): add default style for widget

mthuret:
This PR add the default style for our widget and fix styling issue due to it. 

It also refactor the way we declare style to ease override on the user side.


vvo:
GG

PR #1662, opened 2016-12-05
docs(links): make links works by default in dev and prod

### LINKED ISSUE #1639: Fix all broken links in the documentation
We have a lot of broken links in the documentation, mostly because we do things like `[/guide.html](guide)` and in production the real url is `/instantsearch.js/react/guide.html`.

Two ways to fix this:
- use the [base href](https://developer.mozilla.org/en/docs/Web/HTML/Element/base) attribute on the main webpage to define the base href for all relative (non http://) assets
- try to use h.publicPath helper in markdown files (not sure this is doable)

The base href option seems to be a very convenient way to do what we want without having to use h.publicPath everywhere.

vvo:
- remove the need for publicPath, just use base href
fixes #1639

Always link to ressources (html, css, js) by just putting the ressource
path from the base of the website without any leading slash.
    
Thus "images/test.png" and not publicPath.. nor "/images/test.png".


vvo:
Still need to fix some links to bad page. This PR is mostly about providing a solution and fixing obvious links that would not have worked in the production env (/intantsearch.js/react).

PR #1668, opened 2016-12-06
test(chore): move enzyme tests to main tests since update of React

vvo:


mthuret:
It's a detail, but I would have put the enzyme test at the top of the test file and the snapshot one at the end. (enzyme test are supposed to describe behavior so you might want to read/get that first).

PR #1669, opened 2016-12-06
docs: use widgets term when necessary

mthuret:


docgen/src/Getting started.md
```diff
@@ -308,7 +309,7 @@ In this part, we've seen the following:
 
 ## Next steps
 
-At this point, you know all the basics of react-instantsearch. All the
+At this point, you know all the basics of react-instantsearch. Also, All the
```
bobylito: You left the capital A to All.

mthuret: Thanks :)

Issue #1650, opened 2016-12-02
(Bug)(UX) - star rating - the cursor on the minimal value let me think that I can click (and it might unselect the value)


bobylito:
**What is the current behavior?**

![2016-12-02 18 37 16](https://cloud.githubusercontent.com/assets/393765/20844286/572f053c-b8bf-11e6-84ea-94b046924bdb.gif)

In the gif, I am actually click on the minimal value.

**What is the expected behavior?**

When the minimal value is selected, the cursor should not have an appearance that let the user think that it is clickable.

vvo:
Indeed this is a behavior that changed with react-instantsearch versus instantsearch.js. Basically we do display the minimum rating as being the default rating when you unselect/load the page to reflect the actual filter. What could we do to make it better here?

bobylito:
The behavior is imho perfect right now. The proposes improvement is to set the cursor when hovering the minimal value (when it is selected) to the default, not the hand. Because the hand is for clickable items and when it is the current value, clicking on it does not do anything therefore it is not clickable. Does that make sense? It's hard to explain. I will provide a PR (simple css fix) it should be easier to get :)

vvo:
Oh yes ok indeed shoot

PR #1670, opened 2016-12-06
fix(StarRating): Fix UX bug #1650

bobylito:
**Summary**

Fixes the ability to click even though there is no behavior triggered, when clicking on the default value and it is already selected. #1650 For reference:

**Result**

Now the default value is not clickable if it is selected:

And it also fixes a bug introduced by the highlighting PR :(

### DIFF
--- a/packages/react-instantsearch/src/components/StarRating.js
+++ b/packages/react-instantsearch/src/components/StarRating.js
@@ -31,7 +31,7 @@ class StarRating extends Component {
     }
   }
 
-  buildItem({max, lowerBound, count, translate, createURL}) {
+  buildItem({max, lowerBound, count, translate, createURL, isLowest}) {
     const selected = lowerBound === this.props.currentRefinement.min &&
       max === this.props.currentRefinement.max;
     const disabled = !count;
@@ -48,6 +48,32 @@ class StarRating extends Component {
           disabled && `${iconTheme}Disabled`,
         )}
         />);
+    }
+    if (isLowest && selected) {
+      return (
+        <div {...cx(
+          'ratingLink',
+          'ratingLinkSelected',
+          disabled && 'ratingLinkDisabled')}
+           disabled={disabled}
+           key={lowerBound}
+        >
+          {icons}
+          <span {...cx(
+            'ratingLabel',
+            'ratingLabelSelected',
+            disabled && 'ratingLabelDisabled')}>
+            {translate('ratingLabel')}
+            </span>
+          <span> </span>
+          <span {...cx(
+            'ratingCount',
+            'ratingCountSelected',
+            disabled && 'ratingCountDisabled')}>
+            {count}
+          </span>
+        </div>
+      );
     }
     return (
       <a {...cx(
@@ -92,6 +118,7 @@ class StarRating extends Component {
         count: itemCount,
         translate,
         createURL,
+        isLowest: i === min,
       }));
     }
     return (
--- a/packages/react-instantsearch/src/components/StarRating.js
+++ b/packages/react-instantsearch/src/components/StarRating.js
@@ -49,41 +49,24 @@ class StarRating extends Component {
         )}
         />);
     }
-    if (isLowest && selected) {
-      return (
-        <div {...cx(
-          'ratingLink',
-          'ratingLinkSelected',
-          disabled && 'ratingLinkDisabled')}
-           disabled={disabled}
-           key={lowerBound}
-        >
-          {icons}
-          <span {...cx(
-            'ratingLabel',
-            'ratingLabelSelected',
-            disabled && 'ratingLabelDisabled')}>
-            {translate('ratingLabel')}
-            </span>
-          <span> </span>
-          <span {...cx(
-            'ratingCount',
-            'ratingCountSelected',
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch/src/components/StarRating.js
```diff
@@ -49,6 +49,32 @@ class StarRating extends Component {
         )}
         />);
     }
+    if (isLowest && selected) {
```
vvo: Did you try to merge both this code path with the normal one? I am wondering if this could be a bit less verbose and only changes a into div + avoid setting onClick and href props.

Something like

```js
buildItem('a', commonProps, {href: ...});
buildItem('div', commonProps);
```

vvo: As we discussed via DM, here's the proposed git diff:

```
diff --git a/packages/react-instantsearch/src/components/StarRating.js b/packages/react-instantsearch/src/components/StarRating.js
index abcec5e..486e6b3 100644
--- a/packages/react-instantsearch/src/components/StarRating.js
+++ b/packages/react-instantsearch/src/components/StarRating.js
@@ -31,9 +31,10 @@ class StarRating extends Component {
     }
   }
 
-  buildItem({max, lowerBound, count, translate, createURL, isLowest}) {
+  buildItem({max, lowerBound, count, translate, createURL}) {
     const selected = lowerBound === this.props.currentRefinement.min &&
       max === this.props.currentRefinement.max;
+    const isLowestAndSelected = lowerBound === this.props.min && selected;
     const disabled = !count;
 
     const icons = [];
@@ -49,58 +50,36 @@ class StarRating extends Component {
         )}
         />);
     }
-    if (isLowest && selected) {
-      return (
-        <div {...cx(
-          'ratingLink',
-          'ratingLinkSelected',
-          disabled && 'ratingLinkDisabled')}
-           disabled={disabled}
-           key={lowerBound}
-        >
-          {icons}
-          <span {...cx(
-            'ratingLabel',
-            'ratingLabelSelected',
-            disabled && 'ratingLabelDisabled')}>
-            {translate('ratingLabel')}
-            </span>
-          <span> </span>
-          <span {...cx(
-            'ratingCount',
-            'ratingCountSelected',
-            disabled && 'ratingCountDisabled')}>
-            {count}
-          </span>
-        </div>
-      );
-    }
-    return (
-      <a {...cx(
-        'ratingLink',
-        selected && 'ratingLinkSelected',
-        disabled && 'ratingLinkDisabled')}
-         disabled={disabled}
-         key={lowerBound}
-         onClick={this.onClick.bind(this, lowerBound, max)}
-         href={createURL({lowerBound, max})}
-      >
-        {icons}
-        <span {...cx(
-          'ratingLabel',
-          selected && 'ratingLabelSelected',
-          disabled && 'ratingLabelDisabled')}>
-          {translate('ratingLabel')}
-          </span>
-        <span> </span>
-        <span {...cx(
-          'ratingCount',
-          selected && 'ratingCountSelected',
-          disabled && 'ratingCountDisabled')}>
-          {count}
+
+    const Wrapper = isLowestAndSelected ? 'div' : 'a';
+    const clickProps = isLowestAndSelected ? {} : {
+      onClick: this.onClick.bind(this, lowerBound, max),
+      href: createURL({lowerBound, max}),
+    };
+
+    return <Wrapper {...cx(
+      'ratingLink',
+      selected && 'ratingLinkSelected',
+      disabled && 'ratingLinkDisabled')}
+      disabled={disabled}
+      key={lowerBound}
+      {...clickProps}
+    >
+      {icons}
+      <span {...cx(
+        'ratingLabel',
+        selected && 'ratingLabelSelected',
+        disabled && 'ratingLabelDisabled')}>
+        {translate('ratingLabel')}
         </span>
-      </a>
-    );
+      <span> </span>
+      <span {...cx(
+        'ratingCount',
+        selected && 'ratingCountSelected',
+        disabled && 'ratingCountDisabled')}>
+        {count}
+      </span>
+    </Wrapper>;
   }
```

vvo: If you believe (like I do) this is a ok iteration then I can push the patch

packages/react-instantsearch/src/connectors/connectHighlight.js
```diff
@@ -36,7 +36,7 @@ export default createConnector({
 
   propTypes: {},
 
-  getProps() {
+  getProvidedProps() {
```
vvo: I am moving this in another commit inside v2 if you don't mind since it's not directly linked to the current bug. So that we can release a new beta version and updated the website.

bobylito:
Refactoring done!

packages/react-instantsearch/src/components/StarRating.js
```diff
@@ -49,15 +49,24 @@ class StarRating extends Component {
         )}
         />);
     }
+
+    // The last item of the list (the default item), should not
+    // be clickable if it is selected.
+    const isLastAndSelect = isLowest && selected;
+    const StarsWrapper = isLastAndSelect ? 'div' : 'a';
+    const onClickHandler = isLowest && selected ? {} : {
```
vvo: => isLastAndSelect?

PR #1672, opened 2016-12-06
docs(nav): add active links hints on top nav

### LINKED ISSUE #1563: Active class on header links
Add an active class on a list-item of the header-nav in documentation website so that we always know which page is active (much like hover style)

vvo:
fixes #1563

active

### DIFF
--- a/docgen/plugins/helpers.js
+++ b/docgen/plugins/helpers.js
@@ -13,6 +13,12 @@ export default function helpers(filenames, metalsmith, cb) {
       const lang = opts && opts.lang;
       return highlight(src, lang);
     },
+    maybeActive(navPath, singlePathOrArrayOfPaths) {
+      const pathsToTest = [].concat(singlePathOrArrayOfPaths);
+      return pathsToTest.some(pathToTest => navPath.indexOf(pathToTest) === 0) ?
+        'active' :
+        '';
+    },
   };
 
   cb();
--- a/docgen/src/stylesheets/components/_dropdown.scss
+++ b/docgen/src/stylesheets/components/_dropdown.scss
@@ -24,12 +24,12 @@ ul.simple-dropdown {
   font-size: $sd-font-size;
   text-align: left;
   border: 1px solid $sd-border-color;
-  border-radius: 3px;
+  border-radius: 0 0 3px 3px;
   background: $sd-background-color;
   border: none;
   padding: 0;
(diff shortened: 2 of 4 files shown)


docgen/plugins/helpers.js
```diff
@@ -13,6 +13,12 @@ export default function helpers(filenames, metalsmith, cb) {
       const lang = opts && opts.lang;
       return highlight(src, lang);
     },
+    maybeActive(navPath, singlePathOrArrayOfPaths) {
+      const pathsToTest = [].concat(singlePathOrArrayOfPaths);
```
bobylito: Nice trick!

docgen/src/stylesheets/components/_dropdown.scss
```diff
@@ -24,12 +24,12 @@ ul.simple-dropdown {
...
   text-align: left;
   border: 1px solid $sd-border-color;
-  border-radius: 3px;
+  border-radius: 0 0 3px 3px;
   background: $sd-background-color;
   border: none;
   padding: 0;
   box-shadow: 0 1px 4px 0 rgba(31,59,93,0.5);
-  animation: fadeInDown 0.6s cubic-bezier(0.23, 1, 0.32, 1);
+  // animation: fadeInDown 0.1s cubic-bezier(0.23, 1, 0.32, 1);
```
bobylito: If it's commented maybe we should just remove it? Why remove it btw?

LukyVj: you meant "Why commented it btw" ?

vvo: Oops sorry, I will update this. 0.6s does feel too slow (you really feel the "wait") I think there are recommendations as for animation timings. Will have a look. Right now its instant and not too brutal

docgen/plugins/helpers.js
vvo: I had to test this in console to actually enaure its doing what I want eheh but yes funn

bobylito: ❤️

docgen/src/stylesheets/components/_dropdown.scss
bobylito: Well I trust @LukyVj on this one :)

vvo:
Added back the animation but a faster one

PR #1682, opened 2016-12-07
fix(examples): issues with header & content of the examples

mthuret:


vvo:
z-index :dancing_women:

PR #1684, opened 2016-12-07
docs: add jsx tags to InstantSearch

mthuret:


mthuret:
Also inside the widget.pug layout we have this:

```
    p List of props exposed by the
      code=name
      |  widget
```

This is also used by the <InstantSearch> component. Do we want to add some logic here to make the distinction? 

see: https://community.algolia.com/instantsearch.js/react/widgets/InstantSearch.html

vvo:
> p List of props exposed by the
>       code=name
>       |  widget

Yes and also in the left navigation, that would be kind of awesome :)

vvo:
@mthuret Sorry I was not precise, do you think we could put `<SearchBox>` instead of `SearchBox` and other widgets, in the left nav also?

Maybe by just checking the category (connector or widget), we can add this easily?

vvo:
Merging this one first then will see

PR #1686, opened 2016-12-07
fix(state): when having two widgets of the same type, state were erased when refining

### LINKED ISSUE #1685: React Router new state replace the old state
I install the latest beta version, `beta.20`.

I updated my react-router code from the [example](https://github.com/algolia/instantsearch.js/blob/67c75259c8408bea909a9d2a51404aafd00b8dd9/packages/react-instantsearch/examples/react-router/src/App.js) (you can see the code [here](https://github.com/windtoday/windtoday-app/commit/68685d67aaf2ba90189c91a335428a6a497936f7))

The thing that I detected that happen right now is that the new state replace the old state and it hasn't the previous filters applied:

mthuret:
fixes #1685

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectHierarchicalMenu.js
+++ b/packages/react-instantsearch/src/connectors/connectHierarchicalMenu.js
@@ -148,7 +148,7 @@ export default createConnector({
     const id = getId(props);
     return {
       ...searchState,
-      [namespace]: {[id]: nextRefinement || ''},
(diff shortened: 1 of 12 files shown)


vvo:
Nice find right before release, will publish a new version

vvo:
@kikobeats This can of feedback is gold for us, we would have maybe pushed a completely broken first version.

Kikobeats:
team building! 💪

vvo:
> team building!

We are still recruiting :trollface: :)****

PR #1690, opened 2016-12-07
docs(landing): implement landing page design first iteration

vvo:


vvo:
Some glitches on pages like guide/ other than that it s very good

PR #1693, opened 2016-12-07
fix(theme): Defaults for hits and highlights

bobylito:
**Summary**

 - Move from no font to some sans-serif font (using the same rule as HTML5 boilerplate).
 - Add highlight styling.
 - Adds a wrapper container in InstantSearch

Also a start for a default theme playground to test the default theme.


vvo:
Nice! No more wrapping div needed :heart_eyes_cat:

PR #1694, opened 2016-12-07
update-text

neilrichler:
**Summary**

**Result**

Quick change to text. Alternatively could be searchbox or searchBox.


neilrichler:
What's the best way to address this? Looks like we use camelcase for components in InstantSearch.js.

vvo:
Theres no relation right now between instantsearch.js and react instantsearch. Public is different and will not be confused by the different way to write SearchBox. In react, component names are written like that

PR #1695, opened 2016-12-07
update-text

neilrichler:
**Summary**

**Result**

Text and grammar updates


docgen/src/Getting started.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Getting started
+title: Getting Started Guide
```
vvo: Not sure about the Guide addition. This is really a getting started tutorial. I dont want to confuse users with the actual guide section. Is getting started alone not ok?

PR #1724, opened 2016-12-09
chore(contributions): new issue template

vvo:
- react-instantsearch
- instantsearch.js

**Summary**

We want users to be very precise in their issue descriptions so we can better help them solve their problems. We also want to leverage discourse forum for generic questions.

**Result**

This will trigger a new issue template when opening issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -20,6 +20,8 @@
 
 **Bug: What is the expected behavior?**
 
+**Bug: What browsers are impacted? Which versions?**
+
 **Feature: What is your use case.**
 
 **Feature: What is your proposed API entry, new option, to add**
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,9 +4,7 @@
 
   This form is to report issues or new features.
   As for general questions like "How to do routing using react-instantsearch",
-  please search or post a question to:
-    - Our discourse forum: https://discourse.algolia.com/ (preferred);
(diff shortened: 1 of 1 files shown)


MoOx:
My take

```
<!--
  ***************************
  🎉 Thanks for participating in this project! 🎉

  This form is to report issues or new features.
  As for general questions like "How to do routing using react-instantsearch",
  please search or post a question to our discourse forum: https://discourse.algolia.com/ (preferred).

  ⚠️ In any case,
    - make sure you are using the latest version of the library;
    - do at least one search in current issues or questions, your question might already be answered;
  ***************************

  ❓ Do you want to request a *feature* or report a *bug*?
-->

## 🐛 I want to report a bug

### What is the current behavior?

### What is the expected behavior?

### What browsers are impacted? Which versions?

<!-- ⚠️ or -->

## 🍭 I want to request a feature

### What is your use case?

### What is your proposed API entry, new option, to add if any?

<!--
  ⚠️ *************************** ⚠️
  
  🐛 If the current behavior is a bug, please provide all the steps to reproduce and a minimal
  [JSFiddle](https://jsfiddle.net/) example or a repository on GitHub that we can `npm install`
  and `npm start`.

  🍭 If you are requesting a new feature, we need to understand WHY would you
  need this feature, which means you should give the use case for such feature.

  ⚠️ *************************** ⚠️
-->

## What project are you opening an issue for?

- react-instantsearch
- instantsearch.js

## What is the version you are using?

🚨 Please always use the latest one before opening a bug issue.

<!-- 🚫 Delete any HTML comment and non relevant questions 🚫 -->
```

PR #1727, opened 2016-12-09
fix(range): when unfinite numbers are passed, use search results instead.

mthuret:
Behavior: when using NaN as min/max current values of a range, an infinite loop is triggered. 

This fix check inside the connector if the values are finite, if not, we are defaulting to the one provided to the widget (either by hand ou by using the search results). 

=> We discovered this behavior when using the rheostat slider widget. For an unknown reason when you have the same value for Min/Max, the slider gaves you NaN/NaN instead of the real value. 
The consequence for react-instantsearch is that when the 'shouldComponentUpdate' hook of the range connector is triggered, equality will never be reached as the currentRefinement contains 'null' value.

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectRange.js
+++ b/packages/react-instantsearch/src/connectors/connectRange.js
@@ -103,6 +103,9 @@ export default createConnector({
   },
 
   refine(props, searchState, nextRefinement) {
+    if (!isFinite(nextRefinement.min) || !isFinite(nextRefinement.max)) {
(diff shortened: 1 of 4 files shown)


vvo:
For reference could you maybe add inside the commit when can this happen? In which context. Thanks

mthuret:
Updated.

mthuret:
New behavior is:

- check when calling refine that the values of min/max are finite.
- in the range slider example, hide the slider if min===max (if all the remaining results have the same value, we hide the slider).

PR #1729, opened 2016-12-09
fix(react-native): add View as a container for react-native

bobylito:
The container added in a previous patch in the <InstantSearch /> was not working because strings can't be used in React-Native for createElement. It introduced a bug that [some users found ](https://stackoverflow.com/questions/41064940/using-algolia-react-instantsearch-with-react-native)

In order to fix it and not add a dependency to RN, it fallbacks to using the previous behaviour when not in DOM.


vvo:
> In order to fix it and not add a dependency to RN, it fallbacks to using the previous behaviour when not in DOM.

What about we do not add any dependency and assume people willing to build instantsearch with react native do have react-native installed? This is what we are doing for react-dom, nowhere we have "react-dom" as a dep (not in peer, not in deps).

So that we could keep the same behavior. Would that work?

Ultimately we may have three packages:

react-intantsearch-dom
react-instantsearch-native
react-instantsearch

Just like what facebook does. So we can add any peer/dep we want that makes sense given the use case.

vvo:
Cannot even right a proper test for react-native as this:

```js
/* eslint-env jest */

import 'react-native';
import React from 'react';
// import {InstantSearch} from './native.js';

// Note: test renderer must be required after react-native.
import renderer from 'react-test-renderer';
const algoliaClient = {search() { console.log('test'); }};

it('renders correctly', () => {
  const Root = 'View';
  const App = () => <Root/>;

  const tree = renderer.create(
    <App />
  ).toJSON();
  expect(tree).toMatchSnapshot();
});
```

Just works, might be Jest being too optimistic

bobylito:
I didn't know we didn't have a hard dependency on react-dom 🤔 since it works, like you said, then it's nice because it will be consistent across all platforms👌 making the changes tomorrow 🤓

mthuret:
The change are:

- we passe directly View as a root for the react-native InstantSearch 
- add react-native example folder + simplification of our infinitsearch example

packages/react-instantsearch/examples/reactnative/App.js
```diff
@@ -0,0 +1,101 @@
...
+    flexDirection: 'row',
+    alignItems: 'center',
+    backgroundColor: 'white',
+  },
+});
+
+export default React.createClass({
+  render() {
+    return (
+      <View style={styles.maincontainer}>
```
bobylito: So the View in a View works?

packages/react-instantsearch/examples/reactnative/App.js
```diff
@@ -0,0 +1,101 @@
...
+        >
+          <SearchBox/>
+          <Hits/>
+        </InstantSearch>
+      </View>
+    );
+  },
+});
+
+const SearchBox = connectSearchBox(React.createClass({
```
bobylito: React.createClass is kinda old school. People tend to favor ES6 class or functions #notBlocking :)

packages/react-instantsearch/examples/reactnative/App.js
mthuret: Yes after cleaning up a lot of Views of the Example, everything went fine 👍

packages/react-instantsearch/examples/reactnative/App.js
mthuret: Doing it right now :)

packages/react-instantsearch/examples/reactnative/.gitignore
```diff
@@ -0,0 +1,42 @@
+# OSX
```
vvo: Is this the default gitignore generated by react-native? There's a lot in it.

packages/react-instantsearch/examples/reactnative/App.js
vvo: NICE

PR #1742, opened 2016-12-12
fix(SearchBox): autocomplete was not disabled by default

mthuret:
We were using false instead of "off" in the input of our `<SearchBox/>`


vvo:
love it

PR #1744, opened 2016-12-13
Fix/buttons misc

LukyVj:


docgen/src/stylesheets/components/_hero.scss
```diff
@@ -4,6 +4,23 @@
 
 // Start the hero
 $css: ".st0{ fill:none; stroke:%23FFF; stroke-miterlimit:10;}";
+//   //KEEP ME PLEASE
```
vvo: Can you keep this code on your side for future use? Otherwise we will keep this forever.

PR #1745, opened 2016-12-14
connectInfiniteHits: Fix fetch the first page

Kikobeats:
Because first page index is 0

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectInfiniteHits.js
+++ b/packages/react-instantsearch/src/connectors/connectInfiniteHits.js
@@ -30,7 +30,7 @@ export default createConnector({
     if (page === 0) {
       this._allResults = hits;
     } else {
-      const previousPage = this._allResults.length / hitsPerPage - 1;
+      const previousPage = this._allResults.length - 1 / hitsPerPage;
 
       if (page > previousPage) {
         this._allResults = [


vvo:
The state that is here is the state we "Expose". When doing queries to algolia we then do page - 1 because algolia is 0 based.

But inside the state, state.page = 1 means we will query page 0 on Algolia.

Could you walk us through your usecase/struggle with this maybe? How did you encounter a bug, how can we reproduce it? Thanks

Kikobeats:
I found this behavior playing in [windtoday.co](https://windtoday.co/) with facets. 

Currently is monkey patched on production, but basically in the results:

- I see the page 1.
- When I click in `load more` it injects the results of the page 1 again.

Kikobeats:
@vvo I revert my production build to make more easy reproduce the issue.

Enter in [this url](https://windtoday.co/?range%5Bsail.size%5D%5Bmin%5D=6&range%5Bsail.size%5D%5Bmax%5D=7&p=0).

Now check the title of the results and click in `Load More`. You can find duplicate titles.

packages/react-instantsearch/src/connectors/connectInfiniteHits.js
```diff
@@ -66,7 +66,7 @@ export default createConnector({
     const id = getId();
     const nextPage = searchState[id] ?
       Number(searchState[id]) + 1 :
-      1;
+      0;
```
bobylito: Looks legit based on Algolia API knowledge, but we've had discussion and we decided to fix the Algolia API and provide a user understandable numeration. If there is a bug, it lies where the correction is done not here.

Kikobeats: I'm agree, this is a monkey patch, but looks a problem related with what index is considered as first index

Kikobeats:
Hmm I need to investigate more because after this change I still see duplicate fetch pages 🤔

Kikobeats:
I updated the PR, for me have more sense and resolve the problem.

What do you think, @vvo @bobylito ?

Kikobeats:
nope, looks weird. xD

Kikobeats:
Currently just resolved adding

```
searchParameters={{hitsPerPage: 15}}
```

without modify the code.

vvo:
@Kikobeats Do you think you could either:
- write a unit test
- create a simple example

Using this repository? So that we reduce the test case to the only parts that are relevant?

vvo:
@Kikobeats I fail at seeing if we have a real bug here or what? Having a simple example could help, otherwise we'll close this untill someone find a simple example to reproduce it

Kikobeats:
Sorry currently I fixed following the `searchParameter` setup, I need to investigate more for open a new issue

bobylito:
Thanks @Kikobeats for taking the time!

Issue #1746, opened 2016-12-14
In our examples only use syntaxes compatible with 'latest' and 'react' babel preset 


mthuret:
Goal: having our examples compatible with the largest possible app configurations.

PR #1749, opened 2016-12-14
docs(examples): use only 'latest' and 'react' babel preset syntaxes

mthuret:
see #1746 

Also add our multi-index example into our examples directory.


vvo:
You :metal:

PR #1751, opened 2016-12-14
feat(widget): change default sortBy of refinementList to [count:desc, name:asc]

mthuret:
kapture 2016-12-14 at 17 03 13

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectRefinementList.js
+++ b/packages/react-instantsearch/src/connectors/connectRefinementList.js
@@ -39,7 +39,7 @@ function getValue(name, props, searchState) {
   return nextRefinement;
 }
 
-const sortBy = ['isRefined', 'count:desc', 'name:asc'];
+const sortBy = ['count:desc', 'name:asc'];
 
 /**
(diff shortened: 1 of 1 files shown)


bobylito:
I don't think you should change the defaults because when searching it might make the selected disappears, which is very confusing for users.

vvo:
> when searching it might make the selected disappears, which is very confusing for users

Do you mean when refining the search again in the searchbox, they might disappear? I think I see what you mean but could you precise a bit more the steps that would lead to that bad behavior?

bobylito:
> could you precise a bit more the steps and causes that would lead to that bad behavior?

Yes sure. Here you can't see epiphone in the list of brands http://feigned-queen.surge.sh/?query=&hits=24&idx=instant_search&p=0&dFR%5Bbrand%5D%5B0%5D=Epiphone&is_v=1

vvo:
As seen in the provided example, the facet disappears when sort is only count + alpha, cc @Kikobeats.

There's no silver bullet for now, @bobylito wants to review all options we have and if we need to make changes to the default sort we do.

Kikobeats:
yaaaaaaaay

PR #1753, opened 2016-12-14
[Search for facet values] Refinement list - InstantSearch.js

bobylito:
*This is a preview PR and I would love your input on the implementation, as well as the configuration options. The commits will be squashed to have a proper history.*

**Summary**

This PR provides an implementation of options to be able to search into facet values in the refinementList widget. When activated it adds a search input on top of the list, the results are replacing the values in the list. Pressing return when there are search results, selects the first item.

The API tries to stay consistent with the current implementation with a configuration object that  defaults and with the following structure (using the structure enables the feature):
 - placeholder: string (plain text)
 - templates: object
 - templates.noResults: string (html  template)

**Result**

[See it live](http://bouncy-clam.surge.sh/) 🔥

### DIFF
--- a/dev/app.js
+++ b/dev/app.js
@@ -163,6 +163,30 @@ search.addWidget(
 
 search.addWidget(
   instantsearch.widgets.refinementList({
+    container: '#searchable-brands',
+    attributeName: 'brand',
+    operator: 'or',
+    limit: 10,
+    cssClasses: {
+      header: 'facet-title',
+      item: 'facet-value checkbox',
+      count: 'facet-count pull-right',
+      active: 'facet-active',
+    },
+    templates: {
+      header: 'Searchable brands',
+    },
+    searchForFacetValues: {
+      placeholder: 'Find other brands...',
+      templates: {
+        noResults: 'No results',
+      },
+    },
+  })
+);
+
+search.addWidget(
+  instantsearch.widgets.refinementList({
     collapsible: {
       collapsed: true,
     },
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -6,6 +6,8 @@ import Template from '../Template.js';
 import RefinementListItem from './RefinementListItem.js';
 import isEqual from 'lodash/isEqual';
 
+import SearchBox from '../SearchBox';
+
 class RefinementList extends React.Component {
   constructor(props) {
     super(props);
@@ -121,6 +123,20 @@ class RefinementList extends React.Component {
     this.setState({isShowMoreOpen});
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (this.searchbox && !nextProps.isFromSearch) {
+      this.searchbox.clearInput();
+    }
+  }
+
+  refineFirstValue() {
+    const firstValue = this.props.facetValues[0];
+    if (firstValue) {
+      const actualValue = firstValue[this.props.attributeNameKey];
+      this.props.toggleRefinement(actualValue);
+    }
+  }
+
   render() {
     // Adding `-lvl0` classes
     const cssClassList = [this.props.cssClasses.list];
@@ -144,9 +160,25 @@ class RefinementList extends React.Component {
         /> :
         undefined;
 
+    const searchInput = this.props.searchFacetValues ?
+      <SearchBox ref={i => { this.searchbox = i; }}
+        placeholder={this.props.searchPlaceholder}
+        onChange={this.props.searchFacetValues}
+        onValidate={() => this.refineFirstValue()}/> :
+      null;
+
+    const noResults = this.props.searchFacetValues && this.props.isFromSearch && this.props.facetValues.length === 0 ?
+      <Template
+        templateKey={'noResults'}
+        {...this.props.templateProps}
+      /> :
+      null;
+
     return (
       <div className={cx(cssClassList)}>
+        {searchInput}
         {displayedFacetValues.map(this._generateFacetItem, this)}
+        {noResults}
         {showMoreBtn}
       </div>
     );
@@ -170,6 +202,9 @@ RefinementList.propTypes = {
   showMore: React.PropTypes.bool,
   templateProps: React.PropTypes.object.isRequired,
   toggleRefinement: React.PropTypes.func.isRequired,
+  searchFacetValues: React.PropTypes.func,
+  searchPlaceholder: React.PropTypes.string,
+  isFromSearch: React.PropTypes.bool,
 };
 
 RefinementList.defaultProps = {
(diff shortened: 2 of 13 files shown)


mthuret:
Question about the behavior: if we sort the refinementList by `isRefined` first, shouldn't we sort the search for facet values results list the same way?

bobylito:
@mthuret this is a very valid question indeed. I failed to find a way to do that without extracting and moving the code sorting the values. I think this question is broader, because in instantsearch.js we let developers use custom sort, should we apply this one?

After testing it seems that the order is fine when returned from Algolia because we are searching. I will put an online version to let you test.

vvo:
I managed to break it (Don't get me wrong, this is awesome, I just like to break stuff), steps:
- load http://bouncy-clam.surge.sh/
- search for digipower in facet search
- select digipower
- search for faeta in searchbox
- search for digipower in facet search
- remove characters in facet search one by one
- wait 700ms
- it breaks

bobylito:
/o\
😢 😿 😭

dev/app.js
```diff
@@ -163,6 +163,30 @@ search.addWidget(
...
+    cssClasses: {
+      header: 'facet-title',
+      item: 'facet-value checkbox',
+      count: 'facet-count pull-right',
+      active: 'facet-active',
+    },
+    templates: {
+      header: 'Searchable brands',
+    },
+    searchForValues: {
```
vvo: Would `searchForFacetValues` be more inline with other APIS?

src/components/RefinementList/RefinementList.js
```diff
@@ -6,6 +6,8 @@ import Template from '../Template.js';
 import RefinementListItem from './RefinementListItem.js';
 import isEqual from 'lodash/isEqual';
 
+import SearchBox from '../SearchBox';
```
vvo: When I read this I thought that we were reusing our current SearchBox, turns out we do not but I understand why it's not the case. The current searchbox is only a vanilla JS widget that we made because we had weird use cases that we should not have tried to solve in the first place the way we did. `</LegacyRant>`

src/components/RefinementList/RefinementList.js
```diff
@@ -121,20 +123,44 @@ class RefinementList extends React.Component {
...
   render() {
+    const {
+      cssClasses,
+      limitMin,
+      limitMax,
+      facetValues,
+      searchFacetValues,
+      isFromSearch,
+      searchPlaceholder,
+    } = this.props;
```
vvo: If only React gave props and state as parameter in `render()` so we do not have to put this boilerplate code in place and only do it destructuring arguments! Preact does that.

vvo: I actually never feel the struggle of repeating this.props., I like it it's descriptive about where did we find this data

src/widgets/refinement-list/defaultTemplates.js
```diff
@@ -1,7 +1,7 @@
 export default {
   header: '',
   item: `<label class="{{cssClasses.label}}">
-  <input type="checkbox" class="{{cssClasses.checkbox}}" value="{{name}}" {{#isRefined}}checked{{/isRefined}} />{{name}}
+  <input type="checkbox" class="{{cssClasses.checkbox}}" value="{{name}}" {{#isRefined}}checked{{/isRefined}} />{{{highlighted}}}
```
vvo: This can be seen as a possible breaking change, I would make it only for the new template of search values (and so duplicate the template and provide a searchForFacetValues.templates.item). What do you think?

src/widgets/refinement-list/refinement-list.js
```diff
@@ -63,20 +67,22 @@ refinementList({
...
-  [ collapsible=false ]
+  [ collapsible=false ],
+  [ searchForValues.{placeholder, templates: {noResults}}],
 })`;
 function refinementList({
     container,
     attributeName,
     operator = 'or',
-    sortBy = ['count:desc', 'name:asc'],
+    sortBy = ['isRefined', 'count:desc', 'name:asc'],
```
vvo: Possible breaking change too?

dev/app.js
vvo: Indeed, to reduce similarity maybe we can add "withSearchbox"? Otherwise let's keep it

src/components/RefinementList/RefinementList.js
bobylito: Ok!

src/components/RefinementList/RefinementList.js
vvo: TL;DR; It's good the way you did it

src/widgets/refinement-list/defaultTemplates.js
bobylito: That's OK. Safer is better!

vvo:
One last comment, about design, the icons are not the same style. They seems to come from two different icon sets. The magnifier icon is good, the clear one is a bit too bold/big from my POV

src/widgets/refinement-list/refinement-list.js
bobylito: No it doesn't break, but it does change the behavior.

I had the impression that we discussed the defaults for react-instantsearch. And so I thought that we could move forward on this too. Anyway, if you use search for facet values you will add values that are not the most popular, so they won't be displayed in the list if isRefined is not first.

bobylito:
> One last comment, about design, the icons are not the same style. They seems to come from two different icon sets. The magnifier icon is good, the clear one is a bit too bold/big from my POV

I used @Shipow generator and I think I can tweak that pretty easily. Thanks!

src/widgets/refinement-list/defaultTemplates.js
bobylito: Well after thinking about that, I don't think it's an issue. If users customized the template and they are using `name` it will still work. If they are using search for facet value, the value of `highlighted` will be set to name. I think we're safe actually.

dev/app.js
bobylito: I'll switch for `searchForFacetValues` it's more consistent and obvious what it's about!

src/components/RefinementList/RefinementList.js
bobylito: Using `this.props`!

src/components/SearchBox/index.js
```diff
@@ -6,6 +6,7 @@ export default class SearchBox extends React.Component {
     placeholder: React.PropTypes.string.isRequired,
     onChange: React.PropTypes.func.isRequired,
     onValidate: React.PropTypes.func.isRequired,
+    showSearchIcon: React.PropTypes.bool,
```
vvo: I think we should not have an option to decide to show the search icon or not. Best practice is to have one (cc @Shipow), so lets always have it and if the user really wants to hide it then it can be done with CSS.

vvo:
image

:+1:

PR #1761, opened 2016-12-16
docs(examples): add autocomplete recipe

mthuret:
Those examples shows how to use an external react autocomplete component with react-instantsearch with two use cases: 
* targetting multiple indices 
* displaying both hits and facet values. 

Those examples are boilerplate but will serve to derive the right API to perform the same thing.

kapture 2016-12-16 at 13 57 16

### DIFF
--- a/packages/react-instantsearch/examples/autocomplete/README.md
+++ b/packages/react-instantsearch/examples/autocomplete/README.md
@@ -1,0 +1,14 @@
+This example shows how to use an external autocomplete component with react-instantsearch
+It uses [react-autosuggest](https://github.com/moroshko/react-autosuggest) as the autocomplete external component.
+
+You will find two use cases:
+
+* How to build an autocomplete displaying hits and facet values from the same index.
+* How to build an autocomplete displaying hits from different indices
+
+To start the example:
+
+* npm install
+* npm start
+
+Read more about react-instantsearch [in our documentation](https://community.algolia.com/instantsearch.js/react/).
--- a/packages/react-instantsearch/examples/autocomplete/package.json
+++ b/packages/react-instantsearch/examples/autocomplete/package.json
@@ -1,0 +1,19 @@
+{
+  "private": true,
+  "devDependencies": {
+    "react-scripts": "0.7.0"
+  },
+  "dependencies": {
+    "lodash": "^4.16.6",
+    "react-autosuggest": "^7.0.1",
+    "react": "^15.3.2",
+    "react-dom": "^15.3.2",
+    "react-instantsearch": "file:../../dist",
+    "react-instantsearch-theme-algolia": "file:../../../react-instantsearch-theme-algolia"
+  },
+  "scripts": {
+    "start": "react-scripts start"
+  },
+  "main": "index.js",
+  "license": "MIT"
+}
--- a/packages/react-instantsearch/examples/autocomplete/src/App-Hits-And-Facets.js
+++ b/packages/react-instantsearch/examples/autocomplete/src/App-Hits-And-Facets.js
@@ -1,0 +1,113 @@
+import React, {Component} from 'react';
+import {InstantSearch} from 'react-instantsearch/dom';
+import {createConnector} from 'react-instantsearch';
+import {connectSearchBox, connectRefinementList} from 'react-instantsearch/connectors';
+import Autosuggest from 'react-autosuggest';
+import {forOwn} from 'lodash';
+/* eslint-disable import/no-unresolved */
+import 'react-instantsearch-theme-algolia/style.css';
+/* eslint-enable import/no-unresolved */
+
+class App extends Component {
+  render() {
+    return (
+      <InstantSearch
+        appId="latency"
+        apiKey="6be0576ff61c053d5f9a3225e2a90f76"
+        indexName="bestbuy"
+        searchParameters={{hitsPerPage: 3}}
+      >
+        <div>
+          <VirtualSearchBox/>
+          <VirtualRefinementList attributeName="manufacturer"/>
+          <VirtualRefinementList attributeName="category"/>
+          <ConnectedAutoComplete attributes={['manufacturer', 'category']}/>
+        </div>
+      </InstantSearch>
+    );
+  }
+}
+
+const VirtualSearchBox = connectSearchBox(() => null);
+const VirtualRefinementList = connectRefinementList(() => null);
+
+const connectAutoComplete = createConnector({
+  displayName: 'AutoComplete',
+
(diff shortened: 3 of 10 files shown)


packages/react-instantsearch/examples/autocomplete/src/App-Hits-And-Facets.js
```diff
@@ -0,0 +1,143 @@
...
+  onProps: React.PropTypes.func.isRequired,
+};
+
+const VirtualSearchBox = connectSearchBox(() => null);
+const VirtualRefinementList = connectRefinementList(() => null);
+
+const connectAutoComplete = createConnector({
+  displayName: 'AutoComplete',
+
+  getProvidedProps(props, state, search) {
```
mthuret: We use a custom connector here to be able to gather every props we need before sending them to the higher component. We need to gather all those data because the external autocomplete component (here react-autosuggest) needs to be initialize with the entire state at once.

mthuret:
One thing that could be done to lower the amount of code needed for this autocomplete is to enrich our `searchState` object with both hits and facet values results.

packages/react-instantsearch/examples/autocomplete/src/App-Hits-And-Facets.js
```diff
@@ -0,0 +1,143 @@
+import React, {Component} from 'react';
+import {InstantSearch} from 'react-instantsearch/dom';
+import {createConnector} from 'react-instantsearch';
+import {connectSearchBox, connectRefinementList} from 'react-instantsearch/connectors';
+import Autosuggest from 'react-autosuggest';
+import {isEqual, forOwn} from 'lodash';
+/* eslint-disable import/no-unresolved */
```
bobylito: You can use `eslint-disable-line` to disable the next line.

packages/react-instantsearch/examples/autocomplete/src/App-Hits-And-Facets.js
```diff
@@ -0,0 +1,143 @@
...
+    forOwn(this.state.hits, (value, key) => {
+      hits.push({title: key, hits: value});
+    });
+    return hits;
+  }
+
+  render() {
+    return (
+      <div>
+        <FirstResults onProps={this.onProps.bind(this)}
```
bobylito: You should bind that in the constructor so it doesn't create a new function at each `render`.

packages/react-instantsearch/examples/autocomplete/README.md
```diff
@@ -0,0 +1,8 @@
+This example shows how to build an autocomplete with react-instantsearch
+
+To start the example: 
+
+* npm install
+* npm start
+
+Read more about react-instantsearch [in our documentation](https://community.algolia.com/instantsearch.js/react/).
```
bobylito: I would add more implementation details to let the user know what he/she should focus their attention on.

mthuret: Comments inside the source code of the example would be better no? And maybe a link back to the guide from the README, WDYT?

packages/react-instantsearch/examples/autocomplete/src/App-Hits-And-Facets.js
mthuret: Do you think it would be clearer (I assume you have only one line of comment instead of two?) ? I link to add the precise one for not having to think at which rule was failing.

bobylito: OK no problem, it was just a suggestion (I fail to remember it every time :))

packages/react-instantsearch/examples/autocomplete/README.md
bobylito: Well I think that I would like to have a bit more context before diving in the example itself. But sure, comments in the code would help too! Do you plan on creating a dedicated guide?

mthuret: I'm not sure, at least, not with this solution. 
What I can explain on a guide is the idea behind the example: having a function listening to all changes and then construct the autocomplete state with it. But the way we can achieve this now seems very fragile, that's why I'm not sure about a dedicated guide in the end.

But I can still do that in the README, and then if we come up with a good solution to simplify this kind of use-cases, then we could write a guide. We also need to add all the recipes to the documentation website.

bobylito: > What I can explain on a guide is the idea behind the example: having a function listening to all changes and then construct the autocomplete state with it

OK that sounds of great help! Also I think you can mention that another specificity of this example is that it relies on a custom connector. This is very advanced, so let's be sure it's not too confusing.

About the robustness, it worries me a bit. Maybe it's not worth documenting at all unless we have a more solid solution? I don't know how much documentation are those examples either :) @vvo an idea on the subject?

mthuret: Thanks to @zasmail that tried the example yesterday, there's one part that is complex to understand: the namespacing when gathering all the data. I will detail this using comment in the code. I'll also add details in the README to explain the main idea behind the example, but without going to far as it might heavily depends on the underlying autocomplete component you choose to use.

vvo:
> One thing that could be done to lower the amount of code needed for this autocomplete is to enrich our searchState object with both hits and facet values results.

This is something we will have to discuss soon. I am unsure if that's a good idea but it would simplify things clearly

packages/react-instantsearch/examples/autocomplete/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "private": true,
+  "devDependencies": {
+    "react-scripts": "0.7.0"
+  },
+  "dependencies": {
+    "lodash": "^4.16.6",
+    "react-autosuggest": "^7.0.1",
+    "react": "^15.3.2",
+    "react-dom": "^15.3.2",
+    "react-instantsearch": "file:../../dist",
```
vvo: Can't we use the source files rather than the dist?

vvo:
Can you add it to the guide? Like "Autocomplete menu" guide.

mthuret:
Guide added.

HemalR:
Hi!

Are you able to link to the guide referenced above? My Google searching has only found the non-react autocomplete guide :s

Thanks in advance!

mthuret:
Hi @HemalR you can find the guide here: https://community.algolia.com/instantsearch.js/react/guide/Autocomplete_menu.html

HemalR:
You replied as quick as Algolia searches! Thank you :)

PR #1763, opened 2016-12-16
Fix/responsive

LukyVj:


docgen/assets/js/sidebar.js
```diff
@@ -105,7 +105,7 @@ function toggleDocumentationSidebar() {
   function init() {
     const bodySize = document.body.clientWidth;
     if (bodySize <= 960 && sidebarNav) {
-      trigger.addEventListener('click', () => {
+      trigger.addEventListener('click', (e) => {
```
bobylito: If you don't use the event, you should remove it from the arguments of the function.

docgen/src/stylesheets/components/_sidebar.scss
```diff
@@ -128,13 +130,14 @@ nav.sidebar {
...
   background-size: 60%, 100%;
-  z-index: 11;
+  z-index: 1000;
   border-radius: 50%;
   box-shadow: 0 6px 8px 0 rgba(0,0,0,0.14),0 1px 5px 0 rgba(0,0,0,0.12),0 3px 1px -2px rgba(0,0,0,0.2);
   transition: transform 0.2s ease, left 0.2s ease;
   transform-origin: center center;
   cursor: pointer;
   display: none;
+  pointer-events: all;
```
bobylito: `all` only works on SVG elements. What's the purpose?

([mdn](https://developer.mozilla.org/en/docs/Web/CSS/pointer-events))

docgen/src/stylesheets/components/_sidebar.scss
```diff
@@ -7,10 +7,12 @@ nav.sidebar {
   border-right: 1px solid #d8d8d8;
   overflow: hidden;
   padding-top: 120px;
+  user-select: none;
 
   .sidebar-container {
     width: 100%;
     max-width: $sidebar-width;
+    user-select: all;
```
bobylito: How is `user-select` used? What's the effect on the sidebar?

vvo:
cc @LukyVj comments awaits

docgen/src/stylesheets/components/_sidebar.scss
LukyVj: It's used to re-set back the `user-select: none` of above. 
<details></details>

bobylito: Was not clear... Why do we set it to none in the first place?

docgen/src/stylesheets/components/_sidebar.scss
bobylito: OK... browsers :)

vvo:
Will make the test pass and merge this

PR #1767, opened 2016-12-19
feat(SearchForFacetValues): connector functions + refinementList implementation

mthuret:
sfvv

You can run the storybook, to see the widget live. 

**New API entry:** 

- make the <RefinementList/> widget have the search for facet values: 
```
    <RefinementList attributeName="category" searchForFacetValues/>
```

- use search for facet values inside a custom RefinementList
   * a new exposed prop **searchForFacetValues**: a boolean that if set to true will add the `searchForFacetValues` to the provided props. 

   * a new provided prop **isFromSearch**: a boolean that says if the `items` props contains facets values from the global search, or searchForFacetValues results. 
   * a new provided prop **searchForFacetValues**: a function that take a query as a parameter. To call for searching facet values. 

- create a connector that will propose search for facet values:
  * a new **searchForFacetValuesResults** parameter inside the getProvidedProps
  * a new function **searchForFacetValues** to implement

```
export default createConnector({

  getProvidedProps(props, searchState, searchResults, metadata, searchForFacetValuesResults) {
        ... 
        return props;
  }

  searchForFacetValues(props, searchState, nextRefinement) {
    return {facetName: props.attributeName, query: nextRefinement};
  }
})
```

The next step about this PR are:
* should we keep the facetQuery in our searchState (on another PR if needed)
* [x] Should we merge both searchResults and searchForFacetValuesResults?
* [x] Merge #1765 and add tests to check the call from the helper to SFFV
* [x] Fix the naming once the API is stable
* [x] Add the jsdoc & guide
* [x] Add this feature into one of our example

### DIFF
--- a/packages/react-instantsearch/src/components/RefinementList.js
+++ b/packages/react-instantsearch/src/components/RefinementList.js
@@ -3,13 +3,20 @@ import {pick} from 'lodash';
 import translatable from '../core/translatable';
 import List from './List';
 import classNames from './classNames.js';
-
+import Highlight from '../widgets/Highlight';
+import SearchBox from '../components/SearchBox';
 const cx = classNames('RefinementList');
 
 class RefinementList extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {query: ''};
+  }
+
   static propTypes = {
     translate: PropTypes.func.isRequired,
     refine: PropTypes.func.isRequired,
+    searchForFacetValues: PropTypes.func,
     createURL: PropTypes.func.isRequired,
     items: PropTypes.arrayOf(PropTypes.shape({
       label: PropTypes.string.isRequired,
@@ -17,33 +24,43 @@ class RefinementList extends Component {
       count: PropTypes.number.isRequired,
       isRefined: PropTypes.bool.isRequired,
     })),
+    isFromSearch: PropTypes.bool.isRequired,
     showMore: PropTypes.bool,
     limitMin: PropTypes.number,
     limitMax: PropTypes.number,
   };
 
-  renderItem = item =>
+  selectItem = item => {
+    this.props.refine(item.value);
+  };
+
+  renderItem = item => {
+    const label = item._highlightResult
+      ? <Highlight attributeName="label" hit={item}/>
+      : item.label;
+
+    return (
       <label>
         <input
           {...cx('itemCheckbox', item.isRefined && 'itemCheckboxSelected')}
           type="checkbox"
           checked={item.isRefined}
-          onChange={() => this.props.refine(item.value)}
+          onChange={() => this.selectItem(item)}
         />
         <span {...cx('itemBox', 'itemBox', item.isRefined && 'itemBoxSelected')}></span>
         <span {...cx('itemLabel', 'itemLabel', item.isRefined && 'itemLabelSelected')}>
-          {item.label}
+          {label}
         </span>
         {' '}
         <span {...cx('itemCount', item.isRefined && 'itemCountSelected')}>
           {item.count}
         </span>
-      </label>
-    ;
+      </label>);
+  };
 
   render() {
-    return (
-      <List
+    const facets = this.props.items.length > 0
+      ? <List
         renderItem={this.renderItem}
         cx={cx}
         {...pick(this.props, [
@@ -54,10 +71,38 @@ class RefinementList extends Component {
           'limitMax',
         ])}
       />
+      : <div>{this.props.translate('noResults')}</div>;
+
+    const searchBox = this.props.searchForFacetValues ?
+      <div {...cx('SearchBox')}>
+        <SearchBox
+          currentRefinement={this.props.isFromSearch ? this.state.query : ''}
+          refine={value => {
+            this.setState({query: value});
+            this.props.searchForFacetValues(value);
+          }}
+          translate={this.props.translate}
+          onSubmit={e => {
+            e.preventDefault();
+            e.stopPropagation();
+            if (this.props.isFromSearch) {
+              this.selectItem(this.props.items[0]);
+            }
+          }}
+        />
+      </div> : null;
+
+    return (
+      <div>
+        {searchBox}
+        {facets}
+      </div>
     );
   }
 }
 
 export default translatable({
   showMore: extended => extended ? 'Show less' : 'Show more',
+  noResults: 'No Results',
 })(RefinementList);
+
--- a/packages/react-instantsearch/src/connectors/connectRefinementList.js
+++ b/packages/react-instantsearch/src/connectors/connectRefinementList.js
@@ -39,8 +39,6 @@ function getValue(name, props, searchState) {
   return nextRefinement;
 }
 
-const sortBy = ['isRefined', 'count:desc', 'name:asc'];
-
 /**
  * connectRefinementList connector provides the logic to build a widget that will
  * give the user tha ability to choose multiple values for a specific facet.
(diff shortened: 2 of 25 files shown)


packages/react-instantsearch/src/connectors/connectRefinementList.js
```diff
@@ -100,9 +100,23 @@ export default createConnector({
...
       }));
 
+    let facetValues;
+    if (facetValuesResults) {
+      facetValues = facetValuesResults[attributeName]
+        .slice(0, limit).map(
+          v => ({
+            label: v.value,
+            value: getValue(v.value, props, searchState),
+            _highlightResult: {label: {value: v.highlighted}},
```
mthuret: To use the Highlight widget when displaying the search for values results, I need to cheat a little to mimic the regular hits structure. Do you think it's fine to do that here, or should we:
- have the same structure between hits and facetValues in the helper regarding the highlighting
- have several implementations of the higlight in react-instantsearch

packages/react-instantsearch/src/connectors/connectRefinementList.js
```diff
@@ -100,9 +100,23 @@ export default createConnector({
...
+            count: v.count,
+            isRefined: v.isRefined,
+          }));
+    }
+
     return {
       items,
       currentRefinement: getCurrentRefinement(props, searchState),
+      facetValues,
```
vvo: This is the name where we need to iterate maybe, items are already facet values. We need to emphasis on the fact that those are search results of facet values right?

vvo: Should we also forward the facet search query?

vvo:
Could you add to the PR description the new public API additions and simple examples on how to use them?

vvo:
I am not sure how to name it yet, but in our own code, we need to be very specific as for what are the type of facet values we are talking about. Those are either facetValues of search results of facet values. Today the distinction is not obvious if you read the code.

packages/react-instantsearch/src/core/createConnector.js
```diff
@@ -143,9 +144,10 @@ export default function createConnector(connectorDesc) {
         error,
         widgets,
         metadata,
+        resultsFacetValues,
       } = store.getState();
       const searchState = {results, searching, error};
-      return connectorDesc.getProvidedProps.call(this, props, widgets, searchState, metadata);
+      return connectorDesc.getProvidedProps.call(this, props, widgets, searchState, metadata, resultsFacetValues);
```
vvo: Ideally, it could be nice to be able to get search results for facet values from the searchResults object of the helper. Is that feasible? cc @bobylito

packages/react-instantsearch/src/components/RefinementList.js
```diff
@@ -1,49 +1,94 @@
...
+            renderItem={this.renderItem}
+            items={sortedFacets}
+            cx={cx}
+            {...pick(this.props, [
+              'translate',
+              'limitMin',
+              'limitMax',
+            ])}
+          />
+          : <div>No Results</div>;
```
mthuret: Yes definitely.

packages/react-instantsearch/src/connectors/connectRefinementList.js
mthuret: maybe by just adding everytime the prefix `search`, WDYT?

vvo: Yep should work!

vvo: foundFacetValues
searchedFacetValues
facetValuesFromSeach
searchFacetValues (can be tricky, could be a function!)
itemsFromSearch

mthuret: I like searchedFacetValues but maybe itemsFromSearch is closer than what we already have.

mthuret: To do that we need to add the SFFV query inside the search state. 
Currently the SFFV query is handled directly inside the RefinementList widget by using state. It means that you can't share an url that will bring you to an InstantSearch page with a SFFV displayed. Do we want that?

Also adding it to the search state brings new questions. Basically I assumed we would want a structure like:
```
{
   refinementList: {category: {items: ['Table'], query: 'Ch'}}
}
```

But then when forwarding the `searchState` to the user, how could he easily know that it's not necessary to update the whole <InstantSearch/> (that would trigger a new search for nothing)?

packages/react-instantsearch/src/connectors/connectRefinementList.js
bobylito: Tricky one! I feel it would be better to have one that is distinct. But as you did it works and it is compatible with the current hits so it make more intuitive to use when building a custom version of this widget. Keep it like that it's great!

bobylito:
There is one thing I'm not sure to understand: why do we provide to the component two sets of values? Why can't we provide only items? Wouldn't that simplify the API for the connector user?

packages/react-instantsearch/src/core/createConnector.js
mthuret: One thing we could do for that is merging the `searchResults` object with the searchForFacetValuesResults` as soon as we get one or the other. 

One thing that bothers me though is that the `searchResults` object comes directly from the helper and in our documentation we are actually pointing to the helper documentation to explain its structure. Merging `searchResults` and `searchForFacetValuesResults` wouldn't be a little bit confusing? Can you see another ways?

bobylito: @vvo I'm not sure to get the question :D you propose to merge the results of the search for facet values (`resultsFacetValues`) into the results object (`results`)?

mthuret:
I followed your suggestion @bobylito -> we only have items now. I also added a boolean `isFromSearch` to know if a searchForFacetsValues has been performed or not. 

The next step about this PR are:
* should we keep the facetQuery in our `searchState` (also see my previous comment on that)?
* Should we merge both `searchResults` and `searchForFacetValues`?
* Fix the naming once the API is stable
* Add the jsdoc & guide
* Add this feature into one of our example?

packages/react-instantsearch/src/components/RefinementList.js
```diff
@@ -18,32 +25,44 @@ class RefinementList extends Component {
...
         </span>
-      </label>
-    ;
+      </label>);
+  };
 
   render() {
-    return (
-      <List
+    const sortedFacets = orderBy(this.props.items, ['isRefined', 'count', 'name'], ['desc', 'desc', 'asc']);
```
bobylito: The results of search for facet values is a "search" which means that the order defined is computed by the engine, as it would for a normal search. Therefore I wouldn't sort them again.

mthuret: Sorting here doesn't feel right. As discussed, there's no need to place the 'isRefined' items first in case of a SFFV. I'll remove that :)

packages/react-instantsearch/src/core/createConnector.js
vvo: Yep!

vvo:
Can you update the PR description with the latest API changes?

> should we keep the facetQuery in our searchState (also see my previous comment on that)?

Where is this comment? Cannot find it. Why would we do not want to have the facetQuery?

> Should we merge both searchResults and searchForFacetValuesResults?

I would say yes if that's doable AND if you are convinced too that it makes things easier for the user.

> Add this feature into one of our example?

Yes :)

vvo:
> I followed your suggestion @bobylito -> we only have items now

I was wondering this too. Do you think there was a usecase for having both? The only one I can think of is displaying the currently selected values along with the searchable ones but that's already doable via currentRefinement.

Showing both facet results from the main search AND facet result from facet search does not make sense to me. So we are aligned I guess here.

vvo:
> But then when forwarding the searchState to the user, how could he easily know that it's not necessary to update the whole (that would trigger a new search for nothing)?

Can you precise this a little bit more? Do we have the same issue with the regular instantsearch query?

mthuret:
Providing a `searchState` object trigger a new search after checking that the props are not equals. 
If it's the regular instantsearch query, it's fine because it's what you wan't. 

Now if we put the facetQuery inside the `searchState`, then it will go through the same process, and trigger a new global search. Which is not what we want (we only want the SFFV search). Therefore we need a way to differentiate both search. 

I think it's doable but it will change the way that the search is triggered now.

mthuret:
The feature should be complete. I updated the API inside the PR body so that you can see the last changes.

bobylito:
> should we keep the facetQuery in our searchState (on another PR if needed)

It's a local search state not a global one, we shouldn't keep it in the main state. IMHO.

vvo:
We can iterate on putting the searchQuery inside the state. Ultimately I would say our state should represent the complete search interface so that you can render the whole interface as it was, we can discuss this further in another iteration.

PR #1770, opened 2016-12-20
docs(examples): improve material-ui

mthuret:
UI/UX changes: 

* Add the SearchBox inside the AppBar
* The example can be used on mobile

material-ui

cc @seafoox


vvo:
NICE

PR #1784, opened 2016-12-29
Develop

maxiloc:


julienpa:
Looks good to me :)

julienpa:
Actually, it looks like the `cookie` lib is included twice, since in the next `<script>` there is this: `js-cookie@2.2.0(js.cookie.js)`
What do you think?

PR #1787, opened 2016-12-30
feat(InfiniteHits): Add class to load more button

adlenafane:

**Summary**

When styling load more button from InfiniteHits I want to have a specific class to it so I can specifically target it

**Result**

Tests are passing and are updated and wrapper now get the new classname instead of a generic button


bobylito:
Thanks a lot @adlenafane for your contribution, it means a lot to have a new contributor :)

PR #1788, opened 2016-12-31
docs(getting-started): use yarn instead of npm

l4u:


bobylito:
Hi @l4u Thanks for this PR, it is greatly appreciated :)

I was wondering if `yarn` should be presented as the only way or as one of the way. Not everyone is yet aware of the existence of yarn. What do you think?

l4u:
Related to https://github.com/facebookincubator/create-react-app/pull/1275

vvo:
Did a minor tweak, thanks.

PR #1796, opened 2017-01-03
docs(recipes): add page into docs website

mthuret:
recipes-docs


bobylito:
I've updated the layout to be able to see the title of the demo / recipe at all time. It gives user more context about what's going on.

Issue #1776, opened 2016-12-22
Highlighting for _snippetResult [react-instantsearch]


R4ph-t:
Hard to handle highlighting for **_snippetResult** values.  
The `<ais-highlight-`+`timestamp>` tag format makes things harder (than `<em>`) when you don't have a specialized component like the `Highlight` for `_highlightResult ` that can handle it.

mthuret:
There's two options here to solve this: 

1. We create a new widget called `<Snippet/>` that has the same interface as the `<Highlight/>` one except that it will look for `_snippetResult` instead of `_highlightResult`. 

2. We add an optional prop `path` to the `<Highlight/>` component that allow a user to specify it's own path to the right structure. Default would be `_highlighResult`. 

I tend to prefer the first one, I think it's clearer: if you want to display a snippet, you use the snippet widget. WDYT @vvo and @bobylito?

bobylito:
I vote for 1/ :)

vvo:
1/ for sure

PR #1797, opened 2017-01-03
feat(snippet): add a snippet widget to be able to highlight snippet results

mthuret:
You can now write this: `<Snippet attributeName="description" hit={hit}/>` and have highlighting for snippet attributes.

#1776

### DIFF
--- a/packages/react-instantsearch/src/core/highlight.js
+++ b/packages/react-instantsearch/src/core/highlight.js
@@ -1,29 +1,31 @@
 import {get} from 'lodash';
 
 /**
- * Find an highlighted attribute give a path `attributeName`, parses it,
+ * Find an highlighted attribute given an `attributeName` and an `highlightProperty`, parses it,
  * and provided an array of objects with the string value and a boolean if this
  * value is highlighted.
  *
- * In order to use this feature, highlight must be activated in the configruation of
+ * In order to use this feature, highlight must be activated in the configuration of
  * the index. The `preTag` and `postTag` attributes are respectively highlightPreTag and
  * highligtPostTag in Algolia configuration.
  *
  * @param {string} preTag - string used to identify the start of an highlighted value
  * @param {string} postTag - string used to identify the end of an highlighted value
- * @param {string} attributeName - path to the highlighted attribute in the results
+ * @param {string} highlightProperty - the property that contains the highlight structure in the results
+ * @param {string} attributeName - the highlighted attribute to look for
  * @param {object} hit - the actual hit returned by Algolia.
  * @return {object[]} - An array of {value: string, isDefined: boolean}.
  */
 export default function parseAlgoliaHit({
   preTag = '<em>',
   postTag = '</em>',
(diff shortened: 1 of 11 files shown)


packages/react-instantsearch/src/components/Highlight.js
```diff
@@ -1,19 +1,6 @@
...
-  const reactHighlighted = parsedHighlightedValue.map((v, i) => {
-    const key = `split-${i}-${v.value}`;
-    if (!v.isHighlighted) {
-      return <span key={key} className="ais-Highlight__nonHighlighted">{v.value}</span>;
-    }
-    return <em key={key} className="ais-Highlight__highlighted">{v.value}</em>;
-  });
-  return <span className="ais-Highlight">{reactHighlighted}</span>;
+export default function Highlight(props) {
+  return <Highlighter path="_highlightResult" {...props}/>;
```
bobylito: Did you notice that `path` can be overriden?

packages/react-instantsearch/src/components/Highlight.js
```diff
@@ -1,19 +1,6 @@
...
 }
-
-Highlight.propTypes = {
```
bobylito: How come we don't need to add the proptypes in this case?

packages/react-instantsearch/src/core/highlight.js
```diff
@@ -5,25 +5,27 @@ import {get} from 'lodash';
...
   preTag = '<em>',
   postTag = '</em>',
+  path,
   attributeName,
   hit,
 }) {
   if (!hit) throw new Error('`hit`, the matching record, must be provided');
 
-  const highlightObject = get(hit._highlightResult, attributeName);
+  const highlightObject = get(hit[path], attributeName);
```
bobylito: It's mostly a technicality but here `path` is not really a path but the property that contains the highlight structure. I suggest we change path to a name that explicitly state that the user can't go deep, idea: `highlightProperty`

packages/react-instantsearch/src/components/Highlight.js
mthuret: Yes. The path is not a prop documented in the Highlight/Snippet widget so I figured out it was ok to just forward the props. Do you think it's an issue?

packages/react-instantsearch/src/components/Highlight.js
mthuret: Well if I add them then you get two warnings if a props is missing. From the higlight, and from the highlighter. But I can definitely add them back.

packages/react-instantsearch/src/core/highlight.js
mthuret: Fine with me, I'll change for `highlightProperty`

packages/react-instantsearch/src/components/Highlight.js
bobylito: Should be ok 👍

Issue #1734, opened 2016-12-12
Prevent sort RefinementList onChange


Kikobeats:
Not sure if is possible, but desirable in my case.

Could be possible prevent put on top of the list the marked items?

![](https://d17oy1vhnax1f7.cloudfront.net/items/2J2x1D2V020s2M2z0L0I/Screen%20Recording%202016-12-12%20at%2012.11%20pm.gif?v=3e287e0e)

My RefinementList implementation is [here](https://github.com/windtoday/windtoday-app/blob/master/src/app/RefinementList/index.js).

bobylito:
Thanks @Kikobeats for the issue. This is done when sorting the values. This is not possible to change the order without using a connector.

vvo:
@bobylito @mthuret Do you think we should provide a generic entry to reformat the data? Not a sortBy, not a map. Just a simple method where you can either filter/reorder as you want based on the provided props.

I know we took the path to say "widgets => no options". But since we also made the move from "no css => good css", if we actually want people to use a bit of our widgets, maybe we could be a little bit more flexible here?

We do have SOME options (limit, showMore..). We just need to find the right set of options that handle many use cases, sortBy is limiting, let's make it a complete reduce (well, you will still need to send some array items, but you can do whatever you want with them).

mthuret:
I think this a good idea to bring some flexibility. 

With this, a widget user will also be able to push some pre-defined data. So actually there's is several use case for it. 

I didn't get this part though: 'default transformProvidedProps in lists should be alpha + count instead of isRefined + alpha + count?'?

vvo:
> I didn't get this part though: 'default transformProvidedProps in lists should be alpha + count instead of isRefined + alpha + count?'?

Today the default sort we do in lists is: if it's refined, it's on top, then we use alphanumeric + count sort. We should maybe just use alphanumeric sort so that users are not surprised..?

mthuret:
Maybe on the default behavior we can check before (with CSE?) what would be the one we want to push for? As long as we can override it, it should be fine, no?

bobylito:
The facets are dynamic based on the result set. Removing `isRefined` means that the selected values may disappear, which is also confusing for users. I think that there is no perfect solution here, but at least end-users (not devs) will be less confused by moving values than disappearing values.

bobylito:
Some thoughts on the topic:
[...]

PR #1804, opened 2017-01-04
feat(Menu, connectMenu): add search for facet values

mthuret:
menu-sffv

### DIFF
--- a/packages/react-instantsearch-theme-algolia/styles/_Menu.scss
+++ b/packages/react-instantsearch-theme-algolia/styles/_Menu.scss
@@ -15,6 +15,9 @@
   min-height: 22px;
   line-height: 22px;
 }
+.ais-Menu__item a {
+  outline: 0;
+}
 
 .ais-Menu__itemSelected {
   font-weight: bold;
@@ -50,4 +53,12 @@
 }
 
 .ais-Menu__itemLabelSelected{
+}
+
+.ais-Menu__SearchBox {
+  margin-bottom: 3px;
+}
+
+.ais-Menu__SearchBox .ais-SearchBox {
+  @include searchbox($sffv-searchbox-config...);
 }
--- a/packages/react-instantsearch/src/connectors/connectMenu.js
+++ b/packages/react-instantsearch/src/connectors/connectMenu.js
@@ -28,7 +28,7 @@ function getValue(name, props, searchState) {
   return name === currentRefinement ? '' : name;
 }
 
-const sortBy = ['count:desc', 'name:asc'];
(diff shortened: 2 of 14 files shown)


packages/react-instantsearch/src/connectors/connectMenu.js
```diff
@@ -28,7 +28,7 @@ function getValue(name, props, searchState) {
   return name === currentRefinement ? '' : name;
 }
 
-const sortBy = ['count:desc', 'name:asc'];
+const sortBy = ['isRefined', 'count:desc', 'name:asc'];
```
mthuret: We need to add with SFFV in order to keep track of the newly selected item. 
But for a regular menu it doesn't feel right to be the default behavior. 
WDYT?

bobylito: I think `isRefined` should be the default anytime we have a list of items because we don't know the number of expected items and the widget is not aware if the visibility of selected items is offered by another widget.

packages/react-instantsearch-theme-algolia/styles/_Menu.scss
```diff
@@ -51,3 +54,32 @@
 
 .ais-Menu__itemLabelSelected{
 }
+
+$sffv-searchbox-config:(
```
bobylito: you probably need to rename this variable to make it unique so it does not impact the other searchbox config.

packages/react-instantsearch/src/connectors/connectMenu.js
mthuret: Yes you're right. In any case we should definitely do something about this proposal https://github.com/algolia/instantsearch.js/issues/1734 (transformedProvidedProps), as it will helps to have both behavior easily. 
Also, if we sort by `isRefined` first, it will be a breaking change here.

packages/react-instantsearch-theme-algolia/styles/_Menu.scss
mthuret: Good catch, I'll move that to the style.scss file.

packages/react-instantsearch/src/connectors/connectMenu.js
bobylito: If it's a breaking change and we don't have a fallback, then I guess it's a no go for this change :) Maybe we could that change as a required change in #1734. What do you think @vvo ?

vvo: What we could do is change this sort order only if SFFV is activated, then it's not a breaking change.

vvo:
This is ON HOLD and will be reopened when ready

PR #1805, opened 2017-01-05
chore: remove react-native dependencies from the root

mthuret:


bobylito:
Looks like the lint is not passing anymore. Related to examples. Maybe we could disable the linter on examples?

mthuret:
Arf right, I forgot that it was ok locally, but not ok with travis :(

mthuret:
Well I can do what it's already done for some import: disable the linter only for them. 
I'll prefer to just disable the lint for the import than doing it globally even if those issues might still come up when adding more recipes.

vvo:
This might just be an issue about configuring eslint import well for examples, Ill have a look afterwards

Issue #1807, opened 2017-01-05
setImmediate is not defined on Meteor + React


Kolyasya:
react-instantsearch: 2.1.0
Latest versions of Chrome and FF.

Hello, I've just started setup react-instantsearch on my Meteor project and have strange bug.

I'm following steps from the React Instant Search Getting Started Tutorial and when I insert first widget (`<Hits />` for example) I have an error in console:

```
modules.js?hash=35e875d…:110624 Uncaught ReferenceError: setImmediate is not defined
    at scheduleUpdate (modules.js?hash=35e875d…:110624)
    at Object.registerWidget (modules.js?hash=35e875d…:110633)
    at new Connector (modules.js?hash=35e875d…:109693)
    at modules.js?hash=35e875d…:11541
    at measureLifeCyclePerf (modules.js?hash=35e875d…:11321)
    at ReactCompositeComponentWrapper._constructComponentWithoutOwner (modules.js?hash=35e875d…:11540)
    at ReactCompositeComponentWrapper._constructComponent (modules.js?hash=35e875d…:11526)
    at ReactCompositeComponentWrapper.mountComponent (modules.js?hash=35e875d…:11434)
    at Object.mountComponent (modules.js?hash=35e875d…:4461)
    at ReactDOMComponent.mountChildren (modules.js?hash=35e875d…:10623)
```

It refers to this code:
```
exports.default = createWidgetsManager;                                                                                // 6
function createWidgetsManager(onWidgetsUpdate) {                                                                       // 7
  var widgets = [];                                                                                                    // 8
  // Is an update scheduled?                                                                                           // 9
  var scheduled = false;                                                                                               // 10
                                                                                                                       // 11
  // The state manager's updates need to be batched since more than one                                                // 12
  // component can register or unregister widgets during the same tick.                                                // 13
  function scheduleUpdate() {                                                                                          // 14
    if (scheduled) {                                                                                                   // 15
      return;                                                                                                          // 16
    }                                                                                                                  // 17
    scheduled = true;                                                                                                  // 18
    setImmediate(function () {                                                                                         // 19
      scheduled = false;                                                                                               // 20
      onWidgetsUpdate();                                                                                               // 21
    });                                                                                                                // 22
  }                    
```

mthuret:
Hi @Kolyasya,

[...]

PR #1811, opened 2017-01-06
fix(widgets): replace setImmediate use with Promise use when update is needed

mthuret:
See https://github.com/algolia/instantsearch.js/issues/1807 => usage of react-instantsearch is not possible with Meteor without the use of a `setImmediate` polyfill.

Previously we already add a problem using `process.nextTicks` with react-native.

I tested this version with our react-native example and also a Meteor small app.

### DIFF
--- a/packages/react-instantsearch/src/core/createInstantSearchManager.test.js
+++ b/packages/react-instantsearch/src/core/createInstantSearchManager.test.js
@@ -129,18 +129,19 @@ describe('createInstantSearchManager', () => {
         algoliaClient: client,
       });
 
-      ism.widgetsManager.registerWidget({getMetadata: () => ({id: 'a'})});
-      ism.widgetsManager.registerWidget({getMetadata: () => ({id: 'b'})});
-      ism.widgetsManager.registerWidget({getMetadata: () => ({id: 'c'})});
-      ism.widgetsManager.registerWidget({getMetadata: () => ({id: 'd'})});
-
       const widgetIDsT0 = ism.getWidgetsIds().sort();
       expect(widgetIDsT0).toEqual([]);
 
-      jest.runAllTimers();
+      ism.widgetsManager.registerWidget({getMetadata: () => ({id: 'a'})});
+      ism.widgetsManager.registerWidget({getMetadata: () => ({id: 'b'})});
+      ism.widgetsManager.registerWidget({getMetadata: () => ({id: 'c'})});
 
-      const widgetIDsT1 = ism.getWidgetsIds().sort();
-      expect(widgetIDsT1).toEqual(['a', 'b', 'c', 'd']);
+      ism.widgetsManager.registerWidget({getMetadata: () => ({id: 'd'})});
+
+      return Promise.resolve().then(() => {
+        const widgetIDsT1 = ism.getWidgetsIds().sort();
+        expect(widgetIDsT1).toEqual(['a', 'b', 'c', 'd']);
+      });
     });
   });
 });
--- a/packages/react-instantsearch/src/core/createWidgetsManager.test.js
+++ b/packages/react-instantsearch/src/core/createWidgetsManager.test.js
@@ -20,29 +20,23 @@ describe('createWidgetsManager', () => {
     });
 
     it('schedules an update', () => {
-      jest.useFakeTimers();
       const onUpdate = jest.fn();
       const wm = createWidgetsManager(onUpdate);
-      jest.runAllImmediates();
       wm.registerWidget({});
-      expect(onUpdate.mock.calls.length).toBe(0);
-      jest.runAllImmediates();
-      expect(onUpdate.mock.calls.length).toBe(1);
-      jest.useRealTimers();
+      return Promise.resolve().then(() => {
+        expect(onUpdate.mock.calls.length).toBe(1);
+      });
     });
   });
 
   describe('update', () => {
     it('schedules an update', () => {
-      jest.useFakeTimers();
       const onUpdate = jest.fn();
       const wm = createWidgetsManager(onUpdate);
-      jest.runAllImmediates();
       wm.update();
-      expect(onUpdate.mock.calls.length).toBe(0);
-      jest.runAllImmediates();
-      expect(onUpdate.mock.calls.length).toBe(1);
-      jest.useRealTimers();
+      return Promise.resolve().then(() => {
+        expect(onUpdate.mock.calls.length).toBe(1);
+      });
     });
   });
 });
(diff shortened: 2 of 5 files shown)


packages/react-instantsearch/src/core/createInstantSearchManager.test.js
```diff
@@ -121,26 +121,32 @@ describe('createInstantSearchManager', () => {
...
-      expect(widgetIDsT0).toEqual([]);
-
-      jest.runAllTimers();
+      ism.widgetsManager.registerWidget({getMetadata: () => ({id: 'd'})});
 
-      const widgetIDsT1 = ism.getWidgetsIds().sort();
-      expect(widgetIDsT1).toEqual(['a', 'b', 'c', 'd']);
+      Promise.resolve().then(() => {
+        try {
+          const widgetIDsT1 = ism.getWidgetsIds().sort();
```
bobylito: Why are you using a resolved promise here? Is it to be sure that the code executed after everything else?

mthuret: Yes and you need the try/catch to catch the error if the expect fails.

packages/react-instantsearch/src/core/utils.js
```diff
@@ -58,3 +58,6 @@ export function assertFacetDefined(searchParameters, searchResults, facet) {
 export function getDisplayName(Component) {
   return Component.displayName || Component.name || 'UnknownComponent';
 }
+
+const resolved = typeof Promise !== 'undefined' && Promise.resolve();
+export const defer = resolved ? f => { resolved.then(f); } : setTimeout;
```
bobylito: Why do we base our implementation on promises? Why not just use setTimeout(..., 0)?

```js
defer = f => { setTimeout(f,0); }
```

vvo: After some searching it seems the current fastest way to express "Please execute this code asynchronously as soon as possible" is to actually use Promise.resolve().then().

https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules and https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/#comment-2926206647

I also asked @developit for his advice on this, we can wait a bit before merging.

packages/react-instantsearch/src/core/createInstantSearchManager.test.js
vvo: Did you try to write a simple test case without any of our code just to try return Promise.resolve().then(throw..)? maybe there's a bug somewhere or a misunderstanding of ours on how promise works with jest

packages/react-instantsearch/src/core/utils.js
mthuret: It uses the preact implementation for doing stuff async as soon as possible. The guy even publish a benchmark of the different options.
See: https://github.com/developit/preact/commit/6954d7549f35b5c32421c6855ee29ffda01137d1 and https://esbench.com/bench/55d4d44e2efbca1100bf7251

packages/react-instantsearch/src/core/utils.js
```diff
@@ -58,3 +58,6 @@ export function assertFacetDefined(searchParameters, searchResults, facet) {
 export function getDisplayName(Component) {
   return Component.displayName || Component.name || 'UnknownComponent';
 }
+
+const resolved = typeof Promise !== 'undefined' && Promise.resolve();
```
vvo: We can agree on supporting only browsers with Promise.resolve(), http://caniuse.com/#feat=promises

And delegate promise polyfill to the user.

packages/react-instantsearch/src/core/utils.test.js
```diff
@@ -7,10 +7,13 @@ import {
   capitalize,
   assertFacetDefined,
   getDisplayName,
+  defer,
 } from './utils';
 
 import {SearchParameters, SearchResults} from 'algoliasearch-helper';
 
+jest.useFakeTimers();
```
vvo: Do you need fake timers in the end>?

packages/react-instantsearch/src/core/utils.js
vvo: Note: on the benchmark you might get better results with other implementations but Promise.resolve() is widely supported while others are not

vvo: image

packages/react-instantsearch/src/core/utils.js
mthuret: what about IE 11?

packages/react-instantsearch/src/core/utils.test.js
mthuret: No need here, thanks.

packages/react-instantsearch/src/core/utils.js
vvo: We will explain how to include a polyfill if that's necessary.

packages/react-instantsearch/src/core/utils.js
bobylito: Ok :) I was just curious. It's a very hacky way to use promises :) Especially since we are going to reuse the same promise mutliple times :)

mthuret:
Should be ok now. We'll add a mention for the polyfill in the doc if needed.

developit:
Just wanted to drop this here in case you guys were interested in what Preact did before switching to the Promise method:

https://github.com/developit/preact/commit/6954d7549f35b5c32421c6855ee29ffda01137d1

```js
let ch;
try { ch = new MessageChannel(); } catch (e) {}
export const defer = ch ? ( f => {
  ch.port1.onmessage = f;
  ch.port2.postMessage('');
}) : setTimeout;
```

bobylito:
Thanks @developit for the infos. I am truly amazed by those hacks, very enlightening :D

packages/react-instantsearch/src/core/createWidgetsManager.test.js
```diff
@@ -20,29 +20,23 @@ describe('createWidgetsManager', () => {
...
       const onUpdate = jest.fn();
       const wm = createWidgetsManager(onUpdate);
-      jest.runAllImmediates();
       wm.update();
-      expect(onUpdate.mock.calls.length).toBe(0);
-      jest.runAllImmediates();
-      expect(onUpdate.mock.calls.length).toBe(1);
-      jest.useRealTimers();
+      return Promise.resolve().then(() => {
+        expect(onUpdate.mock.calls.length).toBe(1);
```
vvo: I find this even better than the previous run all timers method

PR #1812, opened 2017-01-06
transfer a user defined object to hitComponent

kizdolf:
- react-instantsearch

**Summary**

When using multi index is usual to want to pass informations of the index to the `hitComponent`.
But there is no way to transfer props from a `Hits` component to a personal `hitComponent`. Only `hits` are passed.  

**Result**

and then you can do stuff like
```js
<Hits userObj={{index: algoliaConf.indexNames[i]}} hitComponent={AlgoliaHit}/>
...
export class AlgoliaHit extends Component {

  constructor(props) {
    super(props)
    this.handleClick = this.handleClick.bind(this)
  }

  handleClick(e) {
    e.preventDefault()
    console.log('click from index: ', this.props.userObj.index);
  }

  render() {
    return (
      <div style={{marginTop: 10}} onClick={this.handleClick}>
        <span className="hit-name">
          <Highlight attributeName="title" hit={this.props.hit} />
        </span>
      </div>
    )
  }
}


```

### DIFF
--- a/packages/react-instantsearch/src/components/Hits.js
+++ b/packages/react-instantsearch/src/components/Hits.js
@@ -5,11 +5,11 @@ const cx = classNames('Hits');
 
 class Hits extends Component {
   render() {
-    const {hitComponent: ItemComponent, hits} = this.props;
+    const {hitComponent: ItemComponent, hits, userObj} = this.props;
     return (
       <div {...cx('root')}>
         {hits.map(hit =>
-          <ItemComponent key={hit.objectID} hit={hit} />
+          <ItemComponent key={hit.objectID} hit={hit} userObj={userObj}/>
         )}
       </div>
     );


mthuret:
Thanks for your contribution @kizdolf 

Did you try to use our [connectors](https://community.algolia.com/instantsearch.js/react/guide/Connectors.html) and more especially the [`connectHits` connector](https://community.algolia.com/instantsearch.js/react/connectors/connectHits.html)? 

I think you can achieve the behavior you want by using it like this: 

```js
import {connectHits} from 'react-instantsearch/connectors';

<InstantSearch
    appId=""
    apiKey=""
    indexName=""
>
    <CustomHits indexName=""/>
</InstantSearch>;

const CustomHits = connectHits(props => {
  const hits = props.hits.map(hit => <p>{hit.name}</p>);
  return <div>
    <p>Index used is : {props.indexName}</p>
    {hits}
  </div>;
});
```

As you can see the `connectHits` connector has the `hits` props to display hits but you can also forward to it any props you want, so I added the `indexName` one. 

Let me know if this can solve your use case :)

kizdolf:
Thanks @mthuret ,

This is a really nice solution, and it solve my issue perfectly. Thanks :)
And now I start to see how connectors can be used, will be usefull!

I close the request, thanks for the example, have a nice day!

PR #1816, opened 2017-01-09
chore(deps): Put dependencies exactly where they need to be

vvo:
Before this commit a lot of needed dependencies were installed at the
root level to avoid repeating them. This leads to issues in linting but
also in running the examples because some dependencies were not linked
properly.


package.json
```diff
@@ -35,67 +35,63 @@
...
-    "material-ui": "^0.16.5",
+    "material-ui": "^0.16.6",
     "metalsmith": "^2.3.0",
     "metalsmith-headings": "^0.1.0",
     "metalsmith-layouts": "^1.7.0",
     "metalsmith-navigation": "^0.2.9",
     "metalsmith-sass": "^1.3.0",
     "ms-webpack": "^2.0.0-alpha.1",
     "mversion": "^1.10.1",
-    "node-sass": "^3.13.0",
```
bobylito: We don't use sass anymore?

packages/react-instantsearch/examples/autocomplete/package.json
```diff
@@ -8,8 +8,8 @@
     "react-autosuggest": "^7.0.1",
     "react": "^15.3.2",
     "react-dom": "^15.3.2",
-    "react-instantsearch": "file:../../dist",
-    "react-instantsearch-theme-algolia": "file:../../../react-instantsearch-theme-algolia"
+    "react-instantsearch": "latest",
```
bobylito: Will this rule make use of linked projects (npm / yarn link)?

packages/react-instantsearch/examples/react-native/package.json
```diff
@@ -2,15 +2,12 @@
 	"name": "reactnative",
 	"version": "0.0.1",
 	"private": true,
-	"scripts": {
-		"start": "node node_modules/react-native/local-cli/cli.js start"
```
bobylito: So we don't need this command anymore? How to run the project?

package.json
vvo: It's bundled in metalsmith-sass

packages/react-instantsearch/examples/autocomplete/package.json
vvo: Never tried yet, I did not yet address the problem of "I want to develop on this example with auto reload of react-instantsearch code". It was not handled before this commit too.

We will need to link dependencies via a script maybe.

packages/react-instantsearch/examples/react-native/package.json
vvo: It's inside the readme, since it's platform dependant (run-ios or run-android). Works?

packages/react-instantsearch/examples/autocomplete/package.json
bobylito: Oh ok, I was just curious of the possibilities of `latest`

packages/react-instantsearch/examples/react-native/package.json
bobylito: WFM indeed.

bobylito:
I relaunched the test but I'm not confident it will make things better :(

packages/react-instantsearch/src/core/utils.test.js
```diff
@@ -46,12 +46,10 @@ describe('utils', () => {
       const searchParameters = new SearchParameters({
         disjunctiveFacets: ['facet'],
       });
-      const searchResults = new SearchResults(searchParameters, {
-        results: [{
-          nbHits: 100,
-          facets: {},
-        }],
-      });
+      const searchResults = new SearchResults(searchParameters, [{
```
vvo: @bobylito FWIW this was needed since recent changes in the helper.

packages/react-instantsearch/src/core/createInstantSearchManager.result.test.js
```diff
@@ -10,7 +10,7 @@ jest.useFakeTimers();
 jest.mock('algoliasearch-helper/src/algoliasearch.helper.js', () => {
   let count = 0;
   const Helper = require.requireActual('algoliasearch-helper/src/algoliasearch.helper.js');
-  Helper.prototype._handleResponse = function(state) {
+  Helper.prototype._dispatchAlgoliaResponse = function(state) {
```
vvo: We rely on internal implementation for tests, need to change that

packages/react-instantsearch/src/core/createInstantSearchManager.test.js
```diff
@@ -24,15 +24,16 @@ const response = {
   ],
 };
 client.search = jest.fn((queries, cb) => {
+  const clonedResponse = JSON.parse(JSON.stringify(response));
```
vvo: Example of tests updates I had to do since new helper cc @bobylito, I cannot explain it right now but without cloning the Algolia response the tests would fail like if something was reusing the response reference

bobylito: Yep the Helper assumes that the response comes from a source that will always return new instances. :( Sorry for the work that it created to you :(

vvo: It's fine thanks for the explanation!

packages/react-instantsearch/src/core/createInstantSearchManager.error.test.js
```diff
@@ -51,21 +48,26 @@ describe('createInstantSearchManager', () => {
 
         expect(ism.store.getState().error).toBe(null);
 
-        jest.runAllTimers();
+        return Promise.resolve().then(() => {
+          jest.runAllTimers();
```
bobylito: Is it really necessary to have both? Is it because of the use of both timeouts and promises?

packages/react-instantsearch/src/core/createInstantSearchManager.result.test.js
bobylito: 👍

scripts/upgrade-deps.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
```
bobylito: I ❤️  this script :)

packages/react-instantsearch/src/core/createInstantSearchManager.error.test.js
vvo: Yes the runAllTimers is for the JS client mock (setImmediate), the return Promise is for the widget manager. Works?

scripts/upgrade-deps.sh
vvo: Cool!

packages/react-instantsearch/src/core/createInstantSearchManager.error.test.js
bobylito: Wfm of course :) let's refactor that later (not this but what is causing this)

vvo: Clearly we will refactor this yes

PR #1817, opened 2017-01-09
Update meta.pug

neilrichler:
updated title and meta description

**Summary**

**Result**

### DIFF
--- a/docgen/layouts/common/meta.pug
+++ b/docgen/layouts/common/meta.pug
@@ -7,7 +7,7 @@ html
     meta(content='width=device-width,initial-scale=1.0,minimum-scale=1.0,maximum-scale=1.0,user-scalable=no', name='viewport')
     link(rel='icon', href="assets/img/react-instant-favicon.png")
     meta(content='IE=edge,chrome=1', http-equiv='X-UA-Compatible')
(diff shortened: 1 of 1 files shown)


docgen/layouts/common/meta.pug
```diff
@@ -24,6 +24,6 @@ html
     meta(content='Boost search on your React apps with Algolia', property='og:description')
     meta(content='React InstantSearch', property='og:site_name')
 
-    title React InstantSearch | #{title}
+    title React InstantSearch | Lightning Fast React and React Native Search
```
bobylito: So we're clear here, this means that all the pages will have the same title. Is that what you want @neilrichler

vvo: Ah yep now probably not what we want, @neilrichler if you want that tagline only on the homepage, we can do that instead. Let us know.

Also the curren tagline on website homepage is "Lightning-fast search for React and React native apps" should we use that instead? In any case both should be consistent.

neilrichler: Got it - yes, let's only update it for the home page for now. I'm OK to use the current tagline, but we might want to adjust this in the future.

vvo: Gonna do it right now then, thanks

PR #1822, opened 2017-01-11
feat(Menu, connectMenu): add search for facet values

mthuret:
menu-sffv

### DIFF
--- a/packages/react-instantsearch/src/components/Menu.js
+++ b/packages/react-instantsearch/src/components/Menu.js
@@ -3,6 +3,7 @@ import {pick} from 'lodash';
 import translatable from '../core/translatable';
 import List from './List';
 import Link from './Link';
+import Highlight from '../widgets/Highlight';
 import classNames from './classNames.js';
 
 const cx = classNames('Menu');
@@ -11,6 +12,7 @@ class Menu extends Component {
   static propTypes = {
     translate: PropTypes.func.isRequired,
     refine: PropTypes.func.isRequired,
+    searchForFacetValues: PropTypes.func,
     createURL: PropTypes.func.isRequired,
     items: PropTypes.arrayOf(PropTypes.shape({
       label: PropTypes.string.isRequired,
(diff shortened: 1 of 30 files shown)


packages/react-instantsearch/src/connectors/connectMenu.js
```diff
@@ -77,19 +77,37 @@ export default createConnector({
...
+    const isFromSearch = Boolean(searchForFacetValuesResults && searchForFacetValuesResults[attributeName]);
+    const searchForFacetValues = props.searchForFacetValues ? this.searchForFacetValues : undefined;
+
+    const sortedItems = !isFromSearch && props.searchForFacetValues ?
+      orderBy(items, ['isRefined', 'count', 'label'], ['desc', 'desc', 'asc']) : items;
+    const transformedItems = props.transformItems ? props.transformItems(sortedItems) : sortedItems;
 
     return {
-      items,
+      items: transformedItems.slice(0, limit),
```
vvo: Since inside transform items we allow users to completely change the items, I would say once the user provides a transformItems method we do not slice based on the limit and we warn that using limit AND transform items is not compatible.

Otherwise we will have usages like:

```
RefinementList
  transformItems=...
  limit=Infinity
```

Plus the fact that it does not make sense to allow the user doing a full filter and map on the whole items to then ditch items he mapped (If this is not clear, yell at me) :D

vvo: After second thought, the behaviour you implemented is the right one because when being given items to reorder, you do need to loop through ALL of them before being able to display only ten of them so we are safe in doing this slice AFTER transformItems

packages/react-instantsearch/src/components/Menu.js
```diff
@@ -26,14 +29,17 @@ class Menu extends Component {
 
   renderItem = item => {
     const {refine, createURL} = this.props;
+    const label = item._highlightResult
```
vvo: Second read: this is not an issue of always using Highlight but an issue of switching to isSearch ? `<Highlight..` : item.label;

mthuret:
@vvo I think this should be ok now :)

Issue #1673, opened 2016-12-06
InstantSearch props are not all updated when they change


mthuret:
**Current behavior**

 the only props that are updated if changed on the `InstantSearch` component are `searchState`, `onSearchStateChange` and `createUrl`.

All the others: `appId`, `apiKey`, `indexName` and `searchParameters` can't be changed once the InstantSearch is mounted. 

This behavior is not natural as you expect changes in props to have repercussion on InstantSearch.

**Expected Behavior**

Props are updated accordingly in new ones are provided.

mthuret:
See this also #1618

bobylito:
Proposed plan:
 - [ ] make indexName configurable
 - [ ] create `<Configure />`
 - [ ] helper.setClient
 - [ ] helper.getClient
 - [ ] forward `setClient` in the InstantSearchManager

PR #1828, opened 2017-01-12
react-instantsearch - Update all props on InstantSearch

### LINKED ISSUE #1825: <Configure distinct={1}/>
Create a widget that has no rendering but provide a declarative API to configuration options of the queries.

bobylito:
**Summary**

This PR aimed at letting the developers update the parameters of `<InstantSearch/>`. Doing so, we also decided to deprecate the searchParameters props and replace it with a `<Configure/>` widget. 

Fixes #1825 #1673 

**Result**

The tourism demo app is now GeoIP based:

screenshot 2017-01-12 15 29 54

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "@kadira/storybook-addon-notes": "^1.0.1",
     "@kadira/storybook-addon-options": "^1.0.1",
     "@kadira/storybook-addons": "^1.6.1",
-    "algoliasearch-helper": "^2.18.0",
     "async": "^2.1.4",
     "autoprefixer": "^6.6.1",
     "babel-cli": "^6.18.0",
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "@kadira/storybook-addon-notes": "^1.0.1",
     "@kadira/storybook-addon-options": "^1.0.1",
     "@kadira/storybook-addons": "^1.6.1",
+    "algoliasearch-helper": "^2.18.0",
     "async": "^2.1.4",
     "autoprefixer": "^6.6.1",
     "babel-cli": "^6.18.0",
--- a/packages/react-instantsearch/src/core/InstantSearch.js
+++ b/packages/react-instantsearch/src/core/InstantSearch.js
@@ -23,7 +23,7 @@ function validateNextProps(props, nextProps) {
  * @propType {string} appId - The Algolia application id.
  * @propType {string} apiKey - Your Algolia Search-Only API key.
  * @propType {string} indexName - The index in which to search.
- * @propType {object} [searchParameters] - Object containing query parameters to be sent to Algolia. It will be overriden by the search parameters resolved via the widgets. Typical use case: setting the distinct setting is done by providing an object like: `{distinct: 1}`. For more information about the kind of object that can be provided on the [official API documentation](https://www.algolia.com/doc/rest-api/search#full-text-search-parameters). Read the [search parameters guide](guide/Search_parameters.html).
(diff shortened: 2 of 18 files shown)


package.json
```diff
@@ -25,6 +25,7 @@
     "@kadira/storybook-addon-notes": "^1.0.1",
     "@kadira/storybook-addon-options": "^1.0.1",
     "@kadira/storybook-addons": "^1.6.1",
+    "algoliasearch-helper": "^2.18.0",
```
vvo: Is this needed inside the root package.json?

packages/react-instantsearch/src/core/InstantSearch.js
```diff
@@ -61,6 +61,11 @@ class InstantSearch extends Component {
 
   componentWillReceiveProps(nextProps) {
     validateNextProps(this.props, nextProps);
+
+    if (this.props.indexName !== nextProps.indexName) {
+      this.aisManager.updateIndex(nextProps.indexName);
```
vvo: 💥

packages/react-instantsearch/src/core/createInstantSearch.js
```diff
@@ -14,15 +14,34 @@ export default function createInstantSearch(defaultAlgoliaClient, root) {
       algoliaClient: PropTypes.object,
       appId: PropTypes.string,
       apiKey: PropTypes.string,
+      children: React.PropTypes.oneOfType([
```
vvo: Nice, how come we never had any warning on this one? Thanks

packages/react-instantsearch/src/core/createInstantSearch.js
```diff
@@ -14,15 +14,34 @@ export default function createInstantSearch(defaultAlgoliaClient, root) {
...
+      super();
+      this.client = props.algoliaClient || defaultAlgoliaClient(props.appId, props.apiKey);
+    }
+
+    componentWillReceiveProps(nextProps) {
+      const props = this.props;
+      if (nextProps.algoliaClient) {
+        this.client = nextProps.algoliaClient;
+      } else if (props.appId !== nextProps.appId || props.apiKey !== nextProps.apiKey) {
+        this.client = defaultAlgoliaClient(nextProps.appId, nextProps.apiKey);
```
vvo: 💥

vvo:
You can use `cd packages/react-instantsearch && yarn upgrade-interactive` sadly the upgrade.sh seems bugged because when we want to quit one upgrade-interactive without making any changes we are not authorized to do so.

packages/react-instantsearch/src/core/createInstantSearch.js
bobylito: because {...props}. We used to forward any props therefore the linter couldn't know about that.

package.json
bobylito: I guess not /o\

bobylito:
Yep doing the proper package changes. :)

packages/react-instantsearch/src/core/createInstantSearchManager.js
```diff
@@ -41,6 +41,11 @@ export default function createInstantSearchManager({
     searching: false,
   });
 
+  function updateClient(client) {
+    helper.setClient(client);
+    search(); // Will it work?
```
mthuret: Being curious: what this comment mean?

bobylito: Note for later: I should remove my own questioning.

Joke aside, I haven't done a proper and complete test on this.

Also I would be interested in your opinion here @vvo @mthuret, should we trigger a new search when the client or the indexName changes. (if not I will cry because I won't know how to test). FYI, we do trigger a new search for new <Configure /> values.

bobylito: And I won't remove the comment until I have your opinion (because GH will hide it if there is a change at this line)

vvo: Yes we should trigger a new search when client and indexName are updated because both can have an impact on the result set:
- changing the index (obvious one)
- updating the client: going from logged off to logged in and having a dedicated API key that will update the result set.

Works?

bobylito: Just wanted to be sure because we've discussions about related topics. Fine with you @mthuret ?

mthuret: Yep all fine 👍

bobylito:
I think I'm done here :) Thanks for the reviews :)

PR #1832, opened 2017-01-13
docs(demo): add new example to illustrate the algolia ris tutorial

mthuret:
demo-react

This is the demo that will illustrate our React InstantSearch tutorial on the Algolia documentation website. 

Some comments:
- I replaced the button that was clearing the facets and the query by our `<ClearAll/>` widget that only clear the facets. I added a note to discuss about adding another widget that will do both. Otherwise we should probably go for a connector here (but that's not that simple anymore).  
- I used the `rc-slider` component for the range. It allowed me to have tooltip out of the box.


bobylito:
Hi @mthuret I don't think that's the right place for this demo :( Sorry I should have mentionned that earlier :( Can you make your PR [here](https://github.com/algolia/examples)?

mario-castle

mthuret:
Will be push in another repo.

PR #1833, opened 2017-01-13
Add React Router v4 example

zackify:

**Summary**

The current example doesn't show how to use v4 of react router, figured I'd make an example that works the exact same as the v3 one :)


bobylito:
Thanks a lot @zackify :)

zackify:
Just excited I could help on something. Maybe I'll be able to do more in the coming weeks!

mthuret:
Hi @zackify, thank you very much for your contribution!! ❤️ 

Can you point us the documentation of `replaceWith` and `transitionTo` of react-router v4? We try to find it but with no luck :(

Also if you can run `yarn fix:lint` on your branch it would be super great :)

zackify:
So I found it in the code itself, it's the only way to programmatically change the path. It's mentioned in this issue and closed by a core contributor. https://github.com/ReactTraining/react-router/issues/4220. I'll run fix lint right now.

mthuret:
The only thing with this is that if it's not officially documented that could go away with the stable version.
Also, it seems that the PR to get the router with an HOC https://github.com/ReactTraining/react-router/pull/4044 has been closed without a merge :(.

There's still some stuff not passing with CI, I'll check and fix it tomorrow :)

zackify:
Yes, but the thing is that many people are using it already so I figured it would be useful. If the api changes I can come back and update this for you. Hope that helps. The CI stuff wasn't from me (pretty sure), I checked that out when it first failed, so that's weird :)

mthuret:
Sure @zackify!  I encountered those linter issues with the previous recipes but it should have been fixed. I'll check :)

PR #1836, opened 2017-01-16
docs(footer): only init twitter tweets on homepage

vvo:


bobylito:
Protip: use ?w=1 to read this PR https://github.com/algolia/instantsearch.js/pull/1836/files?diff=split&w=1

vvo:
> Protip: use ?w=1

This should be the default, did not know this

PR #1838, opened 2017-01-16
Update algolia logos

LukyVj:


LukyVj:
ping @vvo

pixelastic:
Could you add a few screenshots of how it looks like on the website for future PR like this? Makes changes in design much easier to understand.

vvo:
Good point, I am also close to have automatic PR deployed (for RIS)

LukyVj:
Oh yeah my bad. Here is what I updated  :

PR #1840, opened 2017-01-16
fix(react-router): search was triggered too many times

mthuret:
- add deepEqual instead of shallowEqual comparaison when dealing with new widgets props. 
- also compare at react-router example level if the App should update because of a `render` that is triggered even if the `onSearchStateChange`. 

Also I'm not sure if we need yet the check for equality when passing new props to `<InstantSearch/>` so I didn't included it in this PR.

### DIFF
--- a/packages/react-instantsearch/examples/react-router/src/App.js
+++ b/packages/react-instantsearch/examples/react-router/src/App.js
@@ -7,6 +7,7 @@ import {
 import {withRouter} from 'react-router';
 import 'react-instantsearch-theme-algolia/style.css';
 import qs from 'qs';
+import {isEqual} from 'lodash';
 
 class App extends Component {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch/examples/react-router/src/App.js
```diff
@@ -18,6 +19,10 @@ class App extends Component {
     this.setState({searchState: qs.parse(this.props.router.location.query)});
   }
 
+  shouldComponentUpdate(nextProps, nextState) {
+    return !isEqual(this.state.searchState, nextState.searchState);
```
vvo: Do we need to do a deep comparison here or a simple === is sufficient?

mthuret: No it's not but maybe you have another idea to fix it?

vvo: Sorry I did not take enough time to read this PR, this is the right fix.

PR #1844, opened 2017-01-17
fix(InstantSearch): add specific `react-instantsearch ${version}` agent

vvo:
Before this commit we were not identifying react-instantsearch
in Algolia analytics, now we are.

### DIFF
--- a/packages/react-instantsearch/src/core/createInstantSearch.test.js
+++ b/packages/react-instantsearch/src/core/createInstantSearch.test.js
@@ -1,100 +1,61 @@
 /* eslint-env jest, jasmine */
 /* eslint-disable no-console */
 import React from 'react';
-import {mount} from 'enzyme';
-
-import algoliaClient from 'algoliasearch';
+import {shallow} from 'enzyme';
 
 import createInstantSearch from './createInstantSearch';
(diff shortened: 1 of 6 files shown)


vvo:
Also took the opportunity to add more tests on this part and actually check references, usages

packages/react-instantsearch/src/core/createInstantSearch.test.js
```diff
@@ -1,100 +1,61 @@
 /* eslint-env jest, jasmine */
 /* eslint-disable no-console */
 import React from 'react';
-import {mount} from 'enzyme';
-
-import algoliaClient from 'algoliasearch';
+import {shallow} from 'enzyme';
```
mthuret: Was there another reason to switch to `shallow` apart the fact than a dom wasn't needed?

vvo: My reasoning is that if this is a unit test then reducing the unit and targeting it precisely is what I want to do. What do you think?

mthuret: I think it's a good idea 👍 
Was curious about your motivation behind :)

PR #1848, opened 2017-01-17
doc(getting-started): Some links are unclickable

mikaa123:
- react-instantsearch docs

**Summary**
In the docs, [links that are followed by an `h2`](https://community.algolia.com/instantsearch.js/react/Getting_started.html#load-the-theme) are not clickable:

clicking-issue

This prevents the user from accessing them.
This is due to the `h2:before` overflowing the underlying text.

**Result**

After removing `h2:before`, the block doesn't overflow the text anymore:
slice

### DIFF
--- a/docgen/src/stylesheets/components/_documentation.scss
+++ b/docgen/src/stylesheets/components/_documentation.scss
@@ -169,7 +169,7 @@
     }
 
 
-    h2, h3, .api-entry, .css-class {
+    h3, .api-entry, .css-class {
       &:before {
         content: "";
         display: block;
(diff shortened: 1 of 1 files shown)


bobylito:
Hi @mikaa123 thanks for the contribution. I will review that asap :)

vvo:
@mikaa123 Tricky, that code is actually here so that when you use the anchors they are not hidden behind the header, if you try to click an anchor on your update it should fail because it will be hidden behind the fixed header bar.

@LukyVj can you give help maybe?

mikaa123:
@vvo thanks for the feedback, you're completely right!

I used `pointer-events` instead. It won't work in IE < 11, if that's an issue.

vvo:
Ok let's merge this and see if it works

LukyVj:
My bad, didn't saw the notification. Let me know if it works

vvo:
Weirdly @mikaa123 this is not sufficient because now every h2 title cannot bo hovered to get the anchor showing (compare with h3 titles). Can we fix that?

PR #1850, opened 2017-01-17
Fix bug with no props being passed down in 2.2.0, closes #1849

### LINKED ISSUE #1849: RefinementList createURL breaks in 2.2.0
Browsing the changelog myself, if I see the problem I'll comment here with an update. 

Edit: Looks like onSearchStateChanged is never being called either ?

Calling createURL on a refinement just returns `#` in 2.2.0, revert back to 2.1.0, works fine.

For it to work :) 

- react-instantsearch

Edit 2, might want some tests that check for injected variables from createInstanteSearch? :)

zackify:
#1849

### DIFF
--- a/packages/react-instantsearch/src/core/createInstantSearch.js
+++ b/packages/react-instantsearch/src/core/createInstantSearch.js
@@ -41,6 +41,7 @@ export default function createInstantSearch(defaultAlgoliaClient, root) {
     render() {
       return (
         <InstantSearch
+          {...this.props}
           indexName={this.props.indexName}
           root={root}
           algoliaClient={this.client}


bobylito:
Thanks for the contribution @zackify :)

packages/react-instantsearch/src/core/createInstantSearch.js
```diff
@@ -41,6 +41,7 @@ export default function createInstantSearch(defaultAlgoliaClient, root) {
     render() {
       return (
         <InstantSearch
+          {...this.props}
```
bobylito: I don't think we want to go back to this spread way of doing things. Best would be to be explicit about all the parameters that will be passed down. It seems that we're missing two things: 
 - `searchParameters`
 - `createURL`

mthuret: We also need:

* searchState
* onSearchStateChange

bobylito: Did that in another PR https://github.com/algolia/instantsearch.js/pull/1867

PR #1879, opened 2017-01-19
fix(SFFV): translations for searchbox were not applied

### LINKED ISSUE #1843: Search For Facet Values: SearchBox translations
Use Case: I want to change the placeholder of the SearchBox used for SFFV in the RefinementList/Menu

```js
<RefinementList attributeName="category" searchForFacetValues
                    translations={{placeholder: 'Search for products'}}/>
```

Behavior: Nothing happen

---

This is happening because the `translate` function we built for the RefinementList/Menu is overridden by the one built for the `<SearchBox/>`.

I see two ways for fixing this:

- We do not override the first translate function built. It means that if a component A use a component B, then every translations should already have been taken care of by the component A. 

- We forward the `translations` props to every widgets (it's not the case today, we use this props to build the `translate` props, but we do not forward the `translations`object). It means that if a component A uses a component B, then he will need to forward the `translations` keys props to the component B, in order for it to build the correct `translate` function. 

In either case, we need to add the `<SearchBox/>` translations propTypes to the `<RefinementList/>/<Menu/>`.

mthuret:
fix #1843 

```js
<RefinementList attributeName="category"
                    searchForFacetValues
                    translations={{placeholder: 'Search inside categories'}}
    />
```

image


packages/react-instantsearch/src/components/Menu.js
```diff
@@ -76,4 +76,9 @@ class Menu extends Component {
 export default translatable({
   showMore: extended => extended ? 'Show less' : 'Show more',
   noResults: 'No Results',
+  submit: null,
+  reset: null,
+  resetTitle: 'Clear the search query.',
```
vvo: Can you add the keys to the documentation?

mthuret: It was added like that " * @themeKey ais-Menu__SearchBox - the container of the search for facet values searchbox. See [the SearchBox documentation](widgets/SearchBox.html#classnames) for the classnames and translation keys of the SearchBox.
" but I can add them completely instead if you think that's better.

vvo: Ok perfect

PR #1885, opened 2017-01-20
Fix Toggle component used with CurrentRefinements

klacointe:
Steps to reproduce:

1.  Use Toggle with CurrentRefinements
2.  Click on Toggle component
3.  Refinement is shown in CurrentRefinements
4.  Remove Refinement from CurrentRefinements
5.  BUG: Toggle was not unchecked.

### DIFF
--- a/packages/react-instantsearch/src/connectors/connectToggle.js
+++ b/packages/react-instantsearch/src/connectors/connectToggle.js
@@ -44,8 +44,8 @@ export default createConnector({
   },
 
   getProvidedProps(props, searchState) {
-    const checked = getCurrentRefinement(props, searchState);
-    return {checked};
+    const currentRefinement = getCurrentRefinement(props, searchState);
+    return {currentRefinement};
   },
 
   refine(props, searchState, nextChecked) {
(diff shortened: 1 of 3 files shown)


klacointe:
Workaround before this get merged:

```
import Toggle from 'react-instantsearch/src/components/Toggle'
import { connectToggle } from 'react-instantsearch/connectors'

const FixedToggle = (props) => {
  return (
    <Toggle                                                                        
      {...props}
      currentRefinement={props.checked}
    />
  ) 
}

const ConnectedFixedToggle = connectToggle(FixedToggle)
```

Then use `ConnectedFixedToggle` instead of `Toggle`.

mthuret:
Hi @klacointe, thank you very much for reporting this! 

I think the appropriate fix would be to rename the prop `checked` to `currentRefinement` inside the `connectToggle` connector. That way we stay consistent with all the other connectors. 

I can make the fix if you want, or you can modify your PR. Tell me what works best for you :)

klacointe:
Hi @mthuret, thanks for you reply.

I've updated the PR.

klacointe:
Thanks @mthuret !

PR #1889, opened 2017-01-20
feat(panel): add a panel widget

mthuret:
* [See the panel in action](https://deploy-preview-1889--algolia-instantsearch.netlify.com/react/storybook/?selectedKind=Panel&selectedStory=RefinementList%20with%20panel&full=0&down=1&left=1&panelRight=1&downPanel=kadirahq%2Fstorybook-addon-knobs)

* [Panel documentation](https://deploy-preview-1889--algolia-instantsearch.netlify.com/react/widgets/panel)

* [e-commerce example with panel](https://deploy-preview-1889--algolia-instantsearch.netlify.com/react/examples/e-commerce/)

### DIFF
--- a/docgen/src/examples/e-commerce-infinite/App.js
+++ b/docgen/src/examples/e-commerce-infinite/App.js
@@ -12,12 +12,12 @@ import {
   StarRating,
   RangeInput,
   Highlight,
+  Panel,
 } from 'react-instantsearch/dom';
 import {
   connectSearchBox,
   connectRefinementList,
   connectInfiniteHits,
-  connectMultiRange,
 } from 'react-instantsearch/connectors';
 import 'react-instantsearch-theme-algolia/style.scss';
 
@@ -57,9 +57,8 @@ const Facets = () =>
         }}
       />
 
-      <SideBarSection
-        title="Show results for"
-        items={[
+      <section className="facet-wrapper">
+        <div className="facet-category-title facet">Show results for</div>
           <HierarchicalMenu
             key="categories"
             attributes={[
@@ -67,67 +66,28 @@ const Facets = () =>
               'sub_category',
               'sub_sub_category',
             ]}
-          />,
-        ]}
-      />
+          />
+      </section>
 
-      <SideBarSection
-        title="Refine by"
-        items={[
-          <RefinementListWithTitle
-            title="Type"
-            key="Type"
-            item={<RefinementList attributeName="type" operator="or" limitMin={5} searchForFacetValues/>}
-          />,
-          <RefinementListWithTitle
-            title="Materials"
-            key="Materials"
-            item={<RefinementList attributeName="materials" operator="or" limitMin={5} searchForFacetValues/>}
-          />,
-          <RefinementListWithTitle
-            title="Color"
-            key="Color"
-            item={<ConnectedColorRefinementList attributeName="colors" operator="or"/>}
-          />,
-          <RefinementListWithTitle
-            title="Rating"
-            key="rating"
-            item={<StarRating attributeName="rating" max={5}/>}
-          />,
-          <RefinementListWithTitle
-            title="Price"
-            key="Price"
-            item={ <CustomPriceRanges
-              attributeName="price"
-              items={[
-                {end: 10},
-                {start: 10, end: 20},
-                {start: 20, end: 50},
-                {start: 50, end: 100},
-                {start: 100, end: 300},
-                {start: 300, end: 500},
-                {start: 500},
-              ]}
-            />}
-          />,
-          <RangeInput key="price_input" attributeName="price" />,
-        ]}
-      />
+      <section className="facet-wrapper">
+        <div className="facet-category-title facet">RefineBy</div>
+          <Panel title="Type">
+            <RefinementList attributeName="type" operator="or" limitMin={5} searchForFacetValues/>
+          </Panel>
+          <Panel
+            title="Materials">
+            <RefinementList attributeName="materials" operator="or" limitMin={5} searchForFacetValues/>
+          </Panel>
+            <ConnectedColorRefinementList attributeName="colors" operator="or"/>
+          <Panel title="Rating">
+          <StarRating attributeName="rating" max={5}/>
+          </Panel>
+          <Panel title="Price">
+            <RangeInput key="price_input" attributeName="price" />
+          </Panel>
+      </section>
       <div className="thank-you">Data courtesy of <a href="http://www.ikea.com/">ikea.com</a></div>
-    </aside>
-  ;
-
-const SideBarSection = ({title, items}) =>
-  <section className="facet-wrapper">
-    <div className="facet-category-title facet">{title}</div>
-    {items.map(i => i)}
-  </section>;
-
-const RefinementListWithTitle = ({title, item}) =>
-  <div>
-    <div className="facet-title">{title}</div>
-    {item}
-  </div>;
+    </aside>;
 
 const CustomSearchBox = ({currentRefinement, refine}) =>
     <div className="input-group">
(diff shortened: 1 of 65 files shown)


vvo:
That is one of the feature that will help a lot our current and next users, thanks! Reviewing now.

.gitignore
```diff
@@ -8,3 +8,4 @@ npm-debug.log
 .idea/
 *.iml
 preview.zip
+.vscode/
```
vvo: Oww someone is switching to VS code :D

docgen/src/examples/e-commerce-infinite/App.js
```diff
@@ -160,7 +120,8 @@ const ColorItem = ({item, createURL, refine}) => {
 };
 
 const CustomColorRefinementList = ({items, refine, createURL}) =>
-    <div>
+    items.length > 0 ? <div>
+      <h5 className={'ais-Panel__Title'}>Colors</h5>
```
vvo: Did you already thought about how to allow creating your own panels? This could answer this use case in a pretty way. Basically we need to be able to have a connectPanel method that would provide an "isEmpty" or "isRelevant" or "canRefine" boolean to indicate that some widget or custom widget inside has ways to refine.

packages/react-instantsearch-theme-algolia/styles/_Panel.scss
```diff
@@ -0,0 +1,10 @@
+.ais-Panel__empty {
+  display: none;
```
vvo: For a first iteration I would prefer to provide a "disabled" styling to the content of the panel (to indicate it's empty) rather than hiding it completely. What do you think?

We could also think we designers to understand if there's a preferred current way to indicate a widget cannot be used anymore.

I am proposing this because on some widgets it may makes less sense to hide everything:
- slider
- range
- multi ranges

packages/react-instantsearch/src/components/CurrentRefinements.js
```diff
@@ -13,8 +13,21 @@ class CurrentRefinements extends Component {
     })).isRequired,
     refine: PropTypes.func.isRequired,
     transformItems: PropTypes.func,
+    isEmpty: PropTypes.func,
```
vvo: Did you think about other namings for this? Empty is only relevant to list of items I believe, maybe "canRefine"? canRefine true means that the widget can be used to refine, false it means it cannot be used to refine.

packages/react-instantsearch-theme-algolia/styles/_Panel.scss
vvo: I just want to ensure that we tested the "do not hide everything" at least once on our examples and can decide if it's better or not than the hide everything strategy.

packages/react-instantsearch/src/components/HierarchicalMenu.test.js
```diff
@@ -125,4 +125,33 @@ describe('HierarchicalMenu', () => {
...
+      />,
+        {context: {
+          isEmpty,
+        }},
+      );
+
+      expect(isEmpty.mock.calls.length).toBe(1);
+      expect(isEmpty.mock.calls[0][0]).toEqual(false);
+
+      wrapper.unmount();
```
vvo: I feel it would be easier to understand if we updated the props that will actually means "can no more be used to refine" rather than triggering mount/unmount. What do you think?

packages/react-instantsearch/src/components/List.js
```diff
@@ -115,17 +116,17 @@ class List extends Component {
 
   render() {
     const {cx, items, searchForFacetValues} = this.props;
+    const searchBox = searchForFacetValues ? this.renderSearchBox() : null;
     if (items.length === 0) {
       return <div {...cx('root')}>
-        {this.renderSearchBox()}
+        {searchBox}
```
vvo: 👍

packages/react-instantsearch/src/core/createInstantSearchManager.result.test.js
```diff
@@ -91,7 +91,7 @@ describe('createInstantSearchManager', () => {
       });
     });
     describe('on search for facet values', () => {
-      it.only('updates the store and searches', () => {
```
vvo: Small info: I am waiting for jest to release eslint-plugin-jest to check those mistakes automatically

packages/react-instantsearch/src/widgets/Panel.js
```diff
@@ -0,0 +1,37 @@
+import Panel from '../components/Panel.js';
+
+/**
+ * The Panel component will add a title to a widget. It will also handle when a wrapped widget doesn't have any refinement:
```
vvo: Proposal:

```txt
The Panel widget wraps other widgets with a consistent panel design. It also reacts, indicates and set CSS classes when widgets are no more relevant for refining (like when a RefinementList becomes empty because of the current search).
```

I won't list the widgets, ultimately Panel can be used with any widget right?

docgen/src/examples/e-commerce-infinite/App.js
mthuret: I'm not sure that having a `connectPanel` would be easier than just checking for the `items` length (or any condition that works for you).

Also I'm not sure to see an easy way to hide the fact that in order for the panel to work, you need to notify it when your condition is met.

packages/react-instantsearch-theme-algolia/styles/_Panel.scss
mthuret: Fine with me. We can not enforce anything of the panel stylesheet, but decline it depending on the widget. 

Also, I didn't add the Panel feature to the MultiRanges regarding the fact that the list is provided manually by the user. Do you have something in mind for this widget?

packages/react-instantsearch/src/components/CurrentRefinements.js
mthuret: 👍

packages/react-instantsearch/src/widgets/Panel.js
mthuret: As for now, I implemented it for widgets I thought it makes sense that's why I list them here. (but we could also have written a guide). 

I don't now if allowing something like below is pertinent: 

```js
<Panel>
    <HitsPerPage items=[]/>
</Panel>
```

vvo: No but users might use it like that:

```jsx
<Panel>
  <YourCart />
</Panel>
```

```jsx
<Panel>
  <Partners>
     <UPS/>
  </Partners>
</Panel>
```

If there's no constraint on Panel usage (= can be used on any widget) then we do not need to list the widgets that have a specific behavior when used.

Otherwise as a user you might have to parse through the whole list to know if you can use Panel (while you can definitely use it anywhere). And we will forget about maintaining the list.

docgen/src/examples/e-commerce-infinite/App.js
vvo: connectors could forward the isEmpty function down the line when used in standalone?

That's maybe too much and in the end people using connectors may be always be using connectors, not mixing widgets like we do in examples.

mthuret:
I've just updated the PR with last changes: 

* never unmount widgets but instead compute a boolean `canRefine` that is provided to widgets. Those widgets will display a new classname if no refinement present. Some, if a panel wrap them, will notify it. 

I removed the display:none from our default style, but still don't really know how to move forward. It still makes sense to hide panel for some widgets (ex: the RefinementList), and for others maybe not: (ex: RangeInput). Maybe we can do nothing about it in our default theme as for now, and discuss about the right implementation with the design team? End users will still be able to do what they want with our widgets in their side. What do you think?

vvo:
Can we add stories where we have widgets inside panels and canRefine is true (by triggering canRefine with a defaultRefinement maybe?).

So that we can easily work on it.

mthuret:
I think we are good now for this PR. 

As for next iterations we would need:

* Choose and implement the design we think the best when a widgets has no refinement
* Create a `connectPanel` if needed

Kikobeats:
I like that!

PR #1900, opened 2017-01-24
Fix header links, again

LukyVj:
To fix the issue I had to play a lot with margin values, and to increase a bit the distance between titles and paragraphs. 

| Before  | After |
| ------------- | ------------- |
|   |  |

But I don't think it looks that bad after all. 

So now, all the "anchors / links" issues should be fixed :) 
Next time we should consider using Javascript to make anchors, to get rid of this issue.

### DIFF
--- a/docgen/src/stylesheets/components/_fonts.scss
+++ b/docgen/src/stylesheets/components/_fonts.scss
@@ -44,6 +44,10 @@ p {
   }
 }
 
+li {
+  color: $blueish-grey;
+}
+
 /* Big titles
 ======================*/
 h2.title {
--- a/docgen/src/stylesheets/components/_documentation.scss
+++ b/docgen/src/stylesheets/components/_documentation.scss
@@ -45,17 +45,24 @@
     position: relative;
     z-index: 99;
 
+    &:not(.sidebar-header) {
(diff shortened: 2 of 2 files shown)


bobylito:
Yes that looks good on Netlify, https://deploy-preview-1900--algolia-instantsearch.netlify.com/instantsearch.js/react/guide/connectors (love it @vvo)

docgen/src/stylesheets/components/_fonts.scss
```diff
@@ -44,6 +44,10 @@ p {
   }
 }
 
+li {
+  color: $blueish-grey;
```
bobylito: Is this change part of the fix? :)

bobylito:
The anchors still work properly, https://deploy-preview-1900--algolia-instantsearch.netlify.com/instantsearch.js/react/guide/connectors#when-do-i-need-to-use-connectors

GG @LukyVj

docgen/src/stylesheets/components/_fonts.scss
LukyVj: Nope, but as you can see on the "before" picture, the list items were ugly 🙈

PR #1903, opened 2017-01-24
misc(issue template): starting point for jsFiddle

bobylito:
Provides an example to start when providing a live example of a bug.


vvo:
All in for the jsFiddle starting point to ease creating examples. But that one is an advanced one: it's multi index search. Maybe do an even simpler one?

bobylito:
True :)

PR #1912, opened 2017-01-26
fix(Hits): limit the hitComponent to be only a function

mikaa123:

**Summary**
This merge request limits the `hitComponent` prop of `<Hits>` to be only a function. It keeps the API simple, and will now print a warning if the user pass a string.

**Result**
Passing a string to `hitComponent` now makes React complain about it.


bobylito:
@mikaa123 There are some failing tests, can you have a look?

PR #1913, opened 2017-01-26
docs(code samples): change syntax theme, add copy button

vvo:
This commit changes the syntax theme to use GitHub colors and adds a
little header much like on reactjs docs.

It also adds a copy button.

Before:
image

After:
image

Also this:
image

Copy button:
copy-paste


bobylito:
Is it possible to style the copy button? :)

vvo:
Yes, I wondered if maybe a simpler text only button could do it and would be more obvious:

```
[ COPY ]
```

```
[ COPIED ! ]
```

?

vvo:
RIght now this is just an SVG I took from https://clipboardjs.com/ (and it's the same as on GitHub clone copy icon)

vvo:
Ok we are good

bobylito:
\o/

PR #1921, opened 2017-01-26
fix(*List): disable shortcuts in *List SearchBoxes

### LINKED ISSUE #1920: SearchForFacet values searchbox catches shortcut by default, should be disabled

vvo:
Before this commit, when using searchForFacetValues,
the default SearchBox shortcuts would be listening for shortcuts.

Thus when hitting `/` or `s` it would try to focus
the SearchBox of RefinementList instead of something else.

fixes #1920

Note: we are not testing List.js right now, we need to fix that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,9 @@
   "engines": {
     "node": "^6.x.x || ^7.x.x"
   },
+  "jest": {
+    "notify": true
+  },
   "license": "MIT",
   "author": {
     "name": "Algolia, Inc.",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -103,6 +103,9 @@
   "engines": {
     "node": "^6.x.x || ^7.x.x"
   },
+  "jest": {
+    "notify": true
```
vvo: Somehow this ended up here, still very good feature, see:

image

Issue #1871, opened 2017-01-18
Rename searchForFacetValues option to withSearchBox.


vvo:
We tried to avoid speaking about facets at first to users so that they are note confused when they do not know what it means. So we should rename searchForFacetValues to:
- withSearchBox (boolean in widgets) instead of searchForFacetValues
- searchForItems (function in connectors) instead of searchForFacetValues

PR #1922, opened 2017-01-27
refacto(SFFV): use withSearchBox and searchForItems

mthuret:
see #1871


packages/react-instantsearch/src/connectors/connectMenu.js
```diff
@@ -107,15 +110,14 @@ export default createConnector({
           isRefined: v.isRefined,
         }));
 
-    const sortedItems = !isFromSearch && props.searchForFacetValues ?
+    const sortedItems = !isFromSearch && withSearchBox ?
```
bobylito: If the `withSearchBox` is `false` then `isFromSearch` should always be `true`, right? In that case, we could put `withSearchBox` first, no?

mthuret: Yep 👍

packages/react-instantsearch/src/connectors/connectMenu.js
```diff
@@ -56,6 +58,8 @@ export default createConnector({
     limitMax: PropTypes.number,
     defaultRefinement: PropTypes.string,
     transformItems: PropTypes.func,
+    withSearchBox: PropTypes.bool,
+    searchForFacetValues: PropTypes.bool, //deprecated
```
vvo: You can use @deprecated so we can easily search for those tokens for a new major version

packages/react-instantsearch/src/connectors/connectMenu.js
```diff
@@ -68,22 +72,21 @@ export default createConnector({
...
-
     const canRefine =
       Boolean(results) &&
       Boolean(results.getFacetByName(attributeName));
 
     const isFromSearch = Boolean(searchForFacetValuesResults
       && searchForFacetValuesResults[attributeName]
       && searchForFacetValuesResults.query !== '');
-    const searchForFacetValues = props.searchForFacetValues ? this.searchForFacetValues : undefined;
+    const withSearchBox = props.withSearchBox || props.searchForFacetValues;
```
vvo: Can we add a if (process.env.NODE_ENV === 'development') console.warn:

```text
react-instantsearch: `searchForFacetValues` has been renamed to `withSearchBox`, this will break in the next major version.
```

packages/react-instantsearch/src/connectors/connectMenu.js
```diff
@@ -68,22 +72,21 @@ export default createConnector({
...
 
     if (!canRefine) {
       return {
         items: [],
         currentRefinement: getCurrentRefinement(props, searchState),
         isFromSearch,
-        searchForFacetValues,
+        withSearchBox,
```
vvo: This will break for people using props.searchForFacetValues in their connector usage no? We need to pass both properties sadly.

vvo:
> The update is much more complex than expected.

It is! I failed two times at giving this a try ahah

packages/react-instantsearch/src/connectors/connectMenu.js
mthuret: Before the searchForFacetValues that was passed as a prop was the function to call. So I'm not sure to follow here.

mthuret:
I added the warning. If we still want to pursue with backward compatibility, it should be ok :)

packages/react-instantsearch/src/connectors/connectRefinementList.test.js
```diff
@@ -285,7 +294,7 @@ describe('connectRefinementList', () => {
     expect(searchState).toEqual({another: {searchState: 'searchState'}});
   });
 
-  it('calling searchForFacetValues return the right searchForFacetValues parameters', () => {
+  it('calling searchForItems return the right searchForItems parameters', () => {
     const parameters = searchForFacetValues({attributeName: 'ok'}, {}, 'yep');
```
vvo: Do we need a test that actually checks searchForItems usage? Here we test searchForFacet values still?

vvo:
Made some small changes:
- renamed documentation guide file
- put process.env.NODE_ENV test in ifs first instead of last (because in the "optimistic case", production, this will always be false)

packages/react-instantsearch/src/connectors/connectRefinementList.test.js
vvo: Nevermind, we agreed on keep the name `searchForFacetValues` internally in connectors

vvo:
I think we are good to go here, last call! :D

PR #1926, opened 2017-01-30
feat(MultiRange): indicate if a range has no refinements

mthuret:
For the MultiRange, compute a boolean for each item range in order to add classes like:

image

### DIFF
--- a/packages/react-instantsearch/src/components/List.js
+++ b/packages/react-instantsearch/src/components/List.js
@@ -64,6 +64,7 @@ class List extends Component {
         {...this.props.cx(
           'item',
           item.isRefined && 'itemSelected',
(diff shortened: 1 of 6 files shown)


Kikobeats:
fast question, could be possible select more than option?

mthuret:
@Kikobeats I'm not sure to fully understand your question, can give us more details?

Kikobeats:
@mthuret sure, something like:

lol

pick more than one

bobylito:
As of now it's not possible because of a limitation of the underlying JS Helper. I have an [issue opened there](https://github.com/algolia/algoliasearch-helper-js/issues/461) that would make this possible.

Kikobeats:
hmm interesting! then I will waiting for that, thanks for reply 😄

PR #1928, opened 2017-01-31
fix(examples)

mthuret:


docgen/src/examples/tourism/App.js
```diff
@@ -116,7 +117,7 @@ function HitsMap({hits}) {
     nw: {lat: -85, lng: 180},
     se: {lat: 85, lng: -180},
   });
-  const boundsConfig = fitBounds(boundingPoints, availableSpace);
+  const boundsConfig = hits.length > 0 ? fitBounds(boundingPoints, availableSpace) : {};
```
mthuret: This is due to the <Panel/> feature and the change of behavior when now we are not unmounting widgets anymore. We should warn about the fact that now, people might need to handle empty array of items/hits.

docgen/src/examples/tourism/App.js
```diff
@@ -30,6 +30,7 @@ const App = props =>
     onSearchStateChange={props.onSearchStateChange.bind(this)}
   >
     <Configure aroundLatLngViaIP={true}/>
+    <Configure aroundRadius="all"/>
```
vvo: We can reuse the same Configure (otherwise people might read "1 
 Configure = 1 param)

docgen/src/examples/tourism/App.js
vvo: 👍  will add this in release notes and announcement

docgen/src/examples/tourism/App.js
```diff
@@ -17,8 +17,8 @@ import React, {PropTypes} from 'react';
 import GoogleMap from 'google-map-react';
 import {fitBounds} from 'google-map-react/utils';
 
-import Rheostat from 'rheostat';
 import {withUrlSync} from '../urlSync';
+import Range from '../../../../stories/3rdPartiesIntegration.stories';
```
vvo: Would rather have an examples/Slider.js (if we want to "hide" long implementation) that we reuse here and avoid mixing stories with examples. Otherwise we will forget that link between two parts of the code when updating one or the other

PR #1929, opened 2017-01-31
feat(widgets): default design for disabled states

vvo:
When widgets are no more relevant/have no refinements, we were already
adding specific CSS classes. Now the default theme also define more
precise rendering states for widgets.

Basically noRefine panels will get an opacity of 0.7.

Also did some fixes so that MultiRange is working with Panel feature.

### DIFF
--- a/packages/react-instantsearch-theme-algolia/styles/_MultiRange.scss
+++ b/packages/react-instantsearch-theme-algolia/styles/_MultiRange.scss
@@ -55,3 +55,7 @@
 
 .ais-MultiRange__itemLabelSelected{
 }
+
+.ais-MultiRange__itemNoRefinement {
+  opacity: .7;
+}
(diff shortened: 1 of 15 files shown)


bobylito:
I tried on the ecom' demo and when you have no results at all then everything has the same opacity which doesn't make stand out as disabled. I don't know if we should maybe have an opacity that is even less or provide other kinds of hints. 🤔

vvo:
Yes indeed, basically on the link you provided every filter is disabled. Do you have any other idea to make it obvious? Previously they were hidden

bobylito:
What about there was an explicit message?

Also I'm wondering when is there a case where not all the widgets gets grayed? In our datasets, all records have values for the facetted attributes, which means that if for one facetted attribute we have no values -> there are no records in the result set -> all the widgets will be grayed.

vvo:
> What about there was an explicit message?

Yes that could be nice also.

Let's go first with this iteration and then see if we need/want/benefit from more detailed non-refinable states.

> Also I'm wondering when there is a case where not all the widgets gets grayed?

Wondering the same, no 100% sure answer

PR #1935, opened 2017-02-02
feat(Configure): add configure parameters in search state

mthuret:
This PR add to the search state all configure parameters. That way, if you perform some changes to them when listening to `onSearchStateChange` it will be taken into account. 

I used the `transitionState` function of the connector to be able to add those parameters to the state without the need of calling refine (that triggers a new search). It might not be the most obvious name though.


vvo:
Does transition state means "update the internal state without triggering a new search", if so I bet we can find a better name. Any ideas?

mthuret:
Well it's more "modify the search state depending on the previous and next one". 
This function will never be call if no refine are actually perform. So it's not a **standalone** way to update the state without performing the search.

vvo:
I am fine with the usage of transition state for now, it's there let's use it.

Issue #1938, opened 2017-02-02
Pagination widget showNext and showLast disable problem


EZarkov:
**Do you want to request a *feature* or report a *bug*?**
bug
**Bug: What is the current behavior?**
In Pagination widget `showNext ` and  `showLast ` are not disabled when number of hits is 0  
**Bug: What is the expected behavior?**
To be disabled.
**Bug: What browsers are impacted? Which versions?**

**Feature: What is your use case for such a feature?**

**Feature: What is your proposed API entry? The new option to add? What is the behavior?**

Calculation for total pages:
https://github.com/algolia/instantsearch.js/blob/v2.2.5/packages/react-instantsearch/src/components/Pagination.js#L110
`totalPages = 0` because number of hits is 0 ->
https://github.com/algolia/instantsearch.js/blob/v2.2.5/packages/react-instantsearch/src/connectors/connectPagination.js#L48  

https://github.com/algolia/instantsearch.js/blob/v2.2.5/packages/react-instantsearch/src/components/Pagination.js#L110https://github.com/algolia/instantsearch.js/blob/v2.2.5/packages/react-instantsearch/src/components/Pagination.js#L111
lastPage = 0
https://github.com/algolia/instantsearch.js/blob/v2.2.5/packages/react-instantsearch/src/connectors/connectPagination.js#L11
in searchState there is no key = page -> 
https://github.com/algolia/instantsearch.js/blob/v2.2.5/packages/react-instantsearch/src/connectors/connectPagination.js#L49 
`currentRefinement = 1`

https://github.com/algolia/instantsearch.js/blob/v2.2.5/packages/react-instantsearch/src/components/Pagination.js#L153
https://github.com/algolia/instantsearch.js/blob/v2.2.5/packages/react-instantsearch/src/components/Pagination.js#L163
disabled = false
Possible solution:
`disabled: currentRefinement === lastPage || lastPage <= 1`
**What project are you opening an issue for?**
- react-instantsearch

**What is the version you are using? Always use the latest one before opening a bug issue.**
2.2.5

mthuret:
Hi @EZarkov, that's a good catch. 

Classnames are updated correctly to reflect the disabled state but it misses the disabled attribute.

chriscalip:
@vvo @mthuret 
Sorry for ignorance but why is this feature fix not include in HEAD v1?

vvo:
@chriscalip Can you open a new issue about it?

PR #1939, opened 2017-02-02
fix(Pagination): add disabled property

mthuret:
see #1938

### DIFF
--- a/packages/react-instantsearch/src/components/LinkList.js
+++ b/packages/react-instantsearch/src/components/LinkList.js
@@ -53,6 +53,7 @@ export default class LinkList extends Component {
               item.disabled && 'itemDisabled',
               item.modifier
             )}
+            disabled={item.disabled}
           >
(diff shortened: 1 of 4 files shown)


packages/react-instantsearch/src/components/LinkList.js
```diff
@@ -53,6 +53,7 @@ export default class LinkList extends Component {
               item.disabled && 'itemDisabled',
               item.modifier
             )}
+            disabled={item.disabled}
```
vvo: What is the effect of putting disabled on the `<li>`?

mthuret: That way our disabled style mixin is really used. 

pagination-disabled

vvo: Just discovered this, will use it in the previous PR about disabled widget states :)

PR #1948, opened 2017-02-06
feat(custom client): allows to provide a custom JS client instance

olance:

**Summary**
This PR allows one to provide his own Algolia JS client instance to instantsearch.js, instead of letting it create an internal one.  
This can be useful in different scenarios... Some that come to mind:
- when hosts/other options need to be overriden
- when Algolia is needed in another context than instantsearch only

**Result**
You just have to pass a new option to the initialization function:

```js
let search = instantsearch({
  appId: 'appId',
  apiKey: 'apiKey',
  // ...
  clientInstanceFunction: (appID, apiKey) => algoliasearch(appID, apiKey, { timeout: 4000  }),
});
```

Open to feedback of course!
Things we could consider:
- another name for the option?
- provide the same mechanism for the internal JS helper instance?

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -62,7 +62,7 @@ class InstantSearch extends EventEmitter {
     searchParameters = {},
     urlSync = null,
     searchFunction,
-    clientFactory = (app, key) => algoliasearch(app, key),
+    clientFactory = algoliasearch,
   }) {
     super();
     if (appId === null || apiKey === null || indexName === null) {
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -24,8 +24,10 @@ function defaultCreateURL() { return '#'; }
  * @param  {function} [options.searchFunction] A hook that will be called each time a search needs to be done, with the
  * helper as a parameter. It's your responsibility to call helper.search(). This option allows you to avoid doing
  * searches at page load for example.
- * @param   {function} [options.clientFactory] A function called upon initialization with the `options.appId`
- * and `options.apiKey` values, allowing you to return your own instance of the Algolia JS client to be used by instantsearch.js
+ * @param   {function} [options.clientFactory] A function called upon initialization with the original `algoliasearch`
+ * client factory function, the `options.appId` and `options.apiKey` values, allowing you to return your own instance of
+ * the Algolia JS client to be used by instantsearch.js.
+ * Defaults to: `clientFactory = (algoliasearch, app, key) => algoliasearch(app, key)`
  * @param  {Object} [options.searchParameters] Additional parameters to pass to
  * the Algolia API.
  * [Full documentation](https://community.algolia.com/algoliasearch-helper-js/reference.html#searchparameters)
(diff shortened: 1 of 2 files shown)


src/lib/InstantSearch.js
```diff
@@ -24,6 +24,8 @@ function defaultCreateURL() { return '#'; }
  * @param  {function} [options.searchFunction] A hook that will be called each time a search needs to be done, with the
  * helper as a parameter. It's your responsibility to call helper.search(). This option allows you to avoid doing
  * searches at page load for example.
+ * @param   {function} [options.clientInstanceFunction] A function called upon initialization with the `options.appId`
```
bobylito: What about something like `clientFactory`?

src/lib/InstantSearch.js
```diff
@@ -60,6 +62,7 @@ class InstantSearch extends EventEmitter {
     searchParameters = {},
     urlSync = null,
     searchFunction,
+    clientInstanceFunction = (app, key) => algoliasearch(app, key),
```
bobylito: the default could be `algoliasearch` directly:
```js
clientInstanceFunction = algoliasearch,
```

olance: indeed 👍

src/lib/InstantSearch.js
olance: me likes :)

vvo:
What we did in React InstantSearch is letting your provide a really instantiated Algolia client (algoliaClient=client), should we do something similar instead of asking to provide a function that will be called with data you already passed?

src/lib/InstantSearch.js
```diff
@@ -24,6 +24,8 @@ function defaultCreateURL() { return '#'; }
  * @param  {function} [options.searchFunction] A hook that will be called each time a search needs to be done, with the
  * helper as a parameter. It's your responsibility to call helper.search(). This option allows you to avoid doing
  * searches at page load for example.
+ * @param   {function} [options.clientFactory] A function called upon initialization with the `options.appId`
```
vvo: I propose `createAlgoliaClient`, what do you think?

bobylito: It might create less friction than the factory indeed.

olance:
@vvo this solution (with last commit) has the advantage to remove the need to include algoliasearch-client-js twice: one bundled in instantsearch.js and one to instantiate the client on your own

src/lib/InstantSearch.js
olance: Because it's a verb-based name ("createXX") option, if I read the name of this option without its doc, I'd assume it's a boolean to say whether or not the client should be created.

I prefer a noun, which informs on the nature of the option.
Would you prefer `algoliaClientFactory`?

vvo:
This is a good addition, only part I have issue with is the naming, I'd prefer any name that does not have the "factory" name in it. I'd prefer something that does not requires the user to know what is a factory or how does a factory applies here.

The best one I could come up with is `createAlgoliaClient`, open to other suggestions.

vvo:
ping @olance

olance:
Sorry, intense day yesterday 😅 
I can't think of something better so let's go with `createAlgoliaClient`!
I'll try to push that today

src/lib/InstantSearch.js
```diff
@@ -60,6 +66,7 @@ class InstantSearch extends EventEmitter {
     searchParameters = {},
     urlSync = null,
     searchFunction,
+    createAlgoliaClient = (customOrInternalAlgoliasearch, appId, apiKey) => customOrInternalAlgoliasearch(appId, apiKey),
```
bobylito: `customOrInternalAlgoliasearch` this is always our algoliasearch client.

PR #1951, opened 2017-02-06
fix(createConnector): updates with latest props on state change

vinhlh:

**Summary**

Use cases: 
- Use createConnector to create a new custom connector.
- Use this new connector to create a component.
- Update the `props` of this component.
- Trigger a `state` change.
- This component still calls `getProvidedProps` with initial props.

**Result**

- Should calls `getProvidedProps` with latest props.

cc @vvo

### DIFF
--- a/packages/react-instantsearch/src/core/createConnector.js
+++ b/packages/react-instantsearch/src/core/createConnector.js
@@ -64,7 +64,7 @@ export default function createConnector(connectorDesc) {
 
       this.unsubscribe = store.subscribe(() => {
         this.setState({
-          props: this.getProvidedProps(props),
+          props: this.getProvidedProps(this.props),
         });
       });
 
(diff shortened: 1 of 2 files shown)


vvo:
Very nice @vinhlh, I wonder if we already had real work example/cases where this would trigger a bad behaviour.

What was the root event that led you to try and fix this issue? Anything on the UI or the existing widgets usages?

Thanks a lot.

vinhlh:
@vvo my use case is:
When working with `react-router`, it's a change from `/detail/11` to `/detail/12` (`/detail/:newId:`),
I also have a custom `createConnector` to calculate some derivative values based on `searchResults` and props (in this case, it's `newId`).
But `getProvidedProps` still receives the old value `11` after the route is changed, then it's this bug.

PR #1959, opened 2017-02-08
feat(MultiRange): add an all range

mthuret:


mthuret:
Needs to be refactored:
- Use value '' to compute the label instead of the boolean. 
- Handle the case where a user would provided an item with value ''.

mthuret:
I indicate in the commit message that this is a breaking change as it will automatically add an 'All' range. 
You can remove this range through the new classname ais-MultiRange__itemAll.

PR #1961, opened 2017-02-09
feat(InfiniteHits): add new widget

julienpa:
_Note : if the PR gets approved, commits will be squashed into 1 before merging_

@bobylito feel free to assign other reviewers on this one!

## Summary
After hearing many requests for a showMore/infiniteHits widget and given the fact that this feature exists in `React-Instantsearch`, @bobylito helped me bring it to `Instantsearch.js`! 🎉  🍾  🔎 

The **InfiniteHits widget** should be used as a replacement for the regular Hits widget, _not_ along with it.
It will replicate the Hits widget API and rendering and add a button at the bottom that is customizable with a label and CSS classes.
This widget shouldn't be used with the pagination widget since the 2 features overlap.

## Result
You can try the widget here: [demo](http://infinitehits.surge.sh)

GIF preview:


dev/instantsearch.css
```diff
@@ -1 +0,0 @@
-../dist/instantsearch.css
```
vvo: It seems you removed the symlink here to dist/ (maybe by mistake/command). The goal is to load the builded CSS file in the dev env. Can you put it back?

src/widgets/infinite-hits/infinite-hits.js
```diff
@@ -0,0 +1,98 @@
...
+import {
+  bemHelper,
+  prepareTemplateProps,
+  getContainerNode,
+} from '../../lib/utils.js';
+import cx from 'classnames';
+import InfiniteHits from '../../components/InfiniteHits.js';
+import defaultTemplates from './defaultTemplates.js';
+
+const bem = bemHelper('ais-infinitehits');
```
vvo: ais-infinite-hits maybe? This seems to be the way we name CSS classes today.

dev/instantsearch.css
julienpa: I'm thinking about the `staging` command:
```bash
mv ./dev/instantsearch.css ./dev/instantsearch.css.symlink
cp ./dist/instantsearch.css ./dev/instantsearch.css

surge dev

rm ./dev/instantsearch.css
mv ./dev/instantsearch.css.symlink ./dev/instantsearch.css
rm ./dev/bundle.js
rm ./dev/bundle.js.map
```

Is there an easy way to revert this?

vvo: git checkout dev/instantsearch.css

julienpa: All right! Fixed with:
```bash
$ rm dev/instantsearch.css 
$ ln -s dist/instantsearch.css dev/instantsearch.css 
```

julienpa:
@vvo @bobylito thanks so much for the review! 👍 
Updated the PR with requested changes. If it's ok for you, I'll squash it all ;)

bobylito:
Squashing since the changes have been made as requested by @vvo who is on a 🛫

Issue #1860, opened 2017-01-18
Add geo search guide using Configure


vvo:
Guide should explain how to change geo location of a search instance using Configure.

Ultimately adding recipes on how to plug google autocomplete and google maps to react-instantsearch would be awesome.

Recipes:
  - location autocomplete refines search
  - map move refines search

mthuret:
React InstantSearch has been moved here: https://github.com/algolia/react-instantsearch
This issue has either been rewritten or placed inside our backlog.

PR #1967, opened 2017-02-10
fix(Configure): use props a unique source of truth

mthuret:
With this PR I removed the fact that we can read `<Configure>` parameters from the search state, as it's not easy to handle if we also provide a way to change them directly from the props. 

The props passed through `Configure` will only be pushed to search state. If the parameters needs to be modified, then the props of `Configure` needs to be modified. It means that when reading from a url, then the configure parameters must be extracted and passed through the props. 

If three params where passed to the `Configure` at the first rendering, and then as the second one only two props are passed, then the third one will disappear from the `search state`.


vvo:
> It means that when reading from a url, then the configure parameters must be extracted and passed through the props.

Do we have an example on how to do that?

@zackify what do you think? We believe it's tricky to have a Configure widget that read both ways (from props, from state).

When it reads from both props and state, which one should take precedence? @mthuret Can you maybe precise exactly what were the struggles trying to support it both ways?

zackify:
I see the problem. Take from props if the props are null in
componentWillReceiveProps. On initial mount take from the state if it
doesn't exist then try getting from the props? That might fix it

zackify:
Sorry that wasn't explained well. On mount, take from existing state. Any
time componentWillReceiveProps is called, that takes precedence.

mthuret:
One use case illustrating the issue: 

Initial: Configure props is `hitsPerPage=3`

Step1: pass new props to configure to be `hitsPerPage=1`

Now the `search state` has `hitsPerPage=1`. 

Step2: pass new props to configure to be `hitsPerPage=2`

If we look at the search state first, this will never be applied. 

-- 

One obvious use case to read from the `search state` would be when you are loading the page and read the state from the URL. So reading from the state when the component did mount should do the trick for this case, but for now we don't have the architecture for it. 

Before the PR, the way it could have worked was when unmounting systematically the `<Configure/>` widget before passing it new props. 

With the PR, the search state is never taken into account, and you need to pass the right props directly to the `<Configure/>` widget.

vvo:
@mthuret Can you ensure we do have a full example doing an url synchronisation with Configure where we re-read state and apply props dynamically? I am not sure we do have that today while it's critical to users.

mthuret:
@vvo I can modify our examples to include that, even if the props will not changed dynamically. 
I can also make a special story that can change props dynamically but will not highlight the URL part of it. 

Otherwise, we need to build the geo recipes with this PR. What do you think its the best?

mthuret:
So I added a geo-search recipe to show how you can update search parameters and synchronize the URL. I also updated the search parameters guide. 

This geo search example is very basic and only feature:

* a regular search through the searchbox. 
* a map that when we click somewhere perform a new search. 

We still need to do #1860 because with this example:

* the search is not triggered when moving the map. (mostly because I didn't find a `dragEnd` event with the lib used at the moment). 
* No autocomplete.

packages/react-instantsearch/examples/geo-search/src/App.js
```diff
@@ -0,0 +1,143 @@
...
+      );
+    }, updateAfter);
+
+    // when a new query is performed, removed the aroundLatLng property.
+    const aroundLatLng = this.state.searchState.query !== searchState.query ? null : this.state.aroundLatLng;
+    this.setState({searchState, aroundLatLng});
+  };
+
+  render() {
+    const configuration = this.state.aroundLatLng
```
vvo: Nice, that's what I would be looking for as an example. Thx

PR #1980, opened 2017-02-14
docs(storybook): use 'translations' in SearchBox

Haroenv:

**Summary**

It was kind of confusing to realise that the actual prop to add was an object with "translations" as name, since using "translate" would cause "object is not a function" errors.

I haven't yet checked if this is the case with all of the stories, or just this one.

**Result**

the story now shows "translations" as key

### DIFF
--- a/stories/RangeInput.stories.js
+++ b/stories/RangeInput.stories.js
@@ -18,7 +18,7 @@ stories.add('default', () =>
     <RangeInput attributeName="price"
                   min={number('max', 0)}
                   max={number('max', 300)}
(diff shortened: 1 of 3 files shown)


vvo:
> I haven't yet checked if this is the case with all of the stories, or just this one.

Can you check quickly so we do not have to get it back to this afterwards? Thanks!

Haroenv:
There were a few other components where "translation" was visible in the UI as "translate"; I've edited those in this commit.

vvo:
Thanks a lot :)

PR #1981, opened 2017-02-14
fix(searchBox): make svgs inherit color

Haroenv:

**Summary**

right now you'd have to find a way to give the svg a fill (`.ais-SearchBox__reset svg, .ais-SearchBox__button svg { fill: red }`), but in most cases I imagine that in most cases, it should just be the same as whatever the current color is.

**Result**

You can style the searchbox with `.ais-SearchBox__reset, .ais-SearchBox__button { color: red }`, or `.ais-SearchBox { color: red }`, as well as other combinations. 

However this means that explicitly setting the fill might need a higher specificity (we will see when the deploy preview is done if this breaks our default theme, if it does some extra thought needs to be put in this PR)

cc @Shipow for reasoning if this makes sense


vvo:
Do we need anything else here or can this be merged?

Haroenv:
Im on mobile now, so I can't check, but if the deploy preview of searchbox has the correct grey icons, then nothing is broken by this. I'll check in the morning too

Haroenv:
Unfortunately this breaks the Algolia style, so I think the best practice is in your own css, you will keep to use `fill`. Maybe this should be in some guide? 

Closing for now.

PR #1982, opened 2017-02-15
docs(guide): how to style svgs

Haroenv:

**Summary**

it's not so hard, but the actual svg elements don't have a class, so you will have to target them explicitly.

SVG elements also have a different feature set, so this example should clear things up.

**Result**

Extra piece of text in the `Styling widgets` part that explains briefly that we work with svg, so it's styleable and how.

see #1981


vvo:
Thx, documentation is key!

PR #1995, opened 2017-02-21
chore(jest): update and add linter

mthuret:
Add eslint-jest-linter to have an error if some tests are skipped. 

Other new features are described here: https://facebook.github.io/jest/blog/2017/02/21/jest-19-immersive-watch-mode-test-platform-improvements.html

I put error, even for "skipped" tests. They recommend warn. 
What's your thoughts about that? 

One use case I see for the "warn" is: if you have a test that is failing for an unknown (or known but hard to fix) reason, and you still want to merge.


vvo:
Ah nice did not know they released officially the jest plugin!

PR #1999, opened 2017-02-21
Feat/mobile improvements

Haroenv:

**Summary**

There are some things that can be better in a searchbox by default. Since I enabled those in yarn search, I should move them upstream too. 

cc @Shipow 

**Result**

1. keyboard dismisses when you search
2. don’t correct spelling

    this takes up space and is almost never useful in a searchbox

    Google's searchbar and the default Safari search disable autoCorrect, spellCheck and autoComplete too

3. make “return” show as “search” on iOS Safari

    requires an action to the form. Since this is created and controlled by JS, it’s basically a no-op.

before | after
---|---
before | new

### DIFF
--- a/packages/react-instantsearch/src/components/SearchBox.js
+++ b/packages/react-instantsearch/src/components/SearchBox.js
@@ -120,6 +120,7 @@ class SearchBox extends Component {
   onSubmit = e => {
     e.preventDefault();
     e.stopPropagation();
+    this.input.blur();
 
     const {refine, searchAsYouType} = this.props;
     if (!searchAsYouType) {
--- a/packages/react-instantsearch/src/components/SearchBox.js
+++ b/packages/react-instantsearch/src/components/SearchBox.js
@@ -172,6 +172,9 @@ class SearchBox extends Component {
             placeholder={translate('placeholder')}
             autoFocus={autoFocus}
             autoComplete="off"
+            autoCorrect="off"
+            autoCapitalize="off"
+            spellCheck="false"
             required
             value={query}
             onChange={this.onChange}
--- a/packages/react-instantsearch/src/components/SearchBox.js
+++ b/packages/react-instantsearch/src/components/SearchBox.js
@@ -152,6 +152,7 @@ class SearchBox extends Component {
         onSubmit={this.props.onSubmit ? this.props.onSubmit : this.onSubmit}
         onReset={this.onReset}
         {...cx('root')}
+        action=""
       >
         <svg xmlns="http://www.w3.org/2000/svg" style={{display: 'none'}}>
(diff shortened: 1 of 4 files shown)


packages/react-instantsearch/src/components/SearchBox.js
```diff
@@ -120,6 +120,7 @@ class SearchBox extends Component {
   onSubmit = e => {
     e.preventDefault();
     e.stopPropagation();
+    this.input.blur();
```
mthuret: Are we sure we want to change permanently this behaviour, removing the focus each time someone submit a search when `searchAsYouType` is disabled? This can be used even if not on mobile, even if I'm not sure it makes sense. What do you think?

bobylito: I only occurs when you submit so I guess that's ok.

Haroenv: although this *is* a breaking change in a way, a user expects something to happen when pressing "search". You would need to tab back into the input field, but that's how it is in other search that doesn't do search as you type (amazon for ex)

Haroenv:
on iOS the action is needed, because it shows as "return" when there's no action -> http://stackoverflow.com/a/26287843/3185307

Haroenv:
On Android it looks like this, so autosuggest is not disabled, but the return symbol is a search icon. Of course this is very fragmented, so won't be the same everywhere. 

Input type search alone (so this or wasn't needed on android) already shows the icon

img_1223

packages/react-instantsearch/src/components/SearchBox.js
vvo: Same behaviour on Google and Facebook, submit load the search and focusout. should be safe enough

Haroenv:
I'm wondering why you would squash three separate improvements. If you prefer it that way, no big deal, but I'd prefer to know why for example the action was added, without the other attributes mattering at that point.

vvo:
> I'm wondering why you would squash three separate improvements. If you prefer it that way, no big deal, but I'd prefer to know why for example the action was added, without the other attributes mattering at that point.

The way I deal with it is: are any of those commits self sufficient or are they interdependent and only make sense when all together? If none of the commits are providing a "good enough" state then I prefer to have them all in one and label it as

```text
fix(SearchBox): better mobile behaviour

more explanations here
```

Ultimately I would even add a code comment next to action="" because that's a "WTF" moment: there's no way from reading the code to understand why it's here. So adding a comment with a reference to either documentation or a good stackoverflow post is always good.

Haroenv:
Okay! In this case they all work independently, and that's why I made it in three commits in the first place. This PR will be visible in the blame history though, so it would still be useful. You can make the call :smile:

PR #2007, opened 2017-02-25
fix(example): link to instantsearch/react

Haroenv:

**Summary**

link to instantsearch/react instead of instantsearch/ in header of ecom example

**Result**

link is correct


vvo:
👌

PR #2011, opened 2017-02-27
fix(site): deduplicate styles for the site

Haroenv:

**Summary**

there were two css files linked, but they only differed in the `components/header` file, so they could be reused into one

pay attention to doocgen/layouts/example where I replaced header.css to index.css, but it doesn't make a big difference in the size, since index.css should already be cached at this point, and is requested on every other page.

cc @lukyvj

**Result**

css is not longer duplicated, /example/* loads the whole css


Haroenv:
don't merge, this breaks the examples

mthuret:
I don't know if it's related to this, but some time ago I fixed the css collision we had with our examples by only using `header.css` inside them.

Haroenv:
fixed by doing what you suggested in https://github.com/algolia/instantsearch.js/pull/2011/commits/78e2d03c8b8e3200837b1de71420b2021f73111e

Issue #1962, opened 2017-02-09
RIS: Ease multi index/autocomplete like experiences and getting multiple results in a single place/event


mthuret:
Use case: 

Today if you want to target multiple indices at the same time you need to listen to the `search state` and pass it back to each other `<InstantSearch/>` instances. Also if you want to use an external autocomplete component you need to aggregate all the data you need. 

API proposal: 

We want to simplify that. Here's a little code sample for the two use cases mentioned before. 

**Multi Hits** 

```js
<InstantSearch indexName="first"/>
      <Configure hitsPerPage={1}/> //shared between all indices
      <CustomHits /> //only hits of the first index
      <MultiIndexContext indexName="second">
          <Configure hitsPerPage={2}/>
          <CustomHits /> //only hits of the second index
      </MultiIndexContext>
      <MultiIndexContext indexName="third">
          <Configure hitsPerPage={4} />
          <CustomHits /> //only hits of the third index
      </MultiIndexContext>
</InstantSearch>
```

**External Usage of an Autocomplete component**

```js
<InstantSearch indexName="first"/>
      <WrapWithHits>
    <AutoComplete />
    <MultiIndexContext indexName="second">
      <VirtualAutoComplete />
    </MultiIndexContext>
    <MultiIndexContext indexName="third">
      <VirtualAutoComplete />
    </MultiIndexContext>
  </WrapWithHits >
</InstantSearch>

const AutoComplete = connectMultiHits(({hits, query, refine}) => <Autosuggest
  suggestions={hits}
  multiSection={true}
  onSuggestionsFetchRequested={({value}) => refine(value)}
  onSuggestionsClearRequested={() => refine('')}
  getSuggestionValue={hit => hit.name}
  renderSuggestion={hit =>
    <div>
      <div>{hit.name}</div>
    </div>
  }
  inputProps={{
    placeholder: 'Type a product',
    value: query,
    onChange: () => {
    },
  }}
  renderSectionTitle={section => section.index}
  getSectionSuggestions={section => section.hits}
/>);
```

Basically every widgets going under the `<MultiIndexContext/>` will override the main search parameters. Under the hood it means creating a derived helper for each new index to target. 
What `<MultiIndexContext/>` will do is add to each widget context the targeted index. If no multi context is found, then we use the `indexName` declared within `<InstantSearch/>`.

Also, in order to use an external autocomplete component we need to create a new connector that will embed all the hits, the query, and a refine function.  

[...]

PR #2025, opened 2017-03-03
Ease multi index and autocomplete with react-instantsearch

mthuret:
This a preview of what I've done to ease the multi index / autocomplete use case by using react-instantsearch. 

I'd love to have your feedback of what is already done. You can play with the feature by looking at the story "Multi Index". I also updated two recipes for you to see the changes. 

While developing this feature I change some internal implementations: 
* The query is now added as a metadata. The `connectCurrentRefinement` filter them if the prop `clearsQuery` is not their. 
* The pagination reset to first page is now explicitly done by each widget that needs it
* When unmounting a widget, we clear the empty key from the `search state`. Will probably be moved to the `onSearchStateChange` function. 

What I want to do next with this branch: 

* refactor the derivation usage inside `createInstantSearchManager`. I added some basic tests to support this refactoring. @bobylito if you have any advice, pleases shoot :) 
* add a real autocomplete example without the usage of an external component. Don't know if we should go for a recipe, or do a nice demo. 
* update all the documentation => api, guide. 
 
see #1962


Haroenv:
You accidentally made these commits as `Marie-Laure Thuret <mthuret@MacBook-Pro-de-Marie-Laure.local>`

mthuret:
@Haroenv: should be ok now ;) Thanks!

vvo:
Reflecting on the naming of MultiIndexContext, should we go for the shorter IndexContext? Should express the same feature.

docgen/src/guide/Multi_index.md
```diff
@@ -6,22 +6,19 @@ category: guide
...
+* display hits from different indices
+* share a single SearchBox
+* synchronize widgets between different indices
 
-Two props on the [InstantSearch component](widgets/InstantSearch.html) can be used to inject searchState or be notified of searchState changes:
+You can use multiple [`<MultiIndexContext>`](widgets/<MultiIndexContext>.html) instances.
 
-* onSearchStateChange(onSearchStateChange): a function being called every time the [`<InstantSearch>`](guide/<InstantSearch>.html) searchState is updated.
-* [searchState](guide/Search_state.html): a search state
+The `<MultiIndexContext>` component takes one props, the targetted index name.
```
rayrutjes: s/targgeted/targeted

packages/react-instantsearch/src/connectors/connectPagination.test.js
```diff
@@ -5,77 +5,107 @@ import {SearchParameters} from 'algoliasearch-helper';
...
-  it('registers its id in metadata', () => {
-    const metadata = getMetadata({}, {});
-    expect(metadata).toEqual({id: 'page'});
+    it('should return the right searchState when clean up', () => {
+      const newState = cleanUp({}, {page: {searchState: 'searchState'}, another: {searchState: 'searchState'}});
+      expect(newState).toEqual({another: {searchState: 'searchState'}});
+    });
   });
+  describe('multi index', () => {
+    let context = {context: {ais: {mainTargettedIndex: 'first'}, multiIndexContext: {targettedIndex: 'first'}}};
```
rayrutjes: Targetted ;)

docgen/src/guide/Multi_index.md
Haroenv: takes one props -> takes one prop

package.json
```diff
@@ -107,9 +107,6 @@
   "engines": {
     "node": "^6.x.x || ^7.x.x"
   },
-  "jest": {
-    "notify": true
```
Haroenv: Is this removed because we set those settings somewhere else? I like the notifications

packages/react-instantsearch/src/components/__snapshots__/Pagination.test.js.snap
```diff
@@ -62,7 +62,7 @@ exports[`Pagination allows toggling display of the first page button on and off
   >
     <a
       aria-label="Page 9"
-      className="ais-Pagination__itemLink"
+      className="ais-Pagination__itemLink ais-Pagination__itemLinkSelected"
```
Haroenv: Does this test render different things now?

packages/react-instantsearch/src/connectors/connectAutoComplete.js
```diff
@@ -0,0 +1,61 @@
+import createConnector from '../core/createConnector';
+import {omit} from 'lodash';
```
Haroenv: `import omit from 'lodash.omit'` or `'lodash/omit`

packages/react-instantsearch/src/connectors/connectAutoComplete.js
```diff
@@ -0,0 +1,61 @@
...
+
+const getId = () => 'query';
```
Haroenv: what's the point of this function?

packages/react-instantsearch/src/connectors/connectAutoComplete.js
```diff
@@ -0,0 +1,61 @@
...
+ * @kind connector
+ * @providedPropType {array.<object>} hits - the records that matched the search state
+ */
+export default createConnector({
+  displayName: 'AlgoliaAutoComplete',
+
+  getProvidedProps(props, searchState, searchResults) {
+    const hits = [];
+    if (searchResults.results) {
+      Object.keys(searchResults.results).forEach(index => {
```
Haroenv: Does this get polyfilled correctly? I think we're targeting IE9

packages/react-instantsearch/src/connectors/connectConfigure.js
```diff
@@ -1,7 +1,10 @@
 import createConnector from '../core/createConnector.js';
-import {omit, isEmpty, difference, keys} from 'lodash';
+import {omit, difference, keys} from 'lodash';
```
Haroenv: same as earlier, we can also use https://github.com/lodash/babel-plugin-lodash or https://github.com/lodash/lodash-webpack-plugin

packages/react-instantsearch/src/connectors/connectConfigure.js
```diff
@@ -1,7 +1,10 @@
...
+import {hasMultipleIndex, getIndex, refineValue} from '../core/indexUtils';
 
-const namespace = 'configure';
+function getId() {
```
Haroenv: You used an arrow function in the other place, arrow here too then?

packages/react-instantsearch/src/connectors/connectConfigure.test.js
```diff
@@ -4,49 +4,134 @@ import {SearchParameters} from 'algoliasearch-helper';
...
+      );
+      expect(searchState).toEqual({configure: {}});
+    });
   });
+  describe('multi index', () => {
+    let context = {context: {ais: {mainTargettedIndex: 'first'}, multiIndexContext: {targettedIndex: 'second'}}};
+    let cleanUp = connect.cleanUp.bind(context);
+    const getSearchParameters = connect.getSearchParameters.bind(context);
+
+    it('it propagates the props to the SearchParameters without children', () => {
```
Haroenv: you can probably leave out the `it` in this string, since it's implied

packages/react-instantsearch/src/connectors/connectMenu.js
```diff
@@ -87,11 +99,16 @@ export default createConnector({
       console.warn('react-instantsearch: `searchForFacetValues` has been renamed to' +
         '`withSearchBox`, this will break in the next major version.');
     }
+    if (props.withSearchBox && this.context.multiIndexContext) {
+      // eslint-disable-next-line no-console
+      console.error('react-instantsearch: searching in *List is not available when used inside a' +
```
vvo: Is a console.error sufficient or should we throw an error since it's no gonna work in the end? Here the code will try to go further.

vvo: Can you also add a comment for reference to understand why it's not available. And what would be needed to make it working.

packages/react-instantsearch/src/core/createInstantSearchManager.js
```diff
@@ -29,6 +29,9 @@ export default function createInstantSearchManager({
   helper.on('result', handleSearchSuccess);
   helper.on('error', handleSearchError);
 
+  let derivedHelpers = {};
+  let indexMapping = {}; // keep track of the original index where the parameters applied
```
vvo: Maybe the comment and initial values are not exactly the right one because we expect indexMapping to always be manipulated in a way that we "keep track of something between search calls" while it's always being re-assigned to an empty object at every search call.

packages/react-instantsearch/src/core/createInstantSearchManager.js
```diff
@@ -29,6 +29,9 @@ export default function createInstantSearchManager({
   helper.on('result', handleSearchSuccess);
   helper.on('error', handleSearchError);
 
+  let derivedHelpers = {};
```
vvo: if derivedHelpers is always re-assigned then it's initial value is not important and we should remove it so that we do not expect it to be used as a cache. It's just a shared variable between getSearchParameters() and search().

vvo: If I am mistaken because of some of the flow then let me know and we comment it maybe

stories/util.js
```diff
@@ -62,6 +62,7 @@ const WrapWithHits = ({
     appId="latency"
     apiKey="6be0576ff61c053d5f9a3225e2a90f76"
     indexName="ikea"
+    searchParameters={{...searchParameters}}
```
vvo: use Configure

vvo:
Miss:
- [ ] documentation on search state given index context new feature
- [ ] list of any breaking changes for reference
- [ ] when to use connectAutoComplete or IndexContext. So we present a clear path to one or the other solution

package.json
vvo: Because it was slow somehow, we just need to ensure it's a real bug and then track it on our side.

packages/react-instantsearch/src/connectors/connectAutoComplete.js
vvo: We do not have an automatic eslint rule for this so for now the consensus in react-instantsearch is to use the {} form. Ultimately it's preferable IF we can then use babel-lodash to automatically transpile to this form, but last time we checked it was not working I think.

No matter what we can discuss this in another issue no prob.

packages/react-instantsearch/src/connectors/connectAutoComplete.js
vvo: I think it's a notation used in every connector, sometime the id is not a single string, so we always have the same form to get the state id.

packages/react-instantsearch/src/connectors/connectAutoComplete.js
vvo: Object.keys is ES5 so it's safe.

packages/react-instantsearch/src/connectors/connectConfigure.js
vvo: Yes let's try to use that, not in this PR though

packages/react-instantsearch/src/connectors/connectConfigure.js
vvo: 👍  ultimately it would either be fixed with prettier I hope, pure stylistic

Haroenv: I don't think prettier changes functions to arrow functions, it has a difference in semantics

vvo: Indeed, then maybe an eslint rule. Always hard to agree and enforce on style/syntax when it's not possible to automatically check it.

docgen/src/guide/Multi_index.md
```diff
@@ -6,22 +6,19 @@ category: guide
...
 
-* displaying hits from different indices
-* sharing a single SearchBox
-* any use case involving synchronizing widgets between different [`<InstantSearch>`](guide/<InstantSearch>.html) instances
+* display hits from different indices
+* share a single SearchBox
+* synchronize widgets between different indices
 
-Two props on the [InstantSearch component](widgets/InstantSearch.html) can be used to inject searchState or be notified of searchState changes:
+You can use multiple [`<MultiIndexContext>`](widgets/<MultiIndexContext>.html) instances.
```
Haroenv: I don't really like the `MultiIndexContext` as a name. Can a better one be searched?

Maybe just `<Index />`, since the use case is multiple indices, where each of these components implies a separate index? How were you thinking of solving this in vue @rayrutjes

vvo: I proposed IndexContext in another comment, we could even go for Index indeed, very short and still valid.

docgen/src/guide/Multi_index.md
```diff
@@ -6,22 +6,19 @@ category: guide
...
 ---
 
-You can use multiple [`<InstantSearch>`](guide/<InstantSearch>.html) instances for cases like:
+Whenever you want to:
 
-* displaying hits from different indices
-* sharing a single SearchBox
-* any use case involving synchronizing widgets between different [`<InstantSearch>`](guide/<InstantSearch>.html) instances
+* display hits from different indices
+* share a single SearchBox
```
mthuret: Instead of this: "build an autocomplete menu targeting different indices " I would go for "building an autocomplete using an external component". Right now I'm building the demo/recipe to build one without the need of this special connector.

docgen/src/guide/Multi_index.md
```diff
@@ -6,22 +6,19 @@ category: guide
...
+* synchronize widgets between different indices
```
mthuret: I'm not sure yet. Apart from the synchronise that should be removed, we kind of agree that for now we should not highlight the fact that basically any widget can target any indices.

package.json
mthuret: The issue was "for each notif, a new process was created". I need to find the issue again and see if it hasn't been fixed since.

packages/react-instantsearch/examples/autocomplete/src/App-Multi-Index.js
```diff
@@ -1,180 +1,46 @@
...
-const VirtualAutoSuggest = connectAutoComplete(AutoComplete);
+const App = () =>
+  <InstantSearch
+    appId="latency"
+    apiKey="6be0576ff61c053d5f9a3225e2a90f76"
+    indexName="ikea">
+    <AutoComplete />
+    <Configure hitsPerPage={1} />
+    <MultiIndexContext indexName="bestbuy">
+      <VirtualAutoComplete />
```
mthuret: MultiIndexContext is not a widget per se. Then, in order to create a derivated helper, we need to have a widget under it. 

Any idea for improving this?

packages/react-instantsearch/src/components/ClearAll.js
```diff
@@ -10,13 +10,11 @@ class ClearAll extends Component {
     translate: PropTypes.func.isRequired,
     items: PropTypes.arrayOf(PropTypes.object).isRequired,
     refine: PropTypes.func.isRequired,
-    query: PropTypes.string,
   };
 
   render() {
-    const {translate, items, refine, query} = this.props;
```
mthuret: If I remember correctly this is a part when we need to decide if we go for a breaking change or not. It's also related to comment below (about currentRefinements and weird comment :d) 

Basically now queries are present under items (filtered when no usage of clearsQuery). With the introduction of MultiIndex, then you can have several queries at the same time. Thus this "single" query has a meaning only if use with our regular mode. 

Basically I'm in favor for a breaking change here, as it would be cleaner. It will impact those who were not using the "clearAll" widget with the "clearsQuery" props, but redo their own with the connector.

packages/react-instantsearch/src/components/__snapshots__/Pagination.test.js.snap
mthuret: It's due to the fact that I removed the `transitionState` function from the pagination. I need to check for further precision but I believe it comes from the test data context.

packages/react-instantsearch/src/connectors/connectConfigure.js
```diff
@@ -13,22 +16,25 @@ export default createConnector({
     return searchParameters.setQueryParameters(items);
   },
   transitionState(props, prevSearchState, nextSearchState) {
```
mthuret: I still want :) 
Needs more work on the Configure side. Would be done I hope, soon, with another PR.

packages/react-instantsearch/src/connectors/connectCurrentRefinements.js
```diff
@@ -20,10 +22,24 @@ export default createConnector({
...
+          if (props.clearsQuery && meta.id === 'query' && meta.items[0].currentRefinement === '') {
+            return res;
+          }
+          return res.concat(meta.items);
+        }
+      }
+      return res;
+    }, []);
+
+    // legacy
```
mthuret: See previously on "clearAll".

packages/react-instantsearch/src/connectors/connectHits.test.js
```diff
@@ -3,22 +3,45 @@
...
+const {getSearchParameters} = connect;
 
 describe('connectHits', () => {
-  it('provides the current hits to the component', () => {
-    const hits = [{}];
-    const props = getProvidedProps(null, null, {results: {hits}});
-    expect(props).toEqual({hits});
-  });
+  describe('single index', () => {
+    const context = {context: {multiIndexContext: {targettedIndex: 'index'}}};
```
mthuret: Agreed we can nested it. But I also think about removing the multi-index index name from the the context and basically forward this as a prop for each multi index widget. What do you think? 

I am currently working on a manual autocomplete, and I noticed that in order to build the key navigation, somehow I need to have a way to know the hits index. Right now it could have mean using the context from a user side (that's not good), or pass the index name as a props. It's might be hard to understand without any material, so when I'll push the example, I'll show you. 

We could also keep the context, and only for the hits widget, have the index name forward as a props.

vvo:
@mthuret is killing it at React conf so it's expected there will be some delay in this PR answer/deal but not big deal it's already solid and only requires adjustments/docs.

vvo:
Deploy preview ready!

Built with commit 134b3f2c4de771ed532aeefd4d2eba310606012a

https://deploy-preview-2025--algolia-instantsearch.netlify.com

Haroenv:
about the preview: it redirects `Index.html` to `index.html`, is there some way we can avoid that, maybe just by calling it `<Index>.html` for example?

packages/react-instantsearch/src/core/createInstantSearchManager.js
mthuret: If I remove this default value, then, I need to add this " if (derivedHelpers) " when looking for detaching helpers for the first time.

mthuret:
I'm in the middle of the refactoring based on your comments. I believe this is due to the new name "Index".

docgen/src/guide/Multi_index.md
```diff
@@ -6,22 +6,19 @@ category: guide
...
 
-You can use multiple [`<InstantSearch>`](guide/<InstantSearch>.html) instances for cases like:
+Whenever you want to:
 
-* displaying hits from different indices
-* sharing a single SearchBox
-* any use case involving synchronizing widgets between different [`<InstantSearch>`](guide/<InstantSearch>.html) instances
+* display hits from different indices
+* share a single SearchBox
+* build an autocomplete menu targeting different indices. Either manually or using an external autocomplete component (see [`connectAutoComplete`](connectors/connectAutoComplete.html)
```
rayrutjes: Missing a closing parenthesis here ;)
image

mthuret:
In term of changes: 

- Reseting the pagination should be done at each connector level inside the "refine" function when returning the search state. 
- The current page now appears inside the search state when a widget is used
- Query values are part of the items prop of the connectCurrentRefinements connector. Behaviour is unchanged, query will be filtered if `clearsQuery` prop is false.
- Add the index name to all the current refinements items. (not used by our widgets yet, but available if needed).

Pending questions: 

- Should we find a way to avoid the need of a VirtualWidget when using an external autocomplete component?

vvo:
> Should we find a way to avoid the need of a VirtualWidget when using an external autocomplete component?

Should be OK for beta release but ultimately might be not that easy

PR #2029, opened 2017-03-06
chore(helper): update dependency to algoliasearch-helper

bobylito:
Update algoliasearch-helper dependency to 2.19.0 which contains the access to raw results in the SearchResults. CC. @proudlygeek ;)

### DIFF
--- a/packages/react-instantsearch/package.json
+++ b/packages/react-instantsearch/package.json
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "algoliasearch": "^3.21.1",
-    "algoliasearch-helper": "^2.18.1",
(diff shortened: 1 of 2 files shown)


vvo:
To update the helper you need to bump it in packages/react-instantsearch.

To avoid any locking/upgrade issues, restart from v2 branch, checkout it, do `yarn boot`, the yarn.lock file should not change at all, then:

```
cd packages/react-instantsearch
yarn add algoliasearch-helper@2.19.0
```

Then you should be good, right now this PR is rolling back some package to old versions and adds the helper at the root level (instead of in packages/).

Let me know if you need help on this.

bobylito:
Thanks for hints @vvo. Will update that!

PR #2041, opened 2017-03-09
fix(Slider): display disabled slider when `min === max`

iam4x:

**Summary**

Instead of showing nothing when `range.min === range.max` of the slider, we show it on a disabled state.

**Result**

Active
screen shot 2017-03-09 at 13 47 44

Disabled
screen shot 2017-03-09 at 13 48 12

### DIFF
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -23,11 +23,16 @@ class Slider extends React.Component {
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


Haroenv:
Tests are failing though, you should check that

iam4x:
I've updated the tests, it should be all good now :+1:

vvo:
GG

Issue #1868, opened 2017-01-18
Use babel env preset


vvo:
And only compile for last two browsers

PR #2042, opened 2017-03-09
chore: switch to `babel-preset-env`

iam4x:

**Summary**
Use https://github.com/babel/babel-preset-env to target only 2 last versions of browsers and IE9 (sad react only 😭)

see: https://github.com/algolia/instantsearch.js/issues/1868

**Result**

Do we have access to browserstack or similar tool to test it live on other browsers?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,9 @@
 {
-  "presets": ["latest", "react", "stage-2"],
+  "presets": [
+    ["env", { "targets": { "browsers": ["last 2 versions", "ie >= 9"] } }],
+    "react",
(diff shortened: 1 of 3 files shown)


vvo:
> Do we have access to browserstack or similar tool to test it live on other browsers?

We do have a browserstack account, other than that I recommend VirtualBox + https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/.

Install this locally from an optic fiber source and then you can start IE9/10/11/Edge in 10s (literally) and even access a local server on your host machine.

Issue #1851, opened 2017-01-18
MD5 website assets


vvo:
All loaded assets should be MD5 in production otherwise everytime we deploy the website is messed up, some parts are refreshed some are not.

PR #2051, opened 2017-03-15
Feat/hash website assets

iam4x:

**Summary**
see: https://github.com/algolia/instantsearch.js/issues/1851

**Result**
image


vvo:
I am blown-away it just works, congrats!

docgen/plugins/rev-assets.js
```diff
@@ -0,0 +1,69 @@
...
+  const promises = files.map(
+    file => hasha
+      .fromFile(file, {algorithm: 'md5'})
+      .then(hash => ({file, hash}))
+  );
+
+  return Promise.all(promises);
+}
+
+function renameFiles(files: { file: string, hash: string }[]): { oldPath: string, newPath: string }[] {
```
vvo: we need a flow course from you I think.

redox:
nice @iam4x 👍

docgen/plugins/rev-assets.js
```diff
@@ -0,0 +1,69 @@
...
+  return replace({from, to, files: inFiles});
+}
+
+export default function revAssets(): Promise<*> {
+  // COMPUTE CSS
+  // -----------
+  const CSS_PATH = path.join(DIST_PATH, 'stylesheets');
+  const CSS_FILES = glob.sync(`${CSS_PATH}/**/*.css`);
+
+  const computeCSSFiles = () => computeHashForFiles(CSS_FILES)
```
vvo: I think if you want to be safe you should:
- compute MD5 for images
- replace in CSS, JS, HTML
- compute MD5 for CSS
- replace in JS, HTML
- compute MD5 FOR JS
- replace in HTML

This way you should be safe. Maybe from our setup this is already safe the way you did it?

iam4x: It is already safe since we want to remove cache from assets, but it would more "correct" to do it the way you describe. Go for it 👍

iam4x:
Should be all good @vvo now 🙌

PR #2091, opened 2017-04-03
docs(connectClearAll): add an example for connectClearAll

Haroenv:

**Summary**

Add a vanilla js example for connectClearAll.

Problems here: we will add an event listener on every render

'shitty' way around it: use `onclick` instead of event listener

**Result**

A button with Clear All


src/connectors/clear-all/connectClearAll.js
```diff
@@ -46,7 +46,8 @@ const clearAll = ({helper, clearAttributes, hasRefinements}) => () => {
 /**
  * Connects a rendering with the clearAll business logic.
  * @function connectClearAll
- * @param {function(ClearAllRenderingOptions)} renderFn function that renders the clear all widget
+ * @param {function(ClearAllRenderingOptions, boolean)} renderFn function that renders the clear all widget,
+ *        if the boolean is true, then it's the rendering that happens before someone searches
```
Haroenv: this might be added on *every* renderFn

bobylito: We might be able to provide a unique reference to a function that close a variable that we could update with the new values to clear each time there is a rendering. This way we could reference it in the first rendering and bind the correct event handler once.

PR #2092, opened 2017-04-03
feat(Highlight): allow a custom tag

Haroenv:

**Summary**

`<em>` isn't actually very suited for highlighting, a `<mark>` tag is way better, because it brings its own accessibility, default style etc. with it. 

To allow this, a new prop (`HighlightTagName`) was added to Highlight and Highlighter. 

**Result**

I added stories for `Highlight` to make this clear.


vvo:
Was closed since we moved to https://github.com/algolia/react-instantsearch/. Also this will be discussed between @mthuret and @Haroenv

PR #2096, opened 2017-04-05
chore(deps): update dependencies 2017-04-05

Haroenv:
**Summary**

Updated all dependencies except expect-jsx, since that one breaks the tests (cc @vvo)

**Result**

Dependencies are up to date


Haroenv:
too dangerous

PR #2097, opened 2017-04-05
docs(connectHits): make a custom hits example using bel

Haroenv:
**Summary**

[Bel](https://github.com/shama/bel) is an easy 'vanilla'-ish element create thingy, and it's nice to try if we play nice with it for hits. It turns out we do!

Theres's a way to transform these into real `document.createElement` calls with [yo-yoify](https://github.com/shama/yo-yoify#how-this-works), but that's browserify specific, and for production.

It's possible to switch this to [yo-yo](https://github.com/maxogden/yo-yo#dynamic-updates) which is does DOM diffing etc, but that seemed superfluous for this.

**Result**

A custom hits via a connector made with bel

### DIFF
--- a/dev/app/custom-widgets/vanilla/hits.js
+++ b/dev/app/custom-widgets/vanilla/hits.js
@@ -1,0 +1,37 @@
+/* eslint-disable import/default */
+import instantsearch from '../../../../index.js';
+import bel from 'bel';
+
+function render({hits, widgetParams: {containerNode}}) {
+  const content = bel`<div>
+    ${hits.map(hit => {
+      const title = bel`<h4></h4>`;
+      title.innerHTML = hit._highlightResult.name.value;
+      const description = bel`<p></p>`;
+      description.innerHTML = hit._highlightResult.description.value;
+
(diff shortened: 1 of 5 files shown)


dev/customWidgets/hitsBel.js
```diff
@@ -0,0 +1,37 @@
...
+          <div class="hit" id="hit-bel-${hit.objectID}">
+            <div class="media">
+              <a class="pull-left" href="${hit.url}">
+                <img class="media-object" src="${hit.image}">
+              </a>
+              <div class="media-body">
+                <h3 class="pull-right text-right text-info">$${hit.price}</h3>
+                <h4>${title}</h4>
+                <p>${description}</p>
+                ${hit.free_shipping ? bel`<span class="badge pull-right">Free Shipping</span>` : ''}
```
iam4x: I dont think you need to use `bel()` again

Haroenv: You do need to use it again 😄, otherwise it's just a template string, so just text

PR #2107, opened 2017-04-10
Feat/instantsearch.js/unify connector api

iam4x:
**WIP: this list may change in the near future**

* connectMenu
  * [x] easier implementation of `showMore`
  * [x] remove `currentRefinement` which is not used

* connectHits
  * [x] remove `hitsPerPage` config option

* connectClearAll
  * [x] provide `clearAll` fn on first render
  * [x] provide same instance/reference to `clearAll` fn
  * [x] add `{clearsQuery: boolean}` opt-in

* connectCurrentRefinedValues
  * [x] Move `displayOperator` logic into connector, compute `items[i].name` for the user
  * [x] Use `clearRefinement(item)` and `createURL(item)` instead of `clearRefinementURLs[]` and `clearRefinementClicks[]`
  * [x] add `{clearsQuery: boolean}` opt-in

* connectHitsPerPageSelector
  * [x] Rename `options[]` to `items[]`
  * [x] Add `isRefined` to the `items[]`
  * [x] Rename `connectHitsPerPageSelector` to `connectHitsPerPage`

* connectPriceRanges
  * [x] Make `refine()` taking named arguments instead of two arguments => `refine({ from: ?number, to: ?number })`

* connectToggle
  * [x] fix `createURL` fn which is not returning a string
  * [x] fix root level `count` when there's no off value
  * [x] Make `refine()` taking one argument which is the value of the toggle

* Other
  * [x] Remove `currentRefinement` from connectRefinementList, connectNumericRefinementList, connectPriceRanges, connectStarRating
  * [x] Investigate why pagination widget is not working in demo
  * [x] Check that `refine()` and `createURL()` takes `item.value` instead of a different value everywhere
    * [x] connectHierarchicalMenu
    * [x] connectMenu
    * [x] connectNumericRefinementList
    * [x] connectRefinementList
    * [x] connectStarRating


src/connectors/menu/connectMenu.js
```diff
@@ -95,20 +123,25 @@ export default function connectMenu(renderFn) {
...
         }, true);
       },
 
       render({results, instantSearchInstance}) {
         const items = results.getFacetValues(attributeName, {sortBy}).data || [];
+        this.toggleShowMore = this.createToggleShowMore({results, instantSearchInstance});
 
         renderFn({
-          items,
+          items: take(items, this.getLimit()),
```
bobylito: This could be inline as a variable in the body of `render`.

bobylito: What's the rational of using take instead of slice?

src/connectors/menu/connectMenu.js
```diff
@@ -56,13 +61,34 @@ export default function connectMenu(renderFn) {
...
+
+      createToggleShowMore({results, instantSearchInstance}) {
+        return () => {
+          this.isShowingMore = !this.isShowingMore;
+          this.render({results, instantSearchInstance});
+        };
+      },
+
+      getLimit() {
+        return showMoreLimit > limit && this.isShowingMore
```
bobylito: `showMoreLimit > limit` and `this.isShowingMore` should always be consistent right? Or is it to handle the case where showMoreLimit is actually used for showing less values?

bobylito: Oh showMoreLimit is always defined. Maybe if could be {null|number} instead of {number}?

src/connectors/menu/connectMenu.js
```diff
@@ -1,23 +1,25 @@
...
   //   currentRefinement,
+  //   isShowingMore,
+  //   toggleShowMore
   // }
 });
 search.addWidget(
   customMenu({
     attributeName,
     [ limit ],
+    [ showMoreLimit = 0 ]
```
bobylito: I think the default value should be undefined.

src/connectors/menu/connectMenu.js
iam4x: None, good point 👍

src/connectors/menu/connectMenu.js
iam4x: Hmmm, we will still have to check if `showMoreLimit` is greater than `limit` so the ShowMore/ShowLess logic works. We can maybe throw an error before by checking if it's here it should be greater and print `usage` in console?

bobylito: Yes, I think it's the best solution: checking the parameters as we receive them.

src/widgets/current-refined-values/current-refined-values.js
```diff
@@ -36,8 +36,8 @@ const renderer = ({
   attributes,
   clearAllClick,
   clearAllURL,
-  clearRefinementClicks,
-  clearRefinementURLs,
+  clearRefinement,
```
Haroenv: maybe call this `refine` too. maybe `unrefine`?  not so sure

src/widgets/current-refined-values/current-refined-values.js
```diff
@@ -115,6 +119,7 @@ currentRefinedValues({
  * @param {string} [$0.cssClasses.footer] CSS classes added to the footer element
  * @param {object|boolean} [$0.collapsible=false] Hide the widget body and footer when clicking on header
  * @param {boolean} [$0.collapsible.collapsed] Initial collapsed state of a collapsible widget
+ * @param {boolean} [$0.clearsQuery] also clears the active search query
```
Haroenv: maybe give the `$0` a name so that it gets displayed properly in the docs

Haroenv: say that the default is false in jsDoc

src/widgets/hierarchical-menu/__tests__/__snapshots__/hierarchical-menu-test.js.snap
```diff
@@ -126,7 +127,7 @@ exports[`hierarchicalMenu() render sets facetValues to empty array when no resul
       "templates": Object {
         "footer": "",
         "header": "",
-        "item": "<a class=\\"{{cssClasses.link}}\\" href=\\"{{url}}\\">{{name}} <span class=\\"{{cssClasses.count}}\\">{{#helpers.formatNumber}}{{count}}{{/helpers.formatNumber}}</span></a>",
+        "item": "<a class=\\"{{cssClasses.link}}\\" href=\\"{{url}}\\">{{label}} <span class=\\"{{cssClasses.count}}\\">{{#helpers.formatNumber}}{{count}}{{/helpers.formatNumber}}</span></a>",
```
Haroenv: Should we deprecate the templates now or in another major?

src/widgets/hits-per-page-selector/hits-per-page-selector.js
```diff
@@ -17,18 +17,19 @@ const renderer = ({
...
 }) => ({
-  currentRefinement,
-  options,
+  items,
   refine,
   hasNoResults,
 }, isFirstRendering) => {
   if (isFirstRendering) return;
 
+  const {value: currentValue} = items.find(({isRefined}) => isRefined) || {};
```
Haroenv: this should be `filter` instead of `find`, right? 

Also maybe just removed?

src/widgets/numeric-refinement-list/__tests__/numeric-refinement-list-test.js
```diff
@@ -7,6 +7,8 @@ import numericRefinementList from '../numeric-refinement-list.js';
 import RefinementList from '../../../components/RefinementList/RefinementList.js';
 expect.extend(expectJSX);
 
+const encodeValue = (start, end) => window.encodeURI(JSON.stringify({start, end}));
```
Haroenv: same as I said earlier

src/widgets/refinement-list/refinement-list.js
```diff
@@ -66,7 +66,7 @@ const renderer = ({
       showMore={showMoreConfig !== null}
       templateProps={renderState.templateProps}
       toggleRefinement={refine}
-      searchFacetValues={searchForFacetValues && searchForItems}
+      searchFacetValues={searchForFacetValues ? searchForItems : () => {}}
```
Haroenv: Why not undefined as default value?

Haroenv: Since people shouldn't call this in that case, right?

src/widgets/current-refined-values/current-refined-values.js
bobylito: You can't give it a name otherwise, it sorts of breaks the documentation.

src/widgets/hierarchical-menu/__tests__/__snapshots__/hierarchical-menu-test.js.snap
bobylito: The deprecation of templates is not final. Let's not rush it.

src/widgets/hits-per-page-selector/hits-per-page-selector.js
bobylito: HitsPerPage can only have one value so find is fine.

dev/app/custom-widgets/jquery/currentRefinedValues.js
```diff
@@ -4,8 +4,8 @@ import instantsearch from '../../../../index.js';
 const renderFn = ({
   clearAllClick,
   clearAllURL,
-  clearRefinementURLs,
-  clearRefinementClicks,
+  createURL,
+  clearRefinement,
```
iam4x: I don't think so, it's the opposite of `refine()` action?

src/connectors/current-refined-values/connectCurrentRefinedValues.js
```diff
@@ -65,6 +66,7 @@ Full documentation available at https://community.algolia.com/instantsearch.js/c
  * @typedef {Object} CurrentRefinedValuesWidgetOptions
  * @property {CurrentRefinedValuesAttributes[]} attributes specification for the display of refinements per attribute
  * @property {boolean} onlyListedAttributes limit the displayed refinement to the list specified
+ * @property {boolean} clearsQuery also clears the active search query
```
iam4x: 👍

src/connectors/current-refined-values/connectCurrentRefinedValues.js
```diff
@@ -214,3 +218,14 @@ function clearRefinementFromState(state, refinement) {
 function clearRefinement(helper, refinement) {
   helper.setState(clearRefinementFromState(helper.state, refinement)).search();
 }
+
+function computeLabel(value) {
```
iam4x: Do you mean to externalize somewhere else this logic or a way to override this function through connector options?

src/connectors/numeric-refinement-list/connectNumericRefinementList.js
```diff
@@ -77,49 +74,32 @@ export default function connectNumericRefinementList(renderFn) {
...
-        const items = options.map(facetValue =>
-          ({
-            ...facetValue,
-            isRefined: isRefined(helper.state, attributeName, facetValue),
-            attributeName,
-          })
-        );
+        this._prepareItems = state => options.map(({start, end, name: label}) => ({
+          label,
+          value: window.encodeURI(JSON.stringify({start, end})),
```
iam4x: Because a JSON structure breaks DOM node we need to escape some characters, the quick win is to use `window.encodeURI`

For instance this is not valid HTML and it will break markup:

```html
<div data-something="{"foo": "bar"}">
  Foobar
</div>
```

With `window.decodeURI` it becomes:

```html
<div data-something="%7B%22foo%22:%22bar%22%7D">
  Foobar
</div>
````

File: src/widgets/refinement-list/refinement-list.js (hunk 9 above)
Comment (bobylito) [inline src/widgets/refinement-list/refinement-list.js]: It should be undefined if it's not activated. In the current implementation of the refinement list component, the contract is that is there is a function for searchFacetValues, it means that the component should render a search box. If not, no search box.

https://github.com/algolia/instantsearch.js/blob/feat/instantsearch.js/v2/src/components/RefinementList/RefinementList.js#L166-L171

File: src/widgets/clear-all/__tests__/clear-all-test.js [hunk 14]
```diff
@@ -79,7 +79,7 @@ describe('clearAll()', () => {
       widget.render({results, helper, state: helper.state, createURL, instantSearchInstance: {}});
 
       expect(ReactDOM.render.calledTwice).toBe(true, 'ReactDOM.render called twice');
-      expect(ReactDOM.render.firstCall.args[0]).toEqualJSX(<ClearAll {...getProps()} />);
+      //expect(ReactDOM.render.firstCall.args[0]).toEqualJSX(<ClearAll {...getProps()} />);
```
iam4x: Good catch, this look suspicious 👍

src/connectors/current-refined-values/connectCurrentRefinedValues.js
bobylito: What do you mean?

src/widgets/current-refined-values/current-refined-values.js
iam4x: I'm good with `clearRefinement()`, what do you think @bobylito ?

src/connectors/numeric-refinement-list/connectNumericRefinementList.js
Haroenv: ah okay, I was missing what the reason for this was!

src/connectors/current-refined-values/connectCurrentRefinedValues.js
```diff
@@ -214,3 +218,14 @@ function clearRefinementFromState(state, refinement) {
...
+  if (value.hasOwnProperty('operator') && typeof value.operator === 'string') {
+    let displayedOperator = value.operator;
+    if (value.operator === '>=') displayedOperator = '≥';
```
bobylito: We can figure this out later when we have more operators.

dev/app/custom-widgets/jquery/currentRefinedValues.js
Haroenv: yes, maybe this just _is_ the best name.

src/connectors/current-refined-values/connectCurrentRefinedValues.js
Haroenv: not sure if it makes sense, but those symbols might be different in other alphabets etc, so this should be configurable (after we looked some up about it)

also some people might want to have this written in full "less than x more than y".

Might be an edge case though

src/widgets/refinement-list/refinement-list.js
iam4x: Because `RefinementList` is awaiting a `PropTypes.function` for `searchFacetValues` should we change it to `oneOfType([ ... ])` ?

src/widgets/numeric-refinement-list/__tests__/numeric-refinement-list-test.js
iam4x: Explained into https://github.com/algolia/instantsearch.js/pull/2107#discussion_r112229777

src/connectors/hits/connectHits.js
```diff
@@ -9,60 +9,43 @@ var customHits = connectHits(function render(params, isFirstRendering) {
...
 
- /**
-  * Connects a rendering function with the hits business logic.
-  * @param {function(HitsRenderingOptions, boolean)} renderFn function that renders the hits widget
-  * @return {function(CustomHitsWidgetOptions)} a widget factory for hits widget
-  */
+/**
+ * Connects a rendering function with the hits business logic.
+ * @param {function(HitsRenderingOptions, boolean)} renderFn function that renders the hits widget
+ * @return {function} a widget factory for hits widget
```
bobylito: `A function that returns a widget` is what we decided. However it could be done now or during the documentation review (but that would be too bad to miss it)

src/widgets/current-refined-values/current-refined-values.js
Haroenv: I thought we said something about giving it a typedef just for that?

src/widgets/refinement-list/refinement-list.js
Haroenv: maybe there's a way to just _not_ set the prop? 

Can the spread operator help with that or is it overkill

src/connectors/numeric-refinement-list/connectNumericRefinementList.js
bobylito: We should be able to pass this value into html strings, and the user doesn't need to know about the structure of it. So I guess that's fine.

src/widgets/current-refined-values/current-refined-values.js
bobylito: The rule in react-instantsearch is refine everything. Should apply here, no?

src/widgets/refinement-list/refinement-list.js
iam4x: I find it overkill and not beautiful to use the spread operator to achieve this, I can do either:

```jsx
searchFacetValues={searchForFacetValues ? searchForItems : () => {}}
```

or 

```jsx
searchFacetValues={searchForFacetValues ? searchForItems : undefined}
```

src/widgets/current-refined-values/current-refined-values.js
iam4x: If it is, let's do go for it.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -109,12 +109,11 @@ export default function connectHierarchicalMenu(renderFn) {
       _prepareFacetValues(facetValues, state) {
         return facetValues
           .slice(0, limit)
-          .map(subValue => {
+          .map(({name: label, path: value, ...subValue}) => {
```
iam4x: I'm renaming `path` to `label` this way, look:

```js
({ path: label }) => ({ label })
```

src/connectors/current-refined-values/connectCurrentRefinedValues.js
iam4x: Should we do it that way:

```
* @property {boolean = false} clearsQuery also clears the active search query
```

cc @bobylito ?

src/connectors/current-refined-values/connectCurrentRefinedValues.js
iam4x: I think it should come after if it's asked.

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -75,6 +73,9 @@ export default function connectRefinementList(renderFn) {
 
     checkUsage({attributeName, operator, usage});
 
+    const formatItems = ({name: label, ...item}) =>
+      ({...item, label, value: label, highlighted: label});
```
iam4x: It's used here https://github.com/algolia/instantsearch.js/blob/master/src/widgets/refinement-list/defaultTemplates.js#L8 I didn't change this part should be the same as before.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
Haroenv: Ah okay, I was reading it upside down

src/widgets/current-refined-values/current-refined-values.js
Haroenv: It's called `refine` too: https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch/src/components/CurrentRefinements.js

dev/app/custom-widgets/jquery/currentRefinedValues.js
iam4x: Go for `refine()` see: https://github.com/algolia/instantsearch.js/pull/2107#discussion_r112231898

src/connectors/menu/connectMenu.js
```diff
@@ -56,13 +58,35 @@ export default function connectMenu(renderFn) {
...
+      !isNaN(showMoreLimit) && showMoreLimit < limit
+    ) {
       throw new Error(usage);
     }
 
     return {
+      isShowingMore: false,
+
+      // Provide the same function to the `renderFn` so that way the user
+      // has to only bind it once when `isFirstRendering` for instance
```
iam4x: Done ✌️

src/widgets/current-refined-values/current-refined-values.js
iam4x: We should stick to the same API so I went for it 👍

src/widgets/refinement-list/refinement-list.js
bobylito: I think the least impactful is not to change the RefinementList, update the propType and use undefined. If only we had option types 🤓

dev/app/custom-widgets/jquery/currentRefinedValues.js
```diff
@@ -46,17 +46,8 @@ const renderFn = ({
...
-          if (value.operator === '<=') displayedOperator = '&le;';
-          value.name = `${displayedOperator} ${value.name}`;
-        }
-
-        return value;
-      })
-      .map(value => {
-        const {name, count} = value;
+        const {labelWithOperator, count} = value;
+        const name = labelWithOperator || value.name;
```
iam4x: Doesn't have always an operator, renamed `labelWithOperator` to `computedLabel` which defaults to `value.name`

see 0a6dbbf5

Issue #2056, opened 2017-03-17
Allow releasing beta versions


vvo:
Our release process should be able to release beta/rc versions of instantsearch.js and react-instantsearch.

For instantsearch.js, as long as we use jsDelivr auto npm sync we cannot do it. In the next breaking version we will need to stop depending on jsDelivr and start using unpkg.com or jsDelivr soon to be released npm proxy.

For react-instantsearch, for now we are using http://unpkg.com/ so we are good and @Haroenv is on it.

Haroenv:
The deployment of beta versions works for react-instantsearch, but the changelog will consider it as a completely separate version. For now we can update the changelog after it's been taken out of beta I think?

vvo:
For beta releases I think we should not change the changelog that would be better

Haroenv:
Good idea, another is to put it under [unreleased]

Haroenv:
Still needs to stay open for instantsearch.js

jimaek:
The new backend is live. It pulls files directly from npm and github https://github.com/jsdelivr/jsdelivr#usage

Example https://cdn.jsdelivr.net/npm/jquery@3.1.1
Example directory listing https://cdn.jsdelivr.net/npm/jquery@3.1.1/

Haroenv:
That's great, we only need to update our documentation to not use the custom version then :smile:

bobylito:
awesome news @jimaek Just in time for our next release :)

iam4x:
Works like a charm thanks @jimaek <3

bobylito:
Let's make it easy to publish betas.

PR #2109, opened 2017-04-11
fix(CONTRIBUTING): remove section about beta releases

Haroenv:
**Summary**

We don't have beta releases yet (see the [script](https://github.com/algolia/instantsearch.js/blob/develop/scripts/release.sh#L85-L92) and https://github.com/algolia/instantsearch.js/issues/2056), so the contributing section shouldn't be added yet.

We will put this back when `-b` works, and we've found a strategy for making it work with jsDeliver

**Result**

Remove the beta section in CONTRIBUTING for now

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,14 +43,6 @@ Files are automatically formatted with prettier.
 npm run release
 ```
 
-### Beta release
-
-```sh
-npm run release -- -b # or --beta
-```
-
-Append `-beta.x` where x is a number to the version for beta, so 4.0.0-beta.2 for example.
-
 ## Update docs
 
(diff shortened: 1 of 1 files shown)


bobylito:
Actually that would be great if we could do a beta release :)

Haroenv:
Yes, but it's not possible for now because of jsDeliver, it will always take the latest release, regardless of git tag...

bobylito:
Would it be possible to beta release in npm only? (I've had the issue you mention before :/) What if we push explicitly to jsDeliver, and do not rely on npm anymore for that?

Haroenv:
I would be okay with not using jsDeliver anymore, and using unpkg like in react-instantsearch

PR #2112, opened 2017-04-18
Feat/new range slider

iam4x:
⚠️ merge first https://github.com/algolia/instantsearch.js/pull/2107 I will re-open the PR after on the correct branch, this is just for keeping track of what is done and code review.

Use https://github.com/airbnb/rheostat as `range-slider` and should remove the usage of https://github.com/algolia/react-nouislider which is not meant to be maintained.

* fixes range slider issues with negatives values 🎉
* removes the "hacky" JS -> React bridge for nouislider
* tries to keep the same options as before for easy replacement for current client implementations
* fixes the wrong calculation of the "half" value (never got reported?)

### TODO:

* [x] POC
* [x] Tests
* [x] Remove legacy range-slider
* [x] Improve styles

### PREVIEW:

image

(price2)


src/components/Slider2/Pit.js
```diff
@@ -0,0 +1,34 @@
+import React, {PropTypes} from 'react';
+import cx from 'classnames';
+
+const Pit = ({style, children}) => {
```
Haroenv: I don't really understand this name

src/widgets/index.js
```diff
@@ -12,6 +12,7 @@ export {default as pagination} from '../widgets/pagination/pagination.js';
 export {default as priceRanges} from '../widgets/price-ranges/price-ranges.js';
 export {default as searchBox} from '../widgets/search-box/search-box.js';
 export {default as rangeSlider} from '../widgets/range-slider/range-slider.js';
+export {default as rangeSlider2} from '../widgets/range-slider2/range-slider2.js';
```
Haroenv: Why don't we deprecate the other slider and call that one rangeSliderOld or something, while giving this one the `rangeSlider` name?

src/components/Slider2/Pit.js
```diff
@@ -0,0 +1,34 @@
...
+
+  return (
+    <div
+      style={ {...style, marginLeft: positionValue === 100 ? '-2px' : 0} }
+      className={ cx('ais-range-slider--marker ais-range-slider--marker-horizontal', {
+        'ais-range-slider--marker-large': shouldDisplayValue,
+      }) }
+    >
+      { shouldDisplayValue
+          ? <div className="ais-range-slider--value">{ Math.round(children * 100) / 100 }</div>
```
bobylito: What about using [toFixed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) here ?

src/components/Slider2/Slider2.js
```diff
@@ -0,0 +1,101 @@
...
+    ]),
+  }
+
+  handleChange = ({values}) => {
+    const {refine} = this.props;
+    refine(values);
+  }
+
+  // creates an array number where to display a pit point on the slider
+  computeDefaultPitPoints = memoize(({min, max}) => {
```
bobylito: I wouldn't go for memoization unless we hit an identified perf issue.

src/widgets/index.js
iam4x: It's in the TODO 👍  It will be done right after the implementation is correct and tests are passing!

rayrutjes: Why don't we delete the other one all together given we will release a major version anyway?

src/components/Slider2/Pit.js
iam4x: from Slack conversation:

> .toFixed is not rounding to the superior decimal and is returning a string

image

src/components/Slider2/Slider2.js
iam4x: Works for me, during my tests I was using the same min/max values so found it relevant

src/widgets/index.js
iam4x: @rayrutjes Yes it's gonna be the case, like I said above and in the TODO :)

src/widgets/range-slider2/range-slider2.js
```diff
@@ -0,0 +1,123 @@
...
+      templatesConfig: instantSearchInstance.templatesConfig,
+      templates,
+    });
+    return;
+  }
+
+  const shouldAutoHideContainer = autoHideContainer && min === max;
+
+  const minValue = start[0] === -Infinity ? min : start[0];
+  const maxValue = start[1] === Infinity ? max : start[1];
```
iam4x: Should we do this check directly into the React component?

cc @bobylito

bobylito: In the connector probably?

iam4x:
Closing this PR for the new one opened on the correct branch 👍

PR #2120, opened 2017-04-20
fix(sffv): disable sffv input when few facet values FIX #2111

### LINKED ISSUE #2111: When the number of facet values retrieved is less than the number to display, IS should hide the search input
Not showing the "facet search text area" when the # of facets is less than the number being displayed (i.e. a facet returns 2 results, they are all displayed, and yet for some reason we still show the search to add more facets)

cc @seafoox

bobylito:
When the number of items displayed in a refinement list is lower than
the max number of facet displayed, the search for facet input should be
disabled.

This fix disables the input and adds a new classname to identify the
state of the input.

**Result**

[staging default](http://capricious-song.surge.sh/): the search input is not disabled.

fixes #2111 

[staging with a query that limits the items](http://capricious-song.surge.sh/?query=SanDisk%20-%20Ultra%20Plus%2032GB%20microSDHC%20Class%2010%20UHS-1%20Memory%20Card%20-%20Gray%2FRed&hits=3&idx=instant_search&p=0&hierarchical%5BhierarchicalCategories.lvl0%5D%5B0%5D=Cameras%20%26%20Camcorders&nR%5Bpopularity%5D%5B%3E%3D%5D%5B0%5D=0&is_v=1): the input is disabled.

### DIFF
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -165,7 +165,7 @@ class RefinementList extends React.Component {
         placeholder={this.props.searchPlaceholder}
         onChange={this.props.searchFacetValues}
         onValidate={() => this.refineFirstValue()}
-        disabled={displayedFacetValues.length < limit}/> :
+        disabled={!this.props.isFromSearch && displayedFacetValues.length < limit}/> :
       null;
 
     const noResults = this.props.searchFacetValues && this.props.isFromSearch && this.props.facetValues.length === 0 ?
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -164,7 +164,8 @@ class RefinementList extends React.Component {
       <SearchBox ref={i => { this.searchbox = i; }}
         placeholder={this.props.searchPlaceholder}
         onChange={this.props.searchFacetValues}
-        onValidate={() => this.refineFirstValue()}/> :
+        onValidate={() => this.refineFirstValue()}
+        disabled={displayedFacetValues.length < limit}/> :
       null;
 
     const noResults = this.props.searchFacetValues && this.props.isFromSearch && this.props.facetValues.length === 0 ?
--- a/src/components/SearchBox/index.js
+++ b/src/components/SearchBox/index.js
@@ -25,9 +25,7 @@ export default class SearchBox extends React.Component {
 
   render() {
     const {placeholder, onChange} = this.props;
-    const searchInput = this.props.disabled ?
-      <input type="search" name="search" placeholder={placeholder} autoComplete="off" required="required" className="sbx-sffv__input sbx-sffv__input-disabled" onChange={e => onChange(e.target.value)} ref={i => { this.input = i; }} disabled/> :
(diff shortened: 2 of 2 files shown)


src/components/SearchBox/index.js
```diff
@@ -42,7 +46,7 @@ export default class SearchBox extends React.Component {
         </svg>
 
         <div role="search" className="sbx-sffv__wrapper">
-          <input type="search" name="search" placeholder={placeholder} autoComplete="off" required="required" className="sbx-sffv__input" onChange={e => onChange(e.target.value)} ref={i => { this.input = i; }} />
+          {searchInput}
```
Haroenv: what about `{...this.props.disabled} />` instead of passing two different items

src/components/RefinementList/RefinementList.js
```diff
@@ -164,7 +164,8 @@ class RefinementList extends React.Component {
       <SearchBox ref={i => { this.searchbox = i; }}
         placeholder={this.props.searchPlaceholder}
         onChange={this.props.searchFacetValues}
-        onValidate={() => this.refineFirstValue()}/> :
+        onValidate={() => this.refineFirstValue()}
+        disabled={displayedFacetValues.length < limit}/> :
```
Haroenv: this disables it as well when you're searching

bobylito: Good catch 👍

bobylito:
Should be good @Haroenv

gabriel-kaam:
Hi @bobylito ( cc @Haroenv ; @seafoox ),

I'm having an issue with my SearchBox , this PR may be the cause.

See on my website : https://www.shopsquare.fr/mode/femme/bomber .

We have numerous filters with SearchBox on this page, but since this this fix has been deployed ; none of the SearchBox seems to be working.

See how all our filters' SearchBox (couleurs, marques ...) are disabled when they should not ? My users are actually crying :( 

**Why are the inputs disabled ?**

Because, even if we limit the results to 100 items , we still use an css overflow rule so the user is able to see all results by scrolling, while only displaying less items on the page at the same time .

capture d ecran 2017-05-18 a 15 25 14
( here we are limit the requests to only fetch 100 results, but we only display ~8 to the user at the same time ; while providing a way to scroll thru the entire set )

So even if this feature can be useful in some case, I think it should be easily disabled by config. And if you want my full opinion, I do not think it should be the default behaviour. A visible, but disabled input is always a bad experience for the user IMO.

Right now I going to have use a previous or modified version of the library ( without this fix ) on our website, which is not ideal. Could you provide me with a way to recover the initial behaviour so we at Shopsquare can sync with the CDN version ?

Cheers :) 

ping @julienpa

Haroenv:
Hey @gabriel-kaam, this should not cause your inputs to be disabled, we will look into how to support both the use case and the use case this pull request fixed. 

For now the only solution is to use an older version or a custom version of the library. 

Thanks for opening the discussion

Haroenv:
I’d love to be able to replicate how you load the extra facets, could you leave more info in #2156 @gabriel-kaam?

PR #2122, opened 2017-04-21
fix(sffv): add class for disabled state at the form level

bobylito:
The original fix for #2111 didn't add a class name at the form level, which prevents the user from completely hiding the form part of the refinement list when using search for facet values.

This patch adds that class.


seafoox:
Looks good. Thanks for the fix.

PR #2123, opened 2017-04-24
fix(url): link relatively to react-instantsearch

### LINKED ISSUE #2119: link to react-instantsearch in header is treated as a relative link
/instantsearch.js/react-instantsearch because of the _layouts/header.html using {{site.baseurl}}

noticed by @seafoox

Haroenv:
since the header adds `/instantsearch.js` to all urls in the nav (with `{{site.baseurl}}`), this is needed to be able to link to `/react-instantsearch` instead of `/instantsearch.js/react-instantsearch`.

reported by @seafoox

fixes #2119

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -15,7 +15,7 @@ navigation:
  - text: About
    url: /about/
  - text: React InstantSearch
-   url: /react-instantsearch/
+   url: /../react-instantsearch/
 # Build settings
 markdown: kramdown
 


docs/_config.yml
```diff
@@ -15,7 +15,7 @@ navigation:
  - text: About
    url: /about/
  - text: React InstantSearch
-   url: /react-instantsearch/
+   url: /../react-instantsearch/
```
vvo: Does it works if we put an absolute link also? It would fix it for good.

Haroenv: no, it won't, because then it's `/instantsearch.js/https://community.algolia.com/react-instantsearch`. This issue will go away all together in twoish weeks with the new docs anyway 😄

vvo:
changed commit to docs().

quick reminder: feat and fix are actual feat and fix of the library. Not docs, test, release script etc.. (tricky to handle but youll get used to it)

vvo:
This needs to be merged on master to be deployed

PR #2131, opened 2017-05-04
fix(sortby): Consistent across widget / connectors + migration

bobylito:
Add a migration guide for users using the default sortBy, and make the sortBy scheme consistent between widgets and connectors.

### DIFF
--- a/docgen/src/guides/migration.md
+++ b/docgen/src/guides/migration.md
@@ -26,3 +26,33 @@ var search = instantsearch({
   }
 });
 ```
+
+## The items are not sorted like before in the refinementList / menu
+
+We changed the default sort order of those widgets. This might have impacted your implementation
+if you didn't specify them originally. To change back the order use the `sortBy` configuration
+key.
+
+Here are examples of usage of `sortBy` using the previous sorting scheme:
+
+```javascript
+search.addWidget(
+  instantsearch.widgets.refinementList({
+    container: '#brands',
+    attributeName: 'brand',
+    sortBy: ['count:desc', 'name:asc'],
+  })
+);
+
+search.addWidget(
+  instantsearch.widgets.menu({
+    container: '#categories',
+    attributeName: 'categories',
+    sortBy: ['count:desc', 'name:asc']
+  })
+);
+```
+
+If you want to learn more about sorting the values, check out the widget API to see what are
+the valid values for the `sortBy` option of [menu](../widgets/menu.html#struct-MenuWidgetOptions-sortBy) or 
+[refinementList](../widgets/refinementList.html#struct-RefinementListWidgetOptions-sortBy)
--- a/src/connectors/refinement-list/connectRefinementList.js
+++ b/src/connectors/refinement-list/connectRefinementList.js
@@ -18,7 +18,7 @@ search.addWidget(
     attributeName,
     [ operator = 'or' ],
     [ limit ],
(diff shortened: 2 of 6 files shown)


docgen/src/guides/migration.md
```diff
@@ -26,3 +26,24 @@ var search = instantsearch({
   }
 });
 ```
+
+## The items are not sorted like before in the refinementList / menu / hierarchicalMenu widget
+
+We changed the default sort order of those widgets. This might have impacted your implementation
```
Haroenv: changed from what to what?

docgen/src/guides/migration.md
```diff
@@ -26,3 +26,24 @@ var search = instantsearch({
...
+if you didn't specify them originally. To change back the order use the `sortBy` configuration
+key.
+
+```javascript
+search.addWidget(
+  instantsearch.widgets.refinementList({
+    container: '#brands',
+    attributeName: 'brand',
+    sortBy: ['name:desc'],
```
iam4x: I was expecting the old behaviour as example here, `sortBy: ['isRefined', 'count:desc']`

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -38,7 +38,7 @@ export const checkUsage = ({attributeName, operator, usageMessage}) => {
  * @property {string} attributeName The name of the attribute in the records.
  * @property {"and"|"or"} [operator = "or"] How the filters are combined together.
  * @property {number} [limit] The max number of items to display.
- * @property {string[]|function} [sortBy = ['isRefined', 'count:desc']] How to sort refinements. Possible values: `count|isRefined|name:asc|name:desc`.
+ * @property {string[]|function} [sortBy = ['isRefined', 'count:desc', 'name:asc']] How to sort refinements. Possible values: `count|isRefined|name:asc|name:desc`.
```
iam4x: Consistency shouldn't be only `'name:asc'` ?

bobylito: It is now consistent with the widget.

docgen/src/guides/migration.md
bobylito: Do we want to go into the details of it?

bobylito: What about reconcialating both your comments and make the examples for each widget to use the previous scheme?

docgen/src/guides/migration.md
bobylito: See @Haroenv comment :)

src/connectors/refinement-list/connectRefinementList.js
iam4x: Alright 👍

docgen/src/guides/migration.md
Haroenv: Probably fine, but just a small comment will do right

bobylito:
The thing is that with those changes we do not change the sorting behavior of hierarchicalMenu \o/

bobylito:
To give you more context @Haroenv:
 - isRefined used first helps the user not loosing facet values that were selected but are not very important for the current search context
 - we went along on all connectors that were proposing sortBy
 - forgot to proceed on widgets
 - and this change also brings the default scheme meaningful for the type of widget: hierarchical need and menu need name first because they are for navigation (few items). Refinement list is for drilling in the data (more facets)

PR #2133, opened 2017-05-09
InstantSearch.js v2

bobylito:
This PR contains the changes for the V2 of InstantSearch.js.

This new version brings:
 - connectors API, and with it the capability of reusing the logic of the widgets without their rendering
 - better documentation, with more in-depth description of the different aspects of the library, a getting started and guides
 - some clean up in the API


iam4x:
I've implemented every widgets with connector using only the docs and I made a list of mistakes in the docs that I can easily fix before release:

* connectMenu
    * canToggleShowMore is undefined
* connectCurrentRefinedValues
    * doc type of `refinements` rendering option
* connectHierarchicalMenu  
    * typo for `attributes` in documentation of widget option
    * `showParentLevel` wrong type
* connectNumericRefinementList
    * Example uses item.label and docs says use item.name
    * No docs about operator widget params
* connectNumericSelector
    * Default operator is not displayed correctly in doc
    * Still has `currentRefinement` in doc
    * Add the possible operator options
* connectPriceRanges
    * Missing items.url doc

BUT, all the code example of connectors are written in ES5. We should rewrite all connectors examples with ES6, it’s too hard to read when the code is a bit complex. What do you think?

Issue #2134, opened 2017-05-10
[v2] remove all references to "Android" in docs


Haroenv:
- [x] favicon
- [x] meta headers
- [x] title
- [x] docsearch index name
- [x] dist folder name
- [x] header background (mobile --> web)

PR #2135, opened 2017-05-10
Fix/v2/meta updates

Haroenv:
**Summary**

Some references of the Android website remained

**Result**

Rename some safe to rename bits.

Tracking issue: #2134

### DIFF
--- a/docgen/assets/js/main.js
+++ b/docgen/assets/js/main.js
@@ -5,7 +5,7 @@ var alg = require('algolia-frontend-components/javascripts.js');
 
 const docSearch = {
   apiKey: 'fd5e835f5153cad7d5ec0c3595dfa244',
-  indexName: 'instantsearch.js',
+  indexName: 'instantsearch.js-v2',
   inputSelector: '#searchbox',
 };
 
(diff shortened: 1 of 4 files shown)


docgen/assets/js/main.js
```diff
@@ -5,9 +5,9 @@ var alg = require('algolia-frontend-components/javascripts.js');
 
 const docSearch = {
   apiKey: 'fd5e835f5153cad7d5ec0c3595dfa244',
-  indexName: 'instantsearch-android',
-  inputSelector: '#searchbox'
-}
+  indexName: 'instantsearch.js',
```
Haroenv: This might need more configuring somewhere?

docgen/assets/js/main.js
```diff
@@ -5,9 +5,9 @@ var alg = require('algolia-frontend-components/javascripts.js');
 
 const docSearch = {
   apiKey: 'fd5e835f5153cad7d5ec0c3595dfa244',
-  indexName: 'instantsearch-android',
-  inputSelector: '#searchbox'
-}
+  indexName: 'instantsearch.js-v2',
```
Haroenv: now the docsearch won't work until we make the index, but I think that's a fine tradeoff, just remember to make it before launching

PR #2140, opened 2017-05-15
feat(searchFunction): Update API, fix #1924

### LINKED ISSUE #1924: Make search function provide a better API
This is a feature for instantsearch.js V2 -- breaking change

Right now, we have a function that can be given to the instantsearch factory called searchFunction. This function is given a helper that can't be modified, only the search function can be really used. Instead we could only provide a single function `search` as a parameter.

## API proposal

```js
instantsearch({
  searchFunction: (search, searchParameters) => {
     search(searchParameters.setQuery(''));
  };
});
```

In this example, the function will prevent the query to be set.

bobylito:


bobylito:
Let's hope we don't find any weird things afterwards :D

PR #2142, opened 2017-05-16
feat(range-slider): use `rheostat` as slider component

iam4x:
See #2112

### DIFF
--- a/docgen/src/guides/migration.md
+++ b/docgen/src/guides/migration.md
@@ -88,6 +88,16 @@ As of now, we consider the engine used to build the widgets in InstantSearch.js
 as an implementation detail. Since we do not expose it anymore, we'll be able
 to change it and use the best solution for each release.
 
+## RangeSlider widget is using Rheostat as Slider Component
+
+Since we had unclosed issues with the old slider, we switched to a new
+well maintained component [Rheostat](https://github.com/airbnb/rheostat).
+
+If you want to customize the style, some CSS classes have changed. You can find
+examples of stylesheets by clicking [here](https://github.com/airbnb/rheostat/tree/master/css).
+
+We are still providing a look which is similar as the one in the V1.
+
 ## searchFunction can be used to modify parameters
 
 Introduced in 1.3.0, `searchFunction` was originally meant as a way to modify
--- a/docgen/src/guides/migration.md
+++ b/docgen/src/guides/migration.md
@@ -90,8 +90,9 @@ to change it and use the best solution for each release.
 
 ## RangeSlider widget is using Rheostat as Slider Component
 
-Since we had unclosed issues with the old slider, we switched to a new
-well maintained component [Rheostat](https://github.com/airbnb/rheostat).
+Slider compoments are hard to implement and that's why we rely on an external
+component for that. We are taking the opportunity of this new version
+to switch to the current state of the art of sliders: [Rheostat](https://github.com/airbnb/rheostat).
 
 If you want to customize the style, some CSS classes have changed. You can find
 examples of stylesheets by clicking [here](https://github.com/airbnb/rheostat/tree/master/css).
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -36,9 +36,9 @@ class Slider extends Component {
 
   // creates an array number where to display a pit point on the slider
   computeDefaultPitPoints({min, max}) {
-    const totalLenght = max - min;
+    const totalLength = max - min;
     const steps = 34;
-    const stepsLength = totalLenght / steps;
+    const stepsLength = totalLength / steps;
 
     const pitPoints = [min, ...times(steps - 1, step => min + stepsLength * (step + 1)), max]
       // bug with `key={ 0 }` and preact, see https://github.com/developit/preact/issues/642
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -1,92 +1,99 @@
-import React from 'react';
-import omit from 'lodash/omit';
+import times from 'lodash/times';
+import range from 'lodash/range';
+import has from 'lodash/has';
 
-import Nouislider from 'react-nouislider';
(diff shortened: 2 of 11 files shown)


src/components/Slider/Slider.js
```diff
@@ -1,92 +1,99 @@
...
+    refine(values);
   }
 
-  // we are only interested in rawValues
-  handleChange(formattedValues, handleId, rawValues) {
-    this.props.onChange(rawValues);
+  // creates an array number where to display a pit point on the slider
+  computeDefaultPitPoints({min, max}) {
+    const totalLenght = max - min;
+    const steps = 34;
```
bobylito: Why 34?

src/components/Slider/Slider.js
```diff
@@ -1,92 +1,99 @@
...
+    const {refine} = this.props;
+    refine(values);
   }
 
-  // we are only interested in rawValues
-  handleChange(formattedValues, handleId, rawValues) {
-    this.props.onChange(rawValues);
+  // creates an array number where to display a pit point on the slider
+  computeDefaultPitPoints({min, max}) {
+    const totalLenght = max - min;
```
bobylito: should be `totalLength`

src/components/Slider/Slider.js
```diff
@@ -1,92 +1,99 @@
...
+      // bug with `key={ 0 }` and preact, see https://github.com/developit/preact/issues/642
+      .map(pitPoint => pitPoint === 0 ? 0.00001 : pitPoint);
+
+    return pitPoints;
+  }
+
+  // creates an array of values where the slider should snap to
+  computeSnapPoints({min, max, step}) {
+    const r = range(min, max, step);
+    return r.includes(max) ? r : [...r.slice(0, -1), max];
```
bobylito: Range never includes max, https://lodash.com/docs/4.17.4#range Not sure what we're trying to achieve, but there might be an edge case if the last element of `r` is `max - step`.

Also mutating r would be more conservative. `r.slice(0, -1).push(max)` or even using `splice`.

src/components/Slider/Slider.js
```diff
@@ -1,92 +1,99 @@
...
-    React.PropTypes.arrayOf(
-      React.PropTypes.shape({
-        to: React.PropTypes.func,
-      })
-    ),
-  ]),
-};
-
-export default autoHideContainerHOC(headerFooterHOC(RawSlider));
+export default autoHideContainerHOC(headerFooterHOC(Slider));
```
bobylito: 👍  A preview of some changes to come

src/widgets/range-slider/range-slider.js
```diff
@@ -72,15 +70,11 @@ const usage = `Usage:
 rangeSlider({
   container,
   attributeName,
-  [ tooltips=true ],
-  [ templates.{header, footer} ],
```
bobylito: We probably don't want to remove the templates from the usage right? Nor the cssClasses, collapsible or autoHideContainer?

src/components/Slider/Slider.js
iam4x: Backport of react-nouislider number of markers on the range

src/components/Slider/Slider.js
iam4x: Didn't notice it will NEVER include it, we can have in the end:

```js
return [...range(min, max, step), max]
```

redox:
That's awesome 👍

src/components/Slider/Slider.js
bobylito: Mok :)

docgen/src/guides/migration.md
```diff
@@ -88,6 +88,16 @@ As of now, we consider the engine used to build the widgets in InstantSearch.js
 as an implementation detail. Since we do not expose it anymore, we'll be able
 to change it and use the best solution for each release.
 
+## RangeSlider widget is using Rheostat as Slider Component
+
+Since we had unclosed issues with the old slider, we switched to a new
```
Haroenv: I would rephrase this more like `it had more features and we want to use it`, now it’s a bit negative

src/components/Slider/__tests__/Slider-test.js
```diff
@@ -1,76 +1,23 @@
...
-          stepped: true,
-          values: [0, 50, 100],
-        } }
-        range={ {min: props.range.min, max: props.range.min + 0.0001} }
-        start={ props.start }
-        disabled
+describe('Slider', () => {
+  it('should render correctly', () => {
+    const tree = renderer.create(
+      <Slider
```
Haroenv: Why do you explicitly render this, is this for more than shallow rendering?

src/components/Slider/Slider.js
```diff
@@ -1,92 +1,99 @@
-import React from 'react';
-import omit from 'lodash/omit';
+import times from 'lodash/times';
+import range from 'lodash/range';
+import has from 'lodash/has';
 
-import Nouislider from 'react-nouislider';
+import React, {Component, PropTypes} from 'react';
```
iam4x: It's dealt properly with preact for now, we can have an issue open for v2.x 👍

docgen/src/guides/migration.md
iam4x: It's negative for the past v1, so good for us nope?

src/components/Slider/__tests__/Slider-test.js
iam4x: Shallow rendering will only render the top component, so in that case it would have been the `<AutohideContainerHOC />`.

I just render it like this to prevent further regression with snapshot testing, it's a dummy test case. Look for `range-slider-test.js` on the widget itself where I test the rendering and the logic 👍

docgen/src/guides/migration.md
Haroenv: I don’t agree, we can say that it’s better for the reasons that it’s better, but I don’t think we should say that it was bad before.

PR #2144, opened 2017-05-16
Fix for url sync issue in SPA

lggorman:
I'm using Instantsearch in the context of a single page application.  Each time the user hits the page with the search field, a new instance of Instantsearch is created, and whenever they navigate away all handlers are destroyed.  

This generally works, with the exception of the urlSync feature.  Since the `firstRender` variable is defined outside of the URLSync class, once it is set to false the code in lines 127 - 129 never executes again, even when I spin up a new Instantsearch instance.  As a result that the `onHistoryChange` handler is never attached, and the url does not update when the query is changed.  This pr sets the `firstRender` instead to be an attribute on the `URLSync` class, so that it will be reset every time the class is re-instantiated.  This revision is one resolution to the problem, but I'd be interested to hear if you all have other suggestions.  Thanks!

### DIFF
--- a/src/lib/url-sync.js
+++ b/src/lib/url-sync.js
@@ -6,7 +6,6 @@ import assign from 'lodash/assign';
 
 const AlgoliaSearchHelper = algoliasearchHelper.AlgoliaSearchHelper;
 const majorVersionNumber = version.split('.')[0];
-let firstRender = true;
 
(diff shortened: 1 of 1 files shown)


lggorman:
@bobylito Great, thanks for the quick response!

PR #2146, opened 2017-05-16
test(functional): make them work, explain how to run

vvo:
I must say I am unsure about the added value of those functional tests.

It adds a good amount of code and orchestration. Ultimately starting
from now (if we merge this), they should run on Travis and I hope one
day they will crash and we will be able to say "Phew! Thanks we had
those functional tests".

Let me know what you think


bobylito:
> "Phew! Thanks we had
> those functional tests".

I have mixed feelings about making the functional test fail 😅

vvo:
> I have mixed feelings about making the functional test fail

If they fail on Travis it's a WIN for us. If they fail in production they it's a FAIL :D

functional-tests/config.js
```diff
@@ -13,12 +13,12 @@ let conf = {
     timeout: 50000,
     compilers: ['js:babel-core/register'],
   },
-  baseUrl: 'http://localhost:9000',
+  baseUrl: 'http://10.200.10.1:9000',
```
bobylito: What is this IP?

CONTRIBUTING.md
```diff
@@ -42,6 +42,32 @@ Watch mode:
...
+Locally it will use a docker image bundling Selenium and browsers.
+
+```sh
+docker pull elgalu/selenium
+# this helps in loading the host website from the container,
+# comes from https://docs.docker.com/docker-for-mac/networking/#known-limitations-use-cases-and-workarounds
+sudo ifconfig lo0 alias 10.200.10.1/24
+docker run -d --name=grid -p 4444:24444 -p 6080:26080 -p 5900:25900 \
+    -e TZ="US/Pacific" -e NOVNC=true -v /dev/shm:/dev/shm --privileged elgalu/selenium
+npm run test:functional
```
bobylito: Is it something that we should execute?

vvo: Yes this is the basic setup locally to launch the functional tests. I will had more details.

functional-tests/config.js
vvo: It's the one of your HOST machine when the browsers inside docker will try to access the test web server app.

It would have been better to keep "localhost" but docker cannot do that on osx right now sadly.

bobylito: 👌

PR #2147, opened 2017-05-17
chore(travis): don’t build the docs when it’s a test

### LINKED ISSUE #2110: Remove docs build from Travis

both I guess

Travis installs and runs the docs

Travis doesn’t care about the docs 

Have faster builds

Rely on Netlify for the build status of the docs

See also https://github.com/algolia/react-instantsearch/issues/33.

This can be done when we move towards the new docs in v2, since the docs don’t run yet :)

Haroenv:
The Jekyll build is slowing down the travis build, but Netlify already does this build on another machine, and also gives a status when the build fails. 

If it’s a release the build would even run twice, so there’s no need to run it at all in test (see [update-website.sh#L25](https://github.com/algolia/instantsearch.js/blob/feat/instantsearch.js/v2/scripts/docs/update-website.sh#L25))

fixes #2110 

Should I also do this PR on the develop branch for v1?

### DIFF
--- a/scripts/test-ci.sh
+++ b/scripts/test-ci.sh
@@ -4,16 +4,14 @@ set -ev # exit when error
 
 ./scripts/validate-pr-done-on-develop.sh
 npm test
+
 # we need to build to run functional tests
 NODE_ENV=production npm run build
 INDEX_PAGE=index npm run test:functional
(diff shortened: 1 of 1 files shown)


Haroenv:
Lol but Travis doesn’t run on this branch

PR #2151, opened 2017-05-17
Publish v2 docs to github pages

iam4x:
* `$ yarn dev` -> now runs dev app & docs
* `$ yarn docs:publish` -> build for production & publish docs to gh-pages
* `$ yarn docs:build:dev` -> build locally docs (useful for debugging purpose & used by netlify)

Docs are now built in the `/docs` root folder for any env (no more `/docs` and `/docs-preview`)


.gitignore
```diff
@@ -3,4 +3,5 @@ dist/
 dist-es5-module/
 npm-debug.log
 .DS_Store
-docs-preview/
+docs/
+!scripts/docs/
```
bobylito: what does the `!` mean?

package.json
```diff
@@ -7,17 +7,15 @@
...
     "build": "./scripts/build.sh",
-    "dev": "./scripts/dev.sh",
+    "dev": "./scripts/dev-docs.sh",
     "staging": "./scripts/staging-deploy.sh",
-    "dev:docs": "./scripts/dev-docs.sh",
     "doctoc": "doctoc --maxlevel 3 README.md CONTRIBUTING.md",
     "lint": "npm run lint:js && npm run lint:css",
     "lint:js": "eslint .",
     "lint:css": "./scripts/lint-css",
-    "docs:jsdoc": "babel-node ./scripts/docs/gen-widget-doc.js",
```
bobylito: We should remove the script itself.

.gitignore
iam4x: `docs/` match every folder with that name, like for `node_modules/` so we need to exclude from gitignore `/scripts/docs` with `!` 👍

bobylito: Ah ok nice 👍

package.json
Haroenv: Isn’t this also used in `test-ci.sh`, if so it should be removed there too.

PR #2157, opened 2017-05-19
Pass facetsCount to header template of refinementList

timkelty:
**Summary**

I needed to display my refinementList header like so (showing the total selected out of total options):
**My Attribute Label** (3/20)

**Result**

I simply also passed a `facetsCount` prop to the `header` template.

Ideally I could have used `transformData`, but `transformData.header` doesn't exist.
Plus, I think this is sensible default data to pass along.

### DIFF
--- a/src/widgets/refinement-list/refinement-list.js
+++ b/src/widgets/refinement-list/refinement-list.js
@@ -34,7 +34,7 @@ const bem = bemHelper('ais-refinement-list');
  * @param  {object} [options.showMore.templates.inactive] Template used when showMore not clicked
  * @param  {object} [options.showMore.limit] Max number of facets values to display when showMore is clicked
  * @param  {Object} [options.templates] Templates to use for the widget
- * @param  {string|Function} [options.templates.header] Header template, provided with `refinedFacetsCount` data property
+ * @param  {string|Function} [options.templates.header] Header template, provided with `facetsCount`, `refinedFacetsCount` data properties
  * @param  {string|Function} [options.templates.item] Item template, provided with `name`, `count`, `isRefined`, `url` data properties
  * @param  {string|Function} [options.templates.footer] Footer template
  * @param  {Function} [options.transformData.item] Function to change the object passed to the `item` template
@@ -134,10 +134,11 @@ function refinementList({
       return createURL(state.toggleRefinement(attributeName, facetValue));
     }
 
-    // Pass count of currently selected items to the header template
+    // Pass count of total and currently selected items to the header template
+    const facetsCount = facetValues.length;
     const refinedFacetsCount = filter(facetValues, {isRefined: true}).length;
     const headerFooterData = {
-      header: {refinedFacetsCount},
+      header: {facetsCount, refinedFacetsCount},
     };
 
     // Do not mistake searchForFacetValues and searchFacetValues which is the actual search


src/widgets/refinement-list/refinement-list.js
```diff
@@ -134,10 +134,11 @@ function refinementList({
       return createURL(state.toggleRefinement(attributeName, facetValue));
     }
 
-    // Pass count of currently selected items to the header template
+    // Pass count of total and currently selected items to the header template
+    const facetsCount = facetValues.length;
```
bobylito: `facetValues.length` is the number of values that are retrieved from Algolia. (if an attribute has 3  values then it will be 3). Won't the sum of counts be more interesting?

Haroenv: Yes, that's what @jonasba was looking for yesterday too

timkelty: Sure.
My use-case was for an output in the header like so:
**Attribute Name** _(3/20)_

...where _3/20_ is selected `refinedFacetsCount`/`totalFacetsCount`

timkelty:
@bobylito ↓

> Also with the new connector of v2, this wouldn't require a change, just a new rendering.

That sounds intriguing - can you point me to any example or docs to pick apart?

src/widgets/refinement-list/refinement-list.js
timkelty: So do you think we should go through with changes to this PR, or is using a v2 connector the prescribed approach?

bobylito:
Given that it would require to go through all the widget templates and that it is doable in [v2 (which is in beta](https://community.algolia.com/instantsearch.js/v2/)), I don't think we should move on with this PR.

However we really greatly appreciate that you took the time to do the implementation and we hope that you will contribute again :)

timkelty:
Sounds good, thanks for taking the time to review it.

timkelty:
> Also with the new connector of v2, this wouldn't require a change, just a new rendering.

@bobylito Sorry to resurrect this thread, but I dug into this and have some follow-up questions.

So, I ended up getting this working simply by copying https://github.com/algolia/instantsearch.js/blob/feat/instantsearch.js/v2/src/widgets/refinement-list/refinement-list.js into my project, making all the import references non-relative, and tweaking what I needed to.

Works, but still seems like overkill for this particular tweak (since I wanted to keep 99% of the default refinementList rendering). Perhaps there should be an easier way to add data to the templates? Assuming I could get at the facet value counts, it seems like adding `header` to `RefinementListTransforms` could be a way to achieve this: https://community.algolia.com/instantsearch.js/v2/widgets/refinementList.html#struct-RefinementListTransforms

Let me know what you think.

bobylito:
@timkelty no problem. Each pain / roadblock is something that we deeply care about. I'm happy to continue the discussion.

If the information is available to the transformData function for the header, it'll be also available to the templates. I understand that ultimately it feels a lot of work and copying the inner mechanics of InstantSearch.js is not something that we want to promote either. But on the other hand it feels counterproductive to only add a single new piece of information.

What we had in mind when doing the connectors is that making a new widget should be simple enough because you, as a developper / InstantSearch.js user, you wouldn't need to handle the intricacies of managing customisation parameters.

Can you share with us the code that you had to implement, so we could see if we could have done it in a simpler way?

timkelty:
@bobylito Certainly, here's the implementation: https://gist.github.com/timkelty/7186587110e649b210d3f948e5420b5f

> I understand that ultimately it feels a lot of work and copying the inner mechanics of InstantSearch.js is not something that we want to promote either.

To be fair, it was actually quite easy - the unfortunate part is now my refinementList widget has to be maintained separately. It does, of course, still use the connector, so the entire refinementList logic didn't need to be copied.

Issue #2138, opened 2017-05-11
handle xss in templates


bobylito:
cc @alexandremeunier

rayrutjes:
Not sure if this is related, but here is what I did in WordPress to protect against XSS: https://github.com/algolia/algoliasearch-wordpress/commit/42ac2362155a7f90a9c00481d5e05cae107084dc

bobylito:
Thanks @rayrutjes :) Any different view on the implementation is helpful. The main question is what API can we provide to the user? Do we try to magically do it for them? Do we provide hogan helpers?

PR #2158, opened 2017-05-22
Hits xss

iam4x:
* [x] escape HTML entities only from `value` field of an `_highlightResult` or `_snippetResult`
* [x] replace from highlighted result `__ais-highlight__` with `<em>` tags
* [x] opt-in to enable escape of HTML

fix for https://github.com/algolia/instantsearch.js/issues/2138

### DIFF
--- a/src/connectors/hits/__tests__/connectHits-test.js
+++ b/src/connectors/hits/__tests__/connectHits-test.js
@@ -88,4 +88,56 @@ describe('connectHits', () => {
     expect(secondRenderingOptions.hits).toEqual(hits);
     expect(secondRenderingOptions.results).toEqual(results);
   });
+
+  it('escape highlight properties if requested', () => {
+    const rendering = sinon.stub();
+    const makeWidget = connectHits(rendering);
+    const widget = makeWidget({escapeHits: true});
+
+    const helper = jsHelper(fakeClient, '', {});
+    helper.search = sinon.stub();
+
+    widget.init({
+      helper,
+      state: helper.state,
+      createURL: () => '#',
+      onHistoryChange: () => {},
+    });
+
+    const firstRenderingOptions = rendering.lastCall.args[0];
+    expect(firstRenderingOptions.hits).toEqual([]);
+    expect(firstRenderingOptions.results).toBe(undefined);
+
+    const hits = [
+      {
+        _highlightResult: {
+          foobar: {
+            value: '<script>__ais-highlight__foobar__/ais-highlight__</script>',
+          },
+        },
+      },
+    ];
+
+    const results = new SearchResults(helper.state, [{hits}]);
+    widget.render({
+      results,
+      state: helper.state,
+      helper,
+      createURL: () => '#',
+    });
+
+    const escapedHits = [
+      {
+        _highlightResult: {
+          foobar: {
+            value: '&lt;script&gt;<em>foobar</em>&lt;/script&gt;',
+          },
+        },
+      },
+    ];
+
+    const secondRenderingOptions = rendering.lastCall.args[0];
+    expect(secondRenderingOptions.hits).toEqual(escapedHits);
+    expect(secondRenderingOptions.results).toEqual(results);
+  });
 });
--- a/src/connectors/hits/__tests__/connectHits-test.js
+++ b/src/connectors/hits/__tests__/connectHits-test.js
@@ -92,7 +92,7 @@ describe('connectHits', () => {
   it('escape highlight properties if requested', () => {
     const rendering = sinon.stub();
     const makeWidget = connectHits(rendering);
-    const widget = makeWidget({escapeHits: true});
+    const widget = makeWidget();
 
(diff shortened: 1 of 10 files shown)


iam4x:
I added `escapeHits` and `escapeHitsWhitelist` options and documentation 👍 
Should be good now!

src/connectors/hits/__tests__/connectHits-test.js
```diff
@@ -52,7 +55,10 @@ describe('connectHits', () => {
   it('Provides the hits and the whole results', () => {
     const rendering = sinon.stub();
     const makeWidget = connectHits(rendering);
-    const widget = makeWidget();
+    const widget = makeWidget({
+      escapeHits: true,
```
bobylito: What about we use the namespace pattern that we use for other options? This could look like:

```js
const widget = makeWidget({
  escapeHits: true,
});
```

or with the list:

```js
const widget = makeWidget({
  escapeHits: {include: ['title', 'tags']},
});
```

src/connectors/hits/__tests__/connectHits-test.js
```diff
@@ -80,8 +105,31 @@ describe('connectHits', () => {
...
+            wontEscape: '<h1>Not escaped</h1>',
+            value: '<a href="#top"><em>Go to top</em></a>',
+          },
+        },
+      },
+    ];
+
     const secondRenderingOptions = rendering.lastCall.args[0];
-    expect(secondRenderingOptions.hits).toEqual(hits);
+    expect(secondRenderingOptions.hits).toEqual(processedHits);
```
bobylito: We should probably add tests without the escape mechanism and one that tests the white list option.

src/connectors/hits/connectHits.js
```diff
@@ -48,23 +70,126 @@ Full documentation available at https://community.algolia.com/instantsearch.js/c
...
+  };
+}
+
+function replaceWithEm(val) {
+  if (isPlainObject(val)) {
+    return reduce(
+      val,
+      (result, value, key) => {
+        result[key] = replaceWithEm(value);
+        return {};
```
bobylito: Why does it return an empty object here? Does it mean that we could replace `reduce` with a `forEach`?

src/connectors/hits/connectHits.js
```diff
@@ -48,23 +70,126 @@ Full documentation available at https://community.algolia.com/instantsearch.js/c
...
+
+  return val;
+}
+
+function escapeHighlightProperty(highlightResult, whitelist) {
+  return reduce(
+    highlightResult,
+    (result, value, key) => {
+      // Stop here don't escape this key, it's whitelisted.
+      if (whitelist.includes(key)) {
```
bobylito: This should use `lodash.includes`

src/connectors/hits/connectHits.js
```diff
@@ -48,23 +70,126 @@ Full documentation available at https://community.algolia.com/instantsearch.js/c
...
+  );
+}
+
+function recursiveEscape(val, whitelist) {
+  if (isPlainObject(val)) {
+    return reduce(
+      val,
+      (result, value, key) => {
+        // dont escape theses values, they are whitelisted
+        if (whitelist.includes(key)) {
```
bobylito: This should use `lodash.includes`

bobylito:
Could you deploy it on a staging website using `yarn staging`? :)

src/connectors/hits/connectHits.js
iam4x: Good catch, will add a test which fails for it. It should be `return result;`

src/connectors/hits/connectHits.js
iam4x: Let's use `.indexOf` no need for `_.includes` if it's not supported by some browsers.

bobylito: 👍

src/widgets/hits/__tests__/hits-test.js
```diff
@@ -11,6 +11,10 @@ describe('hits call', () => {
   it('throws an exception when no container', () => {
     expect(hits).toThrow();
   });
+
+  it('throws an execption when `escapeHitsWhitelist` is wrong format', () => {
+    expect(() => hits({escapeHitsWhitelist: /foobar/})).toThrow();
+  });
```
iam4x: remove this not needed test anymore

src/widgets/hits/hits.js
```diff
@@ -110,6 +112,8 @@ export default function hits({
   cssClasses: userCssClasses = {},
   templates = defaultTemplates,
   transformData,
+  escapeHits = false,
+  escapeHitsWhitelist = [],
```
iam4x: remove previous option from doc

iam4x:
We should be good @bobylito 👍 

There is now an opt-in for escaping hits, asked by @olance on Slack.

iam4x:
Deployed on staging: http://deadpan-swing.surge.sh/

PR #2161, opened 2017-05-22
fix(show-more): should hide button when show more is not available

### LINKED ISSUE #2160: RefinementListWidgetOptions.showMore should conditionally render
`RefinementListWidgetOptions.showMore` template always renders, regardless of # of facets or `RefinementListWidgetOptions.limit`.

It should only render when the requested # of facets is greater than `RefinementListWidgetOptions.limit` (v1 behavior), or some sort of disabled class should be applied so it can be hidden with CSS.

https://community.algolia.com/instantsearch.js/v2/widgets/refinementList.html#struct-RefinementListWidgetOptions-showMore

Looks like it was brought up here, but I'm not seeing it fixed in `v2.0.0-beta.2`: https://github.com/algolia/instantsearch.js/issues/1455.

This works as expected in v1. The show more block isn't rendered at all if it is not applicable.

`v2.0.0-beta.2`

bobylito:
As noticed by @timkelty in #2160, showMore button should not be rendered
if there are not enough items to display based on the limits. With the
new connector API, this means that we should add a new parameter to
the renderer to give the information that this button should be
displayed.

fixes #2160

http://faithful-anger.surge.sh/?query=Canon%20-%20PowerShot%20SX720%20HS%2020.3-Megapixel%20Digital%20Camera%20-%20Red&idx=instant_search&p=0&dFR%5Bbrand%5D%5B0%5D=Canon&hierarchical%5BhierarchicalCategories.lvl0%5D%5B0%5D=Cameras%20%26%20Camcorders&nR%5Bpopularity%5D%5B%3E%3D%5D%5B0%5D=0&is_v=2

Check the menu widget and the collapsible refinement list.

### DIFF
--- a/src/connectors/menu/connectMenu.js
+++ b/src/connectors/menu/connectMenu.js
@@ -51,6 +51,8 @@ Full documentation available at https://community.algolia.com/instantsearch.js/c
  * @property {Object} widgetParams All original `CustomMenuWidgetOptions` forwarded to the `renderFn`.
  * @property {boolean} isShowingMore True if the menu is displaying all the menu items.
  * @property {function} toggleShowMore Toggles the number of values displayed between `limit` and `showMore.limit`.
+ * @property {boolean} canToggleShowMore `true` if the toggleShowMore button can be activated (enough items to display more or
+ * already displaying more than limit items)
  */
 
  /**
@@ -171,11 +173,13 @@ export default function connectMenu(renderFn) {
           widgetParams,
           isShowingMore: this.isShowingMore,
           toggleShowMore: this.cachedToggleShowMore,
+          canToggleShowMore: false,
         }, true);
       },
 
       render({results, instantSearchInstance}) {
-        const items = (results.getFacetValues(attributeName, {sortBy}).data || [])
+        const facetItems = results.getFacetValues(attributeName, {sortBy}).data || [];
+        const items = facetItems
           .slice(0, this.getLimit())
           .map(({name: label, path: value, ...item}) => ({...item, label, value}));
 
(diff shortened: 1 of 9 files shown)


src/connectors/menu/connectMenu.js
```diff
@@ -51,6 +51,8 @@ Full documentation available at https://community.algolia.com/instantsearch.js/c
  * @property {Object} widgetParams All original `CustomMenuWidgetOptions` forwarded to the `renderFn`.
  * @property {boolean} isShowingMore True if the menu is displaying all the menu items.
  * @property {function} toggleShowMore Toggles the number of values displayed between `limit` and `showMore.limit`.
+ * @property {boolean} canToggleShowMore `true` if the toggleShowMore button can be activated (enough items to display more or
+ * already displaying more than limit items)
```
iam4x: The `isShowingMore` description is not in sync with this one, we should stick to one style only.

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -65,6 +65,7 @@ export const checkUsage = ({attributeName, operator, usageMessage, showMoreLimit
  * @property {function} searchForItems Searches for values inside the list.
  * @property {boolean} isFromSearch `true` if the values are from an index search.
  * @property {boolean} canRefine `true` if a refinement can be applied.
+ * @property {boolean} canToggleShowMore `true` if the show more feature is available and there are enough items to show more
  * @property {Object} widgetParams All original `CustomRefinementListWidgetOptions` forwarded to the `renderFn`.
  * @property {boolean} isShowingMore True if the menu is displaying all the menu items.
```
iam4x: Same let's keep one style only for documentation 👍

iam4x:
Great 💯  !

src/connectors/menu/connectMenu.js
bobylito: Indeed 👍

PR #2163, opened 2017-05-23
fix(sffv): when using a large limit, retain the search

### LINKED ISSUE #2156: Disabling SFFV when there are few results visible keeps it disabled in some cases

bug

Could you make a [jsfiddle](https://jsfiddle.net/vvoyer/0wm3donb/4/) that replicates this issue, it might be hard to find your exact use case otherwise @gabriel-kaam. Mostly how you load the extra facets.

SFFV stays disabled when less than the maximum facets are visible, and more are loaded

SFFV to work when there are more than enough results

all

Make the disabling of sffv opt-in instead of a default.

v1.11.11 I guess, what version do you use @gabriel-kaam?

More detailed description at https://github.com/algolia/instantsearch.js/pull/2120#issuecomment-302405242

bobylito:
In the case we're using the widget with a very big limit (number of
items displayed) in the refinement list, we often still have a huge
number of items. Therefore we should be able to continue searching
in the list, even though all the items are displayed.

Fixes #2156

### DIFF
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -160,19 +160,15 @@ class RefinementList extends React.Component {
         /> :
         undefined;
 
-    let searchInput = null;
-    if (this.props.searchFacetValues) {
-      searchInput = this.props.searchIsAlwaysActive ?
-        <SearchBox ref={i => { this.searchbox = i; }}
-          placeholder={this.props.searchPlaceholder}
-          onChange={this.props.searchFacetValues}
-          onValidate={() => this.refineFirstValue()}/> :
+    const shouldDisableSearchInput =
+      this.props.searchIsAlwaysActive !== true &&
+      !(this.props.isFromSearch || displayedFacetValues.length >= limit);
+    const searchInput = this.props.searchFacetValues ?
         <SearchBox ref={i => { this.searchbox = i; }}
           placeholder={this.props.searchPlaceholder}
           onChange={this.props.searchFacetValues}
           onValidate={() => this.refineFirstValue()}
-          disabled={!this.props.isFromSearch && displayedFacetValues.length < limit}/>;
-    }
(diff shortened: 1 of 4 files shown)


bobylito:
Staging: http://screeching-jar.surge.sh/?query=dx&hits=3&idx=instant_search&p=0&hierarchical%5BhierarchicalCategories.lvl0%5D%5B0%5D=Cameras%20%26%20Camcorders&nR%5Bpopularity%5D%5B%3E%3D%5D%5B0%5D=0&is_v=1

In this example you can see that the filter limits the number of items in the searchable refinement lists. The bottom one doesn't prevent you from searching into it.

src/components/RefinementList/RefinementList.js
```diff
@@ -160,13 +160,19 @@ class RefinementList extends React.Component {
...
+      searchInput = this.props.searchIsAlwaysActive ?
+        <SearchBox ref={i => { this.searchbox = i; }}
+          placeholder={this.props.searchPlaceholder}
+          onChange={this.props.searchFacetValues}
+          onValidate={() => this.refineFirstValue()}/> :
+        <SearchBox ref={i => { this.searchbox = i; }}
+          placeholder={this.props.searchPlaceholder}
+          onChange={this.props.searchFacetValues}
+          onValidate={() => this.refineFirstValue()}
+          disabled={!this.props.isFromSearch && displayedFacetValues.length < limit}/>;
```
Haroenv: Why not change this `this.props.isFromSearch` to be a const we decide somewhere else, based on `searchIsAlwaysActive` as well?

bobylito:
Will merge && publish monday (too dangerous to update that before holidays)

vvo:
Nice!

bobylito:
Whoa that's 3 positive reviews :) Mergin' thanks for taking the time :)

PR #2167, opened 2017-05-29
fix(sass): set correct modifiers and elements

Iuriy-Budnikov:
Fixed default mixins for bem

Can't use default styles

### DIFF
--- a/src/css/_base.scss
+++ b/src/css/_base.scss
@@ -5,13 +5,13 @@
 }
 
 @mixin element($element) {
-  &--#{$element} {
+  &__#{$element} {
     @content;
   }
 }
 
 @mixin modifier($modifier) {
-  &__#{$modifier} {
+  &--#{$modifier} {
     @content;
   }
 }


vvo:
Hi @Iuriy-Budnikov thanks for participating in this repository!

We knew we messed up the BEM notation when we first released InstantSearch.js but thought "Well it's just a notation, people will not really bother".

Since you opened this PR, it must be bothering to you, could you give us more context on how our current BEM notation is affecting your work?

Ultimately this change is a breaking change because most websites would then fail with their current CSS so we cannot change it just like that.

Let me know what you think.

Iuriy-Budnikov:
Hey @vvo,

I tried to override default sass styles in my app.
I just cloned css folder and added a few custom styles to any components.
I noticed that mixins ( element and modifier ) have a typo.

Thanks

vvo:
Thanks, we cannot merge this right now and we will discuss it with the core team. In the meantime, is there any way so that you can continue working?

Maybe create a dedicated sass mixin for Algolia to replicate this naming?

bobylito:
Arguably we can say that something is wrong with our choice. We mixed up the `__` and `--` and that made a divergence from the classic [two dashes style naming convention variant](https://en.bem.info/methodology/naming-convention/#two-dashes-style).

However, reading the convention from the beginning of the article, we can argue that the syntax doesn't really matter as long as the Block Element Modifier pattern is used while naming the css classes. And that's what is implemented in the current state of the css file that we provide.

Also we can see that if we wanted to go for purity we should use the underscore syntax, which is the one advertise from the [start of the article](https://en.bem.info/methodology/naming-convention/#naming-convention) and that we can [see in action](https://en.bem.info/methodology/naming-convention/#example-of-using-the-naming-convention) in the example section.

I think it boils down to letting the user choose the convention, no?

vvo:
> I think it boils down to letting the user choose the convention, no?

Agree if doable, but our CSS will still be "static" even if the convention changes in components no?

bobylito:
> Agree if doable, but our CSS will still be "static" even if the convention changes in components no?

Indeed, our base css is static but it's backed by sass files that can be configurable, right? So one can choose to override the conventions and customise the sass sources?

vvo:
> So one can choose to override the conventions and customise the sass sources?

I guess yes

bobylito:
Thanks for your contribution @Iuriy-Budnikov We are currently working on homogenizing the DOM and CSS of all InstantSearch projects. Your contribution was very valuable and we will make sure that we apply the correct standard on the next iteration.

Issue #2173, opened 2017-05-31
"Hide results on init" example breaks URL sync


Spone:
**Do you want to request a *feature* or report a *bug*?**

A bug

**Bug: What is the current behavior?**

I use the example code from the documentation: https://community.algolia.com/instantsearch.js/documentation/#hide-results-on-init

```js
var search = instantsearch({
  [...],
  searchFunction: function(helper) {
    var searchResults = $('.search-results');
    if (helper.state.query === '') {
      searchResults.hide();
      return;
    }
    helper.search();
    searchResults.show();
  }
}
```

Here is a JSFiddle: https://jsfiddle.net/4jqgawpk/3/

Open the HTML iframe: https://fiddle.jshell.net/4jqgawpk/3/show/ (please try in Firefox, somehow the URL is not updated in Chrome, probably because of JSFiddle)

Type a search, wait for the URL params to update

Empty the search box

The URL params are not reset, the last search query is still there.

**Bug: What is the expected behavior?**

When I empty the search box, I expect the URL to be "clean" (at least the `q` param should be empty).

**Bug: What browsers are impacted? Which versions?**

Firefox latest, Chrome latest

**Bug: What is the proposed solution?**

There should probably be a way to use the JS helper to reset the URL params. If possible, the example from the documentation should be updated to use it.

bobylito:
Hi @Spone 

I came up with a solution to your problem. Indeed now InstantSearch.js relies on the `search` event to update the URL, and in your code you are preventing the search from happening. But there is a way to do that in the [following jsFiddle](https://jsfiddle.net/bobylito/e2jd75oL/1/).

What I did is change the implementation of the searchFunction to have this:

```js
  searchFunction: function(helper) {
    var searchResults = $('#movies');
    if (helper.state.query === '') {
      movies.helper.once('result', function() {
        searchResults.hide();
      })
    }
    helper.search();
    searchResults.show();
  }
```

In this case, there will be a search but as soon as the results come back it hides the content of the hits.

Spone:
Thanks @bobylito for this solution!

It's not ideal as it triggers a search every time the page loads, even if the user is not searching (I'm trying to limit the number of requests to Algolia to stay in our quota).
In my case, the items on the page are already rendered server-side.

Do you see another option?

bobylito:
Not for now. I'll be looking for a fix :/

bobylito:
[...]

Issue #2105, opened 2017-04-10
When using `searchOnEnterKeyPressOnly` the URL is still updated at every keystroke


bobylito:
The correct behavior should be that the URL updates only when there is a new search (on enter key press).

bobylito:
When first fixing this bug, we changed the internal implementation of the URL sync to update the URL when a `search` event instead of a `change` event at the helper level. However this triggered a series of unforeseen side effects:
 - `search` only happens when there is an actual call to the `search()` method. Which prevents the update the URL to be updated when there is no actual search. When we want to hide most of the search interface when the query of the searchbox is empty, this makes it impossible to update the URL when it is empty without actually doing a search that would have no impact because we don't want to display the results then.
 - Using the `search` implies some tricks when using the back button (there is a [PR for implementing a state](https://github.com/algolia/instantsearch.js/pull/2174) for onPopState update)

Overall it feels like using this approach was not the good one as we have to continue fixing the multiple issues. However, we could also try to make sure that we don't trigger a JS Helper state when the `searchOnEnterKeyPressOnly` option is activated. This way we could go back to using the simpler implementation of the URL sync based on the `change` event.

PR #2174, opened 2017-05-31
fix(url-sync): let the user go back in history (fix #2171)

### LINKED ISSUE #2171: When URL sync is active, it requires two click to go back one step
Bug.

**Bug: What is the current behaviour?**
When click on the browser back button, it requires two click to move back one step. Furthermore, when going back, it seems that we can't go forward (which indicates that we might have a new state that is recorded)

On https://demo.algolia.com/instant-search-demo/, follow these steps:
 - select slowly three brands in the brand refinement list
 - go back once --> goes back right away
 - go back again --> nothing happens
 - go back again --> goes a step back (can't go forward from there)

**Bug: What is the expected behaviour?**
One click --> one step back and have the ability to go forward.

tested on chrome, latest.

🤷‍♀️

This bug is reproducible in both v1 and v2.

bobylito:
To test that the behaviour is fixed, go to the [staging website](http://unhealthy-direction.surge.sh/).

Scenario:
 - select slowly three brands in the brand refinement list
 - go back once --> goes back right away
 - go back again --> goes a step back (can move forward from there)

This cannot be reproduced there: https://demo.algolia.com/instant-search-demo/

### DIFF
--- a/src/lib/url-sync.js
+++ b/src/lib/url-sync.js
@@ -88,7 +88,8 @@ function getFullURL(relative) {
 // IE <= 11 has no location.origin or buggy
 function getLocationOrigin() {
   // eslint-disable-next-line max-len
-  return `${window.location.protocol}//${window.location.hostname}${window.location.port ? `:${window.location.port}` : ''}`;
+  const port = window.location.port ? `:${window.location.port}` : '';
+  return `${window.location.protocol}//${window.location.hostname}${port}`;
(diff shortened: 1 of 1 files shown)


bobylito:
After seeing a few feedback, it does not appear clear that continuing to use the `search` event might actually be the right solution. Closing this for now. See #2105 #2173

Issue #2117, opened 2017-04-20
When using searchOnEnterKeyPress + search, the url is not updated after the first query


bobylito:
This a bug affects instantsearch.js 1.x

When using searchOnEnterKeyPress with a custom searchFunction that prevent some cases of search (empty query, query with under n letters), the url is not updated after the first results. The problem is identified [here](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/url-sync.js#L126-L130). The url sync bind the URL mechanics only after the first results have come back from Algolia.

This is a problem when the user blocks the initial first search using a `searchFunction` (which makes sense as they want to delay the first search until a user actually uses the search). This creates a limbo state for instantsearch making the url sync fail.

This is a tough one, because instantsearch.js has been created with the assumption that we're doing an initial search.

Issue #1015, opened 2016-05-12
searchBox: update the query state when using searchOnEnterKeyPressOnly


vvo:
When using searchOnEnterKeyPressOnly and willing to bind an external button to trigger a search (like a submit button), this will fail because the search state when clicking on submit will not be the right one.

From http://stackoverflow.com/questions/37101823/how-to-fire-a-search-with-instantsearch-js-when-clicking-on-a-button

This should work.

Also ultimately this will be solved by providing the full form/binding a search button in the searchBox: #1060

bobylito:
I had a look at the jsFiddle and it seems that the query is not updated at all before the search.

The search is done but with an empty query and when the results come back, the field is updated to the correct value (empty) because the input was blurred. The value is not retrieved from the search box because it is updated once when the enter key is pressed.

The property used in this example is `searchOnEnterKeyPressOnly`. It doesn't sound like it was meant to support the case where we use a button to trigger the search. Anyhow, the solution I see here could be to update the query value as we type even though we do not trigger the search right away.

What do you think @vvo ?

vvo:
> update the query value as we type even though we do not trigger the search right away.

Yes this should work. I think it was the case at some point.

vvo:
Closing in favor of searchBox improvements #1060

vvo:
There's still a bug, updating the issue description

PR #2180, opened 2017-06-02
fix(searchbox): SearchBox should not change the state if SearchOnEnterKeyPress

### LINKED ISSUE #2105: When using `searchOnEnterKeyPressOnly` the URL is still updated at every keystroke

bobylito:
This patch revisits the previous solution to prevent impact on the URL Sync mechanism. It uses the `change` event again and impacts only the behaviour of the SearchBox.

The solution also contains a refactoring that introduces a default value for the queryHook which reduces the control flow complexity.

Staging: http://opposite-girls.surge.sh/

Elements that you can check in the staging website:
 - see that the use of the second search box does not trigger an update of the URL
 - changing the filters creates new history steps, and you can go back and forth using the browser buttons

fix #2105 #2117


vvo:
I had a quick look and search and found this: https://github.com/algolia/instantsearch.js/issues/1015

Is it still relevant?

The searchbox code is brittle because we tried to make it a bit of a component and a bit of a connector early on (lot of cases/customization), any change to it could break other things sadly..

Let me know what you think.

src/widgets/search-box/search-box.js
```diff
@@ -37,6 +37,10 @@ const KEY_SUPPRESS = 8;
  * @return {Object}
  */
 
+const defaultQueryHook = function (query, setQueryAndSearch) {
+  setQueryAndSearch(query);
+};
+
```
iam4x: We use else where named functions, I think should try to enforce to use them when we do not need arrow functions:

```js
function defaultQueryHook(query, setQueryAndSearch) {
 ...
}
```

src/widgets/search-box/search-box.js
```diff
@@ -186,18 +185,12 @@ function searchBox({
         // handle IE8 weirdness where BACKSPACE key will not trigger an input change..
         // can be removed as soon as we remove support for it
         if (INPUT_EVENT === 'propertychange' || window.attachEvent) {
-          addListener(input, 'keyup', ifKey(KEY_SUPPRESS, getInputValueAndCall(setQuery)));
           addListener(input, 'keyup', ifKey(KEY_SUPPRESS, getInputValueAndCall(maybeSearch)));
```
iam4x: Reading the comment saying that we will remove this when we drop IE8 support, isn't this already the case?

iam4x:
Works like a charm 👍 

I dont think it adresses the same issue @vvo from what I've read, the state isn't updated until the search is performed. So it won't still be able to perform a search from an external button without "hack" and getting the value from the input with JavaScript. Did I understood it correctly @bobylito ?

src/widgets/search-box/search-box.js
bobylito: I like named functions too. Let's enforce that!

bobylito:
> I had a quick look and search and found this: #1015

It is related. 

> So it won't still be able to perform a search from an external button without "hack" and getting the value from the input with JavaScript

Yes indeed. It's actually a regression from the #1015 issue. But then I'm wondering if we should have accepted that in the first place. Yes it's hacky when you want to be able to search using a button. But on the other hand it's inconsistent to update the url without searching. Since we update the URL based on the search state, and using the `search` event breaks the back button, I still think this sounds like a valid option.

WDYT? @vvo @iam4x

src/widgets/search-box/__tests__/search-box-test.js
```diff
@@ -411,8 +411,8 @@ describe('searchBox()', () => {
         simulateInputEvent('test', 'tes', widget, helper, state, container);
       });
 
-      it('updates the query', () => {
-        expect(helper.setQuery.calledOnce).toBe(true);
+      it('does not update the query in the state', () => {
+        expect(helper.setQuery.calledOnce).toBe(false);
```
bobylito: Should be based on callCount instead of callOnce to enforce that it's actually never called (and call count has the nice effect of comparing numbers)

src/widgets/search-box/search-box.js
bobylito: It's a v1 update and we are still trying to be ie8 compatible on this version.

src/widgets/search-box/search-box.js
vvo: I like named functions too but not sure when they are assigned to a variable, does it bring any difference at runtime? It should not anymore.

```js
var superFn = function(){}
var superFn = function superFn(){}
```

bobylito: Yes, you see the correct name in the call stack in the debugger :)

vvo:
> WDYT? @vvo @iam4x

We need to be careful, for people using /1/ or magento/wordpress when they update the v1 branch this might break websites I believe.. (since this was requested at some point) I have no particular solution for this.

bobylito:
> We need to be careful, for people using /1/ or magento/wordpress when they update the v1 branch this might break websites I believe.. (since this was requested at some point) I have no particular solution for this.

Yes that's a very good point. We can just say it won't be fixed in v1 because that's a breaking change and adapt this patch in v2.

I should have been more careful in the first place, and not propose a patch for a behaviour that we patched on purpose.

vvo:
> Yes that's a very good point. We can just say it won't be fixed in v1 because that's a breaking change and adapt this patch in v2.

That should do it if there's not so much people impacted by the current bug.

bobylito:
So I'm closing this one and I will open another one that revert back the behavior of the URL sync.

PR #2181, opened 2017-06-05
chore(engine): specify a non breaking node engine

bobylito:
The new node stable version is not compatible with some of our tools that we use for generating doc. Let's keep the one that works.


package.json
```diff
@@ -33,6 +33,9 @@
     "dist",
     "dist-es5-module"
   ],
+  "engines": {
+    "node": "7.10.0"
```
vvo: Sorry I mislead you in adding this line, it should not be here because it will then produce unnecessary warnings when installed: https://docs.npmjs.com/files/package.json#engines

You can safely remove it. The only time this is useful is when you want your dev environment (for devs of the lib), to be tied to a specific nodejs version (this is useful)

PR #2190, opened 2017-06-12
fix(searchFunction): implement another way to express #1924 API

### LINKED ISSUE #2176: v2: AlgoliaSearchHelper events not handled from widget methods
*bug*

Listening for events (`helper.on`) within a widget's `init` method does not work. Worked in v1.

- jsFiddle with 2.0.0.beta.5: https://jsfiddle.net/timkelty/ufaycett
- same jsFiddle, but with 1.x latest: https://jsfiddle.net/timkelty/4s70qkq0/

2.0.0.beta.5

bobylito:
The new approach is less subtle than the previous one: instead of
overloading the search method using the helper as a prototype, we change
the implementation of the search on the fly.

Fixes #2176

**Summary**

We couldn't bind the events `search` and `result` in the previous implementation

**Result**

 - go to http://debonair-current.surge.sh/
 - open the console and see the different log message created by event listeners :)

### DIFF
--- a/src/lib/url-sync.js
+++ b/src/lib/url-sync.js
@@ -88,7 +88,8 @@ function getFullURL(relative) {
 // IE <= 11 has no location.origin or buggy
 function getLocationOrigin() {
   // eslint-disable-next-line max-len
-  return `${window.location.protocol}//${window.location.hostname}${window.location.port ? `:${window.location.port}` : ''}`;
+  const port = window.location.port ? `:${window.location.port}` : '';
+  return `${window.location.protocol}//${window.location.hostname}${port}`;
 }
 
 // see InstantSearch.js file for urlSync options
@@ -126,7 +127,10 @@ class URLSync {
     if (this.firstRender) {
       this.firstRender = false;
       this.onHistoryChange(this.onPopState.bind(this, helper));
-      helper.on('search', state => this.renderURLFromState(state));
+      helper.on('change', state => {
+        console.log('updating url');
+        this.renderURLFromState(state);
+      });
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/url-sync.js
```diff
@@ -126,7 +127,10 @@ class URLSync {
     if (this.firstRender) {
       this.firstRender = false;
       this.onHistoryChange(this.onPopState.bind(this, helper));
-      helper.on('search', state => this.renderURLFromState(state));
+      helper.on('change', state => {
+        console.log('updating url');
```
bobylito: This should be removed.

Haroenv:
to make it clickable: #1924

bobylito:
> Read and re-read again I did not find any obvious flaw that will break.

Good to know. Given the current context, which is modifying the parameters before sending the request, it will be fine indeed. But what you introduce some async in there :) Like for example you do a first query to check another index, and then you trigger the search. The code will go on and change back the reference to the search function. In this case we might end up in a loopy situation :)

Knowing the caveat, we are able to document that though. WDYT? @vvo

vvo:
Now that I think about it, is it still doable to implement debouncing with this? If not that's fine if we provide a way to do debouncing :D

bobylito:
Well it's possible as long as you are aware that the ref to the method will change and that you have to bind the reference to the search method yourself.

A very naive example of delay:

```js
  const search = helper.search.bind(helper);
  setTimeout(function(){
    search();
  }, 10);
}
```

This would work. What about we provide the search method as a standalone function along the helper? (this is one of the point I want to discuss Thursday)

bobylito:
Not good enough!

Issue #2193, opened 2017-06-14
[v2] Revamp of default styles


iam4x:
We want to enhance default styles with a better theme similar to react-instantsearch.

iam4x:
Merged in develop, published in v2.0.0-rc1 🎉

PR #2194, opened 2017-06-15
use dev-novel in dev environment

iam4x:
**Summary**

I really liked the developper experience while using storybook for react-instantsearch. Too bad storybook is sticked to a React environment.

To achieve a similar experience, I used https://github.com/algolia/dev-novel (still private to Algolians) which will enable us to test independently the result of every widgets in a much easier way.

I'll add more use-cases of the widgets with default `SearchParameters` while doing #2193 

**Result**

jun-15-2017 12-39-04

### DIFF
--- a/dev/app/index.js
+++ b/dev/app/index.js
@@ -1,40 +1,42 @@
 /* eslint-disable import/default */
+import {registerInitializer, registerDisposer, start} from 'dev-novel';
+
 import instantsearch from '../../index.js';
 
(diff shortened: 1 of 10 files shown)


Haroenv:
Is this now deployed somewhere with the documentation or still dev-only? I love it!

bobylito:
Only dev for the moment. We'll deploy it for sure :D

PR #2196, opened 2017-06-16
Make search function provide a better API

### LINKED ISSUE #1924: Make search function provide a better API

bobylito:
This implementation offers:
 - a full helper to modify the parameters before sending the request
 - all the other search function (searchOnce, searchForFacetValues)
 - compatibility with async code (no oops effect)

*warning*: This also contains a modification of the slider to be extra sure that we'll never update the state unnecessarily https://github.com/algolia/instantsearch.js/commit/8b7a6eb60fe879c0fa9841931dffbe2d91d1b24b

Fixes #1924 

 - [Deployed version with 2sec lag of all search requests](http://flashy-art.surge.sh/): test the async
 - [Nominal version](http://bite-sized-birth.surge.sh/): just a simple search function, this also serve to test the perf of the search function mechanics
 - [without any search function implementation](http://busy-snail.surge.sh/): let you test the slider modification by itself

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -128,9 +128,9 @@ Usage: instantsearch({
             addAlgoliaAgent: () => {},
             search: () => {},
           },
-          helper.state.index
+          helper.state.index,
+          helper.state
         );
-        helperSearchFunction.overrideStateWithoutTriggeringChangeEvent(helper.state);
         helperSearchFunction.once('search', state => {
           helper.overrideStateWithoutTriggeringChangeEvent(state);
           this._searchToAlgolia();
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -121,19 +121,27 @@ Usage: instantsearch({
     );
 
     if (this._searchFunction) {
-      this.helper = Object.create(helper);
-      this.helper.search = () => {
-        helper.setState(this.helper.state);
-        this._searchFunction(helper);
+      this._searchToAlgolia = helper.search.bind(helper);
+      helper.search = () => {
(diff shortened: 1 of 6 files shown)


src/lib/InstantSearch.js
```diff
@@ -121,19 +121,27 @@ Usage: instantsearch({
     );
 
     if (this._searchFunction) {
-      this.helper = Object.create(helper);
-      this.helper.search = () => {
-        helper.setState(this.helper.state);
-        this._searchFunction(helper);
+      this._searchToAlgolia = helper.search.bind(helper);
```
vvo: Now that I read it, I feel like we could name this "this._mainHelperSearch" rather than searchToAlgolia.

bobylito:
@vvo changed the name :)

iam4x:
When you hit the min boundaries of the price range and toggling some other refinements the min value keep switching. I am not sure this is a bug or is related to this PR:

jun-19-2017 16-18-17

bobylito:
> I am not sure this is a bug or is related to this PR:

That's the expected behavior. If you hit one of the boundary it's equivalent to say that you don't want to select one.

bobylito:
Need to fix the test :/

PR #2198, opened 2017-06-16
chore(packaging): do not add engine instruction to package.json

vvo:
This then warns user installing via npm about their engine not being compatible.


bobylito:
If we don't do that, tests will fail :/

Issue #2200, opened 2017-06-16
SearchBox improvements


bobylito:
Search should implement an easily identifiable input with the ability to clear the query.

Changes needed:
 - follow the same patterns as in http://shipow.github.io/searchbox/
 - provide a clear function in the connector
 - have templates for the magnifying glass (identifying the search field) and the cross (button to clear the query)

[Example on react-instantsearch](https://community.algolia.com/react-instantsearch/storybook/?selectedKind=SearchBox&selectedStory=default&full=0&down=1&left=1&panelRight=1&downPanel=storybooks%2Fstorybook-addon-knobs)

iam4x:
Merged in develop, published in v2.0.0-rc1 🎉

PR #2201, opened 2017-06-19
Searchbox improvements

iam4x:
**Summary**

Linked to #2200 

* [x] Reset button
* [x] Magnifier
* [x] Docs (getting started)

**Result**

Staging URL to see the SearchBox in action: http://obsequious-stomach.surge.sh/?selectedStory=SearchBox.default

image
(Without instantsearch-algolia-theme.css)

image
(With instantsearch-algolia-theme.css)


iam4x:
Staging URL to see the SearchBox in action: http://obsequious-stomach.surge.sh/?selectedStory=SearchBox.default

src/widgets/search-box/search-box.js
```diff
@@ -287,6 +320,59 @@ function addDefaultAttributesToInput(placeholder, input, query, cssClasses) {
...
+  const resetCSSClasses = {root: cx(bem('reset'), reset.cssClasses.root)};
+  const templateData = {cssClasses: resetCSSClasses};
+
+  let stringNode;
+
+  if (isString(resetTemplate)) {
+    stringNode = Hogan.compile(resetTemplate).render(templateData);
+  }
+
+  if (isFunction(resetTemplate)) {
```
bobylito: Maybe use `else if` so it doesn't evaluate if the `resetTemplate` is a string?

src/widgets/search-box/search-box.js
```diff
@@ -287,6 +320,59 @@ function addDefaultAttributesToInput(placeholder, input, query, cssClasses) {
...
+    stringNode = resetTemplate(templateData);
+  }
+
+  const htmlNode = createNodeFromString(stringNode);
```
bobylito: What happens if `stringNode` is `undefined`? Maybe we could throw if `resetTemplate` is neither a string nor a function?

src/widgets/search-box/__tests__/search-box-test.js
```diff
@@ -24,10 +24,17 @@ describe('searchBox()', () => {
       query: '',
     };
     helper = {
+      setState(nextState) {
```
bobylito: 🤦‍♂️ We should not reimplement the helper in the tests. Let's take care of that in the next iteration :)

src/connectors/search-box/connectSearchBox.js
```diff
@@ -75,8 +77,22 @@ export default function connectSearchBox(renderFn) {
   return (widgetParams = {}) => {
     const {queryHook} = widgetParams;
 
+    function clear(helper) {
+      return function() {
+        helper
+          .setState(helper.state.setQuery(''))
```
iam4x: Nope, I've used the same code as https://github.com/algolia/instantsearch.js/blob/feat/instantsearch.js/v2/src/lib/utils.js#L222-L224

src/widgets/search-box/__tests__/search-box-test.js
iam4x: It acts like it was mocked module, I don't think it's an anti-pattern?

src/widgets/search-box/search-box.js
iam4x: We can add more checking for sure, followed the existing code. Will add something more robust, but since we document the options it should not be a problem if it's throw?

src/widgets/search-box/search-box.js
```diff
@@ -25,8 +25,11 @@ const renderer = ({
   autofocus,
   searchOnEnterKeyPressOnly,
   wrapInput,
+  reset,
+  magnifier,
```
iam4x: I used the same naming as `poweredBy` option which accepts a boolean / object of template + css classes.

Using `showXXX` indicate me cannot be more than a boolean.

src/connectors/search-box/connectSearchBox.js
bobylito: You should, the code you linked is used to prevent multiple change event.

src/connectors/search-box/__tests__/connectSearchBox-test.js
```diff
@@ -103,6 +103,49 @@ describe('connectSearchBox', () => {
...
+      helper,
+      createURL: () => '#',
+    });
+
+    { // Second rendering
+      expect(helper.state.query).toBe('bip');
+      const renderOptions = rendering.lastCall.args[0];
+      const {clear, query} = renderOptions;
+      expect(query).toBe('bip');
+      clear();
```
iam4x: Refine with a new query after a clear? Or test that `helper.search()` is called?

src/widgets/search-box/__tests__/search-box-test.js
bobylito: We can discuss about that :)

src/connectors/search-box/__tests__/connectSearchBox-test.js
```diff
@@ -103,6 +103,49 @@ describe('connectSearchBox', () => {
...
+      state: helper.state,
+      createURL: () => '#',
+      onHistoryChange: () => {},
+    });
+
+    { // first rendering
+      expect(helper.state.query).toBe('');
+      const renderOptions = rendering.lastCall.args[0];
+      const {refine} = renderOptions;
+      refine('bip');
```
iam4x: We can, but is it a real use case?

src/widgets/search-box/search-box.js
bobylito: ok.

src/widgets/search-box/search-box.js
```diff
@@ -321,11 +407,14 @@ function addPoweredBy(input, poweredBy, templates) {
...
+
+// Crossbrowser way to create a DOM node from a string. We wrap in
+// a `span` to make sure we have one and only one node.
+function createNodeFromString(stringNode) {
   const tmpNode = document.createElement('div');
   tmpNode.innerHTML = `<span>${stringNode.trim()}</span>`;
-  const htmlNode = tmpNode.firstChild;
-
-  input.parentNode.insertBefore(htmlNode, input.nextSibling);
+  return tmpNode.firstChild;
```
bobylito: `tmpNode.firstChild.firstChild`?

src/connectors/search-box/__tests__/connectSearchBox-test.js
iam4x: After discussion, no need to re-test refine() as it's already tested 👍

src/widgets/search-box/search-box.js
iam4x: Let's throw if the `stringNode` isnt a string 👍

iam4x:
We should be all good now @bobylito 👍

PR #2202, opened 2017-06-19
Fix numeric refinementlist page

### LINKED ISSUE #2195: InfiniteHits continues to add hits when updating numeric refinement list
From SO, here is the content of the message:

----

Currently using the infinite search method that Algolia supplies through their instantSearch.js.

The following scenario is happening:

 1. Search for an item
 2. Refine the list
 3. Load more pages up to page 5
 4. Scroll back to the top and change refinement
 5. Nothing happens

By the looks of it - it's appending the results of the new refinement to the results that are already there. I'm looking to make it resent the results, not sure if this is a bug with instant search itself?

```js
       search.addWidget(
          instantsearch.widgets.numericRefinementList({
            container: '#price',
            attributeName: 'salePrice',
            options: [
              {name: 'All'},
              {end: 20, name: 'less than 20'},
              {end: 50, name: 'less than 50'},
              {start: 50, end: 100, name: 'between 50 and 100'},
              {start: 100, end: 300, name: 'Expensive'},
              {start: 300, name: 'Very Expensive'}
            ],
            templates: {
              header: 'Price'
            }
          })
        );
```
And the infinite search code:

```
    search.addWidget(
      instantsearch.widgets.infiniteHits({
        container: '#infinite-hits-container',
        templates: {
          empty: 'No results',
          item: hitTemplate
        },
        hitsPerPage: 3
      })
    );
```

 ---- 

I was able to reproduce the issue in this jsFiddle: https://jsfiddle.net/bobylito/cbj43209/

iam4x:
Fixes #2195 

The numericRefinementList was not reseting the page when refining, so the infiniteHits could not know it was new results from the first page.

### DIFF
--- a/src/connectors/numeric-refinement-list/__tests__/connectNumericRefinementList-test.js
+++ b/src/connectors/numeric-refinement-list/__tests__/connectNumericRefinementList-test.js
@@ -328,4 +328,40 @@ describe('connectNumericRefinementList', () => {
     const secondRenderingOptions = rendering.mock.calls[1][0];
     expect(secondRenderingOptions.items[0].isRefined).toBe(true);
   });
+
+  it('should reset page on refine()', () => {
+    const rendering = jest.fn();
+    const makeWidget = connectNumericRefinementList(rendering);
+
+    const widget = makeWidget({
+      attributeName: 'numerics',
+      options: [
+        {name: 'below 10', end: 10},
+        {name: '10 - 20', start: 10, end: 20},
+        {name: 'more than 20', start: 20},
+        {name: '42', start: 42, end: 42},
+        {name: 'void'},
+      ],
+    });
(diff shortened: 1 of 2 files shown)


iam4x:
This should be backported to v1 since the bug is present https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/numeric-refinement-list/numeric-refinement-list.js#L158-L205

vvo:
> 36 lines of test for 2 (with a blank line) of fix. GG 👍

Should we export the refinement methods and only test them to avoid having to set a complex fixture to do a one line change?

As soon as we have a stable codebase like now (where we know we will have those refine methods for long), we could test those refine methods by units rather than having to render a full widget.

(thinking out loud)

PR #2207, opened 2017-06-22
Update design page

LukyVj:
**Summary**

**Result**


iam4x:
Awesome work @LukyVj it looks so good! 👏

I have some remarks:

image 
It looks like the whole button should be clickable?

 image 
Did you tried with a white hover text color?


jun-22-2017 19-25-59 
The left side menu has a scrollbar inside itself and does not take the full height space of the page

---

* We lost the syntax color into code examples.
* The get started CTA is only the third one, I expect from a documentation website to be almost the first thing visible to encourage users to start coding!
* I think the get started link does not fit with the no framework needed block but more with the widgets one.

And again thank's a it's so much better <3
If you need any help for anything just ask!

docgen/layouts/common/meta.pug
```diff
@@ -33,4 +33,5 @@ html
     title #{project} | #{title}
     link(rel='stylesheet', href="stylesheets/style.css")
     link(rel="stylesheet", href="https://cdn.jsdelivr.net/docsearch.js/2/docsearch.min.css")
-    link(rel="stylesheet", href="//cdn.jsdelivr.net/npm/instantsearch.js@2.0.0-beta.1/dist/instantsearch.min.css")
+    link(rel="stylesheet", href="//cdn.jsdelivr.net/npm/instantsearch.js@2.0.0-RC.1/dist/instantsearch.min.css")
+    link(rel="stylesheet", href="//cdn.jsdelivr.net/npm/instantsearch.js@2.0.0-RC.1/dist/instantsearch-theme.min.css")
```
iam4x: Nice I missed that 👍

bobylito:
> The get started CTA is only the third one, I expect from a documentation website to be almost the first thing visible to encourage users to start coding!

This one is on me. We'll iterate :)

docgen/assets/img/InstantSearch-Android.svg
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="245px" height="51px" viewBox="0 0 245 51" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
Haroenv: These should be run through svgo or svgomg

bobylito:
Zip for the examples are not accessible.

LukyVj:
Hey guys, thanks for all the comments, I'll take care of everything but for today I'm switching on the meetup video edit! Thx

vvo:
This looks gorgeous, small comment on: 
image

I cannot understand the design of this button easily, so I don't get what this button does.

I like what we did on react-instantsearch, straight to the point 
image. Maybe we could have something similar?

auareyoualgolia:
Hello! 👋

@vvo @bobylito @LukyVj 

I used this icon because it's something I'm used to seeing around to web for both 'copy to clipboard' and 'duplicate' 

I usually go to thenounproject to make sure is not something subjective: 

image

In my opinion, if there's an icon for a simple action like this one it's better to use that than text as it's easier to process quickly and it separates better from the written word which in this case is abundant. However, I can't speak for a user of documentation, if you guys think it's still better to go with text I don't think it's a 'design breaker' 🤷‍♀️

vvo:
Indeed as a dev it was not obvious what this button does, it looks very similar to open in a new window: 
image, from https://thenounproject.com/search/?q=popup%20window&i=539930

We could either have a nice "copy" (no need for copy to clipboard) button, much like we have "edit on github". If we want to go for an icon, that one is more often seen in documentations: 
image, from https://clipboardjs.com/assets/images/clippy.svg

bobylito:
Let's go for `copy` then. It can be revised later.

LukyVj:
So, I arrive a bit after the war, but I don't see the problem of having an icon instead of text. 
Plus, we do it on the dashboard for our customers, most of them are regular profiles I guess ( not so dev in some cases ) and we still show them this icon. 

I'm not strongly for having an icon instead of text, but since we do it on algolia.com, I don't see why it would be a problem on a community website.

bobylito:
So if I sum up:
 - this icon can be confusing (@vvo didn't get the meaning of this icon)
 - icons are better, especially for such icons. But needs to be easily recognizable
 - if something is more universal than the other, let's choose the more universal

@auareyoualgolia do you have another proposal of icon that would be more universal?

vvo:
The one on algolia.com works better to me

LukyVj:
I agree with you @vvo, it's true that the one on the design is more for "open in a new page" but in the end, I think we can find a better one (reusing the one of the dashboard?) and it will bring design consistency across our different pages. 

And as Sebastien says, good design = consistent design = trust

auareyoualgolia:
I think reusing the one from the dashboard would make sense too. I thought it a slightly different use but it is true it's much more clear

LukyVj:
Heyo

So I fixed all the above listed bugs/missing stuff

Do you have a list of what's missing?

PR #2208, opened 2017-06-22
chore(build): Add release script for maintenance

### LINKED ISSUE #2203: Release script for previous version
The v1 now lives in maintenance therefore we can't release new version with the automatic script anymore.

We need to be able to:
 - publish new versions (for fixes)
 - update the documentation

bobylito:
Fixes #2203

This adds a new script to release the maintenance version. It is based on the previous release.sh.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,7 @@ will be done on the maintenance version.
 To run this project, you will need:
 
 - Node.js >= v7.10.0 and <= 8, use nvm - [install instructions](https://github.com/creationix/nvm#install-script)
+- yarn https://yarnpkg.com/en/
 
 ## Development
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -67,8 +68,19 @@ Files are automatically formatted with prettier.
...
+yarn run release
+```
+
+### Maintenance version
+
+For the maintenance version, go on maintenance (`git checkout maintenance`) and use:
+
 ```sh
-npm run release
+yarn run release-maintenance
```
Haroenv: The `run` can be left out

bobylito: but it works too ;)

vvo: Late to the party: I never managed to run a scrip with yarn that will use "npm publish" at some point. I got credentials not being read for publishing to npm. I believe this is still an issue (and that's why previous command was npm run release).

To me that's one of yarn drawbacks that could make us switch back to npm (because if we need to use both yarn and npm, it's a struggle)

Issue #2168, opened 2017-05-30
Include ES6 modules in npm dist


timkelty:
**Do you want to request a *feature* or report a *bug*?**
*feature*

**Feature: What is your use case for such a feature?**
If you are already compiling with Babel/Webpack, you may want to import instantsearch's ES6 modules directly. Currently included are the compiled webpack output and the es5 modules.

**What is the version you are using? Always use the latest one before opening a bug issue.**
2.0.0.beta 4

Haroenv:
Yes, but usually the code itself should still be transformed to lowest common denominator, while only the imports/exports are kept es6. This was already the roadmap after 2 is out of beta.

iam4x:
We will not provide the source files as it on npm, what we do is:

* Providing an UMD build to include in a webpage through a `<script>` tag (/dist) on npm
* Providing ES6 to ES5 files already transpiled through babel

You can easily reduce your final build size by importing the files you only need:

```js
import instantsearch from 'instantsearch.js/dist/instantsearch'
``` 

Results to a build of ~1.14mb (not minified/uglified), but when importing that way:

```js
import instantsearch from 'instantsearch.js/dist-es5-module/src/lib/main';

// you can combine with specific parts
// for instance when using only connectors to build your app
import * as connectors from 'instantsearch.js/dist-es5-module/src/connectors';
```

These two imports (without using widgets in your code but only connectors) results to a ~175kb (not minified/uglified).

Right now this a little bit complicated and not enough documented, the roadmap for the v2.1 is to simplify and update the building tools so you will be abble to import that way:

```js
import instantsearch, { connectors, widgets } from 'instantsearch.js'
```

Can you try to import from `dist-es5-modules` and tell me the result?

timkelty:
> Can you try to import from dist-es5-modules and tell me the result?

@iam4x yep, that's what I'm doing right now, and it works. I just need to exclude it from my `babel-loader`, or I get errors.

However, as I'm already compiling with webpack/babel, ideally I still want to import the ES6 modules. Why? For one, [tree shaking](http://jakewiesler.com/tree-shaking-es6-modules-in-webpack-2/).

iam4x:
It's recommended to exclude any libraries from `node_modules` into babel-loader comfiguration, this is an expected behaviour.

And yes when transpiling from ES6 to ES6 with `babel-present-env` and `modules: false` like @Haroenv said only the imports/exports are kept es6.

timkelty:
Thanks!

Issue #1866, opened 2017-01-18
Update the build stack


vvo:

vvo:
Done on react-instantsearch, to be done on instantsearch.js

bobylito:
Webpack 2 or rollup. Seems like rollup is more fit to libraries cc @Haroenv @rayrutjes

Haroenv:
my vote goes to rollup here

vvo:
From the last experimentations of @rayrutjes, we should use rollup for libraries when feasible (= no heavy webpack loaders we depend on).

bobylito:
Well we did some upgrade in v2. Let's revisit that later. Closing for now.

PR #2218, opened 2017-07-03
Upgrade building tools

iam4x:
This is still work in progress, opening the PR so we can track the progress.

Linked issues #2168 #1866 

## Summary

The build mechanism has not been updated with the v2. Now that version is live, we will upgrade the building tools in order to have the most productive development environment 🚀

I did an experiment by replacing `node-sass` with `postcss` but it's too much overhead, we will sync about this when we want to provide a single shared theme in between instantsearch libraries.

I also did an experiment with rollup instead of webpack, but there's no real advantages of doing this. It adds more dependancies and now webpack3 has reached the state of rollup, maybe goes further with [scope hoisting](https://medium.com/webpack/brief-introduction-to-scope-hoisting-in-webpack-8435084c171f) 🌟

* [x] Upgrade dev & staging build to use webpack v3
* [x] `$ yarn dev` and `$ yarn staging` shares now the same `dev/webpack.config.js`
* [x] Remove the `.babel.js` from webpack config to load faster since we use now nodejs7+ we can use ES6 features without transpiling through babel
* [x] Upgrade docgen to use webpack3
* [x] Upgrade final build to use `babel` CLI for ES6 modules
* [x] Use `babel-preset-env` with `modules: false` (Tree shaking)
* [x] Upgrade final build to use webpack3 for browsers builds
* [x] Simplify and clean every `scripts/*.sh` (since we use almost webpack to do all the stuff, we can have one liner commands into `package.json` scripts)
* [x] Document usage with rollup/wepback2+ for tree shaking

More detailed changes for the dev env:

* use of webpack [style-loader](https://github.com/webpack-contrib/style-loader) + [sass-loader](https://github.com/webpack-contrib/sass-loader) instead of building and watching with `node-sass` CLI
     -> This introduce a speed gain as well of having the "hot-reload" of styles
* use of [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin) to generate an `index.html` with hashed assets
     -> Cache buster when deploying to staging
* use of [autodll-webpack-plugin](https://github.com/asfktz/autodll-webpack-plugin) which build vendors seperatly
     -> Speed up re-build on dev (big external libraries not re-builded)
* use of [happypack](https://github.com/amireh/happypack) which allow to transform multiple files in parallel
      -> Speed up initial build and re-build

## Results

**Staging build** is 50% faster (removed the minification on new branch to have a fair comparison):

```
BEFORE:
➔ time (node-sass -o ./dev/ ./src/css --output-style expanded && webpack --config dev/webpack.staging.config.babel.js --hot --inline --no-info)

~12s

AFTER:
➔ time NODE_ENV=production webpack --config dev/webpack.config.js

~6s
```

**Production build** stays quite the same, no build size changes but everything is bundled by webpack.


dev/webpack.config.js
```diff
@@ -0,0 +1,161 @@
...
+
+  output: {
+    path: path.join(__dirname, 'dist'),
+    filename: '[name].[hash].js',
+    library: '[name]',
+    libraryTarget: 'umd',
+  },
+
+  module: {
+    rules: clean([
```
bobylito: `clean` is not explicit. Why not just filter on a new property that defines if it's dev instead of having `object|false`?

dev/webpack.config.js
```diff
@@ -0,0 +1,161 @@
...
+const ExtractTextPlugin = require('extract-text-webpack-plugin');
+const HtmlWebpackPluging = require('html-webpack-plugin');
+const BabiliPlugin = require('babili-webpack-plugin');
+const AutoDllPlugin = require('autodll-webpack-plugin');
+const HappyPack = require('happypack');
+
+const __DEV__ = process.env.NODE_ENV === 'development';
+const clean = arr => arr.filter(item => item !== false);
+
+module.exports = {
```
bobylito: imho defining three configurations: `common`, `dev`, `prod` is more legible and easy to follow.

vvo:
Is there any feature added for package consumers? (Out of curiosity).
What's the dev reload time before/after?

dev/webpack.config.js
iam4x: I was using different configurations before and merging them, but it was too complicated at some point. 

When you have different depth level of objects and arrays it's just a mess to have something like that to maintain. You are lacking of context, must have multiple files open and since this is only configuration I don't think we need to separate that much the concerns as we do in our codebase.

For instance overriding some options and adding new ones from a common configuration:

prod.webpack.config.js
```js
const commonConfig = require('./common.config.webpack.js');
const prodConfig = Object.assign({}, commonConfig);

prodConfig.entry = {
  bundle: './dev/app'
};

// change the js loader config
commonConfig.module.rules[3].use[1].options = {sourceMap: false};

// remove happypack
commonConfig.plugins.splice(3, 1);

// add minify
commonConfig.plugins.push(new BabiliPlugin());
```

I have also tried the "duplication" strategy like `create-react-app` does. It is also too complicated to maintain when you update the configuration you have the replicate all the changes in between configurations.

If we really want to separate configurations, we should look https://github.com/Fitbit/webpack-config

iam4x:
@vvo 

I will come with real stats in the end, but it's already a big win to have a single webpack handling all the stuff instead of having two CLI tool running for the dev environment (webpack & node-sass) separately.

It might not gain speed in term of bundling, but the end experience of hot reload in the browser is completely worth it while working on styles.

Features for the end users:

* we will ship real ES6 modules so it will be easier for them to have tree-shaking
* webpack3 scope hoisting should speed up the load of the library itself in the browser
* lower the build size of CSS & JS (not sure, investigation at the end)

dev/webpack.config.js
```diff
@@ -0,0 +1,160 @@
...
+  devtool: __DEV__ ? 'cheap-eval-source-map' : 'cheap-module-source-map',
```
vvo: In production why not use source-map the safest one? Performance should not matter that much here.

dev/webpack.config.js
iam4x: We can rename `clean()` to something else, I wasn't inspired when doing it. 

I really like having this syntax: `__DEV__ && ...` it's so explicit that plugin or option is meant for development environment or not. 

We should not bother on how clean is the code for configurations files but on how simple is it to read and understand.

dev/webpack.config.js
vvo: Why not do something in between where you split into different categories:

```js
const bundle = __DEV__ ? ...
const rules = [...];
if (__DEV__) {rules.push()...}
const plugins = [...];
if (__DEV__) plugins.push(autoreload);

module.exports = {...};
```

Not ideal but might ease the reading. I was also in favor for production/dev configs but reading this one, it's not that hard to read it (I am used to webpack still)

iam4x:
@vvo I did a little comparison of build time for staging, dropped from 12s to 6s 💪

dev/webpack.config.js
bobylito: Like everyone I'm a creature of habits. Let's try this syntax :)

iam4x: What about `resolveEnv()` instead of `clean()` ?

dev/webpack.config.js
iam4x: I'm good with `source-map` as well, this is only for staging and not for the final library build 👍

bobylito: I do a lot of debugging in chrome, and a good source map for dev is a requirement :) Can we do so that we have a mode in which this is a reality? (maybe `cheap-eval-source-map` is good enough I need to try, but if it's not it's a pain :D)

vvo:
> I did a little comparison of build time for staging, dropped from 12s to 6s

What's staging (vs dev and prod)? What's the time for the actual dev?

iam4x:
* dev is build of library itself + stories + documentation
* staging is a static build of the dev without documentation
* prod is browser and modules build of the library itself

Will add more metrics when I'll be done with upgrading the documentation tools 👍

iam4x:
I can't upgrade easily the docgen build, `ms-webpack` is explicitly requiring webpack v1 (https://github.com/10xjs/ms-webpack/blob/master/package.json#L21) which is altering the configuration on the fly and make it incompatible with webpack 3 😩

So, I'm removing upgrade of docgen build from the scope of this PR.

dev/webpack.config.js
iam4x: * with `source-map` you get the line and the col of the exception
* with `cheap-module-source-map` you only get the line

vvo:
@iam4x ms-webpack is like 100 lines. Ditch it clearly :)

iam4x:
@vvo Thanks with `ms-webpack` alpha.1 which does not require a specific webpack version it works like a charm 🙌

dev/webpack.config.js
iam4x: Since it's not that hard to read and it's only configuration, do we agree we leave it that way until we experience some issues with this format?

iam4x:
We also now provide an ES6 build to be picked by webpack2+/rollup for tree shaking 🎉 

* We agreed with @vvo on an export strategy:

```js
// provides same instantsearch object as es5 or <script> build
// useful for non-breaking changes on actual apps already using webpack or rollup

import { instantsearch } from 'instantsearch.js';
// provides instantsearch object without connectors and widgets attached.
// only `version` and `createQueryString` are available

import { connectXXX } from 'instantsearch.js'
// provides every available connectors

import { searchBox, menu, refinementList, ... } from 'instantsearch.js'
// provides every available widgets
```

* `.babelrc` file has now only environments, default one picked will be development

iam4x:
@vvo @bobylito 

Some more stats using the new ES6 build with `etch-instantsearch` which is only using the connectors API:

* before: `instantsearch.js: 438.42 KB`
* after: `instantsearch.js: 160.11 KB`

I reduced the size of the instantsearch.js dependency up to **~60%** !!

----

But in order to have this level of optimisation, I had to change the export strategy.

Importing `import main from './src/lib/main.js'` into `index.es6.js` will automatically require React and all the widgets. 

Even if on your side you import instantsearch.js this way: 
```js
import {instantsearch} from 'instantsearch.js'
```

----

In order to resolve this issue, the new exporting strategy now includes breaking changes:

```js
import instantsearch, {connectMenu, searchBox} from 'instantsearch.js';

// throw an error when trying to access to `.connectors` or `.widgets`
instantsearch.connectors.connectMenu(...);
instantsearch.widgets.searchBox(...);

// does not throw
connectMenu(...);
searchBox(...);
```

The error displayed is pretty dumb for now, will add link to the docs after I updated them in this PR:
image

There is also something we did not think about with @vvo is the use-case of importing all the connectors by once which is not intuitive right now: 

```js
// require every connectors available
import * as connectors from 'instantsearch.js/dist-es6-module/src/connectors';
```

Something easier would be:

```js
import {widgets, connectors} from 'instantsearch.js';
```

What do you think?

vvo:
I am not sure to get it but:

```js

instantsearch.connectors.connectMenu(...);
instantsearch.widgets.searchBox(...);
```

Should still work, otherwise it's a breaking change (and major version). If it does not work because we tried to have import {connectSearchBox} from 'instantsearch.js'; instead of having import {connectSearchBox} from 'instantsearch.js/connectors'; Then we should do the later instead (instantsearch.js/connectors).

```js
import * as connectors from 'instantsearch.js/connectors';
```

This should work for all envs but the use case may be edgy.

iam4x:
So to comply with a non-breaking changes update and we want the tree shaking to be working, we have to separate all the "main" modules this way:

```js
// instantsearchjs object without .connectors/.widgets property
import instantsearch from 'instantsearch.js/instantsearch';

import * as widgets from 'instantsearch.js/widgets';
import * as connectors from 'instantsearch.js/connectors';
```

This will imply documentation and won't work "out of the box" for existing projects, but it's non breaking and will provide really good results when you are not importing widgets or connectors.

vvo:
> This will imply documentation and won't work "out of the box" for existing projects, but it's non breaking and will provide really good results when you are not importing widgets or connectors.

Seems fine, in another major version we could ditch the version where we allow for importing the whole instantsearch + instantsearch.widgets* ... So users (of webpack/browserify) would be forced to uniquely ask for each method they want to use. Which is fine when you think about it (react is does that, nobody complains)

iam4x:
* I've added a paragraph about tree-shaking into usage.md
* I've published a `v2.1.0-beta.1` version with tag beta on npm in order to test correctly

asfktz:
HI @iam4x,
how are you?

My name is Asaf, and I'm the author of [autodll-webpack-plugin](https://github.com/asfktz/autodll-webpack-plugin).

I'm here because I searched GitHub for projects which currently implement my plugin to learn from their use cases, and see if they encountered any issues with it.

I noticed that the Plugin's context is incorrect, it should point to the root of your project.

So instead of:

```js
new AutoDllPlugin({
    inject: true,
    context: __dirname,
    filename: '[name].[hash].js',
    entry: {vendor: Object.keys(require('../package.json').dependencies)},
})
```

It should be:

```js
new AutoDllPlugin({
    inject: true,
    context: path.join(__dirname, '..'),
    filename: '[name].[hash].js',
    entry: {vendor: Object.keys(require('../package.json').dependencies)},
})
```

Incorrect context will result in having the same modules in DLL bundle **and** in the main bundle too.

That silent error comes from webpack's own DllPlugin itself.
I plan to add warnings about it soon.

**Stats before the change:**

```
Time: 8901ms
                                Asset       Size  Chunks                    Chunk Names
       bundle.bcfcc724970cd0307929.js    5.96 MB       0  [emitted]  [big]  bundle
instantsearch.bcfcc724970cd0307929.js     4.1 MB       1  [emitted]  [big]  instantsearch
                           index.html  690 bytes          [emitted]
       vendor.df506d98aa0c53bd74ce.js    1.43 MB          [emitted]  [big]
Child html-webpack-plugin for "index.html":
         Asset   Size  Chunks  Chunk Names
    index.html  27 kB       0
```

Average Rebuild:  ~950ms


**Stats before the change:**

```
Time: 5516ms
                                Asset       Size  Chunks                    Chunk Names
       bundle.6873c9409851d5449e9d.js    3.96 MB       0  [emitted]  [big]  bundle
instantsearch.6873c9409851d5449e9d.js     2.1 MB       1  [emitted]  [big]  instantsearch
                           index.html  690 bytes          [emitted]
       vendor.df506d98aa0c53bd74ce.js    1.43 MB          [emitted]  [big]
Child html-webpack-plugin for "index.html":
         Asset   Size  Chunks  Chunk Names
    index.html  27 kB       0
```

Average Rebuild:  ~550ms

You can optimize it even more by adding the "dev-novel" module to the DLL entries too.

Average Rebuild:  ~450ms

Also, please update to version 0.0.8 as it fixes a critical bug.

Tip: add `debug: true` to the Plugin's settings to see more clearly what the plugin is doing.

Please let me know if you need any further help.
Asaf.

vvo:
Thanks @asfktz that's awesome. This PR looks almost complete. Only @iam4x should be able to say so, let's wait a bit (he's currently not available).

Once we have more news, this is a great chore AND feature

asfktz:
@vvo 👍
Let me know if you need me.

bobylito:
To keep fix easy to deploy, I propose that we reopen this PR targeting `feat/2.1` instead of `develop`. Thanks :)

iam4x:
Thank's a lot for your advices and code review @asfktz 👍

iam4x:
Closing it in favor of #2240

PR #2219, opened 2017-07-04
docs(readme): Family is InstantSearch, not InstantSearch.js

Haroenv:
Since we have projects outside of js


bobylito:
Of course 🤦‍♂️

PR #2224, opened 2017-07-08
chore(dep): update algoliasearch and algoliasearch-helper

bobylito:
Update JS Helper and algolia JS client.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,8 +105,8 @@
     ]
   },
   "dependencies": {
-    "algoliasearch": "^3.22.1",
-    "algoliasearch-helper": "^2.20.1",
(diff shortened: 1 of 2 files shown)


vvo:
> Argos test seems to detect the changes in image compression again

cc @oliviertassinari

oliviertassinari:
@vvo I have tried increasing the color threshold without any success in https://github.com/argos-ci/argos/pull/337. I might have to replace ImageMagick with a custom implementation better-supporting imperceptible changes. For now, I have accepted the diff. Another alternative would be to load deterministic jpg or even hiding them in the screenshots.

PR #2234, opened 2017-07-13
docs(readme): list supported browsers

Haroenv:
**Summary**

@jonasba tested, and IE9 and lower does not work. Older Safari, Chrome, FF versions are not checked explicitly

**Result**

add a sentence to show we support the latest two versions

PR #2235, opened 2017-07-14
test(visual): hide flaky jpg issue

oliviertassinari:
This is the temporary fix [I have been referring to](https://github.com/algolia/instantsearch.js/pull/2234#issuecomment-315109588). I need to better handle imperceptible color changes.

### DIFF
--- a/functional-tests/test.js
+++ b/functional-tests/test.js
@@ -1,6 +1,11 @@
 import expect from 'expect';
 import {searchBox, prepareScreenshot} from './utils.js';
 
+const hide = [
+  '.ais-stats--body', // Flaky X ms information.
+  '.media-object', // Flaky jpg rendering
+];
+
 describe('searchBox', () => {
   describe('when there is no query', () => {
     beforeEach(() => searchBox.clear());
@@ -10,9 +15,7 @@ describe('searchBox', () => {
     it('triggers an empty search', () => {
       expect(browser.getText('#hits')).toNotContain('MP3');
       prepareScreenshot();
-      browser.checkDocument({
-        hide: ['.ais-stats--body'], // Flaky X ms information.
-      });
+      browser.checkDocument({ hide });
     });
   });
 
@@ -24,9 +27,7 @@ describe('searchBox', () => {
     it('triggers a new search', () => {
       expect(browser.getText('#hits')).toContain('MP3');
       prepareScreenshot();
-      browser.checkDocument({
-        hide: ['.ais-stats--body'], // Flaky X ms information.
-      });
+      browser.checkDocument({ hide });
     });
   });
 });


vvo:
@oliviertassinari I think we might want to disable argos for now until you have a "good" fix for the jpg issue. Those images have no visual difference from a human eye and this is what we are looking for. We are facing the same issue on react-instantsearch, having to configure DOM selectors is painful.

Can you remind us the exact issue? Is it because we are taking screenshots as JPG instead of PNG?

oliviertassinari:
> Those images have no visual difference from a human eye.

It's not always accurate, look at this diff [base](https://s3.amazonaws.com/argos-screenshots-us/cbef82b9e9e7f1538580470506ca2074)...[compare](https://s3.amazonaws.com/argos-screenshots-us/08df1ca01f03e27ba45539c470246525), the image rendered is definitely different.

> having to configure DOM selectors is painful for an issue that should be handled at the core of your tool.

It depends on the responsibility/problem solved by the tool given the structure used. In our case, Argos-CI has the opportunity to better handle part of the flakiness, that's something we should prioritize as providing more value. But visual regression testing is hard, people will always have to do work on there side, for instance, not rendering non-deterministic data. We have a good example here, we have to hide the roundtrip time.

> Can you remind us the exact issue? Is it because we are taking screenshots as JPG instead of PNG?

I can't tell what's the source of the issue is given the information I have. It could be from the most likely to the less:
- Nondeterministic JPG image rendering, variating hardware like GPU, drivers, chrome versions
- Nondeterministic JPG image source

oliviertassinari:
> until you have a "good" fix for the jpg issue

What would be a "good" fix for you?

bobylito:
> What would be a "good" fix for you?

100% reliable on same web output.

oliviertassinari:
After some time running Argos-CI on the builds here, the only source of flakiness that we have identified is around the non-deterministic JPG rendering. With that PR, we can expect to **address 100% of the flakiness**. It comes with a tradeoff, we won't see JPG regressions, but does that matter? I can submit the same one on *react-instantsearch*.

oliviertassinari:
Do you mean that we could be rendering a png in the search result list instead of a jpg?

Haroenv:
Saving the screenshots as png instead of jpg is what Vincent suggests I think

vvo:
Maybe I am completely off and the underlying issue is that our own images are genereated on the fly constantly as JPG, which mean they change sometime.

oliviertassinari:
@Haroenv Argos-CI is already saving the images as PNG not JPG (`Content-Type:image/png`).

bobylito:
Probably the best would be to not have jpg in our dataset or just in our results...

bobylito:
I opened an other PR that removes the images. #2254

PR #2239, opened 2017-07-17
[v1 Docs] Absolute URLs for meta images

### LINKED ISSUE #2238: [V1 Doc] use absolute link for meta image
right now the link is https://community.algolia.com/instantsearch.js/assets/instantsearch-cover-6630408f1e9eacd08a4209f803be3d43.png which is declared as /instantsearch.js/assets/instantsearch-cover-6630408f1e9eacd08a4209f803be3d43.png in the meta tag, but that doesn't work in discourse for example. 

The meta tags should be changed to absolute links instead

cc @olance

olance:
**Summary**
The `og:image` and `twitter:image:src` meta tags need to provide an absolute URL to be usable from external websites linking to the page.
Today, `asset_path` is used and returns a relative path, making the thumbnail broken on some websites such as our Discourse forum.

**Result**
Using `asset_url` instead of `asset_path` for `og:image` and `twitter:image:src` meta tags will output an absolute URL instead of a relative path.

Fixes #2238

### DIFF
--- a/docs/_plugins/asset_url_tag.rb
+++ b/docs/_plugins/asset_url_tag.rb
@@ -1,0 +1,21 @@
+module Jekyll
+  class AssetUrlTag < Liquid::Tag
+    PARAMS_RE  = / ^
+                     \s*
+                     (?: "(?<path>[^"]+)" | '(?<path>[^']+)' | (?<path>[^ ]+) )
+                     (?<attrs>.*?)
+                     (?: \[(?<options>.*)\] )?
+                     \s*
+                     $
+                   /x
+
+    def render(context)
+      site = context.registers[:site]
+      match = @markup.strip.match(PARAMS_RE)
+
+      File.join(site.config['host'], site.asset_path(match['path']))
+    end
+  end
+end
+
+Liquid::Template.register_tag('asset_url', Jekyll::AssetUrlTag)
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -12,14 +12,14 @@
     <meta content="@algolia" name="twitter:site">
     <meta content="instantsearch.js" name="twitter:title">
     <meta content="Check out instantsearch.js—the new JS library of UI widgets to help you build the best instant-search experience with Algolia" name="twitter:description">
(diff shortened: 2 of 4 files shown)


iam4x:
Fixes #2238

bobylito:
@olance are you sure about `asset_url`?

iam4x:
@bobylito Nope, local and netlify build fail.

bobylito:
Let me know if you need a hand @olance

olance:
@bobylito Just looked around, and the thing is `asset_url` is available in builds of `jekyll-assets` as old as a year ago, but the gem is locked to `v1.0.0` on the project. It was not available then.

bobylito:
Can you try and update that gem?

olance:
@bobylito yes that's what I'm looking at right now :)
I'll keep you updated!

olance:
@bobylito @iam4x so in the end, even in latest versions Jekyll only provides a `asset_url` SASS helper but not a Liquid tag.
So I added my own :)

docs/_plugins/asset_url_tag.rb
```diff
@@ -0,0 +1,21 @@
+module Jekyll
+  class AssetUrlTag < Liquid::Tag
+    PARAMS_RE  = / ^
+                     \s*
+                     (?: "(?<path>[^"]+)" | '(?<path>[^']+)' | (?<path>[^ ]+) )
+                     (?<attrs>.*?)
+                     (?: \[(?<options>.*)\] )?
+                     \s*
+                     $
+                   /x
```
olance: This regexp comes from Jekyll's own code to parse params given to a Liquid tag

olance:
🙌

docs/_plugins/asset_url_tag.rb
bobylito: Ok :)

bobylito:
It's tedious to deploy, I'll take some time probably tomorrow.

iam4x:
Thank's @olance :)

olance:
my pleasure 🙃

PR #2240, opened 2017-07-17
Upgrade building tools

iam4x:
See original PR for description #2218


.babelrc
```diff
@@ -1,14 +1,43 @@
...
+        "react",
+        "stage-2"
+      ],
       "plugins": ["babel-plugin-rewire"]
     },
+    "development": {
+      "presets": [
+        ["env", { "targets": { "browsers": ["last 2 versions", "ie >= 9"] } }],
+        "react",
+        "stage-2"
```
bobylito: Why do we use stage-2 if we specify the targets?

.babelrc
```diff
@@ -1,14 +1,43 @@
...
+      ],
+      "plugins": [
+        "transform-react-remove-prop-types",
+        "transform-react-constant-elements",
+        "transform-react-pure-class-to-function"
+      ]
+    },
+    "production-es6": {
+      "presets": [
+        ["env", { "modules": false, "targets": { "browsers": ["last 2 versions", "ie >= 9"] } }],
```
bobylito: Can we be more conservative on the targets for es6 published modules?

scripts/build.sh
```diff
@@ -1,43 +1,28 @@
...
-# only transpile to ES5 for package.json main entry
-babel -q index.js -o dist-es5-module/index.js
-declare -a sources=("src")
-for source in "${sources[@]}"
-do
-  babel -q --out-dir dist-es5-module/${source} ${source}
-done
+# * allow `import {connectXXX} from 'instantsearch.js/connectors'`
+# * allow `import {searchBox} from 'instantsearch.js/widgets'`
+echo "export * from './dist-es6-module/connectors/index';" >> ./connectors.js &
```
bobylito: Why do we generate these files?

iam4x: We want them only on the npm repository, they are used to allow;

* `import {connectXXX} from 'instantsearch.js/connectors'`
* `import {searchBox} from 'instantsearch.js/widgets'`

It feels weird to have theses files published on github at the root of the project.

.babelrc
iam4x: https://github.com/babel/babel-preset-env#support-all-plugins-in-babel-that-are-considered-latest

.babelrc
iam4x: We still want to support the same browsers and we don't want the end-user to re-preprocess the instantsearch.js library files into his own build system.

scripts/build.sh
bobylito: I think it would be better if they were published because here it's like they exist but only during the build and they are hardcoded in a script which is even weirder to me.

.babelrc
bobylito: So if the features we use come from stage-2 and they are supported by the browsers of env then we should have stage-2 automatically applied, no?

.babelrc
bobylito: wfm

scripts/build.sh
Haroenv: I'm pretty sure that without this change, `instantsearch.js/connectors` would also match `instantsearch.js/connectors/`

.babelrc
Haroenv: InstantSearch doesn't work in IE9, why are we transpiling for it?

scripts/build.sh
bobylito: alright! let's keep that! (because nothing is mature anyway :P)

iam4x:
🚢

PR #2242, opened 2017-07-17
fix(search-box): fix magnifier and reset customization

bgelineau:
**Summary**

Fix the magnifier and reset options of the search-box widget:
 - the ``reset: false`` option currently breaks. The renderer tries to hide the button even when it does not exist.
 - custom templates are not taken into account for both options, despite the documentation saying otherwise.

**Result**

Fixed the problems and added the related tests.


iam4x:
Hello @bgelineau,

Thank's for your contribution, can you re-create this PR on develop branch?

iam4x:
Sorry for closing, thank's to @Haroenv I did not know we could change the target branch from github now.

PR #2243, opened 2017-07-17
fix(theme): enforce box-sizing: border-box

### LINKED ISSUE #2241: Default styling of search box has some issues

bug

Styling of the searchBox widget seems to have odd spacing/appearance: none issues

https://codepen.io/Haroenv/pen/7ea9fcf2b263fbf4c7884e04055c2d02

screen shot 2017-07-17 at 16 47 33
screen shot 2017-07-17 at 16 47 20

look exactly the same as the React InstantSearch Searchbox

chrome, safari, 2.x

fix the styling 😓

iam4x:
fixes #2241

### DIFF
--- a/src/css/theme/_base.scss
+++ b/src/css/theme/_base.scss
@@ -15,18 +15,7 @@
 @import 'stats';
 @import 'toggle';
 
-.ais-clear-all,
-.ais-current-refined-values,
-.ais-hierarchical-menu,
-.ais-menu,
-.ais-pagination,
-.ais-price-ranges,
-.ais-range-slider,
-.ais-refinement-list,
-.ais-search-box,
-.ais-star-rating,
-.ais-stats,
-.ais-toggle {
+[class^=ais-] {
   box-sizing: border-box;
 
   & > *,
--- a/src/css/theme/_base.scss
+++ b/src/css/theme/_base.scss
@@ -15,6 +15,27 @@
 @import 'stats';
 @import 'toggle';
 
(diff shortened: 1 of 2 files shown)


Haroenv:
Does the deploy preview make a build of the css so I can check that?

iam4x:
@Haroenv No, and if I deploy to staging the dev page already includes this rule.

src/css/theme/_base.scss
```diff
@@ -15,6 +15,27 @@
...
+.ais-hierarchical-menu,
+.ais-menu,
+.ais-pagination,
+.ais-price-ranges,
+.ais-range-slider,
+.ais-refinement-list,
+.ais-search-box,
+.ais-star-rating,
+.ais-stats,
+.ais-toggle {
```
Haroenv: I would prefer

```css
[class^=ais-] {
}
```

bobylito: That's more future-proof 🤖 👍

iam4x: I did not know about this CSS3 pattern matching selector. There's a lot of divs created with `.ais-` prefix, do you know what is the best in term of performances?

Haroenv: It's css 2 actually :) https://developer.mozilla.org/en/docs/Web/CSS/Attribute_selectors

I think the performance hit of this is negligible, since most sites have border box on *

iam4x:
Made the changes to use class pattern matching 👍

Issue #2230, opened 2017-07-12
Release scripts should be runnable from develop


bobylito:
Right now when we release new versions in which we update the release scripts, we have to manually update the `master` branch and then launch the release scripts. This is because the `master` never has the latest build scripts, but this could be solved if we were running the scripts from the `develop` branch.

PR #2246, opened 2017-07-18
chore(release): allow releases from develop

iam4x:
related to #2230

* run release from develop
* then merge release from develop into master branch

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-[ -z $HOTFIX ] && HOTFIX='0'
-
 set -e # exit when error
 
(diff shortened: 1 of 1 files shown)


bobylito:
I had in mind that it would be better to have the whole release process happen on master. Something like that:

 0. check if on develop
 1. move to master 
 2. build stuff etc.
 3. publish, tag, etc.
 4. merge back to develop

iam4x:
I've updated the flow, it is the following:

1. check if you are on branch develop
2. git checkout master
3. pull and merge develop into master
4. build, tag, commit on master and publish to npm
5. merge back master into develop and push develop
6. checkout develop

PR #2253, opened 2017-07-20
fix(url-sync): remove is_v from url

### LINKED ISSUE #1105: is_v parameter append to url even if not tracked.
I have the following config:
```js
var search = instantsearch({
            appId: algolia.application_ID,
            apiKey: algolia.search_API_key,
            indexName: algolia.default_search_index,
            urlSync: {
                mapping: {'q': 's'},
                trackedParameters: ['query']
            },

            searchParameters: {
                distinct: '1'
            }
        });
```

I want to only have the `query` tracked in the url as `s`.
A weird `is_v` parameter gets also append even if it is not part of the `trackedParameters` list.

bobylito:
Fixes #1105


vvo:
Wondering if we remove is_v by default this might trigger any breaking change somewhere since this was deeply integrated (everything url related is tricky to remove). Maybe not, since also v2 is somehow new it could be ok completely

Haroenv:
I can't imagine someone using that. The only usecase I can think of is Wordpress, because there supporting both versions is slightly hard, but they didn't upgrade yet

src/lib/url-sync.js
```diff
@@ -88,7 +86,8 @@ function getFullURL(relative) {
 // IE <= 11 has no location.origin or buggy
 function getLocationOrigin() {
   // eslint-disable-next-line max-len
-  return `${window.location.protocol}//${window.location.hostname}${window.location.port ? `:${window.location.port}` : ''}`;
+  const port = window.location.port ? `:${window.location.port}` : '';
+  return `${window.location.protocol}//${window.location.hostname}${port}`;
```
Haroenv: What about the path, is that not relevant too?

package.json
```diff
@@ -70,7 +70,7 @@
     "express": "^4.15.2",
     "gh-pages": "^1.0.0",
     "http-server": "^0.9.0",
-    "jest": "^19.0.2",
+    "jest": "^20.0.4",
```
iam4x: Should we try next time to do dependencies in another commit / PR ?

src/lib/url-sync.js
iam4x: @Haroenv The path is not relevant in the location.origin, this method is just a polyfill for older browsers.

src/lib/__tests__/url-sync-test.js
```diff
@@ -0,0 +1,51 @@
...
+    const helper = jsHelper({addAlgoliaAgent: () => {}});
+    const urlUtils = makeTestUrlUtils();
+    const urlSyncWidget = urlSync({urlUtils, threshold: 0});
+    urlSyncWidget.render({helper});
+    helper.setQuery('query');
+
+    jest.runOnlyPendingTimers();
+
+    expect(urlUtils.lastQs).not.toEqual(expect.stringContaining('is_v'));
+  });
```
iam4x: Can we add a test to check if additional parameters in the URL are not replaced?

iam4x:
Looks good to me 👍  Just added a few notes.

src/lib/url-sync.js
bobylito: Oh and I change that because my syntax highlighting was going crazy with templates litterals in template litterals. #donottroll :)

src/lib/__tests__/url-sync-test.js
bobylito: It's kinda out of scope but since you had the question the other day (what if it's really not working ;)). I'll do that in another PR.

iam4x: LGTM 💯

bobylito:
> Wondering if we remove is_v by default this might trigger any breaking change somewhere since this was deeply integrated

The parameter was treated as an external parameter and the value was not leveraged.

package.json
bobylito: My bad, I was going crazy with some Jest syntax and lack of doc. Let's be more clean yeah next time ;)

bobylito:
Tests are ok and I checked argos. It's all good :)

PR #2255, opened 2017-07-21
feat(doc): Add a guide for opensearch

### LINKED ISSUE #2088: Add a guide on opensearch
OpenSearch provides search for websites from the url bar, you may have been using it without noticing it, works pretty well.

We should add a guide on opensearch in our community websites to show people what can be done when using instantsearch.js/react-instantsearch for OpenSearch.

If opensearch is a good pattern that would benefit our users, then we should have a guide about it.

Example: Try in your browser: "yarn [TAB] lodash", this is opensearch. https://github.com/yarnpkg/website/blob/master/opensearch.xml

It needs to be linked in the `head` like this:

https://github.com/yarnpkg/website/blob/66acf52e8f096bef1d74dce6895874f64b3f44d4/_layouts/default.html#L22

bobylito:
For this guide, we've also added the syntax highlighting for diff patches.

Fix #2088

### DIFF
--- a/docgen/src/data/communityHeader.json
+++ b/docgen/src/data/communityHeader.json
@@ -65,9 +65,14 @@
         "url": "guides/customization.html"
       },
       {
+        "name": "Integrate OpenSearch",
+        "url": "guides/opensearch.html"
+      },
+      {
         "name": "Migrate from V1",
         "url": "guides/migration.html"
       }
+
     ]
   },
   {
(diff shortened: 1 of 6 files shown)


bobylito:
It's not available in the netlify preview but I tested locally :/

bobylito:
> Seems okay, weird that it doesn't show. Did you add a link?

`communityHeader.json` fail. Will fix that :). In the mean time the guide is [here](https://deploy-preview-2255--algolia-instantsearch.netlify.com/guides/opensearch.html)

PR #2263, opened 2017-07-25
feat(publish-script): re-write as js script, allow beta releases

iam4x:
Related to #2056 

I have re-written the release script in javascript to enhance the user experience when doing a new release and also to allow releases from another branch than develop for beta releases.

Here is the flow:

* if branch === develop -> do normal release
* if branch === master -> ask to be on develop for normal release or another branch for beta
* if branch !== develop || branch !== master -> ask if user wants to release a beta

**In case of beta, ask for a new version number with -beta.x suffix**

* [x] Allow stable releases
* [x] Allow beta releases
* [x] Revert changes if process does not go until the end
* [x] Refactoring of duplicated code (for different releases cycles)

Result:

image

image

image


Haroenv:
Is there a reason the release.sh script isn't removed?

bobylito:
> Is there a reason the release.sh script isn't removed?

WIP, maybe?

iam4x:
Need a real world testing on a fork and your review guys 👍

scripts/release.js
```diff
@@ -0,0 +1,214 @@
...
+      console.log(colors.blue('Updating working tree'));
+      execSync('git checkout master');
+      execSync('git pull origin master');
+      execSync('git fetch origin --tags');
+      execSync('git fetch origin develop');
+      execSync('git merge origin/develop');
+    }
+
+    console.log(colors.blue('Install dependencies'));
+    execSync('yarn');
```
bobylito: Do we need to clean up the node_modules folder beforehand with `yarn`?

scripts/release.js
```diff
@@ -0,0 +1,214 @@
...
+      - Current version is "${currentVersion}"
+      - Changelog will be generated only if a fix/feat/performance/breaking is found in git log
+      - You must choose a new ve.rs.ion (semver) ${strategy === 'beta'
+        ? 'with -beta.x suffix'
+        : ''}
+    `)
+    );
+
+    execSync(
+      'conventional-changelog --preset angular --output-unreleased | less'
```
bobylito: In my experience, this step can be annoying because you can't go back to the output when choosing the version number. What I do in the Helper JS script, is that I take the output and display it inline with the script: https://github.com/algolia/algoliasearch-helper-js/blob/develop/scripts/release.js#L60-L64

scripts/release.js
```diff
@@ -0,0 +1,214 @@
...
+          `VERSION=${newVersion} babel-node ./scripts/bump-package-version.js`
+        );
+
+        // build library new version
+        console.log(colors.blue('Building library new version'));
+        execSync(`NODE_ENV=production VERSION=${newVersion} npm run build`);
+
+        // update changelog
+        console.log(colors.blue('Update changelog'));
+        const changelog = execSync('conventional-changelog -p angular')
```
bobylito: Why do we execute it here?

scripts/release.js
```diff
@@ -0,0 +1,214 @@
...
+
+        // regenerate README TOC
+        console.log(colors.blue('Generate TOCS'));
+        execSync('npm run doctoc');
+
+        // regenerate yarn.lock
+        execSync('yarn');
+
+        // git add and tag
+        const commitMessage = `v${newVersion}\n\n${changelog}`;
```
bobylito: Does the changelog contain some information about the version?

scripts/release.js
```diff
@@ -0,0 +1,214 @@
...
+            {
+              type: 'confirm',
+              name: 'publishNpm',
+              message: colors.yellow.underline(
+                'Is everything OK? Continue to push to github and publish the package?'
+              ),
+            },
+          ])
+          .then(({ publishNpm }) => {
+            if (!publishNpm) rollback(newVersion);
```
bobylito: I would use an else here to indicate we're not continuing the flow. Something like:
```js
if (publishNpm) {
  // do the release
} else {
  rollback(newVersion);
}
```

scripts/release.js
```diff
@@ -0,0 +1,214 @@
...
+
+            console.log(colors.blue('Push to github, publish on npm'));
+            if (strategy === 'stable') {
```
bobylito: This part could be a function.

scripts/release.js
```diff
@@ -0,0 +1,214 @@
...
+    It's not possible to release a new version from the master branch.
+    Please checkout any other branch:
+      - develop: release new stable version
+      - xxxxxxx: release new beta version
+  `)
+  );
+  process.exit(1);
+}
+
+const strategy = currentBranch === 'develop' ? 'stable' : 'beta';
```
vvo: I don't quite understand this line, does it means when on develop you can only do stable releases and on any other branch (but master), do a beta release?

scripts/release.js
vvo: Good question, if you mess up with your node_modules/... then yarn should ensure that everything is good, I have no idea if that's the case you will have to test it

scripts/release.js
bobylito: Correct.

scripts/release.js
rayrutjes: No later than yesterday, I realized that yarn is not that clever. If you manually delete something inside of the node_modules folder, it not always catches it. I would go for the safer  "reboot".

scripts/release.js
iam4x: Because we need the changelog for the commit message and I think that a changelog for beta versions could be good as well. Maybe reset it before a new stable version, wdyt?

scripts/release.js
iam4x: I don't understand what you meant.

scripts/release.js
iam4x: Are you okey with:

```js
if (!publishNpm) return rollback(newVersion)
```

The `return` keyword explicitly stops the execution here?

scripts/release.js
iam4x: I did some functions before, it was harder to read because going to reference to references. Since it's a script I wanted something straightforward that you can read like a text from the top to the bottom without jumping elsewhere to avoid mistakes when we will work on it again.

scripts/release.js
iam4x: The "safest" way is to:

```shell
rm -rf node_modules docgen/node_modules && 
    yarn cache clean &&
    yarn

cd docgen && yarn
```

This way we will have no cache, same packages as described into the `yarn.lock` and `package.json`

What do you think?

scripts/release.js
iam4x: Because we want to update the changelog?

scripts/release.js
iam4x: 👍

scripts/release.js
bobylito: That'll do the trick 👍

scripts/release.js
bobylito: What is the content of `changelog`? Are you sure it does not contain a header with the version? I'm pretty it does, can you check?

scripts/release.js
bobylito: Go for it @iam4x :)

bobylito:
> Need a real world testing on a fork and your review guys 👍

Let's do the release with it :D

scripts/release.js
iam4x: Yes it does, what is the issue?

PR #2264, opened 2017-07-27
chore(website): Fix corrupted assets during build

bobylito:
The in-place plugin seems to tamper all the files, even
though it does not transform them >> Scoping this plugin to
only relevant extensions.

Also in this PR: we let the user publish the doc from any branch but we ask for confirmation if not on master.


bobylito:
> lgtm, we should've seen that in the deploy preview last time though 😶

I was careless :) It's fixed now :D

PR #2267, opened 2017-07-31
chore(doc): Creates widget showcase

### LINKED ISSUE #2210: Add a widget showcase to the documentation
The new v2 documentation keep and enhance most aspects of the previous aspects of the previous documentation. However it does not contain a page where we can see all the widgets in a single glance like we could in the previous documentation: discoverability of widgets is less easy.

## Proposal

Add a page where all the widgets are rendered. This page will be underneath the introduction page and before the widgets link (in the sidebar).

bobylito:
This creates a new page that showcases all the widgets of instantsearch.js. For that purpose, this commit also introduces a jsdoc tag: @category.

Fixes #2210


bobylito:
You can see the showcase here: https://deploy-preview-2267--algolia-instantsearch.netlify.com/widgets-showcase.html

iam4x:
Waaahouuu 🎉 

Awesome work @bobylito it's a great page for selling features of InstantSearch.js 👍 
How ever, while testing it I've found severals bugs or small issues:

---------

There is a bug we might have not found before, when you click on the `infiniteHits -> load more` it changes also the results of the `hits` widget.

---------

The `pagination` widget styles are modified from the documentation styles:

```css
.documentation-section .documentation-container ul li, .documentation-section .documentation-container ol li, .examples-section .documentation-container ul li, .examples-section .documentation-container ol li {
    clear: both;
    margin-bottom: 10px;
}
```

You should exclude the `margin-bottom: 10px` rules on any `.ais-pagination` children.

---------

The buttons hover styles are broken as well, the documentation styles override the `a:hover` with:

```css
a:hover, a:focus {
    color: #007ab3;
    text-decoration: underline;
}
```

This should be excluded from widgets on the page.

-----------

Maybe move the `clearAll` widget after the current refined values or at the end of the "Filtering widgets" section. It feels weird to clear all filters when did not played with them yet.

------------

For the `currentRefinedValues` widget, I think that `false` is not a good showcase of an already refined value.

image

------------

The `menu` widget seems to be broken

image

docgen/plugins/documentationjs-data.js
```diff
@@ -70,8 +72,14 @@ function filterSymbolsByType(type, symbols) {
   });
 }
 
+function groupSymbolsByCategories(symbols) {
+  return groupBy(symbols, (s) => {
+    const tag = filter(s.tags, {title: 'category'})[0];
+    return tag && tag.description || 'other';
```
iam4x: ES6 for the win 🥇 

```js
const [ tag ]  = filter(s.tags, { title: 'category' });
return tag && tag.description || 'other';
```

Haroenv:
```js
return [ {description = 'other' } ]  = filter(s.tags, { title: 'category' });
```

This works too, no?

bobylito: w00t thanks. Will try :)

iam4x: @Haroenv Not really, it can fail when the first tag found is undefined. You can fix it this way:

```js
const [ { description = 'other' } = {} ] = filter(s.tags, { title: 'category' });
return description;
```

And even this can fail if the first tag found is not an object or not undefined. I would not recommend to go this far in object destructuring if your are not sure of the result.

bobylito:
I fixed what could be fixed, on the widgets side, tricking the API of the widgets, playing with ~fire~ css. It should be good to go. Also move the content to the introduction and move the common parameters of the widgets into their own page.

bobylito:
Great suggestions @Haroenv. I'm opening a new issue with your suggestion https://github.com/algolia/instantsearch.js/issues/2268

bobylito:
Merging because there is a fix too.

PR #2270, opened 2017-08-02
Feat/2.1 namespace es6 build

iam4x:
As we discussed on Slack, we will namespace the whole ES6 modules build like some popular libraries does (lodash-es, redux/es, ...)

I've updated the usage guide to fix the example:

```js
// instantsearch() function without reference to the widgets or connectors
import instantsearch from 'instantsearch.js/es';

// import connectors individually
import {connectSearchBox} from 'instantsearch.js/es/connectors';

// import widgets individually
import {searchBox} from 'instantsearch.js/es/widgets';

const search = instantsearch({ ... });

search.addWidget(searchBox({ ... }));
search.addWidget(connectSearchBox(function() { ... })({ ... }))
```

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -2,8 +2,8 @@
 
 set -e # exit when error
 
-rm -rf dist dist-es5-module dist-es6-module connectors.js instantsearch.js widgets.js
-mkdir -p dist dist-es5-module dist-es6-module
+rm -rf dist dist-es5-module es connectors.js instantsearch.js widgets.js
+mkdir -p dist dist-es5-module es
 
 echo "➡️  Bundle instantsearch.js to UMD build './dist' via webpack"
 NODE_ENV=production BABEL_ENV=production webpack --config scripts/webpack.config.js --hide-modules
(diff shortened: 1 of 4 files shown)


scripts/build.sh
```diff
@@ -16,13 +16,8 @@ BABEL_ENV=production babel src -d dist-es5-module/src --ignore *-test.js --quiet
...
-BABEL_ENV="production-es6" babel -q index.es6.js -o instantsearch.js &
-BABEL_ENV="production-es6" babel src -d dist-es6-module --ignore *-test.js --quiet &
-
-# * allow `import {connectXXX} from 'instantsearch.js/connectors'`
-# * allow `import {searchBox} from 'instantsearch.js/widgets'`
-echo "export * from './dist-es6-module/connectors/index';" >> ./connectors.js &
-echo "export * from './dist-es6-module/widgets/index';" >> ./widgets.js
+echo "➡️  Bundle instantsearch.js to ES6 build './es' via babel-cli"
+BABEL_ENV="production-es6" babel -q index.es6.js -o es/index.js &
+BABEL_ENV="production-es6" babel src -d es --ignore *-test.js --quiet &
```
Haroenv: How come the files are called `-test.js` and not `.test.js`?

iam4x: This is another problem, but yes it was like because we were not using jest before.

PR #2273, opened 2017-08-03
docs(getting-started): make sure that the template doesn't get replaced

Haroenv:
The `in-place` plugin replaced these tags by their value (undefined), so they didn't show up.

Thanks @ronanlevesque for noticing this!


Haroenv:
This should be deployed too, but I forgot how to do that manually without a release cc @iam4x

iam4x:
I deployed the docs @Haroenv it should be live in a few hours when Github renews the cache 👍

Issue #2275, opened 2017-08-07
How to hide the average value for the V2 slider?


mindbits:
Hi, 

the new slider shows the average value in the middle. 

Which is kind of odd, especially when the average is not a round number like in the example on https://community.algolia.com/instantsearch.js/v2/widgets/rangeSlider.html where the average is 2500.50 for the range of 1-5000. 2500.50 * 2 is 5001

Is there a way to hide the average values completey? Or at least an option to have a template function against it that values can be rounded to sane values?

Thank you

iam4x:
Hello,

Thanks for the feedback @mindbits. We have no option for this a the moment, but I think it's a reasonable use case to implement an option for it.

What do you think @bobylito @Haroenv ?

Haroenv:
I was looking at the widget, and it makes sense to add some configuration about that. What are the options of Rheostat?

iam4x:
There is already the `tooltip` option you can set at `false` to remove all the labels, the min, medium and max. (https://github.com/algolia/instantsearch.js/blob/bcdaa5b58c3c81110287669a71671d2a608259e1/src/components/Slider/Slider.js#L84-L88)

There is no easy way to disable only the medium right now.

Haroenv:
As a workaround, you can do: 

```js
document.querySelectorAll(".ais-range-slider--value")[1].hidden = true
```

mindbits:
Another problem I am having with the slider: 

Eg the slider in the docs: How do you show only items that cost exactly $5,000? 

It's not possible to have only 1 value. In the example the minimum range is $4,942 - $5,000. Which might not matter as the values are high but if you have a slider with a smaller range, eg hotel stars, it might make more sense that customers want to show only 1 value.

This was easily doable with the old slider. 

Is there an easy way to use the old slider instead of the new one?

Haroenv:
For stars you can use the [starRating](https://community.algolia.com/instantsearch.js/v2/widgets/starRating.html) widget, which isn't a slider, but works perfectly for it

mindbits:
I know, this was just an example. But I am sure that there are many use cases where a min/max range slider should be able to return only 1 value?

Haroenv:
Do you mean only integer values?

mindbits:
Not necessarily integer values but single values. 

With the old range slider you could select a range of $5,000 - $5,000 , so just show everything costing $5,000. 

https://community.algolia.com/instantsearch.js/documentation/#rangeslider

[...]

PR #2276, opened 2017-08-07
fix(Pit): round the middle value to the whole number

Haroenv:
Most prices start from > 0, which causes annoying decimal values at the middle.

This PR fixes that by always rounding the middle value to a whole number

Maybe not a complete fix for #2275, but definitely related to the underlying reason.

version|result
---|---
before|screen shot 2017-08-07 at 16 37 02
after|screen shot 2017-08-07 at 16 36 38


Haroenv:
I wanted to add this as a prop first, but it seems like that might be hard with Rheostat

iam4x:
When I did this, I wanted to correct the issues with the rangeSlider from v1 which was not placing the middle pit at the correct middle.

And also, to provide support for negatives values which was a nightmare for the v1.

What I purpose, is to use the `precision` parameter to know at what decimal we should round for the start/middle/end pit value.

What do you think?

Haroenv:
Having this as an argument seems good to me. Precision as number of digits after 0, defaulting to 1?

iam4x:
Yes exactly, the default value is 2 👍

Haroenv:
Pit has now got a `precision` Prop, how do we expose it to the users, via `Slider`?

bobylito:
@Haroenv what's the status on this? If you don't have time to take care of it for the moment, do you mind if I close the PR?

PR #2277, opened 2017-08-07
feat(slider): allow setting which Pits to show  

Haroenv:
A prop `valuesToShow={[0, 50, 100]}` allows you to specify which percentages to show. The annoying part is that not for every % a Pit is rendered. The values are:

```js
const pitValues = [0, 3, 6, 9, 12, 15, 18, 21, 24, 26, 29, 32, 35, 38, 41, 44, 47, 50, 53, 56, 59, 62, 65, 68, 71, 74, 76, 79, 82, 85, 88, 91, 94, 97, 100];
```

I'm not sure what to make of it.

related to #2276

Only for review since I continued on #2276

### DIFF
--- a/src/components/Slider/Pit.js
+++ b/src/components/Slider/Pit.js
@@ -4,10 +4,9 @@ import cx from 'classnames';
 
 import includes from 'lodash/includes';
 
-const Pit = ({style, children}) => {
-  // first, end & middle
+const Pit = ({style, children, valuesToShow = [0, 50, 100]}) => {
   const positionValue = Math.round(parseFloat(style.left));
-  const shouldDisplayValue = includes([0, 50, 100], positionValue);
+  const shouldDisplayValue = includes(valuesToShow, positionValue);
 
   return (
     <div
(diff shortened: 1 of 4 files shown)


src/components/Slider/Pit.js
```diff
@@ -17,7 +16,7 @@ const Pit = ({style, children}) => {
       }) }
     >
       { shouldDisplayValue
-          ? <div className="ais-range-slider--value">{ Math.round(children * 100) / 100 }</div>
+          ? <div className="ais-range-slider--value">{ Math.round(children) }</div>
```
iam4x: Why did you changed that?

iam4x:
Nice option I like it 👍

src/components/Slider/Pit.js
Haroenv: That's from the previous PR, I worked on top of it, can fix later

Haroenv:
Regardless of the wrong history, I don't really like how the possible numbers progress, because if you don't pick a number from that first list, it won't show up.

bobylito:
Hey @Haroenv is there something that we can do to make this PR move on? Or should we just close for the time being, until we have a proper plan of attack?

Haroenv:
This solution is flawed because the numbers you can set aren't predictable, so closing

PR #2282, opened 2017-08-09
docs(getting-started): use `instant_search` as indexName

Haroenv:
This is the one used in the examples of the widgets, which is almost the same as bestbuy, but slightly different.

As noticed by @ronanlevesque, there are a few widgets (range, rangeSlider, hierarchicalMenu and menu) that didn't work with the attribute names in `latency`, but that does work with the ones in `instant_search`.

The API key here is search-only


iam4x:
LGTM 👍 

I had the same stuggle with @marielaures when showing her how to make widgets

PR #2296, opened 2017-08-25
fix(Pagination): add `autohideContainerHOC` to <Pagination />

### LINKED ISSUE #2287: Pagination Widget autoHideContainer does not hide the entire element.
A Bug

When no results returned, `autoHideContainer` on the Pagination widget doesn't hide the entire element. 

You can view this behaviour here by searching for something with no results: https://jsfiddle.net/gmattucc/qxhpm91j/

That the entire pagination element is hidden like the behaviour of other widgets.

All

Have the data react root have `display=none` like all the other widgets.

v2.0.2

iam4x:
The `autohideContainerHOC` was missing from `<Pagination />` widget.

Fixes #2287


bobylito:
Changed the target to `develop` since it's a bug.

PR #2297, opened 2017-08-25
default hits per page

### LINKED ISSUE #2288: hitsPerPageSelector default value in v2
Hi,

No more hitsPerPage in hits since v2.

How do we set a default value for `hitsPerPageSelector` Widget in instantsearch v2?

Thanks

iam4x:
**Summary**

Actually, there is no options to set a default hitsPerPage appart from using the `searchParameters` object when you instantiate `instantsearch`.

My proposal, is to add a key `default: true` into the items[] of the hitsPerPageSelector options to specify on which value you want to start.

```js
instantsearch.widgets.hitsPerPageSelector({
  container,
  items: [
    { value: 3, label: '3 per page' },
    { value: 5, label: '5 per page', default: true },
    { value: 10, label: '10 per page' },
  ],
});
```

In this example the default value will be "5 per page" hits on the first search.

Fixes #2288

### DIFF
--- a/src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.js
+++ b/src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.js
@@ -7,6 +7,17 @@ import connectHitsPerPage from '../connectHitsPerPage.js';
 const fakeClient = { addAlgoliaAgent: () => {} };
 
 describe('connectHitsPerPage', () => {
+  it('should throw when there is two default items defined', () => {
+    expect(() => {
+      connectHitsPerPage(() => {})({
+        items: [
+          { value: 3, label: '3 items per page', default: true },
+          { value: 10, label: '10 items per page', default: true },
+        ],
+      });
+    }).toThrow(/^\[Error\]/);
+  });
+
   it('Renders during init and render', () => {
     // test that the dummyRendering is called with the isFirstRendering
     // flag set accordingly
--- a/src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.js
+++ b/src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.js
@@ -20,6 +20,7 @@ describe('connectHitsPerPage', () => {
     });
 
     expect(typeof widget.getConfiguration).toEqual('function');
+    expect(widget.getConfiguration()).toEqual({});
 
     // test if widget is not rendered yet at this point
     expect(rendering.callCount).toBe(0);
@@ -80,7 +81,7 @@ describe('connectHitsPerPage', () => {
     });
   });
 
-  it('Configures the search with the first item value provided', () => {
+  it('Does not configures the search when there is no default value', () => {
     const rendering = sinon.stub();
     const makeWidget = connectHitsPerPage(rendering);
     const widget = makeWidget({
@@ -90,9 +91,7 @@ describe('connectHitsPerPage', () => {
       ],
     });
 
-    expect(widget.getConfiguration()).toEqual({
-      hitsPerPage: 3,
-    });
+    expect(widget.getConfiguration()).toEqual({});
   });
 
   it('Provide a function to change the current hits per page, and provide the current value', () => {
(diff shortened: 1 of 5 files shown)


vvo:
Small context on why at some point we said that hitsPerPage would be configured at a single point: https://github.com/algolia/instantsearch.js/pull/1653

src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.js
```diff
@@ -19,7 +19,7 @@ describe('connectHitsPerPage', () => {
       ],
     });
 
-    expect(widget.getConfiguration).toEqual(undefined);
+    expect(typeof widget.getConfiguration).toEqual('function');
```
bobylito: Since we have a `getConfiguration` we should test the returned value of this function.

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -115,6 +117,13 @@ export default function connectHitsPerPage(renderFn) {
     }
 
     return {
+      getConfiguration() {
+        const { value: defaultHitsPerPage } =
+          items.find(item => item.default) || items[0];
```
bobylito: This changes the previous behaviour (not default set). By default, we were using the hits per page used by the engine and configured in the index. We should not change that and therefore we should not set any value if there are no `default: true`.

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -115,6 +117,13 @@ export default function connectHitsPerPage(renderFn) {
...
+
+        return { hitsPerPage: defaultHitsPerPage };
```
bobylito: One thing that could be great is to display an error message when there are more than one `default:true`

bobylito:
Also since we have now a default value, we might not want to have the empty value anymore... That could be another fix.

src/connectors/hits-per-page/connectHitsPerPage.js
iam4x: :+1:

src/connectors/hits-per-page/connectHitsPerPage.js
iam4x: 👍

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -118,10 +118,17 @@ export default function connectHitsPerPage(renderFn) {
...
-          items.find(item => item.default) || items[0];
+        if (items.filter(item => item.default).length > 1 && window.console) {
+          window.console.warn(
+            `[Warning][hitsPerPageSelector] more than one default value is specified in \`items[]\`
+  The first one will be picked, you should probably set only one default value`
+          );
+        }
 
-        return { hitsPerPage: defaultHitsPerPage };
+        const defaultHitsPerPage = items.find(item => item.default);
```
bobylito: That's not big but there are 2 loops over items (one with filter and the other with find). We could store the result of the filter and reuse it here.

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -115,6 +117,20 @@ export default function connectHitsPerPage(renderFn) {
     }
 
     return {
+      getConfiguration() {
+        const defaultValues = items.filter(item => item.default);
+        if (defaultValues.length > 1 && window.console) {
```
bobylito: This can be tested when the widget when the widget is instanciated. At this point, we can safely throw an error (instead of just using a warning in the console).

src/widgets/hits-per-page-selector/__tests__/hits-per-page-selector-test.js
```diff
@@ -56,8 +56,9 @@ describe('hitsPerPageSelector()', () => {
     };
   });
 
-  it("doesn't configure anything", () => {
-    expect(widget.getConfiguration).toEqual(undefined);
+  it('does configure the default hits per page if specified', () => {
+    expect(typeof widget.getConfiguration).toEqual('function');
+    expect(widget.getConfiguration()).toEqual({});
```
bobylito: Maybe we should test the actual configuration if there is a default value? Also we should test that the widget throws an error if there is more than one default value.

PR #2303, opened 2017-08-30
Fix `edit this page` link for reference API

bobylito:
This patch fixes the edit this page link. The problem was linked to the environment used and the assumptions that the folder of the project would be `instantsearch.js`. Now we use a more generic solution.

This patch also move the destination of the link to develop to avoid duplicate contributions.


vvo:
@oliviertassinari Starting to see false positive on Argos, any hints?

oliviertassinari:
@vvo Looking at the screenshot, it seems that the Algolia Index changed. Some records have been added. How is this a false positive?

vvo:
Sorry I thought I had already validated those changes (accepted) in another PR and thought this was weird.

PR #2310, opened 2017-09-05
fix(sffv): no error when not providing noResults and no results

### LINKED ISSUE #2087: Instantsearch.js refinementList widget with `searchForFacetValues: true` leads to template/rendering error when there are no results
When adding an instantsearch.js refinementList widget, a user can choose to define a template string/function to render facets. In the case where a user wants to allow searchable facets, whether or not a refinementList template is defined, the widget does a great job of rendering sensible markup when there are results...

However, when searching the refinement list for facet values AND there are NO relevant results returned by the API, an error is thrown in the browser:

`Template must be 'string' or 'function', was 'undefined' (key: noResults)`

The apparent solution here is to simply add the `noResults` property to the templates object that is passed to the refinementList widget declaration. However, I could not find any explanation of this in any guides...

I see that it is mentioned [here](https://community.algolia.com/instantsearch.js/documentation/#templates) but only in the params list.

There is little info in any guide I found concerning the topic of how to properly implement searchableAttributes, attributesForFaceting, or  search-for-facet-values templating
( i.e [here](https://www.algolia.com/doc/api-client/javascript/parameters/searchableAttributes/), [here](https://www.algolia.com/doc/api-client/javascript/parameters/attributesForFaceting/), or [here](https://www.algolia.com/doc/api-client/javascript/search/#search-for-facet-values)).

I think it may be worthwhile to explicitly state somewhere in the instantsearch/templates guide that this `noResults` prop should be specified when building a searchable facets list.

Vincent suggested that this behavior is actually a bug:

"...by default if you do not provide the template for no results in the refinementList it should degrade nicely. If that's not the case then it's a miss"

I agree.

[Here is a JS Fiddle](https://jsfiddle.net/aw30jpud/) I just wrote to demonstrate this issue. See error in browser dev console when searching with a query string that doesn't return results in the "Publisher" refinement list (e.g. a query like `awdawdawd`)

https://jsfiddle.net/aw30jpud/

bobylito:
When there was no results and the template is not provided, is.js would trigger an error.

Fix #2087

### DIFF
--- a/src/widgets/refinement-list/refinement-list.js
+++ b/src/widgets/refinement-list/refinement-list.js
@@ -6,6 +6,7 @@ import filter from 'lodash/filter';
 import RefinementList from '../../components/RefinementList/RefinementList.js';
 import connectRefinementList from '../../connectors/refinement-list/connectRefinementList.js';
 import defaultTemplates from './defaultTemplates.js';
+import sffvDefaultTemplates from './defaultTemplates.searchForFacetValue.js';
 import getShowMoreConfig from '../../lib/show-more/getShowMoreConfig.js';
 
 import {
@@ -253,7 +254,7 @@ export default function refinementList(
     ? prefixKeys('show-more-', showMoreConfig.templates)
     : {};
   const searchForValuesTemplates = searchForFacetValues
-    ? searchForFacetValues.templates
+    ? searchForFacetValues.templates || sffvDefaultTemplates
     : {};
   const allTemplates = {
     ...templates,
--- a/dev/app/init-builtin-widgets.js
+++ b/dev/app/init-builtin-widgets.js
@@ -294,6 +294,25 @@ export default () => {
       })
     )
     .add(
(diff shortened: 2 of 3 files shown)


src/widgets/refinement-list/refinement-list.js
```diff
@@ -253,7 +254,7 @@ export default function refinementList(
     ? prefixKeys('show-more-', showMoreConfig.templates)
     : {};
   const searchForValuesTemplates = searchForFacetValues
-    ? searchForFacetValues.templates
+    ? searchForFacetValues.templates || sffvDefaultTemplates
     : {};
```
Haroenv: This condition seems a bit complex to me. I'm wondering if something like this is more straightforward: 

```js
const searchForValuesTemplates = searchForFacetValues
  ? searchForFacetValues.templates
    ? searchForFacetValues.templates
    : sffvDefaultTemplates
  : {};
```

It's still a double ternary, but this might be easier to read.

Issue #2309, opened 2017-09-05
Let `refine` accept undefined to remove a selected value


bobylito:
When implementing a connector like menu it's hard to know what to do to remove a value. The refine is implemented like a toggle [but it's not obvious](https://stackoverflow.com/questions/46000184/algolia-instantsearch-connectmenu-refine-remove-current-refined-value). If we could give a specific value to `refine` that would help a lot documentation-wise and implementation-wise (no need to store the value).

Impacted connectors:
 - menu
 - hierarchicalMenu
 - refinementList
 - hitsPerPage (this one could work but we'd need to document it)
 - numericRefinementList
 - numericSelector

Haroenv:
Like React InstantSearch, the refinement value in a refinement list should accept the whole list to apply. This way you can easily select multiple values.

PR #2316, opened 2017-09-07
add menu select widget

### LINKED ISSUE #1904: Add a menuSelect widget
feature

Redering a Menu as a select has been asked too many times to be ignored anymore. We need a menuSelect no matter how far we go in providing advanced rendering customisation.

```js
search.addWidget(instantsearch.widgets.menuSelect({
  attributeName: '...',
  limit: ...
  label: '...'
}));
```

This should render:

```html
<label>... <select><option></option></select></label>
```

Ref: https://github.com/algolia/instantsearch.js/issues/868

- instantsearch.js

latest

iam4x:
**Summary**

Works like a menu widget but displays as a select DOM element.

* [x] Widget
* [x] Documentation
* [x] Example
* [x] Tests

**Result**

sep-07-2017 17-54-38

Fixes #1904

### DIFF
--- a/src/components/MenuSelect.js
+++ b/src/components/MenuSelect.js
@@ -33,7 +33,7 @@ class MenuSelect extends Component {
         onChange={this.handleSelectChange}
       >
         <option value="" className={cssClasses.option}>
-          <Template templateKey="seeAllOption" templates={templateProps} />
+          <Template templateKey="seeAllOption" {...templateProps} />
         </option>
 
         {items.map(item =>
@@ -42,11 +42,7 @@ class MenuSelect extends Component {
             value={item.value}
             className={cssClasses.option}
           >
-            <Template
-              data={item}
-              templateKey="item"
-              templates={templateProps}
-            />
+            <Template data={item} templateKey="item" {...templateProps} />
           </option>
         )}
       </select>
--- a/src/components/MenuSelect.js
+++ b/src/components/MenuSelect.js
@@ -33,7 +33,7 @@ class MenuSelect extends Component {
         onChange={this.handleSelectChange}
       >
         <option value="" className={cssClasses.option}>
-          <Template templateKey="seeAllOption" {...templateProps} />
+          <Template templateKey="seeAllOption" templates={templateProps} />
         </option>
 
         {items.map(item =>
@@ -42,7 +42,11 @@ class MenuSelect extends Component {
             value={item.value}
             className={cssClasses.option}
           >
-            <Template data={item} templateKey="item" {...templateProps} />
+            <Template
+              data={item}
+              templateKey="item"
+              templates={templateProps}
+            />
           </option>
         )}
       </select>
--- a/src/components/MenuSelect.js
+++ b/src/components/MenuSelect.js
@@ -1,0 +1,55 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import Template from './Template';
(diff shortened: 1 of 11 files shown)


bobylito:
Nice :)

src/components/MenuSelect.js
```diff
@@ -0,0 +1,55 @@
...
+    cssClasses: PropTypes.shape({
+      select: PropTypes.string,
+      option: PropTypes.string,
+    }),
+    items: PropTypes.array.isRequired,
+    refine: PropTypes.func.isRequired,
+    templateProps: PropTypes.object.isRequired,
+  };
+
+  get selectValue() {
```
bobylito: This is not very idiomatic with the rest of the codebase. Do we want to move forward with that notation?

bobylito: isn't it selectedValue?

src/components/MenuSelect.js
```diff
@@ -0,0 +1,55 @@
...
+    const selectValue = this.props.items.find(({ isRefined }) => isRefined);
+    return selectValue || { value: 'all' };
+  }
+
+  handleSelectChange = ({ target: { value } }) => {
+    this.props.refine(value === 'all' ? this.selectValue.value : value);
```
bobylito: That's interesting. You implement a special value for the selection. I think it should be part of the connector API. I would go for `undefined` or `null` (or both) as `refine(null)` is semantically close to "refine nothing" which is what we want when implementing "no filter / all".

bobylito: For reference, this is what a user tried to do: https://stackoverflow.com/questions/46000184/algolia-instantsearch-connectmenu-refine-remove-current-refined-value

bobylito: https://github.com/algolia/instantsearch.js/issues/2309

src/components/MenuSelect.js
iam4x: In my head it was the value of the <select> DOM element, so `selectValue`. But it works the other way around also.

iam4x: Let's use the JavaScript features we are offered no? (When it makes sense).
I like JavaScript getters, and here it makes sense because I use this value more than once.

src/components/MenuSelect.js
iam4x: Yes, I don't really know if it's the right time, we need to introduce this change a bit everywhere (according to the issue) like on the connectRefinementList connectHieararchicalMenu etc...

We can do this in the next step?

src/connectors/menu/connectMenu.js
```diff
@@ -139,6 +139,23 @@ export default function connectMenu(renderFn) {
         return this.isShowingMore ? showMoreLimit : limit;
       },
 
+      refine({ helper, items }) {
+        return facetValue => {
```
bobylito: When working with @marielaures we found that actually you don't need to use items to figure out the refined value. You can get it using `getHierarchicalFacetBreadcrumb` ([doc](https://community.algolia.com/algoliasearch-helper-js/reference.html#AlgoliaSearchHelper#getHierarchicalFacetBreadcrumb)), the first item will be the one currently selected if any.

iam4x:
With pleasure 🙌

vvo:
GG this has been a long discussed widget over time since the start of IS.js. Now we just did it, because we wanted it. 💥

PR #2320, opened 2017-09-08
Fix range slider dev env

iam4x:
* Remove `rheostat` from the vendors bundled by webpack-autodll-plugin
* Import minified styles & scripts for examples


bobylito:
The tourism demo is un-broken \o/

PR #2323, opened 2017-09-09
docs: fix broken links

rayrutjes:


rayrutjes:
Commit message check does not accept valid `docs: ` format: https://github.com/algolia/instantsearch.js/blob/develop/scripts/validate-commit-msgs.sh#L47

rayrutjes:
Yes @Haroenv , but I think we should accept `docs: ` as well given it is a valid conventional commit syntax.

bobylito:
> Yes @Haroenv , but I think we should accept docs: as well given it is a valid conventional commit syntax.

We might to update tooling. AFAIK, we're using the official tools... For this kind of commit, we use `chore(docs)`.

PR #2336, opened 2017-09-14
Fix edge case slider

### LINKED ISSUE #2331: Slider V2 does not recover from no results, breaks Instantsearch
I do not understand why the slider was replaced for V2. I could spend so many hours writing about the issues we run into with this alpha version of a slider that is unusable on a touch device if I would have the time. 

Here is today's annoyance: 

The slider does not recover if there are no results found in a search

Perfectly working example with V1: https://jsfiddle.net/rata5eeo/

Slider V1 works like a charm , type Sandra into the search box => No Results => Slider goes to min value and beautifully recovers if you type in eg Sarah or at least press backspace

Braindead V2 slider: https://jsfiddle.net/3xkz0vq1/

Type Sandra into the search box => Instantsearch breaks, no way to recover from no results. The page has to be reloaded.

No need to fix this bug, there are so many more, the slider is unusable. Please just replace it.

iam4x:
Fixes #2331

### DIFF
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -28,7 +28,7 @@ class Slider extends Component {
   };
 
   get isDisabled() {
-    return this.props.min === this.props.max;
+    return this.props.min >= this.props.max;
   }
 
   handleChange = ({ min, max, values }) => {


src/components/Slider/Slider.js
```diff
@@ -28,7 +28,7 @@ class Slider extends Component {
   };
 
   get isDisabled() {
-    return this.props.min === this.props.max;
+    return this.props.min === this.props.max || this.props.min > this.props.max;
```
bobylito: Why not >=?

bobylito:
I went ahead of myself and this is of course approved change :)

PR #2337, opened 2017-09-15
feat(analytics): Push pagination

JanPetr:
**Summary**

Currently the analytics widget doesn't trigger `pushFunction` during pagination. This PR adds a parameter to allow the `pushFunction` to be triggered there.

**Result**

`pushFunction` can be triggered on pagination 🎉


src/widgets/analytics/analytics.js
```diff
@@ -64,6 +65,7 @@ function analytics(
     delay = 3000,
     triggerOnUIInteraction = false,
     pushInitialSearch = true,
+    pushPagination = false,
```
JanPetr: `false` to keep backward compatibility

PR #2342, opened 2017-09-19
fix(searchbox): use initial input value if provided in the dom

### LINKED ISSUE #2289: Searchbox with existing input replace the input by a new one
BUG/WEIRD behavior

https://jsfiddle.net/maxiloc/3gd97b2r/

When using the searchbox with an existing input, if the input had a value it disapears.
It seems that we are replacing the input no matter what: https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/search-box/search-box.js#L45

When you have a slow connection or load a website for the first time, then start typing before instantsearch.js has been loaded:
- your query will get erased at some point
- you will loose focus because the initial input is destroyed

Could we do something to keep the initial input? that would avoid weird behaviors at page load

bobylito:
Fix #2289

**Summary**

When we use an input as a container for the searchbox, we do not remove it but just upgrade it to look like a real instantsearch.js widget. However, we do not keep the value of the value if it is set.

**Result**

For consistency, we should keep the value of the input and make it the initial value of the query in instantsearch. This PR does that.

In this screenshot, the input has an initial value and it sets the value of the query.

screen shot 2017-09-19 at 18 45 26


Haroenv:
This doesn't really fix the issue described in https://github.com/algolia/instantsearch.js/issues/2289, since it's a different DOM node the events will be broken.

vvo:
I would love to be able to see the dev-novel examples on PR live so that we can all test the behavior.

> This doesn't really fix the issue described in #2289, since it's a different DOM node the events will be broken.

Can you detail a bit more this?

Haroenv:
Basically what was described in #2289 is that the searchBox widget will create a new input once it's initialising. If you have some input put as that selector, you could have event listeners, or value on it, all before instantSearch starts. Here we're still making a new input DOM Node, and thus not actually taking that original node in account. 

The only solution I see is not having the `input` as a selector, but also a `div`, just like the other elements.

bobylito:
So the exact previous was a bit different from what was described. It was not creating a new input, but was resetting the value to the one in the Helper and applying the new style. Not it reads the value of the input at init if the container is the input and overrides the state with it. This only impact the initialization phase.

bobylito:
@vvo can't make the dev novel deployed... So best for now is to checkout the pr and run `yarn run dev`.

iam4x:
I've tried to fix it and I run into the same issue as you @bobylito, we need to save the initial value somewhere in the instance of the widget in order to apply it after.

Right now when you reload the page here what happens:

sep-20-2017 14-09-19

bobylito:
> Right now when you reload the page here what happens:

I saw that it sometimes happen, sometimes not. I thought it would a problem but, it's rare that there is a input with a value, and even rarer when there are 2 searchbox and with two different way of initializing them. That's why I thought that it would not be interesting to put more complexity in order given the odds. Happy to fix it further if it happens.

bobylito:
Ok to sum up the discussion so far:
 - #2289 was created because it's inconsistent to tell the users they can use their input, and then force the original state on it even if it has a value itself
 - there is an edge case: when there are two searchboxes there is an inconsistent visual state. It happens when one of the searchbox is not providing an input with a value. This is an edge case in a context of an edge-case and we'd be happy to work on it if a user has the issue.

As a side note, this feature (using an input instead of container) is a very early one. Nowadays it can be done with the connectors. This should probably be removed in the next major version.

vvo:
LGTM and

> This should probably be removed in the next major version.

Yes, this has triggered many issues while providing very light value vs connectors today.

iam4x:
🚢

PR #2350, opened 2017-09-20
Fix range slider pips and value 0

### LINKED ISSUE #2343: Range slider pips and zero values
Hello, I have 2 issues with Instantsearch Range Slider.

1) Pips option doesn't work for me. Changing value to false doesn't remove pips. 
2) Zero values don't seem to work as intended.  If I move slider there is no way to display zero value in a search anymore.

JsFiddle: [https://jsfiddle.net/vladislav1979/nt384te0/12/](https://jsfiddle.net/vladislav1979/nt384te0/12/)

iam4x:
**Summary**

* When setting `pips: false` it was still showing pips
* Refine on range boundaries where not correctly applied

Fixes #2343

### DIFF
--- a/src/connectors/range-slider/__tests__/connectRangeSlider-test.js
+++ b/src/connectors/range-slider/__tests__/connectRangeSlider-test.js
@@ -220,4 +220,45 @@ describe('connectRangeSlider', () => {
       expect(helper.getNumericRefinement('price', '<=')).toEqual(undefined);
     }
   });
+
+  it('should refine on boundaries when no min/max defined', () => {
+    const rendering = sinon.stub();
+    const makeWidget = connectRangeSlider(rendering);
+
+    const attributeName = 'price';
+    const widget = makeWidget({ attributeName });
+
+    const helper = jsHelper(fakeClient, '', widget.getConfiguration());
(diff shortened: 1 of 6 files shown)


bobylito:
I have a question @iam4x , why does the header disappear when there is no pips? Is that a behaviour that we want?

iam4x:
I just did not put a header on the second example, it's expected behaviour 👍

src/connectors/range-slider/__tests__/connectRangeSlider-test.js
```diff
@@ -184,4 +184,40 @@ describe('connectRangeSlider', () => {
...
+
+      const renderOptions = rendering.lastCall.args[0];
+      const { refine } = renderOptions;
+      refine([10, 30]);
+
+      expect(helper.getNumericRefinement('price', '>=')).toEqual([10]);
+      expect(helper.getNumericRefinement('price', '<=')).toEqual([30]);
+      expect(helper.search.callCount).toBe(1);
+
+      refine([0, undefined]);
```
bobylito: Just for the sake of being exhaustive, maybe we could try without the bounds. And also do the third case with `[undefined, 20]`

PR #2371, opened 2017-09-21
Maintenance

shan0807:
**Summary**

**Result**


Haroenv:
Hey @shan0807, can I ask why you opened this Pull Request? v1 (maintenance) and v2 are a different documentation website, so all I can think that's useful here is the `asset_url` fix by @olance, which I think is still relevant. Otherwise this change is a bit confusing. 

Hope to get some things cleared up!

bobylito:
Maintenance is not meant to be merged into develop. As @Haroenv mentionned, if a feature is required in v2 and missing from v1 (maintenance) don't hesitate to open an issue and we'll provide a solution.

PR #2374, opened 2017-09-21
[WIP] InstantSearch::removeWidget()

iam4x:
**Summary**

The goal is to be able to mount and unmount widgets from the instantSearch instance dynamically. With such a feature we would be able to build "dynamic" search experiences.

I'm opening this PR to open the discussion around the subject and gather your feedback.

**How it works?**

InstantSearch object has now a new `removeWidgets()` method which takes as only parameter the reference to the widget you want to remove.

1. It removes it from `this.widgets[]` array
2. Call `widget.getConfiguration()` to figure out what parameters we need to remove from state of the helper.
3. Remove the refined facets from the removed widget (for example, if you have a menu and something is refined from the menu. When you unmount the menu widget, the refined value will be removed as well)
4. Re-compute all the configuration from the remaining widgets (for instance if two widgets uses the same facet and one remains)
5. Update the helper state with the new computed one and triggers a new search

**Result**

You can run the dev environment with `$ yarn dev` and go into in your browser to http://localhost:8080/?widgets=unmount where you will be able to unmount widgets via a button and see the behaviour.

**Todo**

* [x] Unmount widgets
* [x] Remove refinement from unmounted widgets
* [ ] Add widgets
* [ ] Documentation
* [ ] Tests

### DIFF
--- a/src/widgets/current-refined-values/current-refined-values.js
+++ b/src/widgets/current-refined-values/current-refined-values.js
@@ -273,15 +273,13 @@ export default function currentRefinedValues({
       specializedRenderer
     );
     return Object.assign(
-      makeCurrentRefinedValues(
-        {
-          attributes,
-          onlyListedAttributes,
-          clearAll,
-          clearsQuery,
-        },
-        { containerNode }
-      )
+      makeCurrentRefinedValues({
+        attributes,
+        onlyListedAttributes,
+        clearAll,
+        clearsQuery,
+      }),
+      { containerNode }
     );
   } catch (e) {
     throw new Error(usage);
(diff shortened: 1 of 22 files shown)


src/widgets/current-refined-values/current-refined-values.js
```diff
@@ -272,12 +272,15 @@ export default function currentRefinedValues({
     const makeCurrentRefinedValues = connectCurrentRefinedValues(
       specializedRenderer
     );
-    return makeCurrentRefinedValues({
-      attributes,
-      onlyListedAttributes,
-      clearAll,
-      clearsQuery,
-    });
+    return Object.assign(
```
vvo: Why not pass containerNode to make...* and then return it back as it was? Would be easier to understand I feel.

iam4x: `makeXXXX` is the connector which returns the widget instance. Since we decided to remove all the view logic from the connector it would be weird as a connector to have an entry for saving the container DOM element of the widget.

I have no real strong opinion on this one, it will probably change 👍

iam4x:
I'm closing this one since we are going to use #2384

PR #2377, opened 2017-09-22
fix: update call to refine

samouss:
**Summary**

The signature of the `refine` method in `PriceRanges` has [recently changed](https://github.com/algolia/instantsearch.js/commit/a5a691649bb7bd33bd0dc933ebbccf365bdc7e5c). But the `PriceRanges` instantiate an other component `PriceRangesForm` that also use the `refine` method as `prop`.

The call of `refine` inside `PriceRangesForm` hasn't be changed, and throw an error when we try to `preventDefault` on the event that is actually a number.

You can try the `PriceRanges` component [in the documentation](https://community.algolia.com/instantsearch.js/v2/widgets/priceRanges.html), press `Go` and the component will throw an error.


Haroenv:
The change with https://deploy-preview-2377--algolia-instantsearch.netlify.com/widgets/priceranges seems correct, no more page reloading

PR #2384, opened 2017-09-25
InstantSearch hot remove/add widgets

iam4x:
Implementation details: https://github.com/algolia/instantsearch.js/issues/2407

* [x] hot `search.addWidget()`
* [x] hot `search.addWidgets()`
* [x] hot `search.removeWidget()`
* [x] hot `search.removeWidgets()`
* [x] `search.dispose()`
* [x] tests
* [x] documentation


src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -163,6 +164,23 @@ export default function connectHierarchicalMenu(renderFn) {
...
         );
       },
+
+      dispose(helper) {
+        // unmount widget from DOM
+        unmountFn();
+
+        // compute nextState for the search
+        const nextState = helper
+          .getState()
```
bobylito: The `dispose` method could provide the state as with `render` and `init`?

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -163,6 +164,23 @@ export default function connectHierarchicalMenu(renderFn) {
...
+          .removeHierarchicalFacetRefinement(hierarchicalFacetName)
+          .removeHierarchicalFacet(hierarchicalFacetName);
+
+        if (nextState.maxValuesPerFacet === limit) {
+          nextState.maxValuesPerFacet = undefined;
```
bobylito: In the spirit of the usage of the SearchParameters, you should use setQueryParameter and not mutate the parameter directly. If you want to avoid yet another copy of the state, you can use `setQueryParameters` which let you pass a complete object.

src/lib/InstantSearch.js
```diff
@@ -90,6 +90,31 @@ Usage: instantsearch({
     this.widgets.push(widget);
   }
 
+  removeWidget(widget) {
+    if (
+      !this.widgets.includes(widget) ||
+      typeof widget.dispose !== 'function'
+    ) {
+      throw new Error('Widget definition is missing dispose method');
```
bobylito: I would go for something like "The widget you tried to remove does not implement the `dispose` method, therefore it is not possible to remove this widget" in order to provide more information and make it less abstract.

src/widgets/clear-all/clear-all.js
```diff
@@ -51,6 +51,10 @@ const renderer = ({
   );
 };
 
+const disposer = containerNode => () => {
```
bobylito: Why not inline this function or make the implementation part of the utils function?

src/widgets/search-box/search-box.js
```diff
@@ -113,6 +113,12 @@ const renderer = ({
   }
 };
 
+const disposer = containerNode => () => {
+  const range = document.createRange(); // IE10+
```
bobylito: This one may be more tricky because the input can be provided or built by the widget.

iam4x: Should we replace it by the initial input? I was expecting user should want to remove the widget/input in any cases from the DOM.

bobylito: That's a good question. I would say that since we enhance the checkbox, we should leave it in the state it was provided. But it's a bit tedious to do... No need to worry to much about that though.

vvo:
Happy to review in detail as soon as you can put some API docs like you did in the previous PR inside a comment

src/lib/InstantSearch.js
```diff
@@ -90,32 +90,38 @@ Usage: instantsearch({
...
 
-  removeWidget(widget) {
-    if (
-      !this.widgets.includes(widget) ||
-      typeof widget.dispose !== 'function'
-    ) {
-      throw new Error(
-        'The widget you tried to remove does not implement the dispose method, therefore it is not possible to remove this widget'
-      );
+  removeWidget(widgets) {
```
iam4x: Should rename it `removeWidgets()` @bobylito WDYT?

bobylito: In my experience I like to have 2 API's and not have polymorhpic methods. In this case, I like to have `removeWidget` which accepts a single widget and `removeWidgets` that accepts an array of widget. The implementation of `removeWidget` is trivial because it can use `removeWidgets`. WDYT?

iam4x: Good 👍

vvo:
Checking, what's the final API features of removeWidget (how to use it?)

iam4x:
I have written a little documentation, where do you think we should update it also @bobylito?

And I have updated the PR description with the tasks done 👍

src/lib/InstantSearch.js
```diff
@@ -82,12 +82,108 @@ Usage: instantsearch({
...
+        });
+
+        this.helper.setState(this.searchParameters);
+      }
+    });
+
+    // no need to retrigger a search if we don't have any widgets left
+    if (this.widgets.length > 0) {
+      this.helper.search();
+    }
```
iam4x: clean URL on dispose as well

PR #2391, opened 2017-09-26
chore(doc): publish doc during the build process

bobylito:
The doc is not auto updated anymore using TravisCI because we don't have to wait for
jsDeliver to publish the packages anymore. This PR makes the release script to publish
the doc just after we've pushed to NPM.


iam4x:
LGTM 🚢

PR #2392, opened 2017-09-26
Fix the connector-docs link + typo

Meschreiber:
Previously the link for more info on connectors led to a 404.

### DIFF
--- a/docgen/src/guides/customization.md
+++ b/docgen/src/guides/customization.md
@@ -261,6 +261,8 @@ With this steps we introduced a way to provide custom parameters:
 
 And voilà, we have covered how to write a simple custom widget using connectors 🎉 !
 
-Now you can [read more about connectors](/connectors.html) in the docs and create more advance usages with the same simplicity.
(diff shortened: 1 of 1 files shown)


Haroenv:
Same here, shouldn't we make the absolute links work with a `<base>` tag maybe?

bobylito:
Relative links work just fine.

Haroenv:
This should also be a relative link then 🤔

Meschreiber:
The relative link was going to https://community.algolia.com/instantsearch.js/connectors.html maybe because it was written as (../connectors.html)

Meschreiber:
PS. Will follow branch naming conventions from now, sorry!

Haroenv:
@Meschreiber, was fixed in another PR, thanks a lot, branch naming doesn't matter much for a patch 😄

bobylito:
Yes indeed branch naming is not enforced. Thanks for contributing that is the most important ❤️

PR #2393, opened 2017-09-26
Add link for custom-widget docs.

Meschreiber:
Adds a link to custom widget documentation.

### DIFF
--- a/docgen/src/connectors.md
+++ b/docgen/src/connectors.md
@@ -16,7 +16,7 @@ all the logic needed for making search widgets. Each one of them is specialized
 to make a certain type of widget.
 
 If you want to create a type of widget that is not available, you should then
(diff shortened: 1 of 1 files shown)


docgen/src/connectors.md
```diff
@@ -16,7 +16,7 @@ all the logic needed for making search widgets. Each one of them is specialized
 to make a certain type of widget.
 
 If you want to create a type of widget that is not available, you should then
-create a [custom widget]()
+create a [custom widget](https://community.algolia.com/instantsearch.js/v2/guides/customization.html).
```
Haroenv: Actually this should better be a relative link so that it works in development 🤔

bobylito: Indeed.

Issue #2388, opened 2017-09-26
Why does instantsearch require React?


millansingh:
I've installed instantsearch via npm, and it installed fine. However, when I include it in a script and compile the script (with Gulp/Browserify), I get console errors saying that it can't find the 'react' module in various places (like 50 places).

I asked my boss, and he told me to just install react, but I have to wonder why this would use react in the first place, especially when there's a whole separate library specifically for react.

Can someone provide some clarification here on why react is a dependency?

Haroenv:
preact is used internally (with an alias to react), but you should normally not be aware of this as an end-user. Could you share your configuration that causes these error messages so that you're correctly using the production (preact) version? 

Thanks!

iam4x:
This is related to #2279 

Until we come up with a real fix, you can try to use https://github.com/tleunen/babel-plugin-module-resolver and resolve `react` to `preact-compat`

millansingh:
Haroenv, it's a default installation of instantsearch. The site does not use React or Preact for anything.

millansingh:
I can't provide something reproducible, since it's a work project. However, I can tell you that we're building front-end files with Gulp and it's within a Laravel instance, using some of Laravel's Elixir Gulp stuff.

iam4x:
Another workaround would be to use the UMD build in the meantime @millansingh. How do you import InstantSearch.js into your code?

millansingh:
It's imported by a simple require statement, and downloaded as an npm package.

hallcyon11:
this is still happening, i have to install react and react-dom for it to work
```
Error: Cannot find module 'react' from '/Users/asdf/node_modules/instantsearch.js/dist-es5-module/src/widgets/menu-select'
    at /Users/asdf/node_modules/browser-resolve/node_modules/resolve/lib/async.js:46:17
    at process (/Users/asdf/node_modules/browser-resolve/node_modules/resolve/lib/async.js:173:43)
    at ondir (/Users/asdf/node_modules/browser-resolve/node_modules/resolve/lib/async.js:188:17)
    at load (/Users/asdf/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
    at onex (/Users/asdf/node_modules/browser-resolve/node_modules/resolve/lib/async.js:92:31)
    at /Users/asdf/node_modules/browser-resolve/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:152:21)
```

[...]

PR #2395, opened 2017-09-26
Remove React occurrences in the codebase

### LINKED ISSUE #2279: Importing `instantsearch.js/dist-es5-module` in a non-react app

It's a bug, when you try to build an application for the browser and it's not using React in it's own dependencies you will get this kind of errors:

```
ERROR in ./node_modules/instantsearch.js/dist-es5-module/src/widgets/range-slider/range-slider.js
Module not found: Error: Can't resolve 'react' in '/Users/iam4x/github/experiement-tree-shaking/node_modules/instantsearch.js/dist-es5-module/src/widgets/range-slider'
 @ ./node_modules/instantsearch.js/dist-es5-module/src/widgets/range-slider/range-slider.js 8:13-29
 @ ./node_modules/instantsearch.js/dist-es5-module/src/widgets/index.js
 @ ./node_modules/instantsearch.js/dist-es5-module/src/lib/main.js
 @ ./node_modules/instantsearch.js/dist-es5-module/index.js
 @ ./index.js
```

We should completely replace any occurrence of `react` in our code with the dependency we are using, which is `preact` now.

iam4x:
Attempt to fix #2279 #2388

I replace `React` and `ReactDOM` by `preact-compat` and fixed the tests which were failing because of the update.

I've also replaced Rheostat `React` with `preact-compat` in the fork.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,7 +6,14 @@
         "react",
         "stage-2"
       ],
-      "plugins": ["babel-plugin-rewire"]
+      "plugins": [
+        "babel-plugin-rewire",
+        ["module-resolver", {
+          "alias": {
+            "preact-compat": "react"
+          }
+        }]
+      ]
     },
     "development": {
       "presets": [
--- a/src/components/Pagination/Pagination.js
+++ b/src/components/Pagination/Pagination.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React from 'preact-compat';
 import forEach from 'lodash/forEach';
 import defaultsDeep from 'lodash/defaultsDeep';
 import { isSpecialClick } from '../../lib/utils.js';
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "babel-core": "6.26.0",
     "babel-eslint": "8.0.0",
     "babel-loader": "7.1.2",
+    "babel-plugin-module-resolver": "^2.7.1",
     "babel-plugin-rewire": "1.1.0",
     "babel-plugin-syntax-jsx": "6.18.0",
     "babel-plugin-transform-react-constant-elements": "6.23.0",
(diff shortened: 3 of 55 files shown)


.babelrc
```diff
@@ -6,7 +6,14 @@
         "react",
         "stage-2"
       ],
-      "plugins": ["babel-plugin-rewire"]
+      "plugins": [
+        "babel-plugin-rewire",
+        ["module-resolver", {
+          "alias": {
+            "preact-compat": "react"
```
bobylito: Why do we alias if we're using preact directly?

src/components/Pagination/Pagination.js
```diff
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types';
-import React from 'react';
+import React from 'preact-compat';
```
bobylito: While we are at it, maybe we could just `import PReact ...` no?

iam4x: It would imply some changes in the React component we wrote. The API of preact is a bit different.

bobylito: I meant to use `import PReact from 'preact-compat` but since there is a different API then I guess it's fine like that.

Haroenv: you'd need to change the jsx calls too then (we can do that though)

.babelrc
bobylito: But the real output that the user will have is the one from preact-compat right? So might as well use this one in the snapshots?

src/components/Pagination/Pagination.js
bobylito: Probably not worth it I guess :)

.babelrc
iam4x: I need to take at least half a day to update theses snapshots, wan't to make sure everything is working for every widgets, let's plan this?

bobylito: https://github.com/algolia/instantsearch.js/issues/2405 ok

PR #2406, opened 2017-09-27
Merge 2.2 into develop

bobylito:
Let's prepare the next version.

It contains some cool new stuff:
 - the analytics now also sends the page (and is triggered accordingly)
 - the hitsPerPage let you select a default page
 - a new menu select widget \o/


bobylito:
The problem with build is a commit format from one that was created for the release of the beta. I'll try to fix it before the release.

bobylito:
Let's move this to monday. I wouldn't like to do support on a sunday :)

bobylito:
Merging because bad commit format...

PR #2408, opened 2017-09-27
Fix range slider boundaries

### LINKED ISSUE #2386: RangeSlider ignores min / max boundaries
If you set boundaries on a `RangeSlider` widget, there are ignored. In the given example the boundaries are set to `2` and `4`. The UI is align with the options but in the response we still have the movies with a rating greater than `4`.

The boundaries are also ignored when we change the current values of the slider, then reset to the initial state. For example set the slider minimum to `3`, reset it, make a research for `Bad Moon` and you will have some movies with rating lower than `2`.

https://jsfiddle.net/tksyn43f/

samouss:
**Summary**

Fix #2386 

- The boundaries were ignored in `getConfiguration` for the first render
- They were ignored when a refinement was triggered and reset

**Result**

You can play with the `RangeSlider` on `dev-novel` (some stories added)

### DIFF
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -2,9 +2,8 @@ import times from 'lodash/times';
 import range from 'lodash/range';
 import has from 'lodash/has';
 
-import PropTypes from 'prop-types';
-
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
 import Rheostat from 'rheostat';
 import cx from 'classnames';
 
@@ -13,7 +12,7 @@ import Pit from './Pit.js';
 import autoHideContainerHOC from '../../decorators/autoHideContainer.js';
 import headerFooterHOC from '../../decorators/headerFooter.js';
 
-class Slider extends Component {
+export class Slider extends Component {
   static propTypes = {
     refine: PropTypes.func.isRequired,
     min: PropTypes.number.isRequired,
@@ -31,17 +30,9 @@ class Slider extends Component {
     return this.props.min >= this.props.max;
   }
 
-  handleChange = ({ min, max, values }) => {
-    // when Slider is disabled, we alter `min, max` values
-    // in order to render a "disabled state" Slider. Since we alter
-    // theses values, Rheostat trigger a "change" event which trigger a new
-    // search to Algolia with wrong values.
+  handleChange = ({ values }) => {
     if (!this.isDisabled) {
-      const { refine } = this.props;
-      refine([
-        min === values[0] ? undefined : values[0],
-        max === values[1] ? undefined : values[1],
-      ]);
+      this.props.refine(values);
     }
   };
 
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -12,7 +12,7 @@ import Pit from './Pit.js';
 import autoHideContainerHOC from '../../decorators/autoHideContainer.js';
 import headerFooterHOC from '../../decorators/headerFooter.js';
 
-export class Slider extends Component {
+export class RawSlider extends Component {
   static propTypes = {
     refine: PropTypes.func.isRequired,
     min: PropTypes.number.isRequired,
@@ -111,4 +111,4 @@ export class Slider extends Component {
   }
 }
 
-export default autoHideContainerHOC(headerFooterHOC(Slider));
+export default autoHideContainerHOC(headerFooterHOC(RawSlider));
--- a/src/connectors/range-slider/__tests__/connectRangeSlider-test.js
+++ b/src/connectors/range-slider/__tests__/connectRangeSlider-test.js
@@ -482,4 +482,319 @@ describe('connectRangeSlider', () => {
       expect(actual).toEqual(expectation);
     });
   });
+
+  describe('_refine', () => {
+    const attributeName = 'price';
+    const rendering = () => {};
+
+    it('expect to refine min and max from empty state', () => {
+      const values = [0, 500];
+      const widget = connectRangeSlider(rendering)({ attributeName });
+      const helper = {
+        getNumericRefinement: jest.fn(() => []),
+        addNumericRefinement: jest.fn(),
+        clearRefinements: jest.fn(),
+        search: jest.fn(),
+      };
+
+      widget._refine(helper)(values);
+
+      expect(helper.addNumericRefinement).toHaveBeenCalledWith(
+        attributeName,
+        '>=',
+        0
+      );
+
+      expect(helper.addNumericRefinement).toHaveBeenCalledWith(
+        attributeName,
+        '<=',
+        500
+      );
+
+      expect(helper.clearRefinements).toHaveBeenCalled();
+      expect(helper.search).toHaveBeenCalled();
+    });
+
+    it('expect to refine min and max from non empty state', () => {
+      const values = [20, 480];
+      const widget = connectRangeSlider(rendering)({ attributeName });
+      const helper = {
+        getNumericRefinement: jest.fn(
+          (_, operation) => (operation === '>=' ? 0 : 500)
+        ),
+        addNumericRefinement: jest.fn(),
+        clearRefinements: jest.fn(),
+        search: jest.fn(),
+      };
+
+      widget._refine(helper)(values);
+
+      expect(helper.addNumericRefinement).toHaveBeenCalledWith(
+        attributeName,
+        '>=',
+        20
+      );
+
+      expect(helper.addNumericRefinement).toHaveBeenCalledWith(
+        attributeName,
+        '<=',
+        480
+      );
+
+      expect(helper.clearRefinements).toHaveBeenCalled();
+      expect(helper.search).toHaveBeenCalled();
+    });
+
+    it('expect to refine min from empty state', () => {
+      const values = [0, null];
+      const widget = connectRangeSlider(rendering)({ attributeName });
+      const helper = {
+        getNumericRefinement: jest.fn(() => []),
+        addNumericRefinement: jest.fn(),
+        clearRefinements: jest.fn(),
+        search: jest.fn(),
+      };
+
+      widget._refine(helper)(values);
+
+      expect(helper.addNumericRefinement).toHaveBeenCalledWith(
+        attributeName,
+        '>=',
+        0
+      );
+
+      expect(helper.addNumericRefinement).not.toHaveBeenCalledWith(
+        attributeName,
+        '<=',
+        expect.any(Number)
+      );
+
+      expect(helper.clearRefinements).toHaveBeenCalled();
+      expect(helper.search).toHaveBeenCalled();
+    });
+
+    it('expect to refine min from non empty state', () => {
+      const values = [20, 500];
+      const widget = connectRangeSlider(rendering)({ attributeName });
+      const helper = {
+        getNumericRefinement: jest.fn(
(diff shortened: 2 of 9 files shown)


src/connectors/range-slider/connectRangeSlider.js
```diff
@@ -186,28 +210,18 @@ export default function connectRangeSlider(renderFn) {
...
-          facet !== undefined && facet.stats !== undefined
-            ? facet.stats
-            : {
-                min: null,
-                max: null,
-              };
-
-        if (userMin !== undefined) stats.min = userMin;
-        if (userMax !== undefined) stats.max = userMax;
+        const facet = find(facetsFromResults, _ => _.name === attributeName);
```
iam4x: `(_) => _.` is really not explicit, and you are using `lodash/find` you can refactor this in a more explicit way:

```js
const facet = find(facetsFromResults, { name: attributeName })
```

src/connectors/range-slider/connectRangeSlider.js
```diff
@@ -101,26 +101,33 @@ export default function connectRangeSlider(renderFn) {
...
-        const conf = {
+      getConfiguration: currentConfiguration => {
+        const configuration = {
           disjunctiveFacets: [attributeName],
         };
 
-        const hasUserBounds = userMin !== undefined || userMax !== undefined;
+        const isMinBound = userMin !== undefined && userMin !== null;
+        const isMaxBound = userMax !== undefined && userMax !== null;
+        const boundsDefined = isMinBound || isMaxBound;
```
iam4x: I think we should be able to refactor more and use this:

```js
const boundsDefined = isSafeInteger(userMin) || isSafeInteger(userMax);
```
http://devdocs.io/lodash~4/index#isSafeInteger

src/connectors/range-slider/connectRangeSlider.js
```diff
@@ -131,7 +131,7 @@ export default function connectRangeSlider(renderFn) {
         return configuration;
       },
 
-      _getCurrentRefinement(helper, stats = {}) {
+      _getCurrentRefinement: (helper, stats = {}) => {
```
iam4x: Do we explicitly need a fat arrow function here?

src/connectors/range-slider/connectRangeSlider.js
```diff
@@ -238,7 +238,7 @@ export default function connectRangeSlider(renderFn) {
 
         renderFn(
           {
-            refine: this._refine,
+            refine: this._refine(helper),
```
iam4x: I'm not a super fan to re instantiate a new function on every render, the helper reference won't change in between two calls we can cache this.

src/components/Slider/Slider.js
```diff
@@ -13,7 +12,7 @@ import Pit from './Pit.js';
 import autoHideContainerHOC from '../../decorators/autoHideContainer.js';
 import headerFooterHOC from '../../decorators/headerFooter.js';
 
-class Slider extends Component {
+export class Slider extends Component {
```
iam4x: Did you need this in order to query the Rheostat component itself in the tests?

src/connectors/range-slider/connectRangeSlider.js
```diff
@@ -34,9 +34,9 @@ Full documentation available at https://community.algolia.com/instantsearch.js/c
 
 /**
  * @typedef {Object} RangeSliderRenderingOptions
- * @property {function({min: number, max: number})} refine Sets a range to filter the results on. Both values
+ * @property {function(Array<number, number>)} refine Sets a range to filter the results on. Both values
  * are optional, and will default to the higher and lower bounds.
- * @property {{min: number, max: number}} numeric Results bounds without the current range filter.
+ * @property {{min: number, max: number}} range Results bounds without the current range filter.
```
iam4x: Can we "patch" this on our side to avoid breaking changes to the public API?

src/connectors/range-slider/connectRangeSlider.js
```diff
@@ -121,7 +121,7 @@ export default function connectRangeSlider(renderFn) {
         } else if (stats.min !== undefined && stats.min !== null) {
           min = stats.min;
         } else {
-          min = -Infinity;
+          min = 0;
```
iam4x: I don't get this change, when you want to select to complete range of results keep in mind they're might be lower than zeros values. Or am I missing something?

iam4x: My bad, it was fixed two commits later: https://github.com/algolia/instantsearch.js/pull/2408/commits/ca2ea036053aeed375db6579b84639b04df23aa6

src/connectors/range-slider/connectRangeSlider.js
```diff
@@ -105,28 +105,24 @@ export default function connectRangeSlider(renderFn) {
       },
 
       _getCurrentRefinement: (helper, stats = {}) => {
-        const minValues = helper.state.getNumericRefinement(
-          attributeName,
-          '>='
-        );
+        const [minValue] =
+          helper.state.getNumericRefinement(attributeName, '>=') || [];
```
iam4x: Me gusta 👍

iam4x:
Nice job @samouss 🥇

src/components/Slider/Slider.js
samouss: Not really but it's more easy to do with the export. It's avoid something like:

```js
shallow(<Enhance {...props} />)
    .find('Slider-HeaderFooter')
    .shallow()
    .find('Slider')
    .shallow()
    .find('Rheostat')
    .simulate('change', {
      values: [0, 100],
    });
```

src/connectors/range-slider/connectRangeSlider.js
samouss: It's not a breaking change, just a fix in the documentation.

- The `refine` function is already called with [an array and not an object](https://github.com/algolia/instantsearch.js/pull/2408/files#diff-6594f5647a134a790a0668875d7b1045L131).
- The property `numeric` doesn't exist on the render function, [it's name is `range`](https://github.com/algolia/instantsearch.js/pull/2408/files#diff-6594f5647a134a790a0668875d7b1045L210).

src/connectors/range-slider/connectRangeSlider.js
samouss: 👍

src/connectors/range-slider/connectRangeSlider.js
samouss: Ok for me. 
But if the user set min / max with a floating number `isSafeInteger` will fail. 
Maybe we should go for `isFinite` ? Or we consider that passing floating numbers for bounds is not valid.

src/connectors/range-slider/connectRangeSlider.js
samouss: Later I add also `range` as parameter, both can be cached but I prefer to rely on arguments for function instead of `this._cachedRange`. IMHO it's easier to reason about.

samouss:
@bobylito I fixed the bug, but I still need to dig a little bit to find a way to allow `refine` on first rendering. Because I based the refinement on the `range` and on the first rendering we have a range of [0, 0] (default value). So, the refinement can't be done properly for the `max` value. It will always be upper than the max bound.

src/components/Slider/Slider.js
bobylito: That's something we do for other components. We've prefixed the component without the wrappers with `Raw` like ClearAll (the one with the headers and autohidecontainer) and RawClearAll (the one without).

src/connectors/range-slider/connectRangeSlider.js
```diff
@@ -76,108 +80,146 @@ export default function connectRangeSlider(renderFn) {
     };
 
     return {
-      getConfiguration: originalConf => {
-        const conf = {
-          disjunctiveFacets: [attributeName],
-        };
+      _isAbleToRefine(min, max) {
+        const isMinValid = _isFinite(min);
```
bobylito: `_isAbleToRefine` is called once with minBounds and maxBounds as the values, right? If yes this seems superfluous.

src/connectors/range-slider/connectRangeSlider.js
```diff
@@ -76,108 +80,146 @@ export default function connectRangeSlider(renderFn) {
...
-          min = min[0];
+      _getCurrentRange(stats = {}) {
+        let min;
+        if (isMinBounds) {
+          min = minBounds;
+        } else if (_isFinite(stats.min)) {
+          min = stats.min;
         } else {
-          min = -Infinity;
+          min = 0;
```
bobylito: I don't get why 0 could be a good value here. Can you explain?

src/connectors/range-slider/connectRangeSlider.js
```diff
@@ -76,108 +80,146 @@ export default function connectRangeSlider(renderFn) {
...
+          }
+        }
+
+        return configuration;
+      },
+
+      init({ helper, instantSearchInstance }) {
+        const stats = {};
+        const range = this._getCurrentRange(stats);
+        const { min, max } = this._getCurrentRefinement(helper);
```
bobylito: `_getCurrentRefinement` should return an array so we don't have to extract the values and create an array.

src/connectors/range-slider/__tests__/connectRangeSlider-test.js
```diff
@@ -250,15 +291,649 @@ describe('connectRangeSlider', () => {
...
+    });
+  });
+
+  describe('_getCurrentRefinement', () => {
+    const attributeName = 'price';
+    const rendering = () => {};
+
+    it('expect to return default refinement', () => {
+      const widget = connectRangeSlider(rendering)({ attributeName });
+      const helper = {
```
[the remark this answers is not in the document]
samouss: No problem, I will update it. 
I ping you if I have trouble.

src/connectors/range-slider/connectRangeSlider.js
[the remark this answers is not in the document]
samouss: I finally use the solution pushed by @iam4x 

```js
const facet = find(facetsFromResults, { name: attributeName });
```

src/connectors/range-slider/connectRangeSlider.js
iam4x: True my bad, thought we were calling `refine({ min, max })`

src/connectors/range-slider/connectRangeSlider.js
samouss: I tried to kept the compatibility.

In the previous version the range was compute with the stats object in the [init](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/range-slider/connectRangeSlider.js#L169) and the [render](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/range-slider/connectRangeSlider.js#L194) functions. Then thoses values are rounded when passed to the rendering function. But it might happen that the stats object can be `{ min: null, max: null }` and `null` is rounded to `0`.

The `0` case is used in `RangeSlider` for [hide it](https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/range-slider/range-slider.js#L44) when there are nothing to display. In upcoming version we can imagine to pass a `disabled` property instead of rely on range values.

samouss:
Now the values are clamp inside the range for `RangeSlider`, avoids some rendering trouble (notice the slider bar when the range changes).

**Before:**
with-out-range

**After:**
without-out-range

bobylito:
Good job :)

Issue #2139, opened 2017-05-15
[currentRefinements] clearsQuery


bobylito:
## Current behaviour

When `clearsQuery` is `true`, the button for clearing the refinements also clears the query.

However if there are no other refinements, the button doesn't appear. Also the query is not considered a current refinement, therefore even though it would cleared by the button it doesn't appear.

## Proposed fix

Make the clearsQuery trigger the following changes:
 - the clear all button clears the query too (and is visible when only the query is set)
 - the query appears as a refinement

bobylito:
cc @iam4x @mthuret @vvo @rayrutjes

Haroenv:
What about making it show up when there’s a query, but not showing that in the refinements, it would look a bit weird.

iam4x:
I'm good with your proposed fix 👍

vvo:
I am fine with your proposal @bobylito. And if the user want to hide the query on display, what would he have to do?

bobylito:
> And if the user want to hide the query on display, what would he have to do?

We have a white list mechanism which could be used. However it doesn't fit the usage of refinements that are not attribute based (query or any query parameters). Maybe we could extend the range of possibilities here?

bobylito:
Ok that's too much of an edge case. Let's implement the clearsQuery that will only make the clear all button clear the query without any rendering of the query in the current refinements.

fabregas4:
Hey guys, customer here... good discussion. Would including the search query as a refinement really be that weird? Consider the attached screenshot - I think this would be handy.
![alg-query-as-refinement](https://cloud.githubusercontent.com/assets/2221888/26114996/65d290dc-3a91-11e7-917e-8274ce23d1f9.png)

bobylito:
> And if the user want to hide the query on display, what would he have to do?

Actually after thinking about it, if you've enabled `clearsQuery` isn't it reasonnable to have the whole package (the button is enabled when the query is not empty and the query appears in the refinements)?

vvo:
I did not quite get what you proposed. If the question is is it ok to have the query appears in refinements then I would say yes, as @fabregas4 proposed and showed us.

With that said, let's see if anyone wants to hide the query afterwards and deal with that once it shows as a need/issue.

bobylito:
What I propose is the following: when you choose to activate `clearsQuery`, we should display the query. On the other hand, if `clearsQuery` is `false` we don't display the query. This brings consistency to the feature: what you can see is what gets cleared.

[...]

PR #2414, opened 2017-09-29
fix(clear-all): display the query when clearQuery is true

ronanlevesque:
This PR provides a fix for https://github.com/algolia/instantsearch.js/issues/2139.


dev/app/init-builtin-widgets.js
```diff
@@ -269,6 +269,25 @@ export default () => {
...
+          );
+        },
+        {
+          searchParameters: {
+            disjunctiveFacetsRefinements: { brand: ['Apple', 'Samsung'] },
+            disjunctiveFacets: ['brand'],
+          },
+        }
+      )
     );
```
bobylito: Since you're modifying the template, can you add a story with the different kind of parameters? So that we can check that everything is going well?

bobylito:
Congratulations on your first PR @ronanlevesque :)

bobylito:
I think you can add a test or two, to check the behavior of the clearsQuery parameter. :)

Haroenv:
When I open the dev-novel, it seems like they don't have the type of refinement before it 🤔 

https://deploy-preview-2414--algolia-instantsearch.netlify.com/dev-novel/?selectedStory=CurrentRefinedValues.with%20clearsQuery

bobylito:
@Haroenv they never had, that's why we have labels. https://community.algolia.com/instantsearch.js/v2/dev-novel/?selectedStory=CurrentRefinedValues.with%20header

dev/webpack.config.js
```diff
@@ -13,7 +13,7 @@ const __DEV__ = process.env.NODE_ENV === 'development';
 const clean = arr => arr.filter(item => item !== false);
 
 module.exports = {
-  devtool: __DEV__ ? 'cheap-eval-source-map' : 'source-map',
+  devtool: 'source-map',
```
iam4x: Why did you introduce this change?

ronanlevesque: From what I recall we had some issues with using this and breakpoints with @bobylito, I can roll this back if needed :)

iam4x: Hmm can we check together? Because this speeds-up re-build time for dev. I had no issues with breakpoints because webpack output sourcemaps also.

src/lib/utils.js
```diff
@@ -171,8 +171,17 @@ function getRefinement(state, type, attributeName, name, resultsFacets) {
   return res;
 }
 
-function getRefinements(results, state) {
-  const res = [];
+function getRefinements(results, state, clearsQuery) {
+  const res =
+    clearsQuery && state.query !== ''
```
iam4x: Should we test also for:

```js
clearsQuery && state.query && state.query.trim()
```
So we ensure it's really not an empty string?

src/widgets/current-refined-values/defaultTemplates.js
```diff
@@ -1,15 +1,29 @@
...
+}) {
+  const computedOperator = operator ? displayOperator : '';
+  const computedLabel = label
+    ? `${label} ${computedOperator || ':'} `
+    : computedOperator;
+  const computedCount = query
+    ? ''
+    : `<span class="${cssClasses.count}">${count === undefined
+        ? 0
+        : count}</span>`;
```
iam4x: This should not eslint pass since we lint for `no-nested-ternary`

ronanlevesque: It seems it didn't raise any error on linting, but I'll replace it 👍

stianmartinsen:
Hey guys! I have a situation where I want to display the search query in the currentRefinedValues widget, but at the same time I want to whitelist the other attributes that are listed (using  `onlyListedAttributes: true`). When this is enabled, the query doesn't show up even if `clearsQuery` is set to true.

Any way to display the query when `onlyListedAttributes` is enabled?

Haroenv:
Sorry, can you make a new issue and a codesandbox describing this behaviour please @stianmartinsen . I don't think I understand it without context like this. Thanks!

There's a template here: https://codesandbox.io/s/x3q0p0vnko

PR #2421, opened 2017-10-02
chore(dependencies): declare docgen deps as devDeps

### LINKED ISSUE #2420: Ensure docgen dependencies are pinned

Bug

When coming to the project and doing `yarn dev` you might end up with different deps than other coworkers (even with yarn.lock, explanation: https://github.com/algolia/react-element-to-jsx-string/pull/93#issuecomment-316967321). Thus reducing the reproducibility of the repo.

Doing `yarn dev` should lead the same result everytime.

Define dependencies from docgen as devDependencies so renovate take them

vvo:
fixes #2420

**Summary**

This will force renovate to come and pin our docgen dependencies, ensuring a stable env

**Result**

### DIFF
--- a/docgen/package.json
+++ b/docgen/package.json
@@ -7,7 +7,7 @@
     "dev": "babel-node start.js",
     "build": "babel-node build.js"
   },
-  "dependencies": {
+  "devDependencies": {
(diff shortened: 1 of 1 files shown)


bobylito:
❤️  @vvo

PR #2422, opened 2017-10-02
fix(webpack.dev): sourcemaps in dev

iam4x:
Sourcemaps are buggy in chrome, let's use a less aggressive option in webpack to restore them until it's fixed in webpack.

-> Impacts a slower re-build time


vvo:
nitpick, this is a "chore(dev)" since there's no actual fix for the consumers of the lib, no big deal it will show up in changelog though (but that can be modified)

Issue #2397, opened 2017-09-27
HitsPerPageSelector root css class being applied twice


Corys8646:
Apologies if this is intended behavior, but the docs make me believe this is possibly a bug. 
Using V2 of Instantsearch and adding a HitsPerPage selector, also adding a css class to the root cssClasses Object.

Problem is, the class is being added to the `<select>` item as well as the parent div.

Trying to add a bootstrap form-control to the select element.

I've got a fiddle here: <https://jsfiddle.net/Corys8646/8mhxqayf/1/>

bobylito:
Hi @Corys8646 thanks for opening this issue. Are you reporting this issue for v2? Also I can't find your link to the jsFiddle you made :)

bobylito:
This should have been fixed in the 2.2 release :)

PR #2423, opened 2017-10-02
fix(selector): root classname is applied twice

### LINKED ISSUE #2396: numericSelector cssClasses option for root applies to a parent div as well as the child <select>
```js
search.addWidget(
    instantsearch.widgets.numericSelector({
        container: '#numericSelector',
        attributeName: 'thing.id',
        options: [{label: 'name of thing', value: 1}],
        cssClasses: {
            root: 'form-control'
        }
    })
);
```

outputs the following HTML:
```html
<div>
    <div class="ais-root ais-numeric-selector form-control">
        <div class="ais-body">
            <select class="ais-numeric-selector form-control">
                <option class="ais-numeric-selector--item" value="1">name of thing</option>
            </select>
        </div>
    </div>
</div>
```

According to the documentation, it should only be applied to the select element. This is problematic in this scenario because I am using with bootstrap and it is making the output undesirable

bobylito:
The implementation of the React component `Selector` was applying the `root` css
classes that were already applied by the headerfooter. In order to
prevent regressions and let user customize the <select>, this patch adds
a new key to customize cssClasses in Selector bases widgets (`select`), and this
new key has the root value to avoid breaking the current implementations.

Fix #2396 #2397


bobylito:
I need push them 🤣

bobylito:
Good thing that we have argos :D

bobylito:
The select was used but not implemented. The change is expected on argos.

src/components/__tests__/__snapshots__/Selector-test.js.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`Selector should render <Selector/> with numbers 1`] = `
 <select
-  className="custom-root"
+  className={undefined}
```
vvo: Should we update this test to not have undefined inside it? And check if select css class is actually read?

src/widgets/sort-by-selector/sort-by-selector.js
```diff
@@ -31,13 +31,14 @@ const usage = `Usage:
...
   indices,
-  [cssClasses.{root,item}={}],
+  [cssClasses.{root,select,item}={}],
   [autoHideContainer=false]
 })`;
 
 /**
  * @typedef {Object} SortByWidgetCssClasses
- * @property {string|string[]} [root] CSS classes added to the parent `<select>`.
+ * @property {string|string[]} [root] CSS classes added to the outer `<div>`.
```
vvo: 👍

PR #2425, opened 2017-10-02
docs(org): default to v2 homepage

### LINKED ISSUE #2236: Make V2 the default homepage?
V2 is officially out. Should we make the https://community.algolia.com/instantsearch.js/ homepage the V2 one and then display a banner for V1 docs (or just a top menu item so it's less intrusive)?

It's not a big priority. But could still be confusing for newcomers searching for InstantSearch.js and ending up on the V1 website.
It's very easy to dismiss a banner because it looks a bit like an ad ("This website use cookies..." effect)

The hierarchy could be:
- / => V2 homepage
- /v2/* v2 docs
- /* others, v1 docs (so there's no need to handle 301 etc. for now)

The goal being to push V2 even more.

vvo:
**Summary**

This PR changes the default homepage of InstantSearch.js to V2.

While the V2 docs/* will still be behind
https://community.algolia.com/instantsearch.js/v2/.

Fixes #2236

We may want to add a link to V1 docs so that:

people still on V1 can access the docs easily
people looking for V1 docs content when using v2 can do it (we had this feedback recently, let's dig it)

**Result**


docgen/src/data/communityHeader.json
```diff
@@ -42,7 +42,7 @@
       "label": "InstantSearch.js",
       "logo": "",
       "caret": "",
-      "url":"https://community.algolia.com/instantsearch.js/v2/",
+      "url":"https://community.algolia.com/instantsearch.js/",
```
vvo: This could be a communityHeader.js and reuse the configuration to make this tailored to the env rather than static.

vvo:
We may want to add a link to V1 docs so that:
- people still on V1 can access the docs easily
- people looking for V1 docs content when using v2 can do it (we had this feedback recently, let's dig it)

vvo:
OK so we are good to go here, once this is merged I can do a deploy test and if it fails, EASY we just git revert inside the gh-pages, boom no one notices :)

vvo:
I will then just activate the cloudflare rules that will redirect old v1 pages to /v1/*

vvo:
If this links works: https://community.algolia.com/instantsearch.js/v1/

You can merge this

Shipow:
yes but @vvo we need https://community.algolia.com/instantsearch.js/ to go on V2 as well

Shipow:
also this link to V2 leads to 404
image

vvo:
@shipow as we discussed homepage of is.js will be V2 by default rather than a 301

vvo:
/ is v2
/v2/* are docs
/v2/ redirects to / (which is v2)
/v1/
Old pages redirects to /v1/*

vvo:
Can you send the link leading to a 404?

package.json
```diff
@@ -16,7 +16,6 @@
     "lint:css": "./scripts/lint-css",
     "docs:build:dev": "./scripts/docs/build-dev.sh",
     "docs:publish": "./scripts/docs/publish.sh",
-    "docs:update-website-search-index": "./scripts/docs/update-website-search-index.sh",
```
bobylito: Yes!

docgen/middlewares.js
```diff
@@ -102,6 +103,20 @@ const common = [
   }),
   // since we use @import, autoprefixer is used after sass
   autoprefixer,
+  copy({
```
bobylito: Can you explain this part, @vvo ? Why do we need to move files to the v2 folder now?

vvo: Those two copy calls turns:
- docs/index.html
- docs/*

into
- docs/v2/*
- docs/index.html

Which is what we want to have in production too.

So that this is exactly the setup that will be used in every env (dev, preview, production). Leaving few differences between those envs, ensuring that if we use paths like a href etc then dev, preview or production will behave the same.

We could even move further and even in dev have /instantsearch.js/v2 to remove some env differences.

Ultimately we then push the docs/ folder to gh-pages.

Other possibilities would have been to only modify the gh-pages script and do the moves there but that would have been a bit too magical I felt and error prone.

Let me know if you want to change it.

bobylito: The iso env seems appealing but it's gonna be less practical (how someone know that they should go to netlify/instantsearch.js when opening a PR). That's a good as of now I think, closer to what we want 👍

vvo: It automatically redirects on netlify :)

vvo: Oh wait not yet, I will add that

vvo: There's no need for any redirection, default netlify preview is the V2 homepage which is what we want.

vvo:
(resolved conflicts)

Shipow:
Good job!

This link is still dead on V1
image

vvo:
Fixed (should be live soon)

PR #2428, opened 2017-10-02
docs(org): move all v1 pages to /v1/

### LINKED ISSUE #2236: Make V2 the default homepage?

vvo:
Fixes #2236

I will publish this one right now because it's non destructive (there's no /v1/ today).


vvo:
I don't really know WHY but clearly with our current setup yarn would not manage to finish on is v1. So I moved to npm5 package-lock.json and it works.

giphy

vvo:
All good

vvo:
Tested the publish script and it works, it publishes to v1 with the right assets paths. I added that to CONTRIBUTING.md in case we want to publish v1 docs again, that's doable.

vvo:
I added the right cloudflare rules as for v1 redirection

PR #2438, opened 2017-10-03
docs(v1): add v1 link

vvo:
This add a link back to v1 docs

image

### DIFF
--- a/docgen/src/data/communityHeader.json
+++ b/docgen/src/data/communityHeader.json
@@ -108,7 +108,10 @@
   {
     "name": "<span class='hidden-sm'>Demos</span>",
     "url": "examples.html"
-  },{
+  },  {
+    "name": "v1 docs",
+    "url": "https://community.algolia.com/instantsearch.js/v1/"
+  }, {
     "name": "Github",
     "image": "<img src='assets/img/github-icon.svg'/>",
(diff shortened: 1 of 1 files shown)


vvo:
@bobylito Had a good idea about putting a link somewhere like "Something is missing" "I can't find what I want" "Give feedback" 

that would work just as the Algolia docs. Maybe we could just link it to "new issue" on GitHub directly. Well that requires a GitHub account but still.

vvo:
Or to a google form, that's an easy win

bobylito:
> @bobylito Had a good idea about putting a link somewhere like "Something is missing" "I can't find what I want" "Give feedback"

Love it. That's a great idea!

> Now where would such link be?

A banner? Or we could have a question mark menu in the bar and have a link in it?

vvo:
> A banner?

I would use banner only for good news, they take too much space for such a "I have a problem" matter (kind of feels like everyone has issues with the doc, which is not the case).

I would put it at the bottom of every webpages. Most probably people will read deeply any webpage to understand that they are missing some information.

So I propose to add "Missing some documentation? [Let us know](LINK TO NEW ISSUE)!"

bobylito:
> So I propose to add "Missing some documentation? [Let us know](LINK TO NEW ISSUE)!" at the end of every page.

Sounds good to me! :)

PR #2440, opened 2017-10-03
Feature: RangeInput

### LINKED ISSUE #2317: New widget: RangeInput
In react-instantsearch, there is a [RangeInput widget](https://community.algolia.com/react-instantsearch/widgets/RangeInput.html) to be able to set a range with two inputs. We can implement with connectRangeSlider. In the process, we should probably alias connectRangeSlider to connectRange and deprecate connectRangeSlider.

samouss:
**Summary**

Fix #2317 

* Finally choose to implement `RangeInput` with the connector `connectRangeSlider` instead of `connectPriceRanges` because the logic was really the same. Maybe try in upcoming versions to manage `PriceRanges` with `connectRange` both share similarities (boundaries, range).

* The markup is the same as RIS

* Deprecation for the connector `connectRangeSlider` for `connectRange`

* Jest enzyme serializer added to be able to handle Snapshots with Enzyme

**Result**

You can play with the `RangeInput` on dev-novel.


samouss:
Due to the changes, now we can pass string values to the `refine` function. This allow the user to implement a range widget without carrying of parsing the numbers from the rendering.

But since the signature of the `refine` function is `(Array<number,number>) => void` we should probably only handle numbers and empty values (the later is used to be able to reset the range).

WDYT?

package.json
```diff
@@ -64,6 +64,7 @@
     "dev-novel": "0.0.4",
     "doctoc": "1.3.0",
     "enzyme": "2.9.1",
+    "enzyme-to-json": "2.0.1",
```
bobylito: Can you explain why did you need this new dependency?

src/components/RangeInput/RangeInput.js
```diff
@@ -0,0 +1,111 @@
...
+    return this.props.min >= this.props.max;
+  }
+
+  render() {
+    const { min: minValue, max: maxValue } = this.state;
+    const { min, max, step, cssClasses, labels } = this.props;
+
+    return (
+      <form className={cssClasses.form} onSubmit={this.onSubmit}>
+        <fieldset className={cssClasses.fieldset}>
```
bobylito: Why is the fieldset necessary?

bobylito:
> Due to the changes, now we can pass string values to the refine function. This allow the user to implement a range widget without caring about parsing the numbers from the rendering.

That's nice and it makes in terms of usage. What is preventing us from supporting both numbers, null and strings? I'm not sure it's consistent with other refine API's that takes numbers as parameters 🤔 but that would be a win in terms of usability.

package.json
samouss: This dependency allow to pass the Enzyme wrapper to the `expect` function to be able to directly use snapshot on it.

```js
import { shallow } from 'enzyme';
import serializer from 'enzyme-to-json/serializer';

expect.addSnapshotSerializer(serializer);

const component = shallow(<div />);

expect(component).toMatchSnapshot();
```

src/components/RangeInput/RangeInput.js
samouss: I just try to stay close as much as possible to RIS markup :)

samouss:
Nothing prevent us to dot it. The other connectors (that I looked) don't do validation on user input, they pass the results to the helper. But in the `connectRange` case we need to do some validations on it to be able to check the given value on boundaries.

IMO we should not trust the user inputs. We should probably do some validation in the connector to avoid to throw an error in the helper if the value is not supported (see [`valToNumber`](https://github.com/algolia/algoliasearch-helper-js/blob/7275a756510f5d7df460ae99cb88af6c2e617424/src/functions/valToNumber.js)).

src/connectors/range/connectRange.js
```diff
@@ -0,0 +1,269 @@
...
+        const facetsFromResults = results.disjunctiveFacets || [];
+        const facet = find(facetsFromResults, { name: attributeName });
+        const stats = facet && facet.stats;
+
+        const range = this._getCurrentRange(stats);
+        const start = this._getCurrentRefinement(helper);
+
+        renderFn(
+          {
+            refine: this._refine(helper, range),
```
bobylito: Maybe for the sake of readability we could rename range into currentRange?

src/connectors/range/connectRange.js
```diff
@@ -0,0 +1,269 @@
...
+        const [maxValue] =
+          helper.state.getNumericRefinement(attributeName, '<=') || [];
+
+        const min = _isFinite(minValue) ? minValue : -Infinity;
+        const max = _isFinite(maxValue) ? maxValue : Infinity;
+
+        return [min, max];
+      },
+
+      _refine(helper, range = {}) {
```
bobylito: Maybe for the sake of readability we could rename range into currentRange?

src/widgets/range-input/__tests__/range-input-test.js
```diff
@@ -0,0 +1,335 @@
+import ReactDOM from 'react-dom';
```
bobylito: Nice test!

src/widgets/range-input/__tests__/range-input-test.js
```diff
@@ -0,0 +1,335 @@
...
+        addAlgoliaAgent() {
+          return {};
+        },
+      },
+      'indexName',
+      { disjunctiveFacets: [attributeName] }
+    );
+
+  afterEach(() => {
+    ReactDOM.render.mockReset();
```
bobylito: Couldn't we rely on snapshots here? I'm not sure that relying on mocks helps much or does it?

dev/app/init-builtin-widgets.js
```diff
@@ -550,6 +550,114 @@ export default () => {
...
+      })
+    )
+    .add(
+      'disabled',
+      wrapWithHits(container => {
+        window.search.addWidget(
+          instantsearch.widgets.rangeInput({
+            container,
+            attributeName: 'price',
+            min: 500,
```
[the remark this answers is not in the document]
samouss: I tried to keep the same behaviour in the `RangeInput` & `RangeSlider`. In the later if the given `min` value is higher or equal than `max` the slider is disabled.

But we can definitely throw an error in the `RangeInput`, and do the same in the `RangeSlider` in the next major release.

src/widgets/range-input/range-input.js
```diff
@@ -0,0 +1,212 @@
...
+ * @property {string} [submit="Go"] Button label.
+ */
+
+/**
+ * @typedef {Object} RangeInputWidgetOptions
+ * @property {string|HTMLElement} container Valid CSS Selector as a string or DOMElement.
+ * @property {string} attributeName Name of the attribute for faceting.
+ * @property {number} [min] Minimal slider value, default to automatically computed from the result set.
+ * @property {number} [max] Maximal slider value, defaults to automatically computed from the result set.
+ * @property {number} [precision = 2] Number of digits after decimal point to use.
```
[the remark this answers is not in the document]
samouss: Yes you are right, it's a bit slow! We can provide `0` as default value to enable only integer.

src/components/RangeInput/RangeInput.js
```diff
@@ -0,0 +1,111 @@
...
+          <label className={cssClasses.labelMin}>
+            <input
+              className={cssClasses.inputMin}
+              type="number"
+              min={min}
```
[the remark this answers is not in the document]
samouss: I think we should validate it, because the `min` & `max` are the range where the user is able to search. It doesn't make sense to let it put an higher or lower. IMO we need to show a hint (in this case the `HTML5` validation) to the user that his search is out of bounds and so irrelevant.

With the actual implementation one thing that it might happen if we don't set the limit on the input is this:

* the user set the input out of bounds
* the user submit the form
* the search will be trigger with original values
* the render method will erase your current out of bounds values and replaced with the actual state

One thing that we could do is to be more precise when we refine. Currently, we just check that [one of the values has changed](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/range-slider/connectRangeSlider.js#L134) if its the case we clear the refinements. Then we set the refinement **IF** [the value are correct](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/range-slider/connectRangeSlider.js#L142). But the search & render will be trigger [even if the value is not set](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/range-slider/connectRangeSlider.js#L162). A solution could be to handle the values separately, in this way we could pass a falsy value without triggering the search & render.

WDYT?

src/widgets/range-input/__tests__/range-input-test.js
samouss: I mock the render method in order to be able to use the snapshot on it.

src/widgets/range-input/range-input.js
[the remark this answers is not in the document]
bobylito: 0 should be fine. They can customize it.

src/widgets/range-input/__tests__/range-input-test.js
bobylito: Ok!

samouss:
After discussion (see #2464), we keep the validation on inputs for a smoother user experience. The main differences from the original implementation is that now **both** inputs **must** be valid to be able to trigger the refinement.

PR #2450, opened 2017-10-04
chore(testing): don't check the commit messages

Haroenv:
We squash PRs anyway, we could test PR titles instead but it's not really needed IMO

It's super annoying to make commits on GitHub otherwise


bobylito:
Thanks for this contribution but it's important that our messages are correctly formatted. It is a cornerstone for automatically building our changelog which is otherwise a pain to write accurately.

Haroenv:
I think we should check for PR titles then, not for commit messages (because we _only_ squash)

bobylito:
> I think we should check for PR titles then, not for commit messages (because we only squash)

That's an interesting idea. Let's discuss that as par of the bi-weekly core sync. Thanks for proposing this PR. I will close it for now.

PR #2451, opened 2017-10-04
feat(breadcrumb): Add the breadcrumb widget

marielaures:
**Summary**

Add a breadcrumb widget to improve navigability.
This has already been implemented on React InstantSearch and this pull request aims to provide the same functionality.

**Result**
image


src/connectors/breadcrumb/connectBreadcrumb.js
```diff
@@ -0,0 +1,194 @@
...
+            currentConfiguration.hierarchicalFacets,
+            ({ name }) => name === hierarchicalFacetName
+          );
+          if (facetSet) {
+            if (
+              !isEqual(facetSet.attributes, attributes) ||
+              facetSet.separator !== separator
+            ) {
+              console.warn(
+                'Using Breadcrumb & HierarchicalMenu on the same facet with different options'
```
bobylito: Maybe we could be more specific on the next steps. What about adding that one of the configuration will be overriden and that they should check their options?

src/connectors/breadcrumb/connectBreadcrumb.js
```diff
@@ -0,0 +1,194 @@
...
+export default function connectBreadcrumb(renderFn) {
+  checkRendering(renderFn, usage);
+  return (widgetParams = {}) => {
+    const { attributes, separator = ' > ', rootPath = null } = widgetParams;
+    const [hierarchicalFacetName] = attributes;
+
+    return {
+      getConfiguration: currentConfiguration => {
+        if (currentConfiguration.hierarchicalFacets) {
+          const facetSet = find(
```
bobylito: Should be named `isFacetSet` instead of `facetSet` because it's a boolean (and I was confused by the usage of a [set](https://en.wikipedia.org/wiki/Set_(abstract_data_type)) here)

src/connectors/breadcrumb/connectBreadcrumb.js
```diff
@@ -0,0 +1,194 @@
...
+      if (Array.isArray(currentItem.data)) {
+        const children = prepareItems(currentItem);
+        result = result.concat(children);
+      }
+    }
+    return result;
+  }, []);
+}
+
+function processArray(array) {
```
bobylito: `processArray` doesn't say much about the process that is applied to the array. In this case you're shifting the values, what about `shiftItemsValues`?

src/connectors/breadcrumb/connectBreadcrumb.js
```diff
@@ -0,0 +1,194 @@
...
+  return array.map((x, idx) => ({
+    name: x.name,
+    value: idx + 1 == array.length ? null : array[idx + 1].value,
```
bobylito: We should only use `==` when we want JS to make [type coercion](https://code.tutsplus.com/articles/the-beginners-guide-to-type-coercion-a-practical-example--cms-21998) automatically. Since we're comparing numbers, it should not be an issue to use `===`.

src/widgets/breadcrumb/breadcrumb.js
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * You can also use a sort function that behaves like the standard Javascript [compareFunction](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Syntax).
+ * @property {BreadcrumbTemplates} [templates] Templates to use for the widget.
+ * @property {BreadcrumbTransforms} [transformData] Set of functions to transform the data passed to the templates.
+ * @property {boolean} [autoHideContainer=true] Hides the container when there are no items in the breadcrumb.
+ * @property {BreadcrumbCSSClasses} [cssClasses] CSS classes to add to the wrapping elements.
+ */
+
+/**
+ * The breadcrumb widget is a secondary navigation scheme that allows the user to see where the current page is in relation to the website's hierarchy.
```
bobylito: It might be a bit confusing to talk about website hierarchy because search is only a part of the website. I would go for something like 'facet hierarchy', WDYT?

src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,82 @@
...
+
+    const breadcrumb = items.map((item, idx) => {
+      const isLast = idx === items.length - 1;
+      const label = isLast
+        ? <a className={`${cssClasses.disabledLabel} ${cssClasses.label}`}>
+            {item.name}
+          </a>
+        : <a
+            className={cssClasses.label}
+            href={createURL(items[idx].value)}
```
bobylito: `items[idx] === item` so the code can be simplfied here.

src/connectors/breadcrumb/connectBreadcrumb.js
```diff
@@ -0,0 +1,194 @@
...
+              hierarchicalFacetName
+            );
+            if (breadcrumb.length > 0) {
+              return createURL(
+                helper.state.toggleRefinement(
+                  hierarchicalFacetName,
+                  breadcrumb[0]
+                )
+              );
+            } else {
```
bobylito: An function which does not explicitly return a value, actually returns `undefined`. So we don't need the last `else` block

src/connectors/breadcrumb/connectBreadcrumb.js
```diff
@@ -0,0 +1,194 @@
...
+        );
+      },
+
+      render({ instantSearchInstance, results, state }) {
+        if (
+          !state.hierarchicalFacets ||
+          (Array.isArray(state.hierarchicalFacets) &&
+            state.hierarchicalFacets.length === 0)
+        ) {
+          throw new Error(usage);
```
bobylito: This message is sent too late. It should be done as soon as we have all the necessary information. Most likely before returning the widget object.

bobylito:
This widget could be a nice addition to the e-commerce demo.

screen shot 2017-10-06 at 12 26 32

bobylito:
I think we're good to go here :) Thanks @marielaures and congrats on your first widget on IS.js, welcome to the contributors 🚀 

Let's do the update of demo in another PR ;)

PR #2462, opened 2017-10-09
style(menuSelect): apply select theme

### LINKED ISSUE #2448: Style / menuselect: should have the same look as the other select with the theme
The widget is not styled for the moment.

screen shot 2017-10-04 at 12 23 25

It should have this style:

screen shot 2017-10-04 at 12 23 48

iam4x:
fixes #2448

### DIFF
--- a/src/css/theme/_hits-per-page.scss
+++ b/src/css/theme/_hits-per-page.scss
@@ -1,5 +1,8 @@
 select.ais-hits-per-page-selector {
+  -webkit-appearance: none;
+  -moz-appearance: none;
   appearance: none;
+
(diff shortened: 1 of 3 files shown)


src/css/theme/_hits-per-page.scss
```diff
@@ -1,5 +1,8 @@
 select.ais-hits-per-page-selector {
+  -webkit-appearance: none;
```
bobylito: Don't we use autoprefixer?

bobylito: humm only on docgen...

iam4x:
We do have autoprefixer, but this is an edge case where Chrome understands it without the prefix but does not applies.

So we can target lower browsers, but it would mean that other properties which we don't want to prefix would be prefixed it we fix it at postcss.config.js level 👍

bobylito:
OK sounds fair @iam4x

PR #2463, opened 2017-10-09
fix(rangeSlider): left handle default position

### LINKED ISSUE #2449: Style / rangeslider: Problems with the alignements
screen shot 2017-10-04 at 12 24 47

I think the handles should not go beyond the verticals defined by the headers. Same goes for the labels.

iam4x:
fixes #2449

### DIFF
--- a/src/css/default/_range-slider.scss
+++ b/src/css/default/_range-slider.scss
@@ -52,7 +52,7 @@ $range-slider-handle-bg: $white;
     }
 
     .rheostat-handle {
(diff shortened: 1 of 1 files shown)


bobylito:
screen shot 2017-10-09 at 15 06 17

That doesn't really fix the issue. it's both the left and right handles and the labels that should be fixed. Let me know if you want to check together.

iam4x:
Hard to take a decision on this one, let's wait for instantsearch.css 👍

PR #2464, opened 2017-10-09
Feature: RangeInput without validation in component

samouss:
**Summary**

As we discuss [here](https://github.com/algolia/instantsearch.js/pull/2440#discussion_r143139279), this PR remove the validations on the inputs inside the component. All the logic is purely handled by the connector. The most significant change is that now **both** inputs must be valid to trigger a search.

The logic for the connector is less permissive for the validation part than the other one. It's a good point, specially for the user using the connector API. But IMO we still need to validate the inputs from the component (via the HTML5 validation) because when the user trigger a search out of range nothing happen. The user as no clue on why the search is invalid.

**Example:**

- use `RangeInput` on `default`
- set `min` input to `10` & `max` input to `35000` (value out of range)
- submit the form

Nothing happen, because the search is out of range. But the user don't have feedbacks of the state of his search. With the HTML5 validation we should see an error that display `Value must be less or equal than {MAX}`.

### DIFF
--- a/src/connectors/range/__tests__/connectRange-test.js
+++ b/src/connectors/range/__tests__/connectRange-test.js
@@ -692,6 +692,40 @@ describe('connectRange', () => {
       expect(helper.search).toHaveBeenCalled();
     });
 
+    it('expect to refine min when user bounds are set and value is not valid', () => {
+      const range = { min: 10, max: 500 };
+      const values = [null, 490];
+      const helper = createHelper();
+      const widget = connectRangeSlider(rendering)({
+        attributeName,
+        min: 10,
+      });
+
+      widget._refine(helper, range)(values);
+
+      expect(helper.getNumericRefinement(attributeName, '>=')).toEqual([10]);
+      expect(helper.getNumericRefinement(attributeName, '<=')).toEqual([490]);
+      expect(helper.clearRefinements).toHaveBeenCalled();
+      expect(helper.search).toHaveBeenCalled();
+    });
+
+    it('expect to refine max when user bounds are set and value is not valid', () => {
+      const range = { min: 10, max: 500 };
+      const values = [20, null];
+      const helper = createHelper();
+      const widget = connectRangeSlider(rendering)({
+        attributeName,
+        max: 500,
+      });
+
+      widget._refine(helper, range)(values);
+
+      expect(helper.getNumericRefinement(attributeName, '>=')).toEqual([20]);
+      expect(helper.getNumericRefinement(attributeName, '<=')).toEqual([500]);
+      expect(helper.clearRefinements).toHaveBeenCalled();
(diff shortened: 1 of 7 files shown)


bobylito:
So do you propose to keep the attributes that allow for validation and also add your changes here, so that the API of the connector can be tighter and also give the user some feedback if the values are wrong?

samouss:
Yep!

bobylito:
> Yep!

Sounds good to me then.

src/connectors/range/__tests__/connectRange-test.js
```diff
@@ -526,8 +526,11 @@ describe('connectRange', () => {
     const rendering = () => {};
     const createHelper = () => {
       const helper = jsHelper(fakeClient);
-      helper.clearRefinements = jest.fn();
       helper.search = jest.fn();
+      const initalClearRefinements = helper.clearRefinements;
+      helper.clearRefinements = jest.fn((...args) =>
```
bobylito: Why wrap the `clearRefinements` method?

src/widgets/range-input/range-input.js
```diff
@@ -41,8 +41,8 @@ const renderer = ({
   const shouldAutoHideContainer = autoHideContainer && rangeMin === rangeMax;
 
   const values = {
-    min: minValue !== -Infinity && minValue !== rangeMin ? minValue : null,
-    max: maxValue !== Infinity && maxValue !== rangeMax ? maxValue : null,
+    min: minValue !== -Infinity && minValue !== rangeMin ? minValue : undefined,
+    max: maxValue !== Infinity && maxValue !== rangeMax ? maxValue : undefined,
```
bobylito: Why `undefined` over `null`?

src/connectors/range/__tests__/connectRange-test.js
samouss: I wrap the `clearRefinements` method to be able to test if she was been called. You are right what matter is the result at the end. But we should check that the `clearRefinements` is called with the given attribute name in order to avoid to clear all the current refinements (see: #2459).

src/widgets/range-input/range-input.js
samouss: `undefined` over `null` in order to avoid to pass `null` to the `refine` function of the `connector`. `null` will be treat as an invalid value instead of a reset action.

bobylito: Good for me.

src/connectors/range/__tests__/connectRange-test.js
bobylito: Let's try that. If we don't like it we can revert back :)

samouss:
Close this, the changes will be integrate [in the original PR](https://github.com/algolia/instantsearch.js/pull/2440).

PR #2467, opened 2017-10-10
chore(publish): fix dev dep install on docgen when publishing

bobylito:
Also adds `yarn install` before building instantsearch.js for the website

### DIFF
--- a/scripts/docs/build-dev.sh
+++ b/scripts/docs/build-dev.sh
@@ -1,6 +1,6 @@
 #! /usr/bin/env bash
 
-set -ev # exit when error
+set -e # exit when error
 
 NODE_ENV=${NODE_ENV:-development}
 VERSION=preview-$(json version < package.json)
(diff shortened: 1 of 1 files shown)


bobylito:
NB: it doesn't work with yarn 1.1.0. But it works with 1.2.0

vvo:
https://github.com/algolia/instantsearch.js/blob/2208f7d93c5c1fa906b3782764fd9ed5de006f09/scripts/docs/publish.sh#L29

Since the publish scripts sets NODE_ENV to production this will not install devDependencies no?

We could force NODE_ENV to development if that's the case in build.sh

bobylito:
It's updated @vvo

scripts/docs/build-dev.sh
```diff
@@ -1,11 +1,12 @@
 #! /usr/bin/env bash
 
-set -ev # exit when error
+set -e # exit when error
```
iam4x: Just for information, what is `v` stand for?

vvo: verbose, every line of the script will be also written on the output (not just the effect, but also the line itself).

doc: http://tldp.org/LDP/abs/html/options.html (not easy to find!)

PR #423, opened 2015-11-03
chore(template): get rid of extra <div /> wrapping the Template if possible

redox:
Fix #422

BREAKING CHANGE: some extra divs have been removed and my break your CSS rules.

PR #2471, opened 2017-10-11
fix(ie11): preact-compat algolia fork with #423

bobylito:
This is to fix an [issue of preact-compat](https://github.com/developit/preact-compat/pull/423) on IE11. It uses an algolia fork of preact-compat, that we need to replace once the new version is out (3.17.1 or 3.18.0)

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "hogan.js": "^3.0.2",
     "lodash": "^4.17.4",
     "preact": "^8.1.0",
-    "preact-compat": "^3.16.0",
+    "preact-compat-algolia": "^3.17.0-beta.0",
     "prop-types": "^15.5.10",
     "rheostat": "algolia/rheostat#dist-preact",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "instantsearch.js",
-  "version": "2.2.0",
+  "version": "2.2.1-beta.0",
```
iam4x: this should only change with the build script

bobylito: That's alright, since it's gonna be 2.2.1 anyway.

iam4x:
Failling on the commit:

```
Checking commit message for 2c5abfc3e7315d6a723b96a539a09857d48e101d... KO (format): v2.2.1-beta.0
```

bobylito:
Well it seems that it might not fix what is supposed to. Will check that tomorrow.

bobylito:
Actually the library works with IE11. No need for this PR.

PR #2472, opened 2017-10-11
[DO NOT MERGE] refactor: use widgetParams from renderer

samouss:
**Summary**

This is an example for the simplification of the widget renderer (see: #2473).

### DIFF
--- a/src/widgets/range-slider/range-slider.js
+++ b/src/widgets/range-slider/range-slider.js
@@ -17,17 +17,22 @@ const defaultTemplates = {
 
 const bem = bemHelper('ais-range-slider');
 
-const renderer = ({
-  containerNode,
-  cssClasses,
-  pips,
-  step,
-  tooltips,
-  autoHideContainer,
-  collapsible,
-  renderState,
-  templates,
-}) => ({ refine, range, start, instantSearchInstance }, isFirstRendering) => {
+const renderer = (
+  { refine, range, start, instantSearchInstance, widgetParams },
+  isFirstRendering
+) => {
+  const {
+    container,
+    cssClasses,
+    pips,
+    step,
+    tooltips,
+    autoHideContainer,
+    collapsible,
+    renderState,
+    templates,
+  } = widgetParams;
+
   if (isFirstRendering) {
     renderState.templateProps = prepareTemplateProps({
       defaultTemplates,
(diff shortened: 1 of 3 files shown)


src/widgets/range-slider/range-slider.js
```diff
@@ -162,47 +167,35 @@ rangeSlider({
...
-    cssClasses,
-  });
+  if (!widgetParams.container) {
+    throw new Error(usage);
+  }
 
   try {
-    const makeWidget = connectRangeSlider(specializedRenderer);
-    return makeWidget({ attributeName, min, max, precision });
+    const makeWidget = connectRangeSlider(renderer);
```
bobylito: The nice thing is that we can even move that out: having a single instance is possible.

src/widgets/range-slider/range-slider.js
```diff
@@ -162,47 +167,35 @@ rangeSlider({
...
+  const widgetParams = {
+    container: '',
+    step: 1,
+    pips: true,
+    tooltips: true,
+    autoHideContainer: true,
+    cssClasses: {},
+    renderState: {},
+    templates: defaultTemplates,
+    ...params,
```
bobylito: Someone can pass a new renderState here.

bobylito:
I think it could really simplifies thinking about the code in the widgets. It works for me because the connect(renderer) that we use will be hidden which is what we want because we have some plumbing that we want to hide (renderState).

src/widgets/range-slider/range-slider.js
samouss: Yep, you are right! But it could be hide, I will update it.

bobylito:
Imho that's a great improvement over the current pattern for implementing widget in InstantSearch.js. We should definetely move forward with it. However this is not high priority as we haven't identified a bug with the current approach. Are you ok @samouss for moving that to 2018?

iam4x:
I agree with @bobylito it would be great to don't move the connectors API until we have a v1 of angular-instantsearch 👍

samouss:
@bobylito No problem, it was just a proposition. I close the PR but I leave the issue open for later.

bobylito:
> @bobylito No problem, it was just a proposition. I close the PR but I leave the issue open for later.

And just to be perfectly clear: I love it :)

PR #2477, opened 2017-10-12
Fix/remove warning in test

samouss:
**Summary**

This PR aims to fixed all the warnings from the tests & PropTypes (when found).

**Result**

You can check the ouput of the on Travis.

**Before**:

screen shot 2017-10-12 at 13 02 21

**After**:

screen shot 2017-10-12 at 12 58 27

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "eslint-plugin-prettier": "2.3.1",
     "eslint-plugin-react": "7.3.0",
     "expect": "1.20.2",
-    "expect-jsx": "2.6.0",
+    "expect-jsx": "4.0.0",
     "express": "4.15.4",
     "extract-text-webpack-plugin": "2.1.2",
     "gh-pages": "1.0.0",
--- a/src/components/Slider/Pit.js
+++ b/src/components/Slider/Pit.js
@@ -11,7 +11,7 @@ const Pit = ({ style, children }) => {
   // Children could be an array, unwrap the value if it's the case
   // see: https://github.com/developit/preact-compat/issues/436
   const value = Array.isArray(children) ? children[0] : children;
-  const pitValue = Math.round(parseFloat(value, 10) * 100) / 100;
+  const pitValue = Math.round(parseFloat(value) * 100) / 100;
 
   return (
     <div
--- a/src/components/Slider/Pit.js
+++ b/src/components/Slider/Pit.js
@@ -1,15 +1,16 @@
-import PropTypes from 'prop-types';
 import React from 'preact-compat';
+import PropTypes from 'prop-types';
+import includes from 'lodash/includes';
 import cx from 'classnames';
 
-import includes from 'lodash/includes';
-
-const Pit = ({ key, style, children }) => {
+const Pit = ({ style, children }) => {
   // first, end & middle
   const positionValue = Math.round(parseFloat(style.left));
   const shouldDisplayValue = includes([0, 50, 100], positionValue);
 
-  const value = children ? children : key.replace('pit-', '');
+  // Children could be an array, unwrap the value if it's the case
+  // see: https://github.com/developit/preact-compat/issues/436
+  const value = Array.isArray(children) ? children[0] : children;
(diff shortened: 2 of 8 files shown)


src/components/Slider/Slider.js
```diff
@@ -48,9 +48,7 @@ export class RawSlider extends Component {
       min,
       ...times(steps - 1, step => min + stepsLength * (step + 1)),
       max,
-    ]
-      // bug with `key={ 0 }` and preact, see https://github.com/developit/preact/issues/642
-      .map(pitPoint => (pitPoint === 0 ? 0.000001 : pitPoint));
```
samouss: It's already fixed in Rheostat fork (see: [rheostat/Slider.jsx#L665](https://github.com/algolia/rheostat/blob/dist-preact/src/Slider.jsx#L665)).

package.json
```diff
@@ -74,7 +74,7 @@
     "eslint-plugin-prettier": "2.3.1",
     "eslint-plugin-react": "7.3.0",
     "expect": "1.20.2",
-    "expect-jsx": "2.6.0",
+    "expect-jsx": "4.0.0",
```
bobylito: That's already one thing that we won't have to do using renovate :)

src/components/Slider/Pit.js
```diff
@@ -30,8 +31,12 @@ const Pit = ({ key, style, children }) => {
 };
 
 Pit.propTypes = {
-  key: PropTypes.string,
```
bobylito: Why isn't the key needed anymore?

package.json
samouss: Yep, but the warnings came from this package. So how we handle that? We use renovate or we can upgrade the dependency manually?

src/components/Slider/Pit.js
samouss: Actually we used the `key` to bypass a bug with `preact`. Now it seems to render correctly, I  don't spot any bugs. I know @iam4x worked on it, if you think that something goes wrong don't hesitate :)

iam4x: It should be good, wasn't too sure about this fix I've made. It removed a warning and it was also used when I tried to moved the tests to a full preact rendering.

src/widgets/numeric-refinement-list/__tests__/numeric-refinement-list-test.js
```diff
@@ -130,7 +130,7 @@ describe('numericRefinementList()', () => {
           header: '',
           // eslint-disable-next-line max-len
           item:
-            '<label class="{{cssClasses.label}}">\n <input type="radio" class="{{cssClasses.radio}}" name="{{attributeName}}" {{#isRefined}}checked{{/isRefined}} />{{label}}\n</label>',
+            '<label class="{{cssClasses.label}}">\n  <input type="radio" class="{{cssClasses.radio}}" name="{{attributeName}}" {{#isRefined}}checked{{/isRefined}} />{{label}}\n</label>',
```
iam4x: Hmmm, this is not really expected isn't it?

samouss: Yes it is, in fact due it's to the indentation in `defaultTemplate` ([link](https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/numeric-refinement-list/defaultTemplates.js#L5)). You can see that at the begining of the line we have two spaces.

iam4x: But, this is not in the PR?

samouss: I just need to change the test, it's probably due some internal change in `react-element-to-jsx-string` that is used by `expect-jsx`.

package.json
bobylito: That's actually a good thing to not have to do it with Renovate. That's probably a strategy that we should adopt more often 👍

PR #2498, opened 2017-10-18
Avoid to always round the range in connectRange

samouss:
**Summary**

In the `connectRange` we always round the range even with a precision greater than 0. Now we round the range only for a precision of 0 and leave the original value in the other case.

**Result**

You can play with the `RangeInput` on dev-novel.

**Before:**

screen shot 2017-10-18 at 11 48 02

**After:**

screen shot 2017-10-18 at 11 40 49

### DIFF
--- a/src/connectors/range/connectRange.js
+++ b/src/connectors/range/connectRange.js
@@ -81,6 +81,8 @@ export default function connectRange(renderFn) {
 
     return {
       _getCurrentRange(stats) {
+        const pow = Math.pow(10, precision);
+
         let min;
         if (hasMinBound) {
           min = minBound;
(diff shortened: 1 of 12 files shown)


src/connectors/range/connectRange.js
```diff
@@ -100,8 +100,8 @@ export default function connectRange(renderFn) {
         }
 
         return {
-          min: Math.floor(min),
-          max: Math.ceil(max),
+          min: precision === 0 ? Math.floor(min) : min,
```
bobylito: What does it mean for a precision of 1 and there 3 digits after the `.` or `,`?

samouss: When I begin to implement it, I thought it was a bit overkill.

bobylito: Well Isn't the purpose to bring consistency? And in this case, we only have consistency for precision = 0, right?

bobylito: But then I might not have a good idea of the complexity...

samouss: Right, it will update it 🙃

Issue #2505, opened 2017-10-19
showParentLevel for hierarchicalMenu doesn't do anything.


Giovanni-Mattucci:
What is the `showParentLevel` property for the `hierarchicalMenu` widget supposed to do?

I do not see any visual difference between `true`, or `false`. 

Also why is the property a `string` when its settings are true/false?

Haroenv:
Where do you see it documented as a string? All references seem to be `boolean` to me 🤔

Giovanni-Mattucci:
Cool, but on another note, what exactly does `showParentLevel` do?

bobylito:
@Giovanni-Mattucci It's very subtle to use actually. Even I had to test it :)

When set to true, it displays the siblings of the selected values. Here is an example:

showParentLevel=true:
![screen shot 2017-10-20 at 14 13 27](https://user-images.githubusercontent.com/393765/31820366-49be13f4-b5a1-11e7-95da-4357b845dac1.png)

showParentLevel=false:
![screen shot 2017-10-20 at 14 13 35](https://user-images.githubusercontent.com/393765/31820312-0d4233e2-b5a1-11e7-973e-ac12f20f6156.png)

Giovanni-Mattucci:
@bobylito ahh I see! Thank you, that visual helps. I didn't have a data set with multiple levels to see its effect.

bobylito:
I tried to give a better description of the behaviour of this parameter [here](https://github.com/algolia/instantsearch.js/blob/bd64acfa57d4a42e3db2c548f29a604b65b14511/src/connectors/hierarchical-menu/connectHierarchicalMenu.js#L40-L41). Can you tell me @Giovanni-Mattucci if it's clearer? Or maybe suggest something that could better help understand it?

Giovanni-Mattucci:
@bobylito that is a bit more clear, but change "levels" to "level" since you can't select more than one parent hierarchical facet:
`Show the siblings of the selected parent level of the current refined value. This does not impact the root level.` :)

In documentation if you could visually illustrate it with some text, that would be good too:

- Parent lvl0
  - **lvl1**
    - **lvl2**
    - lvl2
    - lvl2
  - lvl 1
  - lvl 1
- Parent lvl0
- Parent lvl0

vs

- Parent lvl0
  - **lvl1**
    - **lvl2**
- Parent lvl0
- Parent lvl0

Not sure how best to illustrate, but I think your description is an improvement.

bobylito:
[...]

PR #2506, opened 2017-10-19
docs(hierarchicalMenu): showParentLevel is boolean

Haroenv:
see #2505, thanks @Giovanni-Mattucci


Haroenv:
uh that snapshot failure isn't mine @iam4x

PR #2507, opened 2017-10-20
feat(refinementList): add escapeFacetHits parameter

### LINKED ISSUE #2474: SFFV in refinementList has invalid html for highlighting
see also https://discourse.algolia.com/t/searfforfacetvalue-highlight-bug/3198

bug 😢 

The highlightPreTag is set to the `__ais-highlight__`, like the post tag when the hits widget has the `escapeHits` flag set to `true`. It's not set back to the original tag used by highlighting, or one configurable in the widget.

nobody should notice we are using a different tag against XSS

all

replace the preTag and postTag by default with em, split up the text in pieces and insert text before and after

Maybe an option to insert html, but only when someone asks for it

I think latest?

bobylito:
This parameter let the developer choose to escape the facet values returned by Algolia. This feature is required to keep the independance of widgets, and provide some opt-in safety. It originated when starting to work on #2474 but didn't want to just fix it.

NB: I also updated the Helper to be able to implement this feature.

Fix #2474


src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -457,4 +458,305 @@ describe('connectRefinementList', () => {
...
+      attributeName: 'category',
+      limit: 2,
+    });
+
+    const helper = algoliasearchHelper(
+      fakeClient,
+      '',
+      widget.getConfiguration({})
+    );
+    helper.search = sinon.stub();
```
samouss: Maybe prefer `jest.fn`. We should avoid to add new code with `sinon`. We probably remove it at some point.

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -457,4 +458,305 @@ describe('connectRefinementList', () => {
...
+    expect(sffvQuery).toBe('da');
+    expect(sffvFacet).toBe('category');
+    expect(maxNbItems).toBe(2);
+    expect(paramOverride).toEqual({
+      highlightPreTag: undefined,
+      highlightPostTag: undefined,
+    });
+
+    // because of how promises are resolved we use setImmediate
+    setImmediate(() => {
```
samouss: You should probably use `done` for finish the test. I think the assertions inside `setImmediate` are executed after the test, if they fail all the process will fail instead of just the test case.

You can ensure how many assertions you expect in an async test with:

```js
expect.assertions(numberOfAssertionsExpected);
```

src/widgets/refinement-list/refinement-list.js
```diff
@@ -247,6 +249,9 @@ export default function refinementList(
     throw new Error('showMore.limit configuration should be > than the limit in the main configuration'); // eslint-disable-line
   }
 
+  const escapeFacetValues = searchForFacetValues
+    ? searchForFacetValues.escapeFacetValues || false
```
samouss: Maybe cast the value with `Boolean()`, it will ensure that the value will be a `boolean`. With the actual condition the value can be anything thruthy or a `boolean`.

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
bobylito: I didn't have any issue when the tests are green, it's just that when they fail it breaks jest completely. I will try that, maybe it's gonna be fine.

vvo: I wonder also if we could now use:

```js
await Promise.resolve();
await Promise.resolve();
```

As an easy way to wait for the render fn to be called

samouss: You are right @vvo we could do use an `async` function to `await` the result of the `search` function.

vvo: But you can still use await Promise.resolve() to force the event loop to run the render function. That's what setImmediate does here.

Was just a small comment if it's feasible it might be easier to read than setImmediate (and safer)

bobylito: That's what I call an interesting hack @vvo :) Thanks.

bobylito: After looking at some jest issues, it seems that it handles returned promises and since my main issue with setImmediate is that when the test fail jest breaks, I ended up creating a new promise and in the then do my tests. It's not as smart but `¯\_(ツ)_/¯`

bobylito:
That would be great to fix the snapshots 🤔

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -98,10 +97,10 @@ describe('connectRefinementList', () => {
     });
 
     // test if widget is not rendered yet at this point
-    expect(rendering.callCount).toBe(0);
+    expect(rendering.mock.calls.length).toBe(0);
```
samouss: You could use the shorthand method: `.not.toHaveBeenCalled()`

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -111,11 +110,11 @@ describe('connectRefinementList', () => {
     });
 
     // test that rendering has been called during init with isFirstRendering = true
-    expect(rendering.callCount).toBe(1);
+    expect(rendering.mock.calls.length).toBe(1);
```
samouss: You could use the shorthand method: `.toHaveBeenCalled()`

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -130,10 +129,10 @@ describe('connectRefinementList', () => {
     });
 
     // test that rendering has been called during init with isFirstRendering = false
-    expect(rendering.callCount).toBe(2);
-    expect(rendering.lastCall.args[1]).toBe(false);
+    expect(rendering.mock.calls.length).toBe(2);
```
samouss: You could use the shorthand method: `.toHaveBeenCalledTimes(2)`

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -111,11 +110,11 @@ describe('connectRefinementList', () => {
     });
 
     // test that rendering has been called during init with isFirstRendering = true
-    expect(rendering.callCount).toBe(1);
+    expect(rendering.mock.calls.length).toBe(1);
     // test if isFirstRendering is true during init
-    expect(rendering.lastCall.args[1]).toBe(true);
+    expect(rendering.mock.calls[0][1]).toBe(true);
```
samouss: You could use the shorthand method: `.toHaveBeenLastCalledWith(expect.any(Object), true)`

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
bobylito: I've switched to the number of call counts because it give a finer grain when there is a fail. In this case it would not matter much, but I still prefer consistency of API usage. Is that ok?

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
bobylito: Ok gotcha! Thanks!

PR #2509, opened 2017-10-20
feat(searchbox): provide a loading indicator

### LINKED ISSUE #2315: Loading indicator
When the search has latency it is pretty hard to understand why the results don't update. It's especially true when the search happens on mobile (where there are a lot of latency even on good network). One way to tell the user that something is happening is to display a spinner.

The good thing is that the JS Helper now implements a flag and an event to notify that the search has been completed.

What I propose is a standalone `loading` connector and an implementation in the searchbox (so that we can replace the magnifying glass with a spinner)

bobylito:
When the requests are stalled by the network, it is important to let the user know.

This is an early opening of the PR so that I can start gathering feedback. Current questions: 
 - the magnifying glass is directly in the CSS, but the loading part adds a class. Is it the right approach or should I change the image?
 - I'm not sure about the calibration of the delay, is it too short?

https://deploy-preview-2509--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=SearchBox.With%20loading%20indicator (to test in bad condition, you should set your Chrome console to slow 3g)

fixes #2315


bobylito:
Waiting for the link :)

Haroenv:
Seems like the indicator shows up instantly, do you think it's something that people should add with a transition in css?

Haroenv:
I added a loader image, from https://github.com/SamHerbert/SVG-Loaders which is MIT

src/css/theme/_search-box.scss
```diff
@@ -48,7 +48,8 @@
       svg{
         visibility: hidden;
       }
-      background: red;
+      background-image: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="38" height="38"><defs><linearGradient id="a" x1="8.04%" x2="65.68%" y1="0%" y2="23.86%"><stop stop-color="%23BFC7D8" stop-opacity="0" offset="0%"/><stop stop-color="%23BFC7D8" stop-opacity=".63" offset="63.15%"/><stop stop-color="%23BFC7D8" offset="100%"/></linearGradient></defs><g fill="none" fill-rule="evenodd" transform="translate(1 1)"><path stroke="url(%23a)" stroke-width="2" d="M36 18c0-9.94-8.06-18-18-18"><animateTransform attributeName="transform" type="rotate" from="0 18 18" to="360 18 18" dur="0.9s" repeatCount="indefinite"/></path><circle cx="36" cy="18" r="1" fill="%23BFC7D8"><animateTransform attributeName="transform" type="rotate" from="0 18 18" to="360 18 18" dur="0.9s" repeatCount="indefinite"/></circle></g></svg>');
```
bobylito: Thanks for this @Haroenv. I was not sure where to go on that part, and I was thinking to maybe let the designers have an opinion here (cc @LukyVj)

bobylito:
> I added a loader image, from https://github.com/SamHerbert/SVG-Loaders which is MIT
>  👍 1

Looks good :D

bobylito:
Closing in favor of #2569

PR #2512, opened 2017-10-21
fix(menuSelect): use preact instead of React

### LINKED ISSUE #2388: Why does instantsearch require React?

bobylito:
It seems that we lost this commit before the last release but I can't find out when it happened.

This should fix #2388


bobylito:
Actually this was correctly but that was an incomplete fix :/ (missing the component part)

PR #2513, opened 2017-10-21
fix(MenuSelect): switch from react to preact-compat

bobylito:
Fixing the missing elements on the topic of "removing react in favor of preact".


bobylito:
I'm currently testing it by implementing it on [SearchStone](https://github.com/Shipow/searchstone/)

bobylito:
> Looks good, but is the version change on purpose?

Yes. We have to update the `package.json` file, and for now we commit it when we do the release (for tracking it). Is there a case in which it could be an issue?

PR #2517, opened 2017-10-23
fix(connectRefinementList): set default value for limit

### LINKED ISSUE #2504: connectRefinementList::canToggleShowMore is always true when no limit is set
The condition for `canToggleShowMore` is the following `canToggleShowMore: isShowingMore || !hasExhaustiveItems`

And `hasExhaustiveItems` is computed this way:

```js
const hasExhaustiveItems = items.length < this.getLimit();
```

If `this.getLimit() === undefined` then `hasExhaustiveItems` will be false and at the end `canToggleShowMore` would be true even if no limit has been set and there is not others available items.

This is not happening with the bundled RefinementList widget because we set by default `limit` to 10 (https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/refinement-list/refinement-list.js#L231)

Should we ensure we have a default limit in the connector and update the documentation or fix the way `hasExhaustiveItems` is computed?

iam4x:
Fixes #2504

### DIFF
--- a/src/connectors/refinement-list/connectRefinementList.js
+++ b/src/connectors/refinement-list/connectRefinementList.js
@@ -58,7 +58,7 @@ export const checkUsage = ({
  * @typedef {Object} CustomRefinementListWidgetOptions
  * @property {string} attributeName The name of the attribute in the records.
  * @property {"and"|"or"} [operator = 'or'] How the filters are combined together.
- * @property {number} [limit = undefined] The max number of items to display when
+ * @property {number} [limit = 10] The max number of items to display when
  * `showMoreLimit` is not or if the widget is showing less value.
  * @property {number} [showMoreLimit] The max number of items to display if the widget
  * is showing more items.
@@ -148,7 +148,7 @@ export default function connectRefinementList(renderFn) {
     const {
       attributeName,
       operator = 'or',
-      limit,
+      limit = 10,
       showMoreLimit,
       sortBy = ['isRefined', 'count:desc', 'name:asc'],
     } = widgetParams;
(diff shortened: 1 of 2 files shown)


iam4x:
Found another bug, do not merge yet 👍

iam4x:
Should be good now @bobylito 👍

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -58,7 +58,7 @@ export const checkUsage = ({
  * @typedef {Object} CustomRefinementListWidgetOptions
  * @property {string} attributeName The name of the attribute in the records.
  * @property {"and"|"or"} [operator = 'or'] How the filters are combined together.
- * @property {number} [limit = undefined] The max number of items to display when
+ * @property {number} [limit = 10] The max number of items to display when
  * `showMoreLimit` is not or if the widget is showing less value.
```
bobylito: I feel like there is a missing word here. "showMoreLimit is not **set**"

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -328,12 +330,11 @@ export default function connectRefinementList(renderFn) {
           createURL,
           instantSearchInstance,
         } = renderOptions;
-        const items = results
-          .getFacetValues(attributeName, { sortBy })
-          .slice(0, this.getLimit())
-          .map(formatItems);
 
-        const hasExhaustiveItems = items.length < this.getLimit();
+        const rawItems = results.getFacetValues(attributeName, { sortBy });
```
bobylito: I wouldn't call that `rawItems` because `items` is not an unrefined state of the `items`. Why not `facetValues`?

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -328,12 +330,11 @@ export default function connectRefinementList(renderFn) {
...
+        const items = rawItems.slice(0, this.getLimit()).map(formatItems);
+
+        const hasExhaustiveItems = rawItems.length <= this.getLimit();
```
bobylito: 👍

PR #2521, opened 2017-10-24
Fix connect hierarchical menu

iam4x:
When:

If there is already`currentConfiguration.hierarchicalFacets` in the configuration, we would check if it's the same added and throw a warning.

Bug:

If it's not the same `hierarchicalFacets` we were returning an empty object as well.


bobylito:
Can you add a test for that? And check the breadcrumb implementation too?

PR #2527, opened 2017-10-26
[POC] GeoSearch

samouss:
**Summary**

👋 This is a POC for the GeoSearch (with Google Maps for now). Take a look at the **behaviour only** don't look at the code it's not the **point of the PR**. Since we have preview with Netlify, this is a convenient way to receive feedbacks 🙂

- Initially the search will refine with the IP **or** the given position
- When the map is moved the results will be filter with the bounding box
- The widget don't compute any radius automatically because (IMO) it depends a lot of the use case, you can still configure it at the instantiation.
- When no results are found, we keep the same bounds & zoom that has been set with the last results found. We avoid to lost the context of the search for the user.

Again check the behaviour of the search, I will replicate the Google Maps example with Leaflet (at least).

Some example of GeoSearch:

- [Airbnb](https://www.airbnb.com/s/New-York--NY--United-States/homes?refinement_path=%2Fhomes&allow_override%5B%5D=&place_id=ChIJOwg_06VPwokRYv534QaPC8g&s_tag=SZqkWyWk)
- [Yelp](https://www.yelp.com/search?find_desc=Restaurants&find_loc=New+York,+NY&start=0&l=g:-73.72465136926621,40.7384032871027,-74.15586474817246,40.47776024357566)
- [Foursquare](https://foursquare.com/explore?mode=url&near=New%20York%2C%20NY%2C%20United%20States&nearGeoId=72057594043056517)
- [Booking](https://www.booking.com/searchresults.html?aid=397596;label=gog235jc-index-XX-XX-XX-unspec-fr-com-L%3Axu-O%3AosSx-B%3Achrome-N%3AXX-S%3Abo-U%3Ac-H%3As;sid=2e5374e1379c747207f5be99e2a4a730;city=-1456928;from_idr=1;index_postcard=1&;ilp=1;d_dcp=1;i_upcc=1#map_opened-map-header-cta)

Not yet implemented: 
- clustering
- shape

**Result**

You can play with the `GeoSearch` widget on [dev-novel](https://deploy-preview-2527--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=GeoSearch). You need to reload the page when you switch between the stories otherwise the map is a bit lost.

Ideas of keywords for the search: 
- `studio` 
- `cosy`

### DIFF
--- a/dev/app/built-in-widgets/geo-search-google-maps.stories.js
+++ b/dev/app/built-in-widgets/geo-search-google-maps.stories.js
@@ -27,7 +27,7 @@ export default () => {
     'with IP',
     wrapWithHitsAndConfiguration(container => {
       window.search.addWidget(
-        instantsearch.widgets.geoSearch({
+        instantsearch.widgets.geoSearchWithGoogleMaps({
           container,
         })
       );
@@ -37,7 +37,7 @@ export default () => {
       'with IP & radius',
       wrapWithHitsAndConfiguration(container => {
         window.search.addWidget(
-          instantsearch.widgets.geoSearch({
+          instantsearch.widgets.geoSearchWithGoogleMaps({
             container,
             radius,
           })
@@ -48,7 +48,7 @@ export default () => {
       'with IP & radius & precision',
       wrapWithHitsAndConfiguration(container => {
         window.search.addWidget(
-          instantsearch.widgets.geoSearch({
+          instantsearch.widgets.geoSearchWithGoogleMaps({
             container,
             radius,
             precision,
@@ -62,7 +62,7 @@ export default () => {
     'with position',
     wrapWithHitsAndConfiguration(container => {
       window.search.addWidget(
-        instantsearch.widgets.geoSearch({
+        instantsearch.widgets.geoSearchWithGoogleMaps({
           container,
           position,
         })
@@ -73,7 +73,7 @@ export default () => {
       'with position & radius',
       wrapWithHitsAndConfiguration(container => {
         window.search.addWidget(
-          instantsearch.widgets.geoSearch({
+          instantsearch.widgets.geoSearchWithGoogleMaps({
             container,
             radius,
             position,
@@ -85,7 +85,7 @@ export default () => {
       'with position & radius & precision',
       wrapWithHitsAndConfiguration(container => {
         window.search.addWidget(
-          instantsearch.widgets.geoSearch({
+          instantsearch.widgets.geoSearchWithGoogleMaps({
             container,
             radius,
             precision,
@@ -113,7 +113,7 @@ export default () => {
       );
 
       window.search.addWidget(
-        instantsearch.widgets.geoSearch({
+        instantsearch.widgets.geoSearchWithGoogleMaps({
           container: mapElement,
           radius: 20000,
           enableGeolocationWithIP: false,
@@ -137,7 +137,7 @@ export default () => {
       );
 
       window.search.addWidget(
-        instantsearch.widgets.geoSearch({
+        instantsearch.widgets.geoSearchWithGoogleMaps({
           container: mapElement,
           radius: 20000,
           enableGeolocationWithIP: false,
@@ -152,7 +152,7 @@ export default () => {
     'disable refine on move',
     wrapWithHitsAndConfiguration(container => {
       window.search.addWidget(
-        instantsearch.widgets.geoSearch({
+        instantsearch.widgets.geoSearchWithGoogleMaps({
           container,
           enableRefineOnMapMove: false,
         })
--- a/dev/app/built-in-widgets/geo-search-google-maps.stories.js
+++ b/dev/app/built-in-widgets/geo-search-google-maps.stories.js
@@ -120,31 +120,6 @@ export default () => {
         })
       );
     })
-  ).add(
-    'with position from Places with refine disable on move',
-    wrapWithHitsAndConfiguration(container => {
-      const placesElemeent = document.createElement('input');
-      const mapElement = document.createElement('div');
-
-      container.appendChild(placesElemeent);
-      container.appendChild(mapElement);
-
-      window.search.addWidget(
-        instantsearchPlacesWidget({
-          container: placesElemeent,
-          defaultPosition: [37.7749, -122.4194],
-        })
-      );
-
-      window.search.addWidget(
-        instantsearch.widgets.geoSearchWithGoogleMaps({
-          container: mapElement,
-          radius: 20000,
-          enableGeolocationWithIP: false,
-          enableRefineOnMapMove: false,
-        })
-      );
-    })
   );
 
   // Only UI
--- a/dev/app/built-in-widgets/geo-search-google-maps.stories.js
+++ b/dev/app/built-in-widgets/geo-search-google-maps.stories.js
@@ -149,14 +149,51 @@ export default () => {
 
   // Only UI
   Stories.add(
-    'disable refine on move',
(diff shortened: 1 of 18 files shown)


Haroenv:
Could google maps react be a peer dependency instead so that it isn't included in the build by default?

samouss:
Yep, it's a huge dependency and a lot of users don't have the use case for implement a GeoSearch. But first of all we are not sure to pick Google Maps as default implementation, probably but not sure.

bobylito:
When moving the map around, some pins may be on the exact edge. We should probably have a bounding smaller than the actual viewport of the map, no?

marielaures:
The POC is great!!
Just a small thing: I noticed when that sometimes the bounds move a little too much (screenshot attached)
geosearch

samouss:
@bobylito Completely agree! I didn't implemented it (yet) because when I have searched for this kind of behaviour it looked like a bit difficult to do. But for the final implementation we need it.

mthuret:
Great POC @samouss ! :tada: 

We definitely need a places widget on the other IS libraries as it complete perfectly the geo search use case.  

I have some questions about the future widget itself: 

- do we want to embed the search as you move option inside our Geo Widget? 
- do we want to embed a draw option (like foursquare)?

samouss:
@mthuret 

> do we want to embed the search as you move option inside our Geo Widget?

For the search as you move, IMO the use case is valid because the engine is really fast and the experience is not degraded. But you can provide [an options for disable this behaviour](https://deploy-preview-2527--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=GeoSearch%20%28Google%20Maps%29.disable%20refine%20on%20move). We can also invert the behaviour, disable the search as move by default and pass an options to enable it.

> do we want to embed a draw option (like foursquare)?

IMO it's overkill to support this feature in the widget out of the box. You can already refine with bounding box, drawing a shape on it don't offer many benefits. Maybe we can provide an API through the connector for enable this behaviour. But I'm not really convinced by the usage.

WDYT?

mthuret:
@samouss I'm sorry for the misunderstanding, indeed I was talking about having the option exposed to the user of the website (like the airbnb experience). 

About the draw options, I believe that can be useful when you want to let your user select a very restricted area to find a place for instance (and not programatically defined a bounding box). But agreed probably too overkill for now :).

samouss:
@mthuret

> @samouss I'm sorry for the misunderstanding, indeed I was talking about having the option exposed to the user of the website (like the airbnb experience).

Yes why not expose this option to the user, the experience will probably be better! 👍

bobylito:
> Yes why not expose this option to the user, the experience will probably be better! 👍

If we're talking about adding the option to let the user draw squares of zones to refine, I have a few questions:
 - How much more work does it require regarding the integration with mapping library?
 - Does it couple instantsearch more with the mapping library?
 - Is it something that can be implemented in the next iteration of the widget?
 - Is it core to the feature?
 - Isn't it a new way of interacting with the map altogether?

Obviously I don't have the answers :) Let's just see if it is reasonable to add that at this stage.

samouss:
@bobylito No no it was about adding the option to let the user activate (or not) the "Search as you move" directly on the map like the [Airbnb](https://www.airbnb.com/s/New-York--NY--United-States/homes?refinement_path=%2Fhomes&ne_lat=40.76474044143687&ne_lng=-73.85727574021041&sw_lat=40.55777254873592&sw_lng=-74.06120945603072&zoom=12&search_by_map=true&allow_override%5B%5D=&s_tag=GymkmeYI) component.

IMO we can keep the feature for an upcoming version.

bobylito:
> @bobylito No no it was about adding the option to let the user activate (or not) the "Search as you move" directly on the map like the Airbnb component.
> 
> IMO we can keep the feature of drawing shape for an upcoming version.

Oh ok my bad :) 👍

samouss:
I updated the PR with the Leaflet implementation, you can try it if you want. The behaviour & features are the same as Google Maps.

samouss:
@mthuret I added the opportunity to let the user control the refinement strategy. You can now enable / disable the "Search as you move" directly from the widget. It's available on both implementation (GoogleMaps & Leaflet), again don't look the code the only thing that is interesting is the behaviour.

Haroenv:
A [mapbox-gl](https://www.mapbox.com/mapbox-gl-js/api/) example would be cool too (my preferred map 😄). API should be pretty similar to leaflet, but different. 

Is there some way where we could have people insert their own copy of the mapping API instead of depending on it

samouss:
> A mapbox-gl example would be cool too (my preferred map 😄). API should be pretty similar to leaflet, but different.

@Haroenv It's planned! I'll do it now 😉

> Is there some way where we could have people insert their own copy of the mapping API instead of depending on it, letting them do dependency injection?

Not sure to have understood correctly. But with the `GeoSearch` connector that's the idea. We will deliver by default an implementation with Google Maps I think. Then the user can implement the view for any map provider.

vvo:
Thanks for the POC!

I tried it, here are my remarks:
- do you have an example that reuses Google Places Autocomplete instead of Algolia Places?
- how do you think we could link hits and pins (AirBnB style search). Where when you click on a pin then the UI scrolls to the corresponding hit/activate it and it works both ways (click on hit open PIN). That's what many users have been looking to do with Geo search. Being able to reproduce the AirBnB experience.
- following ^ a big struggle to reproduce this AirBnB experience was being able to have custom pins, with Gmaps it can be a hard task
- when I switch stories from stories the map does not displays, I have to reload the page fully (maybe just a dev-novel issue)
- do you plan on creating an API draft document (options, usage) for widget and connector?
- did you already decide if you wanted to have a full blown widget or only a very smart connector that is easy to link to any mapping lib?

samouss:
> - do you have an example that reuses Google Places Autocomplete instead of Algolia Places?

Nope, I don't have such example. It could be interesting to build a widget like Algolia Places but for Google Maps.

> - how do you think we could link hits and pins (AirBnB style search). Where when you click on a pin then the UI scrolls to the corresponding hit/activate it and it works both ways (click on hit open PIN). That's what many users have been looking to do with Geo search. Being able to reproduce the AirBnB experience.

It's definitely required for the release, I didn't put this logic inside the POC because IMO it's not restricted to the `GeoSearch` widget. Plus having a mechanism in order to communicate between widget is really specific to each implementation of InstantSearch.

> - following a big struggle to reproduce this AirBnB experience was being able to have custom pins, with Gmaps it can be a hard task

For the default widget we can provide an option for the markers customisation. In the case where the user choose the connector, it's his responsibility to customise the markers.

> - when I switch stories from stories the map does not displays, I have to reload the page fully (maybe just a dev-novel issue)

I think it is, I will fix it anyway when I will implement the widget.

> - do you plan on creating an API draft document (options, usage) for widget and connector?

Yep, it's plan! I implemented this example in order to have some feedbacks, I will make a draft of the connector API probably next week.

> - did you already decide if you wanted to have a full blown widget or only a very smart connector that is easy to link to any mapping lib?

The plan is to build a connector that is easy to plug to any mapping library & to build a widget with Google Maps as default implementation. During the draft process I will try to put most of the logic inside the connector. But if the API becomes too complicated to use, I will probably delegate some view logic to the renderer.

samouss:
@Haroenv I have added the Mapbox example! Don't look at the ugly pin 😅 Mapbox doesn't provide a default one, I added a simple one with CSS.

samouss:
Will close this one since we gathered enough feedbacks on the UI.

Thanks all of you for your precious feedbacks  👍

PR #2530, opened 2017-10-28
fix(url-sync): make URLSync consistent even if search is tampered

### LINKED ISSUE #2523: urlSync combined with specific searchFunction does not updates on first letter

bug

Test case: https://jsfiddle.net/8bqdcrg3/8/, this console.log the location.href continuously.

When using this searchFunction:

```
{
    var query = helper.state.query;
    // with this code the url is not updated on first letter
    // while the search is being done
    if (query === '') {
	    return;
    }
    helper.search();
  },
}
```

The urlSync will not be triggered for the first typed letter, but only at the second one.

urlSync must be triggered for the first letter

all

I don't know

2/

bobylito:
When there is a search function that prevents the initial rendering,
the first actual change should be able to update the URL.

Fix #2523

### DIFF
--- a/src/lib/url-sync.js
+++ b/src/lib/url-sync.js
@@ -120,7 +120,7 @@ class URLSync {
     );
   }
 
-  init({state}) {
+  init({ state }) {
     this.initState = state;
   }
 
@@ -144,7 +144,6 @@ class URLSync {
 
       const initStateQs = this.getQueryString(this.initState);
       const stateQs = this.getQueryString(state);
-      console.log(initStateQs, stateQs);
       if (initStateQs !== stateQs) {
         // force update the URL, if the state has changed since the initial URL read
         // We do this in order to make a URL update when there is search function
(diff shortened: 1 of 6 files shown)


bobylito:
The dev-novel should be updated in order to be able to see the action logger :)

src/lib/url-sync.js
```diff
@@ -132,11 +136,22 @@ class URLSync {
...
+  render({ helper, state }) {
     if (this.firstRender) {
       this.firstRender = false;
       this.onHistoryChange(this.onPopState.bind(this, helper));
-      helper.on('change', state => this.renderURLFromState(state));
+      helper.on('change', s => this.renderURLFromState(s));
+
+      const initStateQs = this.getQueryString(this.initState);
+      const stateQs = this.getQueryString(state);
+      console.log(initStateQs, stateQs);
```
samouss: Clean the `console.log` call

bobylito:
I added a command to run eslint with prettier because I don't like autoreformat and it adds a dependency in the editor.

bobylito:
I updated the code @samouss :)

bobylito:
Thanks for the review @samouss I'm just waiting for the next version of dev-novel

PR #2544, opened 2017-10-31
Chore split stories

### LINKED ISSUE #2540: Split stories in different files
What do you think to split `stories` across multiple files? 
IMO it's easier to navigate instead of accumulate hundred of line inside the same file.

ex: `Menu.stories.js`, `RefinementList.stories.js`, ...

samouss:
**Summary**

Fix #2540 

**Result**

screen shot 2017-10-31 at 19 16 04


dev/app/builtin/stories/analytics.stories.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable import/default */
+
+import { action, storiesOf } from 'dev-novel';
+import instantsearch from '../../../../index';
+import { wrapWithHits } from '../../utils/wrap-with-hits.js';
+
+const stories = storiesOf('Analytics');
+
+export default () => {
```
bobylito: Why does the code export a function?

Haroenv: to not have side-effect on import maybe?

samouss: Actually, I just reproduce the same pattern than before. But good question @bobylito , and after some research @Haroenv has right. Since we have different kind of stories (builtin, jquery, vanilla) if we don't apply them explicitly, they will all be loaded by default.

Issue #2552, opened 2017-11-06
RefinementList searchForItems


iam4x:
The connector `connectRefinementList` output two different structure of item, the label is missing when using the searchForItems:

* without search:
![image](https://user-images.githubusercontent.com/893837/32441371-9a120406-c2f7-11e7-8879-baf1adf132c4.png)

* with search:
![image](https://user-images.githubusercontent.com/893837/32441386-a881105e-c2f7-11e7-917e-3c5e896a71da.png)

----

Another issue, in the dev-novel example https://community.algolia.com/instantsearch.js/v2/dev-novel/?selectedStory=RefinementList.with%20search%20inside%20items the searchbar is always disabled.

```jsx
const shouldDisableSearchInput =
      this.props.searchIsAlwaysActive !== true &&
      !(this.props.isFromSearch || !this.props.hasExhaustiveItems);
```

It comes from the `this.props.hasExhaustiveItems` which I changed recently.

bobylito:
I found the second part of the bug on friday. I'm almost done with the fix

bobylito:
Also related to https://github.com/algolia/instantsearch.js/issues/2454

bobylito:
This is fixed in 2.2.3

PR #2554, opened 2017-11-06
fix(refinementList): fix facet exhaustivity check

bobylito:
This check depends on whether or not there is another widget setting the
max value per facet to a bigger value, we can't be sure that if there is
the same number of items than the limit that it is exhaustive.

Fixes the other part of #2552


Haroenv:
Test on CI bailed out because eslint wasn't followed; `yarn lint:js --fix` should do the trick

iam4x:
Nice catch, I didn't thought it would have come from here 👍

PR #2565, opened 2017-11-09
 fix(numericSelector): make default value possible 

bobylito:
When we created the connectNumericSelector, we made sure to support `undefined` as a special value to unset the filters. However because the implementations are DOM based, the value is now a string. This patch fixes that (and adds the documentation to the feature)

I also updated the example because it was super hard to maintain and didn't have much sense in terms of feature. The new example also show case the undefined value :)

### DIFF
--- a/src/components/Selector.js
+++ b/src/components/Selector.js
@@ -20,13 +20,13 @@ export class RawSelector extends React.Component {
       <select
         className={this.props.cssClasses.select}
         onChange={this.handleChange}
-        value={currentValue}
+        value={`${currentValue}`}
       >
         {options.map(option => (
           <option
             className={this.props.cssClasses.item}
             key={option.label + option.value}
-            value={option.value}
+            value={`${option.value}`}
           >
             {option.label}
           </option>
--- a/src/connectors/numeric-selector/__tests__/connectNumericSelector-test.js
+++ b/src/connectors/numeric-selector/__tests__/connectNumericSelector-test.js
@@ -235,4 +235,57 @@ describe('connectNumericSelector', () => {
       refine = renderingParameters.refine;
     });
   });
+
+  it('The refine function can unselect with `undefined` and "undefined"', () => {
+    const rendering = sinon.stub();
+    const makeWidget = connectNumericSelector(rendering);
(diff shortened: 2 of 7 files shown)


src/connectors/numeric-selector/__tests__/connectNumericSelector-test.js
```diff
@@ -235,4 +235,57 @@ describe('connectNumericSelector', () => {
       refine = renderingParameters.refine;
     });
   });
+
+  it('The refine function can unselect with `undefined` and "undefined"', () => {
+    const rendering = sinon.stub();
```
samouss: Prefer usage of `jest.fn`, at some point we should remove `sinon`

src/components/Selector.js
```diff
@@ -20,13 +20,13 @@ export class RawSelector extends React.Component {
       <select
         className={this.props.cssClasses.select}
         onChange={this.handleChange}
-        value={currentValue}
+        value={`${currentValue}`}
```
samouss: Why did you set them to string? If we keep value as number, it's still works. The coercion is made by the browser.

src/connectors/numeric-selector/__tests__/connectNumericSelector-test.js
bobylito: Yes but if I move to jest.fn, I need to do the whole file. You might say "yes exactly do it and move like that each time you encounter sinon" and I would have to agree. OK :)

src/components/Selector.js
bobylito: I will try then. :)

bobylito: Oh know, I have to transform to string because of `undefined`...

samouss: Yes you are right... Maybe at some point we should also support empty string as reset value. With it we don't need to cast number to string in the template, just provide the default value as `""`.

bobylito: Well I think the undefined is the most natural move since you can completely avoid the `value` and I don't think we should support more than one "0" value. However we could change it to `""` if you and the team think it's best.

PR #2566, opened 2017-11-09
chore(linting): target correct folder

### LINKED ISSUE #2443: Avoid linting the generated documentation
The configuration for `sass-lint` run the linter on the generated folder `docs` instead of `docgen`.

Haroenv:
closes #2443

**Summary**

sass-lint was being run on the generated docs instead of the source folder

**Result**

* target correct folder
* ignore correct folders
* make hex rule a warning for now because we aren't consistent with it, we can fix it later when we want less warnings


bobylito:
Will give it a try at fixing the sass lint problems :)

bobylito:
I did a pass on the rules and made the changes accordingly. I just have some questions about the rules around nesting and qualified elements. I'm pretty we should not enforce them. WDYT?

samouss:
@bobylito Nice! IMO the nesting rule is pretty useless it's harder to read with the nested selector in some cases. For the other don't have a strong opinion on it..

bobylito:
@LukyVj @ronanlevesque maybe you can help us decide if we should ditch the last rules?

.sass-lint.yml
```diff
@@ -89,6 +87,7 @@ rules:
   space-before-bang: 1
   space-after-bang: 1
   space-between-parens: 1
+  space-around-operator: 0 # because it's hard to read when you have multiple value set with operators (eg. padding)
```
ronanlevesque: I think it's more readable when set to 1. Any counter-example?

docgen/src/stylesheets/components/_code-highlighting.scss
```diff
@@ -7,41 +7,111 @@
...
+  color: inherit;
+}
+.cm-s-mdn-like .cm-positive {
+  color: #690;
+  background: #EFE;
+}
+.cm-s-mdn-like .cm-negative {
+  color: red;
+  background: #FEE;
+}
```
ronanlevesque: An way to enforce space between selectors with sass-lint? This is not very readable.

ronanlevesque:
I would keep the `no-qualifying-elements` set to `false`. There are some edge cases when one might need to rely on them to enforce specific styles. Same for `nesting-depth`. It's hard to define an arbitrary depth for this as it can depends on a lot of things (eg nesting a modifier, then nesting a hover behaviour on the modifier etc.)

.sass-lint.yml
bobylito: Yes sure:
```sass
padding: 13 * $var 13 * $var; // this is what the rule would enforce
// vs
padding: 13*$var 14*$var;
```

docgen/src/stylesheets/components/_code-highlighting.scss
bobylito: What do you mean?

.sass-lint.yml
ronanlevesque: Oh ok, nevermind, thought `0` would enforce no space. I'm good with this then 👍

docgen/src/stylesheets/components/_code-highlighting.scss
ronanlevesque: This:

```

Comment (ronanlevesque) [reply docgen/src/stylesheets/components/_code-highlighting.scss]: This:

```
 .cm-s-mdn-like.CodeMirror {
   color: #999;
   background-color: #FFF;
 }

 .cm-s-mdn-like .CodeMirror-selected {
   background: #CFC;
 }

 .cm-s-mdn-like .CodeMirror-line::selection, .cm-s-mdn-like .CodeMirror-line > span::selection, .cm-s-mdn-like .CodeMirror-line > span > span::selection {
   background: #CFC;
 }

 .cm-s-mdn-like .CodeMirror-line::-moz-selection, .cm-s-mdn-like .CodeMirror-line > span::-moz-selection, .cm-s-mdn-like .CodeMirror-line > span > span::-moz-selection {
   background: #CFC;
 }
```

Instead of this:

```
 .cm-s-mdn-like.CodeMirror {
   color: #999;
   background-color: #FFF;
 }
 .cm-s-mdn-like .CodeMirror-selected {
   background: #CFC;
 }
 .cm-s-mdn-like .CodeMirror-line::selection, .cm-s-mdn-like .CodeMirror-line > span::selection, .cm-s-mdn-like .CodeMirror-line > span > span::selection {
   background: #CFC;
 }
 .cm-s-mdn-like .CodeMirror-line::-moz-selection, .cm-s-mdn-like .CodeMirror-line > span::-moz-selection, .cm-s-mdn-like .CodeMirror-line > span > span::-moz-selection {
   background: #CFC;
 }
```

Comment (Haroenv) [reply docgen/src/stylesheets/components/_code-highlighting.scss]: https://github.com/sasstools/sass-lint/blob/develop/docs/rules/empty-line-between-blocks.md, right @ronanlevesque

Comment (ronanlevesque) [reply docgen/src/stylesheets/components/_code-highlighting.scss]: Exactly!

Comment (bobylito) [reply docgen/src/stylesheets/components/_code-highlighting.scss]: Ok I will add that

**Comment by bobylito** [comment]
> I would keep the no-qualifying-elements set to false. There are some edge cases when one might need to rely on them to enforce specific styles. Same for nesting-depth.

Ok. What about I disable those rules for some files using `// sass-lint:disable`? @ronanlevesque 

The idea is that we are only using warning as it is the only to get all the errors at once. Of course then we should make sure that we never have warnings :)

**Comment by ronanlevesque** [comment]
Yes, that makes sense.

**Comment by bobylito** [comment]
It's updated @Haroenv @ronanlevesque if you want to have a look :)

File: src/css/default/_price-ranges.scss [hunk 3]
```diff
@@ -1,3 +1,4 @@
+// sass-lint:disable no-empty-rulesets
```
Haroenv: Why are they empty? Could we maybe get rid of the whole css file in some dead code elimination step?

bobylito: It's a template for those who want to use them as a starting point for making their own theme.

PR #2569, opened 2017-11-09
feat(core): provide information about stalled search to widgets

bobylito:
This PR is an alternative implementation of #2509. This is currently a PoC.

At the core of this new implementation is a change when the render part of the lifecycle of the widget. With this implementation, the render can be called when the instantsearch detects that a search is stalled. This means that they will be called more often. However, this is not likely to have a big impact on the performance as it should be the responsability of the widgets to render only when necessary.

On the implementation of the connector and the widget, the work is trivial. It is merely a question of providing the `isSearchStalled` flag to the rendering. The logic is factored in instantsearch, which means that the usage in other widgets will be as easy.

Having done this implementation, I think it's the way to go. WDYT? Happy to answer any questions.

PS: this new implementation has been inspired by what can be done in react-instantsearch and vue-instantsearch. If we were to implement them using this library, this means that we need to fit their logic.

PPS: this can be tested here: https://deploy-preview-2569--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=SearchBox.default


vvo:
I like the implementation, as discussed earlier in a meeting, some questions still:

> However, this is not likely to have a big impact on the performance
Can you try a small manual perf test in the browser to see if there's any obvious issue/diff with this implementation?

> It should be the responsability of the widgets to render only when necessary.

Can you provide an example of a widget rendering that would avoid re-rendering when the widget do not care about the isSearchStalled flag? Just curious here how to do it.

Finally, for reference, can you provide a small concrete example of what's happening under the hood in a scenario where this is implemented? Something like:
- user types 'a'
- flag becomes true
- widget render once..

Something like that. So that we all understand fully what's happening.

Haroenv:
This looks really nice, I think we should go this route. I guess we can add some componentShouldUpate hooks?

bobylito:
> Can you provide an example of a widget rendering that would avoid re-rendering when the widget do not care about the isSearchStalled flag? Just curious here how to do it.

That's a very speculative on my side here I guess. I can try to implement it on a widget / connector to see what happens.

> Something like that. So that we all understand fully what's happening.

Ok I'll do a schema.

> This looks really nice, I think we should go this route. I guess we can add some componentShouldUpate hooks?

I don't know if that's necessary but that can be done, indeed.

bobylito:
I will move with this solution and close the other. Thanks for the feedbacks :)

bobylito:
new rendering step visually

bobylito:
Hello everyone,

This is the final version. Here is the API:

InstantSearch constructor / factory has a new option: [stalledSearchDelay](https://deploy-preview-2569--algolia-instantsearch.netlify.com/v2/instantsearch.html#struct-InstantSearchOptions-stalledSearchDelay) which configures the time before the stalled search kicks in.
Also InstantSearch provides a new searchMetadata attribute with the options sent to the widgets at `render`. It's not provided at `init` because there is no search yet (therefore no metadata about search). The searchMetadata is an object with a single key `isSearchStalled`.

We have a new set of [options for the loading indicator](https://deploy-preview-2569--algolia-instantsearch.netlify.com/v2/widgets/searchbox#struct-SearchBoxWidgetOptions-loadingIndicator) on the search box: 

```js
{
  /* ... other searchbox options */
  loadingIndicator: true,
  // or
  loadingIndicator: {
    template,
    cssClasses: {root: ''}
  } 
}
```

The connector provides a new information to the searchbox rendering: isSearchStalled, if the search is stalled.

Haroenv:
CI failed cause one commit message was too long (merge 1d0aa21feba4607f456385c4606468481a5c893d) so you might need to rebase

bobylito:
Thanks @Haroenv :D

src/widgets/search-box/search-box.js
```diff
@@ -155,6 +183,12 @@ searchBox({
  * @property {{root: string}} [cssClasses] CSS classes added to the reset buton.
  */
 
+/**
+ * @typedef {Object} SearchBoxLoadingIndicatorOption
+ * @property {function|string} template Template used for displaying the button. Can accept a function or a Hogan string.
+ * @property {{root: string}} [cssClasses] CSS classes added to the reset buton.
```
samouss: Should be the loading indicator root element.

src/widgets/search-box/search-box.js
```diff
@@ -173,8 +207,9 @@ searchBox({
  * @property  {string|HTMLElement} container CSS Selector or HTMLElement to insert the widget.
  * @property  {string} [placeholder] Input's placeholder.
  * @property  {boolean|SearchBoxPoweredByOption} [poweredBy=false] Define if a "powered by Algolia" link should be added near the input.
- * @property  {boolean|SearchBoxResetOption} [reset=true] Define if a reset button should be added in the input when there is a query.
+ * @property  {boolean|SearchBoxResetOption} [reset=false] Define if a reset button should be added in the input when there is a query.
```
samouss: Is it an intended change? The default value is still true in the code.

src/connectors/search-box/connectSearchBox.js
```diff
@@ -35,6 +35,8 @@ Full documentation available at https://community.algolia.com/instantsearch.js/c
  * @property {function(string)} refine Sets a new query and searches.
  * @property {function()} clear Remove the query and perform search.
  * @property {Object} widgetParams All original `CustomSearchBoxWidgetOptions` forwarded to the `renderFn`.
+ * @property {boolean} isSearchStalled `true` if the search results takes more than a certain time to come back
```
vvo: Should we document "a certain time"? I would be curious as a dev to know more. This may be too vague

src/widgets/search-box/search-box.js
```diff
@@ -124,6 +131,27 @@ const renderer = ({
...
+}) {
+  const input = getInput(containerNode);
+  const isFocused = document.activeElement === input;
+  if (!isFocused && query !== input.value) {
+    input.value = query;
+  }
+
+  if (loadingIndicator) {
+    if (isSearchStalled) {
+      containerNode.firstChild.classList.add('stalled-search');
```
vvo: Should this CSS class name follow ais- conventions? And be documented somewhere?

src/widgets/search-box/search-box.js
vvo: Is it false by default now to have a reset button?

src/widgets/search-box/search-box.js
```diff
@@ -173,8 +207,9 @@ searchBox({
  * @property  {string|HTMLElement} container CSS Selector or HTMLElement to insert the widget.
  * @property  {string} [placeholder] Input's placeholder.
  * @property  {boolean|SearchBoxPoweredByOption} [poweredBy=false] Define if a "powered by Algolia" link should be added near the input.
- * @property  {boolean|SearchBoxResetOption} [reset=true] Define if a reset button should be added in the input when there is a query.
+ * @property  {boolean|SearchBoxResetOption} [reset=false] Define if a reset button should be added in the input when there is a query.
  * @property  {boolean|SearchBoxMagnifierOption} [magnifier=true] Define if a magnifier should be added at beginning of the input to indicate a search input.
+ * @property  {boolean|SearchBoxLoadingIndicatorOption} [loadingIndicator=false] Define if a loading indicator should be added at beginning of the input to indicate that search is currently stalled.
```
vvo: Should we make it true by default, since it's a new feature and might bring good behaviour by default?

src/connectors/search-box/connectSearchBox.js
bobylito: indeed there is a default value. Will doc that.

src/widgets/search-box/search-box.js
bobylito: Will double check 👍

src/widgets/search-box/search-box.js
bobylito: I was wondering. The idea behind `false` by default is that we don't change any existing behavior / style, especially if the user doesn't use the provided stylesheets.

src/widgets/search-box/search-box.js
bobylito: Oh yes sure. Also it doesn't work if you use an input as container :/

vvo: Maybe document that part, but I won't be surprised as a dev for it not to be supported.

src/widgets/search-box/search-box.js
vvo: Yep that's the safest, if we were to do a new breaking version we would make it true by default, cc @mthuret can we add that to the feedback/product loop?

bobylito: Happy to break in v3 ;)

src/connectors/search-box/connectSearchBox.js
bobylito: Done (only happen a few lines after :))

src/widgets/search-box/search-box.js
Haroenv: agreed, false for now, breaking to `true` in v3 👍

bobylito:
The behavior and loading indicator have been validated by @sebnvzt :D

PR #2575, opened 2017-11-13
fix(connectors): prefer wrappers over bind

bobylito:
Previously we were using bind, which will be applied at each init. This
is ok for one time mount, but since we can mount and unmount dynamically
now, it's a problem.

### DIFF
--- a/src/connectors/price-ranges/connectPriceRanges.js
+++ b/src/connectors/price-ranges/connectPriceRanges.js
@@ -158,13 +158,15 @@ export default function connectPriceRanges(renderFn, unmountFn) {
       },
 
       init({ helper, instantSearchInstance }) {
-        this._refine = this._refine.bind(this, helper);
+        this.refine = (...opts) => {
+          this._refine(helper, ...opts);
+        };
 
         renderFn(
           {
             instantSearchInstance,
             items: [],
-            refine: this._refine,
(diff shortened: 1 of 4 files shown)


Haroenv:
Hey 👋 looks like a good call. You still need to run prettier on this PR :)

bobylito:
> Hey 👋 looks like a good call. You still need to run prettier on this PR :)

Well the only problem is that I haven't touched the files failing...

bobylito:
I need to fix the tests here :/

src/connectors/price-ranges/connectPriceRanges.js
```diff
@@ -158,13 +158,15 @@ export default function connectPriceRanges(renderFn, unmountFn) {
       },
 
       init({ helper, instantSearchInstance }) {
-        this._refine = this._refine.bind(this, helper);
+        this.refine = (...opts) => {
+          this._refine(helper, ...opts);
```
Haroenv: refine only has two arguments, what about not spreading?

bobylito: That's indeed preferable to avoid spreading if we know the number of arguments 👍

bobylito:
Happy to take 5 minutes and go over the code to explain if you want @Haroenv

PR #2577, opened 2017-11-16
chore(doc): add experimental flag about new dynamic is API's 

bobylito:
This PR adds some documentation about the new method to add and remove widgets from InstantSearch. The new methods and behaviors are explicitly marked as experimental so that users are encouraged to give us feedback.

This PR also fixes the problem with the anchors hack that made the links unclickable if they were underneath one of the pseudo elements.


bobylito:
Can you double check the way the writing? @Haroenv 🙏

Haroenv:
yes, I read it and the writing seems okay @bobylito

PR #2585, opened 2017-11-19
fix(searchbox): fix usage of custom reset template

### LINKED ISSUE #2528: SearchBox Widget: Template options are janky/broken
*bug*

2.2.1 (latest as of post)

All I want to do is override the default `magnifier` and `reset` templates (use a different SVG).

There seem to be a number of oddities. First off, the options seems structured differently than most other core widgets. Instead of having a root level `templates`, and `cssClasses` props, it instead has root props of `magnifier` and `reset`. Nothing inherently broken there, just seems inconsistent (at least with the widgets I'm familiar with).

Second, the `cssClasses` prop doesn't really seem to work.

### Attempt 1:
Try and render a custom template for the magnifier

```
searchBox(
  container,
  magnifier: {
    template: myCustomSearchSvg,
  },
);
```

It rendered my svg! However, within an empty span…I was expecting it to still be wrapped by an element with an `.ais-search-box--magnifier` since, I didn't override `cssClasses`.

### Oh well, Attempt 2:
Try and manually add the class? 

```
searchBox(
  container,
    magnifier: {
      template: myCustomSearchSvg,
      cssClasses: [{root: 'ais-search-box--magnifier'}],
    },
);
```

Bubkis. Same result as Attempt 1.

### Attempt 3:
Try and render a custom template for the reset button

```
searchBox(
  container,
  reset: {
    template: 'x',
  },
);
```
Error!

> Uncaught TypeError: Cannot read property 'style' of null
    at search-box.js:124
    at Object.init (connectSearchBox.js:105)
    at InstantSearch.js:213
    at arrayEach (_arrayEach.js:15)
    at forEach (forEach.js:38)
    at InstantSearch._init (InstantSearch.js:211)
    at InstantSearch.start (InstantSearch.js:168)

Then I post this and gave up. :)

bobylito:
fix #2528. If we were to fix that completely we'd probably change the options to make them more consistent with the rest of the library and we'd fix the markup.

What this does:
 - adds two new css classes to help customize the reset and magnifying glass
 - fix the usage of the reset template (was just plain broken)


Haroenv:
Doesn't seem like you changed any test, but I think it was testing the block / none styling: https://travis-ci.org/algolia/instantsearch.js/builds/304400146#L917-L935

bobylito:
> Doesn't seem like you changed any test, but I think it was testing the block / none styling: https://travis-ci.org/algolia/instantsearch.js/builds/304400146#L917-L935

Oops fixing that :)

src/widgets/search-box/__tests__/search-box-test.js
```diff
@@ -243,7 +245,7 @@ describe('searchBox()', () => {
       simulateInputEvent('test', 'tes', widget, helper, state, container);
 
       // When
-      $('button[type="reset"]')[0].click();
+      $('.ais-search-box--reset-wrapper')[0].click();
```
Haroenv: the button is still type reset right?

bobylito: yes. The problem is that it is in the template, so we cannot rely on what's inside it for the tests.

bobylito: This was the source of the bug. If you were to use an emoji, there would not be a button (with type=reset) so instead the event listener is on the outer span.

package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "build": "./scripts/build.sh",
     "dev": "./scripts/dev.sh",
+    "dev-novel": "NODE_ENV=development webpack-dev-server --config dev/webpack.config.js",
```
samouss: We could probably reuse this command from the `dev.sh` script.

bobylito: What do you mean? This line was extracted from `dev.sh`.

samouss: I mean in the `dev.sh` we can do `yarn dev-novel` instead of having the same command in two different place.

bobylito: That's not a big improvement though. Let's iterate on that later.

PR #2589, opened 2017-11-22
docs(angular): add message about angular instantsearch

iam4x:
I have added a message about Angular InstantSearch to redirect users to the new library 👍


bobylito:
Smart move 👍

PR #2593, opened 2017-11-23
chore(travis): add bundlesize token

samouss:
**Summary**

Use the token from `.travis.yml` instead of use it from the Travis dashboard.


bobylito:
What's the benefit of doing this?

samouss:
All token are present in the configuration file. Before that it was set only on the dashboard.

bobylito:
> All token are present in the configuration file. Before that it was set only on the dashboard.

So the argument is consistency?

PR #2603, opened 2017-11-27
2.3 back in develop for the release

bobylito:
- **breadcrumb widget**: a new widget to visualize the path of a hierarchical menu / navigation
 - **range input widget**: an alternative view to the range slider based on inputs (min max)
 - *EXPERIMENTAL* dynamically remove and add widgets. This is tagged as experimental because we are trying it extensively, it's a major change in the core of IS.js. It might contains unforeseen edge case bugs. It is used on [Angular-InstantSearch](https://discourse.algolia.com/t/release-of-angular-instantsearch-beta/3731), go check it out if you haven't already.
 - **loading indicator**: on mobile, it's really frustrating to not have feedbacks when something goes wrong on the network latency. This adds a new spinner in the SearchBox and the information can be used in any widget.


Haroenv:
version and merge commits are too long @bobylito  https://travis-ci.org/algolia/instantsearch.js/builds/307845894#L1081

bobylito:
We still have a few issues with the commit validation. I've fixed some but the rest are from old commits, should be ok for the moment right?

PR #2610, opened 2017-12-01
Fix/widgets dispose

iam4x:
The method of the helper `removeHierarchicalFacetRefinement()` will throw an error if there is no facet refinement set.

We need to perform a check in `dispose()` methods before removing the facet refinement.


src/connectors/menu/connectMenu.js
```diff
@@ -233,9 +233,20 @@ export default function connectMenu(renderFn, unmountFn) {
...
-        const nextState = state
-          .removeHierarchicalFacetRefinement(attributeName)
-          .removeHierarchicalFacet(attributeName);
+        let nextState;
+
+        if (state.isHierarchicalFacetRefined(attributeName)) {
+          nextState = state.removeHierarchicalFacetRefinement(attributeName);
+        }
+
+        nextState = (nextState || state).removeHierarchicalFacet(attributeName);
```
bobylito: If you initialize `nextState` with `state` at l.236 you don't need to use `nextState || state`.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -197,9 +197,17 @@ export default function connectHierarchicalMenu(renderFn, unmountFn) {
...
-          .removeHierarchicalFacet(hierarchicalFacetName);
+        let nextState;
+
+        if (state.isHierarchicalFacetRefined(hierarchicalFacetName)) {
+          nextState = state.removeHierarchicalFacetRefinement(
+            hierarchicalFacetName
+          );
+        }
+
+        nextState = (nextState || state).removeHierarchicalFacet(
```
bobylito: If you initialize `nextState` with `state` at l.200 you don't need to use `nextState || state`.

PR #2615, opened 2017-12-04
fix(poweredBy): minify slightly and make into correct URL

### LINKED ISSUE #2613: Powered by doesn't display on Firefox
Bug

FF:
screen shot 2017-12-03 at 10 07 21

Cr: 
screen shot 2017-12-03 at 10 07 34

FF (tested on 57)

🤷‍♂️

Haroenv:
**Summary**

FireFox is very picky about data URIs and as soon as it's containing `#`, it's interpreting it as a link instead of a data URI. You can escape those `#` with `%23`

check this in FF: https://deploy-preview-2615--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=SearchBox.default

**Result**

fixes #2613


bobylito:
Looks good. Thanks for fixing that @Haroenv

Issue #2612, opened 2017-12-01
Webpack and Typescript, v2.3.0 Error: Can't resolve 'react'


Giovanni-Mattucci:
I'm using Webpack 3 and Typescript, and I have been including `instantsearch.js` in my files like so:
```typescript
import * as instantsearch from 'instantsearch.js';
```
This has been working fine.

I recently tried upgrading from `2.0.2` to `2.3.0` and [I followed your es6 example in your v2 documentation](https://community.algolia.com/instantsearch.js/v2/guides/usage.html#optimize-your-build-with-tree-shaking):
```typescript
import instantsearch from 'instantsearch.js/es';
import { searchBox } from 'instantsearch.js/es/widgets';
const search = instantsearch(/* parameters */);
search.addWidget(
    searchBox({
        container: '#search-input'
    })
);
```
However I am running into trouble.
I get the following errors when I try initializing a widget.
```bash
ERROR in ./node_modules/instantsearch.js/es/components/MenuSelect.js
Module not found: Error: Can't resolve 'react' in 'C:\Users\mattucci\Documents\Visual-Studio-2017\Projects\Ontariocolleges\Ontariocolleges\node_modules\instantsearch.js\es\components'
 @ ./node_modules/instantsearch.js/es/components/MenuSelect.js 11:0-41
 @ ./node_modules/instantsearch.js/es/widgets/menu-select/menu-select.js
 @ ./node_modules/instantsearch.js/es/widgets/index.js
 @ ./Content/Scripts/search/search-programs-instantsearch.ts
 @ ./Content/Scripts/pages/program-search.ts

ERROR in ./node_modules/instantsearch.js/es/widgets/menu-select/menu-select.js
Module not found: Error: Can't resolve 'react' in 'C:\Users\mattucci\Documents\Visual-Studio-2017\Projects\Ontariocolleges\Ontariocolleges\node_modules\instantsearch.js\es\widgets\menu-select'
 @ ./node_modules/instantsearch.js/es/widgets/menu-select/menu-select.js 1:0-26
 @ ./node_modules/instantsearch.js/es/widgets/index.js
 @ ./Content/Scripts/search/search-programs-instantsearch.ts
 @ ./Content/Scripts/pages/program-search.ts

ERROR in ./node_modules/instantsearch.js/es/widgets/menu-select/menu-select.js
[...]

PR #2618, opened 2017-12-06
fix: React reference

samouss:
**Summary**

Both widgets `RangeInput` & `Breadcrumb` still have some references on `react` / `react-dom` instead of using the `preact-compat` import. When the lib is used with a bundler (ex: Webpack) an error is thrown.

```
Uncaught Error: Cannot find module "react"
```

This PR aims to fix the reference of React inside the both widgets.

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.js
+++ b/src/components/Breadcrumb/Breadcrumb.js
@@ -1,4 +1,4 @@
-import React, { PureComponent } from 'react';
+import React, { PureComponent } from 'preact-compat';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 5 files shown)


bobylito:
> Not sure that it will fix this issue (#2612). Because the problem came from MenuSelect and not from RangeInput & Breadcrumb. Maybe the error is pointing the wrong place because there is not trace of react in the MenuSelect widget.

indeed 👍

bobylito:
Btw, tests are failing @samouss . Can you check?

samouss:
I'm on it.

PR #2619, opened 2017-12-06
Update search-box.js

rayrutjes:
Hope this makes sense.


bobylito:
That's indeed better. Thanks :) I would even say something about the fact the input will be wrapped  with some instantsearch.js dom?

PR #2633, opened 2017-12-21
chore(security): remove public tokens

vvo:
Following security reports, removing public tokens for saucelabs and
argos.

For Saucelabs we are now using https://docs.travis-ci.com/user/jwt which
is the recommended way to have external contributors PRs working (read
documentation on travis).

For Argos jwt is not implemented so it will just be disabled, we need to
be careful while reviewing for now (same setup as react-instantsearch).


.travis.yml
```diff
@@ -14,19 +16,16 @@ branches:
   - master
   - develop
   - feat/instantsearch.js/v2
-  - /^feat\/\d+\.\d+$/
+  - "/^feat\\/\\d+\\.\\d+$/"
```
vvo: was done by travis command line automatically

bobylito:
#securityLove Thanks :D

PR #2635, opened 2017-12-22
feat(wrapWithHits): enable async init

samouss:
**Summary**

This PR enable the possibility to init the story asynchronously. It can be useful for preload some assets before adding a widget without loading it for all the stories (ex: Google Maps).

**Result**

We can now call `wrapWithHits` with a function of one or two arguments. One argument means that il will be synchrone, two arguments means asynchronous. The second argument is the start function.

```js
// sync
wrapWithHits(container => {
  window.search.addWidget(
    instantsearch.widgets.hits({
      container: "#hits"
    })
  );
});

// async
wrapWithHits((container, start) => {
  window.search.addWidget(
    instantsearch.widgets.hits({
      container: "#hits"
    })
  );

  start();
});
```


samouss:
#2633 must be merged before in order to run the CI, because the current SauceLabs token has probably be revoked.

Issue #933, opened 2016-03-16
Should urlSync allows attribute names mapping?


fitztrev:
With the URL mapping from #922, is it possible to map hierarchical facets to a different name?

For example, here is my configuration and the query string it generates:

##### Current:

```javascript
mapping: {
    'q': 'query',
    'p': 'page',
    'hPP': 'hits',
    'hFR': 'filter',
},
```

```
?query=&page=0&filter[facets.facets.fits.make][0]=Ford&is_v=1
```

##### Desired:

Without changing the structure of the record inside my index, could I configure it so it's like this?

```
?query=&page=0&filter[make][0]=Ford&is_v=1
```

Or, better yet:

```
?query=&page=0&make[0]=Ford&is_v=1
```

vvo:
I think this is not feasible @fitztrev, you can only map the parameters names (q, query...) not the attribute names unfortunately.

There's no easy way to do so. It would require more changes in the helper: https://github.com/algolia/algoliasearch-helper-js/pull/281

vvo:
This will be provided in react-instantsearch

NicolasPK:
Hi @vvo,
Is this still not feasible? If so, is this feasible only by using the react implementation like you said back then?

Haroenv:
@nicolaspk, can you leave your opinion on the implementation in https://github.com/algolia/instantsearch.js/pull/2636 ? It seems like an interesting usecase that could be covered by that proposal

NicolasPK:
@Haroenv sounds like what I'm looking for indeed.
I guess a simpler configuration would've been appreciated (in the vein of the current `urlSync`), but I noticed that I'll be able to update the title as well so that sounds exactly like what I need ([by looking at this demo](http://weary-plants.surge.sh/app.js)).
This is the very last thing I need to do to fully replace our current (local) system by algolia's 🎉

Haroenv:
great news! feel free to leave a code snippet or fiddle so we can play around and figure out what broke your implementation too, or open a new issue for that

PR #2636, opened 2017-12-22
[POC] New URLSync mechanism

bobylito:
**Proof of concept** - not production ready.

This PR proposes a new API for synchronizing the search parameters with the URL. Here are the key goals of this new implementation:
 - deeper customization of the URL
 - handling the no-value / default value case and do not render any crufts
 - make the state more accessible
 - possibility to handle storages that are not URL's

First, what does not change with this new implementation: the lifecycle itself. The URL is stored when there is a new rendering and a new search is triggered when there is an update of the URL.

This PoC is an effort to make the internals customizable. So anything that is not related to the lifecycle is externalized as API's. We will provide a default implementation that will work but may not be to the taste of everyone, hence the customization part.

## UI State

One of the biggest change is that the widget can now provide UI-oriented views of the state. This is inspired by the state of react-instantsearch. This is this view that will be manipulated by the synchronization. Unlike react-instantsearch, this view is not the search parameters. If you want a widget to be synchronizable, it needs to implement two methods:
 - `getUIState`: this methods transforms the searchParameters into a view of the state of this widget and adds it in an object that is returned (API is like a reducer). The [implementation for the searchbox](https://github.com/algolia/instantsearch.js/blob/c7867ff4e705e009b90b22f145a7aff97340c7b0/src/connectors/search-box/connectSearchBox.js#L155-L162) reads the searchParameters and adds a key `query` in the UI state object with the value of the current query.
 - `getSearchParameters`: this methods does the opposite, reads a UI state and update the search parameters accordingly. Here is an [example with the searchbox](https://github.com/algolia/instantsearch.js/blob/c7867ff4e705e009b90b22f145a7aff97340c7b0/src/connectors/search-box/connectSearchBox.js#L164-L166)

This PoC implements the [synchronization for the refinementList](https://github.com/algolia/instantsearch.js/blob/c7867ff4e705e009b90b22f145a7aff97340c7b0/src/connectors/refinement-list/connectRefinementList.js#L413-L433) as well.

## Storage synchronization

For the synchronization itself, there are two aspects:
 - the mapping
 - the storage

This part has been split in two so that it's always possible to customization the structure without having to reimplement the storage which may be more tedious.

The UI state mapping is an object with two methods:
 - `toSyncable`
 - `fromSyncable`

The syncable object is the simplified structure that will be used directly by the storage. The storage shall not make any changes in this structure. We have currently a [single implementation](https://github.com/algolia/instantsearch.js/blob/c7867ff4e705e009b90b22f145a7aff97340c7b0/src/lib/uiStateMapping/simple.js) that just transfers the UIState without any changes.

The storage part is inspired by both the [work on react-instantsearch](https://github.com/algolia/react-instantsearch/issues/271) for an API to transfer the state of IS into another store and the experience that we gained with the URL sync.

The storage API has 5 methods:
 - `write`: writes a syncable object to the storage
 - `read`: reads the storage and gets a syncable object
 - `onUpdate`: registers a callback that will be triggered when the storage is updated
 - `createURL`: creates a URL. May not be applicable. For now a default implemented must be provided.
 - `dispose`: unregister the storage adapter. Cleans up the event listeners.

This PoC offers a [single implementation](https://github.com/algolia/instantsearch.js/blob/c7867ff4e705e009b90b22f145a7aff97340c7b0/src/lib/storage/modern.js) that is based on the code from the current URL sync and it uses the [modern browser history API.](https://developer.mozilla.org/en-US/docs/Web/API/History)

## Overview

reading from the url 2x
writing to the url 2x

## Demos
 - [with default implementations based on modern history API](https://disastrous-approval.surge.sh) ([source](http://disastrous-approval.surge.sh/app.js))
 - [With non-trivial mappings](http://weary-plants.surge.sh/Cameras%20&%20Camcorders/Binoculars?brand=Canon), brand refinement list data is mapped to brand query string parameter and the hierarchical menu is mapped to the path name of the URL - ([source](http://weary-plants.surge.sh/app.js))
 - [start page count at 1 in the URL](http://nebulous-balls.surge.sh/?page=3) - ([Source](http://nebulous-balls.surge.sh/app.js))

## Next steps
 - [x] Implement debouncing of the write. I'll try to implement this in the data storage in order to not clutter
 - [ ] Implement other widgets. I don't think this matter much in the context of this PoC unless you see some specific widget where it could be a problem. I'll have a look at react-instantsearch also.
 - [x] Try to implement a custom writer that will rewrite the path. What I have in mind is a hierarchical menu path in the URL and some custom query parameters.


Haroenv:
A big change is that it doesn't allow inputting arbitrary api parameters from the URL bar, which I think is a great idea, especially from a security POV, but I'm slightly wondering which usecases will be broken by not allowing refinement from arbitrary elements from the URL. 

Dynamic widgets might also be interesting to consider, I think it's fine here, where the URL is basically the source of truth, so widgets could be created, simply by being there in the URL, others could be left empty. I think this is a step forward for that direction and won't impede future explorations regarding dynamic behaviour.

Haroenv:
not necessarily part of the API, but it would be good if we saved the "default state" of a widget as `undefined`, which won't get serialised by `qs`, as well as making page 1-indexed (rather than 0-indexed). That way `?p=2` will behave like you'd expect and will only get added once you go to the next page

bobylito:
> not necessarily part of the API, but it would be good if we saved the "default state" of a widget as undefined, which won't get serialised by qs, as well as making page 1-indexed (rather than 0-indexed). That way ?p=2 will behave like you'd expect and will only get added once you go to the next page

That's part of the plan. You can see that in action [here](https://github.com/algolia/instantsearch.js/blob/3cb8b1f96ba2f077789ddbd2f9b28e0c8fe282ea/src/connectors/refinement-list/connectRefinementList.js#L415). However I'm not sure for the page because it introduce a shift of value and it's [pretty easy to implement](http://nebulous-balls.surge.sh/?page=3).

Haroenv:
That’s awesome

NicolasPK:
As suggested by @Haroenv in #933 I am trying to implement this feature and I have it half working.
Here's my feedback on this:
- it breaks when using `operator: 'and'` in a `refinementList` widget:
> Error: facet_name is not defined in the disjunctiveFacets attribute of the helper configuration
- I based my implementation on the `non-trivial mapping` example and got it working for one refinement at a time. I still cannot seem to get more that one refinement list parameter added to the URL. The previous refinement gets overwritten by the new one on the url. And example on how to implement this is more than welcome for me 😄 

But at least this is starting to allow us to get fully customizable url and title, thanks for the great job so far 👍

Haroenv:
I think refinementList isn’t done in this PR yet IIRC, thanks for your feedback. Don’t hesitate to give detailed feedback on for example some of the naming

NicolasPK:
@Haroenv in `urlSync`, the configuration could be as follow:
```
mapping: {
  q: 'search',
  fR: 'features',
  dFR: 'choices',
}
```
but this is still an issue for nested values like the `dFR` one: `[url]?dFR[choice_features.Power supply][0]=Battery&dFR[choice_features.Brand][0]=Samsung`

So I was thinking similar but more nested, like:
```
mapping: {
  q: 'search',
  fR: 'features',
  dFR: {
    choice_features: {
      'Brand': 'brand',
      'Power supply': 'power-supply',
    },
  }
}
```

which would give: `[url]?power-supply=Battery&brand=Samsung`.

Is there an ETA for this PoC? If not, is this possible to customize the url and title in [vue instantseach](https://github.com/algolia/vue-instantsearch)? (the only alternative I'm willing to give a shot).

Haroenv:
@NicolasPK this will at least take two weeks more since @bobylito is on holiday now. With Vue InstantSearch you should be able to sync with Vue Router, see https://community.algolia.com/vue-instantsearch/advanced/vue-router-url-sync.html

NicolasPK:
@Haroenv ok, thanks for the ETA, much appreciated. It sounds like I have to give the Vue implementation a shot then 😟 hopefully I will be long done when @bobylito comes back from holidays.
Such a shame this is not feasible as things are, as I've spend hours to get everything working (apart from that then).

Haroenv:
sorry for that Nicolas

vvo:
Some thoughts on naming:
- I would call the modern storage "history". What may seem modern for us might just be the standard for everyone else. And that's the API being used behind
- I would go for page = 1 right now so that the full UI (url, view) now has the same page, which is 1
- proposal API for discussion/bike shedding:


```
instantsearch({
  router: {
    storage: instantsearch.router.history({
      windowTitle() {},
      createURL() {},
      parseURL() {}
    }),
    mapping: {
      uIStateToRouteState() {}
      routeStateToUiState() {}
    }
    // also feasible: mapping: instantsearch.router.mapping.default
  }
});
```

1. rename urlSync2 to router
2. rename instantsearch.storage to instantsearch.router
3. rename modern to history
4. rename UIStateMapping to mapping
5. rename toSyncable to routeStateToUiState
6. from Syncable to uIStateToRouteState

As for the storage API I have no change to propose BUT maybe "write", it seems to only be called at page load time, should we call it "onInit" rather than read? Not sure about this one, it's maybe just documentation and maybe in some situations it will be called 

As for having a "better" default mapping as proposed by Haroen, I am sure this can come later on and live under instantsearch.router.mapping.clean.

bobylito:
After discussing with @vvo, we came up with:
 - `routing` as the name of the feature
 - `router` as the name of the pluggable element that writes and read
 - `mapping` can be `stateMapping` (so that's it's not too generic) and the methods could be named:
   - `stateToRoute`
   - `routeToState`
 - let's keep `write`
 - and we keep the rest of @vvo proposal

bobylito:
If everyone in @algolia/instantsearch-for-websites is in with the latest proposal, then I'll close this PR and open a new one.

NicolasPK:
Thanks for the update @bobylito If this is approved, what's the ETA for the PR and following merge do you think?

Haroenv:
I'd keep `mapping` 😄

vvo:
I like stateMapping too because it's more explicit on the user will have to manipulate.

@bobylito Do you also plan to have a webpage where you have the full ui state explained as a json? A bit like on react-instantsearch: https://community.algolia.com/react-instantsearch/guide/Search_state.html

bobylito:
> Thanks for the update @bobylito If this is approved, what's the ETA for the PR and following merge do you think?

I'd like to have it ready for the release of next month (2.6). Not hard guarantees here :)

bobylito:
Closing this PR for now. I'll open a new one, when I reach a milestone. In the meantime, I'm pushing the current state here: https://github.com/algolia/instantsearch.js/tree/feat/routing

Issue #1050, opened 2016-06-01
Clear cache while search using instantsearch.js 


ChandanaCH:
Hi,
In our project we are using instantserach.js for search criteria.we have a problem with search i.e, when ever search happens that is not giving correct results for example in our page we are having Like and Unlike actions when ever the authenticated user Like an action the count will be increased and the action is changing as Unlike.we are using algolia with c# to update the index with respective changes.though the Index updated, results are not binding as expected.If we refresh the page then results are coming correctly.As per the documentation of algolia it is showing that each and every time in the search algolia won't hit API,the search will happen from cache.In our case also the thing is same, the results are getting from cache.I tried all the ways to clear cache as following but results are not coming.
var client = algoliasearch('applicationId', 'applicationkey');
                    var index = client.initIndex('indexname');
                    index.clearCache();

Please help us to clear the cache at every time search happens.

vvo:
Hello, sorry for late response, you should be able to clear the instantsearch cache by doing:

```js
// search being your instantsearch instance
search.client.clearCache();
```

Then when there's a new search on the IS interface, the cache will be cleared.

Let me know if that works.

ChandanaCH:
Hello vvo,
Thank you for your reply.It is working fine :)

shauntrennery:
How is clearCache called for the new React components?

vvo:
> How is clearCache called for the new React components?

That is not yet feasible, should be feasible this week we are gonna push a new update allowing you to do so

shauntrennery:
@vvo I have a list of users. The list is populated using the Algolia searchBox connector. I'm wanting to refresh the index when a user is added or removed. Hope this helps explain my need.

vvo:
@shauntrennery Most probably what we will expose is this:

```js
import algoliasearch from 'algoliasearch';
var algoliaClient = algoliasearch(...);
const App = () => <InstantSearch algoliaClient={algoliaClient} />;
// ... later on:
// algoliaClient.clearCache();

Would that work for you?
```

ishanray:
@vvo Any updates on clearing cache for the React Components?

vvo:
This is actually doable with the syntax above https://github.com/algolia/instantsearch.js/issues/1050#issuecomment-266767227.

Just submitted a documentation update for the prop.

tahminabs23:
hi, how can I do it with angular component?

Haroenv:
[...]

PR #2637, opened 2017-12-27
feat(reload): add a reload method on the InstantSearch component

marielaures:
**Summary**
Give the ability to the user to clear the cached answers from Algolia.
(see #1050)

**Result**
The user can now clear the cache by calling `clearCache()` on his/her `search` instance.
See the results in the `RefreshCache` story on dev-novel (network tab).


bobylito:
> looks great, what about calling it refresh instead of clearCache?

As a parallel with the browser, `refresh` or `reload` could be more intuitive because what you want when you clear the cache is to reload the result with fresh ones.

bobylito:
It's all good for me. Thanks @marielaures 💯

Issue #818, opened 2016-02-02
Add `hasChildren` item template data to the hierarchical menu widget


vvo:
Hierarchical menu widget should add `hasChildren` template data to the hierarchical menu widget.

This will solve #767

urbanitedesign:
Is there any timeframe for this?

vvo:
No not yet @urbanitedesign, sorry

ethansmith:
+1

bobylito:
Isn't that related to one of the limitations of Hogan as a template system? Isn't this information already encoded in the length of the array containing the children?

vvo:
Yes you could probably compute it from the length, providing it as a pre computed and named property would make it easier I guess.

bobylito:
I've tried to implement this feature in InstantSearch but with the current implementation it's not very useful. You can see the [code / PoC in this PR](https://github.com/algolia/instantsearch.js/pull/2638).

bobylito:
@urbanitedesign @ethansmith Can you have a look at the demo in the PR and let me know if it was what you were looking for? I guess not but I could be wrong :)

bobylito:
Implementing a more useful solution would require to do more searches to discover which values may contain children. Let's reopen this issue or open a new one if we see any solution that would not rely on subsequent search.

PR #2638, opened 2017-12-27
feat(connectHierarchicalMenu): add hasChildren to items

bobylito:
This PR tries to implement the fix for #818. However it appears that we only retrieve values up to the most last open item. This means that even we can add the `hasChildren` data, this is incorrect as it will be true only if it's open and it's not a very useful information for the user since they would have already open the item.

[See it live](https://deploy-preview-2638--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=HierarchicalMenu.with%20custom%20item%20template).

### DIFF
--- a/dev/app/builtin/stories/hierarchical-menu.stories.js
+++ b/dev/app/builtin/stories/hierarchical-menu.stories.js
@@ -67,6 +67,30 @@ export default () => {
       )
     )
     .add(
+      'with custom item template',
+      wrapWithHits(container => {
+        window.search.addWidget(
+          instantsearch.widgets.hierarchicalMenu({
+            container,
+            attributes: [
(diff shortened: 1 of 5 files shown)


bobylito:
IMHO this proves that it's not really a viable option and therefore we should not add it to the project.

vvo:
> IMHO this proves that it's not really a viable option and therefore we should not add it to the project.

Yep at this point only the API could maybe give us this sort of information (which is very valuable), otherwise we'd have to do up to "number of items of hierarchical menu displayed x requests" to know if any item has any sub elements coming afterwards, which on the frontend is a no-go.

cc @mthuret FYI, limitation of hierarchical facets on the frontend: can't inform the user UX wise if an element has or has not subelements before the user refines it.

PR #2641, opened 2017-12-28
fix(theme): fix height of pagination

bobylito:
The height was super wild by default.


Haroenv:
The Argos changes look fine to me, where can I test the change as simple as possible?

bobylito:
http://earthy-protest.surge.sh/ @Haroenv Maybe you will something I don't.

PR #2642, opened 2017-12-28
feat(core): add an error event to monitor error from Algolia

### LINKED ISSUE #1585: feat(error handling): add an `error` event
It would be good to have an `error` event triggered by instantsearch.js when an API/network error occurs.

Right now there's no (official) way to react to any error form Algolia's side

Having the capability to register a handler, something like `search.on('error', (err) => { ... })`

bobylito:
Fix #1585

Adds an event that is emitted when an error is received in place of results.


bobylito:
We had one up until now, `render` which is used to give an information about the end of the rendering step. They are used in order to provide information that we can't provide using widgets.

Haroenv:
the more you know

PR #2646, opened 2018-01-03
Fix: escape highlight

samouss:
**Summary**

The test sample was not correctly structure so the implementation worked with the test but not with the actual Algolia response. I add a story with the same dataset that React InstantSearch use in Storybook. 

**Before**

screen shot 2018-01-03 at 15 26 21

**After**

screen shot 2018-01-03 at 15 26 56

You can also play with the example on [`dev-novel`](https://deploy-preview-2646--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=Hits.with%20highlighted%20array).


iam4x:
Nice thank you @samouss 👍

PR #2651, opened 2018-01-08
Feat/multi index results

iam4x:
Hi team!

This PR is addressing the multi index feature in a simple way, it allows to display results from different Algolia indices. This would be useful to create a similar experience as autocomplete.js for instance.

**How it works?**

1. The connector is creating derived helpers from the original one
2. When results are retrieved by any helper a render is requested

**What is missing?**

* [ ] Better templating option
* [x] Allow `escapeHits` option like `Hits` widget
* [ ] Allow `transformData` scoped to any index
* [ ] Tests

I'm also not fan of the naming, I do not have any strong opinion on it, any help would be appreciated 👍


src/components/MultiIndexResults.js
```diff
@@ -0,0 +1,71 @@
...
+        />
+      </div>
+    );
+  }
+
+  renderResults({ label, hits }) {
+    const { cssClasses: { item } } = this.props;
+
+    const hitsMarkup = hits.map((hit, position) => {
+      const data = { ...hit, __hitIndex: position };
```
Haroenv: probably the same in regular hits, but if you put it already as `__hitIndex` in the `map`, there's no need to rename

src/connectors/multi-index-results/connectMultiIndexResults.js
```diff
@@ -0,0 +1,112 @@
...
+
+        if (
+          widgetParams.escapeHits &&
+          results.hits &&
+          results.hits.length > 0
+        ) {
+          results.hits = escapeHits(results.hits);
+        }
+
+        derivedIndice.results = results;
```
Haroenv: `derivedIndex` is the singular form

src/connectors/multi-index-results/connectMultiIndexResults.js
```diff
@@ -0,0 +1,112 @@
...
+        this.renderWithAllIndices();
```
Haroenv: I don't follow how you actually saved the results 🤔

iam4x: * Every index is stored into an object on the connector instance

```
{ label, index, helper, results }
```

* On every search, I update the `results` and trigger a new render 👍

mthuret:
@iam4x It seems that part of your connector is doing the same thing than the `connectAutocomplete` of RIS => getting all hits?

Did you try to actually build an autocomplete with it?

src/connectors/multi-index-results/connectMultiIndexResults.js
```diff
@@ -0,0 +1,112 @@
...
+        ) {
+          results.hits = escapeHits(results.hits);
+        }
+
+        derivedIndex.results = results;
+        this.renderWithAllIndices();
+      },
+
+      render({ results, instantSearchInstance }) {
+        this.instantSearchInstance = instantSearchInstance;
```
samouss: Why do you need to keep track of the instance twice? Since we save it inside `init` we probably don't need to save it again from the `render` function. The reference should not change between the `init` & `render` lifecycle. Am I missing something?

src/connectors/multi-index-results/connectMultiIndexResults.js
```diff
@@ -0,0 +1,112 @@
...
+export default function connectMultiIndexResults(renderFn, unmountFn) {
+  checkRendering(renderFn, usage);
+
+  return (widgetParams = {}) => {
+    const { indices } = widgetParams;
+
+    if (
+      !indices ||
+      !Array.isArray(indices) ||
+      (Array.isArray(indices) && indices.length < 1)
```
samouss: Not sure about the last condition, we can still render even if we have only one index (the default one). By taking this path we can also add a default value for `indices` and avoid the first condition (expect if we still want to avoid the `null` case).

src/components/MultiIndexResults.js
```diff
@@ -0,0 +1,68 @@
...
+      </div>
+    );
+  }
+
+  render() {
+    const { derivedIndices, cssClasses: { root } } = this.props;
+
+    const markup = derivedIndices.map(
+      ({ label, results }) =>
+        results && results.hits && results.hits.length > 0
```
Haroenv: make hits top level and default to `[]` in the connector

iam4x:
Good idea @mthuret I will add it as an example into `dev-novel` 👍

src/connectors/multi-index-results/connectMultiIndexResults.js
iam4x: True 👍

src/connectors/multi-index-results/connectMultiIndexResults.js
Haroenv: would be nice to leave a comment for the behaviour of doing side effects on render

Haroenv:
connectAutoComplete should also be able to set the query, not just the hits IMO

mthuret:
In term of API naming for the multi indices feature what we have today on react-is: 

* connectAutocomplete: a connector that have the query (and can refine on it) and all the hits. 
* index: to specify stuff related to one index in particular (could be any widgets here, of search parameters). 

The way we use it:

* Autocomplete use case => usage of connectAutoComplete
* Displaying hits from different indices => usage of Index. 

Hope that helps to make a naming choice.

iam4x:
From the meeting: rename to `connectMultiIndices` 👍

bobylito:
Why does the connector and the widget handle multiple indices?

bobylito:
Next steps as seen with @iam4x 👍 
 - [ ] test with a real combo box  / autocomplete
 - [ ] either move for *unit hits from other index*, this implementation or *autocomplete as in React-IS* based on the previous test
 - [ ] refine the API based on the usability

PR #2652, opened 2018-01-08
chore: modernise some tests

Haroenv:
**Summary**

We're using `toMatchJSX` and imported `expect` and `sinon` in the tests. It's easier to rely on what Jest already provides with snapshots and `jest.spyOn`. 

**Result**

1. added the enzyme snapshot format
2. use enzyme in `RefinementListItem`
3. use enzyme in `Stats`
4. use enzyme and modern Jest in `autoHideContainer`

This obviously isn't all tests, but it isn't a straightforward process because of the things renamed in `import expect from "expect"` vs. `global.expect`. For that reason I think it's best to keep this iterative and just use modern techniques in new tests, update tests when we reach them, since there's still quite a lot of them using `toEqualJSX` (39)

### DIFF
--- a/src/decorators/__tests__/autoHideContainer-test.js
+++ b/src/decorators/__tests__/autoHideContainer-test.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { createRenderer } from 'react-test-renderer/shallow';
+import { shallow } from 'enzyme';
 import autoHideContainer from '../autoHideContainer';
 
 class TestComponent extends React.Component {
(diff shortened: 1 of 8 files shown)


src/decorators/__tests__/autoHideContainer-test.js
```diff
@@ -27,11 +23,7 @@ describe('autoHideContainer', () => {
     const AutoHide = autoHideContainer(TestComponent);
     renderer.render(<AutoHide shouldAutoHideContainer {...props} />);
```
samouss: You could also migrate this file to `enzyme`.

samouss:
Inside the `RangeInput-test.js` we add manually the `enzyme-to-json` serializer. Since we add it globally we can remove the line.

https://github.com/algolia/instantsearch.js/blob/6c3f205d4129b72ae1329848a43707599b2eff44/src/components/RangeInput/__tests__/RangeInput-test.js#L6

Haroenv:
done both those @samouss

PR #2658, opened 2018-01-15
chore(pkg): use jsDelivr default file feature

vvo:
We can reduce the urls we give to users for the JavaScript package by
informing jsDelivr what's the right default file to serve when nothing
provided after the @version.

This is also handy as it does not ties the user and us to a particular
schema to follow over time, if we ever decide to move our build, change
it and rename things, it would be transparent for the user.

For v2 that's "too late" but still good to use it for documentation and
ease of use.

For the theme and style ideally it would be in another package too
("IS.css?" :D)

### DIFF
--- a/docgen/src/examples/getting-started-boilerplate/index.html
+++ b/docgen/src/examples/getting-started-boilerplate/index.html
@@ -41,7 +41,7 @@
     </div>
 
     <!-- js -->
-    <script src="https://cdn.jsdelivr.net/npm/instantsearch.js@2.0.0/dist/instantsearch.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/instantsearch.js@2.0.0"></script>
     <script src="app.js"></script>
   </body>
 </html>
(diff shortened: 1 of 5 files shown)


docgen/src/examples/getting-started-boilerplate/index.html
```diff
@@ -41,7 +41,7 @@
     </div>
 
     <!-- js -->
-    <script src="https://cdn.jsdelivr.net/npm/instantsearch.js@2.0.0/dist/instantsearch.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/instantsearch.js@2.0.0"></script>
```
bobylito: Can you also update the version of IS.js used by the boilerplate?

vvo: You mean the CodeSandbox one? We can only do so when the lib is actually re-released and picked up by jsDelivr

bobylito: No I mean this exact line 😅 Which is the source for the boilerplate but has a fixed version number.

PR #2669, opened 2018-01-23
docs(migration): fixed broken link

sammarcus:
**Summary**

fixed a link that was broken

### DIFF
--- a/docgen/src/guides/migration.md
+++ b/docgen/src/guides/migration.md
@@ -31,7 +31,7 @@ search.addWidget(instantsearch.widgets.searchbox({
 }));
 ```
 
-You can read more about these options [here](/widgets/searchBox.html).
+You can read more about these options [here](instantsearch.js/v2/widgets/searchBox.html).
 
 ## No more `hitsPerPage` in `hits` and `infiniteHits`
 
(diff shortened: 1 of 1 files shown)


sammarcus:
Hm, with [widgets.. as it is now it's not loading.

image

docgen/src/guides/migration.md
```diff
@@ -31,7 +31,7 @@ search.addWidget(instantsearch.widgets.searchbox({
 }));
 ```
 
-You can read more about these options [here](/widgets/searchBox.html).
+You can read more about these options [here](instantsearch.js/v2/widgets/searchBox.html).
```
bobylito: The migration page is already on `instantsearch.js/v2` therefore removing the `/` should be enough, no?

bobylito:
@sammarcus do you have time to fix the comments?

docgen/src/guides/migration.md
sammarcus: Sorry I hadn't seen your review earlier -- thank you @bobylito @Haroenv !

bobylito:
Thanks a lot for your contribution @sammarcus 👍

PR #2678, opened 2018-01-26
Correct invalid URL

lachieh:
This is pretty self explanatory, but the URL for the Common Widgets API is incorrect. This PR fixes it.


Haroenv:
Thanks! this is correct 👌

PR #2686, opened 2018-01-31
fix(doc): add maximum width to images (fix #2685)

### LINKED ISSUE #2685: Images don't have a `max-width` in the documentation container

Bug

I'm writing a guide and wanted to add a screenshot. The image is 800px wide and overflows in most cases.

Images in `.documentation-container` don't have any maximum size.

localhost_3002_v2_guides_calendar-widget html ipad pro 1

(rendered as an iPad Pro)

localhost_3002_v2_guides_calendar-widget html ipad pro

All

We should probably set `max-width: 100%;` to images inside `.documentation-container`.

cc @LukyVj probably

francoischalifour:
**Summary**

See discussion in #2685.

**Result**

Images in the documentation container are restricted to `max-width: 100`.

Let me know if this is not the right way to solve this.

### DIFF
--- a/docgen/src/stylesheets/components/_documentation.sass
+++ b/docgen/src/stylesheets/components/_documentation.sass
@@ -106,6 +106,9 @@ $offset-height: 60px
             margin-right: 6px
             color: $logan
 
+    img
+      max-width: 100%
+
     .content
       & p:first-child
         margin-top: 0


francoischalifour:
cc @LukyVj @ronanlevesque

PR #2687, opened 2018-01-31
Add guide "Create a Calendar Widget"

francoischalifour:
**⚠️ Waiting for #2686 and algolia/create-instantsearch-app#8 to be merged.**

---

This PR is the result of the app and the guide: "*Create a Calendar Widget*".

I'm looking for feedback on both the app and the guide 🙌

*Note: the screenshot in the guide is too big because there's an issue I've fixed in #2686.*


docgen/src/guides/calendar-widget.md
```diff
@@ -0,0 +1,198 @@
...
+
+Algolia handles dates as numeric values. We'll store them as JavaScript timestamps, which are the number of milliseconds since the year 1970 ([see `Date.prototype.getTime()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime)).
+
+Through this guide, we'll use a generated concert dataset that contains the following attributes:
+
+* `name`: the name of the music band
+* `location`: the venue of the concert
+* `date`: the date of the event we will refine the results with
+
+Both `name` and `location` will be added to the [searchable attributes](https://www.algolia.com/doc/guides/ranking/searchable-attributes/) in the Algolia dashboard so that the user can filter events on their name *and* their location.
```
Haroenv: no need for italic `and` IMO

docgen/src/guides/calendar-widget.md
```diff
@@ -0,0 +1,198 @@
...
+
+To learn more about creating your own dataset and uploading it to an Algolia index, you can read the guide on [importing from the dashboard](https://www.algolia.com/doc/tutorials/indexing/importing-data/importing-from-the-dashboard/).
+
+## Bootstrapping an InstantSearch app
+
+If you don't yet have an InstantSearch app to work on, you can use the command-line tool [`create-instantsearch-app`](https://github.com/algolia/create-instantsearch-app) that bootstraps an app for you. It will provide a basic JavaScript application including [instantsearch.js](https://github.com/algolia/instantsearch.js) itself. Make sure to create an index in your Algolia dashboard before any further steps.
+
+If you're using [npm](https://www.npmjs.com/):
+
+```console
```
Haroenv: console is not recognised, and is showing it very faint, try `sh`, `bash` or nothing

docgen/src/guides/calendar-widget.md
```diff
@@ -0,0 +1,198 @@
...
+npm install -g create-instantsearch-app
```
Haroenv: `npx create-instantsearch-app [name-of-your-new-app]`

docgen/src/guides/calendar-widget.md
```diff
@@ -0,0 +1,198 @@
...
+create-instantsearch-app [name-of-your-new-app]
+```
+
+If you're using [Yarn](https://yarnpkg.com):
+
+```console
+yarn create instantsearch-app [name-of-your-new-app]
+```
+
+The command-line tool will ask for your credentials:
```
Haroenv: I'm missing here where the dataset is, maybe list the credentials here, will be clearer, and we show them in the example anyway

docgen/src/guides/calendar-widget.md
```diff
@@ -0,0 +1,198 @@
...
+
+![create-instantsearch-app](images/create-instantsearch-app-events.png)
+
+## Customizing the results
+
+The default `item` template from the `hits` widget provided by `create-instantsearch-app` is great, but we're going to enhance it to display the relevant information of our dataset.
+
+```javascript
+search.addWidget(
```
Haroenv: this is `runnable` for some reason, but it doesn't work with the demo `instant_search` index; @bobylito can we make that configurable in the guide metadata 🤔 ?

docgen/src/guides/calendar-widget.md
```diff
@@ -0,0 +1,198 @@
...
+          </h3>
+          <small>on <strong>${new Date(hit.date)}</strong></small>
+        </li>
+      `,
+    },
+  })
+);
+```
+
+We use the function syntax for the `item` template because we want to convert the numeric timestamp from our dataset to a JavaScript date on the fly.
```
Haroenv:
```diff
  on the fly.

+ Be careful here with user-generated data, you can use existing libraries to prevent XSS,
+ or use a template and `transformData` instead.
```

(we don't currently have a guide on XSS I think, otherwise it should be linked

docgen/src/guides/calendar-widget.md
```diff
@@ -0,0 +1,198 @@
...
+      element: $('#calendar'),
+      callback: function() {
+        const start = new Date(this.start_date).getTime();
+        const end = new Date(this.end_date).getTime();
+
+        refine([start, end]);
+      },
+      // Some good parameters based on our dataset:
+      start_date: new Date(),
+      end_date: new Date(`2020 00:00`),
```
Haroenv: Probably better not to mix backticks and normal quotes

docgen/src/guides/calendar-widget.md
```diff
@@ -0,0 +1,198 @@
...
+  }
+);
+```
+
+`connectRange` is passed as its first argument's callback a [`RangeRenderingOptions`](https://community.algolia.com/instantsearch.js/v2/connectors/connectRange.html#struct-RangeRenderingOptions) object which contains the `refine(Array<number,number>)` function. The two numbers that need to be passed are the start and the end dates as numeric timestamps.
+
+In our connector instantiation, we need to attach the calendar and specify its callback function to refine the search. This part will be specific to the calendar library that you use in your app. In the Baremetrics Calendar, this function is passed to the `callback` argument.
+
+```javascript
+const makeRangeWidget = instantsearch.connectors.connectRange(
```
Haroenv: proposal: call the widget `datePicker`, since that's more consistent with the names in IS.js

vvo:
Nice guide!

## Advice on behaviour
Since I am used to the AirBnB like calendar, it took me some time to understand how to use this one. Here's the difference:





If you just blindly click on the two dates and you endup clicking only on the "End date (second date)" then the behaviour is very confusing because you think you're gonna select start then end dates while it's not the case. You're only changing the end date.

Is there a way to have the default demo behaviour where every time the calendar is opened then the action is "select the start date, then the end date" and never "select only the end date".

# For the demo

Could you add a link to the source code in the demo? So the user once he have seen the demo can jump into the code right away.

docgen/src/guides/calendar-widget.md
francoischalifour: You're right, I should expose them. Should I make the JSON dataset downloadable?

docgen/src/guides/calendar-widget.md
francoischalifour: Yeah, it's also more explicit!

docgen/src/guides/calendar-widget.md
bobylito: I feel so old 👴 what is `npx`?

docgen/src/guides/calendar-widget.md
bobylito: Yes the JSON dataset should be provided.

Haroenv: both the json and the credentials should be given IMO, you can put the JSON in https://github.com/algolia/datasets

docgen/src/guides/calendar-widget.md
Haroenv: It's something npm 5 has, doing install globally and execute the binary, like yarn create

docgen/src/guides/calendar-widget.md
francoischalifour: I'm not convinced the dataset is good enough to be officially exposed though. I might work a bit more on that beforehand!

docgen/src/guides/calendar-widget.md
bobylito: Oh :/ It's because we  do:

```js
    if (exampleContent.indexOf("search.addWidget") === 0) {
      initWidgetExample(codeSample, index);
    }
``` 

in the runnable code JS :/

francoischalifour:
I couldn't find a calendar widget without huge dependencies as good as [`react-dates`](https://github.com/airbnb/react-dates) UX-wise. This is definitely the pain point of this guide. I also had trouble getting familiar with the [Baremetrics Calendar](https://github.com/Baremetrics/calendar). I can try a workaround but it probably won't be very elegant.

As for the source code in the demo, I can add a link to the zip file.

docgen/src/guides/calendar-widget.md
```diff
@@ -0,0 +1,198 @@
...
+category: guides
+withHeadings: true
+navWeight: 20
+editable: true
+githubSource: docgen/src/guides/calendar-widget.md
+---
+
+Search results often need to be refined by dates. Calendars are a good visual way to improve the user experience when it comes to filtering dates. InstantSearch provides custom widgets to create reusable components within your app.
+
+By the end of this guide, you'll understand how to implement a calendar widget to refine your results based on a single date and a range of dates. You can [preview](https://community.algolia.com/instantsearch.js/v2/examples/calendar-widget/) and [download](https://community.algolia.com/instantsearch.js/v2/examples/calendar-widget.zip) the concert app that you'll be able to build.
```
bobylito: Links should be relative.

francoischalifour: Right, should I make all links relative or only these two?
I've seen some other guides using absolute links.

docgen/src/guides/calendar-widget.md
Haroenv: It would be nice if the index would be configurable based on the page metadata, do you think that's possible @bobylito ?

vvo:
> I can try a workaround but it probably won't be very elegant.

It's OK for now since we show how to plug a calendar. This one also has some cool features like last 30 days etc.. Which are great

> As for the source code in the demo, I can add a link to the zip file.

I would add a link to zip yes but also the GitHub direct link, since you might just want to read the code without having to download and unzip

francoischalifour:
@vvo What branch should I link it to? `master`? (that would be [this link](https://github.com/algolia/instantsearch.js/tree/master/docgen/src/examples/calendar-widget))

vvo:
master yes since it's the branch tracking the releases (= doc website is master branch doc)

francoischalifour:
All changes have been made and all dependant PRs have been merged 🙂

@Haroenv I've fixed the "Invalid date" issue on Safari.

docgen/src/guides/calendar-widget.md
francoischalifour: Working on it!

francoischalifour: PR sent #2690

bobylito:
LGTM 👍 thanks for this awesome contribution @francoischalifour :)

PR #2689, opened 2018-02-02
chore: update Algolia dependencies

bobylito:
Adds support for insights in the helper.


Haroenv:
Does renovate not work here?

vvo:
Shouldn't the package.json be updated along with the lock file too? package.json says 2.3.0 while it should say the latest of the helper.

vvo:
And yes might be good to know why the helper has not been updated in package.json for so long, maybe @rarkins know why. The branches are protected so there should be PRs being opened (on weekends) but that's maybe no more the case now.

2.3.0 was released long time ago so should have been updated already

rarkins:
@vvo are you referring to `algoliasearch-helper`? Its current version in `dependencies` is `^2.23.0`, and the latest on npmjs is `v2.24.0`, hence it is "in range" and not eligible for an upgrade.

If you want it bumped to the exact latest version every time, there's really no benefit to any end user in having the caret `^` in place in `package.json`, so you can remove it and see it updating regularly.

Another alternative for you is to enable lock file maintenance, this bumps lock files even if the `package.json` remains the same. i.e. would achieve something similar to this PR, for every single package.

vvo:
@rarkins I believe if we wait this weekend, the helper will be bumped to ^2.24.0 by renovate right?

rarkins:
> @rarkins I believe if we wait this weekend, the helper will be bumped to ^2.24.0 by renovate right?

No, that's not the behaviour actually. The logic is: "if the latest npmjs release falls within the existing range, do not update it". This is because in most cases, people use ranges because they *want* to keep a wide range for downstream users.

If you need it bumped to >= v2.24.0 for *functional* reasons (e.g. using a feature of v2.24.0 that's not supported by v2.23.0 then that's something a developer needs to do manually, because only the developer involved could know that. If instead you just think it's best practice to keep your downstream users as updated as possible (e.g. to the latest, once a week), is there any reason why not to remove the caret range and pin it?

Another option is to use tilde `~` ranges instead, then it would bump from `~2.3.0` to `~2.4.0` but not to `~2.3.1` for example.

And finally.. this "logic" I mentioned is not written in stone and I can consider a feature request if you wish to keep bumping the range ever week. I just prefer we both understand "why" things are like they are.. my philosophy is that whenever you set out solving the wrong problem, you're always going to end up with a wrong solution :)

In short: you use ranges if you don't want to overly restrict your downstream users about which (compatible) versions they should use. If you *do* wish to be restrictive, then a narrower range and/or pinning sounds better.

Let me know if the above is clear or any parts are unclear or disagreeable?

samouss:
When we rely on such tools (like Renovate) we should go for exact dependencies. When we don't want to upgrade just close the PR. Correct me if I'm wrong but when you close a PR Renovate will ignore the dependency until a new version is released.

rarkins:
@samouss yes that's right. e.g. if existing is `v2.3.0` and Renovate suggests to upgrade to `v2.4.0` but you close it, then Renovate will ignore `v2.4.0` forever. If `v2.4.1` or `v2.5.0` comes out then it will then suggest it again and you have the choice to merge or ignore.

package.json
```diff
@@ -123,15 +123,15 @@
...
-    "events": "^1.1.0",
-    "hogan.js": "^3.0.2",
-    "lodash": "^4.17.4",
-    "preact": "^8.1.0",
-    "preact-compat": "^3.16.0",
-    "preact-rheostat": "^2.1.1",
-    "prop-types": "^15.5.10",
-    "to-factory": "^1.0.0"
+    "classnames": "2.2.5",
+    "events": "1.1.0",
```
Haroenv: This means that as long as we don't do a release, people will be stuck on the older version of these, while with the ^, they'd still get upgraded, even if we don't update yet. I wonder if that's a problem 🤔

vvo:
Thanks for all the details, now I am in favour of having dependencies pinned in all libraries.

PR #2690, opened 2018-02-02
Support search config override per documentation page

francoischalifour:
This PR adds support for new frontmatter attributes to run code samples on specific Algolia indices depending on the page you're on.

*EDIT: as frontmatter attributes are not documented anywhere, I added a contributing guide explaining what each of them is used for.*

## The problem

Code snippets in the documentation can be run when we detect some patterns like `/widgets.(\S+)\(/g`. They are run on a hard coded Algolia index called `instant_search` on the `latency` app ([see code](https://github.com/algolia/instantsearch.js/blob/5cd002b0c2442cc952d59cecab335a55e61cbd35/docgen/assets/js/bindRunExamples.js#L8)).

Sometimes, you'd like your code snippet to be run on another index, or even another app. This is the case for my guide [Create a Calendar Widget](https://github.com/algolia/instantsearch.js/pull/2687) which relies on the index `concert_events_instantsearchjs`.

## This solution

I'm extracting the search configuration in the frontmatter of the guide/documentation and inject it in the `window` object.

```markdown
---
title: Create a calendar widget
appId: latency
apiKey: 6be0576ff61c053d5f9a3225e2a90f76
index: concert_events_instantsearchjs
---
```

Each article can now use its own search configuration. If no configuration provided, we default to the current one.

—

This is not a great solution but it is the only one that could work as far as I know.

### DIFF
--- a/docgen/CONTRIBUTING.md
+++ b/docgen/CONTRIBUTING.md
@@ -26,7 +26,7 @@ The name of the category of your page.
 
 > `string` | no default
 
-The layout of your page among [`docgen/layouts/`](docgen/layouts/) (i.e. [`main.pug`](docgen/layouts/main.pug)).
+The layout of your page among [`docgen/layouts/`](layouts/) (i.e. [`main.pug`](layouts/main.pug)).
 
 #### `category`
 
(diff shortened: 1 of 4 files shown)


francoischalifour:
As discussed with @bobylito, I wrote a contributing guide for the `docgen` folder to explain what each frontmatter attribute is used for.

Please let me know if I forgot some or got some wrong (I had to guess some since it was not documented so far).

PR #2695, opened 2018-02-05
Specify `indexName` in guide

francoischalifour:
**Summary**

* The `index` attribute in the documentation is actually called `indexName`.
* The `indexName` attribute in the frontmatter of a documentation page is now supported.

### DIFF
--- a/docgen/src/guides/calendar-widget.md
+++ b/docgen/src/guides/calendar-widget.md
@@ -7,6 +7,7 @@ withHeadings: true
 navWeight: 20
 editable: true
 githubSource: docgen/src/guides/calendar-widget.md
+indexName: concert_events_instantsearchjs
 ---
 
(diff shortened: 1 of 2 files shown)


docgen/src/guides/calendar-widget.md
```diff
@@ -7,6 +7,7 @@ withHeadings: true
 navWeight: 20
 editable: true
 githubSource: docgen/src/guides/calendar-widget.md
+indexName: concert_events_instantsearchjs
```
Haroenv: doesn't this need api key and app id too?

francoischalifour: I'm using the default `appId` and `apiKey` values here, only the index name is different. Should I specify all of them?

Haroenv: ah, I thought it was on your own app; all good to go then!

PR #2696, opened 2018-02-07
refactor: move Hogan logic to utils

samouss:
**Summary**

This PR extract the behaviour of the template processing from `<Template>` to `utils`. We can take leverage of the same logic without using the component. It means that non React widget can use the same function for processing the template (ex: SearchBox, Geo).

Still some questions about the usage of `compileOptions`. We do not provide a way to override it properly from `InstantSearch` except:

```js
const search = instantsearch(...)

search.templateConfig.compileOptions = { ... }
```

I didn't find any documentation / issue on the subject , so maybe get rid of this option in the next major release. It will also simplify the `prepareTemplates` function.

### DIFF
--- a/src/components/Template.js
+++ b/src/components/Template.js
@@ -1,13 +1,8 @@
-import PropTypes from 'prop-types';
 import React, { Component } from 'preact-compat';
-import hogan from 'hogan.js';
-
-import curry from 'lodash/curry';
+import PropTypes from 'prop-types';
 import cloneDeep from 'lodash/cloneDeep';
-import mapValues from 'lodash/mapValues';
 import isEqual from 'lodash/isEqual';
-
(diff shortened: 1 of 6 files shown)


samouss:
@bobylito Yep it's only that. But I didn't move the logic related to `useCustomCompileOptions`, it's only available with `<Template>`. So the widget that are using the raw function don't take the compile option into account (it's already the case for the `SearchBox`). Should we? If so I can also move the logic and set some default values for cases like the `SearchBox` widget.

bobylito:
> Should we? If so I can also move the logic and set some default values for cases like the SearchBox widget.

I don't think we should. We can add it later but so far, we've had no one asking for it for the searchbox.

Issue #2697, opened 2018-02-07
RFC: connectConfigure


Haroenv:
* Start Date: 2018-02-05

# Summary

A connector for third-party integrations which doesn't have any rendering to set options for the InstantSearch instance. These options are sent on every `search` or `searchForFacetValues`.

# Motivation

Having `searchParameters` in `instantsearch()` makes sense for a static configuration, but as soon as you want to make it dynamic, it makes more sense as a widget, than as a parameter for InstantSearch. The main reason being that a widget is easier to change than a parameter on InstantSearch, and will be more consistent

# Detailed design

```js
const widgetFactory = connectConfigure(
 (state, isFirstRendering) => {
   // render nothing
 },
 () => {
   // just a necessary noop
 }
);

const widget = widgetFactory({
 analytics: true,
 ruleContexts: ["fancy context", "add dogs"]
});

search.addWidget(widget);

// later, if you want to change something:
const newWidget = widgetFactory({
 analytics: true,
 ruleContexts: ["boring context", "add dogs"]
});
search.addWidget(newWidget);
search.removeWidget(widget);
```

# How We Teach This

Add new entry in docs, implement in Vue InstantSearch and Angular InstantSearch as a widget.

# Drawbacks

This connector doesn't need a widget which renders anything, although in some cases it _could_ make sense as a widget which is reconstructed if the parameters change, as a fallback just on top of a real custom widget. We should not take that usecase too much in account, but rather encourage custom widgets for that use case

# Alternatives

Use the helper directly in the libraries, but this is complicated as that would make components inconsistent

# Unresolved questions

- What happens on unmount
- What happens on conflicting parameters

Haroenv:
As discussed, this is a widget, not a connector

PR #2698, opened 2018-02-07
feat(configure): add the Configure widget

Haroenv:
**Summary**

In component-based frameworks, it makes sense to have search parameters separately from the main instantsearch widget, but in its own widget. This allows for way easier dynamic and static search parameters.

**Result**

This PR adds `widgets.configure` for use in Vue InstantSearch etc.

closes https://github.com/algolia/instantsearch.js/issues/2697 (the RFC for this connector)

see the demo made for this: https://pr-connectconfigure.netlify.com (open the console)

On unmount it will _take ownership_ of its searchParameters and unset them. This won't practically be a problem, since one configure widget should be enough to do what you need, and it's usually used with parameters which don't conflict with widgets


src/connectors/configure/connectConfigure.js
```diff
@@ -0,0 +1,73 @@
+import { checkRendering } from '../../lib/utils.js';
+
+const usage = `Usage:
+var configure = connectConfigure(function render(params, isFirstRendering) {
+  // params is whatever you put as argument to the widget
+  // usually you'll just have a function which does nothing
+  // as rendering for connectConfigure
+});
```
iam4x: I think it might be confusing for users to talk about "render" since this connector should not render anything?
For instance in react-instantsearch the connector is not even exposed to end-users and the Configure component just return `null`

iam4x:
Nice work @Haroenv 👍 

About conflicting parameters, I think that we should print a warning in the console and let the Configure widget to be the one that override? What do you think?

Haroenv:
Having a warning can be an idea. I documented the render function because it’s necessary for a component I thought, but maybe we can find a way around it 🤔

iam4x:
I would remove entirely `renderFn` and `unmountFn` from the API of this connector.

bobylito:
> I would remove entirely renderFn and unmountFn from the API of this connector.

Stumbling on this idea. Actually if you don't need those functions, it means that there is no need for specialization (as a mean to provide custom behavior). Hence there is no need for a connector, as it is a specialization enabler. This means that this should be implemented as a widget factory, not. as a connector. With the new dynamic add / remove API, a widget would even be usable directly in InstantSearch.js.

src/connectors/configure/__tests__/connectConfigure-test.js
```diff
@@ -0,0 +1,83 @@
+import jsHelper from 'algoliasearch-helper';
+import connectConfigure from '../connectConfigure';
+
+const createState = state =>
+  jsHelper({ addAlgoliaAgent: () => {} }, '', state).state;
```
bobylito: You probably want to use [`jsHelper.make`](https://github.com/algolia/algoliasearch-helper-js/blob/develop/src/SearchParameters/index.js#L559) instead of instantiating the helper and getting the state.

bobylito: So much so that you use literrally everywhere you use create state (it's a static function)

Haroenv:
updated it to a widget, and update the demo too

Haroenv:
just not sure about the docs yet, it doesn't really seem to show up 🤔

bobylito:
> just not sure about the docs yet, it doesn't really seem to show up 🤔

I can see one https://deploy-preview-2698--algolia-instantsearch.netlify.com/v2/widgets/configure.html \o/

Haroenv:
I meant the content actually, seems to be more empty than I expected. We’ll take a look Monday

src/widgets/configure/__tests__/configure-test.js
```diff
@@ -0,0 +1,85 @@
+import { SearchParameters } from 'algoliasearch-helper';
+import configure from '../configure';
+
+describe('configure', () => {
+  it('throws when you pass it a non-plain object', () => {
+    [
+      () => configure(new Date()),
+      () => configure(() => {}),
+      () => configure(/ok/),
+    ].map(widget => expect(widget).toThrowError(/Usage/));
```
bobylito: Why a map?

src/widgets/configure/configure.js
```diff
@@ -0,0 +1,50 @@
+import isPlainObject from 'lodash/isPlainObject';
+const usage = `Usage:
+search.addWidget(
+  instantsearch.widgets.configure({
+    // any searchParameter
+  })
+);
+Full documentation available at https://community.algolia.com/instantsearch.js/widgets/configure.html
```
bobylito: The URL is incorrect, it should be https://community.algolia.com/instantsearch.js/v2/widgets/configure.html

src/widgets/configure/configure.js
```diff
@@ -0,0 +1,50 @@
...
+`;
+
+/**
+ * The **Configure** widget provides the logic to build a custom widget that
```
bobylito: It does not provide logic since it's a widget and not a connector anymore. What about:

> The **configure** widget is a headless widget that let you configure the settings of your search using the parameters described by [the general Algolia documentation](https://www.algolia.com/doc/api-reference/search-api-parameters/)

src/widgets/configure/__tests__/configure-test.js
Haroenv: no reason, can be `forEach` too

samouss: We we perform side effects it's better to use a `forEach`. With it the intent is clear: I will perform a side effect. We should not use `map` *only* for iterate on an array. The main purpose of a `map` is to transform an array of `A` into an array of `B`.

PR #2699, opened 2018-02-08
fix(perf): only compute snappoints when step is provided

### LINKED ISSUE #2662: Browsers crash if too many widgets are created.
This is a bug
 
I have a case where i want to have 50 widgets, and the browser crashes. 
Edge (latest) - total page crash , Chrome (latest) - total page crash, Firefox (latest) - popup shows to wait or stop, if you wait it renders the page and up to 24 widgets.
If the widgets are 24 or less it's all good.
Browser doesn't crash if the widgets are more than 25
Edge (latest), Chrome (latest), Firefox (latest)
 
I want to make a page that can filter objects that have many properties
2.4.1

bobylito:
This keeps the snapping but does not generate the snap points if no value is provided. We can do that because rheostat snaps the point to the rounded value by default (if `snap=true`). Until rheostat provides a lazy way to fix the snappoints, there will be a risk of having an "out of memory" error because the range of values can be too big. There is an open issue opened for adding this new API: airbnb/rheostat#14 but it's uncertain if someone will work on it.

Fix #2662


src/components/Slider/Slider.js
```diff
@@ -55,6 +55,7 @@ export class RawSlider extends Component {
 
   // creates an array of values where the slider should snap to
   computeSnapPoints({ min, max, step }) {
+    if(!step) return undefined;
```
iam4x: I think this is not valid to our linter, might do:

```js
if (step) {
  return [...range(min, max, step), max];
}
```

iam4x:
Impressive debugging @bobylito you nailed it 💯

bobylito:
> Anyway, according to travis you still need to update the snapshots

done

> A good workaround for now, but I wonder if an option would be to implement airbnb/rheostat#14 🤔

Yes. That could be the end game. Let's move forward with this one. I dismiss your review @Haroenv ;)

PR #2738, opened 2018-02-15
chore(doc): remove horizontal scroll in sidebar

### LINKED ISSUE #2667: Doc website: Visible scrolling bars
The menu on the left has a horizontal scrollbar visible. It prevents to see all the content of the menu
screenshot 2018-01-22 18 47 34

Not to see them

FF 57

bobylito:
Fix #2667 

Removes the horizontal scroll, can't do anything about fading them or not.


bobylito:
> I don't think this really solves the issue

It does solve the problem for now. It looks better. We don't have to find the ultimate solution to make the situation better.

PR #2739, opened 2018-02-15
fix(yarnrc): use empty string for save-prefix

samouss:
**Summary**

We introduced a `.yarnrc` file recently, but by setting the value to `false` introduce this kind issue:

```
"enzyme-adapter-react-15": "false1.0.5"
``` 

By setting the value to an empty string we get the expected result which is:

```
"enzyme-adapter-react-15": "1.0.5"
```


bobylito:
👍 Can you also fix the package.json? ;)

samouss:
@bobylito Done!

PR #2741, opened 2018-02-16
 refactor(pagination): pass in extra info from connector: pages, isFirstPage, isLastPage

### LINKED ISSUE #2680: RFC: more complete connectPagination

feature

Currently there's still a significant amount of code to write if you want to write your own `pagination` component.

Deal with the whole clamping, +1 / -1 and padding issue in the connector rather than the view

2.4.1

Haroenv:
**Summary**

Moves the Paginator class and its used methods / properties from the Pagination widget to the connector, so it becomes easier to use in other InstantSearch versions.

**Result**

fixes #2680


src/components/Pagination/Pagination.js
```diff
@@ -128,21 +127,33 @@ export class RawPagination extends Component {
...
+        {this.previousPageLink({ isFirstPage, currentPage, createURL })}
+        {this.pages({
+          currentPage,
+          pages,
+          createURL,
+        })}
+        {this.nextPageLink({ isLastPage, currentPage, createURL })}
+        {this.props.showFirstLast
+          ? this.lastPageLink({
+              total: this.props.nbPages,
```
bobylito: This could be extracted like the other props?

src/connectors/pagination/connectPagination.js
```diff
@@ -123,14 +136,42 @@ export default function connectPagination(renderFn, unmountFn) {
         return maxPages !== undefined ? Math.min(maxPages, nbPages) : nbPages;
       },
 
+      getPages({ nbPages, currentRefinement }) {
```
bobylito: Where is this function used?

src/components/Pagination/Pagination.js
```diff
@@ -52,57 +51,57 @@ export class RawPagination extends Component {
...
-      pageNumber: pager.total - 1,
+      pageNumber: total - 1,
       createURL,
     });
   }
 
-  pages(pager, createURL) {
-    const pages = [];
+  pages({ currentPage, pages, createURL }) {
+    const _pages = [];
```
bobylito: I'd try to have a better name for `_pages`. Why not `pageComponents` or `pageElements`?

src/connectors/pagination/connectPagination.js
```diff
@@ -111,6 +121,9 @@ export default function connectPagination(renderFn, unmountFn) {
             currentRefinement: helper.getPage() || 0,
             nbHits: 0,
             nbPages: 0,
+            pages: pager.pages(),
+            isFirstPage: pager.isFirstPage(),
+            isLastPage: pager.isLastPage(),
```
bobylito: These new parameters need to be documented.

src/connectors/pagination/connectPagination.js
```diff
@@ -111,6 +121,9 @@ export default function connectPagination(renderFn, unmountFn) {
             currentRefinement: helper.getPage() || 0,
             nbHits: 0,
             nbPages: 0,
+            pages: pager.pages(),
```
bobylito: During `init`, we know the values as there has been no search yet, so we can hardcode them:
 - pages: `[]`
 - isFirstPage: `true`
 - isLastPage: `true`

src/components/Pagination/Pagination.js
```diff
@@ -128,21 +127,33 @@ export class RawPagination extends Component {
...
     return (
       <ul className={this.props.cssClasses.root}>
-        {this.props.showFirstLast ? this.firstPageLink(pager, createURL) : null}
-        {this.previousPageLink(pager, createURL)}
-        {this.pages(pager, createURL)}
-        {this.nextPageLink(pager, createURL)}
-        {this.props.showFirstLast ? this.lastPageLink(pager, createURL) : null}
+        {this.props.showFirstLast
+          ? this.firstPageLink({ isFirstPage, createURL })
+          : null}
```
samouss: We could use short-circuit evaluation instead of having the ternary with the `null`.

src/components/Pagination/Pagination.js
```diff
@@ -52,57 +51,57 @@ export class RawPagination extends Component {
...
   }
 
-  pages(pager, createURL) {
-    const pages = [];
+  pages({ currentPage, pages, createURL }) {
+    const _pages = [];
 
-    forEach(pager.pages(), pageNumber => {
-      const isActive = pageNumber === pager.currentPage;
+    forEach(pages, pageNumber => {
```
samouss: Not sure about that, but as soon as we refactor some code we can begin to get rid of the `lodash` functions like `forEach`, `map`, `reduce`, ... 

cc @bobylito

bobylito: We can as long as we stay opportunistic.

src/components/Pagination/Pagination.js
bobylito: well yes 👍

src/components/Pagination/Pagination.js
bobylito: maybe the parameter in lastPageLink can be renamed, so that we can forward directly the props?

src/components/Pagination/Pagination.js
```diff
@@ -128,21 +127,33 @@ export class RawPagination extends Component {
...
-    });
-
-    const createURL = this.props.createURL;
+    const {
+      createURL,
+      pages,
+      isFirstPage,
+      isLastPage,
+      currentPage,
+    } = this.props;
```
bobylito: We could probably refactor the code to prevent destructuring? Since the functions that use those values accepts them as an object. It's kinda obfuscating though. A thought on that @samouss ?

src/connectors/pagination/connectPagination.js
Haroenv: sure, just had them here for consistency

src/components/Pagination/Pagination.js
Haroenv: nowhere, was an artefact from an older version of this PR

src/connectors/pagination/connectPagination.js
Haroenv: nowhere, was an artefact from an older version of this PR

src/components/Pagination/Pagination.js
Haroenv: fixed with `map`ping instead of `forEach`ing (was originally going for as little changes as possible to code I didn't touch)

bobylito: You didn't change the name of the variable here. I think we could benefit from that.

src/components/Pagination/Pagination.js
bobylito: If pages is an array, we should go for using the native method as suggested by @samouss

PR #2751, opened 2018-02-26
Upgrade preact-compat

bronzehedwick:
**Summary**

Updrading preact-compat fixes IE 11 + Google Maps clashing. My previous commit (see below) did not address the issue in the preact-compat library, which is also effecting. Upgrading the library resolves this.

**Result**

- https://github.com/algolia/instantsearch.js/pull/2749
- https://github.com/developit/preact-compat/pull/423

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "hogan.js": "3.0.2",
     "lodash": "4.17.4",
     "preact": "8.1.0",
-    "preact-compat": "3.17.0",
+    "preact-compat": "3.18.0",
     "preact-rheostat": "2.1.1",
     "prop-types": "15.5.10",
     "to-factory": "1.0.0"


bronzehedwick:
Folks responding to https://github.com/algolia/instantsearch.js/pull/2749 were correct: I was confused about where the error was actually coming from. While the Symbol checks in instantsearch *might* have also caused a problem, the real issue was the checks in preact-compat, which the mentioned PR resolves.

My apologies for my wrongness.

bobylito:
Hi @bronzehedwick thanks for opening this PR and for taking the time to contribute. However we can't upgrade to this version because it breaks the slider (https://deploy-preview-2751--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=RangeSlider.default) I haven't had time to investigate this specific issue even though encountered it when trying to upgrade the preact-compat lib too.

bronzehedwick:
Ah ok, that's a bummer. I think we're going to go ahead with a fork for now, since we're not using the slider and we'd like to release this coming Tuesday. We can switch back to reference the upstream code once the issues have been worked through.

Thanks so much for your help!

bobylito:
Sorry about that @bronzehedwick Going with your own fork might be the best for now indeed

PR #2757, opened 2018-03-01
feat(Template): add props for the tagName

samouss:
**Summary**

With the current implementation the `<Template>` is always wrapped with a `<div>` element. But in some cases it can lead to incorrect DOM nesting. e.g. When we use the `<Template>` inside a `<label>` or a `<button>` the `<div>` element is forbidden inside those elements. 

By providing a `rootTagName` props we can specify which element the root will be. In the previous example we can use a `<span>` since it's a valid child element for `<label>` & `<button>`.

### DIFF
--- a/src/components/Template.js
+++ b/src/components/Template.js
@@ -14,6 +14,7 @@ export class PureTemplate extends Component {
   }
 
   render() {
+    const RootTagName = this.props.rootTagName;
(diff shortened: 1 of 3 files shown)


Haroenv:
Where do you want to _use_ this?

samouss:
In the *next* GeoSearch widget there are some templates that are used inside a `<label>` or a `<button>` so it will be helpful for this widget at least.

PR #2764, opened 2018-03-05
docs: update Breadcrumb widget example

samouss:
**Before**

screen shot 2018-03-05 at 16 32 10

**After**

screen shot 2018-03-05 at 16 31 58


src/widgets/breadcrumb/breadcrumb.js
```diff
@@ -60,7 +60,7 @@ breadcrumb({
   [ cssClasses.{disabledLabel, home, label, root, separator}={} ],
   [ templates.{home, separator}]
   [ transformData.{item} ],
-  
+
```
Haroenv: what's this empty line for?

samouss: Will remove it.

bobylito:
That's indeed way better than an empty box 😅

PR #2765, opened 2018-03-05
docs: allow static example

### LINKED ISSUE #2694: Better control over runnable examples
Right now, the JS script in the page relies on a test to define if the code of an example should be runnable or not: contains `search.addWidget`? or `/function renderFn\(\S+(, isFirstRendering)?\) {/g.test(exampleContent)`?

This gives poor control over what's runnable or not, which seems magical and also lead to inconsistent experience like the new calendar guide. In this guide, we have one runnable example while the others are not. Given the context, none should be runnable in the guide.

Options could be:
 - have a way to specify runnable js
 - have a front matter option for setting if the js examples on the page should be runnable

samouss:
**Summary**

Right now we don't have to possibility to expose static code example in the documentation. We always have the `run` button in header preview. For some cases it might be interesting to support static code example (e.g. GeoSearch - we don't want to load Leaflet in the page just for the connector example). 

This PR adds a new JSDoc annotation called `@staticExample` to allow this kind of example. This annotation behaves the same way as `@example`, except that the example won't have the `run` button. 

Note that for now we don't support both example types at the same time. It works but the structure of the HTML page will be a bit weird since we duplicate the `h2`. We can probably merge them into only one, but lets interate on that when the needs occur.

fixes #2694

### DIFF
--- a/docgen/assets/js/main.js
+++ b/docgen/assets/js/main.js
@@ -20,11 +20,12 @@ const header = new alg.communityHeader(docSearch);
 
 const container = document.querySelector('.documentation-container');
 const codeSamples = document.querySelectorAll('.code-sample');
+const codeSamplesRunnable = document.querySelectorAll('.code-sample-runnable');
 
 dropdowns();
 move();
 activateClipboard(codeSamples);
-bindRunExamples(codeSamples);
+bindRunExamples(codeSamplesRunnable);
 
 const sidebarContainer = document.querySelector('.sidebar');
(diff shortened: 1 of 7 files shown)


Haroenv:
misc is not allowed by our commit linting, FYI 😄

samouss:
All right, will rebase that.... Thanks!

bobylito:
@samouss don't bother updating the commits 👍

PR #2768, opened 2018-03-05
docs: add DevNovel link

samouss:
**Summary**

This PR adds a "Live example" section in the documentation of the widgets. You can enable the DevNovel link by adding the annotation: `@devNovel NameOfTheWidgetInDevNovel`.

**Before**

screen shot 2018-03-05 at 19 29 41

**After**

screen shot 2018-03-05 at 19 30 51

### DIFF
--- a/src/widgets/numeric-selector/numeric-selector.js
+++ b/src/widgets/numeric-selector/numeric-selector.js
@@ -68,6 +68,7 @@ const usage = `Usage: numericSelector({
  *
  * The values inside this attribute must be JavaScript numbers and not strings.
  * @type {WidgetFactory}
+ * @devNovel NumericSelector
  * @category filter
  * @param {NumericSelectorWidgetOptions} $0 The NumericSelector widget options.
  * @return {Widget} A new instance of NumericSelector widget.
--- a/dev/app/builtin/stories/menu.stories.js
+++ b/dev/app/builtin/stories/menu.stories.js
@@ -9,7 +9,7 @@ const stories = storiesOf('Menu');
 export default () => {
   stories
     .add(
-      'Default',
+      'default',
       wrapWithHits(container => {
         window.search.addWidget(
           instantsearch.widgets.menu({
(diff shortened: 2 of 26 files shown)


src/widgets/numeric-selector/numeric-selector.js
```diff
@@ -68,6 +68,7 @@ const usage = `Usage: numericSelector({
  *
  * The values inside this attribute must be JavaScript numbers and not strings.
  * @type {WidgetFactory}
+ * @devNovel NumericSelector
```
Haroenv: I wonder if we can't edit the dev novel to have the same name as the widget, so there doesn't need to be duplicate config 🤔

bobylito: That's a good thought. Why did you go for this approach @samouss ?

samouss: I thought it was a bit more flexible and less magical to have the name in the file. Note that the annotation is required in order to generate the "Showcase" link because in some cases the widget don't have a story (e.g. `Configure`).

Haroenv: What about leaving the default to "there's a story", making sure the name in the novel is the same as the widget name, and adding something like `@suppressDevNovel` to those that specifically don't have a novel?

samouss: I prefer to have the opt-in in a "positive" way rather than the other. It feels more natural to add them rather than suppress them.

Haroenv: okay, just suggesting it, since I thought it might be easy to forget about it making a new widget, but it's fine like this.

Haroenv:
What about adding the vanilla / jQuery version for the connectors?

src/widgets/numeric-selector/numeric-selector.js
bobylito: I like the opt-in aspect. Sounds good to me.

bobylito:
> What about adding the vanilla / jQuery version for the connectors?

It's already a good improvement. Let's iterate :)

samouss:
@Haroenv Yep it's a good idea but we should make a pass on the example before. Not all the connector have examples, we should first complete them and then we can safely add the links.

PR #2772, opened 2018-03-05
fix(GeoSearch): override button style

samouss:
**Before**

screen shot 2018-03-05 at 20 10 06

**After**

screen shot 2018-03-05 at 20 10 26

### DIFF
--- a/src/css/theme/_geo-search.scss
+++ b/src/css/theme/_geo-search.scss
@@ -4,12 +4,15 @@
 
 .ais-geo-search--clear {
   @extend .ais-clear-all--link;
+  box-shadow: 0 1px 1px 0 rgba(85, 95, 110, 0.2);
+  border: solid 1px #D4D8E3;
+  border-radius: $border-radius;
+  padding: 8px 15px;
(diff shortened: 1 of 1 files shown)


src/css/theme/_geo-search.scss
```diff
@@ -43,6 +46,7 @@
   border: solid 1px #D4D8E3;
   border-radius: $border-radius;
   padding: 8px 15px;
+  outline: none;
```
Haroenv: I don't think we should remove the outline without replacement though, see: [outlinenone.com](http://www.outlinenone.com)

bobylito: Not false. What alternative can you propose @Haroenv ?

samouss: Completely agree! It's just a bad habits 😄

Haroenv: 1. don't change outline
2. we should have a proper outline style which isn't ugly in Chrome; cc @ronanlevesque

PR #2773, opened 2018-03-05
fix(GeoSearch): add API Key for Google Maps

samouss:
This PR adds an API Key for Google Maps to avoid the service to blocked in DevNovel. The key has some domain restrictions so an external user should not be able to use it.


Haroenv:
Did you also add the preview domains to the list in the google dev console?

samouss:
@Haroenv Yep!

PR #2787, opened 2018-03-08
chore: multi package updates

bobylito:
It also contains a fix found by eslint 😅


Haroenv:
The build size seems increased; weird!

bobylito:
> The build size seems increased; weird!

Not cool to upgrade packages. The good news is that the test pass :)

vvo:
Nice chore work! Super weird that the actual size of the lib went up 11kb gz:

image

https://bundlesize-store.now.sh/build?info=%7B%22files%22%3A%5B%7B%22maxSize%22%3A102400%2C%22path%22%3A%22.%2Fdist%2Finstantsearch.min.js%22%2C%22size%22%3A109294%2C%22compression%22%3A%22gzip%22%2C%22master%22%3A97781%7D%5D%2C%22repo%22%3A%22algolia%2Finstantsearch.js%22%2C%22branch%22%3A%22misc%2Fmulti-package-updates%22%2C%22commit_message%22%3A%22Merge%20594e300dcfca7a4046f9c300249ee91cb686c966%20into%2079ebd66bda1b7ccc4f2d8a47ef67f2bb8c5b39c5%22%2C%22sha%22%3A%22594e300dcfca7a4046f9c300249ee91cb686c966%22%7D

Maybe it's just because bundlesize itself was updated. Because I see no other lib that could be responsible for this in the updates.

bobylito:
> Maybe it's just because bundlesize itself was updated. Because I see no other lib that could be responsible for this in the updates.

I'm reading the same size from our own compile script. Digging into the dependencies I have updated: so far keeping the old lodash would have been worse.

bobylito:
Ok let's move on. Maybe we could try to think about some wins that we could implement to reduce the size later.

PR #2808, opened 2018-03-12
chore(functional-tests): mock JS client data

bobylito:
This change implements a fake algoliaserch-client with pre-recorded data. I wished we could have something close to Jest snapshots so that we could easily record and update the data. That's not the case but maybe it could an iteration over that.

The motivation is to be able to make sure that the behaviour of our test never change because of:
 - network
 - index being updated

It should fix the dataset in order to make argos give us consistent tests.

PS: I kept commented the code that let you record the data


functional-tests/app/app.js
```diff
@@ -306,3 +324,19 @@ search.addWidget(
...
...":496,\"34.99\":487},\"categories\":{\"Appliances\":4306,\"Computers & Tablets\":3563,\"Cell Phones\":3291,\"Cell Phone Accessories\":2836,\"Audio\":1570,\"Small Kitchen Appliances\":1510,\"Cameras & Camcorders\":1369,\"Car Electronics & GPS\":1208,\"TV & Home Theater\":1201,\"Cell Phone Cases & Clips\":1195},\"price_range\":{\"1 - 50\":9927,\"50 - 100\":3471,\"100 - 200\":2640,\"200 - 500\":2628,\"500 - 2000\":2467,\"> 2000\":336},\"free_shipping\":{\"true\":18013,\"false\":3456},\"hierarchicalCategories.lvl0\":{\"Appliances\":4306,\"Computers & Tablets\":3563,\"Cell Phones\":3291,\"Audio\":1570,\"Cameras & Camcorders\":1369,\"Car Electronics & GPS\":1208,\"TV & Home Theater\":1201,\"Health, Fitness & Beauty\":923,\"Office & School Supplies\":617,\"Video Games\":505}},\"facets_stats\":{\"price\":{\"min\":1,\"max\":4999.99,\"avg\":242.394,\"sum\":5187220}},\"exhaustiveFacetsCount\":true,\"exhaustiveNbHits\":true,\"query\":\"\",\"params\":\"query=&hitsPerPage=6&maxValuesPerFacet=10&page=0&facets=%5B%22price_range%22%2C%22price%22%2C%22brand%22%2C%22free_shipping%22%2C%22price%22%2C%22categories%22%2C%22hierarchicalCategories.lvl0%22%5D&tagFilters=\",\"index\":\"instant_search\"}]}",
```
Haroenv: What about storing these in files?

Haroenv: oh never mind, I misread!

functional-tests/app/app.js
```diff
@@ -306,3 +307,58 @@ search.addWidget(
...
+ *  - paste the content in the coldQueries
+ *
+ * @param {function} algoliasearch original algoliasearch factory
+ * @param {string} app appID
+ * @param {key} key API Key
+ * @return {object} a JS client
+ */
+function recorder(algoliasearch, app, key) { // eslint-disable-line
+  const client = algoliasearch(app, key);
+  const searchƒ = client.search;
```
Haroenv: `ƒ` is probably from copy pasting?

bobylito: No it's on purpose to differentiate from the other `search`

vvo: I am used when monkey patching to write:

```js
var originalSearch = client.search
```

WDYT?

bobylito: Ok. If 2 devs ask questions better change ;)

Haroenv:
I think this ultimately comes down to the API not giving stable results for the facet counts, since that's what changing every time

samouss:
Agree with @Haroenv the issue always come from the facets count. We have the same issue on React InstantSearch.

vvo:
Nice then we could either:
- have a template where facet count is always the same
- use a smaller index for functional tests/recipes so that we always have it fixed

PR #2812, opened 2018-03-14
docs(tokens): remove [*] which were confusing

Haroenv:
**Summary**

[*] used to mean "not required", but wasn't very obvious (we had two internal questions about it recently) and wasn't applied everywhere.

**Result**

No more `[*]` in the docs


Haroenv:
Weird that there’s a conflict. I’ll check it out tomorrow, I forgot to rebase probably

Haroenv:
@bobylito rebased, now there's no conflict anymore

PR #2814, opened 2018-03-15
fix(main): correctly import EventEmitter

### LINKED ISSUE #2730: InstantSearch imports EventEmitter incorrectly
Bug

We put the instantsearch.js module into our (rollup) build and got an error that internal node module "events" does not export `EventEmitter`.

When I looked there is an ES6 import:
https://github.com/algolia/instantsearch.js/blob/541699d900ca0a20e17c5624baa1a7ab085fbf9a/src/lib/InstantSearch.js#L9
Which is a named import.

But in the relevant module at `node_modules/instantsearch.js/node_modules/events/events.js`, which uses CommonJS exports, the exports is defined as such:
`module.exports = EventEmitter;`
Which is a default export.

I assume in your build system this got forced into resolving, but Rollup is strict with ES imports. To have the project build I quickly changed the line in InstantSearch.js to a default import:
`import EventEmitter from 'events';`
and it builds fine.

I also found a correct, default import being used in one of your tests: https://github.com/algolia/instantsearch.js/blob/3b659783cd659f82539d97aec2e3ca1e691797f5/src/widgets/search-box/__tests__/search-box-test.js#L2

Make the import in InstantSearch.js a default export as shown above for correctness and consistency with the test code.

Haroenv:
**Summary**

fixes #2730 (see that issue for context)

**Result**

Now you can have InstantSearch as a dependency in a rollup **UMD** build

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -6,7 +6,7 @@ import forEach from 'lodash/forEach';
 import mergeWith from 'lodash/mergeWith';
 import union from 'lodash/union';
 import isPlainObject from 'lodash/isPlainObject';
-import { EventEmitter } from 'events';
+import EventEmitter from 'events';
 import urlSyncWidget from './url-sync.js';
 import version from './version.js';
 import createHelpers from './createHelpers.js';
--- a/src/lib/__tests__/__snapshots__/InstantSearch-test.js.snap
+++ b/src/lib/__tests__/__snapshots__/InstantSearch-test.js.snap
@@ -91,7 +91,6 @@ Array [
         "addAlgoliaAgent": [Function],
         "algolia": "client",
       },
-      "domain": null,
       "helper": AlgoliaSearchHelper {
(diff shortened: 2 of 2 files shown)


Haroenv:
The commonjs and es module version still have "`import from events`" in their code, which is good. The source is what jest uses to build, and it's acting as if it's node, so taking the version of `events` in node, and thus having or not having the `domain: null` key (this is why the second commit failed). The UMD build always had `module.exports = EventEmitter` and `EventEmitter.EventEmitter = EventEmitter` in the code, nothing changes here, since we go from "importing EventEmitter.EventEmitter" to "importing EventEmitter". Webpack has allowed "destructuring" of commonJS imports, but rollup doesn't allow that, since the module.exports is just EventEmitter.

Haroenv:
(latest test fails because Revert is not a valid pattern for commitlint, but the actual tests still pass, as initially seen)

Haroenv:
yay

PR #2815, opened 2018-03-16
fix(connectBreadcrumb): allow unmounting

Haroenv:
**Summary**

I got errors while implementing this widget in Vue InstantSearch, so I assume this widget was merged after the adding of `unmountFn`/`dispose`. 

I checked all other connectors, and this was the only one with this missing

**Result**

1. added `unmountFn`
2. nothing at all

There's no implementation of the dispose, since you could imagine someone not showing the breadcrumb anymore, but it's acting on the same info as the hierarchical menu, and you don't want to unrefine that one when you only want to stop showing the breadcrumb

### DIFF
--- a/dev/app/init-unmount-widgets.js
+++ b/dev/app/init-unmount-widgets.js
@@ -112,6 +112,31 @@ export default () => {
     )
   );
 
+  storiesOf('Breadcrumb').add(
+    'default',
+    wrapWithUnmount(
+      container =>
+        instantsearch.widgets.breadcrumb({
+          container,
+          attributes: [
+            'hierarchicalCategories.lvl0',
+            'hierarchicalCategories.lvl1',
+            'hierarchicalCategories.lvl2',
+          ],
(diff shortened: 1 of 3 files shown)


bobylito:
@Haroenv we should be careful when unmounting this widget because it goes in combo with the hierarchical menu. Removing the breadcrumb here might have some interferences with the  hierarchical menu.

Haroenv:
That's a good point, I can change it to not remove the refinement upon dismount. What is the way this is handled in React InstantSearch @samouss?

Haroenv:
Behaviour here is fine (never actually unmount), but I'd like to have the behaviour of the first commit. It isn't doing what I expect though

samouss:
@Haroenv we don't clean up when we unmount this widget in React InstantSearch. This is the same behaviour.

Haroenv:
Makes sense to have this as an exception, since it's not usually the "acting" widget, rather a secondary input

PR #2829, opened 2018-03-19
Feat/routing

bobylito:
This PR provides a new implementation for a URL Synchronisation.

More infos  to come.


Haroenv:
DocumentationJS causes the docs [build](https://app.netlify.com/sites/algolia-instantsearch/deploys/5aafe99c7b6ee81563efc779) to fail: 

```
5:50:32 PM: (node:1844) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: Cannot read property 'name' of null
```

commits 75b0bdf6ec69a8b362438b53994db8b26fe6eb53, b396d123847850c89fc2b225749b3823f80c4340, e38af89cbc3c2bcab223ffcb6387eb3519398835, ba1729e041b585eb433891cd39cfca40a6453b40, 285a26e5f2c1d0848cf073812bad47d89e13ebd9, 0df05a8cf88f0e96a4a70a0f96138d5536ff3dc4, d8b6372cfd68feea48a5859ce3d0a7571fb9b5e6, afdff777eed0336cebc8e491a32f905f4400a246, c7bd05d4405ecbb7ddad1efb00862835e1ef84a1, 6d88dd03afe2759dba1ccff5875bee1bfea19a37, 7685a2826f0250c46c1d95b39c6e492a1e47941d, d7e1c94a8c6a037115d41a89a091978f0285a4ca, 82c60cb3bb73254cbd091266ef616bb217f05155, 4c0a92b698905c0fc8a7fd86586a0692b23d5a27, 297f37f4a1be5f3e2b8e583d353a671bba7c201d, 2423b3b1a25d641267f1cb80dd88ea09234f245e, 489acc0c46631d7e65f8677e691bbc8c88fcbd46, 4f136a802592d978e777f36fea5f43dfb88533b9 are wrong according to commit lint. I've said it before, but since we squash PRs, I don't see much value in having correct titles for PR commits. The script should still run on develop though

bobylito:
Those are old commits :(

src/lib/main.js
```diff
@@ -88,6 +91,8 @@ import * as widgets from '../widgets/index.js';
  */
 const instantsearch = toFactory(InstantSearch);
 
+instantsearch.routing = routing;
```
vvo: I thought (but can't find the discussion, might have been informal) that we would go for instantsearch.routers (since it's an object with the default routers we ship). Wdyt?

vvo: I thought (but can't find the discussion, might have been informal) that we would go for instantsearch.routers (since it's an object with the default routers we ship). Wdyt?

This would also differentiate it from instantsearch({routing})

package.json
```diff
@@ -105,6 +105,7 @@
     "raw-loader": "0.5.1",
     "react-test-renderer": "15.6.1",
     "sass-lint": "1.12.1",
+    "uglifyjs-webpack-plugin": "^1.1.4",
```
vvo: Can you give context on this new package? Thx

vvo:
Thanks for the PR! Can you give some guidelines as for reviewing every widget, what should we be looking for/paying attention to in those code changes?

This is a nice feature, with some good code addition, so releasing a beta and testing it out on multiple already existing websites is very important. So we get a sense of any issue that might happen.

For example, what happens if:
- I am using urlSync and upgrade to the latest, will something break?
- I developed a custom widget and now I will use routing, will something break?

Also worth considering:
- Do we start deprecating urlSync in favour of routing?
- If so, let's provide an easy upgrade guide for users of urlSync (link to doc from the code)

> Find a place in the reference documentation
I would put it into InstantSearch.js if technically that's the easiest, otherwise you'll have to create a specific page for an option of instantsearch(), maybe not easy.

Then about creating your own router or using the feature, I would detail that in two guides, or one guide with two distinct sections:
- routing
  - usage
  - advanced (more details on the API, semantics, behaviour)

src/lib/router/history.js
```diff
@@ -0,0 +1,130 @@
+import qs from 'qs';
```
vvo: Should it be routers/history?

vvo:
Also I am interested in the documentation detail about the default router output (url it outputs) compared to the current urlSync.

src/lib/main.js
bobylito: Yes indeed. 👍

package.json
bobylito: So we have this command called serve that let us serve the package compiled (with auto build and reloading) so that we can use it in a demo implementation (I have a bunch of them). The problem is that this command was broken because the version of this package bundled with webpack is outdated and broken.

src/lib/router/history.js
bobylito: Good question! Probably.

bobylito:
> Can you give some guidelines as for reviewing every widget, what should we be looking for/paying attention to in those code changes?

Done in the description.

> I am using urlSync and upgrade to the latest, will something break?

no. Worth checking for that.

> I developed a custom widget and now I will use routing, will something break?

no. The new methods are optional. You will not benefit from the new routing unless you implement the methods.

> Do we start deprecating urlSync in favour of routing?

Yes. Also the previous implementation is so badly documented that with a guide, it will be pretty obvious what to choose.

> If so, let's provide an easy upgrade guide for users of urlSync (link to doc from the code)

Yep. Next PR as for everything doc.

src/connectors/hierarchical-menu/__tests__/__snapshots__/connectHierarchicalMenu-test.js.snap
```diff
@@ -0,0 +1,93 @@
...
+      "path",
+    ],
+  },
+}
+`;
+
+exports[`connectHierarchicalMenu routing getWidgetState should not overide other entries in the same namespace 1`] = `
+Object {
+  "hierarchicalMenu": Object {
+    "category": Array [
```
vvo: Shouldn't it conains  otherCategory: ['path'], too?

src/connectors/hierarchical-menu/__tests__/__snapshots__/connectHierarchicalMenu-test.js.snap
```diff
@@ -0,0 +1,93 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`connectHierarchicalMenu routing getWidgetSearchParameters should add the refinements according to the UI state provided 1`] = `
+SearchParameters {
+  "advancedSyntax": undefined,
```
vvo: Would it be feasible to not have everything that is undefined and only what's relevant for the test?

bobylito: No unless the object was built in a different way (would require to change the JS Helper for that)

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -125,6 +126,8 @@ The first one will be picked, you should probably set only one default value`
       );
     }
 
+    const defaultValue = find(userItems, item => item.default === true);
```
vvo: what is item.default? (Always thought default was a reserved word in JS)

src/connectors/hierarchical-menu/__tests__/__snapshots__/connectHierarchicalMenu-test.js.snap
bobylito: Yes

src/connectors/menu/__tests__/connectMenu-test.js
```diff
@@ -404,4 +404,117 @@ describe('connectMenu', () => {
...
+        onHistoryChange: () => {},
+      });
+
+      const { refine } = rendering2.mock.calls[0][0];
+
+      return [widget, helper, refine];
+    };
+
+    describe('getWidgetState', () => {
+      test('should give back the object unmodified if there are no refinements', () => {
```
vvo: After seeing many tests that are written in the same way (unmodified, add an entry, do not override namespace entries), I wonder if it would have been feasible to write **one** getWidgetState utility with configuration parameters and unit test that one instead, delegating calls to every widgets afterwards and only checking how we call it from widgets.

bobylito: Some tests are identical indeed. But not all, do you have a proposal for making both special tests and generic ones?

src/lib/url-sync.js
```diff
@@ -111,7 +101,6 @@ class URLSync {
   constructor(urlUtils, options) {
     this.urlUtils = urlUtils;
     this.originalConfig = null;
-    this.timer = timerMaker(Date.now());
```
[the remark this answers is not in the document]
bobylito: Yes exactly and it hadn't been removed.

src/connectors/menu/__tests__/connectMenu-test.js
vvo: After more carefully looking at `getWidgetState` implementations indeed it seems pretty different between widgets and not something easily made generic so all good

src/connectors/hits-per-page/connectHitsPerPage.js
bobylito: It's a boolean flag that expresses the default value of the hitsPerPage. It is a reserved keyword indeed https://docs.microsoft.com/en-us/scripting/javascript/reference/javascript-reserved-words but we're safe https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Reserved_word_usage

src/connectors/pagination/__tests__/__snapshots__/connectPagination-test.js.snap
```diff
@@ -0,0 +1,121 @@
...
+  "synonyms": undefined,
+  "tagFilters": undefined,
+  "tagRefinements": Array [],
+  "typoTolerance": undefined,
+}
+`;
+
+exports[`connectPagination routing getWidgetState should add an entry equal to the refinement 1`] = `
+Object {
+  "page": 5,
```
vvo: should be 4 given the test I believe

vvo: Nevermind, that's the internal +1 being done for ui state to start at 1. 👍

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -962,4 +976,131 @@ describe('connectRefinementList', () => {
...
+
+      test('should add an entry equal to the refinement', () => {
+        const [widget, helper, refine] = getInitializedWidget();
+        refine('value');
+        const uiStateBefore = {};
+        const uiStateAfter = widget.getWidgetState(uiStateBefore, {
+          state: helper.state,
+          helper,
+        });
+        expect(uiStateAfter).toMatchSnapshot();
```
samouss: You use Snapshot for those cases but I find them tedious in that case. Because by reading the tests file we have no clue of what it's expected.

src/connectors/numeric-refinement-list/connectNumericRefinementList.js
```diff
@@ -173,6 +172,85 @@ export default function connectNumericRefinementList(renderFn, unmountFn) {
...
+        const lowerBound =
+          (currentRefinements['>='] && currentRefinements['>='][0]) || '';
+        const upperBound =
+          (currentRefinements['<='] && currentRefinements['<='][0]) || '';
+
+        if (lowerBound !== '' || upperBound !== '') {
+          if (
+            fullState.numericRefinementList &&
+            fullState.numericRefinementList[attributeName] ===
+              `${lowerBound}:${upperBound}`
```
iam4x: If one of the `lowerBound` or `upperBound` is `=== ''` then it will end like this:

* `xx:` or `:xx`

Is it expected?

bobylito: Yes

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
bobylito: OK will had a more precise test.

iam4x:
Bug with `&` character in the facet value:

1. Go to http://disastrous-approval.surge.sh/
2. Select Computers & Tablets in the menu
3. Reload the page, you see in the URL everything after the "&" is removed

src/connectors/pagination/__tests__/connectPagination-test.js
```diff
@@ -245,4 +245,105 @@ describe('connectPagination', () => {
...
+        const searchParametersAfter = widget.getWidgetSearchParameters(
+          searchParametersBefore,
+          { uiState }
+        );
+        expect(searchParametersAfter).toBe(searchParametersBefore);
+      });
+
+      test('should enforce the default value if no value is in the UI State', () => {
+        const [widget, helper, refine] = getInitializedWidget();
+        refine(4);
```
vvo: if we were not even calling refine this test would still work I guess.

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -962,4 +976,131 @@ describe('connectRefinementList', () => {
...
+    describe('getWidgetSearchParameters', () => {
+      test('should return the same SP if no value is in the UI state', () => {
+        const [widget, helper] = getInitializedWidget();
+        const uiState = {};
+        const searchParametersBefore = SearchParameters.make(helper.state);
+        const searchParametersAfter = widget.getWidgetSearchParameters(
+          searchParametersBefore,
+          { uiState }
+        );
+        expect(searchParametersAfter).toBe(searchParametersBefore);
```
samouss: The previous comment the Snapshot is more accurate in that case because the structure of `SearchParameters` is way more complicated than the `uiState` one.

iam4x:
* This URL gets the `&page=2` removed: http://disastrous-approval.surge.sh/?menu[categories]=External%20Hard%20Drives&page=2
* **BUT** this URL does not gets the `&page=1` removed: http://disastrous-approval.surge.sh/?menu[categories]=External%20Hard%20Drives&page=1

iam4x:
1. Go to http://disastrous-approval.surge.sh
2. Select page 2
3. Apply any other filter, the pagination gets reseted

src/connectors/pagination/connectPagination.js
```diff
@@ -164,6 +164,22 @@ export default function connectPagination(renderFn, unmountFn) {
       dispose() {
         unmountFn();
       },
+
+      getWidgetState(fullState, { state }) {
```
vvo: Maybe we can rename this to `getWidgetState(fullUiState, { searchParam })`

src/connectors/pagination/__tests__/connectPagination-test.js
```diff
@@ -245,4 +245,105 @@ describe('connectPagination', () => {
...
+        const uiState = {};
+        const searchParametersBefore = SearchParameters.make(helper.state);
+        const searchParametersAfter = widget.getWidgetSearchParameters(
+          searchParametersBefore,
+          { uiState }
+        );
+        expect(searchParametersAfter).toBe(searchParametersBefore);
+      });
+
+      test('should enforce the default value if no value is in the UI State', () => {
```
vvo: Note for myself and anyone: this test checks this user flow:
- user loads https://website/search
- user clicks on page 4
- user click on back button
- we load page 0

src/lib/RoutingManager.js
```diff
@@ -0,0 +1,195 @@
...
+      'change',
+      this.renderURLFromState
+    );
+    this.router.dispose();
+  }
+
+  getAllSearchParameters({ state, uiState }) {
+    const { widgets } = this.instantSearchInstance;
+    const searchParameters = widgets.reduce((u, w) => {
+      if (!w.getWidgetSearchParameters) return u;
```
samouss: We can get rid of the condition by doing a filter before the reduce.

src/connectors/numeric-selector/connectNumericSelector.js
```diff
@@ -160,6 +160,46 @@ export default function connectNumericSelector(renderFn, unmountFn) {
...
+
+        if (currentRefinement || currentRefinement === 0)
+          return {
+            ...fullState,
+            numericSelector: {
+              ...fullState.numericSelector,
+              [attributeName]: currentRefinement,
+            },
+          };
+        return fullState;
```
iam4x: It's not easy to read without brackets to the condition, I would have written that way:

```js
  if (currentRefinement || currentRefinement === 0) {
    return {
      ...fullState,
      numericSelector: {
        ...fullState.numericSelector,
        [attributeName]: currentRefinement,
      },
    };
  }
    
  return fullState;
```

src/connectors/star-rating/connectStarRating.js
```diff
@@ -196,8 +196,48 @@ export default function connectStarRating(renderFn, unmountFn) {
...
 
+      getWidgetState(fullState, { state }) {
+        const refinedStar = this._getRefinedStar(state);
+        if (
+          refinedStar === undefined ||
+          (fullState &&
+            fullState.starRating &&
+            fullState.starRating[attributeName] === refinedStar)
+        )
+          return fullState;
```
francoischalifour: Could we wrap this condition with brackets?

src/connectors/search-box/__tests__/connectSearchBox-test.js
```diff
@@ -1,5 +1,5 @@
 import jsHelper from 'algoliasearch-helper';
-const SearchResults = jsHelper.SearchResults;
+const { SearchResults, SearchParameters } = jsHelper;
```
samouss: Could be inline with the line above.

vvo:
As discussed as a team, we need to encode the url values by default ([, ], &...) let's be safe since we are now fully customisable if you want cool urls like categories=sf,horror. You can.

src/lib/routers/history.js
```diff
@@ -0,0 +1,130 @@
+import qs from 'qs';
+
+function defaultCreateURL(qsModule, uiState) {
+  const { protocol, hostname, port = '', pathname, hash } = window.location;
+  const queryString = qsModule.stringify(uiState, { encode: false });
```
Haroenv: I think something here is not encoding `&`, while it definitely should

src/connectors/toggle/__tests__/connectToggle-test.js
```diff
@@ -1,7 +1,7 @@
 import sinon from 'sinon';
 
 import jsHelper from 'algoliasearch-helper';
-const SearchResults = jsHelper.SearchResults;
+const { SearchResults, SearchParameters } = jsHelper;
```
samouss: Could be inline with the line above.

src/lib/RoutingManager.js
bobylito: We're gonna do twice the interations... But it's worth considering.

src/connectors/numeric-selector/connectNumericSelector.js
bobylito: Yes to brackets!

src/connectors/star-rating/connectStarRating.js
bobylito: Yes we can!

src/connectors/sort-by-selector/connectSortBySelector.js
```diff
@@ -157,6 +157,30 @@ export default function connectSortBySelector(renderFn, unmountFn) {
...
         return state.setIndex(this.initialIndex);
       },
+
+      getWidgetState(fullState, { state }) {
+        const currentIndex = state.getQueryParameter('index');
+        const isInitialIndex = currentIndex === this.initialIndex;
+
+        if (
+          isInitialIndex ||
+          (fullState && fullState.sortBy === currentIndex)
```
vvo: As said in the slack thread, there's currently an issue on how we compute the default value for the index that we consider when we want to remove the index from the url.

src/connectors/toggle/connectToggle.js
```diff
@@ -275,6 +275,52 @@ export default function connectToggle(renderFn, unmountFn) {
...
+        );
+
+        if (isRefined) {
+          if (hasAnOffValue)
+            return searchParam
+              .removeDisjunctiveFacetRefinement(attributeName, off)
+              .addDisjunctiveFacetRefinement(attributeName, on);
+
+          return searchParam.addDisjunctiveFacetRefinement(attributeName, on);
+        } else {
```
samouss: The else branch is useless, we early return in the condition.

vvo:
Bugs found in http://disastrous-approval.surge.sh/:

---
Pagination bug:
- load
- click on page 2
- refresh
- page 2 not selected and disappears from url

---
Index selector bug:
- load
- change index to Lowest price
- sortBy added to url
- click on page 2
- refresh
- everything removed from url, selected index is lowest price still shown

Also if from there you change back the index to the "default one" (passed to instantsearch), now it's added to the url back
---
Index selector bug 2:
- load http://disastrous-approval.surge.sh/
- select "Lowest price", it gets added to url
- reload
- select back "most relevant", it gets added to url
- select back "lowest price", sortBy is removed to url
- reload, index is now desync with what the user thought it was (edited)
---
This one ^ is very close to the previous one where everything gets removed from url (both page and index) so there might be just one bug in pagination and one in sortBy

src/connectors/toggle/connectToggle.js
bobylito: Old habits :(

src/connectors/sort-by-selector/connectSortBySelector.js
bobylito: The initial index is now computed from the parameter `indexName` given to `instantsearch`

src/connectors/pagination/__tests__/connectPagination-test.js
bobylito: It may be worth adding some comments on this kind of test.

bobylito:
I need to add support for the geosearch widget, right? @samouss

src/connectors/numeric-selector/connectNumericSelector.js
bobylito: But maybe could we add that as a rule in our eslint config and do that in another PR? :)

Haroenv:
travis fails because prettier isn't used :) 

documentationJS still fails for same reason as before: 

```
12:49:03 PM: (node:1847) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: Cannot read property 'name' of null
```

bobylito:
> travis fails because prettier isn't used :)

don't worry, by the end I will have fixed most of the issues

src/connectors/numeric-selector/connectNumericSelector.js
vvo: I am in favor for enforcing brackets in every situation as a linting and formatting rule. Just because it removes those kind of discussion and does not create a dual rule where sometimes it gets written with brackets, sometimes without.

Same thing for arrow function only returning an object without brackets.

WDYT?

Haroenv: -1 for adding a rule against 

```js
const s = () => ({});
```

we'd be forced to write them like 

```js
const s = () => {
  return {
  };
};
```

Which isn't really that valuable IMO. Requiring braces around an if and disallowing inline `if` is definitely +1 from me!

vvo: > const s = () => {
>   return {
>   };
> };

But prettier will do this for you so you won't even see it or have to think about it.

Haroenv: My issue was that it's longer to type

bobylito:
Doc is working again :)

bobylito:
I need to properly rewrite part of the toggle tests because it mocks the helper...

src/connectors/range/connectRange.js
```diff
@@ -292,6 +292,59 @@ export default function connectRange(renderFn, unmountFn) {
 
         return nextState;
       },
+
+      getWidgetState(fullState, { state }) {
```
samouss: Same comments as for `priceRanges` since it's exactly the same logic. Maybe we could refactor a bit this part to reuse the same function across both widgets.

src/lib/InstantSearch.js
```diff
@@ -74,6 +83,12 @@ Usage: instantsearch({
     }
 
     this.urlSync = urlSync === true ? {} : urlSync;
+    if (routing === true) this.routing = ROUTING_DEFAULT_OPTIONS;
```
samouss: It looks like we can use both system at the same time, could be nice to throw an error when both options are provided.

src/lib/routers/history.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+   * This methods sets a callback on the `onpopstate` event of the history API
+   * of the current page. This way, the URL sync can keep track of the changes.
+   * @param {function(object)} cb the callback that will receive the latest uiState.
+   * It is called when the URL is updated.
+   * @returns {undefined}
+   */
+  onUpdate(cb) {
```
samouss: I will go for a different name on this method. The `onUpdate` name intent a function that will be called multiple times. At the end the function is like an `addListener`, we call it to subscribe to udpate.

src/lib/RoutingManager.js
```diff
@@ -0,0 +1,195 @@
...
+
+        if (isEqual(fullHelperState, searchParameters)) return;
+
+        helper
+          .overrideStateWithoutTriggeringChangeEvent(searchParameters)
+          .search();
+      });
+
+      helper.on('change', searchState => {
+        // this.renderURLFromState(s)
```
samouss: We forgot to move this function to `renderURLFromState`, we can't remove the listener on the helper in the `dispose` method with the current implementation.

src/lib/RoutingManager.js
```diff
@@ -0,0 +1,195 @@
...
+        // See: https://github.com/algolia/instantsearch.js/issues/2523#issuecomment-339356157
+        this.router.write(firstRenderState);
+      }
+    }
+  }
+
+  dispose() {
+    this.instantSearchInstance.helper.removeListener(
+      'change',
+      this.renderURLFromState
```
samouss: It's a no-op the function doesn't exist.

src/lib/RoutingManager.js
```diff
@@ -0,0 +1,195 @@
+import algoliasearchHelper from 'algoliasearch-helper';
```
samouss: We should add tests for this module.

src/lib/routers/history.js
```diff
@@ -0,0 +1,130 @@
+import qs from 'qs';
```
samouss: We should add tests for this module.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -213,6 +213,43 @@ export default function connectHierarchicalMenu(renderFn, unmountFn) {
 
         return nextState;
       },
+
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -197,6 +200,38 @@ The first one will be picked, you should probably set only one default value`
       dispose() {
         unmountFn();
       },
+
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/menu/connectMenu.js
```diff
@@ -250,6 +250,49 @@ export default function connectMenu(renderFn, unmountFn) {
 
         return nextState;
       },
+
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/numeric-refinement-list/connectNumericRefinementList.js
```diff
@@ -173,6 +172,85 @@ export default function connectNumericRefinementList(renderFn, unmountFn) {
         unmountFn();
         return state.clearRefinements(attributeName);
       },
+
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/numeric-selector/connectNumericSelector.js
```diff
@@ -160,6 +160,46 @@ export default function connectNumericSelector(renderFn, unmountFn) {
         return state.removeNumericRefinement(attributeName);
       },
 
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/price-ranges/connectPriceRanges.js
```diff
@@ -228,6 +228,60 @@ export default function connectPriceRanges(renderFn, unmountFn) {
 
         return nextState;
       },
+
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/range/connectRange.js
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -409,6 +410,42 @@ export default function connectRefinementList(renderFn, unmountFn) {
             .removeDisjunctiveFacet(attributeName);
         }
       },
+
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/search-box/connectSearchBox.js
```diff
@@ -151,6 +151,23 @@ export default function connectSearchBox(renderFn, unmountFn) {
         unmountFn();
         return state.setQuery('');
       },
+
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/sort-by-selector/connectSortBySelector.js
```diff
@@ -157,6 +157,30 @@ export default function connectSortBySelector(renderFn, unmountFn) {
         unmountFn();
         return state.setIndex(this.initialIndex);
       },
+
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/star-rating/connectStarRating.js
```diff
@@ -196,8 +196,48 @@ export default function connectStarRating(renderFn, unmountFn) {
         return nextState;
       },
 
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/connectors/toggle/connectToggle.js
```diff
@@ -275,6 +278,52 @@ export default function connectToggle(renderFn, unmountFn) {
 
         return nextState;
       },
+
+      getWidgetState(fullState, { state }) {
```
samouss: According to the change in `getWidgetSearchParameters`:

- `fullState` should be renamed `uiState`
- `state` should be renamed `searchParam`

src/lib/routers/history.js
bobylito: I'm not sure to follow here. onUpdate is similar to onClick, the event being update and it happens on the storage. If we go for addListener, it implies that you can have multiple kinds of events, no?

samouss: When I first look at the function I thought the function will be call on each update and then add a listener multiple times. What we want is to listen for the event and call the callback on each update. Right?

src/connectors/range/connectRange.js
bobylito: I'm not in favor to merge part of the logic of two different connectors. This would mean that we would have a third party holding the logic of a connector or even multiple connectors. I have the impression that  would make the code harder to follow.

However this might hint for a generalization of ranges here, right? In this case, we could have only one generalized connector. What about digging into that later?

src/lib/routers/history.js
bobylito: It's a method used by InstantSearch.js and that is  also used for the update callback that you can set in widgets (like the search box). It is used to connect another part of the system to the update of the state of the router. We will 1-n functions plugged into that. It will be strictly limited to updates.

What do you think? Should I still modify it?

src/connectors/range/connectRange.js
samouss: Completely agree, we can dig into that later.

src/lib/routers/history.js
samouss: IMO yes we should but at the end you know the system better than me. Take the path that you find the most understandable. I'm fine with it.

bobylito: Let's talk about that Monday 👍

src/lib/RoutingManager.js
bobylito: Test have been added for this one.

src/connectors/hierarchical-menu/__tests__/__snapshots__/connectHierarchicalMenu-test.js.snap
bobylito: This is fixed.

bobylito: Fixed :)

src/connectors/numeric-refinement-list/connectNumericRefinementList.js
```diff
@@ -213,42 +214,41 @@ export default function connectNumericRefinementList(renderFn, unmountFn) {
...
+
+        const valueAsEqual = value.indexOf(':') === -1 && value;
+
+        if (valueAsEqual) {
+          return clearedParams.addNumericRefinement(
+            attributeName,
+            '=',
+            valueAsEqual
+          );
+        }
```
samouss: We didn't test that case, we only test when it's a range.

src/connectors/numeric-refinement-list/__tests__/connectNumericRefinementList-test.js
```diff
@@ -399,4 +403,127 @@ describe('connectNumericRefinementList', () => {
...
+        });
+
+        expect(uiStateAfter).toMatchSnapshot();
+      });
+
+      test('should not override other values in the same namespace', () => {
+        const [widget, helper] = getInitializedWidget();
+        const uiStateBefore = {
+          numericRefinementList: {
+            'numerics-2': ['27:36'],
```
samouss: I don't think that it's the right structure, it should be `27:36`.

src/connectors/numeric-refinement-list/connectNumericRefinementList.js
bobylito: Test were added

bobylito:
Thanks all for the effort here. And especially @samouss for not giving me a push on the last part :)

PR #2830, opened 2018-03-19
chore: publish beta without tags

bobylito:
The motivation is that conventional changelog is based on tags to generate the latest changelog. The problem is that in our new setup we have two concurrent versions that could add to the changelog. If both did (2.x and develop) then when you have to merge and you will end with interlaced releases because of the data sorting. This is messy and the betas don't matter that much for most of the users.

The proposal prevents the tags from being done so that we ensure that we have a nice automatic changelog for actual releases. Beta releases will still be comited, and we will be able to see them.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -186,7 +186,10 @@ inquirer
           'git add src/lib/version.js yarn.lock package.json CHANGELOG.md README.md CONTRIBUTING.md'
         );
         shell.exec(`git commit -m "${commitMessage}"`);
-        shell.exec(`git tag "v${newVersion}"`);
+        // if tagged, it is not possible to generate a nice changelog without manual updates
+        if (strategy === 'stable') {
+          shell.exec(`git tag "v${newVersion}"`);
+        }
 
         shell.echo(
           colors.yellow.underline(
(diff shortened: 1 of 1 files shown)


iam4x:
I agree, we might not want to have beta changelog in the final one. But, I would appreciate to still have changes for the beta in a changelog.

Can we try to have this flow for beta:

* tag locally the beta
* generate the changelog & update it
* remove the tag locally

And when we have a non-beta, re-generate the changelog from start so it removes the "beta" parts of it? What do you think?

bobylito:
Betas are only relevant for a month max and local tags prevent us from being all able to release. Otherwise we can generate the changelog everytime with tags. But I feel like most people don't care about the betas so it would be just noise.

Haroenv:
What about leaving only most recent beta, and overwriting after the beta is superseded by a regular release?

bobylito:
> What about leaving only most recent beta, and overwriting after the beta is superseded by a regular release? (in changelog)

Too much trouble imho. I'd rather go with something simpler for now and improve the solution afterwards.

bobylito:
Can we move forward with that? @Haroenv @iam4x ?

PR #2831, opened 2018-03-20
fix(removeWidget): check for widgets.length on next tick

iam4x:
I'm using the `removeWidget` API for Angular InstantSearch when the app need to unmount a widget.

If the page is unmounting all the widgets, it will call `removeWidget()` for every widget present on the page and this will result of make multiple requests to Algolia even thought we are unmounting the whole page.

In order to workaround this issue, I'm deferring the execution of the condition `this.widgets.length > 0` on the next tick so every widgets would have been removed.

This is an easy-fix for now until we can maybe think about a better API with InstantSearch core 👍

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -193,10 +193,16 @@ Usage: instantsearch({
       }
     });
 
-    // no need to trigger a search if we don't have any widgets left
-    if (this.widgets.length > 0) {
-      this.helper.search();
-    }
+    // If there's multiple call to `removeWidget()` let's wait until they are all made
+    // and then check for widgets.length & make a search on next tick
+    //
+    // This solves an issue where you unmount a page and removing widget by widget
+    setTimeout(() => {
+      // no need to trigger a search if we don't have any widgets left
+      if (this.widgets.length > 0) {
+        this.helper.search();
+      }
+    }, 0);
   }
 
   /**


Haroenv:
does wrapping it in Promise.resolve also fix it?

iam4x:
Yes it should have the same behaviour 👍

samouss:
@iam4x agree that's an easy fix, we have the same strategy in [React InstantSearch](https://github.com/algolia/react-instantsearch/blob/7c19c75df4856a3e5afe3c79fc9dee470d956633/packages/react-instantsearch/src/core/createWidgetsManager.js#L15-L18). But we should not rely on that since it's not predictable at all. For example, in React when they will launch the async rendering the lib could cause multiple request because of that. We should find a proper API to handle that case, it's a recurring problem.

Haroenv:
How can we handle both “yes we want a new request if a widget is unmourned, but none if we’re also going to delete everything”. Maybe requestIdleCallback, but that doesn’t work in node. All I can think of is that we need to have a queue that must be flushed, but I can’t think of an implementation without timers

vvo:
The only way to do it properly in every flavour, without relying on setTimeout 0, is to know the inner details of the framework (react, angular, vue) behaviour about unmounting. For example maybe in Angular unmounting hooks are called and scheduled using setTimeout 0, maybe with Promise.resolve() or anything..

Once you know that, you can build a small scheduler inside the framework (angular, react, vue). This scheduler must be responsible for stacking widgets mounting/unmounting to avoid calling .search() everytime.

Which also means from my POV that nor removeWidget, nor addWidget should be responsible for calling .search(). At least flavours other than the vanilla one should be able to call removeWidget() without having it triggering a search (example: removeWidget({autoSearch: false})).

The vanilla one could still have great use of this current fix. Knowing the limitations of this fix and it's implication is also important for other frameworks. And we need to check the behaviour when used in Vue (cc @Haroenv).

bobylito:
I have pretty much the same opinion than @vvo on the problem. I'm also for integrating this change  for now and iterate. 

For the glory, allow me to add some elements:

First thing, I think we should define the problem. We want to *batch operations* and yet *minimize* the waiting time before the execution. It's what @vvo mentioned when he talked about scheduling. This problem is inherently based on time. That's why we cannot do without time based callback.

Maybe using some heuristics we  could cut the time. For example, if we've already unmounted all the widgets, maybe it's not meaningful to wait for another x milliseconds to trigger a new search (considering that it's meaningful to do one at all), so we could shortcut the waiting time.

Beyond this definition of the problem, we can also say that in our case we have a lot widget removal and therefore lots of new requests. For 10 removed widgets, we have 10 requests. If we  could batch in 2 of 5, we'll 5x less queries which is already good. In that kind of problem, better is already good.

Eventually when renderers are all asynchronous, it will mean that we'll have time when the system will be stable and unstable. I hope that the frameworks will provide an API for knowing when the system is back to being stable so that we can do the cleanup. Hence, as proposed by @vvo, eventually a method like `start` that triggers the search again could be useful.

But then, here I think that if this change actually helps mitigating the issue on Angular that's a good step. Eventually, as frameworks evolve we should be able to provide more control to the implementors of framework flavors.

Haroenv:
I'm +1 on @vvo's idea of putting this under a flag

bobylito:
> I'm +1 on @vvo's idea of putting this under a flag

I don't think it's necessary.

samouss:
You take a look at this package for the perf about `timeout` / `Promise` etc... 

> https://github.com/jamiebuilds/tickedoff#perf

PR #2837, opened 2018-03-27
chore(doc): document limitations for _geoloc: []

bobylito:


src/widgets/geo-search/geo-search.js
```diff
@@ -119,6 +119,8 @@ Full documentation available at https://community.algolia.com/instantsearch.js/v
  *
  * Note that the GeoSearch widget uses the [geosearch](https://www.algolia.com/doc/guides/searching/geo-search) capabilities of Algolia. Your hits **must** have a `_geoloc` attribute in order to be displayed on the map.
  *
+ * Currently, the feature is not compatible with multiple values in the _geoloc attribute for the moment.
```
Haroenv: `Currently, the feature is not compatible with multiple values in the _geoloc attribute.`

(otherwise you're saying the same thing twice)

bobylito:
Also it was on the wrong widget :/

PR #2839, opened 2018-03-27
Feat/connect configure

### LINKED ISSUE #2816: should configure be a connector?
Right now `configure` is a widget which doesn't render anything, just like `analytics`.

PRO: 

- easier when we split up core and widgets
- other libraries only need to handle `connectors` (not `widgetFactories`)

CON:

- this might imply there's a value in multiple "kinds" of configure
- we might need to handle `refine`

iam4x:
Fixes #2816 

* `renderFn` && `unmountFn` would be only called if they were provided
* `renderFn` gets a `refine()` function to apply new search parameters

Flow of `refine()` :

1. Remove last `searchParameters` applied on the state
2. Compute new state by merging the new `searchParameters` and the actual state
3. Trigger a search with the new state
4. Replace `widgetParams.searchParameters` by the new one

This should cover the cases we discussed in the issue 👍

### DIFF
--- a/src/connectors/configure/__tests__/connectConfigure-test.js
+++ b/src/connectors/configure/__tests__/connectConfigure-test.js
@@ -54,7 +54,8 @@ describe('connectConfigure', () => {
   });
 
   it('should apply new searchParameters on refine()', () => {
-    const makeWidget = connectConfigure();
+    const renderFn = jest.fn();
+    const makeWidget = connectConfigure(renderFn, jest.fn());
     const widget = makeWidget({ searchParameters: { analytics: true } });
 
     helper.setState(widget.getConfiguration());
@@ -63,7 +64,10 @@ describe('connectConfigure', () => {
     expect(widget.getConfiguration()).toEqual({ analytics: true });
     expect(helper.getState().analytics).toEqual(true);
 
-    widget._refine({ hitsPerPage: 3 });
+    const { refine } = renderFn.mock.calls[0][0];
+    expect(refine).toBe(widget._refine);
+
+    refine({ hitsPerPage: 3 });
 
     expect(widget.getConfiguration()).toEqual({ hitsPerPage: 3 });
     expect(helper.getState().analytics).toBe(undefined);
--- a/src/connectors/configure/__tests__/connectConfigure-test.js
+++ b/src/connectors/configure/__tests__/connectConfigure-test.js
@@ -18,13 +18,11 @@ describe('connectConfigure', () => {
     });
 
     it('with a renderFn but no unmountFn', () => {
-      const makeWidget = connectConfigure(jest.fn(), undefined);
-      expect(() => makeWidget({ searchParameters: {} })).toThrow();
+      expect(() => connectConfigure(jest.fn(), undefined)).toThrow();
     });
 
     it('with a unmountFn but no renderFn', () => {
-      const makeWidget = connectConfigure(undefined, jest.fn());
-      expect(() => makeWidget({ searchParameters: {} })).toThrow();
+      expect(() => connectConfigure(undefined, jest.fn())).toThrow();
     });
   });
 
--- a/src/connectors/configure/__tests__/connectConfigure-test.js
+++ b/src/connectors/configure/__tests__/connectConfigure-test.js
@@ -11,24 +11,21 @@ describe('connectConfigure', () => {
     helper = algoliasearchHelper(fakeClient, '', {});
   });
 
-  it('throws on bad usage', () => {
-    // without searchParameters
-    {
+  describe('throws on bad usage', () => {
+    it('without searchParameters', () => {
       const makeWidget = connectConfigure();
       expect(() => makeWidget()).toThrow();
-    }
+    });
 
-    // with a renderFn but no unmountFn
-    {
+    it('with a renderFn but no unmountFn', () => {
       const makeWidget = connectConfigure(jest.fn(), undefined);
       expect(() => makeWidget({ searchParameters: {} })).toThrow();
-    }
+    });
 
-    // with a unmountFn but no renderFn
-    {
+    it('with a unmountFn but no renderFn', () => {
       const makeWidget = connectConfigure(undefined, jest.fn());
       expect(() => makeWidget({ searchParameters: {} })).toThrow();
-    }
+    });
   });
 
   it('should apply searchParameters', () => {
--- a/src/connectors/configure/__tests__/connectConfigure-test.js
+++ b/src/connectors/configure/__tests__/connectConfigure-test.js
@@ -75,7 +75,7 @@ describe('connectConfigure', () => {
     expect(helper.getState().hitsPerPage).toBe(3);
   });
 
-  it('should dispose all the state setted by configure', () => {
+  it('should dispose all the state set by configure', () => {
     const makeWidget = connectConfigure();
     const widget = makeWidget({ searchParameters: { analytics: true } });
 
--- a/src/connectors/configure/__tests__/connectConfigure-test.js
+++ b/src/connectors/configure/__tests__/connectConfigure-test.js
@@ -1,0 +1,92 @@
+import algoliasearchHelper, { SearchParameters } from 'algoliasearch-helper';
+
+import connectConfigure from '../connectConfigure.js';
+
+const fakeClient = { addAlgoliaAgent: () => {}, search: jest.fn() };
+
+describe('connectConfigure', () => {
+  let helper;
+
+  beforeEach(() => {
+    helper = algoliasearchHelper(fakeClient, '', {});
+  });
+
+  it('throws on bad usage', () => {
+    // without searchParameters
+    {
+      const makeWidget = connectConfigure();
+      expect(() => makeWidget()).toThrow();
+    }
+
+    // with a renderFn but no unmountFn
+    {
+      const makeWidget = connectConfigure(jest.fn(), undefined);
+      expect(() => makeWidget({ searchParameters: {} })).toThrow();
+    }
+
+    // with a unmountFn but no renderFn
+    {
+      const makeWidget = connectConfigure(undefined, jest.fn());
+      expect(() => makeWidget({ searchParameters: {} })).toThrow();
+    }
+  });
+
+  it('should apply searchParameters', () => {
+    const makeWidget = connectConfigure();
+    const widget = makeWidget({ searchParameters: { analytics: true } });
+
+    const config = widget.getConfiguration(SearchParameters.make({}));
+    expect(config).toEqual({ analytics: true });
+  });
+
+  it('should apply searchParameters with a higher priority', () => {
+    const makeWidget = connectConfigure();
+    const widget = makeWidget({ searchParameters: { analytics: true } });
+
+    {
+      const config = widget.getConfiguration(
+        SearchParameters.make({ analytics: false })
+      );
+      expect(config).toEqual({ analytics: true });
+    }
+
+    {
+      const config = widget.getConfiguration(
+        SearchParameters.make({ analytics: false, extra: true })
+      );
+      expect(config).toEqual({ analytics: true });
+    }
+  });
+
+  it('should apply new searchParameters on refine()', () => {
+    const makeWidget = connectConfigure();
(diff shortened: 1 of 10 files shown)


Haroenv:
Seems like this got the wrong commits because it was targeting another branch than expected, I guess it needs to be rebased?

iam4x:
I have rebased and removed the extra commits, yes I started my work from develop and not feat/2.7 that's why. It should be good now.

src/connectors/configure/__tests__/connectConfigure-test.js
```diff
@@ -0,0 +1,92 @@
...
+    const widget = makeWidget({ searchParameters: { analytics: true } });
+
+    helper.setState(widget.getConfiguration());
+    widget.init({ helper });
+
+    expect(widget.getConfiguration()).toEqual({ analytics: true });
+    expect(helper.getState().analytics).toBe(true);
+
+    const nextState = widget.dispose({ state: helper.getState() });
+
```
Haroenv: can we have a test for the behaviour when you refine, and after that dismount?

src/connectors/configure/connectConfigure.js
```diff
@@ -0,0 +1,114 @@
...
+ */
+
+/**
+ * @typedef {Object} ConfigureRenderingOptions
+ * @property {function(searchParameters: Object)} refine Sets new `searchParameters` and trigger a search.
+ * @property {Object} widgetParams All original `CustomConfigureWidgetOptions` forwarded to the `renderFn`.
+ */
+
+/**
+ * **Configure** connector provides the logic to build a custom widget
```
Haroenv: the **Configure** [...]

src/connectors/configure/__tests__/connectConfigure-test.js
```diff
@@ -0,0 +1,92 @@
...
+    expect(helper.getState().analytics).toEqual(true);
+
+    widget._refine({ hitsPerPage: 3 });
+
+    expect(widget.getConfiguration()).toEqual({ hitsPerPage: 3 });
+    expect(helper.getState().analytics).toBe(undefined);
+    expect(helper.getState().hitsPerPage).toBe(3);
+  });
+
+  it('should dispose all the state setted by configure', () => {
```
Haroenv: set by configure

src/connectors/configure/connectConfigure.js
```diff
@@ -0,0 +1,114 @@
...
+          );
+        }
+      },
+
+      refine(helper) {
+        return searchParameters => {
+          // remove old `searchParameters` setted by the widget
+          helper.setState(this.removeSearchParameters(helper.getState()));
+
+          // merge new `searchParameters` with the setted one from other widgets
```
Haroenv: with the ones set from other widgets

src/connectors/configure/connectConfigure.js
```diff
@@ -0,0 +1,114 @@
...
+          const actualState = helper.getState();
+          const nextSearchParameters = enhanceConfiguration({})(actualState, {
+            getConfiguration: () => searchParameters,
+          });
+
+          // trigger a search with the new merged searchParameter
```
Haroenv: searchParameters

src/connectors/configure/connectConfigure.js
```diff
@@ -0,0 +1,114 @@
...
+          helper.setState(nextSearchParameters).search();
+
+          // update original `widgetParams.searchParameter` to the new refined one
```
Haroenv: `widgetParams.searchParameters`

src/widgets/configure/__tests__/configure-test.js
```diff
@@ -9,77 +8,4 @@ describe('configure', () => {
       () => configure(/ok/),
     ].forEach(widget => expect(widget).toThrowError(/Usage/));
   });
-
-  it('Applies searchParameters if nothing in configuration yet', () => {
-    const widget = configure({ analytics: true });
-    const config = widget.getConfiguration(SearchParameters.make({}));
-    expect(config).toEqual({
```
Haroenv: did these tests still pass?

iam4x: Yes, just moved them to the connector side since it holds the logic now

src/connectors/configure/connectConfigure.js
```diff
@@ -0,0 +1,114 @@
...
+              refine: this._refine,
+              widgetParams,
+            },
+            false
+          );
+        }
+      },
+
+      dispose({ state }) {
+        if (isFunction(unmountFn)) unmountFn();
```
bobylito: if unmountFn exist, we already that it is a function. We can just test if it's defined.

src/lib/InstantSearch.js
```diff
@@ -3,13 +3,11 @@
...
 import algoliasearchHelper from 'algoliasearch-helper';
 import forEach from 'lodash/forEach';
-import mergeWith from 'lodash/mergeWith';
-import union from 'lodash/union';
-import isPlainObject from 'lodash/isPlainObject';
 import { EventEmitter } from 'events';
 import urlSyncWidget from './url-sync.js';
 import version from './version.js';
 import createHelpers from './createHelpers.js';
+import { enhanceConfiguration } from './utils.js';
```
bobylito: I'd move that either in its own files or in a dedicated package that is less "everything go". What about "lifecycle.js"?

src/widgets/configure/configure.js
```diff
@@ -29,23 +30,12 @@ Full documentation available at https://community.algolia.com/instantsearch.js/v
  *   })
  * );
  */
-export default function configure(searchParameters = {}) {
-  if (!isPlainObject(searchParameters)) {
+export default function configure(searchParameters) {
+  try {
+    // make it clear we do not have default renderFn && unmountFn for this widget
+    const makeWidget = connectConfigure(undefined, undefined);
```
bobylito: I like the explicitness, but isn't a bit redundant? Why not `connectConfigure()`

src/lib/utils.js
```diff
@@ -412,3 +416,31 @@ function parseAroundLatLngFromString(value) {
     lng: parseFloat(pattern[2]),
   };
 }
+
+function enhanceConfiguration(searchParametersFromUrl) {
```
bobylito: I'd add some tests. This a good opportunity since it's now standalone.

src/connectors/configure/__tests__/connectConfigure-test.js
```diff
@@ -0,0 +1,92 @@
...
+    const makeWidget = connectConfigure();
+    const widget = makeWidget({ searchParameters: { analytics: true } });
+
+    helper.setState(widget.getConfiguration());
+    widget.init({ helper });
+
+    expect(widget.getConfiguration()).toEqual({ analytics: true });
+    expect(helper.getState().analytics).toEqual(true);
+
+    widget._refine({ hitsPerPage: 3 });
```
bobylito: You should extract the function through the render function and also test that the reference is the same at init and render.

src/connectors/configure/connectConfigure.js
```diff
@@ -0,0 +1,114 @@
...
+ * @param {function(ConfigureRenderingOptions)} renderFn Rendering function for the custom **Configure** Widget.
+ * @param {function} unmountFn Unmount function called when the widget is disposed.
+ * @return {function(CustomConfigureWidgetOptions)} Re-usable widget factory for a custom **Configure** widget.
+ */
+export default function connectConfigure(renderFn, unmountFn) {
+  return (widgetParams = {}) => {
+    if (
+      !isPlainObject(widgetParams.searchParameters) ||
+      (isFunction(renderFn) && !isFunction(unmountFn)) ||
+      (!isFunction(renderFn) && isFunction(unmountFn))
```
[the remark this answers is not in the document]
iam4x: Ensure we have a renderFn and an unmountFn if one is specified, they should come by pair.

src/connectors/configure/connectConfigure.js
```diff
@@ -0,0 +1,114 @@
...
+      throw new Error(usage);
+    }
+
+    return {
+      getConfiguration() {
+        return widgetParams.searchParameters;
+      },
+
+      init({ helper }) {
+        this._refine = this.refine(helper);
```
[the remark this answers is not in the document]
iam4x: Even if there is no strong convention on this one, it's still this pattern used in some other connectors. I think we should leave it like that until we agree on a pattern.

src/lib/InstantSearch.js
iam4x: Not the scope of the PR, we should keep this on mind and have it in another one or with the instantsearch-core refactoring.

src/lib/utils.js
iam4x: Not the scope of the PR as well, let's try to keep them small 👍

src/connectors/configure/connectConfigure.js
samouss: Same for the case in the `init` & `render`.

src/widgets/configure/__tests__/configure-test.js
samouss: Yes but you can still test that the widget calls the appropriate functions of the connector.

iam4x: @samouss The widget shipped with instantsearch.js is a "dumb" one there's no logic nor rendering, only forwarding options to the connector. Does not makes senses.

samouss: Yep I know but I prefer to always over tests rather than no tests, but I understood your point. I'm fine with it.

src/lib/InstantSearch.js
bobylito: I don't agree on this one. You are moving some code out of a file, but utils.js is not the good file for that.

src/lib/utils.js
bobylito: I do not agree. You should test this function.

vvo: One way to take a decision on this is to ask ourselves "Is this code already tested by other tests". i.e. if I completely break this module, will tests still pass? If yes then this is an issue and we need to add tests on it. If not, less of an issue but we'd have to discuss about which parts we consider needed for unit tests or not.

As a general rule we can agree that, when moving code around for your own needs, if this code is not tested (our fault), let's add tests to it. Especially since it is not a big function and especially if you are moving this to be reused in a new feature you are developing.

On a side note, this code is probably mine so I'd be happy to write some tests with you @iam4x in pair, deal? Book some time with me!

Parts of the code seems to be tested, maybe not the function itself but the way it's used by others. So most of the tests should already be written: https://github.com/algolia/instantsearch.js/commit/0ea3bef85922b876b4efacf64f3025fb730ee982

src/lib/InstantSearch.js
vvo: Which other functions could we put inside lifecycle.js do you think?

bobylito: I definitely didn't give much information here...

Moving this function into utils is not a good move. It makes the situation worse, because the package didn't have a clear purpose and this function is not even consistent with the one that are already there.

That being putting in a separate file doesn't really make for a clear better situation either.

What about keeping this function in the package where it was and exporting it?

src/lib/utils.js
iam4x: Yes this code need to be tested but does it goes with the feature of this PR?

We had a meeting the day I wrote those comments with the team about having too much big PRs that are hard to review and here I'm trying to apply what we discussed.

I totally agree on the fact that we may have identified a flaw of our code because of the lack of unit testing of this function. But I really think it need to be done in another PR with an another issue linked, because here we are raising the issue.

That being said, I see this can be discussable in this case since the PR is not that big yet.

src/lib/InstantSearch.js
iam4x: Yes, good point @bobylito having the function exported from the original file did not crossed my mind.

src/lib/utils.js
iam4x: Let's start to do stack PRs! I'll drop a PR with the tests, this should fit with everybody?

bobylito:
Thanks for taking the time to add the changes and add the test.

PR #2841, opened 2018-03-28
Feat/connect autocomplete

iam4x:
This connector now matches the react-instantsearch API, it exposes `indices` as an array and a `refine` function to update the query 👍

I've build a simple example with https://selectize.github.io/selectize.js/ which integrates well.

* [ ] documentation
* [ ] tests

This PR replaces #2651


Haroenv:
travis build fails because there's a console.log in scripts/release.js

Haroenv:
widget is [here](https://deploy-preview-2841--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=Autocomplete.default&widgets=jquery) if you want to try it :)

src/connectors/autocomplete/connectAutocomplete.js
```diff
@@ -0,0 +1,119 @@
...
+var customAutcomplete = connectAutocomplete(function render(params, isFirstRendering) {
+  // params = {
+  //   indices,
+  //   hits,
+  //   refine,
+  //   currentRefinement
+  // }
+});
+search.addWiget(customAutcomplete({
+  [ indices ],
```
Haroenv: what about also returning `index` (first index of `indices`) for use cases when you know there will only be one index? Seems to be a way how you can have both the simplicity of a single index when you need it, but avoid the two separate return paths RIS has, cc @samouss

samouss: We don't need to handle the single index in a special case. When it's an array both cases are treated the same. You can still loop on array of one item.

iam4x:
> Rebased on feat/2.8 & changed pull request base branch.

src/connectors/autocomplete/connectAutocomplete.js
```diff
@@ -0,0 +1,153 @@
...
+
+      dispose() {
+        // remove `result` listeners for derived helper
+        this.indices
+          .slice(1)
+          .forEach(({ helper }) => helper.removeAllListeners('result'));
+
+        unmountFn();
+      },
+    };
```
bobylito: What about adding the ability to do some synchronize the parameters here?

src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
```diff
@@ -0,0 +1,70 @@
...
+    const makeWidget = connectAutocomplete(renderFn);
+    const widget = makeWidget({ indices: [{ label: 'foo', value: 'foo' }] });
+
+    const helper = jsHelper(fakeClient, '', {});
+    helper.search = jest.fn();
+
+    widget.init({ helper, instantSearchInstance: {} });
+
+    // original helper + derived one
+    expect(widget.indices).toHaveLength(2);
```
bobylito: I'm not sure about relying on a property that is somewhat internal. What about checking that search is called twice if you have two indices?

src/connectors/autocomplete/connectAutocomplete.js
```diff
@@ -0,0 +1,153 @@
...
+          {
+            helper,
+            label: 'primary',
+            index: helper.getIndex(),
+            results: undefined,
+          },
+        ];
+
+        // add additionnal indices into `this.indices`
+        if (indices && Array.isArray(indices)) {
```
samouss: We can get rid of the second condition because we already throw an error when `indices` is not an array. If we move to the default value we can get rid of the complete condition. The value will always be there as an empty array.

src/connectors/autocomplete/connectAutocomplete.js
```diff
@@ -0,0 +1,153 @@
...
+            );
+          });
+        }
+
+        this.instantSearchInstance = instantSearchInstance;
+        this.renderWithAllIndices({ isFirstRendering: true });
+      },
+
+      saveResults({ results, index }) {
+        const derivedIndex = this.indices.find(i => i.index === index);
```
samouss: By using the name of the index as key we can't have an autocomplete with the same index multiple times. Right now it's useless because we can't provide additional search parameters. But we can image to implement this functionality in the future. We have the issue with React InstantSearch, the idea was to require a `key` to avoid relying on the index name.

src/connectors/autocomplete/connectAutocomplete.js
```diff
@@ -0,0 +1,153 @@
...
+        if (
+          widgetParams.escapeHits &&
+          results.hits &&
+          results.hits.length > 0
+        ) {
+          results.hits = escapeHits(results.hits);
+        }
+
+        derivedIndex.results = results;
+        derivedIndex.hits = results ? results.hits : undefined;
```
samouss: Why don't we use an empty array as default value? It means that we always need to check either the `results` or the `hits` in order to render. With a default value we can always loop on the hits.

src/connectors/autocomplete/connectAutocomplete.js
```diff
@@ -0,0 +1,153 @@
...
+          },
+          isFirstRendering
+        );
+      },
+
+      dispose() {
+        // remove `result` listeners for derived helper
+        this.indices
+          .slice(1)
+          .forEach(({ helper }) => helper.removeAllListeners('result'));
```
samouss: Why don't we use `detach` instead? Otherwise the derived helpers are never cleaned up. The main instance will keep them for ever.

src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
iam4x: I guess a way to know if the search has been done is by using `widgets.indices[x].helper.search = jest.fn()` but it's still using `widget.indices` do you have any proposal?

src/connectors/autocomplete/connectAutocomplete.js
iam4x: Can you explain a bit more, I'm not sure to understand what you are meaning with "synchronize the parameters"

src/connectors/autocomplete/connectAutocomplete.js
iam4x: We have a key called `label` as well, I will switch the find using this key it will solve the issue 👍

src/connectors/autocomplete/connectAutocomplete.js
samouss: Good point 👍 I still prefer the empty array but indeed it's the same. But it's only my opinion, we can keep it like that.

iam4x:
> Inside the stories the select component keep the items in memory. It means that the more we search the longer is the list (see below). We can use the clearOptions() to remove all options before each a new render.

Nice catch, internally selectize keeps in memory options added and re-do a search into them.

> Note that the autocomplete widget is always tied with the main search. For example if we want to only target an different index than the main one we can't. We will need to use another instance and sync them together. The use case I'm thinking of is the e-commerce demo with the suggestions.

Yes it's true, good point, as a solution we can also provide another option to do not use the "main index" and so the user has to specify at least one index in the options. What do you think?

src/connectors/autocomplete/connectAutocomplete.js
samouss: We can add a message in the doc who specifies that this value must be unique, since we are using it for retrieve the correct index.

src/connectors/autocomplete/connectAutocomplete.js
Haroenv: I'd prefer a single type and then be able to check on the number of hits rather than if hits is there or not

iam4x: Alright let's go for it 🚢

src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
samouss: You can rely on the arguments of the `renderFn`. The `indices` property is part of the public API on this function.

bobylito: Yes if it is something you get from renderFn.

iam4x:
* Tests are fixed now
* Added an example of multi-index search: https://deploy-preview-2841--algolia-instantsearch.netlify.com/v2/dev-novel/?widgets=jquery&selectedStory=Autocomplete.Multi%20index

bobylito:
> Added an example of multi-index search: https://deploy-preview-2841--algolia-instantsearch.netlify.com/v2/dev-novel/?widgets=jquery&selectedStory=Autocomplete.Multi%20index

Is it expected to not have a search bar?

screen shot 2018-05-25 at 18 33 59

bobylito:
> I don't think it will solve the issue because even with only a different index the query will still be set on the main helper. For the suggestions example it means that the "suggested" hits and the "regular" hits will always be in sync.

If I sum we have 3 cases (with my thoughts):
 - If it's an independant search. Having independant searches seems fine.
 - If if you want to sync the main results and the autocomplete results but with different indices, the idea to have an option for "not the main index" seems to solve it
 - If you have the main index + others and everything is synced, it's the default case

Did I miss something @iam4x @samouss ?

samouss:
>  - If it's an independant search. Having independant searches seems fine.

Yes it's fine.

> - If if you want to sync the main results and the autocomplete results but with different indices, the idea to have an option for "not the main index" seems to solve it

We can also go in the other directions, do not assume that the autocomplete will be on the main index. WDYT? I'm still fine with the current proposal, it's just that I didn't think about it before.

> - If you have the main index + others and everything is synced, it's the default case

Yes indeed, depends on your answers on the question above.

PR #2845, opened 2018-03-28
fix(connectGeoSearch): correctly dispose the connector

samouss:
**Summary**

Right now there are two issues with the connector when the `dispose` function is called. The first one is that the parameter `insideBoundingBox` is not reset. The second one is that we always reset the parameters on the provided `state` argument instead of resetting the parameters on the new created value. It means that only the latest call will be reflected on the `SearchParameters`.

The PR fixes those two issues.


iam4x:
Should we have this PR on feat/2.7 @bobylito ?

bobylito:
> Should we have this PR on feat/2.7 @bobylito ?

why should it be in 2.7?

PR #2861, opened 2018-04-04
feat(size): add sideEffects false to package.json

### LINKED ISSUE #2859: Tree-shaking doesn't seem to be working with latest Webpack
Bug. Repo to reproduce: https://github.com/mariuslazar93/algolia-instantsearch-test . We are only concered about the final bundle size, so just running `npm install` and `npm run build-prod` will generate a dist folder with the production files.

I am trying to create a production build using Webpack 4.5.0 and make use of tree-shaking to remove the unimported widgets/connectors. 

I'm only importing 3 widgets inside `./src/js/search.js` (listed below too), but the generated bundle report shows that all the widgets and connectors are imported into the final (production) bundle. 

`import instantsearch from 'instantsearch.js/es';`
`
import {
  searchBox,
  hits,
  pagination
`

I'm splitting the bundles into app.bundle.js, vendor.bundle.js (what comes from /node_modules) and runtime.bundle.js (webpack runtime). I've added the webpack-bundle-analyzer to generate the dependencies report of the bundles (as you can see in the image).

It also seems like the prod bundle ends up with a duplicated loadash library which adds up to the final size.

image

Only the imported widgets should show in the final bundle. Only one loadash library.

Doesn't apply.

Seems like Webpack 4+ is using a `sideEffect` property in package.json which dictates if the unused code can be removed safely by the minifier (UglifyJS in this case). More on their docs: https://webpack.js.org/guides/tree-shaking/ . Not sure if this is the problem or not.

Latest.

If I'm missing something please let me know.

Haroenv:
Why is this done? see <https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free>

Long story short, modules can do crazy things, but if we don't do any polyfilling, we can opt into better behaviour

fixes #2859 as good as we can for now

technically depends on #2860 being merged before we release this PR, because it counts as a side-effect


Haroenv:
@oliviertassinari do you have any idea why this has an "empty" screenshot, did something get corrupted for "original"?

bobylito:
Well that's super weird what happens with Argos. I have tested with the website and it seems to work. Let's merge :)

oliviertassinari:
The base screenshot on the * develop* branch might have been corrupted 🤔 .

Haroenv:
looks like that, since other PRs also broke @oliviertassinari

PR #2877, opened 2018-04-10
fix(history): provide location and use named parameters

bobylito:
The API between createURL and parseURL was not symetric. Also it add too much parameters for it to be nice.


bobylito:
It's fast enough :)



> Object always wins!

Let's agree to disagree :)

vvo:
> Is this a breaking change technically? Or do you think it's small and fast enough to be a patch?

Since the release was yesterday and is not documented yet, I'd say let's go for a patch and that's it.

PR #2879, opened 2018-04-10
docs(routing): add a routing guide

vvo:
This PR includes:
- the `routing` guide, [review it live](https://deploy-preview-2879--algolia-instantsearch.netlify.com/v2/guides/routing.html)
- deprecation of the urlSync option via console.warn
- minor change to CONTRIBUTING for dev on examples
- routing for all examples (simple one)
- minor fix for no-results guide

### DIFF
--- a/docgen/src/guides/routing.md
+++ b/docgen/src/guides/routing.md
@@ -9,17 +9,13 @@ editable: true
 githubSource: docgen/src/guides/routing.md
 ---
 
-Via the `routing` option, InstantSearch provides the necessary API entries to allow you to synchronize the state of
-your search UI (which widget were refined, what is the current search query ..) with any kind of storage.
-And most probably you want that storage to be the browser url bar.
+Via the `routing` option, InstantSearch provides the necessary API entries to allow you to synchronize the state of your search UI (which widget were refined, what is the current search query ..) with any kind of storage. And most probably you want that storage to be the browser url bar.
 
-Synchronizing your UI with the browser url is a good practice. It allows any of your user to take one of your result page, copy paste the
-browser url and send it to a friend.
+Synchronizing your UI with the browser url is a good practice. It allows any of your users to take one of your result page, copy paste the browser url and send it to a friend. It also allows your users to use the back and next button of their browser and always endup where they were previously.
 
 This guide provides you examples for the most common use cases along with some advanced ones. You can also read the [routing API reference](instantsearch.html#struct-RoutingOptions).
 
-As you will see with advanced examples, creating your own url scheme to map your needs requires some good coding. But keep in mind
-you will need to do this only once, not every day!
+As you will see with advanced examples, creating your own url scheme to map your needs requires a significant amount of code but keep in mind you will need to do this only once, not every day!
 
 ## Simple url synchronization
 
@@ -32,12 +28,11 @@ const search = instantsearch({
 });
 ```
 
-The resulting url in your browser bar will look like this:
+The resulting url in your browser url bar will look like this:
 
 `https://website.com/?query=a&refinementList%5Bbrand%5D%5B0%5D=Drama`
 
-While
-not being pretty it is still very accurate: the query is `a` and the `brand` attribute, which is a `refinementList`, was refined (clicked) to `Drama`. But if you want something custom and clean, let's move on to more user friendly urls.
+While not being pretty it is still very accurate: the query is `a` and the `brand` attribute, which is a `refinementList`, was refined (clicked) to `Drama`. But if you want something custom and clean, let's move on to more user friendly urls.
 
 ## User friendly urls
 
@@ -45,17 +40,13 @@ You might want your urls to look like:
 
 `https://website.com/?query=a&brands=Sony~Samsung&page=2`
 
-This way your users to be able to more easily read them when shared via emails, documents, social media...
+This way your users will be able to read them more easily when shared via emails, documents, social media...
 
-To do so, the `routing` option accepts a simple boolean but also more complex objects to allow for customization. The first customization option
-you want to use is `stateMapping`. It allows you to define more precisely how the state of your search will be synchronized to your
-browser bar (or any other router storage you might have).
+To do so, the `routing` option accepts a simple boolean but also more complex objects to allow customization. The first customization option you want to use is `stateMapping`. It allows you to define more precisely how the state of your search will be synchronized to your browser url bar (or any other router storage you might have).
 
 Here's an example achieving just that ([and here's the live version](https://codesandbox.io/s/005py264m0)):
 
-This example assumes that you have added the [searchBox](widgets/searchBox.html), [refinementList](widgets/refinementList.html)
-and [pagination](widgets/pagination.html) widgets to your search UI. Then the `refinementList` is activated on the `brands` attribute.
-Please adjust given your own data.
+This example assumes that you have added the [searchBox](widgets/searchBox.html), [refinementList](widgets/refinementList.html) and [pagination](widgets/pagination.html) widgets to your search UI. Then the `refinementList` is activated on the `brands` attribute. Please adjust given your own data.
 
 ```javascript
 const search = instantsearch({
@@ -84,12 +75,11 @@ const search = instantsearch({
 });
 ```
 
-There's a lifecycle dictating when the `stateMapping` functions are called:
+There's a lifecycle in which when the `stateMapping` functions are called:
 - `stateToRoute` is called whenever widgets are refined (clicked). It is also called everytime any widget needs to create a URL.
 - `routeToState` is called whenever the user loads, reloads the page or click on back/next buttons of the browser.
 
-To build your own mapping easily, just `console.log(uiState)` and see what you're getting. Note that the object you return in `stateToRoute` will be the one
-you'll receive as an argument in `routeToState`.
+To build your own mapping easily, just `console.log(uiState)` and see what you're getting. Note that the object you return in `stateToRoute` will be the one you'll receive as an argument in `routeToState`.
 
 ## SEO friendly urls
 
@@ -154,20 +144,16 @@ const search = instantsearch({
 });
 ```
 
-As you can see, we are now using the [instantsearch.routers.history](routing.html#instantsearchroutershistory-api) so that we can explicitly set options on the default
-router mechanism used in the previous example. What we see also is that both the `router` and `stateMapping` options can be used
-together as a way to easily map `uiState` to `routeState` and vice versa.
+As you can see, we are now using the [instantsearch.routers.history](routing.html#instantsearchroutershistory-api) so that we can explicitly set options on the default router mechanism used in the previous example. What we see also is that both the `router` and `stateMapping` options can be used together as a way to easily map `uiState` to `routeState` and vice versa.
 
 Using that we can configure:
 - `windowTitle`: This method can be used to map the object (`routeState`) returned from `stateToRoute` to your window title
-- `createURL`: This method is called everytime we need to create a url. When we want to synchronize the `routeState` to the browser url bar,
-when we want to render `<a href>` tags in the `menu` widget or when you call `createURL` in one of your connectors rendering method
+- `createURL`: This method is called everytime we need to create a url. When we want to synchronize the `routeState` to the browser url bar, when we want to render `<a href>` tags in the `menu` widget or when you call `createURL` in one of your connectors's rendering method
 - `parseURL`: This method is called everytime the user loads, reloads the page or click on back/next buttons of the browser
 
 ### About SEO
 
-For your search results to be part of search engines results, you will have to selectively choose them. Trying to have all of your search results
-inside search engines could be considered as spam by them.
+For your search results to be part of search engines results, you will have to selectively choose them. Trying to have all of your search results inside search engines could be considered as spam by them.
 
 To do that, you can create a [`robots.txt`](http://www.robotstxt.org/) and host it at `https://website.com/robots.txt`.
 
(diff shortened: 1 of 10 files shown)


vvo:
One thing I wanted to achieve is to easily handle the case of already existing querystring parameters.

For example let's say I load https://website.com/?loggedIn=1 and I want to use the routing feature. What I'd like to do is:

```js
  // ... appId, apiKey...
  routing: {
    stateMapping: {
      stateToRoute(uiState) {
        return {
		  loggedIn: window.isUserLoggedIn
          q: uiState.query
		};
      },
      routeToState(routeState) {
        return {
          query: routeState.q
        };
      }
    }
  }
});
```

This ^ works and will add loggedIn query parameter to the url. But what if window.isUserLoggedIn state changes from the outside (user logs out, it's an ajax call)? I would like to turn the url to: `https://website.com/?loggedIn=0&q=phone`.

To do so for now I guess I would have to create my own router or own widget?

What I may want as a user:

```js
const search = instantsearch(...);
search.router.write(); // triggers write on the history router
```

There are implications to this so I just want to have your opinion on the proposal.

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,276 @@
...
+withHeadings: true
+editable: true
+githubSource: docgen/src/guides/routing.md
+---
+
+Via the `routing` option, InstantSearch provides the necessary API entries to allow you to synchronize the state of
+your search UI (which widget were refined, what is the current search query ..) with any kind of storage.
+And most probably you want that storage to be the browser url bar.
+
+Synchronizing your UI with the browser url is a good practice. It allows any of your user to take one of your result page, copy paste the
```
Haroenv: your users

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,276 @@
...
+browser url and send it to a friend.
+
+This guide provides you examples for the most common use cases along with some advanced ones. You can also read the [routing API reference](instantsearch.html#struct-RoutingOptions).
+
+As you will see with advanced examples, creating your own url scheme to map your needs requires some good coding. But keep in mind
```
Haroenv: a significant amount of code, but

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,276 @@
...
+  // ... appId, apiKey...
+  routing: true
+});
+```
+
+The resulting url in your browser bar will look like this:
+
+`https://website.com/?query=a&refinementList%5Bbrand%5D%5B0%5D=Drama`
+
+While
```
Haroenv: While not (one line for legibility of the source)

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,276 @@
...
+not being pretty it is still very accurate: the query is `a` and the `brand` attribute, which is a `refinementList`, was refined (clicked) to `Drama`. But if you want something custom and clean, let's move on to more user friendly urls.
+
+## User friendly urls
+
+You might want your urls to look like:
+
+`https://website.com/?query=a&brands=Sony~Samsung&page=2`
+
+This way your users to be able to more easily read them when shared via emails, documents, social media...
```
Haroenv: This way your users will be able to read them more easily when

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,276 @@
...
+
+To do so, the `routing` option accepts a simple boolean but also more complex objects to allow for customization. The first customization option
+you want to use is `stateMapping`. It allows you to define more precisely how the state of your search will be synchronized to your
+browser bar (or any other router storage you might have).
```
Haroenv: URL bar

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,276 @@
...
+Be it for SEO benefits or to align your search UI urls with your current sitemap and existing url scheme.
+
+Here's an example achieving just that ([and here's the live version](https://codesandbox.io/s/l5z6zz4yoq)):
+
+```javascript
+const search = instantsearch({
+  // ... appId, apiKey...
+  routing: {
+    router: instantsearch.routers.history({
+      windowTitle: function(routeState) {
```
Haroenv: we can use either all modern features (object shorthand) or none (no template strings either) so that people can copy-paste it in sites without build process

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,276 @@
...
+```
+
+As you can see, we are now using the [instantsearch.routers.history](routing.html#instantsearchroutershistory-api) so that we can explicitly set options on the default
+router mechanism used in the previous example. What we see also is that both the `router` and `stateMapping` options can be used
+together as a way to easily map `uiState` to `routeState` and vice versa.
+
+Using that we can configure:
+- `windowTitle`: This method can be used to map the object (`routeState`) returned from `stateToRoute` to your window title
+- `createURL`: This method is called everytime we need to create a url. When we want to synchronize the `routeState` to the browser url bar,
+when we want to render `<a href>` tags in the `menu` widget or when you call `createURL` in one of your connectors rendering method
```
Haroenv: `connectors'`

src/lib/InstantSearch.js
```diff
@@ -82,10 +82,28 @@ Usage: instantsearch({
...
+      );
+      console.warn('You can now use the new `routing` option');
+
+      if (urlSync === true) {
+        // when using urlSync: true
+        console.warn('Use it like this: `routing: true`');
+      }
+
+      console.warn(
+        'For advanced usages docs, check out https://community.algolia.com/instantsearch.js/v2/guides/routing.html#migrating-from-urlsync'
```
Haroenv: usage documentation

src/lib/InstantSearch.js
```diff
@@ -82,10 +82,28 @@ Usage: instantsearch({
       this._searchFunction = searchFunction;
     }
 
-    if (urlSync && routing) {
-      throw new Error(
-        'InstantSearch configuration error: it is not possible to use `urlSync` and `routing` at the same time'
```
bobylito: I would still keep the error because the behavior is completely undefined.

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,276 @@
...
+editable: true
+githubSource: docgen/src/guides/routing.md
+---
+
+Via the `routing` option, InstantSearch provides the necessary API entries to allow you to synchronize the state of
+your search UI (which widget were refined, what is the current search query ..) with any kind of storage.
+And most probably you want that storage to be the browser url bar.
+
+Synchronizing your UI with the browser url is a good practice. It allows any of your user to take one of your result page, copy paste the
+browser url and send it to a friend.
```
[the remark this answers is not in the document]
Haroenv: I wouldn't mention it, because it's tricky to say if we handled it correctly (is it correct to have a new history entry for every refinement or not)

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,276 @@
...
+          },
+          page: routeState.page
+        };
+      }
+    }
+  }
+});
+```
+
+There's a lifecycle dictating when the `stateMapping` functions are called:
```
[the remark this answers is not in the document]
Haroenv: There's a lifecycle in which the `stateMapping` functions are called

clear enough IMo

docgen/src/guides/routing.md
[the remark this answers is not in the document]
vvo: Oh yes but Ill definitely mention that you can use back/next buttons and reload without any issue, yes!

docgen/src/guides/routing.md
vvo: I want to keep "routeState" here for the sake of explaining which kind of objects you are receiving. Users can then write that however they want. Works?

src/lib/InstantSearch.js
vvo: The error is right here but below (after the warnings)

vvo: I will put it on top again it's better

vvo:
Thanks a lot to you two for the extensive review. I implemented every change you suggested, they were on point.

I also removed the super weird wrapping I had, sorry about that. I now have configured vscode the right way to handle it.

src/lib/InstantSearch.js
bobylito: Sorry, missed it 👍

bobylito:
🚀

docgen/src/guides/routing.md
Haroenv: I meant:

```diff
- windowTitle: function(routeState) {
+ windowTitle(routeState) {
```

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,258 @@
...
+    }
+  }
+});
+```
+
+There's a lifecycle in which when the `stateMapping` functions are called:
+- `stateToRoute` is called whenever widgets are refined (clicked). It is also called everytime any widget needs to create a URL.
+- `routeToState` is called whenever the user loads, reloads the page or click on back/next buttons of the browser.
+
+To build your own mapping easily, just `console.log(uiState)` and see what you're getting. Note that the object you return in `stateToRoute` will be the one you'll receive as an argument in `routeToState`.
```
francoischalifour: Why not give them the actual object format?

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,258 @@
...
+  }
+});
+```
+
+As you can see, we are now using the [instantsearch.routers.history](routing.html#instantsearchroutershistory-api) so that we can explicitly set options on the default router mechanism used in the previous example. What we see also is that both the `router` and `stateMapping` options can be used together as a way to easily map `uiState` to `routeState` and vice versa.
+
+Using that we can configure:
+- `windowTitle`: This method can be used to map the object (`routeState`) returned from `stateToRoute` to your window title
+- `createURL`: This method is called everytime we need to create a url. When we want to synchronize the `routeState` to the browser url bar, when we want to render `<a href>` tags in the `menu` widget or when you call `createURL` in one of your connectors's rendering method
+- `parseURL`: This method is called everytime the user loads, reloads the page or click on back/next buttons of the browser
```
francoischalifour: `click` → `clicks`

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,258 @@
...
+
+```txt
+User-agent: *
+Allow: /search/q/phones/brands/Samsung/p1
+Allow: /search/q/phones/brands/Apple/p1
+Disallow: /search/
+Allow: *
+```
+
+Now that you know how to create such advanced url synchronization mechanism, the only next step would be to create your own router. But let's first have a look at the full `uiState` reference.
```
francoischalifour: `only next` → `final`?

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,258 @@
...
+
+## `instantsearch.routers.history` API
+
+InstantSearch.js provides a default router under `instantsearch.routers.history`. You can use it when you want to go futher than just aliasing querystring parameters in the url. For example if you want to generate urls like `https://website.com/search/q/phone/brands/Sony~Samsung/p/1`.
```
francoischalifour: `querystring parameters` → `query string parameters`

(also change references below if you decide to change it here)

We should probably add a link to the [definition of a Query String](https://en.wikipedia.org/wiki/Query_string) since this is the first time we talk about it in this guide.

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,258 @@
...
+This function allows you to dynamically customize the window title based on the provided `routeState`.
+This function is called every time the user refines the UI, after the history timer.
+
+You must return a `string`.
+
+- **`opts.createURL: function({qsModule, location, routeState})`**
+This function allows you to directly change the format of urls that will be created and rendered to the browser url bar or widgets.
+This function is called everytime InstantSearch.js needs to create a URL. **The provided options** are:
+  - *`qsModule`: object*, a querystring parsing and stringifying module, [full documentation](https://github.com/ljharb/qs). We use it internally so we provide it to you as a convenience
+  - *`location`: function*, alias to window.location
```
francoischalifour: "window.location" → "`window.location`" (wrap with `code` tag)

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,258 @@
...
+  - *`routeState`: object*, the `routeState` created by the `stateMapping` provided. When no `stateMapping` is provided, this is an untouched `uiState`
+
+You must return a `string`.
+
+- **`opts.parseURL: function({qsModule, location})`**
+This function is responsible for parsing back the url string to a `routeState`. This function must be customized if you customized the `createURL` function.
+This function is called everytime the user loads, reloads or click on back/next buttons of the browser. **The provided options** are:
```
francoischalifour: `click` → `clicks`

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,258 @@
...
+  - *`qsModule`: object*, a querystring parsing and stringifying module, [full documentation](https://github.com/ljharb/qs). We use it internally so we provide it to you as a convenience.
+  - *`location`: function*, alias to window.location.
```
francoischalifour: "window.location" → "`window.location`" (wrap with `code` tag)

docgen/src/guides/routing.md
```diff
@@ -0,0 +1,258 @@
...
+
+You must return an `object`. Which is a `routeState`.
+
+- **`opts.writeDelay: number, default 400`**
+This option controls the number of milliseconds to wait before actually writing the new url to the browse urlr bar. You can think about it this way:
```
francoischalifour: `urlr` → `URL`

bobylito:
> There are implications to this so I just want to have your opinion on the proposal.

I've had the same sort of requirement. This is too awkward that you have to trigger the `change` yourself on the Helper.

Haroenv:
Agree, would be nice to call write in those scenarios

PR #2881, opened 2018-04-11
docs(routing): remove unused parameters

vvo:


vvo:
Since we have an API documentation it's sufficient

PR #2885, opened 2018-04-13
 fix(ie): do not rely on Object.assign 

bobylito:
We have had reports of users wanting to use InstantSearch.js with IE and that was failing because there are some parts of the code base that rely on Object.assign. We were also using _.assign as a polyfill in other parts.

Since both have the same semantic as {...a} and we use the latest already often in the code, let's move to that (and it's correctly [polyfilled by Babel](https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&code_lz=IYAgvCDewFwgLATgL4CgBG4rrgIgCYBm-AzrmgMZaQB0dwANCHTeskA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&lineWrap=false&presets=es2015%2Ces2016%2Ces2017%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3%2Ces2015-loose&prettier=false&targets=&version=6.26.0&envVersion=))


bobylito:
So yeah I removed the record system for faking the search because:
 - it was generating errors (even though it was working locally)
 - we have splitted the build in 3 steps so it's easy to restart only the functional tests

Now we're good :)

bobylito:
I removed again the tests with the recorded data because, they are to update and it's hard to understand what fails. Let's keep that in mind when we want to dig again in this kind of recorded approach (which I think would be great)

PR #2888, opened 2018-04-15
Use url that is pushed via routing for analytics

andymac4182:
**Summary**
The URL is pushed via routing. 

**Result**
This will use the value from the browser instead of trying to replicate the URL.


andymac4182:
@bobylito Personally in the case of Google Analytics it doesn't make sense to send a URL that never existed and might not actually work. If there is extra parameters that aren't shown in the URL that aren't a default I would be surprised. 

I think doing a manual format of the URL should be for advanced users personally since I would imagine most people would be similar to myself and just want the correct page view stats which is based on the routing.

bobylito:
Well I think that's sound, thanks for the explanation @andymac4182

PR #2890, opened 2018-04-17
fix(geosearch): support for routing

bobylito:
Geosearch routing support is missing. This adds it :)


bobylito:
We're good to go now cc @samouss :)

src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -192,9 +193,32 @@ const connectGeoSearch = (renderFn, unmountFn) => {
...
+          lng: parseFloat(boundingBox[3]),
+        },
+      };
+    };
+    let getSearchBoundsInstance;
+
     const init = initArgs => {
       const { state, helper, instantSearchInstance } = initArgs;
       const isFirstRendering = true;
+      getSearchBoundsInstance = getSearchBounds(helper);
```
samouss: I don't think we need to save the instance. We keep the exact same instance only for the function that will trigger a new render (because otherwise we have stalled values in the next render). This function works like `refine` or `clearMapRefinement`.

src/widgets/geo-search/GeoSearchRenderer.js
```diff
@@ -32,6 +32,41 @@ const refineWithMap = ({ refine, paddingBoundingBox, mapInstance }) => {
...
+
+  const ne = mapInstance.getProjection().fromPointToLatLng(northEastPoint);
+  const sw = mapInstance.getProjection().fromPointToLatLng(southWestPoint);
+
+  return {
+    northEast: { lat: ne.lat(), lng: ne.lng() },
+    southWest: { lat: sw.lat(), lng: sw.lng() },
+  };
+};
+
```
samouss: What do you think about trying to merge the two functions that handle bounds?

src/widgets/geo-search/GeoSearchRenderer.js
```diff
@@ -196,14 +232,26 @@ const renderer = (
...
+    });
+    renderState.mapInstance.fitBounds(
+      new googleReference.maps.LatLngBounds(
+        boundsWithPadding.southWest,
+        boundsWithPadding.northEast
+      ),
+      0
+    );
+    renderState.isUserInteraction = true;
+  } else if (hasMarkers && enableFitBounds) {
```
samouss: We can add a test for this case.

src/widgets/geo-search/GeoSearchRenderer.js
```diff
@@ -20,7 +20,7 @@ const refineWithMap = ({ refine, paddingBoundingBox, mapInstance }) => {
     .getProjection()
     .fromLatLngToPoint(mapInstance.getBounds().getSouthWest());
 
-  southWestPoint.x = southWestPoint.x + paddingBoundingBox.right / scale;
+  southWestPoint.x = southWestPoint.x + paddingBoundingBox.left / scale;
```
samouss: Nice catch!

src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -332,6 +358,27 @@ const connectGeoSearch = (renderFn, unmountFn) => {
...
+
+      getWidgetSearchParameters(searchParameters, { uiState }) {
+        if (uiState.boundingBox) {
+          return searchParameters.setQueryParameter(
+            'insideBoundingBox',
+            uiState.boundingBox
+          );
+        }
+        return searchParameters;
+      },
```
samouss: We can add tests for those two functions.

src/widgets/geo-search/GeoSearchRenderer.js
bobylito: The functions works in two different directions, so I'd rather have two functions. I can refactor them though.

src/widgets/geo-search/GeoSearchRenderer.js
bobylito: Ok I will have a look.

src/connectors/geo-search/connectGeoSearch.js
bobylito: The Helper is mutable and will always be up to date. It's not great to generate at each render the exact same function. In other widgets we try to be careful about those, and we cache the refine functions.

src/connectors/geo-search/connectGeoSearch.js
bobylito: Of course!

src/widgets/geo-search/GeoSearchRenderer.js
samouss: Agree it's better to have two separate functions (sorry my comment was not very clear). But since we have almost the same logic in the function we can refactor it a bit I think.

bobylito: agreed

src/connectors/geo-search/connectGeoSearch.js
samouss: In which context it's not great? For performance?

samouss: Since we don't have performance issue I prefer to go for the more simple way, that is (IMO) to create the function every time. Indeed we pay some performance costs but we don't have any issue regarding performance for now. At the end both solutions work, I'm fine with it. Just let's be consistent across the widget.

src/widgets/geo-search/GeoSearchRenderer.js
bobylito: That's the story of my first CS project at university.

src/widgets/geo-search/GeoSearchRenderer.js
bobylito: Can you tell me what kind of test do you have in mind here @samouss ?

samouss: We can check that both cases correctly call fitBounds with the proper values. In the previous implementation the two tests removed does this kind of assertions.

bobylito: I added a test @samouss  :) Not the greatest, but it's okayish

bobylito:
After checking with @samouss we're need to post pone this PR to rethink the flow of control.

bobylito:
We're probably gonna have to do deeper work. Let's close this PR for now.

PR #2894, opened 2018-04-17
feat(search-client): Add support for Custom Search Clients

francoischalifour:
_This PR adds supports for the `searchClient` parameter. This allows any InstantSearch user to plug his own search client (including a custom Algolia backend)._

## Summary

The goal is to add support for this new feature but to still make InstantSearch work as it works right now. This is meant to be released in the next minor version. Checks have been made to make sure both APIs (current and new) don't collide.

```javascript
// With the Algolia JS client
instantsearch({
  indexName: 'concert_events_instantsearchjs',
  searchClient: algoliasearch('latency', '059c79ddd276568e990286944276464a'),
});

// With a custom search client
instantsearch({
  indexName: 'concert_events_instantsearchjs',
  searchClient: {
    search(requests) {
      // perform requests...
      return response;
    },
  },
});
```

### Features

* Add `searchClient` option to plug a custom search client (other than `algoliasearch`)
* Deprecate `createAlgoliaClient` in favor of `searchClient`
* Add tests to make sure the current and new API don't collide

## Releasing strategy

* the next minor version will embed `algoliasearch`, although we ship this new `searchClient` parameter
* the next major version will get rid of `algoliasearch` that will need to be imported beforehand and plugged to `searchClient`

## Demo

You can check the Calendar Widget demo using InstantSearch with `searchClient`: https://instantsearch-search-client.now.sh

---

Let me know if I've forgotten any parameter collision 🙌


src/lib/InstantSearch.js
```diff
@@ -52,22 +89,28 @@ class InstantSearch extends EventEmitter {
...
-  apiKey: 'my_search_api_key',
-  indexName: 'my_index_name'
-});`;
-      throw new Error(usage);
-    }
 
-    const client = createAlgoliaClient(algoliasearch, appId, apiKey);
-    client.addAlgoliaAgent(`instantsearch.js ${version}`);
+    // eslint-disable-next-line prefer-rest-params
+    const [THROWAWAY_PARAMS] = arguments;
```
Haroenv: this line is a bit odd, what about giving the args to this function this name, and destructure after this line?

src/lib/__tests__/search-client-test.js
```diff
@@ -0,0 +1,44 @@
...
+    const search = new InstantSearch({
+      searchClient: searchClientSpy,
+    });
+
+    expect(searchClientSpy.search).not.toHaveBeenCalled();
+
+    search.start();
+
+    expect(search.helper.state.query).toBe('');
+    expect(searchClientSpy.search).toHaveBeenCalledTimes(1);
```
Haroenv: also useful to assert with what it's been called?

src/lib/InstantSearch.js
francoischalifour: It would indeed be cleaner. I wanted to change the fewest possible number of lines since this is meant to be thrown away soon.

Is it still worth it?

src/lib/__tests__/search-client-test.js
francoischalifour: For your information this test was taken from: https://github.com/algolia/instantsearch.js/blob/develop/src/lib/__tests__/InstantSearch-test-2.js#L11

It's called with this:

```json
[
  {
    "indexName": "",
    "params": { "facets": [], "page": 0, "query": "test", "tagFilters": "" }
  }
]
```

Is it worth creating a snapshot for this?

Haroenv: I'd go with a regular test here rather than a snapshot one, but not sure if it needs to be tested if already asserted enough in other spots

src/lib/InstantSearch.js
Haroenv: Maybe leave the code as-is but add a comment that this is the first argument, which will be removed? Not totally sure what to do here

src/lib/__tests__/THROWAWAY-instanciation-test.js
```diff
@@ -0,0 +1,103 @@
...
+        });
+      }).not.toThrow();
+    });
+  });
+
+  describe('with search client', () => {
+    it('and nothing else', () => {
+      expect(() => {
+        new InstantSearch({
+          searchClient: {},
```
bobylito: Miss indexName

src/lib/__tests__/THROWAWAY-instanciation-test.js
```diff
@@ -0,0 +1,103 @@
...
+      expect(() => {
+        new InstantSearch({
+          appId,
+          apiKey,
+          searchClient: {},
+        });
+      }).toThrow();
+    });
+
+    it('and indexname, appId, apiKey', () => {
```
bobylito: Maybe not necessary.

src/lib/__tests__/search-client-test.js
bobylito: Good call on testing the search() parameters here 👍  Considering the black box nature of the parameters, I'd go for a snapshot. Either would work (it's just more maintainance to not use snapshots)

src/lib/__tests__/search-client-test.js
```diff
@@ -0,0 +1,44 @@
...
+      searchClient: searchClientSpy,
+    });
+
+    expect(searchFunctionSpy).not.toHaveBeenCalled();
+    expect(searchClientSpy.search).not.toHaveBeenCalled();
+
+    search.start();
+
+    expect(searchFunctionSpy).toHaveBeenCalledTimes(1);
+    expect(search.helper.state.query).toBe('test');
```
bobylito: I'd use the same strategy as for the previous test

src/lib/InstantSearch.js
```diff
@@ -41,33 +77,53 @@ const defaultCreateAlgoliaClient = (defaultAlgoliasearch, appId, apiKey) =>
...
+
+    if (searchClient) {
+      if (typeof searchClient.search !== 'function') {
+        throw new Error(
+          'InstantSearch configuration error: `searchClient` must implement a `search(requests)` method.'
+        );
+      }
+
+      const client = {
+        __proto__: Object.getPrototypeOf(searchClient),
```
Haroenv: I don't think that this is standardised properly.

What about doing instead:

```js
if (typeof searchClient.clearCache === 'function') {
  searchClient.clearCache = function() {};
}

if (typeof searchClient.addAlgoliaAgent === 'function') {
  searchClient.addAlgoliaAgent = function() {};
}
```

_or_ doing this check in the places in InstantSearch those two functions are used?

francoischalifour: I like your second solution, thanks! I needed `__proto__` because `algoliasearch` relies on it. We come back to the [initial solution](https://github.com/algolia/instantsearch.js/pull/2894/commits/b085d776dc41b4a2cc6c5f828a9467aeb89a58eb#diff-bde03140d12ec15ad28983aea70e6f69R102) but with `Object.create()`.

vvo: > I like your second solution, thanks! I needed __proto__ because algoliasearch relies on it.

Quick question, since this is a custom search client we have here, why do we need to have __proto or Object.create(). Which part of the code will expect this object to be made this way?

src/lib/InstantSearch.js
```diff
@@ -112,6 +168,13 @@ Usage: instantsearch({
         ...ROUTING_DEFAULT_OPTIONS,
         ...routing,
       };
+
+    if (options.createAlgoliaClient) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        'InstantSearch.js: `createAlgoliaClient` option is deprecated and will be removed in the next major version. Please use `searchClient` instead.'
```
vvo: Can you add a link to the documentation that explains how to do this switch? Always good to have more than just an error message, but an actual actionable link

src/lib/InstantSearch.js
vvo: What if you just tested if methods you need are existing and if that's the case then do not even care about creating a new object (most probably this is already a good client, ours). And only create a new object if those methods are missing

src/lib/InstantSearch.js
francoischalifour: Yes, I'm working on a guide "Preparing for v3" that I will link in this warning message.

src/lib/InstantSearch.js
Haroenv: Yes, my preferred option would be to check if `addAlgoliaAgent` exists when we are using it, rather than this workaround of making it always callable

francoischalifour: We could have done this:

```
const client = searchClient;
```

But this kind of assignation is a reference in JavaScript. That means that if ever we do `client.hello = true`, it will mutate `searchClient`. I think we shouldn't risk it. `Object.create()` creates a deep clone of the object.

We could also have done this:

```
const client = { ...searchClient };
```

But this doesn't copy the prototype, on which `algoliasearch` relies (it is not a simple object). We'd lose the reference to `client.search()` (becoming `undefined`).

Therefore, I think that `Object.create()` is the most appropriate thing to do anyway.

Does that answer your question? Do you agree with this decision?

vvo: > Do you agree with this decision?
Yes and thanks for the explanations, it might be good to add a small comment saying "We don't want to use the same reference so we clone the client no matter if it already have the necessary methods"

src/lib/main.js
```diff
@@ -89,10 +95,10 @@ import * as stateMappings from './stateMappings/index.js';
...
  * helper as a parameter. It's your responsibility to call helper.search(). This option allows you to avoid doing
  * searches at page load for example.
- * @property  {function} [createAlgoliaClient] Allows you to provide your own algolia client instead of
- * the one instantiated internally by instantsearch.js. Useful in situations where you need
- * to setup complex mechanism on the client or if you need to share it easily.
- * Usage:
+ * @property  {function} [createAlgoliaClient] _Deprecated in favor of [`searchClient`](instantsearch.html#struct-InstantSearchOptions-searchClient)._<br>
+ * Allows you to provide your own algolia client instead of the one instantiated internally by instantsearch.js.
+ * Useful in situations where you need to setup complex mechanism on the client or if you need to share it easily.
+ * <br>Usage:
```
Haroenv: can you also leave a newline instead of br?

francoischalifour: Yes, didn't realize we could.

src/lib/main.js
```diff
@@ -110,6 +116,37 @@ import * as stateMappings from './stateMappings/index.js';
  * @property {number} [stalledSearchDelay=200] Time before a search is considered stalled.
  * @property {RoutingOptions} [routing] the router configuration used to save the UI State into the URL or
  * any client side persistence.
+ * @property {SearchClient} [searchClient] The search client to plug to instantsearch.js. You should start updating with this
+ * syntax to ease the migration to InstantSearch 3.
```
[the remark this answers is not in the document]
francoischalifour: I'm not sure if the link to the guide is necessary here after all.

francoischalifour:
Waiting for #2903 to get merged to remove these lines: https://github.com/algolia/instantsearch.js/blob/de182ab7386bb442851bc2540fa125b52b68efae/src/lib/InstantSearch.js#L115-L116

src/lib/__tests__/search-client-test.js
```diff
@@ -0,0 +1,72 @@
...
+      expect(() => {
+        // eslint-disable-next-line no-new
+        new InstantSearch({
+          indexName: '',
+          searchClient: {},
+        });
+      }).toThrowErrorMatchingSnapshot();
+    });
+
+    it('should have default `addAlgoliaAgent()` and `clearCache()` methods', () => {
```
francoischalifour: We'll get rid of this test once #2903 is merged.

Haroenv:
@francoischalifour, in that case, is the `object.create` even still necessary?

PR #2899, opened 2018-04-19
docs(api): Rewrite some content on the InstantSearch API page

francoischalifour:
I fixed some inconsistencies I went through when reading the [InstantSearch API](https://community.algolia.com/instantsearch.js/v2/instantsearch.html) page.

* Deprecate `urlSync`
* Add a CSS rule to avoid empty `code` tags to display ([see "widget" here](https://community.algolia.com/instantsearch.js/v2/instantsearch.html#struct-addWidget)).

### DIFF
--- a/docgen/layouts/instantsearch.pug
+++ b/docgen/layouts/instantsearch.pug
@@ -39,7 +39,7 @@ block content
   p
     | Triggered when the rendering of all the widgets is done. This
     | happens after a search result comes back from Algolia - which means that it is
-    | triggered for the first once everything after all the widgets went through all
+    | triggered for the first time once everything after all the widgets went through all
     | there lifecycle steps once (getConfiguration, init, render).
   h3 error
   p
--- a/docgen/layouts/instantsearch.pug
+++ b/docgen/layouts/instantsearch.pug
@@ -37,10 +37,10 @@ block content
   p InstantSearch is an EventEmitter and as such it emits events on specific parts of the lifecycle.
   h3 render
   p
-    | The `render` event is triggered when the rendering of all the widgets is done. This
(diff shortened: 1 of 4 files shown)


docgen/layouts/instantsearch.pug
```diff
@@ -37,10 +37,10 @@ block content
   p InstantSearch is an EventEmitter and as such it emits events on specific parts of the lifecycle.
   h3 render
   p
-    | The `render` event is triggered when the rendering of all the widgets is done. This
+    | Triggered when the rendering of all the widgets is done. This
     | happens after a search result comes back from Algolia - which means that it is
     | triggered for the first once everything after all the widgets went through all
```
francoischalifour: I don't understand this line.

src/lib/main.js
```diff
@@ -101,22 +101,23 @@ import * as stateMappings from './stateMappings/index.js';
  *   }
  * });
  * ```
- * We forward `algoliasearch` which is the original algoliasearch module imported inside instantsearch.js
+ * We forward `algoliasearch`, which is the original Algolia search module imported inside InstantSearch.js
```
Haroenv: was doubting about this change, but I think it's good

docgen/layouts/instantsearch.pug
bobylito: for the first time...

src/lib/main.js
bobylito: algoliasearch refers directly to the npm module. Maybe this could be improved like so: 

```md
which is the original [Algolia search client](https://www.algolia.com/doc/api-client/javascript/getting-started/) imported inside InstantSearch.js
```

PR #2900, opened 2018-04-19
 fix(RangeSlider): go back to original bounds in case of an overflow

### LINKED ISSUE #2404: RangeSlider lost informations about the search
If you set some values on the `RangeSlider` and perform a search with no results the slider will display wrong value.

For example you can set the `RangeSlider` at a minimum of `1500` then trigger a search with the characters `IK`. The results are the product of the query `IK` with a price greater than `1500`. But the slider displays a wrong state with a range from `12` to `180`.

https://jsfiddle.net/tksyn43f/1/

rezak-otmani:
resolves  #2404

Modify the values such that  the minimum bound receives `minRange` and the maximum bound receives `maxRange`, so that bounds return to the initial values of the range in case of overflow

**Summary**

**Result**


Haroenv:
Thanks for contributing!

bobylito:
If I use the scenario of the #2404 in https://deploy-preview-2900--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=RangeSlider.default
 - initial search: no bounds
 - set min bound to 1500 (around): min bound = 1500
 - type I: query = I, min bound = 1500
 - type K: query = IK, min bound = 180
 - remove one character: query = I, min bound = 180

We can see that this does not solve #2404

rezak-otmani:
I think with this scenario we get
* initial search: no bounds
* set min bound to 1500 (around): min bound = 1500
* type I: query = I, min bound = 1500
* type K: query = IK, min bound = 12
* remove one character: query = I, min bound = 1

it's not what expected

francoischalifour:
Thanks for the initiative @rezak-otmani. We'll need more time and research to know what exact behavior we want to achieve regarding this issue.

PR #2903, opened 2018-04-20
feat: Update client and helper

francoischalifour:
This PR updates to `algoliasearch@3.27.0` and `algoliasearch-helper@2.25.1`, using the promisified version of `client.search()`.

This update anticipates the integration of the next minor version of the helper making `addAlgoliaAgent()` and `clearCache()` optional (see algolia/algoliasearch-helper-js#577).

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
   },
   "dependencies": {
     "algoliasearch": "3.27.0",
-    "algoliasearch-helper": "2.25.1",
+    "algoliasearch-helper": "2.26.0",
     "classnames": "2.2.5",
     "events": "1.1.0",
     "hogan.js": "3.0.2",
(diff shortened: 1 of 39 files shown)


package.json
```diff
@@ -130,8 +130,8 @@
     "webpack-dev-server": "2.11.2"
   },
   "dependencies": {
-    "algoliasearch": "3.25.1",
-    "algoliasearch-helper": "2.24.0",
+    "algoliasearch": "3.27.0",
+    "algoliasearch-helper": "2.25.1",
```
Haroenv: I wonder why we pinned dependencies to patch versions here 🤔 I know we have renovate, but it can cause duplication if an enduser has those in their dependencies too with another version specified

samouss: We did this change before we all agree that we should go for pinned dependencies only for the dev one. We should revert the production dependencies to caret range. We can do it in an another PR.

PR #2904, opened 2018-04-20
fix(Pit): round the slider pit value

### LINKED ISSUE #2828: Add an option for control which Pit value is displayed in the RangeSlider

Feature.

Right now we display the values at 0, 50 and 100 percent. In some cases we might want to display more / less values in the pits. We should be able to control where & when the pits are append.

**Without control (current)**

screen shot 2018-03-19 at 16 30 13

**With control (wanted)**

screen shot 2018-03-19 at 16 30 22

Right now the generation is computed from an array `[0, 50, 100]`. Maybe we should expose this an option or a function that return an array like this one.

-----

A bit linked to https://github.com/algolia/instantsearch.js/issues/2755

rezak-otmani:
resolves #2828

always rounding the middle value to a whole number

**Summary**

**Result**


rezak-otmani:
Hi @Haroenv,  it's better I think

bobylito:
Actually the problems lies in prettier (it's actually the check that fails) and parenthesis seems very troublesome as a problem (https://github.com/prettier/prettier/issues/187).

@rezak-otmani can you find a way to rewrite that code so that it doesn't rely on parenthesis (like introducing an intermediate variable that holds the result of parseFloat)?

francoischalifour:
Prettier 1.13 might fix the calculation issue: [Insert more parentheses in math expressions](https://prettier.io/blog/2018/05/27/1.13.0.html#insert-more-parentheses-in-math-expressions-4413-and-4407-by-duailibe).

src/components/Slider/Pit.js
```diff
@@ -11,7 +11,7 @@ const Pit = ({ style, children }) => {
   // Children could be an array, unwrap the value if it's the case
   // see: https://github.com/developit/preact-compat/issues/436
   const value = Array.isArray(children) ? children[0] : children;
-  const pitValue = Math.round(parseFloat(value) * 100) / 100;
+  const pitValue = Math.round(parseInt(value) * 100) / 100;
```
francoischalifour:
```suggestion
  const pitValue = Math.round(parseInt(value, 10) * 100) / 100;
```

francoischalifour:
I changed the base branch so that we can apply changes to your fix. Thank you @rezak-otmani!

PR #2905, opened 2018-04-23
docs(guide): Add guide "Prepare for v3"

francoischalifour:
This PR adds the guide "[Prepare for v3](https://deploy-preview-2905--algolia-instantsearch.netlify.com/v2/guides/prepare-for-v3.html)" to ease the migration towards InstantSearch 3 with search clients.

### DIFF
--- a/docgen/src/guides/prepare-for-v3.md
+++ b/docgen/src/guides/prepare-for-v3.md
@@ -44,3 +44,7 @@ const search = instantsearch({
 
 search.start();
 ```
+
+## Deprecations
+
+* [`createAlgoliaClient`](https://community.algolia.com/instantsearch.js/v2/instantsearch.html#struct-InstantSearchOptions-createAlgoliaClient) becomes deprecated in favor of [`searchClient`](https://community.algolia.com/instantsearch.js/v2/instantsearch.html#struct-InstantSearchOptions-searchClient)
--- a/docgen/src/guides/prepare-for-v3.md
+++ b/docgen/src/guides/prepare-for-v3.md
@@ -9,9 +9,7 @@ editable: true
 githubSource: docgen/src/guides/prepare-for-v3.md
 ---
 
-> Starting with 2.8.0, we are introducing changes and deprecations to prepare for the upcoming InstantSearch.js 3.
-
-InstantSearch has always worked with the [Algolia search client](https://github.com/algolia/algoliasearch-client-javascript) behind the scenes. This will no longer be true for InstantSearch.js 3. To ease the migration to version 3, we recommend applying the changes from this guide now.
+Starting with 2.8.0, we are introducing changes and deprecations to prepare for the upcoming InstantSearch.js 3. InstantSearch has always worked with the [Algolia search client](https://github.com/algolia/algoliasearch-client-javascript) behind the scenes. This will no longer be the default for InstantSearch.js 3. To ease that migration, we recommend applying the changes from this guide now.
 
 ## Initializing InstantSearch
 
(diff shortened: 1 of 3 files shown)


docgen/src/guides/prepare-for-v3.md
```diff
@@ -0,0 +1,42 @@
...
+mainTitle: Guides
+layout: main.pug
+category: guides
+withHeadings: true
+navWeight: 0
+editable: true
+githubSource: docgen/src/guides/prepare-for-v3.md
+---
+
+InstantSearch has always worked with the [Algolia search client](https://github.com/algolia/algoliasearch-client-javascript) behind the scenes. This will no longer be true for InstantSearch 3. To ease the migration from 2.8.x to 3.x.x, we recommend applying the changes from this guide now.
```
bobylito: This will be only a part of this guide, we should probably add a title here like: `Initializing InstantSearch`

bobylito: > To ease the migration from 2.8.x to 3.x.x, we recommend applying the changes from this guide now.

This could be used as a general introduction to the guide.

Also mention 2.x, instead of 2.8.x

Haroenv: `version 2 to version 3` is another, simpler option

francoischalifour: If a user from `<= 2.8.0` follows this guide, it will break though (as `searchClient` won't be available before `2.8.0`). Should we ignore that?

Haroenv: maybe a different sentence in the beginning saying that the guide will only work from 2.8 onwards

bobylito: Actually that's pretty good as an intro for the whole guide:

> Starting with 2.8, we are introducing changes and deprecations to prepare for the upcoming InstantSearch.js 3

WDYT?

francoischalifour: Do you mean instead of "InstantSearch has always worked [...] from this guide now." or before the current intro?

Haroenv: > Starting with 2.8, we are introducing changes and deprecations to prepare for the upcoming InstantSearch.js 3. InstantSearch has always worked with the [Algolia search client](https://github.com/algolia/algoliasearch-client-javascript) behind the scenes. This will no longer be true for InstantSearch 3. To ease that migration, we recommend applying the changes from this guide now.

francoischalifour: Yep, that's what I've got right now. Wanted confirmation, thank you!

docgen/src/guides/prepare-for-v3.md
```diff
@@ -0,0 +1,48 @@
...
+mainTitle: Guides
+layout: main.pug
+category: guides
+withHeadings: true
+navWeight: 0
+editable: true
+githubSource: docgen/src/guides/prepare-for-v3.md
+---
+
+Starting with 2.8.0, we are introducing changes and deprecations to prepare for the upcoming InstantSearch.js 3. InstantSearch has always worked with the [Algolia search client](https://github.com/algolia/algoliasearch-client-javascript) behind the scenes. This will no longer be the default for InstantSearch.js 3. To ease that migration, we recommend applying the changes from this guide now.
```
bobylito: I think we can soften the recommendation :)

> To prepare the future migration to InstantSearch.js v3, you can start using the new options now.

I would still talk about the client only in the introduction of "Initializing InstantSearch".

Haroenv: the upcoming release of

francoischalifour: I get it now!

PR #2906, opened 2018-04-23
fix RangeSlider  alignements

### LINKED ISSUE #2449: Style / rangeslider: Problems with the alignements

rezak-otmani:
resolves #2449

center the tooltip and value  in the handle

**Summary**

**Result**


francoischalifour:
Closing since the styles now depend on [InstantSearch.css](https://yarnpkg.com/en/package/instantsearch.css).

Thanks!

PR #2907, opened 2018-04-23
fix(priceRanges) : convert the entered digits to integers

### LINKED ISSUE #2858: priceRanges entered as floats being truncated to integers

Bug

In instantsearch.js 1.11.7, the priceRanges and priceRange widgets truncate manually entered values to integers.  This makes it hard to search for small values in ranges like 0.00 to 0.50.

instantsearch.js accepts manually entered numbers as integers or floats.

All.

I don't have a patch or test case for you.  You can use the instantsearch.js example on this page to see the problem:

https://community.algolia.com/instantsearch.js/v1/documentation/#priceranges

1.11.7

rezak-otmani:
resolves #2858"

convert the entered digits to integers,  applying the absolute value

**Summary**

**Result**


francoischalifour:
Thanks for your contribution! The `priceRanges` widget is not supported anymore in the new InstantSearch versions, so I'll close this PR.

PR #2912, opened 2018-05-03
chore(issues): Update issue templates

bobylito:


bobylito:
I'm not it won't collide with the previous template 🤔

Haroenv:
Why is the previous template not removed? 

Maybe this can be tried out in a test repo so we see how it behaves?

bobylito:
Or we can try and fix it :)

bobylito:
I will remove the old template :) @Haroenv No need for it actually

bobylito:
@Haroenv Is there anything else I should do?

PR #2914, opened 2018-05-03
New version: 2.8

bobylito:
This version comes with a new argument that will replace the createAlgoliaClient: searchClient. It serves as a standardized way for providing the connectivity to the search engine.

It also comes with a new connector for building autocomplete menus!


bobylito:
Merging that since everything was already reviewed and parts of it were even in beta.

PR #2919, opened 2018-05-11
docs(GeoSearch): add height requirement

samouss:


samouss:
On the element that contains the map, in our case `.ais-geo-search--map`.

PR #2935, opened 2018-05-18
fix(clear-all): apply excludeAttribute correctly with clearsQuery

bobylito:
## Initial bug

When clearsQuery and excludeAttributes are used together, we can clear the excluded attributes when the query is not empty

## Reproducing the issue with the current InstantSearch

http://extra-small-group.surge.sh

 - select a brand (note that the clearall button does not activate)
 - add some text in the search box (clear all button is activable)
 - click clear all
 - the query AND the brand are cleared

brand should not be cleared.

## Notes 

The problem is that fixing only this initial bug had an impact on `currentRefinedValues`. Both widget use the same utility function to provide the clear all feature but they have two opposite list system for excluding attributes: one is a white list, and the other a black list.

This patch refactors the utilities, baking the support of the two types of list in a single place. 

A next step for this could be to use a single connnector for current refined values and clear all widgets.

### DIFF
--- a/src/connectors/clear-all/__tests__/connectClearAll-test.js
+++ b/src/connectors/clear-all/__tests__/connectClearAll-test.js
@@ -1,5 +1,3 @@
-import sinon from 'sinon';
-
 import jsHelper from 'algoliasearch-helper';
 const SearchResults = jsHelper.SearchResults;
 
@@ -8,10 +6,10 @@ import connectClearAll from '../connectClearAll.js';
 describe('connectClearAll', () => {
   it('Renders during init and render', () => {
     const helper = jsHelper({});
-    helper.search = sinon.stub();
+    helper.search = () => {};
     // test that the dummyRendering is called with the isFirstRendering
     // flag set accordingly
-    const rendering = sinon.stub();
+    const rendering = jest.fn();
     const makeWidget = connectClearAll(rendering);
     const widget = makeWidget({
       foo: 'bar', // dummy param to test `widgetParams`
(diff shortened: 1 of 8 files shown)


bobylito:
I could add some stories :)

src/lib/utils.js
```diff
@@ -301,44 +300,64 @@ function getRefinements(results, state, clearsQuery) {
...
-    state = state.clearTags();
-    state = state.clearRefinements();
-    return state;
-  }
+  let finalState = helper.state;
 
-  forEach(attributeNames, attributeName => {
-    if (attributeName === '_tags') {
-      state = state.clearTags();
+  forEach(attributesToClear, attribute => {
```
samouss: We can use the regular `forEach` instead of the lodash version.

src/connectors/clear-all/__tests__/connectClearAll-test.js
```diff
@@ -149,7 +149,7 @@ describe('connectClearAll', () => {
     expect(helper.state.query).toBe('');
```
samouss: We can move this file to Jest.

bobylito: Yes.

vvo:
Thanks for detailed explanation, I read it, tried it, read the new tests that are checking the right behaviour, all good for me

bobylito:
> Thanks for detailed explanation, I read it, tried it, read the new tests that are checking the right behaviour, all good for me

Thanks @vvo for the review.

Points of @samouss are still very valid. I will take this PR as an opportunity for improvement.

bobylito:
Code has been updated @samouss

PR #2942, opened 2018-05-24
docs(routing): Replace urlSync with routing

pixelastic:
The current getting started example will throw a warning about `urlSync` being deprecated in favor of `routing`. This will update the doc to use the `routing` in the examples instead.


bobylito:
Nice catch :)

PR #2943, opened 2018-05-28
 docs(guide): Add "Use your own backend"

francoischalifour:
This PR adds the guide for using a custom backend with InstantSearch.js.

It will get replicated for all flavors. The only part that will change is the code snippet in "Using the search client with InstantSearch".

Note that I named the file `proxy-backend.md`, which is going to be the URL to the guide. Let me know if you think of a better name.

Before merging it'll be great to find somewhere to host the Search Client Specification as I refer to it.

### DIFF
--- a/docgen/src/data/communityHeader.json
+++ b/docgen/src/data/communityHeader.json
@@ -91,6 +91,10 @@
       {
         "name": "Integrate with Angular (2+)",
         "url": "guides/angular-integration.html"
+      },
+      {
+        "name": "Use your own backend",
+        "url": "guides/proxy-backend.html"
       }
     ]
   },
@@ -143,6 +147,10 @@
   {
     "name": "Migrate from V1",
     "url": "guides/migration.html"
+  },
+  {
+    "name": "Use your own backend",
+    "url": "guides/proxy-backend.html"
   },
   {
     "name": "instantsearch()",
--- a/docgen/src/guides/proxy-backend.md
+++ b/docgen/src/guides/proxy-backend.md
@@ -1,0 +1,170 @@
+---
+title: Use your own backend
+mainTitle: Guides
+layout: main.pug
+category: guides
+withHeadings: true
+navWeight: 21
+editable: true
+githubSource: docgen/src/guides/proxy-backend.md
+---
+
+> This guide is compatible with InstantSearch.js ≥ 2.8.0.
+
+Advanced InstantSearch users may have the need to query Algolia’s servers from their backend instead of the frontend, while still being able to reuse InstantSearch widgets. Possible motivations could be for security restrictions, for SEO purposes or to enrich the data sent by the custom server (i.e. fetch Algolia data and data from their own servers). If this sounds appealing to you, feel free to follow this guide. Keep in mind though that we, at Algolia, [recommend doing frontend search](https://www.algolia.com/doc/faq/searching/searching-from-the-front-end-or-the-back-end/#we-definitely-recommend-frontend-search) for performance and high availability reasons.
+
+By the end of this guide, you will have learned how to leverage InstantSearch with your own backend architecture to query Algolia.
+
+## How it’s made possible
+
+InstantSearch is the UI part on top a search client with a state managed by a [JavaScript Helper](https://github.com/algolia/algoliasearch-helper-js). These three layers are composable and can be interchanged to leverage the InstantSearch widgets system with different search clients.
+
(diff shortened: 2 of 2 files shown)


francoischalifour:
I mention that it's a proxy server 3 times, although I never make it really explicit.

Happy to go with `custom-backend.md` too. The downside is that I had in mind to write another guide named `custom-backend` next quarter about using the full potential of the `searchClient` property fetching results without Algolia. I'm not sure it's legit. If not, I can rename this one.

Haroenv:
I'd call that other guide something like `different-backend`

francoischalifour:
You're all right. The mobile team also named it `custom-backend` (see algolia/instantsearch-ios#31). I renamed it.

docgen/src/guides/custom-backend.md
```diff
@@ -0,0 +1,170 @@
...
+      },
+      body: JSON.stringify(requests),
+    }).then(res => res.json());
+  }
+};
+```
+
+We use the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) to query the server; make sure to take note of the [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#Browser_compatibility) before using it in production.
+
+> If you want to transform the data to be passed to your server, you can learn more about all the parameters that the `search()` method supports in the Search Client Specification.
```
Haroenv: if this spec doesn't exist yet, What about "[Algolia documentation](https://www.algolia.com/doc/api-reference/api-parameters/?language=javascript#parameters-list)"

docgen/src/guides/custom-backend.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Searching for facet values
+
+If your server exposes a *Search For Facet Values* endpoint, you can implement this feature in a similar way as the `search()` method with the `searchForFacetValues()` method.
+
+```javascript
```
Haroenv: wonder if the code highlighting is better if you write `typescript`

vvo:
> I mention that it's a proxy server 3 times, although I never make it really explicit.

Just saw that, indeed. I think we found a common ground now, thanks!

docgen/src/guides/custom-backend.md
```diff
@@ -0,0 +1,170 @@
...
+app.post('/search', async (req, res) => {
+  const requests = req.body;
+  const algoliaResults = await algoliaClient.search(requests);
+  const results = {
+    ...algoliaResults,
+    results: algoliaResults.results.map(result => ({
+      ...result,
+      hits: result.hits.map(hit => ({
+        ...hit,
+        downloadUrl: `https://my-website.com/files/${hit.objectID}.zip`,
```
Haroenv: This example could easily be done frontend too, but there's not really a unified way of requesting data from a db or whatever. If you would have that you could do something like: 

```js
result.hits.map(async hit => ({
  ...hit,
  user: await getUser(hit.userID),
});
```

but that requires other things to be set up correctly.

docgen/src/guides/custom-backend.md
francoischalifour: It doesn't seem to be recognized on the documentation website.

docgen/src/guides/custom-backend.md
francoischalifour: Indeed, it would make more sense to fetch data even if the given example is not exhaustive.

We can assume that the `getUser()` functions already exists or create a dummy implementation with something like [Chance.js](https://github.com/chancejs/chancejs). It's more code to digest though.

docgen/src/guides/custom-backend.md
```diff
@@ -0,0 +1,170 @@
...
+The `algoliasearch` package will allow you to query Algolia from your backend. Here's an example using [Express](https://expressjs.com/):
+
+```javascript
+// Instantiate an Algolia client
+const algoliasearch = require('algoliasearch');
+const algoliaClient = algoliasearch('YOUR_APP_ID', 'YOUR_SEARCH_API_KEY');
+
+// Add the search endpoint
+app.post('/search', async (req, res) => {
+  const requests = req.body;
```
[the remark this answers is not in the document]
bobylito: What would be a good alternative @Haroenv ? `queries`? `algoliaQueries`?

docgen/src/guides/custom-backend.md
bobylito: Without assuming anything, you could have a comment that describes what getUser is, just on top of the line.

francoischalifour:
Changes have been made!

Waiting for this PR to be merged before sending PRs for other flavors.

PR #2944, opened 2018-05-28
Documentation: RangeInputLabels button was renamed to submit

michaelscheel:
**Summary**

Documentation was outdated. Parameter button is now named submit.


Haroenv:
Great catch, seems like this is also the case in: 

https://github.com/michaelscheel/instantsearch.js/blob/3f0420d8a57b72bc570c95f03d829446300d859e/src/widgets/range-input/range-input.js#L73

Could that be updated too?

bobylito:
Thanks for your contribution @michaelscheel 🙇. I am integrating your fix in a bigger PR about doc :) Closing this :)

PR #2946, opened 2018-05-28
chore(release): fix release script

bobylito:
NPM doesn't seem to responde to `npm owner add`, this implements
an alternative way to check if the currently logged NPM user is in
the list of contributors on NPM.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -19,11 +19,13 @@ const {
 } = require('./conventionalChangelog.js');
 
 // check if user can publish new version to npm
-const { code: isNotOwner } = shell.exec('$(npm owner add `npm whoami`)', {
-  silent: true,
-});
+const ownersFound = parseFloat(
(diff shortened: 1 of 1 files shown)


vvo:
This seems to be good:

image

vvo:
Unsure this is still used but https://github.com/algolia/instantsearch.js/blob/1bbd6f903aeffca9290fb773ff689c19c45ba74d/scripts/release-maintenance.sh would need an update too

bobylito:
> Unsure this is still used but https://github.com/algolia/instantsearch.js/blob/1bbd6f903aeffca9290fb773ff689c19c45ba74d/scripts/release-maintenance.sh would need an update too

Yes indeed. Will do another PR.

bobylito:
Thanks for the review @vvo

bobylito:
It will match sub part of the name :/ like `boby` in `bobylito`...

PR #2948, opened 2018-05-29
fix(infinite-hits): throw with allItems

### LINKED ISSUE #2543: allItems template shouldn't be available in infiniteHits
Hits allows for the allItems template which allows for greater customization of the templates, for example inside tables and so forth...

When the allItems template is used with infiniteHits, its doesn't give any error, but instead of adding the newly loaded items, it replaces everything with the new page of results.

bobylito:
The infinite hits widget is based internally on the hits widget which
accepts the `allItems` template. However this template is not usable for
the purpose of the infinite hits widget.

In order to prevent users from using this buggy and yet accepted option
this PR generates a exception if we detect this particular template.

Fix #2543

### DIFF
--- a/src/widgets/infinite-hits/__tests__/infinite-hits-test.js
+++ b/src/widgets/infinite-hits/__tests__/infinite-hits-test.js
@@ -76,9 +76,17 @@ describe('infiniteHits()', () => {
     expect(ReactDOM.render.secondCall.args[1]).toEqual(container);
   });
 
-  it('does not accept both item and allItems templates', () => {
-    expect(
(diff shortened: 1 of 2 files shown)


bobylito:
> Would prefer it to be supported instead, but this is a valid problem-stopper

If you want I'm ok broadening the discussion to the team :)

Haroenv:
Can be discussed when we do the template / props review

bobylito:
> Can be discussed when we do the template / props review

Ok then let's merge for now 👍 Also the allItems can be done using the connector, no?

PR #2961, opened 2018-06-01
feat(panel): headerFooter -> panel / is.css

bobylito:
We're gonna change this PR to implement the new panel-as-a-widget approach (see https://github.com/algolia/instantsearch-rfcs/pull/1)

Sub-tasks:
 - [ ] implement panel widget
 - [ ] remove all header / footer / autohide options from widgets
 - [ ] implement a canRefine method on widgets


bobylito:
This is probably not even the right way to start with...

PR #2962, opened 2018-06-01
fix(connectRefinementList): throw error with usage

samouss:
**Summary**

We correctly check display the usage in case of incorrect rendering but not when the provided parameters are wrong. The message is not correctly pass to the function that throw the error.

**Before**

screen shot 2018-06-01 at 15 57 37

**After**

screen shot 2018-06-01 at 16 10 07


samouss:
cc @bobylito

PR #3006, opened 2018-06-19
fix(connectHitsPerPage): default value should not break the API 

### LINKED ISSUE #2732: HitsPerPageSelector, when selecting the empty value, it sends `undefined` to Algolia
**bug**

When using HitsPerPageSelector with no default value, an empty entry is generated. When selecting any non empty value, and then selecting a value, we get an error from the API: 

screen shot 2018-02-12 at 09 32 52

Reproducible here: https://codesandbox.io/s/8qo2m8n89

**Correct behavior**

When selecting back the empty value, it should remove the hitsPerPage from the searchParameters.

**Fix**

Make `refine` accept `"undefined"` or `""` on top of the currently handled `undefined` (in the UI, we'll have strings, it's better if we don't have to convert). When the function receives those values, it should unset the parameter.

bobylito:
Fix #2732 

Reproducible here: https://codesandbox.io/s/8qo2m8n89
With the fix: http://splendid-discussion.surge.sh/

Test case:
 - select 3 hits per page
 - select the empty entry in the menu
 - use the searchbox: the search is broken

## Limitations

What you can notice is that the empty value doesn't fall back to the search parameters value. This is because the correct way to set a default value is to use the `default: true` option. We should maybe not support the empty choice (to be discussed)

### DIFF
--- a/src/connectors/hits-per-page/connectHitsPerPage.js
+++ b/src/connectors/hits-per-page/connectHitsPerPage.js
@@ -157,11 +157,11 @@ The first one will be picked, you should probably set only one default value`
             );
           }
 
-          items = [{ value: undefined, label: '' }, ...items];
+          items = [{ value: '', label: '' }, ...items];
         }
 
         this.setHitsPerPage = value =>
-          value === 'undefined' || value === ''
+          !value && value !== 0
             ? helper.setQueryParameter('hitsPerPage', undefined).search()
             : helper.setQueryParameter('hitsPerPage', value).search();
 
--- a/src/connectors/hits-per-page/connectHitsPerPage.js
+++ b/src/connectors/hits-per-page/connectHitsPerPage.js
@@ -161,7 +161,9 @@ The first one will be picked, you should probably set only one default value`
         }
 
         this.setHitsPerPage = value =>
-          helper.setQueryParameter('hitsPerPage', value).search();
+          value === 'undefined' || value === ''
+            ? helper.setQueryParameter('hitsPerPage', undefined).search()
+            : helper.setQueryParameter('hitsPerPage', value).search();
 
         renderFn(
           {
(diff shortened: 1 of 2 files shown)


src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -161,7 +161,9 @@ The first one will be picked, you should probably set only one default value`
         }
 
         this.setHitsPerPage = value =>
-          helper.setQueryParameter('hitsPerPage', value).search();
+          value === 'undefined' || value === ''
```
samouss: I don't think the `value` could be equal to `'undefined'`. Maybe use `typeof value === 'undefined'`. But IMO we can use the falsy value of `value` for this condition. Every time a falsy value is provided we reset the `hitsPerPage`.

bobylito: As of now, yes it is equal to `'undefined'` because the value for the empty entry is `undefined` and it is used directly in the DOM and stringified in the process.

I can see to change the value for the empty entry.

samouss: Okay my bad though 👍

bobylito: Updated the value and I changed the value for the "not in the list" value.

bobylito:
Let me know what you think of the changes @samouss :)

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -157,11 +157,13 @@ The first one will be picked, you should probably set only one default value`
...
           }
 
-          items = [{ value: undefined, label: '' }, ...items];
+          items = [{ value: '', label: '' }, ...items];
         }
 
         this.setHitsPerPage = value =>
-          helper.setQueryParameter('hitsPerPage', value).search();
+          !value && value !== 0
+            ? helper.setQueryParameter('hitsPerPage', undefined).search()
```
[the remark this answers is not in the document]
samouss: IMO it's not less explicit but we can go merge it as it is. It's not a big deal.

PR #3017, opened 2018-06-29
feat(clearRefinements): implement is.css

bobylito:
**Summary**

This PR implements is.css for the widget clear-all, which is now renamed to clear-refinements.

This also removes the panel from the widget.

**Result**

https://deploy-preview-3017--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=ClearRefinements.default


dev/template.html
```diff
@@ -5,6 +5,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=Edge">
     <title>Instant search demo built with instantsearch.js</title>
     <script src="https://cdn.polyfill.io/v2/polyfill.min.js?features=default,Array.from"></script>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.0.0/themes/reset-min.css">
```
samouss: The theme `algolia` already contains the `reset`. We don't need to include it.

dev/webpack.config.js
```diff
@@ -13,7 +13,7 @@ const __DEV__ = process.env.NODE_ENV === 'development';
 const clean = arr => arr.filter(item => item !== false);
 
 module.exports = {
-  devtool: __DEV__ ? 'cheap-module-source-map' : 'source-map',
+  devtool: 'source-map',
```
samouss: It doesn't slow down too much the build time?

docgen/src/guides/migrate-instantsearchcss.md
```diff
@@ -17,15 +17,17 @@ in v3.
...
+### Renamed connectors
 
 ### Renamed parameters
 
-| In V2 | In V3 |
-| ----- | ----- |
-
+| In V2           | In V3                   |
+| --------------- | ----------------------- |
+| connectClearAll | connectClearRefinements |
```
samouss: It should be in the section above "Renamed connectors" rather than "Renamed parameters".

docgen/src/guides/migrate-instantsearchcss.md
```diff
@@ -72,14 +74,78 @@ The structure is now embeded in a list and each item wraps a separator and the l
...
+</div>
+```
+
+### Options
+
+Options that were renamed (same semantic):
+
+* `excludeAttributes` → `excludedAttributes`
+
+Values for `classNames` have been updated to reflect the markup changes:
```
samouss: The parameter is called `cssClasses` not `classNames` right?

src/components/ClearRefinements/__tests__/ClearRefinements-test.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import sinon from 'sinon';
```
samouss: Do we take advantage of this refactor to drop `sinon`?

src/widgets/clear-refinements/clear-refinements.js
```diff
@@ -0,0 +1,125 @@
+import React, { render, unmountComponentAtNode } from 'preact-compat';
```
samouss: We should align the casing of the files with the rewrite. Components are pascal case, connectors camel case and the widget "kebab" case. IMO it's okay to have the components with a different case for the file but we should at least align the connectors & widgets. WDYT?

src/widgets/clear-refinements/clear-refinements.js
```diff
@@ -0,0 +1,125 @@
...
+import ClearRefinementsWithHOCs from '../../components/ClearRefinements/ClearRefinements.js';
```
samouss: We can rename it, we don't have any HOC anymore.

dev/app/builtin/init-stories.js
```diff
@@ -1,6 +1,6 @@
 import initAnalyticsStories from './stories/analytics.stories';
 import initBreadcrumbStories from './stories/breadcrumb.stories.js';
-import initClearAllStories from './stories/clear-all.stories';
+import initClearRefinementsStories from './stories/clear-refinements.stories.js';
```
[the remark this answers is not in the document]
bobylito: Oh yes agreed. By default, VScode will not add it but that's not the behavior everywhere. Most of the code in is.js has `.js` extensions, but with a rule it can be `--fix`ed easily. No opinions on the topic, happy to do the changes :)

dev/template.html
bobylito: Oh good to know 👍

src/widgets/clear-refinements/clear-refinements.js
bobylito: Indeed we said that the file / folders should have the same case as the exported symbol. So it's gonna be camelCase then for widgets. Right?

samouss: Yes. There is also the `/components/ClearRefinements/ClearRefinements.js` in that case `ClearRefinements` is a bit redundant. WDYT?

bobylito: I don't know. On one hand, it's nicer, but on the other it brings little value and the test are more easy to find with a nested structure.

dev/app/builtin/init-stories.js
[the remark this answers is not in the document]
bobylito: Well `--fix` doesn't work for the rule: `import/extensions`. This makes this change very painful. Also the codebase is mostly using extension except for packages.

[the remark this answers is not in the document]
samouss: But we can enforce the extensions only for InstantSearch but at least statute on something.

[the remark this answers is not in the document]
bobylito: > But we can enforce the extensions only for InstantSearch but at least statute on something.

Ok. If I were to make a choice here, I'd go for no extensions:
 - it's easier to setup the rules in eslint
 - we are inconsistent when it comes to packages (no extensions even for files in packages)
 - vsCode (the editor that we use as far I understand) doesn't add extensions by default

It's a big change but I guess it could be done with a mass replace.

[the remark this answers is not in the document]
bobylito: https://github.com/algolia/instantsearch.js/pull/3022

bobylito:
Please check https://github.com/algolia/instantsearch.js/pull/3022/ before we merge this one.

src/widgets/clear-refinements/clear-refinements.js
bobylito: We can do that in another PR if we decide to move forward. Ok @samouss ?

dev/webpack.config.js
bobylito: Is it a blocker @samouss ?

samouss: Absolutely not!

src/widgets/clear-refinements/clear-refinements.js
samouss: Sure!

bobylito:
Thanks @samouss for taking the time to review :)

PR #3021, opened 2018-07-02
docs: IE11 support (example + mentions)

### LINKED ISSUE #3001: Fix demos for IE11
**Describe the bug 🐛**
Demos do not work on IE11 and the solution we provide is not good enough (using polyfill.io)

However a user offered us a solution: everything is detailed in this [post](https://discourse.algolia.com/t/no-search-result-with-ie11-solved/5391). 

Let's fix that.

bobylito:
FIX #3001

**Summary**

Demos are broken on IE11 when loading a URL that has priceRanges. Such as:

https://community.algolia.com/instantsearch.js/v2/examples/e-commerce/?priceRanges[price]=1%3A10

**Result**

https://deploy-preview-3021--algolia-instantsearch.netlify.com/v2/examples/e-commerce/?priceRanges%5Bprice%5D=1%3A10

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Start by [adding widgets](https://community.algolia.com/instantsearch.js/v2/widg
 
 ## Browser support
 
-We support the last **two versions of major browsers** (Chrome, Edge, Firefox, Safari).
+We support the **last two versions of major browsers** (Chrome, Edge, Firefox, Safari).
 
 To support [IE11](https://en.wikipedia.org/wiki/Internet_Explorer_11), we recommend loading [polyfill.io](https://polyfill.io) with the following URL `https://cdn.polyfill.io/v2/polyfill.min.js?features=default,Array.prototype.includes`
 
--- a/README.md
+++ b/README.md
@@ -117,7 +117,9 @@ Start by [adding widgets](https://community.algolia.com/instantsearch.js/v2/widg
 
 ## Browser support
 
-We support the last **two versions of major browsers**. To support [IE11](https://en.wikipedia.org/wiki/Internet_Explorer_11), we recommend using [polyfill.io](https://polyfill.io).
+We support the last **two versions of major browsers** (Chrome, Edge, Firefox, Safari).
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -117,7 +117,9 @@ Start by [adding widgets](https://community.algolia.com/instantsearch.js/v2/widg
 
 ## Browser support
 
-We support the last **two versions of major browsers**. To support [IE11](https://en.wikipedia.org/wiki/Internet_Explorer_11), we recommend using [polyfill.io](https://polyfill.io).
+We support the last **two versions of major browsers** (Chrome, Edge, Firefox, Safari).
```
francoischalifour: "last" should also probably be `strong` → `**last two versions of major browsers**`.

docgen/src/guides/usage.md.hbs
```diff
@@ -82,3 +82,13 @@ const search = instantsearch({ ... });
 search.addWidget(searchBox({ ... }));
 search.addWidget(connectSearchBox(function() { ... })({ ... }))
 ```
+
+## Browser support
+
+We support the last **two versions of major browsers** (Chrome, Edge, Firefox, Safari).
```
francoischalifour: "last" should also probably be `strong` → `**last two versions of major browsers**`.

README.md
```diff
@@ -117,7 +117,9 @@ Start by [adding widgets](https://community.algolia.com/instantsearch.js/v2/widg
...
+
+To support [IE11](https://en.wikipedia.org/wiki/Internet_Explorer_11), we recommend loading [polyfill.io](https://polyfill.io) with the following URL `https://cdn.polyfill.io/v2/polyfill.min.js?features=default,Array.prototype.includes`
```
samouss: For the usage of `.includes` we should make a choice. Either we consider it as available (via polyfill when it's not) then we use it everywhere and we drop the usage of `lodash/includes`. Or we don't use this feature until it's widely supported and we keep using the lodash version. WDYT?

bobylito: It's about moving towards the direction we set: removing lodash at the end. We're in an intermediary step where we have some parts that will include both, and this PR fixes the state for end-users who do not have the latest version of browsers.

bobylito: It's removed! :)

PR #3022, opened 2018-07-02
chore: unify and enforce import style to no extensions

bobylito:
Following up on [@samouss comment](https://github.com/algolia/instantsearch.js/pull/3017#discussion_r199545330), this PR activates import/extensions and configures to not accepts extensions for JS files.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,6 +3,7 @@ module.exports = {
   rules: {
     "no-param-reassign": 0,
     "import/no-extraneous-dependencies": 0,
-    "react/no-string-refs": 1
+    "react/no-string-refs": 1,
+    "import/extensions": [2, "always", {js: "never"}]
   }
 }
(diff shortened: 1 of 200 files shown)


.eslintrc.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   rules: {
     "no-param-reassign": 0,
     "import/no-extraneous-dependencies": 0,
-    "react/no-string-refs": 1
+    "react/no-string-refs": 1,
+    "import/extensions": [2, "always", {js: "never"}]
```
francoischalifour: What about having this rule part of [eslint-config-algolia](https://github.com/algolia/eslint-config-algolia)?

bobylito: It's a first step :) But sure it's gonna end up there if we're happy about it.

bobylito: https://github.com/algolia/eslint-config-algolia/pull/92 we'll remove the rule when it's merged and the config updated.

PR #3025, opened 2018-07-05
Fix link to hierarchicalMenu.js example

Spone:
**Summary**

The link was broken.

**Result**

Now it's not :tada:


samouss:
I've reword your commit to pass our commit message validation.

PR #3037, opened 2018-07-18
docs(dev-novel): add menu-select

bobylito:
This PR adds a set of stories for menu-select.


bobylito:
> Nice! What about sortBy and other options?

Nice idea but I'd rather work on something else right now.

PR #3055, opened 2018-08-08
chore(release): fix rollback command

bobylito:
Incorrect git command for resetting master during a rollback.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -79,7 +79,7 @@ const strategy = currentBranch !== 'develop' || forceBeta ? 'beta' : 'stable';
 function rollback(newVersion) {
   if (strategy === 'stable') {
     // reset master
-    shell.exec('git reset --hard origin master');
+    shell.exec('git reset --hard origin/master');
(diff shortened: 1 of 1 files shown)


scripts/release.js
```diff
@@ -79,7 +79,7 @@ const strategy = currentBranch !== 'develop' || forceBeta ? 'beta' : 'stable';
 function rollback(newVersion) {
   if (strategy === 'stable') {
     // reset master
-    shell.exec('git reset --hard origin master');
+    shell.exec('git reset --hard origin/master');
```
Haroenv: what about doing `--soft` instead? this command is indeed the fix though

bobylito: It will leave the working files there, and I fail to see any value in that. Can you give your opinion?

Haroenv: not sure either, just wondering if there could be any caution (if you e.g. released before pushing)?

bobylito: There is no rollback possible if the package is published. As far as I can see this is what we expect.

PR #3070, opened 2018-08-20
fix(Stats): let the widget render on all values

### LINKED ISSUE #3056: Avoid to rely on shouldComponentUpdate in component
**Describe the bug 🐛**

The stats widget re-renders only when the `nbHits` or `processingTimeMS` changes. It can lead to issue when you are using the page number in the template for example. When you click on the next page with the Pagination widget only the page value changed not the other ones. The shouldComponentUpdate (sCU) function of the widget will return false and so prevent the re-render.

https://github.com/algolia/instantsearch.js/blob/33fa09835df8678085cb7f382b4973eaa1f9b20f/src/components/Stats/Stats.js#L9-L14

**To Reproduce 🔍**

1. Go to https://codesandbox.io/s/l51or5k9lz
2. Click on the pagination
4. Stats widget is not updated

**Expected behavior 💭**

We should trigger the re-render. The quick fix is to add more values inside the condition. But do we really need the sCU hook on those components? To have a better idea we can track the effect on the performance with / without the hook.

bobylito:
This removes the implementation of shouldComponentUpdate. This can be seen as a regression but given that it's a pretty simple widget, the impact shouldn't be noticeable.

Fix #3056

### DIFF
--- a/src/components/Stats/Stats.js
+++ b/src/components/Stats/Stats.js
@@ -6,13 +6,6 @@ import autoHideContainerHOC from '../../decorators/autoHideContainer.js';
 import headerFooterHOC from '../../decorators/headerFooter.js';
 
 export class RawStats extends Component {
-  shouldComponentUpdate(nextProps) {
-    return (
-      this.props.nbHits !== nextProps.nbHits ||
-      this.props.processingTimeMS !== nextProps.processingTimeMS
-    );
-  }
-
   render() {
     const data = {
       hasManyResults: this.props.nbHits > 1,


samouss:
We implement this hook in several other components `ClearAll`, `PaginationLink`, `RefinementListItem`, ... We can also take a look at this other components. This hook is an escape hatch in most cases we don't need to implement it. Except when we hit a performance issue.

bobylito:
> We implement this hook in several other components ClearAll, PaginationLink, RefinementListItem, ... We can also take a look at this other components. This hook is an escape hatch in most cases we don't need to implement it. Except when we hit a performance issue.

Unlike this component, the others don't create issue as far as we know and they have been implemented as part of a general performance. Let's consider the case of the other on case by case basis.

PR #3082, opened 2018-08-27
fix(refinementList): showMore should continue working after search

### LINKED ISSUE #2478: SFFV in refinement list: when there is no result, the show more doesn't work afterwards
A user on our [community forum](https://discourse.algolia.com/t/show-more-option-not-working-after-facet-value-search-in-instansearch-js/3204) found a bug when using "search for facet values" with "show more".

He did a video in which he shows [how to reproduce the bug](https://www.youtube.com/watch?v=q5gXj-hT2ws&feature=player_embedded).

Steps to reproduce the issue:
 - have a [refinement list with search for facet values and show more activated](https://jsfiddle.net/xe6drzyh/1/)
 - search until no results
 - remove the search
 - show more is not usable anymore

### LINKED ISSUE #3064: Bad Interaction between Refinement List Search and Show More/Less
**Describe the bug 🐛**
We are using the Magento 2 module for integrating Algolia but this issue is not limited to that. It exists in the core of your code. 
If you create a refinement list widget that has both a showMore/limit configuration and a searchForFacetValues configuration, the showMore/limit feature breaks as soon as any text is entered into the search box. 

**To Reproduce 🔍**
1. Create a refinementList widget
2. Define a limit (5) and a showMore (limit 10, template doesn't need to be defined)
3. Define a searchForFacetValues (placeholder can be anything, not important)
4. Load this in the browser.
5. Type anything into the search box, then remove the text there (type a, then hit backspace)
6. Try clicking on the "Show More" text at the bottom of the refinement. Nothing happens.

See this sandbox:
https://codesandbox.io/s/r4wxknv32n
The Show More/Show Less link works just fine until you perform a search. 

**Expected behavior 💭**
After removing the search text the show more and show less links should work again. 

**Additional context**
In the Magento 2 module we note that when you hit Show more, the searchForFacetValues box becomes disabled, and when you hit Show Less it becomes re-enabled. I didn't see that happening in the Codesandbox I linked. But in both environments I can see that once you type a single letter in the search box the event handler for clicking on the Show More/Show Less gets removed. I think this is a similar thing to what I saw with the search box. One thing is attempting to disable the other.

bobylito:
**Summary**

We used to consider that when using `showMore`, you wouldn't use `searchForFacetValues`. Turns out it's been asked and even at Algolia we do so (latest example is the about page). This patch actually fixes the behavior so that both can work and that the end-user will be able to go from one to the other.

**Result**

fixes #2478 
fixes #3064

Scenario:
In a search with a refinement list that has both `showMore` and `searchForFacetValues` activated.
 - click on show more
 - type some search in the query (should hide the show more / show less button)
 - remove the query (should display the show more / show less button with the correct state)
 - toggle show less (should work)

Old implementation:

Fixed implementation:


Haroenv:
Nice, I noticed this in Vue InstantSearch too

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -800,6 +800,118 @@ describe('connectRefinementList', () => {
     });
   });
 
+  it('can search in facet values and show more values', () => {
```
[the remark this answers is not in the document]
bobylito: This PR does not implement a way to use show while you use sffv. It let have both options and correctly goes back and forth from one mode to the other.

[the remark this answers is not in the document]
bobylito: Happy to hear any suggestion of better test scenario :)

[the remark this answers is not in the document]
samouss: What I meant by that is the test does not test the implementation of the PR. The test will pass even without the changes - is it intended?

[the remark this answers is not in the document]
bobylito: Nope I went a bit fast on the test indeed. I'm going to rework the test! Good catch.

francoischalifour:
This is not yet ready because needs more tests.

francoischalifour:
I rebased the PR against the up-to-date `develop` branch to prepare this fix for InstantSearch.js 3.

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -711,6 +711,86 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
     ]);
   });
 
+  it('show more should toggle between two limits after search', () => {
```
francoischalifour: There's possibly a better way to test this.

I rely on the number of renders and the fact that `toggleShowMore` is defined after having searched for items (this test fails when run with the previous `connectRefinementList`).

tkrugg:
I tested this on the story book and it doesn't look like the issue is solved.
Maybe I'm misunderstanding it.
Shouldn't the "Show less" button disappear altogether in the following demo?




https://deploy-preview-3082--instantsearchjs.netlify.com/stories/?selectedKind=RefinementList&selectedStory=with%20show%20more&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

francoischalifour:
What's addressed in this PR is the behavior when [searching for facet values](https://deploy-preview-3082--instantsearchjs.netlify.com/stories/?selectedKind=RefinementList&selectedStory=with%20search%20and%20show%20more&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel) (SFFV).

We can indeed also address this behavior issue.

Haroenv:
If I understand it correctly, this just doesn't show the show more button anymore when in SFFV mode, but when there's actually more results available or when already showing more, wouldn't you expect that limit to be the one in SFFV too?

tkrugg:
I see, that wasn't very clear to me after I read the original issue description: https://github.com/algolia/instantsearch.js/issues/3064#issue-350118877

@sylvain-h: since you've expressed interest in this issue and asked for an ETA, can you confirm your use case involves [searching for facet values](https://user-images.githubusercontent.com/393765/44665394-169a1800-aa16-11e8-8978-6328695e8ebf.gif), (not [searching for hits](https://user-images.githubusercontent.com/2982512/53161552-7d198080-35ca-11e9-9d5a-7a62e0e557c7.gif)).

sylvain-h:
> @francoischalifour I see, that wasn't very clear to me after I read the original issue description: [#3064 (comment)](https://github.com/algolia/instantsearch.js/issues/3064#issue-350118877)
> 
> @sylvain-h: since you've expressed interest in this issue and asked for an ETA, can you confirm your use case involves [searching for facet values](https://user-images.githubusercontent.com/393765/44665394-169a1800-aa16-11e8-8978-6328695e8ebf.gif), (not [searching for hits](https://user-images.githubusercontent.com/2982512/53161552-7d198080-35ca-11e9-9d5a-7a62e0e557c7.gif)).

Initially my request was when searching for hits. It's when you change the input from the main search box that it breaks the refinementList.

tkrugg:
@sylvain-h ok so this a different issue. Is this something easy to address @francoischalifour ?

francoischalifour:
The `toggleShowMore()` method cannot work when SFFV as is because it re-renders the widget, and therefore loses the query in the search box.

Here's what I suggest:

If you've started typing in the SFFV searchbox without having toggled "Show more", display `limit` results. Otherwise, display `limit` results (fixes issue mentioned in @Haroenv's message).

If there's no hits returned in the main search, hide the refinement list "Show more" button (fixes issue mentioned in @tkrugg's message).

I've implemented this behavior that you can test [in this story](https://deploy-preview-3082--instantsearchjs.netlify.com/stories/?selectedKind=RefinementList&selectedStory=with%20search%20and%20show%20more).

Can you confirm that's what we want to go for so I can unit test it fully?

francoischalifour:
@sylvain-h confirmed that this is the expected behavior.

We're good for review @algolia/instantsearch-for-websites.

tkrugg:
@francoischalifour I see one last thing I think can be improved.

Since the latest fix, we have the following situation:
- if we have refinements, the showMore/showLess button is hidden. This is perfect 👍 
- If I have less refinements than `limit`, the showMore button will not show more items. Similarly, the showLess button will not collapse the few refinements displayed.
As a results, theses buttons appear to be broken.



As a user I think it would make more sense that we hide them or disable them. What are your thoughts on that?

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -236,7 +242,12 @@ export default function connectRefinementList(renderFn, unmountFn) {
         };
 
         helper
-          .searchForFacetValues(attribute, query, limit, tags)
+          .searchForFacetValues(
+            attribute,
+            query,
+            isShowingMore ? showMoreLimit : limit,
```
samouss: We can update and use the `getLimit` function.

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -711,6 +711,102 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
...
+    );
+
+    return Promise.resolve().then(() => {
+      expect(rendering).toHaveBeenCalledTimes(3);
+      const renderingOptions3 = rendering.mock.calls[2][0];
+
+      // `searchForItems` triggers a new render
+      renderingOptions3.searchForItems('');
+
+      return Promise.resolve().then(() => {
```
samouss: We could avoid to nest the promise.

samouss:
> As a user I think it would make more sense that we hide them or disable them. What are your thoughts on that?

Agree with @tkrugg we should do something in that case. I would go for the disable button since it's more aligned with the rest of the lifecycle. We only hide it when it's a SFFV or no results. We are not in those cases.

src/connectors/refinement-list/connectRefinementList.js
francoischalifour: `this.getLimit()` is not available at this step.

francoischalifour:
Good catch @tkrugg – will take that into account.

src/connectors/refinement-list/connectRefinementList.js
samouss: Yes but I don't think it's a reason to not use it. We could extract the function to re-use it.

francoischalifour:
The behavior should correspond to what was described above.

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -614,7 +614,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
     secondRenderingOptions.toggleShowMore();
 
     const thirdRenderingOptions = rendering.mock.calls[2][0];
-    expect(thirdRenderingOptions.canToggleShowMore).toBe(true);
+    expect(thirdRenderingOptions.canToggleShowMore).toBe(false);
```
francoischalifour: This behavior changed.

yannickcr:
Should have fixed above bug, there is still the Show more/less button disappearing when using the search box from `refinementList` but if I understand the above conversations correctly then this is how it was designed.

src/components/RefinementList/RefinementList.js
```diff
@@ -171,7 +171,9 @@ class RefinementList extends Component {
         rootTagName="button"
         rootProps={{
           className: showMoreButtonClassName,
-          onClick: this.props.toggleShowMore,
+          onClick: this.props.canToggleShowMore
+            ? this.props.toggleShowMore
+            : noop,
```
samouss: We a `disabled` prop we should be able to avoid the `noop`.

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -193,24 +203,25 @@ export default function connectRefinementList(renderFn, unmountFn) {
...
           toggleShowMore,
           hasExhaustiveItems,
         },
         isFirstSearch
       );
     };
 
-    let lastResultsFromMainSearch;
     let searchForFacetValues;
     let refine;
```
samouss: We can hoist the mutable values together at the top.

Haroenv:
If I understand it correctly, the show more button still disappears when searching for facet values, I don't think it is because there's not enough results

francoischalifour:
@Haroenv This is expected because not possible with the current implementation: https://github.com/algolia/instantsearch.js/pull/3082#issuecomment-467074809. As long as it respects the current limit, the behavior is good IMO (`limit` by default, `showMoreLimit` if we toggle more).

PR #3097, opened 2018-09-07
feat(stats): InstantSearch v3

francoischalifour:
## Specs

https://instantsearch-css.netlify.com/widgets/stats/

### DIFF
--- a/src/components/Stats/Stats.js
+++ b/src/components/Stats/Stats.js
@@ -3,7 +3,6 @@ import React, { Component } from 'preact-compat';
 
 import Template from '../Template.js';
 import autoHideContainerHOC from '../../decorators/autoHideContainer.js';
-import headerFooterHOC from '../../decorators/headerFooter.js';
 
 export class RawStats extends Component {
   shouldComponentUpdate(nextProps) {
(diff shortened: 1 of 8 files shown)


src/components/Stats/Stats.js
```diff
@@ -46,4 +54,4 @@ RawStats.propTypes = {
   templateProps: PropTypes.object.isRequired,
 };
 
-export default autoHideContainerHOC(headerFooterHOC(RawStats));
+export default autoHideContainerHOC(RawStats);
```
bobylito: You should remove the autohideContainer as well. We want to handle that in the panel.

src/widgets/stats/stats.js
```diff
@@ -110,7 +103,7 @@ stats({
  * @typedef {Object} StatsWidgetOptions
  * @property {string|HTMLElement} container Place where to insert the widget in your webpage.
  * @property {StatsWidgetTemplates} [templates] Templates to use for the widget.
- * @property {StatsWidgetTransforms} [transformData] Object that contains the functions to be applied on the data * before being used for templating. Valid keys are `body` for the body template.
+ * @property {StatsWidgetTransforms} [transformData] Object that contains the functions to be applied on the data before being used for templating. Valid keys are `text` for the text template.
  * @property {boolean} [autoHideContainer=true] Make the widget hides itself when there is no results matching.
```
bobylito: It should be removed too

PR #3102, opened 2018-09-10
docs(refinement-list): fix `sortBy` description

francoischalifour:
The `transformItems` documentation was above the `sortBy` definition.

See https://community.algolia.com/instantsearch.js/v2/widgets/refinementList.html#struct-RefinementListWidgetOptions-sortBy


Haroenv:
that empty line before probably isn't necessary?

francoischalifour:
@Haroenv It is to be rendered as another paragraph.

PR #3107, opened 2018-09-11
docs(contributing): fix dev-novel link on deploy

francoischalifour:
Dev Novel is accessible at `/v2/dev-novel` for quite some time, not `/dev-novel`.


bobylito:
Good catch!

PR #3112, opened 2018-09-12
docs(autocomplete): indices is an object array, not a string array

Haroenv:
Not sure if jsdoc will take this in account, but can be changed if necessary once I see the deploy preview

**Summary**

`indices` on `connectAutoComplete` expects an array of label/value and not a string


francoischalifour:
It renders as `[Array<{value: , label: }>]`. We could probably do better.

francoischalifour:
I reckon that square brackets for optional params become confusing for such types.

PR #3127, opened 2018-09-18
feat(searchbox): implement InstantSearch.css

bobylito:
## Specs

https://instantsearch-css.netlify.com/widgets/search-box/

## Stories

https://deploy-preview-3127--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=SearchBox.default


src/widgets/search-box/search-box.js
```diff
@@ -162,66 +142,45 @@ const usage = `Usage:
...
-
-/**
- * @typedef {Object} SearchBoxLoadingIndicatorOption
- * @property {function|string} template Template used for displaying the button. Can accept a function or a Hogan string.
- * @property {{root: string}} [cssClasses] CSS classes added to the loading-indicator element.
+ * @typedef {Ojbect} SearchBoxTemplates
+ * @property {function|string} submit Template used for displaying the submit. Can accept a function or a Hogan string.
+ * @property {function|string} reset Template used for displaying the button. Can accept a function or a Hogan string.
+ * @property {function|string} loadingIndicator Template used for displaying the button. Can accept a function or a Hogan string.
  */
```
francoischalifour: `submitIcon` and `resetIcon` are missing

francoischalifour:
I would also rename `SearchBox/index.js` to `SearchBox/SearchBox.js` for consistency.

samouss:
Now that the `input` is inside a form when we press `enter` the page is reloaded.

src/widgets/search-box/search-box.js
```diff
@@ -376,161 +303,94 @@ function addDefaultAttributesToInput(placeholder, input, query, cssClasses) {
...
-
-  const htmlNode = createNodeFromString(stringNode);
+function wrapInputFn(input, cssClasses) {
+  const wrapper = document.createElement('div');
+  wrapper.className = cx(suit(), cssClasses.root);
 
-  input.parentNode.insertBefore(htmlNode, input.nextSibling);
-}
+  const form = document.createElement('form');
+  form.className = cx(suit({ descendantName: 'form' }), cssClasses.form);
```
Haroenv: is this kept in vanilla JS for combat reasons? 

The form needs an onSubmit handler with preventDefault and the focus moving like in RIS

Haroenv: https://blog.algolia.com/mobile-search-ux-tips/ should have most of it explained :) for the rest you can base on the React InstantSearch one, which I know is correct :)

src/widgets/search-box/search-box.js
```diff
@@ -1,67 +1,43 @@
...
   templates,
   autofocus,
-  searchOnEnterKeyPressOnly,
-  wrapInput,
-  reset,
-  magnifier,
-  loadingIndicator,
+  searchAsYouType,
+  showReset,
+  showSubmit,
```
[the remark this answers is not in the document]
bobylito: When we were writing this component, we thought it was a good idea to let the user choose if they wanted the magnifying glass and the cross (with submit and reset behavior). The former name of the option is magnifier actually, I just renamed it.

Happy to change things here.

src/widgets/search-box/search-box.js
```diff
@@ -244,53 +203,46 @@ searchBox({
...
+  templates,
 } = {}) {
   if (!container) {
     throw new Error(usage);
   }
 
   const containerNode = getContainerNode(container);
 
+  if (containerNode.tagName === 'INPUT') {
+    throw new Error(`container should not be an INPUT`);
```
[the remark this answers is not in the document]
bobylito: OK will provide a link to the current documentation.

bobylito:
> I would also rename `SearchBox/index.js` to `SearchBox/SearchBox.js` for consistency.

This searchbox is used for the refinement list searchable feature. So it's kinda out of scope.

src/widgets/search-box/search-box.js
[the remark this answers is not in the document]
francoischalifour: I see. It seems like React InstantSearch provides [`submit`](https://community.algolia.com/react-instantsearch/widgets/SearchBox.html#default-props-entry-SearchBox-submit) and [`reset`](https://community.algolia.com/react-instantsearch/widgets/SearchBox.html#default-props-entry-SearchBox-reset) options to provide node elements. The way we deal with that in InstantSearch.js for now is having boolean to show/hide these elements.

Let's keep it that way for now.

src/widgets/search-box/search-box.js
```diff
@@ -68,7 +68,10 @@ const renderer = ({
         refine(getValue(e), false);
       });
       addListener(input, 'keyup', e => {
-        if (e.keyCode === KEY_ENTER) refine(getValue(e));
+        if (e.keyCode === KEY_ENTER) {
```
Haroenv: since we have a form now, this can be done in a submit event on the form instead, and the ENTER code can be removed :)

src/widgets/search-box/search-box.js
bobylito: Actually those are `submit` and `reset`. Do you think it would be better with the suffix?

src/widgets/search-box/search-box.js
bobylito: Yes indeed :)

src/widgets/search-box/search-box.js
francoischalifour: Oh ok, nevermind then!

bobylito:
> `autofocus` should be set to `false` by default now according to the specs, not to `"autofocus"`.

Indeed and I completely agree with that :)

francoischalifour:
@bobylito `autofocus` documentation and stories should also be updated.

src/widgets/search-box/search-box.js
bobylito: Do you think it's good now @Haroenv ? :)

Haroenv: I think so based on the code, but the dev novel still has the old code 🤔

src/widgets/search-box/search-box.js
```diff
@@ -309,31 +273,6 @@ export default function searchBox({
 // 'onpropertychange' is the IE8 alternative until we support IE8
```
Haroenv: this code can be removed, IMO we should have used the React version for consistency, but understand that this is less work.

src/widgets/search-box/search-box.js
```diff
@@ -268,11 +269,6 @@ export default function searchBox({
   }
 }
 
-// the 'input' event is triggered when the input value changes
-// in any case: typing, copy pasting with mouse..
-// 'onpropertychange' is the IE8 alternative until we support IE8
-// but it's flawed: http://help.dottoro.com/ljhxklln.php
```
Haroenv: comment was deleted, but not the code that triggered the comment :)

francoischalifour: Done!

src/widgets/search-box/search-box.js
```diff
@@ -376,161 +290,94 @@ function addDefaultAttributesToInput(placeholder, input, query, cssClasses) {
...
-
+function addSubmit(input, cssClasses, templates) {
   const stringNode = renderTemplate({
-    templateKey: 'template',
-    templates: magnifier,
-    data: {
-      cssClasses: magnifierCSSClasses,
-    },
+    templateKey: 'submit',
+    templates,
```
samouss: We can't override the class of the SVG element.

src/widgets/search-box/search-box.js
```diff
@@ -376,161 +290,94 @@ function addDefaultAttributesToInput(placeholder, input, query, cssClasses) {
...
-
+function addReset(input, cssClasses, templates, clearFunction) {
   const stringNode = renderTemplate({
-    templateKey: 'template',
-    templates: reset,
-    data: {
-      cssClasses: resetCSSClasses,
-    },
+    templateKey: 'reset',
+    templates,
```
samouss: We can't override the class of the SVG element.

src/widgets/search-box/search-box.js
```diff
@@ -73,81 +43,82 @@ const renderer = ({
...
-  if (reset) {
-    const resetBtnSelector = `.${cx(bem('reset-wrapper'))}`;
-    // hide reset button when there is no query
-    const resetButton =
-      containerNode.tagName === 'INPUT'
-        ? containerNode.parentNode.querySelector(resetBtnSelector)
-        : containerNode.querySelector(resetBtnSelector);
-    resetButton.style.display = query && query.trim() ? 'block' : 'none';
-  }
+  renderAfterInit({
```
samouss: With the previous implement we trigger this function only on render but now it also trigger on init, not sure about the implications. But we almost never render on the init lifecycle to avoid a flash with the empty state.

src/widgets/search-box/search-box.js
```diff
@@ -1,67 +1,37 @@
...
-      containerNode.appendChild(wrappedInput);
-    }
-
-    if (magnifier) addMagnifier(input, magnifier, templates);
-    if (reset) addReset(input, reset, templates, clear);
-    if (loadingIndicator)
-      addLoadingIndicator(input, loadingIndicator, templates);
+    const input = document.createElement('input');
+    const wrappedInput = wrapInputFn(input, cssClasses);
+    containerNode.appendChild(wrappedInput);
```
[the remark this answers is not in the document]
francoischalifour: According to @bobylito we faced some performance issues having this component written with Preact.
We can try to migrate it in another pass!

dev/app/builtin/stories/search-box.stories.js
```diff
@@ -14,110 +14,66 @@ export default () => {
...
-        container.innerHTML = '<input/>';
-        const input = container.firstChild;
-        container.appendChild(input);
-        window.search.addWidget(
-          instantsearch.widgets.searchBox({
-            container: input,
+            templates: {
+              submit: '<div class="ais-search-box--magnifier">🔍</div>',
+              reset: '<div class="ais-search-box--reset">✖️</div>',
+            },
```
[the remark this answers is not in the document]
francoischalifour: Yes – we'll update all templates in the widgets that haven't been updated yet.

src/widgets/search-box/search-box.js
```diff
@@ -59,18 +65,18 @@ const renderer = ({
     const form = input.parentElement;
 
     if (searchAsYouType) {
-      addListener(input, 'input', getInputValueAndCall(refine));
-      addListener(form, 'submit', e => {
-        e.preventDefault();
+      input.addEventListener('input', refine);
```
samouss: It doesn't work anymore since we provide the event to `refine` rather than the value. It should probably be catch with a test.

src/widgets/search-box/search-box.js
```diff
@@ -59,18 +65,18 @@ const renderer = ({
...
+      form.addEventListener('submit', event => {
+        event.preventDefault();
         input.blur();
       });
     } else {
-      addListener(input, 'input', e => {
-        refine(getValue(e), false);
+      input.addEventListener('input', event => {
+        refine(event, false);
```
samouss: It doesn't work anymore since we provide the event to `refine` rather than the value. It should probably be catch with a test.

src/widgets/search-box/search-box.js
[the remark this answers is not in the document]
francoischalifour: It is documented there, right? https://github.com/algolia/instantsearch.js/blob/feat/3.0-searchbox/src/widgets/search-box/search-box.js#L162

src/widgets/search-box/search-box.js
```diff
@@ -73,84 +48,73 @@ const renderer = ({
...
+
+    if (searchAsYouType) {
+      input.addEventListener('input', event => {
+        refine(event.currentTarget.value);
       });
-      addListener(input, 'keyup', e => {
-        if (e.keyCode === KEY_ENTER) refine(getValue(e));
+      form.addEventListener('submit', event => {
+        event.preventDefault();
+        input.blur();
```
samouss: Is there a reason why we blur?

src/widgets/search-box/search-box.js
```diff
@@ -73,84 +48,73 @@ const renderer = ({
...
-          ifKey(KEY_SUPPRESS, getInputValueAndCall(refine))
-        );
-      }
+      input.addEventListener('input', event => {
+        refine(event.currentTarget.value, false);
+      });
+      form.addEventListener('submit', event => {
+        refine(input.value);
+        event.preventDefault();
+        input.blur();
```
samouss: Is there a reason why we blur?

src/widgets/search-box/__tests__/search-box-test.js
```diff
@@ -304,300 +233,27 @@ describe('searchBox()', () => {
...
-      it('updates the query', () => {
-        expect(helper.setQuery).toHaveBeenCalledTimes(1);
-      });
-
-      it('does not search', () => {
-        expect(helper.search).toHaveBeenCalledTimes(0);
-      });
-    });
-
-    describe('using a queryHook', () => {
```
samouss: Is there a reason why we dropped this test?

src/widgets/search-box/search-box.js
```diff
@@ -162,68 +126,46 @@ const usage = `Usage:
...
   container,
   [ placeholder ],
-  [ cssClasses.{input,poweredBy} ],
-  [ poweredBy=false || poweredBy.{template, cssClasses.{root,link}} ],
-  [ wrapInput ],
-  [ autofocus ],
-  [ searchOnEnterKeyPressOnly ],
+  [ cssClasses.{root, input, reset, submit, loadingIndicator} ],
+  [ autofocus = false ],
+  [ searchAsYouType = true ],
```
samouss: `showLoadingIndicator`, `showReset`, `showSubmit` are missing

src/widgets/search-box/defaultTemplates.js
```diff
@@ -1,59 +1,33 @@
...
-                <animateTransform
-                    attributeName="transform"
-                    type="rotate"
-                    from="0 18 18"
-                    to="360 18 18"
-                    dur="1s"
-                    repeatCount="indefinite"/>
-            </path>
-        </g>
+<svg class="{{cssClasses.loadingIcon}}" width="16" height="16" viewBox="0 0 38 38" xmlns="http://www.w3.org/2000/svg" stroke="#444">
```
samouss: We don't provide this className to the renderer. It should be named `loadingIndicatorIcon` according to the two others `submitIcon`, `resetIcon`.

samouss: Yes we provide the `loadingIndicator` but not the `loadingIcon`.

> https://github.com/algolia/instantsearch.js/pull/3127/files#diff-38bdd34306afd017efc86c093d618177R67

src/widgets/search-box/__tests__/search-box-test.js
samouss: Me neither, it's tested inside the connector though. Let's keep it like that.

src/widgets/search-box/search-box.js
```diff
@@ -243,54 +190,81 @@ searchBox({
...
   if (!container) {
     throw new Error(usage);
   }
 
   const containerNode = getContainerNode(container);
 
-  // Only possible values are 'auto', true and false
-  if (typeof autofocus !== 'boolean') {
-    autofocus = 'auto';
+  if (containerNode.tagName === 'INPUT') {
```
Neekhaulas: Hello,
Why would you force user to not use an already existing input as a search input?

francoischalifour: This change was made to be more consistent across all widgets in all InstantSearch flavors. You probably want to use [`connectSearchBox`](https://www.algolia.com/doc/api-reference/widgets/searchbox/js/#connector) now.

Please [file an issue](https://github.com/algolia/instantsearch.js/issues/new/choose) if you have trouble migrating.

PR #3132, opened 2018-09-20
feat(highlight): implement InstantSearch.css

francoischalifour:
This PR add functionalities to highlight hits without manipulating the Algolia response object (`_highlightResult`). It provides two ways to highlight a hit.

## Specs

[See specs →](https://instantsearch-css.netlify.com/widgets/highlight/)

Note that I didn't add the CSS class `ais-Highlight-nonHighlighted` because it complexifies the parsing for not much added value. This is also what we did for Vue InstantSearch.

### As a function

The hit must be explicitly passed to the helper function to retrieve the highlighted value.

```js
instantsearch.highlight({ attribute, highlightedTagName, hit })
// => <mark class="ais-HighLight-highlighted">Amazon</mark> - Fire TV Stick with Alexa Voice Remote - Black
```

### As a Hogan template helper

The hit is injected leveraging the `data` prop from the `Template` component, allowing to omit the `hit` parameter when using the Hogan template. The parameters are forwarded to the `highlight()` function.

```js
{{#highlight}}{ "attribute": "name", "highlightedTagName": "mark" }{{/highlight}}
// => <mark class="ais-HighLight-highlighted">Amazon</mark> - Fire TV Stick with Alexa Voice Remote - Black
```

If the given object is not parsable with `JSON.parse()`, we return an empty string.

## Stories

- [Function](https://deploy-preview-3132--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=Hits.with%20highlight%20function)
- [Helper](https://deploy-preview-3132--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=Hits.with%20highlight%20helper)


src/lib/highlight.js
```diff
@@ -0,0 +1,22 @@
...
+  attribute,
+  highlightedTagName = 'mark',
+  hit,
+} = {}) {
+  const attributeValue =
+    getPropertyByPath(hit, `_highlightResult.${attribute}.value`) || '';
+
+  return attributeValue
+    .replace(
+      new RegExp('<em>', 'g'),
```
bobylito: What about using a regex litteral instead? `/<em>/g`

src/lib/utils.js
```diff
@@ -436,3 +436,9 @@ function parseAroundLatLngFromString(value) {
     lng: parseFloat(pattern[2]),
   };
 }
+
+export function getPropertyByPath(object, path) {
```
bobylito: I would unit test that :)

Haroenv: https://github.com/algolia/vue-instantsearch/blob/feat/connectors/src/util/__tests__/object.test.js

src/lib/highlight.js
francoischalifour: `/<em>/g` doesn't replace all instances.

francoischalifour: Sorry, I didn't read correctly. Will change that!

francoischalifour:
## Documentation

Should we add another section for "Helpers"? (not sure about the name)

image

- instantsearch()
- Widgets
- Connectors
- Helpers
  - `highlight()`
  - `snippet()`

Haroenv:
helpers sounds good IMO

francoischalifour:
Going for "Helpers" containing:

- `highlight`
- `snippet`
- `formatNumber`

src/lib/createHelpers.js
```diff
@@ -1,7 +1,25 @@
...
     formatNumber(number, render) {
       return Number(render(number)).toLocaleString(numberLocale);
     },
+    highlight(options, render) {
+      let highlightOptions;
+
+      try {
+        highlightOptions = JSON.parse(options);
+      } catch (error) {
+        return render('');
```
francoischalifour: We shouldn't fail silently.

Rather warn in the console or throw an error. WDYT?

Haroenv: that should be an error, since if the JSON is wrong, that means the highlight will never work. 

If the highlighted attribute is not in attributesForHighlighting, that should be a warning (only in dev), like in Vue InstantSearch

src/lib/main.js
```diff
@@ -179,5 +180,6 @@ instantsearch.createQueryString =
 instantsearch.connectors = connectors;
 instantsearch.widgets = widgets;
 instantsearch.version = version;
+instantsearch.highlight = helpers.highlight;
```
samouss: It should also be exported from `index.es6.js` to allow users that don't use the full bundled version to use the `highlight` function.

Haroenv: same in #3134 then :)

francoischalifour: Thanks for spotting that!

PR #3133, opened 2018-09-20
feat(search-client): use search client

francoischalifour:
This PR removes all instances of the old way of using InstantSearch.js:

- `createAlgoliaAgent()`
- `appId` and `apiKey`

`algoliasearch` is now a dev dependency.

## Changes

- Update source code
- Update documentation
- Update README
- Update demos


docgen/src/examples/e-commerce/search.js
```diff
@@ -170,17 +154,17 @@ search.addWidget(
...
-    templates: {
-      link: '<i class="fa fa-eraser"></i> Clear all filters'
-    },
-    cssClasses: {
-      root: 'btn btn-block btn-default'
-    },
-    autoHideContainer: true
-  })
-);
+// search.addWidget(
```
francoischalifour: This would need to be replaced by the new ClearRefinements widget.

docgen/src/getting-started.md.hbs
```diff
@@ -60,9 +60,10 @@ The jsDeliver CDN is highly available with [over 110 locations](https://www.jsde
 If you have a JavaScript build system, you can install **InstantSearch.js** from NPM:
 
 ```javascript
-// `npm install instantsearch.js --save` OR
-// `yarn add instantsearch.js`
+// `npm install instantsearch.js algoliasearch` OR
+// `yarn add instantsearch.js algoliasearch`
 
+const algoliasearch = require('algoliasearch');
```
samouss: Prefer the `algoliasearch/lite` version.

package.json
```diff
@@ -52,6 +52,7 @@
     "es"
   ],
   "devDependencies": {
+    "algoliasearch": "3.30.0",
```
samouss: We should also add it as a `peerDependency`.

francoischalifour: You don't have to depend on `algoliasearch` if you want to create your own search client. Should we still list it as a peer dependency?

samouss:
There is some places when we still have the `appId` & `apiKey` form. The documentation website throws an error on the home page.

https://github.com/algolia/instantsearch.js/blob/ac5d693119ba6e9c8a3373078f192769351b2a0c/docgen/assets/js/bindRunExamples.js#L5-L14

francoischalifour:
@samouss This is solved now.

package.json
samouss: Nope you're right!

docgen/assets/js/bindRunExamples.js
```diff
@@ -4,13 +4,14 @@ import capitalize from "lodash/capitalize";
...
-  urlSync: false,
+  indexName: window.searchConfig.indexName || "instant_search",
+  searchClient: algoliasearch(
+    window.searchConfig.appId || "latency",
+    window.searchConfig.apiKey || "6be0576ff61c053d5f9a3225e2a90f76"
+  ),
   searchParameters: {
-    hitsPerPage: 3
-  },
-  ...window.searchConfig
```
samouss: It there a reason why we don't spread the rest of the arguments anymore?

francoischalifour: In the frontmatter, you can add `appId` and `apiKey` that get into `window.searchConfig`.

InstantSearch doesn't allow passing `appId` and `apiKey` directly, but require a search client.

We can still spread `window.searchConfig` without these two options.

francoischalifour: Note that we only take into account the options `appId`, `apiKey` and `indexName`: https://github.com/algolia/instantsearch.js/blob/feat/3.0-search-client/docgen/plugins/searchConfig.js#L14

PR #3130, opened 2018-09-19
feat(numeric-selector): implement InstantSearch.css

francoischalifour:
## Specs

https://instantsearch-css.netlify.com/widgets/numeric-selector/

## Stories

https://deploy-preview-3130--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=NumericSelector.default

PR #3139, opened 2018-09-24
fix(numericselector): default value can be undefined

bobylito:
Waiting for the netlify build

Conditions of the bug:
 - routing
 - default undefined value (first one in the list)

Can be reproduced in the dev-novel.


francoischalifour:
Related to broken stories here: https://github.com/algolia/instantsearch.js/pull/3130#pullrequestreview-157724749.

PR #3142, opened 2018-09-24
fix(community): fix search config

francoischalifour:
This PR fixes the `searchConfig` object on the community website when it is `undefined`.

I also took the opportunity to refactor the way we extract the search configuration and made it possible to override the whole object (as it was before).

### DIFF
--- a/docgen/assets/js/bindRunExamples.js
+++ b/docgen/assets/js/bindRunExamples.js
@@ -6,8 +6,7 @@ const {
   appId = "latency",
   apiKey = "6be0576ff61c053d5f9a3225e2a90f76",
   ...searchConfig
-} =
-  window.searchConfig || {};
+} = window.searchConfig || {};
 
 window.instantsearch = instantsearch;
 window.search = instantsearch({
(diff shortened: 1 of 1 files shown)


docgen/assets/js/bindRunExamples.js
```diff
@@ -2,16 +2,21 @@ import instantsearch from "../../../src/lib/main.js";
 import algoliasearch from "algoliasearch/lite";
 import capitalize from "lodash/capitalize";
 
+const {
+  appId = "latency",
+  apiKey = "6be0576ff61c053d5f9a3225e2a90f76",
+  ...searchConfig
+} =
+  window.searchConfig || {};
```
samouss: Weird indentation, it should be on the same line no?

francoischalifour: Prettier does that... I'm not sure why either.

samouss: Weird I tried in the REPL and I didn't have this.

francoischalifour:
@samouss I tried with Prettier 1.14.3 and the behavior is normal indeed. Will prettify this line this way. We're not using the same Prettier configuration for this file anyway.

PR #3143, opened 2018-09-24
chore(v3): remove URLSync

bobylito:
This PR removes the URLSync.

This PR depends on #3148 to be fixed.


src/lib/InstantSearch.js
```diff
@@ -226,7 +206,7 @@ class InstantSearch extends EventEmitter {
         // We don't want to re-add URlSync `getConfiguration` widget
         // it can throw errors since it may re-add SearchParameters about something unmounted
         this.searchParameters = this.widgets
-          .filter(w => w.constructor.name !== 'URLSync')
+          .filter(w => w.constructor.name !== 'URLSync') // TODO: check the routing for this
```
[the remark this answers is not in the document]
bobylito: There is no issue as far as I can test with the current behavior. I'm getting rid of it.

src/lib/InstantSearch.js
```diff
@@ -205,9 +205,9 @@ class InstantSearch extends EventEmitter {
       if (nextState) {
         // We don't want to re-add URlSync `getConfiguration` widget
         // it can throw errors since it may re-add SearchParameters about something unmounted
-        this.searchParameters = this.widgets
-          .filter(w => w.constructor.name !== 'URLSync') // TODO: check the routing for this
```
samouss: Just to let you know, we can't use the `name` attribute of function or constructor. Once Uglify pass on the code all values are renamed and the constructor name is changed (except with an Uglify option that is by default to `false`, see [`keep_fnames `](https://github.com/mishoo/UglifyJS2#minify-options)).

bobylito: That's good to know thanks. It might have never worked in production then for the `URLSync`.

bobylito:
The doc is now updated

src/lib/InstantSearch.js
samouss: Yep indeed.

Haroenv:
Linting is failign in this PR :)

bobylito:
> Linting is failign in this PR :)

Linting never sleeps :)

PR #3147, opened 2018-09-25
feat(utils): add warn function

samouss:
**Summary**

This PR introduces a function to display warnings on the page only once (cached by message). I didn't update all the usage of `console.warn` across the codebase to avoid conflict with the 3.0 branch. We can move on latter once the next version is released.

### DIFF
--- a/src/lib/__tests__/utils-test.js
+++ b/src/lib/__tests__/utils-test.js
@@ -899,6 +899,38 @@ describe('utils.deprecate', () => {
   });
 });
 
+describe('utils.warn', () => {
+  it('expect to print a warn message', () => {
+    const message = 'message';
+    const warn = jest.spyOn(global.console, 'warn');
+
+    utils.warn(message);
+
+    expect(warn).toHaveBeenCalled();
+
+    warn.mockReset();
+    warn.mockRestore();
+    utils.warn.cache = {};
+  });
+
+  it('expect to print the same message only once', () => {
+    const message = 'message';
+    const warn = jest.spyOn(global.console, 'warn');
+
+    utils.warn(message);
+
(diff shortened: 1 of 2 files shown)


src/lib/__tests__/utils-test.js
```diff
@@ -899,6 +899,38 @@ describe('utils.deprecate', () => {
...
 });
 
+describe('utils.warn', () => {
+  it('expect to print a warn message', () => {
+    const message = 'message';
+    const warn = jest.spyOn(global.console, 'warn');
+
+    utils.warn(message);
+
+    expect(warn).toHaveBeenCalled();
```
francoischalifour: Shouldn't we assert that the warn function is called with the right string prefixed by `[InstantSearch.js]: `? (see [`toHaveBeenCalledWith()`](https://jestjs.io/docs/en/expect#tohavebeencalledwitharg1-arg2-))

samouss: Yes we can, will do it.

samouss:
@Haroenv we can merge it as it is. But once we release both (Vue & IS) we can take a look to extract some of the core logic of InstantSearch (like connectors, utility function etc...) to share them across all flavours.

francoischalifour:
There's also some work to be done to make our error messages more explicit to users. A first pass could be done when updating all `console.warn` to this warn utility function.

I'd love to make a smarter pass on our error messages and to detect where the error might come from. Say that you're facing a "cannot retrieve `min` of `undefined`" → "Have you declared the attribute `attribute` as a facet in the dashboard?".

This next step would be a good time to create some sort of package to share common errors that our libraries can throw.

PR #3149, opened 2018-09-25
fix(routing): enforce RoutingManager is the last widget

### LINKED ISSUE #3148: Unmount using routing
When unmounting some widget that requires a configuration in a context of an app that leverages the routing we have an error.

More specifically it happens on those widgets:
 - [toggle](https://community.algolia.com/instantsearch.js/v2/dev-novel/?widgets=unmount&selectedStory=Toggle.default)
 - [starRating](https://community.algolia.com/instantsearch.js/v2/dev-novel/?widgets=unmount&selectedStory=StarRating.default)
 - [menu](https://community.algolia.com/instantsearch.js/v2/dev-novel/?widgets=unmount&selectedStory=Menu.default)

To reproduce just mount and remount.

I've tested with a modified dev-novel that doesn't use the routing manager and it works fine. Also the stack trace indicates that the error happens during the routingManager lifecycle (addWidgets).

screen shot 2018-09-25 at 15 14 10

Those widgets have in common that they try to set a value or get information about a facet that is not defined in the search state. If the router widget was always called last we wouldn't have this issue.

bobylito:
fix #3148

See the issue for more context.

Those widgets can't be added back in IS.js:
 - [toggle](https://community.algolia.com/instantsearch.js/v2/dev-novel/?widgets=unmount&selectedStory=Toggle.default)
 - [starRating](https://community.algolia.com/instantsearch.js/v2/dev-novel/?widgets=unmount&selectedStory=StarRating.default)
 - [menu](https://community.algolia.com/instantsearch.js/v2/dev-novel/?widgets=unmount&selectedStory=Menu.default)

But they can with this patch.
 - [toggle](https://deploy-preview-3149--algolia-instantsearch.netlify.com/v2/dev-novel/?widgets=unmount&selectedStory=Toggle.default)
 - [starRating](https://deploy-preview-3149--algolia-instantsearch.netlify.com/v2/dev-novel/?widgets=unmount&selectedStory=StarRating.default)
 - [menu](https://deploy-preview-3149--algolia-instantsearch.netlify.com/v2/dev-novel/?widgets=unmount&selectedStory=Menu.default)

The underlying reason is that the router needs to have a complete state before going through its own `getConfiguration` because it will iterate over all the widgets which expects to have their own `getConfiguration` called first.


src/lib/InstantSearch.js
```diff
@@ -195,6 +195,12 @@ To help you migrate, please refer to the migration guide: https://community.algo
       );
     }
 
+    // The routing manager widget is always added manually at the last position.
+    // By removing it from the last position and adding it back after, we ensure
+    // it keeps this position.
+    // fixes #3148
+    const lastWidget = this.widgets.pop();
```
Haroenv: it's a fine fix I guess, but what if the routing manager came to a different key than `this.widgets`? That would solve the issue too no?

bobylito: Yes indeed. I chose to keep it in the `this.widgets` because otherwise it would require to call the lifecycle steps of the routing manager with specific lines of code. I favored the systematic aspect here.

Haroenv: now that we are talking about this line anyway; I think it would make sense to call it `_widgets`, since it's exposed on the instantsearch instance, and clearly isn't something that should be touched externally. In a later stage we can use the proper private property when they are supported & polyfilled :)

src/lib/InstantSearch.js
```diff
@@ -204,6 +210,9 @@ To help you migrate, please refer to the migration guide: https://community.algo
       this.widgets.push(widget);
     });
 
+    // Second part of the fix for #3148
+    if (lastWidget) this.widgets.push(lastWidget);
```
Haroenv: this breaks the unit tests, not sure if it also breaks the usage? I think it happens because the order might be changed; but not sure. See: 

https://travis-ci.org/algolia/instantsearch.js/jobs/432952845#L1074

bobylito: Yes I saw that. The 0th widget is moved back in the last which is consistent with the purpose of the PR. I will fix the test.

src/lib/InstantSearch.js
bobylito: That sounds like a good idea. Do you mind if I do that in another PR because it involves changes in other places too?

PR #3151, opened 2018-09-26
feat(hierarchical-menu): implement show more feature

bobylito:
This adds the show more feature to the hierachical menu as per the instantsearch.css spec. It does not update the markup and only does the strict minimum in order to have the feature working.

## Stories

https://deploy-preview-3151--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=HierarchicalMenu.with%20show%20more

### DIFF
--- a/src/connectors/hierarchical-menu/connectHierarchicalMenu.js
+++ b/src/connectors/hierarchical-menu/connectHierarchicalMenu.js
@@ -83,6 +83,7 @@ export default function connectHierarchicalMenu(renderFn, unmountFn) {
       rootPath = null,
       showParentLevel = true,
       limit = 10,
+      showMoreLimit,
       sortBy = ['name:asc'],
       transformItems = items => items,
     } = widgetParams;
@@ -97,6 +98,26 @@ export default function connectHierarchicalMenu(renderFn, unmountFn) {
     const [hierarchicalFacetName] = attributes;
 
     return {
+      isShowingMore: false,
+
+      // Provide the same function to the `renderFn` so that way the user
+      // has to only bind it once when `isFirstRendering` for instance
+      toggleShowMore() {},
+      cachedToggleShowMore() {
+        this.toggleShowMore();
+      },
+
+      createToggleShowMore(renderOptions) {
+        return () => {
+          this.isShowingMore = !this.isShowingMore;
+          this.render(renderOptions);
+        };
+      },
+
+      getLimit() {
+        return this.isShowingMore ? showMoreLimit : limit;
+      },
+
       getConfiguration: currentConfiguration => {
         if (currentConfiguration.hierarchicalFacets) {
           const isFacetSet = find(
@@ -118,7 +139,7 @@ export default function connectHierarchicalMenu(renderFn, unmountFn) {
           }
         }
 
-        return {
+        const widgetConfiguration = {
           hierarchicalFacets: [
             {
               name: hierarchicalFacetName,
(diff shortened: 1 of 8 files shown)


src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -97,6 +98,26 @@ export default function connectHierarchicalMenu(renderFn, unmountFn) {
     const [hierarchicalFacetName] = attributes;
 
     return {
+      isShowingMore: false,
+
+      // Provide the same function to the `renderFn` so that way the user
+      // has to only bind it once when `isFirstRendering` for instance
+      toggleShowMore() {},
+      cachedToggleShowMore() {
```
Haroenv: Seems like this is a new way to write this pattern? Seems fine to me, but maybe at a later point we can make that consistent across all connectors?

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -186,13 +232,35 @@ export default function connectHierarchicalMenu(renderFn, unmountFn) {
...
+        // If the limit is the max number of facet retrieved it is impossible to know
+        // if the facets are exhaustive. The only moment we are sure it is exhaustive
+        // is when it is strictly under the number requested unless we know that another
+        // widget has requested more values (maxValuesPerFacet > getLimit()).
+        // Because this is used for making the search of facets unable or not, it is important
+        // to be conservative here.
+        const hasExhaustiveItems =
+          maxValuesPerFacetConfig > currentLimit
+            ? facetValues.length <= currentLimit
+            : facetValues.length < currentLimit;
```
samouss: The engine doesn't handle this case for us?

src/widgets/hierarchical-menu/hierarchical-menu.js
```diff
@@ -46,6 +54,10 @@ const renderer = ({
       shouldAutoHideContainer={shouldAutoHideContainer}
       templateProps={renderState.templateProps}
       toggleRefinement={refine}
+      showMore={showMore}
+      toggleShowMore={toggleShowMore}
+      isShowingMore={isShowingMore}
+      canToggleShowMore={true}
```
samouss: The value should come from the connector right?

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
bobylito: I'm not sure to understand which part could be handled by the engine here. Can you tell me more? I might have missed something.

src/widgets/hierarchical-menu/hierarchical-menu.js
bobylito: Indeed it should be provided but the connector.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
samouss: Yep me too, I messed up!

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -128,14 +149,30 @@ export default function connectHierarchicalMenu(renderFn, unmountFn) {
...
           ],
-          maxValuesPerFacet:
-            currentConfiguration.maxValuesPerFacet !== undefined
-              ? Math.max(currentConfiguration.maxValuesPerFacet, limit)
-              : limit,
         };
+
+        if (limit !== undefined) {
+          const currentMaxValuesPerFacet =
+            currentConfiguration.maxValuesPerFacet || 0;
```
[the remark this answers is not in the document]
bobylito: I added tests for the limit https://github.com/algolia/instantsearch.js/pull/3151/commits/ca2469183884d7073c74f5d7fc7446a90d9c64d2

Let me know if that's what you were looking for

francoischalifour:
Will need to add documentation for this.

PR #3155, opened 2018-10-01
refactor(geo): remove support for paddingBoundingBox [PART-1]

samouss:
**Summary**

This PR removes the support of `paddingBoundingBox`. This option doesn't play nicely with the URLSync - mainly because we are not able to determine the exact same position for the bounding box with a computed padding. For this reason we choose to drop the support of the option to be able to implement a proper URLSync - option that is more valuable.

### DIFF
--- a/docgen/src/guides/migrations-geo-search.md
+++ b/docgen/src/guides/migrations-geo-search.md
@@ -1,0 +1,32 @@
+---
+title: Migration v3 - GeoSearch
+mainTitle: Guides
+layout: main.pug
+category: guides
+withHeadings: true
+navWeight: 0
+editable: true
+githubSource: docgen/src/guides/migrations-geo-search.md
+---
+
+> TODO: move the content of the migration guide inside the v3 migration guide
+
+### GeoSearch
+
+#### Options
+
+| Before               | After    |
+| -------------------- | -------- |
+| `paddingBoundingBox` | Removed  |
+
+- `paddingBoundingBox` was in conflit with the `routing` option - so we removed it to support URLSync for the GeoSearch widget.
+
+### connectGeoSearch
+
+#### Options
+
+| Before               | After    |
+| -------------------- | -------- |
+| `paddingBoundingBox` | Removed  |
+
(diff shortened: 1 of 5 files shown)


docgen/src/guides/migrations-geo-search.md
```diff
@@ -0,0 +1,32 @@
...
+
+### connectGeoSearch
+
+#### Options
+
+| Before               | After    |
+| -------------------- | -------- |
+| `paddingBoundingBox` | Removed  |
+
+- `paddingBoundingBox` was in conflit with the `routing` option - so we removed it to support URLSync for the GeoSearch widget.
```
Haroenv: conflict

docgen/src/guides/migrations-geo-search.md
```diff
@@ -0,0 +1,32 @@
...
+
+### GeoSearch
+
+#### Options
+
+| Before               | After    |
+| -------------------- | -------- |
+| `paddingBoundingBox` | Removed  |
+
+- `paddingBoundingBox` was in conflit with the `routing` option - so we removed it to support URLSync for the GeoSearch widget.
```
Haroenv: conflict

src/widgets/geo-search/__tests__/geo-search-test.js
```diff
@@ -30,17 +30,11 @@ describe('GeoSearch', () => {
...
-        lat: jest.fn(),
-        lng: jest.fn(),
+      getNorthEast: jest.fn(() => ({
+        toJSON: jest.fn(() => ({})),
       })),
-      fromLatLngToPoint: jest.fn(() => ({
-        x: 0,
-        y: 0,
+      getSouthWest: jest.fn(() => ({
+        toJSON: jest.fn(() => ({})),
```
Haroenv: valuable to have real possible values here or not?

samouss: We do have valuable values here later on but not by default - mostly because we don't care about the value.

> https://github.com/algolia/instantsearch.js/pull/3158/files#diff-bc6fb9f0c9db3d39fdc54426badeeed3R1254

PR #3157, opened 2018-10-01
refactor(geo): implement currentRefinement [PART-3]

samouss:
**Summary**

This PR implement the support of `currentRefinement` in the geo-search connector. We now provide the current bounding box to the render function. We can leverage this information to implement a map widget that is "controlled" (i.e. almost always rely on the bounding box to set the map position).

### DIFF
--- a/src/lib/__tests__/utils-test.js
+++ b/src/lib/__tests__/utils-test.js
@@ -931,7 +931,7 @@ describe('utils.warn', () => {
   });
 });
 
-describe('utils.parseAroundLatLngFromString', () => {
+describe('utils.aroundLatLngToPosition', () => {
   it('expect to return a LatLng object from string', () => {
     const samples = [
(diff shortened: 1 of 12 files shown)


src/lib/__tests__/utils-test.js
```diff
@@ -931,26 +931,86 @@ describe('utils.warn', () => {
...
+  ])('expect to return a Position from a string: %j', (input, expectation) => {
+    expect(utils.aroundLatLngToPosition(input)).toEqual(expectation);
   });
 
-  it('expect to throw an error when the parsing fail', () => {
-    const samples = [{ input: '10a,12' }, { input: '10.    12' }];
+  it.each([['10a,12'], ['10.    12']])(
+    'expect to throw an error with: %j',
+    input => {
+      expect(() => utils.aroundLatLngToPosition(input)).toThrow();
```
francoischalifour: Do we need to specify the expected thrown error?

PR #3160, opened 2018-10-01
refactor(geo): implement static map [PART-6]

samouss:
**Summary**

This PR implements the static map support for the geo-search widget. Previously it was not possible to use the widget for display purpose only. You were always force to use the refinement capabilities of the map. We now have the option `enableRefine` that disable the refinement from the map.

**Result**

static

You can play with the live example in [DevNovel](https://deploy-preview-3160--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=GeoSearch.with%20refine%20disabled).

### DIFF
--- a/src/components/GeoSearchControls/GeoSearchControls.js
+++ b/src/components/GeoSearchControls/GeoSearchControls.js
@@ -7,6 +7,7 @@ import GeoSearchToggle from './GeoSearchToggle';
 
 const GeoSearchControls = ({
   cssClasses,
+  enableRefine,
   enableRefineControl,
   enableClearMapRefinement,
   isRefineOnMapMove,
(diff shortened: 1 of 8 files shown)


src/components/GeoSearchControls/GeoSearchControls.js
```diff
@@ -16,67 +17,72 @@ const GeoSearchControls = ({
...
+}) =>
+  enableRefine && (
+    <div>
+      {enableRefineControl && (
+        <div className={cssClasses.control}>
+          {isRefineOnMapMove || !hasMapMoveSinceLastRefine ? (
+            <GeoSearchToggle
+              classNameLabel={cx(
+                cssClasses.toggleLabel,
+                isRefineOnMapMove && cssClasses.toggleLabelActive
```
francoischalifour: You can use the `classnames` feature if needed (we went for this syntax in the version 3 branch).

```js
cx(cssClasses.toggleLabel, {
  [cssClasses.toggleLabelActive]: isRefineOnMapMove,
})
```

samouss: Yep the change is already on the branch `feat/3.0`.

https://github.com/algolia/instantsearch.js/blob/cdea3ab514899f02395e4f08292f859eed967404/src/components/GeoSearchControls/GeoSearchControls.js#L25-L27

PR #3164, opened 2018-10-02
feat(poweredBy): implement InstantSearch.css

francoischalifour:
## Specs

https://instantsearch-css.netlify.com/widgets/powered-by/

## Stories

https://deploy-preview-3164--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=PoweredBy.default


dev/app/builtin/stories/powered-by.stories.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable import/default */
+
+import { storiesOf } from 'dev-novel';
+import instantsearch from '../../../../index';
+import { wrapWithHits } from '../../utils/wrap-with-hits.js';
+
+const stories = storiesOf('PoweredBy');
+
+window.connectPoweredBy = instantsearch.connectors.connectPoweredBy;
+window.poweredBy = instantsearch.widgets.poweredBy;
```
samouss: Why do we copy the reference on the global object?

src/components/PoweredBy/PoweredBy.js
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+class PoweredBy extends Component {
```
samouss: We can use a functional component.

src/components/PoweredBy/PoweredBy.js
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react';
```
samouss: The import should come from `preact-compat`.

src/widgets/powered-by/powered-by.js
```diff
@@ -0,0 +1,94 @@
...
+const suit = component('PoweredBy');
+
+const renderer = ({ containerNode, cssClasses }) => (
+  { widgetParams },
+  isFirstRendering
+) => {
+  if (isFirstRendering) {
+    const { url, theme } = widgetParams;
+
+    render(
```
samouss: It won't have a huge impact here but we don't render at the init step - only at the render.

src/widgets/powered-by/powered-by.js
```diff
@@ -0,0 +1,94 @@
...
+poweredBy({
+  container,
+  [ theme = 'light' ],
+})`;
+
+/**
+ * @typedef {Object} PoweredByWidgetOptions
+ * @property {string|HTMLElement} container Place where to insert the widget in your webpage.
+ * @property {string} [theme] The theme of the logo ("light" or "dark").
+ * @property {PoweredByWidgetCssClasses} [cssClasses] CSS classes to add.
```
samouss: The definition `PoweredByWidgetCssClasses` is missing.

src/widgets/powered-by/powered-by.js
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+/**
+ * The `poweredBy` widget is used to display useful insights about the current results.
```
samouss: Not sure about the description ^^

src/widgets/powered-by/powered-by.js
```diff
@@ -0,0 +1,94 @@
...
+  container,
+  cssClasses: userCssClasses = {},
+  theme = 'light',
+} = {}) {
+  if (!container) {
+    throw new Error(usage);
+  }
+
+  const containerNode = getContainerNode(container);
+  const url =
```
samouss: The URL should be in the connector but it might be possible to override it. We can also make the `utm_source` configurable then Vue & Angular can also use the connector to create their PoweredBy component. See [Vue - PoweredBy](https://github.com/algolia/vue-instantsearch/blob/feat/connectors/src/components/PoweredBy.vue)

dev/app/builtin/stories/powered-by.stories.js
francoischalifour: Oops! Forgot to remove those.

src/widgets/powered-by/powered-by.js
francoischalifour: How would you see it then?

samouss: At the init step you don't (see below). We only call `prepareTemplateProps` at the first render we early return right after. Otherwise we might have a flash of content with the empty state.

https://github.com/algolia/instantsearch.js/blob/92cf470442be7321b9102b8ae7585f302029ff0d/src/widgets/hits/hits.js#L16-L45

src/connectors/powered-by/connectPoweredBy.js
```diff
@@ -30,20 +40,29 @@ export default function connectPoweredBy(renderFn, unmountFn) {
...
+    const defaultUrl =
+      'https://www.algolia.com/?' +
+      'utm_source=instantsearch.js&' +
+      'utm_medium=website&' +
+      `utm_content=${location.hostname}&` +
+      'utm_campaign=poweredby';
+
     return {
-      init() {
+      init({ url = defaultUrl }) {
```
samouss: The URL should come from the widgetParams not the InstantSearch instance to allow other users to override it. Right now we don't have a way to override it from user code. For example in Vue we should be able to do something like:

```js
export default {
  mixins: [createWidget({ connector: connectPoweredBy })],
  computed: {
    widgetParams() {
      return {
        // it would be nice to also support this:
        source: "vue-instantsearch",
        // but also
        URL:
          "https://www.algolia.com/?" +
          "utm_source=vue-instantsearch&" +
          "utm_medium=website&" +
          `utm_content=${location ? location.hostname : ""}&` +
          "utm_campaign=poweredby"
      };
    }
  }
};
```

src/widgets/powered-by/powered-by.js
francoischalifour: Since this widget doesn't depend on the state and isn't meant to be re-rendered, I though it would be appropriate to only render it at first rendering. Am I wrong?

src/connectors/powered-by/connectPoweredBy.js
```diff
@@ -0,0 +1,74 @@
...
+export default function connectPoweredBy(renderFn, unmountFn) {
+  checkRendering(renderFn, usage);
+
+  return (widgetParams = {}) => {
+    const url =
+      widgetParams.url ||
+      'https://www.algolia.com/?' +
+        'utm_source=instantsearch.js&' +
+        'utm_medium=website&' +
+        `utm_content=${location.hostname}&` +
```
Haroenv: this needs a conditional on `location` I think. While usually not used for SSR, probably best to avoid the pattern

PR #3165, opened 2018-10-02
refactor(breadcrumb): refactor widget

francoischalifour:
- Remove unused `cx()`
- Use functional functional components when possible
- Enforce prop types for class names
- Use Enzyme over React Test Renderer
- Convert Sinon to Jest

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.js
+++ b/src/components/Breadcrumb/Breadcrumb.js
@@ -1,14 +1,7 @@
-import React, { PureComponent } from 'preact-compat';
+import React from 'preact-compat';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
 import Template from '../Template.js';
-
-const itemsPropType = PropTypes.arrayOf(
-  PropTypes.shape({
-    name: PropTypes.string,
-    value: PropTypes.string,
-  })
-);
 
 const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
   item,
@@ -22,8 +15,8 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
     <a
       className={cssClasses.link}
       href={createURL(item.value)}
-      onClick={e => {
-        e.preventDefault();
+      onClick={event => {
+        event.preventDefault();
         refine(item.value);
       }}
     >
@@ -31,15 +24,16 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
     </a>
   );
 
-  const itemClassnames = cx(cssClasses.item, {
-    [cssClasses.selectedItem]: isLast,
-  });
-
   return (
-    <li key={item.name + idx} className={itemClassnames}>
+    <li
+      key={item.name + idx}
+      className={cx(cssClasses.item, {
+        [cssClasses.selectedItem]: isLast,
+      })}
+    >
       <Template
         rootTagName="span"
-        rootProps={{ className: cx(cssClasses.separator), ariaHidden: true }}
+        rootProps={{ className: cssClasses.separator, 'aria-hidden': true }}
         templateKey="separator"
         {...templateProps}
       />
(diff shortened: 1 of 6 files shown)


src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -48,57 +42,72 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
...
+    event.preventDefault();
+    refine(null);
+  };
 
-    const homeOnClickHandler = e => {
-      e.preventDefault();
-      refine(null);
-    };
+  const homeUrl = createURL(null);
+  const breadcrumb = items.map(
```
samouss: You can inline the value in the return.

src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -48,57 +42,72 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
...
+    selectedItem: PropTypes.string.isRequired,
+    separator: PropTypes.string.isRequired,
+    link: PropTypes.string.isRequired,
+  }).isRequired,
+  items: PropTypes.arrayOf(
+    PropTypes.shape({
+      name: PropTypes.string,
+      value: PropTypes.string,
+    })
+  ),
```
samouss: `.isRequired`

src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -48,57 +42,72 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
...
-            />
-          </li>
-          {breadcrumb}
-        </ul>
-      </div>
-    );
-  }
-}
+Breadcrumb.propTypes = {
+  createURL: PropTypes.func,
```
samouss: `.isRequired`

src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -48,57 +42,72 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
...
+    link: PropTypes.string.isRequired,
+  }).isRequired,
+  items: PropTypes.arrayOf(
+    PropTypes.shape({
+      name: PropTypes.string,
+      value: PropTypes.string,
+    })
+  ),
+  refine: PropTypes.func.isRequired,
+  separator: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
```
samouss: `.isRequired`

src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -48,57 +42,72 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
...
+    noRefinement: PropTypes.string.isRequired,
+    list: PropTypes.string.isRequired,
+    item: PropTypes.string.isRequired,
+    selectedItem: PropTypes.string.isRequired,
+    separator: PropTypes.string.isRequired,
+    link: PropTypes.string.isRequired,
+  }).isRequired,
+  items: PropTypes.arrayOf(
+    PropTypes.shape({
+      name: PropTypes.string,
```
samouss: `.isRequired`

src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -48,57 +42,72 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
...
+      value: PropTypes.string,
+    })
+  ),
+  refine: PropTypes.func.isRequired,
+  separator: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+  templateProps: PropTypes.object.isRequired,
+  translate: PropTypes.func,
```
samouss: You can remove this one, it's unused.

src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -48,57 +42,72 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
...
+    [cssClasses.noRefinement]: items.length > 0,
+  });
+  const homeClassnames = cx(cssClasses.item, {
+    [cssClasses.selectedItem]: items.length === 0,
+  });
 
-    const rootClassnames = cx(cssClasses.root, {
-      [cssClasses.noRefinement]: items.length > 0,
-    });
+  const homeOnClickHandler = event => {
```
samouss: You can inline it.

src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -48,57 +42,72 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
...
-      [cssClasses.selectedItem]: items.length === 0,
-    });
+const Breadcrumb = ({
+  createURL,
+  items,
+  refine,
+  cssClasses,
+  templateProps,
+}) => {
+  const rootClassnames = cx(cssClasses.root, {
```
samouss: You can inline it.

src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -48,57 +42,72 @@ const renderLink = ({ cssClasses, createURL, refine, templateProps }) => (
...
-
-  render() {
-    const { createURL, items, refine, cssClasses } = this.props;
-
-    const breadcrumb = items.map(renderLink(this.props));
-
-    const homeClassnames = cx(cssClasses.item, {
-      [cssClasses.selectedItem]: items.length === 0,
-    });
+const Breadcrumb = ({
```
tkrugg: i'm curious on why we don't choose not to adopt more PureComponent?
Any reason for this @algolia/instantsearch-for-websites

Haroenv: It hasn't proved necessary / useful to use PureComponent when we tried it the previous time. Note that this is Preact, not React, so they don't behave exactly the same

samouss: Inside the connector we often times create new functions (e.g. refine) on each render so the `PureComponent` check would fail every time. It means that we will do the check for nothing (not in that case though). But as @Haroenv said until we have performance issues then we can stick with functional component.

PR #3169, opened 2018-10-03
refactor: update widgets

francoischalifour:
This is the remaining of other PRs.

- Remove unused `cx()`
- Use functional functional components when possible
- Enforce prop types for class names
- Use Enzyme over React Test Renderer
- Convert Sinon to Jest

### DIFF
--- a/src/components/ClearRefinements/__tests__/ClearRefinements-test.js
+++ b/src/components/ClearRefinements/__tests__/ClearRefinements-test.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import ClearRefinements from '../ClearRefinements';
-import renderer from 'react-test-renderer';
+import { mount } from 'enzyme';
 
 describe('ClearRefinements', () => {
   const defaultProps = {
@@ -20,16 +20,16 @@ describe('ClearRefinements', () => {
   };
 
   it('should render <ClearRefinements />', () => {
-    const tree = renderer
-      .create(<ClearRefinements {...defaultProps} />)
-      .toJSON();
-    expect(tree).toMatchSnapshot();
+    const wrapper = mount(<ClearRefinements {...defaultProps} />);
+
+    expect(wrapper).toMatchSnapshot();
   });
 
   it('should render <ClearRefinements /> with a specific class when no refinements', () => {
-    const tree = renderer
-      .create(<ClearRefinements {...defaultProps} hasRefinements={false} />)
-      .toJSON();
-    expect(tree).toMatchSnapshot();
+    const wrapper = mount(
+      <ClearRefinements {...defaultProps} hasRefinements={false} />
+    );
+
+    expect(wrapper).toMatchSnapshot();
   });
 });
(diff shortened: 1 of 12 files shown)


src/widgets/hits-per-page/hits-per-page.js
```diff
@@ -1,7 +1,6 @@
 import React, { render, unmountComponentAtNode } from 'preact-compat';
 import cx from 'classnames';
```
Haroenv: is this import still used?

src/widgets/sort-by/sort-by.js
```diff
@@ -12,10 +11,12 @@ const renderer = ({ containerNode, cssClasses }) => (
...
   isFirstRendering
 ) => {
-  if (isFirstRendering) return;
+  if (isFirstRendering) {
+    return;
+  }
 
   render(
-    <div className={cx(cssClasses.root)}>
+    <div className={cssClasses.root}>
```
Haroenv: not sure I understand the reasoning of these changes?

src/widgets/hits-per-page/hits-per-page.js
tkrugg: shoudn't the linter detect it otherwise?

src/components/ClearRefinements/__tests__/ClearRefinements-test.js
```diff
@@ -20,16 +20,16 @@ describe('ClearRefinements', () => {
   };
 
   it('should render <ClearRefinements />', () => {
-    const tree = renderer
-      .create(<ClearRefinements {...defaultProps} />)
-      .toJSON();
-    expect(tree).toMatchSnapshot();
+    const wrapper = mount(<ClearRefinements {...defaultProps} />);
+
+    expect(wrapper).toMatchSnapshot();
```
tkrugg: I might have missed it but I'm surprised the snapshot didn't change at all after this. 🤔 
I guess the build would have failed if they didn't match

src/widgets/clear-refinements/__tests__/clear-refinements-test.js
```diff
@@ -29,10 +28,10 @@ describe('clearRefinements()', () => {
     results = {};
     helper = {
       state: {
-        clearRefinements: sinon.stub().returnsThis(),
-        clearTags: sinon.stub().returnsThis(),
+        clearRefinements: jest.fn().mockReturnValue(this),
+        clearTags: jest.fn().mockReturnValue(this),
```
samouss: You can use `mockReturnThis`

src/widgets/sort-by/sort-by.js
samouss: Inside the widget (below) we precompute the class we don't have to run them through the function again. See https://github.com/algolia/instantsearch.js/pull/3169/files#diff-f4506bac0a67ca481cb05bddf147f9c7R96

src/components/ClearRefinements/__tests__/ClearRefinements-test.js
francoischalifour: Emzyme's `mount` is a bit more explicit than React Test Renderer but in this case they output the same component.

src/widgets/hits-per-page/hits-per-page.js
francoischalifour: It is still used to compute the suit CSS classes below.

PR #3175, opened 2018-10-04
feat(widgets): use warn utils

francoischalifour:
Use the `warn` utility function instead of the native `console.warn`.

### DIFF
--- a/src/connectors/hits-per-page/connectHitsPerPage.js
+++ b/src/connectors/hits-per-page/connectHitsPerPage.js
@@ -1,7 +1,7 @@
 import some from 'lodash/some';
 import find from 'lodash/find';
 
-import { checkRendering } from '../../lib/utils.js';
+import { checkRendering, warn } from '../../lib/utils.js';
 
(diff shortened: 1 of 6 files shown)


src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -145,20 +145,18 @@ The first one will be picked, you should probably set only one default value`
...
-  using the searchParameters attribute of the instantsearch constructor.`
-              );
-            }
-          } else if (window.console) {
-            window.console.warn(
-              `[Warning][hitsPerPage] No item in \`items\`
-  with \`value: hitsPerPage\` (hitsPerPage: ${state.hitsPerPage})`
+            warn(
+              `HitsPerPage is not defined.
+You should set the value \`hitsPerPage\` using the \`searchParameters\` attribute of the \`instantsearch\` constructor.`
```
samouss: We should advocate for `configure` rather than `searchParameters`.

PR #3176, opened 2018-10-04
feat: compress templates

### LINKED ISSUE #3095: Clean HTML output
**Is your feature request related to a problem? Please describe 🙏**

The current HTML output of InstantSearch.js contains whitespaces, which makes it messy.

before

**Describe the solution you'd like 🤔**

By trimming the templates [once compiled by Hogan](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/utils.js#L182), we could clean the HTML output by removing the whitespaces.

```diff
 return hogan
     .compile(template, compileOptions)
     .render({
       ...data,
       helpers: transformedHelpers,
     })
+    .trim();
```

after

**Describe alternatives you've considered ✨**

We could also use [Handlebars' whitespace control](http://handlebarsjs.com/expressions.html#whitespace-control) features but I didn't get this to work.

**Additional context**

If we go for this in InstantSearch 3, we'd need to apply this change after having merged the InstantSearch.css specs because it'd also clean up some snapshots.

```diff
 <div
   dangerouslySetInnerHTML={
     Object {
-      "__html": "  $",
+      "__html": "$",
     }
   }
 />
 <div
   dangerouslySetInnerHTML={
     Object {
-      "__html": "  ",
+      "__html": "",
      }
   }
 />
```

francoischalifour:
Fixes #3095

### DIFF
--- a/src/components/Breadcrumb/__tests__/__snapshots__/Breadcrumb-test.js.snap
+++ b/src/components/Breadcrumb/__tests__/__snapshots__/Breadcrumb-test.js.snap
@@ -53,7 +53,7 @@ exports[`Breadcrumb should render <Breadcrumb /> with a single item 1`] = `
         className="separator"
         dangerouslySetInnerHTML={
           Object {
-            "__html": " > ",
+            "__html": ">",
           }
         }
       />
@@ -92,7 +92,7 @@ exports[`Breadcrumb should render <Breadcrumb /> with items 1`] = `
         className="separator"
         dangerouslySetInnerHTML={
           Object {
-            "__html": " > ",
+            "__html": ">",
           }
         }
       />
@@ -112,7 +112,7 @@ exports[`Breadcrumb should render <Breadcrumb /> with items 1`] = `
         className="separator"
         dangerouslySetInnerHTML={
           Object {
-            "__html": " > ",
+            "__html": ">",
           }
         }
       />
--- a/src/widgets/breadcrumb/__tests__/__snapshots__/breadcrumb-test.js.snap
+++ b/src/widgets/breadcrumb/__tests__/__snapshots__/breadcrumb-test.js.snap
@@ -34,7 +34,7 @@ exports[`breadcrumb() render renders transformed items correctly 1`] = `
     Object {
       "templates": Object {
         "home": "Home",
-        "separator": " > ",
+        "separator": ">",
       },
       "templatesConfig": undefined,
       "transformData": undefined,
--- a/src/components/Stats/__tests__/__snapshots__/Stats-test.js.snap
+++ b/src/components/Stats/__tests__/__snapshots__/Stats-test.js.snap
@@ -8,9 +8,7 @@ exports[`Stats should render <Stats /> with custom CSS classes 1`] = `
     className="text"
     dangerouslySetInnerHTML={
       Object {
-        "__html": "
-    1 result
-     found in ms",
+        "__html": "1 result found in ms",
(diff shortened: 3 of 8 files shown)


src/widgets/breadcrumb/__tests__/__snapshots__/breadcrumb-test.js.snap
```diff
@@ -34,7 +34,7 @@ exports[`breadcrumb() render renders transformed items correctly 1`] = `
     Object {
       "templates": Object {
         "home": "Home",
-        "separator": " > ",
+        "separator": ">",
```
francoischalifour: The spaces were ignored.

Haroenv: https://travis-ci.org/algolia/instantsearch.js/jobs/437387566#L689 snapshots for this also need to be updated then

tkrugg: I guess whether they get ignored depends on what's around right?
Because I can think of cases where those spaces matter and this may change them.
```jsx
Home<span dangerouslySetInnerHTML={{ __html: " > " }} />Category => Home > Category
Home <span dangerouslySetInnerHTML={{ __html: ">" }} /> Category => Home > Category
Home<span dangerouslySetInnerHTML={{ __html: ">" }} />Category => Home>Category

```

src/components/Breadcrumb/__tests__/__snapshots__/Breadcrumb-test.js.snap
```diff
@@ -53,7 +53,7 @@ exports[`Breadcrumb should render <Breadcrumb /> with a single item 1`] = `
         className="separator"
         dangerouslySetInnerHTML={
           Object {
-            "__html": " > ",
+            "__html": ">",
```
Haroenv: I actually think that this is not correct, since there won't be spaces around the separator anymore (unless they are manually put?)?

What happens if you copy paste now, vs previous?

src/widgets/breadcrumb/__tests__/__snapshots__/breadcrumb-test.js.snap
francoischalifour: This is how we inject the templates: https://github.com/algolia/instantsearch.js/blob/d72edb32935a3166c146770d914203e3f80b4571/src/components/Breadcrumb/Breadcrumb.js#L34-L40

Whitespaces will get added in any case (I made sure by adding a local story).

src/components/Breadcrumb/__tests__/__snapshots__/Breadcrumb-test.js.snap
francoischalifour: See https://github.com/algolia/instantsearch.js/pull/3176#discussion_r223459777

francoischalifour: Do you mean using [Enzyme's `html()`](https://airbnb.io/enzyme/docs/api/ShallowWrapper/html.html)?

It would output like this:

```html
"<div class="root noRefinement"><ul class="list"><li class="item"><a class="link">home</a></li><li class="item selectedItem"><span class="separator" aria-hidden="true">&gt;</span><!-- react-text: 7 -->name0<!-- /react-text --></li></ul></div>"
```

PR #3178, opened 2018-10-05
fix(InstantSearch): avoid useless search on addWidgets

samouss:
**Summary**

The current implementation of `addWidgets` **always** trigger a search even without any widgets to add on InstantSearch. We should prevent this behaviour to avoid useless request each time a user call `addWidgets` without anything. 

This issue pops out with a dynamic facetting use case (see example below) - we get an infinite loop because we always call `addWidgets` without checking the length of the widgets to add. It should be the default behaviour.

```js
// We should not have to do this check inside user code

if (addWidgets.length) {
  search.addWidgets(addWidgets.map(item => item.widget));
}
```

> https://codesandbox.io/s/nnpvp2x8r4

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -205,7 +205,7 @@ To help you migrate, please refer to the migration guide: https://community.algo
     });
 
     // Init the widget directly if instantsearch has been already started
-    if (this.started) {
+    if (this.started && Boolean(widgets.length)) {
       this.searchParameters = this.widgets.reduce(enhanceConfiguration({}), {
         ...this.helper.state,
       });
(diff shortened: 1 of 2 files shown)


src/lib/InstantSearch.js
```diff
@@ -205,7 +205,7 @@ To help you migrate, please refer to the migration guide: https://community.algo
     });
 
     // Init the widget directly if instantsearch has been already started
-    if (this.started) {
+    if (this.started && Boolean(widgets.length)) {
```
tkrugg: out of curiosity:why not just `&& widgets.length`?

samouss: I'm a type lover I prefer when it's explicit - but agree it's exactly the same.

PR #3179, opened 2018-10-05
feat(refinement-list): implement InstantSearch.css (2)

francoischalifour:
This PR flattens the templates to the top-level `templates` option.

## Specs

https://instantsearch-css.netlify.com/widgets/refinement-list/

## Stories

https://deploy-preview-3179--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=RefinementList.default

### DIFF
--- a/src/widgets/refinement-list/__tests__/refinement-list-test.js
+++ b/src/widgets/refinement-list/__tests__/refinement-list-test.js
@@ -27,6 +27,32 @@ describe('refinementList()', () => {
         refinementList(options);
       }).toThrow(/^Usage:/);
     });
+
+    it('throws an exception when showMoreLimit without showMore option', () => {
+      expect(
+        refinementList.bind(null, {
+          attribute: 'attribute',
+          container,
+          showMoreLimit: 10,
+        })
+      ).toThrowErrorMatchingInlineSnapshot(
+        `"\`showMoreLimit\` must be used with \`showMore\` set to \`true\`."`
+      );
+    });
+
+    it('throws an exception when showMoreLimit is lower than limit', () => {
+      expect(
+        refinementList.bind(null, {
+          attribute: 'attribute',
+          container,
+          limit: 20,
+          showMore: true,
+          showMoreLimit: 10,
+        })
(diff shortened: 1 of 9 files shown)


src/widgets/refinement-list/refinement-list.js
```diff
@@ -308,7 +271,7 @@ export default function refinementList({
       escapeFacetValues,
       transformItems,
     });
-  } catch (e) {
+  } catch (error) {
     throw new Error(usage);
```
tkrugg: It's not related to this PR but it's an occasion to ask the question.
Is there a specific policy regarding catching/displaying errors?

Here for example we choose to forget about the original error but it's also hiding some potentially helpful clues for the support team? Is this a choice?

Haroenv: This is a pattern in InstantSearch specifically. Usually it _does_ get fixed by rereading the input expected, but probably we should put the original error in there too (like on AlgoliaError on the client)

src/widgets/refinement-list/__tests__/refinement-list-test.js
```diff
@@ -27,6 +27,32 @@ describe('refinementList()', () => {
...
+
+    it('throws an exception when showMoreLimit without showMore option', () => {
+      expect(
+        refinementList.bind(null, {
+          attribute: 'attribute',
+          container,
+          showMoreLimit: 10,
+        })
+      ).toThrowErrorMatchingInlineSnapshot(
+        `"\`showMoreLimit\` must be used with \`showMore\` set to \`true\`."`
```
tkrugg: I curious why you prefer not let jest handle the snapshots here with `expect(...).toThrowErrorMatchingSnapshot()`.
(Since we're already using snapshots everywhere)

francoischalifour: This does let Jest handle the snapshot for us, except that it inlines it in the test file (and doesn't output it to a `.snap` file).

I find it handy for one-line snapshots – it avoids context switches when you need to open another file to read the output.

PR #3183, opened 2018-10-08
 feat(numericSelector): remove widget

francoischalifour:
This PR removes the widget NumericSelector as discussed in this PR: https://github.com/algolia/instantsearch.js/pull/3130#pullrequestreview-161232873.


docgen/src/guides/v3-migration.md
```diff
@@ -308,6 +308,10 @@ With the redo button:
 </div>
 ```
 
+### NumericSelector
+
+Widget removed.
```
Haroenv: should say "replaced by `numericMenu`" or something like that

francoischalifour: It's not replaced but rather can be reproduced with NumericMenu. Let's write a guide later on?

Haroenv: good for me, but should at least link to it IMO

francoischalifour: I agree - will do another PR to review the whole migration guide. Let's move forward with this PR.

Issue #984, opened 2016-04-19
Multiple lodash


xarg:
If you already have lodash in your package:

```
{
  "name": "algolia-test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "lodash": "^4.11.1"
  }
}
```

then when you `npm install instantsearch.js` it will download lodash 3 times again:

```
$ ls node_modules/**/node_modules/
node_modules/algoliasearch-helper/node_modules/:
lodash

node_modules/algoliasearch/node_modules/:
lodash

node_modules/instantsearch.js/node_modules/:
lodash
```

Ok.. this is how npm does things when there are version conflicts, but.. then when you're building with webpack it imports all lodash again for every package (almost 2mb redundant stuff in total):

```
...
 [891] ./~/algoliasearch-helper/~/lodash/math/sum.js 1.35 kB {0} [built]
  [892] ./~/algoliasearch-helper/~/lodash/internal/arraySum.js 491 bytes {0} [built]
  [893] ./~/algoliasearch-helper/~/lodash/internal/baseSum.js 569 bytes {0} [built]
  [894] ./~/algoliasearch-helper/~/lodash/internal/toIterable.js 527 bytes {0} [built]
  [895] ./~/algoliasearch-helper/~/lodash/object/values.js 691 bytes {0} [built]
  [896] ./~/algoliasearch-helper/~/lodash/internal/baseValues.js 593 bytes {0} [built]
  [897] ./~/algoliasearch-helper/~/lodash/collection/includes.js 2.05 kB {0} [built]
  [898] ./~/algoliasearch-helper/~/lodash/collection/sortByOrder.js 2.06 kB {0} [built]
  [899] ./~/algoliasearch-helper/~/lodash/internal/baseSortByOrder.js 1.1 kB {0} [built]
  [900] ./~/algoliasearch-helper/~/lodash/internal/baseSortBy.js 543 bytes {0} [built]
  [901] ./~/algoliasearch-helper/~/lodash/internal/compareMultiple.js 1.62 kB {0} [built]
  [902] ./~/algoliasearch-helper/~/lodash/internal/baseCompareAscending.js 1.06 kB {0} [built]
  [903] ./~/algoliasearch-helper/~/lodash/function/partial.js 1.29 kB {0} [built]
[...]

PR #3186, opened 2018-10-09
fix(deps): unpinned production dependencies

tkrugg:
ref: https://github.com/algolia/instantsearch.js/issues/984#issuecomment-427830179


Haroenv:
Yes @francoischalifour, seems to have been a patch change so we can't control that

PR #3207, opened 2018-10-17
fix #3204

### LINKED ISSUE #3204: Usage of deprecated methods on the helper
**Describe the bug 🐛**

Among other things I haven't found yet, `helper.toggleRefinement` -> `helper.toggleFacetRefinement`

**To Reproduce 🔍**
In lots of places, one is in `connectMenu` for example, but it's best to do them all at once, also change the tests etc.

**Expected behavior 💭**
Replace all usage of deprecated helper functions, so we can remove them

prudywsh:
**Summary**

Replace the deprecated helper method `helper.toggleRefinement` by `helper.toggleFacetRefinement`

**Result**

all the tests pass, and there are no regression in the UI


src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -171,7 +171,10 @@ const connectGeoSearch = (renderFn, unmountFn) => {
 
     const toggleRefineOnMapMove = () =>
       widgetState.internalToggleRefineOnMapMove();
-    const createInternalToggleRefinementonMapMove = (render, args) => () => {
+    const createInternalToggleFacetRefinementOnMapMove = (
+      render,
+      args
+    ) => () => {
```
samouss: This one should be reverted, because it's not related with facet refinement. The function change how the widget will trigger the refinement (on map move or not).

src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -199,7 +202,7 @@ const connectGeoSearch = (renderFn, unmountFn) => {
       const { state, helper, instantSearchInstance } = initArgs;
       const isFirstRendering = true;
 
-      widgetState.internalToggleRefineOnMapMove = createInternalToggleRefinementonMapMove(
+      widgetState.internalToggleRefineOnMapMove = createInternalToggleFacetRefinementOnMapMove(
```
samouss: To revert (same reason as before).

src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -251,7 +254,7 @@ const connectGeoSearch = (renderFn, unmountFn) => {
       widgetState.lastRefinePosition = state.aroundLatLng || '';
       widgetState.lastRefineBoundingBox = state.insideBoundingBox || '';
 
-      widgetState.internalToggleRefineOnMapMove = createInternalToggleRefinementonMapMove(
+      widgetState.internalToggleRefineOnMapMove = createInternalToggleFacetRefinementOnMapMove(
```
samouss: To revert (same reason as before).

prudywsh:
`createInternalToggleFacetRefinementOnMapMove` reverted :)

Issue #3205, opened 2018-10-17
There is no way to display the search query in currentRefinedValues while `onlyListedAttributes: true`


stianmartinsen:
**Describe the bug 🐛**
The search query is not shown in the *currentRefinedValues* widget when `onlyListedAttributes` is set to true.

**To Reproduce 🔍**
1. Go to https://codesandbox.io/s/8y2jrkz8o9
2. Filter by brand and see the brand is shown in the *currentRefinedValues* widget.
3. Enter a custom search query
4. See error, the search query is not displayed in the *currentRefinedValues* widget.

**Expected behavior 💭**
The search query should show up in the *currentRefinedValues* widget even when whitelisting the other shown attributes with `onlyListedAttributes: true`. If not by default, we should have a way of enabling it.

_Originally posted by @stianmartinsen in https://github.com/algolia/instantsearch.js/pull/2414#issuecomment-430514274_

samouss:
Thanks for reporting the issue. We can probably fix this issue by providing an attributeName to the query refinement when we create it (see below).

https://github.com/algolia/instantsearch.js/blob/84b65210add282af5962a133a71139124a95f9b6/src/lib/utils.js#L231-L235

francoischalifour:
Hey @stianmartinsen, the next release will contain a fix for this. It will get published in the coming days. Thank you!

PR #3223, opened 2018-10-18
fix(getRefinements): provide attributeName for query

tkrugg:
**Summary**
- added attributeName to fix issue https://github.com/algolia/instantsearch.js/issues/3205 and related
- added tests
- added example showcasing onlyListedAttributes on current-refined-values (maybe unnecessary)

**Result**
screenshot at oct 19 15-41-35

### DIFF
--- a/src/lib/__tests__/utils-test.js
+++ b/src/lib/__tests__/utils-test.js
@@ -342,8 +342,7 @@ describe('utils.getRefinements', () => {
     );
   });
 
-  it('should retrieve one facetRefinement and inject query facet if clearQuery === true', () => {
-    helper.toggleRefinement('facet1', 'facet1val1');
+  it('should inject query facet if clearQuery === true', () => {
     helper.setQuery('my query');
     const expected = [
       {
@@ -352,33 +351,20 @@ describe('utils.getRefinements', () => {
         name: 'my query',
         query: 'my query',
       },
-      { type: 'facet', attributeName: 'facet1', name: 'facet1val1' },
     ];
     const clearsQuery = true;
-    expect(
-      utils.getRefinements(results, helper.state, clearsQuery)
-    ).toHaveLength(2);
-    expect(
-      utils.getRefinements(results, helper.state, clearsQuery)
-    ).toContainEqual(expected[0]);
-    expect(
-      utils.getRefinements(results, helper.state, clearsQuery)
-    ).toContainEqual(expected[1]);
+    expect(utils.getRefinements(results, helper.state, clearsQuery)).toEqual(
+      expected
+    );
   });
 
   it('should retrieve one facetRefinement and not inject query facet if clearQuery === false', () => {
-    helper.toggleRefinement('facet1', 'facet1val1');
     helper.setQuery('my query');
-    const expected = [
-      { type: 'facet', attributeName: 'facet1', name: 'facet1val1' },
-    ];
+    const expected = [];
     const clearsQuery = false;
-    expect(
-      utils.getRefinements(results, helper.state, clearsQuery)
-    ).toHaveLength(1);
-    expect(
-      utils.getRefinements(results, helper.state, clearsQuery)
-    ).toContainEqual(expected[0]);
+    expect(utils.getRefinements(results, helper.state, clearsQuery)).toEqual(
+      expected
+    );
   });
 
   it('should retrieve multiple facetsRefinements on one facet', () => {
--- a/src/lib/__tests__/utils-test.js
+++ b/src/lib/__tests__/utils-test.js
@@ -342,6 +342,45 @@ describe('utils.getRefinements', () => {
     );
   });
 
+  it('should retrieve one facetRefinement and inject query facet if clearQuery === true', () => {
+    helper.toggleRefinement('facet1', 'facet1val1');
+    helper.setQuery('my query');
+    const expected = [
+      {
+        type: 'query',
+        attributeName: 'query',
+        name: 'my query',
+        query: 'my query',
+      },
+      { type: 'facet', attributeName: 'facet1', name: 'facet1val1' },
+    ];
+    const clearsQuery = true;
+    expect(
+      utils.getRefinements(results, helper.state, clearsQuery)
+    ).toHaveLength(2);
+    expect(
+      utils.getRefinements(results, helper.state, clearsQuery)
+    ).toContainEqual(expected[0]);
+    expect(
+      utils.getRefinements(results, helper.state, clearsQuery)
+    ).toContainEqual(expected[1]);
+  });
+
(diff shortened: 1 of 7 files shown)


src/lib/utils.js
```diff
@@ -232,6 +232,7 @@ function getRefinements(results, state, clearsQuery) {
             type: 'query',
             name: state.query,
```
francoischalifour: Do we want to display `"my query"` instead of `my query`?

If so, it'd be nice to wrap the `name` value with double quotes.

tkrugg: ah yes, we've mentioned this in [the related PR I made on ais](https://github.com/algolia/angular-instantsearch/pull/301#discussion_r226354923)
Isn't it a `<q>...</q>` tag IIUC ?

Haroenv: Yep, should be done in a q tag which displays as quotes

francoischalifour: @tkrugg It should be in this file then: https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/current-refined-values/defaultTemplates.js

tkrugg: image

So I can't find where that template rendering is tested. If I read correctly the render method is mocked so changing the default template can't break the tests?

francoischalifour: You can test the rendering in the component itself when rendering a `query`: https://github.com/algolia/instantsearch.js/blob/1510a9439bb22dd31eb4dc0c75bd4316a57db73d/src/components/CurrentRefinedValues/__tests__/CurrentRefinedValues-test.js

tkrugg: I maybe be missing something but this doesn't seem to be testing the default template neither since it's mocked.
https://github.com/algolia/instantsearch.js/blob/1510a9439bb22dd31eb4dc0c75bd4316a57db73d/src/components/CurrentRefinedValues/__tests__/CurrentRefinedValues-test.js#L16-L23

francoischalifour: Right, then getting back to [the widget test file](https://github.com/algolia/instantsearch.js/blob/1510a9439bb22dd31eb4dc0c75bd4316a57db73d/src/widgets/current-refined-values/__tests__/current-refined-values-test.js), you can pass the [default template](https://github.com/algolia/instantsearch.js/blob/1510a9439bb22dd31eb4dc0c75bd4316a57db73d/src/widgets/current-refined-values/defaultTemplates.js) to your widget (it's already imported).

src/widgets/current-refined-values/__tests__/defaultTemplates-test.js
```diff
@@ -0,0 +1,44 @@
...
+        type: 'query',
+        label: 'Query',
+        operator: ':',
+        name: 'Samsu',
+        count: 22, // should be ignored
+        cssClasses: {
+          count: 'ais-current-refined-values--count',
+        },
+      };
+      expect(defaultTemplates.item(item)).toMatchSnapshot();
```
francoischalifour: We shouldn't rely exclusively on snapshots. I think this is something we should move away from. We completely forget what we're testing.

src/widgets/current-refined-values/__tests__/defaultTemplates-test.js
```diff
@@ -0,0 +1,44 @@
...
+    });
+    it('does not show `count` when query refinement', () => {
+      const item = {
+        type: 'query',
+        label: 'Query',
+        operator: ':',
+        name: 'Samsu',
+        count: 22, // should be ignored
```
francoischalifour: This shouldn't be a comment but an assertion.

src/widgets/current-refined-values/__tests__/defaultTemplates-test.js
tkrugg: I agree with you. I hate snapshots, too for the same reasons.
However:
- The snapshotted thing here is a string so there is no very easy way to test it (no enzyme)
- the name of the test `it('wraps query refinements with <q>') can address the concern you have about forgetting the purpose of test
What do you think @francoischalifour ?

src/lib/utils.js
tkrugg: I ended up testing the defaultTemplate separately which I think is more consistent with the ways the concerns have been separated up to now. What do you think?

src/lib/__tests__/utils-test.js
```diff
@@ -342,6 +342,45 @@ describe('utils.getRefinements', () => {
...
+      },
+      { type: 'facet', attributeName: 'facet1', name: 'facet1val1' },
+    ];
+    const clearsQuery = true;
+    expect(
+      utils.getRefinements(results, helper.state, clearsQuery)
+    ).toHaveLength(2);
+    expect(
+      utils.getRefinements(results, helper.state, clearsQuery)
+    ).toContainEqual(expected[0]);
```
samouss: Why using `toContainEqual` on each elements rather than `isEqual` on the global array?

src/lib/__tests__/utils-test.js
```diff
@@ -342,6 +342,45 @@ describe('utils.getRefinements', () => {
     );
   });
 
+  it('should retrieve one facetRefinement and inject query facet if clearQuery === true', () => {
```
samouss: Since we already cover the case with facet refinements, we can focus the test only on query it will simplify it a bit.

src/widgets/current-refined-values/__tests__/defaultTemplates-test.js
samouss: You can test that the expected value are inside inside the string + snapshots. The intent of the test will be a bit clearer.

src/lib/__tests__/utils-test.js
tkrugg: isEqual adds a constraint of order which I didn't think was relevant.

tkrugg: since I simplified the tests and removed the other refinement, I now use isEqual

src/widgets/current-refined-values/__tests__/defaultTemplates-test.js
```diff
@@ -0,0 +1,40 @@
...
+      };
+      expect(defaultTemplates.item(item)).toContain('Query :   <q>Samsu</q>');
```
francoischalifour: There shouldn't be a space before the colon (`:`). Why is there so many spaces after the colon?

tkrugg: it comes from the ${computedExclude} in the returned string `${computedLabel} ${computedExclude} ${computedName} ${computedCount}`. I would change it to a regexp but I didn't think it mattered and regexp get hard to read with so many `<` `>` `/` involved.
We could write it as       `.toMatch(new RegExp("Query +: +<q>Samsu</q>"));`

PR #3225, opened 2018-10-19
Add missing '+' to prevent syntax error.

ecoffman:
Found another instance of missing '+'.


francoischalifour:
We should even convert this example to [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) like other snippets on this guide.

PR #3228, opened 2018-10-23
chore(engine): pin Node and Yarn version

tkrugg:
~upgraded node@8.12.0 and yarn@1.10.1.~
pinned yarn version in travis


tkrugg:
The motivation for doing this is avoid conflicts of integrity sha in yarn.lock. This what react-instantsearch seem to be doing.

I'm not sure however how this is going to play out for user already using instantsearch.js. Would it make there update more painful? what if they have a more recent version of node and yarn, doesn't this constraint them to downgrade?

.travis.yml
```diff
@@ -10,6 +10,7 @@ addons:
 before_install:
 - git config --global user.email "algobot@users.noreply.github.com"
 - git config --global user.name "algobot"
+- curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.10.1
```
francoischalifour: Those are not indented correctly (so are the two previous lines for some reason).

package.json
```diff
@@ -9,6 +9,10 @@
     "search",
     "vanilla"
   ],
+  "engines": {
+    "node": "8.12.0",
```
Haroenv: just make sure that this doesn't end up in the package.json that gets pushed to npm, worst case, filter it out before publishing

francoischalifour: Let's remove it since it doesn't bring value.

samouss: Not sure to understand your comment @Haroenv. Why it would have an impact on the published package? The [`engines`](https://docs.npmjs.com/files/package.json#engines) attribute will only throw an error when a script is ran on that package specifically otherwise it's only a warning. It will only impact contributors.

@francoischalifour It does bring more value because not everyone is using `nvm`. For those who don't, the attribute is useful. One question remains on the strictness of the value. We can probably use a range rather than a pinned version. It will still enforce a value without bothering too much the contributors.

PR #3234, opened 2018-10-26
fix(range-input): fix button classname

francoischalifour:
Related: algolia/instantsearch-specs#92


francoischalifour:
@Haroenv Well spotted, this is fixed.

PR #3235, opened 2018-10-26
 chore: remove CSS source

francoischalifour:
We don't use the CSS source of this project anymore but rely on InstantSearch.css.


francoischalifour:
@Haroenv Dev novel already uses InstantSearch.css: https://github.com/algolia/instantsearch.js/blob/chore/remove-css-source/dev/template.html#L16

Docgen needs InstantSearch styles only for the widget showcase, which relies on InstantSearch 2 for now. We need to publish InstantSearch 3 beta and add the styles there afterwards: https://github.com/algolia/instantsearch.js/blob/chore/remove-css-source/docgen/layouts/widget-showcase.pug#L6

Haroenv:
hmm, indeed for a separate PR, but would be nice if it could come from source instead (although it barely matters since this docs site won't live for very long anymore)

francoischalifour:
I'm not sure it's worth the trouble to be honest.

francoischalifour:
@samouss I tried removing as many top-level dependencies as possible but we actually used them for the `dev/` folder and our stories.

We can still do another pass later.

PR #3190, opened 2018-10-11
 feat(current-refinements): implement InstantSearch.css

francoischalifour:
## Summary

This implements the InstantSearch.css spec for the widget CurrentRefinements.

## Specs

https://instantsearch-css.netlify.com/widgets/current-refinements/

## Stories

https://deploy-preview-3190--algolia-instantsearch.netlify.com/v2/dev-novel/?selectedStory=CurrentRefinements.default

PR #3236, opened 2018-10-26
chore: pin Yarn and Node versions

francoischalifour:
This is the equivalent of #3228 for the branch `feat/3.0`.


Haroenv:
how come this can't be done my merging master into feat/3.0?

francoischalifour:
I was wondering at what point we want to separate the two branches. Is it too early?

I'm especially thinking about the commit 6006fe1c59992f8cb7bf2dd09e004d9999e6e4c2 which is going to be annoying to merge with #3190.

Haroenv:
It's the same for me, but with Vue InstantSearch I'm seeing annoyances now with having two completely separate branches

francoischalifour:
Makes sense. I will try to resolve conflicts from `develop` and treat this PR accordingly.

francoischalifour:
Handled in df23db14e397f0266709b9aa3b0f26d08501a0fd.

PR #3240, opened 2018-10-29
chore(src): remove decorators

francoischalifour:
`autoHideContainer` and `headerFooter` have been removed.


Haroenv:
It looks like some other PR removed the bundlesize test in the travis matrix?

francoischalifour:
@Haroenv It seems like this is a Bundlesize bug which is not triggering in this PR, right?

I will rebase to see if it updates.

PR #3245, opened 2018-10-30
docs(contributing): fix ToC links

cdtinney:


CONTRIBUTING.md
```diff
@@ -20,17 +20,17 @@ don't hesitate to leave a message on the [community forum](https://discourse.alg
...
+- [Folders of the project 🗺](#folders-of-the-project-)
   - [The source folder](#the-source-folder)
   - [The documentation source](#the-documentation-source)
   - [The dev app](#the-dev-app)
 - [Tests 🔬](#tests-)
   - [Unit tests](#unit-tests)
   - [Functional tests](#functional-tests)
 - [Linting ✨](#linting-)
-- [Release 🛳](#release-%F0%9F%9B%B3)
+- [Release 🛳](#release-)
```
Haroenv: 🤔 why did you change these links? When I try it out, the old ones work and your change doesn't 🙂 

Thanks!

francoischalifour:
It seems to work now, indeed.

The thing is that these are generated with [Doctoc](https://github.com/algolia/instantsearch.js/blob/develop/package.json#L22), which will override your changes every time the command is run again.

The easy solution that I see right now would be to either find emojis that work in links, or remove all emojis in headers.

CONTRIBUTING.md
francoischalifour: Some emojis don't work in links, this one included.

cdtinney:
@francoischalifour open issue re: emojis in headers - https://github.com/thlorenz/anchor-markdown-header/issues/36

Seems that the regex doesn't capture all emojis properly thus some work and some don't. I think the cleanest solution is to remove them entirely.

francoischalifour:
Thanks for pointing to this issue (I noticed the faulty behavior but didn't expect it to be an actual issue in the Markdown parser).

I'm for removing the emojis in header if @algolia/instantsearch-for-websites agrees.

Haroenv:
Let's just remove those emoji then, when I checked earlier the old links were working, and the new ones were, but that seemed to depend from which place on GitHub I was viewing it (the diff view did work and the main one doesn't lol). 

Let's remove the confusion, and get rid of the 2+ byte emoji

cdtinney:
PR updated with emojis removed and ToC re-generated via `doctoc`

CONTRIBUTING.md
```diff
@@ -15,29 +15,29 @@ don't hesitate to leave a message on the [community forum](https://discourse.alg
...
+- [Contributing with an new issue](#contributing-with-an-new-issue)
+- [The code contribution process](#the-code-contribution-process)
+- [Commit conventions](#commit-conventions)
+- [Branches organization](#branches-organization)
 - [Requirements](#requirements)
-- [Launch the dev environment 👩‍💻](#launch-the-dev-environment-)
+- [Launch the dev environment](#launch-the-dev-environment)
 - [Updating the examples](#updating-the-examples)
-- [Folders of the project 🗺](#folders-of-the-project-)
+- [Folders of the project 🗺](#folders-of-the-project-%F0%9F%97%BA)
```
francoischalifour: Why did we keep that emoji? It's still broken on my side.

PR #3248, opened 2018-10-30
refactor(components): update components architecture

francoischalifour:
This moves all components to their own folder for consistency.


src/components/RefinementList/RefinementListItem.js
```diff
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { Component } from 'preact-compat';
 import isEqual from 'lodash/isEqual';
-import Template from '../Template.js';
+import Template from '../Template/Template';
```
Haroenv: now that you're moving files around anyway, do you know why we use the name twice, and not index.js?

francoischalifour: I'm not sure, it's a decision made before I started contributing. The goal of this PR is mainly to make it consistent for now.

I think a next step will be to remove folders like the [React InstantSearch component architecture](https://github.com/algolia/react-instantsearch/tree/83ce24508a271e0b0c8bc8e5bcb6ccfc48a75e2b/packages/react-instantsearch-dom/src/components) (but we'll take this decision a bit later).

PR #3253, opened 2018-11-02
feat(panel): add Panel widget

francoischalifour:
The panel widget adds a way to wrap other widgets to add templates (headers, footers) and utility features (collapsing and hiding the widget).

## Motivation

Using headers and footers wrapping a widget is a common need to give more context to the user when interacting with the widget.

React InstantSearch provides a `Panel` widget to wrap any other InstantSearch component:

```js
<Panel header="All brands">
  <RefinementList attributes="brands" />
</Panel>
```

Up until InstantSearch.js 2, widgets had `header` and `footer` templates, in addition to [`autoHideContainer`](https://community.algolia.com/instantsearch.js/v2/widgets-common-api.html#autohidecontainer) and [`collapsible`](https://community.algolia.com/instantsearch.js/v2/widgets-common-api.html#collapsible) options. The panel widget solves these needs, deferring the responsability from the widget itself to the panel widget.

## Usage

```js
const refinementListWithPanel = panel({
  templates: {
    header: 'Header',
    footer: 'Footer',
    // or with functions
    header: ({ results }) => `Header | ${results.nbHits} results`,
  },
  hidden: ({ results }) => results.nbHits === 0,
})(refinementList);

search.addWidget(
  refinementListWithPanel({
    attribute: 'brands',
  })
);
```

The `hidden` property drives the `noRefinement` CSS class.

## Specs

https://instantsearch-css.netlify.com/widgets/panel/

## RFC

https://github.com/algolia/instantsearch-rfcs/pull/9

## Preview

https://deploy-preview-3253--instantsearchjs.netlify.com/stories/?selectedStory=Panel.with%20default

### DIFF
--- a/storybook/app/builtin/stories/panel.stories.js
+++ b/storybook/app/builtin/stories/panel.stories.js
@@ -13,10 +13,11 @@ export default () => {
       window.search.addWidget(
         instantsearch.widgets.panel({
           templates: {
-            header: 'Header',
+            header: ({ results }) =>
+              `Header ${results ? `| ${results.nbHits} results` : ''}`,
             footer: 'Footer',
           },
-          hidden: ({ canRefine }) => !canRefine,
+          hidden: ({ results }) => results.nbHits === 0,
         })(instantsearch.widgets.refinementList)({
           container,
           attribute: 'brand',


src/widgets/panel/panel.js
```diff
@@ -8,20 +8,26 @@ import Template from '../../components/Template/Template';
...
+    >
       {templates.header && (
         <Template
           {...templateProps}
           templateKey="header"
           rootProps={{
             className: cssClasses.header,
           }}
-          data={{ results }}
+          data={renderingOptions}
```
Haroenv: this is cool!

src/widgets/panel/panel.js
```diff
@@ -0,0 +1,145 @@
...
+      containerNode: getContainerNode(container),
+      cssClasses,
+      templates,
+      templateProps,
+    });
+
+    try {
+      const { bodyReference } = renderPanel({
+        renderingOptions: {
+          canRefine: false,
```
francoischalifour: I'm not sure if the default value should be `true` or `false`.

If ever some developer wants to add an fade-out animation when there's no refinement possible, it would flash on load with a default value set to `false`.

francoischalifour: Yep! Let's experiment a little on our side and see what makes more sense.

Keep in mind that if we were to change the default value, it would be a behavior change (and potentially a breaking change).

src/widgets/panel/panel.js
```diff
@@ -0,0 +1,116 @@
...
+
+const myWidget = widgetWithHeaderFooter(widgetOptions)`;
+
+export default function panel({
+  templates = {},
+  hidden = () => false,
+  cssClasses: userCssClasses = {},
+} = {}) {
+  if (typeof hidden !== 'function') {
+    throw new Error(
```
francoischalifour: That should probably be a warning.

src/widgets/panel/__tests__/panel-test.js
```diff
@@ -0,0 +1,46 @@
+import panel from '../panel';
+
+describe('panel call', () => {
+  test('without arguments does not throw', () => {
```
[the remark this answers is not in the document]
francoischalifour: They're similar – we use both in the code base.

src/components/Panel/Panel.js
```diff
@@ -0,0 +1,58 @@
...
+    root: PropTypes.string.isRequired,
+    noRefinementRoot: PropTypes.string.isRequired,
+    body: PropTypes.string.isRequired,
+    header: PropTypes.string.isRequired,
+    footer: PropTypes.string.isRequired,
+  }).isRequired,
+  templateProps: PropTypes.shape({
+    templates: PropTypes.object.isRequired,
+  }).isRequired,
+  hidden: PropTypes.bool.isRequired,
```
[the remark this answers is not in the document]
francoischalifour: This is the boolean attribute passed to the component, computed [here](https://github.com/algolia/instantsearch.js/pull/3253/files/719ce98005ef35bf35a0e572ab119431e59fd003#diff-29c551fae0ce99d951dbd04634b114caR107).

src/widgets/panel/__tests__/panel-test.js
```diff
@@ -0,0 +1,46 @@
...
+
+  test('with templates does not throw', () => {
+    expect(() =>
+      panel({
+        templates: { header: 'header' },
+      })
+    ).not.toThrow();
+  });
+
+  test('with a `hidden` function does not throw', () => {
```
tkrugg:
```suggestion
  test('with `hidden` a function, does not throw', () => {
```

src/widgets/panel/__tests__/panel-test.js
```diff
@@ -0,0 +1,46 @@
...
+    expect(() =>
+      panel({
+        hidden: () => true,
+      })
+    ).not.toThrow();
+  });
+
+  test('with `hidden` as a boolean warns', () => {
```
tkrugg:
```suggestion
  test('with `hidden` a boolean, warns', () => {
```

src/widgets/panel/panel.js
```diff
@@ -0,0 +1,116 @@
...
+
+          renderPanel({
+            renderingOptions,
+            hidden: Boolean(hidden(renderingOptions)),
+          });
+          widget.render(options);
+        },
+      };
+    } catch (error) {
+      throw new Error(usage);
```
tkrugg: This always reads as very abusive to me.
Correct me if I'm wrong, but the only errors the usage instructions are helpful for are type errors in props.
This means a lot of other errors simply swallowed. Doesn't this make it harder to debug?

francoischalifour: That's for sure, it's a pattern that exists in all widgets for now.

I'm working on a proposal for better error management in InstantSearch.js. This is outside of this PR's scope though.

src/components/Panel/Panel.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
samouss:
```suggestion
import React from 'preact-compat';
```

src/components/Panel/Panel.js
```diff
@@ -0,0 +1,58 @@
...
+import PropTypes from 'prop-types';
+import cx from 'classnames';
+import Template from '../Template/Template';
+
+const Panel = ({ cssClasses, hidden, templateProps, data, onRef }) => (
+  <div
+    className={cx(cssClasses.root, {
+      [cssClasses.noRefinementRoot]: !data.canRefine,
+    })}
+    hidden={hidden}
```
samouss: The `hidden` props and `data.canRefine` feels very similar to me. We can achieve the same goal with both of the options. We can use the API of `hidden` to compute the `canRefine` value, then with CSS we can hide by default the Panel `no-refinement` class. The user is able to override the style applied to the class with a CSS override.

src/widgets/panel/panel.js
```diff
@@ -0,0 +1,116 @@
...
+    footer: cx(suit({ descendantName: 'footer' }), userCssClasses.footer),
+  };
+
+  return widgetFactory => (widgetOptions = {}) => {
+    const { container } = widgetOptions;
+
+    if (!container) {
+      throw new Error(
+        `[InstantSearch.js] The \`container\` option is required in the widget "${
+          widgetFactory.name
```
samouss: Careful with this usage, until we have a dedicated development / production bundle. The name of the function is minified by UglifyJS.

src/widgets/panel/panel.js
```diff
@@ -0,0 +1,116 @@
...
+      cssClasses={cssClasses}
+      hidden={hidden}
+      templateProps={templateProps}
+      data={renderingOptions}
+      onRef={ref => (bodyRef = ref)}
+    />,
+    containerNode
+  );
+
+  return { bodyRef };
```
samouss: We heavily rely on the fact that everything is synchronous, it can become an issue at some point. But I don't think it's possible otherwise though.

src/widgets/panel/panel.js
```diff
@@ -0,0 +1,116 @@
...
+      const widget = widgetFactory({
+        ...widgetOptions,
+        container: getContainerNode(bodyRef),
+      });
+
+      return {
+        ...widget,
+        dispose() {
+          unmountComponentAtNode(getContainerNode(container));
+          widget.dispose();
```
samouss: We should check the existence of the `dispose` method it's not a required method (imagine a custom widget built by a user).

src/widgets/panel/panel.js
```diff
@@ -0,0 +1,116 @@
...
+        },
+        render(options) {
+          const renderingOptions =
+            typeof widget.getRenderingOptions === 'function'
```
samouss: We can extract this condition to avoid to do it on each render. Why don't we provide the raw options directly? We can imagine some use case where a user want to implement a clear button inside from the Panel header.

src/widgets/panel/panel.js
```diff
@@ -0,0 +1,116 @@
...
+    const templateProps = prepareTemplateProps({ defaultTemplates, templates });
+
+    const renderPanel = renderer({
+      containerNode: getContainerNode(container),
+      cssClasses,
+      templateProps,
+    });
+
+    try {
+      const { bodyRef } = renderPanel({
```
samouss: It feels odd to have to render this outside of the widget init / render lifecycle. It would be nice to always trigger render at this phase, now some elements will start to render before we call `start` on InstantSearch.

src/components/Panel/Panel.js
francoischalifour: So you mean that the `hidden` option could drive the `canRefine` value?

As an example:

```js
panel({
  hidden: () => true, // → cannot refine
})

panel({
  hidden: () => false, // → can refine
})
```

Default rendering options:

```js
{ canRefine: !hidden }
```

I'm not sure to agree here. Hiding and disabling don't solve the same purposes.

src/widgets/panel/panel.js
francoischalifour: When do you think it could become an issue?

src/widgets/panel/panel.js
francoischalifour: Good point!

src/widgets/panel/panel.js
francoischalifour: I see what you mean. The thing is that this kind of circular dependency is necessary to know where to insert the widget.

Another proposal was to render the widget and then the panel, but this comes with a multitude of other problems.

src/widgets/panel/panel.js
francoischalifour: You're right!

src/widgets/panel/panel.js
francoischalifour: Yeah I agree, I originally wanted to go that way (we could also leverage `userData`, etc.).

@Haroenv thought it was giving the user too much control with a not-so-defined API.

Both arguments are valid, although giving the raw options would be way easier on our side (we wouldn't even need to implement the Panel spec for all widgets).

francoischalifour: Agreed to go back to the full options.

francoischalifour:
The last commit rollbacks to the original behavior:

- we give the full options to the templates and to the `hidden` option
- the evaluated `hidden` option (`() => true`) drives the `noRefinement` CSS class.

samouss:
I didn't realise from the review that we don't have access to the render options inside the Panel but only to the raw render option provided by InstantSearch not the connector. The current implementation is nice but not good enough IMO, we should have access to the same set of options that the wrapped widget provide.

samouss:
> Do you more insights on how you would see it working @samouss?

We the current architecture I have no idea how it would work. 

> AFAIK we never compute the autoHideContainer value anywhere for each widget.

We do compute the autoHideContainer inside each widget (see [RefinementList](https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/refinement-list/refinement-list.js#L68), [Menu](https://github.com/algolia/instantsearch.js/blob/develop/src/widgets/menu/menu.js#L54), ...).

francoischalifour:
Now that we have the concept of UI state in v4, we could pass the `uiState` to the panel so that it's easier to compute the `hidden` and other values.

Haroenv:
good point @francoischalifour, open to this change

PR #3255, opened 2018-11-06
feat(breadcrumb): implement Panel specs

francoischalifour:
This implements the panel specs (#3253) for the `breadcrumb` widget.

## Stories

https://deploy-preview-3255--instantsearchjs.netlify.com/v2/dev-novel/?selectedStory=Panel.with%20breadcrumb


src/connectors/breadcrumb/connectBreadcrumb.js
```diff
@@ -122,7 +140,7 @@ export default function connectBreadcrumb(renderFn, unmountFn) {
           );
         };
 
-        this._refine = function(facetValue) {
+        renderingRefine = function(facetValue) {
```
tkrugg: can be an arrow function.
I think it's best to use them when possible as they don't force you to think about context.

francoischalifour:
Closing because passing the full `options` object to template functions and not only rendering options.

PR #3259, opened 2018-11-12
fix(docs): fix dead links in routing.html

tkrugg:
** Summary **
1. Relative routing does not work as expected because there is `<base href="/v2/" />` at the top.
Added routing.html in the url path to fix it

2.  `#history-router-api` was not the correct anchor. Replaced it with `#instantsearchroutershistory-api`

### DIFF
--- a/docgen/src/guides/routing.md
+++ b/docgen/src/guides/routing.md
@@ -152,7 +152,7 @@ const search = instantsearch({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


francoischalifour:
See preview: https://deploy-preview-3259--algolia-instantsearch.netlify.com/v2/guides/routing.html

PR #3260, opened 2018-11-12
feat(bundle): update bundle strategy

francoischalifour:
This updates the bundle strategy.

## Migrate from Webpack to Rollup

Rollup is now used to bundle InstantSearch.js. Rollup is more suited for libraries. It allows us to create lighter bundles and multiple bundle strategies.

| | Webpack | Rollup
|---|---|---
| Bundle | 372 kB | 301 kB
| GZipped | 97 kB | 84 kB

This sizes are without any further optimizations (will come gradually).

## Different bundles

This is what is exported:

```
dist
├── instantsearch.development.js
├── instantsearch.development.js.map
├── instantsearch.production.min.js
└── instantsearch.production.min.js.map
cjs
├── components
├── connectors
├── helpers
├── index.js
├── lib
├── src
└── widgets
es
├── components
├── connectors
├── helpers
├── index.js
├── lib
└── widgets
```

<details>
  <summary>See outdated (probably for InstantSearch.js version 4)</summary>

```
dist
├── cjs
│   ├── components
│   ├── connectors
│   ├── helpers
│   ├── index.js
│   ├── lib
│   └── widgets
├── es
│   ├── components
│   ├── connectors
│   ├── helpers
│   ├── index.js
│   ├── lib
│   └── widgets
└── umd
    ├── instantsearch.development.js
    ├── instantsearch.development.js.map
    ├── instantsearch.production.min.js
    └── instantsearch.production.min.js.map
```

</details>

### UMD

The files have been renamed to become clearer for the users.

#### Usage

```html
<script src="instantsearch.production.min.js"></script>
```

#### Usage

```js
const instantsearch = require('instantsearch.js').default
```

### ES

No changes.

<details>
  <summary>See outdated (probably for InstantSearch.js version 4)</summary>

The ES bundle now exports `instantsearch`, and all widgets, connectors and helpers (`highlight` and `snippet`) at the root level.

#### Usage

```js
import instantsearch, {
  searchbox,
  connectHits,
  highlight,
} from 'instantsearch.js';

search.addWidget(searchbox());

search.start();
```

The developer won't have to specify the old path:

```js
import { searchbox } from 'instantsearch.js/es/widgets';
```

This is a pattern that we started with React InstantSearch ([see new packages](https://community.algolia.com/react-instantsearch/guide/Migration_guide_package_import.html)).

</details>

## Development and production bundles

Since the bundle is getting heavy, the migration to Rollup has made it easier to create separate UMD bundles:

- `instantsearch.development.js`
- `instantsearch.production.min.js`

These names were inspired by the [React strategy](https://unpkg.com/react-dom@16.6.1/umd/).

The development bundle will include everything that the default bundle has so far. The production bundle will not include the warnings (everything that is runtime) and the verbose widget/connector usages.

Everything under this condition will get stripped in the production bundle:

```js
if (__DEV__) {
  // development only
}
```

In the CJS and ES bundle, it's converted to:

```js
if (process.env.NODE_ENV === 'development') {
  // development only
}
```

## What's next

This new bundle strategy allows to gradually remove development feature from the production bundle.


francoischalifour:
A Babel plugin seems to be missing for the tests to pass as previously. I would appreciate your help to find how to make them pass!

scripts/rollup.config.js
```diff
@@ -0,0 +1,70 @@
...
+    exclude: 'node_modules/**',
+    plugins: [
+      'external-helpers',
+      'transform-react-remove-prop-types',
+      'transform-react-constant-elements',
+      'transform-react-pure-class-to-function',
+    ],
+  }),
+  commonjs(),
+  globals(),
```
Haroenv: why do we need globals? will this put a polyfill of `process` etc in the bundle?

scripts/rollup.config.js
```diff
@@ -0,0 +1,70 @@
...
+  plugins: plugins.concat(
+    [
+      minify &&
+        uglify({
+          output: {
+            preamble: license,
+          },
+        }),
+      replace({
+        __DEV__: mode === 'development',
```
Haroenv: this is a nice pattern, I'll use that instead of the tricky `process.env` we have in Vue InstantSearch

scripts/rollup.config.js
francoischalifour: You're right. You don't need them with the `__DEV__` global.

.babelrc
```diff
@@ -1,50 +1,88 @@
...
+            }
           }
-        }]
+        ],
+        "react",
+        "stage-2"
       ]
     },
-    "development": {
+    "umd": {
```
samouss: Why do we need the `umd` since it's the same as `es`?

.babelrc
```diff
@@ -1,50 +1,88 @@
...
+            }
+          }
+        ],
         "react",
         "stage-2"
       ],
       "plugins": [
-        "transform-react-remove-prop-types",
-        "transform-react-constant-elements",
-        "transform-react-pure-class-to-function"
```
samouss: Why did we remove those?

.babelrc
```diff
@@ -1,50 +1,88 @@
...
+              "browsers": ["last 2 versions", "ie >= 9"]
+            }
+          }
+        ],
         "react",
         "stage-2"
-      ],
-      "plugins": [
-        "babel-plugin-rewire",
-        ["module-resolver", {
```
samouss: Why do we remove this from the `test` env?

.babelrc
```diff
@@ -1,50 +1,88 @@
...
+        ],
         "react",
         "stage-2"
       ],
       "plugins": [
-        "transform-react-remove-prop-types",
-        "transform-react-constant-elements",
-        "transform-react-pure-class-to-function"
+        [
+          "inline-replace-variables",
```
samouss: It's only useful for the `es` env?

package.json
```diff
@@ -13,37 +13,36 @@
   "author": "Algolia <support@algolia.com>",
   "license": "MIT",
   "repository": "algolia/instantsearch.js",
-  "main": "dist-es5-module/index.js",
-  "module": "./es/index.js",
-  "jsdelivr": "dist/instantsearch.min.js",
-  "sideEffects": false,
```
samouss: Why did we remove this?

scripts/rollup.config.js
```diff
@@ -0,0 +1,67 @@
...
+    ],
+  }),
+  commonjs(),
+  filesize({
+    showMinifiedSize: false,
+    showGzippedSize: true,
+  }),
+];
+
+const createConfiguration = ({ mode = 'production', minify = true } = {}) => ({
```
samouss: It seems that the minification depends on the mode so we can drop the minify options.

.babelrc
francoischalifour: They're not the same: the ES bundle use the `inline-replace-variables` plugin. However, `es` and `cjs` are the same.

.babelrc
francoischalifour: I added them to the Rollup configuration instead. But you're right, we need them in CJS and ES too.

.babelrc
francoischalifour: That's why the tests were failing, thank you!

.babelrc
francoischalifour: Yes (`es` env is used for `cjs` and `es`).

In UMD, we replace `__DEV__` depending on the bundle type with Rollup. In CJS and ES, we replace `__DEV__` depending on the Node environment.

package.json
francoischalifour: `sideEffects` is only read by Webpack, right? (not Rollup)

scripts/rollup.config.js
francoischalifour: I was waiting for you to agree on the bundle names. If that's OK for you, I can simplify the function.

package.json
samouss: Yes but it's for consumers of the lib not us.

francoischalifour: Ah I see!

francoischalifour:
@samouss Yes, I created an app for each bundle. I also updated the PR description to add each usage.

package.json
```diff
@@ -13,37 +13,37 @@
...
   "sideEffects": false,
   "scripts": {
-    "build": "./scripts/build.sh",
     "dev": "./scripts/dev.sh",
-    "dev-novel": "NODE_ENV=development webpack-dev-server --config dev/webpack.config.js",
-    "doctoc": "doctoc --no-title --maxlevel 3 README.md CONTRIBUTING.md",
+    "build": "yarn build:cjs && yarn build:es && yarn build:umd",
+    "build:umd": "BABEL_ENV=umd rollup --config scripts/rollup.config.js",
+    "build:cjs": "BABEL_ENV=cjs babel src --out-dir dist/cjs --ignore src/index.es.js,__tests__,__mocks__ --quiet",
+    "build:es": "BABEL_ENV=es babel src --out-dir dist/es --ignore src/index.es.js,__tests__,__mocks__ --quiet && babel src/index.es.js --out-file dist/es/index.js --quiet",
```
samouss: The `BABEL_ENV` is not set for the second `babel` call. It means that the main entry is not ESM compliant. See [output](https://deploy-preview-3260--instantsearchjs.netlify.com/v2/lib/es/index.js).

src/index.es.js
```diff
@@ -0,0 +1,37 @@
...
+To import the connectors:
+
+import { connectSearchBox } from 'instantsearch.js'`
+    );
+  },
+});
+
+export * from './helpers';
+export * from './connectors';
+export * from './widgets';
```
samouss: We have to document those changes inside the migration guide.

francoischalifour: I think we never documented this usage. I will add a section in the migration guide. And we can then promote this usage in the documentation.

package.json
francoischalifour: That's right, thank you!

francoischalifour: That's right, thank you!

francoischalifour:
The branch has been updated to not rename the CJS and ES bundle in order to facilitate the usage in other libraries for now (the original post has been updated).

Happy to review this PR as a group since the changes can seem quite consequent.

docgen/src/guides/v3-migration.md
```diff
@@ -41,6 +41,20 @@ dist
 └── instantsearch.production.min.js.map
 ```
 
+### CJS
```
Haroenv: ### CommonJS

francoischalifour:
:tada:

PR #3263, opened 2018-11-13
feat(refinement-list): implement InstantSearch.css to searchbox

francoischalifour:
This implements InstantSearch.css for the `refinementList` searchbox.

At some point we should refactor the `searchbox` widget to use a single searchbox.

## Specs

https://instantsearch-css.netlify.com/widgets/refinement-list/

## Stories

https://deploy-preview-3263--instantsearchjs.netlify.com/v2/dev-novel/?selectedStory=RefinementList.with%20search%20inside%20items


src/components/SearchBox/Searchbox.js
```diff
@@ -0,0 +1,134 @@
...
+              className={cssClasses.resetIcon}
+              xmlns="http://www.w3.org/2000/svg"
+              viewBox="0 0 20 20"
+              width="10"
+              height="10"
+            >
+              <path d="M8.114 10L.944 2.83 0 1.885 1.886 0l.943.943L10 8.113l7.17-7.17.944-.943L20 1.886l-.943.943-7.17 7.17 7.17 7.17.943.944L18.114 20l-.943-.943-7.17-7.17-7.17 7.17-.944.943L0 18.114l.943-.943L8.113 10z" />
+            </svg>
+          </button>
+          <span className={cssClasses.loadingIndicator} hidden>
```
Haroenv: should it always be hidden?

francoischalifour: For now yes, there's no way to get this information from the refinement list.

We can remove it, or leave it as is waiting for the searchbox refactor.

Haroenv: ah I understand, can be left as-is then for now :)

src/components/SearchBox/Searchbox.js
```diff
@@ -0,0 +1,134 @@
...
+              viewBox="0 0 40 40"
+            >
+              <path d="M26.804 29.01c-2.832 2.34-6.465 3.746-10.426 3.746C7.333 32.756 0 25.424 0 16.378 0 7.333 7.333 0 16.378 0c9.046 0 16.378 7.333 16.378 16.378 0 3.96-1.406 7.594-3.746 10.426l10.534 10.534c.607.607.61 1.59-.004 2.202-.61.61-1.597.61-2.202.004L26.804 29.01zm-10.426.627c7.323 0 13.26-5.936 13.26-13.26 0-7.32-5.937-13.257-13.26-13.257C9.056 3.12 3.12 9.056 3.12 16.378c0 7.323 5.936 13.26 13.258 13.26z" />
+            </svg>
+          </button>
+          <button
+            className={cssClasses.reset}
+            type="reset"
+            title="Clear the search query."
+            hidden
```
samouss: The reset button should not always be hidden, but only when the query is empty.

samouss:
@francoischalifour since the input is uncontrolled the only way to test it is to directly access the input and set a value on it. Indeed they don't have an a proper API to do it but you can probably get a reference to the node and set the value directly on it.

PR #3272, opened 2018-11-21
fix(breadcrumb): rename noRefinement to noRefinementRoot

francoischalifour:
The CSS class was wrongly named `noRefinement` instead of `noRefinementRoot`.

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.js
+++ b/src/components/Breadcrumb/Breadcrumb.js
@@ -51,7 +51,7 @@ const Breadcrumb = ({
 }) => (
   <div
     className={cx(cssClasses.root, {
-      [cssClasses.noRefinement]: items.length > 0,
+      [cssClasses.noRefinementRoot]: items.length > 0,
     })}
   >
     <ul className={cssClasses.list}>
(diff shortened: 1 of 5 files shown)


src/widgets/breadcrumb/breadcrumb.js
```diff
@@ -146,9 +146,9 @@ export default function breadcrumb({
 
   const cssClasses = {
     root: cx(suit(), userCssClasses.root),
-    noRefinement: cx(
-      suit({ modifierName: 'noRefinement' }),
-      userCssClasses.noRefinement
+    noRefinementRoot: cx(
+      suit({ modifierName: 'noRefinementRoot' }),
```
samouss: The name of the modifier should still be `noRefinement` rather than `noRefinementRoot`.

francoischalifour: Nice catch!

src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -51,7 +51,7 @@ const Breadcrumb = ({
 }) => (
   <div
     className={cx(cssClasses.root, {
-      [cssClasses.noRefinement]: items.length > 0,
+      [cssClasses.noRefinementRoot]: items.length > 0,
```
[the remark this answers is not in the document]
francoischalifour: You're right, we seem to have missed that. Will fix in this PR, thanks!

PR #3273, opened 2018-11-21
fix(breadcrumb): rename item's "name" to "label"

francoischalifour:
This renames the breadcrumb items `name` property to `label`.

This is the naming we're going for in v3, as well as [the current naming in React InstantSearch](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-core/src/connectors/connectBreadcrumb.js#L20).


francoischalifour:
@Haroenv Right, I added a note.

PR #3275, opened 2018-11-21
fix(pagination): rename `maxPages` to `totalPages`

francoischalifour:
This renames the previous `maxPages` to `totalPages` ([see Pagination spec](https://instantsearch-css.netlify.com/widgets/pagination/)).


francoischalifour:
@Haroenv It was already mentioned in the migration guide. We forgot to rename it in the source code.

See: https://github.com/algolia/instantsearch.js/blob/5f50c608d6654ad4fa7985cb7d6bf24ea4afdbc2/docgen/src/guides/v3-migration.md#options-9

PR #3276, opened 2018-11-21
ci: use vm-based travis

Haroenv:
this will become the default, so let's see if this breaks now https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration


Haroenv:
back in the day this was a way to make travis use containers instead of on the machine directly, so people used it as default I think

PR #3280, opened 2018-11-23
docs(v3): add migration guide to menus

francoischalifour:
* Add the "Upgrade to InstantSearch 3" guide in the menus
* Remove the "Prepare for v3" guide

## Guide

https://deploy-preview-3280--instantsearchjs.netlify.com/guides/v3-migration.html

## Preview

image


samouss:
I don't think we should remove the migration guide from v1 to v2. We still have users on support asking about question on the v1 of InstantSearch. We don't "support" the v1 but people might want to upgrade from v1.

francoischalifour:
@samouss I meant exclusively on the v3 website.

Haroenv:
I think even then the guide should still exist actually, there are still v1 -> v2 -> v3 use cases (e.g. shopify)

francoischalifour:
@Haroenv In this case, wouldn't it make sense for users to follow this?

1. go on the v1 website and follow the guide: "Migrate to v2"
1. go on the v2 website and follow the guide: "Migrate to v3"

We don't plan to write a guide v1 → v3 directly.

PR #3301, opened 2018-11-26
docs(migration): some fixes

Haroenv:
1. heading level for the new/old usage inside InstantSearch section
2. add frontmatter to make it show up on the right places on the website
3. some typos I spotted
4. syntax highlighting

see /v2/guides/v3-migration.html for the difference


francoischalifour:
Some of these changes have been addressed in #3280.

We should probably merge #3280 first and then apply your changes. WDYT?

PR #3303, opened 2018-11-26
fix(toggleRefinement): provide `name` to templates

francoischalifour:
## Context

> The default template uses the `name` attribute but it’s not provided to the template. It renders something blank on the page. We should find a “good“ default value or we make the `template.labelText` required. Previously it was to the label that was a required option.
> – @samouss 

## This solution

We pass the attribute (called `name`) to the value passed by the connector `connectToggleRefinement`. The refinement now has a `name` property usable in the template.

If your attribute is called `free_shipping`, the default template will display "free_shipping". You will then need to specify `templates.labelText` to change it to "Free shipping".

## Stories

https://deploy-preview-3303--instantsearchjs.netlify.com/stories/?selectedStory=ToggleRefinement.default

### DIFF
--- a/docgen/src/guides/v3-migration.md
+++ b/docgen/src/guides/v3-migration.md
@@ -1252,7 +1252,7 @@ const sortByItem = {
 
 `collapsible` and `autoHideContainer` options have been removed. These options are now implemented as part of the Panel widget wrapper.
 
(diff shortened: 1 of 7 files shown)


francoischalifour:
@Haroenv I added more explanations in the migration guide.

PR #3317, opened 2018-11-29
feat(ratingMenu): merge labels and templates

francoischalifour:
This merges the labels and the templates for the widget `ratingMenu`.

We now pass the templates to the templates themself to access `templates.andUp`.

### DIFF
--- a/docgen/src/guides/v3-migration.md
+++ b/docgen/src/guides/v3-migration.md
@@ -918,7 +918,9 @@ Widget removed.
 | Before          | After             |
 | --------------- | ----------------- |
 | `attributeName` | `attribute`       |
-| `labels.andUp`  | `templates.andUp` |
+| `labels.andUp`  | Removed           |
+
+The value for the label `andUp` is now inlined inside `templates.item`.
 
 ### CSS classes
 
--- a/docgen/src/guides/v3-migration.md
+++ b/docgen/src/guides/v3-migration.md
@@ -915,9 +915,10 @@ Widget removed.
 
 ### Options
 
-| Before          | After       |
-| --------------- | ----------- |
-| `attributeName` | `attribute` |
+| Before          | After             |
+| --------------- | ----------------- |
+| `attributeName` | `attribute`       |
+| `labels.andUp`  | `templates.andUp` |
 
 ### CSS classes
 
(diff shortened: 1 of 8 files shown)


src/widgets/rating-menu/__tests__/__snapshots__/rating-menu-test.js.snap
```diff
@@ -52,14 +56,21 @@ exports[`ratingMenu() calls twice ReactDOM.render(<RefinementList props />, cont
           false,
           false,
         ],
+        "templates": Object {
+          "andUp": "& Up",
+          "item": "{{#count}}<a class=\\"{{cssClasses.link}}\\" aria-label=\\"{{value}} & up\\" href=\\"{{href}}\\">{{/count}}{{^count}}<div class=\\"{{cssClasses.link}}\\" aria-label=\\"{{value}} & up\\" disabled>{{/count}}
```
Haroenv: looking at this like this again, to me it would make sense if the & up was simply part of the item template. If you want to change item, you likely are fine with writing `& Up` yourself. What do you think? Sure, it's a bit more work if you only wan to change that text, but it's not really that bad probably 🤔

francoischalifour: Yeah I thought about it. Don't have any metrics on how many people change only the `andUp` template so I don't have any strong opinion.

samouss: Fair point, inside the next documentation we have the default template part of the snippet so it would even easier to extend. Just c/c the snippet update the label and you're done. WDYT?

francoischalifour: Let's go for that!

francoischalifour:
Will need to merge `item` and `andUp` templates.

storybook/app/init-unmount-widgets.js
```diff
@@ -262,7 +262,7 @@ export default () => {
         container,
         attribute: 'rating',
         max: 5,
-        labels: {
+        templates: {
```
samouss: This template doesn't exist anymore.

docgen/src/guides/v3-migration.md
```diff
@@ -915,9 +915,10 @@ Widget removed.
 
 ### Options
 
-| Before          | After       |
-| --------------- | ----------- |
-| `attributeName` | `attribute` |
+| Before          | After             |
+| --------------- | ----------------- |
+| `attributeName` | `attribute`       |
+| `labels.andUp`  | `templates.andUp` |
```
samouss:
```suggestion
| `labels.andUp`  | Removed |
```

PR #3318, opened 2018-11-29
feat(hierarchicalMenu): merge showMore templates

francoischalifour:
@samouss and I decided to merge `showMoreActive` and `showMoreInactive` by creating a `showMoreText` template which is passed the argument `isShowingMore`.

## Previous usage

```js
instantsearch.widgets.hierarchicalMenu({
  // ...
  templates: {
    showMoreActive: 'Show less',
    showMoreInactive: 'Show more',
  },
});
```

## New usage

```js
instantsearch.widgets.hierarchicalMenu({
  // ...
  templates: {
    showMoreText: `
      {{#isShowingMore}}
        Show less
      {{/isShowingMore}}
      {{^isShowingMore}}
        Show more
      {{/isShowingMore}}
    `,
  },
});
```

## Stories

https://deploy-preview-3318--instantsearchjs.netlify.com/stories/?selectedStory=HierarchicalMenu.with%20show%20more

PR #3329, opened 2018-12-10
feat(refinementList): merge showMore template

francoischalifour:
This implements #3318 for the `refinementList` widget.

It also fixes an error thrown where no templates are given but the `showMore` option is used, by giving a default `showMoreText` template (like for `hierarchicalMenu` and `menu`).


francoischalifour:
@samouss The migration guide is already updated.

PR #3335, opened 2018-12-10
feat(autocomplete): default escapeHTML to true

francoischalifour:
- Rename `escapeHits` to `escapeHTML`
- Default `escapeHTML` to true

### DIFF
--- a/src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
+++ b/src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
@@ -37,6 +37,17 @@ describe('connectAutocomplete', () => {
     expect(renderFn.mock.calls[1][1]).toBeFalsy();
   });
 
+  it('sets the default configuration', () => {
+    const renderFn = jest.fn();
+    const makeWidget = connectAutocomplete(renderFn);
+    const widget = makeWidget();
+
+    expect(widget.getConfiguration()).toEqual({
+      highlightPreTag: TAG_PLACEHOLDER.highlightPreTag,
+      highlightPostTag: TAG_PLACEHOLDER.highlightPostTag,
+    });
+  });
+
   it('creates derived helper', () => {
     const renderFn = jest.fn();
     const makeWidget = connectAutocomplete(renderFn);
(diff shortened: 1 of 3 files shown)


Haroenv:
is this also requiring the helper's usage, like in hits now? cc @samouss

francoischalifour:
What do you mean?

Haroenv:
https://github.com/algolia/instantsearch.js/pull/3323

samouss:
Nope we did not enforce the helper usage yet (even inside `hits`, `infiniteHits`, etc...).

francoischalifour:
Right, I thought you were talking about the JavaScript Helper, @Haroenv.

We may want to call them "formatting helpers" to avoid confusion.

src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
```diff
@@ -69,4 +70,84 @@ describe('connectAutocomplete', () => {
...
+      results,
+      state: helper.state,
+      helper,
+      createURL: () => '#',
+    });
+
+    const rendering = renderFn.mock.calls[1][0];
+
+    expect(rendering.indices[0].hits).toEqual(
+      expect.arrayContaining(escapedHits)
```
francoischalifour: `expect.arrayContaining` is necessary because the structure is not exactly the same. I think the reason is because we internally set the `__escape` object to `true` on the `hits` array after we escape the hits.

I can add a comment about it.

src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
```diff
@@ -69,4 +70,84 @@ describe('connectAutocomplete', () => {
...
+    widget.render({
+      results,
+      state: helper.state,
+      helper,
+      createURL: () => '#',
+    });
+
+    const rendering = renderFn.mock.calls[1][0];
+
+    expect(rendering.indices[0].hits).toEqual(expect.arrayContaining(hits));
```
francoischalifour: I don't think `expect.arrayContaining` is necessary here.

src/connectors/autocomplete/connectAutocomplete.js
```diff
@@ -1,4 +1,4 @@
-import escapeHits, { TAG_PLACEHOLDER } from '../../lib/escape-highlight';
+import escapeHTML, { TAG_PLACEHOLDER } from '../../lib/escape-highlight';
```
samouss:
```suggestion
import escapeHits, { TAG_PLACEHOLDER } from '../../lib/escape-highlight';
```

PR #3339, opened 2018-12-12
fix(configure): make lifecycle functions optional

francoischalifour:
This makes the render and unmount lifecycle functions optional for the `configure` widget. It's more straightforward to use `connectConfigure`.

### DIFF
--- a/src/connectors/configure/__tests__/connectConfigure-test.js
+++ b/src/connectors/configure/__tests__/connectConfigure-test.js
@@ -14,12 +14,12 @@ describe('connectConfigure', () => {
   });
 
   describe('Usage', () => {
-    it('without searchParameters throws', () => {
+    it('throws without searchParameters', () => {
       const makeWidget = connectConfigure();
       expect(() => makeWidget()).toThrow();
     });
 
-    it('with a render function but no unmount function does not throw', () => {
+    it('does not throw with a render function but without an unmount function', () => {
       expect(() => connectConfigure(jest.fn(), undefined)).not.toThrow();
     });
 
@@ -27,8 +27,8 @@ describe('connectConfigure', () => {
       expect(() => connectConfigure(undefined, jest.fn())).not.toThrow();
     });
 
-    it('without unmount and render functions does not throw', () => {
-      expect(() => connectConfigure(undefined, jest.fn())).not.toThrow();
+    it('does not throw without render and unmount functions', () => {
+      expect(() => connectConfigure(undefined, undefined)).not.toThrow();
     });
   });
 
--- a/src/connectors/configure/__tests__/connectConfigure-test.js
+++ b/src/connectors/configure/__tests__/connectConfigure-test.js
@@ -13,18 +13,22 @@ describe('connectConfigure', () => {
     helper = algoliasearchHelper(fakeClient, '', {});
   });
 
-  describe('throws on bad usage', () => {
-    it('without searchParameters', () => {
+  describe('Usage', () => {
+    it('without searchParameters throws', () => {
       const makeWidget = connectConfigure();
       expect(() => makeWidget()).toThrow();
     });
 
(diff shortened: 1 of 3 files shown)


src/connectors/configure/__tests__/connectConfigure-test.js
```diff
@@ -13,18 +13,22 @@ describe('connectConfigure', () => {
     helper = algoliasearchHelper(fakeClient, '', {});
   });
 
-  describe('throws on bad usage', () => {
-    it('without searchParameters', () => {
+  describe('Usage', () => {
+    it('without searchParameters throws', () => {
```
samouss:
```suggestion
    it('throws without searchParameters', () => {
```

src/connectors/configure/__tests__/connectConfigure-test.js
```diff
@@ -13,18 +13,22 @@ describe('connectConfigure', () => {
...
       const makeWidget = connectConfigure();
       expect(() => makeWidget()).toThrow();
     });
 
-    it('with a renderFn but no unmountFn', () => {
-      expect(() => connectConfigure(jest.fn(), undefined)).toThrow();
+    it('with a render function but no unmount function does not throw', () => {
```
samouss:
```suggestion
    it('does not throw with a render function but without an unmount function', () => {
```

src/connectors/configure/__tests__/connectConfigure-test.js
```diff
@@ -13,18 +13,22 @@ describe('connectConfigure', () => {
...
+      expect(() => connectConfigure(jest.fn(), undefined)).not.toThrow();
     });
 
-    it('with a unmountFn but no renderFn', () => {
-      expect(() => connectConfigure(undefined, jest.fn())).toThrow();
+    it('with a unmount function but no render function does not throw', () => {
```
samouss: I don't think that case make sense. We should never be in that case.

src/connectors/configure/__tests__/connectConfigure-test.js
```diff
@@ -13,18 +13,22 @@ describe('connectConfigure', () => {
...
+      expect(() => connectConfigure(undefined, jest.fn())).not.toThrow();
+    });
+
+    it('without unmount and render functions does not throw', () => {
```
samouss:
```suggestion
    it('does not throw without render and unmount functions', () => {
```

src/connectors/configure/__tests__/connectConfigure-test.js
```diff
@@ -13,18 +13,22 @@ describe('connectConfigure', () => {
...
+      expect(() => connectConfigure(undefined, jest.fn())).not.toThrow();
```
samouss:
```suggestion
      expect(() => connectConfigure(undefined, undefined)).not.toThrow();
```

src/connectors/configure/connectConfigure.js
```diff
@@ -1,18 +1,21 @@
...
-    //   widgetParams
+    //   widgetParams,
     // }
-  },
-  function disposeFn() {}
-)
+  }
+);
+search.addWidget(
+  customConfigure({})
```
samouss: We should add `searchParameters` in the usage because it's a required parameter.

src/connectors/configure/connectConfigure.js
```diff
@@ -1,18 +1,21 @@
-import isFunction from 'lodash/isFunction';
+import noop from 'lodash/noop';
```
Haroenv: another time we should either converge on using the lodash or a local noop, it's inconsistent now :)

src/connectors/configure/__tests__/connectConfigure-test.js
francoischalifour: Indeed it doesn't really make sense but it's still a case that should be covered, right?

PR #3340, opened 2018-12-12
fix(connectHits): default escapeHTML to true

francoischalifour:
This makes `connectHits` `escapeHTML` default to `true` at the connector level rather than at the widget level.

It escapes by default the custom hits user implementations.

### DIFF
--- a/src/connectors/hits/connectHits.js
+++ b/src/connectors/hits/connectHits.js
@@ -1,4 +1,4 @@
-import escapeHTML, { TAG_PLACEHOLDER } from '../../lib/escape-highlight.js';
+import escapeHits, { TAG_PLACEHOLDER } from '../../lib/escape-highlight.js';
 import { checkRendering } from '../../lib/utils.js';
 
 const usage = `Usage:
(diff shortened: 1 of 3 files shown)


src/connectors/hits/connectHits.js
```diff
@@ -82,12 +82,8 @@ export default function connectHits(renderFn, unmountFn) {
...
 
       render({ results, instantSearchInstance }) {
-        if (
-          widgetParams.escapeHTML &&
-          results.hits &&
-          results.hits.length > 0
-        ) {
-          results.hits = escapeHTML(results.hits);
+        if (escapeHTML && results.hits && results.hits.length > 0) {
+          results.hits = escapeHits(results.hits);
```
Haroenv: I remember a bug when trying to SSR in Vue due to this rewriting. @samouss, do you remember the context? Maybe we can prevent the modification here?

samouss: The issue was on our side because we did not have the same configuration on the server and the client. But it happen because we cache the escaped hits directly on the results (mutate the results actually). 

We can solve the issue later to force the usage of the highlight helper. We won't have to escape the hits inside the connector but only at the highlight helper level. Note that we should still have the `escapeHTML` option to provide the placeholder tag.

PR #3343, opened 2018-12-13
 feat(hierarchicalMenu): update "show more" behavior

francoischalifour:
- The show more button displays when the option `showMore` is `true`
- `limit` defaults to 10 (no change)
- `showMoreLimit` defaults to 20
- Throw an error if `showMoreLimit <= limit`

### DIFF
--- a/src/connectors/hierarchical-menu/connectHierarchicalMenu.js
+++ b/src/connectors/hierarchical-menu/connectHierarchicalMenu.js
@@ -19,6 +19,8 @@ search.addWidget(
     [ rootPath = null ],
     [ showParentLevel = true ],
     [ limit = 10 ],
+    [ showMore = false ],
+    [ showMoreLimit = limit * 2 ],
     [ sortBy = ['name:asc'] ],
     [ transformItems ],
   })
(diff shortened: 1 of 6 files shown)


src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -82,7 +86,8 @@ export default function connectHierarchicalMenu(renderFn, unmountFn) {
       rootPath = null,
       showParentLevel = true,
       limit = 10,
-      showMoreLimit,
+      showMore,
```
samouss: It should be `false` by default.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -149,22 +158,13 @@ export default function connectHierarchicalMenu(renderFn, unmountFn) {
...
-              showMoreLimit
-            );
-          }
-        }
+        const currentMaxValuesPerFacet =
+          currentConfiguration.maxValuesPerFacet || 0;
+
+        widgetConfiguration.maxValuesPerFacet = Math.max(
+          currentMaxValuesPerFacet,
+          showMoreLimit
```
samouss: We always use the max number for the limit. It means that we always over fetch even if the user don't set `showMore`. We could provide a more accurate value based on the `showMore` option.

```
```

francoischalifour: Good point!

src/widgets/hierarchical-menu/hierarchical-menu.js
```diff
@@ -90,10 +93,12 @@ hierarchicalMenu({
...
  * @property {string[]} attributes Array of attributes to use to generate the hierarchy of the menu.
- * @property {number} [limit=10] How much facet values to get.
- * @property {string} [separator=" > "] Separator used in the attributes to separate level values.
+ * @property {string} [separator = " > "] Separator used in the attributes to separate level values.
  * @property {string} [rootPath] Prefix path to use if the first level is not the root level.
- * @property {boolean} [showParentLevel=true] Show the siblings of the selected parent level of the current refined value. This
+ * @property {boolean} [showParentLevel = true] Show the siblings of the selected parent level of the current refined value. This
+ * @property {number} [limit = 10] Max number of values to display.
+ * @property {boolean} [showMore = false] Whether to display the "show more" button.
+ * @property {number} [showMoreLimit = 10] Max number of values to display when showing more.
```
samouss:
```suggestion
 * @property {number} [showMoreLimit = 20] Max number of values to display when showing more.
```

PR #3347, opened 2018-12-14
fix(ratingMenu): change context-based implementation

francoischalifour:
## The issue

The `connectRatingMenu` connector internally relies on `this` for both `createURL` and `toggleRefinement`.

Relying on `this`  makes the usage of the `ratingMenu` widget impossible within a `panel`.

The `panel` syntax is the following:

```js
const ratingList = panel({
  // ...
})(refinementList) // context is lost

search.addWidget(
  ratingList({
    // ...
  })
);
```

Since we don't call the widget directly, the context (`this`) is lost when the widget is instanciated, throwing errors.

## This solution

This solution gets rid of unnecessary `this` uses to make this widget compatible with any widget instanciation syntax.

---

These fixes must also be implemented for:

- Breadcrumb
- ClearRefinements
- Menu
- NumericMenu
- ToggleRefinements

### DIFF
--- a/src/connectors/rating-menu/connectRatingMenu.js
+++ b/src/connectors/rating-menu/connectRatingMenu.js
@@ -14,7 +14,7 @@ var customStarRating = connectRatingMenu(function render(params, isFirstRenderin
 search.addWidget(
   customStarRatingI({
     attribute,
-    [ max=5 ],
+    [ max = 5 ],
   })
 );
 Full documentation available at https://community.algolia.com/instantsearch.js/v2/connectors/connectRatingMenu.html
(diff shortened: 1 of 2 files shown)


src/widgets/rating-menu/__tests__/rating-menu-test.js
```diff
@@ -90,31 +90,6 @@ describe('ratingMenu()', () => {
     expect(helper.search).toHaveBeenCalledTimes(0);
   });
 
-  it('refines the search', () => {
```
francoischalifour: We do not export `_toggleRefinement` anymore from the connector.

This tested internal implementations already covered at the connector level: https://github.com/algolia/instantsearch.js/blob/bf67509fdf937f15f568ad020b860b1bff8c465d/src/connectors/rating-menu/__tests__/connectRatingMenu-test.js#L135-L141

src/connectors/rating-menu/connectRatingMenu.js
```diff
@@ -178,8 +214,8 @@ export default function connectRatingMenu(renderFn, unmountFn) {
             instantSearchInstance,
             items: facetValues,
             hasNoResults: results.nbHits === 0,
-            refine: this._toggleRefinement,
-            createURL: this._createURL(state),
+            refine: toggleRefinement.bind(this, helper),
```
samouss: It's not the same behaviour as before. Now between `init` & `render` we have two different functions. Previously the function was only created once at the `init` step. We do that to avoid to lost the reference inside the render function.

src/connectors/rating-menu/connectRatingMenu.js
```diff
@@ -114,23 +114,52 @@ export default function connectRatingMenu(renderFn, unmountFn) {
       throw new Error(usage);
     }
 
+    let createURLFn;
```
samouss: The previous implementation didn't cached the function like `toggeRefinement`. On each render a new function is created. We can extract the body of the function and inline the call in both `init` & `render`.

PR #3363, opened 2018-12-18
feat(hitsPerPage): expose createURL to render function

francoischalifour:
The `createURL` parameter was missing from `connectHitsPerPage`.

### DIFF
--- a/src/connectors/hits-per-page/connectHitsPerPage.js
+++ b/src/connectors/hits-per-page/connectHitsPerPage.js
@@ -170,10 +170,13 @@ Learn more: https://community.algolia.com/instantsearch.js/v2/widgets/configure.
             ? helper.setQueryParameter('hitsPerPage', undefined).search()
             : helper.setQueryParameter('hitsPerPage', value).search();
 
-        createURLFn = value => {
-          this.setHitsPerPage(value);
-          return createURL(helper.state);
-        };
+        createURLFn = value =>
+          createURL(
+            helper.state.setQueryParameter(
+              'hitsPerPage',
+              !value && value !== 0 ? undefined : value
+            )
+          );
 
         renderFn(
           {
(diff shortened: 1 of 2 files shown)


src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -167,10 +170,16 @@ Learn more: https://community.algolia.com/instantsearch.js/v2/widgets/configure.
             ? helper.setQueryParameter('hitsPerPage', undefined).search()
             : helper.setQueryParameter('hitsPerPage', value).search();
 
+        createURLFn = value => {
+          this.setHitsPerPage(value);
```
samouss: The `createURL` function should not set the page because otherwise it will trigger a search every time it's called. The function must generate the URL that matches the provided value. You can take a look at the implementation inside [`connectMenu`](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/menu/connectMenu.js#L182-L183).

francoischalifour: Right, it would work as expected if we don't call `.search()`, right? If not, what helper function should it use then?

samouss: We don't have to call methods on the helper directly (because it mutates the state on it). What we have to do is (what is done inside [`connectMenu`](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/menu/connectMenu.js#L182-L183)) call the method on the `state` attribute. It won't mutate the internal helper state.

francoischalifour: Right, I see!

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -167,10 +170,19 @@ Learn more: https://community.algolia.com/instantsearch.js/v2/widgets/configure.
             ? helper.setQueryParameter('hitsPerPage', undefined).search()
             : helper.setQueryParameter('hitsPerPage', value).search();
 
+        createURLFn = value =>
```
samouss: For the moment IMO it's easier to follow the code with the `this` binding. Until we have a proper API to support it I think we should stick to the pattern than we used almost everywhere.

francoischalifour: Yes – I originally went for this because of #3347.

PR #3364, opened 2018-12-18
fix(panel): bind widget methods from panel

francoischalifour:
All widget's methods need to be bound to not lose the `this` context when called from the `panel` widget.
We need to check the type of the widget's attributes before binding them because some are not methods.

An example of a previously failing widget using `panel` is `ratingMenu`.

Closes #3347.

### DIFF
--- a/src/widgets/panel/panel.js
+++ b/src/widgets/panel/panel.js
@@ -129,33 +129,25 @@ export default function panel({
         container: getContainerNode(bodyRef),
       });
 
-      // All widget's methods need to be bound to not lose the `this`
-      // context when called from the `panel` widget.
-      // We need to check the type of the widget's attributes before binding them
-      // because some are not methods.
-      Object.keys(widget)
-        .filter(attribute => typeof widget[attribute] === 'function')
-        .forEach(method => {
-          widget[method] = widget[method].bind(widget);
-        });
-
       return {
         ...widget,
-        dispose(options) {
+        dispose(...args) {
           unmountComponentAtNode(getContainerNode(container));
 
           if (typeof widget.dispose === 'function') {
-            widget.dispose(options);
+            widget.dispose.call(this, ...args);
           }
         },
-        render(options) {
+        render(...args) {
+          const [options] = args;
+
           renderPanel({
             options,
             hidden: Boolean(hidden(options)),
           });
 
           if (typeof widget.render === 'function') {
-            widget.render(options);
+            widget.render.call(this, ...args);
           }
         },
       };
--- a/src/widgets/panel/panel.js
+++ b/src/widgets/panel/panel.js
@@ -129,13 +129,23 @@ export default function panel({
         container: getContainerNode(bodyRef),
       });
 
+      // All widget's methods need to be bound to not lose the `this`
+      // context when called from the `panel` widget.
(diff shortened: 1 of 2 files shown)


src/widgets/panel/panel.js
```diff
@@ -129,13 +129,23 @@ export default function panel({
...
 
+      // All widget's methods need to be bound to not lose the `this`
+      // context when called from the `panel` widget.
+      // We need to check the type of the widget's attributes before binding them
+      // because some are not methods.
+      Object.keys(widget)
+        .filter(attribute => typeof widget[attribute] === 'function')
+        .forEach(method => {
+          widget[method] = widget[method].bind(widget);
+        });
```
samouss: I'm not completely sure it will work in all cases. Because sometimes we create the function on the instance at the `init` step not ahead of time. If I'm not mistaken this code happens before the `init` lifecycle of the connector (like inside [`connectMenu`](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/menu/connectMenu.js#L180-L185)).

francoischalifour: That might be true.
We would either need to refactor those, or dump `this` from the connectors (initiative taken in #3347).

What other solution do you see?

samouss: Actually it does work because it doesn't matter once we have bind the instance to the correct context the one inside `init` is also correct and so on. But I would avoid to bind all the function on the widget. What we can do is to re-use the instance we have to call the appropriate method on the widget. The instance we have in the `panel` is the one that is used so we can re-use it (see below).

```js
return {
  // ...
  dispose(...args) {
    if (typeof widget.dispose === "function") {
      widget.dispose.call(this, ...args);
    }
  }
};
```

francoischalifour: We sometimes rely on `this` in [non-conventional methods](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/menu/connectMenu.js#L135-L140) in the connectors. This would break the implementation if we don't bind these methods.

What's wrong with binding all the methods for now?

francoischalifour: I agree, but the solution with `call`, or binding _only_ the public APIs, will fail in some cases (e.g. [`connectMenu`](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/menu/connectMenu.js#L135-L140)).

samouss: Did you tried the version with `call` and `panel` over `menu` ? I don't think it fails.

francoischalifour: Indeed, it works!

src/widgets/panel/panel.js
```diff
@@ -131,21 +131,23 @@ export default function panel({
 
       return {
         ...widget,
-        dispose() {
+        dispose(...args) {
           unmountComponentAtNode(getContainerNode(container));
 
           if (typeof widget.dispose === 'function') {
-            widget.dispose();
+            widget.dispose.call(this, ...args);
```
samouss: We can add a test on the `panel` to be sure that we don't lose the context rather than only rely on stories.

francoischalifour: I tried but I had difficulties with the `rewire` system, it makes it complicated.

PR #3366, opened 2018-12-19
fix(refinementList): update "show more" behavior

francoischalifour:
Implements #3343 fixes for `refinementList`.

### DIFF
--- a/src/connectors/menu/connectMenu.js
+++ b/src/connectors/menu/connectMenu.js
@@ -38,6 +38,7 @@ Full documentation available at https://community.algolia.com/instantsearch.js/v
  * @typedef {Object} CustomMenuWidgetOptions
  * @property {string} attribute Name of the attribute for faceting (eg. "free_shipping").
  * @property {number} [limit = 10] How many facets values to retrieve.
+ * @property {boolean} [showMore=false] Whether to display a button that expands the number of items.
  * @property {number} [showMoreLimit = 20] How many facets values to retrieve when `toggleShowMore` is called, this value is meant to be greater than `limit` option.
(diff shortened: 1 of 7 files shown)


src/connectors/menu/connectMenu.js
```diff
@@ -38,6 +38,7 @@ Full documentation available at https://community.algolia.com/instantsearch.js/v
  * @typedef {Object} CustomMenuWidgetOptions
  * @property {string} attribute Name of the attribute for faceting (eg. "free_shipping").
  * @property {number} [limit = 10] How many facets values to retrieve.
+ * @property {boolean} [showMore=false] Whether to display a button that expands the number of items.
```
samouss:
```suggestion
  * @property {boolean} [showMore = false] Whether to display a button that expands the number of items.
```

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -323,21 +305,12 @@ export default function connectRefinementList(renderFn, unmountFn) {
...
-              showMoreLimit
-            );
-          }
-        }
+        const currentMaxValuesPerFacet = configuration.maxValuesPerFacet || 0;
+
+        widgetConfiguration.maxValuesPerFacet = Math.max(
+          currentMaxValuesPerFacet,
+          showMore ? showMoreLimit : limit
+        );
```
samouss: We can cover this with a test case. Like the one we have inside [`connectMenu`](https://github.com/algolia/instantsearch.js/blob/v3/src/connectors/menu/__tests__/connectMenu-test.js#L45-L61).

src/connectors/menu/connectMenu.js
francoischalifour: Yeah – this will be removed with the new doc anyway.

PR #3367, opened 2018-12-19
fix(utils): warn only in development environment

francoischalifour:
This is the new signature of the `warning` method:

```js
warning(condition: boolean, message: string)
```

It logs a warning only if the condition is not met.

The condition is now part of the method so that the production UMD build is stripped correctly with less code.

### DIFF
--- a/src/connectors/geo-search/connectGeoSearch.js
+++ b/src/connectors/geo-search/connectGeoSearch.js
@@ -146,9 +146,7 @@ Please consider using the \`Configure\` widget instead:
 
 search.addWidget(
   configure({
-    aroundLatLngViaIP: ${(widgetParams &&
-      widgetParams.enableGeolocationWithIP) ||
-      'true'},
+    aroundLatLngViaIP: ${widgetParams.enableGeolocationWithIP || 'true'},
   })
 );
 


Haroenv:
the boolean trap as first argument seems a bit odd, since I would be surprised if the bundler can actually detect this as unused code when __DEV__ is false. It probably needs to have its own if around each of them

francoischalifour:
The bundler does detect them. If we keep the wrapping `if`, it will transpile to something like:

```js
// ...
{
  // whatever was here on top on the warning
}
// ...
```

With the condition inside the `warning` method, it's completely stripped.

FYI this is what React is doing for their [`invariant`](https://github.com/facebook/react/blob/master/packages/shared/invariant.js) and [`warning`](https://github.com/facebook/react/blob/master/packages/shared/warning.js) functions.

Haroenv:
Ah you're right, it's quite surprising that the bundler inlines those things. I didn't expect that, but that's good news!

src/lib/__tests__/utils-test.js
```diff
@@ -937,34 +937,47 @@ describe('utils.deprecate', () => {
 });
 
 describe('utils.warn', () => {
-  it('expect to print a warn message', () => {
+  it('prints a warn message with a false condition', () => {
     const message = 'message';
     const warn = jest.spyOn(global.console, 'warn');
```
tkrugg: might be shorter if you factorize the mocking and the resetting in a beforeEach & afterEach

src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -186,7 +192,7 @@ http://community.algolia.com/instantsearch.js/migration-guide
     }
 
     if (typeof widgetParams.precision !== 'undefined') {
-      warn(`
+      warning(`
```
Haroenv: What happens if the condition isn't given (i.e. a truthy value / the error message)? There's more like this

francoischalifour: You need to pass `false` to the first argument to always warn. I forgot to change these ones.

src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -138,69 +138,81 @@ const connectGeoSearch = (renderFn, unmountFn) => {
...
+    warning(
+      typeof widgetParams.position === 'undefined',
+      `
 The option \`position\` has been removed from the GeoSearch widget.
 Please consider using the \`Configure\` widget instead:
 
 search.addWidget(
   configure({
-    aroundLatLng: '${widgetParams.position.lat}, ${widgetParams.position.lng}',
+    aroundLatLng: '${widgetParams &&
```
Haroenv: widgetParams is always defined here

src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -138,69 +138,81 @@ const connectGeoSearch = (renderFn, unmountFn) => {
...
+    warning(
+      typeof widgetParams.radius === 'undefined',
+      `
 The option \`radius\` has been removed from the GeoSearch widget.
 Please consider using the \`Configure\` widget instead:
 
 search.addWidget(
   configure({
-    aroundRadius: ${widgetParams.radius},
+    aroundRadius: ${widgetParams && widgetParams.radius},
```
Haroenv: widgetParams is always defined

src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -138,69 +138,81 @@ const connectGeoSearch = (renderFn, unmountFn) => {
...
+    warning(
+      typeof widgetParams.precision === 'undefined',
+      `
 The option \`precision\` has been removed from the GeoSearch widget.
 Please consider using the \`Configure\` widget instead:
 
 search.addWidget(
   configure({
-    aroundPrecision: ${widgetParams.precision},
+    aroundPrecision: ${widgetParams && widgetParams.precision},
```
Haroenv: idem

src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -138,38 +138,47 @@ const connectGeoSearch = (renderFn, unmountFn) => {
...
+    warning(
+      false,
+      `
 The option \`enableGeolocationWithIP\` has been removed from the GeoSearch widget.
 Please consider using the \`Configure\` widget instead:
 
 search.addWidget(
   configure({
-    aroundLatLngViaIP: ${widgetParams.enableGeolocationWithIP || 'true'},
+    aroundLatLngViaIP: ${(widgetParams &&
```
samouss: The value `widgetParams` is always defined we can avoid the check on `widgetParams`.

PR #3395, opened 2019-01-14
fix(usage): update usage error handling

francoischalifour:
This is a proposal for a new error handling of widgets/connectors' bad usage, with a showcase of `poweredBy`.

## The problem

When using widgets, we currently ignore all errors thrown at the connector level and only re-throw the widget's usage. This is very misleading as a user because you never know what you're doing wrong (showing a complete usage doesn't give any clue on what's misused).

Moreover, having the widget's or the connector's usage in the source code makes it error prone as it has a maintenance cost. Now that we have a single place for [our documentation](https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/js/) (the Algolia website), we should redirect all our documentation efforts there. Besides, it increases tremendously the bundle size.

## This solution

It gets rid of the verbose usage and links to the documentation website for more details. An error thrown at the connector level is not ignored anymore by the widget.

We can implement this strategy to all widgets/connectors if it works for you.

#### For a widget

```js
import { createDocumentationMessageGenerator } from '../../lib/utils';

const withUsage = createDocumentationMessageGenerator('powered-by');

export default function poweredBy({ container } = {}) {
  if (!container) {
    throw new Error(withUsage('The "container" option is required.'));
  }
}
```

#### For a connector

```js
// ...

const withUsage = createDocumentationMessageGenerator('powered-by', {
  connector: true,
});

// ...
```

#### Example of error for a widget

```
The "container" option is required.

See documentation: https://www.algolia.com/doc/api-reference/widgets/powered-by/js/
```

#### Example of error for a connector

```
The render function is not valid (got type "undefined").

See documentation: https://www.algolia.com/doc/api-reference/widgets/powered-by/js/#connector
```


src/lib/utils.js
```diff
@@ -492,3 +495,19 @@ function insideBoundingBoxToBoundingBox(value) {
 
   return insideBoundingBoxStringToBoundingBox(value);
 }
+
+function createDocumentationMessageGenerator(
+  widget,
+  { connector = false } = {}
```
Haroenv:
```suggestion
  { type } = {}
```

Where type would be an string `"widget" | "connector"`

src/widgets/powered-by/powered-by.js
```diff
@@ -62,7 +62,7 @@ export default function poweredBy({
   theme = 'light',
 } = {}) {
   if (!container) {
-    throw new Error(withUsage('The "container" option is required.'));
+    throw new Error(withUsage('The `container` option is required.'));
```
Haroenv: at a later point these messages can be factorised out as well I think

francoischalifour: Yeah – we'll see how far we want to go!

PR #3398, opened 2019-01-17
chore(stories): remove hits with highlighted array story

francoischalifour:
This story has never worked with this dataset.

### DIFF
--- a/stories/hits.stories.js
+++ b/stories/hits.stories.js
@@ -105,45 +105,4 @@ storiesOf('Hits', module)
         })
       );
     })
-  )
-  .add(
-    'with highlighted array',
-    withHits(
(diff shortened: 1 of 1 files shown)


stories/hits.stories.js
```diff
@@ -105,45 +105,4 @@ storiesOf('Hits', module)
...
-                    </div>
-                  </div>
-                `,
-            },
-          })
-        );
-      },
-      {
-        indexName: 'highlight_array',
-        appId: 'KY4PR9ORUL',
```
Haroenv: Could we not make a new app which showed it?

francoischalifour: We can do that later on.

PR #3400, opened 2019-01-18
chore(babel): migrate to Babel 7

francoischalifour:
This migrates from Babel 6 to Babel 7.

I took this opportunity to update all Babel and Rollup dependencies.

Note that the production bundle increased from _77.32 KB_ to _77.7 KB_ because Babel 7 becomes more spec-compliant.


francoischalifour:
The remaining issue that I'm aware of is the styling of [the storybook](https://deploy-preview-3400--instantsearchjs.netlify.com/stories):

https://github.com/algolia/instantsearch.js/blob/b23f7f58593a1e69a32492f0454c61b5f85d0640/.storybook/config.js#L3

It's broken when building the storybook, but not during development. I'm not sure why yet.

.babelrc
```diff
@@ -3,78 +3,78 @@
     "development": {
```
samouss: We could move this file to a `.js` file to remove all the duplication we have between the different environment.

samouss:
To avoid the issue with the CSS file you can use the `--static-dir` option with the `preview-head`.

.babelrc
francoischalifour: I actually find it more clear to have the exhaustive configuration hard coded in a file, as opposed to overriding it programmatically for specific environments.

Is configuration duplication a problem for you?

Note that if we go for a `babel.config.js` file, we'd better off do it in another PR because there are a lot of stacked PRs relying on that one.

francoischalifour:
Thank you @samouss – the styles are fixed in production.

.babelrc
samouss: Yep I find it harder to read rather than a plain JS file. It's easy to forgot a place to add / update / remove a plugin or a preset, etc... We can do this change later or not.

Haroenv:
What is the diff in the output?

PR #3411, opened 2019-01-21
chore(typescript): setup Jest (part 1)

francoischalifour:
This is the first step toward our process migrating to TypeScript.

- **Jest**.
  - The Jest configuration has been moved to its own file.
  - It upgrades Jest to version 24 (soon stable) which supports TypeScript by default. The Jest team seems confident enough to [use this alpha version in the React codebase](https://github.com/facebook/react/pull/14634), so I figured let's use it right now. Snapshots have been updated accordingly (a single file needed updates).

- **Renovate**.
  - The Renovate configuration has been moved to its own file.
  - The `types` packages remain unpinned. The corresponding Renovate configuration has been created.

### Related

- Part 2: #3412
- Part 3: #3413
- Part 4: #3414
- Part 5: #3417


src/lib/__tests__/__snapshots__/InstantSearch-test.js.snap
```diff
@@ -25,7 +25,7 @@ Array [
         ],
         "results": Array [
           Object {
-            "isThrow": false,
+            "type": "return",
```
francoischalifour: This is related to the Jest upgrade.

PR #3412, opened 2019-01-21
chore(typescript): setup TSLint (part 2)

francoischalifour:
This is the second step toward our process migrating to TypeScript.

- **ESLint**.
  - Resolve `js`, `ts` and `tsx` files.
- **TSLint**.
  - Add default configuration. This is the [same configuration as React InstantSearch](https://github.com/algolia/react-instantsearch/blob/93c5ce04bd6f26330dbaf77d2e9926ea98f59e96/tsconfig.json) plus the disabled rule `jsx-no-lambda` which we rely on heavily in the InstantSearch.js codebase.

### Related

- Part 1: #3411
- Part 3: #3413
- Part 4: #3414
- Part 5: #3417


Haroenv:
config itself looks fine to me

francoischalifour:
@Haroenv Yeah, I originally added the Travis check in #3414 but changed my mind. It's indeed unnecessary since there's no TS files yet. I will remove this Travis script for now.

PR #3413, opened 2019-01-21
chore(typescript): setup Storybook (part 3)

francoischalifour:
This is the third step toward our process migrating to TypeScript.

This accepts files with `js`, `ts` and `tsx` as extensions and install storybook addons types.

### Related

- Part 1: #3411
- Part 2: #3412
- Part 4: #3414
- Part 5: #3417


francoischalifour:
@Haroenv I tried locally with a dumb TS file. If it makes you feel better you can also try haha

PR #3415, opened 2019-01-21
fix(core): dispose sets helper to null

Haroenv:
Without this change, if you dispose an instance and start it again, a search (not necessarily a request, since the cache isn't cleared) should be done again with a fresh state, and not the state of before the dispose

similar to #3399

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -340,6 +340,7 @@ class InstantSearch extends EventEmitter {
   dispose() {
     this.removeWidgets(this.widgets);
     this.started = false;
+    this.helper = null;
   }
 
   createURL(params) {
--- a/src/lib/__tests__/InstantSearch-test.js
+++ b/src/lib/__tests__/InstantSearch-test.js
@@ -516,6 +516,21 @@ describe('InstantSearch lifecycle', () => {
     expect(helperSearchSpy).toHaveBeenCalledTimes(1);
     expect(search.started).toBe(false);
   });
+
+  it('should set the helper to `null`', () => {
+    search = new InstantSearch({
+      indexName,
+      searchClient: algoliasearch(appId, apiKey),
+    });
+
+    search.start();
+
+    expect(search.helper).not.toBe(null);
+
+    search.dispose();
+
+    expect(search.helper).toBe(null);
+  });
 });
 
 it('Allows to start without widgets', () => {


src/lib/InstantSearch.js
```diff
@@ -340,6 +340,7 @@ class InstantSearch extends EventEmitter {
   dispose() {
     this.removeWidgets(this.widgets);
     this.started = false;
+    this.helper = null;
```
francoischalifour: Can we add a comment explaining why this line is necessary? (so that the new search is performed with a new helper state).

It'd be nice to add some context there for future contributors.

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -516,6 +516,21 @@ describe('InstantSearch lifecycle', () => {
     expect(helperSearchSpy).toHaveBeenCalledTimes(1);
     expect(search.started).toBe(false);
   });
+
+  it('should set the helper to `null`', () => {
```
francoischalifour: Why is this test in the test suite `"When adding widgets after start"`?

francoischalifour:
As a side note, the `dispose` feature should probably not be "experimental" anymore, as [its documentation](https://github.com/algolia/instantsearch.js/blob/fix/dispose-helper/src/lib/InstantSearch.js#L334-L339) states (the same goes for other methods).

src/lib/__tests__/InstantSearch-test.js
Haroenv: I copied it and didn't realise it was in the wrong spot

src/lib/InstantSearch.js
```diff
@@ -339,7 +339,10 @@ class InstantSearch extends EventEmitter {
    */
   dispose() {
     this.removeWidgets(this.widgets);
+    // allow to restart this instance
     this.started = false;
+    // start from a fresh state, not previously set refinements
```
francoischalifour: I had something like this in mind.

```suggestion

    // The helper needs to be reset to perform the next search from a fresh state.
    // If not reset, it would use the state stored before calling `dispose()`.
```

Documenting this way is a pattern that I would like to use more in the codebase. It makes the intent clear and explains why a line is required. WDYT?

Haroenv: As you want, I didn't realise you preferred it written in full

src/lib/InstantSearch.js
```diff
@@ -339,7 +339,12 @@ class InstantSearch extends EventEmitter {
    */
   dispose() {
     this.removeWidgets(this.widgets);
+    // allow to restart this instance
```
```suggestion
    // You can not start an instance two times, therefore a disposed instance needs to set started as false
    // otherwise this can not be restarted at a later point.
```

PR #3420, opened 2019-01-24
fix(InstantSearch): remove event listeners

samouss:
**Summary**

We recently updated the `dispose` function to set the `helper` to `null`. It's indeed the right approach but it might happen that a `render` occurs after the `dispose` e.g `start` → `dispose` right after the start. The search is async so the render is called after we set the helper to null and it throws an error. To avoid this issue we have to remove the listeners on the helper before we lose his reference.

### DIFF
--- a/src/lib/__tests__/InstantSearch-test.js
+++ b/src/lib/__tests__/InstantSearch-test.js
@@ -547,6 +547,33 @@ describe('dispose', () => {
 
     expect(search.helper).not.toBe(null);
   });
+
+  it('should remove the listeners on the helper', done => {
+    const search = new InstantSearch({
+      indexName: '',
+      searchClient: {
+        search() {
+          return Promise.resolve({
+            results: [
+              {
+                query: 'fake query',
+              },
+            ],
+          });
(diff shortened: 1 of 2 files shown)


src/lib/__tests__/InstantSearch-test.js
```diff
@@ -537,6 +547,37 @@ describe('dispose', () => {
...
+      },
+    });
+
+    search.on('error', () => {
+      done.fail('InstantSearch should not throw an error.');
+    });
+
+    search.start();
+
+    expect(search.helper).not.toBe(null);
```
francoischalifour: This assertion is redundant with the previous test, it is not related to this test. We can remove it.

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -537,6 +547,37 @@ describe('dispose', () => {
...
+
+    search.dispose();
+
+    expect(search.helper).toBe(null);
```
francoischalifour: Same.

PR #3422, opened 2019-01-24
 fix(searchbox): update usage errors

francoischalifour:
This updates the usage errors messages for `searchBox`.

- `input`s are not valid containers anymore
- `autofocus` used to accept string values for `"auto"`, which is not a valid value for the HTML attribute `autofocus`.

> Related: https://github.com/algolia/instantsearch.js/pull/3127#pullrequestreview-195958590


src/widgets/search-box/search-box.js
```diff
@@ -217,25 +222,20 @@ export default function searchBox({
 
   const containerNode = getContainerNode(container);
 
-  if (containerNode.tagName === 'INPUT') {
-    // eslint-disable-next-line
-    // FIXME: the link should be updated when the documentation is migrated in the main Algolia doc
-    throw new Error(
-      `[InstantSearch.js] Since in version 3, \`container\` can not be an \`input\` anymore.
+  warning(
```
Haroenv: I'm not really following how this is a warning, if the target is an input it won't work at all right?

francoischalifour: Indeed, but we're not catching all the different `tagName`s.

What we want here is to warn the user that something that used to work in v2 might not work in v3.

robertmogos: I agree with @Haroenv here. If the behaviour is broken and the user would not be able to use it, it is better to throw an error. Warnings can be easily ignored and the user would spend more time to investigate why he doesn't have the desired behaviour. 
Warnings should be use in cases where maybe it is not the recommended way of using or doing something but would still work

robertmogos: > What we want here is to warn the user that something that used to work in v2 might not work in v3.

**Might not**  work or 100% not going to work ?

Haroenv: reason this is an error IMO is because it's already being checked as a compatibility layer. If e.g. attribute was renamed to attributeName (again), we would throw an error, not a warning 🤔

francoischalifour: Here is a counter-argument to yours: https://github.com/algolia/instantsearch.js/blob/f459b3942e51274f1aca1228b21ec3981c17b7c8/src/connectors/geo-search/connectGeoSearch.js#L158-L175.

samouss: Yep I'll go for an error also since it won't work at all. That last example you gave does not prevent you to use the widget. The option is ignored but the widget is fully usable.

francoischalifour: I'll convert it back to an error then, even if I'm not convinced it's semantically correct.

Do we still agree that the `autofocus` case should be a warning?

PR #3424, opened 2019-01-24
test(InstantSearch): drop Rewire usage

samouss:
**Summary**

The PR drops the usage of Rewire for the InstantSearch tests. I didn't rewrote the test I've only replace the usage of Rewire by Jest. I've fix some tests that were not correct. With Rewire it was fine but not with Jest.


src/lib/__tests__/InstantSearch-test.js
```diff
@@ -507,14 +531,23 @@ describe('InstantSearch lifecycle', () => {
     search.start();
 
     expect(search.widgets).toHaveLength(5);
-    expect(helperSearchSpy).toHaveBeenCalledTimes(1);
+    expect(search.helper.search).toHaveBeenCalledTimes(1);
     expect(search.started).toBe(true);
 
+    // Dispose delete the refernce of the `helper`
```
francoischalifour:
```suggestion
    // Calling `dispose()` deletes the reference of the `helper`
```

PR #3478, opened 2019-02-04
fix(breadcrumb): update usage errors

francoischalifour:
This updates the usage errors for `breadcrumb`.

### DIFF
--- a/src/connectors/breadcrumb/connectBreadcrumb.js
+++ b/src/connectors/breadcrumb/connectBreadcrumb.js
@@ -1,26 +1,14 @@
 import find from 'lodash/find';
 import isEqual from 'lodash/isEqual';
-import { checkRendering, warning } from '../../lib/utils';
+import {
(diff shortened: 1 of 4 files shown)


src/connectors/breadcrumb/connectBreadcrumb.js
```diff
@@ -56,20 +44,25 @@ Full documentation available at https://community.algolia.com/instantsearch.js/v
...
-  checkRendering(renderFn, usage);
+  checkRendering(renderFn, withUsage());
   return (widgetParams = {}) => {
     const {
       attributes,
       separator = ' > ',
       rootPath = null,
       transformItems = items => items,
     } = widgetParams;
-    const [hierarchicalFacetName] = attributes;
```
francoischalifour: This line used to fail when `attributes` was not an array. I moved it after the check and added a test.

PR #3479, opened 2019-02-04
fix(currentRefinements): update usage errors

francoischalifour:
This updates the usage errors for `currentRefinements`.


src/connectors/clear-refinements/__tests__/connectClearRefinements-test.js
```diff
@@ -1,21 +1,31 @@
-import jsHelper from 'algoliasearch-helper';
-const SearchResults = jsHelper.SearchResults;
```
Haroenv: I think we moved it like this because rollup doesn't always correctly import these "fake" named exports. If it works here then no problem of course.

francoischalifour: What do you mean with Rollup? Rollup has nothing to do with these tests.

PR #3484, opened 2019-02-05
fix(configure): update usage errors

francoischalifour:
This updates the usage errors for `configure`.


src/connectors/configure/connectConfigure.js
```diff
@@ -45,7 +30,11 @@ Full documentation available at https://community.algolia.com/instantsearch.js/v
 export default function connectConfigure(renderFn = noop, unmountFn = noop) {
   return (widgetParams = {}) => {
     if (!isPlainObject(widgetParams.searchParameters)) {
-      throw new Error(usage);
+      throw new Error(
+        withUsage(
+          `The \`searchParameters\` option must be an object (got type "${typeof widgetParams.searchParameters}")`
```
tkrugg: There is a gotcha
`.isPlainObject(null) === false` but `typeof null === object`.
In this case the error message will not make sense. May be best to remove the typeof part

src/widgets/configure/__tests__/configure-test.js
```diff
@@ -2,10 +2,20 @@ import configure from '../configure';
...
   it('throws when you pass it a non-plain object', () => {
-    [
-      () => configure(new Date()),
-      () => configure(() => {}),
-      () => configure(/ok/),
-    ].forEach(widget => expect(widget).toThrow(/Usage/));
+    expect(() => {
+      configure(new Date());
+    }).toThrowErrorMatchingInlineSnapshot(`
+"The \`searchParameters\` option must be an object (got type \\"object\\")
```
tkrugg: gotcha  😄

francoischalifour: Haha good catch!

francoischalifour:
@Haroenv Displaying the value is not trivial because it depends on its type. Feel free to improve the error messages though!

PR #3488, opened 2019-02-06
fix(hierarchicalMenu): update usage errors

francoischalifour:
This updates the usage errors for `hierarchicalMenu`.


src/widgets/hierarchical-menu/hierarchical-menu.js
```diff
@@ -193,8 +182,8 @@ export default function hierarchicalMenu({
   templates = defaultTemplates,
   cssClasses: userCssClasses = {},
 } = {}) {
-  if (!container || !attributes || !attributes.length) {
```
francoischalifour: This was testing the `attributes` option while it's the connector responsibility.

PR #3489, opened 2019-02-07
fix(searchBox): add reusable SearchBox component

### LINKED ISSUE #2052: Missing indication on how to style the searchForFacetValues searchbar
Feature.

I want to make it match the rest of the design of my search. Neither the `searchForFacetValues.templates`, `templates` nor `cssClasses` options seem to accept anything that could let me change the styling of the searchbar. At least, I couldn't find any mention of it in the doc.

The pre-generated class names also does not seem to follow the `ais-` notation used in the other widgets, which makes me think they are internal classes and shouldn't be relied upon.

Having the class names (currently `searchbox`, `sbx-sffv`, `sbx-sffv__wrapper`, `sbx-sffv__input`, etc) follow the `ais-` naming used in the other widgets would be enough for my use-case. It would give me the signal that I can style them and they won't change in an upcoming release.

image
image

- instantsearch.js

1.11.2

francoischalifour:
This creates a reusable `SearchBox` React component and removes the imperative JavaScript logic that we had.

## Benefits

- Use the `SearchBox` component in the `searchBox` widget
- Use the `SearchBox` component in the `refinementList` SFFV searchbox. This allows to customize the templates and to style the SFFV searchbox (fixes #2052)

### Different modes

The component exists both in controlled and uncontrolled modes:

- **Controlled mode**. This is the default mode that is driven by the connector's query when `searchAsYouType` is `true` (default value).
- **Uncontrolled mode**. This is the mode having an internal state to keep track of the query when `searchAsYouType` is `false`.

### `onHistoryChange` was removed

This is an old API that we decided to not document on the [v3 `searchBox` documentation](https://www.algolia.com/doc/api-reference/widgets/search-box/js/). This use case is handled by React.

### Templates

The `searchBox` widget passes 3 templates to the `SearchBox` component:

- `reset`
- `submit`
- `loadingIndicator`

When using the `SearchBox` component in the SFFV of `RefinementList`, the templates are name as follows:

- `item`
- `showMoreText`
- `searchableNoResults`
- `searchableReset`
- `searchableSubmit`
- `searchableLoadingIndicator`

## Implications

- Update the [`templates` documentation in `refinementList`](https://www.algolia.com/doc/api-reference/widgets/refinement-list/js/#widget-param-templates)
- Update the [`cssClasses` documentation in `refinementList`](https://www.algolia.com/doc/api-reference/widgets/refinement-list/js/#widget-param-cssclasses)

## Stories

- [SearchBox](https://deploy-preview-3489--instantsearchjs.netlify.com/stories/?selectedKind=SearchBox)
- [SFFV](https://deploy-preview-3489--instantsearchjs.netlify.com/stories/?selectedKind=RefinementList&selectedStory=with%20search%20inside%20items)


src/components/SearchBox/__tests__/SearchBox-test.js
```diff
@@ -0,0 +1,347 @@
...
+      test('sets all CSS classes', () => {
+        const wrapper = mount(<SearchBox {...defaultProps} />);
+
+        expect(wrapper.find('div').hasClass('root')).toEqual(true);
+        expect(wrapper.find('form').hasClass('form')).toEqual(true);
+        expect(wrapper.find('input').hasClass('input')).toEqual(true);
+        expect(
+          wrapper.find('button[type="submit"]').hasClass('submit')
+        ).toEqual(true);
+        // expect(
```
francoischalifour: We don't have access to this element since it's in the template.

src/components/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    onSubmit(event);
+
+    return false;
+  };
+
+  onReset = event => {
+    const { searchAsYouType, refine, onReset } = this.props;
+    const query = '';
```
francoischalifour: I decided not to rely on the `clear()` function given by the connector because it's redundant with `refine('')`.

src/components/RefinementList/RefinementList.js
```diff
@@ -177,13 +192,16 @@ class RefinementList extends Component {
     const searchBox = this.props.searchFacetValues && (
       <div className={this.props.cssClasses.searchBox}>
         <SearchBox
-          ref={i => {
-            this.searchbox = i;
-          }}
+          ref={searchBoxRef => (this.searchBox = searchBoxRef)}
```
Haroenv: ~~could use createRef~~ It's not in preact yet

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
```diff
@@ -264,7 +264,7 @@ exports[`RefinementList rendering with facets from search 1`] = `
       <form
         action=""
         className="ais-SearchBox-form"
-        noValidate="novalidate"
+        noValidate={true}
```
Haroenv: apart from this PR, I find it weird to see snapshots of React components; and not of the real HTML that comes on the page. It might be hard to check if something is actually spelled the react-way or the regular way here

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
```diff
@@ -273,91 +273,83 @@ exports[`RefinementList rendering with facets from search 1`] = `
...
-          spellCheck="false"
+          spellCheck={false}
           type="search"
+          value=""
         />
         <button
           className="ais-SearchBox-submit"
+          dangerouslySetInnerHTML={
+            Object {
+              "__html": Object {
```
Haroenv: this surprises me 🤔 It probably shouldn't be an object here

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
```diff
@@ -273,91 +273,83 @@ exports[`RefinementList rendering with facets from search 1`] = `
...
-              d="M26.804 29.01c-2.832 2.34-6.465 3.746-10.426 3.746C7.333 32.756 0 25.424 0 16.378 0 7.333 7.333 0 16.378 0c9.046 0 16.378 7.333 16.378 16.378 0 3.96-1.406 7.594-3.746 10.426l10.534 10.534c.607.607.61 1.59-.004 2.202-.61.61-1.597.61-2.202.004L26.804 29.01zm-10.426.627c7.323 0 13.26-5.936 13.26-13.26 0-7.32-5.937-13.257-13.26-13.257C9.056 3.12 3.12 9.056 3.12 16.378c0 7.323 5.936 13.26 13.258 13.26z"
-            />
-          </svg>
-        </button>
+        />
         <button
           className="ais-SearchBox-reset"
+          dangerouslySetInnerHTML={
+            Object {
+              "__html": Object {
```
Haroenv: idem

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
```diff
@@ -273,91 +273,83 @@ exports[`RefinementList rendering with facets from search 1`] = `
...
-              d="M8.114 10L.944 2.83 0 1.885 1.886 0l.943.943L10 8.113l7.17-7.17.944-.943L20 1.886l-.943.943-7.17 7.17 7.17 7.17.943.944L18.114 20l-.943-.943-7.17-7.17-7.17 7.17-.944.943L0 18.114l.943-.943L8.113 10z"
-            />
-          </svg>
-        </button>
+        />
         <span
           className="ais-SearchBox-loadingIndicator"
+          dangerouslySetInnerHTML={
+            Object {
+              "__html": Object {
```
Haroenv: idem

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
```diff
@@ -462,91 +454,83 @@ exports[`RefinementList rendering without facets from search 1`] = `
...
```
Haroenv: idem

src/components/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,213 @@
+import React, { Component } from 'preact-compat';
+import PropTypes from 'prop-types';
+import noop from 'lodash/noop';
```
Haroenv: I don't think we use this one anywhere else; so in that case just writing it in full once seems better

src/components/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,213 @@
...
+        <form
+          action=""
+          role="search"
+          className={cssClasses.form}
+          noValidate
+          onSubmit={this.onSubmit}
+          onReset={this.onReset}
+        >
+          <input
+            ref={this.onInputRef}
```
Haroenv: I haven't seen the `onInputRef` pattern before. Since we already pass `autoFocus` it seems redundant

src/components/SearchBox/__tests__/SearchBox-test.js
Haroenv: makes sense not to test it

src/connectors/search-box/connectSearchBox.js
```diff
@@ -82,7 +81,7 @@ export default function connectSearchBox(renderFn, unmountFn) {
         this._clear();
       },
 
-      init({ helper, onHistoryChange, instantSearchInstance }) {
+      init({ helper, instantSearchInstance }) {
```
Haroenv: technically a breaking change; but practically this should be fine

src/widgets/search-box/search-box.js
```diff
@@ -24,108 +24,23 @@ const renderer = ({
...
-    // When the page is coming from BFCache
-    // (https://developer.mozilla.org/en-US/docs/Working_with_BFCache)
-    // then we force the input value to be the current query
-    // Otherwise, this happens:
-    // - <input> autocomplete = off (default)
-    // - search $query
-    // - navigate away
-    // - use back button
-    // - input query is empty (because <input> autocomplete = off)
-    window.addEventListener('pageshow', () => {
```
Haroenv: this compatibility layer has been removed

francoischalifour: I wasn't able to reproduce. Were you?

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
francoischalifour: That's true. We can change this in future tests. Which method would you recommend to use?

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
francoischalifour: I return the object from the template in the test: https://github.com/algolia/instantsearch.js/blob/f2f903b51e876d5393331896e6de0d9974ea1a4c/src/components/RefinementList/__tests__/RefinementList-test.js#L263-L268

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
francoischalifour: Idem

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
francoischalifour: Idem

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
francoischalifour: Idem

src/components/SearchBox/SearchBox.js
francoischalifour: We do – we can remove this dependency in another pass.

src/components/SearchBox/SearchBox.js
francoischalifour: Right, I'll simplify this.

I thought it was for compatibility reasons that we force the focus with `focus()` even though we use the `autofocus` attribute.

src/connectors/search-box/connectSearchBox.js
francoischalifour: Yes, the fact that we decided not to document this motivated me to remove it. Even the current team wasn't aware of `onHistoryChange`.

src/widgets/search-box/search-box.js
Haroenv: This was something @bobylito found a while ago only in FireFox, but I'm not too sure when it happens

francoischalifour: I tried to follow the steps in Firefox 64.0.2 but doesn't seem to do anything either.

src/components/RefinementList/__tests__/__snapshots__/RefinementList-test.js.snap
Haroenv: would make more sense to return something that would be as real life? ie. a string and not the whole object?

francoischalifour: You're right. I'll remove them.

src/components/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,213 @@
...
+      autofocus,
+      showSubmit,
+      showReset,
+      showLoadingIndicator,
+      templates,
+      isSearchStalled,
+      searchAsYouType,
+    } = this.props;
+
+    const query = searchAsYouType ? this.props.query : this.state.query;
```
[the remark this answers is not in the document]
francoischalifour: It causes performance issues when typing fast. Some characters are skipped when relying on the internal state.

src/components/RefinementList/RefinementList.js
```diff
@@ -3,9 +3,24 @@ import PropTypes from 'prop-types';
...
 import { isSpecialClick } from '../../lib/utils';
+import { component } from '../../lib/suit';
 import Template from '../Template/Template';
 import RefinementListItem from './RefinementListItem';
-import SearchBox from '../SearchBox/Searchbox';
+import SearchBox from '../SearchBox/SearchBox';
+
+const searchBoxSuit = component('SearchBox');
+
+const searchBoxCSSClasses = {
```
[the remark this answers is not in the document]
francoischalifour: Will do that. We'll need to update the documentation afterwards.

src/components/SearchBox/SearchBox.js
```diff
@@ -0,0 +1,213 @@
...
+    refine: noop,
+  };
+
+  state = {
+    query: this.props.searchAsYouType ? '' : this.props.query,
+  };
+
+  componentWillReceiveProps(nextProps) {
+    // This keeps track of the `searchParameters` and changes the query accordingly.
+    if (!nextProps.searchAsYouType && nextProps.query !== this.props.query) {
```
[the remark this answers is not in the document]
francoischalifour: Indeed, will remove this.

Meschreiber:
Would love to know when this gets merged -- have a customer waiting on it :)

src/components/RefinementList/__tests__/RefinementList-test.js
```diff
@@ -28,6 +28,7 @@ describe('RefinementList', () => {
         ...defaultProps,
         cssClasses: {
           root: 'root',
+          searchable: {},
```
Haroenv: this is likely not necessary right?

src/components/RefinementList/__tests__/RefinementList-test.js
```diff
@@ -263,6 +277,9 @@ describe('RefinementList', () => {
           templates: {
             item: item => item,
             searchableNoResults: x => x,
+            reset: 'reset',
```
Haroenv: I thought we decided on `searchableReset` etc. here?

francoischalifour: It is still called `searchableReset` in the `refinementList` widget, but renamed `reset` [when passed to the component](https://github.com/algolia/instantsearch.js/blob/684cde80f6093c8f10625551c8a4992959711f9f/src/widgets/refinement-list/refinement-list.js#L19-L43). This for the component not to host the logic of the template name (see https://github.com/algolia/instantsearch.js/pull/3489#discussion_r256788379).

src/components/RefinementList/__tests__/RefinementList-test.js
francoischalifour: We use `searchable[cssClass]` in the component (and `undefined[cssClass]` throws) so I'm not sure there's other ways.

src/components/RefinementList/__tests__/RefinementList-test.js
Haroenv: missed that bit.

src/components/RefinementList/__tests__/RefinementList-test.js
Haroenv: You're right, this is a component test, not a widget test!

PR #3492, opened 2019-02-07
fix(menu): update usage errors

francoischalifour:


src/widgets/menu/__tests__/menu-test.js
```diff
@@ -10,38 +10,14 @@ jest.mock('preact-compat', () => {
 });
 
 describe('menu', () => {
-  it('throws usage when no container', () => {
-    expect(menu.bind(null, { attribute: '' })).toThrow(/^Usage/);
-  });
-
-  it('throws an exception when no attribute', () => {
```
francoischalifour: These tests were moved to the connector.

PR #3493, opened 2019-02-07
fix(menuSelect): update usage errors

francoischalifour:


src/widgets/menu-select/menu-select.js
```diff
@@ -96,8 +90,8 @@ export default function menuSelect({
   templates = defaultTemplates,
   transformItems,
 }) {
-  if (!container || !attribute) {
```
francoischalifour: This is checked in the connector (see https://github.com/algolia/instantsearch.js/pull/3492#discussion_r254757352).

PR #3494, opened 2019-02-07
fix(numericMenu): update usage errors

francoischalifour:


src/widgets/numeric-menu/__tests__/numeric-menu-test.js
```diff
@@ -12,23 +12,15 @@ jest.mock('preact-compat', () => {
...
   window.encodeURI(JSON.stringify({ start, end }));
 
-describe('numericMenu call', () => {
-  it('throws an exception when no container', () => {
-    const attribute = '';
-    const items = [];
-    expect(numericMenu.bind(null, { attribute, items })).toThrow(/^Usage/);
-  });
-
-  it('throws an exception when no attribute', () => {
```
francoischalifour: These were moved to the connector.

francoischalifour:
Indeed, it's still an improvement compared to the previous error management. We can improve it again in another pass!

PR #3508, opened 2019-02-11
chore(tests): make main widget render test snapshot smaller

Haroenv:
**Summary**

This test was constantly flaking for me for changed behaviour with keys of `null` which didn't matter for this snapshot.

**Result**

I shortened the snapshot used in this test by using matchers for the relevant parts of this test. 

I wanted to use `expect.any(AlgoliaSearchHelper)` and `expect.any(SearchParameters)`, but that didn't work (I think jest didn't recognise it as classes?).


samouss:
It does not work for `expect.any(SearchParameters)` alone?

Haroenv:
> It does not work for `expect.any(SearchParameters)` alone?

nope: 

```
 TypeError: any() expects to be passed a constructor function. Please pass one or use anything() to match any object.

      186 |             }),
      187 |             instantSearchInstance: expect.any(InstantSearch),
    > 188 |             state: expect.any(SearchParameters)
          |                           ^
      189 |             // state: expect.objectContaining({
      190 |             //   another: {
      191 |             //     config: 'parameter',

      at new Any (node_modules/expect/build/asymmetricMatchers.js:27:13)
      at Function.any (node_modules/expect/build/asymmetricMatchers.js:250:31)
      at Object.any (src/lib/__tests__/InstantSearch-test.js:188:27)
```

I think it might be because it's from `node_modules` 🤔

PR #3514, opened 2019-02-11
fix(range): update error messages

francoischalifour:
Since `connectRange` is used for both `rangeInput` and `rangeSlider`, I had to slightly adjust the function to generate the documentation link. This is not optimal but it works for now.

We could change the signature to something like this later:

```ts
type DocumentationWidget = {
  name: string;
  connector?: boolean;
};

function createDocumentationMessageGenerator(...widgets: DocumentationWidget[]);
```

Usage:

```js
const withUsage = createDocumentationMessageGenerator(
  { name: 'range-input', connector: true },
  { name: 'range-slider', connector: true }
)
```

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -484,11 +484,21 @@ function createDocumentationLink(widget, { connector = false } = {}) {
   ].join('');
 }
 
-function createDocumentationMessageGenerator(widget, { connector } = {}) {
-  const link = createDocumentationLink(widget, { connector });
+function createDocumentationMessageGenerator(
+  widgetOrWidgets,
+  { connector } = {}
+) {
+  const widgets = Array.isArray(widgetOrWidgets)
+    ? widgetOrWidgets
+    : [widgetOrWidgets];
+  const links = widgets
+    .map(widget => createDocumentationLink(widget, { connector }))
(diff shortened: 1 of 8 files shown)


src/lib/utils.js
```diff
@@ -484,11 +484,21 @@ function createDocumentationLink(widget, { connector = false } = {}) {
   ].join('');
 }
 
-function createDocumentationMessageGenerator(widget, { connector } = {}) {
-  const link = createDocumentationLink(widget, { connector });
+function createDocumentationMessageGenerator(
+  widgetOrWidgets,
```
Haroenv: what about avoiding this conditional and writing two functions (one plural, one singular, and then do a single pass to put all singular ones into this plural function?)

francoischalifour: That would also work. The solution in the original post would also remove all conditionals, without writing a new function.

Once all the current PRs are merged (they've all been sent), I can quickly refactor this. WDYT?

Btw I should have originally based all these changes to a specific branch, not `develop`.

Haroenv: Good for me. Maybe you could leave this one for last for that reason to avoid too more changes than necessary

PR #3515, opened 2019-02-11
feat(connectCurrentRefinements): add a root label (PART 1)

Haroenv:
**Summary**

The reason why you'd want a key like this is if you e.g. want to rename a single key or transform all keys to be upper case, without actually changing which attribute will be changed.

**Result**

A new root key `label`, this will always be identical to the attribute, and will be the thing which will be displayed.


francoischalifour:
We need to update the documentation.

PR #3516, opened 2019-02-11
fix: update usage utils signature

francoischalifour:
This updates the signature of the function to generate errors to handle multiple widgets.

The reason is explained in https://github.com/algolia/instantsearch.js/pull/3514#issue-251916573: the `connectRange` connector is shared between [`rangeInput#connector`](https://www.algolia.com/doc/api-reference/widgets/range-input/js/#connector) and [`rangeSlider#connector`](https://www.algolia.com/doc/api-reference/widgets/range-slider/js/#connector) and should link to both.

Function implementation: https://github.com/algolia/instantsearch.js/pull/3516/files#diff-fb54f78b07de422efaa3d2d82bb82db3

### DIFF
--- a/src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
+++ b/src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
@@ -5,8 +5,14 @@ import { TAG_PLACEHOLDER } from '../../../lib/escape-highlight';
 const fakeClient = { addAlgoliaAgent: () => {} };
 
 describe('connectAutocomplete', () => {
-  it('throws without `renderFn`', () => {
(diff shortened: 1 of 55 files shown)


francoischalifour:
@Haroenv We already use this pattern in the codebase. It transpiles to:

```js
function createDocumentationMessageGenerator() {
  for (var _len = arguments.length, widgets = new Array(_len), _key = 0; _key < _len; _key++) {
    widgets[_key] = arguments[_key];
  }

  var links = widgets.map(function (widget) {
    return createDocumentationLink(widget);
  }).join(', ');
  return function (message) {
    return [message, "See documentation: ".concat(links)].filter(Boolean).join('\n\n');
  };
}
```

PR #3523, opened 2019-02-11
fix(panel): update error messages

francoischalifour:


src/widgets/panel/panel.js
```diff
@@ -155,7 +150,7 @@ export default function panel({
         },
       };
     } catch (error) {
-      throw new Error(usage);
+      throw new Error(withUsage('Bad usage of the `panel` widget.'));
```
Haroenv: In which cases can this throw? Wouldn't you lose the inner error by catching it?

francoischalifour: I was undecided because we already were catching it.

I'd be in favor of not catching it also, let's remove it!

PR #3525, opened 2019-02-12
build(rollup): display minified file size

francoischalifour:
We used to only display the Gzipped file size during build. The minified size is also interesting to track.


francoischalifour:
Oh, I see, thanks!

Closing this.

PR #3527, opened 2019-02-13
Update search-box type

Meschreiber:
Change <input> type  to  search

**Summary**

A customer complained  that the <input> type  is text and not search. type='search' makes the CTA button on iOS keyboards read 'search' instead of 'done'. I noticed in React IS this is already the case.


Haroenv:
This is actually already fixed in https://github.com/algolia/instantsearch.js/pull/3489. Unfortunately your fix isn't complete however, for it to show as `search`, it also needs to be in a form with an action.

Thanks for this PR, but it won't be needed with the other existing PR. Thanks!

PR #3543, opened 2019-02-18
fix(instantsearch): update usage errors

francoischalifour:
This updates the `InstantSearch` errors, links to the new documentation and add tests.

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -72,7 +72,7 @@ class InstantSearch extends EventEmitter {
       throw new Error(
         `The \`searchClient\` must implement a \`search\` method.
 
-See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend-search/in-depth/backend-instantsearch/js/`
+See: http://alg.li/isjs-backend`
       );
     }
 
@@ -144,7 +144,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       if (widget.render === undefined && widget.init === undefined) {
         throw new Error(`The widget definition expects a \`render\` and/or an \`init\` method.
 
-See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-own-widgets/js/`);
+See: http://alg.li/isjs-custom-widgets`);
       }
 
       this.widgets.push(widget);
@@ -213,7 +213,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
         throw new Error(
           `The \`dispose\` method is required to remove the widget.
 
-See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-own-widgets/js/`
+See: http://alg.li/isjs-widget-lifecycle`
         );
       }
 
@@ -268,7 +268,9 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
    */
   start() {
     if (this.started) {
-      throw new Error(withUsage('The `start` method can only be called once.'));
+      throw new Error(
+        withUsage('The `start` method has already been called once.')
+      );
     }
 
     let searchParametersFromUrl;
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -8,6 +8,11 @@ import simpleMapping from './stateMappings/simple';
 import historyRouter from './routers/history';
 import version from './version';
 import createHelpers from './createHelpers';
+import { createDocumentationMessageGenerator } from './utils';
+
+const withUsage = createDocumentationMessageGenerator({
+  name: 'instantsearch',
+});
 
 const ROUTING_DEFAULT_OPTIONS = {
   stateMapping: simpleMapping(),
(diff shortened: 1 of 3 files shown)


src/lib/InstantSearch.js
```diff
@@ -249,7 +267,9 @@ class InstantSearch extends EventEmitter {
    * @return {undefined} Does not return anything
    */
   start() {
-    if (this.started) throw new Error('start() has been already called once');
+    if (this.started) {
```
tkrugg: I think the new message does not inform the user that `start` has already been called as clearly. I would keep the `already been called once` at least.

src/lib/InstantSearch.js
```diff
@@ -195,7 +211,9 @@ class InstantSearch extends EventEmitter {
         typeof widget.dispose !== 'function'
       ) {
         throw new Error(
-          'The widget you tried to remove does not implement the dispose method, therefore it is not possible to remove this widget'
+          `The \`dispose\` method is required to remove the widget.
+
+See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-own-widgets/js/`
```
tkrugg: why not point directly the lifecycle sections by adding: `#the-widget-lifecycle-and-api` ?

src/lib/InstantSearch.js
```diff
@@ -47,22 +52,27 @@ class InstantSearch extends EventEmitter {
...
+        )
       );
     }
 
     if (typeof searchClient.search !== 'function') {
       throw new Error(
-        'The search client must implement a `search(requests)` method.'
+        `The \`searchClient\` must implement a \`search\` method.
+
+See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend-search/in-depth/backend-instantsearch/js/`
```
tkrugg: How good of an idea is it to include doc links?
I suspect they might 404 one day and it'll be very hard to hunt for those or find out about it. 
I don't have a better solution other than removing them.

Haroenv: The alternative (which I already use in the JS client) is to use `alg.li` links (which are also shorter). You can log in using okta to https://short.cm

src/lib/InstantSearch.js
```diff
@@ -195,7 +211,9 @@ class InstantSearch extends EventEmitter {
         typeof widget.dispose !== 'function'
       ) {
         throw new Error(
-          'The widget you tried to remove does not implement the dispose method, therefore it is not possible to remove this widget'
+          `The \`dispose\` method is required to remove the widget.
```
tkrugg: Is there a reason why we do not enforce a dispose method, upfront, on init?

src/lib/InstantSearch.js
francoischalifour: This is what most libraries do.

It is pointless to say that it's a bad usage without pointing to how to actually use it. The user is not expected to know that the `search` method must return a promise for instance.

It is our responsibility to make sure this link doesn't break. To prevent that, I think Algolia has a URL shortener. @Haroenv can give us more details here.

src/lib/InstantSearch.js
francoischalifour: Indeed, I'll add the anchor.

src/lib/InstantSearch.js
francoischalifour: I guess that's because at the time `dispose` didn't exist and we didn't want to introduce breaking changes for this.

tkrugg: that makes sense. Thanks

Haroenv: I think the reasoning indeed was that; combined with the fact that most people never `removeWidgets`, so the error would be useless at init

francoischalifour:
I applied changes and created short links that guide to the documentation.

src/lib/InstantSearch.js
```diff
@@ -130,7 +142,9 @@ class InstantSearch extends EventEmitter {
     widgets.forEach(widget => {
       // Add the widget to the list of widget
       if (widget.render === undefined && widget.init === undefined) {
-        throw new Error('Widget definition missing render or init method');
+        throw new Error(`The widget definition expects a \`render\` and/or an \`init\` method.
+
+See: http://alg.li/isjs-custom-widgets`);
```
tkrugg: I still think this is going to be hard to manage if the page in the doc gets moved or refactored.
Short is good for bandwidth but it's even harder to grep through pages we moved in the docs now that they are shortened.

francoischalifour: At some point we need to make a decision. We've discussed many approaches and didn't come to a conclusion yet.

(1) If we start adopting the shorten-URL pattern, we need to educate the team to connect to short.cm and create a shorter URL. We don't need to update the source code if the link changes but only in the short.cm dashboard. Another negligible plus is that the links are shorter and therefore the bundle size get's slightly lighter.

(2) If we inline all the links, we need to update the source code whenever a link breaks (this was the initial solution of this PR).

(3) If we don't include links, errors are completely meaningless for the users.

I don't mind going for either (1) or (2).

francoischalifour:
I updated the links to their raw origin (not shortened), waiting for a better strategy for storing and detecting broken links (cc @tkrugg).

src/lib/InstantSearch.js
Haroenv: agreed with you Francois, and the links are easiest. Maybe later some way to make it more automatic (e.g. generated from some GitHub file or something, but that wouldn't change too much IMO)

PR #3547, opened 2019-02-20
feat(insights): POC click analytics integration

tkrugg:
This POC demonstrating a possible integration between our clickAnalytics and InstantSearchJS.

> How we currently do it: https://www.algolia.com/doc/guides/building-search-ui/going-further/plug-analytics/js/?language=js#using-algolia-insights-with-instantsearchjs

### Test this out
1. Go on https://deploy-preview-3547--instantsearchjs.netlify.com/stories/?selectedKind=ClickAnalytics&selectedStory=with%20function%20template&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel
2. open your browser console
2. click on `add to cart`

### How it is used

For instant search user, sending `click` events (or other events types) can be done with the following way

```js
import * as analytics from "../src/helpers/analytics";

// initialize the search-insights clients
aa("init", {
  appId: "APPLICATION_ID",
  apiKey: "SEARCH_API_KEY",
  userHasOptedOut: false
});
aa("setUserToken", "xxx");

  indexName: "instant_search",
  /* ... */
  insightsClient: aa
});

// add the hits widget
search.addWidgets([
    clickAnalytics: true
  }),
  instantsearch.widgets.hits({
    container,
    templates: {
      item: item => `
        <h3> ${item.name} </h3>
        <button ${analytics.clickedObjectIDsAfterSearch(item.objectID, {
          eventName: "Add to basket"
        })}>
           Add to basket
        </button>
      `
    }
  })
]);
```

This is going to call the [clickedObjectIDsAfterSearch](https://github.com/algolia/search-insights.js#reporting-a-click-event) in the following way:

```js
aa("clickedObjectIDsAfterSearch", {
  eventName: "Add to basket", // user provided
  index: "instant_search", // infered from `results`,
  queryID: "***", // infered from `results`,
  objectIDs: ["1"], // infered from `hit`,
  positions: [1] // computed and infered from `hit` and `results`,
});
```
#### With Hogan

```handlebars
<button
  {{#helpers.clickedObjectIDsAfterSearch}}{
    eventName: 'Add to cart'
  }{{/helpers.clickedObjectIDsAfterSearch}}>
  Add to basket
</button>
```

### How it works

There are three parts:

- `connectHits` is using `createAnalyticsHelpers()`:
  - exposes `analytics` object with methods that call the insights client directly (useful for other flavours)
  - inject in every hit `hit.__analytics = { index: "instant_search", queryID: "xxx", objectIDs: ["1"], positions: [10] }`
- helpers `analytics.clickedObjectIDsAfterSearch` are helpers generates data attributes
- the `WithClickAnalyticsWrapper` wraps the Hits component and sets a click listener that reads data attribute and calls connector analytics methods

### Possible improvements and unresolved questions

- Conversion events are often sent outside the contexte of instantsearch. Can we help in that situation?
We expose the queryID so the user can extract the queryID in the following way:
```html
  <a href="https://mywebsite.com/?queryID=${item.__analytics.queryID}"> Add to basket </a>
```
```handlebars
  <a href="https://mywebsite.com/?queryID={{__analytics.queryID}}"> Add to basket </a>
```
- This POC is intended to work on InfiniteHits as well.

### DIFF
--- a/src/components/Hits/withClickAnalytics.js
+++ b/src/components/Hits/withClickAnalytics.js
@@ -1,0 +1,67 @@
+import React, { Component } from 'preact-compat';
+import PropTypes from 'prop-types';
+
+export default C =>
+  class WithClickAnalyticsWrapper extends Component {
+    static propTypes = {
+      hits: PropTypes.array.isRequired,
+      results: PropTypes.object.isRequired,
+    };
+
+    constructor(props) {
+      super(props);
+      this.ref = null;
+    }
+
+    componentDidMount() {
+      const node = this.ref.getDOMNode();
+      node.addEventListener('click', ev => {
+        const event = ev.target.getAttribute('data-analytics-event');
+        const payload = ev.target.getAttribute('data-analytics-payload');
+        if (event && payload) {
+          // event should be oneOf ['clickedObjectIDsAfterSearch', ...]
+          // payload should be valid json
+          let jsonPayload;
+          try {
+            jsonPayload = JSON.parse(payload);
+          } catch (e) {
+            console.error('Invalid JSON event payload');
+            return;
+          }
+
+          // call aa(eventType, eventPayload)
+          console.log(event, jsonPayload);
+        }
+      });
+    }
+
+    render() {
+      const {
+        results: { index, hitsPerPage, page, queryID },
+      } = this.props;
+
+      const hits = this.props.hits.map((hit, position) => ({
+        ...hit,
+        __hitAbsoluteIndex: hitsPerPage * page + position + 1,
+        __index: index,
+        __queryID: queryID,
+        analytics: {
+          clickedObjectIDsAfterSearch: eventName => {
+            const payload = {
+              eventName,
+              index,
+              queryID,
+              objectIDs: [hit.objectID],
+              positions: [hitsPerPage * page + position + 1],
+            };
+            return `
+              data-analytics-event="clickedObjectIDsAfterSearch"
+              data-analytics-payload='${JSON.stringify(payload)}'
+            `;
+          },
+        },
+      }));
+
+      return <C ref={node => (this.ref = node)} {...this.props} hits={hits} />;
+    }
+  };
--- a/src/components/Hits/withClickAnalytics.js
+++ b/src/components/Hits/withClickAnalytics.js
@@ -1,11 +1,16 @@
 import React, { Component } from 'preact-compat';
 import PropTypes from 'prop-types';
+import {
+  readData as readAnalyticsData,
+  hasData as hasAnalyticsData,
+} from '../../helpers/analytics';
 
 export default C =>
   class WithClickAnalyticsWrapper extends Component {
     static propTypes = {
       hits: PropTypes.array.isRequired,
       results: PropTypes.object.isRequired,
+      analytics: PropTypes.object.isRequired,
     };
 
     constructor(props) {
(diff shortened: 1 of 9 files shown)


stories/clickAnalytics.stories.js
```diff
@@ -0,0 +1,27 @@
+import { storiesOf } from '@storybook/html';
+import { withHits } from '../.storybook/decorators';
+
+storiesOf('ClickAnalytics', module).add(
+  'basic',
```
Haroenv: can you give a hogan story too, or is that not finished yet?

tkrugg: if we want to be able to use it in hogan, we'd need to a specific `clickedObjectIDsAfterSearchHogan`.
It's not impossible it's just not pretty.

```js
        analytics: {
          clickedObjectIDsAfterSearch: eventName => {
            const payload = {
              eventName,
              index,
              queryID,
              objectIDs: [hit.objectID],
              positions: [hitsPerPage * page + position + 1],
            };
            return `
              data-analytics-event="clickedObjectIDsAfterSearch"
              data-analytics-payload='${JSON.stringify(payload)}'
            `;
          },
          clickedObjectIDsAfterSearchHogan: () => {
            return eventName => {
              const payload = {
                eventName,
                index,
                queryID,
                objectIDs: [hit.objectID],
                positions: [hitsPerPage * page + position + 1],
              };
              return `
              data-analytics-event="clickedObjectIDsAfterSearch"
              data-analytics-payload='${JSON.stringify(payload)}'
            `;
            };
          },
```
Would it be fine to enforce the use of function templates in clickAnalytics case?

Haroenv: Seeing as the difference is only a function call, what about making the regular one `clickedObjectIDsAfterSearch()`

tkrugg: ok, so we'd be using it in the following way.
```html
<button ${clickedObjectIDsAfterSearch()('Add to cart'))}>
   Add to basket
</button>
```
```handlebars
<button {{#analytics.clickedObjectIDsAfterSearch}}Add to cart{{/analytics.clickedObjectIDsAfterSearch}}>
   Add to basket
</button>
```
But I'm not sure it's ideal.

src/components/Hits/withClickAnalytics.js
```diff
@@ -0,0 +1,67 @@
...
+      const {
+        results: { index, hitsPerPage, page, queryID },
+      } = this.props;
+
+      const hits = this.props.hits.map((hit, position) => ({
+        ...hit,
+        __hitAbsoluteIndex: hitsPerPage * page + position + 1,
+        __index: index,
+        __queryID: queryID,
+        analytics: {
```
francoischalifour: We should be careful about overriding hits' attributes. We might want to rename it something like `_analytics`?

Haroenv: I suggested this as well, but it should be `__analytics` (one for algolia, one for frontend)

tkrugg: yes let's make it private

src/components/Hits/withClickAnalytics.js
```diff
@@ -0,0 +1,67 @@
...
+          clickedObjectIDsAfterSearch: eventName => {
```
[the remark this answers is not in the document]
tkrugg: If I understand your point well, this would require to [pass `results` to the Template component](https://github.com/algolia/instantsearch.js/blob/2a416f8b6e5309b564d3f97824e9e447829fa2a7/src/components/Hits/Hits.js#L24-L34) inside data so that we can [use its data in the helpers](https://github.com/algolia/instantsearch.js/blob/70b3ecd1690f26482166ddc129adf541a06b52f7/src/lib/utils.js#L162-L166), which also means exposing the `results` in every hit. Is this what you had in mind?

stories/clickAnalytics.stories.js
```diff
@@ -0,0 +1,143 @@
+import { storiesOf } from '@storybook/html';
+import { withHits } from '../.storybook/decorators';
+import * as analyticsHelpers from '../src/helpers/analytics';
+
+const insightsClientMock = (eventName, payload) => {
```
francoischalifour: You can use the `action` from Storybook ([see Analytics story](https://github.com/algolia/instantsearch.js/blob/develop/stories/analytics.stories.js)).

src/helpers/analytics.js
```diff
@@ -0,0 +1,59 @@
+// we could decide to have more specific helpers to make the code lighter for func templates
+// clickedObjectIDsAfterSearch(payload)
+// etc
+export function clickedObjectIDsAfterSearch(objectID, payload) {
```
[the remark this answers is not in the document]
tkrugg: This is the helper layer, we only pass the data we want to override.
the exact payload is build by the connector layer so that it can benefit all the flavours.

src/helpers/analytics.js
```diff
@@ -0,0 +1,59 @@
...
+      `The analytics helper expects payload to be either a string or a JSON object`
+    );
+  }
+
+  // TODO: unserialize payload in base64
+  try {
+    payload = JSON.parse(serializedPayload);
+  } catch (e) {
+    // assume it's the eventName
+    payload = { eventName: serializedPayload };
```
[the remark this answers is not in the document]
tkrugg: I want to allow the shorthand     
```html
<button ${analytics.clickedObjectIDsAfterSearch(item.objectID, 'Add to basket')}>
    Add to basket
</button>
<!-- vs -->
<button ${analytics.clickedObjectIDsAfterSearch(item.objectID, {eventName: 'Add to basket'})}>
    Add to basket
</button>
```
but it's debatable and I'm not holding on it

src/lib/utils.js
```diff
@@ -1,13 +1,15 @@
 import reduce from 'lodash/reduce';
 import forEach from 'lodash/forEach';
 import find from 'lodash/find';
+import findIndex from 'lodash/findIndex';
```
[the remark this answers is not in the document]
tkrugg: you're right but I think it has no extra cost here because we are already using `find` it depends of `findIndex` https://github.com/lodash/lodash/blob/4.17.11-npm/find.js#L2

src/helpers/analytics.js
```diff
@@ -0,0 +1,59 @@
...
+  // TODO: serialize payload in base64
+  let serializedPayload;
+  if (typeof payload === 'string') {
+    serializedPayload = payload;
+  } else if (typeof payload === 'object') {
+    try {
+      serializedPayload = JSON.stringify(payload);
+    } catch (e) {
+      throw new Error(
+        `The analytics helper expects payload to be either a string or a JSON object`
```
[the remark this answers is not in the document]
tkrugg: Answered in your next comment 😄  https://github.com/algolia/instantsearch.js/pull/3547#discussion_r260796957

src/helpers/analytics.js
```diff
@@ -0,0 +1,59 @@
...
+      );
+    }
+  }
+  return `data-analytics-method="${method}" data-analytics-object-id="${objectID}" 
+     data-analytics-payload='${serializedPayload}'`;
+}
+export function hasData(domElement) {
+  return !!domElement.getAttribute('data-analytics-method');
```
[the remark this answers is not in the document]
tkrugg: thanks!

tkrugg:
@francoischalifour 

> There's a lot of assumptions in the code because we want to support multiple data types (string and objects). I think this should be stricter. It makes it confusing for the users and it makes the code hard to follow.

As I answered to your comment:
I want to allow the shorthand
```html
<button ${analytics.clickedObjectIDsAfterSearch(item.objectID, 'Add to basket')}>
    Add to basket
</button>
<!-- vs -->
<button ${analytics.clickedObjectIDsAfterSearch(item.objectID, {eventName: 'Add to basket'})}>
    Add to basket
</button>
```
but it's debatable and I'm not holding on it


> I don't find the way the higher-order component is made intuitive in the widget. We could abstract this logic by creating a proxy hit component that uses `HitsWithAnalytics` if the `analytics` prop is passed, the basic `Hits` component otherwise.

The HOC component here acts like the "proxy", do I'm not sure what you mean.
I wanted to keep this PR short but this same HOC will be used to add analytics on infinite-hits
```js
const HitsWithAnalytics = withClickAnalytics(Hits)
const InfiniteHitsWithAnalytics = withClickAnalytics(InfiniteHits)
```
 
> Do you plan on adding tests

Of course! also rewriting the new code in TS as much as possible.

tkrugg:
closing this. Replaced by #3598

PR #3561, opened 2019-03-01
fix(panel): append panel body as a child element

francoischalifour:
**Summary**

There's an issue where Preact loses the `bodyRef` ref when the widget re-renders because it cleans the tree. This issue exists with both `preact` and `preact-compat` (version < `10.0.0`) but not with `react` (tested by @samouss).

This problem makes the collapsible option impossible to implement in the panel. It also causes other issues:

- [Go to this sandbox](https://codesandbox.io/s/2379j3jv2n)
- Decrease the max price
- Select a filter (e.g. "Machine Washable")

The input range width loses the reference of its container.

**Result**

This solution doesn't rely on `ref`s but on plain DOM manipulation. It inserts the body when the Preact component is mounted.

Note that this solution inserts another `div` inside `div.ais-Panel-body` (this is a recurring issue encountered in the GeoSearch component for example without big consequences).

### DIFF
--- a/src/widgets/panel/panel.js
+++ b/src/widgets/panel/panel.js
@@ -12,24 +12,22 @@ import Panel from '../../components/Panel/Panel';
 const withUsage = createDocumentationMessageGenerator({ name: 'panel' });
 const suit = component('Panel');
 
-const renderer = ({ containerNode, cssClasses, templateProps }) => ({
-  options,
-  hidden,
-}) => {
-  let bodyRef = null;
-
+const renderer = ({
+  containerNode,
+  bodyContainerNode,
+  cssClasses,
+  templateProps,
+}) => ({ options, hidden }) => {
   render(
     <Panel
       cssClasses={cssClasses}
       hidden={hidden}
       templateProps={templateProps}
       data={options}
-      onRef={ref => (bodyRef = ref)}
+      bodyElement={bodyContainerNode}
     />,
     containerNode
   );
-
-  return { bodyRef };
 };
 
 /**
@@ -85,6 +83,7 @@ export default function panel({
     `The \`hidden\` option in the "panel" widget expects a function returning a boolean (received "${typeof hidden}" type).`
   );
 
+  const bodyContainerNode = document.createElement('div');
   const cssClasses = {
     root: cx(suit(), userCssClasses.root),
     noRefinementRoot: cx(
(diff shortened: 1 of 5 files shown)


stories/panel.stories.js
```diff
@@ -21,37 +21,40 @@ storiesOf('Panel', module)
     })
   )
   .add(
-    'with ratingMenu',
+    'with range input',
```
francoischalifour: I updated these stories because they conflicted with the playground.

Haroenv:
As expected, this added an extra div between the body and the widget root. I think it's a reasonable tradeoff in this case, although I wonder if there might be another way

francoischalifour:
@samouss Any clue what would be a good way to test this? It seems to be very specific to Preact, and especially this current version. I'm not sure it would make sense to test that since the test can break if we upgrade the Preact dependency someday.

(I added a test on the lifecycle but it's unrelated)

@Haroenv We could mount the widget in a [`DocumentFragment`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) and [insert this fragment](https://github.com/developit/preact/issues/1022#issuecomment-371373161) into the panel body `div`. However, InstantSearch.js doesn't seem to be compatible with `DocumentFragment`s as containers (it doesn't render anything).

samouss:
I don't think it's [an issue](https://preactjs.com/guide/differences-to-react#what-s-different-), it might be an expected behaviour. Indeed it's very specific to Preact like the issue we have. IMO it worth cover it with a test to avoid any regression in the future. We can reproduce the exact same logic that we have inside the Panel and test that the component is not re-created.

francoischalifour:
FYI we decided not to test https://github.com/algolia/instantsearch.js/pull/3561#issuecomment-468727531. This will be fixed by [Preact X](https://github.com/developit/preact/releases/tag/10.0.0-alpha.0).

src/widgets/panel/panel.js
```diff
@@ -112,18 +111,19 @@ export default function panel({
 
     const renderPanel = renderer({
```
samouss: We can remove the indirection created by the `renderer`. We can call avoid to return a new function from the renderer and provide the full list of argument each time call the `render` function. We do this in widgets because we don't have the control on the `render` call (which can be avoided). Here we do have the control (see below). We can do it in another PR though.

```jsx
const renderer = ({
  containerNode,
  bodyContainerNode,
  cssClasses,
  templateProps,
  options,
  hidden
}) => {
  render(
    <Panel
      cssClasses={cssClasses}
      hidden={hidden}
      templateProps={templateProps}
      data={options}
      bodyElement={bodyContainerNode}
    />,
    containerNode
  );
};
```

PR #3563, opened 2019-03-01
fix(searchbox): unmount component on dispose

francoischalifour:
The `unmount` function in the `searchBox` widget wasn't updated correctly since we moved the component to Preact.

This is a remaining of #3489.


francoischalifour:
@Haroenv There's never been any tests for the `dispose` step in any widgets. Definitely something to add in the future.

PR #3574, opened 2019-03-03
 test: add toWarnDev custom Jest matcher

francoischalifour:
This adds a custom Jest matcher to assert that warnings are reported correctly in development mode. This matcher is called `toWarnDev` and is inspired by [the one developed by the React team](https://github.com/facebook/react/blob/9d756d903fdd07895fc320962c3fab95fa099bd5/scripts/jest/matchers/toWarnDev.js).

It was quite cumbersome to assert that connector/widget warnings were reported and led us to sometimes skip these checks.

This PR aims at reducing this friction and providing a tool to improve DX.

### Before

```js
test('with `hidden` as boolean warns', () => {
  const warn = jest.spyOn(global.console, 'warn');
  warn.mockImplementation(() => {});

  panel({
    hidden: true,
  });

  expect(warn).toHaveBeenCalledWith(
    '[InstantSearch.js]: The `hidden` option in the "panel" widget expects...'
  );

  warn.mockRestore();
});
```

#### Usage

```js
test('with `hidden` as boolean warns', () => {
  expect(() => {
    panel({
      hidden: true,
    });
  }).toWarnDev(
    '[InstantSearch.js]: The `hidden` option in the "panel" widget expects...'
  );
});
```

#### Output

-7655b696-8443-472f-a64e-ba01fb570a61untitled


Haroenv:
Would be amazing if it could be an inline snapshot. Is that possible?

francoischalifour:
That would be a lot more difficult because inline snapshots rely on Prettier, file system and custom parsers.

You can learn more in the [original inline snapshots thread](https://github.com/facebook/jest/pull/6380) and in the [Jest source code](https://github.com/facebook/jest/blob/f206afbda05b3b248a33376afe51b28170056bff/packages/jest-snapshot/src/inline_snapshots.ts).

I'm not aware of any helper for that.

I would suggest sticking with regular expectations.

scripts/jest/matchers/__tests__/toWarnDev-test.ts
```diff
@@ -0,0 +1,50 @@
+/* tslint:disable:no-console */
+
+describe('toWarnDev', () => {
+  describe('usage', () => {
+    test('fails with incorrect type of message', () => {
+      expect(() => {
+        // @ts-ignore:next-line
```
tkrugg: why do you need this?
I would have assume you don't since you've added `no-empty: false` to tslint.json

scripts/jest/matchers/__tests__/toWarnDev-test.ts
```diff
@@ -0,0 +1,50 @@
...
+        expect(() => {
+          expect(() => {
+            console.warn('warning');
+          }).toWarnDev();
+        }).not.toThrow();
+      });
+
+      test('fails if not called', () => {
+        expect(() => {
+          expect(() => {}).toWarnDev();
```
tkrugg: here is another that might need to be ignored if you indeed need `// ts-ignore`

scripts/jest/matchers/__tests__/toWarnDev-test.ts
francoischalifour: That's because `toWarnDev` expects a string or nothing. TypeScript complains that boolean is a wrong type.

Since we still have some tests using plain JavaScript, I figured this test can be useful. At some point when all the tests in the codebase are migrated to TypeScript, it could be removed.

scripts/jest/matchers/__tests__/toWarnDev-test.ts
francoischalifour: See [comment above](https://github.com/algolia/instantsearch.js/pull/3574/files#r262070058).

scripts/jest/matchers/toWarnDev.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const matcher = {
+  toWarnDev: (callback: () => void, expectedMessage: string) => {
+    if (expectedMessage !== undefined && typeof expectedMessage !== 'string') {
+      throw new Error(
+        `toWarnDev() requires a parameter of type string but was given ${typeof expectedMessage}.`
+      );
+    }
+
+    // @ts-ignore
```
Haroenv: is there a particular rule that can be ignored here?

francoischalifour: I created #3577 for this. TSLint isn't aware of the `globals` set for ESLint.

PR #3575, opened 2019-03-04
 feat(panel): implement collapsed feature

### LINKED ISSUE #2083: [Feature] Add option to open collapsible based on isRefined
Feature

Sometimes a refinement list might be collapsed by default, but if it has been refined and the user reloads the page, it would be nice to have it opened instead of collapsed.

issue with collapsible refined

Add a property in the `collapsible` object, like `collapsedIfRefined` with a boolean value (default `true` to not change existing behavior).
Expected usage based on current options would be something like this:
```js
search.addWidget(
    ...
    collapsible: {
      collapsed: true,
      collapsedIfRefined: false
    }
  })
);
```

- instantsearch.js

Latest

francoischalifour:
This adds a way for the panel widget to be collapsed.

This feature was first skipped because of time constraint when releasing InstantSearch.js 3. This is a missing feature that was possible to implement in v2.

## Features

- Set the panel as "collapsible"
- Set the panel as "collapsible" with a collapsed state by default
- Drive the collapsed state with the helper state (fixes #2083)

## Usage

```js
search.addWidget(
  instantsearch.widgets.panel({
    collapsed: options => {
      return options && options.state && options.state.query.length === 0;
    },
    templates: {
      header: 'Collapsible panel',
      footer: 'Footer',
    },
  })(instantsearch.widgets.refinementList)({
    container,
    attribute: 'brand',
  })
);
```

## API

This introduces a new option in the `panel` widget called `collasped`.

`collasped`

> `SearchParameters => boolean` | no default

When this function is defined, it marks the panel as "collapsible". When the panel is collapsible, we add the CSS classes defined in the [Panel specs](https://instantsearch-css.netlify.com/widgets/panel/).

### Uncontrolled mode

To always collapse the panel:

```js
panel({
  // ...
  collapsed: () => false,
})
```

To collapse the panel until a query is typed:

```js
panel({
  collapsed: options => {
    return options && options.state && options.state.query.length === 0;
  },
});
```

### Controlled mode

Once the user starts interacting with the panel (with the ➕ and ➖ buttons), the panel enters in a controlled mode. It no longer reacts to the search parameters sent to the `collapsed` function.

## Stories

- [Collapsed panel](https://deploy-preview-3575--instantsearchjs.netlify.com/stories/?selectedKind=Panel&selectedStory=with%20collapsed&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)


src/components/Panel/Panel.js
```diff
@@ -29,6 +46,21 @@ class Panel extends Component {
...
+            className={cssClasses.collapseButton}
+            aria-expanded={!this.state.collapsed}
+            onClick={() => {
+              this.setState(previousState => ({
+                controlled: true,
+                collapsed: !previousState.collapsed,
+              }));
+            }}
+          >
+            {this.state.collapsed ? '➕' : '➖'}
```
samouss: We should provide a template rather than static values.

src/components/Panel/Panel.js
```diff
@@ -4,17 +4,34 @@ import cx from 'classnames';
 import Template from '../Template/Template';
 
 class Panel extends Component {
+  state = {
+    collapsed: this.props.collapsed,
+    controlled: false,
+  };
+
+  static getDerivedStateFromProps(nextProps, prevState) {
```
samouss: It's only available since `8.3.0` of `preact`. We should at least require this version.

src/components/Panel/Panel.js
```diff
@@ -29,6 +46,21 @@ class Panel extends Component {
           />
         )}
 
+        {collapsible && (
+          <button
+            className={cssClasses.collapseButton}
+            aria-expanded={!this.state.collapsed}
+            onClick={() => {
```
samouss: We don't `preventDefault()` which means that the button might trigger an action e.g. nested inside a form.

src/widgets/panel/panel.js
```diff
@@ -119,6 +139,8 @@ export default function panel({
     renderPanel({
       options: {},
       hidden: true,
+      collapsible: Boolean(collapsed),
```
samouss: We can store this value since it does not change across render.

src/widgets/panel/panel.js
```diff
@@ -141,6 +163,11 @@ export default function panel({
         renderPanel({
           options,
           hidden: Boolean(hidden(options)),
+          collapsible: Boolean(collapsed),
+          collapsed:
+            typeof collapsed === 'function'
```
samouss: We can avoid to this condition at each render.

src/components/Panel/Panel.js
```diff
@@ -29,6 +46,21 @@ class Panel extends Component {
           />
         )}
 
+        {collapsible && (
+          <button
```
samouss: Do we want to make the button the only UI that controlled the collapsed state? e.g. we could make the complete header clickable (without losing a11y of course).

francoischalifour: It's a discussion worth triggering. It's not in the specs right now. We should first investigate if this is a pattern that is used elsewhere.

francoischalifour:
@samouss Follow up on https://github.com/algolia/instantsearch.js/pull/3575#discussion_r262019992, I added two templates:

- `collapseButtonMore`
- `collapseButtonLess`

Note that it inserts `span`s in the `button` now.

Let me know if you have other ideas for these template names.

src/components/Panel/Panel.js
Haroenv: IMO we should start with the button. More than that a custom panel could be made, although I'm not sure if that's very practical in just vanilla JS

src/components/Panel/Panel.js
```diff
@@ -29,6 +46,35 @@ class Panel extends Component {
...
+              this.setState(previousState => ({
+                controlled: true,
+                collapsed: !previousState.collapsed,
+              }));
+            }}
+          >
+            {this.state.collapsed ? (
+              <Template
+                {...templateProps}
+                templateKey="collapseButtonMore"
```
samouss: Why two templates rather than one with the argument `collapsed`? Like we do inside the [`showMore`](https://www.algolia.com/doc/api-reference/widgets/refinement-list/js/?language=js#widget-param-showmoretext) template.

francoischalifour: I forgot we went for this. Updating!

src/components/Panel/Panel.js
```diff
@@ -29,6 +46,28 @@ class Panel extends Component {
...
+
+              this.setState(previousState => ({
+                controlled: true,
+                collapsed: !previousState.collapsed,
+              }));
+            }}
+          >
+            <Template
+              {...templateProps}
+              templateKey="collapseButton"
```
samouss: IIRC we went for `xxxText` in the other template no?

francoischalifour: Indeed, we went for [`showMoreText`](https://github.com/algolia/instantsearch.js/blob/9f2aac2d231456af99d40d8719194beddf8ffa2a/src/components/RefinementList/RefinementList.js#L161) for `refinementList`. The main difference is that here it's a `span` and not a `button`.

It could be a bit misleading to call it `xxxText` because it can be something else than just text (just like in a `button` actually).

We can go for `xxxText` for consistency. I'm not sure it's very appropriate though.

stories/panel.stories.js
```diff
@@ -58,4 +58,44 @@ storiesOf('Panel', module)
         })
       );
     })
+  )
+  .add(
+    'with collapsed',
+    withHits(({ search, container, instantsearch }) => {
+      search.addWidget(
+        instantsearch.widgets.panel({
```
samouss: Prefer the usage of a variable for the `panel`. We can consider Storybook part of our documentation. It might be hard to read at first for users that never saw this kind of pattern.

src/components/Panel/__tests__/Panel-test.js
```diff
@@ -14,49 +17,165 @@ const getDefaultProps = () => ({
...
+        ...getDefaultProps(),
+      };
+
+      const wrapper = mount(<Panel {...props} />);
+
+      expect(wrapper.find(`.${cssClasses.root}`).exists()).toBe(true);
+      expect(wrapper.find(`.${cssClasses.noRefinementRoot}`).exists()).toBe(
+        false
+      );
+      expect(wrapper.find(`.${cssClasses.collapsibleRoot}`).exists()).toBe(
```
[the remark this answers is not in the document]
francoischalifour: I think it makes sense to explicitly test that the `collapsibleRoot` CSS class is _not_ added to the DOM.

src/components/Panel/__tests__/Panel-test.js
```diff
@@ -14,49 +17,165 @@ const getDefaultProps = () => ({
...
+        hidden: true,
+      };
+
+      const wrapper = mount(<Panel {...props} />);
+
+      expect(wrapper.find(`.${cssClasses.root}`).exists()).toBe(true);
+      expect(wrapper.find(`.${cssClasses.noRefinementRoot}`).exists()).toBe(
+        true
+      );
+      expect(wrapper.find(`.${cssClasses.collapsibleRoot}`).exists()).toBe(
```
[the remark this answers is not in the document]
francoischalifour: IMO it makes sense to ensure that the `hidden` option doesn't have an effect on the default `collapsibleRoot` CSS class.

src/components/Panel/__tests__/Panel-test.js
```diff
@@ -14,49 +17,165 @@ const getDefaultProps = () => ({
...
+      );
+      expect(wrapper.find(`.${cssClasses.collapsedRoot}`).exists()).toBe(false);
+      expect(wrapper.find(`.${cssClasses.collapseButton}`).exists()).toBe(true);
+      expect(wrapper.find(`.${cssClasses.collapseButton}`).text()).toBe('Less');
+      expect(
+        wrapper.find(`.${cssClasses.collapseButton}`).prop('aria-expanded')
+      ).toBe(true);
+      expect(wrapper.find(`.${cssClasses.body}`).exists()).toBe(true);
+      expect(wrapper.find(`.${cssClasses.header}`).exists()).toBe(true);
+      expect(wrapper.find(`.${cssClasses.footer}`).exists()).toBe(true);
```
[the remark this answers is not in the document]
francoischalifour: IMO it makes sense to ensure that the `collapsible` option doesn't have an effect on the default these CSS classes.

src/components/Panel/__tests__/Panel-test.js
```diff
@@ -14,49 +17,165 @@ const getDefaultProps = () => ({
...
+      );
+      expect(wrapper.find(`.${cssClasses.collapsedRoot}`).exists()).toBe(true);
+      expect(wrapper.find(`.${cssClasses.collapseButton}`).exists()).toBe(true);
+      expect(wrapper.find(`.${cssClasses.collapseButton}`).text()).toBe('More');
+      expect(
+        wrapper.find(`.${cssClasses.collapseButton}`).prop('aria-expanded')
+      ).toBe(false);
+      expect(wrapper.find(`.${cssClasses.body}`).exists()).toBe(true);
+      expect(wrapper.find(`.${cssClasses.header}`).exists()).toBe(true);
+      expect(wrapper.find(`.${cssClasses.footer}`).exists()).toBe(true);
```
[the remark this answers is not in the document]
francoischalifour: IMO it makes sense to ensure that the collapsible option doesn't have an effect on the default these CSS classes.

francoischalifour:
@samouss See algolia/instantsearch-specs#97 for the styling.

src/widgets/panel/panel.js
```diff
@@ -106,7 +129,11 @@ export default function panel({
       );
     }
 
-    const defaultTemplates = { header: '', footer: '' };
+    const defaultTemplates = {
+      header: '',
+      footer: '',
+      collapseButtonText: isPanelCollased => (isPanelCollased ? '➕' : '➖'),
```
samouss: It keeps not working the value is an `object` not a `boolean`. You've to extract the value from the argument.

francoischalifour:
@Haroenv I updated the spec in algolia/instantsearch-specs#99.

PR #3576, opened 2019-03-04
chore: setup canonical meta

samouss:
**Summary**

This PR setup the meta `canonical` for all the pages of the website. If no meta `canonical` is found for the page we fallback to an `noindex, nofollow` meta. Here is the list of changes:

- allow `canonical` meta or fallback on each pages
- setup `canonical` meta tags for top level pages & guides
- setup `canonical` meta tags for the API Ref

### DIFF
--- a/docgen/src/widgets.md
+++ b/docgen/src/widgets.md
@@ -7,6 +7,7 @@ withHeadings: true
 navWeight: 100
 editable: true
 githubSource: docgen/src/widgets.md
+canonical: https://www.algolia.com/doc/guides/building-search-ui/widgets/showcase/js/
 ---
 
 ## Introduction to widgets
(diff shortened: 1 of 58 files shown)


samouss:
@Haroenv Yes indeed, but like you said it's for Google I think it's fine.

docgen/src/widgets.md
```diff
@@ -7,6 +7,7 @@ withHeadings: true
 navWeight: 100
 editable: true
 githubSource: docgen/src/widgets.md
+canonical: https://www.algolia.com/doc/guides/building-search-ui/widgets/showcase/js/
```
tkrugg: there is also this one that can be used here: https://www.algolia.com/doc/api-reference/widgets/js/
No sure which one we should prefer.

Haroenv: the showcase is a nicer link

samouss: I've asked Marie about this one, definitely the showcase!

PR #3578, opened 2019-03-04
chore(scripts): move scripts to their own folders

francoischalifour:
This moves the scripts to [their own folders](https://github.com/algolia/instantsearch.js/tree/28d565079e58501ce22a967317f4c363cd086d99/scripts). It cleans up the structure a bit and make it more consistent.


scripts/ci/functional-test.sh
```diff
@@ -8,6 +8,3 @@ yarn run build
 
 # push the generated screenshots to Argos-CI.
 yarn run test:argos
-
-# check if the next size of the bundle is inside the threshold
-yarn run test:size
```
francoischalifour: This is already run by the Bundlesize GitHub integration.

samouss: Nope this is the command that triggers the GitHub integration.

francoischalifour: Ah, right! Then I'll add the command directly to the [Travis file scripts](https://github.com/algolia/instantsearch.js/blob/chore/scripts-folders/.travis.yml#L19).

Haroenv:
the bundlesize isn't reported in the "checks" section here 🤔  https://travis-ci.org/algolia/instantsearch.js/jobs/501931671 logs aren't totally clear

francoischalifour:
I've noticed Bundlesize was sometimes skipped on some PRs. I'm not sure if this is what's happening here (because the PR was originally opened without Bundlesize set up), or if it's not picked up.

We can try like this – if Bundlesize is not triggered on other PRs, we can add it back to the `functional-tests` script.

samouss:
`bundlesize` won't ever work with this setup, we run it against the build output. With what we have now we never run the build before `bundlesize`. This is the reason why it was in `functional-test` it avoid to build the source multiple time in multiple jobs. Worst case we have to add the `build` step before the call.

francoischalifour:
Right, it makes sense! Will revert this change.

PR #3596, opened 2019-03-11
chore(types): add basic InstantSearch types

francoischalifour:
> This is the 1st PR in the series "Merchandized Query Rules".

This adds basic InstantSearch types necessary for the `connectQueryRules` connector.

These types are meant to be completed over time.

### DIFF
--- a/src/types/connector.ts
+++ b/src/types/connector.ts
@@ -1,0 +1,11 @@
+import { InstantSearch } from './instantsearch';
+
+export type RenderOptions<T = unknown> = {
+  widgetParams: T;
+  instantSearchInstance: InstantSearch;
+};
+
+export type Renderer<T extends RenderOptions> = (
+  renderOptions: T,
+  isFirstRender: boolean
+) => void;
--- a/src/types/widget.ts
+++ b/src/types/widget.ts
@@ -1,0 +1,23 @@
+import {
+  Helper,
+  HelperState,
+  SearchResults,
+  InstantSearch,
+  SearchParameters,
+} from './instantsearch';
+
+interface WidgetArgs {
+  helper: Helper;
+  state?: HelperState;
+  results?: SearchResults;
+  instantSearchInstance: InstantSearch;
+}
+
+export interface Widget {
+  init(options: WidgetArgs): void;
+  render(options: WidgetArgs): void;
+  dispose(options: any): SearchParameters;
+  getConfiguration?(previousConfiguration: SearchParameters): SearchParameters;
+}
+
+export type WidgetFactory<T> = (widgetParams: T) => Widget;
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -1,0 +1,3 @@
+export * from './instantsearch';
+export * from './connector';
+export * from './widget';
--- a/src/types/instantsearch.ts
+++ b/src/types/instantsearch.ts
@@ -1,0 +1,16 @@
+export type InstantSearch = {
+  templatesConfig?: object;
+};
+
+export type Helper = any;
+
+export type HelperState = any;
+
+export type SearchParameters = any;
+
+export type SearchResults = any;
+
+export type FacetRefinement = {
+  value: string;
+  type: string;
+};


types/widget.ts
```diff
@@ -0,0 +1,23 @@
...
+  results?: SearchResults;
+  instantSearchInstance: InstantSearch;
+}
+
+export interface Widget {
+  init(options: WidgetArgs): void;
+  render(options: WidgetArgs): void;
+  dispose(options: any): SearchParameters;
+  getConfiguration?(previousConfiguration: SearchParameters): SearchParameters;
+}
```
Haroenv: why not add the routing functions here?

types/connector.ts
```diff
@@ -0,0 +1,11 @@
+import { InstantSearch } from './instantsearch';
+
+export type RenderOptions<T = unknown> = {
+  widgetParams: T;
+  instantSearchInstance: InstantSearch;
+};
```
Haroenv: I think helper state & search metadata is also sent by default, no?

types/widget.ts
francoischalifour: The goal was to write the bare minimum to implement the Query Rules connector. We can add them when we need them.

src/types/connector.ts
```diff
@@ -0,0 +1,11 @@
+import { InstantSearch } from './instantsearch';
+
+export type RenderOptions<T = unknown> = {
```
Haroenv: we haven't decided on this yet, but I'd love to ban 1-letter generics, let's discuss this another time when we find a eslint rule for it

src/types/widget.ts
```diff
@@ -0,0 +1,23 @@
...
+  helper: Helper;
+  state?: HelperState;
+  results?: SearchResults;
+  instantSearchInstance: InstantSearch;
+}
+
+export interface Widget {
+  init(options: WidgetArgs): void;
+  render(options: WidgetArgs): void;
+  dispose(options: any): SearchParameters;
```
samouss: We consider the `dispose` method required?

src/types/widget.ts
```diff
@@ -0,0 +1,23 @@
...
+interface WidgetArgs {
+  helper: Helper;
+  state?: HelperState;
+  results?: SearchResults;
+  instantSearchInstance: InstantSearch;
+}
+
+export interface Widget {
+  init(options: WidgetArgs): void;
+  render(options: WidgetArgs): void;
```
samouss: The `init` & `render` methods don't share the same arguments e.g. `searchMetaData`, `results`.

https://github.com/algolia/instantsearch.js/blob/3de59a32a08df71e0d4ba25af060560d4d6495ec/src/lib/InstantSearch.js#L416-L423

https://github.com/algolia/instantsearch.js/blob/3de59a32a08df71e0d4ba25af060560d4d6495ec/src/lib/InstantSearch.js#L392-L402

src/types/widget.ts
```diff
@@ -0,0 +1,23 @@
...
+
+interface WidgetArgs {
+  helper: Helper;
+  state?: HelperState;
+  results?: SearchResults;
+  instantSearchInstance: InstantSearch;
+}
+
+export interface Widget {
+  init(options: WidgetArgs): void;
```
samouss: This is not entirely true a widget can have only a `render` or an `init` method.

https://github.com/algolia/instantsearch.js/blob/3de59a32a08df71e0d4ba25af060560d4d6495ec/src/lib/InstantSearch.js#L144

src/types/widget.ts
francoischalifour: We can reiterate on this widget signature. @tkrugg mentioned working on copying this from the one we've got in Angular InstantSearch.

src/types/widget.ts
francoischalifour: A widget can also have both, but at least one of them.

How would you translate that to TypeScript?

Haroenv: A discriminate union between two types with it required `type Widget = ({ init(): void } | { render(): void }) & { ...commonStuff }`

PR #3597, opened 2019-03-11
feat(queryRules): add connectQueryRules connector

francoischalifour:
_This is the 1st PR in the series "Merchandized Query Rules"._

## Summary

This creates the `connectQueryRules` connector, responsible for handling Algolia Query Rules features.

## Motivation

> Implementing Query Rules inside InstantSearch allows users to promote results and translate search intent with minimal setup and configuration.

Query Rules is a powerful feature in the Algolia engine that is complex to implement with InstantSearch. This feature is often used on E-commerce websites but requires developers to use low-level primitives in our libraries.

Users often need to adapt the search experience based on **specific queries** (e.g. showing a banner to promote a brand new MacBook). This is possible by creating new rules (either on the Algolia dashboard, via the API clients or the REST endpoints). These rules can return custom data as JSON format. The Algolia API returns this data as `userData`.

Banner

<h6 align="center">Banner promoting results on Amazon</h6>

Query Rules offer a **custom experience based on rule contexts**. You might want to customize the users' experience based on the filters of the search (e.g. you're visiting the "Mobile" category). This is achieved by setting a context when creating a rule and providing the API parameter `ruleContexts` when searching. Specifying `ruleContexts` at search time creates a lot of friction without dedicated API for the developer (this is doable via the internal helper or with the `configure` widget).

## Implementation

The connector gets the results from the helper and passes the `userData` array to the widgets. I decided to keep the name `userData` in the connector, and not `items` in case we want to expose more objects that could also be called `items`. If you think this is safe enough, we can rename `userData` to `items`.

The connector is also responsible for manipulating the `ruleContexts` with the helper to configure search parameters.

## Connector usage

Two widgets will be based on this connector:

- [`queryRuleCustomData`](https://github.com/algolia/instantsearch.js/pull/3600)
- [`queryRuleContext`](https://github.com/algolia/instantsearch.js/pull/3602)

## Related

- [`queryRuleCustomData`](https://github.com/algolia/instantsearch.js/pull/3600)
- [`queryRuleContext`](https://github.com/algolia/instantsearch.js/pull/3602)


src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -0,0 +1,237 @@
+import noop from 'lodash/noop';
+import isEqual from 'lodash/isEqual';
```
Haroenv: do we not have babel-plugin-lodash here?

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -0,0 +1,237 @@
...
+  const ruleContexts = transformRuleContexts(allRuleContexts).slice(0, 10);
+
+  return ruleContexts;
+}
+
+function applyRuleContexts({ helper, ruleContexts }) {
+  const previousRuleContexts = helper.getQueryParameter('ruleContexts');
+
+  if (!isEqual(previousRuleContexts, ruleContexts)) {
+    helper.setQueryParameter('ruleContexts', ruleContexts).search();
```
francoischalifour: When refining an item in a refinement list for example, this will trigger 2 network requests because of this implementation.

1. Send request to refine the filter
2. Send request with the new `ruleContexts`

I'm not sure how to avoid that.

Haroenv:
> The connector gets the results from the helper and passes the userData array to the widgets. I decided to keep the name userData in the connector, and not items in case we want to expose more objects that could also be called items. If you think this is safe enough, we can rename userData to items.


Makes sense to me!

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -0,0 +1,237 @@
...
+  // We store the initial rule contexts applied before creating the widget
+  // so that we do not override them with the rules created from `trackedFilters`.
+  let initialRuleContexts: string[] = [];
+
+  // We track the added rule contexts to remove them when the widget is disposed.
+  let addedRuleContexts: string[] = [];
+
+  return widgetParams => {
+    const {
+      trackedFilters = {} as ParamTrackedFilters,
```
francoischalifour: Unsure about this casting pattern. It's a way to tell TypeScript that the widget parameters are not `undefined` at this step.

src/connectors/query-rules/connectQueryRules.ts
francoischalifour: We do not use it in the InstantSearch.js codebase.

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -0,0 +1,237 @@
...
+}: {
+  helper: Helper;
+  trackedFilters: ParamTrackedFilters;
+}) {
+  const ruleContexts = Object.keys(trackedFilters).reduce(
+    (facets: string[], facetName) => {
+      const getFacetValues = trackedFilters[facetName];
+      const facetRefinements: string[] = helper
+        .getRefinements(facetName)
+        .map((refinement: FacetRefinement) => refinement.value);
```
samouss: We only support `facet` but we could receive a `numeric` refinement? It won't be a `string` but `number[]`.

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -0,0 +1,237 @@
...
+
+function getRuleContextsFromTrackedFilters({
+  helper,
+  trackedFilters,
+}: {
+  helper: Helper;
+  trackedFilters: ParamTrackedFilters;
+}) {
+  const ruleContexts = Object.keys(trackedFilters).reduce(
+    (facets: string[], facetName) => {
```
samouss:
```suggestion
    (facets, facetName) => {
```

We can avoid it with the generic of `reduce<string[]>()`.

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -0,0 +1,237 @@
...
+        .getRefinements(facetName)
+        .map((refinement: FacetRefinement) => refinement.value);
+      const facetValues = getFacetValues(facetRefinements);
+
+      return [
+        ...facets,
+        ...facetRefinements
+          .filter(
+            facetRefinement =>
+              facetValues.findIndex(
```
samouss: `findIndex` is not supported in IE and we don't list it inside [the doc](https://www.algolia.com/doc/guides/building-search-ui/installation/js/?language=js#browser-support) browser support. But we list `includes` so we can use it (which simplify a bit the condition).

src/connectors/query-rules/connectQueryRules.ts
samouss: Right now there is none. It's fine for now, we can add the batch to the list of things that the helper can do for us.

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -0,0 +1,237 @@
...
+
+          applyRuleContexts({
+            helper,
+            ruleContexts,
+          });
+        }
+
+        render(
+          {
+            userData: [],
```
samouss: It feels weird to have it called like that specially when `transformItems` is called on it.

francoischalifour: `transformItems` is only called on `render` for now.

Should we rather return this?

```js
render(
  {
    userData: transformItems([]),
    instantSearchInstance,
    widgetParams,
  },
  true
);
```

src/connectors/query-rules/__tests__/connectQueryRules-test.ts
```diff
@@ -0,0 +1,450 @@
...
+      const helper = createFakeHelper();
+      const widget = makeWidget({});
+
+      widget.init({ helper, state: helper.state, instantSearchInstance: {} });
+      widget.dispose({ state: helper.getState() });
+
+      expect(unmountFn).toHaveBeenCalledTimes(1);
+    });
+
+    test('calls the unmount function on dispose with trackedFilters', () => {
```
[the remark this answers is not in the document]
francoischalifour: The reason why I added this test was to make sure that `unmount` is called in `dispose` in all cases because we return the state early.

For example, the test would fail if the implementation was like that:

```js
dispose({ state }) {
  if (Object.keys(trackedFilters).length > 0) {
    const reinitRuleContexts = state
      .getQueryParameter('ruleContexts')
      .filter((rule: string) => !addedRuleContexts.includes(rule));

    return state.setQueryParameter('ruleContexts', reinitRuleContexts);
  }

  // Does not get called because `trackedFilters` was defined.
  unmount();

  return state;
},
```

The test doesn't provide much value in itself. I wrote it because when refactoring the `dispose` function I saw some potential misbehaviors here.

I can remove it if it feels unnecessary.

src/connectors/query-rules/connectQueryRules.ts
francoischalifour: It's not typed at this step because TypeScript doesn't infer the `QueryRulesConnectorParams` types.

image

src/connectors/query-rules/connectQueryRules.ts
francoischalifour: Cool, thank you 👍

src/connectors/query-rules/__tests__/connectQueryRules-test.ts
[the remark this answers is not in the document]
samouss: Ok, maybe a cleare test name would help to understand why it’s there.

src/connectors/query-rules/connectQueryRules.ts
samouss: That's unexpected, it should be inferred I think.

src/connectors/query-rules/connectQueryRules.ts
samouss: Actually I wasn't talking about that but more about the usage of the API. It's not natural to use `transformItems` on something that at the end is called `userData`. But I get the point that since query rules is a larger spectrum maybe we'll have `items` at some point.

src/connectors/query-rules/__tests__/connectQueryRules-test.ts
```diff
@@ -252,6 +252,51 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/query-rules
...
+        });
+
+        widget.render({
+          helper,
+          results: new SearchResults(helper.getState(), [{}, {}]),
+          instantSearchInstance: {},
+        });
+
+        expect(helper.getQueryParameter('ruleContexts')).toEqual([
+          'ais-price-500',
```
samouss: We only send one rule context? Is it expected?

src/connectors/query-rules/connectQueryRules.ts
samouss: Actually after talking a bit with Marie & Julien it might be problematic. We can have a blink:

1. select the refinement -> a request is triggered
2. render with the results, context is applied -> a new request is triggered
3. a query rule is triggered based on the context that set a new filter
4. render with other results, which means that at the step 2 we get stalled results

src/connectors/query-rules/__tests__/connectQueryRules-test.ts
francoischalifour: Yes, the `trackedFilters` value only returns `500` in the test:

```js
const widget = makeWidget({
  trackedFilters: {
    price: () => [500],
  },
});
```

See https://github.com/algolia/instantsearch.js/pull/3597/files#diff-1586d17bb858d5402e71bd206ba2d78fR256

src/connectors/query-rules/connectQueryRules.ts
francoischalifour: Is that an acceptable tradeoff for now? (I would say so)

src/types/instantsearch.ts
```diff
@@ -11,6 +11,6 @@ export type SearchParameters = any;
 export type SearchResults = any;
 
 export type FacetRefinement = {
-  value: string;
+  value: string | number[];
```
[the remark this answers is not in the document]
francoischalifour: You're right, it makes more sense!

src/connectors/query-rules/__tests__/connectQueryRules-test.ts
samouss: Ok but it doesn't seems that we test the behaviour where we return all the value from a refinement. We only test the behaviour with hard coded value.

src/connectors/query-rules/connectQueryRules.ts
samouss: It worth trying with a query rule that alter the results to see the actual impact. We can decide then if it's a tradeoff that is acceptable or not.

src/connectors/query-rules/__tests__/connectQueryRules-test.ts
francoischalifour: We test that the function is called we all the values for this: https://github.com/algolia/instantsearch.js/pull/3597/files#diff-1586d17bb858d5402e71bd206ba2d78fR297.

I will add a test that checks that all the values can effectively be tracked.

src/connectors/query-rules/connectQueryRules.ts
francoischalifour: I promoted the movie "Gone girl" to position 2 in [this movie demo](https://movies-query-rules.now.sh/) when you click on the genre "Drama".

There's indeed a flash, which is noticeable.

francoischalifour:
The `queryRuleContext` widget, responsible for setting `ruleContexts` based on tracked filters, reaches a technical limitation right now that might be problematic.

We internally need to trigger two requests because of the way the [JS helper](https://github.com/algolia/algoliasearch-helper-js) is designed, as explained in https://github.com/algolia/instantsearch.js/pull/3597#discussion_r264310436.

I reproduced the behavior in [the Query Rules demo](https://movies-query-rules.now.sh/). To reproduce:

1. Click on the genre "Drama"
2. See two different result sets loading one after the other

This is an unwanted behavior because:

- There's a flash in the UI (even more noticeable on slow networks)
- It triggers two requests → undesirable for customers

Some work is being done on the helper because of the needs of Federated Search. We could add "batching requests" in the features that we want the helper to support.

The release of the `queryRuleContext` widget will therefore be postponed until the "batching request" feature lands in the helper.

The `queryRuleCustomData` widget can still be shipped without this.

We can either keep or temporarily remove the `ruleContexts`-related feature from the Query Rules connector.

francoischalifour:
⬆️ I removed the `ruleContexts`-related features from the connector and only commented the tests so that we can reuse them when we're ready to reimplement these features (I didn't `skip` the tests because TypeScript was complaining about the now non-existant `trackedFilters` and `transformRuleContexts` options).

src/connectors/query-rules/__tests__/connectQueryRules-test.ts
```diff
@@ -0,0 +1,568 @@
...
+
+        const [renderingParameters] = renderFn.mock.calls[1];
+
+        const { userData } = renderingParameters;
+
+        expect(userData).toEqual({ banner: 'image1.png' });
+      });
+    });
+
+    //     describe('trackedFilters', () => {
```
samouss: We dropped the implementation we can drop the tests too. Save them on a different branch to avoid to loose them, but no need to push them immediately.

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -0,0 +1,78 @@
...
+        );
+      },
+
+      render({ results, instantSearchInstance }) {
+        const { userData: rawUserData = [] } = results;
+        const userData = transformItems(rawUserData);
+
+        render(
+          {
+            userData,
```
samouss: Don't forget to update the name to `items`. 

> https://github.com/algolia/react-instantsearch/pull/2210

PR #3598, opened 2019-03-11
feat(Insights): Insights inside Instantsearch

tkrugg:
**Summary**

This PR follows up on #3547 and aims to introduce 3 things
- `withInsightsClient`: connector wrapper that exposes insights client to connectors
- `helpers/insights`: which adds helpers that generate `data-insight-*` attributes 
- `withInsightsListener`: which wraps Hits and InfiniteHits components and listens to clicks on elements containing `data-insights-*` attributes.

**Result**
WIP


src/lib/insights/client.js
```diff
@@ -0,0 +1,54 @@
+export const _inferPayload = (method, results, objectIDs) => {
```
tkrugg: should have a public path

src/widgets/hits/hits.js
```diff
@@ -28,12 +29,15 @@ const renderer = ({ renderState, cssClasses, containerNode, templates }) => (
     return;
   }
 
+  const HitsWithInsightsListener = withInsightsListener(Hits);
```
tkrugg: to upper scope

src/widgets/hits/hits.js
```diff
@@ -121,7 +125,7 @@ You may want to migrate using \`connectHits\`: ${createDocumentationLink({
     templates,
   });
 
-  const makeHits = connectHits(specializedRenderer, () =>
+  const makeHits = withInsightsClient(connectHits)(specializedRenderer, () =>
```
tkrugg: let's call this withInsights

src/lib/insights/client.js
tkrugg: @Haroenv You mentionned in the review we should expose this function. I would rather not expose it at this point until we get a clearer view of what we should pass to it?
I don't think there is any emergency to expose it, right?

src/types/connector.ts
```diff
@@ -1,11 +1,23 @@
-import { InstantSearch } from './instantsearch';
+import { Hits, InstantSearch } from './instantsearch';
+import { InsightsClient } from './insights';
 
-export type RenderOptions<T = unknown> = {
-  widgetParams: T;
+// export type RenderOptionsWithInsights<T> = RenderOptions<T> implements
```
tkrugg: No sure whether I should do something like this instead of `insights?`

src/lib/__tests__/insights-client-test.js
```diff
@@ -0,0 +1,230 @@
...
+    { objectID: '2', __position: 10 },
+    { objectID: '3', __position: 11 },
+    { objectID: '4', __position: 12 },
+  ];
+  const results = {
+    index: 'theIndex',
+    queryID: 'theQueryID',
+    hits,
+  };
+  describe('payload inferring', () => {
```
francoischalifour: Add an empty line before the `describe`.

src/lib/__tests__/insights-listener-test.js
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
```
francoischalifour: Shouldn't this test be written in TypeScript?

src/lib/insights/client.ts
```diff
@@ -0,0 +1,99 @@
...
+  results: RenderOptions['results'],
+  hits: Hits<unknown>,
+  objectIDs: string[]
+): InsightsClientPayload => {
+  const { index, queryID } = results;
+  switch (method) {
+    case 'clickedObjectIDsAfterSearch': {
+      const positions = objectIDs.map(objectID => {
+        const hit = find(hits, hit => hit.objectID === objectID);
+        // NOTE: we might want to optimize this so as we browse though the
```
francoischalifour:
```suggestion
        // NOTE: we might want to optimize this so as we browse through the
```

src/lib/insights/client.ts
```diff
@@ -0,0 +1,99 @@
...
+        // hits once, which can get pretty big
```
francoischalifour: The maximum number of hits returned by the engine is 1000 I think. In the UI it's very rare that we display more than 100 hits at once so it's not a big concern IMO.

src/lib/insights/client.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const _wrapInsightsClient = (
+  aa: InsightsClient,
+  results: RenderOptions['results'],
+  hits: Hits<unknown>
+): InsightsClient => (
+  method: InsightsClientMethod,
+  payload: InsightsClientPayload
+) => {
+  if (typeof aa !== 'function') {
```
francoischalifour: Isn't this check already performed there?

https://github.com/algolia/instantsearch.js/pull/3598/files#diff-bde03140d12ec15ad28983aea70e6f69R84

src/lib/insights/client.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    case 'convertedObjectIDsAfterSearch':
+      return { index, queryID, objectIDs };
+
+    default:
+      throw new Error(`Unsupported method passed ${method}`);
+  }
+};
+
+const _wrapInsightsClient = (
```
francoischalifour: Not sure about the `_` prefix. We don't use this often in the codebase.

src/lib/insights/client.ts
```diff
@@ -0,0 +1,99 @@
...
+
+type Connector<WidgetParams> = (
+  renderFn: Renderer<RenderOptions<WidgetParams>>,
+  unmountFn: Unmounter
+) => WidgetFactory<WidgetParams>;
+
+export default function withInsights(
+  connector: Connector<unknown>
+): Connector<unknown> {
+  const _wrapRenderFn = (
```
francoischalifour: Not sure about the `_` prefix. We don't use this often in the codebase.

src/lib/insights/client.ts
```diff
@@ -0,0 +1,99 @@
...
+    renderOptions: RenderOptions,
+    isFirstRendering: boolean
+  ) => {
+    const { results, instantSearchInstance, hits } = renderOptions;
+    if (
+      hits &&
+      instantSearchInstance &&
+      instantSearchInstance.insightsClient /* providing the insightsClient is optional */
+    ) {
+
```
francoischalifour:
```suggestion
```

src/lib/insights/client.ts
```diff
@@ -0,0 +1,99 @@
...
+) => WidgetFactory<WidgetParams>;
+
+export default function withInsights(
+  connector: Connector<unknown>
+): Connector<unknown> {
+  const _wrapRenderFn = (
+    renderFn: Renderer<RenderOptions<unknown>>
+  ): Renderer<RenderOptions<unknown>> => (
+    renderOptions: RenderOptions,
+    isFirstRendering: boolean
```
francoischalifour: Recently we took the initiative to rename this by `isFirstRender` (that's the new convention).

src/lib/insights/listener.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'preact-compat';
+import PropTypes from 'prop-types';
```
francoischalifour: We don't need `PropTypes` anymore in InstantSearch.js since it's a TypeScript component.

src/lib/insights/listener.tsx
```diff
@@ -0,0 +1,36 @@
...
+    };
+
+    return (
+      <div onClick={handleClick}>
+        <BaseComponent {...props} />
+      </div>
+    );
+  }
+
+  WithInsightsListener.propTypes = {
```
francoischalifour: We can remove these `propTypes`.

src/lib/insights/listener.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { readDataAttributes, hasDataAttributes } from '../../helpers/insights';
+import { InsightsClient } from '../../types';
+
+type WithInsightsListenerProps = {
+  insights: InsightsClient;
+};
+
+export default BaseComponent => {
+  function WithInsightsListener(props: WithInsightsListenerProps) {
+
```
francoischalifour:
```suggestion
```

src/lib/insights/listener.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const handleClick = ev => {
```
francoischalifour: Can we avoid to shorten variable names and explicitly call this `event`?

src/lib/insights/listener.tsx
```diff
@@ -0,0 +1,36 @@
...
+      if (!hasDataAttributes(ev.target)) return;
```
francoischalifour: Can we avoid inline returns?

We could add an ESlint rule for that.

src/lib/utils/__tests__/index-test.js
```diff
@@ -1165,3 +1166,16 @@ describe('utils.getPropertyByPath', () => {
     expect(utils.getPropertyByPath(object, 'nested.name')).toBe(undefined);
   });
 });
+
+describe('utils.withAbsolutePosition', () => {
```
francoischalifour: For new utils we can create specific test files (e.g. `utils/__tests__/hits-absolute-position-test.js`).

src/lib/insights/listener.tsx
francoischalifour: The event can be typed `React.MouseEvent<HTMLElement>`.

src/lib/insights/listener.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'preact-compat';
+import PropTypes from 'prop-types';
+import { readDataAttributes, hasDataAttributes } from '../../helpers/insights';
+import { InsightsClient } from '../../types';
+
+type WithInsightsListenerProps = {
+  insights: InsightsClient;
+};
+
+export default BaseComponent => {
```
francoischalifour: `BaseComponent` could be typed.

stories/hits.stories.js
```diff
@@ -1,5 +1,7 @@
 import { storiesOf } from '@storybook/html';
 import { withHits } from '../.storybook/decorators';
+import insights from '../src/helpers/insights';
+import { action } from '@storybook/addon-actions';
```
francoischalifour: This import should be on the second line.

stories/infinite-hits.stories.js
```diff
@@ -1,5 +1,7 @@
 import { storiesOf } from '@storybook/html';
 import { withHits } from '../.storybook/decorators';
+import insights from '../src/helpers/insights';
+import { action } from '@storybook/addon-actions';
```
francoischalifour: This import should be on the second line.

stories/infinite-hits.stories.js
```diff
@@ -68,4 +70,39 @@ storiesOf('InfiniteHits', module)
...
+                eventName: 'Add to cart',
+              })} >Add to cart</button>
+            `,
+            },
+          })
+        );
+      },
+      {
+        searchParameters: {
+          clickAnalytics: true,
```
francoischalifour: Since we use `configure` above, can't we set `clickAnalytics` there?

stories/hits.stories.js
```diff
@@ -105,4 +107,39 @@ storiesOf('Hits', module)
...
+                eventName: 'Add to cart',
+              })} >Add to cart</button>
+            `,
+            },
+          })
+        );
+      },
+      {
+        searchParameters: {
+          clickAnalytics: true,
```
francoischalifour: Since we use `configure` above, can't we set `clickAnalytics` there?

src/types/connector.ts
```diff
@@ -1,11 +1,23 @@
...
+  };
+  hits?: Hits<any>;
+  insights?: InsightsClient;
 };
 
 export type Renderer<T extends RenderOptions> = (
   renderOptions: T,
   isFirstRender: boolean
 ) => void;
+export type Unmounter = () => void;
```
francoischalifour:
```suggestion

export type Unmounter = () => void;
```

src/helpers/__tests__/insights-test.ts
```diff
@@ -0,0 +1,165 @@
+import insights, {
+  writeDataAttributes,
+  readDataAttributes,
+  hasDataAttributes,
+} from '../insights';
+
+const makeDomElement = (html: string): HTMLElement => {
+  document.body.innerHTML = `<div id="content">${html}</div>`;
+  const content = document.querySelector('#content');
+  if (!content) return document.body;
```
francoischalifour: Can we avoid inline returns?

src/helpers/__tests__/insights-test.ts
```diff
@@ -0,0 +1,165 @@
...
+        > Add to Cart </button>`
+    );
+
+    expect(hasDataAttributes(domElement)).toBe(false);
+  });
+});
+
+describe('readDataAttributes', () => {
+  describe('on handwritten data-insights-* attributes', () => {
+    let domElement;
```
francoischalifour:
```suggestion
    let domElement: HTMLElement;
```

src/helpers/__tests__/insights-test.ts
```diff
@@ -0,0 +1,165 @@
...
+      expect(() =>
+        readDataAttributes(domElement)
+      ).toThrowErrorMatchingInlineSnapshot(
+        `"Unable to parse \`data-insights-payload\`"`
+      );
+    });
+  });
+
+  describe('on data-insights-* attributes generated with insights helper', () => {
+    let domElement;
```
francoischalifour:
```suggestion
    let domElement: HTMLElement;
```

src/helpers/insights.ts
```diff
@@ -0,0 +1,55 @@
...
+  } catch (e) {
+    throw new Error('Unable to parse `data-insights-payload`');
+  }
+}
+
+export function hasDataAttributes(domElement: HTMLElement): boolean {
+  return domElement.hasAttribute('data-insights-method');
+}
+
+export function writeDataAttributes({
```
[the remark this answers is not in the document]
tkrugg: Having a write vs. read helps keep the logic easy to follow imo.
the insights helpers can look like merely a proxy to writeDataAttributes but the signature is different and closer to the insight client signature, which I think has value.

src/lib/__tests__/insights-client-test.js
```diff
@@ -0,0 +1,230 @@
...
+
+describe('withInsights', () => {
+  let instantSearchInstance;
+  let hits;
+  let results;
+  let widgetParams;
+  let widget;
+  let renderFn;
+
+  beforeEach(() => {
```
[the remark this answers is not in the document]
samouss: Agree, it simpler to reason about IMO.

src/helpers/insights.ts
```diff
@@ -0,0 +1,55 @@
...
+    return {
+      method: <InsightsClientMethod>method,
+      payload: <InsightsClientPayload>payload,
+    };
+  } catch (e) {
+    throw new Error('Unable to parse `data-insights-payload`');
+  }
+}
+
+export function hasDataAttributes(domElement: HTMLElement): boolean {
```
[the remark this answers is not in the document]
tkrugg: I'd prefer to keep it if you don't mind, as it's explicit.

src/lib/__tests__/insights-client-test.js
```diff
@@ -0,0 +1,230 @@
...
+    { objectID: '4', __position: 12 },
+  ];
+  const results = {
+    index: 'theIndex',
+    queryID: 'theQueryID',
+    hits,
+  };
+  describe('payload inferring', () => {
+    it('should infer queryID from results', () => {
+      const payload = inferInsightsPayload(
```
samouss: We should use named parameter it helps for the comprehension.

src/helpers/insights.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export function writeDataAttributes({
+  method,
+  payload,
+}: {
+  method: InsightsClientMethod;
+  payload: InsightsClientPayload;
+}) {
+  let serializedPayload;
```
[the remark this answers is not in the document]
tkrugg: Since the error I want to catch needs to be only JSON serialisation related, I think it's best to keep only that instruction in the try block.
I'll move `if (typeof payload !== 'object')` to the first line of the function.

src/lib/insights/client.ts
```diff
@@ -0,0 +1,99 @@
...
+  ) => {
+    const { results, instantSearchInstance, hits } = renderOptions;
+    if (
+      hits &&
+      instantSearchInstance &&
+      instantSearchInstance.insightsClient /* providing the insightsClient is optional */
+    ) {
+
+      renderOptions.insights = _wrapInsightsClient(
```
samouss: We could avoid to mutate the parameters when possible.

src/lib/utils/hits-absolute-position.ts
```diff
@@ -0,0 +1,12 @@
+import { Hits } from '../../types';
+
+export const addAbsolutePosition = (
```
samouss: We could use named parameters.

src/types/connector.ts
samouss: `RenderOptions` is the most generic type that other types have to extend. We should add on it only the common options.

src/types/insights.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type InsightsClientPayload = {
+  eventName?: string;
+  queryID?: string;
+  index?: string;
+  objectIDs?: string[];
+  positions?: number[];
+};
+
+export type InsightsClient = (
```
samouss: We could enforce which `payload` match which `eventName` with an overload. That way we can have more specific `payload` with less optionals.

src/types/widget.ts
```diff
@@ -20,4 +20,4 @@ export interface Widget {
   getConfiguration?(previousConfiguration: SearchParameters): SearchParameters;
 }
 
-export type WidgetFactory<T> = (widgetParams: T) => Widget;
+export type WidgetFactory<WidgetParams> = (widgetParams: WidgetParams) => Widget;
```
samouss:
```suggestion
export type WidgetFactory<TWidgetParams> = (widgetParams: TWidgetParams) => Widget;
```

Keep the `T` to keep the information that it's a generic.

src/connectors/hits/connectHits.js
```diff
@@ -3,6 +3,7 @@ import {
   checkRendering,
   createDocumentationMessageGenerator,
 } from '../../lib/utils';
+import { addAbsolutePosition } from '../../lib/utils/hits-absolute-position';
```
francoischalifour: If we follow the above pattern, we can export `addAbsolutePosition` in `utils.ts` and import the function from there.

src/helpers/__tests__/insights-test.ts
```diff
@@ -0,0 +1,164 @@
+import insights, {
+  writeDataAttributes,
+  readDataAttributes,
+  hasDataAttributes,
+} from '../insights';
+
+const makeDomElement = (html: string): HTMLElement => {
+  const div = document.createElement('div');
+  div.innerHTML = html;
+  return <HTMLElement>div.firstElementChild || div;
```
francoischalifour:
```suggestion
  return div.firstElementChild as HTMLElement || div;
```

See: https://basarat.gitbooks.io/typescript/docs/types/type-assertion.html#as-foo-vs-foo.

src/helpers/insights.ts
```diff
@@ -0,0 +1,53 @@
...
+export function readDataAttributes(
+  domElement: HTMLElement
+): {
+  method: InsightsClientMethod;
+  payload: InsightsClientPayload;
+} {
+  const method = domElement.getAttribute('data-insights-method');
+  const serializedPayload = domElement.getAttribute('data-insights-payload');
+  if (typeof serializedPayload !== 'string') {
+    throw new Error(`The analytics helper expects payload to be a JSON string`);
```
francoischalifour: Improve message mentioning `data-insights-payload`. There's not much information on how to fix the problem.

src/helpers/insights.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+  try {
+    const payload: InsightsClientPayload = JSON.parse(serializedPayload);
+    return {
+      method: method as InsightsClientMethod,
```
francoischalifour: This cast can be done line 9.

src/helpers/insights.ts
```diff
@@ -0,0 +1,53 @@
...
+  method,
+  payload,
+}: {
+  method: InsightsClientMethod;
+  payload: InsightsClientPayload;
+}): string {
+  if (typeof payload !== 'object') {
+    throw new Error(`The analytics helper expects payload to be an object.`);
+  }
+  let serializedPayload;
```
francoischalifour:
```suggestion
  let serializedPayload: string;
```

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -50,6 +50,29 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
+        indexName: 'indexName',
+        searchClient: { search: () => {} },
+        insightsClient: 'insights',
+      });
+    }).toThrowErrorMatchingInlineSnapshot(
+      `"The provided \`insightsClient\` must be a function."`
+    );
+  });
+
+  it('stores insightsClient on instance', () => {
```
francoischalifour: This is an implementation detail, right?

I don't think we need this test, especially because we don't document this usage.

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,231 @@
...
+    { objectID: '3', __position: 11 },
+    { objectID: '4', __position: 12 },
+  ];
+  const results = {
+    index: 'theIndex',
+    queryID: 'theQueryID',
+    hits,
+  };
+
+  describe('payload inferring', () => {
```
francoischalifour: These tests can be grouped together.

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,231 @@
...
+
+describe('withInsights', () => {
+  let instantSearchInstance: any;
+  let hits: any;
+  let results: any;
+  let widgetParams: any;
+  let widget: any;
+  let renderFn: any;
+
+  beforeEach(() => {
```
francoischalifour: We shouldn't set so many things in the `beforeEach` hook. Tests become hard to follow.

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,231 @@
...
+
+    it('should not expose the insights client wrapper to renderOptions if not passed to instantSearchInstance', () => {
+      delete instantSearchInstance.insightsClient;
+      const [renderProps] = renderFn.mock.calls[0];
+      expect(renderProps).toHaveProperty('insights');
+      expect(renderProps.insights).toBeInstanceOf(Function);
+    });
+  });
+
+  describe('exposed insights client wrapper', () => {
```
francoischalifour: Some tests can be grouped together.

src/lib/__tests__/insights-listener-test.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+import withInsightsListener from '../insights/listener';
+import { mount } from 'enzyme';
+
+describe('withInsightsListener', () => {
+  const setup = Hits => {
```
francoischalifour: `Hits` is not typed at this point.

src/lib/insights/client.ts
```diff
@@ -0,0 +1,102 @@
...
+  objectIDs: string[];
+}): InsightsClientPayload => {
+  const { index, queryID } = results;
+  switch (method) {
+    case 'clickedObjectIDsAfterSearch': {
+      const positions = objectIDs.map(objectID => {
+        const hit = find(hits, hit => hit.objectID === objectID);
+        if (typeof hit === 'undefined') {
+          throw new Error(
+            `Could not find ObjectID ${objectID} in the returned hits.`
```
francoischalifour: Add more context mentioning the method `clickedObjectIDsAfterSearch`.

src/helpers/__tests__/insights-test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+describe('insights', () => {
+  test('default behaviour', () => {
+    expect(
+      insights('clickedObjectIDsAfterSearch', {
+        objectIDs: ['3'],
+        eventName: 'Add to Cart',
+      })
+    ).toEqual(
+      `data-insights-method="clickedObjectIDsAfterSearch" data-insights-payload='{"objectIDs":["3"],"eventName":"Add to Cart"}'`
```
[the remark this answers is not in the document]
tkrugg: I found that the snapshot produced are harder to read due to all the quotes. 
Since we care about having the correct quotes here, I though the inline snapshot was confusing.

```js
   expect(
      insights('clickedObjectIDsAfterSearch', {
        objectIDs: ['3'],
        eventName: 'Add to Cart',
      })
    ).toMatchInlineSnapshot(
      `"data-insights-method=\\"clickedObjectIDsAfterSearch\\" data-insights-payload='{\\"objectIDs\\":[\\"3\\"],\\"eventName\\":\\"Add to Cart\\"}'"`
    );
```

src/helpers/insights.ts
```diff
@@ -0,0 +1,55 @@
...
+  if (typeof payload !== 'object') {
+    throw new Error(`The analytics helper expects payload to be an object.`);
+  }
+  try {
+    serializedPayload = JSON.stringify(payload);
+  } catch (e) {
+    throw new Error(`Could not JSON serialize the payload object.`);
+  }
+  return `data-insights-method="${method}" data-insights-payload='${serializedPayload}'`;
```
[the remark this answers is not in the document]
tkrugg: I just realized my comment on this was never posted

the unsafe case is when there is a single quote in eventNames or indexName.
base64 is straight forward but IMHO it's best to keep the string readable. 
What we can do is escape the single quotes. Any thoughts on this.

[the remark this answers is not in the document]
francoischalifour: A single quote in an `eventName` can happen since there's no restrictions in the naming.

src/types/insights.ts
tkrugg: As discussed offline I ended giving up on overloading for simplicity.

src/lib/insights/client.ts
tkrugg: It looks pretty common here:
https://github.com/algolia/instantsearch.js/blob/9aa5d9a9fde26f6465276749344922f868da566c/src/lib/InstantSearch.js#L289

src/lib/__tests__/InstantSearch-test.js
tkrugg: connectInsights expects it to be here, so it's part of the contract this module is making with other modules of the code. 
if you change something that break this contract, you need to be able to detect it. This test is here for that.
I still think it makes sense to keep it.

src/lib/__tests__/insights-client-test.ts
tkrugg: the alternative is to move what's inside the beforeEach inside the tests themselves.
It's redundant and makes the tests bigger. IMHO it's not helping the readability.

src/lib/__tests__/insights-client-test.ts
tkrugg: could you tell which?
IMHO it's best that each tests is no longer than a couple of lines (or statements).

src/lib/__tests__/insights-client-test.ts
francoischalifour: It does IMO and @samouss agreed in another PR comment.

You should be able to read a test just like you read a book. Seeing [this example](https://github.com/algolia/instantsearch.js/blob/7b0595e0a75a7efb29ee2657215f3cefc982b262/src/lib/__tests__/insights-client-test.ts#L70), I have strictly no idea what the context of the test is without going back to `beforeEach`. Besides, it makes the PR hard to review.

These discussions shouldn't even happen in the first place if we agree beforehand on what and how to test our libraries. That's why we need guidelines about this.

src/lib/__tests__/InstantSearch-test.js
francoischalifour: Right, if other pieces of the codebase depend on it, it makes sense to test it.

src/lib/insights/client.ts
tkrugg: ok, replaced with: 
> `Could not find ObjectID ${objectID} passed to 'clickedObjectIDsAfterSearch' in the returned hits. This is necessary to infer absolute position.`

src/lib/__tests__/insights-listener-test.tsx
tkrugg: Hits is just a mock for functional component
As you know they are rather hard to type and we keep limited typing in tests.

src/lib/__tests__/insights-client-test.ts
tkrugg: ok let's talk about this Wednesday. I'll setup a meeting together with @samouss before the standup, feel free to reschedule.

src/helpers/insights.ts
tkrugg: nice catch, thanks!

src/helpers/__tests__/insights-test.ts
tkrugg: thanks!

src/helpers/insights.ts
tkrugg: ok

src/helpers/insights.ts
tkrugg: changed to :
> The insights helper expects data-insights-payload to be a JSON string

src/lib/insights/listener.tsx
tkrugg: I think we can keep both.
If you look at other codebases using typescript (including [react-instantsearch](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-dom/src/components/Hits.tsx#L2)), that's what they seem to do. https://github.com/apollographql/react-apollo/blob/master/src/ApolloConsumer.tsx#L2

src/lib/insights/listener.tsx
tkrugg: can you suggest a type?

tkrugg: used React.ComponentType<Props>

src/helpers/insights.ts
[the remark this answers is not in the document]
tkrugg: ok, encoding to base64

src/lib/__tests__/insights-listener-test.tsx
tkrugg: changed to `React.ComponentType<any>`

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,261 @@
...
+  const results = {
+    index: 'theIndex',
+    hits,
+  };
+  const widgetParams = { transformItems: jest.fn() };
+
+  const renderFn = jest.fn();
+  const unmountFn = jest.fn();
+
+  const connectHitWithInsights = withInsights(connectHits as any);
```
francoischalifour:
```suggestion
  const connectHitsWithInsights = withInsights(connectHits as any);
```

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,261 @@
...
+  const widget = connectHitWithInsights(renderFn, unmountFn)(widgetParams);
```
francoischalifour:
```suggestion
  const widget = connectHitsWithInsights(renderFn, unmountFn)(widgetParams);
```

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,261 @@
...
+  render({ results, instantSearchInstance }) {
+    const hits = results.hits;
+    renderFn({ hits, results, instantSearchInstance, widgetParams }, false);
+  },
+  dispose() {
+    unmountFn();
+  },
+});
+
+const setup = ({ hits: providedHits, insightsClient }: any = {}) => {
```
francoischalifour: I'm not convinced by this setup. There's a lot of logic and unnecessary abstraction that we should avoid in the tests.

tkrugg: Please explain the solution you have in mind.

src/helpers/insights.ts
```diff
@@ -0,0 +1,59 @@
...
+  domElement: HTMLElement
+): {
+  method: InsightsClientMethod;
+  payload: Partial<InsightsClientPayload>;
+} {
+  const method = domElement.getAttribute(
+    'data-insights-method'
+  ) as InsightsClientMethod;
+  const serializedPayload = domElement.getAttribute('data-insights-payload');
+  if (typeof serializedPayload !== 'string') {
```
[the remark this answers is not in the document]
tkrugg: yes, and typeof null === 'object'.
This check is there to make sure what we give to JSON.parse is a string.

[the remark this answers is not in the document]
francoischalifour: I'm not sure this is doing what you want because `JSON.parse('')` throws an error. We might be better only relying on the try/catch below.

src/lib/__tests__/insights-client-test.ts
francoischalifour: I meant inlining this in the tests themself.

src/helpers/insights.ts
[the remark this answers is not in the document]
tkrugg: are you're suggesting to remove that check altogether?

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,261 @@
...
+      const [renderProps] = renderFn.mock.calls[0];
+      expect(renderProps).toHaveProperty('insights');
+      expect(renderProps.insights).toBeInstanceOf(Function);
+    });
+
+    it('should not expose the insights client wrapper to renderOptions if not passed to instantSearchInstance', () => {
+      const { renderFn } = setup({ insightsClient: undefined });
+      const [renderProps] = renderFn.mock.calls[0];
+      expect(renderProps).toHaveProperty('insights');
+      expect(renderProps.insights).toBeInstanceOf(Function);
```
samouss: This test seems to be incorrect. The `insights` wrapper should **not** be provided, correct? The setup function fallback on `jest.fn()`.

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,261 @@
...
+    });
+  });
+
+  describe('exposed insights client wrapper', () => {
+    beforeEach(() => {});
```
samouss: We can drop it since it's useless.

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    it('should call the insights client under the hood', () => {
+      const { renderFn, instantSearchInstance } = setup();
+      const [renderProps] = renderFn.mock.calls[0];
+      renderProps.insights('clickedObjectIDsAfterSearch', {
+        objectIDs: ['3'],
+        eventName: 'Add to basket',
+      });
+      expect(instantSearchInstance.insightsClient).toHaveBeenCalledTimes(1);
```
samouss: We could avoid to rely on this implementation detail. We can create/provide the `insightsClient` directly to the setup function.

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,261 @@
...
+      const [
+        method,
+        payload,
+      ] = instantSearchInstance.insightsClient.mock.calls[0];
+      expect(method).toEqual('convertedObjectIDsAfterSearch');
+      expect(payload.eventName).toEqual('Add to basket');
+      expect(payload.index).toEqual('theIndex');
+      expect(payload.queryID).toEqual('theQueryID');
+      expect(payload.objectIDs).toEqual(['1']);
+      expect(payload.positions).toBeUndefined();
```
samouss: We could write it with one `toEqual` assertions, it avoid to create assertions for the negation.

```
expect(payload).toEqual({
  eventName: "Add to basket",
  index: "theIndex",
  queryID: "theQueryID",
  objecIDs: ["1"]
});
```

src/lib/insights/client.ts
```diff
@@ -0,0 +1,131 @@
...
+      throw new Error(
+        `Could not find ObjectID ${objectID} passed to 'clickedObjectIDsAfterSearch' in the returned hits. This is necessary to infer absolute position and queryID.`
+      );
+    }
+    return hit;
+  });
+};
+
+const getQueryID = (selectedHits: Hits<unknown>) => {
+  const queryIDs = uniq(selectedHits.map(hit => hit.__queryID));
```
samouss: It's not the first time we use `uniq` but for a simple case like this one `indexOf` work fine.

src/lib/insights/client.ts
```diff
@@ -0,0 +1,131 @@
...
+} from '../../types';
+import find from 'lodash/find';
+import uniq from 'lodash/uniq';
+
+const getSelectedHits = (hits: Hits<unknown>, selectedObjectIDs: string[]) => {
+  return selectedObjectIDs.map(objectID => {
+    const hit = find(hits, h => h.objectID === objectID);
+    if (typeof hit === 'undefined') {
+      throw new Error(
+        `Could not find ObjectID ${objectID} passed to 'clickedObjectIDsAfterSearch' in the returned hits. This is necessary to infer absolute position and queryID.`
```
samouss:
```suggestion
        `Could not find objectID "${objectID}" passed to 'clickedObjectIDsAfterSearch' in the returned hits. This is necessary to infer absolute position and queryID.`
```

src/types/connector.ts
```diff
@@ -1,11 +1,17 @@
-import { InstantSearch } from './instantsearch';
+import { Hits, InstantSearch, SearchResults } from './instantsearch';
+import { InsightsClient } from './insights';
 
-export type RenderOptions<T = unknown> = {
-  widgetParams: T;
+export type RenderOptions<WidgetParams = unknown> = {
```
samouss:
```suggestion
export type RenderOptions<TWidgetParams = unknown> = {
```

src/lib/__tests__/insights-client-test.ts
tkrugg: nice catch, thanks! 💯

src/types/connector.ts
tkrugg: fixed in 92690df0949a76517c46edf4d5da036bd2538a02

src/lib/__tests__/insights-client-test.ts
tkrugg: fixed in 1de2de18f24bee9945916d3dc1ea9652eeb17773

src/lib/__tests__/insights-client-test.ts
tkrugg: fixed in ce61df08bf0278a0c3cbca46c9ad3e17cb490c87

src/lib/__tests__/insights-client-test.ts
tkrugg: fixed in de807bc469293b24920d9d942b55a51a7d99e854

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.js
```diff
@@ -360,6 +364,53 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
+      { hits, queryID: 'theQueryID' },
+    ]);
+    widget.render({
+      results,
+      state: helper.state,
+      helper,
+      createURL: () => '#',
+    });
+
+    const secondRenderingOptions = rendering.mock.calls[1][0];
```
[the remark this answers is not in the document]
tkrugg: fixed in 8dff0a44694ce7b1b5a63ef11bbfa7a5ef4dc2fb

stories/hits.stories.js
tkrugg: fixed here: 0abc39ee4ba7ba9af9443e1eb032cb3cb577012a

src/helpers/insights.ts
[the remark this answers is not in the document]
tkrugg: done here 6b15b1c04beb2741556af332a0b338c80a67b313

src/lib/insights/listener.tsx
francoischalifour: These projects are for React consumers. In InstantSearch.js, this is only internal. This is therefore redundant.

src/lib/insights/client.ts
francoischalifour: Can you explain this check before resolving the conversation?

src/helpers/insights.ts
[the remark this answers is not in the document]
francoischalifour: What gives me the feeling that this check is almost inappropriate is that the error message is (or should be) exactly the same as the try/catch below.

From what I understand this is checking that the payload is not a `boolean` for instance, even though it is JSON parsable.

What we could do is move this check in the try/catch and throw an empty error. Then the top-level catch will throw the correct message error for both.

WDYT?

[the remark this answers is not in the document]
tkrugg: > From what I understand this is checking that the payload is not a boolean for instance, even though it is JSON parsable.
It can only be a string or null, not a boolean.

This checks data-insights-payload is there. It could be `!== null` but since 
 JSON.parse expects a string and reenforced that check. That's all.

Can be rephrased to: `The insights helper expects data-insights-payload attribute` but it's more helpful with hint on the type expected.

If we put everything in the try catch, we can't throw distinct messages for "no attribute found" and "invalid attribute"

src/lib/__tests__/insights-client-test.ts
tkrugg: I think that would be way too verbose. That's 20 lines to copy paste in every tests.

src/lib/insights/client.ts
tkrugg: it's removed here 20545cd

src/lib/insights/listener.tsx
tkrugg: I still think they are helpful as they run at a different time and catch property type mutations for example. 
I've removed it here: b8e063bd

src/lib/insights/listener.tsx
tkrugg: remove here: b8e063bd

stories/infinite-hits.stories.js
tkrugg: this has been resolved here 0abc39ee4ba7ba9af9443e1eb032cb3cb577012a

src/lib/__tests__/insights-client-test.ts
francoischalifour: Verbose sometimes means easier to reason about.

Here it creates so much abstraction with instructions that have very different purposes. Abstraction in tests is also error prone because the abstraction is not tested itself.

You create too much boilerplate that is not needed in many tests. Besides, some (future) tests might not behave correctly because of:

```js
const widgetParams = { transformItems: jest.fn() };
```

The tests are really hard to understand in isolation. There are too much side effects.

Honestly, who cares about having verbose tests? That's something we already have a lot in the codebase anyway.

src/helpers/insights.ts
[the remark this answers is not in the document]
francoischalifour: The message in the try/catch covers both. A missing attribute is in itself an invalid attribute value. I think it's fine removing a code path for this.

We can update the error message in the try/catch to give more information:

```
Unable to parse `data-insights-payload`. This data attribute expects a JSON string.
```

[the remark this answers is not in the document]
tkrugg: reading your comment I think you're looking at the wrong diff.
did you see there `atob`  called now that we serialize to base64?
It's even more important not to have `null` for 2 reasons:
- typescript will raise "null is not assignable to type string"
- atob on chrome will serialize `null`  to `"null"`

src/lib/InstantSearch.js
```diff
@@ -80,7 +81,12 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       searchClient.addAlgoliaAgent(`instantsearch.js (${version})`);
     }
 
+    if (insightsClient && typeof insightsClient !== 'function') {
+      throw new Error(`The provided \`insightsClient\` must be a function.`);
```
francoischalifour:
```suggestion
      throw new Error('The provided `insightsClient` must be a function.');
```

src/lib/insights/client.ts
```diff
@@ -0,0 +1,128 @@
...
+} from '../../types';
+import find from 'lodash/find';
+import uniq from 'lodash/uniq';
+
+const getSelectedHits = (hits: Hits<unknown>, selectedObjectIDs: string[]) => {
+  return selectedObjectIDs.map(objectID => {
+    const hit = find(hits, h => h.objectID === objectID);
+    if (typeof hit === 'undefined') {
+      throw new Error(
+        `Could not find objectID "${objectID}" passed to 'clickedObjectIDsAfterSearch' in the returned hits. This is necessary to infer absolute position and queryID.`
```
francoischalifour:
```suggestion
        `Could not find objectID "${objectID}" passed to \`clickedObjectIDsAfterSearch\` in the returned hits. This is necessary to infer the absolute position and the query ID.`
```

src/lib/insights/client.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+    return hit;
+  });
+};
+
+const getQueryID = (selectedHits: Hits<unknown>) => {
+  const queryIDs = uniq(selectedHits.map(hit => hit.__queryID));
+  if (queryIDs.length > 1) {
+    throw new Error(
+      `Current API allows a single queryID. The objectIDs provided map to multiple queryIDs.`
```
francoischalifour:
```suggestion
      'Insights currently allows a single query ID in `queryIDs`. The `objectIDs` provided map to multiple `queryIDs`.'
```

src/lib/insights/client.ts
```diff
@@ -0,0 +1,128 @@
...
+    );
+  }
+  const queryID = queryIDs[0];
+  if (typeof queryID !== 'string') {
+    throw new Error(
+      `Could not infer queryID. Ensure InstantSearch is configured with clickAnalytics: true `
```
francoischalifour:
```suggestion
      'Could not infer query ID. Ensure `instantsearch` is configured with { clickAnalytics: true }'
```

src/lib/insights/client.ts
```diff
@@ -0,0 +1,128 @@
...
+    case 'clickedObjectIDsAfterSearch': {
+      const positions = getPositions(selectedHits);
+      return { index, queryID, objectIDs, positions };
+    }
+
+    case 'convertedObjectIDsAfterSearch':
+      return { index, queryID, objectIDs };
+
+    default:
+      throw new Error(`Unsupported method passed ${method}`);
```
francoischalifour:
```suggestion
      throw new Error(`Unsupported method passed to Insights: "${method}".`);
```

src/lib/insights/listener.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export default (BaseComponent: React.ComponentType<any>) => {
+  function WithInsightsListener(props: WithInsightsListenerProps) {
+    const handleClick = (event: React.MouseEvent<HTMLElement>) => {
+      if (!hasDataAttributes(event.target)) {
+        return;
+      }
+      if (!props.insights) {
+        throw new Error(
+          `insightsClient was not been provided to InstantSearch instance.`
```
francoischalifour:
```suggestion
          'The `insightsClient` option was not been provided to `instantsearch`.'
```

src/lib/insights/client.ts
```diff
@@ -0,0 +1,128 @@
...
+const wrapInsightsClient = (
+  aa: InsightsClient,
+  results: RenderOptions['results'],
+  hits: Hits<unknown>
+): InsightsClientWrapper => (
+  method: InsightsClientMethod,
+  payload: Partial<InsightsClientPayload>
+) => {
+  if (!Array.isArray(payload.objectIDs)) {
+    throw new TypeError(`Expected objectIDs to be an Array`);
```
francoischalifour:
```suggestion
    throw new TypeError('Expected `objectIDs` to be an array.');
```

stories/infinite-hits.stories.js
```diff
@@ -68,4 +70,37 @@ storiesOf('InfiniteHits', module)
...
+                eventName: 'Add to cart',
+              })} >Add to cart</button>
+            `,
+            },
+          })
+        );
+      },
+      {
+        insightsClient: (method, payload) =>
+          action(`[InsightsClient] sent ${method} with payload`)(payload),
```
francoischalifour:
```suggestion
          action(`[InsightsClient] sent "${method}" with payload`)(payload),
```

src/lib/__tests__/insights-client-test.ts
```diff
@@ -0,0 +1,341 @@
...
+      createWidgetWithInsights({ renderFn, instantSearchInstance, results });
+
+      const [renderProps] = renderFn.mock.calls[0];
+      expect(() => {
+        renderProps.insights('clickedObjectIDsAfterSearch', {
+          objectIDs: ['1', '2'],
+          eventName: 'Add to basket',
+        });
+      }).toThrowErrorMatchingInlineSnapshot(
+        `"Could not infer queryID. Ensure InstantSearch is configured with clickAnalytics: true "`
```
tkrugg:
```suggestion
        '"Could not infer `queryID`. Ensure `instantsearch` is configured with `clickAnalytics: true`"'
```

src/lib/insights/client.ts
tkrugg: I skip this one. queryID is called queryID everywhere so it's best to keep it as it.
I kept the rest. see https://github.com/algolia/instantsearch.js/pull/3598/files/3aa8b4e978530c1784a3be8c7193d34a8a8dc1ed#r273981507

tkrugg:
@algolia/instantsearch-for-websites  will squash this to ease the conflict solving. It shouldn't matter because we squash before merging anyway.
For furture reference, HEAD before squash is at 3d134112a8794150be81728fa1cd26feb113a93c

src/types/utils.ts
```diff
@@ -1 +1,2 @@
 export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+export type Without<T, K> = Pick<T, Exclude<keyof T, K>>;
```
francoischalifour: This is the same as `Omit` AFAICT.

tkrugg: and actually a better one :) thanks

tkrugg: fixed here: c8796b3

src/lib/insights/client.ts
```diff
@@ -0,0 +1,130 @@
+import {
```
francoischalifour: Internal imports should be after external imports btw. Maybe we can enforce that.

tkrugg: I'm not very picky about it but there is a rule for it in eslint https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md

PR #3599, opened 2019-03-12
chore(storybook): add support for playgrounds

francoischalifour:
This adds supports for different playgrounds in the Storybook so that we can use other datasets than `instant_search`.

It's necessary for writing stories about Query Rules.

## Files updates

- Renamed
  - `.storybook/decorators.js` → `.storybook/decorators.ts`
- Edited
  - `.storybook/decorators.ts`: added support for playgrounds
- Added
  - `.storybook/playgrounds/movies.ts`

## Usage

```js
import moviesPlayground from '../.storybook/playgrounds/movies';

storiesOf('MyStory', module).add(
  'default',
  withHits(
    ({ search, container, instantsearch }) => {
      // ...
    },
    {
      appId: '',
      apiKey: '',
      indexName: '',
      playground: moviesPlayground, // this is taken into account
    }
  )
);

```

PR #3600, opened 2019-03-13
feat(queryRules): add queryRuleCustomData widget

francoischalifour:
_This is the 2nd PR in the series "Merchandized Query Rules"._

## Summary

This creates the `queryRuleCustomData` widget, responsible for showing custom data (`userData`) coming from Query Rules.

## Current solution

Right now, we can leverage the [`connectHits`](https://www.algolia.com/doc/api-reference/widgets/hits/js/#connector) connector to show a banner with InstantSearch.js.

```js
const promotedResults = instantsearch.connectors.connectHits(
  (options, isFirstRender) => {
    const { widgetParams, results = {} } = options;
    const { container } = widgetParams;

    if (!results.userData) {
      container.innerHTML = '';
      return;
    }

    const [data] = results.userData;

    container.innerHTML = `<img src="${data.banner}" alt="${data.link}">`;
  }
);

search.addWidget(
  promotedResults({
    container: document.querySelector('#promoted-results'),
  })
);
```

[See live example →](https://codesandbox.io/s/ywy2q74rzv)

The goal of the widget is to simplify this solution.

## Widget usage

Given this `userData`:

```json
{
  "banner": "https://website.com/image.png",
  "title": "This is a revolution.",
  "link": "https://website.com/"
}
```

When the query rule is matched, the engine returns:

```json
[
  {
    "banner": "https://website.com/image.png",
    "title": "This is a revolution.",
    "link": "https://website.com/"
  }
]
```

You would use the `queryRuleCustomData` widget:

```js
instantsearch.widgets.queryRuleCustomData({
  container: widgetContainer,
  transformItems: items => items[0],
  templates: {
    default({ title, banner, link }) {
      if (!banner) {
        return '';
      }

      return `
        <h2>${title}</h2>

        <a href="${link}">
          <img src="${banner}" alt="${title}">
        </a>
      `;
    },
  },
});
```

### `transformItems`

> `object[] => any` – defaults to `items => items`

The function to change the data sent by the engine before giving it to the template.

### `container`

> `string` | `HTMLElement` – no default

The container of the widget.

### `templates`

> `object` – defaults to `{ default: '' }`

The templates to display for the custom data (`userData`).

## Stories

[See stories →](https://deploy-preview-3600--instantsearchjs.netlify.com/stories/?path=/story/queryrulecustomdata--default)

## Related

- [`connectQueryRules`](https://github.com/algolia/instantsearch.js/pull/3597)
- [`queryRuleContext`](https://github.com/algolia/instantsearch.js/pull/3602)


francoischalifour:
GitHub seems to also display the changed introduced in #3599, that's weird.

Haroenv:
These changes seem to be a different PR? Not sure what to review here

Haroenv:
I think the `base` might be wrong :)

francoischalifour:
I said in https://github.com/algolia/instantsearch.js/pull/3600#issuecomment-472837790 I think that's a GitHub bug. You can ignore the changes related to Storybook.

src/components/QueryRuleCustomData/QueryRuleCustomData.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'preact-compat';
+import { QueryRuleCustomDataTemplates } from '../../widgets/query-rule-custom-data/query-rule-custom-data';
+import Template from '../Template/Template';
+
+type QueryRuleCustomDataProps = {
+  templateProps: {
+    templates: QueryRuleCustomDataTemplates;
+  };
+  items: object[];
```
samouss: Would be nice to have generics here, but we can improve that later.

src/lib/utils/getContainerNode.ts
```diff
@@ -23,7 +23,7 @@ function getContainerNode(selectorOrHTMLElement: string | HTMLElement) {
     throw new Error(errorMessage);
   }
 
-  return domElement;
+  return domElement as HTMLElement;
```
samouss: You can avoid the cast with `querySelector<HTMLElement>` inside the function.

src/widgets/query-rule-custom-data/__tests__/query-rule-custom-data-test.ts
```diff
@@ -0,0 +1,120 @@
...
+        // @ts-ignore
+        queryRuleCustomData({});
+      }).toThrowErrorMatchingInlineSnapshot(`
+"The \`container\` option is required.
+
+See documentation: https://www.algolia.com/doc/api-reference/widgets/query-rule-custom-data/js/"
+`);
+    });
+
+    test('throws without container', () => {
```
samouss: We test two times that options without `container` throws an error.

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
```diff
@@ -0,0 +1,79 @@
...
+const withUsage = createDocumentationMessageGenerator({
+  name: 'query-rule-custom-data',
+});
+
+const renderer: QueryRulesRenderer<
+  QueryRuleCustomDataConnectorWidgetParams
+> = ({ userData, instantSearchInstance, widgetParams }) => {
+  const { container, templates } = widgetParams;
+
+  const templateProps = prepareTemplateProps({
```
samouss: Do we really have to trigger this on each render?

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
```diff
@@ -0,0 +1,79 @@
...
+  default: string | ((items: object[]) => string);
+};
+
+type QueryRuleCustomDataWidgetParams = {
+  container: string | HTMLElement;
+  templates?: QueryRuleCustomDataTemplates;
+  transformItems?: (items: object[]) => any;
+};
+
+interface QueryRuleCustomDataConnectorWidgetParams
```
samouss: Why this type is useful? We can directly provide `QueryRuleCustomDataWidgetParams` to the renderer.

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
```diff
@@ -0,0 +1,79 @@
...
+    container
+  );
+};
+
+const queryRuleCustomData: QueryRuleCustomData = (
+  {
+    container,
+    templates: userTemplates = {},
+    transformItems = items => items,
+  } = {} as QueryRuleCustomDataWidgetParams
```
samouss: We have to investigate why we have to cast here, but definitely not a blocker.

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
```diff
@@ -0,0 +1,79 @@
...
+) => {
+  if (!container) {
+    throw new Error(withUsage('The `container` option is required.'));
+  }
+
+  const defaultTemplates = { default: '' };
+  const templates: QueryRuleCustomDataTemplates = {
```
samouss: Actually I think the `prepareTemplateProps` already do [the merge of the `templates`](https://github.com/algolia/instantsearch.js/blob/60a965df3443d14759657e282172cf3440621cdd/src/lib/utils/prepareTemplateProps.ts#L16-L18).

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
```diff
@@ -0,0 +1,79 @@
...
+    ...defaultTemplates,
+    ...userTemplates,
+  };
+
+  const containerNode = getContainerNode(container);
+  const makeContextualHits = connectQueryRules(renderer, () => {
```
samouss: The name does not seem to match the widget.

stories/query-rule-custom-data.stories.ts
```diff
@@ -0,0 +1,95 @@
...
+
+      container.appendChild(description);
+      container.appendChild(widgetContainer);
+
+      search.addWidget(
+        instantsearch.widgets.queryRuleCustomData({
+          container: widgetContainer,
+          transformItems: (items: CustomDataItem[]) => items[0],
+          templates: {
+            default({ title, banner, link }: CustomDataItem) {
```
samouss: Actually we have no guarantee of the type safety here. The widget should accept a generic type that would be pass to `transformItems`. Then the data provided to the template could be inferred from the return of type for `transformItems`. I'm saying that it's easy though, we can improve it later.

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
francoischalifour: Nope, because at the renderer level `container` _must_ be an `HTMLElement`, and not `string | HTMLElement`.

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
francoischalifour: In this case it's because `container` is required in `QueryRuleCustomDataWidgetParams`.

Since we want to keep the pattern of `const widget = (options = {})` and `container` doesn't exist in `{}`, it's necessary to cast `{}` to `QueryRuleCustomDataWidgetParams`. Does that make sense?

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
francoischalifour: Indeed, I'll simplify this. I wonder if we still need `prepareTemplateProps` actually. Thoughts?

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
francoischalifour: Good catch – that's a remaining of my POC.

stories/query-rule-custom-data.stories.ts
francoischalifour: Exactly – that's a discussion that we've had already.

src/lib/utils/getContainerNode.ts
francoischalifour: Then the signature becomes:

```js
getContainerNode(selectorOrHTMLElement: string | HTMLElement): HTMLElement | null
```

While we don't want (and it can't) be `null`.

samouss: With a proper typing of `isDOMElement` we should not have the issue, we can ensure that at this stage the selector is a DOM element.

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
samouss: Yes indeed, a different name would have helped to understand IMO. The `container` has nothing to do with the connector it's only used for the rendering. It's confusing to see it in the name.

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
samouss: Yes.

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
samouss: We can keep it for now and see what is the impact of change. It does add some information on the object like `useCustomCompileOptions`. Not sure why it's useful though.

francoischalifour: I removed `prepareTemplateProps` in the last change.

> Not sure why it's useful though.

Exactly my thought too.

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
francoischalifour: I renamed it `QueryRuleCustomDataRendererWidgetParams`.

src/lib/utils/getContainerNode.ts
Haroenv: Yep, looks like `isDomElement` should be a type guard: https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types

(and the post by @wittydeveloper [medium](https://medium.com/@wittydeveloper/typescript-make-types-real-the-type-guard-functions-814364e8dbe3))

francoischalifour: Indeed, that's what we're looking for!

francoischalifour:
Without custom CSS classes, the inserted `div` container is not easily customizable because nothing identifies it.

I think we should create an [InstantSearch.css](https://github.com/algolia/instantsearch-specs) spec for this widget with the following markup:

```html
<div class="ais-QueryRuleCustomData"></div>
```

It would imply adding a `cssClass` option to the widget, like in most widgets.

WDYT?

Haroenv:
makes sense to me!

src/lib/utils/isDomElement.ts
```diff
@@ -1,5 +1,6 @@
-function isDomElement(object) {
+function isDomElement(object: any): object is HTMLElement {
   return (
+    // @ts-ignore
     object instanceof window.HTMLElement ||
```
samouss: I don't think the `window` is really useful in that case. We can simplify the condition and remove the `@ts-ignore`. Note that the rest of the condition does not entirely satisfy the type constraint. `nodeType` greater than 0 can represent `Element`, `Text`, `Comment`, ... Which is incompatible with our type but anyway it's working.

francoischalifour: Indeed – I didn't want to change this condition in a minor because some users probably rely on this check.

samouss: I think React/Preact throw an error if the container is not an Element.

francoischalifour: I checked online the other day and apparently the second check is for browsers not supporting `window.HTMLElement`. We can still change these checks on another iteration.

PR #3601, opened 2019-03-13
feat(voiceSearch): add connector and widget

eunjae-lee:
This POC shows an implementation of voice search widget on instantsearch.js.
See more: https://github.com/algolia/instantsearch-rfcs/pull/15

This PR is a draft, not intended to be merged right away.


src/lib/voiceSearchHelper/index.js
```diff
@@ -0,0 +1,83 @@
+const STATUS_INITIAL = 'initial';
+const STATUS_ASKING_PERMISSION = 'askingPermission';
+const STATUS_WAITING_FOR_AUDIO = 'waitingForAudio';
+const STATUS_WAITING_FOR_VOICE = 'waitingForVoice';
+const STATUS_RECOGNIZING = 'recognizing';
+const STATUS_FINISHED = 'finished';
+
+export default function voiceSearchHelper({ onQueryChange, onStateChange }) {
+  window.SpeechRecognition =
```
tkrugg: So we actually need to write to window? I think it be just `const SpeechRecognition = ...` and there is no footprint on the global scope

src/connectors/voice-search/connectVoiceSearch.js
```diff
@@ -0,0 +1,87 @@
...
+            voiceListeningState: this._voiceListeningState,
+            widgetParams,
+          },
+          true
+        );
+      },
+      render({ helper }) {
+        this._rerender = () => {
+          this.render({ helper });
+        };
```
tkrugg: This is rather tricky 😅 I understand why need it but I wonder if there is anyway to avoid it.
Did you find similar patterns in part of the code base that deal with the same kind of re-rendering issue?

src/components/VoiceSearch/VoiceSearch.js
```diff
@@ -0,0 +1,111 @@
...
+    } = this.props;
+    const isListening = this.props.isListening();
+    const {
+      status,
+      transcript,
+      isSpeechFinal,
+      errorCode,
+    } = voiceListeningState;
+    return (
+      isSupportedBrowser() && (
```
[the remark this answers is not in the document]
Haroenv: Yes, this could be unexpected for UI, I think making it the same size etc, but visually hidden could make sense too

src/components/VoiceSearch/VoiceSearch.js
```diff
@@ -0,0 +1,111 @@
...
+    searchAsYouSpeak: true,
+    query: '',
+  };
+
+  constructor(props) {
+    super(props);
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick() {
```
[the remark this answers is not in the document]
Haroenv: `handleClick = () => {`

src/connectors/voice-search/connectVoiceSearch.js
Haroenv: This is the same pattern as used for showMore

src/lib/voiceSearchHelper/index.js
eunjae-lee: @tkrugg Updated.

src/components/VoiceSearch/VoiceSearch.js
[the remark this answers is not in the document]
eunjae-lee: Good point. I think doing this two makes sense:

- Disable the button if it's not on supported browser.
- Expose `hideOnUnsupportedBrowsers` with a default value `false`.

What should happen if our user didn't put `hideOnUnsupportedBrowsers={true}` and end-user tries on unsupported browser? For now, they will see a disabled button. Should we put a tooltip telling why it's disabled?

src/components/VoiceSearch/VoiceSearch.js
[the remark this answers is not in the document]
eunjae-lee: 👍

src/components/VoiceSearch/VoiceSearch.js
```diff
@@ -0,0 +1,111 @@
...
+    this.button.blur();
```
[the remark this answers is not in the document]
eunjae-lee: 👍

src/connectors/voice-search/connectVoiceSearch.js
```diff
@@ -0,0 +1,87 @@
...
+    return {
+      init({ helper }) {
+        this._refine = (() => {
+          let previousQuery;
+          const setQueryAndSearch = query => {
+            if (query !== helper.state.query) {
+              previousQuery = helper.state.query;
+              helper.setQuery(query);
+            }
+            if (previousQuery !== undefined && previousQuery !== query) {
```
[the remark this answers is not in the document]
eunjae-lee: @tkrugg could you give me more explanation on this or a link to a good post?

src/components/VoiceSearch/VoiceSearch.js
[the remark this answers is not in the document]
Haroenv: Simply a disabled button in our case should be fine I think, since users can still customise that with the template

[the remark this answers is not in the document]
eunjae-lee: `isSupportedBrowser` is not exposed to templates. For users to customise, should it be exposed?

src/connectors/voice-search/connectVoiceSearch.js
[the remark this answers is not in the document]
eunjae-lee: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#Typeof_operator_and_undefined

https://stackoverflow.com/a/2703122

This?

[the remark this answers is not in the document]
francoischalifour: He meant:

```js
if (typeof previousQuery !== 'undefined' && previousQuery !== query) {
  // ...
}
```

src/components/VoiceSearch/VoiceSearch.js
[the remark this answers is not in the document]
francoischalifour: We could actually have a specific template for when the browser is not supported.

src/connectors/voice-search/connectVoiceSearch.js
samouss: We use en external function for the `render` in the [other](https://github.com/algolia/instantsearch.js/blob/develop/src/connectors/refinement-list/connectRefinementList.js#L359-L371) implementations. Here we heavily rely on `this` which doesn't help to follow how the code is actually executed.

src/connectors/voice-search/connectVoiceSearch.js
```diff
@@ -0,0 +1,87 @@
...
+
+          return setQueryAndSearch;
+        })();
+
+        this._rerender = () => {};
+
+        this._voiceSearchHelper = voiceSearchHelper({
+          onQueryChange: query => this._refine(query),
+          onStateChange: state => {
+            this._voiceListeningState = state;
```
samouss: We could expose a `getState` on the `voiceSearchHelper` that would avoid to store the value on the instance.

src/connectors/voice-search/connectVoiceSearch.js
eunjae-lee: That looks clearer 👍

Haroenv:
The lint job indicates you used `Function` as a type somewhere in TS, and suggests to avoid it by a real annotation: `() => {}` (with types filled in)

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,104 @@
...
+const STATUS_RECOGNIZING = 'recognizing';
+const STATUS_FINISHED = 'finished';
+const STATUS_ERROR = 'error';
+
+type VoiceSearchHelperParams = {
+  onQueryChange: (query: string) => any;
+  onStateChange: () => any;
+};
+
+export default function voiceSearchHelper({
```
Haroenv: I know this one is hard to test, but it probably should be done with a mocked `window. SpeechRecognition`, and different ways it can behave

Haroenv: plus some test cases with `webkitSpeechRecogntion` and `undefined`

stories/voice-search.stories.js
```diff
@@ -146,13 +146,12 @@ storiesOf('VoiceSearch', module)
           left: 0;
           right: 0;
           z-index: 2;
-          display: flex;
```
BenoitPerrot: Squash this in the previous commit, to simplify the history

stories/voice-search.stories.js
```diff
@@ -146,13 +146,12 @@ storiesOf('VoiceSearch', module)
...
           align-items: center;
           justify-content: center;
           display: none;
         }`,
         `.voice-search-transcript .layer.started-true {
-          display: block;
```
BenoitPerrot: Squash this in the previous commit, to simplify the history

src/components/VoiceSearch/VoiceSearch.js
```diff
@@ -0,0 +1,187 @@
...
+      this.state.status === STATUS_FINISHED
+    );
+  }
+
+  startRecognition() {
+    this.setState({
+      status: STATUS_ASKING_PERMISSION,
+    });
+    this.recognition = new window.SpeechRecognition();
+    // this.recognition.continuous = true;
```
BenoitPerrot: Prefer not committing code-in-comment

src/components/VoiceSearch/VoiceSearch.js
```diff
@@ -0,0 +1,187 @@
...
+        !this.props.searchAsYouSpeak
+      ) {
+        this.props.refine(this.state.transcript);
+      }
+      this.setState({ status: STATUS_FINISHED });
+    };
+    this.recognition.onerror = e => {
+      this.setState({ errorCode: e.error });
+    };
+    // this.recognition.onnomatch = e => console.log('onnomatch', e);
```
BenoitPerrot: Prefer not committing code-in-comment

stories/voice-search.stories.js
Haroenv: We squash all PRs as a single commit on merging to avoid many rebases on working branches. This allows for making the commits within the branch however necessary without needing to think about history 🙂

src/components/VoiceSearch/VoiceSearch.js
```diff
@@ -94,7 +94,6 @@ class VoiceSearch extends Component {
     this.recognition.onerror = e => {
       this.setState({ errorCode: e.error });
     };
-    // this.recognition.onnomatch = e => console.log('onnomatch', e);
```
BenoitPerrot: This is unrelated; prefer not mixing several intents in a same patch (here: remove the comment in another patch, or just search back in the history of your not-yet-merged commits to not introduce comments in the first place)

stories/voice-search.stories.js
eunjae-lee: @BenoitPerrot This whole PR is going to be squashed on merge.

stories/voice-search.stories.js
```diff
@@ -99,7 +99,7 @@ storiesOf('VoiceSearch', module)
               if (isSpeechFinal) {
                 return `Final Transcript: ${transcript}`;
               } else {
-                return `Interim Transcript: ${transcript}`;
```
BenoitPerrot: Prefer squashing this in the first commit introducing the error message (making the history cleaner)

src/components/VoiceSearch/VoiceSearch.js
```diff
@@ -127,7 +127,7 @@ class VoiceSearch extends Component {
   render() {
     const { cssClasses, templates } = this.props;
     const { status, transcript, isSpeechFinal, errorCode } = this.state;
-    const isStarted = status !== STATUS_INITIAL && status !== STATUS_FINISHED;
```
BenoitPerrot: Perform this renaming in the first commit - to make the history cleaner

src/components/VoiceSearch/VoiceSearch.js
eunjae-lee: It's removed in the latest.

stories/voice-search.stories.js
BenoitPerrot: OK clear! never mind then!

BenoitPerrot: (Why not squashing _before_ pushing the review then?)

Haroenv: This is more work for not really much benefit we decided, especially for third-party contributors. Especially since we can enforce a squash commit in the settings of the repository 🙂

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,104 @@
+const STATUS_INITIAL = 'initial';
+const STATUS_ASKING_PERMISSION = 'askingPermission';
+const STATUS_WAITING = 'waiting';
+const STATUS_RECOGNIZING = 'recognizing';
+const STATUS_FINISHED = 'finished';
+const STATUS_ERROR = 'error';
+
+type VoiceSearchHelperParams = {
+  onQueryChange: (query: string) => any;
+  onStateChange: () => any;
```
francoischalifour: This doesn't return anything.

```suggestion
  onStateChange: () => void;
```

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -0,0 +1,105 @@
...
+              typeof previousQuery !== 'undefined' &&
+              previousQuery !== query
+            ) {
+              helper.search();
+            }
+          };
+          return setQueryAndSearch;
+        })();
+        (this as any)._voiceSearchHelper = voiceSearchHelper({
+          onQueryChange: (query: string) => (this as any)._refine(query),
```
francoischalifour: The `query` typed is inferred at this point, you can omit it.

src/widgets/voice-search/voice-search.tsx
```diff
@@ -0,0 +1,105 @@
...
+  voiceListeningState: VoiceListeningState;
+};
+
+const renderer = ({
+  container,
+  cssClasses,
+  templates,
+  searchAsYouSpeak,
+}: VoiceSearchWidgetParams) => ({
+  isSupportedBrowser,
```
francoischalifour: We can remove one level of indirection by getting these params from `widgetParams`.

src/components/VoiceSearch/VoiceSearch.tsx
```diff
@@ -0,0 +1,95 @@
...
+          errorCode,
+          isListening: listening,
+          transcript,
+          isSpeechFinal,
+        }}
+        templates={templates}
+      />
+      <Template
+        templateKey="status"
+        rootTagName="div"
```
francoischalifour: It's a `div` by default.

src/components/VoiceSearch/VoiceSearch.tsx
```diff
@@ -0,0 +1,95 @@
+import React from 'preact-compat';
+import PropTypes from 'prop-types';
```
francoischalifour: We don't need PropTypes now that we use TypeScript.

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -0,0 +1,105 @@
...
+import voiceSearchHelper from '../../lib/voiceSearchHelper';
+import { VoiceListeningState } from '../../widgets/voice-search/voice-search';
+
+const withUsage = createDocumentationMessageGenerator({
+  name: 'voice-search',
+  connector: true,
+});
+
+export interface VoiceSearchRenderOptions extends RenderOptions {
+  isSupportedBrowser: () => boolean;
```
francoischalifour: Why isn't it a simple boolean?

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -0,0 +1,105 @@
...
+  isListening: () => boolean;
```
francoischalifour: Why isn't it a simple boolean?

src/components/VoiceSearch/VoiceSearch.tsx
```diff
@@ -0,0 +1,95 @@
...
+          transcript,
+          isSpeechFinal,
+        }}
+        templates={templates}
+      />
+    </div>
+  );
+};
+
+VoiceSearch.propTypes = {
```
francoischalifour: We don't need PropTypes now that we use TypeScript.

src/components/VoiceSearch/VoiceSearch.tsx
```diff
@@ -0,0 +1,95 @@
...
+  isSupportedBrowser: PropTypes.func.isRequired,
+  isListening: PropTypes.func.isRequired,
+  toggleListening: PropTypes.func.isRequired,
+  voiceListeningState: PropTypes.object.isRequired,
+  searchAsYouSpeak: PropTypes.bool,
+  templates: PropTypes.object.isRequired,
+};
+
+VoiceSearch.defaultProps = {
+  searchAsYouSpeak: false,
```
francoischalifour: You can assign the default value at the function signature.

src/components/VoiceSearch/VoiceSearch.tsx
```diff
@@ -0,0 +1,95 @@
...
+const VoiceSearch = ({
+  cssClasses,
+  isSupportedBrowser,
+  isListening,
+  toggleListening,
+  voiceListeningState,
+  searchAsYouSpeak,
+  templates,
+}: VoiceSearchProps) => {
+  const handleClick = (event: any) => {
```
francoischalifour:
```suggestion
  const handleClick = (event: React.MouseEvent<HTMLElement>) => {
```

src/lib/voiceSearchHelper/index.ts
eunjae-lee: Done

stories/voice-search.stories.ts
```diff
@@ -0,0 +1,181 @@
...
+              <p>transcript: {{transcript}}</p>
+              <p>isSpeechFinal: {{isSpeechFinal}}</p>
+          `,
+          },
+        })
+      );
+    })
+  )
+  .add(
+    'disabled by default on unsupported browser (open this in Safari, Firefox, ...)',
```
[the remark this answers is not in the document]
eunjae-lee: Right, but I explicitly made it to remind me to test the disabled status on unsupported browser. Is there any better way to indicate it?

stories/voice-search.stories.ts
```diff
@@ -0,0 +1,181 @@
...
+          display: none;
+        }`,
+        `.voice-search-status .layer.listening-true {
+          display: flex;
+        }`,
+        `.voice-search-status .layer span {
+          font-size: 2rem;
+          color: #555;
+        }`,
+      ].forEach(rule => (style!.sheet! as CSSStyleSheet).insertRule(rule));
```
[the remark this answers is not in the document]
eunjae-lee: Without it, my VSCode complaint, but now it doesn't.... It's weird. Anyway I've removed it.

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export default function voiceSearchHelper({
+  onQueryChange,
+  onStateChange,
+}: VoiceSearchHelperParams) {
+  const SpeechRecognition =
+    (window as any).webkitSpeechRecognition ||
+    (window as any).SpeechRecognition;
+  const defaultState = (status: string) => ({
```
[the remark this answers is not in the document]
eunjae-lee: Done

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  const stop = () => {
+    if (recognition) {
+      recognition.stop();
+      recognition = undefined;
+    }
+    resetState();
+  };
+
+  const start = (searchAsYouSpeak: boolean) => {
```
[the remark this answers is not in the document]
eunjae-lee: Done

src/connectors/voice-search/connectVoiceSearch.ts
eunjae-lee: VoiceSearchHelper has a function with same name, but here it doesn't have to be a function. So I've changed it to a boolean variable.

src/connectors/voice-search/connectVoiceSearch.ts
eunjae-lee: Done

src/components/VoiceSearch/VoiceSearch.tsx
eunjae-lee: As far as I know, PropTypes provides runtime type checking, but assuming our users will not use the component directly, so there won't be any case of wrong props delivered to the component, we can remove Proptypes and stick to TypeScript for compile time type checking. Is this correct?

francoischalifour: Yes – PropTypes don't bring much more than TypeScript in this project since we're the only consumers (our components are not exposed). It's not worth carrying that redundancy for new components.

eunjae-lee: Clear. Thanks!

src/components/VoiceSearch/VoiceSearch.tsx
eunjae-lee: Done

stories/voice-search.stories.ts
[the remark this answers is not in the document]
francoischalifour: It shouldn't really be a story since you don't use an option to make the widget disabled. I'm not sure I understand why it's here.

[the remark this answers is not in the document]
eunjae-lee: Okay, got your point.

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,104 @@
...
+    (window as any).webkitSpeechRecognition ||
+    (window as any).SpeechRecognition;
+  const defaultState = (status: string) => ({
+    status,
+    transcript: '',
+    isSpeechFinal: undefined,
+    errorCode: undefined,
+  });
+  let state = defaultState(STATUS_INITIAL);
+  let recognition: any;
```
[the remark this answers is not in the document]
eunjae-lee: You mean `speechRecognition` (with first letter lowercased)?

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,104 @@
...
+type VoiceSearchHelperParams = {
+  onQueryChange: (query: string) => any;
+  onStateChange: () => any;
+};
+
+export default function voiceSearchHelper({
+  onQueryChange,
+  onStateChange,
+}: VoiceSearchHelperParams) {
+  const SpeechRecognition =
```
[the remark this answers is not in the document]
eunjae-lee: Done

src/types/template.ts
```diff
@@ -0,0 +1 @@
+export type Template = string | ((items: object[]) => string);
```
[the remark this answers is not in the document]
eunjae-lee: Done

src/components/VoiceSearch/VoiceSearch.tsx
eunjae-lee: Done

src/components/VoiceSearch/__tests__/VoiceSearch-test.js
```diff
@@ -0,0 +1,81 @@
...
+      expect(mount(<VoiceSearch {...defaultProps} />)).toMatchSnapshot();
+    });
+
+    it('button disabled in unsupported browser', () => {
+      const props = {
+        ...defaultProps,
+        isSupportedBrowser: false,
+      };
+      const wrapper = mount(<VoiceSearch {...props} />);
+      expect(wrapper.find('button').props().disabled).toBeTruthy();
```
francoischalifour: Better be strict.

```suggestion
      expect(wrapper.find('button').props().disabled).toBe(true);
```

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,107 @@
...
+  const getDefaultState = (status: string) => ({
+    status,
+    transcript: '',
+    isSpeechFinal: undefined,
+    errorCode: undefined,
+  });
+  let state = getDefaultState(STATUS_INITIAL);
+  let recognition: any;
+
+  const isSupportedBrowser = () => Boolean(SpeechRecognition);
```
francoischalifour: This can be a boolean instead of a function. A more correct name could be `isBrowserSupported`.

src/lib/voiceSearchHelper/__tests__/index-test.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+  );
+
+describe('VoiceSearchHelper', () => {
+  beforeEach(() => {
+    delete window.webkitSpeechRecognition;
+    delete window.SpeechRecognition;
+  });
+
+  it('test nothing', () => {
```
francoischalifour: What is this for?

src/lib/voiceSearchHelper/__tests__/index-test.js
```diff
@@ -0,0 +1,154 @@
...
+    expect(true).toBeTruthy();
+  });
+
+  it('has initial state correctly', () => {
+    const helper = getHelper();
+    expect(helper.getState()).toMatchSnapshot();
```
francoischalifour: Could use an actual object with `toEqual({ ... })`.

src/lib/voiceSearchHelper/__tests__/index-test.js
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it('is not supported', () => {
+    const helper = getHelper();
+    expect(helper.isSupportedBrowser()).toBeFalsy();
```
francoischalifour: Could use `toBe(false)`.

src/lib/voiceSearchHelper/__tests__/index-test.js
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it('is not listening', () => {
+    const helper = getHelper();
+    expect(helper.isListening()).toBeFalsy();
```
francoischalifour: Could use `toBe(false)`.

src/lib/voiceSearchHelper/__tests__/index-test.js
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it('is supported with webkitSpeechRecognition', () => {
+    window.webkitSpeechRecognition = () => {};
+    const helper = getHelper();
+    expect(helper.isSupportedBrowser()).toBeTruthy();
```
francoischalifour: Could use `toBe(true)`.

src/lib/voiceSearchHelper/__tests__/index-test.js
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it('is supported with SpeechRecognition', () => {
+    window.SpeechRecognition = () => {};
+    const helper = getHelper();
+    expect(helper.isSupportedBrowser()).toBeTruthy();
```
francoischalifour: Could use `toBe(true)`.

(same everywhere else)

src/lib/voiceSearchHelper/__tests__/index-test.js
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it('works with mock SpeechRecognition (searchAsYouSpeak:false)', () => {
+    let that;
```
francoischalifour: This is called `recognition`, right? It's more descriptive than `that`.

src/lib/voiceSearchHelper/__tests__/index-test.js
```diff
@@ -0,0 +1,154 @@
...
+    }));
+    const onQueryChange = jest.fn();
+    const onStateChange = jest.fn();
+    const helper = getHelper({
+      onQueryChange,
+      onStateChange,
+    });
+    const { getState } = helper;
+    helper.toggleListening({ searchAsYouSpeak: false });
+    expect(onStateChange).toHaveBeenCalled();
```
francoischalifour: Could be stricter with `toHaveBeenCalledTimes(1)`.

src/lib/voiceSearchHelper/__tests__/index-test.js
```diff
@@ -0,0 +1,154 @@
...
+          ];
+          obj.isFinal = true;
+          return obj;
+        })(),
+      ],
+    });
+    expect(getState().status).toEqual('recognizing');
+    expect(getState().transcript).toEqual('Hello World');
+    expect(getState().isSpeechFinal).toBeTruthy();
+    expect(onQueryChange).not.toHaveBeenCalled();
```
francoischalifour: `toHaveBeenCalledTimes(0)` usually gives better error messages and is easier to update in the long run.

src/widgets/voice-search/voice-search.tsx
eunjae-lee: Nice! I had a hard time with that renderer and now it's clearer.

eunjae-lee:
@francoischalifour Thanks for your feedbacks! They're applied. I'd appreciate your feedbacks on tests.

eunjae-lee:
I need to discuss something. `VoiceSearchHelper` is instantiated in `connectVoiceSearch`. The instance stores bunch of statuses. However the connector is called only once when creating VoiceSearch widget. It means if user puts VoiceSearch widget more than one on a same page, all the widgets will have same statuses. Is this correct?
Having more than one VoiceSearch widget doesn't make any sense. And even if user does, the widgets cannot be used at the same time. Do we consider statuses being shared in several widgets as an issue? If so, the instantiation of the helper should be done in creation of the component. IMO, it's not an issue. What do you think? @Haroenv @francoischalifour

src/lib/voiceSearchHelper/index.ts
eunjae-lee: @francoischalifour Right, this can be a boolean. Is `isBrowserSupported` okay to be a boolean while `isListening` is a function?

src/lib/voiceSearchHelper/__tests__/index-test.js
eunjae-lee: I should've removed it.

src/lib/voiceSearchHelper/__tests__/index-test.js
eunjae-lee: Done

src/lib/voiceSearchHelper/__tests__/index-test.js
eunjae-lee: Done

src/lib/voiceSearchHelper/__tests__/index-test.js
eunjae-lee: Done

src/lib/voiceSearchHelper/__tests__/index-test.js
eunjae-lee: Done

src/lib/voiceSearchHelper/index.ts
francoischalifour: That's a good point. It would make sense to keep the same type for both of them.

src/widgets/voice-search/voice-search.tsx
```diff
@@ -0,0 +1,106 @@
...
+  searchAsYouSpeak?: boolean;
+};
+
+type VoiceSearchWidgetRenderParams = {
+  isSupportedBrowser: boolean;
+  isListening: boolean;
+  toggleListening: (toggleListeningParams: {
+    searchAsYouSpeak: boolean;
+  }) => void;
+  voiceListeningState: VoiceListeningState;
```
samouss: What is the purpose to scope the state under an attribute vs top level?

src/widgets/voice-search/voice-search.tsx
```diff
@@ -0,0 +1,106 @@
...
+  widgetParams: VoiceSearchWidgetParams;
+};
+
+const renderer: VoiceSearchRenderer<VoiceSearchWidgetParams> = ({
+  isSupportedBrowser,
+  isListening,
+  toggleListening,
+  voiceListeningState,
+  widgetParams,
+}: VoiceSearchWidgetRenderParams) => {
```
samouss: You don't have to provide the type, values are already inferred from the `VoiceSearchRenderer`.

src/widgets/voice-search/voice-search.tsx
```diff
@@ -0,0 +1,106 @@
...
+};
+
+type VoiceSearchWidgetParams = {
+  container: string | HTMLElement;
+  cssClasses: VoiceSearchCSSClasses;
+  templates: VoiceSearchTemplates;
+  searchAsYouSpeak?: boolean;
+};
+
+type VoiceSearchWidgetRenderParams = {
```
samouss: This type is not useful, those information are already available through the renderer. The type that we have to define here is the `widgetParams` that is provided to the `render` function. The type is different from the one provided to the widget because some values are now required (like `cssClasses` & `templates`). You can take example on the [`QueryRulesCustomData`](https://github.com/algolia/instantsearch.js/blob/577e4ccb33155d0d4678b04505132d39a2918f22/src/widgets/query-rule-custom-data/query-rule-custom-data.tsx#L22-L36) widget.

stories/voice-search.stories.ts
```diff
@@ -0,0 +1,186 @@
...
+      const descContainer = document.createElement('div');
+      const realContainer = document.createElement('div');
+      container.appendChild(descContainer);
+      container.appendChild(realContainer);
+      descContainer.innerHTML = `
+        <p>To see this button disabled, test it on unsupported browsers like Safari, Firefox, etc.</p>
+      `;
+
+      search.addWidget(
+        instantsearch.widgets.voiceSearch({
```
samouss: Until we have a proper definition for `widgets` we should use the source of the widget. Otherwise we don't leverage TypeScript to warn us about the usage of the widget. Here the usage does not match the definition of the types since the widget is inferred as `any`. cc @francoischalifour

src/lib/voiceSearchHelper/index.ts
eunjae-lee: Okay, then I'll leave both of them as function.

src/components/VoiceSearch/VoiceSearch.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+type VoiceSearchProps = {
+  cssClasses: VoiceSearchCSSClasses;
+  isSupportedBrowser: boolean;
+  isListening: boolean;
+  toggleListening: (toggleListeningParams: {
+    searchAsYouSpeak: boolean;
+  }) => void;
+  voiceListeningState: VoiceListeningState;
+  searchAsYouSpeak?: boolean;
```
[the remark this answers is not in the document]
eunjae-lee: Ahah right. It will be filled in widget level.

src/components/VoiceSearch/VoiceSearch.tsx
```diff
@@ -0,0 +1,78 @@
...
+const VoiceSearch = ({
+  cssClasses,
+  isSupportedBrowser,
+  isListening,
+  toggleListening,
+  voiceListeningState,
+  searchAsYouSpeak = false,
+  templates,
+}: VoiceSearchProps) => {
+  const handleClick = (event: React.MouseEvent<HTMLElement>) => {
```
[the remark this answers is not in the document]
eunjae-lee: @samouss okay, I will change it to `any`. BTW, could you help me understand more why it doesn't work in the current setup?

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -0,0 +1,112 @@
...
+          isFirstRendering: false,
+          instantSearchInstance,
+          voiceSearchHelper: (this as any)._voiceSearchHelper,
+        });
+      },
+      dispose({ state }) {
+        unmountFn();
+        return state.setQuery('');
+      },
+    };
```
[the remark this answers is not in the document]
eunjae-lee: I didn't know. After reading [this](https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-own-widgets/js/?language=react#the-widget-lifecycle-and-api), I guess I need to define `getWidgetState` and `getWidgetSearchParameters`, right?
And in case of VoiceSearch, the implementations will be the same as SearchBox.
Let me know if I'm wrong about this. @samouss cc @Haroenv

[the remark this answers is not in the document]
samouss: You're right. We can postpone it to a latter PR though because this one is already consequent.

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -0,0 +1,112 @@
...
+  isListening: boolean;
+  toggleListening: (toggleListeningParams: {
+    searchAsYouSpeak: boolean;
+  }) => void;
+  voiceListeningState: VoiceListeningState;
+}
+
+export type VoiceSearchRenderer<T> = Renderer<VoiceSearchRenderOptions<T>>;
+
+export type VoiceSearchWidgetFactory<T> = WidgetFactory<T>;
```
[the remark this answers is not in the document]
eunjae-lee: Thanks. Done!

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,110 @@
...
+};
+
+export default function voiceSearchHelper({
+  onQueryChange,
+  onStateChange,
+}: VoiceSearchHelperParams) {
+  const SpeechRecognition: SpeechRecognition =
+    (window as any).webkitSpeechRecognition ||
+    (window as any).SpeechRecognition;
+  const getDefaultState = (status: string) => ({
```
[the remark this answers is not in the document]
eunjae-lee: Done

src/widgets/voice-search/voice-search.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+export type VoiceSearchCSSClasses = {
+  root: string;
+  button: string;
+  status: string;
+};
+
+export type VoiceSearchTemplates = {
+  buttonText: Template<object>;
+  status: Template<object>;
```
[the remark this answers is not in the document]
eunjae-lee: Done

src/widgets/voice-search/voice-search.tsx
eunjae-lee: Done

src/widgets/voice-search/voice-search.tsx
```diff
@@ -0,0 +1,106 @@
...
+};
+
+export type VoiceListeningState = {
+  status: string;
+  transcript?: string;
```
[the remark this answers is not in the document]
eunjae-lee: Once speech recognition is started, `transcript` can have some value, but before it's started, it is impossible for `transcript` to have any value. That's why I made it optional. What do you think?

src/widgets/voice-search/voice-search.tsx
```diff
@@ -0,0 +1,106 @@
...
+  isSpeechFinal?: boolean;
```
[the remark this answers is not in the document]
eunjae-lee: Same with `transcript`.

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,110 @@
...
+      setState({
+        status: STATUS_WAITING,
+      });
+    };
+
+    recognition.start();
+  };
+
+  const toggleListening = ({
+    searchAsYouSpeak,
```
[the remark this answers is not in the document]
eunjae-lee: You're right. `searchAsYouSpeak` can be provided from widgetParams!
I'm removing `searchAsYouSpeak` from component props and provide it directly to voiceSearchHelper function.

src/lib/voiceSearchHelper/__tests__/index-test.js
```diff
@@ -0,0 +1,158 @@
...
+      },
+    }));
+    const onQueryChange = jest.fn();
+    const onStateChange = jest.fn();
+    const helper = getHelper({
+      onQueryChange,
+      onStateChange,
+    });
+    const { getState } = helper;
+    helper.toggleListening({ searchAsYouSpeak: false });
```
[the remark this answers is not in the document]
eunjae-lee: Done

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  const getState = () => state;
+
+  const resetState = (status = STATUS_INITIAL) => {
+    setState(getDefaultState(status));
+  };
+
+  const stop = () => {
+    if (recognition) {
+      recognition.stop();
```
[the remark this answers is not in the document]
eunjae-lee: @samouss I'm not confident with this. Before the reference to `recognition` is lost by `recognition = undefined;`, do we need to do something to clean up the listeners, when listeners are registered as `recognition.onxxx = function () {}`?

src/widgets/voice-search/voice-search.tsx
eunjae-lee: Thanks for the explanation and the example link! 👍

src/widgets/voice-search/voice-search.tsx
eunjae-lee: `voiceListeningState` is an object returned from `getState` function at `VoiceSearchHelper`. There is no strong opinion why I let it scoped or why I didn't spread it.

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,110 @@
...
+    recognition.onend = () => {
+      if (!state.errorCode && state.transcript && !searchAsYouSpeak) {
+        onQueryChange(state.transcript);
+      }
+      if (state.status !== STATUS_ERROR) {
+        setState({ status: STATUS_FINISHED });
+      }
+    };
+    recognition.onerror = (event: SpeechRecognitionError) => {
+      setState({ status: STATUS_ERROR, errorCode: event.error });
```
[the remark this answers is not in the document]
eunjae-lee: If user triggers 'toggleListening' again, it will start a new speech recognition.

src/connectors/voice-search/connectVoiceSearch.ts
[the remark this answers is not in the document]
eunjae-lee: Done

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -0,0 +1,82 @@
...
+  const helper = jsHelper({});
+
+  return {
+    renderFn,
+    widget,
+    helper,
+  };
+}
+
+describe('connectVoiceSearch', () => {
```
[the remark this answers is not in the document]
eunjae-lee: Added more tests.

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,110 @@
...
+  onStateChange,
+}: VoiceSearchHelperParams) {
+  const SpeechRecognition: SpeechRecognition =
+    (window as any).webkitSpeechRecognition ||
+    (window as any).SpeechRecognition;
+  const getDefaultState = (status: string) => ({
+    status,
+    transcript: '',
+    isSpeechFinal: undefined,
+    errorCode: undefined,
```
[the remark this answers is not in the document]
eunjae-lee: These are used to overwrite the states to initial values.

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,123 @@
...
+};
+
+export type ToggleListening = () => void;
+
+export default function voiceSearchHelper({
+  searchAsYouSpeak,
+  onQueryChange,
+  onStateChange,
+}: VoiceSearchHelperParams) {
+  const SpeechRecognitionClass: SpeechRecognition =
```
francoischalifour:
```suggestion
  const SpeechRecognitionAPI: SpeechRecognition =
```

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,123 @@
...
+  const getDefaultState = (status: string): VoiceListeningState => ({
+    status,
+    transcript: undefined,
+    isSpeechFinal: undefined,
+    errorCode: undefined,
+  });
+  let state: VoiceListeningState = getDefaultState(STATUS_INITIAL);
+  let recognition: SpeechRecognition | undefined;
+
+  const isBrowserSupported = () => Boolean(SpeechRecognitionClass);
```
francoischalifour:
```suggestion
  const isBrowserSupported = () => Boolean(SpeechRecognitionAPI);
```

.storybook/preview-head.html
```diff
@@ -1,2 +1,5 @@
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.1.1/themes/algolia.min.css">
-<link rel="stylesheet" href="storybook.css">
+<link
+  rel="stylesheet"
+  href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.3.0/themes/algolia.min.css"
```
francoischalifour: InstantSearch.css 7.3.0 is damaged and doesn't include the changes made in 7.2.0 (previous button styles).

- 7.2.0: https://unpkg.com/instantsearch.css@7.2.0/themes/algolia.css
- 7.3.0: https://unpkg.com/instantsearch.css@7.3.0/themes/algolia.css

We released a new patch including the fix.

```suggestion
  href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.3.1/themes/algolia.min.css"
```

src/widgets/voice-search/defaultTemplates.js
```diff
@@ -0,0 +1,12 @@
+export default {
+  buttonText({ status, errorCode, isListening }) {
+    if (status === 'error' && errorCode === 'not-allowed') {
+      return `<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="feather feather-mic-off"><line x1="1" y1="1" x2="23" y2="23"></line><path d="M9 9v3a3 3 0 0 0 5.12 2.12M15 9.34V4a3 3 0 0 0-5.94-.6"></path><path d="M17 16.95A7 7 0 0 1 5 12v-2m14 0v2a7 7 0 0 1-.11 1.23"></path><line x1="12" y1="19" x2="12" y2="23"></line><line x1="8" y1="23" x2="16" y2="23"></line></svg>`;
```
francoischalifour: Can we prettify those SVGs?

We should be able to not repeat the outer `svg` element, only the inner elements differ.

src/widgets/voice-search/defaultTemplates.js
```diff
@@ -5,13 +5,11 @@ const getButtonInnerElement = (status, errorCode, isListening) => {
             <path d="M17 16.95A7 7 0 0 1 5 12v-2m14 0v2a7 7 0 0 1-.11 1.23"></path>
             <line x1="12" y1="19" x2="12" y2="23"></line>
             <line x1="8" y1="23" x2="16" y2="23"></line>`;
-  } else if (isListening) {
-    return `<path d="M12 1a3 3 0 0 0-3 3v8a3 3 0 0 0 6 0V4a3 3 0 0 0-3-3z" fill="currentColor"></path>
-            <path d="M19 10v2a7 7 0 0 1-14 0v-2"></path>
-            <line x1="12" y1="19" x2="12" y2="23"></line>
-            <line x1="8" y1="23" x2="16" y2="23"></line>`;
   } else {
```
Haroenv: else isn't needed if the `if` returns

src/components/VoiceSearch/__tests__/VoiceSearch-test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    it('with custom template for buttonText (1)', () => {
+      const props = {
+        ...defaultProps,
+        isListening: true,
+        templates: {
+          buttonText: ({ isListening }) => (isListening ? 'Stop' : 'Start'),
+          status: ``,
+        },
+      };
+      expect(mount(<VoiceSearch {...props} />)).toMatchSnapshot();
```
francoischalifour: We can get rid of the overall snapshot and test the actual content of the template. It makes the intent of the test clearer:

```js
const wrapper = mount(<VoiceSearch {...props} />);
expect(wrapper.find('.button').text()).toBe('Stop');
```

src/components/VoiceSearch/__tests__/VoiceSearch-test.tsx
```diff
@@ -0,0 +1,99 @@
...
+    it('with custom template for buttonText (2)', () => {
+      const props = {
+        ...defaultProps,
+        isListening: false,
+        templates: {
+          buttonText: ({ isListening }) => (isListening ? 'Stop' : 'Start'),
+          status: ``,
+        },
+      };
+      expect(mount(<VoiceSearch {...props} />)).toMatchSnapshot();
```
francoischalifour: We can get rid of the overall snapshot and test the actual content of the template. It makes the intent of the test clearer.

src/components/VoiceSearch/__tests__/VoiceSearch-test.tsx
```diff
@@ -0,0 +1,99 @@
...
+            <p>status: {{status}}</p>
+            <p>errorCode: {{errorCode}}</p>
+            <p>isListening: {{isListening}}</p>
+            <p>transcript: {{transcript}}</p>
+            <p>isSpeechFinal: {{isSpeechFinal}}</p>
+            <p>isBrowserSupported: {{isBrowserSupported}}</p>
+          `,
+        },
+      };
+      expect(mount(<VoiceSearch {...props} />)).toMatchSnapshot();
```
francoischalifour: We can reduce the scope of the snapshot with:

```js
expect(wrapper.find('.status')).toMatchSnapshot();
```

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -0,0 +1,188 @@
...
+
+  helper.search = () => {};
+  widget.init({ helper });
+
+  return {
+    renderFn,
+    unmountFn,
+    widget,
+    helper,
+    refine: query => widget._refine(query),
```
francoischalifour: Do we need to rely on the `_.refine()` internal? We should be able to get it from the rendering parameters.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -0,0 +1,188 @@
...
+See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-search/js/#connector"
+`);
+    });
+  });
+
+  describe('lifecycle', () => {
+    it('calls renderFn during init and render', () => {
+      const { renderFn, widget, helper } = getDefaultSetup();
+      widget.init({ helper });
+      expect(renderFn).toHaveBeenCalled();
```
francoischalifour: Let's be more strict.

```suggestion
      expect(renderFn).toHaveBeenCalledTimes(1);
```

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -0,0 +1,188 @@
...
+      widget.dispose({ helper, state: helper.getState() });
+      expect(unmountFn).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  it('render triggered when state changes', () => {
+    const { renderFn, widget, helper } = getDefaultSetup();
+    widget.init({ helper });
+    expect(renderFn).toHaveBeenCalled();
+    widget._voiceSearchHelper.changeState();
```
francoischalifour: It'd be nice not to touch the internals in our tests. Not sure how to achieve that here.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -0,0 +1,188 @@
...
+      widget.init({ helper });
+      widget.dispose({ helper, state: helper.getState() });
+      expect(unmountFn).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  it('render triggered when state changes', () => {
+    const { renderFn, widget, helper } = getDefaultSetup();
+    widget.init({ helper });
+    expect(renderFn).toHaveBeenCalled();
```
francoischalifour:
```suggestion
    expect(renderFn).toHaveBeenCalledTimes(1);
```

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -0,0 +1,188 @@
...
+
+    it('calls unmount on dispose', () => {
+      const { unmountFn, widget, helper } = getDefaultSetup();
+      widget.init({ helper });
+      widget.dispose({ helper, state: helper.getState() });
+      expect(unmountFn).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  it('render triggered when state changes', () => {
```
francoischalifour:
```suggestion
  it('triggers render when state changes', () => {
```

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -0,0 +1,188 @@
...
+      const uiState = {
+        query: 'my search',
+      };
+      const searchParametersBefore = SearchParameters.make(helper.state);
+      const searchParametersAfter = widget.getWidgetSearchParameters(
+        searchParametersBefore,
+        { uiState }
+      );
+      expect(searchParametersAfter.query).toEqual(uiState.query);
+      expect(searchParametersAfter).toMatchSnapshot();
```
francoischalifour: What do you expect to catch with this snapshot? Is it necessary?

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -0,0 +1,188 @@
...
+      const { widget, helper, refine } = getInitializedWidget();
+      refine('previous search');
+      const uiState = {};
+      const searchParametersBefore = SearchParameters.make(helper.state);
+      const searchParametersAfter = widget.getWidgetSearchParameters(
+        searchParametersBefore,
+        { uiState }
+      );
+      expect(searchParametersAfter.query).toBe('');
+      expect(searchParametersAfter).toMatchSnapshot();
```
francoischalifour: What do you expect to catch with this snapshot? Is it necessary?

src/widgets/voice-search/__tests__/voice-search-test.ts
```diff
@@ -0,0 +1,103 @@
...
+      },
+      createURL: () => '',
+    });
+  };
+
+  return { container, widget, widgetInit, widgetRender };
+}
+
+describe('voiceSearch()', () => {
+  let helper;
```
francoischalifour: Could be typed `Partial<Helper>` if needed.

src/components/VoiceSearch/__tests__/VoiceSearch-test.tsx
eunjae-lee: That's a good idea.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
eunjae-lee: `connectVoiceSearch` doesn't provide `refine` as a rendering parameter to its component. All the state changes are happening in the instance of `voiceSearchHelper` and it's inside the connector. So it triggers refinement inside the connector.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
eunjae-lee:
```
                                    (A)                             (B)
voiceSearchHelper internal logics  ----> `onStateChange` triggered ----> render triggered
```

I want to test (B) and now the voiceSearchHelper is mocked. So I had to fake (A). I don't know if it's possible without touching the internals of voiceSearchHelper.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
eunjae-lee: You're right. It's not necessary and doesn't give any benefit. I've removed them.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
eunjae-lee: You're right. It's not necessary and doesn't give any benefit. I've removed them.

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -0,0 +1,138 @@
...
+const withUsage = createDocumentationMessageGenerator({
+  name: 'voice-search',
+  connector: true,
+});
+
+export type VoiceSearchConnectorParams = {
+  searchAsYouSpeak: boolean;
+};
+
+export interface VoiceSearchRenderOptions<T> extends RenderOptions<T> {
```
francoischalifour: If you update this branch you will have linting errors because of the generic name `T`.

src/lib/voiceSearchHelper/__tests__/index-test.ts
```diff
@@ -0,0 +1,170 @@
...
+    window.SpeechRecognition = () => {};
+    const helper = getHelper();
+    expect(helper.isBrowserSupported()).toBe(true);
+  });
+
+  it('works with mock SpeechRecognition (searchAsYouSpeak:false)', () => {
+    let recognition;
+    window.SpeechRecognition = jest.fn().mockImplementation(() => ({
+      start() {
+        /* eslint-disable-next-line consistent-this */
```
francoischalifour: This doesn't seem to be necessary anymore.

src/lib/voiceSearchHelper/__tests__/index-test.ts
```diff
@@ -0,0 +1,170 @@
...
+    recognition.onend();
+    expect(onQueryChange).toHaveBeenCalledWith('Hello World');
+    expect(getState().status).toEqual('finished');
+  });
+
+  it('works with mock SpeechRecognition (searchAsYouSpeak:true)', () => {
+    let recognition;
+    window.SpeechRecognition = jest.fn().mockImplementation(() => ({
+      start() {
+        /* eslint-disable-next-line consistent-this */
```
francoischalifour: This doesn't seem to be necessary anymore.

src/lib/voiceSearchHelper/__tests__/index-test.ts
```diff
@@ -0,0 +1,170 @@
...
+    recognition.onend();
+    expect(onQueryChange).toHaveBeenCalledTimes(1);
+    expect(getState().status).toEqual('finished');
+  });
+
+  it('works with onerror', () => {
+    let recognition;
+    window.SpeechRecognition = jest.fn().mockImplementation(() => ({
+      start() {
+        /* eslint-disable-next-line consistent-this */
```
francoischalifour: This doesn't seem to be necessary anymore.

src/lib/voiceSearchHelper/index.ts
```diff
@@ -0,0 +1,123 @@
...
+  const stop = () => {
+    if (recognition) {
+      recognition.stop();
+      recognition = undefined;
+    }
+    resetState();
+  };
+
+  const start = () => {
+    recognition = new SpeechRecognitionAPI();
```
francoischalifour: Are there any drawbacks at recreating a new speech recognition every time time we start listening?

src/lib/voiceSearchHelper/__tests__/index-test.ts
eunjae-lee: Without disabling the rule, it still gives me an error.

```
Unexpected alias 'recognition' for 'this'  consistent-this
```

src/lib/voiceSearchHelper/index.ts
eunjae-lee: I have no idea. I haven't read on the [spec](https://w3c.github.io/speech-api) anything about reusing the instance, cleaning it up, etc. I guess recreating a new speech recognition shouldn't be a burden.

src/lib/voiceSearchHelper/__tests__/index-test.ts
francoischalifour: Indeed, my bad!

PR #3602, opened 2019-03-14
feat(queryRules): add queryRuleContext widget

francoischalifour:
_This is the 3rd PR in the series "Merchandized Query Rules"._

## Summary

This **headless** widget sets the rule contexts (called [`ruleContexts`](https://www.algolia.com/doc/api-reference/api-parameters/ruleContexts/)) of the search parameters based on the filters.

Query rules are by default _generic_ when not provided any context. If you provide a context, they become _contextual_ and are only triggered when the context is matched.

## Usage

```js
instantsearch.widgets.queryRuleContext({
  container,
  trackedFilters: {
    genre: () => ['Thriller', 'Drama'],
  },
});
```

This will set the following `ruleContexts` dynamically:

- `ais-genre-Thriller`
- `ais-genre-Drama`

### `trackedFilters`

> `object` – no default

The filters to track to add rule contexts.

### `transformRuleContexts`

> `string[] ⇒ string[]` – defaults to identity function

The function to apply to the rule contexts before sending them to Algolia.

## Stories

[See stories →](https://deploy-preview-3602--instantsearchjs.netlify.com/stories/?path=/story/queryrulecontext--default)

## Related

- [`connectQueryRules`](https://github.com/algolia/instantsearch.js/pull/3597)
- [`queryRuleCustomData`](https://github.com/algolia/instantsearch.js/pull/3600)

### DIFF
--- a/src/widgets/query-rule-context/__tests__/query-rule-context-test.ts
+++ b/src/widgets/query-rule-context/__tests__/query-rule-context-test.ts
@@ -1,0 +1,18 @@
+import queryRuleContext from '../query-rule-context';
+
+describe('queryRuleContext', () => {
+  describe('Usage', () => {
+    test('does not throw without options', () => {
+      expect(() => {
+        // @ts-ignore
+        queryRuleContext();
+      }).not.toThrow();
+    });
+
+    test('does not throw with empty options', () => {
+      expect(() => {
+        queryRuleContext({});
+      }).not.toThrow();
+    });
+  });
+});
--- a/src/widgets/query-rule-context/__tests__/query-rule-context-test.ts
+++ b/src/widgets/query-rule-context/__tests__/query-rule-context-test.ts
@@ -2,17 +2,26 @@ import queryRuleContext from '../query-rule-context';
 
 describe('queryRuleContext', () => {
   describe('Usage', () => {
-    test('does not throw without options', () => {
+    test('throws trackedFilters error without options', () => {
       expect(() => {
         // @ts-ignore
         queryRuleContext();
-      }).not.toThrow();
+      }).toThrowErrorMatchingInlineSnapshot(`
+"The \`trackedFilters\` option is required.
+
+See documentation: https://www.algolia.com/doc/api-reference/widgets/query-rule-context/js/"
+`);
     });
 
-    test('does not throw with empty options', () => {
+    test('throws trackedFilters error with empty options', () => {
       expect(() => {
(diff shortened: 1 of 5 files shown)


stories/query-rule-context.stories.ts
```diff
@@ -0,0 +1,63 @@
...
+        trackedFilters: {
+          genre: () => ['Thriller', 'Drama'],
+        },
+      })
+    );
+
+    search.addWidget(
+      instantsearch.widgets.queryRuleCustomData({
+        container: widgetContainer,
+        transformItems: (items: CustomDataItem[]) => items[0],
```
Haroenv: this story proves it was good that we put this API here, otherwise we'd have two banners visible. Deciding which one to choose is still totally easy to do and in user land 🙌

src/widgets/query-rule-context/__tests__/query-rule-context-test.ts
```diff
@@ -0,0 +1,18 @@
+import queryRuleContext from '../query-rule-context';
+
+describe('queryRuleContext', () => {
+  describe('Usage', () => {
+    test('does not throw without options', () => {
```
Haroenv: do we not want it to throw without `trackedFilters`?

francoischalifour: That's a fair point – what's your thoughts @algolia/instantsearch-for-websites?

samouss: Right now it's a noop in the connector so we can throw an error. 

But it means that we cannot track everything by default. Worth asking to the SE if it's a use case or not.

francoischalifour: We couldn't track everything by default anyway. The default behavior is to track nothing.

Besides, we don't recommend tracking everything because there's a limitation of 10 `ruleContexts` sent to the engine.

I will update the widget to throw if no `trackedFilters` are passed.

francoischalifour:
This feature is on hold – see https://github.com/algolia/instantsearch.js/pull/3597#issuecomment-474809122.

francoischalifour:
⬆️ I rebased against #3617 with the workaround for setting the context and other search parameters in a single request. You can [check the story](https://deploy-preview-3602--instantsearchjs.netlify.com/stories/?path=/story/queryrulecontext--default).

PR #3614, opened 2019-03-18
build(travis): fix matrix scripts

francoischalifour:
I realized that the CI matrix scripts run with `yarn` were not actually executed since #3481 (for two weeks)... Quotes are needed for the scripts to be run (it ran only `yarn`).

I linted the files that didn't pass at the same time.

Once this PR is merged, please rebase/merge your PRs against `develop` so that CI checks are effective again.

_Lesson to self: always check the CI tests when the configuration changes, even when they pass._

:facepalm:

### DIFF
--- a/src/lib/utils/prefixKeys.ts
+++ b/src/lib/utils/prefixKeys.ts
@@ -2,7 +2,7 @@ import mapKeys from 'lodash/mapKeys';
 
 function prefixKeys(prefix: string, obj: object) {
   if (obj) {
-    return mapKeys(obj, (_value, key) => prefix + key);
+    return mapKeys(obj, (_0, key) => prefix + key);
   }
 
   return undefined;
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,10 @@ script:
   - '${SCRIPT}'
 env:
   matrix:
-    - SCRIPT=./scripts/ci/functional-test.sh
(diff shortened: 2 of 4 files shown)


Haroenv:
Argos is failing: https://travis-ci.org/algolia/instantsearch.js/jobs/508115130#L646

cc @samouss, this is probably the same as in React InstantSearch

francoischalifour:
Note also that functional tests were disabled since v3 development.

samouss:
Not really the same issue, we don't upload the screenshots anymore. Not sure we even run them. For React InstantSearch they are uploaded the diff is even working. We don't get the check on the PR that's all.

src/lib/utils/prefixKeys.ts
```diff
@@ -2,7 +2,7 @@ import mapKeys from 'lodash/mapKeys';
 
 function prefixKeys(prefix: string, obj: object) {
   if (obj) {
-    return mapKeys(obj, (_value, key) => prefix + key);
+    return mapKeys(obj, (_0, key) => prefix + key);
```
tkrugg: _value was fine no?

francoischalifour: TSLint can't ignore underscore leading parameters right now. It complains when not following the format `_{number}`.

Reference: https://github.com/palantir/tslint/issues/3442.

PR #3616, opened 2019-03-22
fix(ua): Update the User-Agent to use the new format

robertmogos:
**Summary**
Update the UA from `instantsearch.js ${version}` to `instantsearch.js (${version})`

### DIFF
--- a/src/lib/__tests__/InstantSearch-test.js
+++ b/src/lib/__tests__/InstantSearch-test.js
@@ -2,6 +2,7 @@ import range from 'lodash/range';
 import times from 'lodash/times';
 import algoliaSearchHelper from 'algoliasearch-helper';
 import InstantSearch from '../InstantSearch';
+import version from '../version'
 
 jest.mock('algoliasearch-helper', () => {
   const module = require.requireActual('algoliasearch-helper');
@@ -151,6 +152,7 @@ describe('InstantSearch lifecycle', () => {
           results: requests.map(() => ({})),
         });
       },
+      addAlgoliaAgent: jest.fn()
     };
 
     algoliasearch = jest.fn().mockReturnValue(client);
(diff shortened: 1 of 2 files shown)


francoischalifour:
Can we add a test for this?

An appropriate place could be [`InstantSearch-test.js`](https://github.com/algolia/instantsearch.js/blob/5f41fc6ae8d03c9ddd76dff5e54e73fecc395900/src/lib/__tests__/InstantSearch-test.js#L138).

robertmogos:
> Can we add a test for this?
> 
> An appropriate place could be [`InstantSearch-test.js`](https://github.com/algolia/instantsearch.js/blob/5f41fc6ae8d03c9ddd76dff5e54e73fecc395900/src/lib/__tests__/InstantSearch-test.js#L138).

Yes, it is on my todo list. I might need some help with that

francoischalifour:
You can see in the test that [we mocked the JavaScript client](https://github.com/algolia/instantsearch.js/blob/5f41fc6ae8d03c9ddd76dff5e54e73fecc395900/src/lib/__tests__/InstantSearch-test.js#L148-L154), with only a `search` method for now.

The official JavaScript client also exposes an `addAlgoliaAgent` method. You can therefore add this method to the client to mock the `addAlgoliaAgent` function with [`jest.fn()`](https://jestjs.io/docs/en/jest-object#jestfnimplementation) and check that it has been called correctly at instantiation.

Let me know if you need further help.

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -2,6 +2,7 @@ import range from 'lodash/range';
 import times from 'lodash/times';
 import algoliaSearchHelper from 'algoliasearch-helper';
 import InstantSearch from '../InstantSearch';
+import version from '../version'
```
francoischalifour: The file is not formatted: missing `;`.

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -151,6 +152,7 @@ describe('InstantSearch lifecycle', () => {
           results: requests.map(() => ({})),
         });
       },
+      addAlgoliaAgent: jest.fn()
```
francoischalifour: The file is not formatted: missing `,`.

robertmogos: 🤦‍♂️

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -182,6 +185,11 @@ describe('InstantSearch lifecycle', () => {
     expect(algoliasearch).toHaveBeenCalledWith(appId, apiKey);
   });
 
+  it('calls addAlgoliaAgent', () => {
+    expect(client.addAlgoliaAgent).toHaveBeenCalledTimes(1);
+    expect(client.addAlgoliaAgent).toHaveBeenCalledWith(`instantsearch.js (${version})`);
```
francoischalifour: This line also needs to be formatted. Running `yarn lint --fix` should fix this kind of problem.

If you use Vscode, you can toggle [format on save](https://marketplace.visualstudio.com/items?itemName=cbrevik.toggle-format-on-save).

Issue #3497, opened 2019-02-08
[RoutingManager]: getConfiguration uses a staled uiState


samouss:
**Describe the bug 🐛**

The `RoutingManager.getConfiguration` uses a stale `uiState` to compute the parameters. It always uses the initial `uiState`. It leads to issue with dynamic add / remove of widgets. Each time one or the other occurs we call `getConfiguration` on every widgets to create the next search parameters. But the `uiState` might have changed since the instance start. It means that the search parameters are always computed like if it was a fresh start.

The issue does not occurs with every type of refinement. It only occurs on top level refinement like `query`, `hitsPerPage`, etc... It's because the implementation of `enhanceConfiguration` is not correct. With the correct behaviour the issue would happen on every refinements.

**To Reproduce 🔍**

1. Go to https://codesandbox.io/s/n15lol465j
2. Type a query e.g. "Apple"
3. Click outside of the `searchBox`
3. Wait for the timeout (5000ms)
4. The `rangeInput` widget is added

Note that the `query` disappeared and the refinement is lost.

**Expected behavior 💭**

The search parameters must stay the same after the add / remove widgets operation.

**Additional context**

> [Discourse post issue with Vue InstantSearch](https://discourse.algolia.com/t/preventing-multiple-api-hits-when-updating-geolocation-via-browser-api-query-parsing/7030)

PR #3630, opened 2019-03-28
fix(RoutingManager): avoid stale uiState

samouss:
**Summary**

Closes https://github.com/algolia/instantsearch.js/issues/3497

This PR aims to fix the issue related to the stale `uiState` inside the `RoutingManager` (see the issue for more information). Previously the state was only captured at the creation of the manager and never updated. Now we update it each time an update occurs (state change, router update). It means that now once a widget is added/removed the lifecycle read the up to date `uiState`.

**Before**

before

**After**

after

### DIFF
--- a/src/lib/__tests__/RoutingManager-test.js
+++ b/src/lib/__tests__/RoutingManager-test.js
@@ -1,18 +1,18 @@
-import algoliasearch from 'algoliasearch';
 import instantsearch from '../main';
 import RoutingManager from '../RoutingManager';
 import simpleMapping from '../stateMappings/simple';
 
-const fakeAlgoliaClient = {
+const createFakeSearchClient = () => ({
   search: () => Promise.resolve({ results: [{}] }),
-};
+});
 
 describe('RoutingManager', () => {
   describe('getAllUIStates', () => {
     test('reads the state of widgets with a getWidgetState implementation', () => {
+      const searchClient = createFakeSearchClient();
       const search = instantsearch({
         indexName: '',
-        searchClient: fakeAlgoliaClient,
+        searchClient,
       });
 
       const widgetState = {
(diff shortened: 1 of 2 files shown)


src/lib/__tests__/RoutingManager-test.js
```diff
@@ -312,5 +382,167 @@ describe('RoutingManager', () => {
...
+      });
+
+      await runAllMicroTasks();
+
+      // Trigger an update - Apple iPhone → Apple
+      history.back();
+
+      await runAllMicroTasks();
+
+      // Trigger getConfigurartion
```
francoischalifour:
```suggestion
      // Trigger getConfiguration
```

src/lib/__tests__/RoutingManager-test.js
```diff
@@ -312,5 +382,167 @@ describe('RoutingManager', () => {
...
+      search.start();
+
+      await runAllMicroTasks();
+
+      expect(router.write).toHaveBeenCalledTimes(1);
+      expect(router.write).toHaveBeenLastCalledWith({
+        query: 'Apple iPhone',
+      });
+
+      // Trigger getConfigurartion
```
francoischalifour:
```suggestion
      // Trigger getConfiguration
```

src/lib/__tests__/RoutingManager-test.js
```diff
@@ -312,5 +382,167 @@ describe('RoutingManager', () => {
...
+      fakeSearchBox.refine('Apple');
+
+      expect(router.write).toHaveBeenCalledTimes(1);
+      expect(router.write).toHaveBeenLastCalledWith({
+        query: 'Apple',
+      });
+
+      await runAllMicroTasks();
+
+      // Trigger getConfigurartion
```
francoischalifour:
```suggestion
      // Trigger getConfiguration
```

PR #3632, opened 2019-03-28
refactor(InstantSearch): remove dead variable

samouss:
**Summary**

This PR removes dead code inside `InstantSearch`. Here is the change:

- The `enhanceConfiguration` was previously used with the `SearchParameters` crafted from the URL with the [`URLSync`](https://github.com/algolia/instantsearch.js/blob/48a8d66746eb51fc0bc135f38182f84bc92ced37/src/lib/InstantSearch.js#L333). Under the hood the value was provided to [`getConfiguration`](https://github.com/algolia/instantsearch.js/blob/48a8d66746eb51fc0bc135f38182f84bc92ced37/src/lib/InstantSearch.js#L480-L483) along with the actual configuration. I didn't find any documentation about this argument. We drop the support of `URLSync` with InstantSearch 3.x. Since its introduction, the `routing` system always call the `enhanceConfiguration` function with `undefined` on start and an empty object on add/remove. The behavior is not consistent between the lifecycles. I've dropped the support for the argument which means that now the `getConfiguration` is always called with one argument. The impact should be none: we don't use it internally and it's not documented at all.

### DIFF
--- a/src/connectors/configure/connectConfigure.js
+++ b/src/connectors/configure/connectConfigure.js
@@ -57,7 +57,7 @@ export default function connectConfigure(renderFn = noop, unmountFn = noop) {
         return searchParameters => {
           // merge new `searchParameters` with the ones set from other widgets
           const actualState = this.removeSearchParameters(helper.getState());
-          const nextSearchParameters = enhanceConfiguration({})(
+          const nextSearchParameters = enhanceConfiguration()(
             { ...actualState },
             {
               getConfiguration: () => searchParameters,
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -155,7 +155,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
 
     // Init the widget directly if instantsearch has been already started
     if (this.started && Boolean(widgets.length)) {
-      this.searchParameters = this.widgets.reduce(enhanceConfiguration({}), {
+      this.searchParameters = this.widgets.reduce(enhanceConfiguration(), {
         ...this.helper.state,
       });
 
@@ -227,7 +227,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
       // re-compute remaining widgets to the state
       // in a case two widgets were using the same configuration but we removed one
       if (nextState) {
-        this.searchParameters = this.widgets.reduce(enhanceConfiguration({}), {
+        this.searchParameters = this.widgets.reduce(enhanceConfiguration(), {
           ...nextState,
         });
 
@@ -273,8 +273,6 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
       );
     }
 
-    let searchParametersFromUrl;
-
     if (this.routing) {
       const routingManager = new RoutingManager({
         ...this.routing,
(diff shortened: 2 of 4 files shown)


src/connectors/configure/connectConfigure.js
```diff
@@ -57,7 +57,7 @@ export default function connectConfigure(renderFn = noop, unmountFn = noop) {
         return searchParameters => {
           // merge new `searchParameters` with the ones set from other widgets
           const actualState = this.removeSearchParameters(helper.getState());
-          const nextSearchParameters = enhanceConfiguration({})(
+          const nextSearchParameters = enhanceConfiguration()(
```
Haroenv: if this argument isn't used anymore, why isn't it simply one function?

samouss: Yep indeed will drop the first level function.

src/lib/InstantSearch.js
```diff
@@ -426,32 +424,29 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
...
+    // always create a unified array for facets refinements
+    if (Array.isArray(a)) {
+      return union(a, b);
+    }
 
-      return undefined;
-    };
+    // avoid mutating objects
+    if (isPlainObject(a)) {
+      return mergeWith({}, a, b, customizer);
```
Haroenv: We already use Object.assign in other parts of the code right? Worth checking if we can use it here.

src/lib/InstantSearch.js
```diff
@@ -426,32 +424,29 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
...
+  const partialConfiguration = widgetDefinition.getConfiguration(configuration);
 
-      // avoid mutating objects
-      if (isPlainObject(a)) {
-        return mergeWith({}, a, b, customizer);
-      }
+  const customizer = (a, b) => {
+    // always create a unified array for facets refinements
+    if (Array.isArray(a)) {
+      return union(a, b);
```
```suggestion
      return [...a, ...b];
```

I think `b` could be not an array at this point, but it seems simpler (and less lodash)

tkrugg:
Apparently that variable appeared here: https://github.com/algolia/instantsearch.js/pull/1154
I'm not sure I understand well enough the test case the PR targeting but I might be lacking some context. Did you manage to understand it?

samouss:
@tkrugg Like I said in the PR description it was used inside the previous URLSync implementation called `URLSync`. The translation of the URL into `SearchParameters` was attached to this attribute inside the [`URLSync`](https://github.com/algolia/instantsearch.js/blob/48a8d66746eb51fc0bc135f38182f84bc92ced37/src/lib/url-sync.js#L115-L119) widget. With the `routing` system this step is performed directly inside the manager with the [`getAllSearchParameters`](getAllSearchParameters) function that is called at the configuration phase. The lifecycle `getConfiguration` is not tied to the URL anymore, it's only one purpose is to create the "configuration" (setup the facets, limit, etc...) for the widget.

src/lib/InstantSearch.js
samouss: I prefer to let this function as is (even though its implementation is not correct). This is a sensitive part of the code that I really don't want to alter at the moment. We'll get rid of the function with the federated search implementation anyway.

src/lib/InstantSearch.js
samouss: See https://github.com/algolia/instantsearch.js/pull/3632#discussion_r270817793.

PR #3645, opened 2019-04-02
feat(infiniteHits): add previous button

### LINKED ISSUE #2835: RFC: InfiniteHits should load previous pages
When reloading a page with urlSync and page > 0, the infinite hits doesn't load the previous hence it doesn't reproduce the exact same state. We decided to close https://github.com/algolia/instantsearch.js/issues/2750 but we've had other users asking for it.

### Proposed API

```diff
const widget = instantsearch.widgets.infiniteHits({ 
  container: string|HTMLElement, 
  templates: [InfiniteHitsTemplates], 
  showMoreLabel: [string], 
  transformData: [InfiniteHitsTransforms], 
  cssClasses: [InfiniteHitsCSSClasses], 
  escapeHits: [boolean],
+  loadPreviousPagesOnReload: [boolean=false]
}: InfiniteHitsWidgetOptions);
```

### Behaviour

The widget is not aware of the nature of the rendering. Therefore it should check at each rendering if the current page matches the cache of results. If it doesn't it should reload the missing pages. 
Cool side effect of that, is that it could handle the changes of hits per page.

### Raise your voice ;)

Because we might have a lot of pages,  we should probably add anchors, and go directly to the current page?

The name is long but explicit 🤔 

Please comment if you see a better name or you see any other issue.

yannickcr:
**Summary**

The [InfiniteHits widget](https://www.algolia.com/doc/api-reference/widgets/infinite-hits/js/) is used to display a list of results with a “Show more” button. The user can click on this button to display more results without leaving its current page context.

Each time the user click on the “Show more” button the user will navigate to a new page (provided that [the routing option is enabled in InstantSearch](https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/#widget-param-routing)).

However, in the case the user navigates directly to a particular page (direct link, previous button, page refresh) then the widget only displays the last result page.

This change add a previous button to the Infinite Hits widget to give the possibility to the user to load the previous result pages.

For implementation details see [the full RFC](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/infinite-hits-previous-page.md).

Before being mergeable, this change requires an update in https://github.com/algolia/instantsearch-specs to add the new CSS classes:

* [x] Merge https://github.com/algolia/instantsearch-specs/pull/101
* [x] Publish new InstantSearch.css release: [Published in release v7.2.0](https://github.com/algolia/instantsearch-specs/releases/tag/v7.2.0)

**Result**

A "with previous button enabled" story has been added for the Instant Hits widget.

https://deploy-preview-3645--instantsearchjs.netlify.com/stories/iframe.html?id=infinitehits--with-previous-button-enabled&page=3

fixes #2835


Haroenv:
It's pretty annoying that we chose to do the styling for each button individually, this way you need to update the style to see a correct styling here

src/widgets/infinite-hits/__tests__/infinite-hits-test.js
```diff
@@ -110,4 +110,38 @@ describe('infiniteHits()', () => {
...
+      results: { ...results, page: 0, nbPages: 2 },
+      state,
+    });
+    widget.render({
+      results: { ...results, page: 1, nbPages: 2 },
+      state,
+    });
+
+    expect(render).toHaveBeenCalledTimes(2);
+    expect(render.mock.calls[0][0]).toMatchSnapshot();
```
francoischalifour: I know these tests were originally written like this, but we can make them more clear now. We shouldn't overuse snapshots. Asserting only the snapshot doesn't translate the intent of the test. We should add an expectation with the actual `isFirstPage` value.

src/widgets/infinite-hits/__tests__/infinite-hits-test.js
```diff
@@ -110,4 +110,38 @@ describe('infiniteHits()', () => {
...
+    widget.init({ helper, instantSearchInstance: {} });
+
+    const state = { page: 1 };
+    widget.render({
+      results: { ...results, page: 1, nbPages: 2 },
+      state,
+    });
+
+    expect(render).toHaveBeenCalledTimes(1);
+    expect(render.mock.calls[0][0]).toMatchSnapshot();
```
francoischalifour: Same as above with `isFirstPage` to `false`.

src/widgets/infinite-hits/__tests__/defaultTemplates-test.js
```diff
@@ -22,4 +22,8 @@ describe('hits defaultTemplates', () => {
 
     expect(defaultTemplates.item(item)).toBe(expected);
   });
+
+  it('has a `showPreviousText` default template', () => {
+    expect(defaultTemplates.showPreviousText).toBe('Show previous results');
```
francoischalifour: We can use an inline snapshot here.

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -70,23 +70,39 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
   return (widgetParams = {}) => {
     const { escapeHTML = true, transformItems = items => items } = widgetParams;
     let hitsCache = [];
+    let firstReceivedPage = Infinity;
     let lastReceivedPage = -1;
 
-    const getShowMore = helper => () => helper.nextPage().search();
+    const getShowPrevious = helper => () => {
+      helper
+        .overrideStateWithoutTriggeringChangeEvent({
```
francoischalifour: Is there a reason for not using `setPage` as below? If so, can we add a comment explaining why?

src/components/InfiniteHits/__tests__/__snapshots__/InfiniteHits-test.js.snap
```diff
@@ -72,6 +72,96 @@ exports[`InfiniteHits markup should render <InfiniteHits /> on last page 1`] = `
 </div>
 `;
 
+exports[`InfiniteHits markup should render <InfiniteHits /> with "Show previous" button on first page 1`] = `
+<div
+  className="root"
+>
+  <button
+    className="undefined"
```
francoischalifour: We need to add the `loadPrevious` and `disabledLoadPrevious` CSS classes to the test.

src/components/InfiniteHits/__tests__/InfiniteHits-test.js
```diff
@@ -114,5 +122,71 @@ describe('InfiniteHits', () => {
...
+            showMoreText: 'showMoreText',
+            showPreviousText: 'showPreviousText',
+          },
+        },
+        cssClasses,
+      };
+
+      const tree = mount(<InfiniteHits {...props} />);
+
+      expect(tree).toMatchSnapshot();
```
francoischalifour: On top of the snapshot, we could add the expectations that are specific to this test. This would avoid missing errors like we have now about the CSS classes. For instance, check that the previous button is added, that it's disabled, that the CSS classes are here, etc. What do you think?

src/components/InfiniteHits/__tests__/__snapshots__/InfiniteHits-test.js.snap
```diff
@@ -72,6 +72,96 @@ exports[`InfiniteHits markup should render <InfiniteHits /> on last page 1`] = `
...
+  />
+</div>
+`;
+
+exports[`InfiniteHits markup should render <InfiniteHits /> with "Show previous" button on last page 1`] = `
+<div
+  className="root"
+>
+  <button
+    className=""
```
francoischalifour: Let's add the class name to the test to update this.

src/components/InfiniteHits/InfiniteHits.js
```diff
@@ -69,8 +86,11 @@ InfiniteHits.propTypes = {
   }).isRequired,
```
francoischalifour: Missing new CSS classes.

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.js
```diff
@@ -148,27 +155,65 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
-    expect(fourthRenderingOptions.results).toEqual(thirdResults);
+    const secondRenderingOptions = rendering.mock.calls[1][0];
+    const { showPrevious } = secondRenderingOptions;
+    expect(secondRenderingOptions.hits).toEqual(hits);
+    expect(secondRenderingOptions.results).toEqual(results);
+    showPrevious();
+    expect(helper.search).toHaveBeenCalledTimes(1);
+
+    // the results should be prepended if there is an decrement in page
+    const otherHits = [{ fake: 'data 2' }, { sample: 'infos 2' }];
```
francoischalifour: This could be called `previousHits`.

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.js
```diff
@@ -148,27 +155,65 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
+    const otherResults = new SearchResults(helper.state, [
```
francoischalifour: This could be called `previousResults`.

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -130,6 +147,22 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
...
+        return {
+          ...uiState,
+          page: page + 1,
+        };
+      },
+
+      getWidgetSearchParameters(searchParameters, { uiState }) {
+        const uiPage = uiState.page;
+        if (uiPage)
+          return searchParameters.setQueryParameter('page', uiState.page - 1);
```
francoischalifour: The variable `uiPage` is declared but not used here.

```suggestion
          return searchParameters.setQueryParameter('page', uiPage - 1);
```

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.js
```diff
@@ -148,27 +155,65 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     const thirdRenderingOptions = rendering.mock.calls[2][0];
     expect(thirdRenderingOptions.hits).toEqual([...hits, ...otherHits]);
     expect(thirdRenderingOptions.results).toEqual(otherResults);
+  });
+
+  it('Provides the hits and prepend results on previous page', () => {
```
francoischalifour:
```suggestion
  it('Provides the hits and prepends results on previous page', () => {
```

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.js
```diff
@@ -96,7 +103,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     });
   });
 
-  it('Provides the hits and the whole results', () => {
+  it('Provides the hits and accumulate results on next page', () => {
```
francoischalifour:
```suggestion
  it('Provides the hits and accumulates results on next page', () => {
```

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.js
```diff
@@ -148,27 +155,65 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
 
-    const fourthRenderingOptions = rendering.mock.calls[3][0];
-    expect(fourthRenderingOptions.hits).toEqual(thirdHits);
-    expect(fourthRenderingOptions.results).toEqual(thirdResults);
+    const secondRenderingOptions = rendering.mock.calls[1][0];
+    const { showPrevious } = secondRenderingOptions;
+    expect(secondRenderingOptions.hits).toEqual(hits);
+    expect(secondRenderingOptions.results).toEqual(results);
+    showPrevious();
+    expect(helper.search).toHaveBeenCalledTimes(1);
```
francoischalifour: Might be worth checking that the helper is on the right page before?

```js
expect(helper.getPage()).toBe(0);
```

src/connectors/infinite-hits/connectInfiniteHits.js
yannickcr: Yes, here we're using `overrideStateWithoutTriggeringChangeEvent` to update the internal state without updating the UI since we do not want to change the page number in the URL ([RFC](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/infinite-hits-previous-page.md#detailed-design)).

I will add a comment to make it clearer.

francoischalifour: I see! This is not caught by any tests though. We should add one.

src/components/InfiniteHits/__tests__/InfiniteHits-test.js
yannickcr: Yeah. I copied the previous test and adapted it for the previous button but maybe that's not the more effective way to test it.

What do you think of this?

```js
      // Test that we have two buttons (previous and next)
      expect(tree.find('button').length).toEqual(2);

      // Test that the first one (previous) is disabled since we already are on the first page
      expect(
        tree
          .find('button')
          .first()
          .props().disabled
      ).toBeDefined();

      // Test that the first one (previous) has the correct CSS classes
      expect(
        tree
          .find('button')
          .first()
          .hasClass('loadPrevious disabledLoadPrevious')
      ).toEqual(true);
```

(but in that case I think the snapshot test is no longer useful and could be removed)

francoischalifour: We should keep the `.loadMore` button apart from this test, so I wouldn't test that there are two button there.

This "jQuery-like" way of testing is very fragile IMO because it would break easily if we were to change the order of the generated DOM, and we would lose the intent of the test again.

We could test it more declaratively and keep the snapshot:

- expect that `.loadPrevious` exists
- expect that it has the `.disabledLoadPrevious` class
- expect that it has the disabled prop

That would be my guess. I might be wrong though, let me know what you think.

src/connectors/infinite-hits/connectInfiniteHits.js
yannickcr: Added the following assertion in `connectInfiniteHits-test.js` test suite:

```js
expect(helper.emit).not.toHaveBeenCalled();
```

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -130,6 +151,22 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
       dispose() {
         unmountFn();
       },
+
+      getWidgetState(uiState, { searchParameters }) {
```
samouss: We've implemented the `routing` but we always run it even when `showPrevious` is disabled. It means that previous implementation that use the widget have now their URL sync with it. We should only enable it when the previous button is present otherwise we might "break" their implementation.

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -70,23 +70,43 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
...
+        firstReceivedPage = helper.state.page;
+        lastReceivedPage = helper.state.page;
 
         renderFn(
           {
             hits: hitsCache,
             results: undefined,
+            showPrevious: this.showPrevious,
             showMore: this.showMore,
+            isFirstPage: true,
```
samouss: We don't know whether or not we are on the last page but we do for the first since it's always 0.

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -96,29 +116,30 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
       },
 
       render({ results, state, instantSearchInstance }) {
-        if (state.page === 0) {
```
samouss: We removed this part of the implementation but do we find why it was implemented?

src/components/InfiniteHits/InfiniteHits.js
```diff
@@ -64,13 +81,18 @@ InfiniteHits.propTypes = {
...
     item: PropTypes.string.isRequired,
+    loadPrevious: PropTypes.string.isRequired,
+    disabledLoadPrevious: PropTypes.string.isRequired,
     loadMore: PropTypes.string.isRequired,
     disabledLoadMore: PropTypes.string.isRequired,
   }).isRequired,
   hits: PropTypes.array.isRequired,
   results: PropTypes.object.isRequired,
+  hasShowPrevious: PropTypes.bool.isRequired,
+  showPrevious: PropTypes.func,
```
samouss: Both functions are required, we always provide both function.

stories/infinite-hits.stories.js
```diff
@@ -68,4 +69,34 @@ storiesOf('InfiniteHits', module)
...
+            showPrevious: true,
+            templates: {
+              item: '{{name}}',
+            },
+          })
+        );
+      },
+      {
+        routing: {
+          stateMapping: {
```
samouss: We should update this part to make the story usable from the Storybook. We can implement a small memory router that saves the different `routeState`. It would be nice to also logs the value into the [action](https://github.com/algolia/instantsearch.js/blob/577e4ccb33155d0d4678b04505132d39a2918f22/.storybook/decorators.ts#L37-L46) tab (like we do for the rest).

src/components/InfiniteHits/InfiniteHits.js
yannickcr: `showMore` was not marked as required so I did not maked `showPrevious` as required.

If I update `showPrevious`, do I also update `showMore` to mark it as required?

src/connectors/infinite-hits/connectInfiniteHits.js
yannickcr: Not sure why (seems to be here from the first Instant Hits implementation https://github.com/algolia/instantsearch.js/commit/cdf867544a9acb9a4973f700fdc5230393ca3d51).

It was flushing the cache if you were reaching the first page. But even in the original implementation the widget had a weird behavior with this if you were using your browser to go back to previous pages: when reaching the first page the results list was emptied and only the last page was displayed.

This was even more problematic with the previous button, so I removed it.

src/components/InfiniteHits/InfiniteHits.js
samouss: Yep!

src/connectors/infinite-hits/connectInfiniteHits.js
Haroenv: it's indeed already in the initial implementation: https://github.com/algolia/instantsearch.js/pull/1961/files#diff-db508ea4bf1e589da868d7cc605ac30bR77

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -68,7 +68,11 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
   checkRendering(renderFn, withUsage());
 
   return (widgetParams = {}) => {
-    const { escapeHTML = true, transformItems = items => items } = widgetParams;
+    const {
+      escapeHTML = true,
+      transformItems = items => items,
+      showPrevious,
```
samouss: We should now provide the default value at this level rather than inside the widget.

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -154,14 +158,19 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
 
       getWidgetState(uiState, { searchParameters }) {
         const page = searchParameters.page;
-        if (page === 0 || page + 1 === uiState.page) return uiState;
+        if (!showPrevious || page === 0 || page + 1 === uiState.page) {
```
samouss: We should cover this case with a test.

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -154,14 +158,19 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
...
+          return uiState;
+        }
         return {
           ...uiState,
           page: page + 1,
         };
       },
 
       getWidgetSearchParameters(searchParameters, { uiState }) {
+        if (!showPrevious) {
```
samouss: We should cover this case with a test.

stories/infinite-hits.stories.js
```diff
@@ -68,4 +70,35 @@ storiesOf('InfiniteHits', module)
...
+          router: new MemoryRouter({ page: 3 }),
+          stateMapping: {
+            stateToRoute(uiState) {
+              const query = qs.parse(location.search.slice(1));
+              return {
+                ...query,
+                page: uiState.page,
+              };
+            },
+            routeToState: routeState => routeState,
```
samouss: We can filter out the value append by storybook.

stories/infinite-hits.stories.js
```diff
@@ -68,4 +70,35 @@ storiesOf('InfiniteHits', module)
...
+            },
+          })
+        );
+      },
+      {
+        routing: {
+          router: new MemoryRouter({ page: 3 }),
+          stateMapping: {
+            stateToRoute(uiState) {
+              const query = qs.parse(location.search.slice(1));
```
samouss: This line is useless with the `MemoryRouter`.

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -130,6 +155,27 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
...
+          page: page + 1,
+        };
+      },
+
+      getWidgetSearchParameters(searchParameters, { uiState }) {
+        if (!showPrevious) {
+          return searchParameters;
+        }
+        const uiPage = uiState.page;
+        if (uiPage)
```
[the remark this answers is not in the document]
yannickcr: I don't mind :innocent: 

But IMHO it should be something picked up by the linter rules :wink:

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -130,6 +155,27 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
       dispose() {
         unmountFn();
       },
+
+      getWidgetState(uiState, { searchParameters }) {
+        const page = searchParameters.page;
+        if (!showPrevious || page === 0 || page + 1 === uiState.page) {
```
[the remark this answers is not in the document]
yannickcr: Same as above :wink:

src/widgets/infinite-hits/__tests__/infinite-hits-test.js
```diff
@@ -110,4 +112,41 @@ describe('infiniteHits()', () => {
...
+      results: { ...results, page: 1, nbPages: 2 },
+      state,
+    });
+
+    expect(render).toHaveBeenCalledTimes(2);
+
+    expect(render.mock.calls[0][0].props.isFirstPage).toEqual(true);
+    expect(render.mock.calls[0][1]).toEqual(container);
+
+    expect(render.mock.calls[0][0].props.isFirstPage).toEqual(true);
```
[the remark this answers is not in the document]
yannickcr: Ho yeah, small error here.

I updaded the test to use the intermediate variables, it is more readable that way :+1:

src/connectors/infinite-hits/connectInfiniteHits.js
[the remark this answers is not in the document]
francoischalifour: That's for sure!

We have [the `curly` rule in our ESLint config](https://github.com/algolia/eslint-config-algolia/blob/f1039d64a03ccbbcee37b7089a1e0ec97c1a1500/rules/base.js#L45) but it's apparently disabled by the Prettier ESLint config. That's something we should investigate.

src/connectors/infinite-hits/connectInfiniteHits.js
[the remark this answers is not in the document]
francoischalifour: I agree! I'm not aware of any tools for that however. Do you know any?

src/connectors/infinite-hits/connectInfiniteHits.js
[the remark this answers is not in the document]
yannickcr: Apparently the `curly` rule is part of what `eslint-config-prettier` calls ["Special rules"](https://github.com/prettier/eslint-config-prettier#special-rules) and have some options that can conflict with prettier. We can re-enable it manually if we do not uses these options.

src/connectors/infinite-hits/connectInfiniteHits.js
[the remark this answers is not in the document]
yannickcr: Yes, it can be done using the [`padding-line-between-statements`](https://eslint.org/docs/rules/padding-line-between-statements) ESLint rule.

PR #3646, opened 2019-04-03
chore(queryRules): update stories

francoischalifour:
When playing around with Query Rules widgets and writing their documentation (algolia/doc#2896), I saw a few patterns emerging.

That's my proposition for their stories:

- I don't think it's nice to showcase mutating the structure of the items:

```js
transformItems(item[]): item
```

- `transformItems` can be used to filter out custom data to avoid having to add checks to the template

These stories are released to the production storybook once merged to master (when we release a new library version for instance).

If you agree with these changes we can release the new minor version.


francoischalifour:
Closing in favor of a future PR.

PR #3648, opened 2019-04-03
fix(connectQueryRules): improve tracked refinement type

francoischalifour:
The tracked filters assumed a refinement can be a `number` or a `string`. It can also be a `boolean`.

See:

- https://github.com/algolia/react-instantsearch/pull/2258/files#diff-61853eb4221589c33badeb16d8725534R12
- https://github.com/algolia/angular-instantsearch/pull/483#discussion_r271643098

### DIFF
--- a/src/connectors/query-rules/connectQueryRules.ts
+++ b/src/connectors/query-rules/connectQueryRules.ts
@@ -18,10 +18,12 @@ import {
   NumericRefinement as InternalNumericRefinement,
 } from '../../lib/utils/getRefinements';
 
+type TrackedFilterRefinement = string | number | boolean;
+
 export type ParamTrackedFilters = {
   [facetName: string]: (
-    facetValues: Array<string | number>
-  ) => Array<string | number>;
+    facetValues: TrackedFilterRefinement[]
+  ) => TrackedFilterRefinement[];
 };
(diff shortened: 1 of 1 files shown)


src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -18,10 +18,12 @@ import {
   NumericRefinement as InternalNumericRefinement,
 } from '../../lib/utils/getRefinements';
 
+type TrackedFilterRefinement = string | number | boolean;
+
 export type ParamTrackedFilters = {
   [facetName: string]: (
-    facetValues: Array<string | number>
-  ) => Array<string | number>;
+    facetValues: TrackedFilterRefinement[]
```
tkrugg: I think maybe this parameter should be optional. We want to consider this valid:
```ts
{ genre: () => ['Comedy'] }
```

francoischalifour: That's a good point!

Haroenv:
(looks like tsc doesn't like your change yet)

src/connectors/query-rules/connectQueryRules.ts
francoischalifour: Actually that's wrong. This function gets called with the computed tracked filters (it's never `undefined`).

PR #3649, opened 2019-04-03
refactor(typescript): migrate lib files

francoischalifour:
This migrates some lib files to TypeScript and add typings for Algolia hits.


src/helpers/highlight.ts
```diff
@@ -2,15 +2,24 @@ import { getPropertyByPath } from '../lib/utils';
 import { TAG_REPLACEMENT } from '../lib/escape-highlight';
 import { component } from '../lib/suit';
 
+export type HighlightOptions = {
+  attribute: string;
+  highlightedTagName?: string;
+  hit: object;
```
Haroenv:
```suggestion
  hit: {
    _highlightResult: {
      [attribute: string]: {
        value: string;
      },
    },
  };
```

This way you don't need to do "as string" I think

francoischalifour: Actually I created the type `Hit` so I can reuse it here 👍

francoischalifour: I don't necessarily need the "as string" but `getPropertyBypath` doesn't ensure the return type. It gives any type given by the path that resolved in the object. That's why I explicitly casted it.

src/helpers/snippet.ts
```diff
@@ -1,16 +1,26 @@
...
-  attribute,
-  highlightedTagName = 'mark',
-  hit,
-} = {}) {
+export default function snippet(
+  {
+    attribute,
+    highlightedTagName = 'mark',
+    hit,
+  }: SnippetOptions = {} as SnippetOptions
```
Haroenv: if it is `= {}` then doing `as SnippetOptions` can cause undefined errors. In fact I think it probably should be removed like this:

```suggestion
  }: SnippetOptions
```

src/lib/createHelpers.ts
```diff
@@ -21,9 +31,9 @@ The highlight helper expects a JSON object of the format:
 { "attribute": "name", "highlightedTagName": "mark" }`);
       }
     },
-    snippet(options, render) {
+    snippet(options: string, render: HoganRenderer) {
       try {
-        const snippetOptions = JSON.parse(options);
+        const snippetOptions: Omit<SnippetOptions, 'hit'> = JSON.parse(options);
```
Haroenv: Why do we do this? To copy the value or something else?

src/lib/escape-highlight.ts
```diff
@@ -13,7 +14,7 @@ export const TAG_REPLACEMENT = {
   highlightPostTag: '</mark>',
 };
 
-function replaceTagsAndEscape(value) {
+function replaceTagsAndEscape(value: string): string {
```
Haroenv: I'm still not convinced of writing the return type here, it can easily be inferred from the fact that .replace is called on escape (which always returns a string)

src/lib/createHelpers.ts
francoischalifour: In the `snippet` (and `highlight`) Hogan helper, you don't pass the `hit` because you don't have access to it and we can infer it. Therefore the options that you pass are the snippet options, without the hit.

src/lib/escape-highlight.ts
francoischalifour: We explained it here: https://github.com/algolia/instantsearch.js/pull/3617#discussion_r269201029.

Let's chat about it IRL if you're not convinced or need more details.

src/helpers/snippet.ts
francoischalifour: Good point – will remove this default value.

src/lib/createHelpers.ts
Haroenv: My question was about the JSON.parse

francoischalifour:
I experimented with the following rule but it turns out to be very annoying in certain situations. Even though the types are declared in interfaces, we need to rewrite them every time. IMO we shouldn't go with this rule for now.

```
"typedef": [true, "call-signature", "property-declaration"]
```

francoischalifour:
I updated this PR so that it reflects only changes in the lib files.

src/lib/escape-highlight.ts
samouss: One of the purposes of the types is to reduce the cognitive overhead that reading code might produce. It's also about the contract, with the type definition the intent of the function is clearer.

src/types/instantsearch.ts
```diff
@@ -10,6 +12,50 @@ export type SearchParameters = any;
 
 export type SearchResults = any;
 
+export type Hit = {
+  [attribute: string]: any;
+  objectID: string;
+  _highlightResult?: {
+    [attribute: string]: {
```
samouss: This definition is not 100% correct. It's possible to have a recursive object or even an array.

```json
{
  "categories": [
    { "value": "", "matchLevel": "none" },
    { "value": "", "matchLevel": "none" }
  ]
}
```

```json
{
  "hierarchical": {
    "lvl0": {
      "value": "",
      "matchLevel": "none"
    }
  }
}
```

Haroenv: would make sense to add HighlightResult as a type and then write all three possibilities with the result as an end factor

francoischalifour: Thanks! I fixed that in d551209ae664e717685ef7d43150d83c716f679c for both highlight and snippet.

src/types/instantsearch.ts
```diff
@@ -12,23 +12,37 @@ export type SearchParameters = any;
 
 export type SearchResults = any;
 
+type HitAttributeHighlightResult = {
+  value: string;
+  matchLevel: 'none' | 'partial' | 'full';
+  matchedWords: string[];
+  fullyHighlighted?: boolean;
```
samouss: Is it really optional?

francoischalifour: When it's not fully highlighted, the key is not returned:

image

When it is fully highlighted, the attribute is `true`:

image

Haroenv: wouldn't in that case the type be `fullyHighlighted?: true`? Although I guess technically it could still be false, and best to leave that case open

PR #3650, opened 2019-04-04
fix(queryRuleCustomData): add default template

francoischalifour:
This adds a default template to the `queryRuleCustomData` widget. This default template is a JSON dump of the custom data returned by the connector. It can help users understand what data structure to interact with before creating a custom template.

This is the behavior we went for in the [Vue InstantSearch implementation](https://github.com/algolia/vue-instantsearch/pull/652) and in the [Angular InstantSearch implementation](https://github.com/algolia/angular-instantsearch/pull/482).

[See story →](https://deploy-preview-3650--instantsearchjs.netlify.com/stories/?path=/story/queryrulecustomdata--without-template)


francoischalifour:
@Haroenv This is the template we have originally for the [hits](https://github.com/algolia/instantsearch.js/blob/e16ad5759c9f2f27b64017a3acf7fa92b5a8a6c6/src/widgets/hits/defaultTemplates.js) for instance. I think should keep this pattern in the project.

PR #3651, opened 2019-04-05
fix(types): improve InstantSearch types

francoischalifour:
This completes the InstantSearch, Helper, Client and Widget types, and updates the code impacted.


src/types/widget.ts
```diff
@@ -6,18 +6,55 @@ import {
...
+type UiState = {
+  [stateKey: string]: any;
+};
+
 export interface Widget {
-  init(options: WidgetArgs): void;
-  render(options: WidgetArgs): void;
-  dispose(options: any): SearchParameters;
-  getConfiguration?(previousConfiguration: SearchParameters): SearchParameters;
+  init?(options: InitOptions): void;
```
francoischalifour: This definition is not entirely true because a widget must at least have an `init` or a `render` method, or both ([see reference](https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-own-widgets/js#the-widget-lifecycle-and-api)).

This is however the easiest we can do right now, because it becomes complex to type otherwise.

src/widgets/query-rule-custom-data/__tests__/query-rule-custom-data-test.ts
```diff
@@ -59,7 +64,11 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/query-rule-
           container: document.createElement('div'),
         });
 
-        widget.init({ helper, state: helper.state, instantSearchInstance: {} });
+        widget.init!({
```
francoischalifour: Since there's no guarantee that the `init` or `render` method of the widget exists, we need to force them with `!` in the tests.

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -227,7 +234,13 @@ const connectQueryRules: QueryRulesConnector = (render, unmount = noop) => {
         );
       },
 
-      render({ results, instantSearchInstance }) {
+      render({
+        results,
+        instantSearchInstance,
+      }: {
+        results: SearchResults;
```
francoischalifour: It's necessary at this point to add the definition again to remove the `undefined` type to access values.

src/widgets/query-rule-custom-data/__tests__/query-rule-custom-data-test.ts
yannickcr: It is not possible to solve this using [conditional types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#conditional-types) to get the correct return type from `queryRuleCustomData`? (my fear is that the users get the same issue when calling `widget.init`)

francoischalifour: We tried some stuff like conditional types with @samouss but each solution brought other complications. @samouss has more input than me on this.

I don't think users have to manipulate `widget.init` except in tests as we do.

Another solution would be to enforce `init`, `render` and `dispose` from now on. WDYT?

francoischalifour:
I updated this PR so that it reflects only changes in the source files.

src/widgets/query-rule-custom-data/__tests__/query-rule-custom-data-test.ts
yannickcr: Well, I still lack context on the lib to see the impact on enforcing these methods.

But I would be very interested to see what complications you encountered with @samouss with conditional types (for my own curiosity :wink: )

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -191,7 +190,15 @@ const connectQueryRules: QueryRulesConnector = (render, unmount = noop) => {
...
-      init({ helper, state, instantSearchInstance }) {
+      init({
+        helper,
+        state,
+        instantSearchInstance,
+      }: {
+        helper: Helper;
+        state: HelperState;
+        instantSearchInstance: InstantSearch;
+      }) {
```
samouss: TypeScript is not able to infer it from the `Widget`?

src/connectors/query-rules/connectQueryRules.ts
samouss: I don't understand why because `results`, `instantSearchInstance` are required at this point.

src/types/instantsearch.ts
```diff
@@ -22,3 +25,15 @@ export type NumericRefinement = {
...
 
 export type Refinement = FacetRefinement | NumericRefinement;
+
+export type Client = AlgoliaSearchClient;
+
+export type Helper = AlgoliaSearchHelper;
+
+// We complete the original helper state with the custom SearchParameters
+// not yet typed in the original `algoliasearch-helper`.
+export type HelperState = AlgoliaSearchHelper['state'] & SearchParameters;
```
samouss: Both `HelperState` and `SearchParameters ` describe the same object. Why do we have both?

src/types/widget.ts
```diff
@@ -6,18 +6,55 @@ import {
...
+interface RenderOptions {
   instantSearchInstance: InstantSearch;
+  templatesConfig?: object;
+  results?: SearchResults;
+  state?: HelperState;
+  helper?: Helper;
+  searchMetadata?: {
+    isSearchStalled: boolean;
+  };
+  createURL?(state: HelperState): string;
```
samouss: Almost all the attributes are required.

https://github.com/algolia/instantsearch.js/blob/d553502f13e7eb9b6d423f85983c19e5732cc7a5/src/lib/InstantSearch.js#L390-L400

src/types/widget.ts
```diff
@@ -6,18 +6,55 @@ import {
...
 
-interface WidgetArgs {
-  helper: Helper;
+interface InitOptions {
+  instantSearchInstance: InstantSearch;
   state?: HelperState;
-  results?: SearchResults;
+  helper?: Helper;
+  templatesConfig?: object;
+  createURL?(state: HelperState): string;
```
samouss: Almost all the attributes are required.

https://github.com/algolia/instantsearch.js/blob/d553502f13e7eb9b6d423f85983c19e5732cc7a5/src/lib/InstantSearch.js#L414-L421

src/types/widget.ts
```diff
@@ -6,18 +6,55 @@ import {
...
+  state?: HelperState;
+  helper?: Helper;
+  searchMetadata?: {
+    isSearchStalled: boolean;
+  };
+  createURL?(state: HelperState): string;
+}
+
+interface DisposeOptions {
+  helper?: Helper;
```
samouss: `helper` is required.

https://github.com/algolia/instantsearch.js/blob/d553502f13e7eb9b6d423f85983c19e5732cc7a5/src/lib/InstantSearch.js#L222-L225

src/types/widget.ts
```diff
@@ -6,18 +6,55 @@ import {
...
+};
+
 export interface Widget {
-  init(options: WidgetArgs): void;
-  render(options: WidgetArgs): void;
-  dispose(options: any): SearchParameters;
-  getConfiguration?(previousConfiguration: SearchParameters): SearchParameters;
+  init?(options: InitOptions): void;
+  render?(options: RenderOptions): void;
+  dispose?(options: DisposeOptions): HelperState;
```
samouss: The function can return `undefined`.

https://github.com/algolia/instantsearch.js/blob/d553502f13e7eb9b6d423f85983c19e5732cc7a5/src/lib/InstantSearch.js#L222-L229

src/types/widget.ts
```diff
@@ -6,18 +6,55 @@ import {
...
+  getConfiguration?(
+    previousConfiguration?: SearchParameters
```
samouss: The function does not accept a `SearchParameters` but a plain object that represents `SearchParameters`. It's the same for the return type.

src/widgets/query-rule-custom-data/__tests__/query-rule-custom-data-test.ts
samouss: > But I would be very interested to see what complications you encountered with @samouss with conditional types (for my own curiosity 😉 )

We've tried with both union types and conditional types. With the latter we're almost there but not yet. Once we put the widgets inside an array we're not able to find which kind of widget it is. It means that we only have access to all the common attributes of each kind of widget (which makes perfect sense). Here is [an example](https://www.typescriptlang.org/play/index.html#src=type%20InitOptions%20%3D%20any%3B%0D%0Atype%20RenderOptions%20%3D%20any%3B%0D%0Atype%20DisposeOptions%20%3D%20any%3B%0D%0Atype%20SearchParameters%20%3D%20any%3B%0D%0A%0D%0Aexport%20interface%20BaseWidget%20%7B%0D%0A%20%20dispose%3F()%3A%20SearchParameters%3B%0D%0A%20%20getConfiguration%3F()%3A%20SearchParameters%3B%0D%0A%7D%0D%0A%0D%0Aexport%20interface%20WidgetWithInit%20extends%20BaseWidget%20%7B%0D%0A%20%20init(options%3A%20InitOptions)%3A%20void%3B%0D%0A%7D%0D%0A%0D%0Aexport%20interface%20WidgetWithRender%20extends%20BaseWidget%20%7B%0D%0A%20%20render(options%3A%20RenderOptions)%3A%20void%3B%0D%0A%7D%0D%0A%0D%0Aexport%20interface%20WidgetWithInitRender%20extends%20WidgetWithInit%2C%20WidgetWithRender%20%7B%0D%0A%0D%0A%7D%0D%0A%0D%0Aexport%20type%20Widget%20%3D%0D%0A%20%20%7C%20WidgetWithInit%0D%0A%20%20%7C%20WidgetWithRender%0D%0A%20%20%7C%20WidgetWithInitRender%0D%0A%0D%0Aexport%20type%20CreateWidget%3CTWidget%20extends%20Widget%20%3D%20Widget%3E%20%3D%20()%20%3D%3E%20%0D%0A%20%20TWidget%20extends%20WidgetWithInitRender%20%3F%20WidgetWithInitRender%20%3A%0D%0A%20%20TWidget%20extends%20WidgetWithInit%20%3F%20WidgetWithInit%20%3A%0D%0A%20%20WidgetWithRender%0D%0A%0D%0Aconst%20createWidget%3A%20CreateWidget%3CWidgetWithInitRender%3E%20%3D%20()%20%3D%3E%20(%7B%0D%0A%20%20init()%20%7B%7D%2C%0D%0A%20%20render()%20%7B%7D%2C%0D%0A%7D)%3B%0D%0A%0D%0Aconst%20createWidgetInit%3A%20CreateWidget%3CWidgetWithInit%3E%20%3D%20()%20%3D%3E%20(%7B%0D%0A%20%20init()%20%7B%7D%2C%0D%0A%7D)%3B%0D%0A%0D%0Aconst%20createWidgetRender%3A%20CreateWidget%3CWidgetWithRender%3E%20%3D%20()%20%3D%3E%20(%7B%0D%0A%20%20render()%20%7B%7D%2C%0D%0A%7D)%3B%0D%0A%0D%0Aconst%20w%20%3D%20createWidget()%0D%0Aconst%20wInit%20%3D%20createWidgetInit()%3B%0D%0Aconst%20wRender%20%3D%20createWidgetRender()%3B%0D%0Aconst%20ws%20%3D%20%5Bw%2C%20wInit%2C%20wRender%5D%3B%0D%0A%0D%0Aw.init(%7B%7D)%0D%0Aw.render(%7B%7D)%0D%0A%0D%0AwInit.init(%7B%7D)%0D%0AwInit.render(%7B%7D)%0D%0A%0D%0AwRender.init(%7B%7D)%0D%0AwRender.render(%7B%7D)%0D%0A%0D%0Aws.forEach(w%20%3D%3E%20%7B%0D%0A%20%20if%20(typeof%20w.init%20!%3D%3D%20'function')%20%7B%0D%0A%20%20%20%20w.init()%3B%0D%0A%20%20%7D%0D%0A%7D)). Note that I might be completely wrong about the subject though.

> my fear is that the user get the same issue when calling `widget.init`

This is not an issue because we call the methods for the users. They never have to call it directly.

> Another solution would be to enforce init, render and dispose from now on. WDYT?

Yes that would solve the issue but I would avoid to do it right now. I would rather suggest to take this into consideration once we revamp the core of InstantSearch. Until now we can treat them as optional inside the codebase it was (almost) already the case.

https://github.com/algolia/instantsearch.js/blob/ba6e2e61c16c1b524d24cf72a66cbb354cf81b55/src/lib/InstantSearch.js#L387

https://github.com/algolia/instantsearch.js/blob/ba6e2e61c16c1b524d24cf72a66cbb354cf81b55/src/lib/InstantSearch.js#L413

src/connectors/query-rules/connectQueryRules.ts
francoischalifour: Now that we made them required, no need for this!

src/connectors/query-rules/connectQueryRules.ts
francoischalifour: Now that we made them required, no need for this!

src/types/instantsearch.ts
francoischalifour: I kept only `SearchParameters` in the end.

PR #3669, opened 2019-04-08
fix(routing): apply windowTitle on first load

### LINKED ISSUE #3667: windowTitle does not trigger when page loads
**Describe the bug 🐛**
The `windowTitle` option for routing does not trigger when the page first loads.

**To Reproduce 🔍**
This issue can be easily reproduced via official SEO friendly URLs demo

1. Open demo via [CodeSandbox](https://codesandbox.io/s/github/algolia/doc-code-samples/tree/master/InstantSearch.js/routing-seo-friendly)
2. Run the demo and click "Open in New Window"
3. Type something to search, notice the title change
4. Refresh the page using browser reload.
5. Title does not change.

**Expected behavior 💭**
The page title should contains correct queries from `windowTitle` after page reload.

**Screencast 🎬**
https://www.useloom.com/share/96c3c8f3329041c7bc6d74e232812ad4

**Environment:**
 - OS: macOS
 - Browser: Google Chrome
 - Version: 73.0.3683.86

**Additional context**

francoischalifour:
_This PR is a fix for #3667._

The [`windowTitle`](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/js/#windowtitle-functionroutestate-string) param of the routing option didn't apply on the first page load. 

I fixed this by setting the document title also in the `RoutingManager` constructor, based on the URL query parameters which represent the route state.

This feature remained untested so I added a test for the first page load.

Note that Jest doesn't support updating `window.location` in tests (see facebook/jest#890 and facebook/jest#5124). I therefore needed to introduce two new JSDom dev dependencies to change `window.location` in separate tests, which give access to `jsdom.reconfigure({ url })`. I believe these dependencies will be needed in the feature while we refactor/test/reimplement the routing option.

The [Jest team has at some point considered giving access to `reconfigure`](https://github.com/facebook/jest/issues/890#issuecomment-352585259). This would mean that the new dependencies become unnecessary.

Closes #3667.

### DIFF
--- a/src/lib/__tests__/RoutingManager-test.js
+++ b/src/lib/__tests__/RoutingManager-test.js
@@ -1,5 +1,7 @@
+/* globals jsdom */
 import instantsearch from '../main';
 import RoutingManager from '../RoutingManager';
+import historyRouter from '../routers/history';
 
 const runAllMicroTasks = () => new Promise(setImmediate);
 
@@ -543,6 +545,56 @@ describe('RoutingManager', () => {
       expect(router.write).toHaveBeenLastCalledWith({
         query: 'Apple',
       });
+    });
+  });
+
+  describe('windowTitle', () => {
+    let setWindowTitle = jest.fn();
+
+    Object.defineProperty(window.document, 'title', {
+      get() {
+        return '';
+      },
+      set(value) {
+        setWindowTitle(value);
+      },
+    });
+
+    beforeEach(() => {
+      setWindowTitle = jest.fn();
+    });
+
+    test('should update the window title with URL query params on first render', async () => {
+      jsdom.reconfigure({
+        url: 'https://website.com/?query=query',
+      });
+
+      const searchClient = createFakeSearchClient();
+      const stateMapping = createFakeStateMapping();
+      const router = historyRouter({
+        windowTitle(routeState) {
+          return `Searching for "${routeState.query}"`;
+        },
+      });
+
+      const search = instantsearch({
+        indexName: 'instant_search',
+        searchClient,
+        routing: {
+          router,
+          stateMapping,
+        },
+      });
+
+      const fakeSearchBox = createFakeSearchBox();
+
+      search.addWidget(fakeSearchBox);
+      search.start();
+
+      await runAllMicroTasks();
+
+      expect(setWindowTitle).toHaveBeenCalledTimes(1);
+      expect(setWindowTitle).toHaveBeenLastCalledWith('Searching for "query"');
     });
   });
 });
--- a/src/lib/__tests__/RoutingManager-test.js
+++ b/src/lib/__tests__/RoutingManager-test.js
@@ -549,26 +549,12 @@ describe('RoutingManager', () => {
   });
 
   describe('windowTitle', () => {
-    let setWindowTitle = jest.fn();
-
-    Object.defineProperty(window.document, 'title', {
-      get() {
-        return '';
-      },
-      set(value) {
-        setWindowTitle(value);
-      },
-    });
-
-    beforeEach(() => {
-      setWindowTitle = jest.fn();
-    });
-
     test('should update the window title with URL query params on first render', async () => {
       jsdom.reconfigure({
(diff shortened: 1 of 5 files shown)


src/lib/__tests__/RoutingManager-test.js
```diff
@@ -545,4 +547,54 @@ describe('RoutingManager', () => {
...
+    let setWindowTitle = jest.fn();
+
+    Object.defineProperty(window.document, 'title', {
+      get() {
+        return '';
+      },
+      set(value) {
+        setWindowTitle(value);
+      },
+    });
```
tkrugg: There is an easier way to mock getters/setters
https://jestjs.io/docs/en/jest-object.html#jestspyonobject-methodname-accesstype

The nice thing about this way, is that you can later restore the mock, which you don't at the moment.

francoischalifour: That's great, I didn't think of this usage. Thanks!

src/lib/routers/history.js
```diff
@@ -44,6 +50,13 @@ class BrowserHistory {
     this.writeDelay = writeDelay;
     this._createURL = createURL;
     this.parseURL = parseURL;
+
+    const routeState = this.parseURL({
```
samouss: You can directly call `this.read()` it would avoid to duplicate the call.

src/lib/routers/history.js
```diff
@@ -44,6 +50,13 @@ class BrowserHistory {
...
+      qsModule: qs,
+      location: window.location,
+    });
+    const title = this.windowTitle && this.windowTitle(routeState);
```
samouss: You can inline the call `this.read()` inside `this.windowTitle` (or make a condition) to avoid a useless parse when `windowTitle` is not provided.

src/lib/routers/history.js
```diff
@@ -15,6 +15,12 @@ function defaultParseURL({ qsModule, location }) {
   return qsModule.parse(location.search.slice(1));
 }
 
+function setWindowTitle(title) {
+  if (title) {
+    window.document.title = title;
```
Haroenv: we aren't checking that these variables are available, however that's the same in other parts of this module too

samouss: You mean `window`?

src/lib/routers/history.js
```diff
@@ -44,6 +50,9 @@ class BrowserHistory {
     this.writeDelay = writeDelay;
     this._createURL = createURL;
     this.parseURL = parseURL;
+
+    const title = this.windowTitle && this.windowTitle(this.read());
```
samouss:
```suggestion
    const title = this.windowTitle && this.windowTitle(this.read());
    
```

src/lib/routers/history.js
Haroenv: yes, if window is not defined (i.e. in server side rendering) this code will throw

francoischalifour: We've got at least two more occurrences of `window` without check:

- https://github.com/algolia/instantsearch.js/blob/577e4ccb33155d0d4678b04505132d39a2918f22/src/connectors/numeric-menu/connectNumericMenu.js#L133
- https://github.com/algolia/instantsearch.js/blob/577e4ccb33155d0d4678b04505132d39a2918f22/src/widgets/analytics/analytics.js#L176

We can handle those when we plan to be SSR-ready.

src/lib/__tests__/RoutingManager-test.js
```diff
@@ -1,5 +1,7 @@
+/* globals jsdom */
```
tkrugg: what is this used for? looks like your tests are passing without it.

francoischalifour: `jest-environment-jsdom-global` makes `jsdom` available in our tests ([read more](https://github.com/simon360/jest-environment-jsdom-global#readme)). However, ESLint is not aware of this global variable because this is not part of [`eslint-plugin-jest`](https://github.com/jest-community/eslint-plugin-jest).

PR #3670, opened 2019-04-09
fix(queryRules): fix types and stories

francoischalifour:
Until we have a proper way to provide generic types to InstantSearch.js widgets, the most accurate type that we can provide for items is `any`. The `object` type wasn't correct.

This PR now manipulates the source of the widgets in the stories, which provides typings.


stories/query-rule-custom-data.stories.ts
```diff
@@ -1,6 +1,7 @@
 import { storiesOf } from '@storybook/html';
 import { withHits } from '../.storybook/decorators';
 import moviesPlayground from '../.storybook/playgrounds/movies';
+import { queryRuleCustomData } from '../src/widgets';
```
samouss: We didn't fix the issue with this import, the widgets is inferred as `any`.

PR #3671, opened 2019-04-09
fix(routing): support parsing URLs with up to 100 refinements

francoischalifour:
## Description

An issue raised by a customer is that when selecting more than 20 values with the `routing` option activated, the data structure returned by the `parseURL` function changes.

`qs` by default converts arrays with more than 20 items to an object. We want to avoid this because the data structure manipulated can therefore vary.

Setting the limit to `100` seems a good number because the engine's default is 100. It can go up to 1000 but it is very unlikely to select more than 100 items in the UI.

## Reference

- `qs` doc: https://github.com/ljharb/qs#parsing-arrays
- Algolia `maxValuesPerFacet` doc: https://www.algolia.com/doc/api-reference/api-parameters/maxValuesPerFacet/
- Original customer issue: https://algolia.slack.com/archives/C2XP6HBKQ/p1553005092109200


src/lib/routers/history.js
```diff
@@ -12,7 +12,14 @@ function defaultCreateURL({ qsModule, routeState, location }) {
...
 function defaultParseURL({ qsModule, location }) {
-  return qsModule.parse(location.search.slice(1));
+  // `qs` by default converts arrays with more than 20 items to an object.
+  // We want to avoid this because the data structure manipulated can therefore vary.
+  // Setting the limit to `100` seems a good number because the engine's default is 100
+  // (it can go up to 1000 but it is very unlikely to select more than 100 items in the UI).
+  // See:
+  //   - https://github.com/ljharb/qs#parsing-arrays
+  //   - https://www.algolia.com/doc/api-reference/api-parameters/maxValuesPerFacet/
+  return qsModule.parse(location.search.slice(1), { arrayLimit: 99 });
```
Haroenv: this comment could explain why it's 99 and not 100

src/lib/__tests__/RoutingManager-test.js
```diff
@@ -585,4 +586,67 @@ describe('RoutingManager', () => {
...
+          (_v, i) => `refinementList[brand][${i}]=brand-${i}`
+        ).join('&'),
+      ].join('?');
+
+    test('should parse refinements with more than 20 filters per category as array', () => {
+      const router = historyRouter();
+      const url = createFakeUrl({ length: 22 });
+      const parsedUrl = router.parseURL({
+        qsModule: qs,
+        location: new URL(url),
```
samouss: Here we use URL, but it's not the same object. Why not a mocked value?

samouss: Nope I was more thinking about a fake value, but this also work.

PR #3673, opened 2019-04-10
fix(infiniteHits): clear cache on query or refinement changes

yannickcr:
**Summary**

In https://github.com/algolia/instantsearch.js/pull/3645 we removed [a condition](https://github.com/algolia/instantsearch.js/blob/db9b91e140f3172545d76b37e1c153421ca20799/src/connectors/infinite-hits/connectInfiniteHits.js#L99-L102) that reseted the widget internal cache when reaching the first page. It was removed since it was conflicting with the new previous button feature.

Sadly, this reset step was essential to reset the widget internal cache on any refinement or query update so we don't append the new results with the ones from the previous search.

This fix re-introduces this code but with a different condition: since we can not rely on the page index anymore we now compare the full search state for any difference, and reset the cache if any difference (except for the page) is found.

While working correctly this fix is not very clean IMHO (it serialize the full state to do the comparison :nauseated_face: ), so if anyone has a better idea I am open to any suggestion :hugs: )

**Result**

### DIFF
--- a/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.js
+++ b/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.js
@@ -219,6 +219,54 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     expect(thirdRenderingOptions.results).toEqual(previousResults);
   });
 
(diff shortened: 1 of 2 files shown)


yannickcr:
Merged in https://github.com/algolia/instantsearch.js/pull/3675

PR #3675, opened 2019-04-11
feat(infiniteHits): add previous button

### LINKED ISSUE #2835: RFC: InfiniteHits should load previous pages

yannickcr:
**Summary**

The [InfiniteHits widget](https://www.algolia.com/doc/api-reference/widgets/infinite-hits/js/) is used to display a list of results with a “Show more” button. The user can click on this button to display more results without leaving its current page context.

Each time the user click on the “Show more” button the user will navigate to a new page (provided that [the routing option is enabled in InstantSearch](https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/#widget-param-routing)).

However, in the case the user navigates directly to a particular page (direct link, previous button, page refresh) then the widget only displays the last result page.

This change add a previous button to the Infinite Hits widget to give the possibility to the user to load the previous result pages.

For implementation details see [the full RFC](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/infinite-hits-previous-page.md).

Before being mergeable, this change requires an update in https://github.com/algolia/instantsearch-specs to add the new CSS classes:

* [x] Merge https://github.com/algolia/instantsearch-specs/pull/101
* [x] Publish new InstantSearch.css release: [Published in release v7.2.0](https://github.com/algolia/instantsearch-specs/releases/tag/v7.2.0)

**Note**

This PR was previously merged #3645 but was ultimately reverted #3674.

The reason is that it removed [a condition](https://github.com/algolia/instantsearch.js/blob/db9b91e140f3172545d76b37e1c153421ca20799/src/connectors/infinite-hits/connectInfiniteHits.js#L99-L102) that reseted the widget internal cache when reaching the first page. It was removed since it was conflicting with the new previous button feature.

Sadly, this reset step was essential to reset the widget internal cache on any refinement or query update so we don't append the new results with the ones from the previous search.

This was fixed by re-introducing this code but with a different condition: since we can not rely on the page index anymore we now compare the full search state for any difference, and reset the cache if any difference (except for the page) is found.

While working correctly this fix is not very clean IMHO (it serializes the full state to do the comparison :nauseated_face: ), so if anyone has a better idea I am open to any suggestion :hugs: )

**Result**

A "with previous button enabled" story has been added for the Instant Hits widget.

https://deploy-preview-3675--instantsearchjs.netlify.com/stories/?path=/story/infinitehits--with-previous-button-enabled

fixes #2835


src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -96,29 +119,45 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
...
       render({ results, state, instantSearchInstance }) {
-        if (state.page === 0) {
+        // Reset cache and received pages if anything changes in the
+        // search state, except for the page.
+        //
+        // We're doing this to "reset" the widget if a refinement or the
+        // query changes between renders, but we want to keep it as is
+        // if we only change pages.
+        const { page, ...stateWithoutPage } = state;
+        const currentState = JSON.stringify(stateWithoutPage);
```
francoischalifour: Is a comparaison with a stringified object safe in that case? Are we sure that the helper state always orders its state keys in the same order?

Since we use `lodash.isEqual` in the codebase, it might be better to rely on their implementation. According to [this benchmark](https://www.measurethat.net/Benchmarks/Show/1854/0/lodash-isequal-test), there are no important performance concerns to have.

WDYT?

yannickcr: From what I saw the state is an instance of `SearchParameters` and it's always constructed the same way, so it the keys are always in the same order.

For lodash, I though we were in the process of removing it from our dependencies?

francoischalifour: We're in the process of removing Lodash in the JavaScript helper, not in InstantSearch.js (not even sure we plan this).

Even if we were to remove Lodash, I think it's safer to use it here since we already often rely on their `isEqual` util function. We can still later think of creating our own `isEqual` util and replacing the import.

Would that make sense?

yannickcr: > We're in the process of removing Lodash in the JavaScript helper, not in InstantSearch.js (not even sure we plan this).

Well it seems logical to me to remove it from InstantSearch.js after the helper :wink:

> Even if we were to remove Lodash, I think it's safer to use it here since we already often rely on their isEqual util function. We can still later think of creating our own isEqual util and replacing the import.

I'm ok with that :slightly_smiling_face:

francoischalifour:
The latest fix doesn't seem to solve the issue. [In the story](https://deploy-preview-3645--instantsearchjs.netlify.com/stories/?path=/story/infinitehits--with-previous-button-enabled), the InfiniteHits still don't react to the new query.

yannickcr:
Mmm, the version on netlify seems to be an old one and does not contains the last changes.

What can explain that? Is there a way to force a new deployment?

francoischalifour:
You can re-trigger a deployment with a clear cache in the Netlify interface.

yannickcr:
Figured it out: the link in the description was from the old PR, the one with the bug :woman_facepalming: 

I updated it to link to the good deployment :innocent:

.storybook/preview-head.html
```diff
@@ -1,2 +1,2 @@
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.1.1/themes/algolia.min.css">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.2.0/themes/algolia.min.css">
```
francoischalifour:
```suggestion
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.3.1/themes/algolia.min.css">
```

src/connectors/infinite-hits/connectInfiniteHits.js
```diff
@@ -96,29 +121,44 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
...
+        // query changes between renders, but we want to keep it as is
+        // if we only change pages.
+        const { page, ...currentState } = state;
+        if (!isEqual(currentState, prevState)) {
           hitsCache = [];
-          lastReceivedPage = -1;
+          firstReceivedPage = page;
+          lastReceivedPage = page;
         }
+        prevState = currentState;
```
francoischalifour: This assignment can be moved to the above condition.

website/examples/calendar-widget/index.html
```diff
@@ -1,7 +1,7 @@
 <html>
 
 <head>
-  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.1.1/themes/algolia.min.css">
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.2.0/themes/algolia.min.css">
```
francoischalifour:
```suggestion
  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.3.1/themes/algolia.min.css">
```

website/examples/e-commerce/index.html
```diff
@@ -14,7 +14,7 @@
     />
     <link
       rel="stylesheet"
-      href="https://unpkg.com/instantsearch.css@7.1.1/themes/algolia-min.css"
+      href="https://unpkg.com/instantsearch.css@7.2.0/themes/algolia-min.css"
```
francoischalifour:
```suggestion
      href="https://unpkg.com/instantsearch.css@7.3.1/themes/algolia-min.css"
```

website/examples/tourism/index.html
```diff
@@ -14,7 +14,7 @@
     />
     <link
       rel="stylesheet"
-      href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.1.1/themes/reset-min.css"
+      href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.2.0/themes/reset-min.css"
```
francoischalifour:
```suggestion
      href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.3.1/themes/reset-min.css"
```

PR #3677, opened 2019-04-12
feat(Insights): expose `insights`  helper on instantsearch instance

tkrugg:
**Summary**

This is mimicking the way we allow usage for `highlight` and `snippet`.

**Result**

```js
  Instantsearch.widgets.hits({
    // ...
    templates: {
      item(hit) {
        return `
          <h2> ${hit.name} </h2>
          <button ${
            Instantsearch.insights('clickedObjectIDsAfterSearch', {
              eventName: "Add to favorite",
              objectIDs: [hit.objectID]
            })
          }>
            Add to favorite
          </button>
        `;
      },
    },
  });
```

### DIFF
--- a/src/helpers/index.ts
+++ b/src/helpers/index.ts
@@ -2,3 +2,4 @@ export * from './highlight';
 export * from './snippet';
 export { default as highlight } from './highlight';
 export { default as snippet } from './snippet';
+export { default as insights } from './insights';
--- a/src/lib/main.js
+++ b/src/lib/main.js
@@ -129,5 +129,6 @@ instantsearch.widgets = widgets;
 instantsearch.version = version;
 instantsearch.highlight = helpers.highlight;
 instantsearch.snippet = helpers.snippet;
+instantsearch.insights = helpers.insights;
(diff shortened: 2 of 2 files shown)


src/helpers/index.ts
```diff
@@ -2,3 +2,4 @@ export * from './highlight';
 export * from './snippet';
```
tkrugg: I don't see the use for theses. @francoischalifour I see you've added them 3 days ago. Can you explain the reason?

francoischalifour: This is used to [export the types](https://github.com/algolia/instantsearch.js/blob/92966391b1b2d2870827f164f1046d9a0c544efb/src/helpers/highlight.ts#L6-L10) to reuse them in the [Hogan helpers](https://github.com/algolia/instantsearch.js/blob/92966391b1b2d2870827f164f1046d9a0c544efb/src/lib/createHelpers.ts#L22).

tkrugg: ok thanks!

PR #3678, opened 2019-04-12
feat(Insights): expose connectHitsWithInsights and connectInfiniteHitsWithInsights

tkrugg:
**Summary**
The intention is to make it easier and more straight forward to create
custom InfiniteHits and Hits with the connector.

**Result**

If we exposed `withInsights` directly, this is what we'd need to decide
on a definitive name that implies it works only on Hits (like withHitsInsights)

```js
const connectHitsWithInsights = Instantsearch.connectors.withHitsInsights(Instantsearch.connectors.connectHits);
const connectInfiniteHitsWithInsights = Instantsearch.connectors.withHitsInsights(Instantsearch.connectors.connectInfiniteHits);
```

If we expose `connectHitsWithInsights` and `connectInfiniteHitsWithInsights` directly, we can keep `withInsights` totally private, and all custom components example are more simple.


Haroenv:
looks like it didn't pass lint

src/connectors/hits/__tests__/connectHitsWithInsights-test.js
```diff
@@ -0,0 +1,46 @@
...
+      state: helper.state,
+      helper,
+      createURL: () => '#',
+      instantSearchInstance: {
+        insightsClient: jest.fn(),
+      },
+    });
+
+    const secondRenderingOptions = rendering.mock.calls[1][0];
+    expect(typeof secondRenderingOptions.insights).toBe('function');
```
francoischalifour: You can use [`toBeInstanceOf`](https://jestjs.io/docs/en/expect#tobeinstanceofclass) for better error messages.

src/connectors/hits/__tests__/connectHitsWithInsights-test.js
```diff
@@ -0,0 +1,46 @@
+import jsHelper, { SearchResults } from 'algoliasearch-helper';
+import connectHitsWithInsights from '../connectHitsWithInsights';
+
+jest.mock('../../../lib/utils/hits-absolute-position', () => ({
+  addAbsolutePosition: hits => hits,
+}));
+
+
```
francoischalifour: The file isn't prettified.

src/connectors/hits/__tests__/connectHitsWithInsights-test.js
```diff
@@ -0,0 +1,46 @@
...
+      state: helper.state,
+      createURL: () => '#',
+      onHistoryChange: () => {},
+      instantSearchInstance: {
+        insightsClient: jest.fn(),
+      },
+    });
+
+    const firstRenderingOptions = rendering.mock.calls[0][0];
+    expect(typeof firstRenderingOptions.insights).toBe('function');
```
francoischalifour: You can use [`toBeInstanceOf`](https://jestjs.io/docs/en/expect#tobeinstanceofclass) for better error messages.

src/connectors/infinite-hits/__tests__/connectInfiniteHitsWithInsights-test.js
```diff
@@ -0,0 +1,46 @@
...
+      state: helper.state,
+      createURL: () => '#',
+      onHistoryChange: () => {},
+      instantSearchInstance: {
+        insightsClient: jest.fn(),
+      },
+    });
+
+    const firstRenderingOptions = rendering.mock.calls[0][0];
+    expect(typeof firstRenderingOptions.insights).toBe('function');
```
francoischalifour: Same here.

src/connectors/infinite-hits/__tests__/connectInfiniteHitsWithInsights-test.js
```diff
@@ -0,0 +1,46 @@
...
+      state: helper.state,
+      helper,
+      createURL: () => '#',
+      instantSearchInstance: {
+        insightsClient: jest.fn(),
+      },
+    });
+
+    const secondRenderingOptions = rendering.mock.calls[1][0];
+    expect(typeof secondRenderingOptions.insights).toBe('function');
```
francoischalifour: Same here.

francoischalifour:
The linting error comes all the way from #3598.

src/connectors/hits/__tests__/connectHitsWithInsights-test.js
tkrugg: fixed here 9a005ab

src/connectors/hits/__tests__/connectHitsWithInsights-test.js
tkrugg: fixed here 9a005ab

Haroenv:
that's likely already covered under the withInsights tests no?

tkrugg:
Rigorously speaking every test on connectInfinitHitsWithInsights should pass connectInfiniteHits test suite.
I was about to do that but it looked like a lot of tests to maintain and it's thoroughly covered by withInsights (which mocks a connector and specifically checks its props are preserved [here](https://github.com/algolia/instantsearch.js/blob/insights-connector/src/lib/__tests__/insights-client-test.ts#L44-L67)).

Let's add that test since it's something exposed and if it's breaks for any reason it's best to know.

PR #3695, opened 2019-04-16
fix(storybook): fix Hierarchical menu separator in Breadcrumb story

tkrugg:
If a separator is chosen which isn't the same as the corresponding hierarchical menu on the page, it will cause a conflict between those, and it will no longer be split correctly.

### DIFF
--- a/stories/breadcrumb.stories.js
+++ b/stories/breadcrumb.stories.js
@@ -135,7 +135,6 @@ storiesOf('Breadcrumb', module)
       search.addWidget(
         instantsearch.widgets.breadcrumb({
           container: breadcrumb,
-          separator: ' / ',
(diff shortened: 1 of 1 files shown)


stories/breadcrumb.stories.js
```diff
@@ -135,7 +135,7 @@ storiesOf('Breadcrumb', module)
       search.addWidget(
         instantsearch.widgets.breadcrumb({
           container: breadcrumb,
-          separator: ' / ',
+          separator: ' > ',
```
samouss: The option is really useful? Did you tried without it?

tkrugg: correct, removed altogether.

PR #3696, opened 2019-04-16
fix(stories): remove custom separator from breadcrumb.

Haroenv:
If a separator is chosen which isn't the same as the corresponding hierarchical menu on the page, it will cause a conflict between those, and it will no longer be split correctly.

spotted by @tkrugg initially

Check [story →](https://deploy-preview-3696--instantsearchjs.netlify.com/stories/?path=/story/breadcrumb--with-hierarchical-menu) (vs [before this change →](https://instantsearchjs.netlify.com/stories/?path=/story/breadcrumb--with-hierarchical-menu))


Haroenv:
duplicate of #3695

PR #3701, opened 2019-04-18
refactor(lodash): remove noop usage

francoischalifour:
This removes `lodash/noop` for a custom `noop` util function.


Haroenv:
Do you think we should enforce this? Maybe possible with eslint

src/lib/utils/noop.ts
```diff
@@ -0,0 +1,3 @@
+function noop() {}
+
+export default noop;
```
tkrugg: typing wise, do you knwo why they have `noop(...args: any[]): void;` in lodash and if there is any benefit to it?
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/06d8677cde28467e830bf77bc2cbfdaca37740aa/types/lodash/common/util.d.ts#L885

Haroenv: this means that you can call a noop without issues; I think we might not allow our noop to be called

francoischalifour: I guess it's to be more flexible. I'll update that.

francoischalifour: Hmm I didn't see your comment @Haroenv. I updated the signature.

src/lib/InstantSearch.js
```diff
@@ -8,7 +8,7 @@ import simpleMapping from './stateMappings/simple';
 import historyRouter from './routers/history';
```
samouss: Inside the file we have a line with:

```js
this._onHistoryChange = function() {};
```

We could also replace it with `noop`.

src/lib/utils/__tests__/noop-test.js
```diff
@@ -0,0 +1,10 @@
+import noop from '../noop';
+
+describe('noop', () => {
+  test('does nothing', () => {
+    expect(noop).toBeInstanceOf(Function);
+    expect(noop()).toBeUndefined();
+    // Checks that `noop` acts like a function.
+    expect(noop.toString).toBe(Function.prototype.toString);
+  });
+});
```
tkrugg: it's nitpicking but this file can be typescript

PR #3702, opened 2019-04-19
refactor(lodash): remove forEach usage

francoischalifour:
This removes `lodash/forEach` for the native `Array.prototype.forEach` method.

### DIFF
--- a/src/lib/__tests__/main-test.js
+++ b/src/lib/__tests__/main-test.js
@@ -1,9 +1,6 @@
 import instantsearch from '../main';
-import forEach from 'lodash/forEach';
 
 describe('instantsearch()', () => {
-  // to ensure the global.window is set
-
   it('includes a version', () => {
     expect(instantsearch.version).toMatch(
       /^(\d+\.)?(\d+\.)?(\*|\d+)(-beta.\d+)?$/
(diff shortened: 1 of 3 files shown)


src/lib/__tests__/main-test.js
```diff
@@ -1,34 +1,25 @@
...
-      );
+    Object.values(instantsearch.connectors).forEach(connector => {
+      expect(connector).toBeInstanceOf(Function);
     });
   });
 
   it('includes the highlight helper function', () => {
-    expect(typeof instantsearch.highlight).toEqual(
-      'function',
-      'THe highlight helper must be a function'
```
Haroenv: 👍

src/lib/utils/getRefinements.ts
```diff
@@ -85,66 +84,89 @@ function getRefinements(
   clearsQuery: boolean = false
 ): Refinement[] {
   const res: Refinement[] = [];
-
-  forEach(state.facetsRefinements, (refinements, attributeName) => {
-    forEach(refinements, name => {
+  const {
+    facetsRefinements = {},
```
Haroenv: just checking, when these are not set, it's undefined right (not `null`)?

francoischalifour: From the helper typings coming from DefinitelyTyped, these can be `undefined` (not `null`).

From my experiments, these are set by default in the search parameters (accordingly to `{}` or `[]`). I'm not sure in which case they can be `undefined` in the real world. I figured I would replicate almost exactly what Lodash did.

PR #3706, opened 2019-04-19
refactor(lodash): remove find usage

francoischalifour:
This removes all `lodash/find` usages. We need to be careful with `undefined` collections being loop over since Lodash used to deal with that for us.

Note that we already use `Array.prototype.find` in the codebase and that we document [using a polyfill to support browsers](https://www.algolia.com/doc/guides/building-search-ui/installation/js/#browser-support).

### DIFF
--- a/src/connectors/hits-per-page/connectHitsPerPage.js
+++ b/src/connectors/hits-per-page/connectHitsPerPage.js
@@ -1,4 +1,3 @@
-import find from 'lodash/find';
 import {
   checkRendering,
   warning,
@@ -114,7 +113,7 @@ export default function connectHitsPerPage(renderFn, unmountFn) {
       );
     }
 
-    const defaultValue = find(userItems, item => item.default === true);
+    const defaultValue = userItems.find(item => item.default === true);
 
     return {
       getConfiguration() {
--- a/src/connectors/hits-per-page/connectHitsPerPage.js
+++ b/src/connectors/hits-per-page/connectHitsPerPage.js
@@ -2,6 +2,7 @@ import {
   checkRendering,
   warning,
   createDocumentationMessageGenerator,
+  find,
 } from '../../lib/utils';
 
 const withUsage = createDocumentationMessageGenerator({
@@ -113,7 +114,7 @@ export default function connectHitsPerPage(renderFn, unmountFn) {
       );
     }
 
-    const defaultValue = userItems.find(item => item.default === true);
+    const defaultValue = find(userItems, item => item.default === true);
 
     return {
       getConfiguration() {
--- a/src/connectors/sort-by/connectSortBy.js
+++ b/src/connectors/sort-by/connectSortBy.js
@@ -1,6 +1,7 @@
 import {
   checkRendering,
   createDocumentationMessageGenerator,
+  find,
 } from '../../lib/utils';
 
 const withUsage = createDocumentationMessageGenerator({
@@ -100,7 +101,7 @@ export default function connectSortBy(renderFn, unmountFn) {
     return {
       init({ helper, instantSearchInstance }) {
         const currentIndex = helper.getIndex();
-        const isIndexInList = items.find(item => item.value === currentIndex);
+        const isIndexInList = find(items, item => item.value === currentIndex);
 
         if (!isIndexInList) {
           throw new Error(
--- a/src/connectors/sort-by/connectSortBy.js
+++ b/src/connectors/sort-by/connectSortBy.js
@@ -91,7 +91,7 @@ export default function connectSortBy(renderFn, unmountFn) {
   return (widgetParams = {}) => {
     const { items, transformItems = x => x } = widgetParams;
 
-    if (!items) {
+    if (!Array.isArray(items)) {
       throw new Error(
         withUsage('The `items` option expects an array of objects.')
       );
--- a/src/connectors/sort-by/connectSortBy.js
+++ b/src/connectors/sort-by/connectSortBy.js
@@ -1,4 +1,3 @@
-import find from 'lodash/find';
 import {
   checkRendering,
   createDocumentationMessageGenerator,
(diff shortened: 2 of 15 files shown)


src/lib/utils/getRefinements.ts
```diff
@@ -47,18 +46,22 @@ function getRefinement(
...
   let count: number;
 
   if (type === 'hierarchical') {
     const facetDeclaration = state.getHierarchicalFacetByName(attributeName);
-    const split = name.split(facetDeclaration.separator);
+    const nameParts = name.split(facetDeclaration.separator);
 
-    for (let i = 0; facet !== undefined && i < split.length; ++i) {
-      facet = find(facet.data, { name: split[i] });
+    for (let i = 0; facet !== undefined && i < nameParts.length; ++i) {
```
Haroenv: we should probably use a modern loop here too? wdyt?

francoischalifour: For sure – I was unsure if I should change this in this PR. I updated that part too.

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -115,7 +114,7 @@ export default function connectHitsPerPage(renderFn, unmountFn) {
       );
     }
 
-    const defaultValue = find(userItems, item => item.default === true);
+    const defaultValue = userItems.find(item => item.default === true);
```
Haroenv: I think userItems can be undefined

src/connectors/sort-by/connectSortBy.js
```diff
@@ -101,7 +100,7 @@ export default function connectSortBy(renderFn, unmountFn) {
     return {
       init({ helper, instantSearchInstance }) {
         const currentIndex = helper.getIndex();
-        const isIndexInList = find(items, item => item.value === currentIndex);
+        const isIndexInList = items.find(item => item.value === currentIndex);
```
Haroenv: items can be undefined here?

Haroenv: actually no it can't there's an error thrown earlier

src/lib/utils/getRefinements.ts
```diff
@@ -47,18 +46,22 @@ function getRefinement(
...
+      facet = Object.keys(facet.data)
```
Haroenv: I think data can be null here?

src/connectors/hits-per-page/connectHitsPerPage.js
samouss: We throw an error above if `items` (which is `userItems`) is not an array. It should be safe to use `find`.

https://github.com/algolia/instantsearch.js/blob/9d33599d0d95946bd8fba871b7ef0477e57c3614/src/connectors/hits-per-page/connectHitsPerPage.js#L104-L108

src/connectors/sort-by/connectSortBy.js
samouss: We throw an error for falsy value not for arrays. We should enforce the condition.

src/lib/utils/getRefinements.ts
```diff
@@ -47,18 +46,22 @@ function getRefinement(
   type: Refinement['type'],
   attributeName: Refinement['attributeName'],
   name: Refinement['name'],
-  resultsFacets: SearchResults['facets' | 'hierarchicalFacets']
+  resultsFacets: SearchResults['facets' | 'hierarchicalFacets'] = []
 ): Refinement {
   const res: Refinement = { type, attributeName, name };
-  let facet: any = find(resultsFacets, { name: attributeName });
+  let facet: any = (resultsFacets as Array<{ name: string }>).find(
```
samouss: `resultsFacets` is not correctly inferred?

src/lib/utils/getRefinements.ts
samouss: Worth checking at least.

src/connectors/hits-per-page/connectHitsPerPage.js
francoischalifour: It cannot because of those lines:

https://github.com/algolia/instantsearch.js/blob/9d33599d0d95946bd8fba871b7ef0477e57c3614/src/connectors/hits-per-page/connectHitsPerPage.js#L102-L108

The flow is hard to follow though.

src/lib/utils/getRefinements.ts
francoischalifour: What did you go for in the helper in the end? Would be nice to clean all these data structures to know exactly what we manipulate at some point.

src/lib/utils/getRefinements.ts
francoischalifour: No, it seems like TypeScript cannot infer it with the syntax:


```ts
SearchResults['facets' | 'hierarchicalFacets']
```

samouss: We should probably inline the value then.

francoischalifour:
@samouss Do you mean that it would be safer to create our own util?

samouss:
@francoischalifour Yep. @algolia/instantsearch-for-websites opinions?

Haroenv:
Yes, I'd go for polyfill here too for the reasons discussed by @samouss. This is likely going to break implementations since we didn't use it visibly before

francoischalifour:
That would make sense to me.

When you say "polyfill", you mean an actual polyfill or a wrapper function?

Haroenv:
We shouldn't polyfill, but instead use our own util function which uses indexOf. The real polyfill on MDN seems more advanced, but not things we really need: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#Polyfill (it also has a note about not polyfilling)

francoischalifour:
I created a `find` util function that uses the native `find` method if available, otherwise an implementation with `filter`.

I chose to keep the native one when available because it's optimized to return once the item is found (it doesn't go through all items).

If you're concerned about the `reduce` implementation which goes through all the items, we can refactor it with a basic loop.

Once this is merged, we can remove the `find` mention in the Browser support section of the documentation.

src/lib/utils/find.ts
```diff
@@ -0,0 +1,13 @@
+function find<TItem>(
```
Haroenv: maybe add a comment why we don't use native find always?

```suggestion
// We aren't using the native Array.find, because introducing the refactor away from lodash in a minor/patch
// adding a new required polyfill at this step, which before was only required for niche use-cases
// was decided as too risky. 
// @MAJOR  this can be replaced by native Array#find
function find<TItem>(
```

PR #3707, opened 2019-04-19
fix(hitsPerPage): improve warning for missing state value

francoischalifour:
I stumbled across this development warning message and couldn't understand it. I tried to rephrase it.

### DIFF
--- a/src/connectors/hits-per-page/connectHitsPerPage.js
+++ b/src/connectors/hits-per-page/connectHitsPerPage.js
@@ -144,7 +144,7 @@ Learn more: https://community.algolia.com/instantsearch.js/v2/widgets/configure.
           warning(
             false,
             `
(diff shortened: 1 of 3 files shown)


src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -143,9 +143,12 @@ Learn more: https://community.algolia.com/instantsearch.js/v2/widgets/configure.
 
           warning(
             false,
-            `No items in HitsPerPage \`items\` with \`value: hitsPerPage\` (hitsPerPage: ${
+            `
+The \`hitsPerPage\` widget \`items\` option does not contain the "hits per page" value coming from the state: ${
```
```suggestion
The \`items\` option of \`hitsPerPage\` does not contain the "hits per page" value coming from the state: ${
```

PR #3709, opened 2019-04-19
refactor(lodash): remove isFinite usage

francoischalifour:
This removes `lodash/isFinite` usage.

### DIFF
--- a/src/lib/utils/isFiniteNumber.ts
+++ b/src/lib/utils/isFiniteNumber.ts
@@ -1,0 +1,5 @@
+function isFiniteNumber(value: any): boolean {
+  return isFinite(value) && !isNaN(parseFloat(value));
+}
+
+export default isFiniteNumber;
--- a/src/lib/utils/isFiniteNumber.ts
+++ b/src/lib/utils/isFiniteNumber.ts
@@ -1,4 +1,4 @@
-function isFiniteNumber(value: any): boolean {
+function isFiniteNumber(value: any): value is number {
   return isFinite(value) && !isNaN(parseFloat(value));
 }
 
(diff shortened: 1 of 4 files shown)


Haroenv:
Number.isFinite isn't supported in IE11, where should we add a note for this? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite#Browser_compatibility

Otherwise this is fine, because lodash "polyfills" Number.isFinite: https://github.com/lodash/lodash/blob/4.17.11/lodash.js#L11628

francoischalifour:
Since I read that Lodash's `isFinite` relies on `Number.isFinite`, I thought that would be fine. Good catch.

We can use the top-level [`isFinite`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite) function which is supported on IE.

src/lib/utils/isFiniteNumber.ts
```diff
@@ -0,0 +1,5 @@
+function isFiniteNumber(value: any): boolean {
```
samouss:
```suggestion
function isFiniteNumber(value: any): value is number {
```

src/lib/utils/isFiniteNumber.ts
```diff
@@ -0,0 +1,5 @@
...
+  return isFinite(value) && !isNaN(parseFloat(value));
```
samouss: Why don't we use the MDN version, it should be enough?

> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite#Polyfill

src/lib/utils/isFiniteNumber.ts
```diff
@@ -0,0 +1,5 @@
+function isFiniteNumber(value: any): value is number {
```
samouss: We can add a comment to the polyfill version, since we don't provide a test.

> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite#Polyfill

PR #3711, opened 2019-04-19
chore(tools): replace tslint by typescript-eslint

yannickcr:
**Summary**

TypeScript support was added to `eslint-config-algolia@13.3.0` ([changelog](https://github.com/algolia/eslint-config-algolia/blob/master/CHANGELOG.md#1330-2019-04-16)), allowing us to drop TSLint and use ESLint to lint both JavaScript and TypeScript files.

* Update `eslint-config-algolia` to `v13.3.0`
* Enable `algolia/typescript` configuration
* Remove TSLint
* Update `lint` and `lint:fix` npm scripts to only use ESLint on JavaScript and TypeScript files
* Update files to match new rules:
  * Update ignore comments to target the new ESLint rules
  * Add return type to functions (new `@typescript-eslint/explicit-function-return-type` rule)
  * Update naming of generic type variables (new `@typescript-eslint/generic-type-naming` rule)
  * Fix errors in TypeScript files (not validated by ESLint before)

**Result**

Tooling only, no changes are visible to the users.
For developers the development environment is simplified with one less tool to setup.


src/helpers/__tests__/highlight-test.js
```diff
@@ -1,6 +1,6 @@
 import highlight from '../highlight';
 
-/* eslint-disable camelcase */
+/* eslint-disable @typescript-eslint/camelcase */
```
francoischalifour: The associated closing command must also be updated to `@typescript/eslint`. It still is:

```js
/* eslint-enable camelcase */
```

src/helpers/__tests__/snippet-test.js
```diff
@@ -1,6 +1,6 @@
 import snippet from '../snippet';
 
-/* eslint-disable camelcase */
+/* eslint-disable @typescript-eslint/camelcase */
```
francoischalifour: The associated closing command must be updated to `@typescript-eslint`.

src/lib/createHelpers.ts
```diff
@@ -8,16 +8,22 @@ import {
 
 type HoganRenderer = (value: any) => string;
 
+interface HoganHelpers {
```
francoischalifour: Was the ESLint configuration complaining about the structure not being created prior to usage here?

src/lib/suit.ts
```diff
@@ -1,12 +1,16 @@
...
   modifierName?: string;
-} = {}) => {
+}
+
+type SuitSelector = (names?: SuitNames) => string;
+
+export const component = (componentName: string): SuitSelector => ({
+  descendantName,
+  modifierName,
+}: SuitNames = {}) => {
```
francoischalifour: Again, do we want to allow inline types? It might be handy.

src/lib/utils/hits-query-id.ts
```diff
@@ -1,6 +1,6 @@
 import { Hits } from '../../types';
 
-export const addQueryID = (hits: Hits, queryID: string) => {
+export const addQueryID = (hits: Hits, queryID: number): Hits => {
```
francoischalifour: Why was the `queryID` converted to a number?

stories/query-rule-custom-data.stories.ts
```diff
@@ -37,7 +37,7 @@ storiesOf('QueryRuleCustomData', module)
                   const { title, banner, link } = item;
 
                   if (!banner) {
-                    return;
+                    return undefined;
```
francoischalifour: We can return the same data type.

```suggestion
                    return '';
```

src/lib/suit.ts
Haroenv: Interfaces and types are useful when you repeat, but would feel over the top for me if necessary everywhere

src/lib/createHelpers.ts
yannickcr: The ESLint configuration wants us to explicitly describe the return type of the `hoganHelpers` function. I could have put it inline without creating an interface but I found it clearer that way. 

Rule: [explicit-function-return-type](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md)

src/lib/suit.ts
yannickcr: There is no rules that restrict the usage of inline types, so you're free to use what is best for your usage. In this specific case I just found clearer to use an interface and a type.

src/lib/utils/hits-query-id.ts
yannickcr: For what I understand it was an error since `Hit.__queryID` is typed as a `number` in [src/types/instantsearch.ts](https://github.com/algolia/instantsearch.js/blob/develop/src/types/instantsearch.ts#L45-L70) and `addQueryID` is returning an array of `Hit`.

src/components/PoweredBy/PoweredBy.tsx
```diff
@@ -12,7 +12,7 @@ type Props = {
   cssClasses: CSSClasses;
 };
 
-const PoweredBy = ({ url, theme, cssClasses }: Props) => (
+const PoweredBy = ({ url, theme, cssClasses }: Props): React.ReactNode => (
```
Haroenv:
```suggestion
const PoweredBy React.FC<Props> = ({ url, theme, cssClasses })  => (
```

this style seems less repetitive to me, WDYT?

francoischalifour: This doesn't work with our Preact setup.

yannickcr: No opinion on this but one issue is that `eslint-plugin-react`'s `prop-types` rule does not support this pattern and will mark `url`, `theme` and `cssClasses` as missing in proptypes. 

```js
const PoweredBy: React.FC<Props> = ({ url, theme, cssClasses }: Props) => (
```
would works but it makes us repeat `Props` :neutral_face:

src/lib/utils/hits-query-id.ts
francoischalifour: Let's see if @tkrugg can confirm.

src/lib/suit.ts
```diff
@@ -1,12 +1,16 @@
 const NAMESPACE = 'ais';
 
-export const component = (componentName: string) => ({
-  descendantName,
-  modifierName,
-}: {
+interface SuitNames {
```
francoischalifour: This name is confusing because it implies that it's an array. Would `SuitOptions` make more sense here? (that's one drawback of naming things)

yannickcr: No objection to rename it (I was not convinced by the name myself :neutral_face:  )

src/components/PoweredBy/PoweredBy.tsx
yannickcr: (I should definitively add the support for this pattern in the plugin)

src/lib/utils/hits-query-id.ts
francoischalifour: I tried it in the storybook and it is a string:

image

The type in `instantsearch.ts`  is therefore wrong.

yannickcr: Ok, I fix it the other way around then :+1:

tkrugg: yes `queryID` and `__queryID` should be of type string.

src/types/widget.ts
```diff
@@ -37,7 +37,7 @@ type UiState = {
 export interface Widget {
   init?(options: InitOptions): void;
   render?(options: RenderOptions): void;
-  dispose?(options: DisposeOptions): SearchParameters | undefined;
+  dispose?(options: DisposeOptions): SearchParameters | void;
```
francoischalifour: From what I understand on [TypeScript `void` doc](https://www.typescriptlang.org/docs/handbook/basic-types.html#void), `void` accepts both `undefined` and `null`. We never return `null` in `dispose()`. This change is not a big deal though.

yannickcr: With `undefined` I had this error on the `dispose` method in [`src/lib/__tests__/insights-client-test.ts`](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/__tests__/insights-client-test.ts#L3-L12):

> Type '() => void' is not assignable to type '(options: DisposeOptions) => SearchParameters | undefined'.
  Type 'void' is not assignable to type 'SearchParameters | undefined'.ts(2322)

```ts
const connectHits = (renderFn: any, unmountFn: any): WidgetFactory<unknown> => (
  widgetParams = {}
): Widget => ({
  init() {},
    const hits = results.hits;
    renderFn({ hits, results, instantSearchInstance, widgetParams }, false);
  },
  dispose() {
    unmountFn();
  },
});
```

PR #3721, opened 2019-04-23
refactor(lodash): remove range and times usage

francoischalifour:
This removes `lodash/range` and `lodash/times` from the codebase.

I created a single custom `range` util for both functions because `lodash/times` is a subset of `lodash/range`.

I decided to change the `range` function signature to something more clear:

**Before**

`start` and `step` are optional, which makes it confusing because arguments change depending on what's given to the function.

```ts
function range(start = 0, end, step = 1): number[] | undefined;
```

**After**

```ts
type RangeOptions = {
  start?: number;
  end: number;
  step?: number;
};

function range({ start = 0, end, step = 1 }: RangeOptions): number[];
```

### DIFF
--- a/src/components/Slider/Slider.js
+++ b/src/components/Slider/Slider.js
@@ -2,8 +2,7 @@ import React, { Component } from 'preact-compat';
 import Rheostat from 'preact-rheostat';
 import PropTypes from 'prop-types';
 import cx from 'classnames';
-import times from 'lodash/times';
-import range from 'lodash/range';
+import { range } from '../../lib/utils';
 import Pit from './Pit';
 
 class Slider extends Component {
@@ -42,7 +41,9 @@ class Slider extends Component {
 
     const pitPoints = [
       min,
-      ...times(steps - 1, step => min + stepsLength * (step + 1)),
+      ...range({
+        end: steps - 1,
+      }).map(step => min + stepsLength * (step + 1)),
       max,
     ];
 
@@ -52,7 +53,7 @@ class Slider extends Component {
   // creates an array of values where the slider should snap to
   computeSnapPoints({ min, max, step }) {
     if (!step) return undefined;
-    return [...range(min, max, step), max];
+    return [...range({ start: min, end: max, step }), max];
   }
 
   createHandleComponent = tooltips => props => {
--- a/src/connectors/pagination/Paginator.js
+++ b/src/connectors/pagination/Paginator.js
@@ -1,4 +1,4 @@
-import range from 'lodash/range';
+import { range } from '../../lib/utils';
 
 class Paginator {
   constructor(params) {
@@ -13,7 +13,9 @@ class Paginator {
     if (total === 0) return [0];
 
     const totalDisplayedPages = this.nbPagesDisplayed(padding, total);
-    if (totalDisplayedPages === total) return range(0, total);
+    if (totalDisplayedPages === total) {
+      return range({ start: 0, end: total });
+    }
 
     const paddingLeft = this.calculatePaddingLeft(
       currentPage,
@@ -26,7 +28,7 @@ class Paginator {
     const first = currentPage - paddingLeft;
     const last = currentPage + paddingRight;
 
-    return range(first, last);
+    return range({ start: first, end: last });
   }
 
   nbPagesDisplayed(padding, total) {
--- a/src/connectors/pagination/Paginator.js
+++ b/src/connectors/pagination/Paginator.js
@@ -14,7 +14,7 @@ class Paginator {
 
     const totalDisplayedPages = this.nbPagesDisplayed(padding, total);
     if (totalDisplayedPages === total) {
-      return range({ start: 0, end: total });
+      return range({ end: total });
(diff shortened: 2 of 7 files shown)


src/connectors/pagination/__tests__/connectPagination-test.js
```diff
@@ -115,7 +115,11 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/pagination/
     }
 
     widget.render({
-      results: new SearchResults(helper.state, [{}]),
+      results: new SearchResults(helper.state, [
+        {
+          nbPages: 250,
```
francoischalifour: There was an issue swallowed by `lodash/range` implementation with this test. `nbPages` was `undefined` so the Paginator `end` value when `undefined` and the `end` value computed and given to `lodash/range` was `NaN`. Adding this value to the search results makes the test more consistent with real-life cases.

src/connectors/pagination/Paginator.js
```diff
@@ -13,7 +13,9 @@ class Paginator {
     if (total === 0) return [0];
 
     const totalDisplayedPages = this.nbPagesDisplayed(padding, total);
-    if (totalDisplayedPages === total) return range(0, total);
+    if (totalDisplayedPages === total) {
+      return range({ start: 0, end: total });
```
Haroenv: should we consistently leave out start = 0 or consistently pass it?

src/components/Slider/Slider.js
```diff
@@ -42,7 +41,9 @@ class Slider extends Component {
 
     const pitPoints = [
       min,
-      ...times(steps - 1, step => min + stepsLength * (step + 1)),
```
Haroenv: I love your change here, much better!

src/lib/utils/range.ts
```diff
@@ -0,0 +1,26 @@
...
+  // Example:
+  //   { start: 1, end: 5000, step: 500 }
+  //   => Array length = (5000 - 1) / 500 = 9.998
+  const arrayLength = Math.round((end - start) / limitStep);
+
+  return Array.apply(null, Array(arrayLength)).map((_, current) =>
+    // If the step is 0, the values shouldn't get incremented.
+    // Example:
+    //   range({ start: 1, end: 4, step: 0 }) => [1, 1, 1]
+    step === 0 ? start : (start + current) * step
```
Haroenv: do we really need step 0 in our code? I'd rather see that replaced by `Array(length).fill(x)`

src/lib/utils/range.ts
```diff
@@ -0,0 +1,26 @@
...
+  const limitStep = step === 0 ? 1 : step;
+
+  // In some cases the array to create has a decimal length.
+  // We therefore need to round the value.
+  // Example:
+  //   { start: 1, end: 5000, step: 500 }
+  //   => Array length = (5000 - 1) / 500 = 9.998
+  const arrayLength = Math.round((end - start) / limitStep);
+
+  return Array.apply(null, Array(arrayLength)).map((_, current) =>
```
Haroenv: What's the support for `Array.from({length}, (_,i) => i)`?

Haroenv: Edge 12 / Safari 9: no go :)

src/lib/utils/range.ts
Haroenv: I've looked around, and only see it if someone passes `step:  0` which doesn't really make sense?

francoischalifour: That's to be completely in sync with the previous Lodash behavior. I bet nobody used `{ step: 0 }` in our widgets... I'm OK to remove this.

Note that `Array.prototype.fill` is not supported by IE.

src/connectors/pagination/Paginator.js
francoischalifour: Good point – I can remove it here.

src/lib/utils/range.ts
Haroenv: let's remove this, maybe add an error on `step: 0` on the range constructor?

src/lib/utils/range.ts
```diff
@@ -0,0 +1,23 @@
...
+function range({ start = 0, end, step = 1 }: RangeOptions): number[] {
+  // We can't divide by 0 so we re-assign the step to 1 if it happens.
+  const limitStep = step === 0 ? 1 : step;
+
+  // In some cases the array to create has a decimal length.
+  // We therefore need to round the value.
+  // Example:
+  //   { start: 1, end: 5000, step: 500 }
+  //   => Array length = (5000 - 1) / 500 = 9.998
+  const arrayLength = Math.round((end - start) / limitStep);
```
Haroenv:
```suggestion
  const arrayLength = Math.round((end - start) / step);
```

since we no longer support step=0 this can be simplified too

francoischalifour: We should keep this check because using `{ step: 0 }` with the current implementation works but will increment as if `{ step: 1 }`. With the change you suggest, it will break the widget.

PR #3723, opened 2019-04-23
refactor(lodash): remove isEqual usage

francoischalifour:
This removes `lodash/isEqual` usage for our own custom `isEqual` function.

The `isEqual` function takes into account object references, primitive objects and nested objects. This covers all our usages.

### DIFF
--- a/src/lib/utils/isEqual.ts
+++ b/src/lib/utils/isEqual.ts
@@ -1,0 +1,31 @@
+function isPrimitive(obj: any): boolean {
+  return obj !== Object(obj);
+}
+
+function isEqual(first: any, second: any): boolean {
+  if (first === second) {
+    return true;
+  }
+
+  if (isPrimitive(first) || isPrimitive(second)) {
+    return first === second;
+  }
+
+  if (Object.keys(first).length !== Object.keys(second).length) {
+    return false;
+  }
+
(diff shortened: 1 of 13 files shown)


src/lib/utils/isEqual.ts
```diff
@@ -0,0 +1,31 @@
...
+function isEqual(first: any, second: any): boolean {
+  if (first === second) {
+    return true;
+  }
+
+  if (isPrimitive(first) || isPrimitive(second)) {
+    return first === second;
+  }
+
+  if (Object.keys(first).length !== Object.keys(second).length) {
```
Haroenv: This can throw in IE if first isn’t an object right?

francoischalifour: I tried from IE 9 to IE 11 and they all work as expected.

Haroenv: TIL you can use Object.keys on Arrays!

samouss: @Haroenv `undefined` & `null` fall inside the `isPrimitive` check.

francoischalifour:
@samouss Great call – added functions comparison support in 5e8e0c43f197db6c03ca4199c0e4936b82f8eb23.

PR #3729, opened 2019-04-24
refactor(lodash): remove mapValues usage

francoischalifour:
This removes `lodash/mapValues` usage from the codebase.

### DIFF
--- a/src/lib/utils/renderTemplate.js
+++ b/src/lib/utils/renderTemplate.js
@@ -1,4 +1,3 @@
-import mapValues from 'lodash/mapValues';
 import curry from 'lodash/curry';
 import hogan from 'hogan.js';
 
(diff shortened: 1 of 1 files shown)


src/lib/utils/renderTemplate.js
```diff
@@ -1,17 +1,22 @@
...
-  return mapValues(helpers, method =>
-    curry(function(text) {
-      const render = value => hogan.compile(value, compileOptions).render(this);
-      return method.call(data, text, render);
-    })
+function transformHelpersToHogan(helpers = {}, compileOptions, data) {
+  return Object.keys(helpers).reduce(
+    (acc, helperKey) => ({
+      ...acc,
+      [helperKey]: curry(function(text) {
```
Haroenv: do you get rid of curry in a separate PR? seems like it could cause conflicts easily

francoischalifour: I will take care of it in another PR. This is not a trivial change.

src/lib/utils/renderTemplate.js
```diff
@@ -1,17 +1,22 @@
...
-function transformHelpersToHogan(helpers, compileOptions, data) {
-  return mapValues(helpers, method =>
-    curry(function(text) {
-      const render = value => hogan.compile(value, compileOptions).render(this);
-      return method.call(data, text, render);
-    })
+function transformHelpersToHogan(helpers = {}, compileOptions, data) {
+  return Object.keys(helpers).reduce(
+    (acc, helperKey) => ({
+      ...acc,
```
Haroenv: perf-wise it's probably basically the same because of the low number of helpers, but this could be done with inner mutation too

PR #3730, opened 2019-04-24
refactor(lodash): remove escape usage

francoischalifour:
This removes `lodash/escape` usage from our codebase to use our own util. I basically copied their implementation and their test suite.

### DIFF
--- a/src/lib/utils/escape.ts
+++ b/src/lib/utils/escape.ts
@@ -1,6 +1,6 @@
 /**
  * This implementation is taken from Lodash implementation.
- * See: https://github.com/lodash/lodash/blob/master/escape.js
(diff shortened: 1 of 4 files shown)


src/lib/utils/escape.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * This implementation is taken from Lodash implementation.
+ * See: https://github.com/lodash/lodash/blob/master/escape.js
```
Haroenv:
```suggestion
 * See: https://github.com/lodash/lodash/blob/6b1d83c968b769206302c440ba0cc894cbd5209d/escape.js
```

or a tag

PR #3734, opened 2019-04-25
refactor(lodash): remove mergeWith usage

francoischalifour:
This removes `lodash/mergeWith` usage and introduces a custom `deepMerge` util. It also simplifies the logic of the `enhanceConfiguration` function.

### DIFF
--- a/src/lib/utils/mergeDeep.ts
+++ b/src/lib/utils/mergeDeep.ts
@@ -7,7 +7,7 @@ import isPlainObject from './isPlainObject';
  * - Primitive values are replaced
  * - Arrays a concatenated and their values are made unique
  */
-function mergeDeep(...values: any[]): object {
+function mergeDeep(...values: object[]): object {
   return values.reduce((acc, source = {}) => {
     Object.keys(source).forEach(key => {
       const previousValue = acc[key];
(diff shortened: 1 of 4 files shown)


src/lib/utils/mergeDeep.ts
```diff
@@ -0,0 +1,29 @@
+import uniq from './uniq';
+import isPlainObject from './isPlainObject';
+
+/**
+ * Deeply merges all the object values in a new object.
+ *
+ * - Primitive values are replaced
+ * - Arrays a concatenated and their values are made unique
+ */
+function mergeDeep(...values: any[]): object {
```
Haroenv: replace the any by:

```
type Item = {
  [key: string]: Array<Item> | Item
}
```

Something like that at least, will make it much clearer that this function only works for that

Haroenv: eh, never mind, I forgot you can't make circular types in TS easily (and I forgot the end type), so any might be fine

francoischalifour: That wouldn't work in this line:

```js
mergeDeep(previousValue, nextValue)
```

We can use `object` instead of `any`. WDYT?

Haroenv: yes, possible, it wasn't too clear what the types are to me, just based on this typing

francoischalifour:
@samouss I first broke the implementation to check the tests fail – this function is covered in our tests. It's used to merge all widgets' search parameters. It seems like it's working as expected both in our tests and in our stories. The only bottleneck that I could see is a lower performance compared to the implementation with Lodash but I'd argue this is meaningless.

PR #3737, opened 2019-04-26
refactor(lodash): remove Lodash deps

francoischalifour:
This is the base branch of the Lodash removal. I just removed the dependency `lodash` and the development dependency `@types/lodash`.

All changes were approved already, only 6941ca9fb101295deaec932705a6f7b97cb0ab00 needs to be reviewed.


.eslintrc.js
```diff
@@ -6,7 +6,19 @@ module.exports = {
     'react/no-string-refs': 1,
     // Avoid errors about `UNSAFE` lifecycles (e.g. `UNSAFE_componentWillMount`)
     'react/no-deprecated': 0,
+    '@typescript-eslint/no-unused-vars': [
+      'error',
+      { argsIgnorePattern: '^_', ignoreRestSiblings: true },
```
francoischalifour: The option `argsIgnorePattern` should be in our shared config.

.eslintrc.js
```diff
@@ -6,7 +6,19 @@ module.exports = {
...
+    ],
   },
+  overrides: [
+    {
+      files: ['*.ts', '*.tsx'],
+      rules: {
+        'valid-jsdoc': 0,
```
francoischalifour: Valid JSDoc is redundant with TypeScript signatures. I disabled this rule for TS files because it's still convenient to use JSDoc-style comments to explain a function.

Example:

```js
/**
 * Blabla...
 */
function (...args) {}
```

src/lib/utils/getRefinements.ts
```diff
@@ -58,13 +58,16 @@ function getRefinement(
...
+      facetData[refinementKey];
+
     for (let i = 0; facet !== undefined && i < nameParts.length; ++i) {
       facet =
         facet.data &&
         find(
-          Object.keys(facet.data).map(
-            refinementKey => facet.data[refinementKey]
-          ),
+          Object.keys(facet.data).map(getFacetRefinement(facet.data)),
```
francoischalifour: This removes the function creation in the loop to respect ESLint's `no-loop-func` rule.

src/lib/utils/range.ts
```diff
@@ -15,7 +15,7 @@ function range({ start = 0, end, step = 1 }: RangeOptions): number[] {
   //   => Array length = (5000 - 1) / 500 = 9.998
   const arrayLength = Math.round((end - start) / limitStep);
 
-  return Array.apply(null, Array(arrayLength)).map(
```
Haroenv: Seems the same, but why did lint not allow it?

francoischalifour: I forgot to add a note on this one. It was conflicting with the rule [`prefer-spread`](https://eslint.org/docs/rules/prefer-spread).

PR #3738, opened 2019-04-26
feat(voice): add additional query parameters

marieglr:
Two new arguments:
- `additionalQueryParameters: () => Partial<SearchParameters>` (only applied when voice search is mounted)
- `language: string` (iso 639-1), the parameter sent to Algolia will be always the short version

known limitation: additional query parameters will stay applied as long as the Voice Search widget is mounted, meaning they can cause stale values if you switch input method without unmounting (limitation of the architecture of InstantSearch)

### DIFF
--- a/src/connectors/voice-search/connectVoiceSearch.ts
+++ b/src/connectors/voice-search/connectVoiceSearch.ts
@@ -1,3 +1,4 @@
+import { PlainSearchParameters } from 'algoliasearch-helper';
 import {
   checkRendering,
   createDocumentationMessageGenerator,
@@ -15,7 +16,11 @@ const withUsage = createDocumentationMessageGenerator({
 });
 
 export type VoiceSearchConnectorParams = {
-  searchAsYouSpeak?: boolean;
+  searchAsYouSpeak: boolean;
+  language?: string;
+  additionalQueryParameters?: (params: {
+    query: string;
+  }) => PlainSearchParameters | void;
 };
 
 export interface VoiceSearchRendererOptions<TVoiceSearchWidgetParams>
@@ -71,7 +76,11 @@ const connectVoiceSearch: VoiceSearchConnector = (
       );
     };
 
-    const { searchAsYouSpeak = false } = widgetParams;
+    const {
+      searchAsYouSpeak = false,
+      language,
+      additionalQueryParameters,
+    } = widgetParams;
 
     return {
       $$type: 'ais.voiceSearch',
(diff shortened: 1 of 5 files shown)


marieglr:
Failed linting is expected, that needs to be fixed in #3601

vvo:
Nice :)

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -77,6 +90,20 @@ const connectVoiceSearch: VoiceSearchConnector = (
           const setQueryAndSearch = (query: string) => {
             if (query !== helper.state.query) {
               previousQuery = helper.state.query;
+              if (typeof additionalQueryParameters === 'function') {
+                const queryLanguages = language
+                  ? [language.split('-')[0]]
+                  : undefined;
+                helper.setState(
+                  helper.state.setQueryParameters({
+                    queryLanguages,
```
eunjae-lee: Is there any reason why `queryLanguages` is set only when there's `additionalQueryParameters`?

Haroenv: It might counteract with queryLanguages set with a `configure`, so it should be something you enable. Also I found it easier to implement that way

eunjae-lee:
There was a conflict, so I merged develop to this branch and now it looks ugly 😅

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -77,6 +90,20 @@ const connectVoiceSearch: VoiceSearchConnector = (
...
+                    ignorePlurals: true,
```
eunjae-lee: @Haroenv 
When these three properties need to be set? When there's queryLanguages? When there's additionalQueryParams? or `queryLanguages || additionalQueryParams`?

```
                    ignorePlurals: true,
                    removeStopWords: true,
                    optionalWords: query,
```

Haroenv: when additionalParameters is defined

eunjae-lee: Can we replace 

```js
if (typeof additionalQueryParameters === 'function') {
  const queryLanguages = language
    ? [language.split('-')[0]]
    : undefined;
  helper.setState(
    helper.state.setQueryParameters({
      queryLanguages,
      ignorePlurals: true,
      removeStopWords: true,
      optionalWords: query,
      ...additionalQueryParameters({ query }),
    })
  );
}
```

with

```js
const queryLanguages = language
  ? [language.split('-')[0]]
  : undefined;

if (typeof queryLanguages !== 'undefined' || typeof additionalQueryParameters === 'function') {
  helper.setState(
    helper.state.setQueryParameters({
      queryLanguages,
      ...(typeof additionalQueryParameters === 'function' && {
        ignorePlurals: true,
        removeStopWords: true,
        optionalWords: query,
        ...additionalQueryParameters({ query }),
      })
    })
  )
}
```

Since `language` is a widget param, user should be able to use it without having to use `additionalQueryParameters`. I don't know how to make it prettier, though.

PR #3739, opened 2019-04-26
refactor(RoutingManager): migrate to TypeScript

francoischalifour:
This migrates the `RoutingManager` to TypeScript. It will make this quarter's bigger refactors easier.


src/lib/RoutingManager.ts
```diff
@@ -102,55 +136,62 @@ export default class RoutingManager {
...
 
-  dispose() {
-    if (this.renderURLFromState) {
-      this.instantSearchInstance.helper.removeListener(
-        'change',
-        this.renderURLFromState
-      );
-    }
-    this.router.dispose();
+  public init({ state }: { state: SearchParameters }): void {
```
francoischalifour: For some reasons the lifecycles hooks are not picked up by TypeScript even with:

```ts
... implement Widget {
 // ...
}
```

This is why these methods are typed.

src/types/instantsearch.ts
```diff
@@ -91,11 +92,26 @@ export type NumericRefinement = {
...
+export type Router = {
+  onUpdate(callback: (route: string) => void): void;
+  read(): UiState;
+  write(route: string): void;
+  createURL(state: Partial<SearchParameters>): string;
+  dispose(): Widget['dispose'];
+};
+
+export type StateMapping = {
+  stateToRoute(state: UiState): string;
```
samouss: The return value is not a `string` but an object that represents the `uiState` for the URL. 

By default it's our `UIState` but it can be different if `StateMapping` is implemented by the user.

src/types/instantsearch.ts
```diff
@@ -91,11 +92,26 @@ export type NumericRefinement = {
...
+  routeToState(route: Partial<SearchParameters>): UiState;
```
samouss: The `route` parameter is not a `SearchParameters` but an object that represents the `uiState` for the URL.

By default it's our `UIState` but it can be different if `StateMapping` is implemented by the user.

src/lib/RoutingManager.ts
samouss: https://github.com/Microsoft/TypeScript/issues/1373

Haroenv:
I wonder if this was the case before the PR already, but TS complains in the voice helper tests

francoischalifour:
@Haroenv That's new to this PR because the `InstantSearch` type is getting stricter. These issues arrived when I rebased. I'll fixed those.

src/lib/RoutingManager.ts
```diff
@@ -0,0 +1,213 @@
...
+      });
+    }, {});
+  }
+
+  private setupRouting(state: SearchParameters): void {
+    const helper = this.instantSearchInstance.helper!;
+
+    this.router.onUpdate(route => {
+      const nextUiState = this.stateMapping.routeToState(route);
+      const widgetsUIState = this.getAllUIStates({
```
samouss: We still have inconsistencies with `uiState` vs `UIState`.

src/types/instantsearch.ts
```diff
@@ -91,11 +92,26 @@ export type NumericRefinement = {
...
+export type Router = {
+  onUpdate(callback: (route: Partial<SearchParameters>) => void): void;
+  read(): UiState;
+  write(route: Partial<SearchParameters>): void;
+  createURL(state: UiState): string;
+  dispose(): Widget['dispose'];
+};
+
+export type StateMapping = {
+  stateToRoute(state: UiState): UiState;
```
samouss: This is not 100% correct, we should define a `RouteState` generics with a default of `UiState`. 

This is also true for the `routeToState` function.

src/types/instantsearch.ts
```diff
@@ -91,11 +92,26 @@ export type NumericRefinement = {
 
 export type Refinement = FacetRefinement | NumericRefinement;
 
+export type Router = {
+  onUpdate(callback: (route: Partial<SearchParameters>) => void): void;
+  read(): UiState;
+  write(route: Partial<SearchParameters>): void;
```
samouss: The function accepts a `RouteState` not a `Partial<SearchParameters>`.

src/types/instantsearch.ts
```diff
@@ -91,11 +92,26 @@ export type NumericRefinement = {
 
 export type Refinement = FacetRefinement | NumericRefinement;
 
+export type Router = {
+  onUpdate(callback: (route: Partial<SearchParameters>) => void): void;
```
samouss: The function provides a `RouteState` not a `Partial<SearchParameters>`.

src/types/instantsearch.ts
```diff
@@ -91,11 +92,26 @@ export type NumericRefinement = {
 
 export type Refinement = FacetRefinement | NumericRefinement;
 
+export type Router = {
```
samouss: This should probably be an interface that is implemented by the `history` router.

src/types/instantsearch.ts
```diff
@@ -91,11 +92,26 @@ export type NumericRefinement = {
...
+  onUpdate(callback: (route: Partial<SearchParameters>) => void): void;
+  read(): UiState;
+  write(route: Partial<SearchParameters>): void;
+  createURL(state: UiState): string;
+  dispose(): Widget['dispose'];
+};
+
+export type StateMapping = {
```
samouss: This should probably be an interface that is implemented by the `simple` mapping.

src/types/instantsearch.ts
francoischalifour: Yes – we'll be able to do this once this PR is merged.

src/types/instantsearch.ts
francoischalifour: Yes – that's for another PR.

src/types/instantsearch.ts
```diff
@@ -93,9 +93,9 @@ export type NumericRefinement = {
 export type Refinement = FacetRefinement | NumericRefinement;
 
 export type Router = {
-  onUpdate(callback: (route: Partial<SearchParameters>) => void): void;
+  onUpdate(callback: (route: RouteState) => void): void;
   read(): UiState;
-  write(route: Partial<SearchParameters>): void;
+  write(route: RouteState): void;
   createURL(state: UiState): string;
```
samouss: The `createURL` function also takes a `RouteState` and not an `UiState`.

PR #3740, opened 2019-04-26
refactor(history): migrate to TypeScript

francoischalifour:
This migrates the browser history router to TypeScript.

Since Git loses track of the file, you can review the commit 32aa9bc6b4aceb0902e1f67298dfe049e40a425d which is the actual migration after the file rename.


src/lib/routers/history.ts
```diff
@@ -67,19 +112,24 @@ class BrowserHistory {
...
+  private write(routeState: UiState): void {
     const url = this.createURL(routeState);
     const title = this.windowTitle && this.windowTitle(routeState);
 
     if (this.writeTimer) {
       window.clearTimeout(this.writeTimer);
     }
 
-    this.writeTimer = setTimeout(() => {
+    this.writeTimer = window.setTimeout(() => {
```
Haroenv: this code now will no longer run on node, is there a reason you added window.?

src/lib/routers/history.ts
```diff
@@ -88,50 +138,39 @@ class BrowserHistory {
...
-
-  /**
-   * This methods sets a callback on the `onpopstate` event of the history API
-   * of the current page. This way, the URL sync can keep track of the changes.
-   * @param {function(object)} cb the callback that will receive the latest routeState.
-   * It is called when the URL is updated.
-   * @returns {undefined}
-   */
-  onUpdate(cb) {
+  public onUpdate(callback: (routeState: UiState) => void): void {
```
Haroenv: How can routeState === uiState?

src/lib/routers/history.ts
```diff
@@ -23,38 +54,53 @@ function defaultParseURL({ qsModule, location }) {
   //   - https://github.com/ljharb/qs#parsing-arrays
   //   - https://www.algolia.com/doc/api-reference/api-parameters/maxValuesPerFacet/
   return qsModule.parse(location.search.slice(1), { arrayLimit: 99 });
-}
+};
 
-function setWindowTitle(title) {
+const setWindowTitle = (title?: string): void => {
```
Haroenv: no advantage having this as arrow IMO, but it's changed now so no big deal

src/lib/routers/history.ts
```diff
@@ -1,17 +1,48 @@
...
   const queryString = qsModule.stringify(routeState);
   const portWithPrefix = port === '' ? '' : `:${port}`;
-  // IE <= 11 has no location.origin or buggy. Therefore we don't rely on it
-  if (!routeState || Object.keys(routeState).length === 0)
+
+  // IE <= 11 has no proper `location.origin` so we cannot rely on it.
+  if (!routeState || Object.keys(routeState).length === 0) {
     return `${protocol}//${hostname}${portWithPrefix}${pathname}${hash}`;
-  else
+  } else {
```
Haroenv: this can just be return with the else removed

Haroenv:
I'd create a separate type for RouteState

src/lib/routers/history.ts
francoischalifour: TypeScript types this as `NodeJS.Timer` otherwise, which is not assignable to a number. See can [read more on this thread](https://github.com/Microsoft/TypeScript/issues/842).

Otherwise we can keep the previous version and type `writeTimer` as any.

src/lib/routers/history.ts
francoischalifour: `UiState` is rather generic but you're right, let's create another type.

src/lib/routers/history.ts
francoischalifour: It was to be consistent with the two above functions.

src/lib/routers/history.ts
Haroenv: We don't actually care what the type of the number is, I'd put it as any

francoischalifour: Note that we rely on `window` everywhere else in this file, so we might want to change that later on. Besides, this is a browser history.

src/lib/routers/history.ts
```diff
@@ -0,0 +1,200 @@
+import qs from 'qs';
+import { RouteState } from '../../types';
+
+type BrowserHistoryProps = {
+  windowTitle?: (routeState: RouteState) => string;
+  writeDelay: number;
+  createURL({
```
samouss: We could extra the signature of the function outside of the type definition. That would be easier to follow.

src/lib/routers/history.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    const title = this.windowTitle && this.windowTitle(this.read());
+
+    setWindowTitle(title);
+  }
+
+  /**
+   * Reads the URL and returns a syncable UI search state.
+   */
+  private read(): RouteState {
```
samouss: read/write are also public, we call them from the `RoutingManager`.

https://github.com/algolia/instantsearch.js/blob/21e4d81b3e37c03686acaf3a8dc2c73c3c519570/src/lib/RoutingManager.js#L78

https://github.com/algolia/instantsearch.js/blob/21e4d81b3e37c03686acaf3a8dc2c73c3c519570/src/lib/RoutingManager.js#L12

src/lib/routers/history.ts
```diff
@@ -0,0 +1,200 @@
...
+  parseURL({
+    qsModule,
+    location,
+  }: {
+    qsModule: typeof qs;
+    location: Location;
+  }): RouteState;
+};
+
+const defaultCreateURL: BrowserHistoryProps['createURL'] = ({
```
samouss: IIRC we lose the inference with this notation sometimes. We could rather inline the type directly. It won't compile if there is a mismatch anyway. This is especially true if we extract the value of the function into its own type.

src/lib/routers/history.ts
```diff
@@ -0,0 +1,197 @@
+import qs from 'qs';
+import { RouteState } from '../../types';
+
+type CreateUrlOptions = {
```
samouss: IIRC we use a lot `URL` vs `Url` inside the codebase. We should stay consistent.

src/lib/routers/history.ts
```diff
@@ -0,0 +1,197 @@
...
+
+type ParseUrlOptions = {
+  qsModule: typeof qs;
+  location: Location;
+};
+
+type BrowserHistoryProps = {
+  windowTitle?: (routeState: RouteState) => string;
+  writeDelay: number;
+  createURL({ qsModule, routeState, location }: CreateUrlOptions): string;
```
samouss: I was more thinking about extracting the whole function to be able to re-use it later in the file.

```ts
type CreateURL = ...
```

PR #3748, opened 2019-04-29
chore: disable canonical link

samouss:
**Summary**

We introduce [the canonical URLs](https://github.com/algolia/instantsearch.js/pull/3576) to keep backlinks to the algolia.com/doc. But it's not enough to make the Community website disappear from the search results. We have to hide them from the search results otherwise it's confusing for users.


Haroenv:
I guess you did it on purpose, but is there a reason you wouldn't keep both tags if available?

samouss:
Yes, it's not recommended to mix both attributes on the same page. The intent of both attributes is contradictory and robots seems to take the canonical links over the other instructions.

PR #3749, opened 2019-04-29
fix(numericMenu): prevent refinement reset on checked radio click

yannickcr:
**Summary**

**Summary**

In the `NumericMenu` we bind an `onClick` event listener on the list item and change the radio button status at each click.

While this behavior is perfect for a list containing checkboxes (like a `RefinementList`) it does not fit for a list containing radio button, like the `NumericMenu` since it will reset the refinement if you click on an already selected radio button.

A way to fix this would be to use an `onChange` event listener on the radio button instead of the `onClick` on the list item (like the Angular and Vue implementations), but sadly in our case the radio button is made using a template and we cannot bind the event to it without affecting all widgets using `RefinementList`  :disappointed:.

The used solution was to prevent the refinement if the user click on an already checked radio button (so we do not affect `RefinementList` with checkboxes). This change is made at two places in the `RefinementList` component, depending if you directly clicked on the radio button or on one of it's parent.

A new unit test was added to test the new behavior, but I was not able to test a direct click on the input element (since it is created using a template we cannot simulate a click on it :cry:).

**Result**


src/components/RefinementList/RefinementList.js
```diff
@@ -100,6 +100,14 @@ class RefinementList extends Component {
       return;
     }
 
+    if (
+      originalEvent.target.matches('input[type="radio"]:checked') &&
+      isRefined
+    ) {
+      // Prevent refinement for being reset if the user click on an already checked radio button
```
francoischalifour:
```suggestion
      // Prevent refinement for being reset if the user clicks on an already checked radio button
```

src/components/RefinementList/RefinementList.js
```diff
@@ -125,6 +133,14 @@ class RefinementList extends Component {
 
     originalEvent.stopPropagation();
 
+    if (
+      originalEvent.target.querySelector('input[type="radio"]:checked') &&
+      isRefined
+    ) {
+      // Prevent refinement for being reset if the user click on an already checked radio button
```
francoischalifour:
```suggestion
      // Prevent refinement for being reset if the user clicks on an already checked radio button
```

src/components/RefinementList/RefinementList.js
```diff
@@ -100,6 +100,14 @@ class RefinementList extends Component {
       return;
     }
 
+    if (
+      originalEvent.target.matches('input[type="radio"]:checked') &&
```
francoischalifour: According to MDN, [`Element.matches()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/matches#Browser_compatibility) isn't supported by IE9.

yannickcr: Oh, checked too quickly and didn't saw the icon in the compatibility table. I will update the code to also test `msMatchesSelector`.

francoischalifour: I wasn't able to try the story on IE9 because storybook doesn't work on this browser. It seems like we never used `Element.matches()` but rather `event.target.querySelector()`:

https://github.com/algolia/instantsearch.js/blob/e610946c77088e6bdbb49916caff0b0ebb2a57d9/src/components/RefinementList/RefinementList.js#L118-L122

yannickcr: I rewrote the condition to use `event.target.parentNode.querySelector`.

src/components/RefinementList/RefinementList.js
```diff
@@ -99,6 +99,16 @@ class RefinementList extends Component {
       return;
     }
 
+    if (
+      originalEvent.target.parentNode.querySelector(
+        'input[type="radio"]:checked'
+      ) &&
+      isRefined
```
francoischalifour: We can check for `isRefined` first to not do a DOM operation every time.

PR #3751, opened 2019-05-03
 refactor(infiniteHits): migrate to TypeScript 

yannickcr:
**Summary**

Following up the current TypeScript migration this change migrate the InfiniteHits widget to TypeScript, including the component, connector and tests.

**Result**

No visible changes.


src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -91,27 +131,27 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
...
-        return escapeHTML ? TAG_PLACEHOLDER : undefined;
+        return escapeHTML ? TAG_PLACEHOLDER : {};
       },
 
       init({ instantSearchInstance, helper }) {
-        this.showPrevious = getShowPrevious(helper);
-        this.showMore = getShowMore(helper);
-        firstReceivedPage = helper.state.page;
-        lastReceivedPage = helper.state.page;
+        (this as any).showPrevious = getShowPrevious(helper);
```
yannickcr: Is there a better way to do this ?

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -91,27 +131,27 @@ export default function connectInfiniteHits(renderFn, unmountFn) {
...
+        (this as any).showMore = getShowMore(helper);
+        firstReceivedPage = helper.state.page as number;
```
yannickcr: According to the `algoliasearch-helper` type definition `page` can be `undefined` https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/algoliasearch-helper/index.d.ts#L429

Is this true or is this an error in the type definition? Because it is a case that is not handled by the widget at this time (this is why we force to `number` type here)

src/lib/insights/client.ts
```diff
@@ -94,12 +94,12 @@ const wrapInsightsClient = (
...
 
 type Connector<TWidgetParams> = (
-  renderFn: Renderer<RenderOptions<TWidgetParams>>,
+  renderFn: Renderer<any>,
   unmountFn: Unmounter
 ) => WidgetFactory<TWidgetParams>;
 
 export default function withInsights(
-  connector: Connector<unknown>
+  connector: Connector<any>
```
yannickcr: Didn't find a better way to do this. Any idea?

src/connectors/infinite-hits/connectInfiniteHits.ts
francoischalifour: We don't need to store it on `this`. We can declare it in an upper-level in a variable.

src/connectors/infinite-hits/connectInfiniteHits.ts
francoischalifour: You can use `helper.state.page!`. I don't think it can be `undefined`. If needed, you can override the type in:
https://github.com/algolia/instantsearch.js/blob/9741445905265ffb0f3e770fbea9c8e0b0ac4849/src/types/instantsearch.ts#L110

src/widgets/infinite-hits/infinite-hits.tsx
```diff
@@ -12,20 +14,58 @@ import {
...
+export type InfiniteHitsTemplates = {
+  empty: Template<void>;
+  showPreviousText: Template<void>;
+  showMoreText: Template<void>;
+  item: Template<Hit>;
+};
+
+export type InfiniteHitsRendererWidgetParams = {
+  escapeHTML: boolean;
+  transformItems: (items: any[]) => any;
```
francoischalifour: I can't find any reasons for the return type not to be an array for this widget.

yannickcr: I don't know either, but it was typed that way in https://github.com/algolia/instantsearch.js/blob/9dca548/src/connectors/query-rules/connectQueryRules.ts#L29.

Can be good to be stricter on that if we agree that we can only return an Array here.

francoischalifour: For Query Rules it was a bit different. We wanted to allow people to use `transformItems` to process the `items` as a single one because most of the times, this is what you want. I'm not sure we should recommend this pattern though.

yannickcr: Oh, ok. I did not saw that the context was different for the two widgets. I updated the definition to assert that we always return an Array.

src/types/widget.ts
```diff
@@ -11,6 +11,7 @@ interface InitOptions {
   helper: Helper;
   templatesConfig: object;
   createURL(state: SearchParameters): string;
+  onHistoryChange?(state: Partial<SearchParameters>): void;
```
francoischalifour: `onHistoryChange` is not used by any widget anymore. We can remove it from here and the tests.

src/connectors/hits/__tests__/connectHitsWithInsights-test.ts
```diff
@@ -7,37 +7,52 @@ jest.mock('../../../lib/utils/hits-absolute-position', () => ({
 }));
 
 describe('connectHitsWithInsights', () => {
+  const defaultInitOptions = {
+    instantSearchInstance: {
+      helper: null,
+      widgets: [],
+      insightsClient: jest.fn(),
```
francoischalifour: This Jest function will always be the same. It would usually better to create a function to create the default init options (`getDefaultInitOptions()`) so that we get a new spy instance every time.

However, in this case, we do not need a spy. We can go with `() => {}`.

src/types/widget.ts
yannickcr: Oh ok, I thought that it was missing from the interface, I'll remove it from here and the tests.

src/widgets/infinite-hits/infinite-hits.tsx
```diff
@@ -12,20 +14,58 @@ import {
...
+};
+
+export type InfiniteHitsTemplates = {
+  empty: Template<void>;
+  showPreviousText: Template<void>;
+  showMoreText: Template<void>;
+  item: Template<Hit>;
+};
+
+export type InfiniteHitsRendererWidgetParams = {
```
samouss: Why do we have a type related to the renderer at this level? It should probably be into the connector. Those values are indeed required by the renderer but not by the widget itself. At the moment we can't use the widget without any of those options.

PR #3758, opened 2019-05-06
fix(rangeSlider): round the slider pit value

### LINKED ISSUE #2828: Add an option for control which Pit value is displayed in the RangeSlider

francoischalifour:
**Summary**

This adds a missing base to the [`parseInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) function.

**Result**

Before:

image

After:

image

### DIFF
--- a/src/components/Slider/Pit.js
+++ b/src/components/Slider/Pit.js
@@ -10,7 +10,7 @@ const Pit = ({ style, children }) => {
   // Children could be an array, unwrap the value if it's the case
   // see: https://github.com/developit/preact-compat/issues/436
   const value = Array.isArray(children) ? children[0] : children;
-  const pitValue = Math.round(parseFloat(value) * 100) / 100;
+  const pitValue = Math.round(parseInt(value) * 100) / 100;
 
   return (
     <div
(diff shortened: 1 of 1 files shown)


francoischalifour:
@Haroenv The original PR stated that. Indeed, it doesn't, but it still solves an issue.

francoischalifour:
@tkrugg We're always in control of the values because the `Slider` component from Rheostat passes them. This won't ever be breaking. As for the test, I didn't find any way to test it because mounting the component within a `Slider` would mount the whole external Rheostat component.

francoischalifour:
This was already approved in #2904 by the way.

PR #3762, opened 2019-05-07
refactor(RoutingManager): migrate tests to TypeScript

francoischalifour:
This refactors the `RoutingManager` test suite to TypeScript. It ensures that our types are correct.


Haroenv:
looks like the test you changed fails now :)

src/lib/__tests__/RoutingManager-test.ts
```diff
@@ -402,8 +438,8 @@ describe('RoutingManager', () => {
         },
       });
 
-      const fakeSearchBox = createFakeSearchBox();
-      const fakeHitsPerPage = createFakeHitsPerPage();
+      const fakeSearchBox: any = createFakeSearchBox({});
```
francoischalifour: `any` is used because we dynamically create a `refine()` function on the widget to test it. This is not part of the original widget definition.

src/lib/RoutingManager.ts
```diff
@@ -36,7 +36,7 @@ class RoutingManager implements Widget {
     this.currentUiState = this.stateMapping.routeToState(this.router.read());
   }
 
-  private getAllSearchParameters({
+  public getAllSearchParameters({
```
Haroenv: I wonder if we should even have these public / private markers 🤔 since they aren't _actually_ protecting anything from consumers from using it, except if they use TS

src/lib/__tests__/RoutingManager-test.ts
```diff
@@ -3,23 +3,42 @@ import qs from 'qs';
...
-
-const createFakeSearchClient = () => ({
+const createFakeSearchClient: () => FakeSearchClient = () => ({
   search: () => Promise.resolve({ results: [{}] }),
 });
 
-const createFakeRouter = (args = {}) => ({
-  onUpdate() {},
-  write() {},
+const createFakeRouter: (args: Partial<Router>) => Router = (args = {}) => ({
```
Haroenv:
```suggestion
const createFakeRouter = (args = {}: Partial<Router>):  Router => ({
```

I forgot the exact syntax, but something more like that makes more sense to me

src/lib/RoutingManager.ts
yannickcr: The presence of those markers is enforced by [@typescript-eslint/explicit-member-accessibility](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-member-accessibility.md).

I think they are useful since they are protecting us (and some consumers) to use private methods. They also could potentially make our lives easier if we migrate to native [private methods](https://github.com/tc39/proposal-private-methods) in the future.

But for consumers that do not use TS I think we should also prefix the method names with a `_` since it is common to indicate a private property/method like this in JavaScript (this could be enforced by a tool like ESLint).

yannickcr: In the context of this PR, are you changing the accessibility of the methods just for the tests? If yes, I do not think this is a good idea :neutral_face:  . We should limit our tests to (real) public methods and potential side effects and not trying to test internal state.

WDYT?

francoischalifour: Yes, that's why I changed the visibility. I don't think that's good either but I didn't want to dramatically update the tests/implementation because of the migration.

Do you have an idea in mind how to rewrite the test while keeping the private visibility? Maybe this doesn't need to be tested at this level.

yannickcr: Since these tests are testing specifically the `getAllSearchParameters` method I would have removed them. The conditions that are tested here should ideally be covered by the other tests, but I don't know if it is the case :thinking: 

In doubt, for now I would have added a `// @ts-ignore` comment in the tests, not perfect but I find this better than publicly exposing something that should not be used.

WDYT?

francoischalifour: That's indeed an acceptable solution for now!

src/lib/routers/history.ts
```diff
@@ -76,12 +76,12 @@ class BrowserHistory {
...
    * The storage adaptor maps all syncable keys to the query string of the URL.
    */
-  private readonly _createURL: BrowserHistoryProps['createURL'];
+  public readonly _createURL: BrowserHistoryProps['createURL'];
   /**
    * Parses the URL into a route state.
    * It should be symetrical to `createURL`.
    */
-  private readonly parseURL: BrowserHistoryProps['parseURL'];
+  public readonly parseURL: BrowserHistoryProps['parseURL'];
```
yannickcr: Same as in `RoutingManager.ts` (`private` then `// ts-ignore` in tests`)

src/lib/routers/history.ts
```diff
@@ -76,12 +76,12 @@ class BrowserHistory {
    * Creates a full URL based on the route state.
    * The storage adaptor maps all syncable keys to the query string of the URL.
    */
-  private readonly _createURL: BrowserHistoryProps['createURL'];
+  public readonly _createURL: BrowserHistoryProps['createURL'];
```
yannickcr: Same as in `RoutingManager.ts` (`private` then `// ts-ignore` in tests`)

src/lib/routers/history.ts
```diff
@@ -179,7 +179,7 @@ class BrowserHistory {
   /**
    * Removes the event listener and cleans up the URL.
    */
-  public dispose(): void {
+  public dispose(): undefined {
```
yannickcr: I pretty sure that you should keep `void` here and that it is the `Router.dispose` signature in `src/types/instantsearch.ts` that is incorrect.

I think this is what you wanted:

```diff
 export type Router<TRouteState = UiState> = {
   onUpdate(callback: (route: TRouteState) => void): void;
   read(): UiState;
   write(route: TRouteState): void;
   createURL(state: TRouteState): string;
-  dispose(): Widget['dispose'];
+  dispose: Widget['dispose'];
 };
```

Then there is multiple differences between `Router.dispose` and `Widget.dispose`: 

* mandatory on `Router` and optional on `Widget`
* `options` parameter on `Widget` and no parameters on `Router`

So the final definition should be more like:

```diff
 export type Router<TRouteState = UiState> = {
   onUpdate(callback: (route: TRouteState) => void): void;
   read(): UiState;
   write(route: TRouteState): void;
   createURL(state: TRouteState): string;
-  dispose(): Widget['dispose'];
+  dispose: SearchParameters | void;
 };
```

WDYT?

francoischalifour: You're right, this is not optimal. I tried another approach in 611090bff85eb37e8b5f8d99dbc65d2cb3561fa6: it's made more generic and the `Router` extends the `Widget` type. I believe this is more accurate. Let me know what you think.

yannickcr: LGTM :+1:

PR #3763, opened 2019-05-07
fix(types): improve UiState types

francoischalifour:
This adds types for the UI state.

### DIFF
--- a/src/types/widget.ts
+++ b/src/types/widget.ts
@@ -31,7 +31,45 @@ interface DisposeOptions {
 }
 
 export type UiState = {
-  [stateKey: string]: any;
+  query?: string;
+  menu?: {
+    [attribute: string]: string;
+  };
+  /**
+   * The list of hierarchical menus.
+   * Nested levels must contain the record separator.
+   *
+   * @example ['Audio', 'Audio > Headphones']
+   */
+  hierarchicalMenu?: {
+    [attribute: string]: string[];
+  };
+  refinementList?: {
+    [attribute: string]: string[];
+  };
+  numericRefinementList?: {
+    [attribute: string]: number;
+  };
+  numericSelector?: {
+    [attribute: string]: number;
+  };
+  /**
+   * The range as a tuple.
+   *
+   * @example '100:500'
+   */
+  range?: {
+    [attribute: string]: string;
+  };
+  starRating?: {
+    [attribute: string]: number;
+  };
+  toggle?: {
+    [attribute: string]: boolean;
+  };
+  sortBy?: string;
+  page?: number;
+  hitsPerPage?: number;
 };
 
 export interface Widget {
--- a/src/lib/RoutingManager.ts
+++ b/src/lib/RoutingManager.ts
@@ -61,7 +61,7 @@ class RoutingManager implements Widget {
     const widgets = this.instantSearchInstance.widgets;
     const helper = this.instantSearchInstance.helper!;
 
-    return widgets.reduce((state, widget) => {
+    return widgets.reduce<UiState>((state, widget) => {
       if (!widget.getWidgetState) {
         return state;
       }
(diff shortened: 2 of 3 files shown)


src/types/widget.ts
```diff
@@ -31,7 +31,45 @@ interface DisposeOptions {
 }
 
 export type UiState = {
-  [stateKey: string]: any;
+  query?: string;
+  menu?: {
```
Haroenv: I'm not sure this should be hardcoded though 🤔

Haroenv: Would be more doing something should be imported from the widgets instead

francoischalifour: Do you mean that the type should be extended when a widget is added to InstantSearch?

Haroenv: what I mainly mean is that this type could easily be "lying" if there's new widgets or change in the things it exposes, this will easily be missed in this type. Maybe it needs tests

francoischalifour: The type won't be "lying" because it's checked in each `getWidgetState()` connector method:

https://github.com/algolia/instantsearch.js/blob/adf8220064dd46b5b6d6b06ef0e858bf476f219b/src/connectors/infinite-hits/connectInfiniteHits.ts#L226-L237

If you were to create another widget, leading to the creation of a new key in the UiState, this would be easily spotted.

eunjae-lee: @Haroenv could you help me understand what you mean? This is what I think. Let me know if I'm wrong.

If there's new widgets or change in already existing widgets, when we try to read/write something to/from uiState, the editor will show type error message. Then we realise the type is outdated and can update it to whatever we want it to be.

Haroenv: good for me then I guess, but what about custom widgets in TS (later) who can't extend this type?

eunjae-lee: That's a good point. When user creates a custom widget in their own project, can they extend this type, add their things, and overwrite it somehow (like in vue-shims.d.ts) ?

francoischalifour: That would be a good way to handle this in the future, indeed!

For now this PR is still an improvement over what we have right now and makes our data flow easier to understand.

eunjae-lee: Right. We can proceed like this, and in the future users can extend this type if they build their own custom widgets in their code base like [this](https://stackoverflow.com/questions/43952198/how-to-override-or-extend-a-libary-type-definition-in-typescript).

src/types/widget.ts
```diff
@@ -31,7 +31,45 @@ interface DisposeOptions {
...
+   *
+   * @example ['Audio', 'Audio > Headphones']
+   */
+  hierarchicalMenu?: {
+    [attribute: string]: string[];
+  };
+  refinementList?: {
+    [attribute: string]: string[];
+  };
+  numericRefinementList?: {
```
samouss: We don't have a `numericRefinementList` widget.

src/types/widget.ts
```diff
@@ -31,7 +31,45 @@ interface DisposeOptions {
...
+  };
+  /**
+   * The range as a tuple.
+   *
+   * @example '100:500'
+   */
+  range?: {
+    [attribute: string]: string;
+  };
+  starRating?: {
```
samouss: The name is `ratingMenu` not `starRating`.

src/types/widget.ts
```diff
@@ -31,7 +31,45 @@ interface DisposeOptions {
 }
 
 export type UiState = {
```
samouss: The type does not contain: `geoSearch`, `numericMenu`.

src/types/widget.ts
```diff
@@ -31,7 +31,45 @@ interface DisposeOptions {
...
+  hierarchicalMenu?: {
+    [attribute: string]: string[];
+  };
+  refinementList?: {
+    [attribute: string]: string[];
+  };
+  numericRefinementList?: {
+    [attribute: string]: number;
+  };
+  numericSelector?: {
```
samouss: We don't have a `numericSelector` widget.

PR #3765, opened 2019-05-10
chore(jest): put config top-level

Haroenv:
Some tools (like the jest plugin for vscode) otherwise would need to be configured for this path too, which is a bit of a hassle.

The other option is of course to set those settings in the workspace, but it's easy to forget since we don't all use the tool.


francoischalifour:
Wouldn't renaming `jest/config.js` to `jest/jest.config.js` be enough? ([see reference](https://github.com/jest-community/vscode-jest#how-to-start-the-jest))

Haroenv:
No, it works the same as if you didn't pass a `--config` argument to jest.

PR #3774, opened 2019-05-14
refactor(helper): replace .getState() by .state

Haroenv:
Since https://github.com/algolia/algoliasearch-helper-js/pull/707 getState is the same as fetching state, we decided we want to get rid of it. This makes sure InstantSearch will be ready for that change


francoischalifour:
To avoid later confusions we can rename the `helper` instances in [this file](https://github.com/algolia/instantsearch.js/blob/57b3269c350bd7d46b2174003b0cd764eb848eaf/src/lib/voiceSearchHelper/__tests__/index-test.ts) to `voiceSearchHelper` (same with `getHelper` to `getVoiceSearchHelper`) because it exposes a `getState` method.

Haroenv:
yes, was also confused by why it broke a test first

Haroenv:
As discussed, the type definition can stay as the one on DefinitelyTyped for now, we can see how we fix this the easiest once we start releasing

PR #3777, opened 2019-05-14
fix(types): update UiState type

francoischalifour:
This updates the `UiState` type as addressed in https://github.com/algolia/instantsearch.js/pull/3763#pullrequestreview-237282004.

The keys were taken from the current documentation which is outdated.

### DIFF
--- a/src/types/widget.ts
+++ b/src/types/widget.ts
@@ -32,6 +32,9 @@ interface DisposeOptions {
 
 export type UiState = {
   query?: string;
+  refinementList?: {
+    [attribute: string]: string[];
+  };
   menu?: {
     [attribute: string]: string;
   };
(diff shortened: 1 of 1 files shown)


src/types/widget.ts
```diff
@@ -61,12 +61,18 @@ export type UiState = {
...
   };
-  starRating?: {
-    [attribute: string]: number;
-  };
   toggle?: {
     [attribute: string]: boolean;
   };
+  geoSearch?: {
+    /**
+     * The rectangular area in geo coordinates.
```
francoischalifour: The description is taken from the [`insideBoundingBox`](https://www.algolia.com/doc/api-reference/api-parameters/insideBoundingBox/?language=javascript) API ref.

src/types/widget.ts
```diff
@@ -44,13 +47,10 @@ export type UiState = {
   hierarchicalMenu?: {
     [attribute: string]: string[];
   };
-  refinementList?: {
-    [attribute: string]: string[];
-  };
-  numericRefinementList?: {
+  numericMenu?: {
     [attribute: string]: number;
```
samouss: If I'm not mistaken the `numericMenu` takes a `string` not a `number`.

francoischalifour: Ah, you're right.

PR #3778, opened 2019-05-15
fix(types): improve types for voiceSearch

eunjae-lee:
**Summary**

This makes the types of the two things non-optional.

* `isSpeechFinal`
* `transcript`

The related discussion is [here](https://github.com/algolia/react-instantsearch/pull/2316#discussion_r283886705)


eunjae-lee:
I'm doing the change here first before React. @samouss

PR #3779, opened 2019-05-15
fix(voiceSearch): remove event listeners on dispose

eunjae-lee:
**Summary**

This removes event listeners attached to SpeechRecognition instance when the widget is disposed.
The related discussion is [here](https://github.com/algolia/react-instantsearch/pull/2316#discussion_r283821410).

**Result**

It shouldn't change any user behaviour.

### DIFF
--- a/src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
+++ b/src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
@@ -78,14 +78,14 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-searc
       );
     });
 
-    it('calls unmount on disposal', () => {
+    it('calls unmount on `dispose`', () => {
       const { unmountFn, widget, helper } = getDefaultSetup();
       widget.init({ helper });
       widget.dispose({ helper, state: helper.state });
       expect(unmountFn).toHaveBeenCalledTimes(1);
     });
 
-    it('removes event listeners on disposal', () => {
+    it('removes event listeners on `dispose`', () => {
       const { widget, helper } = getDefaultSetup();
       widget.init({ helper });
       widget.dispose({ helper, state: helper.state });
--- a/src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
+++ b/src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
@@ -8,7 +8,7 @@ jest.mock('../../../lib/voiceSearchHelper', () => {
       isBrowserSupported: () => true,
       isListening: () => false,
       toggleListening: () => {},
-      removeEventListeners: jest.fn(),
+      dispose: jest.fn(),
       // ⬇️ for test
       changeState: () => onStateChange(),
       changeQuery: query => onQueryChange(query),
@@ -78,20 +78,18 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-searc
       );
     });
 
-    it('calls unmount on dispose', () => {
+    it('calls unmount on disposal', () => {
       const { unmountFn, widget, helper } = getDefaultSetup();
       widget.init({ helper });
       widget.dispose({ helper, state: helper.state });
       expect(unmountFn).toHaveBeenCalledTimes(1);
     });
 
-    it('remove event listeners on dispose', () => {
+    it('removes event listeners on disposal', () => {
       const { widget, helper } = getDefaultSetup();
       widget.init({ helper });
       widget.dispose({ helper, state: helper.state });
-      expect(
-        widget._voiceSearchHelper.removeEventListeners
-      ).toHaveBeenCalledTimes(1);
+      expect(widget._voiceSearchHelper.dispose).toHaveBeenCalledTimes(1);
     });
   });
 
--- a/src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
+++ b/src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
@@ -8,6 +8,7 @@ jest.mock('../../../lib/voiceSearchHelper', () => {
       isBrowserSupported: () => true,
       isListening: () => false,
       toggleListening: () => {},
+      removeEventListeners: jest.fn(),
       // ⬇️ for test
       changeState: () => onStateChange(),
       changeQuery: query => onQueryChange(query),
@@ -82,6 +83,15 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-searc
       widget.init({ helper });
       widget.dispose({ helper, state: helper.state });
       expect(unmountFn).toHaveBeenCalledTimes(1);
+    });
+
+    it('remove event listeners on dispose', () => {
+      const { widget, helper } = getDefaultSetup();
+      widget.init({ helper });
+      widget.dispose({ helper, state: helper.state });
+      expect(
+        widget._voiceSearchHelper.removeEventListeners
+      ).toHaveBeenCalledTimes(1);
     });
   });
 
--- a/src/connectors/voice-search/connectVoiceSearch.ts
+++ b/src/connectors/voice-search/connectVoiceSearch.ts
@@ -116,8 +116,8 @@ const connectVoiceSearch: VoiceSearchConnector = (
         });
       },
       dispose({ state }) {
+        (this as any)._voiceSearchHelper.dispose();
         unmountFn();
-        (this as any)._voiceSearchHelper.removeEventListeners();
         return state.setQuery('');
       },
       getWidgetState(uiState, { searchParameters }) {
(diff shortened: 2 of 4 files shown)


src/lib/voiceSearchHelper/__tests__/index-test.ts
```diff
@@ -79,9 +85,9 @@ describe('VoiceSearchHelper', () => {
     voiceSearchHelper.toggleListening();
     expect(onStateChange).toHaveBeenCalledTimes(1);
     expect(voiceSearchHelper.getState().status).toEqual('askingPermission');
-    recognition.onstart();
+    listeners.start();
```
samouss: We could add a comment to inform that we simulate the `start` event of the `SpeechRecognition`.

src/lib/voiceSearchHelper/__tests__/index-test.ts
```diff
@@ -39,38 +29,54 @@ describe('VoiceSearchHelper', () => {
...
   });
 
   it('works with mock SpeechRecognition (searchAsYouSpeak:false)', () => {
-    let recognition;
+    const listeners: any = {};
     window.SpeechRecognition = jest.fn().mockImplementation(() => ({
-      start() {
-        /* eslint-disable-next-line consistent-this */
-        recognition = this;
+      start() {},
```
samouss: We re-use the exact same implementation multiple times inside the test. We could have a function that create the mock version of `SpeechRecognition` (like we do for [the client or the helper](https://github.com/algolia/instantsearch.js/blob/61ae52c9b1b29f812844c00d1b44eca3be929a9f/src/connectors/query-rules/__tests__/connectQueryRules-test.ts#L31-L43)). You can store the [`listeners`](https://github.com/algolia/instantsearch.js/pull/3779/files#diff-7fd428b5c85af5b43d66125872408774R70) directly on the `SpeechRecognition` mock to avoid to return multiple values from the function.

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -117,6 +117,7 @@ const connectVoiceSearch: VoiceSearchConnector = (
       },
       dispose({ state }) {
         unmountFn();
+        (this as any)._voiceSearchHelper.removeEventListeners();
```
samouss: We should also stop listening once we dispose the widget.

src/lib/voiceSearchHelper/__tests__/index-test.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/3779/commits/c611f9dddb53d997cc210f077e28712d399fa88d

src/lib/voiceSearchHelper/__tests__/index-test.ts
eunjae-lee: I'm not sure if [this](https://github.com/algolia/instantsearch.js/pull/3779/commits/aa4f1bbb9540fcd8b86d1577e43d42c6600e09fe) is what you meant by storing the listeners directly on the SpeechRecognition mock.

src/connectors/voice-search/connectVoiceSearch.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/3779/commits/d95a9bcaf0f77b357b5b649272aa96ef193297f4

src/lib/voiceSearchHelper/__tests__/index-test.ts
```diff
@@ -85,7 +85,7 @@ describe('VoiceSearchHelper', () => {
     voiceSearchHelper.toggleListening();
     expect(onStateChange).toHaveBeenCalledTimes(1);
     expect(voiceSearchHelper.getState().status).toEqual('askingPermission');
-    listeners.start();
+    listeners.start(); // This way, we're simulating `start` event of the SpeechRecognition.
```
samouss: I didn't mention it into my original comment, but it worth adding it on the others listeners calls.

src/lib/voiceSearchHelper/index.ts
```diff
@@ -119,14 +119,17 @@ export default function voiceSearchHelper({
...
     if (!recognition) {
       return;
     }
+    recognition.stop();
     recognition.removeEventListener('start', onStart);
     recognition.removeEventListener('error', onError);
     recognition.removeEventListener('result', onResult);
     recognition.removeEventListener('end', onEnd);
+    recognition = undefined;
+    resetState();
```
samouss: Not sure the `resetState` call is useful, the Helper is destroyed with the widget anyway.

src/lib/voiceSearchHelper/__tests__/index-test.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/3779/commits/af65fcd0e9c1ce8122a3e1f39a3aaaa2ef68c97a

src/lib/voiceSearchHelper/index.ts
eunjae-lee: I wasn't sure about this. I've removed it now and changed the way to test it.
https://github.com/algolia/instantsearch.js/pull/3779/commits/ae1f20e93ceffacbd4e67be83b6baaefae858f24

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -77,12 +78,19 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-searc
       );
     });
 
-    it('calls unmount on dispose', () => {
+    it('calls unmount on disposal', () => {
```
Haroenv: reads pretty weird, I think we can keep `dispose` (maybe even in backticks to signify the fact that it's a function)

src/lib/voiceSearchHelper/__tests__/index-test.ts
```diff
@@ -39,38 +46,49 @@ describe('VoiceSearchHelper', () => {
...
   it('works with mock SpeechRecognition (searchAsYouSpeak:false)', () => {
-    let recognition;
-    window.SpeechRecognition = jest.fn().mockImplementation(() => ({
-      start() {
-        /* eslint-disable-next-line consistent-this */
-        recognition = this;
-      },
-    }));
+    window.SpeechRecognition = createFakeSpeechRecognition();
+    const { listeners } = window.SpeechRecognition as any;
```
Haroenv: I think the comments might become redundant if we call this `simulateListener`

src/lib/voiceSearchHelper/__tests__/index-test.ts
```diff
@@ -189,6 +192,6 @@ describe('VoiceSearchHelper', () => {
     });
     voiceSearchHelper.toggleListening();
     voiceSearchHelper.dispose();
-    expect(voiceSearchHelper.isListening()).toBe(false);
```
samouss: What was wrong with this assertion? IMO it was better because we weren't relying on implementation detail.

src/lib/voiceSearchHelper/__tests__/index-test.ts
```diff
@@ -8,11 +8,14 @@ declare global {
...
 }
 
+const start = jest.fn();
+const stop = jest.fn();
+
 const createFakeSpeechRecognition = (): jest.Mock => {
   const listeners: any = {};
   const mock = jest.fn().mockImplementation(() => ({
-    start() {},
-    stop() {},
```
samouss: What was wrong with this?

src/lib/voiceSearchHelper/__tests__/index-test.ts
eunjae-lee: `isListening` relies on `voiceSearchHelper.getState().status`.
Since we're not updating the `state` of `voiceSearchHelper` on `dispose`, the `status` will remain as `recognizing` which results `isListening` returning `true`. And that way we cannot test this. So I had to check if `stop` has been called.

src/lib/voiceSearchHelper/__tests__/index-test.ts
eunjae-lee: I had to make them a mock function due to [this](https://github.com/algolia/instantsearch.js/pull/3779/files/af65fcd0e9c1ce8122a3e1f39a3aaaa2ef68c97a..ae1f20e93ceffacbd4e67be83b6baaefae858f24#r284716504)

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
eunjae-lee: haha okay. I've reverted it back to disposal with backticks 👍 
https://github.com/algolia/instantsearch.js/pull/3779/commits/8aaab1c492e7d76e7af4b603a4bb33f5a2b74ab3

src/lib/voiceSearchHelper/__tests__/index-test.ts
eunjae-lee: That's brilliant.
https://github.com/algolia/instantsearch.js/pull/3779/commits/6bd29b40d93d1f8fa38d8972e9e0a949fd6523e8

PR #3783, opened 2019-05-17
fix(types): fix cssClasses of voiceSearch

eunjae-lee:
**Summary**

This allows the widget can accept an array of string as css classes.
This was missing in the first place, allowing only strings.


eunjae-lee:
fixing the lint errors now..

eunjae-lee:
Here's a headache. The reason why the lint failed is because the type of `cssClasses` changes.

At the widget, `cssClasses` should be

```js
export type VoiceSearchCSSClasses = {
  root: string | string[];
  button: string | string[];
  status: string | string[];
};
```

At the component, `cssClasses` should be

```js
export type VoiceSearchComponentCSSClasses = {
  root: string;
  button: string;
  status: string;
};
```

[Here](https://github.com/algolia/instantsearch.js/blob/c4f3e5ba83aef2cf63de53a1034ae8ad13fdab11/src/widgets/voice-search/voice-search.tsx#L91), `cssClasses` are being modified from `VoiceSearchCSSClasses` to `VoiceSearchComponentCSSClasses`, but at the [renderer](https://github.com/algolia/instantsearch.js/blob/c4f3e5ba83aef2cf63de53a1034ae8ad13fdab11/src/widgets/voice-search/voice-search.tsx#L66) it's still inferred as `VoiceSearchCSSClasses` because of the type of the widget params.

So as a workaround, I casted `cssClasses` to `VoiceSearchComponentCSSClasses` at rendering the component.

Do you have any better idea?

src/widgets/voice-search/voice-search.tsx
```diff
@@ -61,7 +63,7 @@ const renderer: VoiceSearchRenderer<VoiceSearchRendererWidgetParams> = ({
 
   render(
     <VoiceSearch
-      cssClasses={cssClasses}
+      cssClasses={cssClasses as VoiceSearchComponentCSSClasses}
```
francoischalifour: You don't need to cast `cssClasses` here if you change the type of `VoiceSearchRendererWidgetParams['cssClasses']` to `VoiceSearchComponentCSSClasses`.

eunjae-lee: @francoischalifour If I don't cast it, it doesn't pass `yarn type-check`.

image

francoischalifour: It should work if you change [this line](https://github.com/algolia/instantsearch.js/blob/c4f3e5ba83aef2cf63de53a1034ae8ad13fdab11/src/widgets/voice-search/voice-search.tsx#L49):

```diff
interface VoiceSearchRendererWidgetParams extends VoiceSearchWidgetParams {
  container: HTMLElement;
- cssClasses: VoiceSearchCSSClasses;
+ cssClasses: VoiceSearchComponentCSSClasses;
  templates: VoiceSearchTemplates;
}
```

eunjae-lee: Right. That's where I can change the type! I was lost in params hell. Thanks for saving me.

PR #3793, opened 2019-05-20
refactor(helper): replace getQueryParameter usage [PART-2]

samouss:
The PR replaces the usage of the [`getQueryParameter`](https://github.com/algolia/algoliasearch-helper-js/blob/develop/src/SearchParameters/index.js#L1552-L1560) function. This function rely on the previous structure of the `SearchParameters`. It does check if the given attribute exists on the state, if it doesn't it throws an error. This function will be removed from the [Helper](https://github.com/algolia/algoliasearch-helper-js/pull/713). We've replace the call to a direct access to the state. You can verify that the tests pass for the changes with (note that the tests fail but it's only snapshots):

```
yarn test \
src/connectors/hierarchical-menu/__tests__/connectHierarchicalMenu-test.js \
src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.js \
src/connectors/refinement-list/__tests__/connectRefinementList-test.js \
src/connectors/sort-by/__tests__/connectSortBy-test.js
```


tkrugg:
side question: is `setQueryParameter` kept in helper ?

samouss:
@tkrugg Yes `setQueryParameter` is kept inside both Helper & `SearchParameters`. The reason is that it does some additional checks (like [shallow equality](https://github.com/algolia/algoliasearch-helper-js/blob/ffa13f45dd4a71d0a148ecd72fa1da6fb8da4c4f/src/SearchParameters/index.js#L1323-L1331)) to avoid creating a new instance for no reason. Its usage is not equivalent to: `helper.state.value = 'myValue';`. On the other hand, the usage of the getter is equivalent.

PR #3794, opened 2019-05-20
refactor(helper): update tests [PART-3]

samouss:
This PR mostly update the snapshots of the `SearchParameters`. We don't have [default values](https://github.com/algolia/algoliasearch-helper-js/pull/670) anymore which means that the all the snapshots are outdated. It also updates tests that was relying on this behavior.


samouss:
@Haroenv Nope the inline snapshots require Prettier and we don't have it set up on the project.

src/connectors/hierarchical-menu/__tests__/__snapshots__/connectHierarchicalMenu-test.js.snap
```diff
@@ -39,36 +24,10 @@ SearchParameters {
       "path",
     ],
   },
-  "highlightPostTag": undefined,
-  "highlightPreTag": undefined,
-  "hitsPerPage": undefined,
-  "ignorePlurals": undefined,
   "index": "",
```
Haroenv: is it expected that index is still defined?

src/widgets/current-refinements/__tests__/__snapshots__/current-refinements-test.js.snap
```diff
@@ -53,6 +38,21 @@ exports[`currentRefinements() render() DOM output renders correctly 1`] = `
           },
         ],
       },
+      Object {
```
Haroenv: is this different order expected, or something which changed in the helper?

francoischalifour:
@samouss The InstantSearch.js setup supports Prettier. We use inline snapshots in [some tests](https://github.com/algolia/instantsearch.js/blob/f4f05bdc30e3e6a88d6761731b303e85cbf3501b/src/helpers/__tests__/highlight-test.js#L77) already. I think here we should use objects comparison with `toEqual` now that the search parameters are less noisy. Bouncing back and forth from test files to snapshots is a pain.

samouss:
@francoischalifour Yep, my bad I thought I was in the Helper...  I agree on the idea but I don't think it's a good idea to do it inside this PR though. We can improve that latter.

src/widgets/current-refinements/__tests__/__snapshots__/current-refinements-test.js.snap
Haroenv: perfect! I'm glad you can see a change like that reflected in these tests.

Haroenv:
looks like the history is wrong here too

samouss:
@Haroenv Yes because I'm rebasing the PR that's why.

PR #3796, opened 2019-05-20
refactor(helper): update listeners [PART-4]

samouss:
This PR updates how we subscribe/use the listeners on the Helper. We moved away from [a list of arguments to an object](https://github.com/algolia/algoliasearch-helper-js/pull/700). The PR address those changes on the listeners. It also updates how the `error` event is emitted at the InstantSearch level. It now to match the shape of the Helper.

```js
// With list of arguments
search.on('error', error => {
  console.log(error);
});

// With an object
search.on('error', event => {
  console.log(event.error);
});
```

### DIFF
--- a/src/connectors/query-rules/connectQueryRules.ts
+++ b/src/connectors/query-rules/connectQueryRules.ts
@@ -18,6 +18,10 @@ import {
   NumericRefinement as InternalNumericRefinement,
 } from '../../lib/utils/getRefinements';
 
+type HelperChangeEvent = {
+  state: SearchParameters;
+};
+
 type TrackedFilterRefinement = string | number | boolean;
 
 export type ParamTrackedFilters = {
@@ -126,7 +130,7 @@ function applyRuleContexts(
     trackedFilters: ParamTrackedFilters;
     transformRuleContexts: ParamTransformRuleContexts;
   },
-  sharedHelperState: SearchParameters
+  event: HelperChangeEvent
 ): void {
   const {
     helper,
@@ -135,6 +139,7 @@ function applyRuleContexts(
     transformRuleContexts,
   } = this;
 
+  const sharedHelperState = event.state;
   const previousRuleContexts: string[] = sharedHelperState.ruleContexts || [];
   const newRuleContexts = getRuleContextsFromTrackedFilters({
     helper,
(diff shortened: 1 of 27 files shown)


src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -207,11 +212,13 @@ const connectQueryRules: QueryRulesConnector = (render, unmount = noop) => {
...
             Boolean(widgetParams.transformRuleContexts)
           ) {
-            onHelperChange(state);
+            onHelperChange({ state });
           }
 
           // We track every change in the helper to override its state and add
           // any `ruleContexts` needed based on the `trackedFilters`.
+          // @ts-ignore
+          // @TODO: we have to update the definition to make it work
```
Haroenv: We could ship the d.ts with the helper & make the modifications?

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -18,6 +18,10 @@ import {
   NumericRefinement as InternalNumericRefinement,
 } from '../../lib/utils/getRefinements';
 
+type HelperChangeEvent = {
```
francoischalifour: Since we'll reuse this type we can store it in [instantsearch.ts](https://github.com/algolia/instantsearch.js/blob/6b64d6b0dd2153b5f40fb0121e3a19224c011420/src/types/instantsearch.ts).

src/connectors/query-rules/connectQueryRules.ts
francoischalifour: An alternative is to update the type we use in InstantSearch: https://github.com/algolia/instantsearch.js/blob/6b64d6b0dd2153b5f40fb0121e3a19224c011420/src/types/instantsearch.ts#L109

samouss: I didn't planed to do it in this PR at first, but make sense to do it now in fact. Will update the PR.

src/connectors/query-rules/connectQueryRules.ts
samouss: https://github.com/algolia/instantsearch.js/pull/3796/commits/d0ac107ccea87dd53238b5210e719a2e77a6265c

PR #3797, opened 2019-05-21
refactor(helper): update page usage [PART-5]

samouss:
This PR updates the usage of the `page` attribute from the `SearchParameters`. Its value is now optional which implies that we have to fall back on a default value at the connector level. The default value is not always required though, it depends on the use case. I've tried to make as few changes as possible to stick with the current behavior.

### DIFF
--- a/src/connectors/infinite-hits/connectInfiniteHits.ts
+++ b/src/connectors/infinite-hits/connectInfiniteHits.ts
@@ -146,8 +146,8 @@ const connectInfiniteHits: InfiniteHitsConnector = (
       init({ instantSearchInstance, helper }) {
         showPrevious = getShowPrevious(helper);
         showMore = getShowMore(helper);
-        firstReceivedPage = helper.state.page!;
-        lastReceivedPage = helper.state.page!;
+        firstReceivedPage = helper.state.page || 0;
+        lastReceivedPage = helper.state.page || 0;
 
         renderFn(
           {
@@ -171,11 +171,12 @@ const connectInfiniteHits: InfiniteHitsConnector = (
         // We're doing this to "reset" the widget if a refinement or the
         // query changes between renders, but we want to keep it as is
         // if we only change pages.
-        const { page, ...currentState } = state;
+        const { page = 0, ...currentState } = state;
+
         if (!isEqual(currentState, prevState)) {
           hitsCache = [];
-          firstReceivedPage = page!;
-          lastReceivedPage = page!;
+          firstReceivedPage = page;
+          lastReceivedPage = page;
           prevState = currentState;
         }
 
@@ -193,12 +194,12 @@ const connectInfiniteHits: InfiniteHitsConnector = (
 
         results.hits = transformItems(results.hits);
 
-        if (lastReceivedPage < page! || !hitsCache.length) {
+        if (lastReceivedPage < page || !hitsCache.length) {
           hitsCache = [...hitsCache, ...results.hits];
-          lastReceivedPage = page!;
-        } else if (firstReceivedPage > page!) {
+          lastReceivedPage = page;
+        } else if (firstReceivedPage > page) {
           hitsCache = [...results.hits, ...hitsCache];
-          firstReceivedPage = page!;
+          firstReceivedPage = page;
         }
 
         const isFirstPage = firstReceivedPage === 0;
@@ -224,7 +225,7 @@ const connectInfiniteHits: InfiniteHitsConnector = (
       },
 
       getWidgetState(uiState, { searchParameters }) {
-        const page = searchParameters.page!;
+        const page = searchParameters.page || 0;
 
         if (!hasShowPrevious || page === 0 || page + 1 === uiState.page) {
           return uiState;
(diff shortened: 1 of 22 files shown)


src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -240,11 +241,14 @@ const connectInfiniteHits: InfiniteHitsConnector = (
...
         }
+
         const uiPage = uiState.page;
+
         if (uiPage) {
           return searchParameters.setQueryParameter('page', uiPage - 1);
         }
-        return searchParameters.setQueryParameter('page', 0);
+
+        return searchParameters.setQueryParameter('page');
```
Haroenv: I wonder if it's worth explicitly passing `undefined` as second argument. Not sure how consistent we are about that in the code base

src/connectors/pagination/connectPagination.js
```diff
@@ -165,9 +170,12 @@ export default function connectPagination(renderFn, unmountFn) {
...
       getWidgetSearchParameters(searchParameters, { uiState }) {
         const uiPage = uiState.page;
-        if (uiPage)
+
+        if (uiPage) {
           return searchParameters.setQueryParameter('page', uiState.page - 1);
-        return searchParameters.setQueryParameter('page', 0);
+        }
+
+        return searchParameters.setQueryParameter('page');
```
Haroenv: this implies we no longer need `setPage` on the helper right?

src/types/instantsearch.ts
```diff
@@ -9,10 +9,13 @@ import { Widget, UiState } from './widget';
...
 export type InstantSearchOptions = any;
 
-// That's a proxy to avoid manipulating the original `algoliasearch-helper` SearchParameters
-// typings and to add newer search parameters not yet documented.
+// That's a proxy to avoid manipulating the original `algoliasearch-helper`
+// SearchParameters typings and to add newer search parameters not yet
+// documented or wrongly typed.
 export type SearchParameters = AlgoliaSearchHelperSearchParameters & {
   ruleContexts?: string[];
+  // The value is optional not required
```
francoischalifour:
```suggestion
  // The value is optional
```

src/connectors/infinite-hits/connectInfiniteHits.ts
francoischalifour: Passing `undefined` would indeed be more explicit. We could perhaps enforce that by removing the `?` optional type in the added definition.

samouss: This part is not exactly true. It doesn't matter much at the moment because it will be updated once we do the modifications inside `getConfiguration` to "mount" a default value on the state that indicates that a widget "own" this part of the state.

src/connectors/pagination/connectPagination.js
samouss: https://github.com/algolia/instantsearch.js/pull/3797/files#r286121194

samouss:
I'll change the PR to have the exact same behaviour as today, it's slightly different at the moment.

src/connectors/infinite-hits/connectInfiniteHits.ts
samouss: IMO the argument `undefined` is useless and might be even counter-intuitive. Here, the intent is to omit the value on the `SearchParameters` which is what the call translate. With the value `undefined` we might expect that the final object will actually contain `undefined` which is not the case. I'm fine with both solution though, if you both agree that it's better I'll update.

src/connectors/pagination/connectPagination.js
samouss: We never _really_ have the need for `setPage` since it forwards the call. The removal of those methods could be part of the process to remove the unused methods of the Helper.

src/types/instantsearch.ts
samouss: https://github.com/algolia/instantsearch.js/pull/3797/commits/3b15222a9dd803318be0faa91a4b493171d6a624

src/connectors/infinite-hits/connectInfiniteHits.ts
francoischalifour: I'm not used to seeing setters without values, hence my reply. We could have a method in the helper called `resetQueryParameter(parameter: string)` which would make it clearer. In the end, it doesn't really matter.

samouss: Ok, I will update the call. We should avoid creating methods on the helper that don't have "real" value. Otherwise, we have an API surface that became very broad. Like what we have at the moment. By multiplying the methods we might also introduce multiple ways to do the exact same operation, which doesn't help for consistency.

samouss: https://github.com/algolia/instantsearch.js/pull/3797/commits/cb6ea7153e18d948620a65a902dcda91642b40e8, https://github.com/algolia/instantsearch.js/pull/3797/commits/f7f239cdc17b838ccb856b045355d9c3498bfb9a

PR #3798, opened 2019-05-21
chore(examples): build examples before serving

francoischalifour:
This PR changes the flow of how the examples are built. This will be required for the new demos we're building.

## Before

- The examples used to be static and directly serve at e.g. https://instantsearchjs.netlify.com/examples/e-commerce.

- Those examples were then zipped to be downloaded at https://instantsearchjs.netlify.com/examples/e-commerce.zip.

## After

Since we're building new examples based on [Create InstantSearch App](https://github.com/algolia/create-instantsearch-app), these projects are bundled and the build files are contained in e.g. `e-commerce/dist`.

Before serving the examples, we're now running the `build` command of each example located in `/examples`, and copy the dist files in `/website/examples` so that the served URL remains unchanged. `/website/examples` files are generated and untracked on Git.

Since we're still in the process of creating the new demos, I added a `package.json` file to the old ones with a `build` script copying the content of the demo in a `dist` folder. This is temporary until we migrate all the examples.

__The zip functionality was removed because barely used and harder to put in place with the new system since several directories were to ignore (`.cache`, `dist` and `node_modules`).__

## Preview

- E-commerce demo: https://deploy-preview-3798--instantsearchjs.netlify.com/examples/e-commerce/


website/src/media/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "instantsearch-media",
+  "version": "1.0.0",
+  "private": true,
+  "scripts": {
+    "build": "rm -rf ./dist/ && rsync -r . ./dist/"
```
Haroenv: what's the reason for using rsync instead of cp here, is it faster?

We don't have an issue not supporting windows now, but that might cause issues later. For now I guess it's fine either way

francoischalifour: It worked well in that case. Those build scripts are temporary and should be run only via Netlify anyway.

PR #3799, opened 2019-05-21
chore(examples): add e-commerce widgets [part 1]

francoischalifour:
_This is part 1 of the series "e-commerce demo"._

## Description

This PR scaffolds the e-commerce demo with InstantSearch.js widgets. It was generated using [Create InstantSearch App](https://github.com/algolia/create-instantsearch-app).

## Notes

- Everything in the demo was achieved without connectors
- `algoliasearch` and `instantsearch.js` deps are added as dependencies to trigger less network requests (not loading from a CDN) and to leverage Renovate
- The ES build of `instantsearch.js` is used
- It doesn't include voice search because the designs are not done yet

## Preview

There's currently no styles in this part 1.

[See preview →](https://deploy-preview-3799--instantsearchjs.netlify.com/examples/e-commerce/)

## Related

- [E-commerce demo design](https://app.zeplin.io/project/5acc8becfc0c980068fc12cc/screen/5cd1a7992e4f02039b5a8336)

---

The styles will come in a stacked PR.


website/src/e-commerce/src/search.js
```diff
@@ -0,0 +1,46 @@
...
+
+const searchClient = algoliasearch(
+  'latency',
+  '6be0576ff61c053d5f9a3225e2a90f76'
+);
+
+const search = instantsearch({
+  searchClient,
+  indexName: 'instant_search',
+  routing: true,
```
francoischalifour: The routing will later be improved based on algolia/doc-code-samples#102.

website/src/e-commerce/src/widgets/PriceSlider.js
```diff
@@ -0,0 +1,13 @@
+import { panel, rangeSlider } from 'instantsearch.js/es/widgets';
+
+// The price range slider has a dummy panel to be collapsed
+// at the same time as `rangeInput`.
+const priceRangeSlider = panel({})(rangeSlider);
```
francoischalifour: As you can see in the [design](https://app.zeplin.io/project/5acc8becfc0c980068fc12cc/screen/5cd1a7992e4f02039b5a8336), the `rangeInput` is wrapped in the same panel as `rangeSlider`. However, InstantSearch.js only supports wrapping a single widget in a panel.

This will be required once the styles come.

website/src/e-commerce/src/widgets/index.js
```diff
@@ -0,0 +1,13 @@
+export { default as brands } from './Brands';
```
Haroenv: I'm not sure if this pattern is really necessary in an example like this

francoischalifour: This is fixed in 49167efc5e365472fd851b5971e7a233e6319c63.

francoischalifour: What would you have used? It's to export them all from `widgets/index.js`.

francoischalifour:
@Haroenv This is fixed in 49167efc5e365472fd851b5971e7a233e6319c63.

website/src/e-commerce/src/widgets/index.js
Haroenv: since all widgets are always mounted, I would've gone for the regular way of calling the functions inline in the main file. I understand why you split it up here, but I fear like it might confuse people

francoischalifour: The goal of the demo is also to educate people how to build apps with InstantSearch. I tried several patterns and this once seem like an easy structure to work with. It will be easy to migrate this demo to React/Vue/etc. 

We've often neglected how to integrate InstantSearch into apps and most of our examples don't make usage of modern JavaScript feature that we usually like working with.

eunjae-lee:
> @Haroenv This is fixed in [49167ef](https://github.com/algolia/instantsearch.js/commit/49167efc5e365472fd851b5971e7a233e6319c63).

I don't still see any style from https://deploy-preview-3799--instantsearchjs.netlify.com/examples/e-commerce/
Should I try different url?

website/src/e-commerce/.gitignore
```diff
@@ -1 +1,22 @@
+# See https://help.github.com/ignore-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+
+# testing
+/coverage
```
Haroenv: seems like most of these aren't necessary here

francoischalifour:
@eunjae-lee The style is in a next PR. This loaded the internal resources.

website/src/e-commerce/README.md
```diff
@@ -0,0 +1,21 @@
+# E-commerce demo
+
+_This project was generated with [create-instantsearch-app](https://github.com/algolia/create-instantsearch-app) by [Algolia](https://algolia.com)._
+
+## Get started
```
Haroenv: Shouldn't we mention production build here too? 🤔

website/src/e-commerce/src/search.js
```diff
@@ -0,0 +1,46 @@
...
+  priceSlider,
+  freeShipping,
+  ratings,
+  sortBy,
+  hitsPerPage,
+  products,
+  pagination,
+]);
+
+export default search;
```
Haroenv: since this isn't an entry point, I wonder if it should be wrapped into a function 🤔

website/src/e-commerce/src/widgets/Products.js
```diff
@@ -0,0 +1,36 @@
...
+      <span class="hit-em hit-rating">
+      <svg xmlns="http://www.w3.org/2000/svg" width="8" height="8" viewBox="0 0 16 16">
+        <path fill="#e2a400" fill-rule="evenodd" d="M10.472 5.008L16 5.816l-4 3.896.944 5.504L8 12.616l-4.944 2.6L4 9.712 0 5.816l5.528-.808L8 0z"/>
+      </svg>
+        {{rating}}
+      </span>
+    </p>
+  </footer>
+</article>
+      `,
```
Haroenv:
```suggestion
`,
```

I wonder if we should be indenting these instead?

website/src/e-commerce/src/search.js
eunjae-lee: You mean like `getSearch = () => ...` ?
Can you tell me more about this?
Does it have something to do with "entry point"?
In my last project, I had some issues exporting an object like that.
So I had to wrap it in a function but couldn't understand why I had to do it.

Haroenv: if you'd import this twice for two search instances next to each other (weird, but not impossible), these would be the same instance. A similar remark can be made for the widgets themselves, which is why I was more a fan of not splitting it up

website/src/e-commerce/README.md
francoischalifour: You mean explaining the build script? I'm open to suggestions for the README because I'd like to make it more exhaustive.

website/src/e-commerce/src/search.js
eunjae-lee: I get what you mean. Speaking of splitting it up, if we're going to wrap `search` instance in a function, IMO we need to wrap all the widgets in function too. Otherwise, wrapping `search` only doesn't make any change. And I'm not sure if we need to do that.

website/src/e-commerce/src/widgets/Products.js
francoischalifour: I don't really mind as long as we're consistent.

With time I came to prefer aligning them on the left because it's less indented on the right (easier to read on the IDE) and it's not really part of the JavaScript around.

francoischalifour:
@Haroenv 

> looks like text is overlapping, we probably don't want that, even if there's no styles

I think this is because of Rheostat.

website/src/e-commerce/src/search.js
francoischalifour: The reason why this search is in `search.js` and not directly in `app.js` is that `app.js` will grow with more JavaScript (UI features in #3802, sticky header is to come, etc.).

We can indeed export a `startSearch()` function and call it from `app.js`.

francoischalifour:
@eunjae-lee Those additions will be taken care of in the future because we don't have the behavior/design for now.

PR #3804, opened 2019-05-21
refactor(helper): update query usage [PART-6]

samouss:
This PR updates the usage of the `query` attribute from the `SearchParameters`. Its value is now optional which implies that we have to fall back on a default value at the connector level. The default value is not always required though, it depends on the use case. I've tried to make as few changes as possible to stick with the current behavior.

### DIFF
--- a/src/connectors/search-box/connectSearchBox.js
+++ b/src/connectors/search-box/connectSearchBox.js
@@ -149,7 +149,7 @@ export default function connectSearchBox(renderFn, unmountFn) {
       },
 
       getWidgetSearchParameters(searchParameters, { uiState }) {
-        return searchParameters.setQuery(uiState.query);
+        return searchParameters.setQueryParameter('query', uiState.query);
       },
     };
   };
--- a/src/connectors/search-box/connectSearchBox.js
+++ b/src/connectors/search-box/connectSearchBox.js
@@ -103,7 +103,7 @@ export default function connectSearchBox(renderFn, unmountFn) {
 
         renderFn(
           {
-            query: helper.state.query,
+            query: helper.state.query || '',
             refine: this._refine,
             clear: this._cachedClear,
             widgetParams,
@@ -118,7 +118,7 @@ export default function connectSearchBox(renderFn, unmountFn) {
 
         renderFn(
           {
-            query: helper.state.query,
+            query: helper.state.query || '',
             refine: this._refine,
             clear: this._cachedClear,
             widgetParams,
@@ -131,11 +131,12 @@ export default function connectSearchBox(renderFn, unmountFn) {
 
       dispose({ state }) {
         unmountFn();
+
         return state.setQuery('');
       },
 
       getWidgetState(uiState, { searchParameters }) {
-        const query = searchParameters.query;
+        const query = searchParameters.query || '';
 
         if (query === '' || (uiState && uiState.query === query)) {
           return uiState;
@@ -148,7 +149,7 @@ export default function connectSearchBox(renderFn, unmountFn) {
       },
 
       getWidgetSearchParameters(searchParameters, { uiState }) {
-        return searchParameters.setQuery(uiState.query || '');
+        return searchParameters.setQuery(uiState.query);
       },
     };
   };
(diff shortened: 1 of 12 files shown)


src/connectors/search-box/connectSearchBox.js
```diff
@@ -150,7 +153,7 @@ export default function connectSearchBox(renderFn, unmountFn) {
       },
 
       getWidgetSearchParameters(searchParameters, { uiState }) {
-        return searchParameters.setQuery(uiState.query || '');
+        return searchParameters.setQuery(uiState.query);
```
samouss: This part is not exactly true. It doesn't matter much at the moment because it will be updated once we do the modifications inside `getConfiguration` to "mount" a default value on the state that indicates that a widget "own" this part of the state.

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -133,7 +131,7 @@ const connectVoiceSearch: VoiceSearchConnector = (
         };
       },
       getWidgetSearchParameters(searchParameters, { uiState }) {
-        return searchParameters.setQuery(uiState.query || '');
+        return searchParameters.setQuery(uiState.query);
```
samouss: https://github.com/algolia/instantsearch.js/pull/3804/files#r286120796

samouss:
I'll change the PR to have the exact same behaviour as today, it's slightly different at the moment.

src/connectors/search-box/connectSearchBox.js
```diff
@@ -94,8 +94,10 @@ export default function connectSearchBox(renderFn, unmountFn) {
               previousQuery = helper.state.query;
               helper.setQuery(q);
             }
-            if (doSearch && previousQuery !== undefined && previousQuery !== q)
+
+            if (doSearch && previousQuery !== q) {
```
samouss: This part is going away once the branch is rebased with `next`.

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -82,10 +82,8 @@ const connectVoiceSearch: VoiceSearchConnector = (
               previousQuery = helper.state.query;
               helper.setQuery(query);
             }
-            if (
-              typeof previousQuery !== 'undefined' &&
-              previousQuery !== query
-            ) {
+
+            if (previousQuery !== query) {
```
samouss: https://github.com/algolia/instantsearch.js/pull/3804/files#r286342747

src/connectors/search-box/connectSearchBox.js
```diff
@@ -131,11 +131,12 @@ export default function connectSearchBox(renderFn, unmountFn) {
 
       dispose({ state }) {
         unmountFn();
+
         return state.setQuery('');
```
francoischalifour: We don't want to reset the query state to `undefined` on dispose?

stories/panel.stories.js
```diff
@@ -88,7 +88,7 @@ storiesOf('Panel', module)
     withHits(({ search, container, instantsearch }) => {
       const brandList = instantsearch.widgets.panel({
         collapsed: options => {
-          return options && options.state && options.state.query.length === 0;
+          return !((options && options.state && options.state.query) || '');
```
francoischalifour: This is slightly more complicated. Would this make sense?

```suggestion
          return (options && options.state && (!options.state.query || options.state.query === ''));
```

src/connectors/search-box/connectSearchBox.js
samouss: Yes we do. I didn't changed it (yet) because there are multiple places where this call is neither incorrect or not even implemented e.g `connectAutocomplete`. We'll create another round of PRs to align the calls to `dispose` and `getWidgetSearchParameters`.

stories/panel.stories.js
samouss: https://github.com/algolia/instantsearch.js/pull/3804/commits/80418777984a5366e30066b693d9098b39a715b6

PR #3815, opened 2019-05-27
feat(connectAutocomplete): clear the state on dispose

samouss:
This PR fix the implementation of the `dispose` method for the connector `connectAutocomplete`. It omits `query` & `highlightTags` once the widget is removed. I also cover the rest of the `dispose` function with tests.

### DIFF
--- a/src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
+++ b/src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
@@ -269,5 +269,30 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/autocomplet
 
       expect(nextState.query).toBeUndefined();
     });
+
+    it('removes the TAG_PLACEHOLDER from the `SearchParameters`', () => {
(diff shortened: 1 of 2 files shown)


src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
```diff
@@ -167,4 +167,155 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/autocomplet
...
+      widget.init({ helper, instantSearchInstance: {} });
+
+      const nextState = widget.dispose({ helper, state: helper.state });
+
+      expect(nextState.highlightPreTag).toBeUndefined();
+      expect(nextState.highlightPostTag).toBeUndefined();
+    });
+
+    it('removes the TAG_PLACEHOLDER from the `SearchParameters` only with `
escapeHTML`', () => {
```
francoischalifour:
```suggestion
    it('does not remove the TAG_PLACEHOLDER from the `SearchParameters` with `escapeHTML`', () => {
```

PR #3816, opened 2019-05-27
feat(connectHits): clear the state on dispose

samouss:
This PR fix the implementation of the `dispose` method for the connector `connectHits`. It omits `highlightTags` once the widget is removed. I also cover the rest of the `dispose` function with tests.

### DIFF
--- a/src/connectors/hits/__tests__/connectHits-test.js
+++ b/src/connectors/hits/__tests__/connectHits-test.js
@@ -348,4 +348,19 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
       expect.anything()
     );
   });
+
+  describe('dispose', () => {
+    it('calls the unmount function', () => {
+      const renderFn = () => {};
+      const unmountFn = jest.fn();
+      const makeWidget = connectHits(renderFn, unmountFn);
+      const widget = makeWidget();
+
+      expect(unmountFn).toHaveBeenCalledTimes(0);
+
+      widget.dispose();
+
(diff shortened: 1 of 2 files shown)


src/connectors/hits/__tests__/connectHits-test.js
```diff
@@ -348,4 +344,77 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
...
+      );
+
+      const nextState = widget.dispose({ helper, state: helper.state });
+
+      expect(nextState.highlightPreTag).toBeUndefined();
+      expect(nextState.highlightPostTag).toBeUndefined();
+    });
+
+    it('removes the TAG_PLACEHOLDER from the `SearchParameters` only with `
escapeHTML`', () => {
```
francoischalifour:
```suggestion
    it('does not remove the TAG_PLACEHOLDER from the `SearchParameters` with `escapeHTML`', () => {
```

src/connectors/hits/connectHits.js
```diff
@@ -99,8 +100,22 @@ export default function connectHits(renderFn, unmountFn) {
...
-      dispose() {
+      dispose({ state }) {
         unmountFn();
+
+        if (!escapeHTML) {
+          return state;
+        }
+
+        return state.setQueryParameters(
+          Object.keys(TAG_PLACEHOLDER).reduce(
```
Haroenv: Object.fromEntries would make this cleaner, but needs to be polyfilled, so this lgtm

PR #3819, opened 2019-05-27
feat(connectInfiniteHits): clear the state on dispose

samouss:
This PR fix the implementation of the `dispose` method for the connector `connectInfiniteHits`. It omits `page` once the widget is removed. I also cover the rest of the `dispose` function with tests.

### DIFF
--- a/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
+++ b/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
@@ -44,10 +44,8 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
   });
 
   it('Renders during init and render', () => {
-    // test that the dummyRendering is called with the isFirstRendering
-    // flag set accordingly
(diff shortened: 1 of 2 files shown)


src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -596,23 +591,121 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
+        helper,
+        state: helper.state,
+      }) as SearchParameters;
+
+      expect(nextState.highlightPreTag).toBeUndefined();
+      expect(nextState.highlightPostTag).toBeUndefined();
+    });
+
+    it('removes the TAG_PLACEHOLDER from the `SearchParameters` only with `
escapeHTML`', () => {
```
francoischalifour:
```suggestion
    it('does not remove the TAG_PLACEHOLDER from the `SearchParameters` with `escapeHTML`', () => {
```

PR #3820, opened 2019-05-27
fix(getTag): use object version of toString

### LINKED ISSUE #3814: calling toString on random objects can cause errors in IE
**Describe the bug 🐛**

**To Reproduce 🔍**
1. Go to IE11
2. see an error

A live example helps a lot!   We have a simple online template for you to use for your explanations:

for now: https://nopproj.computeruniverse.net/de/search?cid=0&query=

**Expected behavior 💭**

No errors

**Screenshots 🖥**

Screenshot 2019-05-27 at 15 49 22

**Environment:**
 - OS: Windows
 - Browser: IE
 - Version: 11

**Additional context**

toString.call can not be called on a string in IE, we should check for that first https://github.com/algolia/instantsearch.js/blob/9dca5480d27265a92f9e0f0cc7405b1c33badd0e/src/lib/utils/isPlainObject.ts#L11

Haroenv:
IE11: global `toString` can not be called as `Object.prototype.toString`, so we need to use the Object version, like done in lodash (not doing this initially was likely a simple omission)

fixes #3814 

references: 

- https://github.com/lodash/lodash/pull/4115 (remove core-js to test more regular code)
- https://github.com/lodash/lodash/blob/master/.internal/getTag.js#L1

Screenshot 2019-05-27 at 17 23 43
Screenshot 2019-05-27 at 15 49 22

### DIFF
--- a/src/lib/utils/isPlainObject.ts
+++ b/src/lib/utils/isPlainObject.ts
@@ -8,7 +8,7 @@ function getTag(value: any): string {
     return value === undefined ? '[object Undefined]' : '[object Null]';
   }
 
-  return toString.call(value);
+  return Object.prototype.toString.call(value);
 }
 
 function isObjectLike(value: any): boolean {


Haroenv:
other versions of `toString` were not the global call, but directly called on an object / string @samouss

PR #3822, opened 2019-05-27
feat(connectSearchBox): clear the state on dispose

samouss:
This PR fix the implementation of the `dispose` method for the connector `connectSearchBox`. It omits `query` once the widget is removed. I also cover the rest of the `dispose` function with tests.

### DIFF
--- a/src/connectors/search-box/connectSearchBox.js
+++ b/src/connectors/search-box/connectSearchBox.js
@@ -133,7 +133,7 @@ export default function connectSearchBox(renderFn, unmountFn = noop) {
       dispose({ state }) {
         unmountFn();
 
-        return state.setQuery('');
+        return state.setQueryParameter('query', undefined);
       },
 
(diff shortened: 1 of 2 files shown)


src/connectors/search-box/connectSearchBox.js
```diff
@@ -63,7 +64,7 @@ const withUsage = createDocumentationMessageGenerator({
  *   })
  * );
  */
-export default function connectSearchBox(renderFn, unmountFn) {
+export default function connectSearchBox(renderFn, unmountFn = noop) {
```
Haroenv: Is this consistent with the other connectors?

samouss: For some of them - Yes. It should be the norm though. Otherwise, the connector throws an error on `dispose`. Hence the test on each PR I've opened about the `dispose` function.

PR #3824, opened 2019-05-28
fix(connectSortBy): do not throw with wrong indexes

### LINKED ISSUE #3397: [connectSortBy] avoid to throw an error when the index does not exist
**Describe the bug 🐛**

The current implementation does throw an error when the current index is not in the list of provided indices. It can be an issue with the URL sync. It's easy for a users to mistyped the name of the index. The errors throw and block the render of the InstantSearch App.

> https://rrz1pyz83n.codesandbox.io/?sortBy=instantt_search_price_asc

**Expected behavior 💭**

Fallback to the default index when the current one does not exist?

**Additional context**

> https://discourse.algolia.com/t/how-to-catch-errors/6938

francoischalifour:
> This PR closes #3397.

The `sortBy` connector used to throw an error when the current index wasn't listed in the `items` of the widget.

This broke some users' apps when using `routing`. If users type a URL with a wrong index name for the `sortBy` widget, it will throw straight away and break the InstantSearch lifecycle.

This PR removes the error thrown and rathers warns only in development. The initial index (from the helper) is used if that happens.

I slightly changed the implementation because we were relying on the InstantSearch instance index (`instantSearch.indexName`) and now use the helper index (`helper.getIndex()`). This shouldn't break anything and is more consistent within the connector.

The deleted tests in the widget are now covered in the connector.

### DIFF
--- a/src/connectors/sort-by/connectSortBy.js
+++ b/src/connectors/sort-by/connectSortBy.js
@@ -2,6 +2,7 @@ import {
   checkRendering,
   createDocumentationMessageGenerator,
   find,
+  warning,
 } from '../../lib/utils';
 
 const withUsage = createDocumentationMessageGenerator({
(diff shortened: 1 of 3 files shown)


src/connectors/sort-by/connectSortBy.js
```diff
@@ -100,21 +101,25 @@ export default function connectSortBy(renderFn, unmountFn) {
 
     return {
       init({ helper, instantSearchInstance }) {
-        const currentIndex = helper.getIndex();
-        const isIndexInList = find(items, item => item.value === currentIndex);
+        const initialIndex = helper.getIndex();
```
Haroenv: not sure if this is needed, but we might remove these kind of functions from the helper, instead it can be read like `helper.index` I think. cc @samouss

samouss: Yes I agree with @Haroenv. We can use `helper.state.index`.

PR #3826, opened 2019-05-28
 refactor(analytics): migrate to TypeScript

francoischalifour:
This PR migrates the `analytics` widget to TypeScript.

I added a `dispose` method to clear the side effects created during `init`.

To ease the review, you can check https://github.com/algolia/instantsearch.js/commit/7cf252631bfcd84e8fce8df3f748f7c57e9fca82 which is the diff after renaming the files.


src/widgets/analytics/analytics.ts
```diff
@@ -0,0 +1,202 @@
...
+          )}_${encodeURIComponent(values)}`
+        );
+      }
+    }
+
+    return refinements.join('&');
+  };
+
+  const serializeNumericRefinements = function(
+    numericRefinements: any
```
yannickcr: To be more specific, it could be typed as `QueryParameters['numericRefinements']` (from `algoliasearch-helper`)

src/widgets/analytics/analytics.ts
```diff
@@ -0,0 +1,202 @@
...
+        stringifiedParams,
+        analyticsState.state,
+        analyticsState.results
+      );
+
+      lastSentData = dataToSend;
+    }
+  };
+
+  let pushTimeout: any;
```
yannickcr: To be more specific, it could be typed as `number` and use `window.setTimeout` (to be coherent with `history.ts`)

src/widgets/analytics/analytics.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  type AnalyticsState = {
+    results: SearchResults;
+    state: SearchParameters;
+  } | null;
+
+  let cachedState: AnalyticsState = null;
+
+  const serializeRefinements = function(parameters: object): string {
```
yannickcr: To be more specific, instead of `object` it could be typed as

```js
{
  [key: string]: string[];
}
```

PR #3830, opened 2019-05-31
fix(escape): make sure that __escaped does not get removed

Haroenv:
When mapping over an array, extra non-enumerable properties don't get copied into the new array. This means that the `__escaped` gets removed due to transformItems, or addAbsolutePosition.

We fix this by setting that mark back just before calling the render function. We want transformItems to happen after escaping, so it's no option to put the mark before.

Note also that if we would put the escaping before, we would still need to capture the marker, since it will be removed by the earlier transformations.

The new tests I added fail without my modification.

See also https://github.com/algolia/vue-instantsearch/issues/639 which highlighted 😅 this issue.

### DIFF
--- a/src/connectors/hits/__tests__/connectHits-test.js
+++ b/src/connectors/hits/__tests__/connectHits-test.js
@@ -219,7 +219,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
 
     const expectedHits = [{ name: 'transformed' }, { name: 'transformed' }];
     expectedHits.__escaped = true;
-    results.hits.__escaped = true;
 
     expect(rendering).toHaveBeenNthCalledWith(
       2,
@@ -348,7 +347,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
       },
     ];
     expectedHits.__escaped = true;
-    results.hits.__escaped = true;
 
     expect(rendering).toHaveBeenNthCalledWith(
       2,
--- a/src/connectors/hits/__tests__/connectHits-test.js
+++ b/src/connectors/hits/__tests__/connectHits-test.js
@@ -3,7 +3,7 @@ import { TAG_PLACEHOLDER } from '../../../lib/escape-highlight';
 import connectHits from '../connectHits';
 
 jest.mock('../../../lib/utils/hits-absolute-position', () => ({
-  addAbsolutePosition: hits => hits,
+  addAbsolutePosition: hits => hits.map(x => x),
 }));
 
 describe('connectHits', () => {
@@ -50,7 +50,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
     );
 
     widget.render({
-      results: new SearchResults(helper.state, [{}]),
+      results: new SearchResults(helper.state, [{ hits: [] }]),
       state: helper.state,
       helper,
       createURL: () => '#',
@@ -217,10 +217,14 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
       createURL: () => '#',
     });
 
+    const expectedHits = [{ name: 'transformed' }, { name: 'transformed' }];
+    expectedHits.__escaped = true;
+    results.hits.__escaped = true;
+
     expect(rendering).toHaveBeenNthCalledWith(
       2,
       expect.objectContaining({
-        hits: [{ name: 'transformed' }, { name: 'transformed' }],
+        hits: expectedHits,
         results,
       }),
       expect.anything()
@@ -254,13 +258,16 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
       createURL: () => '#',
     });
 
+    const expectedHits = [
+      { name: 'name 1', __queryID: 'theQueryID' },
+      { name: 'name 2', __queryID: 'theQueryID' },
+    ];
+    expectedHits.__escaped = true;
+
     expect(rendering).toHaveBeenNthCalledWith(
       2,
       expect.objectContaining({
-        hits: [
-          { name: 'name 1', __queryID: 'theQueryID' },
-          { name: 'name 2', __queryID: 'theQueryID' },
-        ],
+        hits: expectedHits,
       }),
       expect.anything()
     );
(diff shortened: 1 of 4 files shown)


src/connectors/hits/__tests__/connectHits-test.js
```diff
@@ -3,7 +3,7 @@ import { TAG_PLACEHOLDER } from '../../../lib/escape-highlight';
 import connectHits from '../connectHits';
 
 jest.mock('../../../lib/utils/hits-absolute-position', () => ({
-  addAbsolutePosition: hits => hits,
+  addAbsolutePosition: hits => hits.map(x => x),
```
Haroenv: I changed this so that my test would fail before the modification, to keep it closer to the real implementation

src/connectors/hits/connectHits.js
```diff
@@ -74,10 +74,12 @@ export default function connectHits(renderFn, unmountFn) {
       },
 
       render({ results, instantSearchInstance }) {
-        if (escapeHTML && results.hits && results.hits.length > 0) {
```
Haroenv: results.hits can never be something else than an array

src/connectors/hits/connectHits.js
```diff
@@ -74,10 +74,12 @@ export default function connectHits(renderFn, unmountFn) {
...
+        if (escapeHTML && results.hits.length > 0) {
           results.hits = escapeHits(results.hits);
         }
 
+        const oldEscaped = results.hits.__escaped;
```
francoischalifour: Suggestion name: `initialEscaped`.

src/connectors/hits/connectHits.js
```diff
@@ -88,6 +90,11 @@ export default function connectHits(renderFn, unmountFn) {
 
         results.hits = transformItems(results.hits);
 
+        // make sure the escaped tag stays, even after mapping over the hits
```
francoischalifour:
```suggestion
        // Make sure the escaped tag stays, even after mapping over the hits.
```

src/connectors/hits/connectHits.js
```diff
@@ -88,6 +90,11 @@ export default function connectHits(renderFn, unmountFn) {
...
+        // this prevents the hits from being double-escaped if there are multiple
```
francoischalifour:
```suggestion
        // This prevents the hits from being double-escaped if there are multiple
```

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -9,7 +9,7 @@ import connectInfiniteHits from '../connectInfiniteHits';
 import { Client } from '../../../types';
 
 jest.mock('../../../lib/utils/hits-absolute-position', () => ({
-  addAbsolutePosition: hits => hits,
+  addAbsolutePosition: hits => hits.map(x => x),
```
francoischalifour: We should add a comment explaining why we map because out of context, it doesn't make much sense (→ to create a new array instance).

src/connectors/hits/__tests__/connectHits-test.js
francoischalifour: We should add a comment explaining why we map because out of context, it doesn't make much sense (→ to create a new array instance). It will also make it explicit that this part can be removed once we change the implementation.

src/connectors/hits/__tests__/connectHits-test.js
```diff
@@ -217,10 +217,14 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
       createURL: () => '#',
     });
 
+    const expectedHits = [{ name: 'transformed' }, { name: 'transformed' }];
+    expectedHits.__escaped = true;
+    results.hits.__escaped = true;
```
francoischalifour: What's this for? I believe this is unused.

src/connectors/hits/__tests__/connectHits-test.js
```diff
@@ -322,30 +329,62 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
...
+        name: 'halloween',
+        _highlightResult: {
+          name: {
+            value: 'HA<MARK>LLO</MARK>WEEN',
+          },
+        },
+      },
+    ];
+    expectedHits.__escaped = true;
+    results.hits.__escaped = true;
```
francoischalifour: What's this for? I believe this is unused.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -179,9 +179,10 @@ const connectInfiniteHits: InfiniteHitsConnector = (
           prevState = currentState;
         }
 
-        if (escapeHTML && results.hits && results.hits.length > 0) {
+        if (escapeHTML && results.hits.length > 0) {
           results.hits = escapeHits(results.hits);
         }
+        const oldEscaped = (results.hits as any).__escaped;
```
francoischalifour: Suggestion name: `initialEscaped`.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -193,6 +194,11 @@ const connectInfiniteHits: InfiniteHitsConnector = (
 
         results.hits = transformItems(results.hits);
 
+        // make sure the escaped tag stays, even after mapping over the hits
```
francoischalifour:
```suggestion
        // Make sure the escaped tag stays after mapping over the hits.
```

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -193,6 +194,11 @@ const connectInfiniteHits: InfiniteHitsConnector = (
...
+        // this prevents the hits from being double-escaped if there are multiple
```
francoischalifour:
```suggestion
        // This prevents the hits from being double-escaped if there are multiple
```

src/connectors/hits/__tests__/connectHits-test.js
Haroenv: Best expects arrays to have extra properties as well, do fails without this

src/connectors/hits/connectHits.js
Haroenv: Can we add a lint rule instead

francoischalifour: It's more that the sentence was not correct at first because it needed a break in between (a fullstop here), so then we should capitalize the next sentence.

src/connectors/hits/__tests__/connectHits-test.js
francoischalifour: I tried and it doesn't fail without this.

Haroenv: ah yes, it's not needed because the reference is the same

PR #3842, opened 2019-06-04
fix(getRefinements): check for facet before accessing its data

francoischalifour:
This regression was introduced while removing Lodash from the codebase. We need to check that `facet` is defined before accessing `facet.data`, otherwise some query might fail and break InstantSearch lifecycle.

Steps to reproduce the issue:

- Select "Cameras & Camcorders" from the HierarchicalMenu
- Select "Cameras & Camcorders > Binoculars" from the HierarchicalMenu
- Type the query: "pokem"

Example on [this deployment](https://deploy-preview-3802--instantsearchjs.netlify.com/examples/e-commerce).

### DIFF
--- a/src/lib/utils/__tests__/index-test.js
+++ b/src/lib/utils/__tests__/index-test.js
@@ -699,6 +699,37 @@ describe('utils.getRefinements', () => {
     );
   });
 
+  it('should retrieve hierarchicalFacetsRefinements on multiple levels without any data', () => {
+    helper.toggleRefinement(
+      'hierarchicalFacet2',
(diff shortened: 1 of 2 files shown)


francoischalifour:
@samouss @eunjae-lee I wasn't able to replicate in our test environment. If you can give it a go, I'd appreciate!

eunjae-lee:
I had a URL with query params which replicated the issue, but now I've lost it... :(

samouss:
@francoischalifour I can't add you to the list of reviewers since you're the owner.

PR #3843, opened 2019-06-04
chore(examples): update e-commerce design [part 5]

francoischalifour:
This updates the e-commerce demo design after reviewing it with Nico M.

[Preview →](https://deploy-preview-3843--instantsearchjs.netlify.com/examples/e-commerce/)


examples/e-commerce/src/app.css
```diff
@@ -281,3 +267,9 @@ mark {
 .toggle-refinement-checkbox:checked::before {
   content: 'Yes';
 }
+
+/* RatingMenu */
+
+.rating-menu-item:not(.rating-menu-item--selected) {
```
francoischalifour: It would be best to apply this style only when a RatingMenu item is refined but I didn't get to do it with only CSS.

examples/e-commerce/src/widgets/Configuration.js
```diff
@@ -1,7 +1,7 @@
 import { configure } from 'instantsearch.js/es/widgets';
 
 const configuration = configure({
-  attributesToSnippet: ['description'],
+  attributesToSnippet: ['name:10', 'description:10'],
```
francoischalifour: We want smaller titles and descriptions that take about two lines.

examples/e-commerce/src/widgets/SearchBox.js
```diff
@@ -3,7 +3,6 @@ import { searchBox as searchBoxWidget } from 'instantsearch.js/es/widgets';
 const searchBox = searchBoxWidget({
   container: '#searchbox',
   placeholder: 'Search for a product, brand, color, …',
-  showReset: false,
```
francoischalifour: Since we don't showcase voice search yet, we can show the reset button.

examples/e-commerce/src/widgets/Pagination.js
```diff
@@ -3,7 +3,7 @@ import { pagination as paginationWidget } from 'instantsearch.js/es/widgets';
 const pagination = paginationWidget({
   container: '#pagination',
   scrollTo: '.container',
-  totalPages: 4,
+  totalPages: 5,
```
samouss: This option limit the number of pages to 5, this is really what we want?

Screenshot 2019-06-05 at 14 44 51

examples/e-commerce/src/widgets/Configuration.js
samouss: With the snippet on the `name` attribute we might have the `...` at the beginning. It feels a bit weird for a product name.

Screenshot 2019-06-05 at 14 47 49

francoischalifour: That's true, we might want to only use a snippet for the description.

examples/e-commerce/src/widgets/Pagination.js
francoischalifour: From our discussion with Nico, there shouldn't be a reason to go beyond because the first pages should already be relevant to the query. Do you think using `padding` would be a better alternative?

samouss: Ok, it doesn't matter much at the end it's an example. On real website though (it depends on your strategy) I would have used padding. Users might want to be able to browse the full content of the catalog e.g. Amazon, Google, leboncoin.

examples/e-commerce/src/widgets/Pagination.js
```diff
@@ -3,7 +3,7 @@ import { pagination as paginationWidget } from 'instantsearch.js/es/widgets';
 const pagination = paginationWidget({
   container: '#pagination',
   scrollTo: '.container',
-  totalPages: 4,
+  padding: 2,
```
samouss: With a `padding` we probably want to display first & last page. Otherwise it's cumbersome to go back on the first page with only previous arrow.

francoischalifour: That's not part of the design and this will break it.

samouss: Yes I know that it's not, worth asking Nico though.

PR #3845, opened 2019-06-05
fix(voiceSearch): remove event listeners on stop

eunjae-lee:
**Summary**

This PR contains began with copy&pasting the latest version of voiceSearchHelper from react-IS.

- Better typing: 046de085
- Removes event listeners on stop: 5cefe7c2
- Fixing the side effect from the above: 84cc9f9c

**Result**

While waiting/recognizing, if user triggers 'stop', the status will be

... -> `waiting(or recognizing)` -> (user clicks button) -> `finished`


src/lib/voiceSearchHelper/index.ts
```diff
@@ -131,6 +134,14 @@ export default function createVoiceSearchHelper({
     recognition = undefined;
   };
 
+  const stop = (): void => {
+    dispose();
+    // Because `dispose` removes event listeners, `end` listener is not called.
+    // So we're setting the `status` as `finished` here.
+    // If we don't do it, it will be still `waiting` or `recognizing`.
+    resetState('finished');
```
francoischalifour: Can we add a test to check the status at this point?

eunjae-lee: Good idea. Thanks!
851de97

src/lib/voiceSearchHelper/__tests__/index-test.ts
```diff
@@ -191,4 +191,19 @@ describe('VoiceSearchHelper', () => {
...
+    const onQueryChange = (): void => {};
+    const onStateChange = (): void => {};
+    const voiceSearchHelper = createVoiceSearchHelper({
+      searchAsYouSpeak: true,
+      onQueryChange,
+      onStateChange,
+    });
+
+    voiceSearchHelper.toggleListening();
+    voiceSearchHelper.toggleListening();
```
samouss: IIIRC the main problem was that `setState` was called multiple times with different values. From the `onEnd` listener and inside `stop` (via `resetState`). From what I've understood, now they are called with the same values. But the function is still called two times no?

eunjae-lee: Yes, the `status` changed like this:
`initial` -> `waiting` -> (button click) -> `initial` (which shouldn't be) -> `finished`.

It's supposed to be like this:
`initial` -> `waiting` -> (button click) -> `finished`.

Now with these commits, after recognition started, if user clicks the button (which triggers `stop`),
the `stop()` will call `dispose()` and `resetState('finished');`.

`dispose()` will simply stop the recognition and remove all the event listeners. Now that the listeners are gone, `onEnd` will not be called.
and `resetState('finished')` will set the `status` to `'finished'`.

To sum up, in this version, after `(button click)`, `setState` is called only once.

src/lib/voiceSearchHelper/index.ts
```diff
@@ -1,21 +1,32 @@
...
   onStateChange: () => void;
 };
 
+export type Status =
+  | 'initial'
+  | 'askingPermission'
+  | 'waiting'
+  | 'recognizing'
+  | 'finished'
+  | 'error';
```
yannickcr: Could be replaced by an enum, so you could continue to use constants instead of duplicating the different values:

```js
enum Status {
  INITIAL = 'initial',
  ASKING_PERMISSION = 'askingPermission',
  WAITING = 'waiting',
  RECOGNIZING = 'recognizing',
  FINISHED = 'finished',
  ERROR = 'error',
}

// ...

setState({
   status: Status.WAITING,
});
```

src/lib/voiceSearchHelper/index.ts
```diff
@@ -1,21 +1,32 @@
...
+  | 'bad-grammar'
+  | 'language-not-supported';
+
 export type VoiceListeningState = {
-  status: string;
+  status: Status;
   transcript: string;
   isSpeechFinal: boolean;
-  errorCode?: string;
+  errorCode?: ErrorCode;
```
yannickcr: Could be typed as a `SpeechRecognitionErrorCode`

src/lib/voiceSearchHelper/index.ts
samouss: The advantage of the union type is that it completely disappears at runtime, [it's not the case with an enum](https://www.typescriptlang.org/play/index.html#src=enum%20EStatus%20%7B%0D%0A%20%20INITIAL%20%3D%20'initial'%2C%0D%0A%20%20ASKING_PERMISSION%20%3D%20'askingPermission'%2C%0D%0A%20%20WAITING%20%3D%20'waiting'%2C%0D%0A%20%20RECOGNIZING%20%3D%20'recognizing'%2C%0D%0A%20%20FINISHED%20%3D%20'finished'%2C%0D%0A%20%20ERROR%20%3D%20'error'%2C%0D%0A%7D%0D%0A%0D%0Aconst%20eValue%3A%20EStatus%20%3D%20EStatus.INITIAL%3B%0D%0A%0D%0A%2F**%0D%0A%20*%20---%0D%0A%20*%2F%0D%0A%0D%0Atype%20TStatus%20%3D%0D%0A%20%20%7C%20'initial'%0D%0A%20%20%7C%20'askingPermission'%0D%0A%20%20%7C%20'waiting'%0D%0A%20%20%7C%20'recognizing'%0D%0A%20%20%7C%20'finished'%0D%0A%20%20%7C%20'error'%3B%0D%0A%0D%0Aconst%20tValue%3A%20TStatus%20%3D%20'initial'%3B%20). We do have a kind of repetition with the enum too. We duplicate the different values but the compiler helps us with this if the type changes or we do a typo an error is thrown. One advantage of the enum is to be able to enumerate all the possibles values at runtime. But we don't use the feature.

src/lib/voiceSearchHelper/__tests__/index-test.ts
samouss: Yes, thanks!

src/lib/voiceSearchHelper/index.ts
eunjae-lee: Thanks for the thorough explanation.

src/lib/voiceSearchHelper/index.ts
eunjae-lee: Right, it works fine :)

PR #3866, opened 2019-06-19
chore(examples): implement e-commerce demo mobile design

francoischalifour:
## Description

This implements the mobile design of the e-commerce demo. Since the theme we provide is desktop-first, I wrote the demo style desktop-first.

## Preview

[See preview →](https://deploy-preview-3866--instantsearchjs.netlify.com/examples/e-commerce)

## Implementation

I took some decisions on how to develop the mobile adaptation to the e-commerce demo. Let me know if you think that some are not appropriate.

- I used `data-` attributes for multiple purposes.
    - `data-widget="widget-name"` to be explicit on which element is related to an InstantSearch widget (I didn't specified an `id`)
    - `data-layout="mobile|desktop"` to filter out the elements that shouldn't be displayed on mobile or desktop
- SVGs have a `viewbox` attribute so that they can be scaled via CSS
- Once the user refines the filters, it scrolls to the results container
- `position: sticky;` is used for the button at the bottom of the filters panel but is not supported on iOS

## Notes

Here are some notes on things that we can improve.

- The correct breakpoint to trigger to the mobile design is to be defined by Nico
- The animation is not great and I plan to revisit it. I'd appreciate some suggestions on how to make it look better.
- The hit layout overflows on small screens (see page 3)
- I'm not convinced with the chosen CSS naming so I'm in for suggestions

## Related

- [Design](https://app.zeplin.io/project/5acc8becfc0c980068fc12cc/screen/5d03a7333916305e0ad8e492)


examples/e-commerce/index.html
```diff
@@ -29,12 +30,7 @@
     <header class="header">
       <p class="header-logo">
         <a href="https://algolia.com" aria-label="Go to the Algolia website">
-          <svg
-            xmlns="http://www.w3.org/2000/svg"
-            width="93"
-            height="24"
-            viewBox="0 0 93 24"
-          >
+          <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 93 24">
```
francoischalifour: The SVG sizes are now specified in CSS.

examples/e-commerce/src/theme.css
```diff
@@ -10,6 +10,7 @@ a[class^='ais-'] {
   text-decoration: none;
 }
 
+[class^='ais-'][class$='--disabled'],
```
francoischalifour: This adds compatibility to Angular components.

eunjae-lee:
I'm not sure about the two buttons at the bottom. Whatever filters I modify, it's taken into account instantly. It means there's nothing to "save". It's kind of already saved. Intead of "Save and close", it should be just "Close" IMO.
If there should be a "save" button, I think it should work like this:
- Filter A is applied.
- User opens up the filter popup.
- User applies Filter B.
- User doesn't clicks "save" button but "close" button.
- The only filter applied now is still A.

According to current behaviors of our widgets, "reset" and "close" buttons seem to be enough. What do you think?

francoischalifour:
@eunjae-lee I agree with you. Here is my proposal to Nico, waiting for his approval.

Filters overlay

What do you think?

eunjae-lee:
I think it's better. I also thought about not having the bottom buttons view at all and having a reset button somewhere around "Filters" title and having a simple "X" close button somewhere top-right corner, but haven't got any proper proposal. I'm fine with yours.

eunjae-lee:
Could you explain to me more about why we're using `data-widget="widget-name"` instead of `id`? What do we get from this?

francoischalifour:
It's mainly to make it clear that a container is used for an InstantSearch widget. With `id` we don't know if it's used for styling, JavaScript interactions, etc.

It's not super important but I'm trying different patterns while building this demo.

Besides, we're becoming a `data-driven` company... haha

eunjae-lee:
haha okay. let's give it a go!

eunjae-lee:
FYI, [here](https://github.com/algolia/instantsearch.js/pull/3866/files#diff-e8c7e43cd3cdd0ff527be8f88d1d02acR80), `.container-body > div` is changed to `.container-body .ais-Panel` because the root element of `ais-panel` widget is not a `div` in Angular InstantSearch.

francoischalifour:
The dynamic save button with the number of hits was approved by Nico so I implemented it.

examples/e-commerce/src/widgets/Products.js
```diff
@@ -1,31 +1,33 @@
...
-      </svg>
-        {{rating}}
-      </span>
-    </p>
-  </footer>
+    <footer>
+      <p>
+        <span class="hit-em">$</span> <strong>{{price}}</strong>
+        <span class="hit-em hit-rating">
+        <svg xmlns="http://www.w3.org/2000/svg" width="8" height="8" viewBox="0 0 16 16">
```
eunjae-lee: You may want to get rid of `width` and `height` attributes here.

examples/e-commerce/index.html
```diff
@@ -48,45 +44,72 @@
         Stop looking for an item — find it.
       </p>
 
-      <div id="searchbox"></div>
+      <div data-widget="searchbox"></div>
```
tkrugg: is there a reason for not using ids?
ids should probably be the best practice we push for, as they are often faster by x100
https://jsperf.com/id-vs-class-vs-data-attribute/6

Not a blocker at all.

eunjae-lee: fyi https://github.com/algolia/instantsearch.js/pull/3866#issuecomment-504025276

examples/e-commerce/src/theme.css
tkrugg: Since the goal is to have the same DOM output everywhere (eventually), shouldn't we add here a comment reminding to remove this when no longer necessary?

examples/e-commerce/src/theme.css
```diff
@@ -10,6 +10,7 @@ a[class^='ais-'] {
...
 [class^='ais-'][class$='--disabled'] * {
```
eunjae-lee: Do we need this?
I've tested the vanillia with only `[class^='ais-'][class$='--disabled'] { ... }` and it works fine.
And I think it's supposed to work without `*`.
What do you think?

examples/e-commerce/src/widgets/Products.js
```diff
@@ -1,31 +1,33 @@
...
-      <svg xmlns="http://www.w3.org/2000/svg" width="8" height="8" viewBox="0 0 16 16">
-        <path fill="#e2a400" fill-rule="evenodd" d="M10.472 5.008L16 5.816l-4 3.896.944 5.504L8 12.616l-4.944 2.6L4 9.712 0 5.816l5.528-.808L8 0z"/>
-      </svg>
-        {{rating}}
-      </span>
-    </p>
-  </footer>
+    <footer>
+      <p>
+        <span class="hit-em">$</span> <strong>{{price}}</strong>
```
tkrugg: This should use `{{#helpers.formatNumber}}` on the price, helps the readability.

examples/e-commerce/src/ui.js
```diff
@@ -0,0 +1,37 @@
+const filtersButtons = [
+  ...document.querySelectorAll('[data-action="open-overlay"]'),
+];
+const closeOverlayButtons = [
+  ...document.querySelectorAll('[data-action="close-overlay"]'),
+];
+const resultsContainer = document.querySelector('.container-results');
```
eunjae-lee: FYI, I've move these queries into functions in Angular and Vue to avoid querying before rendering.

eunjae-lee:
image

unnecessary gap there.

examples/e-commerce/src/app.mobile.css
```diff
@@ -0,0 +1,261 @@
...
+  /* RefinementList */
+
+  .ais-RefinementList-list {
+    display: grid;
+    grid-auto-flow: column;
+    grid-template-rows: repeat(5, 1fr);
+  }
+
+  .ais-RefinementList-item {
+    flex: 50%;
```
tkrugg: image

examples/e-commerce/src/app.mobile.css
```diff
@@ -0,0 +1,261 @@
...
+  }
+
+  .container-filters-footer {
+    background-color: #fff;
+    border-top: 1px solid #ebecf3;
+    bottom: 4rem;
+    display: flex;
+    justify-content: space-between;
+    padding: 1rem;
+    position: sticky;
```
tkrugg: it looks like this didn't work on iphoneX (found on react-instantsearch example too)

eunjae-lee: I guess we need `position: -webkit-sticky` to make it work. But doing so introduces another problem on my iPhone XR. 

RPReplay_Final1561380846 2 2019-06-24 15_19_58

Look at the bottom. The bottom bars are not aligned to bottom.

examples/e-commerce/src/widgets/Products.js
francoischalifour: I left them here because they're not part of the theme, they don't change size based on the screen size.

examples/e-commerce/src/theme.css
francoischalifour: We need this line for widgets like RangeSlider which contain multiple components with several cursors. Ohterwise, the "not-allowed" cursor gets overridden.

I can add a comment to explain why in the theme.

examples/e-commerce/src/widgets/Products.js
francoischalifour: I didn't because we don't in other flavors. I can add it and add a custom formatter in other demos.

examples/e-commerce/index.html
francoischalifour: @tkrugg See the link above sent by @eunjae-lee.

Speed shouldn't be an issue here. I tried to make the code more descriptive with this approach. Not entirely satisfied but I found it easier to understand where the widgets are injected. I can change if you have concerns.

examples/e-commerce/src/theme.css
eunjae-lee: Ah I see. It'd be good to have that comment since this one is `theme.css`, but not `app.css`.

examples/e-commerce/src/widgets/Products.js
eunjae-lee: okay, got it.

examples/e-commerce/src/theme.css
francoischalifour: WDYM?

As I understand it, it's needed for Angular because it creates web components. Those are not targetable with the `*` selector.

examples/e-commerce/src/app.mobile.css
francoischalifour: Fixed in 1ddb6fe9343ae223bf4f4206fd42eb55eff6012d using `position: fixed;`.

examples/e-commerce/src/app.mobile.css
```diff
@@ -159,6 +159,15 @@
     max-width: 90%;
   }
 
+  .ais-SearchBox .ais-SearchBox-input,
+  .ais-RefinementList .ais-SearchBox-input {
+    font-size: 1rem;
```
francoischalifour: This makes the inputs font size at `16px`, which is the fix for mobile browsers on iOS to not zoom on the inputs.

examples/e-commerce/src/theme.css
```diff
@@ -196,52 +200,41 @@ a[class^='ais-'] {
...
   -webkit-appearance: none;
   -moz-appearance: none;
   appearance: none;
+  display: flex;
   height: 100%;
+  justify-content: center;
   position: absolute;
   top: 50%;
   transform: translateY(-50%);
+  width: 48px;
```
francoischalifour: This makes the buttons in the input more accessible on mobile: it increases the touch area.

examples/e-commerce/src/app.mobile.css
francoischalifour: Fixed in c7622d086cc465125468d9e94797c0dee69b739b.

PR #3882, opened 2019-06-24
feat(utils): implement defer

samouss:
**Summary**

This PR introduce the utility function `defer`. It allows a provided function to be delayed until the next micro task i.e. `Promise.resolve()`. It also ensure that only one call to this function can occur inside the same micro task. We'll use this function to "batch" the calls to `search` and `render` in the coming PRs.  Here is an example of its usage:

```js
const fn1 = defer(() => {
  console.log('Hello');
});

const fn2 = defer(() => {
  console.log('Bye');
});

fn1();
fn2();
fn1();
fn2();
fn1();
fn2();

console.log('Start');
Promise.resolve().then(() => {
  console.log('Done');
});

// Output:
// Start
// Hello
// Bye
// Done
```

### DIFF
--- a/src/lib/utils/defer.ts
+++ b/src/lib/utils/defer.ts
@@ -1,0 +1,19 @@
+const nextMicroTask = Promise.resolve();
+
+type Callback = (...args: any[]) => void;
+
+const defer = (callback: Callback): Callback => {
+  let progress: Promise<void> | null = null;
+  return (...args) => {
+    if (progress !== null) {
+      return;
+    }
+
+    progress = nextMicroTask.then(() => {
+      callback(...args);
+      progress = null;
+    });
+  };
+};
+
+export default defer;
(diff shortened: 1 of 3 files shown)


Haroenv:
What's this necessary for?

samouss:
> We'll use this function to "batch" the calls to `.search` in the upcoming PRs.

src/lib/utils/defer.ts
```diff
@@ -0,0 +1,19 @@
...
+type Callback = (...args: any[]) => void;
+
+const defer = (callback: Callback): Callback => {
+  let progress: Promise<void> | null = null;
+  return (...args) => {
+    if (progress !== null) {
+      return;
+    }
+
+    progress = nextMicroTask.then(() => {
```
Haroenv: is this not going to be shared between different calls of defer now?

samouss: `progress` is scoped inside each call to `defer` to track whether or not we have a "work" in progress. `nextMicroTask` is shared but it doesn't matter since we only use it to defer calls to the next microtask.

PR #3884, opened 2019-06-24
refactor(InstantSearch): avoid to store searchParameters

samouss:
This PR removes the `searchParameters` from the `InstantSearch` instance. We don't have to keep track of the them because we already have access to them through the `helper.state`. The PR only removes the reference on the `InstantSearch` not the usage of the `searchParameters` option.

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -91,13 +91,17 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     this.insightsClient = insightsClient;
     this.helper = null;
     this.indexName = indexName;
-    this.searchParameters = { ...searchParameters, index: indexName };
     this.widgets = [];
     this.templatesConfig = {
       helpers: createHelpers({ numberLocale }),
       compileOptions: {},
     };
+
     this._stalledSearchDelay = stalledSearchDelay;
+    this._searchParameters = {
+      ...searchParameters,
+      index: indexName,
+    };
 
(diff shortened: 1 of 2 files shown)


src/lib/InstantSearch.js
```diff
@@ -298,15 +302,15 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
...
-      this.searchParameters
+      this._searchParameters
     );
 
     const helper = algoliasearchHelper(
       this.client,
-      this.searchParameters.index || this.indexName,
-      this.searchParameters
+      initialSearchParameters.index || this.indexName,
+      initialSearchParameters
```
Haroenv: I'm not sure from this code why this is initialSearchParameters, instead of `this.helper.state`, can a comment be put for the reason?

samouss: We use `initialSearchParameters` to create the `helper`, we can't use `helper.state` to create it. Which part do you think is unclear? Note that almost nothing changed from the previous implementation.

Haroenv: I just mean that in the earlier parts of the code you managed to replace it with `this.helper`

samouss: Yes but inside the others parts of the code the `helper` is already created. Here it's not the case, we're inside the `start` method the `helper` does not exist yet. I could have inlined the returned value of the `reduce` to avoid the extra variable but we have to branch on the `indexName`.

PR #3885, opened 2019-06-24
feat(InstantSearch): switch to DerivedHelper only

samouss:
This PR switches the `.search` implementation to `. searchOnlyWithDerivedHelpers`. We override the `.search` method to keep the exact same API for the consumer of the libraries. Only the module `InstantSearch` is impacted by the change. 

This switch allow a full control on each request made by the Helper, which is not the case with the regular `.search` function. We're able to alter the request without any impact on the internal state of the Helper (via the derivation function). Now the "main" Helper is only in charge of managing the state, the derived one is in charge of the requests.

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -90,6 +90,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     this.client = searchClient;
     this.insightsClient = insightsClient;
     this.helper = null;
+    this.derivedHelper = null;
     this.indexName = indexName;
     this.widgets = [];
     this.templatesConfig = {
@@ -313,6 +314,16 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
       initialSearchParameters
     );
 
+    const derivedHelper = helper.derive(() => helper.state);
+
+    helper.search = () => {
+      // This solution allows us to keep the exact same API for the users but
+      // under the hood, we have a different implementation. It should be
+      // completely transparent for the rest of the codebase. Only this module
+      // is impacted.
+      helper.searchOnlyWithDerivedHelpers();
+    };
+
     if (this._searchFunction) {
       this._mainHelperSearch = helper.search.bind(helper);
       helper.search = () => {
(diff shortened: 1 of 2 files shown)


src/lib/InstantSearch.js
```diff
@@ -332,14 +343,17 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
     }
 
     this.helper = helper;
+    this.derivedHelper = derivedHelper;
 
-    this._init(helper.state, this.helper);
+    this._init(helper, helper.state);
```
samouss: I've switched the order of the parameter to match the one from `_render`.

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -281,7 +281,9 @@ describe('InstantSearch lifecycle', () => {
       });
 
       it('calls helper.search()', () => {
-        expect(search.helper.search).toHaveBeenCalledTimes(1);
+        expect(
+          search.helper.searchOnlyWithDerivedHelpers
```
samouss: The tests are brittle and rely on implementation details. I didn't change them to avoid too many changes but we'll have to reconsider this practice it makes them a bit useless for such changes. I don't have a good answer but it's something that we have to investigate (rely on something that is not as much tied to our core logic, client?).

Haroenv: testing the client indeed seems more our contract now, especially considering we might want to integrate the helper more closely

src/lib/InstantSearch.js
```diff
@@ -395,6 +413,21 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
...
+          templatesConfig: this.templatesConfig,
+          createURL: this._createAbsoluteURL,
+          onHistoryChange: this._onHistoryChange,
+          instantSearchInstance: this,
+        });
+      }
+    });
+  }
+
   _render(helper, results, state) {
```
Haroenv: unrelated to this pr, but why do we even bother splitting up the arguments here, isn't it always helper.results & helper.state?

samouss: It won't be there in the next PRs anyway.

samouss:
@Haroenv Most of the tests will be rewritten or removed with the next PRs.

PR #3892, opened 2019-06-26
feat(index): add widget [PART-1]

samouss:
The current implementation manages all the aspect of the search at the InstantSearch level (lifecycle, widgets, render, ...). This PR introduces an `index` widget that will take some of those responsibilities. 

The InstantSearch instance will be responsible to expose global state/methods for the `index` widgets. Each `index` widget own a local Helper to manage the state of the index with the widgets attach to it. By default, each InstantSearch instance will be created with an `index`. The calls to `addWidgets` & `removeWidgets` are forwarded to the `index`. One important thing to note is that the `index` does not make the queries. It delegates this operation with a DerivedHelper to the "main" Helper created at the InstantSearch level.

This split of responsibilities will allow multiple `index` inside the same instance. Currently, the feature is not "enabled" because the derivation function only looks up for [the local state](https://github.com/algolia/instantsearch.js/pull/3892/files#diff-f82164a0b1338f78fd22259b6552aed0R197). The idea is to resolve the parent until the root and create the `SearchParameters` for the current index. This will be possible once we're able to merge the `SearchParameters`.

At the moment the `index` widget is not connected to anything. The next PR will introduce its usage at the `InstantSearch` level. We'll take this opportunity to remove all the duplicated tests that we have between `InstantSearch` & `index`.


src/types/instantsearch.ts
```diff
@@ -119,10 +120,19 @@ export type DerivedHelper = EventEmitter & {
...
+  mainIndex: Index;
+  insightsClient: AlgoliaInsightsClient | null;
+  templatesConfig: object;
+  _isSearchStalled: boolean;
+  _searchParameters: Partial<SearchParameters>;
+  _createAbsoluteURL(state: Partial<SearchParameters>): string;
+  _onHistoryChange(callback: (state: Partial<SearchParameters>) => void): void;
+  scheduleSearch(): void;
+  scheduleRender(): void;
+  scheduleStalledRender(): void;
```
samouss: Those methods are implemented in a next PR.

src/lib/__tests__/index-test.ts
```diff
@@ -0,0 +1,855 @@
...
+          templatesConfig: instantSearchInstance.templatesConfig,
+          createURL: instantSearchInstance._createAbsoluteURL,
+          onHistoryChange: instantSearchInstance._onHistoryChange,
+        });
+      });
+
+      it('schedules a search to take the added widgets into account', () => {
+        const instance = index({ indexName: 'index_name' });
+        const instantSearchInstance = createInstantSearch({
+          scheduleSearch: jest.fn(),
```
samouss: Those tests are brittle, but the method doesn't exist yet on InstantSearch. We might want to actually use the instance when it's available. We won't "unit" test anymore but at least we'll test a real behaviour.

src/types/widget.ts
```diff
@@ -7,10 +8,12 @@ import {
 
 export interface InitOptions {
   instantSearchInstance: InstantSearch;
+  parent: Index | null;
```
samouss: This value is provided to the `index` in a next PR.

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
...
+
+      // The `mainHelper` is already defined at this point. The instance is created
+      // inside InstantSerach at the `start` method, which occurs before the `init`
+      // step.
+      const mainHelper = instantSearchInstance.mainHelper!;
+
+      const initialSearchParameters =
+        // Uses the `searchParameters` for the top level index only, it allow
+        // us to have the exact same behaviour than before for the mono-index.
+        parent === null ? instantSearchInstance._searchParameters : {};
```
samouss: The support of `searchParameters` is (still) required for default values. The `configure` widget does not behave exactly the same when a widget is added or removed. If the default value is removed by the user with a `configure`  widget (+ routing) the filters are re-applied on add/remove.

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
...
+      return this;
+    },
+
+    init({ instantSearchInstance, parent }) {
+      localInstantSearchInstance = instantSearchInstance;
+
+      // The `mainHelper` is already defined at this point. The instance is created
+      // inside InstantSerach at the `start` method, which occurs before the `init`
+      // step.
+      const mainHelper = instantSearchInstance.mainHelper!;
```
samouss: One solution to avoid the `!` assertion is to provide the "main" Helper through the lifecycle options. It's a solution that we might consider at some point. At the moment it's not the case because it's easier to add a parameter rather than removing one (since it's a public API).

src/lib/__tests__/index-test.ts
```diff
@@ -0,0 +1,855 @@
...
+
+        instance.addWidgets([]);
+
+        expect(instantSearchInstance.scheduleSearch).toHaveBeenCalledTimes(0);
+      });
+    });
+  });
+
+  describe('removeWidgets', () => {
+    it('removes given widgets from the instance', () => {
```
francoischalifour: It's worth also adding a test removing only _some_ widgets (not all).

src/lib/__tests__/index-test.ts
```diff
@@ -0,0 +1,855 @@
...
+      expect(searchBox.init).toHaveBeenCalledTimes(0);
+      expect(pagination.init).toHaveBeenCalledTimes(0);
+
+      instance.init!(
+        createInitOptions({
+          instantSearchInstance,
+        })
+      );
+
+      expect(searchBox.init).toHaveBeenCalledTimes(1);
```
francoischalifour: We don't test what `searchBox.init` was called with?

src/lib/__tests__/index-test.ts
```diff
@@ -0,0 +1,855 @@
...
+      expect(searchBox.render).toHaveBeenCalledTimes(0);
+      expect(pagination.render).toHaveBeenCalledTimes(0);
+
+      instance.render!(
+        createRenderOptions({
+          instantSearchInstance,
+        })
+      );
+
+      expect(searchBox.render).toHaveBeenCalledTimes(1);
```
francoischalifour: We don't test what `searchBox.render` was called with?

src/lib/__tests__/index-test.ts
```diff
@@ -0,0 +1,855 @@
...
+      );
+
+      expect(instance.getHelper()).toBeDefined();
+
+      instance.dispose!(createDisposeOptions());
+
+      expect(instance.getHelper()).toBe(null);
+    });
+
+    it('removes the DerivedHelper', async () => {
```
francoischalifour: Is this test doing what it's supposed to do?

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
+import algoliasearchHelper from 'algoliasearch-helper';
```
francoischalifour: Why isn't `src/lib/index.ts` in `src/widgets/index/`?

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
...
+        widgets.forEach(widget => {
+          if (localInstantSearchInstance && widget.init) {
+            widget.init({
+              helper: helper!,
+              parent: this,
+              instantSearchInstance: localInstantSearchInstance,
+              state: helper!.state,
+              templatesConfig: localInstantSearchInstance.templatesConfig,
+              createURL: localInstantSearchInstance._createAbsoluteURL,
+              onHistoryChange: localInstantSearchInstance._onHistoryChange,
```
francoischalifour: Is this still used?

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
...
+      }
+
+      return this;
+    },
+
+    init({ instantSearchInstance, parent }) {
+      localInstantSearchInstance = instantSearchInstance;
+
+      // The `mainHelper` is already defined at this point. The instance is created
+      // inside InstantSerach at the `start` method, which occurs before the `init`
```
francoischalifour:
```suggestion
      // inside InstantSearch at the `start` method, which occurs before the `init`
```

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
...
+      // step.
+      const mainHelper = instantSearchInstance.mainHelper!;
+
+      const initialSearchParameters =
+        // Uses the `searchParameters` for the top level index only, it allow
```
francoischalifour:
```suggestion
        // Uses the `searchParameters` for the top level index only, it allows
```

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
...
+      // `searchClient`. Only the "main" Helper created at the `InstantSearch`
+      // level is aware of the client.
+      helper = algoliasearchHelper(
+        {} as Client,
+        indexName,
+        localWidgets.reduce(enhanceConfiguration, initialSearchParameters)
+      );
+
+      // We forward the call to `search` to the "main" instance of the Helper
+      // which is repsonsible to manage the queries (it's the only one that is
```
francoischalifour:
```suggestion
      // which is responsible for managing the queries (it's the only one that is
```

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
...
+        // responsibility of the main instance. It does not make sense to manage
+        // it at the index level because it's either: all of them or none of them
+        // that are stalled. The queries are performed into a single network request.
+        instantSearchInstance.scheduleStalledRender();
+      });
+
+      // We have to use `!` at the moment because `dervive` is not correctly typed.
+      derivedHelper!.on('result', () => {
+        // The index does not render the results it schedules a new render
+        // to let all the other indices emit their own results. It allow us to
```
francoischalifour:
```suggestion
        // to let all the other indices emit their own results. It allows us to
```

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
...
+          });
+        }
+      });
+    },
+
+    dispose() {
+      localWidgets.forEach(widget => {
+        if (widget.dispose) {
+          // The dispose function is always called once the instance is started
+          // (it's an effect of  `removeWidgets`). The index is initialized and
```
francoischalifour:
```suggestion
          // (it's an effect of `removeWidgets`). The index is initialized and
```

src/types/instantsearch.ts
```diff
@@ -119,10 +120,19 @@ export type DerivedHelper = EventEmitter & {
...
-  widgets: Widget[];
+  widgets: Widget[]; // @TODO: remove once the RoutingManger uses the index
+  mainHelper: Helper | null;
+  mainIndex: Index;
+  insightsClient: AlgoliaInsightsClient | null;
+  templatesConfig: object;
+  _isSearchStalled: boolean;
+  _searchParameters: Partial<SearchParameters>;
+  _createAbsoluteURL(state: Partial<SearchParameters>): string;
+  _onHistoryChange(callback: (state: Partial<SearchParameters>) => void): void;
```
francoischalifour: Why did we add `_onHistoryChange` back?

src/types/widget.ts
```diff
@@ -7,10 +8,12 @@ import {
...
   state: SearchParameters;
   helper: Helper;
   templatesConfig: object;
   createURL(state: SearchParameters): string;
+  onHistoryChange(callback: (state: Partial<SearchParameters>) => void): void;
```
francoischalifour: Why did we add `onHistoryChange` back?

src/lib/index.ts
Haroenv: this is also confusing since the name is "index" and it's not an index file

samouss: The widgets are the binding between the connector + the Vanilla renderer. It's not the case for this one which is renderer agnostic. That's why it's inside `lib` rather than `widgets` it's part of the "core" of InstantSearch.

src/types/instantsearch.ts
samouss: AFAIC we never removed it, neither on [`develop`](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/InstantSearch.js#L179) nor on [`next`](https://github.com/algolia/instantsearch.js/blob/0a21759db095ad912de9ce1fbada5e6b7d786927/src/lib/InstantSearch.js#L184). We can drop it inside its own PR.

src/types/widget.ts
samouss: https://github.com/algolia/instantsearch.js/pull/3892#discussion_r297723400

src/lib/__tests__/index-test.ts
samouss: d0fd1aa6d069a7d48f22976ed7825cbc2e886542

src/lib/__tests__/index-test.ts
samouss: bd2b7f77ad8405f4b862d0929064896905ef03ad

src/lib/__tests__/index-test.ts
samouss: bd2b7f77ad8405f4b862d0929064896905ef03ad

src/lib/__tests__/index-test.ts
samouss: Yes, it does. We do not expose the `derivedHelper` through the `index` widget. This test case checks that the listeners are correctly removed. It assumes that it's enough to consider the `DerivedHelper` removed. I've updated the test name to be more clear. I also added another one that uses the Helper to checks whether or not the `DerivedHelper` has been removed.

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
...
+        throw new Error(
+          withUsage('The widget definition expects a `dispose` method.')
+        );
+      }
+
+      localWidgets = localWidgets.filter(
+        widget => widgets.indexOf(widget) === -1
+      );
+
+      if (localInstantSearchInstance && widgets.length) {
```
eunjae-lee:
```suggestion
      if (localInstantSearchInstance && Boolean(widgets.length)) {
```

src/lib/index.ts
```diff
@@ -0,0 +1,277 @@
...
+      const lastWidget = localWidgets.pop();
+
+      localWidgets = localWidgets.concat(widgets);
+
+      if (lastWidget) {
+        // Second part of the fix for #3148
+        localWidgets = localWidgets.concat(lastWidget);
+      }
+
+      if (localInstantSearchInstance && Boolean(widgets.length)) {
```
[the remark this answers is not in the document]
samouss: Both are equivalent if we remove the check against `localInstantSearchInstance` for `helper`, we remove the `!` for one but we'll have to add it on the other. I used this approach at first for almost all the places where we use `!` inside the module. I wrapped all usages inside a condition but it's "incorrect". The values **are** defined at that point, it adds a bit of complexity that is not correct. We only add those conditions to make the compiler happy, it doesn't change the behavior of the code. 

Then the problem is how the code is structured. I agree with this. The current architecture does not scale very well with the approach I took for multiple indices, but it works for now and it's "simple" enough. If we revamp the core at some point we have to take those constraints into consideration.

src/lib/index.ts
francoischalifour: It think that the reasoning behind @samouss' decision is that when we'll create several InstantSearch packages (e.g. `instantsearch-core`, `instantsearch-vanilla`, etc.), `index` will be exported from `instantsearch-core` because it doesn't include anything render. Thus it makes sense not to consider it as a widget but still to export it in `instantsearch-vanilla`, which right now is the `widgets/` folder.

Is that right @samouss?

samouss: Yes it is.

yannickcr: Well, we also have some widgets in `widgets` that doesn't include anything related to rendering (ex: `analytics` or `configure`). For me is is more about usage than about what is rendered or not.

samouss: The widget `analytics` is tied to the renderer since we rely on the DOM e.g `location`. This is true for `configure` though. Anyway, we won't debate for ages about such small details. I'm fine with a path change, it doesn't matter.

samouss: https://github.com/algolia/instantsearch.js/pull/3892/commits/a8e6d3df4856cb1021f9539f3eaef4176eb77317

src/widgets/index/__tests__/index-test.ts
```diff
@@ -0,0 +1,871 @@
...
+    });
+
+    describe('with a started instance', () => {
+      it('updates the internal state with added widgets', () => {
+        const instance = index({ indexName: 'index_name' });
+        const instantSearchInstance = createInstantSearch();
+
+        instance.addWidgets([createSearchBox()]);
+
+        instance.init!(
```
tkrugg: IIRC, `!` is necessary here because `Index = {...} & Widget` and `Widget` has optional `init` method.
I wonder if it wouldn't have been preferable to enforce `init` (and `render`) on Index (AFAIKT they are always there).
Not sure it brings a lot of value but you like to hear your opinion on this.

src/types/widget.ts
```diff
@@ -1,3 +1,4 @@
+import { Index } from '../widgets/index/index';
```
tkrugg: good call on using the full "path" 😄

src/widgets/index/index.ts
```diff
@@ -0,0 +1,277 @@
...
+      }
+
+      localWidgets = localWidgets.filter(
+        widget => widgets.indexOf(widget) === -1
+      );
+
+      if (localInstantSearchInstance && Boolean(widgets.length)) {
+        const nextState = widgets.reduce((state, widget) => {
+          // the `dispose` method exists at this point we already assert it
+          const next = widget.dispose!({ helper: helper!, state });
```
tkrugg: `dispose` is optional on widget, so shouldn't we check if dispose exists before calling it?
Or maybe we should add spit an error message if `dispose` is not implemented on an added widget.

src/widgets/index/__tests__/index-test.ts
samouss: Yes we can enforce it. Like you said it doesn't bring that much value though. The `init`/`render` methods are only (almost) called directly inside the tests. Most of the time the widgets are inside an array and we have to check whether or not the methods exist before calling them.

samouss: https://github.com/algolia/instantsearch.js/pull/3892/commits/e4e4152031388cf07e5984bc129c83abd96d440b

PR #3893, opened 2019-06-26
feat(index): use widget [PART-2]

samouss:
This PR introduces the usage of the `index` widget through `InstantSearch`. With the previous PR, we saw that the widget is now responsible for managing widgets & state. We now leverage the widget inside `InstantSearch`.

By default, all `InstantSearch` instance has a "main" index that replaces the logic that was previously inside the instance. The instance exposes functions (like `scheduleXXX`) to be able to manage multiple indices at the same time. The `index` widgets are self-contained and they don't have that much knowledge about the external world. The instance is here to "sync" their behavior (like a `render`). The change should be invisible for regular users since the API didn't change.

With these changes, I've merged all the tests related to `InstantSearch` inside the same module. I also get rid of a bunch of them because we cover most of them through the `index` widget. Inside the PR I've tried to linked which test is replacing what to ease the review.


src/lib/__tests__/InstantSearch-test.js
```diff
@@ -1,19 +1,38 @@
...
+    );
+
+    helper.searchOnlyWithDerivedHelpers = jest.fn((...searchArgs) => {
+      return searchOnlyWithDerivedHelpers(...searchArgs);
+    });
 
     return helper;
   });
+
+  Object.entries(module).forEach(([key, value]) => {
```
samouss: Since the import is mocked we were not able to access the static values on `algoliasearchHelper`.

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -151,742 +162,890 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
...
+    // eslint-disable-next-line no-new
+    new InstantSearch({
+      indexName: 'index_name',
+      searchClient: createSearchClient(),
+    });
+
     expect(algoliasearchHelper).not.toHaveBeenCalled();
   });
 
-  it('does not fail when passing same references inside multiple searchParameters props', () => {
```
samouss: https://github.com/algolia/instantsearch.js/blob/802dc7d354e7ccc3e9db0f572877c697dff3c10b/src/lib/__tests__/InstantSearch-test.js#L237-L267

src/lib/__tests__/search-client-test.js
```diff
@@ -1,62 +0,0 @@
-import InstantSearch from '../InstantSearch';
-
-describe('InstantSearch Search Client', () => {
-  describe('Properties', () => {
-    it('throws if no `search()` method', () => {
```
samouss: https://github.com/algolia/instantsearch.js/blob/802dc7d354e7ccc3e9db0f572877c697dff3c10b/src/lib/__tests__/InstantSearch-test.js#L61-L70

src/lib/__tests__/search-client-test.js
```diff
@@ -1,62 +0,0 @@
...
-        new InstantSearch({
-          indexName: '',
-          searchClient: {},
-        });
-      }).toThrowErrorMatchingSnapshot();
-    });
-  });
-
-  describe('Lifecycle', () => {
-    it('gets called on search', () => {
```
samouss: https://github.com/algolia/instantsearch.js/blob/802dc7d354e7ccc3e9db0f572877c697dff3c10b/src/lib/__tests__/InstantSearch-test.js#L445-L457

src/lib/__tests__/search-client-test.js
```diff
@@ -1,62 +0,0 @@
...
-      expect(searchClientSpy.search).not.toHaveBeenCalled();
-
-      search.start();
-
-      expect(search.helper.state.query).toBeUndefined();
-      expect(searchClientSpy.search).toHaveBeenCalledTimes(1);
-      expect(searchClientSpy.search.mock.calls[0][0]).toMatchSnapshot();
-    });
-
-    it('calls the provided searchFunction when used', () => {
```
samouss: https://github.com/algolia/instantsearch.js/blob/802dc7d354e7ccc3e9db0f572877c697dff3c10b/src/lib/__tests__/InstantSearch-test.js#L367-L384

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -151,742 +162,890 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
...
+        searchMetadata: {
+          isSearchStalled: false,
+        },
+      })
+    );
   });
 
-  it('should detach the DerivedHelper', () => {
+  // https://github.com/algolia/instantsearch.js/pull/2623
+  it('does not trigger a re-`render` without results', () => {
```
francoischalifour: This feature is meant for InstantSearch to not render anything until it receives some results, right? (only the first load the first time)

samouss: Yes, otherwise on the first `search` InstantSearch schedules a stalled render. If the delay is reached we call the render, but we don't have the results yet which throws an error.

src/lib/InstantSearch.js
```diff
@@ -296,87 +236,70 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
...
+    const mainHelper = algoliasearchHelper(this.client);
 
-    helper.search = () => {
+    mainHelper.search = () => {
       // This solution allows us to keep the exact same API for the users but
       // under the hood, we have a different implementation. It should be
       // completely transparent for the rest of the codebase. Only this module
       // is impacted.
-      helper.searchOnlyWithDerivedHelpers();
+      return mainHelper.searchOnlyWithDerivedHelpers();
```
Haroenv: seems like this was from a previous PR, but what about:

```
mainHelper.search = mainHelper.searchOnlyWithDerivedHelpers
```

src/lib/InstantSearch.js
```diff
@@ -296,87 +236,70 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
...
+    this.mainIndex.init({
+      instantSearchInstance: this,
+      parent: null,
     });
 
+    mainHelper.search();
+
+    // Keep the previous reference for legacy purpose, some pattern use
+    // the direct Helper access `search.helper` (e.g mutli-index).
+    this.helper = this.mainIndex.getHelper();
```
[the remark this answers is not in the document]
Haroenv: Which kind of warning would you give? "there's no need for this" or "it is wrong when you are using the index widget"?

src/lib/InstantSearch.js
```diff
@@ -389,75 +312,69 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
...
     // The helper needs to be reset to perform the next search from a fresh state.
     // If not reset, it would use the state stored before calling `dispose()`.
-    this.helper.removeAllListeners();
+    this.removeAllListeners();
+    this.mainHelper.removeAllListeners();
+    this.mainHelper = null;
     this.helper = null;
   }
 
+  scheduleSearch = defer(() => {
```
Haroenv: where's this function called?

src/lib/InstantSearch.js
```diff
@@ -389,75 +312,69 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
...
-      });
-    });
-
-    /**
-     * Render is triggered when the rendering of the widgets has been completed
-     * after a search.
-     * @event InstantSearch#render
-     */
-    this.emit('render');
+    this.mainHelper.clearCache().search();
```
Haroenv: This clears all the cache since there's only one search client, so we should be fine :)

src/lib/InstantSearch.js
```diff
@@ -84,21 +85,30 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
+
+    this.started = false;
     this.templatesConfig = {
       helpers: createHelpers({ numberLocale }),
       compileOptions: {},
     };
 
     this._stalledSearchDelay = stalledSearchDelay;
+    this._searchStalledTimer = null;
+    this._isSearchStalled = true;
```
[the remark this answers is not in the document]
francoischalifour: It needs to be stalled at first. See #2623.

src/lib/InstantSearch.js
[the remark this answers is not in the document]
samouss: Both are valid. Users probably don't have to access the `helper` on the instance. There are different way to access it, connectors, widgets, events. The second is also valid, because within a multi-index context the `helper` won't be the full picture of the state anymore.

src/lib/InstantSearch.js
```diff
@@ -296,87 +236,70 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
       );
       this._createURL = routingManager.createURL.bind(routingManager);
       this._createAbsoluteURL = this._createURL;
-      this.widgets.push(routingManager);
+      // We don't use `addWidgets` because we have to ensure that `RoutingManager`
+      // is the last widget added. Otherwise we have an issue with the `routing`.
+      // https://github.com/algolia/instantsearch.js/pull/3149
+      this.mainIndex.getWidgets().push(routingManager);
```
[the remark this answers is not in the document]
samouss: Not sure I understood the statement. Indeed `.widgets` is not available anymore since it's now scoped under the "main" index. The schema provided is incorrect. The `router` has to be scoped under `index` otherwise it does not have access to the rest of the widgets.

Note that for compatibly purpose we can still expose it. I didn't do it by default because (I think) the usage of `.widget` isn't that common. We can mirror the value of the main index if required (like we do for the [`helper`](https://github.com/algolia/instantsearch.js/pull/3893#discussion_r297712527)).

src/lib/InstantSearch.js
samouss: It is called on [`addWidget`](https://github.com/algolia/instantsearch.js/blob/8881841f28240766dded1a49b3dcb89ef3565327/src/lib/index.ts#L106) & [`removeWidget`](https://github.com/algolia/instantsearch.js/blob/8881841f28240766dded1a49b3dcb89ef3565327/src/lib/index.ts#L144) at the index level once the instance is started. An `index` widget could contain one (or many) index widgets. It means that all of them will trigger a new search. To avoid multiple searches we defer them to trigger only one.

src/lib/InstantSearch.js
samouss: Both are equivalent.

src/lib/InstantSearch.js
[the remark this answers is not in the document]
Haroenv: If the router needs to be scoped under the main index, how does it work for any lower indices then? This is why I ask how come it can't be at the root instead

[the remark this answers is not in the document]
samouss: Yes that's right it doesn't work. At the moment the multi-index does not work anyway (we don't merge the parameters). The changes inside this PR mirror the current implementation with the usage of a main index. The routing will be updated to support multiple levels.

PR #3894, opened 2019-06-26
chore(examples): add empty results design to e-commerce demo

francoischalifour:
This adds the empty state when no hits are returned by the engine. It doesn't change the SFFV no results design.

## Implementation

If no filter are applied, the empty message asks the user to change the query. If there are filters applied, the empty message shows a different message and offers the user to clear the filters.

Since it's not possible to add a widget within another widget template, I used the `ClearRefinements` widget within a panel that is hidden when there are results.

I added `removeWordsIfNoResults` to `allOptional` so that results are returned most of the time.

## Preview

- [Preview with only query leading to empty results](https://deploy-preview-3894--instantsearchjs.netlify.com/examples/e-commerce/?query=hellll)
- [Preview with filters leading to empty results](https://deploy-preview-3894--instantsearchjs.netlify.com/examples/e-commerce/?query=helll&hierarchicalMenu%5BhierarchicalCategories.lvl0%5D%5B0%5D=Cameras%20%26%20Camcorders)
- [Default preview](https://deploy-preview-3894--instantsearchjs.netlify.com/examples/e-commerce/)

## Related

- [Desktop design](https://app.zeplin.io/project/5acc8becfc0c980068fc12cc/screen/5d133b0a7695220487ab06e7)
- [Mobile design](https://app.zeplin.io/project/5acc8becfc0c980068fc12cc/screen/5d133b6fc0b0e869459947ca)


examples/e-commerce/src/widgets/Configuration.js
```diff
@@ -3,6 +3,7 @@ import { configure } from 'instantsearch.js/es/widgets';
 const configuration = configure({
   attributesToSnippet: ['description:10'],
   snippetEllipsisText: '…',
+  removeWordsIfNoResults: 'allOptional',
```
tkrugg: I don't understand `allOptional` it would have made more sense to me if you've set it to `none`.
This makes it harder to actually see the "no results/empty" state right ?

francoischalifour:
@tkrugg The error you see was fixed in #3842, it's not from this PR. The base branch wasn't rebased, which is why the fix wasn't propagated here.

examples/e-commerce/src/widgets/Configuration.js
francoischalifour: Yes, which is the goal for an e-commerce use case usually. You don't want to show that there's no results often.

Haroenv:
I can't try the demo out with that error 🤔

francoischalifour:
@Haroenv Try the links in the description. Interacting with the menu can cause the error.

PR #3896, opened 2019-06-27
chore(storybook): add stories with lifecycle

samouss:
It's always painful to test the behavior of a widget when it's added or removed. This PR aims to fix this issue with a decorator that creates a preview to dynamically add/remove the widget. You have to call the decorator inside the `withHits` decorator. I didn't implement a way to chain them because they don't have the same signature (we'll have to implement an adapter for the last one). This approach keeps the flexibility.

I added the `withLifecycle` to the stories when it's relevant: 
- when the `dispose` function actually does something (remove the refinements)
- when the previewed widget does not conflict with the one of the playground

**Usage**

```js
storiesOf("searchBox", module).add(
  "with add/remove",
  withHits(({ search, container, instantsearch }) => {
    withLifecycle(search, container, node =>
      instantsearch.widgets.searchBox({
        container: node
      })
    );
  })
);
```

You can test it live on the [preview](https://deploy-preview-3896--instantsearchjs.netlify.com/stories/?path=/story/menu--with-add-remove).

### DIFF
--- a/.storybook/decorators/withLifecycle.ts
+++ b/.storybook/decorators/withLifecycle.ts
@@ -1,3 +1,5 @@
+import { Widget } from '../../src/types';
+
 const setDisabledState = (element: HTMLButtonElement, state: boolean) => {
   element.disabled = state;
   element.classList.toggle('ais-ClearRefinements-button--disabled');
@@ -6,7 +8,7 @@ const setDisabledState = (element: HTMLButtonElement, state: boolean) => {
 export const withLifecycle = (
   search: any,
   container: HTMLElement,
-  fn: (node: HTMLElement) => void
+  fn: (node: HTMLElement) => Widget
 ) => {
   const actions = document.createElement('div');
   actions.style.marginBottom = '15px';
--- a/.storybook/decorators/withLifecycle.ts
+++ b/.storybook/decorators/withLifecycle.ts
@@ -1,0 +1,54 @@
+const setDisabledState = (element: HTMLButtonElement, state: boolean) => {
+  element.disabled = state;
+  element.classList.toggle('ais-ClearRefinements-button--disabled');
+};
+
+export const withLifecycle = (
+  search: any,
+  container: HTMLElement,
+  fn: (node: HTMLElement) => void
+) => {
+  const actions = document.createElement('div');
+  actions.style.marginBottom = '15px';
+  actions.style.paddingBottom = '15px';
+  actions.style.borderBottom = '1px solid #e4e4e4';
+
+  const description = document.createElement('p');
+  description.textContent =
+    'Click on one of the button to add/remove the widget.';
+
+  const add = document.createElement('button');
+  add.style.marginRight = '10px';
+  add.className = 'ais-ClearRefinements-button';
+  add.textContent = 'Add widget';
+
+  const remove = document.createElement('button');
+  remove.className = 'ais-ClearRefinements-button';
+  remove.textContent = 'Remove widget';
+
+  const node = document.createElement('div');
+
+  actions.appendChild(description);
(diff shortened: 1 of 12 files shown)


stories/breadcrumb.stories.js
```diff
@@ -189,4 +189,30 @@ storiesOf('Breadcrumb', module)
         },
       });
     })
+  )
+  .add(
+    'with add/remove',
+    withHits(
+      ({ search, container, instantsearch }) => {
```
Haroenv:
```suggestion
      ({ search, container, instantsearch }) => 
```

since you leave out the braces in the other one

nbd either way

Haroenv: since it's not in our eslint rule, let's ignore this though

.storybook/decorators/withLifecycle.ts
```diff
@@ -0,0 +1,54 @@
+const setDisabledState = (element: HTMLButtonElement, state: boolean) => {
+  element.disabled = state;
+  element.classList.toggle('ais-ClearRefinements-button--disabled');
+};
+
+export const withLifecycle = (
+  search: any,
+  container: HTMLElement,
+  fn: (node: HTMLElement) => void
```
Haroenv:
```suggestion
  fn: (node: HTMLElement) => Widget
```

.storybook/decorators/withLifecycle.ts
```diff
@@ -0,0 +1,54 @@
...
+  const node = document.createElement('div');
+
+  actions.appendChild(description);
+  actions.appendChild(add);
+  actions.appendChild(remove);
+
+  container.appendChild(actions);
+  container.appendChild(node);
+
+  const widget = fn(node);
```
Haroenv: weird that TS doesn't complain about this, since it should be void

.storybook/decorators/withLifecycle.ts
```diff
@@ -0,0 +1,54 @@
+const setDisabledState = (element: HTMLButtonElement, state: boolean) => {
+  element.disabled = state;
+  element.classList.toggle('ais-ClearRefinements-button--disabled');
+};
+
+export const withLifecycle = (
```
francoischalifour: Any reasons to not go with a similar signature as `withHits` (with a single object)?

.storybook/decorators/withLifecycle.ts
samouss: The widget has a type `void` which is completely valid. The `addWidget` & `removeWidget` are not typed which means that they are inferred as `any`. That's why it does not complain.

.storybook/decorators/withLifecycle.ts
samouss: The main reason is readability. The `withHits` decorator only takes a function as an argument but it's not the case with this one. It feels like with the list of parameters reads better, but we can change it I don't have a strong opinion.

```js
// with list of parameters
withLifecycle(search, container, () =>
    hitsPerPage: 1
  })
);

// with named parameters
withLifecycle({
  search,
  container,
  fn(node) {
    instantsearch.widgets.searchBox({
      container: node
    });
  }
});
```

eunjae-lee: For me, it has to be either `exactly same` or `don't care`. If we will introduce more decorators, then they'd better look the same, so that we can do

```js
withHits(...
  withLifecycle(...
    withFoo(...
      withBar(...
```

It will look cleaner. And ultimately we could do like this:

```js
withDecorators([withHits, withLifecycle, ...], ({ container, instantsearch, search }) => {
  ...
});
```

but if we're not going that far, I'm okay with the current signature.

PR #3897, opened 2019-06-27
chore(examples): e-commerce mobile design part-2

eunjae-lee:
**Summary**

This adds more styles to e-commerce mobile design.

**Result**

This PR will

- force the size for resetIcon in SearchBox as `10x10`, because in Vue-IS and Angular-IS, the sizes are hardcoded and there is no option to override them.
- have `z-index: 5` instead of `1` at `.filtering .container-filters-footer` because there was a collision between the filters footer and the slide UIs in Vue-IS and Angular-IS.
- add class names to the children of `.container-options` because `.container-options div:...` didn't select correctly at Angular-IS.
- move slider-related styles to dedicated `PriceSlider.css` file as same as the flavors are doing.

### DIFF
--- a/examples/e-commerce/src/app.css
+++ b/examples/e-commerce/src/app.css
@@ -90,7 +90,7 @@ h2 {
   padding: 30px 0;
 }
 
-.container-options .container-option:not(:first-of-type) {
+.container-options .container-option:not(:first-child) {
   margin-left: 48px;
 }
 
--- a/examples/e-commerce/src/app.css
+++ b/examples/e-commerce/src/app.css
@@ -279,15 +279,6 @@ mark {
   padding: 16px 24px;
 }
 
-/* RangeSlider */
-
-.ais-RangeSlider .rheostat-tooltip::before {
-  color: #e2a400;
-  content: '$';
-  font-size: 0.6;
-  margin-right: 4px;
-}
-
 /* ToggleRefinement */
 
 .ais-ToggleRefinement-label {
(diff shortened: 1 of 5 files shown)


examples/e-commerce/src/app.css
```diff
@@ -90,7 +90,7 @@ h2 {
   padding: 30px 0;
 }
 
-.container-options div:not(:first-of-type) {
+.container-options .container-option:not(:first-of-type) {
```
eunjae-lee: This doesn't work for Angular-IS.

image

Do you have any idea?
`:first-child` works, though.
I don't get the difference here.

francoischalifour: > Using :first-child is very similar to :first-of-type but with one critical difference: it is less specific. :first-child will only try to match the immediate first child of a parent element, while first-of-type will match the first occurrence of a specified element, even if it doesn't come absolutely first in the HTML.

Source: https://css-tricks.com/almanac/selectors/f/first-child/

Angular probably creates some sub-elements making `:first-of-type` ineffective.

eunjae-lee: Yes according to the descriptions, it's supposed to work.
I'll change it to `:first-child`.

francoischalifour: Angular creates web components and I don't think we can use pseudo elements on web components. Hence, only `:first-child` works on the first child of the web component.

PR #3914, opened 2019-07-02
feat(index): add story [PART-3]

samouss:
This PR adds a minimal story to work on the `index` widget. We can review the design later, right now it's mostly to be able to have a playground to work on. The current feature set is limited to only display a static list of hits (it trigger the correct number of requests). We only take into account the local `helper` of the `index` which means that parent widgets are ignored. We'll cover that soon.

~~**Currently the story does not work because the previous branches are not correctly rebased.**~~

> https://deploy-preview-3914--instantsearchjs.netlify.com/stories/?path=/story/index--default

### DIFF
--- a/.storybook/playgrounds/default.ts
+++ b/.storybook/playgrounds/default.ts
@@ -1,5 +1,25 @@
 import instantsearch from '../../src/index';
 
+export const hitsItemTemplate = `
+<div
+  class="hits-image"
+  style="background-image: url({{image}})"
+></div>
+<article>
+  <header>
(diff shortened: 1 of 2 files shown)


Haroenv:
I'm not sure if this is expected, but a query in the main index doesn't yet propagate up right?

samouss:
Yes it's expected see the PR description:

> The current feature set is limited to only display a static list of hits (it trigger the correct number of requests). We only take into account the local `helper` of the `index` which means that parent widgets are ignored. We'll cover that soon.

PR #3915, opened 2019-07-02
chore(examples): add media demo

francoischalifour:
This updates the media demo with a new dataset and a new design.

## Notes

- ~~I need to create replicas of the index for the `SortBy` widget.~~
- ~~The "No results" designs are missing.~~
- ~~The `SearchBox` `showReset` option is disabled, waiting for the design.~~

## Preview

[See preview →](https://deploy-preview-3915--instantsearchjs.netlify.com/examples/media/)

## Related

- [Design](https://app.zeplin.io/project/5d1393093588af695d431e9e/screen/5d1393af1ab3c4042c7a14d5)


Haroenv:
looks like the filter dropdown's transition isn't synchronised to the hits :)

Haroenv:
If you don't type a query  or refine in any way, the Date filter isn't doing anything because the latest results (which are within the last day) are all on the first page by default. Maybe it would be interesting to have some results from other days which do rank on the first page

Haroenv:
Screenshot 2019-07-03 at 11 27 23

If the languages overflow, the design breaks for the other elements too, probably because it's a grid?

examples/media/src/widgets/Articles.ts
```diff
@@ -0,0 +1,151 @@
...
+import { getCurrentDate } from '../utils';
+
+const currentDate = getCurrentDate();
+
+function createHit(hit, { isHighlighted }) {
+  return `
+<li class="ais-InfiniteHits-item${
+    isHighlighted ? ' infinite-hits-item--highlighted' : ''
+  }">
+  <article class="card">
```
tkrugg: This could looks better with placeholders to minimize motion on image load.
https://css-tricks.com/building-skeleton-screens-css-custom-properties/

francoischalifour: Definitely! I'll work on this to make progress on researching how to better handle low-network connections with InstantSearch.

francoischalifour: I added a basic implementation in 51c6f413f29e19d8d474011ed1a38e6632d36fbb.

examples/media/src/widgets/Articles.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+let globalIsLastPage = false;
+
+const infiniteHits = connectInfiniteHits(
+  ({ results, hits, showMore, isLastPage, widgetParams }, isFirstRender) => {
+    const { container } = widgetParams;
+    const containerNode = document.querySelector(container);
+
+    globalIsLastPage = isLastPage;
```
yannickcr: Why do you need `globalIsLastPage`?

francoischalifour: Since we're in the `isFirstRender` branch, `isLastPage` is only captured once (during `init`) and is therefore always `true` in the intersection condition. We thus need to store it outside of this branch to be able to update it for other renders.

yannickcr: I see now, thanks :+1:

robertmogos (dismissed):

francoischalifour:
⚠️ Adding the "DO NOT MERGE" tag to still be able to make changes to the deployment URL that has been communicated.

examples/media/.htmlnanorc
```diff
@@ -0,0 +1,7 @@
+{
+	"minifySvg": {
+		"plugins": [
+			{ "cleanupIDs": false }
```
eunjae-lee: FYI: https://github.com/parcel-bundler/parcel/issues/2474

devdreamsolution (approved):

dhayab:
I don't know why the codesandbox build failed or how to restart it if necessary. Is there a way to get more detailed logs from them?

examples/media/src/search.ts
```diff
@@ -0,0 +1,61 @@
...
+
+const searchClient = algoliasearch(
+  '1QDAWL72TQ',
+  '47700f55d95d23f5a57744b9a027ea83'
+);
+
+const search = instantsearch({
+  searchClient,
+  indexName: 'PROD_algolia_blog',
+  routing: true,
```
Haroenv, line 24:
```suggestion
  routing: { stateMapping: singleIndex('PROD_algolia_blog') },
```

Haroenv:
We are tracking the page in the URL, but don't have showPrevious and/or infiniteHitsCache enabled, meaning https://deploy-preview-3915--instantsearchjs.netlify.app/examples/media/?PROD_algolia_blog%5Bpage%5D=3 for examples doesn't show the first results at all

examples/media/index.html
```diff
@@ -1,63 +1,221 @@
...
-    <script src="search.js"></script>
+      </div>
+
+      <section class="container-results">
+        <div data-widget="hits"></div>
+      </section>
+    </main>
+
+    <aside class="back-to-top">
+      <a href="#" title="Back to top">
```
Haroenv, line 197: as the href is #, it removes the search parameters, not sure what can be used to keep search though

dhayab: I can try to prevent its default behavior, so it won't follow through the route change with the #.

examples/media/src/search.ts
```diff
@@ -21,7 +21,7 @@ const searchClient = algoliasearch(
 const search = instantsearch({
   searchClient,
   indexName: 'PROD_algolia_blog',
-  routing: true,
+  routing: { stateMapping: singleIndex('PROD_algolia_blog') },
```
Haroenv, line 24: sorry for not mentioning in the earlier suggestion, but this needs to be imported still

dhayab: no worries, i added it in c8de92a.

examples/media/index.html
dhayab: I changed the "back to top" behavior in 6561751.

Haroenv (dismissed):
good as a first start, but adding the dates back would be useful IMO

examples/media/src/search.ts
```diff
@@ -0,0 +1,64 @@
...
+  container: '[data-widget="clear-filters-mobile"]',
+});
+const authorsDesktop = createAuthors({
+  container: '[data-widget="categories-desktop"]',
+});
+const authorsMobile = createAuthors({
+  container: '[data-widget="categories-mobile"]',
+});
+
+search.addWidgets([
```
Haroenv, line 49: according to @LorrisSaintGenez, there's an attribute `created_at_timestamp` that can be used for the date

dhayab: Cool 👍 , I'll re-add the date widgets.

dhayab:
> We are tracking the page in the URL, but don't have showPrevious and/or infiniteHitsCache enabled, meaning https://deploy-preview-3915--instantsearchjs.netlify.app/examples/media/?PROD_algolia_blog%5Bpage%5D=3 for examples doesn't show the first results at all

I didn't notice that. I'll add a button to show previous results in this case.

examples/media/src/search.ts
dhayab: Done in 4c280c2.

examples/media/src/widgets/Dates.ts
```diff
@@ -0,0 +1,50 @@
...
+        end: getUnixTime(endOfDay(currentDate)),
+      },
+      {
+        label: 'This week',
+        start: getUnixTime(startOfWeek(currentDate)),
+        end: getUnixTime(endOfWeek(currentDate)),
+      },
+      {
+        label: 'This month',
+        start: getUnixTime(startOfMonth(currentDate)),
```
Haroenv, line 41: I first thought this didn't work, but as it's the beginning of the month, there were only a few days for items to show up. Maybe the items should be "today", "last 7 days", "last 30 days", "last year"?

Haroenv (commented):
it's working!

examples/media/src/widgets/Dates.ts
dhayab: Haha this puzzled me as well for a bit, as I also expected a "last X" behavior. I could borrow the wording from Hacker News, as it's used widely in production. Wdyt?

image

Haroenv: yes, that wording & behaviour seems more consistent

Haroenv:
This is a nitpick, but the date & category filters don't look the same, because of the padding on the search box. Not sure how to solve that tbh, as the padding is *inside*

Screenshot 2021-11-05 at 10 09 27

Haroenv (approved):
all this taken in account, I think this lgtm

examples/media/.htmlnanorc
```diff
@@ -0,0 +1,7 @@
+{
```
francoischalifour, line 1: Nit: there are tabs instead of spaces in this file.

examples/media/src/search.ts
```diff
@@ -0,0 +1,75 @@
...
+  },
+});
+
+const datesDesktop = createDates({
+  container: '[data-widget="dates-desktop"]',
+  header: 'Date',
+});
+const datesMobile = createDates({
+  container: '[data-widget="dates-mobile"]',
+  header: ' Upload Time',
```
francoischalifour, line 37: Upload time sounds a bit "tech-y" for a blog. Maybe "Date" is enough?

examples/media/src/search.ts
```diff
@@ -0,0 +1,75 @@
...
+const selectedTopicsMobile = createSelectedTopics({
+  container: '[data-widget="selected-topics-mobile"]',
+});
+const clearFiltersDesktop = createClearFilters({
+  container: '[data-widget="clear-filters-desktop"]',
+});
+const clearFiltersMobile = createClearFilters({
+  container: '[data-widget="clear-filters-mobile"]',
+});
+const authorsDesktop = createAuthors({
```
francoischalifour, line 51: On desktop, the Authors refinement are before the Date refinement. On mobile, it's the opposite. I would suggest we always put "Date" first.

dhayab:
I made some adjustments related to alignment and the "previous results" CTA, thanks to input from Nana.

PR #3917, opened 2019-07-03
feat(index): add mergeSearchParameters function

Haroenv:
```js
mergeSearchParameters(a, b, c) // returns a new search parameters
```

**Summary**

- overrides left-to-right on uncontrolled parameters
- merges controlled parameters
- overrides left-to-right on duplicate attributes**Result**

note: linting fails, but opening a separate PR to fix that


src/lib/utils/mergeSearchParameters.ts
```diff
@@ -0,0 +1,125 @@
...
+    },
+  });
+
+const mergeHierarchicalFacetsRefinements: Merger = right => left =>
+  left.setQueryParameters({
+    hierarchicalFacetsRefinements: {
+      ...left.hierarchicalFacetsRefinements,
+      ...right.hierarchicalFacetsRefinements,
+    },
+  });
```
yannickcr: Could worth it to make a generic merge function to reduce duplication and save on bundle size. WDYT?
```js
type GenericMerger = (
  key: keyof SearchParameters,
  right: SearchParameters
) => Merge;

const mergeGeneric: GenericMerger = (key, right) => left =>
  left.setQueryParameters({
    [key]: {
      ...left[key],
      ...right[key],
    },
  });

// ...

compose(
  mergeGeneric('hierarchicalFacetsRefinements', right),
  mergeGeneric('numericRefinements', right),
  // ...
)(left)
```

samouss: Yep I did it at first but revert it afterwards because it wasn't worth it IMO.

src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -0,0 +1,370 @@
...
+          colors: ['Red'],
+        },
+      })
+    );
+
+    expect(actual).toEqual(
+      algoliasearchHelper.SearchParameters.make({
+        facets: ['brand', 'categories', 'colors'],
+        facetsRefinements: {
+          brand: ['Apple'],
```
francoischalifour: Are there any cases where we wouldn't want `Samsung` to be overridden by `Apple` but rather appended? (same question for other objects containing arrays)

src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -0,0 +1,370 @@
+import algoliasearchHelper from 'algoliasearch-helper';
+import merge from '../mergeSearchParameters';
+
+describe('mergeSearchParameters', () => {
+  it('merges non-managed parameters', () => {
```
Haroenv:
```suggestion
  it('overrides non-managed parameters', () => {
```

src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -0,0 +1,370 @@
...
+    );
+
+    expect(actual).toEqual(
+      algoliasearchHelper.SearchParameters.make({
+        disjunctiveFacets: ['brand', 'categories', 'colors'],
+      })
+    );
+  });
+
+  it('merges `facetsRefinements` parameters', () => {
```
Haroenv:
```suggestion
  it('merges `facetsRefinements` parameters, overrides conflicts', () => {
```

I think a title like this would explain more the behaviour we want. It could also be split up into separate tests

src/lib/utils/__tests__/mergeSearchParameters-test.ts
samouss: @Haroenv had the same question. The reasoning behind this implementation is to keep the behavior consistent between primitives and managed parameters. Applied to widgets it helps to understand:

```js
search.addWidgets([
  index({ indexName: "topLevel" }).addWidgets([
    // Top level searchBox
    searchBox(),
    index({ indexName: "subLevel" }).addWidgets([
      // Sub level searchBox
      searchBox()
    ])
  ])
]);
```

The two widgets `search box` are independent from each other. Each widget **controls** its part of the state where it is mounted. The sub-level `searchBox` does not inherit from the top-level one, it overrides it. We apply the same rule to the managed parameters that are on the **same attribute**. Note that this rule only applies to managed parameters shared across multiple indices. Within the same index both widgets can read/write on the internal state.

With React InstantSearch I've never seen a use case where multiple widgets inherit from each other within a multi-index context. The behavior of React InstantSeach is the same as the one described above. I agree that it's a tradeoff but to it's the behavior that makes sense (IMO) with the idea of controlled/uncontrolled widgets.

src/lib/utils/mergeSearchParameters.ts
```diff
@@ -0,0 +1,125 @@
+import { SearchParameters } from '../../types';
+
+type Merge = (value: SearchParameters) => SearchParameters;
+
+type Merger = (right: SearchParameters) => Merge;
+
+const compose = <TValue>(...fns: Array<(x: TValue) => TValue>) => (
```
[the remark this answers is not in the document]
francoischalifour: If you have other pieces of code in mind where to apply this pattern, we can extract it as a util. Otherwise, we can rewrite the code without this.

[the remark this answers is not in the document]
samouss: Nope I don't.

src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -0,0 +1,370 @@
...
+        facetsRefinements: {
+          brand: ['Apple'],
+          categories: ['TVs'],
+          colors: ['Red'],
+        },
+      })
+    );
+  });
+
+  it('merges `facetsExcludes` parameters', () => {
```
Haroenv:
```suggestion
  it('merges `facetsExcludes` parameters, overrides conflicts', () => {
```

src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -0,0 +1,370 @@
...
+        facetsExcludes: {
+          brand: ['Apple'],
+          categories: ['TVs'],
+          colors: ['Red'],
+        },
+      })
+    );
+  });
+
+  it('merges `disjunctiveFacetsRefinements` parameters', () => {
```
Haroenv:
```suggestion
  it('merges `disjunctiveFacetsRefinements` parameters, overrides conflicts', () => {
```

src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -0,0 +1,370 @@
...
+        disjunctiveFacetsRefinements: {
+          brand: ['Apple'],
+          categories: ['TVs'],
+          colors: ['Red'],
+        },
+      })
+    );
+  });
+
+  it('merges `numericRefinements` parameters', () => {
```
Haroenv:
```suggestion
  it('merges `numericRefinements` parameters, overrides conflicts', () => {
```

src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -0,0 +1,370 @@
...
+          },
+          vote: {
+            '>=': [50],
+          },
+        },
+      })
+    );
+  });
+
+  it('merges `tagRefinements` parameters', () => {
```
Haroenv:
```suggestion
  it('merges `tagRefinements` parameters, overrides conflicts', () => {
```

src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -0,0 +1,370 @@
...
+    );
+
+    expect(actual).toEqual(
+      algoliasearchHelper.SearchParameters.make({
+        tagRefinements: ['brand', 'categories', 'colors'],
+      })
+    );
+  });
+
+  it('merges `hierarchicalFacets` parameters', () => {
```
Haroenv:
```suggestion
  it('merges `hierarchicalFacets` parameters, overrides conflicts', () => {
```

src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -0,0 +1,370 @@
...
+          {
+            name: 'folders',
+            attributes: ['folders.lvl0', 'folders.lvl1'],
+          },
+        ],
+      })
+    );
+  });
+
+  it('merges `hierarchicalFacetsRefinements` parameters', () => {
```
Haroenv:
```suggestion
  it('merges `hierarchicalFacetsRefinements` parameters, overrides conflicts', () => {
```

src/lib/utils/__tests__/findIndex-test.ts
```diff
@@ -0,0 +1,33 @@
+import findIndex from '../findIndex';
+
+describe('findIndex', () => {
+  describe('with polyfill', () => {
+    test('with empty array', () => {
```
francoischalifour: The case without an array is not covered while it's implemented.

src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -0,0 +1,370 @@
...
+      })
+    );
+  });
+
+  it('merges `disjunctiveFacetsRefinements` parameters, overrides conflicts', () => {
+    const actual = merge(
+      algoliasearchHelper.SearchParameters.make({
+        disjunctiveFacets: ['brand'],
+        disjunctiveFacetsRefinements: {
+          brand: ['Samsung'],
```
francoischalifour: Since we now consider default `disjunctivefacetsRefinements` to be empty arrays (e.g. algolia/algoliasearch-helper-js#725), we should add empty arrays in our test cases (valid for all array refinements).

src/lib/utils/mergeSearchParameters.ts
```diff
@@ -0,0 +1,145 @@
...
+const mergeFacets: Merger = (left, right) =>
+  right.facets!.reduce((_, name) => _.addFacet(name), left);
+
+const mergeDisjunctiveFacets: Merger = (left, right) =>
+  right.disjunctiveFacets!.reduce(
+    (_, name) => _.addDisjunctiveFacet(name),
+    left
+  );
+
+const mergeTagRefinements: Merger = (left, right) =>
```
samouss: This should be grouped with the "Merge facets refinements".

src/lib/utils/mergeSearchParameters.ts
```diff
@@ -0,0 +1,145 @@
...
+      facetsExcludesMerged,
+      right
+    );
+    const disjunctiveFacetsMerged = mergeDisjunctiveFacets(
+      facetRefinementsMerged,
+      right
+    );
+    const facetsMerged = mergeFacets(disjunctiveFacetsMerged, right);
+
+    return mergeWithRest(facetsMerged, right);
```
samouss: The previous order was more understandable IMO, even though in the end it doesn't matter.

src/lib/utils/__tests__/mergeSearchParameters-test.ts
samouss: The change you've pointed at does not impact the function. The default value is only returned by the getter, it's not stored like this on the `SearchParameters`. We don't use getter in this situation we use the "raw" attributes of `SearchParameters` which does not contain default values for each attributes.

francoischalifour: Ah, right!

francoischalifour:
`mergeSearchParameters` probably shouldn't be in `lib/utils` but rather `lib/`. This is part of the core (same for `enhanceConfiguration`). We can move them another time though. WDYT?

Haroenv:
I'm not too sure what the distinction between both is, let's deal with that another time, since I think there might be more inconsistencies @francoischalifour

src/widgets/index/index.ts
```diff
@@ -205,17 +206,9 @@ const index = (props: IndexProps): Index => {
...
-
-        // @TODO: replace this dummy merge with the correct function
-        return parameters.reduce((previous, current) =>
-          algoliasearchHelper.SearchParameters.make({
-            ...previous,
-            ...current,
-          })
-        );
-      });
+      derivedHelper = mainHelper.derive(() =>
```
```suggestion
      derivedHelper = mainHelper.derive(() =>
```

PR #3919, opened 2019-07-03
chore(deps): eslint-config-algolia 13.4.0

Haroenv:
**Summary**

This version removes the requirement to type return parameters.
**Result**

#3917 will no longer have linting errors


francoischalifour:
Why was this targeting `next`? It should have been `develop`.

PR #3921, opened 2019-07-03
chore(examples): add routing to e-commerce demo

francoischalifour:
## Description

This is my take at adding a routing system to the e-commerce demo.

This is definitely not easy to implement and we probably have improvements to do inside the library.

## Implementation

I based the implementation on the [SEO-friendly guide](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/js/#seo-friendly-urls).

The routing values read in the URL might lead to a wrong UI state. I added fallback strategies to overcome that.

- `SortBy`: when we detect that the `sortBy` value in the URL is not valid, we don't accept the value and fallback to `undefined`. Otherwise, InstantSearch tries to target the invalid index and throws.
- `HitsPerPage`: we fallback to `undefined` if it's not listed in the items.
- `RatingMenu`: we fallback to `undefined` if it's not listed in the items.
- `RangeSlider`: there's no easy way to know if we're out of range. It could be possible to access it via `facets_stats` inside the library but not with the public API.

This fallback behaviors could be implemented in the library routing system.

## Preview

[See preview →](http://deploy-preview-3921--instantsearchjs.netlify.com/examples/e-commerce/)


Haroenv:
We should set up netlify so /examples/*/search shows the same as examples/*

Haroenv:
Netlify now correctly redirects, but the urls for JS & CSS are wrong (relative, instead of relative towards root). Solution might be a base tag, but that might break the urls we create with createURL if those are relative too

francoischalifour:
URLs now work on Netlify with:

- 0d5c2d1f9183e475eafc870013cee40db69ef83d
- e547b0282bbab28bdb50eb4cc1080f3d48d7e193

This is ready for review.

PR #3930, opened 2019-07-08
fix(InstantSearch): cancel scheduled operations

samouss:
Some operations are now asynchronous, which means that we have to clean up those operations when the `InstantSearch` instance is disposed. Otherwise, the functions are called and might refer to props that have been pruned. This PR "cancel" the pending operations on `dispose`.

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -312,6 +312,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
    * @return {undefined} This method does not return anything
    */
   dispose() {
+    this.scheduleSearch.cancel();
     this.scheduleRender.cancel();
 
     this.removeWidgets(this.mainIndex.getWidgets());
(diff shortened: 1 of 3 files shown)


src/lib/InstantSearch.js
```diff
@@ -312,6 +312,10 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
    * @return {undefined} This method does not return anything
    */
   dispose() {
+    this.scheduleSearch.cancel();
+    this.scheduleRender.cancel();
+    clearTimeout(this._searchStalledTimer);
```
samouss: We could use the defer approach for the stalled timer, the implementation would be the exact same one. But right now, the `defer` function does not accept a timeout since it relies on `Promise`. It's easy to change its implementation to use `setTimeout` but will have to update the tests. For the moment this implementation is good enough.

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -506,6 +508,82 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
...
+    await runAllMicroTasks();
+
+    // The call to `addWidgets` schedules a new search
+    search.addWidgets([createWidget()]);
+
+    search.dispose();
+
+    // Without the cancel operation, the function call throws an error which
+    // prevents the test to complete. We can't assert that the function throws
+    // because we don't have access to the promise that throws in the first place.
```
eunjae-lee: I tried to assert somehow, but it was difficult. So, we could go on with this way :)

PR #3931, opened 2019-07-08
refactor(test): POC integration tests

yannickcr:
**Summary**

Our current test suite is way too close to the implementation details and the actual widgets usage is not truthfully tested, making it harder for us to be confident in the test suite. Any change in the codebase can potentially break multiple tests even if we did not break anything in reality. And at the same time its "freezing" some projects details making internal changes more complicated than it should be.

IMHO it would benefit us to write less unit tests and more integration tests, closer on how our widgets and connectors are used by real users.

This PR is a POC to use [Testing Library](https://testing-library.com) on InstantSearch.js to write such integration tests.

For this POC I tested only one flavor but the same tools and principles can be applied on other flavors as well.

## Guiding principles

> The more your tests resemble the way your software is used, the more confidence they can give you. - Testing Library guiding principles

### Widgets

* **Usage**: The widget must be used the same way a user will use it. It means initializing an `instantsearch` instance, using the `addWidget` method then calling the `start` method).
* **Interactions**: Changes are only triggered using simulated user interactions (using [Testing Library events API](https://testing-library.com/docs/dom-testing-library/api-events).
* **Multiple widgets**: Multiple widgets can be initialized in a test to see if a particular widget interact correctly with another one.
* **Validation**: Results must only be tested by checking the widget output (DOM) or, in some cases, the potential side effects (API request).

### Connectors

* **Usage**: The connector must be used the same way a user will use it. It means calling `instantsearch.connectors.connect*` with a custom render function (a [Jest mock function](https://jestjs.io/docs/en/jest-object#jestfnimplementation) for our case).
* **Interactions**: Changes are only triggered using methods passed to the rendering function (captured using our [Jest mock function](https://jestjs.io/docs/en/jest-object#jestfnimplementation)).
* **Multiple widgets**: Multiple widgets can be initialized in a test to see if a particular widget interact correctly with another one.
* **Validation**: Results must only be tested by checking parameters passed to the rendering function or, in some cases, the potential side effects (API request).

Using theses principles we limit ourselves to use and interact with our widgets/connectors like a real user, limiting the needs for mocking and the risk to rely on internals for testing.

Resources:
 * [Testing Library](https://testing-library.com)
 * [jest-dom](https://github.com/testing-library/jest-dom)
 * [Kent C. Dodds – Write tests. Not too many. Mostly integration](https://www.youtube.com/watch?v=Fha2bVoC8SE)

**PS**: Other tests are failing since I removed the `preact-compat` -> `react` alias in the Babel configuration. Both test suites can coexists but I did not take the time to tweak the configuration for this POC.


package.json
```diff
@@ -108,6 +109,7 @@
     "inquirer": "6.4.1",
     "jest": "24.8.0",
     "jest-diff": "24.8.0",
+    "jest-dom": "3.5.0",
```
tkrugg: shouldn't this be @testing-library/jest-dom ?
It has been moved apparently: https://www.npmjs.com/package/jest-dom

yannickcr: Nice catch. Apparently their documentation is not up to date yet https://testing-library.com/docs/ecosystem-jest-dom

src/widgets/breadcrumb/__tests__/breadcrumb-testing-library-test.js
```diff
@@ -0,0 +1,286 @@
...
+
+      await wait();
+
+      queryByText(container, 'Cameras & Camcorders').click();
+
+      await wait();
+
+      expect(queryByText(container, 'Home')).toBeTruthy();
+      expect(queryByText(container, 'Cameras & Camcorders')).toBeTruthy();
+      expect(queryByText(container, 'Digital Cameras')).toBeNull();
```
tkrugg: It could be nice to snapshot the calls to `.search` during each interaction.
One of the issues we had with the hierarchicalMenu is that it triggered too many requests

tkrugg:
Looks really good.

it's still not a real browser, so still not as close as what we ship to the end user but will definitely catch a lot of bugs!

I see there is a certain level of compatibility with react and vue. It'd be be nice to check we can reach the same level of feature coverage in every flavour we have. 
There is nothing for angular but maybe we can work something out with puppeteer or cypress.

src/widgets/breadcrumb/__tests__/breadcrumb-testing-library-test.js
francoischalifour: Adding an assertion with the number of requests (i.e. how many times the `searchClient.search()` method was called) might make more sense than snapshotting its calls. In the end, this testing strategy is about what is the state of the app at each step, not necessarily how we got there (this can be considered as an implementation detail). Keeping track of the number of requests can still be an important metric because it has a business impact and should be considered as a feature.

src/connectors/breadcrumb/__tests__/connectBreadcrumb-testing-library-test.js
```diff
@@ -0,0 +1,245 @@
...
+      indexName: 'instant_search',
+      searchClient,
+    });
+  });
+
+  afterEach(() => {
+    try {
+      search.dispose();
+    } catch (e) {
+      // Fail if the search was never started, ignoring
```
Haroenv: can wrap in if (search.started) instead

francoischalifour: We don't want to use `search.started` because it's an internal.

src/widgets/breadcrumb/__tests__/breadcrumb-testing-library-test.js
yannickcr: I think it makes sense to count and/or snapshots API calls if we consider these requests as a side effect from the widgets.

This could simply be asserted during a test with something like:

```js
// ...

search.start();

await wait();

expect(searchClient.search).toHaveBeenCalledTimes(1);
expect(searchClient.search.mock.calls[0]).toMatchSnapshot();

queryByText(container, 'Home').click();

await wait();

expect(searchClient.search).toHaveBeenCalledTimes(2);
expect(searchClient.search.mock.calls[1]).toMatchSnapshot();

yannickcr:
@tkrugg 

> I see there is a certain level of compatibility with react and vue. It'd be be nice to check we can reach the same level of feature coverage in every flavour we have.

It should be pretty similar but I can try to reproduce these tests in react/vue/angular to compare the different solutions.

> There is nothing for angular but maybe we can work something out with puppeteer or cypress.

There is also an [Angular Testing Library](https://testing-library.com/docs/angular-testing-library/intro)

@francoischalifour 

> Did you spot some cases where this strategy hits limitation compared to the way we currently test our exposed API?

Not yet, but I only tested a very simple widget for now. Do you have something in mind?

@Haroenv 

> I think this is useful, but it needs a document where we write which kind of tests should be tested in which way to have a consistent experience

Yes, I think it could be documented the same way @francoischalifour  did for our [current testing guidelines](https://docs.google.com/document/d/1EUi72bcz9RzrlNrRrkIsLNDjsygpD0yN9ptjp8v57Hc/edit?pli=1).

src/connectors/breadcrumb/__tests__/connectBreadcrumb-testing-library-test.js
yannickcr: It seems publicly exposed, but yes it is not documented :neutral_face: 

Also while making the change to use `search.started` I found an issue in `connectBreadcrumb`. If you do not provide an `unmountFn` during widget creation then `search.dispose()` crashes (no default value).

francoischalifour: @yannickcr Some connectors still don't provide a default value for the unmount function. This is known but we haven't fixed that yet.

Regarding `search.started`, I don't think we should use it because it is not supposed to be exposed. This is an internal variable accessible because we use a class for `InstantSearch` and we store `started` on `this`. If we were to refactor this class to a function or even rename `started`, this will fail. I believe we should _strictly_ restrict ourselves to the public API with this move.

yannickcr: Agree to restrict ourselves to the public API

yannickcr:
> Last year, I took this class from Kent C. Dodds.

Lucky you! I'd love to take that class. :slightly_smiling_face: 

We can also test directly the production build with the integration tests (tested it [here](https://github.com/algolia/instantsearch.js/pull/3931/commits/5d91e78f7fbfb2eda8be7bf491d3e64adb8a9111)) but we need to see how to make it works with tests running during development.

Agree for E2E tests, but it asks more efforts to be put in place so I see it more like a second step to do after the integration tests.

eunjae-lee:
> Agree for E2E tests, but it asks more efforts to be put in place so I see it more like a second step to do after the integration tests.

Sure, e2e test is way out of the scope for now :)

PR #3932, opened 2019-07-09
fix(index): prevent render without results

samouss:
The `index` assumes that the `lastResults` prop is set which is not always the case. The issue arises only when the stalled render kicks in. The previous implementation (without `index`) fixed this issue by turning the default value of `_isSearchStalled` to `true` to avoid extra `render` until we have results (https://github.com/algolia/instantsearch.js/pull/2623). This fix does not work anymore since we don't have a single Helper. 

This PR revert the change to a default value of `false` which makes more sense (it's not stalled by default). Only the `index` is aware that we have results or not, then the condition belongs to its render function.

**Before**

before

**After**

after

### DIFF
--- a/src/widgets/index/index.ts
+++ b/src/widgets/index/index.ts
@@ -238,16 +238,18 @@ const index = (props: IndexProps): Index => {
     render({ instantSearchInstance }) {
       localWidgets.forEach(widget => {
         // At this point, all the variables used below are set. Both `helper`
-        //  and `derivedHelper` has been created at the `init` step. The attribute
-        // `lastResults` is set before the event `result` is emitted. At this stage,
-        // the event has emitted hence the value is already set.
+        // and `derivedHelper` has been created at the `init` step. The attribute
+        // `lastResults` might be `null` though. It's possible that a stalled render
+        // happen before the result e.g with a dynamically added index the request might
+        // be delayed. The render is triggered for the complete tree but some parts do
+        // not have results yet.
 
-        if (widget.render) {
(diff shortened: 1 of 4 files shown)


Haroenv:
The issue with `<mark>` in your screencast, is that related or #3830?

samouss:
> The issue with `<mark>` in your screencast, is that related or #3830?

Yes, the branch is not up to date with `develop`.

src/widgets/index/index.ts
```diff
@@ -238,16 +238,18 @@ const index = (props: IndexProps): Index => {
     render({ instantSearchInstance }) {
       localWidgets.forEach(widget => {
         // At this point, all the variables used below are set. Both `helper`
-        //  and `derivedHelper` has been created at the `init` step. The attribute
-        // `lastResults` is set before the event `result` is emitted. At this stage,
-        // the event has emitted hence the value is already set.
+        // and `derivedHelper` has been created at the `init` step. The attribute
```
francoischalifour:
```suggestion
        // and `derivedHelper` have been created at the `init` step. The attribute
```

src/widgets/index/index.ts
```diff
@@ -238,16 +238,18 @@ const index = (props: IndexProps): Index => {
...
+        // `lastResults` might be `null` though. It's possible that a stalled render
+        // happen before the result e.g with a dynamically added index the request might
```
francoischalifour:
```suggestion
        // happens before the result e.g with a dynamically added index the request might
```

PR #3933, opened 2019-07-09
fix(defer): recover from error

samouss:
This PR fixes an issue when the deferred function throws an error. The next calls to the function shouldn't be impacted by the error that happens inside a previous call. Previously it was the case because the value of `progress` was reset only if the call to the callback succeeds. It's not the case anymore the value is always reset whether or not the call succeeds.

### DIFF
--- a/src/lib/utils/__tests__/defer.ts
+++ b/src/lib/utils/__tests__/defer.ts
@@ -128,4 +128,36 @@ describe('defer', () => {
       `"The deferred function should be called before calling \`wait()\`"`
     );
   });
+
+  it('recovers a deferred function that throws an error', async () => {
+    const fn = jest.fn();
+    const deferred = defer(fn);
+
+    fn.mockImplementation(() => {
+      throw new Error('FAIL');
+    });
+
+    deferred();
+
+    expect(fn).toHaveBeenCalledTimes(0);
+
+    try {
+      await deferred.wait();
+    } catch {
+      // The test verifies that the function is able to recover. We don't want
+      // to terminate the test on this expected error.
+    }
+
+    expect(fn).toHaveBeenCalledTimes(1);
+
+    fn.mockImplementation();
+
+    deferred();
+
+    expect(fn).toHaveBeenCalledTimes(1);
+
+    await deferred.wait();
+
+    expect(fn).toHaveBeenCalledTimes(2);
+  });
 });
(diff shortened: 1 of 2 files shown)


src/lib/utils/defer.ts
```diff
@@ -1,35 +1,48 @@
...
         return;
       }
 
       callback(...args);
-      progress = null;
     });
   };
 
+  fn.wait = () => {
+    if (progress === null) {
```
francoischalifour: Do we need this check in production code? If not, we can wrap it in a `__DEV__` condition.

samouss: WDYM? If we don't perform the check the the function will throw an unexpected the error.

src/lib/utils/__tests__/defer.ts
```diff
@@ -104,8 +104,60 @@ describe('defer', () => {
...
+  it('waits for the deferred function', async () => {
+    const fn = jest.fn();
+    const deferred = defer(fn);
+
+    deferred();
+
+    await deferred.wait();
+
+    expect(fn).toHaveBeenCalledTimes(1);
+  });
```
tkrugg: Is this test really necessary?
it seems redundant with https://github.com/algolia/instantsearch.js/pull/3933/files#diff-7ad0c7bb4b40181d3ab0d95a0a4b1678R4

src/lib/utils/defer.ts
francoischalifour: From what I understand, this check is only performed for us to never call `defer` before `wait`, right? I guess my reasoning is that this shouldn't happen at runtime in the user bundle so it might not be worth including it in the end bundle.

src/lib/utils/__tests__/defer.ts
samouss: Yes indeed wasn't sure about it. If it breaks then all of them break. I'll remove it.

src/lib/utils/defer.ts
samouss: I wouldn't care that much, it will save ~90 bytes. I can remove it and let it throw an error but that could bite us later I think.

src/lib/utils/__tests__/defer.ts
samouss: https://github.com/algolia/instantsearch.js/pull/3933/commits/fb1a34c908aee8b5f5cc6fb8e00f960391422ad9

PR #3935, opened 2019-07-10
fix(routing): support parsing URLs with up to 100 refinements

yannickcr:
## Description

Backport of #3671 for InstantSearch v2.

Original description:

> An issue raised by a customer is that when selecting more than 20 values with the `routing` option activated, the data structure returned by the `parseURL` function changes.
> 
> `qs` by default converts arrays with more than 20 items to an object. We want to avoid this because the data structure manipulated can therefore vary.
> 
> Setting the limit to `100` seems a good number because the engine's default is 100. It can go up to 1000 but it is very unlikely to select more than 100 items in the UI.
> 
> ## Reference
> 
> - `qs` doc: https://github.com/ljharb/qs#parsing-arrays
> - Algolia `maxValuesPerFacet` doc: https://www.algolia.com/doc/api-reference/api-parameters/maxValuesPerFacet/
> - Original customer issue: https://algolia.slack.com/archives/C2XP6HBKQ/p1553005092109200

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,6 +98,8 @@
     "html-webpack-plugin": "2.30.1",
     "inquirer": "5.2.0",
     "jest": "23.6.0",
+    "jest-environment-jsdom": "24.8.0",
+    "jest-environment-jsdom-global": "1.2.0",
     "jest-watch-typeahead": "0.2.0",
     "jsdom-global": "3.0.2",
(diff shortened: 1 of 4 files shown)


francoischalifour:
@Haroenv Since this is an easy fix and that this one was implemented already, I guess it's fine?

PR #3936, opened 2019-07-10
fix(clearRefinements): reset page to 0

eunjae-lee:
**Summary**

`clearRefinements` is not resetting the page to 0.
This is not an expected behaviour and brings a confusion to users.
When everything is reset, the page should be reset to 0.

**Result**

Now it resets page to 0.


src/connectors/clear-refinements/__tests__/connectClearRefinements-test.js
```diff
@@ -517,5 +517,26 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/clear-refin
...
+      widget.init({
+        helper,
+        state: helper.state,
+        createURL: () => '#',
+      });
+      const clearRefinements = rendering.mock.calls[0][0].refine;
+
+      helper.setPage(2);
+      clearRefinements();
+      expect(helper.getPage()).toBe(0);
```
samouss: Prefer the usage of `helper.state.page`. We never know we might remove those getters at some point.

src/lib/utils/clearRefinements.js
```diff
@@ -21,6 +21,7 @@ function clearRefinements({ helper, attributesToClear = [] }) {
   if (attributesToClear.indexOf('query') !== -1) {
     finalState = finalState.setQuery('');
   }
+  finalState = finalState.setPage(0);
```
samouss: We could have inlined the call into the first assignment.

eunjae-lee: 1fc7375

src/connectors/clear-refinements/__tests__/connectClearRefinements-test.js
eunjae-lee: 43f9c39

PR #3937, opened 2019-07-10
feat(index): resolve parent SearchParameters

samouss:
This PR introduces a function to resolve the parent `SearchParameters` from the `index`. It allows the index to find the list of parameters that it will have to merge. Currently, the function to merge those parameters is not merged yet that's why we have a naive implementation. We'll migrate to the correct function once it's available.

At the moment the function to resolve the `SearchParameters` is fairly simple: it recursively bubbles up the hierarchy to find the root `index`. At each iteration, we collect the current state of the `index` where we are. Note that we have to expose the `parent` on the `index` instance otherwise we're not able to retrieve the parents' state.

The [examples](https://deploy-preview-3937--instantsearchjs.netlify.com/stories/?path=/story/index--default) now inherits from the `searchBox`, only the primitive values are merged (because of the naive merge function e.g. `refinementList` does not update the children indices).

**Before**

before

**After**

after

### DIFF
--- a/src/lib/utils/resolveSearchParameters.ts
+++ b/src/lib/utils/resolveSearchParameters.ts
@@ -1,0 +1,16 @@
+import { Index } from '../../widgets/index/index';
+import { SearchParameters } from '../../types';
+
+const resolveSearchParameters = (current: Index): SearchParameters[] => {
+  let parent = current.getParent();
+  let states = [current.getHelper()!.state];
+
+  while (parent !== null) {
+    states = [parent.getHelper()!.state].concat(states);
+    parent = parent.getParent();
+  }
+
+  return states;
+};
+
+export default resolveSearchParameters;
--- a/src/widgets/index/__tests__/index-test.ts
+++ b/src/widgets/index/__tests__/index-test.ts
@@ -582,6 +582,108 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
       );
     });
 
+    it('inherits from the parent states for the queries', () => {
+      const level0 = index({ indexName: 'level_0_index_name' });
+      const level1 = index({ indexName: 'level_1_index_name' });
+      const level2 = index({ indexName: 'level_1_index_name' });
+      const searchClient = createSearchClient();
+      const mainHelper = algoliasearchHelper(searchClient, '', {});
+      const instantSearchInstance = createInstantSearch({
+        mainHelper,
+      });
+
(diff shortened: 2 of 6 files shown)


src/lib/utils/resolveSearchParameters.ts
```diff
@@ -0,0 +1,16 @@
+import { Index } from '../../widgets/index/index';
+import { SearchParameters } from '../../types';
+
+const resolveSearchParameters = (current: Index): SearchParameters[] => {
+  let parent = current.getParent();
+  let states = [current.getHelper()!.state];
+
+  while (parent !== null) {
```
Haroenv: could also have been done recursively and then you wouldn't need the duplicated declaration, but this is easier to follow I guess

src/widgets/index/__tests__/index-test.ts
```diff
@@ -582,6 +582,108 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
       );
     });
 
+    it('inherits from the parent states for the queries', () => {
+      const level0 = index({ indexName: 'level_0_index_name' });
+      const level1 = index({ indexName: 'level_1_index_name' });
+      const level2 = index({ indexName: 'level_1_index_name' });
```
francoischalifour: The index name wasn't changed on purpose?

src/lib/utils/resolveSearchParameters.ts
samouss: Yep, I wrote it recursively at first. I think it's easier to follow like this.

```ts
const resolveSearchParameters = (current: Index): SearchParameters[] => {
  const _resolveSearchParameters = (
    state: SearchParameters[],
    parent: Index | null
  ) => {
    if (parent === null) {
      return state;
    }

    return _resolveSearchParameters(
      [parent.getHelper()!.state].concat(state),
      parent.getParent()
    );
  };

  return _resolveSearchParameters(
    [current.getHelper()!.state],
    current.getParent()
  );
};
```

src/widgets/index/__tests__/index-test.ts
samouss: Nope, will fix it.

samouss: https://github.com/algolia/instantsearch.js/pull/3937/commits/10d548e16691a1e03ed31b0b565e438aa466f3d2

PR #3940, opened 2019-07-11
chore(helper): update to latest canary

Haroenv:
**Summary**

This removes possible errors in favour of empty values, as well as adding type definitions

**Result**

- [x] updated type definitions to use the definitions of the helper instead
- [x] checked all changed previously error cases (#3934)


package.json
```diff
@@ -126,7 +126,7 @@
     "scriptjs": "2.5.9",
     "semver": "6.0.0",
     "shelljs": "0.8.3",
-    "typescript": "3.4.5",
+    "typescript": "3.5.3",
```
Haroenv: needed for `Omit` used by the helper's types

test/utils/createAlgoliaResponse.ts
```diff
@@ -0,0 +1,38 @@
...
+    page = 0,
+    hitsPerPage = 20,
+    nbHits = 0,
+    nbPages,
+    processingTimeMS = 0,
+    hits = [],
+    query = '',
+    params = '',
+    exhaustiveNbHits = true,
+    exhaustiveFacetsCount = true,
```
Haroenv: TS fails here, but that's fixed by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36859

src/connectors/query-rules/__tests__/connectQueryRules-test.ts
```diff
@@ -136,7 +140,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/query-rules
           instantSearchInstance,
           state: helper.state,
           results: new SearchResults(helper.state, [
-            { hits: [], userData: [{ banner: 'image.png' }] },
+            createAlgoliaResponse({ userData: [{ banner: 'image.png' }] }),
```
Haroenv: TS fails, but that's fixed by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36859

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -1,10 +1,12 @@
+import {
```
Haroenv: TS fails on 

https://github.com/algolia/instantsearch.js/pull/3940/files#diff-571e9df908c4237e29ced537895105b7R92

but I don't know what the solution is here, making the `getRefinements` function accept any doesn't really make sense to me

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -495,7 +500,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     ];
 
     const results = new SearchResults(helper.state, [
-      { hits, queryID: 'theQueryID' },
+      createAlgoliaResponse({ hits, queryID: 'theQueryID' }),
```
Haroenv: TS also fails here, but fixed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36859

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -88,7 +87,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
         showMore: expect.any(Function),
         results: expect.any(Object),
         isFirstPage: true,
-        isLastPage: false,
```
Haroenv: without the mock of the search response it got this wrong, since nbPages should be the same as page (both 0)

src/lib/utils/getRefinements.ts
```diff
@@ -165,7 +165,8 @@ function getRefinements(
   Object.keys(numericRefinements).forEach(attributeName => {
     const operators = numericRefinements[attributeName];
 
-    Object.keys(operators).forEach(operator => {
+    Object.keys(operators).forEach(operatorOriginal => {
+      const operator = operatorOriginal as SearchParameters.Operator;
```
Haroenv: putting this in the arguments didn't work

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -87,7 +90,9 @@ function getRuleContextsFromTrackedFilters({
   const ruleContexts = Object.keys(trackedFilters).reduce<string[]>(
     (facets, facetName) => {
       const facetRefinements: TrackedFilterRefinement[] = getRefinements(
-        helper.lastResults || {},
+        // Note that this isn't correct, but getRefinements only uses properties,
```
francoischalifour: We should make it explicit that this comment is about the typing.

```suggestion
        // An empty object is technically not a `SearchResults` but `getRefinements`
        // only accesses properties, meaning it will not throw with an empty object.
```

Can't we work around that by changing the `getRefinements` signature with only partial search parameters?

src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -87,7 +90,9 @@ function getRuleContextsFromTrackedFilters({
...
+        // meaning it will not throw with an empty object
```
francoischalifour:
```suggestion
```

src/types/instantsearch.ts
```diff
@@ -1,25 +1,12 @@
...
-// That's a proxy to avoid manipulating the original `algoliasearch-helper`
-// SearchParameters typings and to add newer search parameters not yet
-// documented or wrongly typed.
-export type SearchParameters = AlgoliaSearchHelperSearchParameters & {
-  ruleContexts?: string[];
-};
-
-export type SearchResults = AlgoliaSearchHelperSearchResults;
-
+// @TODO: can this be written some other way?
```
francoischalifour: WDYM?

test/mock/createAPIResponse.ts
```diff
@@ -1,21 +1,39 @@
...
+    processingTimeMS = 0,
+    hits = [],
+    query = '',
+    params = '',
+    exhaustiveNbHits = true,
+    exhaustiveFacetsCount = true,
+    ...rest
+  } = subset;
+  const fallbackNbPages =
+    nbPages !== undefined ? nbPages : Math.ceil(nbHits / hitsPerPage);
```
francoischalifour: That's kind of meta but it could be worth testing this test util since many of our tests depend on it.

test/mock/createAPIResponse.ts
```diff
@@ -1,21 +1,39 @@
...
+    query,
+    params,
+    exhaustiveNbHits,
+    exhaustiveFacetsCount,
+    ...rest,
+  };
+}
 
-export const createMutliSearchResponse = (
+export const createMultiSearchResponse = (
```
francoischalifour: Good catch haha

test/mock/createAPIResponse.ts
Haroenv: agree, but not sure if this is really worth testing, should we test only that line?

francoischalifour: That's the one where there's more logic so I'd say this is the most "critical". No strong opinion on the rest though.

src/types/instantsearch.ts
Haroenv: I am wondering if it could be extracted from the helper's types, but it isn't really essential

src/lib/RoutingManager.ts
```diff
@@ -43,7 +42,7 @@ class RoutingManager implements Widget {
   private getAllSearchParameters({
     currentSearchParameters,
     uiState,
-  }): Partial<SearchParameters> {
+  }): PlainSearchParameters {
```
samouss: This is incorrect, the return type expected is `SearchParameters`.

src/types/instantsearch.ts
samouss: Yes it would be interesting rather than inlining the types inside the listeners.

test/mock/createAPIResponse.ts
```diff
@@ -1,21 +1,41 @@
...
-  page: 0,
-  nbPages: 0,
-  hitsPerPage: 4,
-  processingTimeMS: 1,
-  query: '',
-  params: '',
-  index: 'index_name',
-  ...args,
-});
+export function createSingleSearchResponse<THit = any>(
```
samouss: You can keep the function expression signature with generic:

```ts
export const createSingleSearchResponse = <THit = any>(
  subset: Partial<Response<THit>> = {}
): Response<THit> => {
  // ...
}
```

test/mock/createAPIResponse.ts
```diff
@@ -1,21 +1,41 @@
...
+    hits = [],
+    query = '',
+    params = '',
+    exhaustiveNbHits = true,
+    exhaustiveFacetsCount = true,
+    ...rest
+  } = subset;
+  const fallbackNbHits = nbHits !== undefined ? nbHits : hits.length;
+  const fallbackNbPages =
+    nbPages !== undefined ? nbPages : Math.ceil(fallbackNbHits / hitsPerPage);
```
samouss: I don't think the tests are useful for this. We could leverage the default value to make them easier to understand. 

```ts
export function createSingleSearchResponse() {
  const {
    hits = [],
    hitsPerPage = 20,
    nbHits = hits.length,
    nbPages = Math.ceil(nbHits / hitsPerPage)
  } = args;

  // ...
}
```

test/mock/createAPIResponse.ts
Haroenv: was trying to find how to write it, but decided function was also fine, thanks!

test/mock/createAPIResponse.ts
Haroenv: had it there earlier, but thought it might have been too complicated. Seeing it written out again, I'll flip back to that form

PR #3941, opened 2019-07-12
feat: drop support for onHistoryChange

samouss:
This PR drops the support for `onHistoryChange`. This API have been introduced to fix [an issue with the `searchBox`](https://github.com/algolia/instantsearch.js/issues/729). The codebase has evolve since then and we don't use the option anymore. The options is not documented on [algolia.com/doc](https://www.algolia.com/doc/), IMO it's safe enough to remove it on the next major.


samouss:
@Haroenv Not yet, will do a pass once at the end.

PR #3944, opened 2019-07-12
feat(index): merge `ruleContexts` search parameter

francoischalifour:
This adds support for the [`ruleContexts`](https://www.algolia.com/doc/api-reference/api-parameters/ruleContexts/) search parameter in the merge function.


src/lib/utils/__tests__/mergeSearchParameters-test.ts
```diff
@@ -367,4 +367,31 @@ describe('mergeSearchParameters', () => {
       })
     );
   });
+
+  it('merges and dedupes `ruleContexts` parameters', () => {
+    const actual = merge(
+      algoliasearchHelper.SearchParameters.make({
+        ruleContexts: ['ais-genre-comedy'],
```
francoischalifour: The `make` function is not yet properly typed in the JS helper at this point and is not aware of the `ruleContexts` search parameter. Rebasing should work.

Haroenv: see https://github.com/algolia/instantsearch.js/pull/3940 which will have a fix for that

src/lib/utils/mergeSearchParameters.ts
```diff
@@ -103,6 +105,23 @@ const mergeHierarchicalFacetsRefinements: Merger = (left, right) =>
...
 
+const mergeRuleContexts: Merger = (left, right) => {
+  const ruleContexts: string[] = uniq(
+    ([] as any)
+      .concat(left.ruleContexts)
+      .concat(right.ruleContexts)
+      .filter(Boolean)
+  );
+
+  if (ruleContexts.length > 0) {
```
francoischalifour: Note that I add the `ruleContexts` key in the search parameters only when not empty. Do we want to always add it?

src/lib/utils/mergeSearchParameters.ts
```diff
@@ -103,6 +105,23 @@ const mergeHierarchicalFacetsRefinements: Merger = (left, right) =>
     },
   });
 
+const mergeRuleContexts: Merger = (left, right) => {
+  const ruleContexts: string[] = uniq(
+    ([] as any)
+      .concat(left.ruleContexts)
+      .concat(right.ruleContexts)
+      .filter(Boolean)
```
Haroenv: why do you need to filter empty values out here?

francoischalifour: `right.ruleContexts` might be `undefined`.

src/lib/utils/mergeSearchParameters.ts
samouss: I would say omit when it doesn't make sense.

PR #3945, opened 2019-07-12
 chore(enhanceConfiguration): swap mergeDeep for mergeSearchParameters

### LINKED ISSUE #3278: Duplicate facets inside the request
Original Vue InstantSearch issue: https://github.com/algolia/vue-instantsearch/issues/587 

The issue seems to come from the [`enhanceConfiguration`](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/InstantSearch.js#L484-L510) function that compute the initial search parameters. Inside this function we do a union on of both value when it's an array. It can leads to duplicated value like we have inside this issue. In most cases the problem won't be visible unless in that case with the hot reload.

https://github.com/algolia/instantsearch.js/blob/1e86b2e74064c2334f2eb3c8793c3250b78fe27b/src/lib/InstantSearch.js#L496-L498

cc @dulnan

> https://codesandbox.io/s/1qoq47086j

-----

**Bug 🐞**

### What is the current behavior?
When using an ais-refinement-list and ais-menu, while also having routing (using historyRouter and simpleMapping) enabled, the hierarchical facets are added multiple times to the search parameters.

A small CodeSandbox showing this behavior:
https://codesandbox.io/s/9324rjjz3p
Upon page load, using the browser's network activity inspector, the request to Algolia servers shows that `categories` has been added to `facets` twice.

The bug is especially noticeable when working locally, as every time one changes something in a Vue file and a refresh happens, the duplicated hierarchical facets are duplicated again. Which, due to routing with state mapping being enabled, will result in the app hanging for several seconds or even crashing.

image
image

This can best be reproduced by opening components/Facet.vue and adding a line break in the template, so a refresh happens.

### What is the expected behavior?
The hierarchicalFacets should only be added once.

### Does this happen only in specific situations?
When using ais-menu or ais-refinement-list together with routing enabled.

### What is the proposed solution?
Unfortunately I haven't figured out exactly where the bug is coming from.

### What is the version you are using?
2.0.0-beta.1

Haroenv:
This is a follow-up on #3917, to easily be able to see the perf impact this has. It's expected that we lose some performance, but it's a tradeoff we can make to fix some bugs in wrongly merging parameters

fixes #3278


src/lib/utils/enhanceConfiguration.ts
```diff
@@ -1,5 +1,6 @@
-import { SearchParameters, Widget } from '../../types';
-import mergeDeep from './mergeDeep';
+import { Widget } from '../../types';
+import mergeSearchParameters from './mergeSearchParameters';
+import { SearchParameters } from 'algoliasearch-helper';
```
Haroenv: looks like we will need #3940 for this to fully work

samouss: Nope it's not linked. It's because the Helper is written in cjs (the [error](https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module) come from Rollup). To avoid the error without any Rollup configuration use:

```js
import algoliasearchHelper from 'algoliasearch-helper';

// ...

new algoliasearchHelper.SearchParameters()
```

Haroenv: Good call! There’s more places than this with the destructuring pattern. Should we just move them all to the dot style or only the ones the build fails for?

samouss: Inside the source code we have other one? It would surprise me that they don't fail. Inside the test is fine though since we don't run Rollup on them.

src/lib/utils/enhanceConfiguration.ts
```diff
@@ -13,7 +15,10 @@ function enhanceConfiguration(
   // Get the relevant partial configuration asked by the widget
   const partialConfiguration = widget.getConfiguration(configuration);
 
-  return mergeDeep(configuration, partialConfiguration);
+  return mergeSearchParameters(
+    new algoliasearchHelper.SearchParameters(configuration),
```
samouss: We could have updated the signature to avoid the creation of this `SearchParameters` on each call.

```ts
type enhance = (configuration: SearchParameters, widget: Widget) => SearchParameters
```

Haroenv: And call the constructor before passing it to enhance?

samouss: Yes.

PR #3954, opened 2019-07-15
chore(examples): fix back-to-top button

eunjae-lee:
**Summary**

- This PR fixes the problem where the backgroud of the button was not visible on desktop Chrome.
- This PR hides the button when the filter modal is open.

### DIFF
--- a/examples/media/index.html
+++ b/examples/media/index.html
@@ -216,13 +216,13 @@
       <a href="#" title="Back to top">
         <svg viewBow="0 0 40 40">
           <defs>
-            <linearGradient id="a" x1="100%" x2="0%" y1="100%" y2="0%">
+            <linearGradient id="d" x1="100%" x2="0%" y1="100%" y2="0%">
               <stop offset="0%" stop-color="#764ba2" />
               <stop offset="100%" stop-color="#667eea" />
             </linearGradient>
           </defs>
           <g fill="none" fill-rule="evenodd">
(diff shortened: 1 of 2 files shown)


examples/media/index.html
```diff
@@ -216,13 +216,13 @@ <h2 class="container-filters-title">
       <a href="#" title="Back to top">
         <svg viewBow="0 0 40 40">
           <defs>
-            <linearGradient id="a" x1="100%" x2="0%" y1="100%" y2="0%">
+            <linearGradient id="d" x1="100%" x2="0%" y1="100%" y2="0%">
```
eunjae-lee: https://github.com/algolia/instantsearch.js/blob/8f9a52feb0691f7511e84e3fa18cdc804538e408/examples/media/index.html#L58-L61

There is another `id="a"`. I think it collided with this. So I had to rename the id to display the background. I thought it was scoped somehow even though it's an `id`, but apparently, it's not..?
It's weird that this duplicated `id`s, it still works fine on Safari.

francoischalifour: Ah, good catch! We probably should have renamed them more explicitly.

eunjae-lee:
> The "back to top" button doesn't appear on Chrome on my machine. It does on Safari. Do you have the same behavior?

It looks like it's not deployed properly.

image

This should be replaced with `id="d"`.

Let me re-deploy it.

PR #3955, opened 2019-07-15
fix: pass noop as default value to unmountFn at connectors

eunjae-lee:
**Summary**

Many of the connectors doesn't have default value of `unmountFn`. When users create custom widget using the connectors and if they don't provide `unmountFn`, removing the widget will throw an error.
So this PR adds `noop` as the default value of `unmountFn` to all the connectors.

**Result**

All the connectors have `noop` as default value of `unmountFn`.


eunjae-lee:
> this is indeed consistent with the existing ones which have the pattern, and I'll trust you that you didn't forget one

Yup, I went through all the connectors one by one.

PR #3957, opened 2019-07-15
chore: merge develop into next

francoischalifour:
This syncs the `next` branch with the latest changes on the `develop` branch. Doing this from time to time will help maintaining the two different codebases.


francoischalifour:
I'm gonna need to change some imports and that should be good.

PR #3958, opened 2019-07-17
fix: remove useless types 

samouss:
We leverage the types of the Helper now.


samouss:
> IIUC we can also remove the HelperChangeEvent type now

Nope we can't we use the type inside connectQueryRules.

https://github.com/algolia/instantsearch.js/blob/e2ed6c7b07161f9f6827263afed5744959873bb3/src/connectors/query-rules/connectQueryRules.ts#L196

Issue #2084, opened 2017-03-31
instantsearch.js/hierarchicaMenu rootPath option buggy behaviour


vvo:
**Do you want to request a *feature* or report a *bug*?**

A bug.

Reproducible case: https://jsfiddle.net/1mfdwqcg/18/

**Bug: What is the current behavior?**

When using the hiearchicalMenu and the rootPath option, I get this behaviour:

![2017-03-31 15 14 00](https://cloud.githubusercontent.com/assets/123822/24551953/e7f16782-1624-11e7-85dc-f8696e5e4433.gif)

**Bug: What is the expected behavior?**

It should be like this:

![2017-03-31 15 14 32](https://cloud.githubusercontent.com/assets/123822/24551958/ed7bdfb6-1624-11e7-9c0e-33bbb107b777.gif)

**Bug: What browsers are impacted? Which versions?**

All

**What project are you opening an issue for?**
- instantsearch.js

**What is the version you are using? Always use the latest one before opening a bug issue.**

Latest

Haroenv:
This issue is really old, but has good SEO. I'm adding on here a possible new case where rootPath will behave wrongly (in v4):

1. have a hierarchical refinement on the page which is not the root refinement
2. add a hierarchical menu with a rootPath to the page
3. notice it does not take the rootPath in account, and keeps the old refinement

This happens because we might not be calling `.make` on `SearchParameters` (only place where rootPath "exists" / is put back where something else is refined).

This bug is mostly hypothetical, and should be confirmed if someone has it

PR #3959, opened 2019-07-17
chore(enhanceConfiguration): prevent helper to be superfluously created

Haroenv:
follow-up on #3945, enhanceConfiguration now accepts a SearchParameters, and all usages have been changed to instantiate this helper themselves.

I wonder if we could force the result of getConfiguration be a SearchParameters too so we can avoid that instantiation? 🤔

### DIFF
--- a/src/connectors/infinite-hits/connectInfiniteHits.ts
+++ b/src/connectors/infinite-hits/connectInfiniteHits.ts
@@ -1,5 +1,5 @@
 import escapeHits, { TAG_PLACEHOLDER } from '../../lib/escape-highlight';
-import algoliasearchHelper, {
+import {
   AlgoliaSearchHelper as Helper,
   SearchParameters,
 } from 'algoliasearch-helper';
@@ -143,16 +143,16 @@ const connectInfiniteHits: InfiniteHitsConnector = (
     return {
       $$type: 'ais.infiniteHits',
 
-      getConfiguration() {
+      getConfiguration(config) {
         const parameters = {
           page: 0,
         };
 
         if (!escapeHTML) {
-          return new algoliasearchHelper.SearchParameters(parameters);
+          return config.setQueryParameters(parameters);
         }
 
-        return new algoliasearchHelper.SearchParameters({
+        return config.setQueryParameters({
           ...parameters,
           ...TAG_PLACEHOLDER,
         });
--- a/src/connectors/infinite-hits/connectInfiniteHits.ts
+++ b/src/connectors/infinite-hits/connectInfiniteHits.ts
@@ -1,5 +1,5 @@
 import escapeHits, { TAG_PLACEHOLDER } from '../../lib/escape-highlight';
-import {
+import algoliasearchHelper, {
   AlgoliaSearchHelper as Helper,
   SearchParameters,
 } from 'algoliasearch-helper';
@@ -149,13 +149,13 @@ const connectInfiniteHits: InfiniteHitsConnector = (
         };
 
         if (!escapeHTML) {
-          return parameters;
+          return new algoliasearchHelper.SearchParameters(parameters);
         }
 
-        return {
+        return new algoliasearchHelper.SearchParameters({
           ...parameters,
           ...TAG_PLACEHOLDER,
-        };
+        });
       },
 
       init({ instantSearchInstance, helper }) {
--- a/src/connectors/voice-search/connectVoiceSearch.ts
+++ b/src/connectors/voice-search/connectVoiceSearch.ts
@@ -77,7 +77,7 @@ const connectVoiceSearch: VoiceSearchConnector = (
       $$type: 'ais.voiceSearch',
 
       getConfiguration(config) {
-        return config.setQuery(config.query || '');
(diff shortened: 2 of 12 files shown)


src/lib/utils/__tests__/enhanceConfiguration-test.js
```diff
@@ -85,4 +90,20 @@ describe('enhanceConfiguration', () => {
       })
     );
   });
+
+  it('does not duplicate hierarchicalFacets (object in array)', () => {
```
Haroenv: this test failed with the previous implementation and is added to prove #3278 is fixed

src/lib/utils/enhanceConfiguration.ts
```diff
@@ -16,7 +14,7 @@ function enhanceConfiguration(
   const partialConfiguration = widget.getConfiguration(configuration);
```
samouss: The function `getConfiguration` expects a `PlainSearchParameters`, but I guess it's fine since both type are compatible.

src/lib/RoutingManager.ts
```diff
@@ -136,7 +136,7 @@ class RoutingManager implements Widget {
   }
 
   public getConfiguration(
-    currentConfiguration: PlainSearchParameters
+    currentConfiguration: SearchParameters
```
samouss: The previous type was correct.

src/types/widget.ts
```diff
@@ -93,7 +93,7 @@ export interface Widget {
   render?(options: RenderOptions): void;
   dispose?(options: DisposeOptions): SearchParameters | void;
   getConfiguration?(
-    previousConfiguration?: PlainSearchParameters
+    previousConfiguration?: SearchParameters
```
samouss: The previous type was correct.

src/widgets/index/index.ts
```diff
@@ -174,7 +168,11 @@ const index = (props: IndexProps): Index => {
       const initialSearchParameters =
         // Uses the `searchParameters` for the top level index only, it allows
         // us to have the exact same behaviour than before for the mono-index.
-        parent === null ? instantSearchInstance._searchParameters : {};
+        parent === null
+          ? new algoliasearchHelper.SearchParameters(
+              instantSearchInstance._searchParameters
```
samouss: We could have nest the condition inside the argument, that would ease the reading.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -147,13 +147,13 @@ const connectInfiniteHits: InfiniteHitsConnector = (
         };
 
         if (!escapeHTML) {
-          return parameters;
+          return new algoliasearchHelper.SearchParameters(parameters);
```
samouss: Why don't we use the given value? It would avoid to use the constructor.

src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -74,10 +74,8 @@ const connectVoiceSearch: VoiceSearchConnector = (
     const { searchAsYouSpeak = false } = widgetParams;
 
     return {
-      getConfiguration() {
-        return {
-          query: '',
-        };
+      getConfiguration(config) {
+        return config.setQuery(config.query || '');
```
samouss: The behaviour with the previous value should be covered by a test. IMO it's better to implement the "previous" behaviour in a separate PR. We can do a pass on each widget at the same time.

src/lib/RoutingManager.ts
```diff
@@ -137,20 +134,18 @@ class RoutingManager implements Widget {
   }
 
   public getConfiguration(
-    currentConfiguration: PlainSearchParameters
-  ): PlainSearchParameters {
+    currentConfiguration: SearchParameters
+  ): SearchParameters {
     // We have to create a `SearchParameters` because `getAllSearchParameters`
     // expects an instance of `SearchParameters` and not a plain object.
     const currentSearchParameters = algoliasearchHelper.SearchParameters.make(
```
samouss: This call is useless since now the value should always be a `SearchParameters`.

src/connectors/infinite-hits/connectInfiniteHits.ts
Haroenv: setting any parameter calls the constructor anyway

src/widgets/index/index.ts
```diff
@@ -171,10 +165,11 @@ const index = (props: IndexProps): Index => {
       // step.
       const mainHelper = instantSearchInstance.mainHelper!;
 
-      const initialSearchParameters =
+      const initialSearchParameters = new algoliasearchHelper.SearchParameters(
```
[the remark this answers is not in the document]
Haroenv: We decided to leave rootPath as is, and use `.make` nowhere in the codebase until it can be proved necessary: https://github.com/algolia/instantsearch.js/issues/2084#issuecomment-512862729

Haroenv:
Yes, the typescript widgets alone have been changed to match the signature, doing the pass on all widgets is part of https://algolia.atlassian.net/browse/IFW-874

PR #3960, opened 2019-07-17
feat(widgets): add `$$type` to widgets definition

francoischalifour:
Adding a type to the `index` widget is necessary to filter it out from the list of the widgets. This is needed to implement the "reset page" behavior to get all the `index`-type children and reset their own page.

I took this opportunity to add a type to all widgets for debugging purpose. The widgets sharing the same connector have the same type. This is fine for now. We can provide an option to override it from the widget parameters later if needed.

The `$$type` property is not required in TypeScript because once we expose the widget interface, users might not need to mark them.

### DIFF
--- a/src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
+++ b/src/connectors/autocomplete/__tests__/connectAutocomplete-test.js
@@ -15,6 +15,24 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/autocomplet
 `);
   });
 
+  it('is a widget', () => {
+    const render = jest.fn();
+    const unmount = jest.fn();
+
+    const customAutocomplete = connectAutocomplete(render, unmount);
+    const widget = customAutocomplete({});
+
+    expect(widget).toEqual(
+      expect.objectContaining({
+        $$type: 'ais.autocomplete',
+        init: expect.any(Function),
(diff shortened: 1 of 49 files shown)


francoischalifour:
@Haroenv I added an assertion ensuring that widgets are "somehow widgets" in each file: https://github.com/algolia/instantsearch.js/blob/6bbce4111b31a04fecd15e7bf4b21d649258cc73/src/connectors/hierarchical-menu/__tests__/connectHierarchicalMenu-test.js#L56-L66

The goal would be to make this assertion stricter with time by removing the `expect.objectContaining()` because we sometimes return more objects in this scope (some are used for testing purpose).

I'm not a fan of generated tests. We only do that to loop over all exported widgets to check that they are functions. Do you think these should also add an assertion for `$$type`?

https://github.com/algolia/instantsearch.js/blob/6bbce4111b31a04fecd15e7bf4b21d649258cc73/src/lib/__tests__/main-test.js#L10-L14

These checks seem to be here to compensate the lack of integration tests.

---

Another solution would be to make `$$type` required and to not export the internal widget type to users.

Haroenv:
Yes, but that would still be only catch those written in TS which actually correctly implement the type, which is already most ideally. I thought we could quickly test something like this, but I understand your POV

PR #3961, opened 2019-07-17
chore(routing): correct types

Haroenv:
This is a follow-up on #3940 requested by @samouss

- correct the type for `getAllSearchParameters`
- move defaults  in response mock to not have intermediary variables
- consistent function style in mock

sorry for the 3-in-1 PR, but wasn't really worth splitting IMO

### DIFF
--- a/test/mock/createAPIResponse.ts
+++ b/test/mock/createAPIResponse.ts
@@ -1,30 +1,27 @@
 import { SearchForFacetValues, Response, MultiResponse } from 'algoliasearch';
 
-export function createSingleSearchResponse<THit = any>(
+export const createSingleSearchResponse = <THit = any>(
   subset: Partial<Response<THit>> = {}
-): Response<THit> {
+): Response<THit> => {
   const {
+    query = '',
     page = 0,
     hitsPerPage = 20,
-    nbHits,
-    nbPages,
-    processingTimeMS = 0,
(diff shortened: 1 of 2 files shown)


test/mock/createAPIResponse.ts
```diff
@@ -1,30 +1,27 @@
...
+    query = '',
     page = 0,
     hitsPerPage = 20,
-    nbHits,
-    nbPages,
-    processingTimeMS = 0,
     hits = [],
-    query = '',
+    nbHits = hits.length,
+    nbPages = Math.ceil(nbHits / hitsPerPage),
```
samouss: I would have drop the tests of this file too, unless you think they are critical.

Haroenv: not critical, but @francoischalifour preferred it to have a test

francoischalifour: There was more logic on our side before this change and many tests are depending on it. I guess it's OK to keep them.

samouss: Many tests depend on them but they will fail anyway if the implementation is incorrect. The original implementation was using hard-coded values to avoid those issues. At the moment the logic is fairly simple, we should stick with this level of complexity to avoid testing the logic of the helpers.

PR #3964, opened 2019-07-18
feat(index): provide scoped results to render hook

francoischalifour:
## Description

This introduces the concept of _scoped results_ during the `render` hook.

## Why

Since InstantSearch.js now handles multi-index search, we will need to revisit the behavior of some widgets:

- `autocomplete` may need to access the scope of its current index, as well as all the siblings and children
- `currentRefinements` may need to show the refinements coming from the current index and all its siblings/children
- `clearRefinement` may need to reset the refinements coming from the current index and all its siblings/children

### Behavior

An index now has access to all its children and siblings (+ siblings children) results. As illustrated in the image below, the lower in the tree, the fewer results the index has access to.

instantsearch-scoped-results

The more transparent a level is, the less view it has on the global results.

### Render options interface

This information is given in the `scopedResults` key of the render options. The new interface is:

```ts
interface ScopedResult {
  indexId: string;
  results: SearchResults;
}

interface RenderOptions {
  templatesConfig: object;
  results: SearchResults;
  scopedResults: ScopedResult[]; // ← this is new
  state: SearchParameters;
  helper: Helper;
  searchMetadata: {
    isSearchStalled: boolean;
  };
}
```

A scoped result contains an `indexId` and a `results` key. It always has a least one entry in `scopedResults`, being the current index and its own result. Once you start nesting indexes, more results appear in `scopedResults`.

- `indexId`: unique identifier of the results. This is not true for now because the  `indexId` is the `indexName` and you can use the same index several times. We might later need to introduce an `indexId` option in the `index` widget.
- `results`: the results of the given index

### DIFF
--- a/src/widgets/index/__tests__/index-test.ts
+++ b/src/widgets/index/__tests__/index-test.ts
@@ -1257,6 +1257,12 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
         expect(widget.render).toHaveBeenCalledWith({
           instantSearchInstance,
           results: expect.any(algoliasearchHelper.SearchResults),
+          scopedResults: [
+            {
+              indexId: 'index_name',
+              results: (widget.render as jest.Mock).mock.calls[0][0].results,
+            },
+          ],
           state: expect.any(algoliasearchHelper.SearchParameters),
           helper: instance.getHelper(),
           templatesConfig: instantSearchInstance.templatesConfig,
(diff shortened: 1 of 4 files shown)


eunjae-lee:
Btw, what did you use to draw that? That's 💯

francoischalifour:
@eunjae-lee Good old [Sketch](https://www.sketch.com/) haha

src/widgets/index/index.ts
```diff
@@ -53,10 +90,18 @@ const index = (props: IndexProps): Index => {
   return {
     $$type: 'ais.index',
 
+    getIndexId() {
```
Haroenv: `get xxx` is ES5, so we could use it if we want, unrelated to this PR

src/widgets/index/index.ts
```diff
@@ -37,6 +40,40 @@ export type Index = Widget & {
...
 };
 
+function isIndexWidget(widget: Widget): widget is Index {
+  return widget.$$type === 'ais.index';
+}
+
+function resolveScopedResultsFromWidgets(widgets: Widget[]): ScopedResult[] {
+  const indexWidgets = widgets.filter(isIndexWidget);
+
+  if (indexWidgets.length === 0) {
```
samouss: This branch is useless since we reduce below. The returned value will be the initial state.

src/widgets/index/index.ts
```diff
@@ -37,6 +40,40 @@ export type Index = Widget & {
...
+    return [];
+  }
+
+  return indexWidgets.reduce<ScopedResult[]>((scopedResults, current) => {
+    const currentDerivedHelper = current.getDerivedHelper()!;
+
+    scopedResults.push(
```
samouss: We can use `concat` to avoid the intermediate step.

src/widgets/index/__tests__/index-test.ts
```diff
@@ -896,6 +902,132 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
         expect(widget.render).toHaveBeenCalledTimes(0);
       });
     });
+
+    it('calls `render` with `scopedResults` coming from siblings and children', async () => {
+      /* eslint-disable @typescript-eslint/camelcase */
+      const level0 = index({ indexName: 'level0_index_name' });
+      const level1 = index({ indexName: 'level1_index_name' });
+      const level2 = index({ indexName: 'level2_index_name' });
+      const level2_1 = index({ indexName: 'level2_1_index_name' });
```
samouss: We don't cover by an assertions the use case with siblings. We could add an assertion for the `level2_1`.

src/widgets/index/index.ts
```diff
@@ -53,10 +90,18 @@ const index = (props: IndexProps): Index => {
...
+    getIndexId() {
+      return indexName;
+    },
+
     getHelper() {
       return helper;
     },
 
+    getDerivedHelper() {
+      return derivedHelper;
```
samouss: Do we really want to give an access to the `derivedHelper` instance? Its usage is mostly internal I don't think we have a use case to access its instance. We could only expose the results at the index level.

PR #3967, opened 2019-07-19
docs(examples): make sure e-commerce example can be ran on codeSandBox

Haroenv:
This wasn't possible before because the array generated from spreading a form element got transformed into .slice which it didn't have. If downlevelIteration is true & dom.iterable is true the code will work however!

see https://github.com/codesandbox/codesandbox-client/issues/2166#issuecomment-513210419

you can verify this fix by comparing:

https://codesandbox.io/s/github/algolia/instantsearch.js/tree/develop/examples/e-commerce

to the fixed version:

https://codesandbox.io/s/github/algolia/instantsearch.js/tree/docs/fix-codesandbox/examples/e-commerce

### DIFF
--- a/examples/e-commerce/tsconfig.json
+++ b/examples/e-commerce/tsconfig.json
@@ -1,0 +1,10 @@
+{
+  "compilerOptions": {
+    "lib": [
+      "es6",
+      "dom",
+      "dom.iterable"
+    ],
+    "downlevelIteration": true
+  }
+}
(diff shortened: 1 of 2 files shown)


examples/e-commerce/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "lib": [
+      "es6",
+      "dom",
+      "dom.iterable"
+    ],
+    "downlevelIteration": true,
+    "esModuleInterop": true
```
Haroenv: I don't know why we didn't need this in the root config, but I need it for the sandbox to work

PR #3975, opened 2019-07-22
feat(autocomplete): leverage scoped results

francoischalifour:
## Description

This is a revamp of the `connectAutocomplete` connector leveraging scoped results (#3964). 

Since this is a complete rewrite of how the connector works, I took this opportunity to migrate it to TypeScript.

### Options changes

The [`indices`](https://www.algolia.com/doc/api-reference/widgets/autocomplete/js/#connector-param-render-indices) option doesn't exist anymore and is inferred by the widgets tree. If the option is used in development mode, a warning is raised.

### Structure change of an autocomplete index

```diff
 interface AutocompleteIndex {
-  label: string;
   index: string;
   hits: Hits;
   results: SearchResults | undefined;
 }
```

Once we introduce the concept of index IDs, we might want to add it so that users can make a difference from indexes having the same name:

```diff
 interface AutocompleteIndex {
-  label: string;
+  indexId: string;
-  index: string;
+  indexName: string;
   hits: Hits;
   results: SearchResults | undefined;
 }
```

## Preview

[See story →](https://deploy-preview-3975--instantsearchjs.netlify.com/stories/?path=/story/autocomplete--default)

## Related

- [Current `connectAutocomplete` documentation](https://www.algolia.com/doc/api-reference/widgets/autocomplete/js/)
- [Scoped results feature](https://github.com/algolia/instantsearch.js/pull/3964)


src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,171 @@
...
+
+      render({ helper, scopedResults }) {
+        const indices = scopedResults.map(scopedResult => {
+          scopedResult.results.hits = escapeHTML
+            ? escapeHits(scopedResult.results.hits)
+            : scopedResult.results.hits;
+
+          return {
+            index: scopedResult.results.index,
+            hits: scopedResult.results.hits,
```
Haroenv: something which the previous version also doesn't implement yet, and thus could be done in a separate PR, is adding `queryID` & absolute position to the hits array so it can be used for insights, like it can for Hits & InfiniteHits; cc @tkrugg 

see also https://github.com/algolia/react-instantsearch/pull/2687

Haroenv:
Which way do you see an index which is only used for an autocomplete being used, something like this?

```js
search.addWidgets([
  index({ indexName: 'query_suggestions' })
    .addWidgets([
       autocomplete(), // only widget which reads from this index
    ]),
  hits(),
]);
```

francoischalifour:
@Haroenv Yep!

Haroenv:
TS is failing because of #3959 now requiring `getConfiguration` to be passed a SearchParameters object

samouss:
@Haroenv Yes it's solution, depends on the use-case but this should work too:

```js
search.addWidgets([
  index({ indexName: 'query_suggestions' })
  autocomplete(),
]);
```

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,171 @@
...
+interface AutocompleteConnectorParams {
+  /**
+   * Escapes HTML entities from hits string values.
+   *
+   * @default `true`
+   */
+  escapeHTML?: boolean;
+}
+
+type RefineFunction = (query: string) => Helper;
```
samouss: We don't have to repeat the type inside the name, the information is already carried by the type.

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,171 @@
...
+) => {
+  checkRendering(renderFn, withUsage());
+
+  return widgetParams => {
+    const { escapeHTML = true } = widgetParams || {};
+
+    warning(
+      !(widgetParams as any).indices,
+      `Since InstantSearch.js 4, \`connectAutocomplete\` infers the indices from the tree of widgets.
+The \`indices\` option is ignored.`
```
samouss: We could print the alternative usage WDYT?

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,171 @@
...
+    };
+
+    const connectorState: ConnectorState = {};
+
+    return {
+      $$type: 'ais.autocomplete',
+
+      getConfiguration(previousParameters) {
+        const parameters = {
+          query: '',
```
samouss: The function should read from the previous parameter then fallback.

src/connectors/autocomplete/connectAutocomplete.ts
samouss: The return value should be `void` we don't have to expose the Helper through the connector.

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,171 @@
...
+            refine: connectorState.refine,
+            instantSearchInstance: connectorState.instantSearchInstance,
+          },
+          true
+        );
+      },
+
+      render({ helper, scopedResults }) {
+        const indices = scopedResults.map(scopedResult => {
+          scopedResult.results.hits = escapeHTML
```
samouss: We could add a comment that explain why we mutate the results.

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,171 @@
...
+
+const withUsage = createDocumentationMessageGenerator({
+  name: 'autocomplete',
+  connector: true,
+});
+
+interface AutocompleteIndex {
+  index: string;
+  hits: Hits;
+  results: SearchResults | undefined;
```
samouss: For which cases the value is `undefined`?

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,171 @@
...
+  Unmounter,
+} from '../../types';
+
+const withUsage = createDocumentationMessageGenerator({
+  name: 'autocomplete',
+  connector: true,
+});
+
+interface AutocompleteIndex {
+  index: string;
```
samouss: We can already name it `indexName`.

stories/autocomplete.stories.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    search.addWidgets([
+      instantsearch.widgets
+        .index({ indexName: 'instant_search_price_asc' })
+        .addWidgets([
+          instantsearch.widgets.configure({
+            hitsPerPage: 3,
+          }),
+
+          instantsearch.widgets.searchBox({
```
samouss: We don't really showcase the full power of the connector. IMO we should use the `refine` behaviour of the connector, rather than relying on the `searchBox` (for the design we can give it the same class).

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -0,0 +1,421 @@
...
+    helper.search = jest.fn();
+
+    widget.init!(createInitOptions({ helper }));
+
+    expect(renderFn).toHaveBeenCalledTimes(1);
+    expect(renderFn).toHaveBeenLastCalledWith(
+      expect.objectContaining({
+        currentRefinement: '',
+        indices: [],
+        refine: expect.any(Function),
```
samouss: We can add the `widgetParams` and `instantSearchInstance` to the test.

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -0,0 +1,421 @@
...
+      }),
+      false
+    );
+  });
+
+  it('consumes the correct indices', () => {
+    const searchClient = createSearchClient();
+    const renderFn = jest.fn();
+    const makeWidget = connectAutocomplete(renderFn);
+    const widget = makeWidget({ escapeHTML: false });
```
samouss:
```suggestion
    const widget = makeWidget();
```

It doesn't make sense to add the option for test, we don't use it.

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -0,0 +1,421 @@
...
+
+    const helper = algoliasearchHelper(searchClient, '', {});
+    helper.search = jest.fn();
```
samouss: We should create a mock helper for this we do it thousand times across the codebase.

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,171 @@
...
+            widgetParams,
+            currentRefinement: helper.state.query || '',
+            indices,
+            refine: connectorState.refine!,
+            instantSearchInstance: connectorState.instantSearchInstance!,
+          },
+          false
+        );
+      },
+
```
samouss: The widget does not implement the routing lifecycle, we can add them in a separate PR though.

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,171 @@
...
+>;
+
+export type AutocompleteWidgetFactory<
+  TAutocompleteWidgetParams
+> = WidgetFactory<AutocompleteConnectorParams & TAutocompleteWidgetParams>;
+
+export type AutocompleteConnector = <TAutocompleteWidgetParams>(
+  render: AutocompleteRenderer<TAutocompleteWidgetParams>,
+  unmount?: Unmounter
+) => AutocompleteWidgetFactory<TAutocompleteWidgetParams>;
```
samouss: Those types are a nightmare to follow, would be nice to simplify them at some point. No idea how though.

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
francoischalifour: This is needed otherwise we cannot do this assertion:

```ts
expect(secondRenderOptions.indices[0].hits).toEqual(firstIndexHits);
```

Because we add `__escaped`, to the array with `escapeHTML` to `true`. Jest considers them as different arrays.

src/connectors/autocomplete/connectAutocomplete.ts
francoischalifour: WDYM? A link to the doc?

src/connectors/autocomplete/connectAutocomplete.ts
francoischalifour: None, it's mistyped.

src/connectors/autocomplete/connectAutocomplete.ts
francoischalifour: We mutate the results also in `connectHits` and `connectInfiniteHits`. What kind of comment would you expect?

stories/autocomplete.stories.ts
francoischalifour: What could we showcase more with the search box connector that we don't with the widget?

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
francoischalifour: Let's do it in another PR.

src/connectors/autocomplete/connectAutocomplete.ts
samouss: Convert the usage of `indices` into `index` This what we did for [the `geoSearch` warnings](https://github.com/algolia/instantsearch.js/blob/509513c0feafaad522f6f18d87a441559f4aa050/src/connectors/geo-search/connectGeoSearch.js#L117-L132).

src/connectors/autocomplete/connectAutocomplete.ts
samouss: Yes we do, but it's not obvious why we do it. Something that mention highlighting.

stories/autocomplete.stories.ts
samouss: The story do not use `refine`, `currentRefinement` to implement the autocomplete. We leverage the `searchBox` widget. Most of the time people will build their autocomplete with the connector only, specially if it's an integration with a pre-built component. IMO it worth showcasing the full API.

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
samouss: My bad.

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,188 @@
...
+            instantSearchInstance: connectorState.instantSearchInstance!,
+          },
+          false
+        );
+      },
+
+      dispose({ state }) {
+        unmountFn();
+
+        const stateWithoutQuery = state.setQueryParameter('query', undefined);
```
Haroenv: This is fine, but we should probably avoid calling state functions multiple times, and can join them all in a single setQueryParameters, WDYT?

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,188 @@
...
+          query: previousParameters.query || '',
+        };
+
+        if (!escapeHTML) {
+          return previousParameters.setQueryParameters(parameters);
+        }
+
+        return previousParameters.setQueryParameters({
+          ...parameters,
+          ...TAG_PLACEHOLDER,
```
francoischalifour: Do we want to support previous highlighting tags?

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -0,0 +1,188 @@
...
+   *
+   * @default `true`
+   */
+  escapeHTML?: boolean;
+}
+
+export interface AutocompleteRendererOptions<TAutocompleteWidgetParams>
+  extends RendererOptions<TAutocompleteWidgetParams> {
+  currentRefinement: string;
+  indices: AutocompleteIndex[];
```
francoischalifour: Do we keep `indices` or rename it to `items` like the other widgets?

src/connectors/autocomplete/connectAutocomplete.ts
francoischalifour: We do that in other connectors. They don't mutate the global state, they return a new object.
Why would you recommend not calling these functions multiple times?

src/connectors/autocomplete/connectAutocomplete.ts
samouss: I don't think so, otherwise it might breaks our highlighting feature.

src/connectors/autocomplete/connectAutocomplete.ts
samouss: `items` refers most of the time to a list of facet values, `hits` refers to a list of hits, `indices` could make sense. Both are valid, I don't think it matters much.

francoischalifour: Let's keep it that way, then!

src/connectors/autocomplete/connectAutocomplete.ts
samouss: Here it's to avoid the duplication of the logic for removing the query. At some point if we have perf issue we could revisit those parts, but for now I guess it's fine.

PR #3976, opened 2019-07-22
feat(federated): keep a consistent state in the RefinementList life cycle

Haroenv:
IFW-875

**Summary**

RefinementList no longer sets any type of refinements to an empty object, always keeping this attribute as an empty array in refinements.

- getConfiguration sets empty array to XRefinement
- getConfiguration reads from existing state to set value from other identical widgets in index
- getConfiguration returns a SearchParameters
- getFacetValues should have a fallback
- Refine (via toggleXRefinement in https://github.com/algolia/algoliasearch-helper-js/pull/738) doesn’t remove refinement if empty (helper)
- Dispose removes XRefinement, using only helper.removeXFacet

**Result**

~~tests will fail until https://github.com/algolia/algoliasearch-helper-js/pull/738 is used here~~

Dispose & get Configuration work as expected

### DIFF
--- a/src/connectors/refinement-list/__tests__/connectRefinementList-test.js
+++ b/src/connectors/refinement-list/__tests__/connectRefinementList-test.js
@@ -455,7 +455,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
     ]);
   });
 
-  it('Provide a function to clear the refinements at each step', () => {
+  it('Provide a function to clear the refinements at each step (or)', () => {
     const { makeWidget, rendering } = createWidgetFactory();
     const widget = makeWidget({
       attribute: 'category',
@@ -479,9 +479,9 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
     const firstRenderingOptions = rendering.mock.calls[0][0];
     const { refine } = firstRenderingOptions;
     refine('value');
-    expect(helper.hasRefinements('category')).toBe(false);
+    expect(helper.state.disjunctiveFacetsRefinements.category).toHaveLength(0);
     refine('value');
-    expect(helper.hasRefinements('category')).toBe(true);
+    expect(helper.state.disjunctiveFacetsRefinements.category).toHaveLength(1);
 
     widget.render({
       results: new SearchResults(helper.state, [{}, {}]),
@@ -493,9 +493,53 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
     const secondRenderingOptions = rendering.mock.calls[1][0];
     const { refine: renderToggleRefinement } = secondRenderingOptions;
     renderToggleRefinement('value');
-    expect(helper.hasRefinements('category')).toBe(false);
+    expect(helper.state.disjunctiveFacetsRefinements.category).toHaveLength(0);
     renderToggleRefinement('value');
-    expect(helper.hasRefinements('category')).toBe(true);
+    expect(helper.state.disjunctiveFacetsRefinements.category).toHaveLength(1);
+  });
+
+  it('Provide a function to clear the refinements at each step (and)', () => {
+    const { makeWidget, rendering } = createWidgetFactory();
+    const widget = makeWidget({
+      attribute: 'category',
+      operator: 'and',
+    });
+
+    const helper = jsHelper(
+      {},
+      '',
+      widget.getConfiguration(new SearchParameters({}))
+    );
+    helper.search = jest.fn();
+
+    helper.toggleRefinement('category', 'value');
+
+    widget.init({
+      helper,
+      state: helper.state,
+      createURL: () => '#',
+    });
+
+    const firstRenderingOptions = rendering.mock.calls[0][0];
+    const { refine } = firstRenderingOptions;
+    refine('value');
+    expect(helper.state.facetsRefinements.category).toHaveLength(0);
+    refine('value');
+    expect(helper.state.facetsRefinements.category).toHaveLength(1);
+
+    widget.render({
+      results: new SearchResults(helper.state, [{}, {}]),
+      state: helper.state,
+      helper,
+      createURL: () => '#',
+    });
+
+    const secondRenderingOptions = rendering.mock.calls[1][0];
+    const { refine: renderToggleRefinement } = secondRenderingOptions;
+    renderToggleRefinement('value');
+    expect(helper.state.facetsRefinements.category).toHaveLength(0);
+    renderToggleRefinement('value');
+    expect(helper.state.facetsRefinements.category).toHaveLength(1);
   });
 
   it('If there are too few items then canToggleShowMore is false', () => {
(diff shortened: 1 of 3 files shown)


src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -299,7 +313,7 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
           showMore ? showMoreLimit : limit
         );
 
-        return widgetConfiguration;
+        return new algoliasearchHelper.SearchParameters(widgetConfiguration);
```
samouss: You can use the argument provided by the `getConfiguration` function.

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -374,11 +388,9 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
         unmountFn();
 
         if (operator === 'and') {
-          return state.removeFacetRefinement(attribute).removeFacet(attribute);
+          return state.removeFacet(attribute);
         } else {
-          return state
-            .removeDisjunctiveFacetRefinement(attribute)
-            .removeDisjunctiveFacet(attribute);
+          return state.removeDisjunctiveFacet(attribute);
```
samouss: We can cover this by a test that ensure that the value is not there anymore (`dispose` is never tested).

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -374,11 +388,9 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
         unmountFn();
 
         if (operator === 'and') {
-          return state.removeFacetRefinement(attribute).removeFacet(attribute);
+          return state.removeFacet(attribute);
         } else {
```
samouss: The `else` branch is useless since we early return inside the `if`.

src/connectors/refinement-list/connectRefinementList.js
Haroenv: d51f2c670bcef53e4f28ad1f8c0e7a58d7658ba4

src/connectors/refinement-list/connectRefinementList.js
Haroenv: 92904a8de

src/connectors/refinement-list/connectRefinementList.js
Haroenv: 7414f92,  noticed that we weren't removing the maxValuesPerFacet too

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -375,13 +388,14 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
...
-          return state.removeFacetRefinement(attribute).removeFacet(attribute);
+          newState = state.removeFacet(attribute);
         } else {
-          return state
-            .removeDisjunctiveFacetRefinement(attribute)
-            .removeDisjunctiveFacet(attribute);
+          newState = state.removeDisjunctiveFacet(attribute);
         }
+
+        return newState.setQueryParameter('maxValuesPerFacet', undefined);
```
Haroenv: we need to make sure we only  unset maxValuesPerFacet if it was set by this widget

Haroenv: decided that it's ok, because the other mounted widgets will set their value back

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -375,13 +388,14 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
       dispose({ state }) {
         unmountFn();
 
+        let newState;
```
samouss: We can avoid the assignment with the call to `setQueryParameter` first, then we early return inside the branch.

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -1422,4 +1568,174 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
       });
     });
   });
+
+  describe('dispose', () => {
+    it('removes refinements completely on dispose (and)', () => {
```
samouss: https://github.com/algolia/instantsearch.js/pull/3986#discussion_r307675529

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -289,10 +289,23 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
...
 
-      getConfiguration: (configuration = {}) => {
-        const widgetConfiguration = {
-          [operator === 'and' ? 'facets' : 'disjunctiveFacets']: [attribute],
-        };
+      getConfiguration(configuration) {
+        let widgetConfiguration;
+        if (operator === 'and') {
+          widgetConfiguration = {
+            facets: [attribute],
```
samouss: https://github.com/algolia/instantsearch.js/pull/3986#discussion_r307677462

francoischalifour: How do we want to deal with facets duplication here? Call `uniq` at this step?

samouss: Actually I thought that the `SearchParameters` managed that for us but I was wrong it only does it for the facets refinements. Whether the class does it for us or not the behaviour should be consistent. Keep it like this then the merge function will do it.

Once we merge the two lifecycles (`getConfiguration` & `getWidgetSearchParameters`) we should be able get rid of the previous value since we'll always clear the previous value to let the `uiState` completely control the refinements.

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -388,14 +388,15 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
...
 
-        let newState;
+        const newState = state.setQueryParameter(
+          'maxValuesPerFacet',
+          undefined
+        );
         if (operator === 'and') {
-          newState = state.removeFacet(attribute);
+          return newState.removeFacet(attribute);
         } else {
```
samouss: The `else` branch is useless.

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -388,14 +388,15 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
       dispose({ state }) {
         unmountFn();
 
-        let newState;
+        const newState = state.setQueryParameter(
```
samouss: We can name with a clearer name like `withoutMaxValuesPerFacet`.

src/connectors/refinement-list/connectRefinementList.js
samouss: Actually I was correct. `SearchParameters` manage it for us, at least for [`facets`](https://github.com/algolia/algoliasearch-helper-js/blob/6ac260dfcbdc0a94059eb93b8ad8b4552f63abf5/src/SearchParameters/index.js#L689-L697), [`disjunctiveFacets`](https://github.com/algolia/algoliasearch-helper-js/blob/6ac260dfcbdc0a94059eb93b8ad8b4552f63abf5/src/SearchParameters/index.js#L705-L713) but not for [`hierarchicalFacets`](https://github.com/algolia/algoliasearch-helper-js/blob/6ac260dfcbdc0a94059eb93b8ad8b4552f63abf5/src/SearchParameters/index.js#L722-L731). The deduplication for facet refinements are also managed by the `SearchParameters` for each type of facets. 

cc @Haroenv @francoischalifour

PR #3977, opened 2019-07-23
fix(hits): update lifecycle state

francoischalifour:
## Description

This fixes the lifecycle of `connectHits`.

## Changes

### DIFF
--- a/src/connectors/hits/connectHits.js
+++ b/src/connectors/hits/connectHits.js
@@ -65,9 +65,7 @@ export default function connectHits(renderFn, unmountFn = noop) {
           return state;
         }
 
-        return state.setQueryParameters({
-          ...TAG_PLACEHOLDER,
-        });
+        return state.setQueryParameters(TAG_PLACEHOLDER);
       },
 
       init({ instantSearchInstance }) {
--- a/src/connectors/hits/connectHits.js
+++ b/src/connectors/hits/connectHits.js
@@ -60,8 +60,14 @@ export default function connectHits(renderFn, unmountFn = noop) {
     return {
       $$type: 'ais.hits',
 
-      getConfiguration() {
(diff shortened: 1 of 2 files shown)


src/connectors/hits/connectHits.js
```diff
@@ -60,8 +60,14 @@ export default function connectHits(renderFn, unmountFn = noop) {
...
 
-      getConfiguration() {
-        return escapeHTML ? TAG_PLACEHOLDER : undefined;
+      getConfiguration(state) {
+        if (!escapeHTML) {
+          return state;
+        }
+
+        return state.setQueryParameters({
+          ...TAG_PLACEHOLDER,
```
Haroenv: feel free to ignore this; but `setQueryParameters(TAG_PLACEHOLDER)` is an option too

src/connectors/hits/connectHits.js
```diff
@@ -60,8 +60,21 @@ export default function connectHits(renderFn, unmountFn = noop) {
...
-        return escapeHTML ? TAG_PLACEHOLDER : undefined;
+      getConfiguration(state) {
+        if (!escapeHTML) {
+          return state;
+        }
+
+        return state.setQueryParameters({
+          highlightPreTag:
+            state.highlightPreTag === undefined
+              ? TAG_PLACEHOLDER.highlightPreTag
```
Haroenv: I feel like we should be always overriding it, but I'm not fully sure

francoischalifour: I removed it.

src/connectors/hits/connectHits.js
samouss: Agree with @Haroenv the spread is useless in that case.

PR #3978, opened 2019-07-23
fix(hitsPerPage): update lifecycle state

francoischalifour:
## Description

This fixes the lifecycle of `connectHitsPerPage`.

## Changes

### DIFF
--- a/src/widgets/hits-per-page/__tests__/hits-per-page-test.js
+++ b/src/widgets/hits-per-page/__tests__/hits-per-page-test.js
@@ -1,4 +1,5 @@
 import { render } from 'preact-compat';
+import { SearchParameters } from 'algoliasearch-helper';
 import hitsPerPage from '../hits-per-page';
 
 jest.mock('preact-compat', () => {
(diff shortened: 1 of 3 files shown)


src/widgets/hits-per-page/__tests__/hits-per-page-test.js
```diff
@@ -74,9 +77,13 @@ describe('hitsPerPage()', () => {
       ],
     });
 
-    expect(widgetWithDefaults.getConfiguration()).toEqual({
-      hitsPerPage: 20,
-    });
+    expect(
+      widgetWithDefaults.getConfiguration(new SearchParameters({}))
+    ).toEqual(
+      expect.objectContaining({
```
Haroenv: I wonder if we should create `toEqualSearchParameters` or use `toEqual(new SearchParameters({`, I'm fine with `objectContaining` too since we aren't yet consistent about this

francoischalifour: I replaced this with `new SearchParameters({ ... })` to check the returned type as well.

PR #3979, opened 2019-07-23
fix(pagination): update lifecycle state

francoischalifour:
## Description

This fixes the lifecycle of `connectPagination`.

## Changes

### DIFF
--- a/src/connectors/pagination/connectPagination.js
+++ b/src/connectors/pagination/connectPagination.js
@@ -102,7 +102,7 @@ export default function connectPagination(renderFn, unmountFn = noop) {
 
       getConfiguration(state) {
         return state.setQueryParameters({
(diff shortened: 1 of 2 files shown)


src/connectors/pagination/connectPagination.js
```diff
@@ -100,10 +100,10 @@ export default function connectPagination(renderFn, unmountFn = noop) {
     return {
       $$type: 'ais.pagination',
 
-      getConfiguration() {
-        return {
+      getConfiguration(state) {
+        return state.setQueryParameters({
           page: 0,
```
francoischalifour: Should this take into account the previous `page` state now?

Haroenv: Yes, it should

PR #3984, opened 2019-07-24
chore(core): migrate core file to TS

Haroenv:
**Summary**

migrated the InstantSearch.js file to TypeScript to have better inference in other places

**Result**

- main.ts & InstantSearch.ts are now in typescript
- mocked version of InstantSearch now has all other properties that weren't mocked before
  - including deferred functions which were previously implemented as just a jest.fn, meaning some change in tests

this depends on:

- [x] https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37105
- [x] https://github.com/algolia/algoliasearch-helper-js/pull/739


src/lib/InstantSearch.ts
```diff
@@ -285,10 +314,11 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
 
     this.mainHelper = mainHelper;
 
-    this.mainIndex.init({
+    // @TODO: do we want this to be correct arguments?
+    this.mainIndex.init(({
       instantSearchInstance: this,
       parent: null,
-    });
+    } as any) as InitOptions);
```
Haroenv: we could change the signature of Index to only need these options

src/lib/InstantSearch.ts
```diff
@@ -307,13 +337,14 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
-  dispose() {
+  public dispose(): void {
     this.scheduleSearch.cancel();
     this.scheduleRender.cancel();
     clearTimeout(this._searchStalledTimer);
 
     this.removeWidgets(this.mainIndex.getWidgets());
-    this.mainIndex.dispose();
+    // @TODO: do we want this to be correct arguments?
+    this.mainIndex.dispose(({} as any) as DisposeOptions);
```
Haroenv: here also, we call this with less than what index asks for

src/lib/InstantSearch.ts
```diff
@@ -323,30 +354,31 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
       this._searchStalledTimer = null;
       this._isSearchStalled = false;
     }
 
-    this.mainIndex.render({
+    // @TODO: do we want this to be correct arguments?
+    this.mainIndex.render(({
       instantSearchInstance: this,
-    });
+    } as any) as RenderOptions);
```
Haroenv: here also, we call this with less than what index asks for

test/mock/createInstantSearch.ts
```diff
@@ -2,26 +2,65 @@ import algoliasearchHelper from 'algoliasearch-helper';
...
     _searchParameters: {},
+    _createURL: jest.fn(() => '#'),
     _createAbsoluteURL: jest.fn(() => '#'),
+    createURL: jest.fn(() => '#'),
+    routing: null, // @TODO: good definition
+    addWidget: jest.fn(),
+    addWidgets: jest.fn(),
+    removeWidget: jest.fn(),
+    removeWidgets: jest.fn(),
+    // @TODO: can be inherited from EventEmitter?
```
Haroenv: I tried spreading out `new EventEmitter()`, but that didn't work for TS

src/lib/InstantSearch.ts
```diff
@@ -64,24 +95,24 @@ class InstantSearch extends EventEmitter {
...
       throw new Error(
         `The \`searchClient\` must implement a \`search\` method.
 
 See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend-search/in-depth/backend-instantsearch/js/`
       );
     }
 
-    if (typeof searchClient.addAlgoliaAgent === 'function') {
-      searchClient.addAlgoliaAgent(`instantsearch.js (${version})`);
+    if (typeof (searchClient as Client).addAlgoliaAgent === 'function') {
```
Haroenv: requires https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37105

src/lib/InstantSearch.ts
```diff
@@ -256,21 +281,25 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     };
```
Haroenv: line above this requires https://github.com/algolia/algoliasearch-helper-js/pull/739

package.json
```diff
@@ -68,8 +68,7 @@
     "@storybook/addon-actions": "5.1.9",
     "@storybook/html": "5.1.9",
     "@storybook/theming": "5.1.9",
-    "@types/algoliasearch": "3.30.14",
-    "@types/algoliasearch-helper": "2.26.1",
```
Haroenv: was no longer required since we used the canary version of the helper, but it seems like it was merged back in wrongly

src/lib/InstantSearch.ts
Haroenv: could be done with a type guard as well, but didn't seem necessary here

src/lib/InstantSearch.ts
```diff
@@ -64,24 +95,24 @@ class InstantSearch extends EventEmitter {
...
+    if (typeof (options as any).urlSync !== 'undefined') {
       throw new Error(
         withUsage(
           'The `urlSync` option was removed in InstantSearch.js 3. You may want to use the `routing` option.'
         )
       );
     }
 
-    if (typeof searchClient.search !== 'function') {
+    if (typeof (searchClient as any).search !== 'function') {
```
francoischalifour: `searchClient` should always be typed with a `search` method. Why do you need this cast?

src/lib/InstantSearch.ts
```diff
@@ -133,23 +164,21 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
    * widget after InstantSearch started is considered **EXPERIMENTAL** and therefore
```
francoischalifour: We use and test those methods in the new major version. WDYT about removing the "experimental" mentions?

src/lib/InstantSearch.ts
```diff
@@ -256,21 +281,25 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     };
 
     if (this._searchFunction) {
+      // this client isn't used to actually search, but required for the helper
+      // to not throw errors
+      const fakeClient = ({
```
francoischalifour: It would be less of a lie to type it as `Partial<Client>` and to cast it as `Client` line 294.

```ts
const fakeClient = {
  search: () => new Promise(noop),
} as Partial<Client>;
```

src/lib/InstantSearch.ts
francoischalifour: Since we correctly type our widget lifecycle now, it would make sense to either create a 
`IndexInitOptions` type or to pass all options.

I would go for passing all init options because we might need them at some point in the `index` methods and the type checking might break if we keep casting.

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -349,7 +350,7 @@ describe('start', () => {
     search.start();
 
     expect(algoliasearchHelper).toHaveBeenCalledTimes(2);
-    expect(algoliasearchHelper).toHaveBeenCalledWith(searchClient);
+    expect(algoliasearchHelper).toHaveBeenCalledWith(searchClient, indexName);
```
francoischalifour: Since we do not need this index (it has worked without), do you actually need to check that it's passed?

src/types/instantsearch.ts
```diff
@@ -4,11 +4,22 @@ import {
...
-import { Index } from '../widgets/index/index';
 import { InsightsClient as AlgoliaInsightsClient } from './insights';
 import { Widget, UiState } from './widget';
-
-export type InstantSearchOptions = any;
+export { default as InstantSearch } from '../lib/InstantSearch';
+
+export type SearchClient = Pick<Client, 'search' | 'searchForFacetValues'>;
+
+export type InstantSearchOptions = {
```
francoischalifour: We can copy the descriptions and default values (with `@default`) from the [`instantsearch` doc](https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/#options). That's helpful when working on the codebase and to onboard people.

src/types/instantsearch.ts
```diff
@@ -4,11 +4,22 @@ import {
...
+  indexName: string;
+  searchClient: SearchClient | Client;
+  numberLocale?: string;
+  searchFunction?: (helper: AlgoliaSearchHelper) => void;
+  searchParameters?: PlainSearchParameters;
+  routing?: any;
```
francoischalifour: Any reasons for not typing `routing`?

test/mock/createInstantSearch.ts
francoischalifour: We want to assign all values to `jest.fn()` anyway so I guess it's fine like this?

src/lib/InstantSearch.ts
```diff
@@ -42,7 +55,25 @@ function defaultCreateURL() {
...
+  public mainIndex: Index;
+  public started: boolean;
+  public templatesConfig: object;
+  public _stalledSearchDelay: number;
+  public _searchStalledTimer: any;
+  public _isSearchStalled: boolean;
+  public _searchParameters: PlainSearchParameters;
+  public _searchFunction?: InstantSearchOptions['searchFunction'];
+  public _createURL?: (params: SearchParameters) => string;
+  public _createAbsoluteURL?: (params: SearchParameters) => string;
```
francoischalifour: All these class variables don't need to be public. Perhaps we can make some private and add getters?

src/lib/InstantSearch.ts
Haroenv: TS would complain about a needless comparison if search is always defined

src/lib/__tests__/InstantSearch-test.js
Haroenv: The assertion fails if not all arguments are listed

src/types/instantsearch.ts
Haroenv: I did not find a good existing type for it

src/lib/InstantSearch.ts
Haroenv: you're right, but `_createAbsoluteURL` is actually being used by the index component, do you suggest making it properly public without the prefix?

Haroenv: Note also that in this PR I didn't change any semantics of the existing JS, which is why I didn't make changes to the visibility

src/lib/__tests__/InstantSearch-test.js
francoischalifour: Would using `expect.anything()` work in both cases?

src/lib/InstantSearch.ts
Haroenv: agree, but can be done in another major version :)

src/lib/InstantSearch.ts
```diff
@@ -42,7 +55,25 @@ function defaultCreateURL() {
  * @fires Instantsearch#render This event is triggered each time a render is done
  */
 class InstantSearch extends EventEmitter {
-  constructor(options) {
+  public client: InstantSearchOptions['searchClient'];
+  public insightsClient: InsightsClient | null;
```
samouss:
```suggestion
  public insightsClient: InstantSearchOptions['insightsClient'] | null;
```

For the properties that come from the options we should be consistent (whatever solution we take).

src/lib/InstantSearch.ts
```diff
@@ -42,7 +55,25 @@ function defaultCreateURL() {
...
+  public indexName: string;
+  public helper: AlgoliaSearchHelper | null;
+  public mainHelper: AlgoliaSearchHelper | null;
+  public mainIndex: Index;
+  public started: boolean;
+  public templatesConfig: object;
+  public _stalledSearchDelay: number;
+  public _searchStalledTimer: any;
```
samouss:
```suggestion
  public _searchStalledTimer?: number;
```

src/lib/InstantSearch.ts
```diff
@@ -256,21 +281,25 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
       mainHelper.search = () => {
         const mainIndexHelper = this.mainIndex.getHelper();
         const searchFunctionHelper = algoliasearchHelper(
-          {
-            search: () => new Promise(noop),
-          },
-          mainIndexHelper.state.index,
-          mainIndexHelper.state
+          fakeClient,
+          mainIndexHelper!.state.index,
```
samouss: We can remove the `!` if we put it after the `getHelper` call. It would be less redundant.

src/lib/InstantSearch.ts
samouss: `IndexInitOptions` seems a good compromise, the type should be a subset of `InitOptions` anyway.

src/lib/InstantSearch.ts
samouss: We can provide the `Helper` + `state` but we don't use them.

src/lib/InstantSearch.ts
```diff
@@ -323,30 +354,31 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
-  scheduleSearch = defer(() => {
-    this.mainHelper.search();
+  public scheduleSearch = defer(() => {
+    this.mainHelper!.search();
   });
 
-  scheduleRender = defer(() => {
-    if (!this.mainHelper.hasPendingRequests()) {
+  public scheduleRender = defer(() => {
+    if (this.mainHelper!.hasPendingRequests() === false) {
```
samouss:
```suggestion
    if (this.mainHelper!.hasPendingRequests()) {
```

Is it really useful since `hasPendingRequests()`  return a boolean already?

src/lib/InstantSearch.ts
samouss: The suggestion for `init` also apply for `render`.

src/types/instantsearch.ts
samouss: This option can be inlined inside the module now that it's a TypeScript one.

src/lib/InstantSearch.ts
Haroenv: it's not a number in Node, and TS gets confused about this (won't accept number, since we have @types/node)

src/lib/InstantSearch.ts
Haroenv: this is not the same, input is undefined or client, instance is client / null

src/lib/InstantSearch.ts
Haroenv: If I move this, I still need the exclamation mark, where do you mean to move it exactly?

samouss: https://github.com/algolia/instantsearch.js/pull/3984/files/cecc940c97b45bb614e42dedc870eac620b04a55#diff-1fc36c4ea042d2ecb5a498416e5e0357R292

src/types/widget.ts
```diff
@@ -102,10 +102,24 @@ export type UiState = {
...
+  /**
+   * Called once before the first search
+   */
   init?(options: InitOptions): void;
+  /**
+   * Called after each search response has been received
+   */
   render?(options: RenderOptions): void;
+  /**
+   * Called when  this widget is unmounted. Used to remove refinements set by
```
Haroenv:
```suggestion
   * Called when this widget is unmounted. Used to remove refinements set by
```

PR #3986, opened 2019-07-25
chore(menu): migrate to new life cycles

Haroenv:
also fixes dispose (yay)!

IFW-882

**Summary**

**Result**


src/connectors/menu/__tests__/connectMenu-test.js
```diff
@@ -681,4 +707,93 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/menu/js/#co
       });
     });
   });
+
+  describe('dispose', () => {
+    it('removes hierarchical refinements', () => {
+      const widget = makeWidget({
```
samouss: Until we have a proper setup for integration tests we could focus only on `dispose`. Here we are lost inside the setup boilerplate. We can craft a `SearchParameters` by hand, it's not likely to change anytime soon.

src/connectors/menu/connectMenu.js
```diff
@@ -164,7 +164,7 @@ export default function connectMenu(renderFn, unmountFn = noop) {
           showMore ? showMoreLimit : limit
         );
 
-        return widgetConfiguration;
+        return configuration.setQueryParameters(widgetConfiguration);
```
samouss: Since now we use a `SearchParameters` we can use the function to add the facets and avoid the call to the merge function inside `enhanceConfiguration`. It means that `getConfiguration` should not override the "controlled" parameters e.g `hierarchicalFacets`.

src/connectors/menu/connectMenu.js
```diff
@@ -232,22 +232,9 @@ export default function connectMenu(renderFn, unmountFn = noop) {
       dispose({ state }) {
         unmountFn();
 
-        let nextState = state;
-
-        if (state.isHierarchicalFacetRefined(attribute)) {
```
samouss: The condition prevent the function `removeHierarchicalFacetRefinement` to throw an error if the no refinement are present (https://github.com/algolia/instantsearch.js/pull/2610). It worth adding a test where we remove the widget without any refinement to ensure that we don't break this behaviour.

src/connectors/menu/connectMenu.js
samouss: My bad I thought `SearchParameters` would do it for us https://github.com/algolia/instantsearch.js/pull/3976#discussion_r308155858.

src/connectors/menu/__tests__/connectMenu-test.js
```diff
@@ -619,7 +645,11 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/menu/js/#co
           helper,
         });
 
-        expect(uiStateAfter).toMatchSnapshot();
+        expect(uiStateAfter).toEqual({
+          menu: {
+            category: 'pants',
+          },
+        });
```
samouss: 👍

samouss:
You have the fix the merge issue, the `menu` uses `getConfiguration({})` somewhere.

PR #3987, opened 2019-07-26
fix(ratingMenu): update lifecycle state

francoischalifour:
## Description

This fixes the lifecycle of `connectRatingMenu`.

## Changes

- `getConfiguration` returns a search parameter object
- `getConfiguration` support previously set disjunctive facets refinements
- `getFacetValues()` fallbacks to an empty array
- `dispose` doesn't call `removeDisjunctiveFacetRefinement()`

### DIFF
--- a/src/connectors/rating-menu/__tests__/connectRatingMenu-test.js
+++ b/src/connectors/rating-menu/__tests__/connectRatingMenu-test.js
@@ -58,10 +58,15 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/rating-menu
       attribute,
     });
 
-    const config = widget.getConfiguration({});
-    expect(config).toEqual({
-      disjunctiveFacets: [attribute],
-    });
+    const config = widget.getConfiguration(new SearchParameters({}));
+    expect(config).toEqual(
+      new SearchParameters({
+        disjunctiveFacets: [attribute],
+        disjunctiveFacetsRefinements: {
+          grade: [],
+        },
+      })
+    );
 
     const helper = jsHelper({}, '', config);
     helper.search = jest.fn();
@@ -154,7 +159,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/rating-menu
       attribute,
     });
 
-    const config = widget.getConfiguration({});
+    const config = widget.getConfiguration(new SearchParameters({}));
 
     const helper = jsHelper({}, '', config);
     helper.search = jest.fn();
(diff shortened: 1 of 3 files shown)


src/connectors/rating-menu/connectRatingMenu.js
```diff
@@ -106,8 +106,13 @@ export default function connectRatingMenu(renderFn, unmountFn = noop) {
     return {
       $$type: 'ais.ratingMenu',
 
-      getConfiguration() {
-        return { disjunctiveFacets: [attribute] };
+      getConfiguration(state) {
+        return state.setQueryParameters({
+          disjunctiveFacets: [attribute],
```
francoischalifour: See if we want to support previous facets at this level instead of in `enhanceConfiguration`.

src/connectors/rating-menu/connectRatingMenu.js
```diff
@@ -180,9 +185,7 @@ export default function connectRatingMenu(renderFn, unmountFn = noop) {
       dispose({ state }) {
         unmountFn();
 
-        const nextState = state
-          .removeDisjunctiveFacetRefinement(attribute)
-          .removeDisjunctiveFacet(attribute);
+        const nextState = state.removeDisjunctiveFacet(attribute);
```
Haroenv: looks like there's no tests yet for `dispose`

This could also be directly returned

francoischalifour: WDYM? I created a test suite.

Haroenv: I missed it in the diff, sorry

src/connectors/rating-menu/__tests__/connectRatingMenu-test.js
```diff
@@ -248,6 +253,95 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/rating-menu
...
+      const indexName = 'indexName';
+      const helper = jsHelper({}, indexName, {});
+      helper.search = jest.fn();
+
+      const attribute = 'grade';
+      const widget = makeWidget({
+        attribute,
+      });
+
+      const nextState = widget.dispose({ state: helper.state });
```
samouss: It looks like the previous state does not contain `disjunctiveFacets` neither `disjunctiveFacetsRefinements`. It means that we don't actually test that the `dispose` function removes the refinements. We should apply refinements to the state then call the function.

src/connectors/rating-menu/connectRatingMenu.js
samouss: https://github.com/algolia/instantsearch.js/pull/3976#discussion_r308155858

PR #3993, opened 2019-07-29
fix(toggleRefinement): update lifecycle state

francoischalifour:
## Description

This fixes the lifecycle of `connectToggleRefinement`.

## Changes

- `getConfiguration` sets default `disjunctiveFacetsRefinements`
- `getFacetValues` falls back to an empty array

### DIFF
--- a/src/connectors/toggleRefinement/__tests__/connectToggleRefinement-test.js
+++ b/src/connectors/toggleRefinement/__tests__/connectToggleRefinement-test.js
@@ -589,13 +589,17 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/toggle-refi
       const render = jest.fn();
       const makeWidget = connectToggleRefinement(render);
       const indexName = 'indexName';
-      const helper = jsHelper({}, indexName, {});
+      const helper = jsHelper({}, indexName, {
+        disjunctiveFacets: ['freeShipping'],
+      });
       helper.search = jest.fn();
 
       const attribute = 'freeShipping';
       const widget = makeWidget({
         attribute,
       });
+
+      helper.addDisjunctiveFacetRefinement('freeShipping', ['true']);
 
       const nextState = widget.dispose({ state: helper.state });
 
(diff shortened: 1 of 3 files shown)


src/connectors/toggleRefinement/__tests__/connectToggleRefinement-test.js
```diff
@@ -521,6 +546,67 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/toggle-refi
...
+      const indexName = 'indexName';
+      const helper = jsHelper({}, indexName, {});
+      helper.search = jest.fn();
+
+      const attribute = 'freeShipping';
+      const widget = makeWidget({
+        attribute,
+      });
+
+      const nextState = widget.dispose({ state: helper.state });
```
samouss: It looks like the previous state does not contain `disjunctiveFacets` neither `disjunctiveFacetsRefinements`. It means that we don't actually test that the `dispose` function removes the refinements. We should apply refinements to the state then call the function.

src/connectors/toggleRefinement/connectToggleRefinement.js
```diff
@@ -105,10 +105,13 @@ export default function connectToggleRefinement(renderFn, unmountFn = noop) {
...
       $$type: 'ais.toggleRefinement',
 
-      getConfiguration() {
-        return {
+      getConfiguration(state) {
+        return state.setQueryParameters({
           disjunctiveFacets: [attribute],
-        };
+          disjunctiveFacetsRefinements: {
+            [attribute]: [],
```
samouss: We don't use the previous value, it's inconsistent with the other changes currently made. But we won't use it once we merge the lifecycles so it's fine.

src/connectors/toggleRefinement/__tests__/connectToggleRefinement-test.js
francoischalifour: Fixed in 05fc47855cd48ec79f4253b4de9a5c049e727f38.

src/connectors/toggleRefinement/connectToggleRefinement.js
Haroenv: Not sure what you mean here, why is the Current refinement not taken in account?

PR #3994, opened 2019-07-29
fix(configure): update lifecycle state

francoischalifour:
## Description

This fixes the lifecycle of `connectConfigure`.

## Changes

- `refine` relies on `mergeSearchParameters` directly instead of passing a virtual widget to `enhanceConfiguration`
- `removeSearchParameters` was removed from the widget scope to not pollute it
- `connectConfigure` tests now also check that search parameters are resolved correctly (e.g. with facets as an array)

### DIFF
--- a/src/connectors/configure/connectConfigure.js
+++ b/src/connectors/configure/connectConfigure.js
@@ -1,7 +1,8 @@
+import algoliasearchHelper from 'algoliasearch-helper';
 import {
   createDocumentationMessageGenerator,
-  enhanceConfiguration,
   isPlainObject,
+  mergeSearchParameters,
   noop,
 } from '../../lib/utils';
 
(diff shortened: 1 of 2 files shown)


src/connectors/configure/connectConfigure.js
```diff
@@ -60,10 +76,14 @@ export default function connectConfigure(renderFn = noop, unmountFn = noop) {
...
           // merge new `searchParameters` with the ones set from other widgets
-          const actualState = this.removeSearchParameters(helper.state);
-          const nextSearchParameters = enhanceConfiguration(actualState, {
-            getConfiguration: () => searchParameters,
-          });
+          const actualState = getInitialSearchParameters(
+            helper.state,
+            widgetParams
+          );
+          const nextSearchParameters = mergeSearchParameters(
```
Haroenv: good call on changing this

src/connectors/configure/connectConfigure.js
```diff
@@ -60,10 +76,14 @@ export default function connectConfigure(renderFn = noop, unmountFn = noop) {
...
+            actualState,
+            new algoliasearchHelper.SearchParameters(searchParameters)
```
Haroenv: just like before, it won't be removed I think if you add a parameter which wasn't in widgetParams, can we add a test for that?

francoischalifour: Can you elaborate the test case?

PR #3995, opened 2019-07-30
refactor(configure): migrate to TypeScript

francoischalifour:
## Description

This refactors `connectConfigure` and `configure` to TypeScript.

The changes are mainly type definitions which are as long as the source so Git thinks they're whole new files.

I added some basic tests for the `configure` widget.

## Result

- I removed the widget scope pollution by moving `this._refine` to a variable.
- The connector/widget are now typed


src/widgets/configure/configure.ts
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * A list of [search parameters](https://www.algolia.com/doc/api-reference/search-api-parameters/)
+ * to enable when the widget mounts.
+ */
+type ConfigureWidgetParams = PlainSearchParameters;
+
+type Configure = WidgetFactory<ConfigureWidgetParams>;
+
+const configure: Configure = (searchParameters: ConfigureWidgetParams) => {
```
samouss:
```suggestion
const configure: Configure = (widgetParams: ConfigureWidgetParams) => {
```

IMO it would be simpler to understand.

src/connectors/configure/__tests__/connectConfigure-test.ts
```diff
@@ -157,7 +166,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/configure/j
 
     refine({ hitsPerPage: 3, facets: ['rating'] });
 
-    expect(widget.getConfiguration(new SearchParameters({}))).toEqual(
+    expect(widget.getConfiguration!(new SearchParameters({}))).toEqual(
```
samouss: To avoid the `!` inside the test we can override the definition to force the presence of `getConfiguration`. This is what we do inside the `index` widget to avoid the problem. The `index` widget doesn't use the factory though so it might be a bit more complicated to implement.

src/widgets/configure/configure.ts
francoischalifour: Since we already use this pattern in other connectors tests, I think this can wait until we have a proper widget definition interface with only required lifecycle hooks. WDYT?

src/connectors/configure/__tests__/connectConfigure-test.ts
francoischalifour: Since we already use this pattern in other connectors tests, I think this can wait until we have a proper widget definition interface with only required lifecycle hooks. WDYT?

samouss: Yep sure! It was not for this PR anyway?

PR #3996, opened 2019-07-30
chore: merge develop into next

francoischalifour:
This syncs the `next` branch with the latest changes on the develop branch. Doing this from time to time will help maintaining the two different codebases.

- We made the code a bit more robust on `develop` and added some tests that we can leverage on `next`.
- An annoying part was the upgrade to Prettier 18 that doesn't format the same on the two branches.
- The e-commerce example works on CodeSandbox on the `develop` branch.
- Etc.


samouss:
@francoischalifour You've tried to rebase?

francoischalifour:
@samouss I didn't because merging seemed to be easier here. Do you think rebasing would have been more straightforward in that case?

samouss:
@francoischalifour I've no idea but it would make the history better if we merge the branch once we release the branch `next` as `stable`.

francoischalifour:
@samouss That's true but I think we'd be more likely to mess up some changes because we miss the big picture when rebasing that many changes.

Here we're still mostly aware of the branch differences (as opposed to commits) so it's easier to keep them sync.

Next time we can try rebasing.

PR #3998, opened 2019-07-30
feat(test): POC e2e tests

yannickcr:
**Summary**

This PR contains a sample End-2-End test and a configuration for each tested service.

See related gist for details and instructions about how to run tests on each service.


e2e/test.js
```diff
@@ -0,0 +1,51 @@
...
+    browser.maximizeWindow();
+  });
+
+  it('must work', async () => {
+    await browser.url(
+      'https://deploy-preview-3967--instantsearchjs.netlify.com/examples/e-commerce/'
+    );
+
+    // Select "Apple" brand in list
+    const brand = await browser.$('span=Apple');
```
Haroenv: is there a reason this isn't `querySelector`?

e2e/test.js
```diff
@@ -0,0 +1,51 @@
...
+    await brand.click();
+
+    // Fill search input with "macbook"
+    const searchInput = await browser.$('[type=search]');
+    await searchInput.setValue('macbook');
+
+    // Wait for the results list to be updated (wait for the "macbook" word to be highlighted)
+    await browser.waitUntil(
+      async () => (await browser.$$('mark=MacBook')).length > 0,
+      10000
```
Haroenv: can we not rely on the search callback instead of a magic number?

e2e/test.js
yannickcr: We are not in the browser context here so we have to use the selector from WebdriverIO. It is a bit different than `querySelector` https://webdriver.io/docs/selectors.html

e2e/test.js
yannickcr: What do you mean by search callback? You can find the API documentation here https://webdriver.io/docs/api.html

e2e/test.js
francoischalifour: Can we rely on something than `span`? This seems flaky and we shouldn't describe where to get the information, but rather what information to get. In that case we want the `label` (this is what drives accessibility) element having as value `"Apple"`.

e2e/test.js
```diff
@@ -0,0 +1,51 @@
...
+    );
+
+    // Get title for each result
+    const hits = await browser.$$('.hit h1');
+    const hitsText = await Promise.all(hits.map(hit => hit.getText()));
```
francoischalifour: Why do we extract only the text and not the HTML for also the highlighted parts? This seems to cover better the actual usage.

wdio.browserstack.config.js
```diff
@@ -0,0 +1,62 @@
+const config = {
+  hostname: 'hub.browserstack.com',
+  port: 443,
+  baseUrl: '',
+  user: process.env.BROWSERSTACK_USER,
```
francoischalifour: Not sure yet if these E2E tests are also meant to be run on our local machines. If so, we can consider adding [`dotenv`](https://github.com/motdotla/dotenv) to our stack.

wdio.browserstack.config.js
```diff
@@ -0,0 +1,62 @@
...
+  // Capabilities Generator
+  // https://www.browserstack.com/automate/capabilities?tag=selenium-4
+  commonCapabilities: {
+    'browserstack.debug': true,
+    'browserstack.video': true,
+    'browserstack.console': 'verbose',
+    'browserstack.networkLogs': true,
+  },
+
+  capabilities: [
```
francoischalifour: What in the time impact of adding multiple other combinations?

e2e/test.js
Haroenv: We can wait for the `result` event of `search` for example. However that isn't very portable between frameworks. Since we know there's only one search done, the search client could be injected & waited for maybe

e2e/test.js
yannickcr: The `label` element have the value `Apple\n442` and the linebreak seems to trigger some issues with the selector (I did not take the time to try to fix it).

But we can also use CSS class selectors if needed. https://webdriver.io/docs/selectors.html

e2e/test.js
yannickcr: Here I'm only testing that I got the right list of results, I do not test that the text is correctly highlighted.

wdio.browserstack.config.js
yannickcr: It depends of the offer on the test service.

With an offer like [Automate Pro on BrowserStack](https://www.browserstack.com/accounts/subscriptions) we can run 5 tests in parallel, so 5 combinations in total without impacting the test duration.

e2e/test.js
yannickcr: Since it is an E2E test we're not supposed to interact with InstantSearch differently from a user.

We should only wait for a visual feedback that the search is done, like a real user.

wdio.browserstack.config.js
eunjae-lee: When the e2e tests fail on CI server, we'll probably want to try it again on our local machines. So let's consider adding it :)

e2e/test.js
eunjae-lee: I wish to be able to use [ByLabelText](https://testing-library.com/docs/dom-testing-library/api-queries#bylabeltext) here :(

If `span` sounds flaky, IMO, the class names are less flaky since we try to maintain all the same class names across the flavors(which is far from a good way to test in accessibility-wise, and which is testing implementation details).

samouss: Since we use them for E2E tests we can imagine the usage for `[test-id]` attribute in the DOM.

yannickcr: Yes, that's also a possibility.

tkrugg: test- attributes could be nice by I don't see how we can do it without altering every the DOM of every widget.
We're pretty lucky our css classes are pretty stable across flavours, using them should be good enough

e2e/test.js
```diff
@@ -0,0 +1,49 @@
+const expect = require('chai').expect;
+
+describe('InstantSearch e-commerce demo', () => {
+  before(() => {
+    browser.maximizeWindow();
+  });
+
+  it('must work', async () => {
```
tkrugg: For end to end test, I'd tend to use `it` in a slightly different way that in a unit test, for readability of the report and to clarify the intent of the commands that failed.
Example in this case:

```js
it('navigates to the e-commerce demo', () => {
   await browser.url('http://localhost:5000/examples/e-commerce/');
});
it('selects Apple in the brands refinement list', () => {
   const brand = await browser.$('span=Apple');
   await brand.click();
});
```

I'm just suggesting. This test is small enough that it's doesn't beg for it 😄

e2e/test.js
```diff
@@ -0,0 +1,49 @@
...
+      10000
+    );
+
+    // Get title for each result
+    const hits = await browser.$$('.hit h1');
+    const hitsText = await Promise.all(hits.map(hit => hit.getText()));
+
+    // Compare them to expected titles
+    expect(hitsText).to.deep.equal([
+      'Apple - MacBook Air® (Latest Model) - 13.3" Display - Intel Core i5 - 8GB Memory - 128GB Flash Storage - Silver',
```
tkrugg: Here I think what we expect is to have Apple in all the hits text so maybe we should just check that?
There are less changes the test will fail if for some reason the ranking a or hit changed.

examples/calendar-widget/package.json
```diff
@@ -3,6 +3,6 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "build": "rm -rf ./dist/ && rsync -r . ./dist/"
+    "build": "rm -rf ./dist/ && mkdir dist && cp -r `ls -A | grep -v 'dist'` ./dist/"
```
tkrugg: what was wrong with rsync?

yannickcr: `rsync` is not available on CircleCI

e2e/test.js
yannickcr: I do not have any opinion on this, it makes the test a bit longer but one advantage of this is that it annotates the test, for example on Sauce Labs you'll get:

Screen Shot 2019-08-01 at 18 30 52

(this is similar to the use of [`sauce:context`](https://wiki.saucelabs.com/display/DOCS/Annotating+Tests+with+Selenium%27s+JavaScript+Executor ))

e2e/test.js
yannickcr: Mmm, I prefer to test that we get exactly the results we expect, and checking for only "Apple" is too vague IMO.

I find this as a good thing to have the test to fail if the ranking or a hit change :innocent:

e2e/test.js
francoischalifour: Maybe this can help: https://github.com/testing-library/testcafe-testing-library.

yannickcr: Sadly this is not compatible with WebdriverIO :disappointed:

e2e/test.js
Haroenv: This is something we can do with a dedicated demo: we can show the value of the attribute that will be filtered in the UI so we can test it

examples/calendar-widget/package.json
Haroenv: are all examples built on circle? don't we just want the one for the test?

e2e/test.js
yannickcr: That would be equivalent to exposing some internals for testing, not a good practice and it would ruin the purpose of having end-2-end tests.

Haroenv: Not really true, the attribute is what proves that it's filtered, since the filter isn't related to the search query per se, and can be in multiple attributes

yannickcr: Let's not forget that we are supposed to test from a end user POV, so what matter at the end is that the result list is correct.

yannickcr:
Implementation: 
* https://github.com/algolia/instantsearch-e2e-tests 
* https://github.com/algolia/instantsearch.js/pull/4035

PR #4000, opened 2019-07-31
refactor(currentRefinements): migrate to TypeScript

francoischalifour:
_This anticipates the implementation of the scoped results for displaying items of siblings and child indexes in `CurrentRefinements`._

This refactors the `CurrentRefinements` connector/widget/component to TypeScript.

TypeScript is useful for this widget because we manipulate multiple data structures at different steps.


src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -0,0 +1,320 @@
...
+    (includedAttributes || []).indexOf('query') !== -1 ||
+    (excludedAttributes || []).indexOf('query') === -1;
+
+  const filterFunction = includedAttributes
+    ? item => includedAttributes.indexOf(item.attributeName) !== -1
+    : item => excludedAttributes.indexOf(item.attributeName) === -1;
+
+  const items = getRefinements(results, state, clearsQuery)
+    .filter(filterFunction)
+    // @ts-ignore: @TODO the types don't match exactly, this will need to get refactored.
```
francoischalifour: This was already not completely accurate. We'll need to reiterate on the final types once the helper is fully typed anyway (wrong types are coming from `getRefinement.ts`).

src/widgets/current-refinements/__tests__/current-refinements-test.ts
```diff
@@ -64,39 +73,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/current-ref
       });
     });
 
-    describe('options.templates', () => {
```
francoischalifour: These tests were useless because we don't support any `templates` option.

src/widgets/current-refinements/__tests__/current-refinements-test.ts
```diff
@@ -158,31 +132,32 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/current-ref
...
+      );
 
       const renderParameters = {
-        results: {
-          facets: [
-            {
-              name: 'facet',
-              data: {
-                'facet-val1': 1,
+        results: new SearchResults(helper.state, [
```
francoischalifour: I am missing something with the `SearchResults` constructor. How do we pass the counts for the facets? (you can see that they're missing in the snapshots, but not in the stories)

src/connectors/current-refinements/connectCurrentRefinements.ts
francoischalifour: Fixed in 92d44525461948622e91bafac6b8b9b69666c74b.

francoischalifour:
So... while digging in the Current Refinements code, I noticed several unused and broken features that [we document in the connector](https://www.algolia.com/doc/api-reference/widgets/current-refinements/js/#create-a-render-function).

Top-level `refine` and `createURL` were never used and most likely broken. They should only exist in the items' refinements themselves:

```ts
export type ConnectorRefinement = {
  attribute: string;
  type: string;
  value: string | number;
  label: string;
  operator?: string;
  count?: number;
  exhaustive?: boolean;
};

export type Item = {
  attribute: string;
  label: string;
  refinements: ItemRefinement[];
  createURL(state: ItemRefinement): string;
};
```

These were not tested so I took the opportunity to at least cover basic cases.

I think we're now on the right track to implement the scoped results inside Current Refinements.

src/components/CurrentRefinements/CurrentRefinements.tsx
```diff
@@ -1,14 +1,30 @@
...
+  type: string;
+  operator?: string;
+}): string =>
   [attribute, type, value, operator]
     .map(key => key)
     .filter(Boolean)
     .join(':');
 
-const handleClick = callback => event => {
+const handleClick = (callback: () => void) => (event: any) => {
```
Haroenv: we might be able to inline this, instead of this HOF

src/connectors/clear-refinements/connectClearRefinements.js
```diff
@@ -179,7 +179,7 @@ function getAttributesToClear({
 
   return transformItems(
     getRefinements(helper.lastResults || {}, helper.state, clearsQuery)
-      .map(refinement => refinement.attributeName)
+      .map(refinement => refinement.attribute)
```
Haroenv: great catch! How long was this broken then?

src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
```diff
@@ -41,25 +51,25 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/current-ref
...
-      const helper = jsHelper({});
-      helper.search = () => {};
+      const helper = algoliasearchHelper(createSearchClient(), '', {});
+      helper.search = jest.fn();
       // test that the dummyRendering is called with the isFirstRendering
       // flag set accordingly
       const rendering = jest.fn();
       const customCurrentRefinements = connectCurrentRefinements(rendering);
-      const widget = customCurrentRefinements({
-        foo: 'bar', // dummy param to test `widgetParams`
```
Haroenv: type here was wrong indeed, but could be useful to test with non-null arguments

src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
```diff
@@ -394,118 +439,112 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/current-ref
...
     beforeEach(() => {
-      helper = jsHelper({}, '', {
+      helper = algoliasearchHelper(createSearchClient(), '', {
         facets: ['facet1', 'facet2', 'facet3'],
       });
-      helper.search = () => {};
+      helper.search = jest.fn();
     });
 
-    it('provides a `refine` function', () => {
```
Haroenv: this is because you first removed the root refine function right?

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -0,0 +1,340 @@
...
+      );
+    case 'hierarchical':
+      return state.removeHierarchicalFacetRefinement(refinement.attribute);
+    case 'exclude':
+      return state.removeExcludeRefinement(
+        refinement.attribute,
+        refinement.value
+      );
+    case 'numeric':
+      // @ts-ignore @TODO: fix non-existant `removeNumericRefinement` type in the helper typings
```
Haroenv: https://github.com/algolia/algoliasearch-helper-js/pull/742

src/widgets/current-refinements/__tests__/__snapshots__/current-refinements-test.ts.snap
```diff
@@ -22,16 +22,12 @@ exports[`currentRefinements() render() DOM output renders correctly 1`] = `
         "refinements": Array [
           Object {
             "attribute": "facet",
-            "count": 1,
-            "exhaustive": true,
```
Haroenv: what caused these changes?

src/widgets/current-refinements/__tests__/__snapshots__/current-refinements-test.ts.snap
```diff
@@ -102,8 +92,8 @@ exports[`currentRefinements() render() DOM output renders correctly 1`] = `
         "refinements": Array [
           Object {
             "attribute": "hierarchicalFacet",
-            "count": 6,
-            "exhaustive": true,
+            "count": null,
+            "exhaustive": null,
```
Haroenv: doesn't look correct either TBH

src/connectors/clear-refinements/connectClearRefinements.js
francoischalifour: That wasn't broken. The internal implementation created a key called `attributeName`. I changed it to `attribute` to type this data structure more easily. I think it's less confusing this way.

src/widgets/current-refinements/__tests__/__snapshots__/current-refinements-test.ts.snap
francoischalifour: See https://github.com/algolia/instantsearch.js/pull/4000#discussion_r309115051. I think that the `SearchResults` constructor is misused in the tests but I'm not sure how.

src/widgets/current-refinements/__tests__/__snapshots__/current-refinements-test.ts.snap
francoischalifour: See https://github.com/algolia/instantsearch.js/pull/4000#discussion_r309115051.

src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
francoischalifour: I tested it here: https://github.com/algolia/instantsearch.js/blob/db681a241507c5f9caab63fb45c63f605e817080/src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts#L81-L82

src/widgets/current-refinements/__tests__/current-refinements-test.ts
Haroenv: the counts are taken from the second response, bu there you are only having one, and seems to be a wrong facets interface

`[ { facets: { xxx: 5 }}, { facets: { xxxTaken: 5000 }}]`

src/components/CurrentRefinements/CurrentRefinements.tsx
francoischalifour: That's a pattern we use [elsewhere](https://github.com/algolia/instantsearch.js/blob/bc0f9e245f28fe7ee6f9578c86ae833607d36145/src/components/Pagination/Pagination.js#L102-L110). We can refactor this later.

src/widgets/current-refinements/__tests__/__snapshots__/current-refinements-test.ts.snap
francoischalifour: I fixed it for other facets than hierarchical. I'm not sure what syntax is missing for this one.

src/connectors/current-refinements/connectCurrentRefinements.ts
Haroenv: https://github.com/algolia/instantsearch.js/pull/4034

src/widgets/current-refinements/__tests__/__snapshots__/current-refinements-test.ts.snap
francoischalifour: Fixed in 132015b9b4869fdc42fe53ec03cd065142099429.

src/connectors/current-refinements/connectCurrentRefinements.ts
Haroenv:
```suggestion
```

this should work now that the branch is updated

PR #4002, opened 2019-07-31
feat(connectSearchBox): update getWidgetSearchParameters

samouss:
This PR updates the lifecycle `getWidgetSearchParameters` to `connectSearchBox` to take the add the default value without `uiState`. We don't apply the previous value on purpose, the `uiState` drives completely the `SearchParameters`.

### DIFF
--- a/src/connectors/search-box/__tests__/__snapshots__/connectSearchBox-test.js.snap
+++ b/src/connectors/search-box/__tests__/__snapshots__/connectSearchBox-test.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`connectSearchBox routing getWidgetSearchParameters should add the refinements according to the UI state provided 1`] = `
+exports[`connectSearchBox getWidgetSearchParameters should add the refinements according to the UI state provided 1`] = `
 SearchParameters {
   "disjunctiveFacets": Array [],
   "disjunctiveFacetsRefinements": Object {},
@@ -16,7 +16,7 @@ SearchParameters {
 }
 `;
 
-exports[`connectSearchBox routing getWidgetSearchParameters should enforce the default value if no value is the ui state 1`] = `
+exports[`connectSearchBox getWidgetSearchParameters should enforce the default value if no value is the ui state 1`] = `
 SearchParameters {
   "disjunctiveFacets": Array [],
   "disjunctiveFacetsRefinements": Object {},
@@ -31,7 +31,7 @@ SearchParameters {
 }
 `;
 
-exports[`connectSearchBox routing getWidgetState should add an entry equal to the refinement 1`] = `
+exports[`connectSearchBox getWidgetState should add an entry equal to the refinement 1`] = `
 Object {
   "query": "some query",
 }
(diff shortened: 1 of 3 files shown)


src/connectors/search-box/__tests__/__snapshots__/connectSearchBox-test.js.snap
```diff
@@ -1,37 +1,6 @@
...
-  "index": "",
-  "numericRefinements": Object {},
-  "tagRefinements": Array [],
-}
-`;
-
-exports[`connectSearchBox routing getWidgetState should add an entry equal to the refinement 1`] = `
+exports[`connectSearchBox getWidgetState should add an entry equal to the refinement 1`] = `
 Object {
   "query": "some query",
```
Haroenv: let's remove this snapshot here as well

src/connectors/search-box/connectSearchBox.js
```diff
@@ -158,8 +158,8 @@ export default function connectSearchBox(renderFn, unmountFn = noop) {
         };
       },
 
-      getWidgetSearchParameters(searchParameters, { uiState }) {
-        return searchParameters.setQueryParameter('query', uiState.query);
+      getWidgetSearchParameters(state, { uiState }) {
+        return state.setQueryParameter('query', uiState.query || '');
```
Haroenv: Do you just want this to be consistent with the getConfiguration here? Not sure what flow will work best for all widgets.

samouss: WDYM?

src/connectors/search-box/__tests__/__snapshots__/connectSearchBox-test.js.snap
samouss: I didn't to avoid to touch too much code. It's not related to this PR at all.

src/connectors/search-box/connectSearchBox.js
Haroenv: Cleared it up in person, I was talking about the goal of the PR here, where we make sure the changes done in getConfiguraiton are replicated in getWidgetSearchParameters

src/connectors/search-box/__tests__/connectSearchBox-test.js
```diff
@@ -377,105 +399,89 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/search-box/
...
           helper,
-        });
-        expect(uiStateAfter).toBe(uiStateBefore);
+        }
+      );
+      const uiStateAfter = widget.getWidgetState(uiStateBefore, {
+        searchParameters: helper.state,
+        helper,
       });
+      expect(uiStateAfter).toBe(uiStateBefore);
```
francoischalifour: Do we care about getting the same instance? We just want to have the same content, right?

src/connectors/search-box/__tests__/connectSearchBox-test.js
```diff
@@ -377,105 +399,89 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/search-box/
...
-
-      return [widget, helper, refine];
-    };
-
-    describe('getWidgetState', () => {
-      test('should give back the object unmodified if the default value is selected', () => {
-        const [widget, helper] = getInitializedWidget();
-        const uiStateBefore = {};
-        const uiStateAfter = widget.getWidgetState(uiStateBefore, {
+    test('should give back the same instance if the value is alreay in the uiState', () => {
```
francoischalifour:
```suggestion
    test('should give back the same instance if the value is already in the uiState', () => {
```

src/connectors/search-box/__tests__/connectSearchBox-test.js
Haroenv: if it's ever updated we can change this assertion I guess?

PR #4003, opened 2019-07-31
fix(index): subscribe to state change only after init for uiState

samouss:
We've now introduced the `uiState` at the index level. One thing we want to avoid is to subscribe to state change before the `init` step. This step has always been considered as a configuration step for InstantSearch, especially for custom widgets. We don't want to notify of an update at this step. One side effect could be to push a useless entry into the URL history. 

To avoid this issue we delayed the subscription once all the widgets are already initialised. Inside the [`RoutingManager`](https://github.com/algolia/instantsearch.js/blob/299ffe28006deb516bdc969da8089e703623be6b/src/lib/RoutingManager.ts#L158-L163) we have a similar approach expect that it's performed at the first `render` not after `init`. Not sure to understand why but if we do have issues in the future about sync we could replicate the first render subscription inside the index.

> https://github.com/algolia/instantsearch.js/pull/994/commits/4a672ae3fd78809e213de0368549ef12e9dc9454


src/widgets/index/__tests__/index-test.ts
```diff
@@ -858,6 +858,43 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
...
+        new SearchParameters({
+          index: 'index_name',
+          query: 'Apple iPhone',
+          page: 5,
+        })
+      );
+
+      expect(instance.getWidgetState({})).toEqual({
+        // eslint-disable-next-line @typescript-eslint/camelcase
+        index_name: {},
```
Haroenv: could we just name this something not in camel case?

samouss: Yes we can, will do it in the next PR (which is about testing).

PR #4004, opened 2019-07-31
feat(connectPagination): update getWidgetSearchParameters

samouss:
This PR updates the lifecycle `getWidgetSearchParameters` in `connectPagination` to add the default value without `uiState`. We don't apply the previous value on purpose, the `uiState` drives completely the `SearchParameters`.


src/connectors/pagination/connectPagination.js
```diff
@@ -180,13 +180,9 @@ export default function connectPagination(renderFn, unmountFn = noop) {
       },
 
       getWidgetSearchParameters(searchParameters, { uiState }) {
-        const uiPage = uiState.page;
+        const page = uiState.page ? uiState.page - 1 : 0;
```
francoischalifour: It seamed to be there before but can you explain why we set the page to `page - 1` when there's a UI state for page? Do we considerer `page: 1` to be the first page (i.e. `0` in code)?

If so, it's worth adding a comment.

src/connectors/pagination/__tests__/connectPagination-test.js
```diff
@@ -361,113 +380,76 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/pagination/
...
+      const uiStateBefore = {};
+      const uiStateAfter = widget.getWidgetState(uiStateBefore, {
+        searchParameters: helper.state,
         helper,
-        state: helper.state,
-        createURL: () => '#',
       });
 
-      const { refine } = renderFn.mock.calls[0][0];
+      expect(uiStateAfter).toBe(uiStateBefore);
```
francoischalifour: We don't care about it being the same reference. We only want to check the value, right?

src/connectors/pagination/__tests__/connectPagination-test.js
```diff
@@ -361,113 +380,76 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/pagination/
...
 
-    describe('getWidgetState', () => {
-      test('should give back the object unmodified if the default value is selected', () => {
-        const [widget, helper] = getInitializedWidget();
-        const uiStateBefore = {};
-        const uiStateAfter = widget.getWidgetState(uiStateBefore, {
-          searchParameters: helper.state,
-          helper,
-        });
+      expect(uiStateAfter).toMatchSnapshot();
```
francoischalifour: Can we not use a snapshot for this? That's small enough to use `toEqual` and it creates less context switch.

src/connectors/pagination/__tests__/connectPagination-test.js
```diff
@@ -361,113 +380,76 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/pagination/
...
 
-      test('should add an entry equal to the refinement', () => {
-        const [widget, helper, refine] = getInitializedWidget();
-        refine(4);
-        const uiStateBefore = {};
-        const uiStateAfter = widget.getWidgetState(uiStateBefore, {
-          searchParameters: helper.state,
-          helper,
-        });
+      expect(uiStateAfter).toBe(uiStateBefore);
```
francoischalifour: Same comment regarding the value instead of the reference.

src/connectors/pagination/__tests__/connectPagination-test.js
samouss: At some point IIRC it was important. But anyway my PR move this code but didn't change it at all. Only the code related to `getWidgetSearchParameters` changes (implem + tests).

src/connectors/pagination/__tests__/connectPagination-test.js
samouss: https://github.com/algolia/instantsearch.js/pull/4004#discussion_r309580941

src/connectors/pagination/__tests__/connectPagination-test.js
samouss: https://github.com/algolia/instantsearch.js/pull/4004#discussion_r309580941

src/connectors/pagination/connectPagination.js
samouss: Yes it is. At the API/`SearchParameters` level page is indexed on 0. Inside the `uiState` it's indexed on 1. Most of the time users expect the first page to be `1` not `0`, especially inside their URL. That's why we have this (see the [`getWidgetState`](https://github.com/algolia/instantsearch.js/blob/299ffe28006deb516bdc969da8089e703623be6b/src/connectors/pagination/connectPagination.js#L158-L165)).

src/connectors/pagination/__tests__/connectPagination-test.js
francoischalifour: It's still an opportunity to improve this, that's a quick win and a small effort both for you and the reviewers.

samouss: https://github.com/algolia/instantsearch.js/pull/4004/commits/72adb233342dd1e476ecd00864144e8fbc0df0c4

PR #4007, opened 2019-07-31
test(sortby): remove snapshots

Haroenv:
This is just a small thing, but going through the tests I'm noticing this indirection is annoying.


Haroenv:
There’s more of these assertions, but I’ll make individual PRs as I have the bandwidth to update them or if the files are updated anyway

PR #4008, opened 2019-07-31
feat(connectRatingMenu): update getWidgetSearchParameters 

samouss:
This PR updates the lifecycle `getWidgetSearchParameters` in `connectRatingMenu` to add the configuration of facets and default value. We don't apply the previous value on purpose, the `uiState` drives completely the `SearchParameters`.


src/connectors/rating-menu/__tests__/connectRatingMenu-test.js
```diff
@@ -370,116 +395,159 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/rating-menu
...
-      const rendering = jest.fn();
-      const makeWidget = connectRatingMenu(rendering);
-
-      const attribute = 'grade';
-      const widget = makeWidget({
-        attribute,
-        ...config,
+  describe('getWidgetState', () => {
+    test('should give back the object unmodified if the default value is selected', () => {
+      const [widget, helper] = getInitializedWidget();
```
eunjae-lee: Not a big thing at all, but if `getInitializedWidget` returned `{widget, helper, refine}` instead of `[widget, helper, refine]`, it would be easier to destructure it without having to think about the returned order, like
`const { helper, widget } = getInitializedWidget();`.

And `const {widget, refine} = ...` instead of `const [widget, , refine] = ...`.

samouss: Indeed it's a fair point, we'll revisit this function anyway because we can't even use it right now inside the tests for the `getWidgetSearchParameters`. We can do this change at the same time (the function is not part of the PR, it just moved).

src/connectors/rating-menu/__tests__/connectRatingMenu-test.js
```diff
@@ -370,116 +395,159 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/rating-menu
...
-      helper.search = jest.fn();
+    test('should add an entry equal to the refinement', () => {
+      const [widget, helper, refine] = getInitializedWidget();
+      refine('3');
+      const uiStateBefore = {};
+      const uiStateAfter = widget.getWidgetState(uiStateBefore, {
+        searchParameters: helper.state,
+        helper,
+      });
+      expect(uiStateAfter).toMatchSnapshot();
```
francoischalifour: Same comment as other PRs about the snapshot, I think we can afford to change it to `toEqual` in these PRs.

samouss: https://github.com/algolia/instantsearch.js/pull/4008/commits/46fd50601f8f5a47728166baa1aab0920eb535ba

PR #4012, opened 2019-08-01
feat(currentRefinements): support multiple indices

francoischalifour:
## Description

This implements multi-index context in `CurrentRefinements` leveraging `scopedResults` (#3964).

A `CurrentRefinements` widget now displays the refinements coming from its siblings and its children.

## Internal changes

The `index` now calls the `render` hook with an enhanced `scopedResults` containing the helper for this index. This is needed to call `refine` on each item coming from the `CurrentRefinements`. Otherwise, calling `setState` on the helper doesn't change the correct helper state (and thus breaks).

The new structure is:

```ts
interface ScopedResult {
  indexId: string;
  results: SearchResults;
  helper: Helper;
}
```

A new property `indexName` is added to the CurrentRefinements items. This property is given to the renderer. This `indexName` can be used to display the index where the attribute is coming from, or compute a node key for component-based flavors (e.g. React). The new structure is:

```diff
 type Item = {
+  indexName: string;
   attribute: string;
   label: string;
   refinements: ItemRefinement[];
   refine(refinement: ItemRefinement): void;
 };
```

## Example

```js
search.addWidgets([
  // Displays both the refinements for `index1` and `index2`
  instantsearch.widgets.currentRefinements({
    container,
  }),

  instantsearch.widgets
    .index({
      indexName: 'index1',
    })
    .addWidgets([
        container,
        attribute: 'brand',
      }),
    ]),

  instantsearch.widgets
    .index({
      indexName: 'index2',
    })
    .addWidgets([
        container,
        attribute: 'locations',
      }),
    ]),
]);
```

## Preview

[See story →](http://deploy-preview-4012--instantsearchjs.netlify.com/stories/?path=/story/currentrefinements--with-multi-indices)

### DIFF
--- a/src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
+++ b/src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
@@ -490,9 +490,17 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/current-ref
 
       widget.render!(
         createRenderOptions({
-          results: new SearchResults(helper.state, [
-            createSingleSearchResponse(),
-          ]),
+          scopedResults: [
+            {
+              indexId: 'firstIndex',
+              helper,
+              results: new SearchResults(helper.state, [
+                createSingleSearchResponse({
+                  index: 'firstIndex',
+                }),
+              ]),
+            },
+          ],
           state: helper.state,
           helper,
         })
@@ -500,6 +508,8 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/current-ref
 
       const secondRenderingOptions = rendering.mock.calls[1][0];
       const items: Item[] = secondRenderingOptions.items;
+
+      expect(items).toHaveLength(2);
       expect(items).toEqual([
         {
           attribute: 'facet1',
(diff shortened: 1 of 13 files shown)


src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
```diff
@@ -547,5 +557,109 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/current-ref
...
+              label: 'facetValue',
+              type: 'facet',
+              value: 'facetValue',
+            },
+          ],
+          refine: expect.any(Function),
+        },
+        {
+          attribute: 'facet1',
+          label: 'facet1',
```
Haroenv: aren't these duplicate? This will mean we can no longer use `attribute` or `label` as key

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -166,16 +165,19 @@ const connectCurrentRefinements: CurrentRefinementsConnector = (
...
-            includedAttributes,
-            excludedAttributes,
-          })
-        );
+      render({ scopedResults, helper, createURL, instantSearchInstance }) {
+        const items = scopedResults.reduce<Item[]>((results, scopedResult) => {
+          return results.concat(
+            transformItems(
+              getItems({
+                results: scopedResult.results,
```
Haroenv: simply  calling it like this will not remove duplicates, as shown in the test, we need to make sure we fetch "top-down" and ignore ones we already have. Then we need to be careful, because they could have different values, in which case there would still be two.

src/widgets/current-refinements/__tests__/current-refinements-test.ts
```diff
@@ -389,39 +417,45 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/current-ref
...
+                        name: 'extraFacet',
+                        exhaustive: true,
+                        data: {
+                          'extraFacet-val1': 42,
+                          'extraFacet-val2': 42,
+                        },
+                      },
+                    ],
+                    // @ts-ignore @TODO: missing types for facets in the response
+                    disjunctiveFacets: [
```
Haroenv: same remark here and in the other instances, we need to make sure we don't ignore these errors, since they actually behave wrong (skip helper steps, which won't happen IRL)

src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
francoischalifour: Can you be more explicit by "duplicate"?

If you mean UI duplicate, this is what we already do in React InstantSearch and what you plan to do in the first iteration for InstantSearch.js too. We don't plan to prefix the label by the index name because it's not useful from the user POV.

Note that it is rather unlikely that you've got the same attribute in nested indices.

src/connectors/current-refinements/connectCurrentRefinements.ts
francoischalifour: See https://github.com/algolia/instantsearch.js/pull/4012#discussion_r312417432.

src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
Haroenv: I mean totally  identical. We can't avoid showing them twice in the UI, but we can avoid React errors if we have enough information to make them unique for the key

francoischalifour: Right, that makes sense! I'll add the `indexName` key to the refinement items.

src/connectors/current-refinements/connectCurrentRefinements.ts
francoischalifour: Fixed f994a628ca7d29896e6cd25c2219a1c6ec4eff5f.

src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
francoischalifour: Fixed f994a628ca7d29896e6cd25c2219a1c6ec4eff5f.

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -230,6 +231,7 @@ function getItems({
         (item: Item) => item.attribute !== currentItem.attribute
       ),
       {
+        indexName: helper.state.index,
```
Haroenv: good place for this!

PR #4013, opened 2019-08-02
feat(connectNumericMenu): update state lifecycle 

samouss:
This PR updates the `getWidgetState`/`getWidgetSearchParameters` in `connectNumericMenu`. 

The `getWidgetState` function is simplified, we don't preserve the reference equality anymore. We perform a deep inside the `RoutingManager` anyway. Agree that with with a reference check it would be faster, but for 99% of cases the `uiState` will actually change. It means that even with the ref check will go through the deep equal comparison. I've also removed the snapshots.

The `getWidgetSearchParamters` function adds default value when the `uiState` is empty. We don't apply the previous value on purpose, the `uiState` drives completely the `SearchParameters`.


src/connectors/numeric-menu/connectNumericMenu.js
```diff
@@ -170,80 +170,67 @@ export default function connectNumericMenu(renderFn, unmountFn = noop) {
...
+        const isExact = value.indexOf(':') === -1;
 
-        if (valueAsEqual) {
-          return clearedParams.addNumericRefinement(
-            attribute,
-            '=',
-            valueAsEqual
-          );
+        if (isExact) {
+          return withoutRefinements.addNumericRefinement(attribute, '=', value);
```
samouss: We should have the same path than for `min`/`max` i.e. `parseFloat` -> `isFiniteNumber `.

src/connectors/numeric-menu/connectNumericMenu.js
```diff
@@ -170,80 +170,67 @@ export default function connectNumericMenu(renderFn, unmountFn = noop) {
...
-        if (isFiniteNumber(lowerBound)) {
-          clearedParams = clearedParams.addNumericRefinement(
-            attribute,
-            '>=',
-            lowerBound
-          );
-        }
+        const withMinRefinement = isFiniteNumber(min)
+          ? withoutRefinements.addNumericRefinement(attribute, '>=', min)
+          : withoutRefinements;
```
samouss: Here the value won't have the default value but it should be. We have to perform all the checks at the top to ensure that if any of the value is incorrect we return the default value. We can implement the two fix at the same time since they are related.

src/connectors/numeric-menu/__tests__/connectNumericMenu-test.js
```diff
@@ -573,246 +599,284 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/numeric-men
...
-      const widget = makeWidget({
-        attribute: 'numerics',
-        items: [
-          { label: 'below 10', end: 10 },
-          { label: '10 - 20', start: 10, end: 20 },
-          { label: 'more than 20', start: 20 },
-        ],
-      });
+  describe('getWidgetState', () => {
+    test('returns the `uiState` emtpy', () => {
```
francoischalifour: "empty"

src/connectors/numeric-menu/connectNumericMenu.js
francoischalifour: In a separate PR then?

samouss: Yes.

src/connectors/numeric-menu/__tests__/connectNumericMenu-test.js
samouss: https://github.com/algolia/instantsearch.js/pull/4013/commits/ef00be3a939f5aa5bbd3783f5f84a875da25d44d

PR #4015, opened 2019-08-02
fix(connectSortBy): never update the initial index

samouss:
We've updated the `sortBy` inside a recent [PR](https://github.com/algolia/instantsearch.js/pull/3824). We indeed fix an issue with it but we also introduced one. We've removed the `currentIndex` to always use the `initialIndex`. But the two value aren't the same. The `currentIndex` is the current value, which might come from the top-level options or the URLSync. The `initialIndex` is **always** the one provided at the top-level options. 

The `getWidgetState` function compares both values to determine whether or not the current value should go inside the `uiState`. We never push the value of the `initialIndex` to avoid to pollute the URL. But if the value depends on the one provided at `init` we'll lose the information inside the `uiState` since now the `initialIndex` is equal to the `currentIndex`.

This PR restores the usage of `instantSearchInstance.indexName`.

**Before**

before

**After**

after

### DIFF
--- a/src/connectors/sort-by/connectSortBy.js
+++ b/src/connectors/sort-by/connectSortBy.js
@@ -103,7 +103,7 @@ export default function connectSortBy(renderFn, unmountFn = noop) {
     return {
       init({ helper, instantSearchInstance }) {
         const currentIndex = helper.state.index;
-        const isInitialIndexInItems = find(
+        const isCurrentIndexInItems = find(
           items,
           item => item.value === currentIndex
         );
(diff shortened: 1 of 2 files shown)


src/connectors/sort-by/connectSortBy.js
```diff
@@ -102,25 +102,35 @@ export default function connectSortBy(renderFn, unmountFn = noop) {
 
     return {
       init({ helper, instantSearchInstance }) {
-        const initialIndex = helper.state.index;
+        const currentIndex = helper.state.index;
         const isInitialIndexInItems = find(
```
eunjae-lee:
```suggestion
        const isCurrentIndexInItems = find(
```

what do you think?

PR #4024, opened 2019-08-07
chore(hierarchical): consistent state

Haroenv:
IFW-882

also fixes maxValuesPerFacet on dispose, which wasn't tested.

this requires: https://github.com/algolia/algoliasearch-helper-js/pull/743

**Summary**

**Result**


Haroenv:
This is missing the "refinement" default value still

PR #4025, opened 2019-08-07
fix(RefinementList): remove root css class on sublists

tkrugg:
**Summary**

The RefinementList is the component used for rendering of the HierarchicalMenu widget.
When a given item of the HierarchicalMenu has children, the RefinementList is recursively called, with the same cssClasses prop.

As a result, the `cssClasses.root` is appearing on every list of children.

This change is proposed because:

1. This is not the specified behaviour.
2. This is not the behaviour observed on InstantSearch V2.
  see difference between [InstantSearch V3](https://codesandbox.io/s/ais-ecommerce-demo-app-8snic) and [InstantSearch V2](https://codesandbox.io/s/instantsearchjs-template-7q78t)
  In V2, the `cssClasses.root` belonged to the [headerFooter HOC](https://github.com/algolia/instantsearch.js/blob/v2/src/decorators/headerFooter.js#L22).
  The problem emerged in InstantSearch V3 as the `cssClasses.root` property was moved
  inside the RefinementList widget.

**Result**
```js
throw new Error(NOT_IMPLEMENTED)
```

### DIFF
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -30,7 +30,7 @@ class RefinementList extends Component {
     let subItems;
     const hasChildren = facetValue.data && facetValue.data.length > 0;
     if (hasChildren) {
-      const cssClasses = { ...this.props.cssClasses, root: '' };
+      const { root, ...cssClasses } = this.props.cssClasses;
       subItems = (
         <RefinementList
           {...this.props}
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -30,9 +30,11 @@ class RefinementList extends Component {
     let subItems;
     const hasChildren = facetValue.data && facetValue.data.length > 0;
     if (hasChildren) {
+      const cssClasses = { ...this.props.cssClasses, root: '' };
       subItems = (
         <RefinementList
           {...this.props}
+          cssClasses={cssClasses}
           depth={this.props.depth + 1}
           facetValues={facetValue.data}
           showMore={false}
--- a/src/components/RefinementList/__tests__/RefinementList-test.js
+++ b/src/components/RefinementList/__tests__/RefinementList-test.js
@@ -19,7 +19,7 @@ describe('RefinementList', () => {
       facetValues: [],
       ...extraProps,
     };
-    return shallow(React.createElement(RefinementList, props));
+    return shallow(<RefinementList {...props} />);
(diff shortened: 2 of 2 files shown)


src/components/RefinementList/RefinementList.js
```diff
@@ -30,9 +30,11 @@ class RefinementList extends Component {
     let subItems;
     const hasChildren = facetValue.data && facetValue.data.length > 0;
     if (hasChildren) {
+      const cssClasses = { ...this.props.cssClasses, root: '' };
```
francoischalifour: Removing the `root` property is more suited here because it can be type-checked.

```ts
const { root, ...cssClasses } = this.props.cssClasses;
```

src/components/RefinementList/__tests__/RefinementList-test.js
```diff
@@ -218,6 +223,32 @@ describe('RefinementList', () => {
       expect(root.props().children[2].props.className).toContain('depth-0');
       expect(subList.props().children[2].props.className).toContain('depth-1');
     });
+
+    it('should not add root class for on sub lists', () => {
```
francoischalifour:
```suggestion
    it('should not add root class on sub lists', () => {
```

tkrugg:
> This should target `next` because it's technically a breaking change.

It doesn't have to be a breaking change since this is isn't the spec we documented. 
https://www.algolia.com/doc/api-reference/widgets/hierarchical-menu/js/#widget-param-cssclasses

francoischalifour:
@tkrugg A bug can turn into an undocumented feature: https://hackernoon.com/its-not-a-bug-it-s-a-feature-the-problem-with-semver-cb067de7843c.

tkrugg:
@francoischalifour haha okay I get the joke 😄

src/components/RefinementList/RefinementList.js
Haroenv:
```suggestion
      const { root, ...cssClasses } = this.props.cssClasses;
```

suggestion to keep it simple, I agree by the way

Haroenv:
i changed the branch from `develop` to master, but looks like you'll need to rebase now, sorry!

francoischalifour:
This PR has been pending for a while. Did you have the chance to read the feedback @tkrugg?

tkrugg:
@Haroenv made an argument not to release this on 3.x but only on next. 
closing this in favour of #4117

PR #4026, opened 2019-08-07
chore(breadcrumb): update life cycles

Haroenv:
i only updated the return value of getConfiguration, since this seems like the only requirement.

IFW-877

### DIFF
--- a/src/connectors/breadcrumb/connectBreadcrumb.js
+++ b/src/connectors/breadcrumb/connectBreadcrumb.js
@@ -94,6 +94,12 @@ export default function connectBreadcrumb(renderFn, unmountFn = noop) {
               rootPath,
             },
           ],
+          hierarchicalFacetsRefinements: {
+            [hierarchicalFacetName]:
+              currentConfiguration.hierarchicalFacetsRefinements[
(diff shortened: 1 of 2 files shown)


src/connectors/breadcrumb/connectBreadcrumb.js
```diff
@@ -81,11 +81,11 @@ export default function connectBreadcrumb(renderFn, unmountFn = noop) {
               'Using Breadcrumb and HierarchicalMenu on the same facet with different options overrides the configuration of the HierarchicalMenu.'
             );
 
-            return {};
+            return currentConfiguration;
           }
         }
 
-        return {
+        return currentConfiguration.setQueryParameters({
```
francoischalifour: No need to add `hierarchicalFacetsRefinements`?

Haroenv: good point, this is missing. I originally didn't add it because someone usually uses it with a hierarchical or a configure which sets that, but it's still better to avoid the merging issues completely

Haroenv: done in c195b49ed

PR #4028, opened 2019-08-08
chore(range): update life cycles

Haroenv:
1. add tests for dispose
2. make getConfiguration accept & return SearchParameters

IFW-891

### DIFF
--- a/src/connectors/range/connectRange.js
+++ b/src/connectors/range/connectRange.js
@@ -218,7 +218,7 @@ export default function connectRange(renderFn, unmountFn = noop) {
           }
         }
 
-        return configuration;
+        return currentConfiguration.setQueryParameters(configuration);
       },
 
       init({ helper, instantSearchInstance }) {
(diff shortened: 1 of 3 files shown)


src/connectors/range/connectRange.js
```diff
@@ -275,11 +275,9 @@ export default function connectRange(renderFn, unmountFn = noop) {
       dispose({ state }) {
         unmountFn();
 
-        const nextState = state
+        return state
           .removeNumericRefinement(attribute)
```
francoischalifour: We've removed the call to `removeXRefinement` in other refactors.

Haroenv: There's no function to remove a numeric facet, just a  single function that takes (or does not take) operator & value, so this is equivalent. Note that the disjunctive facet is only for the counts, and separate from the numeric

francoischalifour: Right!

PR #4030, opened 2019-08-09
chore(examples): update demos to use instantsearch from repo

yannickcr:
**Summary**

To improve InstantSearch releases reliability we plan to add some [End-2-End tests](https://github.com/algolia/instantsearch-e2e-tests) in addition to our existing unit test suite.

Since we'll use the [e-commerce demo](https://github.com/algolia/instantsearch.js/tree/develop/examples/e-commerce) to run our tests then we need to update it to use the `instantsearch.js` build from the current branch.

To achieve this, this PR makes the following changes:
* Remove `instantsearch.js` from dependencies
* Alias `instantsearch.js` to use the build in the parent directory (using [Parcel aliases](https://parceljs.org/module_resolution.html#alias) and [TypeScript path mapping](https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping))
* Update `website:build` command to build `instantsearch` before the demos


Haroenv:
Seems good, but should we not do that for all examples? Downside is now that it won’t work standalone in csb anymore. I’d really rather have a purpose-built example for the tests

yannickcr:
> Seems good, but should we not do that for all examples?

Mmm, I don't think we'll run some tests on them (maybe on the new media demo but it is not merged yet https://github.com/algolia/instantsearch.js/pull/3915) but I can update them to stay consistent :slightly_smiling_face: 

> Downside is now that it won’t work standalone in csb anymore. I’d really rather have a purpose-built example for the tests

Is it a problem? I never saw these demos used in csb.
I think we agreed to use this one for now to go faster but to build some test specific demos later on.

francoischalifour:
I don't remember if we agreed to run it on the actual example. I thought the reasoning was that we should create a minimal example (e.g. no Google Fonts, no image header to speed up the page load). We could add these minimal examples in the [`e2e/` folder](https://github.com/algolia/instantsearch.js/tree/6dc0b3a2b921687f0add0decb8d17dc9888a3b7b/e2e). It's indeed nice to be able to run the example in CodeSandbox (some people asked for it already).

WDYT?

yannickcr:
> I don't remember if we agreed to run it on the actual example. I thought the reasoning was that we should create a minimal example (e.g. no Google Fonts, no image header to speed up the page load). We could add these minimal examples in the e2e/ folder. 

Yes, but it was planned to do this in a second time since already having a running test suite on the existing examples was way faster to set up and already bring us a lot of value to improve InstantSearch reliability.

> It's indeed nice to be able to run the example in CodeSandbox (some people asked for it already).

New examples (like the e-commerce demo) are already not runnable in CodeSansbox anyway, no? Or I am missing something here?

Haroenv:
it works now: https://codesandbox.io/s/github/algolia/instantsearch.js/tree/develop/examples/e-commerce

yannickcr:
Ok, I missed that CodeSandbox runs the parcel build.

Still not convinced that keeping this "feature" worth the effort tho.

yannickcr:
As we decided IRL last week, to go faster we'll use the existing demos for now and see in the future if some dedicated demos are required.

PR #4035, opened 2019-08-12
test: add End-2-End test suite

yannickcr:
**Summary**

To improve InstantSearch releases reliability we plan to add some [End-2-End tests](https://github.com/algolia/instantsearch-e2e-tests) in addition to our existing unit test suite.

Based on https://github.com/algolia/instantsearch.js/pull/4030, this PR integrates this test suite to InstantSearch.js.

This PR makes the following changes:
* Add `algolia/instantsearch-e2-tests` as devDependency (since I'm not sure we want to publish it on npm I link to the GitHub repository).
* Add `algolia/instantsearch-e2-tests` peerDependencies.
* Add 2 WebdriverIO configurations: one to run tests locally and one to run them on Sauce Labs.
* Add `.env` file to `.gitignore` so you can use it to store Sauce Labs credentials (see https://github.com/algolia/instantsearch-e2e-tests#credentials ).
* Add npm scripts to run End-2-End tests:
  * `test:e2e`: Alias for `test:e2e:local`.
  * `test:e2e:local`: Run the test suite on Chrome browser on your local machine.
  * `test:e2e:saucelabs`: Run the test suite on multiple browsers on the Sauce Labs service.
* Update CircleCI configuration to run the new `test_e2e` task.
* Update examples to use `cp` instead of `rsync` (which is unavailable on CircleCI).


scripts/website/build-examples.sh
```diff
@@ -12,6 +12,6 @@ for example in examples/*; do
     name=$(basename "$example")
     echo "Building $name example..."
 
-    (cd "$example" && yarn && yarn build && rsync -r "dist/" "../../website/examples/$name/")
```
Haroenv: what's this change for?

yannickcr: `rsync` is unavailabled on CircleCI and we need to execute this script on it before running the tests.

wdio.local.conf.js
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable import/no-commonjs, no-console */
```
Haroenv:
```suggestion
/* eslint-disable import/no-commonjs */
```

wdio.saucelabs.conf.js
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable import/no-commonjs, no-console */
```
Haroenv:
```suggestion
/* eslint-disable import/no-commonjs */
```

wdio.local.conf.js
Haroenv: https://github.com/algolia/eslint-config-algolia/issues/112

yannickcr:
> looks like storybook is built for e2e, doesn't seem necessary.

I used the `website:build` script that also build storybook for Netlify deployment. I can update this to only build storybook for Netlify. 

> logs contain "RUNNING", which likely isn't useful on CI?

It depends, since the tests can be long to run it can be good to see when they started in the logs. Also there is no option to suppress this in WebdriverIO (still displayed even with `logLevel` set to `silent`)

> looks like we're only testing on windows? I guess fine for now

No objections to test on multiple platforms but multiplying the browser/platform combinations can seriously lengthen test time.

Right now one test take ~30sec, and we can run 5 parallel tests on SauceLabs

```js
// x = number of tests
// y = number of browser/platform combinations
// the results is the number of minutes to run the e2e test suite
Math.ceil((x * y) / 5) * 0.5
```

francoischalifour:
> looks like we're only testing on windows? I guess fine for now

I think that's the best minimal effort that we can do now because we use both macOS and Linux on the team, but not Windows. If it doesn't add too much overhead, we can add other platforms.

rarkins:
@Haroenv yes, Renovate will update this type of reference. Example: https://github.com/renovate-tests/npm8/pull/2

yannickcr:
Great! Thank you @rarkins :+1:

package.json
```diff
@@ -106,6 +116,7 @@
     "eslint-plugin-prettier": "3.1.0",
     "eslint-plugin-react": "7.14.3",
     "inquirer": "6.5.1",
+    "instantsearch-e2e-tests": "algolia/instantsearch-e2e-tests#v1.1.0",
```
tkrugg: dumb question, since this is a private repository, is this going to break the dev environment of external contributors?

package.json
```diff
@@ -80,6 +83,13 @@
     "@types/storybook__addon-actions": "^3.4.2",
     "@typescript-eslint/eslint-plugin": "1.13.0",
     "@typescript-eslint/parser": "1.13.0",
+    "@wdio/cli": "5.11.14",
```
tkrugg: could theses be subdependencies of instantsearch-e2e-tests#v1.1.0 ?

package.json
yannickcr: The repository is public, the package is just not published on npm. It should work fine for external contributors.

examples/calendar-widget/package.json
```diff
@@ -3,6 +3,6 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "build": "rm -rf ./dist/ ./dist/ && rsync -r . ./dist/ && rsync -r ../../dist/ ./dist/"
+    "build": "rm -rf ./dist/ && mkdir dist && cp -r `ls -A | grep -v 'dist'` ./dist/ && cp ../../dist/* ./dist/"
```
tkrugg: unrelated but yarn workspace could be nice here, right ?
This is something more or less every flavour is handling in a different way

package.json
yannickcr: They could be, but I choose to put them as `peerDependencies` since the `wdio` command will be run from `instantsearch.js`. Also it gives us more control having them as top dependencies.

PR #4036, opened 2019-08-13
feat(clearRefinements): support multiple indices

francoischalifour:
## Description

This implements multi-index context in ClearRefinements leveraging `scopedResults` (#3964).

A ClearRefinements widget now clears the refinements coming from its siblings and its children.

### `init` hook is now a noop

Since we now rely on the `scopedResults` given during `render`, `init` is a noop:

- It considers there are no refinements
- `refine` is a noop at this step
- `createURL` is a noop at this step

This is the reason why some tests have change from `hasRefinements: true` to `hasRefinements: false` at `init`.

### [Attributes were not deduped](https://github.com/algolia/instantsearch.js/blob/2ec09814941b9e1bbec0bdbeea31680f800a0112/src/connectors/clear-refinements/connectClearRefinements.js#L181-L196)

This didn't cause any issues for the usage of the connector but I still dedup them in the new implementation to avoid inconsistent state if we're to build on top on this feature.

In the story, if you select "Samsung", "Apple", and "Cell Phones", the result is:

```
// Old implementation
['brand', brand', 'categories']  // Length: 3 => hasRefinements === true
// New implementation
['brand', 'categories] // Length: 2 => hasRefinements === true
```

This is using `uniq` which is not very costly for the number of attributes that we need to dedup.

## Example

```js
search.addWidgets([
  instantsearch.widgets.clearRefinements({
    container,
  }),

  instantsearch.widgets
    .index({
      indexName: 'instant_search_price_asc',
    })
    .addWidgets([
        container,
        attribute: 'brand',
      }),
    ]),

  instantsearch.widgets
    .index({
      indexName: 'instant_search_rating_asc',
    })
    .addWidgets([
        container,
        attribute: 'categories',
      }),
    ]),
]);
```

## Preview

[See stories →](https://deploy-preview-4036--instantsearchjs.netlify.com/stories/?path=/story/clearrefinements--with-multi-indices)


samouss:
Could you rebase? The history is a bit weird.

francoischalifour:
I think @Haroenv played around for Netlify to report the deployments. It should be fine now.

src/connectors/clear-refinements/connectClearRefinements.js
```diff
@@ -89,70 +91,62 @@ export default function connectClearRefinements(renderFn, unmountFn = noop) {
...
 
         renderFn(
           {
-            hasRefinements,
-            refine: this._refine,
-            createURL: this._createURL,
+            hasRefinements: attributesToClear.some(
+              attributeToClear => attributeToClear.items.length > 0
+            ),
+            refine: () => {
```
samouss: With those changes we provide a new function on each render which was not the case before. It can lead to issues when we the render function have a branch for the first render. This problem arises because the `scopedResults` is not provided at the `init` step. A solution might be to provide it, even empty it doesn't matter.

With an empty state for `scopedResults` we can create the function at the `init` step like before. It will have the same behavior than the proposed implementation the call to `refine`/`createURL` will be a noop.

src/connectors/clear-refinements/connectClearRefinements.js
```diff
@@ -89,70 +91,62 @@ export default function connectClearRefinements(renderFn, unmountFn = noop) {
...
+                      helper: indexHelper,
+                      attributesToClear: items,
+                    })
+                  )
+                  .search();
+              });
+            },
+            createURL: () =>
+              createURL(
+                mergeSearchParameters(
```
samouss: From what I've understood the implementation won't behave like expected. With the usage of `merge` we apply the state of the children on the current index. Only its URL of will be computed but this is not what we want. We want to compute the URL for the current index + siblings + children. With the current implementation of `createURL` I don't think it's possible to implement the expected behaviour. 

The idea would be to provide a list of "patch" to the `createURL` function. Those "patches" describe which indices have to be updated with their next `SearchParameters`. We're then able to compute the `uiState` of each of the index. By default the `createURL` would be called with one patch for the current index, when we have a widget that alter multiple indices we can provide multiple patches. This is the raw idea we can see how it would fit with the current architecture.

src/connectors/clear-refinements/connectClearRefinements.js
francoischalifour: That's right, I fixed that in 07a010772b1c3fc3c6c5e3a715dd703bb0fa5e4f. I went for storing this in the connector state and assigning noops like this implementation.

What do you mean by this though?

> It can lead to issues when we the render function have a branch for the first render.

On a side note, it would be nice to have the same signature for both `init` and `render` at some point.

samouss: > What do you mean by this though?

The ref change between render which means that if we have a render function that branch on `isFirstRender` to register events, the function created at the render step will never be called.

samouss: The fix made inside https://github.com/algolia/instantsearch.js/commit/07a010772b1c3fc3c6c5e3a715dd703bb0fa5e4f does not solve the problem. The reference is different since the function gets re-assigned. You can take a look at [`refinementList`](https://github.com/algolia/instantsearch.js/blob/a23c301f1909d67516664108ff8e015cdea21516/src/connectors/refinement-list/connectRefinementList.js#L285-L290) or [`geoSearch`](https://github.com/algolia/instantsearch.js/blob/a23c301f1909d67516664108ff8e015cdea21516/src/connectors/geo-search/connectGeoSearch.js#L220-L226), both connectors use different strategies to keep the reference.

francoischalifour: I get it now. This is addressed in 72a39a6fb7f293c512ef3974d492e8a4ec514470.

src/connectors/clear-refinements/connectClearRefinements.js
francoischalifour: This will be addressed in another PR because some prior work needs to be done to support `createURL` in a multi-index context.

PR #4037, opened 2019-08-14
feat(toggleRefinement): support new routing system

francoischalifour:
This updates `toggleRefinement` lifecycle methods to support the new routing system.

## `getWidgetState` changes

This method is self-contained. It sets both `disjunctiveFacets` and `disjunctiveFacetsRefinements`.

---


src/connectors/toggleRefinement/connectToggleRefinement.js
```diff
@@ -277,23 +274,38 @@ export default function connectToggleRefinement(renderFn, unmountFn = noop) {
...
+            [attribute]: [],
+          },
+        });
+
         const isRefined = Boolean(uiState.toggle && uiState.toggle[attribute]);
 
         if (isRefined) {
-          if (hasAnOffValue)
-            return searchParameters
+          if (hasAnOffValue) {
```
samouss: We don't have to branch to remove the value we can always call remove even when the value isn't there. The function can be simplified if we move the removal part above the branches. Right now it's a bit tedious to follow.

src/connectors/toggleRefinement/connectToggleRefinement.js
```diff
@@ -277,23 +274,38 @@ export default function connectToggleRefinement(renderFn, unmountFn = noop) {
       },
 
       getWidgetSearchParameters(searchParameters, { uiState }) {
+        const withFacetConfiguration = searchParameters.setQueryParameters({
+          disjunctiveFacets: [attribute],
+          disjunctiveFacetsRefinements: {
```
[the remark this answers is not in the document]
francoischalifour: 5e9460009e7f2177e0c46d9d7b2a353f7dec6240

src/connectors/toggleRefinement/connectToggleRefinement.js
francoischalifour: a0642b475ec76a6e9fb0bcd7a77d44e2194fdef2

I didn't touch this logic though. Hopefully that's what you mean.

src/connectors/toggleRefinement/connectToggleRefinement.js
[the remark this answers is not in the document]
samouss: We still lose the configuration of the facets.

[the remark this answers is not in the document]
francoischalifour: 289457c533051e3389c630a6965fdffa8de801ee

src/connectors/toggleRefinement/connectToggleRefinement.js
```diff
@@ -277,23 +274,28 @@ export default function connectToggleRefinement(renderFn, unmountFn = noop) {
...
+        const state = searchParameters
+          .setQueryParameters({
+            disjunctiveFacetsRefinements: {
+              ...searchParameters.disjunctiveFacetsRefinements,
+              [attribute]: [],
+            },
+          })
+          .addDisjunctiveFacet(attribute)
+          .removeDisjunctiveFacetRefinement(attribute, off)
+          .removeDisjunctiveFacetRefinement(attribute, on);
```
samouss: The `remove` calls are redundant with the `setQueryParameters` call. We can remove one or the other but we have to preserve the empty array when the `uiState` is empty. Here is how I would have done it:

```js
const widget = {
    const withFacetConfiguration = searchParameters
      .clearRefinements(attribute)
      .addDisjunctiveFacet(attribute);

    const isRefined = Boolean(uiState.toggle && uiState.toggle[attribute]);

    // case where it's refined
    if (isRefined) {
      return withFacetConfiguration.addDisjunctiveFacetRefinement(
        attribute,
        on
      );
    }

    // case where it's not refined but with an off value
    if (hasAnOffValue) {
      return withFacetConfiguration.addDisjunctiveFacetRefinement(
        attribute,
        off
      );
    }

    // case where it's not refined withtout an off value
    return withFacetConfiguration.setQueryParameters({
      disjunctiveFacetsRefinements: {
        ...searchParameters.disjunctiveFacetsRefinements,
        [attribute]: []
      }
    });
  }
};
```

src/connectors/toggleRefinement/__tests__/connectToggleRefinement-test.js
```diff
@@ -634,143 +634,271 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/toggle-refi
...
+      expect(actual.disjunctiveFacetsRefinements).toEqual({
+        freeShipping: [],
+      });
+    });
+
+    test('returns the `SearchParameters` without overriding the other refinements', () => {
+      const render = jest.fn();
+      const makeWidget = connectToggleRefinement(render);
+      const helper = jsHelper({}, '', {
+        disjunctiveFacets: ['freeShipping'],
```
samouss:
```suggestion
        disjunctiveFacets: ['onSale', 'freeShipping'],
```

It's better to have a valid `SearchParameters`.

src/connectors/toggleRefinement/connectToggleRefinement.js
```diff
@@ -277,23 +274,57 @@ export default function connectToggleRefinement(renderFn, unmountFn = noop) {
...
+
+        // if (isRefined) {
+        //   return state.addDisjunctiveFacetRefinement(attribute, on);
+        // }
+
+        // if (hasAnOffValue) {
+        //   return state.addDisjunctiveFacetRefinement(attribute, off);
+        // }
+
+        // return state;
```
samouss: Not sure the comment is useful.

PR #4038, opened 2019-08-14
feat(hitsPerPage): support new routing system

francoischalifour:
This updates `hitsPerPage` lifecycle methods to support the new routing system.

## `getWidgetState` changes

This is only a refactor.

---

### DIFF
--- a/src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.js
+++ b/src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.js
@@ -521,23 +521,77 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits-per-pa
     });
   });
 
-  it('does not throw without the unmount function', () => {
-    const rendering = jest.fn();
-    const makeWidget = connectHitsPerPage(rendering);
-    const widget = makeWidget({
-      items: [
-        { value: 3, label: '3 items per page', default: true },
-        { value: 10, label: '10 items per page' },
-      ],
+  describe('getWidgetState', () => {
+    test('returns the `uiState` empty', () => {
+      const render = jest.fn();
+      const makeWidget = connectHitsPerPage(render);
+      const helper = algoliasearchHelper({}, 'indexName');
+      const widget = makeWidget({
+        items: [
+          { value: 3, label: '3 items per page' },
+          { value: 10, label: '10 items per page' },
+        ],
+      });
+
+      const actual = widget.getWidgetState(
+        {},
+        {
+          searchParameters: helper.state,
+        }
+      );
+
+      expect(actual).toEqual({});
     });
-    const helper = algoliasearchHelper({}, '', {});
-    expect(() => widget.dispose({ helper, state: helper.state })).not.toThrow();
+
+    test('returns the `uiState` with a refinement', () => {
+      const render = jest.fn();
+      const makeWidget = connectHitsPerPage(render);
+      const helper = algoliasearchHelper({}, 'indexName', {
+        hitsPerPage: '22',
+      });
+      const widget = makeWidget({
+        items: [
+          { value: 3, label: '3 items per page' },
+          { value: 10, label: '10 items per page' },
+        ],
+      });
+
+      const actual = widget.getWidgetState(
+        {},
+        {
+          searchParameters: helper.state,
+        }
+      );
+
+      expect(actual).toEqual({
+        hitsPerPage: 22,
+      });
+    });
   });
 
-  describe('routing', () => {
-    const getInitializedWidget = () => {
-      const renderFn = jest.fn();
-      const makeWidget = connectHitsPerPage(renderFn);
+  describe('getWidgetSearchParameters', () => {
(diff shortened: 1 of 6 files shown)


Haroenv:
What do you mean by "This method ignores previous UI state values. Otherwise, new values gets skipped."?

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -108,25 +108,25 @@ export default function connectHitsPerPage(renderFn, unmountFn = noop) {
...
-    if (defaultValues.length > 1) {
+    const defaultItems = items.filter(item => item.default);
+    if (defaultItems.length > 1) {
       throw new Error(
         withUsage('More than one default value is specified in `items`.')
       );
     }
 
-    const defaultValue = find(userItems, item => item.default === true);
+    const defaultItem = find(userItems, item => item.default === true);
```
```suggestion
    const defaultItem = defaultItems[0];
```

no big deal though

francoischalifour:
@Haroenv I meant the previous `uiState.hitsPerPage` values. Imagine `getWidgetState` returns `{ hitsPerPage: 10 }`. Then the user refines the state to `{ hitsPerPage: 20 }`. The next `getWidgetState` will still return `{ hitsPerPage: 10 }` if we keep the previous value.

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -229,19 +226,12 @@ You may want to add another entry to the \`items\` option with this value.`
...
-            'hitsPerPage',
-            defaultValue.value
-          );
-        }
-        return searchParameters.setQueryParameter('hitsPerPage', undefined);
+        const hitsPerPage =
+          uiState.hitsPerPage || (defaultItem && defaultItem.value);
+
+        return searchParameters.setQueryParameters({
+          hitsPerPage,
```
samouss: IMO the widget should be deterministic. Here it's not the case, when we don't provide a default value we don't know how many hits per page we'll have. It depends on the context where the widget is added. I think that we should fallback on the first item when the default value is not provided or enforce the default value (this is what we do inside React InstantSearch).

francoischalifour: That's a change in the way the widget acts.

I'll be in favor of having another option that determines the default value as well. Should we go for `defaultValue`?

See [React InstantSearch reference to `defaultRefinement`](https://www.algolia.com/doc/api-reference/widgets/hits-per-page/react/#widget-param-defaultrefinement)

samouss: We already have a mechanism for the default value. We don't have to introduce another option. We have to either enforce it or find a fallback. If you really want to introduce another option we can but I don't see the value.

francoischalifour: The value would be a less confusing API. Right now it's easy to miss or to add several `default` values. Then we get into a tricky situation of making a choice for the user to select the right one. If we add an option to select the default one, it's more straightforward for users.

I think this should be part of another PR. Let's not diverge too much here.

src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.js
```diff
@@ -521,129 +521,131 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits-per-pa
...
 
-  describe('routing', () => {
-    const getInitializedWidget = () => {
-      const renderFn = jest.fn();
-      const makeWidget = connectHitsPerPage(renderFn);
+    test('returns the `uiState` with a refinement', () => {
+      const render = jest.fn();
+      const makeWidget = connectHitsPerPage(render);
+      const helper = algoliasearchHelper({}, 'indexName', {
+        hitsPerPage: '22',
```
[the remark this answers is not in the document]
francoischalifour: b8ed7ab14d5a37373f3e94abe4a359da66948bfb

src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.js
```diff
@@ -521,129 +521,131 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits-per-pa
...
 
-    describe('getWidgetState', () => {
-      test('should give back the object unmodified if there are no refinements', () => {
-        const [widget, helper] = getInitializedWidget();
-        const uiStateBefore = {};
-        const uiStateAfter = widget.getWidgetState(uiStateBefore, {
-          searchParameters: helper.state,
-          helper,
-        });
+      expect(actual.hitsPerPage).toEqual(undefined);
```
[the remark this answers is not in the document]
francoischalifour: 08b1ac7b634347cb0f480c58cbd26eb6f2ae859c

src/connectors/hits-per-page/connectHitsPerPage.js
francoischalifour: See #4069.

samouss:
Would be better to first merge #4069, then rebase the branch on `next`.

PR #4039, opened 2019-08-14
feat(range): support new routing system

francoischalifour:
This updates `range` lifecycle methods to support the new routing system.

## `getWidgetState` changes

This method is self-contained. It sets both `disjunctiveFacets` and `numericRefinements`. It doesn't overwrite other `numericRefinements` but only sets the one coming from `widgetParams.attribute` to an empty object.

---


src/connectors/range/connectRange.js
```diff
@@ -305,43 +300,39 @@ export default function connectRange(renderFn, unmountFn = noop) {
...
 
-        const {
-          '>=': previousMin = [NaN],
-          '<=': previousMax = [NaN],
-        } = searchParameters.getNumericRefinements(attribute);
-        let clearedParams = searchParameters.clearRefinements(attribute);
         const [lowerBound, upperBound] = value.split(':').map(parseFloat);
 
-        if (
-          previousMin.includes(lowerBound) &&
```
Haroenv: makes sense to remove this, and rely on the helper to avoid duplicates. does this have a test?

src/connectors/range/__tests__/connectRange-test.js
```diff
@@ -1283,3 +1036,287 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/range-input
...
+    const actual = widget.getWidgetState(
+      {},
+      {
+        searchParameters: helper.state,
+      }
+    );
+
+    expect(actual).toEqual({
+      range: {
+        price: '100:1000',
```
samouss: We could add more tests to cover all the possible usage: `x:`, `:x`, `x:x`.

src/connectors/range/connectRange.js
```diff
@@ -305,43 +300,39 @@ export default function connectRange(renderFn, unmountFn = noop) {
       },
 
       getWidgetSearchParameters(searchParameters, { uiState }) {
-        const value = uiState && uiState.range && uiState.range[attribute];
+        let withFacetConfiguration = searchParameters.setQueryParameters({
+          disjunctiveFacets: [attribute],
```
samouss: We'll lose the previous configuration sets by other widgets (see [`refinementList`](https://github.com/algolia/instantsearch.js/blob/f69ea2614a9d92fb48ba5535fae11537beb7818d/src/connectors/refinement-list/connectRefinementList.js#L432-L475)).

src/connectors/range/connectRange.js
```diff
@@ -305,43 +300,39 @@ export default function connectRange(renderFn, unmountFn = noop) {
...
-        if (
-          previousMin.includes(lowerBound) &&
-          previousMax.includes(upperBound)
-        ) {
-          return searchParameters;
-        }
-
         if (isFiniteNumber(lowerBound)) {
-          clearedParams = clearedParams.addNumericRefinement(
+          withFacetConfiguration = withFacetConfiguration.addNumericRefinement(
```
samouss: I would have use a variable for each block, it's easier to follow IMO: `withLowerBound`, `withUpperBrand`.

src/connectors/range/connectRange.js
francoischalifour: Not sure to follow. `disjunctiveFacets` are merged with the `mergeSearchParameters` method, only facet refinements are not.

src/connectors/range/connectRange.js
francoischalifour: Here we need to mutate them so I'm not sure it makes them more readable. Besides, it creates a naming burden.

src/connectors/range/__tests__/connectRange-test.js
francoischalifour: 0e606b0ac0fb28e8b280981786e89326d4ba2953

src/connectors/range/connectRange.js
samouss: The function `mergeSearchParameters` won't be used to at this level (see #4059). We leverage the `SearchParameters` structure to do that for us. The merge function is only used once we have the "final" `SearchParameters` for each of the indices (see [`index`](https://github.com/algolia/instantsearch.js/blob/f69ea2614a9d92fb48ba5535fae11537beb7818d/src/widgets/index/index.ts#L285)).

francoischalifour: 021a62545ef80d606f56605ef4520200794a6fa3

PR #4040, opened 2019-08-16
feat(infiniteHits): support new routing system

francoischalifour:
This updates `infiniteHits` lifecycle methods to support the new routing system.

I'm not aware of how the new routing system will exactly work, so this behavior needs to be reviewed.

## `getWidgetState` changes

This method now also supports highlighting tags.

---


src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -262,30 +262,40 @@ const connectInfiniteHits: InfiniteHitsConnector = (
...
+          const uiPage = uiState.page;
 
-        if (uiPage) {
-          return searchParameters.setQueryParameter('page', uiPage - 1);
+          if (uiPage) {
+            // The page in the search parameters is decremented by one
+            // to get to the actual parameter value from the UI state.
+            state = searchParameters.setQueryParameter('page', uiPage - 1);
+          } else {
+            state = searchParameters.setQueryParameter('page', undefined);
```
francoischalifour: Is this branch even necessary? I left it here because I don't know why it was here before. When do we need to reset the `page` search parameter? (I suppose when option changes during the widget lifecycle but we don't support such a feature)

Haroenv: I think this turns page 0 into undefined

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -842,220 +843,146 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
-          createInitOptions({
-            state: helper.state,
-            helper,
-          })
-        );
-
-        const uiStateBefore = { page: 1 };
-        const uiStateAfter = widget.getWidgetState!(uiStateBefore, {
+  describe('getWidgetState', () => {
+    test('returns the `uiState` empty without `showPrevious` option', () => {
```
samouss: This test always pass even with `showPrevious: true` since we don't provide a page.

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -842,220 +843,146 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
+      );
 
-        const { showMore } = renderFn.mock.calls[0][0];
+      expect(actual).toEqual({
+        page: 1,
+      });
+    });
 
-        showMore();
+    test('returns the `uiState` containing a page number incremented by one with `showPrevious` option and `page` search parameter', () => {
```
samouss:
```suggestion
    test('returns the `uiState` containing a page number incremented by one with `page` search parameter', () => {
```

We don't have to mention it on every tests, it's our default case.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -262,30 +262,40 @@ const connectInfiniteHits: InfiniteHitsConnector = (
...
-        const page = searchParameters.page || 0;
-
-        if (!hasShowPrevious || page === 0 || page + 1 === uiState.page) {
+        if (!hasShowPrevious) {
           return uiState;
         }
 
+        // The page in the UI state is incremented by one
+        // to expose the user value (not `0`).
+        const page = (searchParameters.page || 0) + 1;
```
samouss: We're not suppose to sync the zero value (see [`pagination`](https://github.com/algolia/instantsearch.js/blob/f69ea2614a9d92fb48ba5535fae11537beb7818d/src/connectors/pagination/connectPagination.js#L170-L179)).

src/connectors/infinite-hits/connectInfiniteHits.ts
samouss: It's not necessary since we have to provide a default value, otherwise the widget is controlled by its parent (see [`pagination`](https://github.com/algolia/instantsearch.js/blob/f69ea2614a9d92fb48ba5535fae11537beb7818d/src/connectors/pagination/connectPagination.js#L182-L185)).

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -262,30 +262,40 @@ const connectInfiniteHits: InfiniteHitsConnector = (
...
           ...uiState,
-          page: page + 1,
+          page,
         };
       },
 
       getWidgetSearchParameters(searchParameters, { uiState }) {
-        if (!hasShowPrevious) {
-          return searchParameters;
+        let state = searchParameters;
```
samouss: I would have kept the same pattern used in `dispose` with named variable vs `let`.

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
francoischalifour: The test made sense before https://github.com/algolia/instantsearch.js/pull/4040#discussion_r316740878, I didn't have the context.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -264,28 +264,36 @@ const connectInfiniteHits: InfiniteHitsConnector = (
...
-          return searchParameters.setQueryParameter('page', uiPage - 1);
+          if (uiPage) {
+            // The page in the search parameters is decremented by one
+            // to get to the actual parameter value from the UI state.
+            state = searchParameters.setQueryParameter('page', uiPage - 1);
+          }
         }
 
-        return searchParameters.setQueryParameter('page', undefined);
+        return state;
```
samouss: The widget have to enforce a default value like pagination does otherwise its value is controlled by its parent.

francoischalifour: So you mean setting an arbitrary page of `0`?

samouss: Yes.

francoischalifour: f390b1c737e93de0bf6b3aceea1ca4576a1493cd

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -842,220 +843,188 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
-          }
-        );
-        // Applying an empty state, should force back to page 0
-        expect(searchParametersAfter).toMatchSnapshot();
-        expect(searchParametersAfter.page).toBeUndefined();
+      expect(actual.highlightPreTag).toBeUndefined();
+      expect(actual.highlightPostTag).toBeUndefined();
+    });
+
+    test('return Search Parameters with empty page when `showPrevious` without `page` in the UI state', () => {
```
samouss: The name of the test does not match the expectation (the same problem is present on the next cases). The page is not empty but `0`.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -264,28 +264,38 @@ const connectInfiniteHits: InfiniteHitsConnector = (
...
-        const uiPage = uiState.page;
+        if (hasShowPrevious) {
+          const uiPage = uiState.page;
 
-        if (uiPage) {
-          return searchParameters.setQueryParameter('page', uiPage - 1);
+          if (uiPage) {
+            // The page in the search parameters is decremented by one
+            // to get to the actual parameter value from the UI state.
+            return widgetSearchParameters.setQueryParameter('page', uiPage - 1);
```
samouss: We don't check for negative values etc.. but I guess it's fine for now. We never have any issues about that.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -264,28 +264,38 @@ const connectInfiniteHits: InfiniteHitsConnector = (
...
+          );
         }
 
-        const uiPage = uiState.page;
+        if (hasShowPrevious) {
+          const uiPage = uiState.page;
 
-        if (uiPage) {
-          return searchParameters.setQueryParameter('page', uiPage - 1);
+          if (uiPage) {
```
samouss: We could avoid the second nesting level with the condition merged inside the if above.

PR #4041, opened 2019-08-16
docs(infiniteHits): update documentation

francoischalifour:
This removes the outdated JSDoc and leverages TypeScript's documentation for the `infiniteHits` stack.

### DIFF
--- a/src/widgets/infinite-hits/infinite-hits.tsx
+++ b/src/widgets/infinite-hits/infinite-hits.tsx
@@ -46,9 +46,24 @@ export type InfiniteHitsTemplates = {
 };
 
 export type InfiniteHitsRendererWidgetParams = {
+  /**
+   * Escapes HTML entities from hits string values.
+   *
+   * @default `true`
+   */
   escapeHTML: boolean;
-  transformItems: (items: any[]) => any[];
+  /**
+   * Enable the button to load previous results.
+   *
+   * @default `false`
+   */
   showPrevious: boolean;
+  /**
+   * Receives the items, and is called before displaying them.
+   * Should return a new array with the same shape as the original array.
+   * Useful for mapping over the items to transform, and remove or reorder them.
+   */
+  transformItems: (items: any[]) => any[];
 };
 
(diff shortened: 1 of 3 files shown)


src/widgets/infinite-hits/infinite-hits.tsx
```diff
@@ -46,9 +46,24 @@ export type InfiniteHitsTemplates = {
...
+   *
+   * @default `false`
+   */
   showPrevious: boolean;
+  /**
+   * Receives the items, and is called before displaying them.
+   * Should return a new array with the same shape as the original array.
+   * Useful for mapping over the items to transform, and remove or reorder them.
+   */
+  transformItems: (items: any[]) => any[];
```
eunjae-lee: Can we see this kind of comments on editor like.. on tooltip when hovering? or they're just commented there?

francoischalifour: They work when the import is coming from the TypeScript file itself (`src/widgets/infinite-hits/infinite-hits.ts`). That means that you can read them from the test files and from the stories if using the direct file (not `instantsearch.widgets.infiniteHits`).

It doesn't work when loaded from other files because we don't export the types yet.

<details>
  <summary>From a test</summary>

image

</details>

<details>
  <summary>From a story (import modified locally</summary>

image

</details>

eunjae-lee: Thanks for the screen shot!
I've checked and saw it working on `infinite-hits-test.ts`.
Something I don't understand is,

image

This is from `infinite-hits.tsx` and why doesn't it work? It looks like it should work.

francoischalifour: I think this is a bug in the TypeScript VS Code extension because the definition is showing up for `cssClasses`, which is using a renaming with the `actualName: otherName` syntax.

image

eunjae-lee: You're right. I've tested a few times and it only works for the renamed keys.

PR #4054, opened 2019-08-21
feat(connectMenu): update getWidgetSearchParameters

samouss:
This PR updates the lifecycle `getWidgetSearchParameters` in `connectMenu` to add the configuration of facets and default value. We don't apply the previous value on purpose, the `uiState` drives completely the `SearchParameters`.

### DIFF
--- a/src/connectors/menu/__tests__/connectMenu-test.js
+++ b/src/connectors/menu/__tests__/connectMenu-test.js
@@ -709,51 +709,223 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/menu/js/#co
   });
 
   describe('getWidgetSearchParameters', () => {
-    test('should return the same SP if there are no refinements in the UI state', () => {
-      const [widget, helper] = getInitializedWidget();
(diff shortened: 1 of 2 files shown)


src/connectors/menu/__tests__/connectMenu-test.js
```diff
@@ -617,144 +643,311 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/menu/js/#co
...
-        const uiStateBefore = {};
-        const uiStateAfter = widget.getWidgetState(uiStateBefore, {
-          searchParameters: helper.state,
-          helper,
-        });
-
-        expect(uiStateAfter).toEqual({
-          menu: {
-            category: 'pants',
+    test('returns the `uiState` without namespace overridden', () => {
```
Haroenv: not sure what this test name means

samouss: The sentence is now used in a couple of places. We can update them all at once into a dedicated PR.

PR #4055, opened 2019-08-21
feat(connectVoiceSearch): update getWidgetSearchParameters

samouss:
This PR updates the lifecycle `getWidgetSearchParameters` in `connectVoiceSearch` to add the  default value. We don't apply the previous value on purpose, the `uiState` drives completely the `SearchParameters`.

### DIFF
--- a/src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
+++ b/src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
@@ -272,40 +272,48 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-searc
   });
 
   describe('getWidgetSearchParameters', () => {
-    it('should return the same SearchParameters if no value is in the UI state', () => {
+    test('returns the `SearchParameters` with the value from `uiState`', () => {
       const { widget, helper } = getInitializedWidget();
-      const uiState = {};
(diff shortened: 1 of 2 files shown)


src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -232,80 +232,80 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-searc
...
-      const { widget, helper, refine } = getInitializedWidget();
-      const uiStateBefore = { foo: 'bar' };
-      refine('');
-      const uiStateAfter = widget.getWidgetState(uiStateBefore, {
-        searchParameters: helper.state,
-      });
-      expect(uiStateAfter).toBe(uiStateBefore);
-    });
+    test('returns the `uiState` empty', () => {
+      const { widget, helper } = getInitializedWidget();
```
Haroenv: for other reviewers, I checked and `getInitializedWidget` doesn't call `getConfiguration`

samouss:
@haroen not sure I’ve understood the comment.

Haroenv:
I misread this: ` (uiState && uiState.query === query)`, removing it is indeed identical, with the only difference that the object gets recreated.

PR #4056, opened 2019-08-21
 chore(configure): participate in routing

Haroenv:
IFW-944

this adds simple routing for configure, where it will override the state on getWidgetState, and merge current state with configuration on getWidgetSearchParameters

This also changes the behaviour of the simple state mapping to explicitly filter out configure. While we don't expect someone to explicitly add a configure to the route, we can't be sure, and people need to explicitly override the statemapping to identity functions if they want to keep it. We can document that in the routing API documentation once we have it.


src/lib/stateMappings/__tests__/simple-test.ts
```diff
@@ -0,0 +1,122 @@
...
+        } as SimpleRouteState)
+      ).toEqual({
+        query: 'zamboni',
+        refinementList: {
+          color: ['red'],
+        },
+      });
+    });
+
+    // @TODO: do we want to change this?
```
Haroenv: I think we don't want to change this, because we shouldn't filter out custom widgets

src/lib/stateMappings/simple.ts
```diff
@@ -1,15 +1,19 @@
...
+  return {
+    stateToRoute(uiState: UiState): SimpleRouteState {
+      const { configure, ...trackedRouteState } = uiState;
+      return trackedRouteState;
+    },
 
-export default function(): SimpleUIStateMapping {
-  return new SimpleUIStateMapping();
+    routeToState(routeState: SimpleRouteState): UiState {
+      // users can still make a url containing the key "configure", we don't want
```
samouss: Not sure the comment is useful.

src/lib/stateMappings/simple.ts
```diff
@@ -1,15 +1,19 @@
...
+export default function SimpleStateMapping(): StateMapping<SimpleRouteState> {
+  return {
+    stateToRoute(uiState: UiState): SimpleRouteState {
+      const { configure, ...trackedRouteState } = uiState;
+      return trackedRouteState;
+    },
 
-export default function(): SimpleUIStateMapping {
-  return new SimpleUIStateMapping();
+    routeToState(routeState: SimpleRouteState): UiState {
```
samouss: The type is incorrect then if the users are still able to provide it.

src/connectors/configure/connectConfigure.ts
```diff
@@ -140,6 +140,21 @@ const connectConfigure: ConfigureConnector = (
...
+          ...uiState.configure,
+          ...widgetParams.searchParameters,
+        });
+      },
+
+      getWidgetState(uiState) {
+        return {
+          ...uiState,
+          // widgetParams gets updated by refine
+          configure: widgetParams.searchParameters,
```
samouss: We should be able to have multiple widgets.

src/lib/stateMappings/__tests__/simple-test.ts
```diff
@@ -0,0 +1,122 @@
...
+    it('passes non-UiState through', () => {
```
samouss: At the moment I guess it's fine, the keys are ignored by the widgets anyway.

src/lib/stateMappings/__tests__/simple-test.ts
samouss: We have to keep them indeed, both tests aren't that useful IMO.

Haroenv: @francoischalifour insisted there should be tests for stateMapping

src/lib/stateMappings/simple.ts
```diff
@@ -1,17 +1,16 @@
 import { UiState, StateMapping } from '../../types';
 
-export type SimpleRouteState = Omit<UiState, 'configure'>;
-
-export default function SimpleStateMapping(): StateMapping<SimpleRouteState> {
+// technically a URL could contain any key, since users provide it,
+// which is why the input to this function is UiState, not something
+// which excludes "configure" as this function does.
+export default function SimpleStateMapping(): StateMapping<UiState> {
```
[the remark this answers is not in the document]
Haroenv:
```suggestion
export default function simpleStateMapping(): StateMapping<UiState> {
```

src/connectors/configure/connectConfigure.ts
```diff
@@ -140,6 +140,24 @@ const connectConfigure: ConfigureConnector = (
...
+          ...widgetParams.searchParameters,
+        });
+      },
+
+      getWidgetState(uiState) {
+        return {
+          ...uiState,
+          configure: {
+            ...uiState.configure,
+            // widgetParams gets updated by refine
```
samouss:
```suggestion

```

src/lib/stateMappings/simple.ts
```diff
@@ -1,15 +1,18 @@
...
-  public stateToRoute(uiState: UiState): RouteState {
-    return uiState;
-  }
+// technically a URL could contain any key, since users provide it,
+// which is why the input to this function is UiState, not something
+// which excludes "configure" as this function does.
+export default function simpleStateMapping(): StateMapping<UiState> {
+  return {
+    stateToRoute(uiState) {
+      const { configure, ...trackedRouteState } = uiState;
```
samouss: We have to keep in mind that we have to update the logic once we have the new format of the `uiState`.

Haroenv: the type UiState will change (https://github.com/algolia/instantsearch.js/pull/4071, but different) so this should fail because of that :)

PR #4059, opened 2019-08-22
feat(index): use uiState driven SearchParameters

samouss:
**Summary**

This PR replaces the usage of `getConfiguration` for `getWidgetSearchParameters`. The widgets are now driven by the `uiState`. Note that widgets that don't implement the lifecycle can still write on `SearchParameters`, but if a widget that implements the lifecycle use the same parameter it will be always overridden.

The lifecycle is now called on `init`, `add` (when the instance is started), `remove` (when the instance is started). On `init` we currently always provide an empty `uiState` (but it will change soon). On `add`/`remove` we provide the local `uiState` of the index to drive the `SearchParameters`.

### DIFF
--- a/src/widgets/index/index.ts
+++ b/src/widgets/index/index.ts
@@ -2,6 +2,7 @@ import algoliasearchHelper, {
   AlgoliaSearchHelper as Helper,
   DerivedHelper,
   PlainSearchParameters,
+  SearchParameters,
   SearchResults,
 } from 'algoliasearch-helper';
 import { Client } from 'algoliasearch';
@@ -12,6 +13,7 @@ import {
   InitOptions,
   RenderOptions,
   WidgetStateOptions,
+  WidgetSearchParametersOptions,
   ScopedResult,
 } from '../../types';
 import {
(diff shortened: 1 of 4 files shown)


src/types/widget.ts
```diff
@@ -50,6 +51,7 @@ export interface WidgetSearchParametersOptions {
 
 export type UiState = {
   query?: string;
+  configure?: PlainSearchParameters;
```
Haroenv: this is already added in #4056

src/widgets/index/index.ts
```diff
@@ -12,12 +13,12 @@ import {
   InitOptions,
   RenderOptions,
   WidgetStateOptions,
+  WidgetSearchParametersOptions,
   ScopedResult,
 } from '../../types';
 import {
   createDocumentationMessageGenerator,
   resolveSearchParameters,
-  enhanceConfiguration,
```
Haroenv: it can be removed now, right?

samouss: Yes but want to do it inside a dedicated PR (only RED!!!!).

src/types/widget.ts
samouss: First one that gets merged win!

PR #4069, opened 2019-08-26
feat(hitsPerPage): enforce default value

francoischalifour:
In InstantSearch.js, we don't enforce a default value in the [`hitsPerPage`](https://www.algolia.com/doc/api-reference/widgets/hits-per-page/js/#widget-param-items) widget. This means that we pick the first in the list when no `default` is specified.

The problem with not always having a default value is that it's not deterministic with multi-indexing: `getWidgetSearchParameters`' `hitsPerPage` value can be inherited from the parent index whereas we want it to be self contained (see #4038).

This PR enforces a default value (and a single one).

Long term, we can think moving toward an API similar to the [React version](https://www.algolia.com/doc/api-reference/widgets/hits-per-page/react/#widget-param-defaultrefinement), where another option (e.g. `defaultValue`) picks the right item in the list. The benefit of such an API is that it's less error prone for users and better type-checkable.

### DIFF
--- a/src/connectors/hits-per-page/connectHitsPerPage.js
+++ b/src/connectors/hits-per-page/connectHitsPerPage.js
@@ -114,7 +114,9 @@ export default function connectHitsPerPage(renderFn, unmountFn = noop) {
       throw new Error(
         withUsage(`A default value must be specified in \`items\`.`)
       );
-    } else if (defaultItemsCount > 1) {
+    }
+
+    if (defaultItemsCount > 1) {
       throw new Error(
         withUsage('More than one default value is specified in `items`.')
       );
--- a/src/connectors/hits-per-page/connectHitsPerPage.js
+++ b/src/connectors/hits-per-page/connectHitsPerPage.js
@@ -110,7 +110,7 @@ export default function connectHitsPerPage(renderFn, unmountFn = noop) {
     const defaultItems = items.filter(item => item.default === true);
     const defaultItemsCount = defaultItems.length;
 
-    if (items.length && defaultItemsCount === 0) {
+    if (defaultItemsCount === 0) {
       throw new Error(
         withUsage(`A default value must be specified in \`items\`.`)
       );
--- a/src/connectors/hits-per-page/connectHitsPerPage.js
+++ b/src/connectors/hits-per-page/connectHitsPerPage.js
@@ -108,8 +108,13 @@ export default function connectHitsPerPage(renderFn, unmountFn = noop) {
     }
 
     const defaultItems = items.filter(item => item.default === true);
+    const defaultItemsCount = defaultItems.length;
 
-    if (defaultItems.length > 1) {
+    if (items.length && defaultItemsCount === 0) {
+      throw new Error(
+        withUsage(`A default value must be specified in \`items\`.`)
+      );
+    } else if (defaultItemsCount > 1) {
       throw new Error(
         withUsage('More than one default value is specified in `items`.')
       );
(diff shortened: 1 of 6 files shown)


src/widgets/hits-per-page/__tests__/hits-per-page-test.js
```diff
@@ -61,14 +61,7 @@ describe('hitsPerPage()', () => {
...
 
-  it('does not configure the default hits per page if not specified', () => {
-    expect(typeof widget.getConfiguration).toEqual('function');
-    expect(widget.getConfiguration(new SearchParameters({}))).toEqual(
-      new SearchParameters({})
-    );
-  });
-
-  it('does configures the default hits per page if specified', () => {
+  it('does configures the default hits per page', () => {
```
Haroenv:
```suggestion
  it('configures the default hits per page', () => {
```

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -108,8 +108,13 @@ export default function connectHitsPerPage(renderFn, unmountFn = noop) {
     }
 
     const defaultItems = items.filter(item => item.default === true);
+    const defaultItemsCount = defaultItems.length;
 
-    if (defaultItems.length > 1) {
+    if (items.length && defaultItemsCount === 0) {
```
Haroenv: why is there a  check for items.length here?

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -108,8 +108,13 @@ export default function connectHitsPerPage(renderFn, unmountFn = noop) {
...
+      throw new Error(
+        withUsage(`A default value must be specified in \`items\`.`)
+      );
+    } else if (defaultItemsCount > 1) {
```
```suggestion
    }
    if (defaultItemsCount > 1) {
```

throwing closes a branch

src/connectors/hits-per-page/connectHitsPerPage.js
francoischalifour: Otherwise an empty array would error:

```js
hitsPerPage({
  items: [],
});
```

Haroenv: is this a valid use case though?

francoischalifour: It works in React InstantSearch. I'm not sure it makes too much sense though.

Haroenv: Personally I'm okay with throwing an error, since I don't see what use it can have. Even if you'd set the defaultRefinement, you still need items to apply it

francoischalifour: Addressed in 3f744e2d4dfabd1b4c3ce61cc76c8fadae53ec7e.

src/connectors/hits-per-page/connectHitsPerPage.js
```diff
@@ -108,8 +108,15 @@ export default function connectHitsPerPage(renderFn, unmountFn = noop) {
     }
 
     const defaultItems = items.filter(item => item.default === true);
+    const defaultItemsCount = defaultItems.length;
```
samouss: Note sure the variable is really useful here.

stories/hits-per-page.stories.js
```diff
@@ -9,7 +9,7 @@ storiesOf('HitsPerPage', module)
         instantsearch.widgets.hitsPerPage({
           container,
           items: [
-            { value: 3, label: '3 per page' },
+            { value: 3, label: '3 per page', default: true },
```
samouss: We can remove the example "with default hitPerPage to 5" it's redundant with the default one.

francoischalifour: Not really, it shows that the second option in the select can be the one selected by default.

samouss: Yes... but I mean it's expected I think.

PR #4070, opened 2019-08-26
feat(index): accept indexId

Haroenv:
The indexid is used in the already existing `getIndexId` function, so all that was needed to be done there was to add a new test to ensure the usage is correct.

There can still be conflicts within a scope, but that would be if an indexId is shared

IFW-930

I didn't make a story highlighting this use, since it only has an effect in the route state, which isn't yet being listened to for nested indices, so is hard to see in effect for now

### DIFF
--- a/src/widgets/index/__tests__/index-test.ts
+++ b/src/widgets/index/__tests__/index-test.ts
@@ -1335,6 +1335,28 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
         });
       });
 
+      it('indexId is used for scope key', () => {
+        const instance = index({ indexName: 'indexName', indexId: 'indexId' });
+        const widgets = [createSearchBox(), createPagination()];
+
+        instance.addWidgets(widgets);
+
+        instance.init(createInitOptions());
+
+        // Simulate a state change
+        instance
+          .getHelper()!
+          .setQueryParameter('query', 'Apple')
+          .setQueryParameter('page', 5);
+
+        expect(instance.getWidgetState({})).toEqual({
+          indexId: {
+            query: 'Apple',
+            page: 5,
+          },
+        });
+      });
+
       it('does not update the local `uiState` on state changes in `init`', () => {
         const instance = index({ indexName: 'indexName' });
         const widgets = [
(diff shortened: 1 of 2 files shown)


src/widgets/index/__tests__/index-test.ts
```diff
@@ -1557,20 +1579,20 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
       const level0 = index({ indexName: 'level0IndexName' });
       const level1 = index({ indexName: 'level1IndexName' });
       const level2 = index({ indexName: 'level2IndexName' });
-      const level21 = index({ indexName: 'level21IndeName' });
+      const level21 = index({ indexName: 'level21IndexName' });
```
Haroenv: fixed some typos at the same time as well, inconsequential for the main change

src/widgets/index/__tests__/index-test.ts
```diff
@@ -1335,6 +1335,28 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
         });
       });
 
+      it('indexId is used for scope key', () => {
```
```suggestion
      it('uses `indexId` for scope key', () => {
```

src/widgets/index/index.ts
```diff
@@ -106,7 +107,7 @@ function resolveScopedResultsFromIndex(widget: Index): ScopedResult[] {
 }
 
 const index = (props: IndexProps): Index => {
-  const { indexName = null } = props || {};
+  const { indexName = null, indexId = indexName } = props || {};
```
francoischalifour: These types are inferred as `any`. A way to get it typed correctly would be like so:

```ts
// ...

const { indexName = null } = props || ({} as IndexProps);

if (indexName === null) {
  throw new Error(withUsage('The `indexName` option is required.'));
}

// At this point, `indexId` and `indexName` are strings
const { indexId = indexName } = props || ({} as IndexProps);
```

src/widgets/index/index.ts
```diff
@@ -105,8 +106,12 @@ function resolveScopedResultsFromIndex(widget: Index): ScopedResult[] {
   return resolveScopedResultsFromWidgets(widgetSiblings);
 }
 
-const index = (props: IndexProps): Index => {
-  const { indexName = null } = props || {};
+const index = (props?: IndexProps): Index => {
```
samouss: The switch to optional is not required we lose the type-safety. An alternative is to check the value even though the compiler ensure the value is always available. At least TypeScript users (mainly us) don't lose type-safety.

```ts
const index = (props: IndexProps): Index => {
  if (!props) {
    throw new Error(withUsage('Cannot be called without args.'));
  }

  const { indexName = null, indexId = indexName } = props;

  if (indexName === null) {
    throw new Error(withUsage('The `indexName` option is required.'));
  }

  if (indexId === null) {
    // Should not happen but at least the compiler is happy.
    throw new Error(withUsage("The `indexId` option can't be `null`."));
  }
   
  // ...
```

Haroenv: will remove the ?

Haroenv: db4f325c7

PR #4071, opened 2019-08-26
feat(index): add child indices to UiState correctly

Haroenv:
Before this change, UiState was merging the indexIds directly with the other state, meaning it's very hard to do changes like IFW-933, which filter out the current UiState, as well as making routing later more easy to write (IFW-941 specifically)

Merging this will prompt an update to #4070 to have correct UiState in the tests

### DIFF
--- a/src/types/widget.ts
+++ b/src/types/widget.ts
@@ -101,6 +101,9 @@ export type UiState = {
   sortBy?: string;
   page?: number;
   hitsPerPage?: number;
+  indices?: {
+    [indexId: string]: UiState;
+  };
 };
 
 /**
--- a/src/widgets/index/__tests__/index-test.ts
+++ b/src/widgets/index/__tests__/index-test.ts
@@ -1007,21 +1007,23 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
           .setQueryParameter('page', 4);
 
         expect(level0.getWidgetState({})).toEqual({
-          level0IndexName: {
-            query: 'Apple',
-            page: 1,
-          },
-          level1IndexName: {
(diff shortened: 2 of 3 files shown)


francoischalifour:
One of the most annoying thing in the React InstantSearch codebase is to conditionally support single and multi index states. I thought we were going to a different structure in InstantSearch.js that will make development easier: always have the index ID as key and remove all top-level states.

```ts
interface IndexUiState {
  query?: string;
  // ...
}

interface UiState {
  [indexId: string]: IndexUiState;
}
```

An example would be:

```ts
const state: UiState = {
  movies: {
    query: 'hammer',
  },
  songs: {
    query: 'burning',
  },
};
```

The role of state mapping would be to transform any single-index mapping to a `IndexUiState` to keep routing backward compatible.

Haroenv:
made some changes here, but the story doesn't seem complete: https://github.com/algolia/instantsearch.js/commit/b0c1acddc97cf5ea24baf407e6ba4241c8c39f28

samouss:
> Before this change, UiState was merging the indexIds directly with the other state,

Not sure to understand the meaning of the comment. Could you explain? The current format of `uiState` is already the one described by @francoischalifour. I don't think we have to update its structure. The types inside the codebase are not up-to-date though, this is the idea:

```ts
type UiState = {
  [indexId: string]: IndexUiState;
};

const movies: IndexUiState = {
  query: "hammer"
};

const songs: IndexUiState = {
  query: "burning"
};

const uiState: UiState = {
  songs,
  movies
};
```

Haroenv:
will do this later when other changes are integrated, now it requires too many changes

PR #4074, opened 2019-08-27
feat: introduce initialUiState option

samouss:
**Summary**

This PR introduces the `initialUiState` option at the `InstantSearch` level. The option allows users to provide an `uiState` used for the initial request. Only for the **initial request**. I've chosen to name the option `initialXXX` rather than `uiState` to have a clear intent. At some point, we might want to introduce the concept of controlled state (like React InstantSearch) that could use the option `uiState`.

At the moment the type definitions are incorrect. We have only one type that defines the `uiState` but we have to define two different types. The first is one is the "global" `uiState` and the second one is the "local" to each index. The "global" `uiState` is built on multiple "local" `uiState`. We'll update the definition in a separate PR.

**Usage**

```js
  // ...
  initialUiState: {
    instant_search: {
      query: "Apple"
    },
    instant_search_price_asc: {
      refinementList: {
        brand: ["Apple"]
      }
    }
  }
});
```

Checkout the example on [Storybook](https://deploy-preview-4074--instantsearchjs.netlify.com/stories/?path=/story/instantsearch--with-initialuistate).

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -392,6 +392,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     this.mainIndex.init({
       instantSearchInstance: this,
       parent: null,
+      uiState: {},
     });
 
     mainHelper.search();
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -102,6 +102,14 @@ export type InstantSearchOptions<TRouteState = UiState> = {
   searchParameters?: PlainSearchParameters;
 
   /**
+   * Injects an `uiState` to the `instantsearch` instance. You can use this option
+   * to provide an initial state to a widget. Note that the state is only used
+   * for the first search. To unconditionally pass additional parameters to the
+   * Algolia API, take a look at the `configure` widget.
+   */
+  initialUiState?: UiState;
+
+  /**
    * Time before a search is considered stalled. The default is 200ms
    */
   stalledSearchDelay?: number;
@@ -137,6 +145,7 @@ class InstantSearch extends EventEmitter {
   public _searchStalledTimer: any;
   public _isSearchStalled: boolean;
   public _searchParameters: PlainSearchParameters;
+  public _initialUiState: UiState;
   public _searchFunction?: InstantSearchOptions['searchFunction'];
   public _createURL?: (params: SearchParameters) => string;
   public _createAbsoluteURL?: (params: SearchParameters) => string;
@@ -150,6 +159,7 @@ class InstantSearch extends EventEmitter {
       indexName = null,
       numberLocale,
       searchParameters = {},
+      initialUiState = {},
       routing = null,
       searchFunction,
       stalledSearchDelay = 200,
@@ -215,6 +225,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     this._stalledSearchDelay = stalledSearchDelay;
     this._searchStalledTimer = null;
     this._isSearchStalled = false;
+    this._initialUiState = initialUiState;
     this._searchParameters = {
       ...searchParameters,
       index: indexName,
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -403,7 +403,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     this.mainIndex.init({
       instantSearchInstance: this,
       parent: null,
-      uiState: {},
+      uiState: this._initialUiState,
     });
 
     mainHelper.search();
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -102,7 +102,7 @@ export type InstantSearchOptions<TRouteState = UiState> = {
   searchParameters?: PlainSearchParameters;
 
   /**
-   * Injects an `uiState` to the `instantsearch` instance. You can use this option
+   * Injects a `uiState` to the `instantsearch` instance. You can use this option
    * to provide an initial state to a widget. Note that the state is only used
    * for the first search. To unconditionally pass additional parameters to the
    * Algolia API, take a look at the `configure` widget.
--- a/src/lib/__tests__/InstantSearch-test.js
+++ b/src/lib/__tests__/InstantSearch-test.js
@@ -172,7 +172,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
 
   it('throws if createURL is called before start', () => {
     const search = new InstantSearch({
-      indexName: 'index_name',
+      indexName: 'indexName',
       searchClient: createSearchClient(),
     });
 
@@ -185,7 +185,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
 
   it('throws if refresh is called before start', () => {
     const search = new InstantSearch({
-      indexName: 'index_name',
+      indexName: 'indexName',
       searchClient: createSearchClient(),
     });
 
@@ -205,7 +205,7 @@ describe('InstantSearch', () => {
 
     // eslint-disable-next-line no-new
     new InstantSearch({
-      indexName: 'index_name',
+      indexName: 'indexName',
       searchClient,
     });
 
@@ -218,7 +218,7 @@ describe('InstantSearch', () => {
   it('does not call algoliasearchHelper', () => {
     // eslint-disable-next-line no-new
     new InstantSearch({
-      indexName: 'index_name',
+      indexName: 'indexName',
       searchClient: createSearchClient(),
     });
 
@@ -228,7 +228,7 @@ describe('InstantSearch', () => {
   it('does store insightsClient on the instance', () => {
     const insightsClient = () => {};
     const search = new InstantSearch({
-      indexName: 'index_name',
+      indexName: 'indexName',
       searchClient: createSearchClient(),
       insightsClient,
     });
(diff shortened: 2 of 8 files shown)


src/widgets/index/index.ts
```diff
@@ -212,6 +216,7 @@ const index = (props: IndexProps): Index => {
             widget.init({
               helper: helper!,
               parent: this,
+              uiState: {},
```
samouss: The `initialUiState` is only provided on the first search. Dynamically added widgets ignore the option.

src/lib/InstantSearch.ts
```diff
@@ -101,6 +101,14 @@ export type InstantSearchOptions<TRouteState = UiState> = {
    */
   searchParameters?: PlainSearchParameters;
 
+  /**
+   * Injects an `uiState` to the `instantsearch` instance. You can use this option
```
```suggestion
   * Injects a `uiState` to the `instantsearch` instance. You can use this option
```

src/widgets/index/index.ts
Haroenv: By this you mean that it's only used by the `init` of the index itself, right?

stories/instantsearch.stories.js
```diff
@@ -1,17 +1,31 @@
...
-);
+        helper.search();
+      },
+    })
+  )
+  .add(
+    'with initialUiState',
+    withHits(() => {}, {
+      initialUiState: {
+        // eslint-disable-next-line @typescript-eslint/camelcase
```
Haroenv: was thinking on allowing the string `instant_search` specifically, or else we can do it with indexId, but also fine to just ignore here

src/lib/InstantSearch.ts
```diff
@@ -137,6 +145,7 @@ class InstantSearch extends EventEmitter {
   public _searchStalledTimer: any;
   public _isSearchStalled: boolean;
   public _searchParameters: PlainSearchParameters;
+  public _initialUiState: UiState;
```
Haroenv: would  be nice if we could avoid storing it, but since we need it in `start`, not in the constructor, it isn't possible

Haroenv:
How does it behave when combined with routing on initial load with a state in the url?

src/widgets/index/__tests__/index-test.ts
```diff
@@ -1367,15 +1369,73 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
...
+                  brand: ['Apple'],
+                },
+              },
+              // @ts-ignore
+              subLevelIndexName: {
+                configure: {
+                  hitsPerPage: 2,
+                },
+                menu: {
+                  categgories: 'Phone',
```
```suggestion
                  categories: 'Phone',
```

src/widgets/index/__tests__/index-test.ts
```diff
@@ -1367,15 +1369,73 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
...
+            refinementList: {
+              brand: ['Apple'],
+            },
+          },
+          subLevelIndexName: {
+            configure: {
+              hitsPerPage: 2,
+            },
+            menu: {
+              categgories: 'Phone',
```
```suggestion
              categories: 'Phone',
```

src/widgets/index/index.ts
samouss: I mean, only on the first loop over the `init` lifecycle. An `index` dynamically added won't take into account the `initialUiState`. We can provide it since we have access to the `instantsearchInstance` but I'm not sure which behavior is the one expected: ignored or taken into account?

stories/instantsearch.stories.js
samouss: Not sure it's really useful, until now we don't have it that much inside the codebase.

samouss:
> How does it behave when combined with routing on initial load with a state in the url?

Nothing is implemented right now since routing is not updated (yet). For a first version I think an override (or shallow merge) will be enough otherwise we have to implement a merge function for `UiState`. This is something that I would like to avoid as much as possible. Another alternative is to allow the user to provide its own function to perform the merge.

src/widgets/index/__tests__/index-test.ts
```diff
@@ -1367,15 +1369,73 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
...
+          indexId: {
             query: 'Apple',
             page: 5,
           },
         });
       });
 
-      it('uses `indexId` for scope key', () => {
-        const instance = index({ indexName: 'indexName', indexId: 'indexId' });
+      it('uses `uiState` for the local `uiState`', () => {
```
francoischalifour: Can we make this test name clearer? Do you mean "forwards the init `uiState` to the widget state"?

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -421,17 +421,41 @@ describe('start', () => {
...
-        index: 'index_name',
+        index: 'indexName',
         hitsPerPage: 5,
         disjunctiveFacetsRefinements: { brand: ['Apple'] },
         disjunctiveFacets: ['brand'],
       })
     );
   });
 
+  it('forwards the `initialUiState` to the main index', () => {
```
francoischalifour: We don't test for nested indices?

samouss: No we don't, it's already covered inside [`index`](https://github.com/algolia/instantsearch.js/blob/271fc7b2cd905d3ca388a5e31ecd2b4c2e3ca27e/src/widgets/index/__tests__/index-test.ts#L1379-L1435). From the InstantSearch perspective it's the same, it doesn't matter if there is 1 or 100 indices.

src/widgets/index/__tests__/index-test.ts
samouss: `uses the provided `uiState` for the local `uiState` is it better?

src/widgets/index/__tests__/index-test.ts
samouss: https://github.com/algolia/instantsearch.js/pull/4074/commits/4f9a1df26b4a4c9da0bd60a9b9ad750a31c79e39

src/widgets/index/__tests__/index-test.ts
samouss: https://github.com/algolia/instantsearch.js/pull/4074/commits/4f9a1df26b4a4c9da0bd60a9b9ad750a31c79e39

src/widgets/index/__tests__/index-test.ts
francoischalifour: Yep!

samouss: https://github.com/algolia/instantsearch.js/pull/4074/commits/2cf0ac103c47dcc864c14a6a3b87ca1ec0e31660

fabregas4:
Any evidence this actually works in v4? Can't get it to work at all myself. No console errors, but also zero impact on results. Have followed the example at https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/#widget-param-initialuistate

Haroenv:
Note that you need to mount a (could be virtual / non-rendering) widget too. We have a warning for that case normally, so I’d like to see your reproduction on codesandbox

fabregas4:
@Haroenv Thanks for responding. Here is some code on codesandbox:
https://codesandbox.io/s/wizardly-jackson-ovx2f?fontsize=14

Not sure what you mean about mounting a widget. The hits widget should be enough right?

You'll see in the sample that there are results produced but they don't have the query or Designer refinement applied. Have tried various options but none are making a difference. Have tried just the query and not the refinementList, and vice versa. Not 100% sure if initialUiState is meant to have a property of "indexName" or "products" (which is the actual index name). Have tried both in any case. Have also tried not including this node, i.e. just:
`        initialUiState: {
          refinementList: {
            Designer: ["Zimmermann"]
          }
        }`
which is how the code example says its done on this page: https://www.algolia.com/doc/guides/building-search-ui/upgrade-guides/js/#upgrade-from-v3-to-v4

I'm not the most advanced coder but the documentation in this area seems to be very light.

Shouldn't be any issue on the index side, as falling back to v3 and specifying searchParameters produces expected results. E.g.
`
searchParameters: {
disjunctiveFacetsRefinements: {"Designer":["Zimmermann"]}
}
`

Any ideas?

Haroenv:
```js
  indexName: "products",
  searchClient,
  initialUiState: {
    products: {
      query: "fakequerythatshouldreturnnoresults",
      refinementList: {
        Designer: ["Zimmermann"]
      }
    }
  }
});

const virtualRefinementList = instantsearch.connectors.connectRefinementList(
  () => null
);
const virtualSearchBox = instantsearch.connectors.connectSearchBox(
  () => null
);

search.addWidgets([
  virtualSearchBox(),
  virtualRefinementList({ attribute: "Designer" }),
  instantsearch.widgets.hits({
    container: "#hits",
    templates: {
      item: `<div>{{Designer}} - {{ProductTitle}}</div>`
    }
  })
]);
```

This is what's meant with "virtual widget". 

Thanks for pointing out that [this section](https://www.algolia.com/doc/guides/building-search-ui/upgrade-guides/js/#searchparameters) is indeed insufficient to follow if  you didn't already  use advanced InstantSearch features.

Another option is the following:

```js
search.addWidgets([
  instantsearch.widgets.configure(
      { disjunctiveFacetsRefinements: {"Designer":["Zimmermann"]} }
  ),
]);
```

Haroenv:
see my example: https://codesandbox.io/s/determined-wood-quvoq

fabregas4:
Thanks, that seems to have done the job. I must say it does seem more difficult than using `searchParameters` from previous versions, so I'm not sure why that would be deprecated in favour of this, although I'm sure you guys have your reasons :) Cheers

Haroenv:
I think you missed the first part of that note, saying the default replacement is the configure widget. I've updated the docs to be clearer by the way

fabregas4:
I did miss that, my bad. Looking at the docs now...

https://www.algolia.com/doc/api-reference/widgets/configure/js/
points me to:
https://www.algolia.com/doc/api-reference/search-api-parameters/

There is no mention of "disjunctiveFacetsRefinements" as a possible parameter in the searchParameters object.

Haroenv:
True, that’s also somewhat confusing. The arguments aren’t just search parameters, but also helper ones: https://community.algolia.com/algoliasearch-helper-js/reference.html

PR #4075, opened 2019-08-28
feat: update UiState definition

samouss:
This PR updates the type definitions to match the new format of the `UiState`. We introduce a new type called `IndexUiState`. The `UiState` is the global structure that represents the InstantSearch instance with each indices. The `IndexUiState` is the previous version of the `UiState`, it represents the state of an "index". Here is their definitions:

```ts
export type IndexUiState = {
  query?: string;
  hitsPerPage?: number;
  // ...
};

export type UiState = {
  [indexId: string]: IndexUiState;
};
```

The introduction of this type mainly impact the `StateMapping` & `RoutingManager`. Both have been updated. The changes applied to `RoutingManager` are there to make the tests pass don't pay too much attention to them since we'll rework this part soon.

**Note**: Name are just a proposition, we could use something else like `WidgetsState`, ... ideas are welcome! We could rename `UiState` for consistency if we find a better name for both.

### DIFF
--- a/src/lib/RoutingManager.ts
+++ b/src/lib/RoutingManager.ts
@@ -3,6 +3,7 @@ import { isEqual } from './utils';
 import {
   InstantSearch,
   UiState,
+  IndexUiState,
   Router,
   StateMapping,
   Widget,
@@ -21,6 +22,7 @@ class RoutingManager implements Widget {
   private readonly stateMapping: StateMapping;
 
   private isFirstRender: boolean = true;
+  private indexId: string;
   private currentUiState: UiState;
   private initState?: UiState;
   private renderURLFromState?: (event: HelperChangeEvent) => void;
@@ -33,6 +35,7 @@ class RoutingManager implements Widget {
     this.router = router;
     this.stateMapping = stateMapping;
     this.instantSearchInstance = instantSearchInstance;
+    this.indexId = this.instantSearchInstance.indexName;
     this.currentUiState = this.stateMapping.routeToState(this.router.read());
   }
 
@@ -44,6 +47,7 @@ class RoutingManager implements Widget {
     uiState: UiState;
   }): SearchParameters {
     const widgets = this.instantSearchInstance.mainIndex.getWidgets();
+    const indexUiState = uiState[this.indexId] || {};
 
     return widgets.reduce((parameters, widget) => {
       if (!widget.getWidgetSearchParameters) {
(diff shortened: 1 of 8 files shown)


src/lib/RoutingManager.ts
```diff
@@ -44,14 +47,15 @@ class RoutingManager implements Widget {
...
+    const indexUiState = uiState[this.indexId] || {};
 
     return widgets.reduce((parameters, widget) => {
       if (!widget.getWidgetSearchParameters) {
         return parameters;
       }
 
       return widget.getWidgetSearchParameters(parameters, {
-        uiState,
+        uiState: indexUiState,
```
samouss: I didn't update the name of the argument but its type changed (`UiState` → `IndexUiState`). We can change it (BC) but IMO it's not worth it. We already have the type to warn us about its usage. We can update it if you think that it's too confusing.

src/types/widget.ts
```diff
@@ -47,10 +47,10 @@ export interface WidgetStateOptions {
 }
 
 export interface WidgetSearchParametersOptions {
-  uiState: UiState;
+  uiState: IndexUiState;
```
samouss: https://github.com/algolia/instantsearch.js/pull/4075/files#r318454539

src/lib/RoutingManager.ts
Haroenv: We can also change the route state name to something else, keeping UiState for the focused one, and "GlobalUiState" for global issues

src/lib/stateMappings/simple.ts
```diff
@@ -1,18 +1,33 @@
-import { UiState, StateMapping } from '../../types';
+import { UiState, IndexUiState, StateMapping } from '../../types';
+
+function indexStateWithoutConfigure(uiState: IndexUiState): IndexUiState {
```
francoischalifour: `getIndexStateWithoutConfigure` would better reflect the intent of this function.

src/lib/RoutingManager.ts
samouss: The needs to put "global" in the name, feels odd to me.

francoischalifour: I think this "mis-naming" is fine for now. We can still add it to the list of things to think about/check before releasing v4 stable.

src/lib/stateMappings/simple.ts
samouss: https://github.com/algolia/instantsearch.js/pull/4075/commits/0df9e8ca4ffebff2c4a47fef7d2813d5ee35911d

PR #4077, opened 2019-08-28
docs(example): fix html for hits

Haroenv:
There can only be one main element per page:

> A document mustn't have more than one <main> element that doesn't have the hidden attribute specified.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main

Found  this running an audit on the demo: 

Screenshot 2019-08-28 at 12 53 55

(to check, do we use `main` in css? let's see in the preview)

### DIFF
--- a/examples/e-commerce/src/widgets/Products.ts
+++ b/examples/e-commerce/src/widgets/Products.ts
@@ -9,7 +9,7 @@ export const products = hits({
     <img src="{{image}}" alt="{{name}}" class="hit-image">
   </header>
 
-  <main class="hit-info-container">
+  <div class="hit-info-container">
     <p class="hit-category">{{categories.0}}</p>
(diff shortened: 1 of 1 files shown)


Haroenv:
fixed in Vue InstantSearch & React InstantSearch: 

https://github.com/algolia/vue-instantsearch/commit/3c5bd4a9a6ffa8c8432fa83f71dfd55c8df13813
https://github.com/algolia/react-instantsearch/commit/c66a39cde3c1508ad6b895fe533133da7b6c376c

Angular InstantSearch does not have this example as far as I can tell

francoischalifour:
@Haroenv Angular InstantSearch does have it on the `v3` branch: https://github.com/algolia/angular-instantsearch/tree/v3/examples/e-commerce/src.

Haroenv:
fixed in https://github.com/algolia/angular-instantsearch/commit/154f90d40eb534b4822e56d75bf7ad2635d0d096

PR #4078, opened 2019-08-28
feat(index): replicate searchFunction hack

samouss:
This PR replicates [the hack](https://github.com/algolia/instantsearch.js/blob/509513c0feafaad522f6f18d87a441559f4aa050/src/lib/RoutingManager.ts#L113-L130) implemented inside `RoutingManager` for the `searchFunction`. The `uiState` is now a first-class citizen so we have to apply this logic to the index. I've tried to keep the same process: the update is triggered once all the widgets of the index are rendered like it was the case inside the manager (since it was always the last widget).

At the moment the call to notify the instance that a change occurs isn't implemented because we don't expose the function (yet). Once the API is available I'll update the PR. The pieces that need to be updated are commented.

### DIFF
--- a/src/widgets/index/__tests__/index-test.ts
+++ b/src/widgets/index/__tests__/index-test.ts
@@ -1483,7 +1483,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
         expect(instantSearchInstance.onStateChange).not.toHaveBeenCalled();
       });
 
-      it('updates the local `uiState` only with widgets not indices', () => {
+      it('updates the local `uiState` only with widgets', () => {
         const level0 = index({ indexName: 'level0IndexName' });
         const level1 = index({ indexName: 'level1IndexName' });
         const widgets = [createSearchBox(), createPagination()];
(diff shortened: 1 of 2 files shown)


src/widgets/index/__tests__/index-test.ts
```diff
@@ -1508,6 +1510,88 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
...
+        // Simulate a change that does not emit (like `searchFunction`)
+        instance.getHelper()!.overrideStateWithoutTriggeringChangeEvent({
+          ...instance.getHelper()!.state,
+          query: 'Apple iPhone',
+        });
+
+        instance.render(createRenderOptions());
+
+        // @TODO: assert that onChange is called once
+        // expect(...onChange).toHaveBeenCalledTimes(1);
```
samouss: Wait for IFW-932

src/widgets/index/__tests__/index-test.ts
```diff
@@ -1508,6 +1510,88 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
...
+        // Simulate a change that does not emit (like `searchFunction`)
+        instance.getHelper()!.overrideStateWithoutTriggeringChangeEvent({
+          ...instance.getHelper()!.state,
+          query: 'Apple iPhone XS',
+        });
+
+        instance.render(createRenderOptions());
+
+        // @TODO: assert that onChange is called once
+        // expect(...onChange).toHaveBeenCalledTimes(1);
```
samouss: Wait for IFW-932

src/widgets/index/__tests__/index-test.ts
```diff
@@ -1508,6 +1510,88 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
...
+          // the DerivedHelper does not contain the results. The `lastResults` attribute
+          // is set once the `result` event is emitted.
+          .search();
+
+        await runAllMicroTasks();
+
+        topLevelInstance.render(createRenderOptions());
+
+        // @TODO: assert that onChange is not called
+        // expect(...onChange).not.toHaveBeenCalled()
```
samouss: Wait for IFW-932

src/widgets/index/index.ts
```diff
@@ -408,6 +410,29 @@ const index = (props: IndexProps): Index => {
...
+        });
+
+        if (!isEqual(localUiState, firstRenderState)) {
+          // Force update the URL, if the state has changed since the initial read.
+          // We do this in order to make the URL update when there is `searchFunction`
+          // that prevents the search of the initial rendering.
+          // See: https://github.com/algolia/instantsearch.js/issues/2523#issuecomment-339356157
+          localUiState = firstRenderState;
+
+          // @TODO: call onChange to trigger
```
samouss: Wait for IFW-932

src/widgets/index/index.ts
```diff
@@ -408,6 +410,29 @@ const index = (props: IndexProps): Index => {
...
         }
       });
+
+      // Hack for backward compatibily with `searchFunction` + `routing`
+      // https://github.com/algolia/instantsearch.js/blob/509513c0feafaad522f6f18d87a441559f4aa050/src/lib/RoutingManager.ts#L113-L130
+      if (localParent === null && isFirstRender) {
+        isFirstRender = false;
+        // Compare initial state and first render state to see if the query has been
+        // changed by the `searchFunction`. It's required because the helper of the
+        // `searchFunction` does not trigger change event (not the same instance).
```
francoischalifour:
```suggestion
        // `searchFunction` does not trigger change events (not the same instance).
```

src/widgets/index/index.ts
```diff
@@ -408,6 +410,29 @@ const index = (props: IndexProps): Index => {
...
+        const firstRenderState = getLocalWidgetsState(localWidgets, {
+          helper: helper!,
+          searchParameters: helper!.state,
+        });
+
+        if (!isEqual(localUiState, firstRenderState)) {
+          // Force update the URL, if the state has changed since the initial read.
```
francoischalifour:
```suggestion
          // Force update the URL if the state has changed since the initial read.
```

src/widgets/index/index.ts
```diff
@@ -408,6 +410,29 @@ const index = (props: IndexProps): Index => {
...
+          // We do this in order to make the URL update when there is `searchFunction`
```
francoischalifour:
```suggestion
          // We do this to trigger a URL update when `searchFunction` prevents
```

src/widgets/index/__tests__/index-test.ts
```diff
@@ -1508,6 +1510,88 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
...
+
+        // Simulate a change that does not emit (like `searchFunction`)
+        instance.getHelper()!.overrideStateWithoutTriggeringChangeEvent({
+          ...instance.getHelper()!.state,
+          query: 'Apple iPhone XS',
+        });
+
+        instance.render(createRenderOptions());
+
+        // @TODO: assert that onChange is called once
```
```suggestion
        // @TODO: assert that onChange is called second time
```

src/widgets/index/__tests__/index-test.ts
```suggestion
        // expect(...onChange).toHaveBeenCalledTimes(2);
```

src/widgets/index/index.ts
```diff
@@ -408,6 +410,29 @@ const index = (props: IndexProps): Index => {
...
+          // that prevents the search of the initial rendering.
```
francoischalifour:
```suggestion
          // the search on the initial render.
```

src/widgets/index/__tests__/index-test.ts
samouss: Actually no it has to be called only once. We only update the `uiState` on the first render.

PR #4079, opened 2019-08-28
fix(sortBy): read initial index name from parent index

francoischalifour:
## Description

This fixes a bug where the `sortBy` widget used to read the `indexName` from the InstantSearch instance. It now reads the `indexName` from its parent index.

## Changes

- The index widget now exposes its name with `getIndexName`.
- The test mocks were updated and used to get closer to a real environment. In a real environment, [all widgets receive a `parent` at the `init` step](https://github.com/algolia/instantsearch.js/blob/bf0497864d880e41007e80ac163d270fcc3f83a0/src/widgets/index/index.ts#L219). The mock didn't reflect that. It is updated to be assigned the main index coming from the InstantSearch instance by default (this is the default behavior when using InstantSearch without nested indices). The tests for the `index` widget now need to explicitly set the `parent` to `null` (this is fine because this is a rarer case compared to all other widgets).


src/connectors/sort-by/connectSortBy.js
```diff
@@ -103,26 +103,26 @@ export default function connectSortBy(renderFn, unmountFn = noop) {
...
+      init({ helper, instantSearchInstance, parent }) {
+        const currentIndex = helper.state.index;
+        const isCurrentIndexInItems = find(
           items,
-          item => item.value === initialIndex
+          item => item.value === currentIndex
         );
 
-        this.initialIndex = initialIndex;
+        this.initialIndex = parent.getIndexName();
```
Haroenv: good call, there's indeed always a parent!

src/connectors/sort-by/connectSortBy.js
```diff
@@ -103,26 +103,26 @@ export default function connectSortBy(renderFn, unmountFn = noop) {
     return {
       $$type: 'ais.sortBy',
 
-      init({ helper, instantSearchInstance }) {
-        const initialIndex = helper.state.index;
-        const isInitialIndexInItems = find(
+      init({ helper, instantSearchInstance, parent }) {
+        const currentIndex = helper.state.index;
+        const isCurrentIndexInItems = find(
```
Haroenv: is this dead code eliminated in the prod build? it seems to be just for dev mode

src/connectors/sort-by/connectSortBy.js
Haroenv: good call, there's indeed always a parent!

src/connectors/sort-by/__tests__/connectSortBy-test.js
```diff
@@ -389,19 +405,110 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/sort-by/js/
...
-        expect(uiStateAfter).toBe(uiStateBefore);
+
+        expect(uiStateAfter).toEqual(uiStateBefore);
+      });
+
+      test('should use the top-level `indexName` for the initial index', () => {
+        const render = jest.fn();
+        const makeWidget = connectSortBy(render);
+        const instantSearchInstance = createInstantSearch({
+          indexName: 'initial_index_name',
```
samouss: Prefer the usage of camel-case. The comment apply for the all file.

test/mock/createInstantSearch.ts
```diff
@@ -8,7 +8,7 @@ export const createInstantSearch = (
   args: Partial<InstantSearch> = {}
 ): InstantSearch => {
   const searchClient = createSearchClient();
-  const indexName = 'index_name';
+  const indexName = args.indexName === undefined ? 'indexName' : args.indexName;
```
samouss: You can simplify this logic with the rest/spread.

https://github.com/algolia/instantsearch.js/blob/bf0497864d880e41007e80ac163d270fcc3f83a0/test/mock/createWidget.ts#L14

src/connectors/sort-by/connectSortBy.js
francoischalifour: That's worth checking, but really related to this PR.

Haroenv: I have checked, and both the warning and the find are present in the prod build

PR #4080, opened 2019-08-28
feat(instantsearch): add onStateChange method

yannickcr:
**Summary**

IFW-932

Implement the (for now empty) `onStateChange` method on `InstantSearch`.

This method is called by the `index` widget every time its `localUiState` is updated and will in the future notify the middlewares (like the routing) of any change in the widgets state.

The tests for the `index` widget were updated to check if `onStateChange` was called as expected.

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -471,7 +471,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     }
   }
 
-  public onChange = () => {
+  public onStateChange = () => {
     // @TODO: Provide `nextUiState` to all middlewares (eg. routing)
     // const nextUiState = this.mainIndex.getWidgetState({});
   };
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -471,6 +471,11 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     }
   }
 
+  public onChange = defer(() => {
+    // @TODO: Provide `nextUiState` to all middlewares (eg. routing)
+    // const nextUiState = this.mainIndex.getWidgetState({});
+  });
+
   public createURL(params: PlainSearchParameters): string {
     if (!this._createURL) {
       throw new Error(
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -471,10 +471,10 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     }
   }
 
-  public onChange = defer(() => {
+  public onChange = () => {
     // @TODO: Provide `nextUiState` to all middlewares (eg. routing)
     // const nextUiState = this.mainIndex.getWidgetState({});
-  });
+  };
 
   public createURL(params: PlainSearchParameters): string {
     if (!this._createURL) {
--- a/src/widgets/index/__tests__/index-test.ts
+++ b/src/widgets/index/__tests__/index-test.ts
@@ -443,12 +443,10 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
           })
         );
 
-        // `instantSearchInstance` must have been notified 4 times of the `uiState` changes:
-        // 1. By the widget, for the widget initialization
-        // 2. By the helper `change` event callback, for the change to the query parameters
-        // 3. By the widget, for the children widget disposal
-        // 4. By the helper `change` event callback, also for the children widget disposal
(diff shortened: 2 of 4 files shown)


src/widgets/index/__tests__/index-test.ts
```diff
@@ -434,6 +442,13 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
             distinct: true,
           })
         );
+
+        // `instantSearchInstance` must have been notified 4 times of the `uiState` changes:
+        // 1. By the widget, for the widget initialization
+        // 2. By the helper `change` event callback, for the change to the query parameters
+        // 3. By the widget, for the children widget disposal
+        // 4. By the helper `change` event callback, also for the children widget disposal
```
yannickcr: Is it expected? Not critical since the call to onChange is deferred, but maybe the call to `onChange` in `removeWidgets` is unnecessary.

WDYT?

src/widgets/index/index.ts
```diff
@@ -281,6 +282,7 @@ const index = (props: IndexProps): Index => {
       localInstantSearchInstance = instantSearchInstance;
       localParent = parent;
       localUiState = uiState[indexId] || {};
+      instantSearchInstance.onChange();
```
samouss: We don't have to call it on `init`, nothing have changed yet.

src/lib/InstantSearch.ts
```diff
@@ -471,6 +471,11 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     }
   }
 
+  public onChange = defer(() => {
```
samouss: Not sure about `defer` for this function. The deferred function does trigger a side effect (like a network request) that's why we batch them. But this callback is a way to notify the "middleware" that something happens. IMO the decision to pick or not the change belongs to the "middleware", not InstantSearch.

src/widgets/index/__tests__/index-test.ts
samouss: Indeed I think the it's useless, the call below (`setState`) trigger the `change` event which already contains `onChange`. You may have noticed that we compute two times the `uiState` on remove. Once before `setState` to be able to correctly get the `SearchParameters`, once after when the event is emitted (regular flow). We might be able to avoid the two consecutive calls but at the moment it's good enough.

francoischalifour:
Calling this callback `onStateChange` would make sense.

A few other APIs exposing an `onStateChange` function:

- [Downshift](https://github.com/downshift-js/downshift#onstatechange)
- [ServiceWorker](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorker/onstatechange)
- [YouTube](https://developers.google.com/youtube/iframe_api_reference)

src/lib/InstantSearch.ts
yannickcr: The idea was more to avoid recomputing `nextUiState` too often and avoid exposing intermediate states to the middlewares.

But since the middlewares are not really defined yet then we can keep a simple function for now and see later if it is useful to defer it.

src/widgets/index/__tests__/index-test.ts
```diff
@@ -434,6 +440,11 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index/js/"
             distinct: true,
           })
         );
+
+        // `instantSearchInstance` must have been notified 2 times of the `uiState` changes:
+        // 1. By the helper `change` event callback, for the change to the query parameters
+        // 2. By the helper `change` event callback, for the children widget disposal
```
Haroenv:
```suggestion
        // 2. By the helper `change` event callback, for the child widgets being disposed
```

src/lib/InstantSearch.ts
```diff
@@ -471,6 +471,11 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     }
   }
 
+  public onChange = () => {
```
```suggestion
  public onStateChange = () => {
```

this makes sense for me :) but we probably could add an _ in front of it to make it "internal"?

samouss: I agree for the name of the function it's better like this. We don't have the concept of `state` though but `uiState`. It would be more consistent to name it: `onUiStateChange`. 

Not sure about the internal, it doesn't matter at the moment. We already have multiple attributes that are exposed publicly but only used internally (`helper`, `mainHelper`, `mainIndex`, ...).

francoischalifour: This function answers to any change event sent by the helper, so it's not specifically coupled with a `uiState` change, but rather any change. Saying that it's a "UI state change" is a lie. Later, we can think about using this hook for other purposes. Hence, `onStateChange` is generic yet accurate enough IMO.

PR #4081, opened 2019-08-29
feat: drop support of searchParameters for initialUiState

samouss:
This PR drops the support of `searchParameters` for `initialUiState`. Here is the changes:

- `InstantSearch`: remove the support of the option
- `index`: remove the logic for the option
- `stories`: replace the usage of `searchParamters` by `initialUiState` when applicable
- `stories`: replace the usage of `searchParamters` by `configure` when applicable

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,10 @@ module.exports = {
       'error',
       { argsIgnorePattern: '^_', ignoreRestSiblings: true },
     ],
+    '@typescript-eslint/camelcase': [
+      'error',
+      { allow: ['instant_search', 'instant_search_movies'] },
+    ],
   },
   overrides: [
     {
--- a/.storybook/decorators/withHits.ts
+++ b/.storybook/decorators/withHits.ts
@@ -19,7 +19,6 @@ export const withHits = (
     appId = 'latency',
     apiKey = '6be0576ff61c053d5f9a3225e2a90f76',
     indexName = 'instant_search',
-    searchParameters = {},
     playground = defaultPlayground,
     ...instantsearchOptions
   } = searchOptions || {};
@@ -28,12 +27,6 @@ export const withHits = (
   const search = instantsearch({
     indexName,
     searchClient: algoliasearch(appId, apiKey),
-    searchParameters: {
-      hitsPerPage: 4,
-      attributesToSnippet: ['description:15'],
-      snippetEllipsisText: '[…]',
-      ...searchParameters,
-    },
     routing: {
       router: {
         write: (routeState: object) => {
@@ -66,6 +59,14 @@ export const withHits = (
   const rightPanelPlaygroundElement = document.createElement('div');
   rightPanelPlaygroundElement.classList.add('panel-right');
   playgroundElement.appendChild(rightPanelPlaygroundElement);
+
+  search.addWidget(
+    instantsearch.widgets.configure({
+      hitsPerPage: 4,
+      attributesToSnippet: ['description:15'],
+      snippetEllipsisText: '[…]',
+    })
+  );
 
   playground({
     search,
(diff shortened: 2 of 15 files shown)


.eslintrc.js
```diff
@@ -10,6 +10,10 @@ module.exports = {
       'error',
       { argsIgnorePattern: '^_', ignoreRestSiblings: true },
     ],
+    '@typescript-eslint/camelcase': [
+      'error',
+      { allow: ['instant_search', 'instant_search_movies'] },
```
samouss: It's convenient since we use them a lot in the stories.

.storybook/decorators/withHits.ts
```diff
@@ -19,7 +19,6 @@ export const withHits = (
     appId = 'latency',
     apiKey = '6be0576ff61c053d5f9a3225e2a90f76',
     indexName = 'instant_search',
-    searchParameters = {},
```
samouss: Any additional parameters can be provided to a `configure` widget at the story level.

stories/breadcrumb.stories.js
```diff
@@ -1,5 +1,17 @@
 import { storiesOf } from '@storybook/html';
 import { withHits, withLifecycle } from '../.storybook/decorators';
+import { connectHierarchicalMenu } from '../src/connectors';
+import { noop } from '../src/lib/utils';
+
+const virtualHierarchicalMenu = (args = {}) =>
+  connectHierarchicalMenu(noop, noop)({
```
samouss: The widget is hidden so I could have used a `configure` widget, but at the moment it's doesn't work. The widget uses `setQueryParameters` which means that it will override the previous value. We have to update the behavior of the widget to use `mergeSearchParameters` (in a separate PR). It will ease the migration for our users since we advocate a lot for the usage of `configure`.

.storybook/decorators/withHits.ts
Haroenv: we could change this to use a configure widget automatically if the parameter is given, if it's sufficient amount of work?

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -392,12 +361,11 @@ describe('start', () => {
...
-        helper.search();
-      },
-      searchParameters: {
-        disjunctiveFacetsRefinements: { brand: ['Apple'] },
-        disjunctiveFacets: ['brand'],
+        const nextState = helper.state
+          .addDisjunctiveFacet('brand')
+          .addDisjunctiveFacetRefinement('brand', 'Apple');
+
+        helper.setState(nextState).search();
```
Haroenv: can be done without touching the state methods, since both exist on helper "root"

stories/breadcrumb.stories.js
Haroenv: you mean that the logic here:

https://github.com/algolia/instantsearch.js/blob/bf0497864d880e41007e80ac163d270fcc3f83a0/src/connectors/configure/connectConfigure.ts#L145

should be like here:

https://github.com/algolia/instantsearch.js/blob/bf0497864d880e41007e80ac163d270fcc3f83a0/src/connectors/configure/connectConfigure.ts#L94

?

.storybook/decorators/withHits.ts
samouss: It already uses a `configure` widget for the static parameters. If you want to add more values you have to create another `configure` widget. I don't see the value of adding an extra parameter only for this.

stories/breadcrumb.stories.js
samouss: Yes.

```js
getWidgetSearchParameters(state, { uiState }) {
  return mergeSearchParameters(
    state,
    new algoliasearchHelper.SearchParameters({
      ...uiState.configure,
      ...widgetParams.searchParameters
    })
  );
}
```

src/lib/__tests__/InstantSearch-test.js
samouss: You can't call `addDisjunctiveFacet` on the root, it's only with a configuration object.

PR #377, opened 2015-10-28
feat(urls): ability to create an URL from a set of params

redox:
Fix #372

BREAKING CHANGE: the instantsearch.createURL method is now taking a
simple JS object and not a SearchParameter instance anymore.

Issue #372, opened 2015-10-27
Url generation from parameters


maxiloc:
For some integration like Magento, to load the instant search page with good params, the easiest way is to generate the (url) hash of the query and redirect to it. That way instantsearch will load on it's on.

Right now the url is hardcoded in the code which is not the cleaner way to do it.

A solution can be to instantiate a helper but that's not perfect to.
Adding this function to instantsearch could be useful

For a set of refinements like ```{query: "madison}``` or ```[{query: "", refinements: ["category:Shirt"]}``` the idea would be to generate something like ```#q=madison&idx=...```

Also if this could be done without instantiating instantsearch it would be even better

redox:
Yes, the widgets are already using a `createURL` method to handle such use-case. @bobylito any chance we could expose it in the `instantsearch` object we're returning?

redox:
So the resulting instantsearch instance has already a `createURL` method; the only limitation is that it takes a `SearchParameters` (state) instance in argument. I'll propose something.

redox:
So you can do something like:

```js
// with an instantsearch instance:
var search = instantsearch(....);
search.createURL({facetsRefinements: {category: 'Home'}});

// or by just calling the static method:
window.location.href = '....' + instantsearch.createQueryString({facetsRefinements: {category: 'Home'}});
```

PR #4082, opened 2019-08-29
fix(createURL): support multi-index

francoischalifour:
## Description

This updates the signature of the `createURL` function given to widgets' lifecycle (`init` and `render`). It also updates the undocumented top-level `createURL` to match the same signature.

## New signature

The signature at the widget level (given to the `init` and `render` hooks) stays the same:

```ts
createURL(nextState: SearchParameters): string;
```

However, the internal signature in the `index` widget changes and uses the UI state.

```ts
createURL(nextState: UIState): string;
```

This UI state is accumulated with all the widgets mounted on the instance, which results in the complete UI state of the sub indices.

This change is especially needed for generating exhaustive URLs for both `currentRefinement` and `clearRefinements`.

## Current architecture limitations

The new architecture for v4 proceeds top to bottom. This means that if you call `createURL` in a top-level widget `init` hook, it ignores the `uiState` of all children. This results in an incomplete URL string.

This likely will never happen in user code because we rarely only use `init` without `render`. Still, if we've got complains about incomplete URLs, we can suggest adding a `render` method to fix the issue.

## Notes on top-level `createURL`

The top-level `createURL` method has never been documented and was initially introduced for our Magento integration. This is not used anymore so we might consider removing this method by the time we go stable.

See #372 and #377.

## What's remaining

There are some checks that we need to proceed and to fix in others PRs if the result is not expected.

- Check the `createURL` usage in `currentRefinements`
- Check the `createURL` usage in `clearRefinements`


src/lib/InstantSearch.ts
```diff
@@ -147,8 +146,7 @@ class InstantSearch extends EventEmitter {
   public _searchParameters: PlainSearchParameters;
   public _initialUiState: UiState;
   public _searchFunction?: InstantSearchOptions['searchFunction'];
-  public _createURL?: (params: SearchParameters) => string;
-  public _createAbsoluteURL?: (params: SearchParameters) => string;
```
samouss: 👍

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -1007,56 +1024,97 @@ describe('createURL', () => {
...
+        },
+      },
       routing: {
         router,
       },
     });
 
-    search.addWidget(
+    search.addWidgets([
       createWidget({
```
samouss: You can use a `connectSearchBox` rather than the mocked version.

src/widgets/index/index.ts
```diff
@@ -221,7 +221,13 @@ const index = (props: IndexProps): Index => {
               instantSearchInstance: localInstantSearchInstance,
               state: helper!.state,
               templatesConfig: localInstantSearchInstance.templatesConfig,
-              createURL: localInstantSearchInstance._createAbsoluteURL!,
+              createURL: (nextState: SearchParameters) =>
```
samouss: The function gets created at each render which wasn't the case previously.

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -995,10 +999,23 @@ describe('createURL', () => {
     createURL: jest.fn(() => '#'),
   });
 
-  it('returns the URL for the main index state', () => {
+  it('at top-level returns the default URL for the main index state', () => {
     const router = createRouter();
-    router.createURL.mockImplementation(() => 'http://algolia.com');
```
samouss: I would have kept this, at least we know where the value come from.

src/lib/__tests__/InstantSearch-test.js
francoischalifour: 26cd7a040e01a8602714995cf652927573279834

src/lib/__tests__/InstantSearch-test.js
francoischalifour: 82c05556460545f474bc507267f7218198ba284c

src/widgets/index/index.ts
francoischalifour: 20f1e8d2de4e66ac7d2da04787cb81d70d39e076

Haroenv:
merged, since @francoischalifour can't :)

PR #4087, opened 2019-08-30
feat(routing): add a "single index" compatibility mode

Haroenv:
This mode has as goal to flatten the URL structure to avoid the indexName of the root index in the URL when there's only one URL.

Usage:

```js
import { history } from 'instantsearch.js/lib/router';
import { singleIndex } from 'instantsearch.js/lib/stateMappings';

instantsearch({
  // ...
  indexName: "blabla",
  routing: {
    router: history(),
    stateMapping: singleIndex("blabla"),
  },
});
```

Right now it's not yet possible to use this router in TS because InstantSearch itself isn't generic, and trying to make this happen was causing me a cascade of Generic issues. For now you need to cast it (`stateMapping: singleIndex("blabla") as StateMapping<UiState>`).

BREAKING CHANGE: Default URLs have changed, and now will contain the full state, which includes the index name. You can use the old behaviour by using the `singleIndex` stateMapping, and passing the index to synchronise in the argument.

https://algolia.atlassian.net/browse/IFW-941


src/lib/stateMappings/__tests__/singleIndex-test.ts
```diff
@@ -0,0 +1,133 @@
...
+  describe('stateToRoute', () => {
+    it('passes normal state through', () => {
+      const stateMapping = singleIndexStateMapping('indexName');
+      const actual = stateMapping.stateToRoute({
+        indexName: {
+          query: 'zamboni',
+          refinementList: {
+            color: ['red'],
+          },
+        },
```
francoischalifour: We can add other indices in there to make sure that it only picks up the provided index properties.

src/lib/stateMappings/__tests__/singleIndex-test.ts
```diff
@@ -0,0 +1,133 @@
+import singleIndexStateMapping from '../singleIndex';
+
+describe('singleIndexStateMapping', () => {
+  describe('stateToRoute', () => {
```
francoischalifour: - What if the index name is not defined?
- What if the index name doesn't exist?

We should cover these cases.

src/lib/stateMappings/singleIndex.ts
```diff
@@ -0,0 +1,21 @@
+import { StateMapping, IndexUiState } from '../../types';
+
+function getIndexStateWithoutConfigure(uiState: IndexUiState): IndexUiState {
+  const { configure, ...trackedUiState } = uiState;
+  return trackedUiState;
+}
+
+export default function singleIndexStateMapping(
+  rootIndexName: string
```
francoischalifour: Practically, this is not always the root index name but could be any index name. We can call the parameter `indexName`.

src/lib/stateMappings/__tests__/singleIndex-test.ts
Haroenv: If the index isn’t defined an error will be thrown. That should probably be one we add ourselves. 

Test for that should indeed be given (undefined output now but might better be an empty object)

Haroenv:
I found opt-in the most elegant. We could make a version which has access to the instance & the tree, but the logic seemed too inflexible to me

src/lib/stateMappings/__tests__/singleIndex-test.ts
```diff
@@ -0,0 +1,215 @@
...
+          spy: ['stealing', 'all', 'your', 'searches'],
+        },
+      });
+    });
+
+    it('returns wrong data if used with nested state', () => {
+      const stateMapping = singleIndexStateMapping('indexName');
+      const actual = stateMapping.routeToState({
+        // @ts-ignore (we are passing wrong data)
+        indexName: {
```
tkrugg: can you tell me what parts of this object are "wrong data"?

Haroenv: routeToState expects a single index here, but this is a multi-index state (which doesn't work for "singleIndexStateMapping"

src/lib/InstantSearch.ts
```diff
@@ -218,11 +213,16 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       this._searchFunction = searchFunction;
     }
 
+    const defaultRoutingOptions = {
+      stateMapping: simpleStateMapping(),
+      router: historyRouter(),
+    };
```
tkrugg: Is there a reason for moving this inside the function (declared here vs declared at the top) ? Is it depending on some kind of internal state?

Haroenv: Originally  we wanted to use `singleIndexStateMapping` here, but it wasn't possible after all

PR #4099, opened 2019-09-02
 feat(RoutingManager): remove RoutingManager from the render process 

yannickcr:
**Summary**

IFW-935

This change removes RoutingManager from the render process, therefore it is no longer a widget. Updating state on external route update (popState) was not implemented, it will be done later in IFW-936.

List of changes:
* Remove all widget related methods from `RoutingManager`.
* Remove the hack to always have the `RoutingManager` widget last in the widget list.
* Replace https://github.com/algolia/instantsearch.js/pull/4078 by another fix (https://github.com/algolia/instantsearch.js/compare/next...feat/fs-routingmanager?expand=1#diff-1fc36c4ea042d2ecb5a498416e5e0357R366-R369. The reason is that with the original fix there was an issue with the configure widget (present in `firstRenderState` but absent from `localUiState`) forcing the URL to always be updated.
* Remove some tests: 
  * Tests about `getAllUiStates` and `getAllSearchParameters` (both methods were removed)
  * Tests about updating state on external route update (since it is not implemented yet), but they will be re-added in IFW-936.

**Note**: This PR includes https://github.com/algolia/instantsearch.js/pull/4082 and will be rebased once this one is merged. Since it can make harder to review the changes I advise you to start your review from https://github.com/algolia/instantsearch.js/commit/b55a1fa39f5ec61c1ac664527b93553914cf723e

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -139,7 +139,7 @@ class InstantSearch extends EventEmitter {
   public _createURL?(nextState: UiState): string;
   public _mainHelperSearch?: AlgoliaSearchHelper['search'];
   public routing?: Routing;
-  public routingManager?;
+  private _routingManager?;
 
   public constructor(options: InstantSearchOptions) {
     super();
@@ -322,11 +322,11 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     }
 
     if (this.routing) {
-      this.routingManager = new RoutingManager({
+      this._routingManager = new RoutingManager({
         ...this.routing,
         instantSearchInstance: this,
       });
-      this._createURL = this.routingManager.createURL;
+      this._createURL = this._routingManager.createURL;
     } else {
       this._createURL = defaultCreateURL;
     }
@@ -389,7 +389,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     });
 
     if (this.routing) {
-      this.routingManager.setupRouting();
+      this._routingManager.setupRouting();
     }
 
     mainHelper.search();
@@ -430,7 +430,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     this.helper = null;
 
     if (this.routing) {
-      this.routingManager.dispose();
+      this._routingManager.dispose();
     }
   }
 
@@ -464,7 +464,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
   public onStateChange = () => {
     const nextUiState = this.mainIndex.getWidgetState({});
     if (this.routing) {
-      this.routingManager.write({ state: nextUiState });
+      this._routingManager.write({ state: nextUiState });
     }
   };
 
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -139,6 +139,7 @@ class InstantSearch extends EventEmitter {
   public _createURL?(nextState: UiState): string;
   public _mainHelperSearch?: AlgoliaSearchHelper['search'];
   public routing?: Routing;
+  public routingManager?;
 
   public constructor(options: InstantSearchOptions) {
     super();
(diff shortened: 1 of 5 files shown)


src/lib/InstantSearch.ts
```diff
@@ -321,15 +322,11 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     }
 
     if (this.routing) {
-      const routingManager = new RoutingManager({
+      this.routingManager = new RoutingManager({
```
```suggestion
      this._routingManager = new RoutingManager({
```
We aren't consistent with public/private yet, but we want to keep currently hidden things hidden

src/lib/RoutingManager.ts
```diff
@@ -35,139 +34,44 @@ class RoutingManager implements Widget {
...
-    searchParameters: SearchParameters;
-  }): UiState {
-    const widgets = this.instantSearchInstance.mainIndex.getWidgets();
-    const helper = this.instantSearchInstance.mainIndex.getHelper()!;
-
-    return {
-      [this.indexId]: widgets.reduce<IndexUiState>((state, widget) => {
-        if (!widget.getWidgetState) {
-          return state;
+  public setupRouting(): void {
```
Haroenv: what about `applyDefaultParameters`? It's supposed to be set on initialisation of InstantSearch only, and it's there to apply the search parameters from the URL. If you have another idea that's also cool :)

yannickcr: What about `applyStateFromRoute`?

Haroenv: sounds good too for me! It's of course a little confusing because route isn't passed as an argument, but good name

yannickcr: Agree, since route is passed to the constructor it is kind of OK I guess.

src/widgets/index/index.ts
```diff
@@ -419,29 +406,6 @@ const index = (props: IndexProps): Index => {
           });
         }
       });
-
-      // Hack for backward compatibily with `searchFunction` + `routing`
-      // https://github.com/algolia/instantsearch.js/blob/509513c0feafaad522f6f18d87a441559f4aa050/src/lib/RoutingManager.ts#L113-L130
-      if (localParent === null && isFirstRender) {
```
Haroenv: i don't think this is expected to be removed now?

yannickcr: It's replaced by https://github.com/algolia/instantsearch.js/compare/next...feat/fs-routingmanager?expand=1#diff-1fc36c4ea042d2ecb5a498416e5e0357R366-R369

I discussed it with @samouss and it should work the same way.

Haroenv: Ahh that makes sense. How do I best test it?

yannickcr: You can see this related story: https://deploy-preview-4099--instantsearchjs.netlify.com/stories/?path=/story/instantsearch--with-searchfunction-to-prevent-search
And this unit test: https://github.com/algolia/instantsearch.js/blob/d49ca2bc5caf5a7d85a5a331907766ba25f82743/src/lib/__tests__/RoutingManager-test.ts#L516-L565

src/lib/InstantSearch.ts
```diff
@@ -453,8 +462,10 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
   }
 
   public onStateChange = () => {
-    // @TODO: Provide `nextUiState` to all middlewares (eg. routing)
-    // const nextUiState = this.mainIndex.getWidgetState({});
+    const nextUiState = this.mainIndex.getWidgetState({});
```
francoischalifour: This is computed all the time while it should only be if `routing` is defined. It should have been part of the `if` branch below.

src/lib/RoutingManager.ts
```diff
@@ -35,139 +32,45 @@ class RoutingManager implements Widget {
...
   }
 
-  public init({ state }: { state: SearchParameters }): void {
-    // Store the initial state from the storage to compare it with the state on next renders
-    // in case the `searchFunction` has modified it.
-    this.initState = this.getAllUiStates({
-      searchParameters: state,
-    });
-  }
+  public write({ state }: { state: UiState }) {
```
francoischalifour: For signature consistency:

```suggestion
  public write({ state }: { state: UiState }): void {
```

src/lib/RoutingManager.ts
```diff
@@ -35,139 +32,45 @@ class RoutingManager implements Widget {
...
-    }
+    this.router.write(route);
   }
 
   public dispose({ helper, state }): void {
-    if (this.renderURLFromState) {
-      helper.removeListener('change', this.renderURLFromState);
-    }
-
     if (this.router.dispose) {
```
francoischalifour: In our quest to becoming stricter and more type-prone, should we rather throw earlier when the interface is not respected?

PR #4100, opened 2019-09-02
feat(RoutingManager): update state on route update

yannickcr:
**Summary**

This PR add the event listener to update the state on any route change and keep both synchronized. It also re-add the related unit tests that were removed in https://github.com/algolia/instantsearch.js/pull/4099.

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -389,7 +389,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     });
 
     if (this.routing) {
-      this._routingManager.applyStateFromRoute();
+      this._routingManager.applyStateFromRoute(this.routing.router.read());
     }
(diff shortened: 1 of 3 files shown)


src/lib/InstantSearch.ts
```diff
@@ -389,7 +389,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     });
 
     if (this.routing) {
-      this._routingManager.applyStateFromRoute();
+      this._routingManager.applyStateFromRoute(this.routing.router.read());
```
samouss: We already have read the URL at the creation of the RoutingManager. I don't think we have to read again from the router since the value is already inside the `initialUiState`. Thoughts?

francoischalifour: That would make sense to me to use `initialUiState`.

PR #4101, opened 2019-09-03
chore(typescript): migrate connectNumericMenu

Haroenv:
I didn't change the behaviour here at all, but there seems to be something either wrong in the helper types, or in its usage here.

This PR is less important to review than other open ones.


src/connectors/numeric-menu/__tests__/connectNumericMenu-test.js
```diff
@@ -78,7 +78,10 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/numeric-men
       const unmount = jest.fn();
 
       const customNumericMenu = connectNumericMenu(render, unmount);
-      const widget = customNumericMenu({ attribute: 'facet', items: [] });
+      const widget = customNumericMenu({
+        attribute: 'facet',
+        items: [{ label: 'x', start: 0 }],
```
Haroenv: this is due to us now also throwing on empty array

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -233,12 +248,20 @@ export default function connectNumericMenu(renderFn, unmountFn = noop) {
 
         return withMaxRefinement;
       },
+
+      // @TODO: remove this (used in widget test now)
+      _getRefine() {
```
Haroenv: this was simpler than refactoring the tests

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -233,12 +248,20 @@ export default function connectNumericMenu(renderFn, unmountFn = noop) {
...
     };
   };
-}
+};
 
-function isRefined(state, attribute, option) {
-  const currentRefinements = state.getNumericRefinements(attribute);
+function isRefined(state: SearchParameters, attribute: string, option: Option) {
+  // @TODO: same as another spot, why is this mixing arrays & elements?
+  const currentRefinements = state.getNumericRefinements(
```
Haroenv: getNumericRefinements returns an array (according to typings), but we use it as a single OperatorList, which one is correct?

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -260,15 +283,22 @@ function isRefined(state, attribute, option) {
...
+  facetValue: string
+) {
   let resolvedState = state;
 
-  const refinedOption = JSON.parse(window.decodeURI(facetValue));
+  const refinedOption = JSON.parse((window as any).decodeURI(facetValue));
 
-  const currentRefinements = resolvedState.getNumericRefinements(attribute);
+  // @TODO: why is array / element mixed here & hasRefinements; seems wrong?
+  const currentRefinements = resolvedState.getNumericRefinements(
```
Haroenv: getNumericRefinements returns an array (according to typings), but we use it as a single OperatorList, which one is correct?

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -154,10 +165,10 @@ export default function connectNumericMenu(renderFn, unmountFn = noop) {
       render({ results, state, instantSearchInstance }) {
         renderFn(
           {
-            createURL: this._createURL(state),
-            items: transformItems(this._prepareItems(state)),
+            createURL: connectorState.createURL!(state),
+            items: transformItems(connectorState.prepareItems!(state)),
             hasNoResults: results.nbHits === 0,
-            refine: this._refine,
+            refine: connectorState.refine!,
```
tkrugg: why not use getRefine() here ?

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -4,146 +4,157 @@ import {
...
+   */
+  attribute: string;
+  /**
+   * List of all the items
+   */
+  items: Option[];
+  /**
+   * Function to transform the items passed to the templates
+   */
+  transformItems: ParamTransformItems;
```
tkrugg: This should be made optional, no?

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -4,146 +4,157 @@ import {
...
 
-    if (!attribute) {
+    if (attribute === '') {
       throw new Error(withUsage('The `attribute` option is required.'));
     }
 
-    if (!items) {
+    if (!items || items.length === 0) {
       throw new Error(
         withUsage('The `items` option expects an array of objects.')
```
tkrugg: The error message here is not longer compatible with the condition:
when `items = []`, it says: "we expect an array"

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -4,146 +4,157 @@ import {
...
+        connectorState.createURL = state => facetValue =>
+          createURL(refine(state, attribute, facetValue));
 
-        this._prepareItems = state =>
+        connectorState.prepareItems = state =>
           items.map(({ start, end, label }) => ({
             label,
-            value: window.encodeURI(JSON.stringify({ start, end })),
-            isRefined: isRefined(state, attribute, { start, end }),
+            value: (window as any).encodeURI(JSON.stringify({ start, end })),
```
tkrugg: is as any really needed to call encoreURI ?

tkrugg: I've removed it an it looks like it's still compiling fine.

src/connectors/numeric-menu/connectNumericMenu.ts
Haroenv: We need ! here because the object is initialized empty, and thus all keys are optional, some better options are outlined in #4102

src/connectors/numeric-menu/connectNumericMenu.ts
Haroenv:
```suggestion
  transformItems?: ParamTransformItems;
```

src/connectors/numeric-menu/connectNumericMenu.ts
Haroenv: wasn't for me, would need to double check

francoischalifour:
This PR needs to be rebased.

Haroenv:
done @francoischalifour

src/types/connector.ts
```diff
@@ -3,6 +3,9 @@ import { Hits, InstantSearch } from './instantsearch';
 import { InsightsClient } from './insights';
 
 export type RendererOptions<TWidgetParams = unknown> = {
+  /**
+   * Original parameters for this widget, in case rendering depends on them.
```
francoischalifour:
```suggestion
   * Original parameters for this widget.
   * Useful for giving back the render parameters to the renderer.
```

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -4,146 +4,157 @@ import {
...
+            isRefined: isRefined(state, attribute, { start, end, label }),
           }));
 
         renderFn(
           {
-            createURL: this._createURL(helper.state),
-            items: transformItems(this._prepareItems(helper.state)),
+            createURL: connectorState.createURL(helper.state),
+            items: transformItems(connectorState.prepareItems!(helper.state)),
```
francoischalifour: `!` shouldn't be required here.

```suggestion
            items: transformItems(connectorState.prepareItems(helper.state)),
```

Haroenv:
I removed those tests, because they were written in a too implementation-dependent way, as well as being duplicate to the connector ones

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -4,146 +4,157 @@ import {
...
           helper.setState(refinedState).search();
         };
 
-        this._createURL = state => facetValue =>
-          createURL(refine(state, attribute, items, facetValue));
+        connectorState.createURL = state => facetValue =>
+          createURL(refine(state, attribute, facetValue));
 
-        this._prepareItems = state =>
+        connectorState.prepareItems = state =>
```
francoischalifour: Why is this stored on the connector state? It's a util function so we can pass both `state` and `items` as arguments and remove it from the connector scope.

src/connectors/numeric-menu/connectNumericMenu.ts
Haroenv: I don't really think this is a problem

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -0,0 +1,378 @@
...
+          },
+          true
+        );
+      },
+
+      render({ results, state, instantSearchInstance }) {
+        renderFn(
+          {
+            createURL: connectorState.createURL!(state),
+            items: transformItems(prepareItems!(state)),
```
francoischalifour:
```suggestion
            items: transformItems(prepareItems(state)),
```

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -0,0 +1,378 @@
...
+        withUsage('The `items` option expects an array of objects.')
+      );
+    }
+
+    type ConnectorState = {
+      refine?: Refine;
+      createURL?: (state: SearchParameters) => (facetValue: string) => string;
+    };
+
+    const prepareItems = (state: SearchParameters) =>
```
francoischalifour: This could actually be outside of the connector itself by passing `items` in the arguments.

PR #4104, opened 2019-09-03
fix(connectInfiniteHits): fix page state when adding or removing widgets

yannickcr:
**Summary**

IFW-980

This PR fix multiple pagination reset issues with InfiniteHits:

* When requesting the `SearchParameters` from the InfiniteHits widget the page was always reset to `0` if `showPrevious` was `false`. We now return the original `searchParameters` (like it was the case before https://github.com/algolia/instantsearch.js/pull/4040).
* When comparing the currentState with the prevState (to know if we need to reset the widget internal cache) we now exlude `hierarchicalFacets`, `disjunctiveFacets` and filter `hierarchicalFacetsRefinements` and `disjunctiveFacetsRefinements` content to only keep the actual refinements. This is very hacky :nauseated_face: . Any idea to make this cleaner and/or more reliable is welcome :wink: .

### DIFF
--- a/src/connectors/infinite-hits/connectInfiniteHits.ts
+++ b/src/connectors/infinite-hits/connectInfiniteHits.ts
@@ -141,7 +141,11 @@ const connectInfiniteHits: InfiniteHitsConnector = (
     };
     const filterEmptyRefinements = (refinements = {}) => {
       return Object.keys(refinements)
-        .filter(key => refinements[key].length)
+        .filter(key =>
+          Array.isArray(refinements[key])
+            ? refinements[key].length
+            : Object.keys(refinements[key]).length
+        )
         .reduce((obj, key) => {
           obj[key] = refinements[key];
           return obj;
@@ -206,6 +210,9 @@ const connectInfiniteHits: InfiniteHitsConnector = (
         );
         currentState.disjunctiveFacetsRefinements = filterEmptyRefinements(
           currentState.disjunctiveFacetsRefinements
+        );
+        currentState.numericRefinements = filterEmptyRefinements(
+          currentState.numericRefinements
         );
 
         if (!isEqual(currentState, prevState)) {
--- a/src/connectors/infinite-hits/connectInfiniteHits.ts
+++ b/src/connectors/infinite-hits/connectInfiniteHits.ts
@@ -139,7 +139,7 @@ const connectInfiniteHits: InfiniteHitsConnector = (
     const getShowMore = (helper: Helper): (() => void) => () => {
       helper.setPage(lastReceivedPage + 1).search();
     };
-    const filterEmptyRefinements = refinements => {
+    const filterEmptyRefinements = (refinements = {}) => {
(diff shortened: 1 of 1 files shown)


src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -186,7 +194,19 @@ const connectInfiniteHits: InfiniteHitsConnector = (
...
+          page = 0,
+          hierarchicalFacets,
+          disjunctiveFacets,
+          ...currentState
+        } = state;
+
+        currentState.hierarchicalFacetsRefinements = filterEmptyRefinements(
+          currentState.hierarchicalFacetsRefinements
+        );
+        currentState.disjunctiveFacetsRefinements = filterEmptyRefinements(
```
Haroenv: how come this isn't necessary for numeric?

yannickcr: Oh, it's necessary for numeric too, just forgot to add it. Fixed in https://github.com/algolia/instantsearch.js/pull/4104/commits/31d7cdba144f3fd9c3205382d6808604ba49727a

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -285,7 +312,11 @@ const connectInfiniteHits: InfiniteHitsConnector = (
           );
         }
 
-        if (hasShowPrevious && uiState.page) {
+        if (!hasShowPrevious) {
+          return widgetSearchParameters;
+        }
```
samouss: It doesn't have the expected behavior anymore. The widget has to set a default value otherwise it gets controlled by its parent. Here is an example of a structure that wouldn't behave like expected:

```js
search.addWidgets([
  hits(),
  pagination(),

  index().addWidgets([
    infiniteHits(),
  ]),
]);
```

Here is the live [example](https://flimsy-treatment.surge.sh/) (CodeSandbox doesn't work with the ES version at the moment). You can select a page inside the `pagination` it shouldn't impact the `infiniteHits` but it does. The widget doesn't have a default value anymore. We have to keep the page of `0` no matter what when the widget is mounted without refinement.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -186,7 +198,22 @@ const connectInfiniteHits: InfiniteHitsConnector = (
...
+
+        currentState.hierarchicalFacetsRefinements = filterEmptyRefinements(
+          currentState.hierarchicalFacetsRefinements
+        );
+        currentState.disjunctiveFacetsRefinements = filterEmptyRefinements(
+          currentState.disjunctiveFacetsRefinements
+        );
+        currentState.numericRefinements = filterEmptyRefinements(
+          currentState.numericRefinements
+        );
```
samouss: I think we forgot to handle `facets` & `facetsRefinements`.

PR #4110, opened 2019-09-11
docs(v4): add migration guide

Haroenv:
This is a temporary location, but will allow us to link towards it until it's in the main documentation.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -96,7 +96,7 @@ The `configure` widget is now included in the UiState. If you want to exclude th
 
 ## Helper
 
-This release includes v3 of the `algoliasearch-helper` package. If you are using the built-in widgets, nothing will change for you. This version no longer includes `lodash`, having two main places of impact for users. The main difference will be in bundle size, which is now significantly smaller. If you are using any methods from the `helper` or `searchResults`, please see the detailed changelog [here](https://github.com/algolia/algoliasearch-helper-js/blob/next/documentation-src/metalsmith/content/upgrade.md).
+This release includes v3 of the `algoliasearch-helper` package. If you are using the built-in widgets, nothing will change for you. This version no longer includes `lodash`, having two main places of impact for users. The main difference will be in bundle size, which is now significantly smaller (`algoliasearch-helper` standalone has gone from 27.5 kB gz to now 9.1 kB gz). If you are using any methods from the `helper` or `searchResults`, please see the detailed changelog [here](https://github.com/algolia/algoliasearch-helper-js/blob/next/documentation-src/metalsmith/content/upgrade.md).
 
 ## Custom widgets
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -113,6 +113,21 @@ const widget = {
       disjunctiveFacets: ['myAttribute'],
     };
   },
+  getWidgetSearchParameters(searchParameters, { uiState }) {
+    return searchParameters
+      .addDisjunctiveFacetRefinement(
+        'myAttribute',
+        uiState.myWidgetName.myAttribute
+      );
+  },
+  getWidgetState(uiState, { searchParameters }) {
+    return {
+      ...uiState,
+      myWidgetName: {
+        myAttribute: searchParameters.getDisjunctiveRefinements('myAttribute')
+      }
+    };
+  }
 };
 ```
 
@@ -138,8 +153,6 @@ const widget = {
   }
 };
 ```
-
-While this looks longer, more of the functions are now together than before, and it isn't possible anymore to get into a state where `getConfiguration` is called multiple times and isn't behaving as you would want.
 
 ## connectAutoComplete
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -45,7 +45,50 @@ Is now:
 
 If you are using the default stateMapping with the current version, you can replace that with `singleIndexStateMapping('yourIndexName')` (import from `instantsearch.js/es/lib/stateMappings`).
 
-If you were using a custom stateMapping, you need to loop over the outer level of index, and add this extra level back in `routeToState`. You can check the [source](https://github.com/algolia/instantsearch.js/blob/next/src/lib/stateMappings/singleIndex.ts) for reference on how to implement this.
+If you were using a custom stateMapping, you need to loop over the outer level of index, and add this extra level back in `routeToState`. You can check the [source](https://github.com/algolia/instantsearch.js/blob/next/src/lib/stateMappings/singleIndex.ts) for reference on how to implement this. A stateMapping where you map only some of the properties for example would change like this:
+
+```js
+// before
+const stateMapping = {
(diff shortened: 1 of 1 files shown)


MIGRATION.md
```diff
@@ -0,0 +1,130 @@
...
+
+search.addWidgets([
+  index({ indexName: "additional" }),
+  autocomplete()
+]);
+```
+
+## onHistoryChange
+
+This was a function which wasn't documented, and has been a no-op since InstantSearch v2, it has now been removed. Listening to changes in the URL can be done by creating a custom `router` and listening in the `write` hook.
```
tkrugg: I think later we'd want to example for this as I don't find it super trivial TBH.

```js
const router = historyRouter()
const originalWrite = router.write.bind(router)
router.write = state => {
    // listening 
    originalWrite(state)
}

{
   routing: {
      router,
      stateMapping: ...
   }
}

```

MIGRATION.md
```diff
@@ -0,0 +1,130 @@
...
+Will be replaced with this:
+
+```js
+const autocomplete = connectAutocomplete(() => {/* ... */});
+
+search.addWidgets([
+  index({ indexName: "additional" }),
+  autocomplete()
+]);
+```
```
tkrugg: as discussed offline, I would love to have an example where this cannot be entirely replaced with a simple searchBox. I'm still missing it.

MIGRATION.md
```diff
@@ -0,0 +1,130 @@
...
+      ...uiState,
+      myWidgetName: {
+        myAttribute: searchParameters.getDisjunctiveRefinements('myAttribute')
+      }
+    };
+  }
+};
+```
+
+While this looks longer, more of the functions are now together than before, and it isn't possible anymore to get into a state where `getConfiguration` is called multiple times and isn't behaving as you would want.
```
tkrugg: > more of the functions are now together than before 
could double check this phrase, I don't understand it maybe it's missing a word?

MIGRATION.md
```diff
@@ -0,0 +1,130 @@
...
+  getWidgetState(uiState, { searchParameters }) {
+    return {
+      ...uiState,
+      myWidgetName: {
+        myAttribute: searchParameters.getDisjunctiveRefinements('myAttribute')
+      }
+    };
+  }
+};
+```
```
tkrugg: Thanks for adding the example! 👍

MIGRATION.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Configure
+
+The `configure` widget is now included in the UiState. If you want to exclude this from the URL (because it's usually static, or it allows users to add arbitrary search parameters), you can either use the default stateMappings which exclude it, or exclude it yourself in a custom state mapping. This has to be done both in `stateToRoute` to prevent it appearing in the URL, and `routeToState` to prevent it from applying from the URL. You can check the [source](https://github.com/algolia/instantsearch.js/blob/next/src/lib/stateMappings/simple.ts) for inspiration.
+
+## Helper
+
+This release includes v3 of the `algoliasearch-helper` package. If you are using the built-in widgets, nothing will change for you. This version no longer includes `lodash`, having two main places of impact for users. The main difference will be in bundle size, which is now significantly smaller. If you are using any methods from the `helper` or `searchResults`, please see the detailed changelog [here](https://github.com/algolia/algoliasearch-helper-js/blob/next/documentation-src/metalsmith/content/upgrade.md).
```
tkrugg: Might be worth mentioning the bundlesize difference

MIGRATION.md
```diff
@@ -0,0 +1,130 @@
+# Migrating to InstantSearch v4
+
+This latest version of InstantSearch is focused mostly around Federated Search, as well as bundle size. To achieve this, we have made some changes which impact your apps. These changes are detailed here:
```
tkrugg: 👍

MIGRATION.md
```diff
@@ -0,0 +1,130 @@
...
+  "indexName": {
+    "query": "value",
+    "page": 5
+  }
+}
+```
+
+If you are using the default stateMapping with the current version, you can replace that with `singleIndexStateMapping('yourIndexName')` (import from `instantsearch.js/es/lib/stateMappings`).
+
+If you were using a custom stateMapping, you need to loop over the outer level of index, and add this extra level back in `routeToState`. You can check the [source](https://github.com/algolia/instantsearch.js/blob/next/src/lib/stateMappings/singleIndex.ts) for reference on how to implement this.
```
eunjae-lee: What do you think of having a small example for before/after?

```js
// before
const stateMapping = {
  stateToRoute(uiState) {
    return {
      query: uiState.query,
      page: uiState.page,
      // ...
    };
  },

  routeToState(routeState) {
    return {
      query: routeState.query,
      page: routeState.page,
      // ...
    };
  },
};

// after
const stateMapping = {
  stateToRoute(uiState) {
    const indexUiState = uiState[indexName];
    return {
      query: indexUiState.query,
      page: indexUiState.page,
      // ...
    };
  },

  routeToState(routeState) {
    return {
      [indexName]: {
        query: routeState.query,
        page: routeState.page,
        // ...
      },
    };
  },
};
```

MIGRATION.md
Haroenv: Would add this, but not sure if it’s used

MIGRATION.md
Haroenv: makes sense, I'll add this example

MIGRATION.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Helper
+
+This release includes v3 of the `algoliasearch-helper` package. If you are using the built-in widgets, nothing will change for you. This version no longer includes `lodash`, having two main places of impact for users. The main difference will be in bundle size, which is now significantly smaller. If you are using any methods from the `helper` or `searchResults`, please see the detailed changelog [here](https://github.com/algolia/algoliasearch-helper-js/blob/next/documentation-src/metalsmith/content/upgrade.md).
+
+## Custom widgets
+
+As mentioned earlier, any reference of the helper now contains the "new helper", as well as the `getConfiguration` life cycle no longer being used. You can replace its usage with `getWidgetSearchParameters` and `getWidgetState`. Note that this means that your custom widget will also take part in routing from then on. You can still exclude it from the URL via `stateMapping`.
```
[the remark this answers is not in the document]
Haroenv: both are correct but mean something else, but `to` might make more sense

MIGRATION.md
eunjae-lee: What about attaching a link to the source code of the storybook and its live example?
I just tried storybook and I think it explain well how it works.

Haroenv: you have access to the hits of both of the indices in autocomplete, so it's not the same as a search box (if the autocomplete component you're wrapping is using both hits & query at the same time, which is common)

MIGRATION.md
Haroenv: Agree the sentence isn't that good, do you know what it should say?

PR #4117, opened 2019-09-16
fix(RefinementList 4.x): remove root css class on sublists

tkrugg:
⚠️ this is port of #4025 on the next branch, to be released in 4.x
 
 **Summary**

The RefinementList is the component used for rendering of the HierarchicalMenu widget.
When a given item of the HierarchicalMenu has children, the RefinementList is recursively called, with the same cssClasses prop.

As a result, the `cssClasses.root` is appearing on every list of children.

This change is proposed because:

1. This is not the specified behaviour.
2. This is not the behaviour observed on InstantSearch V2.
  see difference between [InstantSearch V3](https://codesandbox.io/s/ais-ecommerce-demo-app-8snic) and [InstantSearch V2](https://codesandbox.io/s/instantsearchjs-template-7q78t)
  In V2, the `cssClasses.root` belonged to the [headerFooter HOC](https://github.com/algolia/instantsearch.js/blob/v2/src/decorators/headerFooter.js#L22).
  The problem emerged in InstantSearch V3 as the `cssClasses.root` property was moved
  inside the RefinementList widget.

**Result**
```js
throw new Error(NOT_IMPLEMENTED)
```

### DIFF
--- a/src/components/RefinementList/RefinementList.js
+++ b/src/components/RefinementList/RefinementList.js
@@ -30,9 +30,11 @@ class RefinementList extends Component {
     let subItems;
     const hasChildren = facetValue.data && facetValue.data.length > 0;
     if (hasChildren) {
+      const { root, ...cssClasses } = this.props.cssClasses;
       subItems = (
         <RefinementList
           {...this.props}
+          cssClasses={cssClasses}
           depth={this.props.depth + 1}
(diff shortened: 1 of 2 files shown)


francoischalifour:
I cleaned up the Git history to merge this because it's been pending for a while now (since #4025).

PR #4119, opened 2019-09-16
fix(urlSync): remove outdated `urlSync` error

francoischalifour:
We used to notice users coming from InstantSearch v2 that `urlSync` was replaced by `routing` in v3. We can remove this error in InstantSearch v4.


Haroenv:
looks like i broke tsc unrelated to this PR yesterday

PR #4120, opened 2019-09-17
refactor(router): remove widget dependency

samouss:
This PR removes the dependency of `Widget` on the type `Router`. The only method that both have in common is `dispose` but with a different signature. The `Router` doesn't have to receive `Helper`, `State`. The only data structures it has to care about are: `UiState`, `RouteState`. 

The update of the signature is indeed a breaking change but it shouldn't have an impact on user-land. The implementation of a custom router is rare. The usage of the arguments of `dispose` would be even rarer.

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -134,7 +134,7 @@ class InstantSearch extends EventEmitter {
   public _createURL?(nextState: UiState): string;
   public _mainHelperSearch?: AlgoliaSearchHelper['search'];
   public routing?: Routing;
-  private _routingManager?;
+  private _routingManager?: RoutingManager;
 
   public constructor(options: InstantSearchOptions) {
     super();
@@ -396,7 +396,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       uiState: this._initialUiState,
     });
 
-    if (this.routing) {
+    if (this._routingManager && this.routing) {
       this._routingManager.applyStateFromRoute(this.routing.router.read());
     }
 
@@ -437,7 +437,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     this.mainHelper = null;
     this.helper = null;
 
-    if (this.routing) {
+    if (this._routingManager) {
       this._routingManager.dispose();
     }
   }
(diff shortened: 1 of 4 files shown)


src/types/instantsearch.ts
```diff
@@ -116,6 +116,11 @@ export interface Router<TRouteState = UiState> extends Widget {
    * return a string. It may return an empty string.
    */
   createURL(state: TRouteState): string;
+
+  /**
+   * Called when InstantSearch is disposed. Used to remove subscriptions.
+   */
+  dispose?(): void;
```
samouss: We could make it required though since all the other methods are required.

src/lib/InstantSearch.ts
```diff
@@ -134,7 +134,7 @@ class InstantSearch extends EventEmitter {
   public _createURL?(nextState: UiState): string;
   public _mainHelperSearch?: AlgoliaSearchHelper['search'];
   public routing?: Routing;
-  private _routingManager?;
+  private _routingManager?: RoutingManager;
```
samouss: The type was inferred as `any`.

src/lib/RoutingManager.ts
```diff
@@ -70,9 +70,9 @@ class RoutingManager {
     this.router.write(route);
   }
 
-  public dispose({ helper, state }): void {
+  public dispose(): void {
     if (this.router.dispose) {
-      this.router.dispose({ helper, state });
```
Haroenv: we need to mention this in the migration guide

src/types/instantsearch.ts
Haroenv: agreed, seems like a good method to make required

samouss: https://github.com/algolia/instantsearch.js/pull/4120/commits/fe3e21315dea1f00bcc12b18158c8f8721553c11

src/lib/RoutingManager.ts
```diff
@@ -70,10 +70,8 @@ class RoutingManager {
     this.router.write(route);
   }
 
-  public dispose({ helper, state }): void {
-    if (this.router.dispose) {
```
samouss: I've removed the guard since now the method is required. Note that he method might be missing since the router could be implemented in user-land. At the moment it's fine since we already use the others function without a check: `read`, `write`, `createURL`, ... We can add validation at some point on development mode.

samouss:
The issue with the test is on `next`. Wait until the branch is fixed.

src/lib/RoutingManager.ts
samouss: https://github.com/algolia/instantsearch.js/pull/4120/commits/dcb3c72f2c70cca582224b8a49683a0e594c6ef2

PR #4122, opened 2019-09-17
refactor(InstantSearch): RoutingManager integration

samouss:
This updates how the `RoutingManager` and `InstantSearch` interact together. 

`InstantSearch` is aware of `stateMapping` and `RouteState` which should not be the case. The only part that is concerned by those details is `RoutingManager`. The `InstantSearch` instance only cares about `UiState`. The `write` part of the  `router` is correctly abstracted behind a `write` method. We have now a `read` function exposed by the manager to avoid manipulate `router` and `stateMapping`.

The subscription to `onUpdate` function is a side effect of `RoutingManager` constructor. While it might be ok, it blocks us to move the declaration of the manager earlier inside the constructor. We already subscribe to the router a bit earlier than before but we don't want to subscribe if the instance does not start. We have now a `subscribe` function exposed by the manager that triggers the subscription to `onUpdate`.

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -398,6 +398,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
 
     if (this._routingManager && this.routing) {
       this._routingManager.applyStateFromRoute(this.routing.router.read());
+      this._routingManager.subscribe();
     }
 
     mainHelper.search();
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -130,10 +130,9 @@ class InstantSearch extends EventEmitter {
   public _searchStalledTimer: any;
   public _isSearchStalled: boolean;
   public _initialUiState: UiState;
+  public _createURL: (nextState: UiState) => string;
   public _searchFunction?: InstantSearchOptions['searchFunction'];
-  public _createURL?(nextState: UiState): string;
   public _mainHelperSearch?: AlgoliaSearchHelper['search'];
-  public routing?: Routing;
   private _routingManager?: RoutingManager;
 
   public constructor(options: InstantSearchOptions) {
@@ -209,6 +208,9 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
     this._searchStalledTimer = null;
     this._isSearchStalled = false;
 
+    this._createURL = defaultCreateURL;
+    this._initialUiState = initialUiState;
+
     if (searchFunction) {
       this._searchFunction = searchFunction;
     }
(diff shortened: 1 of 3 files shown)


src/lib/InstantSearch.ts
```diff
@@ -396,8 +393,9 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       uiState: this._initialUiState,
     });
 
-    if (this._routingManager && this.routing) {
-      this._routingManager.applyStateFromRoute(this.routing.router.read());
+    if (this._routingManager) {
+      this._routingManager.applySearchParameters(this._routingManager.read());
+      this._routingManager.subscribe();
```
samouss: Those two calls are redundant but one of them will be removed in the next PR.

src/lib/InstantSearch.ts
```diff
@@ -130,10 +130,9 @@ class InstantSearch extends EventEmitter {
   public _searchStalledTimer: any;
   public _isSearchStalled: boolean;
   public _initialUiState: UiState;
+  public _createURL: (nextState: UiState) => string;
```
samouss: The function is now required since by default we have a dummy implementation.

src/lib/RoutingManager.ts
```diff
@@ -34,17 +34,13 @@ class RoutingManager {
...
-    this.applyStateFromRoute = this.applyStateFromRoute.bind(this);
-
-    this.router.onUpdate(this.applyStateFromRoute);
+    this.applySearchParameters = this.applySearchParameters.bind(this);
   }
 
-  public applyStateFromRoute(route: UiState): void {
-    const currentUiState = this.stateMapping.routeToState(route);
-
+  public applySearchParameters(uiState: UiState): void {
```
samouss: I've renamed the function to a broader name, we already have the type for more information.

src/lib/InstantSearch.ts
```diff
@@ -218,22 +220,27 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
+    if (routingOptions) {
+      this._routingManager = new RoutingManager({
+        ...routingOptions,
+        instantSearchInstance: this,
+      });
+
+      this._createURL = this._routingManager.createURL;
       this._initialUiState = {
         ...initialUiState,
-        ...this.routing.stateMapping.routeToState(this.routing.router.read()),
```
Haroenv: great call!

src/lib/RoutingManager.ts
```diff
@@ -64,12 +60,26 @@ class RoutingManager {
...
     const route = this.stateMapping.stateToRoute(state);
 
     this.router.write(route);
   }
 
+  public subscribe(): void {
+    this.router.onUpdate(route => {
+      const uiState = this.stateMapping.routeToState(route);
+
+      this.applySearchParameters(uiState);
```
Haroenv: i like that it was moved

PR #4124, opened 2019-09-17
fix(pagination): update no refinement behavior

francoischalifour:
_This targets the `next` branch because this is a breaking change._

## Why

The logic behind the `noRefinement` CSS class differs from one flavor to another. Right now, `noRefinement` is added when on the first page of the search. This doesn't make sense and is incorrect. We want to add the CSS class when no refinements are possible (i.e. there's a single page).

Aligning these behaviors will make creating CSS themes easier because we can hide the pagination widget in CSS when there's no refinement possible. This is a request that we've had several times and using a Panel is necessary for this in InstantSearch.js. We should align these behaviors before next major releases.

## Inconsistencies

There are inconsistencies in the way we consider Pagination to be refinable in our flavors.

- InstantSearch.js: we add the `noRefinement` CSS class when on the first page ❌ ([see source](https://github.com/algolia/instantsearch.js/blob/9bb18cf6f9072044666c303b02496af8b9981f0f/src/components/Pagination/Pagination.js#L116))

- React InstantSearch: we add the `noRefinement` CSS class when it's impossible to refine ✅ ([see source](https://github.com/algolia/react-instantsearch/blob/7e34e69c59bb630ff710e6286120ec47a7171679/packages/react-instantsearch-dom/src/components/Pagination.js#L167))

- Vue InstantSearch: we don't add the `noRefinement` CSS class ❌ ([see source](https://github.com/algolia/vue-instantsearch/blob/3c8c6f18b6c8c3cfa427237c27e5755f1f12ffae/src/components/Pagination.vue#L4))

- Angular InstantSearch: we don't add the noRefinement CSS class ❌ ([see source](https://github.com/algolia/angular-instantsearch/blob/11ad85256dffcc3ea223a3aacec7933f103ff61e/src/pagination/pagination.ts#L11))

## This solution

The CSS class is computed based on the `nbPages` given by the connector. I didn't introduce another rendering option (e.g. `canRefine`) because we're not consistent yet.

## Related

- Pagination spec: https://instantsearch-css.netlify.com/widgets/pagination/

### DIFF
--- a/src/components/Pagination/Pagination.js
+++ b/src/components/Pagination/Pagination.js
@@ -113,7 +113,7 @@ class Pagination extends Component {
     return (
       <div
         className={cx(this.props.cssClasses.root, {
-          [this.props.cssClasses.noRefinementRoot]: this.props.isFirstPage,
+          [this.props.cssClasses.noRefinementRoot]: this.props.nbPages <= 1,
         })}
       >
         <ul className={this.props.cssClasses.list}>
--- a/src/components/Pagination/__tests__/Pagination-test.js
+++ b/src/components/Pagination/__tests__/Pagination-test.js
@@ -45,6 +45,15 @@ describe('Pagination', () => {
   it('should display the first/last link', () => {
     const wrapper = mount(<Pagination {...defaultProps} showFirst showLast />);
 
+    expect(wrapper.find('.firstPageItem')).toHaveLength(1);
+    expect(wrapper.find('.lastPageItem')).toHaveLength(1);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should add the noRefinement CSS class with a single page', () => {
+    const wrapper = mount(<Pagination {...defaultProps} nbPages={1} />);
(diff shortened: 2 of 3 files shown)


src/components/Pagination/Pagination.js
```diff
@@ -113,7 +113,7 @@ class Pagination extends Component {
     return (
       <div
         className={cx(this.props.cssClasses.root, {
-          [this.props.cssClasses.noRefinementRoot]: this.props.isFirstPage,
+          [this.props.cssClasses.noRefinementRoot]: this.props.nbPages <= 1,
```
Haroenv: can this be returned by the connector?

Haroenv:
this should be mentioned in the migration guide

src/components/Pagination/Pagination.js
francoischalifour: As stated in the description, I didn't change the connector widget render option because we're not consistent yet in this flavor.

PR #4126, opened 2019-09-18
fix(sortBy): ensure a return value for getWidgetSearchParameters

samouss:
This PR ensures that the `sortBy` widget always return a value for `index` (or at least don't remove the current one). The call to `getWidgetSearchParameters` might happen once before the `init` step. The current implementation relies on the `init` step (`this.initialIndex`) when no value is found in the `uiState`. When the widget is not initialized the value set is `undefined` which means that it removes the current index.

To avoid this problem we fall back on the current index when none of the conditions are met. It's the responsibility of the widget to set a value. It doesn't make sense to return `SearchParameters` without an index.

### DIFF
--- a/src/connectors/sort-by/__tests__/connectSortBy-test.js
+++ b/src/connectors/sort-by/__tests__/connectSortBy-test.js
@@ -530,7 +530,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/sort-by/js/
         );
       });
 
-      test('should enforce the default value on empty UiState', () => {
+      test('should retrun the initial index on empty UiState with widget initialized', () => {
         const [widget, helper, refine] = getInitializedWidget();
 
         refine('other');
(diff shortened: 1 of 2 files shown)


src/connectors/sort-by/__tests__/connectSortBy-test.js
```diff
@@ -530,7 +530,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/sort-by/js/
         );
       });
 
-      test('should enforce the default value on empty UiState', () => {
+      test('should retrun the initial index on empty UiState with widget initialized', () => {
```
eunjae-lee:
```suggestion
      test('should return the initial index on empty UiState with widget initialized', () => {
```

src/connectors/sort-by/__tests__/connectSortBy-test.js
```diff
@@ -549,6 +549,32 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/sort-by/js/
           })
         );
       });
+
+      test('should retrun the current index on empty UiState without widget initialized', () => {
```
eunjae-lee:
```suggestion
      test('should return the current index on empty UiState without widget initialized', () => {
```

PR #4128, opened 2019-09-18
refactor(showMore): introduce `withShowMore` enhancer

francoischalifour:
_This is a proposal for sharing and composing functionalities across connectors._

## Description

The logic for our "show more" feature is kind of all over the place and is part of the connector logic itself:

- Binds `toggleShowMore` to get a single function reference
- Takes care of keeping track of the `isShowingMore` state
- Pollutes the widget scope: `isShowingMore`, `toggleShowMore`, `cachedToggleShowMore`, `createToggleShowMore`, `getLimit` ([see source](https://github.com/algolia/instantsearch.js/blob/531bbe883ab956063db6d7af59840b92addb8fff/src/connectors/hierarchical-menu/connectHierarchicalMenu.js#L98-L116))

https://github.com/algolia/instantsearch.js/blob/531bbe883ab956063db6d7af59840b92addb8fff/src/connectors/hierarchical-menu/connectHierarchicalMenu.js#L98-L116

This PR introduces a new pattern for subscribing a widget to a "show more" behavior with a dedicated logic.

This refactors the usage in:

- `connectRefinementList`
- `connectMenu`
- `connectHierarchicalMenu`

## Usage

```ts
function connectHierarchicalMenu(render, unmount) {
  // Instantiate the higher-order function
  const withShowMore = createWithShowMore();

  return widgetParams => {
    // ...

    return withShowMore(
      ({ toggleShowMore, isShowingMore, onToggleShowMore }) => ({
        // ...

        render(renderProps) {
          onToggleShowMore(this.render.bind(this, renderProps));

          // ...
        },
      })
    );
  };
}
```

## Alternative names

The current working name of this category of features is `enhancers`.

Other names I have in mind:

- `wrappers`
- `hofs` (higher-order functions)

Feel free to suggest some more.

## Next steps

If you'd like to use this pattern in our codebase, it can be composed with other enhancers like `withSearchForFacetValues`:

```ts
function connectHierarchicalMenu(render, unmount) {
  // Instantiate the higher-order function
  const withShowMore = createWithShowMore();
  const withSearchForFacetValues = createWithSearchForFacetValues();

  return widgetParams => {
    // ...

    return withSearchForFacetValues(
      ({ searchForFacetValues, onSearchForFacetValues }) =>
        withShowMore(({ toggleShowMore, isShowingMore, onToggleShowMore }) => ({
          // Widget definition
        }))
    );
  };
}
```


src/lib/enhancers/withShowMore.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  function toggleShowMore() {
+    state.isShowingMore = !state.isShowingMore;
+
+    if (onToggle) {
+      onToggle();
+    }
+  }
+
+  function onToggleShowMore(listener: Listener) {
```
eunjae-lee: `setOnToggleShowMore` wdyt?

src/connectors/menu/connectMenu.js
```diff
@@ -114,165 +117,148 @@ export default function connectMenu(renderFn, unmountFn = noop) {
...
-      getLimit() {
-        return this.isShowingMore ? showMoreLimit : limit;
-      },
-
-      refine(helper) {
-        return facetValue => {
-          const [refinedItem] = helper.getHierarchicalFacetBreadcrumb(
-            attribute
+    return withShowMore(
+      ({ toggleShowMore, isShowingMore, onToggleShowMore }) => ({
```
Haroenv: I'd rather have a way to share behaviour, which can influence every part of the widget, including search parameters. As @samouss said, the issue is broader than just here, and thus might not be the pattern that fits us best

francoischalifour:
Thanks for the feedback!

To be clear about this proposal, it's not meant to be merged as is but is meant to understand what we all have in mind.

> The `showMore` abstraction should manage the feature from A to Z. Here we only abstract the UI state management of `showMore` but it's not only about visible/hidden. We also have to change the number of `items` rendered on the page. This logic is still managed by the connector.

That precisely what I had in mind by "sharing functionalities". I didn't want to introduce some magic helper that changes the connector logic as a blackbox, but rather give helpers to abstract the logic, which can be complex ([check `connectRefinementList` source code...](https://github.com/algolia/instantsearch.js/tree/90ad88ca5de40b5cc05a8fa05c08cabd783bfa1d/src/connectors/refinement-list)). So in summary, give the minimal amount of information to create a `showMore` behavior in the connectors themselves.

> IMO the current state brings a bit more indirection about how the `showMore` logic works. It is now spread across different modules.

If we introduce an abstraction (which is what we want to do), it will create indirection anyway. What solution do you have in mind that would not do so?

> @tkrugg implemented a similar solution for insights with the [`withInsights`](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/insights/client.ts#L100-L128) function. It's a function that augments the capabilities of the render function. This concept would be a better fit I think for this use case. I don't say that it will perfectly fit but the concept looks more appropriate. We can read values, update values, add behaviors to the render function.

That's right, I wanted to try another approach here where you're more in control of what the connector does. The approach with `withInsights` resembles the patterns we have in React InstantSearch, which are _not_ easy to build on top of IMO. Thus, I tried another approach on purpose.

> Before any implementation of such solutions, we should rethink the render process (`init`/`render`, `isFirstRender`, etc...). Those problems are part of the public API. You can't avoid them when connectors are used. I agree that would be nice to share more behavior between our connectors, but right now it's not a critical problem. We don't build new widgets every weeks, neither our users. Once we have a simpler/streamlined render process, solutions to share behavior across connectors will be simpler to think about.

That's for sure! That proposal is only for sharing what I had in mind that day so that we iterate, share everybody's opinions, background-think and gather the pros/cons of different approaches. Again, this is not meant to be merged as is.

I'm not sure we tackle the removal of the `init` hook before introducing `searchFoFacetValues` in `Menu` though, which is why I started thinking about new patterns (we don't want `connectMenu` to be as messy as `connectRefinementList`).

src/connectors/menu/connectMenu.js
francoischalifour: This is precisely what _I didn't want_ to do in this proposal – I tried another approach (see https://github.com/algolia/instantsearch.js/pull/4128#issuecomment-533633177). Not saying that one is better than another, just wanted to try something new and trigger discussion.

Here I wanted the connector to still be in control of everything with the given information (e.g. "is it showing more right now?"). I tried to avoid the black box architecture that React InstantSearch creates.

francoischalifour:
Closing this until we need a proper solution.

PR #4131, opened 2019-09-19
feat(core): deprecate addWidget & removeWidget

Haroenv:
Migration guide already mentions this since dc75d2fc64c68876d91e9daf847038607c1215aa, discussed with @samouss

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -310,7 +310,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
    * Removes multiple widgets from the search instance.
    * @param widgets Array of widgets instances to remove from InstantSearch.
    *
-  * The widgets must implement a `dispose()` method to clear their states.
+   * The widgets must implement a `dispose()` method to clear their states.
    */
   public removeWidgets(widgets: Widget[]) {
     if (!Array.isArray(widgets)) {
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -293,7 +293,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
 
   /**
    * Removes a widget from the search instance.
-   * @deprecated This method will still be supported in 4.x releases, but not further. It is replaced by `addWidgets([widget])`
+   * @deprecated This method will still be supported in 4.x releases, but not further. It is replaced by `removeWidgets([widget])`
    * @param widget The widget instance to remove from InstantSearch.
    *
    * The widget must implement a `dispose()` method to clear its state.
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -12,6 +12,7 @@ import {
   isPlainObject,
   defer,
   noop,
+  warning,
 } from './utils';
 import {
   InsightsClient as AlgoliaInsightsClient,
@@ -245,23 +246,22 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
   }
 
   /**
-   * Adds a widget. This can be done before and after InstantSearch has been started. Adding a
-   * widget after InstantSearch started is considered **EXPERIMENTAL** and therefore
-   * it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20hot%20addWidget).
-   * @param  widget The widget to add to InstantSearch. Widgets are simple objects
-   * that have methods that map the search life cycle in a UI perspective. Usually widgets are
-   * created by [widget factories](widgets.html) like the one provided with InstantSearch.js.
+   * Adds a widget. This can be done before and after InstantSearch has been started.
+   * @param widget The widget to add to InstantSearch.
+   *
+   * This method is deprecated, and `addWidget(widget)` can be replaced by `addWidgets([widget])`
    */
   public addWidget(widget: Widget) {
+    warning(
+      false,
+      'addWidget(widget) is replaced by addWidgets([widgets]) to help see the structure'
+    );
     this.addWidgets([widget]);
   }
 
   /**
-   * Adds multiple widgets. This can be done before and after the InstantSearch has been started. This feature
-   * is considered **EXPERIMENTAL** and therefore it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20addWidgets).
-   * @param {Widget[]} widgets The array of widgets to add to InstantSearch.
+   * Adds multiple widgets. This can be done before and after the InstantSearch has been started.
+   * @param widgets The array of widgets to add to InstantSearch.
    */
   public addWidgets(widgets: Widget[]) {
     if (!Array.isArray(widgets)) {
@@ -290,20 +290,26 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
   }
 
   /**
-   * Removes a widget. This can be done after the InstantSearch has been started. This feature
-   * is considered **EXPERIMENTAL** and therefore it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20removeWidget).
-   * @param  {Widget} widget The widget instance to remove from InstantSearch. This widget must implement a `dispose()` method in order to be gracefully removed.
+   * Removes a widget. This can be done after the InstantSearch has been started.
+   *
+   * @param widget The widget instance to remove from InstantSearch.
+   *
+   * This widget must implement a `dispose()` method in order to be gracefully removed.
    */
   public removeWidget(widget: Widget) {
+    warning(
+      false,
+      'removeWidget(widget) is replaced by removeWidgets([widgets]) to help see the structure'
+    );
     this.removeWidgets([widget]);
   }
 
   /**
-   * Removes multiple widgets. This can be done only after the InstantSearch has been started. This feature
-   * is considered **EXPERIMENTAL** and therefore it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20addWidgets).
-   * @param {Widget[]} widgets Array of widgets instances to remove from InstantSearch.
+   * Removes multiple widgets. This can be done only after the InstantSearch has been started.
+   *
+   * @param widgets Array of widgets instances to remove from InstantSearch.
+   *
+   * These widget must implement a `dispose()` method in order to be gracefully removed.
    */
   public removeWidgets(widgets: Widget[]) {
     if (!Array.isArray(widgets)) {
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -246,21 +246,23 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
   }
 
   /**
-   * Adds a widget. This can be done before and after InstantSearch has been started.
+   * Adds a widget to the search instance.
+   * A widget can be added either before or after InstantSearch has started.
    * @param widget The widget to add to InstantSearch.
    *
-   * This method is deprecated, and `addWidget(widget)` can be replaced by `addWidgets([widget])`
+   * @deprecated This method will still be supported in 4.x releases, but not further. It is replaced by `addWidgets([widget])`.
    */
   public addWidget(widget: Widget) {
     warning(
       false,
-      'addWidget(widget) is replaced by addWidgets([widgets]) to help see the structure'
+      'addWidget will still be supported in 4.x releases, but not further. It is replaced by `addWidgets([widget])`'
     );
     this.addWidgets([widget]);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/lib/InstantSearch.ts
```diff
@@ -245,23 +246,22 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
-   * created by [widget factories](widgets.html) like the one provided with InstantSearch.js.
+   * Adds a widget. This can be done before and after InstantSearch has been started.
+   * @param widget The widget to add to InstantSearch.
+   *
+   * This method is deprecated, and `addWidget(widget)` can be replaced by `addWidgets([widget])`
    */
   public addWidget(widget: Widget) {
+    warning(
+      false,
+      'addWidget(widget) is replaced by addWidgets([widgets]) to help see the structure'
```
samouss:
```suggestion
      '`addWidget(widget)` is deprecated, prefer the use of `addWidgets([widgets])`'
```

Would be nicer for the user. Or something similar.

src/lib/InstantSearch.ts
```diff
@@ -290,20 +290,26 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
   }
 
   /**
-   * Removes a widget. This can be done after the InstantSearch has been started. This feature
-   * is considered **EXPERIMENTAL** and therefore it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20removeWidget).
-   * @param  {Widget} widget The widget instance to remove from InstantSearch. This widget must implement a `dispose()` method in order to be gracefully removed.
+   * Removes a widget. This can be done after the InstantSearch has been started.
+   *
```
samouss:
```suggestion

```

src/lib/InstantSearch.ts
```diff
@@ -290,20 +290,26 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
     this.removeWidgets([widget]);
   }
 
   /**
-   * Removes multiple widgets. This can be done only after the InstantSearch has been started. This feature
-   * is considered **EXPERIMENTAL** and therefore it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20addWidgets).
-   * @param {Widget[]} widgets Array of widgets instances to remove from InstantSearch.
+   * Removes multiple widgets. This can be done only after the InstantSearch has been started.
+   *
```
samouss:
```suggestion

```

src/lib/InstantSearch.ts
```diff
@@ -245,23 +246,22 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
 
   /**
-   * Adds a widget. This can be done before and after InstantSearch has been started. Adding a
-   * widget after InstantSearch started is considered **EXPERIMENTAL** and therefore
-   * it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20hot%20addWidget).
-   * @param  widget The widget to add to InstantSearch. Widgets are simple objects
-   * that have methods that map the search life cycle in a UI perspective. Usually widgets are
-   * created by [widget factories](widgets.html) like the one provided with InstantSearch.js.
+   * Adds a widget. This can be done before and after InstantSearch has been started.
```
francoischalifour:
```suggestion
   * Adds a widget to the search instance.
   * A widget can be added either before or after InstantSearch has started.
```

We could even add more information here. Example:

- Adding a widget sets the according search parameters to communicate with the server.
- A widget is responsible for setting and reading the UI state.

src/lib/InstantSearch.ts
```diff
@@ -245,23 +246,22 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
+   * @param widget The widget to add to InstantSearch.
+   *
+   * This method is deprecated, and `addWidget(widget)` can be replaced by `addWidgets([widget])`
```
francoischalifour:
```suggestion
   * @deprecated This method will still be supported in 4.x releases, but not further. It is replaced by `addWidgets([widget])`.
```

src/lib/InstantSearch.ts
```diff
@@ -290,20 +290,26 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
   /**
-   * Removes a widget. This can be done after the InstantSearch has been started. This feature
-   * is considered **EXPERIMENTAL** and therefore it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20removeWidget).
-   * @param  {Widget} widget The widget instance to remove from InstantSearch. This widget must implement a `dispose()` method in order to be gracefully removed.
+   * Removes a widget. This can be done after the InstantSearch has been started.
+   *
+   * @param widget The widget instance to remove from InstantSearch.
+   *
+   * This widget must implement a `dispose()` method in order to be gracefully removed.
```
francoischalifour:
```suggestion
   * The widget must implement a `dispose()` method to clear its state.
```

src/lib/InstantSearch.ts
francoischalifour:
```suggestion
   *
   * @deprecated This method will still be supported in 4.x releases, but not further. It is replaced by `removeWidgets([widget])`.
```

src/lib/InstantSearch.ts
```diff
@@ -245,23 +246,22 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
+    );
     this.addWidgets([widget]);
   }
 
   /**
-   * Adds multiple widgets. This can be done before and after the InstantSearch has been started. This feature
-   * is considered **EXPERIMENTAL** and therefore it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20addWidgets).
-   * @param {Widget[]} widgets The array of widgets to add to InstantSearch.
+   * Adds multiple widgets. This can be done before and after the InstantSearch has been started.
```
francoischalifour:
```suggestion
   * Adds multiple widgets to the search instance.
   * Widgets can be added either before or after InstantSearch has started.
```

src/lib/InstantSearch.ts
```diff
@@ -245,23 +246,22 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
+   * @param widgets The array of widgets to add to InstantSearch.
    */
   public addWidgets(widgets: Widget[]) {
```
francoischalifour: What about this syntax?

```ts
// Signature
addWidgets(...widgets: Widget[])

// Usage
search.addWidgets(searchBox, hits, pagination)
```

src/lib/InstantSearch.ts
```diff
@@ -290,20 +290,26 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
   }
 
   /**
-   * Removes a widget. This can be done after the InstantSearch has been started. This feature
-   * is considered **EXPERIMENTAL** and therefore it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20removeWidget).
-   * @param  {Widget} widget The widget instance to remove from InstantSearch. This widget must implement a `dispose()` method in order to be gracefully removed.
+   * Removes a widget. This can be done after the InstantSearch has been started.
```
francoischalifour:
```suggestion
   * Removes a widget from the search instance.
```

src/lib/InstantSearch.ts
```diff
@@ -290,20 +290,26 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
+    );
     this.removeWidgets([widget]);
   }
 
   /**
-   * Removes multiple widgets. This can be done only after the InstantSearch has been started. This feature
-   * is considered **EXPERIMENTAL** and therefore it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20addWidgets).
-   * @param {Widget[]} widgets Array of widgets instances to remove from InstantSearch.
+   * Removes multiple widgets. This can be done only after the InstantSearch has been started.
```
francoischalifour:
```suggestion
   * Removes multiple widgets from the search instance.
```

We can also mention that it clears the state that the widget is responsible for.

src/lib/InstantSearch.ts
```diff
@@ -290,20 +290,26 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
+   *
+   * @param widgets Array of widgets instances to remove from InstantSearch.
+   *
+   * These widget must implement a `dispose()` method in order to be gracefully removed.
```
francoischalifour:
```suggestion
  * The widgets must implement a `dispose()` method to clear their states.
```

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -251,12 +251,12 @@ describe('addWidget(s)', () => {
 
     expect(search.mainIndex.getWidgets()).toHaveLength(0);
 
-    search.addWidget(createWidget());
+    expect(() => search.addWidget(createWidget())).toWarnDev();
```
francoischalifour: We should track the message.

francoischalifour:
We shouldn't recommend using `addWidget()` anymore then. There are still some spots (e.g. [`connectGeoSearch`](https://github.com/algolia/instantsearch.js/blob/432f833c9bc3100f89bfafa80b4a14e1862b4517/src/connectors/geo-search/connectGeoSearch.js#L123-L127)).

src/lib/InstantSearch.ts
Haroenv: I'm not sure people can even see these comments, let's not get in a discussion totally changing those just because I removed an "experimental" warning

src/lib/InstantSearch.ts
Haroenv: Interesting idea, but i'm not sure how to not make that a big breaking change for anyone now

Haroenv: Maybe open an issue about it

samouss: Agee with @Haroenv. The idea with 4.x.x is to ease the transition as much as possible. If it was the introduction of the function I would have been for it, but it's not the case. The function is already there for quite some time now. I would prefer stick the current version.

samouss:
> We shouldn't recommend using `addWidget()` anymore then. There are still some spots (e.g. [`connectGeoSearch`](https://github.com/algolia/instantsearch.js/blob/432f833c9bc3100f89bfafa80b4a14e1862b4517/src/connectors/geo-search/connectGeoSearch.js#L123-L127)).

We remove most of the warnings from 3.x.x, the one on `geoSeach` should be removed too. We don't have to update them.

src/lib/InstantSearch.ts
francoischalifour: Yep, just wanted to share this idea. We don't document the array usage as much as the `addWidget` so I figured I'd trigger the discussion.

francoischalifour:
@samouss Done in #4134.

src/lib/InstantSearch.ts
Haroenv:
```suggestion
      'This method will still be supported in 4.x releases, but not further. It is replaced by `addWidgets([widget])`'
```

src/lib/InstantSearch.ts
Haroenv:
```suggestion
   * @deprecated This method will still be supported in 4.x releases, but not further. It is replaced by `addWidgets([widget])`
```

src/lib/InstantSearch.ts
```diff
@@ -290,20 +290,26 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
+   * Removes a widget. This can be done after the InstantSearch has been started.
+   *
+   * @param widget The widget instance to remove from InstantSearch.
+   *
+   * This widget must implement a `dispose()` method in order to be gracefully removed.
    */
   public removeWidget(widget: Widget) {
+    warning(
+      false,
+      'removeWidget(widget) is replaced by removeWidgets([widgets]) to help see the structure'
```
```suggestion
      'removeWidget will still be supported in 4.x releases, but not further. It is replaced by `removeWidgets([widget])`'
```

src/lib/InstantSearch.ts
```diff
@@ -290,20 +292,25 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
   }
 
   /**
-   * Removes a widget. This can be done after the InstantSearch has been started. This feature
-   * is considered **EXPERIMENTAL** and therefore it is possibly buggy, if you find anything please
-   * [open an issue](https://github.com/algolia/instantsearch.js/issues/new?title=Problem%20with%20removeWidget).
-   * @param  {Widget} widget The widget instance to remove from InstantSearch. This widget must implement a `dispose()` method in order to be gracefully removed.
+   * Removes a widget from the search instance.
+   * @deprecated This method will still be supported in 4.x releases, but not further. It is replaced by `addWidgets([widget])`
```
samouss:
```suggestion
   * @deprecated This method will still be supported in 4.x releases, but not further. It is replaced by `removeWidgets([widget])`
```

PR #4133, opened 2019-09-19
Implement partials, render function parameter

glassfordm:
**Summary**

1. Implement the mustache partial feature.
2. Add a render parameter to render functions. 

**Result**

1. Allows scenarios like this:

```
instantsearch.widgets.hits({
      ...
      templates: {
        item: 'Template using {{> partial }}',
        partial: 'Partial template {{variable }}'
      }
  });
```

2. Allows scenarios like this:

```
instantsearch.widgets.hits({
     ...
      templates: {
        item: (function($item, $render) {
          var template = generic_template;
          if ($item.type == "special") {
            template = special_template
          }
          return $render(template);
      })
  });
```

Or this:

```
instantsearch.widgets.hits({
     ...
      templates: {
        item: (function($data, $render) {
          return $render('My template with a {{> partial }}', data.partial_data);
       }),
       partial: partial_template,
      }
  });
```

### DIFF
--- a/src/lib/utils/__tests__/index-test.js
+++ b/src/lib/utils/__tests__/index-test.js
@@ -149,6 +149,82 @@ describe('utils.renderTemplate', () => {
     expect(actual).toBe(expectation);
   });
 
+  it('expect to process templates as string using a partial', () => {
+    const templateKey = 'test';
+    const templates = {
+      test: '{{> partial}}',
+      partial: 'it works with {{type}} using partial',
+    };
+    const data = { type: 'strings' };
+
+    const actual = utils.renderTemplate({
+      templateKey,
+      templates,
+      data,
+    });
+
+    const expectation = 'it works with strings using partial';
+
+    expect(actual).toBe(expectation);
+  });
+
+  it('expect to process templates as function using render', () => {
+    const templateKey = 'test';
+    const templates = {
+      test: (data, render) => render(`it works with {{ type }} using render`),
+    };
+    const data = { type: 'functions' };
+
+    const actual = utils.renderTemplate({
+      templateKey,
+      templates,
+      data,
+    });
+
+    const expectation = 'it works with functions using render';
+
+    expect(actual).toBe(expectation);
+  });
+
+  it('expect to process templates as function using render to render a string using a partial', () => {
+    const templateKey = 'test';
+    const templates = {
+      test: (data, render) =>
+        render(`it works with {{> partial }} using render and partial`),
+      partial: `{{ type }}`,
+    };
+    const data = { type: 'functions' };
+
+    const actual = utils.renderTemplate({
+      templateKey,
+      templates,
+      data,
+    });
+
+    const expectation = 'it works with functions using render and partial';
+
+    expect(actual).toBe(expectation);
+  });
+
+  it('expect to process templates as function using render passing new data context', () => {
+    const templateKey = 'test';
+    const templates = {
+      test: (data, render) =>
+        render(`it works with {{ typeName }} using render and data`, data.type),
+    };
+    const data = { type: { typeName: 'functions' } };
+
+    const actual = utils.renderTemplate({
+      templateKey,
+      templates,
+      data,
+    });
+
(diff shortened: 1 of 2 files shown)


Haroenv:
Is there a reason you prefer this towards "function" templates?

glassfordm:
Perhaps this PR is complicated by the fact that it actually introduces 2 changes. I decided to do it that way because both changes were to the same section of code, so if I made separate PRs they would likely conflict with each other.

My motivation for adding support for partial templates is fairly obvious, since it's the reason behind why they were invented in the first place: it allows the template creator to modularize and reduce duplication within a template (for example, if a template displays a name in several places, it can use a "name" partial template to render it in each case).

My main motivation for adding a "render" parameter to function templates was that I was looking for some way for the hits widget to be able to use a different template for each row based on the data in the row (which is a feature others have asked for in the past). The cleanest way I could come up with was to create a function template that was able to choose the template based on the data and then render it--but to do the rendering, it needed all the original context, much of which wasn't available to the function, only to its caller. So I created a render parameter that allowed the render to happen with all the original context.

src/lib/utils/__tests__/index-test.js
```diff
@@ -149,6 +149,82 @@ describe('utils.renderTemplate', () => {
     expect(actual).toBe(expectation);
   });
 
+  it('expect to process templates as string using a partial', () => {
+    const templateKey = 'test';
+    const templates = {
+      test: '{{> partial}}',
```
tkrugg: worth nothing this can accidentally cause infinite loops if `test: {{> test}}` or partial calls test itself. 
Is there a way it can be prevented?

glassfordm: As far as I can see this is a general issue with Mustache partials more than it is a problem with this PR, and would more properly be handled in the Hogan code, if anywhere.

Still, it might be possible to change the InstantSearch.js renderTemplate() function to keep track of what templates it has rendered and prevent any template from being used more than a few times in a given call chain, or alternatively to keep track of a total recursion depth and impose some limit, if you're interested in me attempting it at the expense of the added complexity.

francoischalifour:
That's an interesting addition to the current templating system.

> The cleanest way I could come up with was to create a function template that was able to choose the template based on the data and then render it--but to do the rendering, it needed all the original context, much of which wasn't available to the function, only to its caller.

This might mean that we didn't modularize our templates sections enough, which is likely. Could you explain what wasn't possible with a function template exactly here?

Eventually, we want to move away from Hogan and rely on a more powerful renderer though. This change would have been gracefully accepted earlier but we don't think it makes sense to add this feature if we want to get rid of Hogan all together soon.

Thanks a lot for your contribution though – feel free to open an issue first next time to make sure that a change is appropriate to land in the codebase.

glassfordm:
> This might mean that we didn't modularize our templates sections enough, which is likely. Could you explain what wasn't possible with a function template exactly here?

What I was trying to do is have two (or more) completely different templates for rendering search hits and choose one based on the value of one of the fields of the item being rendered. I was leaning towards using Hogan templates because they seemed less verbose for fairly complex output, but I guess I'd be open to using a function template instead if that would do what I want. I was also hoping to avoid combining all the templates into a single large function template with an if statement or other logic to choose the relevant part to render.

> Thanks a lot for your contribution though – feel free to open an issue first next time to make sure that a change is appropriate to land in the codebase.

No problem. I thought since I had figured out how to do it, it made sense to submit it as a PR. So you're not interested in either part of the PR? The Mustache partial implementation could be pretty useful until you actually do move away from Hogan, and the render parameter could be adapted to provide something like partials for function templates even after you do move away from Hogan (by changing it so that you pass a template name instead of the full content of the template to the render function).

francoischalifour:
Thanks for the efforts @glassfordm, we really appreciate it.

As explained in https://github.com/algolia/instantsearch.js/pull/4133#issuecomment-534290126, since what you're trying to achieve can be done with JavaScript functions and that we don't plan to invest more in Hogan, we'll close this PR.

Issue #4129, opened 2019-09-18
Allow routing without a search box (regular JS / non-react version)


t0byman:
**Is your feature request related to a problem? Please describe 🙏**
No.

**Describe the solution you'd like 🤔**
I would like to be able to do routing without having a search box on my page.

**Describe alternatives you've considered ✨**
I added routing using the parseURL method. It works great, but only when I have a search box on my page. This is because of `const { query = queryName } = qsModule.parse(
                            location.search.slice(1)
                        );` the location.search part is telling the query to go into the search box so that my results can be filtered. But I want them to be filtered without a search box, when the page loads. Is this possible? If not, here is my feature request :)

**Additional context**
Here is my complete routing code:
```js
routing: {
                router: instantsearch.routers.history({
                    parseURL({ qsModule, location }) {
                        const pathnameMatches = location.pathname.match(/zoeken\/(.*?)\/?$/);
                        const queryName = getQueryName(
                            (pathnameMatches && pathnameMatches[1]) || ''
                        );
                        const { query = queryName } = qsModule.parse(
                            location.search.slice(1)
                        );

                        return {
                            query: decodeURIComponent(query),
                        };
                    }
                }),

                stateMapping: {
                    stateToRoute(uiState) {
                        return {
                            query: uiState.query,
                        };
                    },

                    routeToState(routeState) {
                        return {
                            query: routeState.query,
                        };
                    }
                }
            }
```

francoischalifour:
Mounting a search box widget on the page is necessary for the `query` search parameter to be sent to Algolia. Widgets drive search parameters.

You don't have to display a search box though, you can create a virtual widget without rendering with [`connectSearchBox`](https://www.algolia.com/doc/api-reference/widgets/search-box/js/#connector).

t0byman:
Thanks!

PR #4140, opened 2019-09-23
fix(index): warn for inconsistent UI state in development mode

francoischalifour:
This adds a development warning when the UI state is inconsistent with the widgets mounted on the instance.

## Why

We often have users not understanding why certain parameters are not sent to the API. The reason is because widgets drive search parameters. Certain widgets need to be added to the instance for some search parameters to be applied. See a typical issue in #4129.

On support, we often advise to create "virtual widgets" to activate the search parameter, without displaying the widget on the page. This development warning is meant for them to understand why something might look like broken, *and* provide a solution for them.

I suspect this error to happen even more often with InstantSearch.js v4, where UI state is a predominant concept. My hope is that such a warning reduces to support load that we have regarding this use case.

## Warning preview

[InstantSearch.js]: The UI state for the index "instant_search" is not consistent with the widgets mounted.

This can happen when the UI state is specified via `initialUiState` or `routing` but that the widgets responsible for this state were not added. This results in query parameters not being sent to the API.

To fully reflect the state, some widgets need to be added to the index "instant_search":

- `page` needs one of these widgets: "pagination", "infiniteHits"
- `refinementList` needs one of these widgets: "refinementList"
- `hierarchicalMenu` needs one of these widgets: "hierarchicalMenu"

If you do not wish to display widgets but still want to support their search parameters, you can mount "virtual widgets" that don't render anything:

```
const virtualPagination = connectPagination(() => null);
const virtualRefinementList = connectRefinementList(() => null);
const virtualHierarchicalMenu = connectHierarchicalMenu(() => null);

search.addWidgets([
  virtualPagination({ /* ... */ }),
  virtualRefinementList({ /* ... */ }),
  virtualHierarchicalMenu({ /* ... */ })
]);
```

If you're using custom widgets that do set these query parameters, we recommend using connectors instead.

See https://www.algolia.com/doc/guides/building-search-ui/widgets/customize-an-existing-widget/js/#customize-the-complete-ui-of-the-widgets

## Implementation details

The implementation is at the index level. When the `search` event is triggered by the helper, the index reports if its UI state is not synced with its widgets.

This is detected with the `$$type` property that defines the type of the widget. If a UI state key is present but that none of the mounted widgets support this key, the warning appears.

## Cons of this solution

- The mapping between the UI parameters and the widgets that control them is written manually. We need to maintain that list. Could we make this a private connector API?
- Users creating custom widgets will see this warning because they aren't aware of the `$$type` property. I don't think this is a bad thing because the warning give them clues that they should use connectors instead (this is what we recommend).
- Users creating custom connectors will likely see this message because they aren't aware of the `$$type` property.

## Possible improvements

- We could think of improving this DX later by even checking if the attributes match in the UI state (not possible right now because we cannot read the `attribute` option from outside the widget)

### DIFF
--- a/src/types/widget.ts
+++ b/src/types/widget.ts
@@ -115,7 +115,31 @@ export type UiState = {
  * have at least a `render` or a `init` function.
  */
 export interface Widget {
-  $$type?: string;
+  $$type?:
+    | 'ais.autocomplete'
+    | 'ais.breadcrumb'
+    | 'ais.clearRefinements'
+    | 'ais.configure'
+    | 'ais.currentRefinements'
+    | 'ais.geoSearch'
+    | 'ais.hierarchicalMenu'
+    | 'ais.hits'
+    | 'ais.hitsPerPage'
+    | 'ais.index'
+    | 'ais.infiniteHits'
+    | 'ais.menu'
+    | 'ais.numericMenu'
+    | 'ais.pagination'
+    | 'ais.poweredBy'
+    | 'ais.queryRules'
+    | 'ais.range'
+    | 'ais.ratingMenu'
+    | 'ais.refinementList'
+    | 'ais.searchBox'
+    | 'ais.sortBy'
+    | 'ais.stats'
+    | 'ais.toggleRefinement'
+    | 'ais.voiceSearch';
   /**
    * Called once before the first search
    */
--- a/src/widgets/index/index.ts
+++ b/src/widgets/index/index.ts
@@ -348,6 +348,7 @@ const index = (props: IndexProps): Index => {
         if (__DEV__) {
           // Some connectors are responsible for multiple widgets so we need
           // to map them.
+          // eslint-disable-next-line no-inner-declarations
           function getWidgetNames(connectorName: string): string[] {
             switch (connectorName) {
               case 'range':
--- a/src/widgets/index/index.ts
+++ b/src/widgets/index/index.ts
@@ -346,6 +346,18 @@ const index = (props: IndexProps): Index => {
         instantSearchInstance.scheduleStalledRender();
 
         if (__DEV__) {
+          // Some connectors are responsible for multiple widgets so we need
+          // to map them.
+          function getWidgetNames(connectorName: string): string[] {
+            switch (connectorName) {
+              case 'range':
+                return ['rangeInput', 'rangeSlider'];
+
+              default:
+                return [connectorName];
+            }
+          }
+
           type StateToWidgets = {
             [TParameter in keyof IndexUiState]: Array<Widget['$$type']>;
           };
@@ -370,8 +382,10 @@ const index = (props: IndexProps): Index => {
             .map(widget => widget.$$type)
             .filter(Boolean);
 
+          type MissingWidgets = Array<[string, Array<Widget['$$type']>]>;
+
           const missingWidgets = Object.keys(localUiState).reduce<
-            Array<Partial<StateToWidgets>>
+            MissingWidgets
           >((acc, parameter) => {
             const requiredWidgets: Array<Widget['$$type']> =
               stateToWidgetsMap[parameter];
@@ -381,9 +395,13 @@ const index = (props: IndexProps): Index => {
                 mountedWidgets.includes(requiredWidget)
               )
             ) {
-              acc.push({
-                [parameter]: stateToWidgetsMap[parameter],
-              });
+              acc.push([
+                parameter,
+                stateToWidgetsMap[parameter].map(
+                  (widgetIdentifier: string) =>
+                    widgetIdentifier.split('ais.')[1]
+                ),
+              ]);
             }
 
             return acc;
@@ -398,15 +416,11 @@ This can happen when the UI state is specified via \`initialUiState\` or \`routi
 To fully reflect the state, some widgets need to be added to the index "${this.getIndexId()}":
 
 ${missingWidgets
-  .map(widget => {
-    const stateParameter = Object.keys(widget)[0];
-    const neededWidgets = stateToWidgetsMap[stateParameter]
-      .map(
-        (widgetIdentifier: string) => `"${widgetIdentifier.split('ais.')[1]}"`
-      )
-      .join(', ');
-
-    return `- \`${stateParameter}\` needs one of these widgets: ${neededWidgets}`;
+  .map(([stateParameter, widgets]) => {
+    return `- \`${stateParameter}\` needs one of these widgets: ${([] as string[])
+      .concat(...widgets.map(name => getWidgetNames(name!)))
+      .map((name: string) => `"${name}"`)
+      .join(', ')}`;
   })
   .join('\n')}
 
@@ -414,11 +428,8 @@ If you do not wish to display widgets but still want to support their search par
 
 \`\`\`
 ${missingWidgets
-  .map(widget => {
-    const stateParameter = Object.keys(widget)[0];
-    const capitalizedWidget = capitalize(
-      stateToWidgetsMap[stateParameter][0].split('ais.')[1]
-    );
+  .map(([_stateParameter, widgets]) => {
+    const capitalizedWidget = capitalize(widgets[0]!);
 
     return `const virtual${capitalizedWidget} = connect${capitalizedWidget}(() => null);`;
   })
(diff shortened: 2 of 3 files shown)


src/widgets/index/index.ts
```diff
@@ -342,6 +344,105 @@ const index = (props: IndexProps): Index => {
         // it at the index level because it's either: all of them or none of them
         // that are stalled. The queries are performed into a single network request.
         instantSearchInstance.scheduleStalledRender();
+
+        if (__DEV__) {
+          type StateToWidgets = {
+            [TParameter in keyof IndexUiState]: Array<Widget['$$type']>;
```
francoischalifour: The rule `@typescript-eslint/generic-type-naming` should also allow `^K[A-Z][a-zA-Z]+$` for `KParameter` here.

src/widgets/index/index.ts
```diff
@@ -342,6 +344,105 @@ const index = (props: IndexProps): Index => {
...
+            }
+
+            return acc;
+          }, []);
+
+          warning(
+            missingWidgets.length === 0,
+            `The UI state for the index "${this.getIndexId()}" is not consistent with the widgets mounted.
+
+This can happen when the UI state is specified via \`initialUiState\` or \`routing\` but that the widgets responsible for this state were not added. This results in query parameters not being sent to the API.
```
```suggestion
This can happen when the UI state is specified via \`initialUiState\` or \`routing\` but the widgets responsible for this state were not added. This results in those query parameters not being sent to the API.
```

src/widgets/index/index.ts
```diff
@@ -21,6 +21,8 @@ import {
   createDocumentationMessageGenerator,
   resolveSearchParameters,
   mergeSearchParameters,
+  warning,
+  capitalize,
```
Haroenv: just wondering here, what have we used capitalize for before?

src/widgets/index/index.ts
```diff
@@ -342,6 +344,105 @@ const index = (props: IndexProps): Index => {
...
+            range: ['ais.range'],
+            toggle: ['ais.toggleRefinement'],
+            geoSearch: ['ais.geoSearch'],
+            sortBy: ['ais.sortBy'],
+            page: ['ais.pagination', 'ais.infiniteHits'],
+            hitsPerPage: ['ais.hitsPerPage'],
+            configure: ['ais.configure'],
+          };
+
+          const mountedWidgets = this.getWidgets()
```
Haroenv: since this happens per index, you could get multiple of these warnings right? Another option is to put this whole warning in InstantSearch and travel down the tree to compare. Might be rare enough to give this warning that just here is good

src/widgets/index/index.ts
```diff
@@ -342,6 +344,105 @@ const index = (props: IndexProps): Index => {
...
+            .filter(Boolean);
+
+          const missingWidgets = Object.keys(localUiState).reduce<
+            Array<Partial<StateToWidgets>>
+          >((acc, parameter) => {
+            const requiredWidgets: Array<Widget['$$type']> =
+              stateToWidgetsMap[parameter];
+
+            if (
+              requiredWidgets.every(
```
Haroenv: perf doesn't really matter in dev, but you can swap this condition around and it will be triggered as soon as something is off:

```js
if (requiredWidgets.some(requiredWidget => mountedWidgets.includes(requiredWidget) ===  false) === false)
```

This looks somewhat more ugly, which is why i  used ===  false, instead of !

src/widgets/index/index.ts
francoischalifour: We use it to compute the CurrentRefinements label:

https://github.com/algolia/instantsearch.js/blob/9b0d79f0915bcb0fa9662097530b9448a1538997/src/components/CurrentRefinements/CurrentRefinements.tsx#L44

src/widgets/index/index.ts
francoischalifour: That's right. I first wanted to see if this feature is well received. We can still improve it later.

src/widgets/index/index.ts
francoischalifour: Are you sure about removing "that"? It's harder to read to me.

src/widgets/index/index.ts
francoischalifour: This can actually be simplified to this:

```ts
!requiredWidgets.some(requiredWidget =>
  mountedWidgets.includes(requiredWidget)
)
```

src/types/widget.ts
```diff
@@ -115,7 +115,31 @@ export type UiState = {
  * have at least a `render` or a `init` function.
  */
 export interface Widget {
-  $$type?: string;
+  $$type?:
```
samouss: The type is set at the connector level not at the widget level. Currently this type is incorrect.

src/widgets/index/index.ts
```diff
@@ -342,6 +344,105 @@ const index = (props: IndexProps): Index => {
...
+            const requiredWidgets: Array<Widget['$$type']> =
+              stateToWidgetsMap[parameter];
+
+            if (
+              !requiredWidgets.some(requiredWidget =>
+                mountedWidgets.includes(requiredWidget)
+              )
+            ) {
+              acc.push({
+                [parameter]: stateToWidgetsMap[parameter],
```
samouss: I wonder if the current shape of the object is appropriate. We can only have an object with one key which is non defined in advance. It makes it a bit complicated to manipulate it below ([L402](https://github.com/algolia/instantsearch.js/pull/4140/files#diff-a07b6c7d4c8f3ebd0777d2622f2dc912R402)). Why not an object with a pre-defined structure where we can easily access the values? Or even a tuple? 

It would be simpler to traverse at the next step.

src/widgets/index/index.ts
```diff
@@ -342,6 +344,105 @@ const index = (props: IndexProps): Index => {
...
+            `The UI state for the index "${this.getIndexId()}" is not consistent with the widgets mounted.
+
+This can happen when the UI state is specified via \`initialUiState\` or \`routing\` but that the widgets responsible for this state were not added. This results in those query parameters not being sent to the API.
+
+To fully reflect the state, some widgets need to be added to the index "${this.getIndexId()}":
+
+${missingWidgets
+  .map(widget => {
+    const stateParameter = Object.keys(widget)[0];
+    const neededWidgets = stateToWidgetsMap[stateParameter]
```
samouss: We already have access to the needed widgets inside `widget` if I'm not mistaken.

src/widgets/index/index.ts
```diff
@@ -342,6 +344,105 @@ const index = (props: IndexProps): Index => {
...
+      .map(
+        (widgetIdentifier: string) => `"${widgetIdentifier.split('ais.')[1]}"`
```
samouss: There is some cases where we don't have a 1:1 mapping for connector/widget e.g. `range`, we don't have a `range` widget. We already have a hard-coded list we could create a real mapping of widgets that maps to the `uiState`.

src/types/widget.ts
francoischalifour: Would you see this type in `WidgetFactory` instead?

src/widgets/index/index.ts
francoischalifour: Fixed in 0ffe81e8c3ecad6c14868b6b7b4c60b763b69b6c.

src/widgets/index/index.ts
francoischalifour: Fair point! This is fixed in 0ffe81e8c3ecad6c14868b6b7b4c60b763b69b6c.

src/types/widget.ts
samouss: Yes it could be.

src/widgets/index/index.ts
```diff
@@ -346,6 +346,18 @@ const index = (props: IndexProps): Index => {
         instantSearchInstance.scheduleStalledRender();
 
         if (__DEV__) {
+          // Some connectors are responsible for multiple widgets so we need
+          // to map them.
+          function getWidgetNames(connectorName: string): string[] {
+            switch (connectorName) {
+              case 'range':
```
samouss: `menu` should also be inside the switch, both `menu` and `menuSelect` use it.

src/types/widget.ts
francoischalifour: Let's change that in another PR if you don't mind since it's not related to this change and could lead to broader TypeScript changes.

src/widgets/index/index.ts
```diff
@@ -348,11 +348,15 @@ const index = (props: IndexProps): Index => {
         if (__DEV__) {
           // Some connectors are responsible for multiple widgets so we need
           // to map them.
+          // eslint-disable-next-line no-inner-declarations
```
Haroenv: why do we even have this eslint rule? seems useless to me

francoischalifour: It's part of `eslint:recommended`. I agree that's not a great rule, especially here since we create that function for each index (i.e. not often).

PR #4142, opened 2019-09-25
feat(autocomplete): provide indexId

samouss:
This PR adds the `indexId` attribute to the `indices` provided by `autocomplete`.

Only the `indexName` was provided but its value is not stable. The value changes when the index changes e.g. with `sortBy`. A possible use case for stable value is a list of labels. Users might want to have labels on their indices. Without stable values, you can't predict which label you'll have.  The `indexId` solve this problem.


src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -183,13 +183,15 @@ search.addWidgets([
 
     expect(render).toHaveBeenCalledTimes(2);
     expect(secondRenderOptions.indices).toHaveLength(2);
```
Haroenv: I wonder if this assertion isn't simpler as an inline snapshot or as a toEqual

PR #4143, opened 2019-09-25
fix(instantsearch): return instance in widgets methods

francoischalifour:
An `index` widget returns its own reference when calling `addWidgets()` and `removeWidgets()`. This makes chainable calls possible.

https://github.com/algolia/instantsearch.js/blob/443cb0610e77ff87b9d054e3b3bf43586fc31810/src/widgets/index/index.ts#L53-L54

However, `instantsearch` widgets methods didn't return the search instance, which was not intuitive when you're used to the `index` behavior.

This PR returns the search instance when calling these methods, so that this becomes possible:

```ts
const search = instantsearch({ /* ... */ });

search
  .addWidgets(widgets)
  .start();
```

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -258,9 +258,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       'addWidget will still be supported in 4.x releases, but not further. It is replaced by `addWidgets([widget])`'
     );
 
-    this.addWidgets([widget]);
-
-    return this;
+    return this.addWidgets([widget]);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/lib/InstantSearch.ts
```diff
@@ -257,7 +257,10 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       false,
       'addWidget will still be supported in 4.x releases, but not further. It is replaced by `addWidgets([widget])`'
     );
+
     this.addWidgets([widget]);
+
+    return this;
```
samouss: The singular methods can return the call to `addWidgets` rather than `this`.

PR #4150, opened 2019-09-30
fix: display warnings only in development

francoischalifour:
This fixes warnings being part of the production bundle. This usage is similar to how React strips their warnings in production.

Rollup couldn't infer from the `warning` usage that they should be stripped from the bundle. This PR adds a Babel plugin for wrapping the `warning` calls with the `__DEV__` check, that our tools evaluate to `false` in production and therefore strip in the bundle.

This fix can get backported to v3 so that next versions' bundles are also smaller.

### DIFF
--- a/scripts/babel/wrap-warning-with-dev-check.js
+++ b/scripts/babel/wrap-warning-with-dev-check.js
@@ -1,0 +1,68 @@
+/**
+ * Babel plugin that wraps `warning` calls with development check to be
+ * completely stripped from the production bundle.
+ *
+ * In the development bundle, warnings get wrapped with their condition
+ * and their condition becomes false to trigger them without evaluating twice.
+ *
+ * Input:
+ *
+ * ```
+ * warning(condition, message);
+ * ```
+ *
+ * Output:
+ *
+ * ```
+ * if (__DEV__) {
+ *   if (!condition) {
+ *     warning(false, message);
+ *   }
+ * }
+ * ```
+ */
+
+function wrapWarningInDevCheck(babel) {
+  const t = babel.types;
+
+  const DEV_EXPRESSION = t.identifier('__DEV__');
+  const SEEN_SYMBOL = Symbol('expression.seen');
+
+  return {
+    visitor: {
+      CallExpression: {
+        exit(path) {
+          const node = path.node;
+
+          if (node[SEEN_SYMBOL]) {
+            return;
+          }
+
+          if (path.get('callee').isIdentifier({ name: 'warning' })) {
+            const [condition, message] = node.arguments;
+            const newNode = t.callExpression(
+              node.callee,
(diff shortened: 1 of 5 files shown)


scripts/babel/wrap-warning-with-dev-check.js
```diff
@@ -0,0 +1,68 @@
...
+ * ```
+ * warning(condition, message);
+ * ```
+ *
+ * Output:
+ *
+ * ```
+ * if (__DEV__) {
+ *   if (!condition) {
+ *     warning(false, message);
```
Haroenv:
```suggestion
 *     warning(condition, message);
```

could this not be sufficient in output?

francoischalifour: WDYM?

In development, we want the reverse of the condition to be in `if` and the condition in warning to be `false`. Remember: the warning message is displayed if the condition evaluates to `false` (that can be confusing and we might want to change that at some point).

Haroenv: I meant that this should be enough:

```
warning(condition, message);

// becomes

__DEV__ ? warning(condition, message) : void 0;

// instead of

__DEV__ ? !condition ? warning(false, message) : void 0 : void 0;
```

Not sure why that extra wrapper is necessary :)

francoischalifour: That should indeed be enough in our case, thanks!

francoischalifour: See 656b44da1f956909853662a6fe38137a46063695.

francoischalifour:
@Haroenv That's a fair point. For now we can still rely on `rollup-plugin-replace` since we use it for other replacements until we invest more heavily on custom Babel plugins.

Haroenv:
That's another option. I think we can do some nice things with babel plugins, so I don't mind that the first one we write is replaceable

Haroenv:
we can lower the bundlesize limit, seems to be 1kb lower than `next`: https://bundlesize-store.now.sh/build?info=%7B%22files%22%3A%5B%7B%22maxSize%22%3A70656%2C%22path%22%3A%22.%2Fdist%2Finstantsearch.production.min.js%22%2C%22size%22%3A67452%2C%22compression%22%3A%22gzip%22%7D%2C%7B%22maxSize%22%3A163840%2C%22path%22%3A%22.%2Fdist%2Finstantsearch.development.js%22%2C%22size%22%3A160214%2C%22compression%22%3A%22gzip%22%7D%5D%2C%22repo%22%3A%22algolia%2Finstantsearch.js%22%2C%22branch%22%3A%22fix%2Fwarning-babel-plugin%22%2C%22commit_message%22%3A%22%22%2C%22sha%22%3A%228104a686bb3b901f8d7d289a7292d5ab23c66ae4%22%7D

PR #4151, opened 2019-09-30
fix(instantsearch): warn deprecated usage of `searchParameters`

francoischalifour:
> This PR targets the `develop` branch, thus InstantSearch.js 3.

This displays a development warning when using `searchParameters` on InstantSearch.js 3. This option doesn't exist anymore on v4 and therefore it's better to warn users before reading the migration guide. We instead advise for using the `configure` widget.

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -7,10 +7,12 @@ import version from './version';
 import createHelpers from './createHelpers';
 import {
   createDocumentationMessageGenerator,
+  createDocumentationLink,
   findIndex,
   isPlainObject,
   mergeDeep,
   noop,
+  warning,
 } from './utils';
 
 const withUsage = createDocumentationMessageGenerator({
@@ -48,7 +50,7 @@ class InstantSearch extends EventEmitter {
     const {
       indexName = null,
       numberLocale,
-      searchParameters = {},
+      searchParameters,
       routing = null,
       searchFunction,
       stalledSearchDelay = 200,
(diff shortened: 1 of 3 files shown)


src/lib/InstantSearch.js
```diff
@@ -88,6 +90,15 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       throw new Error('The provided `insightsClient` must be a function.');
     }
 
+    warning(
+      !searchParameters,
+      `The \`searchParameters\` option is deprecated and will not be supported in InstantSearch.js 4.x.
+
+You can replace it with the \`configure\` widget: ${createDocumentationLink({
+        name: 'configure',
+      })}`
```
```suggestion
      })}
      
      instantsearch.widgets.configure({${JSON.stringify(searchParameters, null, 2)}});
      
      `
```

this is an option

francoischalifour: Yeah I spent some time trying to format this but in the end it's not possible. It won't be formatted properly. Not sure it's worth the trouble anyway.

Haroenv: This isn't about formatting, I'm adding some text (example, needs to be refined) to show people which search parameters to apply themselves

francoischalifour: What I meant is that formatting is off that way:

[InstantSearch.js]: The `searchParameters` option is deprecated and will not be supported in InstantSearch.js 4.x.
      
You can replace it with the `configure` widget:

```
configure({
  "disjunctiveFacetsRefinements": {
    "brand": [
      "Samsung"
    ]
  }
});
```

See https://www.algolia.com/doc/api-reference/widgets/configure/js/

--- 

If we value guidance over formatting, we can still go for it. WDYT?

Haroenv: What you wrote in that comment looks quite good to me. Maybe can be wrapped in an `addWidgets`, but that could be clear enough

francoischalifour: Ideally it would be like this:

```
configure({
  disjunctiveFacetsRefinements: {
    brand: ['Samsung'],
  },
});
```

But I guess it's fine that way. I'll update this.

PR #4153, opened 2019-10-01
feat(transformers): add tests

Haroenv:
Edited the existing codemod to add tests, and ran it on the codebase. This is as a reference for other transformers we can add in the future.

I could not find how to merge two calls unfortunately :(

Note: this PR used to include a CLI, which is now removed, since we disagree whether it should be part of this project.


scripts/transforms/README.md
```diff
@@ -1,16 +1,18 @@
 ## Codemods
 
-These codemods should be ran with [`jscodeshift`](https://github.com/facebook/jscodeshift).
+These codemods (code transformers) can be ran with `npx`. If you installed the `instantsearch.js` package already with `yarn`, you can ru`yarn instantsearch-scripts`).
```
Haroenv:
```suggestion
These codemods (code transformers) can be ran with `npx`. If you installed the `instantsearch.js` package already with `yarn`, you can run them like this: `yarn instantsearch-scripts`).
```

package.json
```diff
@@ -115,7 +121,7 @@
     "eslint-plugin-jest": "22.15.2",
     "eslint-plugin-prettier": "3.1.0",
     "eslint-plugin-react": "7.14.3",
-    "inquirer": "6.5.2",
+    "inquirer": "7.0.0",
```
Haroenv: should be moved to `dependencies`, not sure why it was already explicitly added

scripts/transforms/index.js
```diff
@@ -0,0 +1,219 @@
...
+    }
+  }
+
+  if (!clean) {
+    if (force) {
+      console.log(`WARNING: ${errorMessage}. Forcibly continuing.`);
+    } else {
+      console.log('Thank you for using InstantSearch codemods!');
+      console.log(
+        chalk.yellow(
```
Haroenv: only used here

scripts/transforms/index.js
```diff
@@ -0,0 +1,219 @@
+/* eslint-disable no-process-exit,import/no-commonjs,no-console */
+const globby = require('globby');
+const inquirer = require('inquirer');
+const meow = require('meow');
+const path = require('path');
+const execa = require('execa');
+const chalk = require('chalk');
```
Haroenv: can be replaced by the manual version of the script

scripts/transforms/index.js
```diff
@@ -0,0 +1,219 @@
+/* eslint-disable no-process-exit,import/no-commonjs,no-console */
+const globby = require('globby');
+const inquirer = require('inquirer');
+const meow = require('meow');
+const path = require('path');
```
Haroenv: should be at the top, since it's node global

package.json
```diff
@@ -141,6 +147,9 @@
     "typescript": "3.5.3",
     "webpack": "4.39.2"
   },
+  "bin": {
+    "instantsearch-scripts": "./scripts/index.js"
```
Haroenv: we need to make sure these are published on npm via the `files` key

package.json
eunjae-lee: It's being used in `publish.js`.

package.json
eunjae-lee: `./scripts/index.js` is not inside `files` key...?

scripts/transforms/index.js
```diff
@@ -0,0 +1,219 @@
...
+
+  inquirer
+    .prompt([
+      {
+        type: 'input',
+        name: 'files',
+        message: 'On which files or directory should the codemods be applied?',
+        when: !cli.input[1],
+        default: '.',
+        // validate: () =>
```
eunjae-lee:
```suggestion
```

package.json
```diff
@@ -141,6 +148,9 @@
...
```
francoischalifour: Why not have this index file in `scripts/transforms/bin.js` and only include this folder in `files`?

scripts/transforms/README.md
```diff
@@ -1,16 +1,18 @@
...
 
 ```
-yarn jscodeshift --transform scripts/transforms/addWidget-addWidgets.js --extensions='ts,js,tsx' examples stories .storybook
+npx instantsearch.js addWidget-to-addWidgets <path>
 ```
 
-Make sure to run Prettier after!
+### Notes
+
+If you are using Prettier or eslint, make sure to run its autofixing after this transformation, since it's possible that code should be formatted differently after it has been transformed. For example in our repository using prettier that would be:
```
francoischalifour:
```suggestion
If you are using Prettier or ESLint, make sure to run its autofixing after this transformation, since code can be formatted differently after it has been transformed. For example, in our repository, the Prettier command would be:
```

package.json
Haroenv: It’s possible, I was thinking of potentially adding more cli later, but I guess that’s premature Optimization

francoischalifour:
@Haroenv As expressed in https://github.com/algolia/instantsearch.js/pull/4153#pullrequestreview-295442598, I wouldn't be comfortable introducing a new usage of the package for exposing a first codemod that does not fully address the problem. Right now, it makes the code work with v4 but leaves it in an "unstable" state where you still call multiple times `addWidgets()`. In my mind, we should either solve the problem fully and not provide  a non-complete solution.

Haroenv:
Merging function calls isn't something that seems to be in any way trivial with codemods, since you can't really go to the "next call" as far as I can tell

scripts/transforms/README.md
```diff
@@ -1,16 +1,18 @@
 ## Codemods
 
-These codemods should be ran with [`jscodeshift`](https://github.com/facebook/jscodeshift).
+These codemods (code transformers) can be ran with `npx`. If you installed the `instantsearch.js` package already with `yarn`, you can run them like this: `yarn instantsearch-scripts`).
 
-### addWidget / removeWidget
+### `addWidget-to-addWidgetss`
```
tkrugg: there is a typo here, no?

Haroenv: the title is what it does, not the name of the transform

Haroenv:
```suggestion
### `addWidget-to-addWidgets`
```

Haroenv:
correct, CLI was removed in a later commit since we disagreed :)

PR #4156, opened 2019-10-04
refactor: migrate to Preact X

### LINKED ISSUE #3828: Upgrade preact-compat
**Is your feature request related to a problem? Please describe 🙏**
This is by no means urgent, but integrating instantsearch into a project using Preact X (e.g. 10.0.0-beta.1) causes issues during page load (see https://github.com/preactjs/preact/issues/1655)

So just raising a potential issue that might arise more often as and when Preact X is released and more Algolia consumers upgrade. And I expect it will be taken up quickly given it adds support for latest React 16 features (hooks, context)

**Describe the solution you'd like 🤔**
Move to new `preact/compat` instead of `preact-compat` (perhaps when X is stable)

**Describe alternatives you've considered ✨**
Our interim solution was to alias `preact-compat` to `preact/compat`

francoischalifour:
This updates InstantSearch.js to Preact X. This version was [released as stable](https://github.com/preactjs/preact/releases/tag/10.0.0) and will generate a lighter bundle and a more stable development environment. The challenging part of this migration was the way we tested some components.

This history is quite messy because I started this work a few months ago and we rebased the base branch in the meantime. Rebasing is now quite hard. I can spend some time rebasing later but not sure it's worth the trouble for now.

## Why

We want to migrate to Preact X for the following reasons:

- Recent and future-proof React public API
    - Fragment support: some component could not follow the InstantSearch.css spec because Fragments were not supported: additional `div`s were added.
- Smaller bundle size
- Less confusion in our setup ([we used to use Preact components but test them with React](https://github.com/algolia/instantsearch.js/blob/21b1ef9223f6b0d327e608f96e343d157518815c/src/components/Template/__tests__/Template-test.js))

### `unmountComponentAtNode` is not exported by Preact anymore

We now rely on rendering `null` (which is the internal implementation in Preact) because `unmountComponentAtNode` is part of `preact/compat` which we don't need anymore.

### `onChange` becomes `onInput` on input elements

This makes it hard to test with Enzyme. Besides Enzyme's `mount` doesn't mount children (it shallow-render them). Therefore I introduced [`preact-testing-library`](https://github.com/testing-library/preact-testing-library) which correctly trigger DOM events that can be caught up in tests.

### React is fully removed from the codebase

We used to depend on Preact for our components but React in the testing environment, which led to inconsistent results. We now use Preact everywhere and use the `h` pragma. This should make the setup easier and more consistent.

### Preact Rheostat dependency is gone

Having Preact Rheostat as a dependency caused some troubles: hard to remove its proptypes in production, depends on other versions of Preact, unmaintained source. Besides, the version we had wasn't compatible with Preact X (it used string refs). The Rheostat component is now part of InstantSearch.js code and will be able to evolve with time. With this change, we're also in control of what goes into our library (I removed some unused part).

### Snapshot format has changed

The `mount` function of Enzyme with the `enzyme-adapter-preact-pure` adapter now always returns arrays. Component snapshots have all been updated. Some were migrated using the snapshot version of `preact-testing-library` (we might want to use this version for new component tests).

### Snapshot to Preact's `render` function have changed

We used to snapshot the whole call to the Preact `render` function but now only snapshot the props object. It'd be nice at some point to get rid of these snapshots.

### Less usage of `shouldComponentUpdate`

`shouldComponentUpdate` acts differently in Preact X and cause issues where used like this:

    shouldComponentUpdate(nextProps) {
      return !isEqual(this.props, nextProps);
    }

I removed cases where it was used this way. The performance gains were minimal if not inexistent.

### Remove usage of `componentWillMount`

[`componentWillMount`](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount) is a deprecated lifecycle hook that was trivial to remove.

## Browser support

Preact X supports IE11, like InstantSearch.js.

## Skipped tests

Some tests don't work with the current setup and are skipped. I suspect this is because of underlying libraries not being updated yet. If we don't plan to touch the `SearchBox` component, we can wait for libraries to be updated and keep these skipped tests for now.

### SearchBox autofocus

The test environment doesn't reflect the `autofocus` prop like it does in production.

### SearchBox reset

The reset event doesn't seem to be caught by the test setup.

## How to review

A lot of snapshots were updated so there are a lot of files to check. Reviewing components' behavior on the [Storybook](https://deploy-preview-4156--instantsearchjs.netlify.com/stories) and on the examples will help make sure that everything behaves as expected.

## Next steps

- Migrate away from deprecated React lifecycles

## Preview

- [Storybook](https://deploy-preview-4156--instantsearchjs.netlify.com/stories)
- [E-commerce demo](https://deploy-preview-4156--instantsearchjs.netlify.com/examples/e-commerce/)

## Related

- Closes #3828

## References

- [https://preactjs.com/guide/v10/whats-new/](https://preactjs.com/guide/v10/whats-new/)
- [https://preactjs.com/guide/v10/upgrade-guide/#notes-for-library-authors](https://preactjs.com/guide/v10/upgrade-guide/#notes-for-library-authors)


Haroenv:
You can update the dev bundle size, it's 10kb smaller than the limit

.eslintrc.js
```diff
@@ -1,11 +1,14 @@
 module.exports = {
   extends: ['algolia', 'algolia/jest', 'algolia/react', 'algolia/typescript'],
+  plugins: ['react-hooks'],
```
Haroenv: can you open an issue or PR to add this to our react config?

scripts/jest/matchers/toWarnDev.ts
```diff
@@ -12,12 +12,12 @@ declare global {
...
     // eslint-disable-next-line @typescript-eslint/generic-type-naming
     interface Matchers<R> {
-      toWarnDev(expectedMessage?: string): R;
+      toWarnDev(expectedMessage?: string): CustomMatcherResult;
     }
   }
 }
 
-const matcher = {
+const matcher: jest.ExpectExtendMap = {
```
Haroenv: this file change seems unrelated, but would be `chore` otherwise, so fine likely

src/components/CurrentRefinements/__tests__/CurrentRefinements-test.tsx
```diff
@@ -1,7 +1,9 @@
-import React from 'react';
+/** @jsx h */
```
Haroenv: can we set this globally instead of per file, or is the problem that the import will disappear or be "unused" otherwise?

src/components/CurrentRefinements/__tests__/CurrentRefinements-test.tsx
```diff
@@ -167,9 +169,9 @@ describe('CurrentRefinements', () => {
         ],
       };
 
-      const tree = mount(<CurrentRefinements {...props} />);
+      const { container } = render(<CurrentRefinements {...props} />);
```
Haroenv: does container have a method that could be used? seems like that's likely what causes it to be an array

src/components/GeoSearchControls/__tests__/__snapshots__/GeoSearchControls-test.js.snap
```diff
@@ -15,7 +15,13 @@ exports[`GeoSearchControls Clear button expect to render the button when the ref
       data={Object {}}
       rootTagName="span"
       templateKey="reset"
-      templates={Object {}}
+      templates={
+        Object {
+          "redo": "redo",
```
Haroenv: great addition

src/components/GeoSearchControls/__tests__/__snapshots__/GeoSearchControls-test.js.snap
```diff
@@ -166,4 +208,4 @@ exports[`GeoSearchControls Control enabled expect to render the toggle unchecked
 </div>
 `;
 
-exports[`GeoSearchControls expect to render nothing with refine disabled 1`] = `""`;
+exports[`GeoSearchControls expect to render nothing with refine disabled 1`] = `null`;
```
Haroenv: replace with a normal assertion (toBeNull)?

src/components/Pagination/PaginationLink.js
```diff
@@ -1,46 +1,43 @@
...
 
-  shouldComponentUpdate(nextProps) {
-    return !isEqual(this.props, nextProps);
-  }
+import { h } from 'preact';
+import PropTypes from 'prop-types';
 
-  handleClick(e) {
-    this.props.handleClick(this.props.pageNumber, e);
+function PaginationLink({
```
Haroenv: good refactor!

src/components/RangeInput/__tests__/RangeInput-test.js
```diff
@@ -119,165 +122,115 @@ describe('RangeInput', () => {
...
 
-      component
-        .find('input[type="number"]')
-        .last()
-        .simulate('change', {
-          currentTarget: {
-            value: 480,
-          },
-        });
+      fireEvent.input(maxInput, { target: { value: 480 } });
```
Haroenv: good refactor on this test IMO

src/components/SearchBox/__tests__/SearchBox-test.js
```diff
@@ -153,14 +156,19 @@ describe('SearchBox', () => {
       });
     });
 
-    test('sets focus with autofocus to true', () => {
+    // @TODO Since the Preact X migration and new testing environment, this test
+    // doesn't pass.
+    // eslint-disable-next-line jest/no-disabled-tests
+    test.skip('sets focus with autofocus to true', () => {
```
Haroenv: can you try changing `autoFocus` to `autofocus` on the component?

src/components/SearchBox/__tests__/SearchBox-test.js
```diff
@@ -258,50 +273,59 @@ describe('SearchBox', () => {
...
 
         expect(onSubmit).toHaveBeenCalledTimes(1);
       });
     });
 
-    describe('onReset', () => {
+    // @TODO Since the Preact X migration and new testing environment, this test
+    // suite doesn't pass.
+    // eslint-disable-next-line jest/no-disabled-tests
+    describe.skip('onReset', () => {
```
Haroenv: where should we open an issue for this?

src/components/Selector/Selector.js
```diff
@@ -1,37 +1,27 @@
...
-            key={option.label + option.value}
-            value={`${option.value}`}
-          >
-            {option.label}
-          </option>
-        ))}
-      </select>
-    );
-  }
+function Selector({ currentValue, options, cssClasses, setValue }) {
```
Haroenv: should we call this `Select` instead?

src/components/Slider/Rheostat.js
```diff
@@ -0,0 +1,709 @@
...
+              </PitComponent>
+            );
+          })}
+        {children}
+      </div>
+    );
+  }
+}
+
+Rheostat.propTypes = {
```
Haroenv: could this have moved to TS, or do you want to do it in another PR?

src/components/Template/__tests__/Template-test.js
```diff
@@ -11,92 +32,109 @@ describe('Template', () => {
...
     it('calls render when templateKey changes', () => {
-      props.templateKey += '-rerender';
-      props.templates = { [props.templateKey]: '' };
-      ReactDOM.render(<Template {...props} />, container);
-      expect(component.render).toHaveBeenCalled();
+      const props = getProps({});
+      const wrapper = shallow(<Template {...props} />, container);
+      const onRender = jest.spyOn(wrapper.instance(), 'render');
+
+      wrapper.setProps({
```
Haroenv: good change on moving from mutation

src/components/VoiceSearch/__tests__/__snapshots__/VoiceSearch-test.tsx.snap
```diff
@@ -2,38 +2,51 @@
 
 exports[`VoiceSearch Rendering with custom template for status 1`] = `
 <div
```
Haroenv: how come this isn't wrapped in an array?

src/widgets/breadcrumb/__tests__/__snapshots__/breadcrumb-test.js.snap
```diff
@@ -1,47 +1,41 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`breadcrumb() render renders transformed items correctly 1`] = `
-<Breadcrumb
-  canRefine={true}
-  createURL={[Function]}
-  cssClasses={
+Object {
```
Haroenv: are you aware the JSX didn't work here?

src/widgets/breadcrumb/__tests__/breadcrumb-test.js
```diff
@@ -109,7 +109,9 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
         instantSearchInstance: {},
       });
 
-      expect(render.mock.calls[0][0]).toMatchSnapshot();
+      const [firstRender] = render.mock.calls;
+
+      expect(firstRender[0].props).toMatchSnapshot();
```
Haroenv: shouldn't this be done by test-renderer like the other tests?

src/widgets/breadcrumb/breadcrumb.js
```diff
@@ -164,7 +166,7 @@ export default function breadcrumb({
   });
 
   const makeBreadcrumb = connectBreadcrumb(specializedRenderer, () =>
-    unmountComponentAtNode(containerNode)
+    render(null, containerNode)
```
Haroenv: have we migrated to this in every component?

tsconfig.json
```diff
@@ -3,9 +3,10 @@
     "target": "ESNEXT",
     "module": "es2015",
     "moduleResolution": "node",
-    "jsx": "preserve",
     "noEmit": true,
     "strict": true,
+    "jsx": "react",
```
Haroenv: should this not be preact?

scripts/jest/matchers/toWarnDev.ts
francoischalifour: I converted `setupTest.js` to TypeScript and it raised some TS issues so I took this opportunity to fix them.

src/components/CurrentRefinements/__tests__/CurrentRefinements-test.tsx
francoischalifour: At first I wanted to avoid having to declare this pragma everywhere but VS code was unaware of the pragma when it was not present in the file and thus raised errors. We can remove it later with Babel plugins if it's too annoying.

src/components/CurrentRefinements/__tests__/CurrentRefinements-test.tsx
francoischalifour: The snapshots from `preact-testing-library` are _not_ arrays, while the ones from `enzyme` are.

src/components/SearchBox/__tests__/SearchBox-test.js
francoischalifour: It doesn't work either.

src/components/Selector/Selector.js
francoischalifour: We can refactor further in other PRs.

src/components/Slider/Rheostat.js
francoischalifour: We can refactor further in other PRs.

src/components/VoiceSearch/__tests__/__snapshots__/VoiceSearch-test.tsx.snap
francoischalifour: https://github.com/algolia/instantsearch.js/pull/4156#discussion_r331562006

src/widgets/breadcrumb/breadcrumb.js
francoischalifour: Yep

src/widgets/breadcrumb/__tests__/breadcrumb-test.js
francoischalifour: WDYM?

src/widgets/breadcrumb/__tests__/__snapshots__/breadcrumb-test.js.snap
francoischalifour: See the section "Snapshot to Preact's `render` function have changed" in the original post.

tsconfig.json
francoischalifour: Nope, the way to tell TS that it's a Preact environment is to have:

```json
{
  "compilerOptions": {
    "jsx": "react",
    "jsxFactory": "h"
  }
}
```

francoischalifour:
@Haroenv Good call about the `utils.ts` file, it an unused file that comes from a merge issue.

src/components/SearchBox/__tests__/SearchBox-test.js
yannickcr: Seems to be an issue in `jsdom` https://github.com/testing-library/react-testing-library/issues/393
An issue was open, but without any information it was promptly closed https://github.com/jsdom/jsdom/issues/2622
We can complete this issue and ask to re-open it.

src/components/SearchBox/__tests__/SearchBox-test.js
yannickcr: Seems to be a limitation of `testing-library`/`jsdom` (not sure, I did not find a lot of information on `autofocus`). I will open an issue for this.

I think that right now the only thing we can test is the presence of the `autofocus` attribute on the element:

```js
expect(input).toHaveAttribute('autofocus', 'true');
```

src/components/SearchBox/__tests__/SearchBox-test.js
francoischalifour: Great find – so I was not going crazy! I'll take care of opening an issue on `jsdom`.

.eslintrc.js
francoischalifour: Will open a PR!

src/components/SearchBox/__tests__/SearchBox-test.js
francoischalifour: That's a good temporary workaround!

francoischalifour:
@yannickcr I didn't want to migrate the files that didn't need to be updated but we should indeed at some point get rid of one solution (likely Enzyme). From my experience it's way easier to write tests with the Testing Library, except for the few problems that I mentioned and that I'm gonna write issues about on their respective projects.

@Haroenv @yannickcr Are you guys confident enough to include this in the first v4 release? (i.e. merging now)

src/widgets/breadcrumb/__tests__/breadcrumb-test.js
Haroenv: I misunderstood what these tests are doing, i thought they checked the output, but instead it's just asserting the props. I think we can remove the tests fully (in a stacked PR), wdyt?

.eslintrc.js
francoischalifour: algolia/eslint-config-algolia#118

src/widgets/breadcrumb/__tests__/breadcrumb-test.js
francoischalifour: At some point we want to remove snapshots at this level. This is however not the only widget using those (all of them use snapshots).

yannickcr:
@francoischalifour Yes, my intent was not to migrate in the other files right now (this PR is already big :yum:) but more about thinking for the future. I agree for Testing Library vs Enzyme :innocent:.

We are currently at `4.0.0-beta.1` so I suppose this can still be added to the release. I'm relatively confident but I would be more at ease with at least a beta and an RC before the final release.

Haroenv:
I'm confident enough too once all comments have been solved

src/components/SearchBox/__tests__/SearchBox-test.js
francoischalifour: From my tests, it's not coming from `jsdom` because the `reset` event is triggered:

```js
const { JSDOM } = require('jsdom');

const dom = new JSDOM(`
<body>
  <form>
    <button type="submit">Submit</button>
    <button type="reset">Reset</button>
  </form>
</body>
`);

const document = dom.window.document;

document.querySelector('form').addEventListener('reset', event => {
  event.preventDefault();

  console.log('Reset event received');
});

document.querySelector('button[type="reset"]').click();
```

You can reproduce that in a Node environment, by installing `jsdom`.

I suggest we move forward with this PR and find a workaround after. WDYT?

yannickcr: > I suggest we move forward with this PR and find a workaround after.

I agree.

Haroenv: https://github.com/algolia/instantsearch.js/issues/4160

PR #4165, opened 2019-10-14
fix(index): support custom UI params in inconsistent UI state warning

francoischalifour:
## Description

When developing the new [Places.js](https://github.com/algolia/places) widget for InstantSearch 4, I realized that having a non-supported key in the UI state mapping throws in development mode because of the inconsistent UI state warning. This fix makes sure that custom widgets that enhance the UI state does not throw in development mode.

## Related

- #4140

### DIFF
--- a/src/widgets/index/index.ts
+++ b/src/widgets/index/index.ts
@@ -394,6 +394,10 @@ const index = (props: IndexProps): Index => {
             const requiredWidgets: Array<Widget['$$type']> =
               stateToWidgetsMap[parameter];
 
+            if (!requiredWidgets) {
+              return acc;
+            }
+
             if (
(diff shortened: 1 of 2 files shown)


src/widgets/index/index.ts
```diff
@@ -394,6 +394,10 @@ const index = (props: IndexProps): Index => {
             const requiredWidgets: Array<Widget['$$type']> =
               stateToWidgetsMap[parameter];
```
Haroenv: a separate solution  could be 
```suggestion
              stateToWidgetsMap[parameter] || [];
```

this logic still works for unknown ui state

Haroenv: the TS annotation is lying here, since it can also be undefined

PR #4167, opened 2019-10-14
feat(places): add Places widget

francoischalifour:
## Description

The [Places.js widget for InstantSearch](https://github.com/algolia/places/blob/f9f7aa47e73d46866d71dd4d32f80d31234f86f8/src/instantsearch/widget.js) was not longer compatible with InstantSearch 4 because the widget lifecycle changed.

We'll no longer maintain the widget on the Places.js repo but rather provide a Places widget as part of InstantSearch itself. The reasoning behind this decision is that the widget hosted in Place.js couldn't be updated without also bumping Place.js to a major version. We are now more in control of when to release this widget.

This is a port of the existing widget with updated lifecycles and technical stack (TypeScript).

## Usage

```ts
import algoliaPlaces from 'places.js';

  /* ... */
});

search.addWidgets([
  places({
    placesReference: algoliaPlaces,
    container: '#places',
    defaultPosition: ['37.7793', '-122.419'],
  }),
]);
```

## API

So that Places.js isn't loaded with InstantSearch even though you don't use this widget, Places reference is now passed to the widget itself (similar to what we do for `searchClient`, `insightsClient` and `googleReference`).

### `placesReference`

> Places reference | required

The Algolia Places reference to use.

### `defaultPosition`

> `string[]` | optional

The default position when the input is empty.

## Preview

[See stories →](https://deploy-preview-4167--instantsearchjs.netlify.com/stories/?path=/story/results-geosearch--with-position-from-places)

## Impact

- Document the widget API on [the documentation website](https://www.algolia.com/doc/)
- Update the documentation for Place.js with InstantSearch.js 4

### DIFF
--- a/.storybook/static/storybook.css
+++ b/.storybook/static/storybook.css
@@ -115,7 +115,7 @@ body {
 
 .algolia-places {
   display: block !important;
-  width: 850px;
+  width: 100%;
   margin: 0 auto;
 }
 
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "jscodeshift": "0.6.4",
     "jsdom-global": "3.0.2",
     "mversion": "1.13.0",
-    "places.js": "1.16.4",
+    "places.js": "1.16.6",
     "prettier": "1.18.2",
     "rollup": "1.21.4",
     "rollup-plugin-babel": "4.3.3",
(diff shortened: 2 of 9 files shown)


Haroenv:
(approve here is for the code alone, we need to make sure we don't rush this release of course)

francoischalifour:
@Haroenv We could call that `placesReference` like we do for [`googleReference`](https://www.algolia.com/doc/api-reference/widgets/geo-search/js/#widget-param-googlereference).

Haroenv:
yes, that works better than client for me

PR #4170, opened 2019-10-22
fix(helper): expose .lastResults to .helper

Haroenv:
**Summary**

In v3, the helper exposed on the instance is the same helper as the one who searches. This means that lastResults is set as expected.

However, with v4 we split this up in the derived helper which actually searches and the exposed helper.

This is fixed now by just setting the last results once the search is done on the derived helper.

**Result**

`search.helper.lastResults` is defined in cases like these: https://codesandbox.io/s/instantsearchjs-app-tkp7h

### DIFF
--- a/src/widgets/index/index.ts
+++ b/src/widgets/index/index.ts
@@ -459,11 +459,16 @@ See https://www.algolia.com/doc/guides/building-search-ui/widgets/customize-an-e
         }
       });
 
-      derivedHelper.on('result', () => {
+      derivedHelper.on('result', ({ results }) => {
         // The index does not render the results it schedules a new render
         // to let all the other indices emit their own results. It allows us to
         // run the render process in one pass.
         instantSearchInstance.scheduleRender();
+
+        // the derived helper is the one which actually searches, but the 'main'
+        // helper is exposed e.g. via instance.helper, which needs to have the
+        // results accessible as well.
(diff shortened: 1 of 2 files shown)


src/widgets/index/index.ts
```diff
@@ -459,11 +459,16 @@ See https://www.algolia.com/doc/guides/building-search-ui/widgets/customize-an-e
...
-      derivedHelper.on('result', () => {
+      derivedHelper.on('result', ({ results }) => {
         // The index does not render the results it schedules a new render
         // to let all the other indices emit their own results. It allows us to
         // run the render process in one pass.
         instantSearchInstance.scheduleRender();
+
+        // the derived helper is the one which actually searches, but the 'main'
+        // helper is exposed e.g. via instance.helper, which needs to have the
+        // results accessible as well.
```
samouss: The comment is misleading. We expose the [`mainIndex.helper`](https://github.com/algolia/instantsearch.js/blob/57cd9d7c4806239a36cc7886177c2698bf9052a0/src/lib/InstantSearch.ts#L430-L432) via `instantSearchInstance.helper`  not the `mainHelper`. The `mainHelper` is the one used to trigger the requests, it is not tied to any states nor results.

Haroenv: thanks, that's a good point. Reworded it in be8885a

PR #4172, opened 2019-10-24
fix: widget name in documentation link for index

tkrugg:
**Summary**

For technical reason we were not able to use `index` as widget name in
the docs, so we picked instead `index-widget`.

When you misuse a widget, we output an error message with a link to the
documentation generated from the name of the widget.

**Result**

N/A


Haroenv:
correct, if the tests are updated!

PR #4191, opened 2019-11-07
feat: detect window.aa when available on global scope and a function

tkrugg:
**Summary**

This detects the insights client when available on the global scope.
it purposely only works on the browser, there no value as far as I know, in having this behaviour server side. 

We should think of a more selective way of detecting it. Not sure how to do it without forcing the update of the snippet

**Result**

N/A


src/lib/InstantSearch.ts
```diff
@@ -184,6 +186,10 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       );
     }
 
+    if (!insightsClient && Boolean(getDetectedInsightsClient())) {
```
Haroenv: what about:

```js
const possibleDetectedInsightsClient = getDetectedInsightsClient();

if (!insightsClient && possibleDetectedInsightsClient) {
  insightsClient = possibleDetectedInsightsClient;
}
```

since it explicitly returns false, I don't see the need of wrapping it in a boolean call (&& already does it)

tkrugg: I should make it return `Function | undefined` instead

tkrugg: I changed it to `hasDetectedInsightsClient` so it's even more obvious now that it returns a boolean.

src/lib/InstantSearch.ts
```diff
@@ -184,6 +185,18 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       );
     }
 
+    warning(
+      Boolean(insightsClient) || !hasDetectedInsightsClient(),
```
Haroenv: I feel like this reads complicated, but it's because `warning` accepts a falsy argument

src/lib/utils/detect-insights-client.ts
```diff
@@ -0,0 +1,6 @@
+export default function hasDetectedInsightsClient(): boolean {
+  return (
+    typeof window !== 'undefined' &&
+    Boolean(window && (window as any).AlgoliaAnalyticsObject)
```
francoischalifour: Why do you recheck `window` here? It's already checked on the line above.

```suggestion
    Boolean((window as any).AlgoliaAnalyticsObject)
```

src/lib/InstantSearch.ts
```diff
@@ -23,6 +23,7 @@ import {
   Router,
   UiState,
 } from '../types';
+import hasDetectedInsightsClient from './utils/detect-insights-client';
```
francoischalifour: This file could go in `lib/insights/`.

src/lib/utils/detect-insights-client.ts
```diff
@@ -0,0 +1,6 @@
+export default function hasDetectedInsightsClient(): boolean {
```
francoischalifour: If we want to completely strip it out of the production bundle, we might have to only implement the function in development mode (within a `__DEV__` check).

src/lib/InstantSearch.ts
```diff
@@ -184,6 +185,18 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       );
     }
 
+    warning(
+      Boolean(insightsClient) || !hasDetectedInsightsClient(),
+      `InstantSearch detected the insights client in the global scope.
```
francoischalifour: We capitalize "Insights" [everywhere in the doc](https://www.algolia.com/doc/guides/getting-insights-and-analytics/search-analytics/click-through-and-conversions/in-depth/cookie-strategy/#identifying-users-with-a-cookie-and-a-user-token).

```suggestion
      `InstantSearch detected the Insights client in the global scope.
```

src/lib/InstantSearch.ts
```diff
@@ -184,6 +185,18 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
       );
     }
 
+    warning(
+      Boolean(insightsClient) || !hasDetectedInsightsClient(),
+      `InstantSearch detected the insights client in the global scope. 
+To connect InstantSearch to insights client, make sure you provide it to your InstantSearch instance:
```
francoischalifour:
```suggestion
To connect InstantSearch to the Insights client, make sure to specify the `insightsClient` option:
```

src/lib/InstantSearch.ts
```diff
@@ -184,6 +185,18 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
+
+const search = instantsearch({
+  /* ... */
+  insightsClient: window.aa,
+});
+`
```
francoischalifour: We could link to the option in the [API reference](https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/). However I realize that we haven't documented `insightsClient` there.

src/lib/utils/detect-insights-client.ts
tkrugg: I was worried that `window === null` for example

src/lib/InstantSearch.ts
tkrugg: yes both are legit places. Any good argument for having it in `lib/insights` rather than `utils/`?

src/lib/utils/detect-insights-client.ts
tkrugg: where would you add `if (__DEV__)` condition? inside this function?

src/lib/InstantSearch.ts
tkrugg: > We could link to the option in the API reference.

True, but I wasn't sure how to point to:
- `https://www.algolia.com/doc/api-reference/widgets/instantsearch/js` for vanilla
- `https://www.algolia.com/doc/api-reference/widgets/instantsearch/angular` for angular
- etc.

> I realize that we haven't documented insightsClient there.

Good point. Filed a task here: https://algolia.atlassian.net/browse/IFW-1104

src/lib/InstantSearch.ts
tkrugg: I think there might be a typo in your comment. Do you mean _"`warning` accepts a boolean argument_"?

If yes, then indeed you get typescript compile error if the `condition` argument of the warning is not a strictly a boolean.

src/lib/utils/detect-insights-client.ts
francoischalifour: Is there a reason for writing this in a defensive programming way? This is usually not what we do in JavaScript and in this codebase.

src/lib/utils/detect-insights-client.ts
francoischalifour: You can see how it's done for the [`warning`](https://github.com/algolia/instantsearch.js/blob/50f4af3006a44cd08dd99b3a72bd410340c2e48a/src/lib/utils/logger.ts) function for example.

src/lib/InstantSearch.ts
francoischalifour: Linking to the right flavor documentation is a recurring problem that we already have. The same goes for widget vs. connector. We might have to introduce an error boundary system at the widgets' level at some point but this is beyond this scope. For consistency, we could link to the vanilla documentation once available (that's not a blocker for this PR anyway).

src/lib/InstantSearch.ts
francoischalifour: This file is related to the Insights feature. It's not a util meant to be shared and reused across the codebase. I would encourage you to read [this post about colocation](https://kentcdodds.com/blog/colocation/).

tkrugg: Nice read, thank you.

It is a util, the fact that it's not used in multiple place or specialised does not make it less of a util. 
See other specialised utils it'd be colocated with: [hits-query-id](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/utils/hits-query-id.ts), [geo-search](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/utils/geo-search.ts#L68), [renderTemplate](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/utils/renderTemplate.js#L24), [resolveSearchParameters](https://github.com/algolia/instantsearch.js/blob/develop/src/lib/utils/resolveSearchParameters.ts#L4).
One argument for putting it in `utils` is to keep  `lib/insights` stuff for public stuff (like we do for `lib/history` and `lib/stateMapping`).

src/lib/utils/detect-insights-client.ts
tkrugg: Make sense, thanks!

src/lib/utils/detect-insights-client.ts
tkrugg: it's not about being defensive. My concerns when writing this were:
1. having clear error messages: if you don't do this you get `TypeError: Cannot read property 'AlgoliaAnalyticsObject' of null`, which doesn't say much about what happened. 
2. not deceiving the reader thinking such a line can't fail `typeof window !== 'undefined' && window. AlgoliaAnalyticsObject`:
  Maybe writing it as `typeof window !== 'undefined' && window !== null` would have been better. 
  Or we can decide the window global can't be null, which is a safe assumption to my knowledge.

francoischalifour: That's what I meant. `window` cannot be `null`, otherwise the contract is not respected and everything would break (outside of InstantSearch).

src/lib/utils/detect-insights-client.ts
francoischalifour: We could even inline this check in the `warning`, this would make it easier to strip it out in production.

tkrugg: I checked the production build and I can confirm for the current PR, the function `hasDetectedInsightsClient ` has been completely tree shaked, without needing to inline it or add a similar logic as `warning`.

This is smarter than I expected and ideal, as if someone starts using the function for anything else than dev warnings, it will be included in the bundle no effort added.

francoischalifour: That's great! That's because of [this Babel plugin](https://github.com/algolia/instantsearch.js/blob/develop/scripts/babel/wrap-warning-with-dev-check.js) that wraps the warning inside a `__DEV__` check, then Rollup sees it's unused and doesn't include it in the bundle.

src/lib/InstantSearch.ts
tkrugg: done using `withUsage` here  b4b38a9

src/lib/utils/detect-insights-client.ts
tkrugg: ok, it's a fair assumption. I've removed it here: https://github.com/algolia/instantsearch.js/pull/4191/commits/0246b92043145e6821d7c4203bb78b84e15d3cb7

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -90,6 +90,61 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
...
+
+    expect(() => {
+      // eslint-disable-next-line no-new
+      new InstantSearch({
+        indexName: 'indexName',
+        searchClient: createSearchClient(),
+        insightsClient: jest.fn(),
+      });
+    }).not
+      .toWarnDev(`[InstantSearch.js]: InstantSearch detected the insights client in the global scope.
```
francoischalifour: This assertion is outdated now because the message expectation differs from the real one. We could use `.not.toWarnDev()` here to avoid that.

tkrugg: The message has been updated here: https://github.com/algolia/instantsearch.js/pull/4191/commits/64b4133917f57d5bde9e964bd0af5e1cf1388fc1

francoischalifour: For this test _not_ to send a false positive in the future, we should either store the message in a variable to use it in the two tests or simply use `.not.toWarnDev()`.

tkrugg:
> It'd be nice to address [#4191 (comment)](https://github.com/algolia/instantsearch.js/pull/4191#discussion_r353636200) before merging to avoid future false positives.

@francoischalifour Good idea! it's addressed here.

PR #4195, opened 2019-11-13
chore(TS): convert MenuSelect component to TypeScript

Haroenv:
**Summary**

I transformed it as an exercise, so that we have more examples on how to transform class components to functional ones in TS, notably without the prop types and "classCallCheck"

**Result**

MenuSelect still works exactly the same as now, except it's written in TS


src/components/MenuSelect/MenuSelect.tsx
```diff
@@ -0,0 +1,92 @@
...
+    templates: MenuSelectTemplates;
+  };
+};
+
+function MenuSelect({ cssClasses, templateProps, items, refine }: Props) {
+  const { value: selectedValue } = find(items, item => item.isRefined) || {
+    value: '',
+  };
+
+  const rootClassNames = cx(cssClasses.root, {
```
francoischalifour: We can inline this class name.

PR #4196, opened 2019-11-13
refactor(utils): migrate remaining utils to TypeScript

Haroenv:
**Summary**

Some utils were written in JS still, so now they can be written in TypeScript instead

**Result**

All utils, except renderTemplate are in TS now
removed unused utils folder


src/types/templates.ts
```diff
@@ -0,0 +1,7 @@
+type StringTemplate = string;
+
+type FunctionTemplate = (data: unknown) => string;
+
+export type Templates = {
+  [key: string]: StringTemplate | FunctionTemplate;
```
francoischalifour: Templates are already typed in a generic way:

https://github.com/algolia/instantsearch.js/blob/c7c53de037959070422c7be85882423078dc1253/src/types/widget.ts#L194-L196

Haroenv: ah I missed it. I'll use that one then

src/lib/utils/prepareTemplateProps.ts
```diff
@@ -1,6 +1,28 @@
 import uniq from './uniq';
+import { Template } from '../../types';
 
-function prepareTemplates(defaultTemplates = {}, templates = {}) {
+type TemplatesConfig = object;
+
+type Templates = {
+  [key: string]: Template;
```
francoischalifour: We don't leverage the generics here but not sure we need them.

Haroenv: Yes, as far as I can tell we don't know anything about the type in this place, so I left it like this. Future usage (unlikely) can precise this further

PR #4197, opened 2019-11-13
fix(insights): detect clicks on children of [data-insights] HTML elements

tkrugg:
**Summary**

When the insights integration is used in the following way:

```js
<a href="#" ${insights('clickObjectIDsAfterSearch', {...})}>
   <h1> ${hit.name} </h1>
</a>
```
and click occurs on `h1`, insight client is not called.

This PR addresses this problem.


src/lib/insights/listener.tsx
```diff
@@ -9,23 +9,37 @@ type WithInsightsListenerProps = {
...
-
       if (!props.insights) {
         throw new Error(
           'The `insightsClient` option has not been provided to `instantsearch`.'
         );
       }
-
-      const { method, payload } = readDataAttributes(
-        event.target as HTMLElement
+      const realTarget = findRealTarget(
```
tkrugg: A better naming for this variable is welcome 😄

src/lib/__tests__/insights-listener-test.tsx
```diff
@@ -61,6 +61,62 @@ describe('withInsightsListener', () => {
...
+    const HitsWithInsightsListener: any = withInsightsListener(Hits);
+    const { container } = render(
+      <HitsWithInsightsListener
+        hits={hits}
+        results={results}
+        instantSearchInstance={instantSearchInstance}
+        insights={insights}
+      />
+    );
+    const innerChild = container.querySelector('h1') as HTMLElement;
```
francoischalifour:
```suggestion
    const innerChild = container.querySelector<HTMLElement>('h1');
```

src/lib/insights/listener.tsx
francoischalifour: `insightsRoot` seems fine.

src/lib/insights/listener.tsx
```diff
@@ -9,23 +9,37 @@ type WithInsightsListenerProps = {
   insights: InsightsClientWrapper;
 };
 
+const findRealTarget = (
```
francoischalifour: We could name this `findInsightsRoot`.

src/lib/insights/listener.tsx
```diff
@@ -9,23 +9,37 @@ type WithInsightsListenerProps = {
...
+  startElement: HTMLElement | null,
+  endElement: HTMLElement | null
+): HTMLElement | null => {
+  let element: HTMLElement | null = startElement;
```
francoischalifour:
```suggestion
  let element = startElement;
```

src/lib/insights/listener.tsx
tkrugg: what would `root` mean here?

francoischalifour: What `findRealTarget` (current naming) is doing here is resolving the root HTML element where data attributes are defined. The root is where the data needs to be read from.

`insightsRootTarget` if you want to keep the "target" name.

tkrugg: Ok I understand, but actually the code we're adding here is trying to find a DOM elements (not necessarily a root element) that carries the data-insights attribute, so I'm not sure how `root` adds meaning here.
Is `insightsTarget` fine or would it be something missing?

francoischalifour: They're both fine. Root just means it's the highest parent in this context.

tkrugg: ok, renamed to `insightsTarget` in 115c932f

src/lib/insights/listener.tsx
tkrugg: renamed to `findInsightsTarget` in 115c932f as agreed at https://github.com/algolia/instantsearch.js/pull/4197#discussion_r351394715

Issue #4102, opened 2019-09-03
Use a more typesafe pattern for "static" values in render parameters


Haroenv:
Currently we use three patterns for storing functions like `refine` etc. to both init and render. We need to make sure we never reassign the function reference directly, because that means if someone with a custom connector has an event listener in a conditional, they "lose" the reference (this is the same problem as with useEffect in React, for which they have the dependency array).

```js
const customWidget = connectSomething(
  ({ refine, widgetParams: { container } }, firstRender) => {
    if (firstRender) {
      const btn = document.createElement('button');
      container.appendChild(btn);
      btn.addEventListener('click', () => refine());
    }
  }
);
```

this previous code only works, because we make sure the "refine" reference can become state without issue, and always stays the same for us.

The problem is that we use many methods to do this. As far as I can tell we have the following methods:

1. "public" method & reassign private method

```js
const widget = {
  _refine() {
    return refine();
  },
  init() {
    refine = xxx;
    renderFn({
      refine: this._refine,
    });
  },
  render() {
    renderFn({
      refine: this._refine,
    });
  },
};
```

2. "private" container & reassign within that

I'm not actually sure if this works as we think it does, we need to check that this doesn't erase the reference

```js
const createWidget = () => {
  const connectorState = {};
  return {
    init() {
      connectorState.refine = xxx;
      renderFn({
        refine: connectorState.refine,
      });
    },
    render() {
      renderFn({
        refine: connectorState.refine,
      });
    },
  };
};
```

3. Set in `init` and never override

```js
const widget = {
  init() {
    this._refine = xxx;
    renderFn({
      refine: this._refine,
    });
  },
  render() {
    renderFn({
      refine: this._refine,
    });
  },
};
```

---

Personally I think that all of these methods have big flaws, because they're:

1. inconsistent
2. hard to remember
3. hard to test for

A good solution would be more automatic, harder to forget, easy to convert all functions to.

There's two suggestions I have:

1. remove `init`

We don't really have much use for it (we don't use this function to render in vanilla, Vue, nor Angular), and could do with the simplicity of purely a render function. The downside is that this would become hard to pass `isFirstRender`, but that can be solved like this:

```js
const createWidget = () => {
  let isFirstRender = true;
  return {
    render() {
      renderFn({}, isFirstRender);
      // maybe this can be done more efficient if we see it hurts perf
      isFirstRender = true;
    },
  };
};
```

1. return an object in `init`.

[...]

PR #4199, opened 2019-11-14
chore(ts): migrate connectPagination

Haroenv:
**Summary**

migrate both the Paginator class and connectPagination. Mostly copied from other connectors. I wish we had #4102 and less boilerplate though

**Result**

connectPagination is fully available from TS


src/widgets/pagination/__tests__/pagination-test.js
```diff
@@ -73,7 +73,7 @@ describe('pagination()', () => {
   });
 
   it('sets the page', () => {
-    widget.refine(helper, 42);
+    widget.refine(42);
```
tkrugg: was this a bug that was there all this time?
There is another call to refine to fix on [line 104](https://github.com/algolia/instantsearch.js/pull/4199/files#diff-d23004038675fb0d352846b17da97790L104)

src/connectors/pagination/Paginator.ts
```diff
@@ -31,11 +41,12 @@ class Paginator {
     return range({ start: first, end: last });
   }
 
-  nbPagesDisplayed(padding, total) {
-    return Math.min(2 * padding + 1, total);
-  }
-
-  calculatePaddingLeft(current, padding, total, totalDisplayedPages) {
+  public calculatePaddingLeft(
```
tkrugg: This is rather an internal method no?
It's only used within this class (in `pages`) so it can be private?

src/connectors/pagination/Paginator.ts
```diff
@@ -47,11 +58,15 @@ class Paginator {
     return padding;
   }
 
-  isLastPage() {
+  public nbPagesDisplayed(padding: number, total: number) {
```
tkrugg: This looks like it could be private too since it's only used within this class?

src/widgets/pagination/__tests__/pagination-test.js
Haroenv: this was a bug in the test, yes. I'll fix the other occurrance too.

src/connectors/pagination/Paginator.ts
Haroenv:
```suggestion
  private nbPagesDisplayed(padding: number, total: number) {
```

src/connectors/pagination/Paginator.ts
Haroenv:
```suggestion
  private calculatePaddingLeft(
```

src/connectors/pagination/connectPagination.ts
```diff
@@ -0,0 +1,191 @@
...
+   */
+  createURL: (page: number) => string;
+}
+
+export type PaginationRenderer<TPaginationWidgetParams> = Renderer<
+  PaginationRendererOptions<PaginationConnectorParams & TPaginationWidgetParams>
+>;
+
+export type PaginationWidgetFactory<TPaginationWidgetParams> = WidgetFactory<
+  PaginationConnectorParams & TPaginationWidgetParams
```
tkrugg: I would have expected to read here, am I missing something?
```suggestion
  TPaginationWidgetParams
```

src/connectors/pagination/connectPagination.ts
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * When using the pagination with Algolia, you should be aware that the engine won't provide you pages
+ * beyond the 1000th hits by default. You can find more information on the [Algolia documentation](https://www.algolia.com/doc/guides/searching/pagination/#pagination-limitations).
+ */
+const connectPagination: PaginationConnector = (renderFn, unmountFn = noop) => {
+  checkRendering(renderFn, withUsage());
+
+  return widgetParams => {
+    const { totalPages, padding = 3 } =
+      widgetParams || ({} as PaginationConnectorParams);
```
tkrugg: I would have expected this value to be casted as
```suggestion
      widgetParams || ({} as TPaginationWidgetParams);
```

src/connectors/pagination/connectPagination.ts
```diff
@@ -0,0 +1,191 @@
...
+  /**
+   * true if the current page is also the last page
+   */
+  isLastPage: boolean;
+  /**
+   * Sets the current page and trigger a search.
+   */
+  refine: (page: number) => void;
+  /**
+   * createURL Creates URLs for the next state, the number is the page to generate the URL for.
```
francoischalifour:
```suggestion
   * Creates URLs for the next state, the number is the page to generate the URL for.
```

src/connectors/pagination/connectPagination.ts
```diff
@@ -0,0 +1,191 @@
...
+      getWidgetSearchParameters(searchParameters, { uiState }) {
+        const page = uiState.page ? uiState.page - 1 : 0;
+
+        return searchParameters.setQueryParameter('page', page);
+      },
+
+      // for tests, remove when refactored
+      refine: page => connectorState.refine!(page),
+      createURL: helper => connectorState.createURL!(helper),
+      getMaxPage,
```
tkrugg: I would have prefered we include the refactoring of the tests before merging, because I sense it's the kind of thing we end up never doing.
Is there a reason (other than PR bloat) not to include them in this diff?

Haroenv: PR bloat and I wasn’t sure what the alternative way of testing this could be.

src/connectors/pagination/connectPagination.ts
```diff
@@ -0,0 +1,191 @@
...
+  PaginationRendererOptions<PaginationConnectorParams & TPaginationWidgetParams>
+>;
+
+export type PaginationWidgetFactory<TPaginationWidgetParams> = WidgetFactory<
+  PaginationConnectorParams & TPaginationWidgetParams
+>;
+
+export type PaginationConnector = <TPaginationWidgetParams>(
+  render: PaginationRenderer<TPaginationWidgetParams>,
+  unmount?: () => void
```
francoischalifour: There's a type called `Unmounter`.

src/connectors/pagination/connectPagination.ts
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * When using the pagination with Algolia, you should be aware that the engine won't provide you pages
+ * beyond the 1000th hits by default. You can find more information on the [Algolia documentation](https://www.algolia.com/doc/guides/searching/pagination/#pagination-limitations).
+ */
+const connectPagination: PaginationConnector = (renderFn, unmountFn = noop) => {
+  checkRendering(renderFn, withUsage());
+
+  return widgetParams => {
+    const { totalPages, padding = 3 } =
+      widgetParams || ({} as PaginationConnectorParams);
```
francoischalifour:
```suggestion
      widgetParams || ({} as typeof widgetParams);
```

src/connectors/pagination/connectPagination.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    const getMaxPage = ({ nbPages }) =>
+      totalPages !== undefined ? Math.min(totalPages, nbPages) : nbPages;
+
+    return {
+      $$type: 'ais.pagination',
+
+      init({ helper, createURL, instantSearchInstance }) {
+        connectorState.refine = page => {
+          helper.setPage(page);
```
francoischalifour: We can chain the calls.

src/connectors/pagination/connectPagination.ts
```diff
@@ -0,0 +1,191 @@
...
+          helper.search();
+        };
+
+        connectorState.createURL = state => page =>
+          createURL(state.setPage(page));
+
+        renderFn(
+          {
+            createURL: connectorState.createURL!(helper.state),
```
francoischalifour:
```suggestion
            createURL: connectorState.createURL(helper.state),
```

src/connectors/pagination/connectPagination.ts
```diff
@@ -0,0 +1,191 @@
...
+            currentRefinement: helper.state.page || 0,
+            nbHits: 0,
+            nbPages: 0,
+            pages: [],
+            isFirstPage: true,
+            isLastPage: true,
+            refine: connectorState.refine!,
```
francoischalifour:
```suggestion
            refine: connectorState.refine,
```

src/connectors/pagination/connectPagination.ts
francoischalifour: I agree with @tkrugg.

There's no need to test `widget.refine()` at the widget level. We should rather mount the widget and interact with it to trigger the expected behaviors.

src/connectors/pagination/connectPagination.ts
Haroenv: I copied from QueryRules connector, meaning that that one is wrong too

src/widgets/pagination/__tests__/pagination-test.js
francoischalifour: After fully reviewing, this test doesn't make sense here in the end. See https://github.com/algolia/instantsearch.js/pull/4199#discussion_r349163769.

src/connectors/pagination/connectPagination.ts
francoischalifour: It's not wrong, it was introduced meanwhile.

Issue #4141, opened 2019-09-25
queryHook is broken 


samouss:
**Describe the bug 🐛**

The option `queryHook` on `SearchBox` is broken since the version [3.2.0](https://github.com/algolia/instantsearch.js/blob/develop/CHANGELOG.md#320-2019-03-14). The widget is not updated correctly, stalled data override the latest keystrokes. I've tested with version 3.1.1 and we don't have the issue, it made me think that the `SearchBox` rewrite is the problem.

**To Reproduce 🔍**

1. Go to https://codesandbox.io/s/instantsearchjs-queryhook-issue-n04f7
2. Enable the network throttling
3. Type into the `SearchBox`
4. After some delay the latest character are overridden

Without the network throttling, the issue is present too. But it's easier to reproduce with it. 

**Screenshots 🖥**

![Untitled](https://user-images.githubusercontent.com/6513513/65584414-22da0700-df81-11e9-812e-6fee6a4e6642.gif)

PR #4202, opened 2019-11-20
fix(queryHook): restore behaviour of queryHook

tkrugg:
**Summary**

There has been many complaints from user on our [debouncing guide](https://www.algolia.com/doc/guides/building-search-ui/going-further/improve-performance/js/#debouncing) no longer working as expected.

This issue was introduced when the SearchBox component was converted into a Preact, and the following condition removed:
https://github.com/algolia/instantsearch.js/commit/c073a9acb51fff3c15278fcd563e47fec55c8365#diff-530222e0c4597f2110dc6ba173a306b0L98

- **Before SearchBox refactor**:
  - when input was focused, the query exposed by the connector didn't update the input value
  - when input was not focused, the query exposed by the connector did update the input value

- **After SearchBox refactor**:
  - regardless on whether input was focused, the query exposed by the connector updates the input value

This PR restores the logic as found before the refactor.

**Result**

fixes https://github.com/algolia/instantsearch.js/issues/4141

**How to test this**:
- Checkout the current branch
- `yarn  start` to start the storybook
- go to "http://localhost:6006/?path=/story/basics-searchbox--with-debounced-queryhook" (or [online](https://deploy-preview-4202--instantsearchjs.netlify.com/stories/?path=/story/basics-searchbox--with-debounced-queryhook))
- set your network throttling to `Fast 3G`.
- type really fast!

**Before**

**After**


src/components/SearchBox/SearchBox.js
```diff
@@ -72,13 +73,18 @@ class SearchBox extends Component {
...
       refine(query);
-    } else {
-      this.setState({ query });
     }
+    this.setState({ query });
 
     onChange(event);
   };
 
+  componentWillReceiveProps(nextProps) {
```
Haroenv: i researched, and it doesn't seem like this method is deprecated in preact.

src/components/SearchBox/SearchBox.js
```diff
@@ -72,13 +73,18 @@ class SearchBox extends Component {
...
+    if (!this.state.focused && nextProps.query !== this.state.query) {
```
Haroenv: why can't we always apply the new query? (even when not focused)

Haroenv: never mind, i didn't see the `!`

src/components/SearchBox/SearchBox.js
tkrugg: yes, I found the same.
Also I tried using getDerivedPropsFromState. It worked in the browser, but it seems the renderer we use in the tests does not fire it, and a lot of tests failed as a result.

src/components/SearchBox/SearchBox.js
francoischalifour: We should add an explanation that mentions why this branch is necessary (because of the `queryHook` option). This is not straightforward to understand.

src/components/SearchBox/__tests__/SearchBox-test.js
```diff
@@ -184,6 +184,37 @@ describe('SearchBox', () => {
   });
 
   describe('Events', () => {
+    describe('focus/blur', () => {
```
francoischalifour: I'd add a note saying that this test suite is necessary for `queryHook` to work as documented.

tkrugg: Does the next test description specify this? 
> - does not derive value from prop when focused
> - derives value from prop when not focused

src/components/SearchBox/SearchBox.js
tkrugg: agreed, it's done here: e205c2169766de15b853f3772edd7c674855510e

src/components/SearchBox/__tests__/SearchBox-test.js
francoischalifour: This is a technical definition. I would mention in a comment something along:

```suggestion
    // This test suite is necessary for the `queryHook` option to be synced
    // with InstantSearch.
    // The query coming from the refinements should not take over the query
    // from the state when the input is focused because it would override
    // the query coming from the `queryHook` function.
    describe('focus/blur', () => {
```

tkrugg: I think this formulation is vague, and does not explain what's happening: `the queryHook to be synced`, `take over the query from the state`.
Stated more precisely, the `query` is not "coming" from the `queryHook`, the `query` is exposed by the connector.

How about this?

> This makes sure we don't override the user's input while (s)he's typing. This issue is more obvious when using `queryHook` to add a debouncing.

francoischalifour: Yeah go ahead! If I can't phrase it clearly it means it definitely needs some explanation!

- `(s)he's typing` → `they're typing`

tkrugg: ok, done here: 51e21a5eaf865fbad19eae9982fcd7ef2b16e62c

PR #4203, opened 2019-11-21
chore(stories): use `addWidgets()` instead of `addWidget()`

francoischalifour:
We still had some occurrences of `addWidget()` in the stories while it's deprecated in favor of `addWidgets()` now. This resulted in warnings when running the Storybook in development mode.


Haroenv:
this is a duplicate of https://github.com/algolia/instantsearch.js/pull/4153/files :)

francoischalifour:
@Haroenv No it's not:

https://github.com/algolia/instantsearch.js/blob/bf0e57632fae3328fc8e6aead6ea39beb8fcc738/.storybook/decorators/withHits.ts#L43-L49

PR #4204, opened 2019-11-22
refactor(showMore): introduce `createShowMore` enhancer

francoischalifour:
This is me again, with another proposal similar to #4128 to share logic between connectors.

## Context

The InfiniteHits has been receiving some frustrations lately because we've made it more opinionated since we support [`showPrevious`](https://www.algolia.com/doc/api-reference/widgets/infinite-hits/js/#widget-param-showprevious) and users cannot recreate the widget easily.

Loading more items is _really_ complicated in our current codebase because we need to think about caching, paging, etc. I believe having a pattern like this enhancer proposal is a first step before wanting to implement it in other widgets.

This PR _is not_ about loading more items, it's about the **show more behavior**. This pattern is slightly easier than loading more items, which is why I start with this one.

## Why

As explained in #4128, the show more behavior was duplicated in all connectors and was hard to keep track of during the widget lifecycle.

This PR is a proposal to use "enhancers" that take care of managing this scoped state of each connector.

## Solution

This is a rework of #4128 with a simpler API that in the end resembles [React hooks](https://reactjs.org/docs/hooks-intro.html).

### Usage

```ts
function connectHierarchicalMenu(renderFn, unmountFn) {
  return widgetParams => {
    const {
      // ...
      limit = 10,
      showMoreLimit = 20,
    } = widgetParams;

    // We get the updated state of the show more behavior with as initial state
    // the `limit` and `showMoreLimit` coming from the connector options.
    const {
      toggleShowMore,
      setToggleShowMore,
      isShowingMore,
      getCurrentLimit,
    } = createShowMore({
      limit,
      showMoreLimit,
    });

    return {
      // ...

      init() {
        renderFn({
          toggleShowMore,
          isShowingMore: isShowingMore(),
        });
      },

      render(renderOptions) {
        setToggleShowMore(() => this.render(renderOptions));

        renderFn({
          toggleShowMore,
          isShowingMore: isShowingMore(),
          items: items.slice(0, getCurrentLimit()),
        });
      },
    };
  };
}
```

### Reference

```ts
interface ShowMoreState {
  isShowingMore: boolean;
  limit: number | null;
  showMoreLimit: number | null;
}

createShowMore(initialState?: Partial<ShowMoreState>)

// =>
//  {
//    toggleShowMore,
//    setToggleShowMore,
//    isShowingMore,
//    getCurrentLimit,
//  }
```

## Next steps

The previous proposal (#4128) received some push back because we thought we didn't need such a solution before removing the `init` lifecycle step. I believe we do.

**If we expose this kind of enhancers to users, they will be able to create their own advanced widgets easily**. Think about creating enhancers for:

- Search For Facet Values
- Caching hits
- Show more
- etc.

If you're keen on trying this new pattern, we can try to refactor the "show previous" and "show next" behavior in an enhancer so that users can leverage it to overcome the frustrations they have with the current InfiniteHits widget.

### DIFF
--- a/src/connectors/hierarchical-menu/connectHierarchicalMenu.js
+++ b/src/connectors/hierarchical-menu/connectHierarchicalMenu.js
@@ -1,3 +1,4 @@
+import { createShowMore } from '../../lib/enhancers';
 import {
   checkRendering,
   warning,
@@ -92,31 +93,20 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
     // we use the first attribute name
     const [hierarchicalFacetName] = attributes;
 
+    const {
+      toggleShowMore,
+      setToggleShowMore,
+      isShowingMore,
+      getCurrentLimit,
+    } = createShowMore({
+      limit,
+      showMoreLimit,
+    });
+
     return {
       $$type: 'ais.hierarchicalMenu',
 
-      isShowingMore: false,
-
-      // Provide the same function to the `renderFn` so that way the user
-      // has to only bind it once when `isFirstRendering` for instance
-      toggleShowMore() {},
-      cachedToggleShowMore() {
-        this.toggleShowMore();
-      },
-
-      createToggleShowMore(renderOptions) {
-        return () => {
-          this.isShowingMore = !this.isShowingMore;
-          this.render(renderOptions);
-        };
-      },
-
-      getLimit() {
-        return this.isShowingMore ? showMoreLimit : limit;
-      },
-
       init({ helper, createURL, instantSearchInstance }) {
-        this.cachedToggleShowMore = this.cachedToggleShowMore.bind(this);
         this._refine = function(facetValue) {
           helper.toggleRefinement(hierarchicalFacetName, facetValue).search();
         };
@@ -135,26 +125,17 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
             refine: this._refine,
             instantSearchInstance,
             widgetParams,
-            isShowingMore: false,
-            toggleShowMore: this.cachedToggleShowMore,
+            isShowingMore: isShowingMore(),
+            toggleShowMore,
             canToggleShowMore: false,
           },
           true
         );
       },
 
-      _prepareFacetValues(facetValues, state) {
-        return facetValues
-          .slice(0, this.getLimit())
-          .map(({ name: label, path: value, ...subValue }) => {
-            if (Array.isArray(subValue.data)) {
-              subValue.data = this._prepareFacetValues(subValue.data, state);
-            }
-            return { ...subValue, label, value };
-          });
-      },
-
       render(renderOptions) {
+        setToggleShowMore(this.render.bind(this, renderOptions));
+
         const {
           results,
           state,
@@ -162,11 +143,11 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
           instantSearchInstance,
         } = renderOptions;
 
+        const currentLimit = getCurrentLimit();
         const facetValues =
           results.getFacetValues(hierarchicalFacetName, { sortBy }).data || [];
         const items = transformItems(
-          this._prepareFacetValues(facetValues),
-          state
+          formatFacetValues(facetValues, currentLimit)
         );
 
         // Bind createURL to this specific attribute
@@ -177,19 +158,16 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
         }
 
         const maxValuesPerFacetConfig = state.maxValuesPerFacet;
-        const currentLimit = this.getLimit();
         // If the limit is the max number of facet retrieved it is impossible to know
         // if the facets are exhaustive. The only moment we are sure it is exhaustive
         // is when it is strictly under the number requested unless we know that another
-        // widget has requested more values (maxValuesPerFacet > getLimit()).
+        // widget has requested more values (maxValuesPerFacet > getCurrentLimit()).
         // Because this is used for making the search of facets unable or not, it is important
         // to be conservative here.
         const hasExhaustiveItems =
           maxValuesPerFacetConfig > currentLimit
             ? facetValues.length <= currentLimit
             : facetValues.length < currentLimit;
-
-        this.toggleShowMore = this.createToggleShowMore(renderOptions);
 
         renderFn(
           {
@@ -198,10 +176,10 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
             createURL: _createURL,
             instantSearchInstance,
             widgetParams,
-            isShowingMore: this.isShowingMore,
-            toggleShowMore: this.cachedToggleShowMore,
+            isShowingMore: isShowingMore(),
+            toggleShowMore,
             canToggleShowMore:
-              showMore && (this.isShowingMore || !hasExhaustiveItems),
+              showMore && (isShowingMore() || !hasExhaustiveItems),
           },
           false
         );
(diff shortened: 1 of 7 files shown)


src/lib/enhancers/__tests__/createShowMore-test.ts
```diff
@@ -0,0 +1,54 @@
+import { createShowMore } from '../createShowMore';
+
+describe('createShowMore', () => {
+  test('goes through the lifecycle', () => {
```
Haroenv: what's meant by this test name?

src/lib/enhancers/createShowMore.ts
```diff
@@ -0,0 +1,69 @@
...
+    limit: null,
+    showMoreLimit: null,
+    ...initialState,
+  };
+
+  function toggleShowMore() {
+    state.isShowingMore = !state.isShowingMore;
+
+    if (onToggle) {
+      onToggle();
```
Haroenv: where is onToggle used?

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -135,38 +125,29 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
...
           instantSearchInstance,
         } = renderOptions;
 
+        const currentLimit = getCurrentLimit();
         const facetValues =
           results.getFacetValues(hierarchicalFacetName, { sortBy }).data || [];
         const items = transformItems(
-          this._prepareFacetValues(facetValues),
-          state
+          formatFacetValues(facetValues, currentLimit)
```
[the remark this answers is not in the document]
francoischalifour: I'm not sure to follow. WDYM?

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -177,31 +158,28 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
...
             createURL: _createURL,
             instantSearchInstance,
             widgetParams,
-            isShowingMore: this.isShowingMore,
-            toggleShowMore: this.cachedToggleShowMore,
+            isShowingMore: isShowingMore(),
+            toggleShowMore,
             canToggleShowMore:
-              showMore && (this.isShowingMore || !hasExhaustiveItems),
+              showMore && (isShowingMore() || !hasExhaustiveItems),
```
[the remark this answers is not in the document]
francoischalifour: Yes, we could pass the information like this:

```js
// 1.
const {
  toggleShowMore,
  setToggleShowMore,
  isShowingMore,
  getCurrentLimit,
  canToggleShowMore,
} = createShowMore({
  limit,
  showMoreLimit,
  showMore,
});

// 2. 
renderFn(
  {
    canToggleShowMore: canToggleShowMore({ state }), // `state` is the helper state
  }
)
```

src/lib/enhancers/__tests__/createShowMore-test.ts
francoischalifour: It tests that it goes through the subscription lifecycle of the enhancer as expected. We can split this test once we determine the final API.

src/lib/enhancers/createShowMore.ts
francoischalifour: That's the subscriber of the enhancer. It's the render method bound to its render params in the widgets.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -296,3 +274,19 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
     };
   };
 }
+
+function formatFacetValues(facetValues, currentLimit) {
+  return facetValues
+    .slice(0, currentLimit)
```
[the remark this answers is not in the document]
francoischalifour: That'd be optimal. What API would you see?

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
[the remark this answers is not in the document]
Haroenv: the one function both maps & slices, which I feel should be two tasks.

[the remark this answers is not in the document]
francoischalifour: It does it recursively though, so that's tricky.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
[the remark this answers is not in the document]
Haroenv: hmm, the recursiveness in hierarchical indeed poses a problem. That means that you can not really do much more than wrap the slice? Maybe someone else has an idea

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -296,3 +274,19 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
     };
   };
 }
+
```
Haroenv: should https://github.com/algolia/instantsearch.js/pull/4204/files#diff-6ea8d65722bc8ade374d1b33d23d90caR252 not use `getCurrentLimit()` too? 🤔

francoischalifour: No because `showMore` is not passed to the enhancer. It can be used if we address https://github.com/algolia/instantsearch.js/pull/4204#discussion_r350227854 and that we add `getFacetsLimit()` on top of `getCurrentLimit()`.

- `getCurrentLimit()` would get the UI limit
- `getFacetsLimit()` would get the search parameter computed based on whether `showMore` is provided

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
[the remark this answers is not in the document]
eunjae-lee: > hmm, the recursiveness in hierarchical indeed poses a problem. That means that you can not really do much more than wrap the slice? Maybe someone else has an idea

can you elaborate the problem? I don't understand how it can be a problem.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
Haroenv: That seems to be ideal to me

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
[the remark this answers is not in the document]
francoischalifour: Actually the logic is quite different in each connector for now, so we might not want to extract that yet.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
francoischalifour: Done in a4ccbb2ba43233b08b0b126a8ae4d51377e83b8e.

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -96,7 +96,7 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
     const {
       toggleShowMore,
       setToggleShowMore,
-      isShowingMore,
+      getIsShowingMore,
```
eunjae-lee: What do you think of `checkIsShowingMore`?
I always struggle with naming boolean variable and function returning boolean.
Not sure what's the best.

Haroenv: it could maybe also be a getter on the helper, so it can stay `isShowingMore` instead of a function

francoischalifour:
Closing because tracked in algolia/instantsearch-rfcs#32.

PR #4220, opened 2019-11-25
feat(typescript): add declaration files

### LINKED ISSUE #3562: Add typings for TypeScript
Hi!
It is mostly impossible to use instantsearch.js without typing in Angular4+
Please, add them to @types

yannickcr:
**Summary**

Fixes #3562 

Since we recently migrated to TypeScript 3.7 (https://github.com/algolia/instantsearch.js/pull/4097), we can now generate declaration files for the library consumers even if the TypeScript migration is not finished yet.

It is possible thanks to the new [`--declaration` and `--allowJS` support](http://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#--declaration-and---allowjs).

Even with a partial migration we get some pretty good declaration files even if we still have 197 `any` on the public API (the existing JSDocs helped here). What is mainly missing is inline documentation tho, but this is still way better than what currently exists IMHO.

Since we heavily rely on JSDoc in JavaScript files the risk is to have some incorrect types declared, this is not an issue a long as it is not in the public API (we do not use `checkJS` in development).

To check the public API I used [`api-extractor`](https://api-extractor.com) to generate some report files: https://gist.github.com/yannickcr/e46632add9a98578269616613a5bad56 Let me know if you see any issue.

(**Theses files are not part of this PR**, this is just an extraction of what was generated to help to find some incorrect types.)


Haroenv:
Interesting that it all shows (undocumented), while almost all those things have jsdoc. Do you know why that could be?

package.json
```diff
@@ -18,13 +18,15 @@
...
+  "types": "es/index.d.ts",
   "scripts": {
     "start": "yarn run dev",
     "dev": "yarn run storybook",
-    "build": "yarn run build:cjs && yarn run build:es && yarn run build:umd",
+    "build": "yarn run build:cjs && yarn run build:es && yarn build:types && yarn run build:umd",
     "build:umd": "BABEL_ENV=umd rollup --config scripts/rollup/config.js",
     "build:cjs": "BABEL_ENV=cjs babel src --extensions '.js,.ts,.tsx' --out-dir cjs/ --ignore 'src/index.es.ts','**/__tests__','**/__mocks__' --quiet",
     "build:es": "BABEL_ENV=es babel src --extensions '.js,.ts,.tsx' --out-dir es/ --ignore 'src/index.es.ts','**/__tests__','**/__mocks__' --quiet && BABEL_ENV=es babel src/index.es.ts --out-file es/index.js --quiet",
+    "build:types": "tsc -p tsconfig.declaration.json --outDir es/; mv es/index.es.d.ts es/index.d.ts",
```
Haroenv: Does this mean the declaration is only for ESM, not for cjs?

yannickcr: You cannot specify multiple declaration in `types` so we have to choose between CJS and ESM.

But I also noted that, even with the declaration files available, if you import `instantsearch.js` as CJS module you do not have any typing, do you confirm?

yannickcr:
@Haroenv Do you have some examples of things that are documented and shown as `undocumented`? From what I saw most of it are types and very few types are documented in `instantsearch.js`

Haroenv:
I was thinking of e.g. the actual wigets and connectors, but it seems like they're totally missing? Maybe because they aren't re-exported at the root. eg:

- https://github.com/algolia/instantsearch.js/blob/3df7bd6b8e7505be0a66ab9201a9072b33d22ad9/src/connectors/autocomplete/connectAutocomplete.ts#L29
- https://github.com/algolia/instantsearch.js/blob/3df7bd6b8e7505be0a66ab9201a9072b33d22ad9/src/connectors/numeric-menu/connectNumericMenu.ts#L30

yannickcr:
Oh, you're right.

I only generated the report from the `index.d.ts` so we are missing the widgets and connectors. I will properly regenerate the file so we can have a better overview of the public API (this does not affect the declaration files generated in this PR).

yannickcr:
Added the reports for the widgets and connectors to the gist.

The public API surface is waaaayyyy larger than what I thought, I think it will be difficult for us to identify incorrect types even with this flattened declarations :neutral_face: .

francoischalifour:
Exposing wrong types is worse than no types. I think we should only export the types coming from TypeScript files, and continue the TypeScript migration as best effort.

Haroenv:
On a quick glance the types don't seem wrong, but it does seem hard to check these. I wonder if there's an option to create more files (maybe without api extractor?)

Haroenv:
There seem to be duplicate types there though, `Hit` seems  to be there three times?

yannickcr:
@Haroenv There is 3 files in the gist (1 for `src/index.es.ts`, 1 for `src/widgets/index.ts` and 1 for `src/connectors/index.ts`), each of them containing `Hit`.

To get multiple files you can checkout this branch and run `yarn build` and check the `.d.ts` files in the `es` directory. May be it would be clearer?

yannickcr:
@francoischalifour It is also possible, but since we are not sure they are wrong it would be too bad to skip all JSDocs and TypeScript type-check on all JavaScript files from this assumption.

package.json
```diff
@@ -18,13 +18,15 @@
   "jsdelivr": "dist/instantsearch.production.min.js",
   "unpkg": "dist/instantsearch.production.min.js",
   "sideEffects": false,
+  "types": "es/index.d.ts",
   "scripts": {
     "start": "yarn run dev",
     "dev": "yarn run storybook",
-    "build": "yarn run build:cjs && yarn run build:es && yarn run build:umd",
+    "build": "yarn run build:cjs && yarn run build:es && yarn build:types && yarn run build:umd",
```
Haroenv: remove this while experimental

package.json
```diff
@@ -18,13 +18,15 @@
   "jsdelivr": "dist/instantsearch.production.min.js",
   "unpkg": "dist/instantsearch.production.min.js",
   "sideEffects": false,
+  "types": "es/index.d.ts",
```
Haroenv: put this back manually when you are publishing experimental

tsconfig.declaration.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "./tsconfig",
+  "compilerOptions": {
+    "noEmit": false,
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "allowJs": true
```
Haroenv: add a check with `checkJS: true`, `yarn tsc` in another PR (non-breaking)

eunjae-lee:
When we publish the typescript version for the first time manually, we need to make sure the version in `version.ts` is modified like `x.y.z-typescript-experimental`, so that it will have a correct user agent, but it shouldn't be committed to git.

Haroenv:
Correct, for publishing an experimental, the flow would be:

1. `VERSION=4.X.X-typescript.X yarn build`
2. `yarn build:types`
3. ` npm publish --tag typescript`

(or `typescript-experimental`)

yannickcr:
As decided together, we'll publish this as experimental release on the npm tag `experimental-typescript` to avoid potentially breaking users builds (stricter typecheck) while being able to let internal and external users to experiment with the generated typings.

This release will be handled manually at first (with instructions in the `CONTRIBUTING.md` file) and then will be automated with Ship.js.

Additionally to the standard release this experimental one will contains: 
* declaration files (generated with the `yarn build:types` command)
* `types` entry in `package.json`

This PR will be updated accordingly.

In a separate PR, we will add an additional type check with `tsc` and the `allowJs` + `checkJs` flags to `true`. This will allows us to see the remaining errors in the JavaScript files and fix them progressively (by updating JSDoc or migrating JS files to TS). This check will be ran by the CI and will be allowed to fail.

eunjae-lee:
> Correct, for publishing an experimental, the flow would be:
> 
>     1. `VERSION=4.X.X-typescript.X yarn build`
> 
>     2. `yarn build:types`
> 
>     3. ` npm publish --tag typescript`
> 
> 
> (or `typescript-experimental`)

Shouldn't it be something like

1. `yarn build`
2. `npm publish`
3. `VERSION=4.X.X-typescript.X yarn build:types`
4. `npm publish --tag typescript`

Haroenv:
the version is used in the yarn-build

eunjae-lee:
> the version is used in the yarn-build

okay, then it'd be like this, right?

1. `VERSION=4.X.X yarn build`
2. `npm publish`
3. `VERSION=4.X.X-typescript.X yarn build:types`
4. `npm publish --tag typescript`

Haroenv:
I'll put my comments on things I noticed here:

- `connectGeoSearch` exposes a `Connector` type which isn't defined.
- connect hierarchical menu & stats seem to be defined differently, and wasn't passing the general "Connector" type in Angular InstantSearch

CONTRIBUTING.md
```diff
@@ -217,3 +216,21 @@ npm run release -- --beta
 ```
 
 _Make sure to use `npm run` instead of `yarn run` to avoid issues._
+
+#### Experimental TypeScript version
+
+An experimental version containing the TypeScript declaration files is available.
+
+Since some of these declaration files are generated from the JSDoc comments they can still contains some typing errors. This version will stay experimental until we are confident enough in the generated declarations to put them in the stable release.
```
francoischalifour:
```suggestion
Since some of these declaration files are generated from the JSDoc comments, they can contain some typing errors. This version will stay experimental until we are confident enough in the generated declarations to put them in a stable release.
```

CONTRIBUTING.md
```diff
@@ -217,3 +216,21 @@ npm run release -- --beta
 ```
 
 _Make sure to use `npm run` instead of `yarn run` to avoid issues._
+
+#### Experimental TypeScript version
+
+An experimental version containing the TypeScript declaration files is available.
```
francoischalifour:
```suggestion
An experimental version containing the TypeScript declaration files is available on the npm tag `experimental-typescript`.
```

CONTRIBUTING.md
```diff
@@ -217,3 +216,21 @@ npm run release -- --beta
...
+
+Since some of these declaration files are generated from the JSDoc comments they can still contains some typing errors. This version will stay experimental until we are confident enough in the generated declarations to put them in the stable release.
+
+To generate the experimental TypeScript version for a particular release run:
```
francoischalifour:
```suggestion
To generate the experimental TypeScript version for a particular release, run:
```

scripts/release/bump-package-version.js
```diff
@@ -19,7 +19,10 @@ if (!semver.valid(newVersion)) {
   );
 }
 
-if (semver.gte(currentVersion, newVersion)) {
+if (
+  semver.gte(currentVersion, newVersion) &&
+  semver.diff(newVersion, currentVersion) !== 'prerelease'
```
francoischalifour: Why do we need this?

CONTRIBUTING.md
```diff
@@ -217,3 +216,21 @@ npm run release -- --beta
...
+
+```sh
+git checkout v4.X.X
+VERSION=4.X.X-experimental-typescript.X node ./scripts/release/bump-package-version.js
+yarn build
+yarn build:types
+sed -i '/"main":/ a \ "types": "es/index.d.ts",' package.json # Add `types` entry in `package.json`
+npm publish --tag experimental-typescript
+git checkout .
+```
```
francoischalifour: Would it make sense to also push the code of the release on a GitHub tag? This will allow us to track what's been released.

scripts/release/bump-package-version.js
yannickcr: We need this change because in semver `4.X.X-experimental-typescript.X < 4.X.X` and `semver.gte` will then return false. This is a problem if we want to do the TypeScript release after the stable one (what we'll most likely do).

With this change I allow us to publish prereleases of the `currentVersion`.

CONTRIBUTING.md
yannickcr: I don't think so. Since we will not push the changes for `experimental-typescript` to GitHub, the tag would always be on the same commit as for the stable one (ex: on `v4.X.X` in the script above).

scripts/release/bump-package-version.js
francoischalifour: I see – I think that'd be worth adding a comment explaining why this line is here then.

CONTRIBUTING.md
francoischalifour: Are you reluctant to this because it's more work to do on our side or because it doesn't provide value?

Ideally, you should be able to map an npm release with a GitHub tag. If we release without publishing the source code related to this release, it's like we "blindly" release without having the ability to check (even later), what this release contained.

I see this as an improvement (that could be handled by Ship.js later), not as a blocker.

yannickcr: FMPOV it just doesn't provide value.

The source code for `v4.X.X-experimental-typescript.X` will be the same as the one of `v4.X.X` (minus the `version` and `types` changes in `package.json` that are not committed) so no need for a new tag.

tsconfig.declaration.json
yannickcr: Done in https://github.com/algolia/instantsearch.js/pull/4250

PR #4222, opened 2019-11-26
fix(infiniteHits): correct widget options types

francoischalifour:
Some options types were marked as required while they were not. In the stories, we should use the direct import to make sure that we leverage our types and spot this kind of issues.


stories/infinite-hits.stories.ts
```diff
@@ -17,33 +18,11 @@ storiesOf('Results|InfiniteHits', module)
       ]);
     })
   )
-  .add(
```
francoischalifour: I removed this story because it was not working anymore and we unit test this behavior anyway.

Haroenv:
E2E test failure is unexpected here @yannickcr

yannickcr:
Relaunched the tests, it passes now (flacky tests ☹️)

PR #4224, opened 2019-11-28
feat(router): plug router as a middleware

francoischalifour:
This PR introduces the concept of middleware in InstantSearch. It's leveraged internally to plug the routing system.

## Why

When building tools on top of InstantSearch, you had no way to subscribe and detect changes.

Since Federated Search came in v4, you can build more complex apps, but they're sometimes hard to debug. You can now dynamically add and remove indices, which impacts the UI state and the search parameters sent to Algolia. To help debug apps in this situation, I plan to build a dev tool to **visualize the InstantSearch tree**.

This visualizer could be plugged as a Storybook addon (I built a POC based on this PR already) or as a browser extension (think like React Devtools).

Building such integrations was not possible before this middleware system.

## Middleware definition

Middleware can be attached to InstantSearch with the `use` method.

```ts
function router({ instantSearchInstance }) {
  return {
    onStateChange({ state }) {
      // write in the history
    },
    subscribe() {},
    unsubscribe() {},
  };
}

function logger({ instantSearchInstance }) {
  return {
    onStateChange({ state }) {
      // log changes
    },
    subscribe() {},
    unsubscribe() {},
  };
}

  /* ... */
});

search.use(router, logger).start();
```

Middleware can also be attached after the search starts.


src/lib/InstantSearch.ts
```diff
@@ -159,14 +156,6 @@ class InstantSearch extends EventEmitter {
       throw new Error(withUsage('The `searchClient` option is required.'));
     }
 
-    if (typeof (options as any).urlSync !== 'undefined') {
-      throw new Error(
-        withUsage(
-          'The `urlSync` option was removed in InstantSearch.js 3. You may want to use the `routing` option.'
```
Haroenv: we still have a reasonable amount of people who go from InstantSearch v1 -> v4, can we keep this, but as a development warning?

src/lib/InstantSearch.ts
```diff
@@ -234,33 +223,29 @@ See ${createDocumentationLink({
...
-        ...this._routingManager.read(),
-      };
+      return newMiddleware;
+    });
+
+    // If the instance has already started, we directly subscribe the
+    // middleware so they're notified of changes.
+    if (this.started) {
+      newMiddlewareList.forEach(m => {
+        m.subscribe();
```
Haroenv: does the middleware not have an unsubscribe possibility? (could be returned from the subscribe call, not sure if useful)

src/lib/InstantSearch.ts
```diff
@@ -234,33 +223,29 @@ See ${createDocumentationLink({
...
 
-    if (routingOptions) {
-      this._routingManager = new RoutingManager({
-        ...routingOptions,
-        instantSearchInstance: this,
-      });
+  public use(...middleware: Middleware[]): this {
+    const newMiddlewareList = middleware.map(fn => {
+      const newMiddleware = fn({ instantSearchInstance: this });
+      this.middleware.push(newMiddleware);
```
```suggestion
      this.middlewares.push(newMiddleware);
```

(also needs to be updated in more places if this is chosen

src/lib/InstantSearch.ts
Haroenv: Ah no, I see it's separate, makes sense!

src/middleware/createRouter.ts
```diff
@@ -0,0 +1,92 @@
...
+        instantSearchInstance.mainIndex.getWidgetState({})
+      );
+
+      const route = stateMapping.stateToRoute(uiState);
+
+      return router.createURL(route);
+    }
+
+    instantSearchInstance._createURL = topLevelCreateURL;
+    instantSearchInstance._initialUiState = {
```
Haroenv: this makes sense, but I don't see it in the previous implementation, what am I missing?

Haroenv:
I wonder if there also should be either a parameter on InstantSearch or a widget-like interface for Vue InstantSearch etc.

src/lib/InstantSearch.ts
francoischalifour: We already removed this error in #4119. I was surprised it was still here. We might have messed up the history.

src/lib/InstantSearch.ts
francoischalifour: "Middlewares" is not correct english though.

src/middleware/createRouter.ts
francoischalifour: It used to be here:

https://github.com/algolia/instantsearch.js/blob/49ffca2d781d2125250bdcb111d2f1dc0c43e740/src/lib/InstantSearch.ts#L258-L262

It's now part of the middleware. Note that this is a side effect though, so not that great.

francoischalifour:
I think all framework-based flavors (Vue InstantSearch, React InstantSearch, Angular InstantSearch) will have their dedicated APIs but this flexible top-level InstantSearch.js should enable all those to use middleware.

src/middleware/createRouter.ts
```diff
@@ -0,0 +1,92 @@
...
+    .getWidgets()
+    .filter(function(widget): widget is Index {
+      return widget.$$type === 'ais.index';
+    })
+    .forEach(innerIndex => {
+      walk(innerIndex, callback);
+    });
+};
+
+interface RouterProps {
```
yannickcr: The `Routing` type defined in `InstantSearch.ts` could be merged with this interface, no?

src/middleware/createRouter.ts
yannickcr: Could work for now but like you said this is not great. Maybe we could introduce more handlers to address these cases later?

src/middleware/index.ts
```diff
@@ -0,0 +1,13 @@
...
+  onStateChange(options: { state: UiState }): void;
+  subscribe(): void;
+  unsubscribe(): void;
+}
+
+export type Middleware = ({
+  instantSearchInstance: InstantSearch,
+}) => MiddlewareDefinition;
+
+export { createRouter } from './createRouter';
```
yannickcr: Would have gone with simply `router` instead of `createRouter` here. I found this more natural when using it with `this.use()`: `this.use(router(options))`

WDYT?

src/middleware/createRouter.ts
francoischalifour: Yes, I'll think it through. What do you mean by "handlers"?

We could still replicate the logic in the original `InstantSearch.ts` but I was reluctant to it at first because I didn't want this entry to be dependant on the router. The technical solution would be to mutate those in the constructor by importing the basic `stateMapping` and reading it from here. But I wanted it completely detached so that we don't need to import this package if we don't use it. What I'm aiming here is to make the package more easily modular so people load only the parts of the code they need. Do you think that's an OK tradeoff?

src/middleware/index.ts
francoischalifour: Sure! I didn't know if there were any naming convention for middleware.

francoischalifour: Note that the word `router` will be repeated:

```js
router({ router, stateMapping })
```

src/middleware/createRouter.ts
francoischalifour: Indeed.

src/middleware/createRouter.ts
yannickcr: Sorry if I wasn't clear, I was thinking about some additional methods on the object returned by the middleware. Something like this (not tested, just for the idea):

```js
function router({ instantSearchInstance }) {
  return {
    onStateChange() {},
    subscribe() {},
    unsubscribe() {},
    getInitialUiState() {},
  };
}
```

```js
this._initialUiState = this.middleware.reduce((initialUiState, m) => {
  return m.getInitialUiState(initialUiState);
}, {});
```

francoischalifour: I see. Ideally I think we shouldn't give the capability of setting the initial UI state in the middleware though.

I renamed the method `EXEPRIMENTAL_use` for now. We can reiterate on this once we roll it out and find more use cases.

francoischalifour:
The renamed the method to hook middleware `EXPERIMENTAL_use`. This is meant to discourage users to use for now, as it's marked experimental.

PR #4227, opened 2019-11-29
fix(widgets): override connectors' `$$type`

francoischalifour:
Since InstantSearch.js v4, we started introducing the `$$type` property on each widget. This property allows to differentiate index widgets from other widgets, as well as facilitate debugging.

Since the `$$type` property is defined in the connectors, widgets sharing the same connector had the same type. This PR fixes that by overriding the type in the widgets.

This will make it more clear what widgets we're manipulating. This can even be used for development tools (like the index tree visualizer I talked about) and reporting widgets usage in the future.


src/widgets/query-rule-custom-data/__tests__/query-rule-custom-data-test.ts
```diff
@@ -54,6 +54,17 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/query-rule-
...
 `);
     });
+
+    it('is a widget', () => {
+      const container = document.createElement('div');
+      const widget = queryRuleCustomData({ container });
+
+      expect(widget).toEqual(
+        expect.objectContaining({
+          $$type: 'ais.queryRuleCustomData',
```
Haroenv: What about `ais.queryRule.customData`? Kinda like the connect pattern on display name in react

francoischalifour: I think I'd rather see another `$$connectorType` attribute when we need it in the future.

PR #4231, opened 2019-12-02
fix(insights): show docs URL when  is missing.

tkrugg:
**Summary**

When clickAnalytics: true is missing, we already fire an error.
In this PR, we guide the user even more by adding a link to the docs showing how to add it.

**Result**

N/A


src/lib/insights/client.ts
```diff
@@ -33,8 +33,9 @@ const getQueryID = (selectedHits: Hits): string => {
   }
   const queryID = queryIDs[0];
   if (typeof queryID !== 'string') {
+    const docsUrl = `https://www.algolia.com/doc/guides/getting-insights-and-analytics/search-analytics/click-through-and-conversions/how-to/send-click-and-conversion-events-with-instantsearch/js/#connect-instantsearch-with-the-insights-client-for-javascript`;
```
tkrugg: should be use a short url here instead ?

Haroenv: ; I'm not exactly sure if this is the right URL. It's clarifying for this case, but it doesn't seem very permanent

francoischalifour: We should use a URL shortener for these, otherwise:

- bundle size will be impacted
- URL might disappear

tkrugg: Is there a page you can think of that could be more appropriate ?

I thought about just adding the docs from the `configure` `https://www.algolia.com/doc/api-reference/widgets/configure/js/` but I didn't know how to point to the current flavour (for Angular and Vue InstantSearch), whereas we can easily add Angular/Vue/React snippets to https://www.algolia.com/doc/guides/getting-insights-and-analytics/search-analytics/click-through-and-conversions/how-to/send-click-and-conversion-events-with-instantsearch/js/#connect-instantsearch-with-the-insights-client-for-javascript

tkrugg: @francoischalifour ok adding it

tkrugg: done here: https://github.com/algolia/instantsearch.js/pull/4231/commits/1bb98184058d81c4f3f5ddf7b0f846cb66865e4b

src/lib/insights/client.ts
```diff
@@ -34,7 +34,8 @@ const getQueryID = (selectedHits: Hits): string => {
   const queryID = queryIDs[0];
   if (typeof queryID !== 'string') {
     throw new Error(
-      'Could not infer `queryID`. Ensure InstantSearch is configured with `clickAnalytics: true`'
+      `Could not infer \`queryID\`. Ensure InstantSearch \`clickAnalytics: true\` was added with Configure widget.
+https://alg.li/lNiZZ7`
```
francoischalifour:
```suggestion

See: https://alg.li/lNiZZ7`
```

src/lib/insights/client.ts
```diff
@@ -34,7 +34,8 @@ const getQueryID = (selectedHits: Hits): string => {
   const queryID = queryIDs[0];
   if (typeof queryID !== 'string') {
     throw new Error(
-      'Could not infer `queryID`. Ensure InstantSearch is configured with `clickAnalytics: true`'
+      `Could not infer \`queryID\`. Ensure InstantSearch \`clickAnalytics: true\` was added with Configure widget.
```
francoischalifour:
```suggestion
      `Could not infer \`queryID\`. Ensure InstantSearch \`clickAnalytics: true\` was added with the Configure widget.
```

tkrugg:
@francoischalifour that's fixed, thank you.

src/lib/insights/client.ts
```diff
@@ -34,7 +34,9 @@ const getQueryID = (selectedHits: Hits): string => {
   const queryID = queryIDs[0];
   if (typeof queryID !== 'string') {
     throw new Error(
-      'Could not infer `queryID`. Ensure InstantSearch is configured with `clickAnalytics: true`'
+      `Could not infer \`queryID\`. Ensure InstantSearch \`clickAnalytics: true\` was added with the Configure widget.
+
+See: https://alg.li/lNiZZ7`
```
Haroenv: I normally make a  rememberable link, but I think we can do a redo on the url minification another time (something with identifiers and stored together with the code so it can be tested?)

tkrugg: I thought the same but I was undecided on the "rememberable" part as it has to be explicit and short. Maybe you can help:
How about `https://alg.li/click-analytics` ?

Haroenv: Makes sense, I don't think `click-analytics` fits that well here, so let's think about generalising these urls another time

tkrugg: ok, works for me!

PR #4233, opened 2019-12-02
feat(widgets): introduce `configureRelatedItems` as experimental

francoischalifour:
This PR introduces a new set of experimental widgets:

- `EXPERIMENTAL_configureRelatedItems`
- `EXPERIMENTAL_connectConfigureRelatedItems`

Combined with the `hits` component, these widgets allow to create related items experiences as seen in e-commerce and media use cases:

Example

<details>

<summary>See more examples</summary>

Example

Example

</details>

## Usage

Given this hit:

```json
{
  "name": "Amazon - Fire TV Stick with Alexa Voice Remote - Black",
  "description": "Enjoy smart access to videos [...]",
  "brand": "Amazon",
  "categories": ["TV & Home Theater", "Streaming Media Players"],
  "hierarchicalCategories": {
    "lvl0": "TV & Home Theater",
    "lvl1": "TV & Home Theater > Streaming Media Players"
  },
  "type": "Streaming media plyr",
  "price": 39.99,
  "price_range": "1 - 50",
  "image": "https://cdn-demo.algolia.com/bestbuy-0118/5477500_sb.jpg",
  "url": "https://api.bestbuy.com/click/-/5477500/pdp",
  "free_shipping": false,
  "rating": 4,
  "popularity": 21469,
  "objectID": "5477500",
  "__position": 1
}
```

And these options:

```js
configureRelatedItems({
  hit,
  matchingPatterns: {
    brand: { score: 3 },
    type: { score: 10 },
    categories: { score: 2 },
  },
});
```

The following search parameters will be computed:

```json
{
  "facets": [],
  "disjunctiveFacets": [],
  "hierarchicalFacets": [],
  "facetsRefinements": {},
  "facetsExcludes": {},
  "disjunctiveFacetsRefinements": {},
  "numericRefinements": {},
  "tagRefinements": [],
  "hierarchicalFacetsRefinements": {},
  "index": "instant_search",
  "hitsPerPage": 5,
  "sumOrFiltersScores": true,
  "filters": "NOT objectID:5477500",
  "optionalFilters": [
    ["brand:Amazon<score=3>"],
    ["type:Streaming media plyr<score=10>"],
    [
      [
        "categories:TV & Home Theater<score=2>",
        "categories:Streaming Media Players<score=2>"
      ]
    ]
  ]
}
```

[Learn more about how scores work in Algolia →](https://www.algolia.com/doc/guides/managing-results/refine-results/merchandising-and-promoting/in-depth/optional-filter-scoring/)

The search parameters will then be set in the index to display related items with a `hits` widget.

## Implementation

Since this is an experimental widget, I didn't add logic in `connectConfigure` but rather created a connector called `connectConfigureRelatedItems` that Vue InstantSearch and Angular InstantSearch can rely on.

## Preview

[See story →](https://deploy-preview-4233--instantsearchjs.netlify.com/stories/?path=/story/basics-configurerelateditems--default)

## Related

- [RFC](https://github.com/algolia/instantsearch-rfcs/blob/7871af03c80cb6ab87c47d82cc8c1f78da0a5e51/accepted/related-items.md)


src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    const optionalFilters = Object.keys(matchingPatterns).reduce<
+      Array<string | string[]>
+    >((acc, attributeName) => {
+      const attribute = matchingPatterns[attributeName];
+      const attributeValue = hit[attributeName];
+      const attributeScore = attribute.score;
+
+      if (Array.isArray(attributeValue)) {
+        acc.push(
```
tkrugg: we might as well keep `acc` immutable with a `concat`

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,155 @@
...
+            attributeName,
+            attributeValue,
+            attributeScore,
+          })
+        );
+      } else {
+        warning(
+          false,
+          `
+The \`matchingPatterns\` option returned a value of type \`${typeof attributeValue}\` for the "${attributeName}" key. This value was not sent to Algolia because it's not supported by \`optionalFilters\`.
```
tkrugg: since both `typeof {} === "object"` and `typeof [] === "object"` this formulation feels a bit strange
```
The `matchingPatterns` option returned a value of type `object` 
```
I'm not sure how to improve it though

src/widgets/configure-related-items/configure-related-items.ts
```diff
@@ -0,0 +1,16 @@
+import connectConfigureRelatedItems, {
+  ConfigureRelatedItemsConnectorParams,
+} from '../../connectors/configure-related-items/connectConfigureRelatedItems';
+import { WidgetFactory } from '../../types';
+
+type ConfigureRelatedItems = WidgetFactory<
```
tkrugg: actually why not import `ConfigureRelatedItemsWidgetFactory` from `../../connectors/configure-related-items/connectConfigureRelatedItems` here directly ?

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,155 @@
...
+  matchingPatterns: MatchingPatterns;
+  /**
+   * Function called to transform the search parameters generated.
+   */
+  transformSearchParameters?(
+    searchParameters: SearchParameters
+  ): PlainSearchParameters;
+}
+
+type ConfigureConfigureRelatedItemsWidgetFactory<
```
tkrugg: It looks like there a typo here and you meant Configure~Configure~RelatedItemsWidgetFactory

francoischalifour: 8a9b9196f2b360d8675e647dc3fadaac46c367f2

src/widgets/configure-related-items/configure-related-items.ts
francoischalifour: 95b754ec794f39bb3e5c4b457476d255e9741bf4

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
francoischalifour: I used spread: 52a140c69d3165e23e366e67ac430c8f0ca1117c.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
tkrugg: That's a an option. Note that although generic, it doesn't solve the issue for class instances other than Array. `typeof new Date() === "object"`.
Something using `Object.prototype.toString.call(attributeValue)` would be more reliable.

I just noticed we don't say what the types expected: `this value is expected to be a string or an array`. It is useful information for someone seeing this error.

francoischalifour: That's a good first step, we can use `Object.prototype.toString.call` which is more accurate.

About the expected types, that's why I link to the documentation. I can add a sentence that makes it more explicit.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,157 @@
...
+      throw new Error(withUsage('The `hit` option is required.'));
+    }
+
+    if (!matchingPatterns) {
+      throw new Error(withUsage('The `matchingPatterns` option is required.'));
+    }
+
+    const optionalFilters = Object.keys(matchingPatterns).reduce<
+      Array<string | string[]>
+    >((acc, attributeName) => {
```
tkrugg: I know they are fully inferred but I find it more explicit to add types on the callback.
I let you judge if it's necessary.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,157 @@
...
+   * The reference hit to extract the filters from.
+   */
+  hit: ResultHit;
+  /**
+   * The schema to create the optional filters.
+   * Each key represents an attribute from the hit.
+   */
+  matchingPatterns: MatchingPatterns;
+  /**
+   * Function called to transform the search parameters generated.
```
Haroenv:
```suggestion
   * Function called to transform the generated search parameters.
```

Haroenv:
```suggestion
   * Function to transform the generated search parameters.
```

(called is optional here imo)

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,157 @@
...
+}) {
+  return `${attributeName}:${attributeValue}<score=${attributeScore || 1}>`;
+}
+
+const connectConfigureRelatedItems: ConfigureRelatedItemsConnector = (
+  renderFn,
+  unmountFn
+) => {
+  warning(
+    false,
```
Haroenv: ideally you could opt-in to this widget without warning, maybe via an argument?

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,157 @@
...
+    'ConfigureRelatedItems is an experimental widget that is subject to change in next minor versions.'
+  );
+
+  return widgetParams => {
+    const { hit, matchingPatterns, transformSearchParameters = x => x } =
+      widgetParams || ({} as typeof widgetParams);
+
+    if (!hit) {
```
Haroenv: if these errors are done before the destructuring, the cast isn't needed I think.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,157 @@
...
+          // See https://github.com/algolia/algoliasearch-helper-js/pull/754
+          optionalFilters,
+        })
+      ),
+    };
+
+    const makeConfigure = connectConfigure(renderFn, unmountFn);
+
+    return {
+      ...makeConfigure({ searchParameters }),
```
Haroenv: interesting pattern, but seems to make for a clean implementation, so lgtm

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -0,0 +1,157 @@
...
+
+See https://www.algolia.com/doc/api-reference/api-parameters/optionalFilters/
+            `
+      );
+
+      return acc;
+    }, []);
+
+    const searchParameters: PlainSearchParameters = {
+      ...transformSearchParameters(
```
Haroenv: Since we have no API for modifying props, we can indeed assume this all is static, but that will need to be rethought once we have an update api

src/connectors/configure-related-items/__tests__/connectConfigureRelatedItems-test.ts
```diff
@@ -0,0 +1,255 @@
...
+      expect(() => {
+        connectConfigureRelatedItems();
+      }).toWarnDev(
+        '[InstantSearch.js]: ConfigureRelatedItems is an experimental widget that is subject to change in next minor versions.'
+      );
+    });
+  });
+
+  describe('options', () => {
+    test('sets search parameters optionalFilters based on matchingPatterns', () => {
```
Haroenv:
```suggestion
    test('sets the optionalFilters search parameter based on matchingPatterns', () => {
```

src/types/instantsearch.ts
```diff
@@ -64,6 +64,8 @@ export type Hit = {
   __queryID?: string;
 };
 
+export type ResultHit = Omit<Hit, '__position' | '__queryID'>;
```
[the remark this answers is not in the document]
tkrugg: I agree this way of formulating it could be more intuitive.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
yannickcr: IMHO it would be good to always warn about it in development. Users could set the argument, forget about it and then be surprised by a future breaking change in a minor version.

src/types/instantsearch.ts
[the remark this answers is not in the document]
francoischalifour: I'll rename the base one `AlgoliaHit` and the InstantSearch one (with `__position`, etc.) `Hit`.

@tkrugg Here we just don't want users to have to pass a hit containing `__position`. If you pass it, it's still fine because it's satisfied with `[key: string]: any;`.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
francoischalifour: Or name the connector `EXPERIMENTAL_connectConfigureRelatedItems`?

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
francoischalifour: So like most of the connectors.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
francoischalifour: `reduce` is generic and therefore they're typed.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
francoischalifour: Adding more explanation about the warning in 168942fc59b76ed35159f77564126c42813a54a6.

@tkrugg I'll send a separate PR that address more correct types in a more global way.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
francoischalifour: We'll need the cast anyway because we want to support passing an empty object (`{}`).

Haroenv: isn't hit required?

francoischalifour: It is. Not sure to follow what you're suggesting. Can you elaborate what you have in mind?

Note that destructuring is also useful here to assign a default value to `transformSearchParameters`.

Haroenv: If done after the ifs, we don’t need a default object anymore, since it’s always going to be a value. It’s more of a general pattern, so not too important to land with this PR

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
francoischalifour: PR sent: #4249.

src/types/instantsearch.ts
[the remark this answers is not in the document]
francoischalifour: Fixed in 0619e161e9a871a8c5035771c11f0a4d02cbdea5.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
francoischalifour: Implemented in 6644967387b759548dd953acfc27b29e779cdf9e.

stories/configure-related-items.stories.ts
```diff
@@ -0,0 +1,160 @@
...
+    const pagination = connectPagination(
+      ({
+        currentRefinement,
+        refine,
+        isFirstPage,
+        isLastPage,
+        nbPages,
+        widgetParams,
+      }) => {
+        widgetParams.previousPageContainer.innerHTML = '';
```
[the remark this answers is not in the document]
francoischalifour: Yep – I didn't want to write too much code for this pagination widget which was not the main point of the story.

francoischalifour:
The last commit changes the exported names of these widgets:

- `EXPERIMENTAL_configureRelatedItems`
- `EXPERIMENTAL_connectConfigureRelatedItems`

They do _not_ display a warning in the console anymore, because their name is explicit about being experimental.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
francoischalifour: Done in 8c29d7e4f8e1f6f8a9767a8e7c551b03970850ad.

Haroenv:
You need to allow `EXPERIMENTAL_` for `new-cap` in eslint too: "A function with a name starting with an uppercase letter should only be used as a constructor  new-cap"

PR #4250, opened 2019-12-10
chore(typescript): add type-check:js command

yannickcr:
**Summary**

Following https://github.com/algolia/instantsearch.js/pull/4220 this change adds a new `type-check:js` command that run `tsc` on all TypeScript and JavaScript files. This will allow us to see the remaining errors in the JavaScript files and fix them progressively (by updating the JSDocs or migrating JS files to TS).

This command is run by the CI and always pass ([We cannot allow a test to fail on CircleCI](https://discuss.circleci.com/t/can-we-allow-a-job-to-fail-in-a-workflow/23343)) so we do not block the CI with the existing errors.

When all errors will be fixed the `tsconfig.checkjs.json` config will be merged with the default `tsconfig.json` and this command removed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "lint": "eslint --ext .js,.ts,.tsx .",
     "lint:fix": "eslint --ext .js,.ts,.tsx --fix .",
     "type-check": "tsc",
+    "type-check:js": "tsc -p tsconfig.checkjs.json",
     "type-check:watch": "yarn type-check --watch",
     "test": "jest",
     "test:watch": "jest --watch --bail",
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "lint": "eslint --ext .js,.ts,.tsx .",
     "lint:fix": "eslint --ext .js,.ts,.tsx --fix .",
     "type-check": "tsc",
-    "type-check:js": "tsc -p tsconfig.checkjs.json",
+    "type-check:js": "tsc --project tsconfig.checkjs.json",
     "type-check:watch": "yarn type-check --watch",
     "test": "jest",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,6 +33,7 @@
     "lint": "eslint --ext .js,.ts,.tsx .",
     "lint:fix": "eslint --ext .js,.ts,.tsx --fix .",
     "type-check": "tsc",
+    "type-check:js": "tsc -p tsconfig.checkjs.json",
```
francoischalifour: Always nice to be more explicit?

```suggestion
    "type-check:js": "tsc --project tsconfig.checkjs.json",
```

yannickcr:
We cannot allow a test to fail on CircleCI (https://discuss.circleci.com/t/can-we-allow-a-job-to-fail-in-a-workflow/23343). The only solution is to override the return code to make it pass, but we still have the log with the number of errors (less visible but still).

Haroenv:
I meant putting it in a whole separate check until it passes, `ci/circleci: test_typescript_js` (or something like that), and make it not required

yannickcr:
Oh, that's a better solution indeed. I'll update the PR accordingly.

yannickcr:
Done, but one drawback is that all PRs will be marked as failed then :neutral_face:

eunjae-lee:
> Done, but one drawback is that all PRs will be marked as failed then 😐

which will drive us to finish the task...? 😅

yannickcr:
Well that can be something to motivate us :smile:

yannickcr:
Apparently it works with space and parenthesis, I updated the name to "type check js (optional)".

PR #4252, opened 2019-12-10
ci(circle): update job naming

Haroenv:
**Summary**

test naming of the jobs is pretty repetitive, but it can be avoided, as noticed in https://github.com/algolia/instantsearch.js/pull/4250#issuecomment-564073193

**Result**

We need to update the required tests after this, so merge whenever that feels well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ defaults: &defaults
 
 version: 2
 jobs:
-  test_build:
+  'build size':
     <<: *defaults
     steps:
       - checkout
@@ -41,7 +41,7 @@ jobs:
             yarn run build
             yarn run test:size
 
-  test_lint:
+  lint:
     <<: *defaults
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ defaults: &defaults
 
 version: 2
 jobs:
-  test_build:
+  'build size':
```
francoischalifour: We also test the build here.

```suggestion
  'build and size':
```

.circleci/config.yml
```diff
@@ -98,8 +98,8 @@ workflows:
   version: 2
   ci:
     jobs:
-      - test_build
-      - test_unit
-      - test_lint
+      - build size
```
francoischalifour:
```suggestion
      - build and size
```

Haroenv:
There's just one other PR open now (excl renovate), so I'll merge now, and change the required tasks

PR #4254, opened 2019-12-13
fix(insights): move 'insightsClient not provided error' to wrapper level

### LINKED ISSUE #4269: The `insightsClient` option has not been provided to `instantsearch`
When I click anywhere on my search result it show a error on console 

**The `insightsClient` option has not been provided to `instantsearch`** 

The same in your netlify example as well - https://instantsearchjs.netlify.com/examples/e-commerce/search/?query=books

**Steps to reproduce**
1. Search anything
2. Click on the search result items

How can I fix this? 

image

tkrugg:
**Summary**

Prior to this change, this error was fired in the click listeners. As a result, users who used a custom component (and therefore a custom listeners) could never see it (they saw instead: insights is `TypeError: insights is not a function`).

Now the insightsClient wrapper itself is responsible for throwing this error.

This also fixes a bug introduced in https://github.com/algolia/instantsearch.js/pull/4197 when the `if (!props.insights) { ... }` was accidentaly brought on top of the handler.

This is not breaking change.

fixes #4269 

**Result**

[View Story →](https://deploy-preview-4254--instantsearchjs.netlify.com/stories/?path=/story/results-hits--with-insights-helper-when-insightsclient-has-not-been-provided)

### DIFF
--- a/src/lib/__tests__/insights-client-test.ts
+++ b/src/lib/__tests__/insights-client-test.ts
@@ -86,10 +86,11 @@ describe('withInsights', () => {
       expect(renderProps).toHaveProperty('insights');
       expect(renderProps.insights).toBeInstanceOf(Function);
     });
-
-    it('should not expose the insights client wrapper to renderOptions if not passed to instantSearchInstance', () => {
+    it('should expose the insights client wrapper even when insightsClient was not provided', () => {
       const renderFn = jest.fn();
-      const instantSearchInstance = { insightsClient: undefined };
+      const instantSearchInstance = {
+        /* insightsClient was not passed */
+      };
       const results = {
         index: 'theIndex',
         hits: [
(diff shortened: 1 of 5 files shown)


src/lib/__tests__/insights-client-test.ts
```diff
@@ -102,7 +123,14 @@ describe('withInsights', () => {
...
 
       const [renderProps] = renderFn.mock.calls[0];
-      expect(renderProps).not.toHaveProperty('insights');
+      expect(() => {
+        renderProps.insights('clickedObjectIDsAfterSearch', {
+          eventName: 'add to favorites',
+          objectIDs: ['1'],
+        });
+      }).toThrowErrorMatchingInlineSnapshot(
+        `"The \`insightsClient\` option has not been provided to \`instantsearch\`."`
```
eunjae-lee: What if we also provide a url to the doc? Too much?

https://www.algolia.com/doc/guides/getting-insights-and-analytics/search-analytics/click-through-and-conversions/how-to/send-click-and-conversion-events-with-instantsearch/js/

(btw, that's a one long url 😅)

tkrugg: Good call

tkrugg: Actually I added a URL to instantsearch docs. it's a generated URL and just as informative.

stories/hits.stories.js
```diff
@@ -140,4 +140,36 @@ storiesOf('Results|Hits', module)
           action(`[InsightsClient] sent ${method} with payload`)(payload),
       }
     )
+  )
+  .add(
+    'with insights helper when insightsClient has not been provided',
```
francoischalifour: What's the point of this story if it errors and users can't try the feature?

PR #4267, opened 2019-12-18
chore(major): add a dev comment to hasMapMoveSinceLastRefine

Haroenv:
This should go from `hasMapMoveSinceLastRefine` to `hasMapMovedSinceLastRefine`, but changing that now would be a breaking change.

### DIFF
--- a/src/connectors/geo-search/connectGeoSearch.js
+++ b/src/connectors/geo-search/connectGeoSearch.js
@@ -112,6 +112,7 @@ const connectGeoSearch = (renderFn, unmountFn = noop) => {
 
     const widgetState = {
       isRefineOnMapMove: enableRefineOnMapMove,
+      // @MAJOR hasMapMoveSinceLastRefine -> hasMapMovedSinceLastRefine
       hasMapMoveSinceLastRefine: false,
       lastRefinePosition: '',
       lastRefineBoundingBox: '',


src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -112,6 +112,7 @@ const connectGeoSearch = (renderFn, unmountFn = noop) => {
 
     const widgetState = {
       isRefineOnMapMove: enableRefineOnMapMove,
+      // @MAJOR hasMapMoveSinceLastRefine -> hasMapMovedSinceLastRefine
```
tkrugg: it's the first time i see this annotation. How don't forget about it?
We already have todos lurking around, we never actually "end up" working on them.

Haroenv: we have two major markers in InstantSearch.js now (since v4) https://github.com/algolia/instantsearch.js/search?l=TypeScript&q=major&type=

francoischalifour: We used this technique for v3 → v4 and it worked well.

tkrugg: in this example: https://github.com/algolia/instantsearch.js/search?l=TypeScript&q=major&type=Code there are two `@Major` tags which have been there for 8 months (way before v4 release). 

So although `@Major` was added we didn't include it in the v4. This means either:
1. `@Major` should be removed for from those places and should not be set
2. `@Major` is not very effective in marking theses kind of changes.

If it's 2., I'm questioning the effectivity of this process. A more effective approach would be to make a PR and merge it in a v5 branch. It's much better than building up an invisible pile of work that adds up to your next release.

PR #4268, opened 2019-12-18
refactor(panel): migrate Panel widget stack to TypeScript

francoischalifour:
This migrates the Panel widget stack to TypeScript. This is a non-conventional widget that benefits from being typed.

This anticipates new APIs that we might want to introduce in v5 regarding simplifying its usage (perhaps passing UI state on top of the whole render options).

## Changes

- Migrate the Preact class component to a function (TypeScript + Preact class components don't work with our setup which is why hooks were used)
- The tests haven't changed. They were migrated to `@testing-library/preact`

## Impact

There is a small bundle size increase because we now import `preact/hooks`. This will get "absorbed" once we remove classes from the codebase.

## Preview

[See stories →](https://deploy-preview-4268--instantsearchjs.netlify.com/stories/?path=/story/basics-panel--default)


yannickcr:
I would have put the refactoring to hooks in a separate PR, it would make easier for us to see what changed during the migration.

francoischalifour:
@yannickcr I forgot to mention, class components do not work with our TypeScript + Preact setup, so that was not possible.

yannickcr:
Oh, I'm interested in more details :slightly_smiling_face: 

But that's a good reason to make both at the same time then, worth mentioning it in the PR description tho :wink:

src/components/Panel/Panel.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const [isControlled, setIsControlled] = useState<boolean>(false);
+  const bodyRef = useRef<HTMLElement | null>(null);
+
+  useEffect(() => {
+    if (!bodyRef.current) {
+      return;
+    }
+
+    bodyRef.current.appendChild(props.bodyElement);
+  }, [bodyRef, props.bodyElement]);
```
eunjae-lee: If I understand correctly,
whenever either `bodyRef` or `props.bodyElement` changes, this effect runs.
`bodyRef` itself is not going to change, but what if `props.bodyElement` changes?
Will it append it again? or it won't happen by design?

src/widgets/panel/panel.tsx
```diff
@@ -19,76 +80,44 @@ const renderer = ({
...
+  widgetFactory: WidgetFactory<NestedWidgetOptions>
+) => (widgetOptions: NestedWidgetOptions) => Widget;
 
 /**
- * @typedef {Object} PanelWidgetOptions
- * @property {function} [hidden] This function is called on each render to determine from the render options if the panel have to be hidden or not. If the value is `true` the CSS class `noRefinementRoot` is applied and the wrapper is hidden.
- * @property {PanelTemplates} [templates] Templates to use for the widgets.
- * @property {PanelWidgetCSSClasses} [cssClasses] CSS classes to add.
+ * The panel widget wraps other widgets in a consistent panel design.
+ * It also reacts, indicates and sets CSS classes when widgets are no more relevant for refining.
```
Haroenv:
```suggestion
 * It also reacts, indicates and sets CSS classes when widgets are no longer relevant for refining.
```

src/components/Panel/Panel.tsx
francoischalifour: It won't be design but you're right, we should probably clean the effect:

```ts
return () => {
  bodyRef.current.removeChild(props.bodyElement);
};
```

src/components/Panel/Panel.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const [isCollapsed, setIsCollapsed] = useState<boolean>(props.isCollapsed);
+  const [isControlled, setIsControlled] = useState<boolean>(false);
+  const bodyRef = useRef<HTMLElement | null>(null);
+
+  useEffect(() => {
+    if (!bodyRef.current) {
+      return;
+    }
+
+    bodyRef.current.appendChild(props.bodyElement);
```
Haroenv: I'm not sure if simply appending is correct here, do you not need to clear the DOM inside the body before adding a new bodyElement? Seems like we didn't handle that before either though

src/components/Panel/Panel.tsx
francoischalifour: Addressed in e983babb853b56e5ff47f7622d13afa2287392a8.

src/components/Panel/Panel.tsx
Haroenv: dealt with before I could submit my review 🐭 👒

francoischalifour:
@Haroenv Nothing relevant in the tests, I just migrated to `@testing-library/preact`. You can try the [stories](https://deploy-preview-4268--instantsearchjs.netlify.com/stories/?path=/story/basics-panel--default).

PR #4270, opened 2019-12-20
chore(examples): make hero image jpg

Haroenv:
JPG is a more fitting image format for pictures, since it compresses better. Instead of 700 & 790 kB, it's now 80 & 50 kB

This change has already happened on Vue, but not yet React & Angular


Haroenv:
This brought the performance score on lighthouse for the e-commerce demo from 81 -> 87

PR #4279, opened 2020-01-06
feat(insights): add getInsightsAnonymousUserToken helper

tkrugg:
**Summary**

This PR adds getInsightsAnonymousUserToken to instantsearch.js.
Related RFC: https://github.com/algolia/instantsearch-rfcs/pull/27

```js
// usage
  enablePersonalization: true, 
  userToken: instantsearch.getInsightsAnonymousUserToken()
});
```


src/helpers/__tests__/get-insights-anonymous-user-token-test.ts
```diff
@@ -0,0 +1,31 @@
+import getInsightsAnonymousUserToken, {
+  ANONYMOUS_TOKEN_COOKIE_KEY,
+} from '../get-insights-anonymous-user-token';
+
+const DAY = 86400000; /* 1 day in ms*/
+const DATE_TOMORROW = new Date(Date.now() + DAY).toUTCString();
+const DATE_YESTERDAY = new Date(Date.now() - DAY).toUTCString();
+
+const resetCookie = cookieKey => {
+  document.cookie = `${cookieKey}=;expires=Thu, 01-Jan-1970 00:00:01 GMT;`;
```
Haroenv: can you add a test for a malformed cookie, like in search-insights.js?

src/helpers/get-insights-anonymous-user-token.ts
```diff
@@ -0,0 +1,21 @@
+export const ANONYMOUS_TOKEN_COOKIE_KEY = '_ALGOLIA';
```
Haroenv: this is exported just for the test right?

src/helpers/get-insights-anonymous-user-token.ts
```diff
@@ -0,0 +1,21 @@
...
+  for (let i = 0; i < cookies.length; i++) {
+    let cookie = cookies[i];
+    while (cookie.charAt(0) === ' ') {
+      cookie = cookie.substring(1);
+    }
+    if (cookie.indexOf(prefix) === 0) {
+      return cookie.substring(prefix.length, cookie.length);
+    }
+  }
+  return '';
```
Haroenv: should we not return undefined here instead?

src/helpers/get-insights-anonymous-user-token.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export default function getInsightsAnonymousUserToken(): string | null {
+  const anonymousUserToken = getCookie(ANONYMOUS_TOKEN_COOKIE_KEY);
+  return anonymousUserToken || null;
```
Haroenv: I'd keep it as undefined in the return, I think the Algolia API will throw if you pass it `null`, but it will be fine on `undefined`, since that doesn't get serialised.

src/helpers/__tests__/get-insights-anonymous-user-token-test.ts
tkrugg: excellent point, added here: a68cb7a4

src/helpers/get-insights-anonymous-user-token.ts
tkrugg: yes

src/helpers/get-insights-anonymous-user-token.ts
tkrugg: I thought about returning undefined first, but we have this ESlint rule `consistent-return`.
I can override it, would you recommend it in this case?

src/helpers/get-insights-anonymous-user-token.ts
tkrugg: same here, returning `null` here instead of `undefined` because  ESlint rule `consistent-return`.

src/helpers/get-insights-anonymous-user-token.ts
francoischalifour: If you explicitly return `undefined` it should work.

tkrugg: awesome! thanks!

tkrugg: fixed aee4083

src/helpers/get-insights-anonymous-user-token.ts
tkrugg: fixed here aee4083

src/helpers/__tests__/get-insights-anonymous-user-token-test.ts
```diff
@@ -0,0 +1,37 @@
...
+    document.cookie = `${ANONYMOUS_TOKEN_COOKIE_KEY}=anonymous-uuid;expires=${DATE_YESTERDAY};`;
+    expect(getInsightsAnonymousUserToken()).toBe(undefined);
+  });
+
+  it('should return the anonymous uuid when cookie present and valid', () => {
+    document.cookie = `${ANONYMOUS_TOKEN_COOKIE_KEY}=anonymous-uuid;expires=${DATE_TOMORROW};`;
+    expect(getInsightsAnonymousUserToken()).toBe('anonymous-uuid');
+  });
+
+  it("should not care about other cookies and fail if they're malformed", () => {
```
francoischalifour: What does "fail" mean in this test? Do you mean "not fail"?

src/helpers/__tests__/get-insights-anonymous-user-token-test.ts
```diff
@@ -0,0 +1,37 @@
...
+    document.cookie = `${ANONYMOUS_TOKEN_COOKIE_KEY}=anonymous-uuid;expires=${DATE_TOMORROW};`;
```
francoischalifour: I think this line is not intended.

tkrugg: is it intended. 
All cookies "live" in the same string, which needs to be parsed.
The spirit of this test is to check `getCookie(cookieName_1)` will not fail if `cookieName_2` is malformed: 
- it should not care about other cookie 
- it should not fail if other cookies are malformed.
Should the description of the test be improved?

src/helpers/__tests__/get-insights-anonymous-user-token-test.ts
tkrugg: Answered in https://github.com/algolia/instantsearch.js/pull/4279#discussion_r364192158

francoischalifour: What does "fail" mean then? "Throw"?

A suggestion would be:

```suggestion
  it('should return the anonymous uuid when other cookies are invalid', () => {
```

src/helpers/__tests__/get-insights-anonymous-user-token-test.ts
francoischalifour: https://github.com/algolia/instantsearch.js/pull/4279/files#r364202991

PR #4286, opened 2020-01-10
Move from uglify to closure

kristoferbaxter:
Move from Uglify to Closure Compiler

**Result**

`dist/instantsearch.production.min.js: 241.09 KB/63.25 KB` => `dist/instantsearch.production.min.js: 238.58 KB/62.34 KB`


francoischalifour:
Thanks for this initiative!

I don't get the same size results for the Minified + GZipped bundle with [Bundlesize](https://github.com/siddharthkp/bundlesize):

- Before (Uglify): 63.24 KB
- After (Closure): 65.06 KB

This would mean that Uglify leads to a better GZipped bundle.

Is that something that you would have expected?

Haroenv:
It's indeed pretty surprising. Maybe closure needs to be configured to remove more things?

kristoferbaxter:
Apologies for delay: here are some results with different minfiers.

### Results
| compiler + terser | terser | compiler | uglify |
| :--- | :--- | :--- | :--- |
| **62.39KB** | 63.64KB | 64.12KB | 63.24KB |

**Based on this I'd recommend moving to compiler + terser**.

### Future Opportunities

This configuration uses Closure Compiler with `SIMPLE_OPTIMIZATIONS` (see [documentation](https://developers.google.com/closure/compiler/docs/compilation_levels) for differences between optimization levels).

With a few tweaks I was able to build the source with `ADVANCED_OPTIMIZATIONS`, and the result was `56.17KB`. A significant improvement.

To move to `ADVANCED_OPTIMIZATIONS`, you'd want to ensure that valid Closure types or externs are present for the entire source. (This is doable, but significantly more work).

For instance, right now this fails due to invalid syntax usage in `hogan.js`.

```
if (partial.subs) {
        // Make sure we consider parent template now
        if (!partials.stackText) partials.stackText = {};
        for (key in partial.subs) {
          if (!partials.stackText[key]) {
            partials.stackText[key] = (this.activeSub !== undefined && partials.stackText[this.activeSub]) ? partials.stackText[this.activeSub] : this.text;
          }
        }
        template = createSpecializedPartial(template, partial.subs, partial.partials,
          this.stackSubs, this.stackPartials, partials.stackText);
      }
```

Closure would like `key` to be declared (`const key in ...`).

francoischalifour:
Thanks for this analysis! We decided not to move forward with this change for now because it involves some changes in our setup (dependencies, banners in the final bundles, stripping out comments, etc.) and the overall bundle size doesn't decrease significantly.

I'm personally really excited that you made Google Closure Compiler accessible to Rollup users because it was a pain to integrate in a frontend library without this plugin.

Thank you for the PR – we'll reevaluate this tool when we plan to rework on our build script.

PR #4294, opened 2020-01-13
feat(algoliasearch): add support for algoliasearch v4

yannickcr:
**Summary**

This change add support for algoliasearch v4 to InstantSearch.js.

No changes were necessary in the code. Also since the client is mocked in the unit tests we did not add an extra job on CIrcleCI to test with the 2 versions of the client.

The examples were updated to use the latest v4 beta.


examples/calendar-widget/index.html
```diff
@@ -41,7 +41,7 @@ <h3>Dates</h3>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.3.1/jquery.min.js"></script>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.20.1/moment.min.js"></script>
   <script src="https://unpkg.com/BaremetricsCalendar@1.0.11/public/js/Calendar.js"></script>
-  <script src="https://cdn.jsdelivr.net/npm/algoliasearch@3.32.0"></script>
+  <script src="https://cdn.jsdelivr.net/npm/algoliasearch@4.0.0-beta.13/dist/algoliasearch-lite.umd.min.js"></script>
```
francoischalifour: I'm wondering if it'd make sense to change the `jsdelivr` and `unpkg` entries to `algoliasearch/lite`.

// @nunomaduro

nunomaduro: I think we are good point the CDN to the default build.

examples/e-commerce/index.html
```diff
@@ -25,8 +25,7 @@
     <link rel="stylesheet" href="./src/app.css" />
     <link rel="stylesheet" href="./src/app.mobile.css" />
     <link rel="stylesheet" href="./src/widgets/PriceSlider.css" />
-    <script src="https://polyfill.io/v3/polyfill.min.js?features=default,Array.prototype.find,Array.prototype.includes"></script>
-
+    <script src="https://polyfill.io/v3/polyfill.min.js?features=default%2CArray.prototype.find%2CArray.prototype.includes%2CPromise%2CObject.assign%2CObject.entries"></script>
```
tkrugg: IIUC, the new client makes use of new polyfills. 
When it's fully released, it might be worth updating this section for each flavour: https://www.algolia.com/doc/guides/building-search-ui/installation/js/#browser-support

Haroenv: Yes, let's not wait for v4 to update that

package.json
```diff
@@ -90,7 +90,7 @@
     "@wdio/selenium-standalone-service": "5.16.5",
     "@wdio/spec-reporter": "5.16.5",
     "@wdio/static-server-service": "5.16.5",
-    "algoliasearch": "3.35.1",
+    "algoliasearch": ">= 3.35.1 < 5",
```
Haroenv: devDependencies should be pinned

examples/e-commerce/index.html
nunomaduro: Yes. Noted here: https://github.com/algolia/doc/pull/4053.

nunomaduro: Is noted here: https://github.com/algolia/doc/pull/4053.

nunomaduro:
@Haroenv @tkrugg Feedback addressed here: https://github.com/algolia/instantsearch.js/pull/4294/commits/5959b420e65000b4dfb01f7dc41a7f3f87e9ff48.

src/types/instantsearch.ts
```diff
@@ -153,7 +178,20 @@ export type RouteState = {
   [stateKey: string]: any;
 };
 
-export type SearchClient = Pick<
-  AlgoliaSearchClient,
-  'search' | 'searchForFacetValues'
->;
+export type SearchClient = Pick<Client, 'search' | 'searchForFacetValues'>;
```
Haroenv: I feel like these additions could be on their own `algoliasearch` file, so some more explanation on the reasons for the complicated code can be explained. Now it seem to make this file more complicated

nunomaduro: You mean a definitions file inside `instantsearch.js`?

Haroenv: I mean all the definitions related to the client probably should go in an adjacent `algoliasearch.ts` file, since it's not per se related to the instantsearch constructor

Haroenv:
`PlainSearchParameters` seems to have an issue in the helper still

yannickcr:
Yes, we need to new helper v3 to fix this issue (same on React InstantSearch)

src/types/algoliasearch.ts
```diff
@@ -0,0 +1,45 @@
...
+  results: Array<SearchResponse<THit>>;
+}
+
+export type SearchResponse<THit> = ReturnType<
+  typeof algoliasearch
+> extends DummySearchClientV4
+  ? SearchResponseV4<THit>
+  : SearchResponseV3<THit>;
+
+export type SearchForFacetValuesResponse = ReturnType<
```
Haroenv: we can do `Client extends DummySearchClient` here too

PR #4296, opened 2020-01-14
chore(release): put correct reviewer config on release PRs

eunjae-lee:
**Summary**

Previously I put a wrong reviewer config in the `ship.config.js` file.
And Ship.js was handling it in a wrong way as well.
It turned out the API from GitHub takes separate parameters: `reviewers` & `team_reviewers`.

This way, the new release PRs will have the team automatically assigned as team_reviewers.

### DIFF
--- a/ship.config.js
+++ b/ship.config.js
@@ -13,7 +13,7 @@ module.exports = {
   beforeCommitChanges: ({ exec }) => {
     exec('yarn doctoc');
   },
-  pullRequestReviewer: ['@algolia/instantsearch-for-websites'],
+  pullRequestTeamReviewer: ['instantsearch-for-websites'],
   slack: {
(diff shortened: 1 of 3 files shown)


ship.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
   beforeCommitChanges: ({ exec }) => {
     exec('yarn doctoc');
   },
-  pullRequestReviewer: ['@algolia/instantsearch-for-websites'],
+  pullRequestTeamReviewer: ['instantsearch-for-websites'],
```
francoischalifour: Side note: since it's an array I would expect the option to be named `pullRequestTeamReviewers`.

eunjae-lee: Thanks for the feedback. It started as a string at first, so its name was singular. Now that it supports array, I should deprecate the old name and make a new name for later version.

PR #4298, opened 2020-01-14
chore: use tagName instead of releaseTag in link to release in slack notification

jeetiss:
**Summary**

Fix link to release in slack notification as in 
https://github.com/algolia/shipjs/pull/609

### DIFF
--- a/ship.config.js
+++ b/ship.config.js
@@ -22,7 +22,7 @@ module.exports = {
     releaseSuccess: ({
       appName,
       version,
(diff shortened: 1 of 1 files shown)


eunjae-lee:
@jeetiss Hey, nice catch!
Thanks for the PR.

For more context, I've copied and overrode some config to send release message to Slack.
There has been a change in Ship.js, so this PR updates the change here.

PR #4306, opened 2020-01-21
feat(highlight): add cssClasses to snippet & highlight helper

Haroenv:
This is related to: https://github.com/algolia/vue-instantsearch/issues/759. Just adding this doesn't totally solve the situation yet, but it at least allows a hook for adding extra class names.

What's interesting to note is that the suit helper in InstantSearch only concatenates names, and thus the keys for cssClasses are just the modifier. In Vue InstantSearch the keys for classNames are the generated classes, since the suit helper will check those keys directly. I wonder if we could move to such a system in InstantSearch.js in the future. This will allow us to remove these mappings, and deal with it at call site (which likely has a decrease in bundle size)

**Summary**

Before this change, it was impossible to override class names for the highlighted part of an ais-highlight.

**Result**

A new `cssClasses` option to the highlight helper, which accepts a key `highlighted`.


src/helpers/highlight.ts
```diff
@@ -15,14 +18,16 @@ export default function highlight({
...
+  cssClasses = {},
 }: HighlightOptions): string {
   const attributeValue =
     (getPropertyByPath(hit, `_highlightResult.${attribute}.value`) as string) ||
     '';
 
-  const className = suit({
-    descendantName: 'highlighted',
-  });
+  const className =
```
francoischalifour: We can use `cx` here:

```ts
import cx from 'classnames';

// ...

const className = cx(
  suit({
    descendantName: 'highlighted',
  }),
  cssClasses.highlighted
);
```

src/helpers/snippet.ts
```diff
@@ -15,14 +18,16 @@ export default function snippet({
...
+  cssClasses = {},
 }: SnippetOptions): string {
   const attributeValue =
     (getPropertyByPath(hit, `_snippetResult.${attribute}.value`) as string) ||
     '';
 
-  const className = suit({
-    descendantName: 'highlighted',
-  });
+  const className =
```
francoischalifour: We can use `cx` here.

src/helpers/highlight.ts
Haroenv: cx is currently not a dependency to the connectors, using it here would make it a dependency for Vue & Angular

francoischalifour: Ah, makes sense! Might be worth explaining that we don't want to use the `classnames` dependency here because it's used in other flavors, then?

Another suggestion would be to use:

```ts
const className = [
  suit({
    descendantName: 'highlighted',
  }),
  cssClasses.highlighted,
]
  .filter(Boolean)
  .join(' ');
```

That's rather a matter of taste, so feel free to keep it that way.

src/helpers/snippet.ts
Haroenv:
```suggestion
  // cx is not used, since it would be bundled as a dependency for Vue & Angular
  const className =
```

src/helpers/highlight.ts
Haroenv:
```suggestion
  // cx is not used, since it would be bundled as a dependency for Vue & Angular
  const className =
```

PR #4307, opened 2020-01-21
chore(release): publish `experimental-typescript`

eunjae-lee:
**Summary**

This PR updates the release flow to publish `experimental-typescript` build by default after publishing a normal version.

The addition in `ship.config.js` is the following:

```js
  afterPublish: ({ exec }) => {
    exec('./scripts/release/build-experimental-typescript.js');
    exec(
      `yarn publish --no-git-tag-version --non-interactive --tag experimental-typescript`
    );
  },
```

If you want, you can run `./scripts/release/build-experimental-typescript.js` on your computer. It will create a build with the typescript definition files.

To actually publish it, you can run either

`yarn publish --no-git-tag-version --non-interactive --tag experimental-typescript`
or
`npm publish --tag experimental-typescript`

We normally use yarn, so I put the yarn version in `ship.config.js`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -224,10 +224,13 @@ To generate the experimental TypeScript version for a particular release, run:
 
 ```sh
 git checkout v4.X.X
-VERSION=4.X.X-experimental-typescript.X node ./scripts/release/bump-package-version.js
-yarn build
-yarn build:types
-sed -i '/"main":/ a \ "types": "es/index.d.ts",' package.json # Add `types` entry in `package.json`
+./scripts/release/build-experimental-typescript.js
+```
+
+To publish it, run:
+
+```
 npm publish --tag experimental-typescript
-git checkout .
+# or
+yarn publish --no-git-tag-version --non-interactive --tag experimental-typescript
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -220,10 +220,9 @@ An experimental version containing the TypeScript declaration files is available
 
 Since some of these declaration files are generated from the JSDoc comments, they can contain some typing errors. This version will stay experimental until we are confident enough in the generated declarations to put them in a stable release.
 
-To generate the experimental TypeScript version for a particular release, run:
+To generate the experimental TypeScript version for a particular (stable) release, run:
 
 ```sh
-git checkout v4.X.X
 ./scripts/release/build-experimental-typescript.js
 ```
 
--- a/scripts/release/build-experimental-typescript.js
+++ b/scripts/release/build-experimental-typescript.js
@@ -1,0 +1,22 @@
+#!/usr/bin/env node
+
+/* eslint-disable import/no-commonjs */
+const fs = require('fs');
+const path = require('path');
+const shell = require('shelljs');
+
+const packageJsonPath = path.resolve('package.json');
+const packageJson = JSON.parse(fs.readFileSync(packageJsonPath).toString());
+const { version: currentVersion } = packageJson;
+const newVersion = `${currentVersion}-experimental-typescript`;
+packageJson.version = newVersion;
+packageJson.types = 'es/index.d.ts';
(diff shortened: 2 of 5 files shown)


ship.config.js
```diff
@@ -14,6 +14,12 @@ module.exports = {
     exec('yarn doctoc');
   },
   pullRequestTeamReviewer: ['instantsearch-for-websites'],
+  afterPublish: ({ exec }) => {
```
francoischalifour: We don't need to clean Git's state because it's only supposed to be run on CircleCI, right?

CONTRIBUTING.md
```diff
@@ -224,10 +224,13 @@ To generate the experimental TypeScript version for a particular release, run:
 
 ```sh
 git checkout v4.X.X
```
francoischalifour: This is not needed now, right? Besides, it'll fail if the `v4.X.X` branch doesn't previously exist (and it shouldn't exist).

ship.config.js
francoischalifour: Do we want to release on this tag only for stable versions?

scripts/release/build-experimental-typescript.js
```diff
@@ -0,0 +1,22 @@
...
+
+/* eslint-disable import/no-commonjs */
+const fs = require('fs');
+const path = require('path');
+const shell = require('shelljs');
+
+const packageJsonPath = path.resolve('package.json');
+const packageJson = JSON.parse(fs.readFileSync(packageJsonPath).toString());
+const { version: currentVersion } = packageJson;
+const newVersion = `${currentVersion}-experimental-typescript`;
```
francoischalifour: This will fail when we release twice non-stable because multiple versions will be named the same.

- Actual
  - `4.2.0-experimental-typescript`
  - `4.2.0-experimental-typescript`
- Expected
  - `4.2.0-experimental-typescript.0`
  - `4.2.0-experimental-typescript.1`

We can ignore this if we only release for stable releases, but then we should append the version with `.0`.

ship.config.js
Haroenv: Good call, we don't want a prerelease, or a release of a previous major to override this

eunjae-lee: Right, I didn't put code to clean it up (kind of on purpose). And if you ever run this on your computer, you'll see the change.

CONTRIBUTING.md
eunjae-lee: done https://github.com/algolia/instantsearch.js/pull/4307/commits/124555ee339060f7faa7d907df23fcb142e254b2

ship.config.js
eunjae-lee: It makes sense. Now it's releasing types only with v4 stable version.
https://github.com/algolia/instantsearch.js/pull/4307/commits/1e51fea844e2ae6f074efe82a574b1916ec00ea1

eunjae-lee:
I'm merging this even though e2e failed, because there is nothing to do with e2e in this PR since it's updating the release script and a markdown file.

scripts/release/build-experimental-typescript.js
francoischalifour: This comment was not addressed @eunjae-lee.

eunjae-lee: @francoischalifour I thought it's resolved, because I updated this PR to release only for stable releases.

> We can ignore this if we only release for stable releases

https://github.com/algolia/instantsearch.js/pull/4307/commits/1e51fea844e2ae6f074efe82a574b1916ec00ea1

francoischalifour: > [...], but then we should suffix the version with `.0`.

To be confirmed, but I think tags need to be versioned.

Haroenv: Yes, the current syntax isn't valid semver

eunjae-lee: ah.. I get what you mean.

PR #4321, opened 2020-01-27
chore: add VERSION env var before yarn build

eunjae-lee:
**Summary**

This PR fix the release flow to include `VERSION` env var before `yarn build`.

https://unpkg.com/instantsearch.js@4.2.0/dist/instantsearch.production.min.js

> /*! InstantSearch.js UNRELEASED (Thu, 23 Jan 2020 15:32:51 GMT) | © Algolia, Inc. and contributors; MIT License | https://github.com/algolia/instantsearch.js */

With the lack of `VERSION`, the last version was released with `UNRELEASED` comment at the top of the bundle.

Now it's attached to both normal release and experimental-typescript release.

**One question**:
Do we just want to throw an error at rollup when `VERSION` is not given?
Or, do we still need `UNRELEASED` for some reason? If so, what is a valid case for that?

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "scriptjs": "2.5.9",
     "semver": "6.3.0",
     "shelljs": "0.8.3",
-    "shipjs": "0.14.1",
+    "shipjs": "0.14.2",
     "typescript": "3.7.5",
     "webpack": "4.41.5"
   },
--- a/scripts/release/build-experimental-typescript.js
+++ b/scripts/release/build-experimental-typescript.js
@@ -18,5 +18,5 @@ fs.writeFileSync(
   `export default '${newVersion}';\n`
 );
 
-shell.exec('yarn build');
+shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`);
 shell.exec('yarn build:types');
--- a/ship.config.js
+++ b/ship.config.js
@@ -14,6 +14,8 @@ module.exports = {
     exec('yarn doctoc');
   },
   pullRequestTeamReviewer: ['instantsearch-for-websites'],
+  buildCommand: ({ version }) =>
+    `NODE_ENV=production VERSION=${version} yarn build`,
   testCommandBeforeRelease: () => 'echo "No need to test again."',
   afterPublish: ({ exec, version, releaseTag }) => {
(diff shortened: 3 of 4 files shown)


eunjae-lee:
@francoischalifour @Haroenv 

1. Do we just want to throw an error at rollup when VERSION is not given?

2. Or, do we still need UNRELEASED for some reason? If so, what is a valid case for that?

If you're not sure about (2), we could go for (1) first and see if we get any errors. wdyt?

francoischalifour:
Sorry, I forgot to address these comments.

(1) That's a good idea to throw if `VERSION` is not defined when releasing, but not as-is in the Rollup build script because it means that we wouldn't be able to build locally.

(2) Historically we had this "UNRELEASED" tag to be able to see where the build came from, and to tag it as not official.

Does Ship.js inject an environment variable somewhere? CircleCI (and all CIs) [inject `CI=true`](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables) to give context to scripts. We would imagine injecting `RELEASE=true`in Ship.js when the job is eligible to releasing (the commit name matches), and throw in the Rollup script if `RELEASE=true` _and_ `VERSION` is `undefined`.

eunjae-lee:
> Does Ship.js inject an environment variable somewhere? CircleCI (and all CIs) [inject `CI=true`](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables) to give context to scripts. We could imagine injecting `RELEASE=true`in Ship.js when the job is eligible to releasing (the commit name matches), and throw in the Rollup script if `RELEASE=true` _and_ `VERSION` is `undefined`.

It's a good idea to pass an environment variable.

1. I'd name it `SHIPJS` rather than `RELEASE`, because even if I search for `RELEASE`, I wouldn't be able to find it in our code base which will later confuse us or someone in charge who doesn't know this decision.

2. Letting Ship.js to inject the env var vs. Updating our package.json to have 
```js
"scripts": {
  ...
  "release:prepare": "SHIPJS=true shipjs prepare"
}
```
Hmm. After writing this, I guess it'd be better to have in out-of-the-box.

What do you think?

francoischalifour:
Yeah it should be out of the box IMO.

PR #4322, opened 2020-01-28
fix(middleware): subscribe middleware after `init`

francoischalifour:
Subscribing the middleware before `mainIndex.init` is fine for now but it won't be as soon as we expose this API officially and that the middleware relies on a index instance variables and events because they're not yet set.

I stubble across this issue while developing the InstantSearch DevTools, so users could have unintended behaviors if they also build similar middleware.

Note that in the [original implementation of the router](https://github.com/algolia/instantsearch.js/pull/4224/files#diff-1fc36c4ea042d2ecb5a498416e5e0357L421-L429), it was also subscribed after `init`.


Haroenv:
This looks good, but the test isn't really clear _why_ the order matters. Can we try reading a value that would otherwise be undefined?

francoischalifour:
@Haroenv Open to suggestions but I couldn't find an easy way to rely on something better because they're mostly callbacks.

eunjae-lee:
If `why` isn't clear, we could add some comments in the test case to explain why order matters there?

PR #4324, opened 2020-01-31
WIP: pattern router

Haroenv:
This is a proof of concept of a new router, its advantage is that it's way more readable than a classic SEO history router. I will write an RFC on this once there's a first batch of rough feedback.

The example usage is in the media example, so that

Things to think about:

- how do we deal with this dependency?
- is this something we want to offer
- technically: clean up the "omit" part for which params go to query string
- do people want more control over e.g. hash parameters?
- do we need a polyfill of URL in the docs, or do we rewrite it to not need it?
- naming?  we can't both call the parameter "pattern", as well as the router
- does this become an option in historyRouter instead of a separate router?

**Summary**

**Result**

### DIFF
--- a/examples/media/index.html
+++ b/examples/media/index.html
@@ -15,7 +15,7 @@
       rel="stylesheet"
       href="https://fonts.googleapis.com/css?family=Roboto"
     />
-    <link rel="stylesheet" href="main.css" />
+    <link rel="stylesheet" href="/main.css" />
   </head>
 
   <body>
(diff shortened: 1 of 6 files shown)


Haroenv:
Bundlesize impact is too bad, I wonder if it should be shipped as a separate bundle maybe?

eunjae-lee:
👍

devdreamsolution (approved):

Haroenv:
You can now use [@algolia/instantsearch-pattern-router](https://github.com/algolia/instantsearch-pattern-router)

PR #4330, opened 2020-02-03
chore: validate conditions before release

eunjae-lee:
**Summary**

This PR actually does two things but they're small enough, so I put it in a same PR and both of them depends on the new update of Ship.js@0.15.0.

## 1. Missing `VERSION`

When SHIPJS=true and VERSION is undefined, rollup will throw an exception to prevent having a `UNRELEASE` comment in released bundle.

https://unpkg.com/instantsearch.js@4.2.0/dist/instantsearch.production.min.js

## 2. Skipping preparation when there's nothing important

When it's a patch release and if there are only chores but no fixes, then `yarn release:prepare` will skip and won't create a PR. What do you think of this behavior? With rules like this, we can later automate `yarn release:prepare`.

### DIFF
--- a/scripts/rollup/config.js
+++ b/scripts/rollup/config.js
@@ -5,7 +5,7 @@ import replace from 'rollup-plugin-replace';
 import { uglify } from 'rollup-plugin-uglify';
 import filesize from 'rollup-plugin-filesize';
 
-if (process.env.SHIPJS === 'true' && !process.env.VERSION) {
+if (process.env.NODE_ENV === 'production' && !process.env.VERSION) {
   throw new Error(
     'You need to specify an environment variable `VERSION` to run the build process.'
   );
--- a/scripts/rollup/config.js
+++ b/scripts/rollup/config.js
@@ -6,16 +6,13 @@ import { uglify } from 'rollup-plugin-uglify';
 import filesize from 'rollup-plugin-filesize';
 import semver from 'semver';
 
-if (process.env.NODE_ENV === 'production') {
-  if (!process.env.VERSION) {
-    throw new Error(
-      'You need to specify an environment variable `VERSION` to run the build process.'
-    );
-  }
-
-  if (!semver.valid(process.env.VERSION)) {
-    throw new Error(`${process.env.VERSION} is not a valid semver.`);
-  }
+if (
+  process.env.NODE_ENV === 'production' &&
+  !semver.valid(process.env.VERSION)
+) {
+  throw new Error(
+    `You need to specify a valid semver environment variable 'VERSION' to run the build process (received: '${process.env.VERSION}').`
+  );
 }
 
 const version =
--- a/scripts/rollup/config.js
+++ b/scripts/rollup/config.js
@@ -5,6 +5,12 @@ import replace from 'rollup-plugin-replace';
 import { uglify } from 'rollup-plugin-uglify';
 import filesize from 'rollup-plugin-filesize';
 
+if (process.env.SHIPJS === 'true' && !process.env.VERSION) {
+  throw new Error(
+    'You need to specify an environment variable `VERSION` to run the build process.'
+  );
+}
+
 const version =
   process.env.VERSION || `UNRELEASED (${new Date().toUTCString()})`;
 const algolia = '© Algolia, Inc. and contributors; MIT License';
--- a/scripts/rollup/config.js
+++ b/scripts/rollup/config.js
@@ -4,11 +4,18 @@ import babel from 'rollup-plugin-babel';
 import replace from 'rollup-plugin-replace';
 import { uglify } from 'rollup-plugin-uglify';
 import filesize from 'rollup-plugin-filesize';
+import semver from 'semver';
 
-if (process.env.NODE_ENV === 'production' && !process.env.VERSION) {
-  throw new Error(
-    'You need to specify an environment variable `VERSION` to run the build process.'
-  );
+if (process.env.NODE_ENV === 'production') {
+  if (!process.env.VERSION) {
+    throw new Error(
+      'You need to specify an environment variable `VERSION` to run the build process.'
+    );
+  }
+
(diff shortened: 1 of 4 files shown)


ship.config.js
```diff
@@ -4,6 +4,13 @@ const path = require('path');
 
 module.exports = {
   mergeStrategy: { toSameBranch: ['master', 'next'] },
+  shouldPrepare: ({ releaseType, commitNumbersPerType }) => {
+    const { fix = 0 } = commitNumbersPerType;
+    if (releaseType === 'patch' && fix === 0) {
```
Haroenv:
```suggestion
    return releaseType === 'patch' && fix === 0;
```

note sure if that's better or worse tbh 😂

scripts/rollup/config.js
```diff
@@ -5,6 +5,12 @@ import replace from 'rollup-plugin-replace';
 import { uglify } from 'rollup-plugin-uglify';
 import filesize from 'rollup-plugin-filesize';
 
+if (process.env.SHIPJS === 'true' && !process.env.VERSION) {
```
Haroenv: I can't imagine this ever matching again, we'd have to explicitly break the release script no?

eunjae-lee: @Haroenv sorry I don't understand what you mean. Could you rephrase it?

ship.config.js
eunjae-lee: We still need to return true when `releaseType` is 'minor' or 'major' 😅

scripts/rollup/config.js
Haroenv: Since we added the env variable in shipjs configuration, and this check relies on the SHIPJS env variable, the chance that we ever get back into this branch is minimal IMO (much more likely to be forgotten in some other way. What about explicitly making version required, or asking for it interactively if not given (with the option to go for UNRELEASED)

eunjae-lee: > What about explicitly making version required

I don't know what kind of stuff you have in mind,

but after your comment, I read the code again.

```js
if (process.env.SHIPJS === 'true' && !process.env.VERSION) {
```

That condition might not be understandable in a few months.
What if we just change it to

```js
if (process.env.NODE_ENV === 'production' && !process.env.VERSION) {
```

Ship.js prepends `NODE_ENV=production` and our previous release scripts did the same.
Only when we run `yarn build` on local machine, NODE_ENV is not `production`.
So it may be more explicit. What do you think?

scripts/rollup/config.js
```diff
@@ -5,6 +5,12 @@ import replace from 'rollup-plugin-replace';
 import { uglify } from 'rollup-plugin-uglify';
 import filesize from 'rollup-plugin-filesize';
 
+if (process.env.NODE_ENV === 'production' && !process.env.VERSION) {
```
francoischalifour: Since we're at it, we should even check that `process.env.VERSION` follows semver to avoid any surprises. We already use the `semver` package in the codebase. Wdyt?

Haroenv: that's a good idea!

eunjae-lee: On it !

scripts/rollup/config.js
```diff
@@ -4,11 +4,18 @@ import babel from 'rollup-plugin-babel';
...
+import semver from 'semver';
 
-if (process.env.NODE_ENV === 'production' && !process.env.VERSION) {
-  throw new Error(
-    'You need to specify an environment variable `VERSION` to run the build process.'
-  );
+if (process.env.NODE_ENV === 'production') {
+  if (!process.env.VERSION) {
+    throw new Error(
+      'You need to specify an environment variable `VERSION` to run the build process.'
```
francoischalifour: We can merge the conditions to a single with `process.env.NODE_ENV === 'production'` (like previously):

```suggestion
      'You need to specify a valid semver environment variable `VERSION` to run the build process (received ${process.env.VERSION}).'
```

eunjae-lee: It's a good idea. I've updated it.

---

BTW, TIL `1.2.3-beta` is a valid semver.

```
> semver.valid('1.2.3-beta.')
null
> semver.valid('1.2.3-beta')
'1.2.3-beta'
> semver.valid('1.2.3-beta.1')
'1.2.3-beta.1'
```

scripts/rollup/config.js
```diff
@@ -4,6 +4,16 @@ import babel from 'rollup-plugin-babel';
...
 import { uglify } from 'rollup-plugin-uglify';
 import filesize from 'rollup-plugin-filesize';
+import semver from 'semver';
+
+if (
+  process.env.NODE_ENV === 'production' &&
+  !semver.valid(process.env.VERSION)
+) {
+  throw new Error(
+    `You need to specify a valid semver environment variable 'VERSION' to run the build process (received: '${process.env.VERSION}').`
```
francoischalifour: Perhaps this fits better:

```suggestion
    `You need to specify a valid semver environment variable 'VERSION' to run the build process (received: JSON.stringify(${process.env.VERSION})).`
```

scripts/rollup/config.js
```diff
@@ -4,6 +4,16 @@ import babel from 'rollup-plugin-babel';
...
 import { uglify } from 'rollup-plugin-uglify';
 import filesize from 'rollup-plugin-filesize';
+import semver from 'semver';
+
+if (
+  process.env.NODE_ENV === 'production' &&
+  !semver.valid(process.env.VERSION)
+) {
+  throw new Error(
+    `You need to specify a valid semver environment variable 'VERSION' to run the build process (received: JSON.stringify(${process.env.VERSION})).`
```
francoischalifour: My bad, made a mistake here!

```suggestion
    `You need to specify a valid semver environment variable 'VERSION' to run the build process (received: ${JSON.stringify(process.env.VERSION)}).`
```

eunjae-lee: Then my bad too 😅

PR #4350, opened 2020-03-12
chore(ts-migration): hits per page rename files

nunomaduro:
This pull request renames the files that migrate **hits per page** to TypeScript.


nunomaduro:
@francoischalifour Nice catch, let me catch.

PR #4351, opened 2020-03-12
chore(ts-migration): hits per page connector and widget

nunomaduro:
This pull request migrates both the connector and the widget of **hits per page**.


src/widgets/hits-per-page/hits-per-page.tsx
```diff
@@ -38,56 +43,57 @@ const renderer = ({ containerNode, cssClasses }) => (
...
+
+  /**
+   * CSS classes added to each `<option>`.
+   */
+  option?: string | string[];
+};
+
+export type HitsPerPageItems = HitsPerPageConnectorParams['items'];
+
+export type HitsPerPageWidgetOptions = {
```
francoischalifour: Wdyt about reusing (i.e. "extending") `HitsPerPageConnectorParams` here?

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -4,101 +4,120 @@ import {
...
+   */
+  render: HitsPerPageRenderer<THitsPerPageWidgetParams>,
+
+  /**
+   * Unmount function called when the widget is disposed.
+   */
+  unmount?: () => void
+) => HitsPerPageWidgetFactory<THitsPerPageWidgetParams>;
+
+const connector: HitsPerPageConnector = function connectHitsPerPage(
```
francoischalifour: We could be explicit as we are in the widget file by naming the variable `connectHitsPerPage`.

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -4,101 +4,120 @@ import {
...
+  renderFn,
+  unmountFn = noop
+) {
   checkRendering(renderFn, withUsage());
 
-  return (widgetParams = {}) => {
-    const { items: userItems, transformItems = items => items } = widgetParams;
+  return widgetParams => {
```
francoischalifour: Since we forward the `widgetParams` to the render function, defaulting the widget params to `{}` is different than destructuring the widget params with a `{}` fallback.

We should probably add a test to make this behavior consistent across connectors.

src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.ts
```diff
@@ -19,10 +30,13 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits-per-pa
     });
 
     it('throws with empty items', () => {
+      const items: HitsPerPageConnectorParamsItem[] = [];
```
francoischalifour: TS is supposed to throw once you pass the options with a wrong type. Wdyt about not using explicit types in the tests?

src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.ts
```diff
@@ -116,16 +130,19 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits-per-pa
...
-    widget.init({
-      helper,
-      state: helper.state,
-      createURL: () => '#',
-    });
+    widget.init!(
+      createInitOptions({
+        helper,
+        state: helper.state,
+        createURL: () => '#',
```
francoischalifour: You can remove the default `createURL` implems since they're handled in `createInitOptions`.

src/connectors/hits-per-page/__tests__/connectHitsPerPage-test.ts
```diff
@@ -308,27 +343,34 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits-per-pa
...
-    widget.init({
-      helper,
-      state: helper.state,
-      createURL: state => state,
-    });
+    widget.init!(
+      createInitOptions({
+        helper,
+        state: helper.state,
+        createURL: state => state as any,
```
francoischalifour: If you want to keep type safety, you can return `JSON.stringify(state)` and parse it line 360.

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -4,101 +4,120 @@ import {
...
+
+  /**
+   * Creates the URL for a single item name in the list.
+   */
+  createURL: (value: HitsPerPageRendererOptionsItem['value']) => string;
+
+  /**
+   * Sets the number of hits per page and trigger a search.
+   */
+  refine: (number: number) => any;
```
Haroenv:
```suggestion
  refine: (number: number) => void;
```

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -4,101 +4,120 @@ import {
...
+const connector: HitsPerPageConnector = function connectHitsPerPage(
+  renderFn,
+  unmountFn = noop
+) {
   checkRendering(renderFn, withUsage());
 
-  return (widgetParams = {}) => {
-    const { items: userItems, transformItems = items => items } = widgetParams;
+  return widgetParams => {
+    const { items: userItems, transformItems = items => items } =
```
Haroenv: I don't understand why we do the aliasing here? I don't see the usage, so it could just as well be `items` all the way, no?

src/connectors/hits-per-page/connectHitsPerPage.ts
nunomaduro: @Haroenv Using `void`, we are enforcing a return type public API that we don't care actually. Not allowing people to use short closures that actually return something.

nunomaduro: Actually, having `void` or `any` works the same way:

```ts
type Refine = (number: Number) => void;
const myArray: Number[] = [];
const myRefine: Refine = (number) => myArray.push(number); // works
```

playground: http://www.typescriptlang.org/play/#code/C4TwDgpgBAShBmBLAdtAvFAFMgrgWwCMIAnALigDl8jiBKKNAPigDcB7RAEwG4AoAYzbIAzsCh4QAQWLEAhiHJVCJANoBdBlHV9BIsRLhJU5QynRZcyug2YTpckADowOYQAts1ErT5A

nunomaduro: Actually, having `void` or `any` works the same way:

```ts
type Refine = (number: Number) => void;
const myArray: Number[] = [];
const myRefine: Refine = (number) => myArray.push(number); // works
```

playground: http://www.typescriptlang.org/play/#code/C4TwDgpgBAShBmBLAdtAvFAFMgrgWwCMIAnALigDl8jiBKKNAPigDcB7RAEwG4AoAYzbIAzsCh4QAQWLEAhiHJVCJANoBdBlHV9BIsRLhJU5QynRZcyug2YTpckADowOYQAts1ErT5A

francoischalifour: Ah nice!

francoischalifour: I though we had an ESLint rule that forbids the use of argument name similar to a type name. I'd be in favor of removing this confusion.

```suggestion
  refine: (value: number) => any;
```

src/widgets/hits-per-page/hits-per-page.tsx
nunomaduro: Good idea!

src/connectors/hits-per-page/connectHitsPerPage.ts
Haroenv: true, but refine is our function no?

src/connectors/hits-per-page/connectHitsPerPage.ts
nunomaduro: It's because items is being used on the line `transformItems = items => items`. And narrowing a type is bad practice.

nunomaduro:
@francoischalifour @Haroenv all comments addressed, and stories are now migrated too.

PR #4357, opened 2020-03-18
chore: bump dependencies

nunomaduro:
This pull request bumps all dependencies to their latest **minor** versions. Let me know what do you think, and if is a good idea.

I do this kind of work, on the api client javascript, once a week.


.eslintrc.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   extends: ['algolia', 'algolia/jest', 'algolia/react', 'algolia/typescript'],
   plugins: ['react-hooks'],
   rules: {
+    'import/extensions': 0,
```
francoischalifour: Why is this needed?

package.json
```diff
@@ -59,41 +59,41 @@
     "classnames": "^2.2.5",
     "events": "^1.1.0",
     "hogan.js": "^3.0.2",
-    "preact": "^10.0.0",
+    "preact": "^10.3.4",
```
francoischalifour: We shouldn't bump the `dependencies` because they restrict the available range for our users unnecessarily.

nunomaduro: Ups, nice catch. Indeed!

src/lib/InstantSearch.ts
```diff
@@ -111,7 +111,7 @@ export type InstantSearchOptions<TRouteState = UiState> = {
  * created using the `instantsearch` factory function.
  * It emits the 'render' event every time a search is done
  */
-class InstantSearch extends EventEmitter {
+class InstantSearch extends EventEmitter.EventEmitter {
```
Haroenv: what made this change necessary? It doesn't seem like you've updated that dependency

package.json
```diff
@@ -140,12 +140,12 @@
...
     "shipjs": "0.16.0",
     "typescript": "3.7.5",
-    "webpack": "4.41.5"
+    "webpack": "4.42.0"
   },
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "63.5 kB"
+      "maxSize": "63.97 kB"
```
Haroenv: do you know what caused this regression? It isn't big, but still surprising

src/lib/InstantSearch.ts
francoischalifour: Since the lock file was regenerated, the internal `events` dependency is now `3.1.0`:

https://github.com/algolia/instantsearch.js/blob/3b831507df3a5ba70b1db3952c45701551ceab86/yarn.lock#L6064-L6067

We probably want to revert this change.

package.json
francoischalifour: Some `dependencies` were updated and the lock file was regenerated. Since the UMD relies on these locked `dependencies`, their latest versions are a bit bigger. We should be careful with regenerating the lock file. I think we should revert that.

src/lib/InstantSearch.ts
Haroenv: hmm, IIRC we want events@2, not events@3, there's a quite big difference in size, and the helper is using @2

nunomaduro:
Delaying this for later.

PR #4359, opened 2020-03-18
fix(index): support adding index widget with initial UI state

Haroenv:
Sorry for this larger pull request, if it proves too complicated for a single PR, we can split this up in three smaller PRs

1. when hydrating (server or client) we add widgets, these shouldn't search yet, until started

since every added widget calls scheduleSearch if `init` has happened, this used to do a search, even before `start` was called, and thus with not yet the correct state

2. make sure local ui state _keeps_ "not yet mounted" widgets on change

e.g. initialUiState contains a few widgets, if you mount those one by one, after `init` on its index, each widget would cause a change event on the helper, which used to erase non-valid state, since it starts from an empty ui state. This is now fixed by using the previous local ui state as base, but a special provision for `isPageReset`

3. prevent mutation in escapeHits

It's okay to do `results.hits = ...` inside connectHits, since that overrides just the value on `SearchResults`, however if we mutate within the inner hits, we are actually mutating `_rawResults`, which isn't what we want to achieve for server side rendering, since `_rawResults` gets serialised.

Further changes are just minor for tests

## How to test this?

Two options:

1. the tests themselves show a use case of uiState which shouldn't be removed
2. https://github.com/algolia/vue-instantsearch/pull/764 has instructions


src/widgets/index/index.ts
```diff
@@ -403,6 +406,10 @@ const index = (props: IndexProps): Index => {
       // configuration of the widget is pushed in the URL. That's what we want to avoid.
       // https://github.com/algolia/instantsearch.js/pull/994/commits/4a672ae3fd78809e213de0368549ef12e9dc9454
       helper.on('change', ({ state }) => {
+        if (!instantSearchInstance.started) {
+          return;
+        }
```
Haroenv, line 411: adding these lines will make a bunch of test fail related specifically to this behaviour before start. The tests in index seem to make sense to need .started, but the ones in RoutingManager seem like legit bug fixes

I've considered doing something like `_isSsrServer` (passing Vue's `this.$isServer` to instance), but not sure if that makes full sense.

src/widgets/index/index.ts
```diff
@@ -403,6 +406,13 @@ const index = (props: IndexProps): Index => {
       // configuration of the widget is pushed in the URL. That's what we want to avoid.
       // https://github.com/algolia/instantsearch.js/pull/994/commits/4a672ae3fd78809e213de0368549ef12e9dc9454
       helper.on('change', ({ state }) => {
+        if (
+          instantSearchInstance.__isServerRendering &&
+          !instantSearchInstance.started
+        ) {
+          return;
+        }
```
Haroenv, line 414: I'm not too sure how to test this, since localUiState isn't exposed outside

src/widgets/index/index.ts
```diff
@@ -357,6 +361,7 @@ const index = (props: IndexProps): Index => {
       // listener on `change` below, once `init` is done.
       helper.on('change', ({ isPageReset }) => {
         if (isPageReset) {
+          localUiState.page = undefined;
```
samouss, line 364: I would have put this condition where we mutate the `localUiState` that way it's simpler to reason about.

Haroenv:
E2E test failure seems to be a fluke (cc @yannickcr), locally the URL updates as expected in those scenarios

src/lib/escape-highlight.ts
```diff
@@ -46,13 +46,15 @@ function recursiveEscape(input: any): any {
 
 export default function escapeHits(hits: Hit[]): Hit[] {
   if ((hits as any).__escaped === undefined) {
-    hits = hits.map(hit => {
-      if (hit._highlightResult) {
-        hit._highlightResult = recursiveEscape(hit._highlightResult);
+    // We don't overidde the value on hit because it will mutate the raw results
```
nunomaduro, line 49:
```suggestion
    // We don't override the value on hit because it will mutate the raw results
```

src/lib/escape-highlight.ts
```diff
@@ -46,13 +46,15 @@ function recursiveEscape(input: any): any {
...
+    // instead we make a shallow copy and we assigin the escaped values on it.
```
nunomaduro, line 50:
```suggestion
    // instead we make a shallow copy and we assign the escaped values on it.
```

Haroenv:
The e2e test seems to actually catch a bug; if a refinement is set again to the default value of the widget (usually nothing) the previous value for uiState is kept by the widget. This is expected on first render, but not later.

Haroenv:
The problem is related to this line

https://github.com/algolia/instantsearch.js/blob/9ff71e2341acc5b09133e095c3285a2de162b846/src/widgets/index/index.ts#L418-L425

Before this pull request, every change to the helper, will erase the uiState which is "still there", since it was reduced from an empty object. This was causing issues in Vue InstantSearch, since every widget added on server side erased the initialUiState (except the first widget to be mounted).

The solution I saw was to base the new state of the existing state. The issue that pops up because of this is that when a widget refines, and later goes back to its default / unrefined state, all widgets have a guard that check "if not refined, return uiState", and thus they _keep_ the previously refined state in the URL & uiState.

A change I can imagine is in all widgets themselves removing that condition and returning a uiState with explicitly `undefined` for those values, but that likely would break intialUiState & routing.

src/widgets/index/index.ts
```diff
@@ -402,11 +410,19 @@ const index = (props: IndexProps): Index => {
...
       // configuration of the widget is pushed in the URL. That's what we want to avoid.
       // https://github.com/algolia/instantsearch.js/pull/994/commits/4a672ae3fd78809e213de0368549ef12e9dc9454
-      helper.on('change', ({ state }) => {
-        localUiState = getLocalWidgetsState(localWidgets, {
-          searchParameters: state,
-          helper: helper!,
-        });
+      helper.on('change', ({ state, isPageReset }) => {
+        if (isPageReset) {
+          localUiState.page = undefined;
```
francoischalifour, line 415: Could this be passed in the `getLocalWidgetsState` 3rd argument call below? That would be one less mutation and code branch to think about.

src/widgets/index/index.ts
```diff
@@ -402,11 +410,19 @@ const index = (props: IndexProps): Index => {
...
+        }
+
+        localUiState = getLocalWidgetsState(
+          localWidgets,
+          {
+            searchParameters: state,
+            helper: helper!,
+          },
+          localUiState
```
francoischalifour, line 424: If I understand https://github.com/algolia/instantsearch.js/pull/4359#issuecomment-623964835 well, you want to pass `localUiState` only at first render, and `{}` next times. Is that right? If not, can you explain when we need to use the current UI state and when we don't want to use it?

src/widgets/index/index.ts
Haroenv: how would you move it to the lower argument? I could only see this as an alternative, but I like it less:


```
const { page, ...stateWithoutPage } = localUiState;

// ...
  isPageReset ? stateWithoutPage : localUiState
```

Haroenv: or else thought of putting it in here: https://github.com/algolia/instantsearch.js/blob/9ff71e2341acc5b09133e095c3285a2de162b846/src/widgets/index/index.ts#L364-L368

src/widgets/index/index.ts
```diff
@@ -111,8 +140,7 @@ function resetPageFromWidgets(widgets: Widget[]): void {
   indexWidgets.forEach(widget => {
     const widgetHelper = widget.getHelper()!;
 
-    // @ts-ignore remove once resetPage is a helper method
-    widgetHelper._change({
+    privateHelperSetState(widgetHelper, {
```
eunjae-lee, line 143: If I understand correctly, this doesn't have to be `privateHelperSetState` since we're not providing `_isFromAddWidget` here.

What about just reverting it back to `widgetHelper._change({ ... })`?
If we do that, then we use `privateHelperSetState` only in one place and we could rename it to something like `setStateFromAddWidget` to be more explicit.

What do you think?

Haroenv: yes, but that would require us to use _change still, which is a private, so I'd rather use that only in a single function, if that makes sense :)

eunjae-lee: I didn't know _change was used only here. Then it makes sense to me.

src/widgets/index/index.ts
```diff
@@ -402,11 +439,25 @@ const index = (props: IndexProps): Index => {
...
+          localUiState.page = undefined;
+        }
+
+        localUiState = getLocalWidgetsState(
+          localWidgets,
+          {
+            searchParameters: state,
+            helper: helper!,
+          },
+          _isFromAddWidget ? localUiState : {}
```
Haroenv, line 459:
```suggestion
          // @MAJOR in a next version we can always use localUiState, instead of usually empty object), but it requires every single widget to set an empty value (or remove the state key's value) if they are not refined.
          _isFromAddWidget ? localUiState : {}
```

src/widgets/index/index.ts
```diff
@@ -68,9 +69,38 @@ export function isIndexWidget(widget: Widget): widget is Index {
...
+    state,
+    isPageReset,
+    _isFromAddWidget,
+  }: {
+    state: SearchParameters;
+    isPageReset?: boolean;
+    _isFromAddWidget?: boolean;
+  }
+) {
+  if (state !== helper.state) {
```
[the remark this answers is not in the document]
Haroenv: this is code in the helper: https://github.com/algolia/algoliasearch-helper-js/blob/develop/src/algoliasearch.helper.js#L1315

src/widgets/index/index.ts
```diff
@@ -213,19 +246,23 @@ const index = (props: IndexProps): Index => {
...
         widgets.forEach(widget => {
           if (localInstantSearchInstance && widget.init) {
             widget.init({
               helper: helper!,
               parent: this,
-              uiState: {},
+              // Only index widget uses this key. This means that the initial value
+              // is sufficient, since that index didn't yet exist. If we see wrong
+              // behavior with this, use mainIndex.getWidgetState()
+              uiState: localInstantSearchInstance._initialUiState,
```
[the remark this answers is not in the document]
Haroenv: I've added a test that highlights how global ui state before this change gets "lost". Without this change nested indices don't receive a UiState to start from

src/widgets/index/index.ts
```diff
@@ -68,9 +69,38 @@ export function isIndexWidget(widget: Widget): widget is Index {
...
+/**
+ * This is the same content as helper._change / setState, but allowing for extra
+ * _isFromAddWidget, which we use to decide whether local ui state invalidates.
+ */
+function privateHelperSetState(
+  helper: AlgoliaSearchHelper,
+  {
+    state,
+    isPageReset,
+    _isFromAddWidget,
```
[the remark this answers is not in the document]
Haroenv: Good suggestion, I've changed it to `_uiState`, where we pass the local ui state in this case

src/widgets/index/index.ts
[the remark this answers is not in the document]
francoischalifour: Got it! Can you add [this link](https://github.com/algolia/algoliasearch-helper-js/blob/6b835ffd07742f2d6b314022cce6848f5cfecd4a/src/algoliasearch.helper.js#L1311-L1324) to the above comment then?

src/widgets/index/index.ts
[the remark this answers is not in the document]
francoischalifour: I got confused because of the comment—I think it's fair to remove it.

Although the name is `uiState`, it has always only acted as an initial UI state from the beginning since it's used to initialize the index UI state.

src/widgets/index/index.ts
```diff
@@ -402,11 +439,25 @@ const index = (props: IndexProps): Index => {
...
+          localUiState.page = undefined;
+        }
+
+        localUiState = getLocalWidgetsState(
+          localWidgets,
+          {
+            searchParameters: state,
+            helper: helper!,
+          },
+          // @MAJOR in a next version we can always use localUiState, instead of usually empty object), but it requires every single widget to set an empty value (or remove the state key's value) if they are not refined.
```
francoischalifour, line 458: Those seemed to be your thoughts when you were writing code but they're confusing from the outside 😅

Wdyt about removing this?

src/widgets/index/index.ts
```diff
@@ -68,9 +69,38 @@ export function isIndexWidget(widget: Widget): widget is Index {
   return widget.$$type === 'ais.index';
 }
 
+/**
+ * This is the same content as helper._change / setState, but allowing for extra
+ * UiState to be synchronized.
+ */
```
Haroenv, line 75:
```suggestion
 * see: https://github.com/algolia/algoliasearch-helper-js/blob/6b835ffd07742f2d6b314022cce6848f5cfecd4a/src/algoliasearch.helper.js#L1311-L1324
 */
```

src/widgets/index/index.ts
```diff
@@ -213,19 +246,23 @@ const index = (props: IndexProps): Index => {
...
 
         widgets.forEach(widget => {
           if (localInstantSearchInstance && widget.init) {
             widget.init({
               helper: helper!,
               parent: this,
-              uiState: {},
+              // Only index widget uses this key. This means that the initial value
+              // is sufficient, since that index didn't yet exist. If we see wrong
+              // behavior with this, use mainIndex.getWidgetState()
```
Haroenv, line 264:
```suggestion
```

src/widgets/index/index.ts
Haroenv:
```suggestion
```

PR #4360, opened 2020-03-19
docs(rangeSlider): update story

francoischalifour:
This updates the [`rangeSlider`](https://www.algolia.com/doc/api-reference/widgets/range-slider/js/) stories to make them more representative of what the widget offers:

- remove non-existant `templates.header` option
- remove custom `tooltips` option by default
- update the `step` value to be more usable and remove `pips` for the story so that it's less confusing (the step is not aligned with the pip)

### DIFF
--- a/stories/range-slider.stories.js
+++ b/stories/range-slider.stories.js
@@ -9,14 +9,6 @@ storiesOf('Refinements|RangeSlider', module)
         instantsearch.widgets.rangeSlider({
           container,
           attribute: 'price',
-          templates: {
-            header: 'Price',
-          },
-          tooltips: {
-            format(rawValue) {
-              return `$${Math.round(rawValue).toLocaleString()}`;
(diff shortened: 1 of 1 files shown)


stories/range-slider.stories.js
```diff
@@ -28,12 +20,8 @@ storiesOf('Refinements|RangeSlider', module)
...
           container,
           attribute: 'price',
-          step: 500,
-          tooltips: {
-            format(rawValue) {
-              return `$${Math.round(rawValue).toLocaleString()}`;
-            },
-          },
+          step: 100,
+          pips: false,
```
Haroenv: I guess since pips is the number, but we don't know the range is why they can't line up with step?

francoischalifour: Yes, I explained in the post.

PR #4361, opened 2020-03-19
chore(ts-migration): refactors connector types

nunomaduro:
This pull request reduces a little bit the number of types needed while migrating a connector to typescript. In addition, it also makes a the minimum amount of work to have widgets in this new type system.

Screenshot 2020-03-23 at 14 53 53

Note that some new lines are comments to uncommented types before.

### DIFF
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -6,7 +6,7 @@ import {
   noop,
   warning,
 } from '../../lib/utils';
-import { Hits, InstantSearch, Connector } from '../../types';
+import { Hits, Connector } from '../../types';
 
 const withUsage = createDocumentationMessageGenerator({
   name: 'autocomplete',
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -6,27 +6,14 @@ import {
   noop,
   warning,
 } from '../../lib/utils';
-import {
-  RendererOptions,
-  Renderer,
-  WidgetFactory,
-  Hits,
-  InstantSearch,
-  Unmounter,
-} from '../../types';
+import { Hits, InstantSearch, Connector } from '../../types';
 
 const withUsage = createDocumentationMessageGenerator({
   name: 'autocomplete',
   connector: true,
 });
 
-type AutocompleteIndex = {
-  indexName: string;
-  hits: Hits;
-  results: SearchResults;
-};
-
-type AutocompleteConnectorParams = {
+export type AutocompleteConnectorParams = {
   /**
    * Escapes HTML entities from hits string values.
    *
@@ -35,32 +22,46 @@ type AutocompleteConnectorParams = {
   escapeHTML?: boolean;
 };
 
-export type AutocompleteRendererOptions<TAutocompleteWidgetParams> = {
+export type AutocompleteRendererOptions = {
+  /**
+   * The current value of the query.
+   */
   currentRefinement: string;
-  indices: AutocompleteIndex[];
+
+  /**
+   * The indices this widget has access to.
+   */
+  indices: Array<{
+    /**
+     * The name of the index
+     */
+    indexName: string;
+
+    /**
+     * The resolved hits from the index matching the query.
+     */
+    hits: Hits;
+
+    /**
+     * The full results object from the Algolia API.
+     */
+    results: SearchResults;
+  }>;
+
+  /**
+   * The current instant search instance.
+   *
+   * @internal
+   */
   instantSearchInstance: InstantSearch;
+
+  /**
+   * Searches into the indices with the provided query.
+   */
   refine: (query: string) => void;
-} & RendererOptions<TAutocompleteWidgetParams>;
+};
 
-export type AutocompleteRenderer<TAutocompleteWidgetParams> = Renderer<
-  AutocompleteRendererOptions<
-    AutocompleteConnectorParams & TAutocompleteWidgetParams
-  >
->;
-
-export type AutocompleteWidgetFactory<
-  TAutocompleteWidgetParams
-> = WidgetFactory<AutocompleteConnectorParams & TAutocompleteWidgetParams>;
-
-export type AutocompleteConnector = <TAutocompleteWidgetParams>(
-  render: AutocompleteRenderer<TAutocompleteWidgetParams>,
-  unmount?: Unmounter
-) => AutocompleteWidgetFactory<TAutocompleteWidgetParams>;
-
-const connectAutocomplete: AutocompleteConnector = (
-  renderFn,
-  unmountFn = noop
-) => {
+export default (function connectAutocomplete(renderFn, unmountFn = noop) {
   checkRendering(renderFn, withUsage());
 
   return widgetParams => {
@@ -194,6 +195,4 @@ search.addWidgets([
       },
     };
   };
-};
-
-export default connectAutocomplete;
+} as Connector<AutocompleteRendererOptions, AutocompleteConnectorParams>);
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -54,7 +54,15 @@ export type AutocompleteRendererOptions = {
   refine: (query: string) => void;
 };
 
-export default (function connectAutocomplete(renderFn, unmountFn = noop) {
+export type AutocompleteConnector = Connector<
+  AutocompleteRendererOptions,
+  AutocompleteConnectorParams
+>;
+
+const connectAutocomplete: AutocompleteConnector = function connectAutocomplete(
+  renderFn,
+  unmountFn = noop
+) {
   checkRendering(renderFn, withUsage());
 
   return widgetParams => {
@@ -188,4 +196,6 @@ search.addWidgets([
       },
     };
   };
-} as Connector<AutocompleteRendererOptions, AutocompleteConnectorParams>);
+};
+
+export default connectAutocomplete;
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -49,13 +49,6 @@ export type AutocompleteRendererOptions = {
   }>;
 
   /**
-   * The current instant search instance.
-   *
-   * @internal
-   */
-  instantSearchInstance: InstantSearch;
-
-  /**
    * Searches into the indices with the provided query.
    */
   refine: (query: string) => void;
--- a/src/connectors/configure-related-items/connectConfigureRelatedItems.ts
+++ b/src/connectors/configure-related-items/connectConfigureRelatedItems.ts
@@ -54,7 +54,16 @@ function createOptionalFilter({
   return `${attributeName}:${attributeValue}<score=${attributeScore || 1}>`;
 }
 
-export default (function connectConfigureRelatedItems(renderFn, unmountFn) {
+type ConfigureRelatedItemsConnector = Connector<
+  ConfigureRendererOptions,
+  ConfigureRelatedItemsConnectorParams
+>;
+
+export default (function connectConfigureRelatedItems<
+  TWidgetParams extends {
+    searchParameters: algoliasearchHelper.PlainSearchParameters;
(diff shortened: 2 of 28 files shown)


src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -35,32 +22,46 @@ type AutocompleteConnectorParams = {
...
+     */
+    results: SearchResults;
+  }>;
+
+  /**
+   * The current instant search instance.
+   *
+   * @internal
+   */
   instantSearchInstance: InstantSearch;
```
francoischalifour: This should already be handled by the `RendererOptions` type: https://github.com/algolia/instantsearch.js/blob/ad1b5c6e912bf41026fae55fb451c2e4cb28f36a/src/types/connector.ts#L20

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -66,10 +54,7 @@ function createOptionalFilter({
   return `${attributeName}:${attributeValue}<score=${attributeScore || 1}>`;
 }
 
-const connectConfigureRelatedItems: ConfigureRelatedItemsConnector = (
-  renderFn,
-  unmountFn
-) => {
+export default (function connectConfigureRelatedItems(renderFn, unmountFn) {
```
francoischalifour: Note for other reviewers: the render function is now required for consistency. If it causes trouble, we can create an alternative connector type that makes this render function optional.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -148,11 +133,8 @@ See https://www.algolia.com/doc/api-reference/api-parameters/optionalFilters/
     const makeConfigure = connectConfigure(renderFn, unmountFn);
 
     return {
-      ...makeConfigure({ searchParameters }),
-
+      ...makeConfigure({ searchParameters } as any),
```
francoischalifour: Do you mind explaining why casting as `any` is required here? There should be a way to avoid this.

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -63,25 +41,23 @@ export type HitsPerPageConnectorParams = {
...
 
-export type HitsPerPageRendererOptions<
-  THitsPerPageWidgetParams extends HitsPerPageConnectorParams = HitsPerPageConnectorParams
-> = {
+export type HitsPerPageRendererOptions = {
   /**
    * Array of objects defining the different values and labels.
    */
-  items: HitsPerPageRendererOptionsItem[];
+  items: Array<HitsPerPageConnectorParamsItem & { isRefined: boolean }>;
```
francoischalifour: The connector and renderer items are not the same. It's a coincidence that they've a similar shape. The ones passed to the renderer don't have the `default` property. Here we should avoid reusing the connector type and create another `HitsPerPageRendererParamsItem` type.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
nunomaduro: This is not a easy one. I may need to explain you an zoom, and write here the feedback later.

Haroenv: I'm also surprised here, I assume that a configure widget should accept searchParams

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -194,6 +188,4 @@ search.addWidgets([
       },
     };
   };
-};
-
-export default connectAutocomplete;
+} as Connector<AutocompleteRendererOptions, AutocompleteConnectorParams>);
```
Haroenv: could we make this instead:

```
const connectAutocomplete: Connector<AutocompleteRendererOptions, AutocompleteConnectorParams> = function connectAutocomplete() {};
export default connectAutocomplete;
```

(or even `export const` so we don't need to rename default back to the variable later)

thanks for suggesting this @yannickcr

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
nunomaduro: fixed.

src/connectors/autocomplete/connectAutocomplete.ts
francoischalifour: I also find the variable syntax less awkward than the function "cast".

In favor of `export const` as well, if you believe this is the right PR to do so. More generally we should aim at reducing the amount of default exports in the codebase.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -24,37 +18,19 @@ export type InfiniteHitsConnectorParams = Partial<
...
 const withUsage = createDocumentationMessageGenerator({
   name: 'infinite-hits',
   connector: true,
 });
 
-const connectInfiniteHits: InfiniteHitsConnector = (
-  renderFn,
-  unmountFn = noop
-) => {
+export default (function connectInfiniteHits(renderFn, unmountFn = noop) {
```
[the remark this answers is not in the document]
francoischalifour: It'd make sense to type it at the Index level, especially when you manipulate scoped results (i.e. children and siblings).

src/types/connector.ts
```diff
@@ -1,22 +1,73 @@
...
+  /**
+   * The original search results.
+   */
   results?: SearchResults;
+
+  /**
+   * The mutable list of hits. The may change depending
+   * of the given transform items function.
+   */
   hits?: Hits;
```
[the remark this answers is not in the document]
francoischalifour: These are some misused typings. They shouldn't be here after the migration.

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -84,42 +79,18 @@ type CurrentRefinementsConnectorParams = {
   transformItems?: (items: Item[]) => any;
 };
 
-export type CurrentRefinementsRendererOptions<
-  TCurrentRefinementsWidgetParams
-> = {
+export type CurrentRefinementsRendererOptions = {
   items: Item[];
   refine(refinement: ItemRefinement): void;
```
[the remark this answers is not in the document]
nunomaduro: Noted for a future pull request.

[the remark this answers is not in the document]
francoischalifour: @Haroenv This vocabulary is confusing but `ItemRefinement` makes more sense in this context because an item contains refinements:

```
item
 - refinement
 - refinement
```

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -24,37 +18,19 @@ export type InfiniteHitsConnectorParams = Partial<
   InfiniteHitsRendererWidgetParams
```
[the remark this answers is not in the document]
nunomaduro: Noted for eventually be removed later.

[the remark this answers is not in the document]
nunomaduro: Noted to, if needed, removed lated.

src/types/connector.ts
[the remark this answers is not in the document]
nunomaduro: Noted to be refactored later.

src/types/widget.ts
```diff
@@ -192,8 +192,14 @@ export type Widget = {
...
-export type WidgetFactory<TWidgetParams> = (
-  widgetParams: TWidgetParams
+/**
+ * The function that creates a new widget.
+ */
+export type WidgetFactory<TConnectorParams, TWidgetParams> = (
+  /**
+   * The params of the widget.
+   */
+  widgetParams: TConnectorParams & TWidgetParams
```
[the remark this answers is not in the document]
nunomaduro: Indeed, in that case I prefer we actually pass explicitly `{}` so the reader knows that there is no connector params there.

src/types/connector.ts
```diff
@@ -1,22 +1,73 @@
...
+
+  /**
+   * The current instant search instance.
+   */
   instantSearchInstance: InstantSearch;
+
+  /**
+   * The original search results.
+   */
   results?: SearchResults;
```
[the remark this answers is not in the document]
nunomaduro: Not sure, noted to be refactored later.

src/connectors/autocomplete/connectAutocomplete.ts
nunomaduro: Personally, I found the way I did more clean ( less typing too ).

Haroenv: I always thought that `as` was a bad practice in typescript, but I might have made it up, since I don't see any reason why it's different

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -130,11 +139,11 @@ See https://www.algolia.com/doc/api-reference/api-parameters/optionalFilters/
       ),
     };
 
-    const makeConfigure = connectConfigure(renderFn, unmountFn);
+    const makeConfigure = connectConfigure<TWidgetParams>(renderFn, unmountFn);
 
     return {
-      ...makeConfigure({ searchParameters } as any),
+      ...makeConfigure({ searchParameters } as TWidgetParams),
```
Haroenv:
```suggestion
      ...makeConfigure({ searchParameters }),
```

What TS error do you get if you do it just like this?

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -54,7 +54,16 @@ function createOptionalFilter({
...
 
-export default (function connectConfigureRelatedItems(renderFn, unmountFn) {
+type ConfigureRelatedItemsConnector = Connector<
+  ConfigureRendererOptions,
+  ConfigureRelatedItemsConnectorParams
+>;
+
+export default (function connectConfigureRelatedItems<
+  TWidgetParams extends {
+    searchParameters: algoliasearchHelper.PlainSearchParameters;
```
Haroenv: I'm confused, I think this makes searchParameters a widget param of configure related items, no?

nunomaduro: @Haroenv I will only find time to work on this pull request next Friday, I will back to you then.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
nunomaduro: I am going to alias to any here. I propose we re-visit this topic once we analyse if searchParameters should be a type of the connector or widget or wtv.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
nunomaduro: I am going to alias to any here. I propose we re-visit this topic once we analyse if searchParameters should be a type of the connector or widget or wtv.

src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
```diff
@@ -32,10 +32,11 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/current-ref
       const render = jest.fn();
       const unmount = jest.fn();
 
-      const customCurrentRefinements = connectCurrentRefinements(
+      const customCurrentRefinements = connectCurrentRefinements<{}>(
```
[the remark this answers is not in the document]
nunomaduro: yeah, prefer being explicit rather than implicit.

src/connectors/configure-related-items/connectConfigureRelatedItems.ts
```diff
@@ -148,8 +141,7 @@ See https://www.algolia.com/doc/api-reference/api-parameters/optionalFilters/
     const makeConfigure = connectConfigure(renderFn, unmountFn);
 
     return {
-      ...makeConfigure({ searchParameters }),
-
+      ...makeConfigure({ searchParameters } as any),
```
Haroenv:
```suggestion
      // required, since widget parameters differ between these connectors
      // and we don't want to have the parameters of configure here
      ...makeConfigure({ searchParameters } as any),
```

PR #4362, opened 2020-03-19
feat: introduce controlled mode

francoischalifour:
_This PR introduces a controlled mode in InstantSearch.js._

## Summary

Introducing a controlled mode in InstantSearch allows more powerful state updates coming from internal events and one-time external events.

> [👉 Try controlled mode in the playground](https://codesandbox.io/s/instantsearchjs-controlled-mode-o6lz0)!

## Definition

"Controlled mode" is a generic term that, in InstantSearch, is defined as:

- (1) An external event can update the internal state
- (2) An internal state change is delegated to a public function

## Usage

There are two API additions to allow (1) and (2):

1. `setUiState`: available on the search instance
2. `onStateChange`: available as an `instantsearch` option

```ts
  onStateChange({ uiState, setUiState }) {
    logger.info(uiState);

    setUiState(uiState);
  },
});

button.addEventListener('click', () => {
  search.setUiState({
    instant_search: {
      query: 'phone',
    },
  });
});
```

### `setUiState`

> `(uiState: UiState | ((uiState: UiState) => UiState)) ⇒ void`

#### Description

This function allows to **inject a UI state** in the search experience, **without relying on an internal event** (e.g., connector's `refine` or widget interaction). This can be used to set the internal state from an external event.

Note that for the UI state to be effective, the widgets responsible for each UI state attribute need to be mounted (same behavior as `initialUiState` and `routing`).

#### Usage

```ts
  /* ... */
});

// This can happen after an interaction outside of the InstantSearch scope.
search.setUiState({
  instant_search: {
    query: 'phone',
    refinementList: {
      brand: ['Samsung'],
    },
  },
});

// You can compute a UI state based on a previous one with the function form.
search.setUiState(prevUiState => ({
  ...prevUiState,
  instant_search: {
    ...prevUiState.instant_search,
    refinementList: {
      brand: ['Samsung'],
    },
  },
}));
```

### `onStateChange`

> `(params: { uiState: UiState, setUiState(uiState: UiState): void }) ⇒ void`

#### Description

This function is a hook passed to the `instantsearch` option. If it's given to the options, **the experience becomes fully controlled**. This can be used to hook some custom logic whenever the internal state changes (e.g., a logger, avoiding a specific state, etc.).

If you use it, you become responsible for setting the UI state with `setUiState`.

#### Usage

```ts
  onStateChange({ uiState, setUiState }) {
    logger.info(uiState);

    // Sets the internal state as the uncontrolled behavior would do.
    setUiState(uiState);
  },
});
```

## Implementation

- The internal `onStateChange` function of the experimental middleware param `state` was renamed `uiState` for consistency.
- The internal `onStateChange` was renamed `onInternalStateChange` to avoid confusion
- The development warning for inconsistent UI state was moved to its own function because it's used in multiple places now (I'll send a PR later to move the warning at the InstantSearch instance level, rather than at the index level).

## References

- [RFC](https://github.com/algolia/instantsearch-rfcs/pull/33)
- [Playground](https://codesandbox.io/s/instantsearchjs-controlled-mode-o6lz0)


Haroenv:
A "forceUpdate" would look like this right: `setUiState(state => state)`. I don't think we need a different paradigm for it, but just wanted to mention it. Since all helper methods send a change event I don't think there's a use case though.

src/lib/InstantSearch.ts
```diff
@@ -80,6 +81,17 @@ export type InstantSearchOptions<TRouteState = UiState> = {
    */
   searchFunction?: (helper: AlgoliaSearchHelper) => void;
 
+  /**
+   * Function called when the state changes.
+   *
+   * Using this function turns the experience controlled. This means that you
```
```suggestion
   * Using this function makes the instance controlled. This means that you
```

src/lib/InstantSearch.ts
```diff
@@ -80,6 +81,17 @@ export type InstantSearchOptions<TRouteState = UiState> = {
...
+   * become in charge of updating the UI state with the `setUiState` function.
```
Haroenv: What happens if you call the function async. Is that something you tried?

Haroenv: https://codesandbox.io/s/instantsearchjs-controlled-mode-so5qw, looks like it makes a global debounce possible, but seems a bit glitchy to me 🤔

src/lib/InstantSearch.ts
```diff
@@ -494,12 +509,55 @@ See ${createDocumentationLink({
...
+    this.onInternalStateChange();
+  };
+
+  public onInternalStateChange = () => {
     const nextUiState = this.mainIndex.getWidgetState({});
 
     this.middleware.forEach(m => {
       m.onStateChange({
-        state: nextUiState,
+        uiState: nextUiState,
```
Haroenv: Not sure if that matters since middleware is not yet documented, but it's a breaking change I think

src/lib/utils/checkIndexUiState.ts
```diff
@@ -0,0 +1,181 @@
+import { warning, capitalize } from '.';
+import { Index } from '../../widgets/index/index';
+import { Widget, IndexUiState } from '../../types';
```
Haroenv: this file doesn't seem to individually be tested, but that is ok I think

src/lib/InstantSearch.ts
francoischalifour: It's not documented and under an `EXPERIMENTAL_` prefix.

src/lib/InstantSearch.ts
francoischalifour: It behaves as I would expect to: [see sandbox](https://codesandbox.io/s/instantsearchjs-controlled-mode-3c26u). What seems glitchy to you?

Haroenv: all updates happen still at the same speed as you initially did, but they are delayed. It's clearer to see with a higher value for the timeout.

I wonder if the older values maybe need to be cancelled, I'll try that

Haroenv: https://codesandbox.io/s/instantsearchjs-controlled-mode-so5qw, yes this way it's way smoother. 

```js
let timeoutHandle;

  indexName: 'instant_search',
  searchClient,
  // Try controlled mode here!
  onStateChange({ uiState, setUiState }) {
    console.group('onStateChange');
    console.log({ uiState });
    console.groupEnd();

    if (timeoutHandle) {
      clearTimeout(timeoutHandle);
    }

    timeoutHandle = setTimeout(() => setUiState(uiState), 500);
  },
});
```

If we document this in the future, we need to make sure to cancel the stale updates. I don't think InstantSearch can do that out of the box.

Haroenv:
I looked, and couldn't find an eslint rule to prefer types over interfaces, or the other way round :(

yannickcr:
@Haroenv  were you looking for something like this? https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-definitions.md

PR #4363, opened 2020-03-25
feat(voice): allow custom voice helper

Haroenv:
**Summary**

If the `createVoiceHelper` function can be overridden, other third-party voice solutions can be used, without them needing to redo all the rendering & templating.

**Result**

Allow custom `createVoiceHelper` to be passed to the widget & connector.

cc @dylanbfox

todo:
- [x] test
- [ ] allow this in React InstantSearch, Vue InstantSearch & Angular InstantSearch too


src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -21,6 +21,7 @@ export type VoiceSearchConnectorParams = {
   additionalQueryParameters?: (params: {
     query: string;
   }) => PlainSearchParameters | void;
+  createVoiceSearchHelper?: typeof builtInCreateVoiceSearchHelper;
```
francoischalifour, line 24: We should create an interface rather than relying on the existing code inference.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -65,6 +65,28 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-searc
...
+      const voiceHelper = {
+        isBrowserSupported: () => true,
+        dispose: () => {},
+        getState: () => ({
+          isSpeechFinal: true,
+          status: 'askingPermission',
+          transcript: '',
+        }),
+        isListening: () => true,
+        toggleListening: () => {},
```
francoischalifour, line 79: If we make this API public, the options should be explicit. I'd suggest remplacing `toggleListening` by `onListenStart` and `onListenEnd`.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -65,6 +65,28 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-searc
         })
       );
     });
+
+    it('creates custom voice helper', () => {
+      const voiceHelper = {
+        isBrowserSupported: () => true,
+        dispose: () => {},
+        getState: () => ({
+          isSpeechFinal: true,
```
francoischalifour, line 74: Does this state translate to other voice solutions? Can you copy/paste a usage with another solution so that we better understand how the API can work?

src/connectors/voice-search/connectVoiceSearch.ts
Haroenv: why?

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
Haroenv: toggle listening is a function to toggle the listening, not event listeners

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
Haroenv: other api didn't need isSpeechFinal since it didn't give intermediary results (only will transcribe once text is finished, and thus isSpeechFinal will be true). A fully local transcribing service would be able to use the variable though

src/connectors/voice-search/connectVoiceSearch.ts
francoischalifour: Because it's a contract that we now need to respect internally too. It makes it obvious when we break it, and will make it easier to export.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
francoischalifour: Than perhaps `start` and `stop`.

It seems like `toggleListening` is redundant with `dispose` too, since it calls `dispose` despite `dispose` being exported too.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
francoischalifour: Do you have an external API implementation usage so we better understand?

Haroenv: it's not yet public, so I sent it on slack

src/connectors/voice-search/connectVoiceSearch.ts
francoischalifour: You wouldn't write `typeof true`, but you'll explicitly write `boolean`.

Would it make sense to allow `getState` to return a custom state shape? If so, that's another limitation of this line.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
Haroenv: dispose removes listeners, so it's different. We could expose start & stop from the helper & put toggleListening inside the connector indeed

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
```diff
@@ -65,6 +68,28 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-searc
...
+          status: 'askingPermission',
+          transcript: '',
+        }),
+        isListening: () => true,
```
francoischalifour, line 81: `isListening` seems to share lots of similarities with state variables, like `isSpeechFinal`. Wdyt about having `isListening` in `getState`?

src/lib/voiceSearchHelper/index.ts
```diff
@@ -139,22 +109,14 @@ export default function createVoiceSearchHelper({
...
-      start();
-    }
-  };
-
   return {
     getState,
     isBrowserSupported,
     isListening,
-    toggleListening,
+    startListening: start,
```
francoischalifour, line 116: We can rename our internal functions `startListening` and `stopListening` to use the same vocabulary. Wdyt?

src/lib/voiceSearchHelper/types.ts
```diff
@@ -0,0 +1,34 @@
...
+  | 'waiting'
+  | 'recognizing'
+  | 'finished'
+  | 'error';
+
+export type VoiceListeningState = {
+  status: Status;
+  transcript: string;
+  isSpeechFinal: boolean;
+  errorCode?: string;
```
francoischalifour, line 13: When we expose a value that is explicitly not defined in our codebase, we usually set it to `null` ([examples](https://github.com/algolia/instantsearch.js/blob/1a6dc518763132eb3223c078cb38841bca83b602/src/widgets/index/index.ts#L47-L65)).

I think we should do the same here.

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
eunjae-lee: Our implementation for `isListening()` is 

```js
  const isListening = (): boolean =>
    state.status === 'askingPermission' ||
    state.status === 'waiting' ||
    state.status === 'recognizing';
```

For our case, we could have it as a boolean variable in `getState`.
However, in terms of using third-party helpers, `isListening` might have to be a function on their side (idk).
And `isListening, startListening, stopListening` sounds consistent to me.

src/lib/voiceSearchHelper/index.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4363/commits/9fe79ab80fd29595839a5364cd9939821ffe9d2e

src/lib/voiceSearchHelper/types.ts
Haroenv: I don't know what you're talking about, error code is defined in our implementation, but since we don't use it ourselves (only exposed to users, so breaking change to change that), it's either a string or undefined. This is not a scoped variable, but just some key in an object.

Do I misunderstand what you're suggesting?

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
Haroenv: In the custom helper I wrote the condition is

```js
  const isListening = () =>
    state.status === 'askingPermission' || state.status === 'recognizing';
```

since there's no 'waiting' phase. Other custom flows will have their own statuses as well

src/lib/voiceSearchHelper/types.ts
francoischalifour: I meant that it would have made sense to have it be a `string | null` but didn't realise it's already exposed to user, so breaking change 👍

src/connectors/voice-search/__tests__/connectVoiceSearch-test.js
francoischalifour: Right! I don't have a complete understanding of ways to use the voice search widget so I'm only suggesting ideas, they might not make total sense—feel free to ignore.

src/lib/voiceSearchHelper/types.ts
```diff
@@ -0,0 +1,34 @@
+export type Status =
+  | 'initial'
+  | 'askingPermission'
+  | 'waiting'
+  | 'recognizing'
+  | 'finished'
+  | 'error';
+
+export type VoiceListeningState = {
```
francoischalifour, line 9: To make sure I understand: are this state properties necessary for all Voice solutions, or only for our default implementation? In other words, should we accept unknown properties in the state?

Haroenv: We could accept arbitrary props here as well, but these are the props we are using in our own component, documentation & template, so at least those are required (and should always be able to be present, at least with a static value).

I don't really see how to combine arbitrary props with a set of required ones, but will try out

eunjae-lee: I don't know if there is any standard spec or something, but `VoiceListeningState` and `Status` just came from our default implementation which is based on SpeechRecognition web api.
These fit at least our implementation and this new custom helper one. We could see later if we need to accept unknown properties for other third party helpers.

src/lib/voiceSearchHelper/types.ts
```diff
@@ -0,0 +1,34 @@
...
+  status: Status;
+  transcript: string;
+  isSpeechFinal: boolean;
+  errorCode?: string;
+};
```
Haroenv, line 14:
```suggestion
  [key: string]: any;
};
```

src/lib/voiceSearchHelper/types.ts
francoischalifour: Then I would suggest making it a generic that extends the required properties: `VoiceListeningState<TState>`.

dustincoates: Recapping our discussions earlier, we decided that we don't need arbitrary states at the moment and this shouldn't block folding this into the other flavors.

Haroenv:
Thanks @dustincoates, @eunjae-lee & @francoischalifour, this is now ready for review again

dustincoates:
@francoischalifour would you mind giving this a review? :)

francoischalifour:
@haroenv Do we plan to roll this out in all flavors at once or in IS.js first?

Haroenv:
InstantSearch.js was enough for now, but will add other flavours soon I guess :)

PR #4365, opened 2020-03-27
chore(fixes-yarn-lint-hangs): adds ecommerce example to eslintignore

nunomaduro:
This pull request fixes the reason why `yarn lint` takes a year to lint the project. The reason is because it was trying to linting the ecommerce example.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,3 @@
-/dist
 /cjs
 /es
 /website
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,6 +6,7 @@ node_modules/
 
 # TODO: don't ignore these examples once they're updated
 /examples/calendar-widget
-/examples/e-commerce
+/examples/e-commerce/.cache
+/examples/e-commerce/dist
 /examples/media
 /examples/tourism
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,6 +5,7 @@
 node_modules/
 
 # TODO: don't ignore these examples once they're updated
+/examples/e-commerce
 /examples/calendar-widget
 /examples/media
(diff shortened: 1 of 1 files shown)


.eslintignore
```diff
@@ -6,5 +6,6 @@ node_modules/
 
 # TODO: don't ignore these examples once they're updated
 /examples/calendar-widget
+/examples/e-commerce
```
francoischalifour: We shouldn't ignore linting the examples, but only their built files: `.cache` and `dist`.

nunomaduro: @francoischalifour That's the comment on the line 7 right? Can we address that in other pull request?

francoischalifour: No, the end goal is to not ignore any examples. The ones listed here are the ones that have not been migrated to Parcel yet, and are not compliant with our lint configuration. `examples/e-commerce` is.

Ignoring `.cache` and `dist` (not `/dist`) should work.

nunomaduro: @francoischalifour Thanks for the feedback. Just updated the pull request.

.eslintignore
```diff
@@ -6,5 +6,7 @@ node_modules/
 
 # TODO: don't ignore these examples once they're updated
 /examples/calendar-widget
+/examples/e-commerce/.cache
```
francoischalifour: More globally, we're safe to remove the `/examples/e-commerce/` prefix for these two lines so that we're not surprised again when we migrate the other examples. We can change those two lines above:

```diff
- /dist
+ dist
+ .cache
```

nunomaduro: @francoischalifour Thanks for the feedback. Just updated the pull request.

PR #4368, opened 2020-04-01
chore(ts-migration): renders explicit the client type

nunomaduro:
This pull request makes explicit the client type used along instantsearch. Also improves a little bit the type itself, saying that the `addAlgoliaAgent` is optional.

### DIFF
--- a/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
+++ b/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
@@ -11,7 +11,7 @@ import { createSearchClient } from '../../../../test/mock/createSearchClient';
 import { createSingleSearchResponse } from '../../../../test/mock/createAPIResponse';
 import connectAutocomplete from '../connectAutocomplete';
 import { TAG_PLACEHOLDER } from '../../../lib/escape-highlight';
-import { Client } from '../../../types';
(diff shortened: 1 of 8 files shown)


src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -22,7 +22,7 @@ describe('connectAutocomplete', () => {
     });
 
     const initialConfig = {};
-    const helper = algoliasearchHelper({} as Client, '', initialConfig);
+    const helper = algoliasearchHelper({} as SearchClient, '', initialConfig);
```
francoischalifour, line 25: Instead of casting we can pass an actual client with [`createSearchClient`](https://github.com/algolia/instantsearch.js/blob/1a6dc518763132eb3223c078cb38841bca83b602/test/mock/createSearchClient.ts#L9-L21). This will be more consistent with the real usage.

nunomaduro: @francoischalifour Indeed. Do you mind if we do those changes in another pull request that actually tackles that problem?

francoischalifour: Sure, if you prefer!

PR #4369, opened 2020-04-01
chore(ts-migration): exports missing types at top level

nunomaduro:
This pull request fixes most of the errors reported by API Extractor. They are related to the fact that some types weren't exported a top level making the API difficult to use. Here is an example that **was not** possible in typescript before this pull request:

```ts
const client: SearchClient = { ... }; // my own special search client that looks like algolia one.

const options: InstantSearchOptions = {
  indexName: 'instant_search',
  searchClient: client,
};

const search: InstantSearch = instantsearch(options);
```

**Warning**: I am still validating locally if we are not exporting things that we shouldn't. So this pull request is not ready to merge. Just wanted to know your opinion about this change.

### DIFF
--- a/src/index.es.ts
+++ b/src/index.es.ts
@@ -41,4 +41,7 @@ import { connectSearchBox } from 'instantsearch.js/es/connectors'`
   },
 });
 
+export * from './types';
+export * from './middleware';
+
 export default instantsearch;
--- a/src/types/instantsearch.ts
+++ b/src/types/instantsearch.ts
@@ -10,26 +10,26 @@ export type HelperChangeEvent = {
   state: SearchParameters;
 };
 
-type HitAttributeHighlightResult = {
+export type HitAttributeHighlightResult = {
   value: string;
   matchLevel: 'none' | 'partial' | 'full';
   matchedWords: string[];
   fullyHighlighted?: boolean;
 };
 
-type HitHighlightResult = {
+export type HitHighlightResult = {
   [attribute: string]:
     | HitAttributeHighlightResult
     | HitAttributeHighlightResult[]
     | HitHighlightResult;
 };
 
-type HitAttributeSnippetResult = Pick<
+export type HitAttributeSnippetResult = Pick<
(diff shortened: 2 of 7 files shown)


src/index.es.ts
```diff
@@ -41,4 +41,7 @@ import { connectSearchBox } from 'instantsearch.js/es/connectors'`
   },
 });
 
+export * from './types';
+export * from './middleware';
```
francoischalifour, line 45: We can omit this folder for now. It's never been public.

src/types/instantsearch.ts
```diff
@@ -10,26 +10,26 @@ export type HelperChangeEvent = {
   state: SearchParameters;
 };
 
-type HitAttributeHighlightResult = {
+export type HitAttributeHighlightResult = {
```
francoischalifour, line 13: Note: we should make sure that these types are exact if we export them.

src/types/widget.ts
```diff
@@ -7,6 +7,12 @@ import {
 } from 'algoliasearch-helper';
 import { InstantSearch } from './instantsearch';
 
+export {
```
francoischalifour, line 10: These are internal.

src/types/widget.ts
```diff
@@ -7,6 +7,12 @@ import {
...
+  Index,
+  IndexInitOptions,
+  IndexRenderOptions,
+} from '../widgets/index/index';
+
 export type InitOptions = {
```
francoischalifour, line 16: The following types are also internal.

src/widgets/index/index.ts
```diff
@@ -33,12 +33,12 @@ type IndexProps = {
   indexId?: string;
 };
 
-type IndexInitOptions = Pick<
+export type IndexInitOptions = Pick<
```
francoischalifour, line 36: This and the following type are internal.

nunomaduro:
Replaced by https://github.com/algolia/instantsearch.js/pull/4374.

PR #4373, opened 2020-04-07
fix(setUiState): display experimental warning

francoischalifour:
## Description

Before releasing the new set of APIs to control the UI state, we mark them as experimental to gather feedback and change their APIs in a next minor if needed.

The warning only happens in `setUiState` because `onStateChange` relies on `setUiState`.

## Related

- API documentation: algolia/doc#4352

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -518,6 +518,15 @@ See ${createDocumentationLink({
       );
     }
 
+    warning(
+      false,
+      `
+\`setUiState\` provides a powerful way to manage the UI state. This is considered experimental as the API might change in a next minor version.
+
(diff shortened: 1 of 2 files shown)


francoischalifour:
@Haroenv We could, but since `onStateChange` is always used with `setUiState`, the warning will likely be unnecessary.

Haroenv:
Yes, but the warning would be called only once, versus now on every call of setUiState. No big difference of course, since we cache warnings

francoischalifour:
@Haroenv Yep but if we only use the warning in `onStateChange`, users using the `setUiState` API (for injecting a UI state only once) won't see the warning.

francoischalifour:
Merging since doc PR was merged.

PR #4374, opened 2020-04-07
chore(ts-migration): adds api extrator for public definitions files and strip Internals

nunomaduro:
This pull request **now** uses api extractor for validating the public files we are exporting instead of a single file.

Here is an example:

Screenshot 2020-04-08 at 14 47 41

In addition, it adds **stripInternals** option the tsconfig file, so we can hide some internals from the public API.


Haroenv:
Did you think the warnings from api-extractor weren't useful after all?

nunomaduro:
@Haroenv They are useful - but the goal of having API Extrator is detect missing exported types on a a **single file public api**. With the current approach, and to move faster, we are going to have around 120 files.

nunomaduro:
@Haroenv Maybe in the future, and with more time, once we have root level exports and can have again api extrator 👍

francoischalifour:
Will this configuration be enough to catch error types like when `InstantSearchOptions` is not exported? If not, how important do you think it is?

nunomaduro:
> Will this configuration be enough to catch error types like when InstantSearchOptions is not exported? If not, how important do you think it is?

You just gave me a wonderful idea. Give me a second.

package.json
```diff
@@ -23,9 +23,9 @@
     "start": "yarn run dev",
     "dev": "yarn run storybook",
     "build": "yarn run build:cjs && yarn run build:es && yarn run build:umd && yarn run build:types",
-    "build:umd": "BABEL_ENV=umd rollup --config scripts/rollup/config.js",
```
nunomaduro, line 26: Prevents builds of keep trash files from previous builds.

src/widgets/geo-search/geo-search.js
```diff
@@ -93,7 +93,6 @@ const suit = component('GeoSearch');
  *
  * Don't forget to explicitly set the `height` of the map container (default class `.ais-geo-search--map`), otherwise it won't be shown (it's a requirement of Google Maps).
  *
- * @type {WidgetFactory}
```
nunomaduro, line 96: API Extrator can't find this type, really don't know why. This file will be migrated to typescript anyways.

tsconfig.declaration.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig",
   "compilerOptions": {
+    "stripInternal": true,
```
nunomaduro, line 4: Prevents internals of being exposed.

nunomaduro:
cc @Haroenv @francoischalifour pull request updated.

PR #4378, opened 2020-04-08
chore(ts-migration): renames breadcrumb files

nunomaduro:
As usual, this is the first pull request while migrating breadcrumb files. This pull request renames the files to typescript.

The actual changes are in this pull request: https://github.com/algolia/instantsearch.js/pull/4379


yannickcr:
Seems you also need to rename some jest snapshots :wink:

nunomaduro:
@yannickcr Those are ok with we loose the history right? I have done it on the second pull request.

yannickcr:
@nunomaduro yes we are ok with loosing the history on these files, it was just to make the tests to pass.

But I don't know how "clean" you want this PR to be, because right now nothing is passing (tests, lint, type check :yum: )

nunomaduro:
@yannickcr I see, this pull request is just file renaming. The important one is : https://github.com/algolia/instantsearch.js/pull/4379

yannickcr:
As seen together, since everything is fixed in https://github.com/algolia/instantsearch.js/pull/4379 that will be merged just after this PR no need to fix this here :wink:

PR #4379, opened 2020-04-08
chore(ts-migration): migrates breadcrumb files content

nunomaduro:
This pull request migrates breadcrumb ( connectors and widget ) to typescript. Please take in consideration my comments on the pull request.

Internally it also adds new types to make refactor easy: TransformItems, and CreateURL.

### DIFF
--- a/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/src/components/Breadcrumb/Breadcrumb.tsx
@@ -32,10 +32,7 @@ export type BreadcrumbProps = {
   };
   createURL(value: string | undefined): string;
   refine(value: string | undefined): void;
-  /**
-   * @internal
-   */
-  canRefine?: boolean;
+  canRefine: boolean;
 };
 
 const Breadcrumb = ({
--- a/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/src/components/Breadcrumb/Breadcrumb.tsx
@@ -32,7 +32,7 @@ export type BreadcrumbProps = {
   };
   createURL(value: string | undefined): string;
   refine(value: string | undefined): void;
-  canRefine: boolean;
+  canRefine?: boolean;
 };
 
 const Breadcrumb = ({
--- a/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/src/components/Breadcrumb/Breadcrumb.tsx
@@ -32,6 +32,10 @@ export type BreadcrumbProps = {
   };
   createURL(value: string | undefined): string;
   refine(value: string | undefined): void;
+  /**
+   * @internal
+   */
+  canRefine?: boolean;
 };
 
 const Breadcrumb = ({
--- a/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
+++ b/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
@@ -3,9 +3,7 @@ import jsHelper, {
   SearchParameters,
 } from 'algoliasearch-helper';
 import { warning } from '../../../lib/utils';
-import connectBreadcrumb, {
-  BreadcrumbConnectorParams,
-} from '../connectBreadcrumb';
+import connectBreadcrumb from '../connectBreadcrumb';
 import { createSearchClient } from '../../../../test/mock/createSearchClient';
 import { createSingleSearchResponse } from '../../../../test/mock/createAPIResponse';
 import {
--- a/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
+++ b/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
@@ -185,6 +185,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
             name: 'category',
             attributes: ['category', 'sub_category'],
             separator: ' > ',
+            // @TODO Add missing type to js helper
             // @ts-ignore
             rootPath: null,
           },
--- a/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
+++ b/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
@@ -149,8 +149,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
             name: 'country',
             attributes: ['country', 'sub_country'],
             separator: ' > ',
-            // @ts-ignore
-            rootPath: null,
           },
         ],
       });
@@ -168,7 +166,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
           name: 'country',
           attributes: ['country', 'sub_country'],
           separator: ' > ',
-          rootPath: null,
         },
         {
           name: 'category',
--- a/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
+++ b/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
@@ -3,12 +3,21 @@ import jsHelper, {
   SearchParameters,
 } from 'algoliasearch-helper';
 import { warning } from '../../../lib/utils';
-import connectBreadcrumb from '../connectBreadcrumb';
+import connectBreadcrumb, {
+  BreadcrumbConnectorParams,
+} from '../connectBreadcrumb';
+import { createSearchClient } from '../../../../test/mock/createSearchClient';
+import { createSingleSearchResponse } from '../../../../test/mock/createAPIResponse';
+import {
+  createInitOptions,
+  createRenderOptions,
+} from '../../../../test/mock/createWidget';
 
 describe('connectBreadcrumb', () => {
   describe('Usage', () => {
     it('throws without render function', () => {
       expect(() => {
+        // @ts-ignore
         connectBreadcrumb()({});
       }).toThrowErrorMatchingInlineSnapshot(`
 "The render function is not valid (received type Undefined).
@@ -20,6 +29,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
     it('throws with undefined `attributes`', () => {
       expect(() => {
         connectBreadcrumb(() => {})({
+          // @ts-ignore
           attributes: undefined,
         });
       }).toThrowErrorMatchingInlineSnapshot(`
@@ -67,12 +77,12 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
     it('returns the `SearchParameters` with default value', () => {
       const render = () => {};
       const makeWidget = connectBreadcrumb(render);
-      const helper = jsHelper({}, '');
+      const helper = jsHelper(createSearchClient(), '');
       const widget = makeWidget({
         attributes: ['category', 'sub_category'],
       });
 
-      const actual = widget.getWidgetSearchParameters(helper.state, {
+      const actual = widget.getWidgetSearchParameters!(helper.state, {
         uiState: {},
       });
 
@@ -89,13 +99,13 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
     it('returns the `SearchParameters` with default a custom `separator`', () => {
       const render = () => {};
       const makeWidget = connectBreadcrumb(render);
-      const helper = jsHelper({}, '');
+      const helper = jsHelper(createSearchClient(), '');
       const widget = makeWidget({
         attributes: ['category', 'sub_category'],
         separator: ' / ',
       });
 
-      const actual = widget.getWidgetSearchParameters(helper.state, {
+      const actual = widget.getWidgetSearchParameters!(helper.state, {
         uiState: {},
       });
 
@@ -112,13 +122,13 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
     it('returns the `SearchParameters` with default a custom `rootPath`', () => {
       const render = () => {};
       const makeWidget = connectBreadcrumb(render);
-      const helper = jsHelper({}, '');
+      const helper = jsHelper(createSearchClient(), '');
       const widget = makeWidget({
         attributes: ['category', 'sub_category'],
         rootPath: 'TopLevel > SubLevel',
       });
 
-      const actual = widget.getWidgetSearchParameters(helper.state, {
+      const actual = widget.getWidgetSearchParameters!(helper.state, {
         uiState: {},
       });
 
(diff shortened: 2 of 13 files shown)


src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -32,6 +32,10 @@ export type BreadcrumbProps = {
   };
   createURL(value: string | undefined): string;
   refine(value: string | undefined): void;
+  /**
+   * @internal
+   */
+  canRefine?: boolean;
```
nunomaduro, line 38: Seems that `canRefine` property exists somewhere, but is not public. Do you know where it comes from? And why is used?

src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -5,54 +5,97 @@ import {
...
+
+  /**
+   * Sets the path of the hierarchical filter and triggers a new search.
+   */
+  refine: (value: BreadcrumbConnectorParamsItem['value']) => void;
+
+  /**
+   * @internal
+   */
+  canRefine: boolean;
```
nunomaduro, line 71: Seems that canRefine property exists somewhere, but is not public. Do you know where it comes from? And why is used?

src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -5,54 +5,97 @@ import {
...
+>;
+
+const connectBreadcrumb: BreadcrumbConnector = function connectBreadcrumb(
+  renderFn,
+  unmountFn = noop
+) {
   checkRendering(renderFn, withUsage());
-  return (widgetParams = {}) => {
+
+  type ConnectorState = {
```
nunomaduro, line 85: As usual, we also remove `this` references on the migration.

src/widgets/breadcrumb/breadcrumb.tsx
```diff
@@ -42,97 +45,88 @@ const renderer = ({ containerNode, cssClasses, renderState, templates }) => (
...
+   * CSS class to add to the links in the items.
+   */
+  link?: string | string[];
+};
+
+export type BreadcrumbTemplates = {
+  /**
+   * Label of the breadcrumb's first element.
+   */
+  home?: string | ((value: any) => string);
```
nunomaduro, line 89: Can you confirm that this also accepts a function?

src/widgets/breadcrumb/breadcrumb.tsx
```diff
@@ -42,97 +45,88 @@ const renderer = ({ containerNode, cssClasses, renderState, templates }) => (
...
+
+  /**
+   * Symbol used to separate the elements of the breadcrumb.
+   */
+  separator?: string | ((value: any) => string);
```
nunomaduro, line 94: Can you confirm that this also accepts a function?

src/types/connector.ts
```diff
@@ -71,3 +71,13 @@ export type Connector<TRendererOptions, TConnectorParams> = <TWidgetParams>(
    */
   unmountFn?: Unmounter
 ) => WidgetFactory<TConnectorParams, TWidgetParams>;
+
+/**
+ * Transforms the given items.
+ */
+export type TransformItems<TItem> = (objects: TItem[]) => TItem[];
```
nunomaduro, line 78: Is everyone ok with this name? I can also make it ItemsTransformer or something.

src/types/connector.ts
```diff
@@ -71,3 +71,13 @@ export type Connector<TRendererOptions, TConnectorParams> = <TWidgetParams>(
...
+
+/**
+ * Creates the URL for the given value.
+ */
+export type CreateURL<TValue> = (value: TValue) => string;
```
nunomaduro, line 83: Is everyone ok with this name? I can also make it URLCreator or something.

src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -116,17 +162,20 @@ export default function connectBreadcrumb(renderFn, unmountFn = noop) {
       render({ instantSearchInstance, results, state }) {
         const [{ name: facetName }] = state.hierarchicalFacets;
 
-        const facetValues = results.getFacetValues(facetName);
+        const facetValues = results.getFacetValues(
+          facetName,
+          {}
```
nunomaduro, line 167: Missing default value right?

src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -66,26 +109,29 @@ export default function connectBreadcrumb(renderFn, unmountFn = noop) {
...
-        this._createURL = facetValue => {
+        connectorState.createURL = facetValue => {
           if (!facetValue) {
             const breadcrumb = helper.getHierarchicalFacetBreadcrumb(
               hierarchicalFacetName
             );
             if (breadcrumb.length > 0) {
               return createURL(
-                helper.state.toggleRefinement(
+                helper.state.toggleFacetRefinement(
```
nunomaduro, line 119: helper.state.toggleRefinement got renamed to helper.state.toggleFacetRefinement

src/widgets/breadcrumb/breadcrumb.tsx
Haroenv: you can use the `Template` or `TemplateData` type we have. Since no value is given to those you can use a string too (for any template), but there won't be any value given, so it should be `() => string`

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -58,9 +58,7 @@ export type HitsPerPageConnectorParams = {
   /**
    * Function to transform the items passed to the templates.
    */
-  transformItems?: <TItem extends HitsPerPageConnectorParamsItem>(
-    objects: TItem[]
-  ) => TItem[];
+  transformItems?: TransformItems<HitsPerPageConnectorParamsItem>;
```
Haroenv, line 61: this now no longer allows the user to pass another type which has more values since it's no longer generic. I don't think that's a big deal, but something to keep in mind.

src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -102,11 +148,11 @@ export default function connectBreadcrumb(renderFn, unmountFn = noop) {
 
         renderFn(
           {
-            createURL: this._createURL,
+            createURL: connectorState.createURL,
             canRefine: false,
```
Haroenv, line 152: here's & in render where canRefine comes from

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
```diff
@@ -4,11 +4,18 @@ import jsHelper, {
...
+import {
+  createInitOptions,
+  createRenderOptions,
+} from '../../../../test/mock/createWidget';
 
 describe('connectBreadcrumb', () => {
   describe('Usage', () => {
     it('throws without render function', () => {
       expect(() => {
+        // @ts-ignore
```
Haroenv, line 18: can be @ts-expect-error now, right?

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
```diff
@@ -135,12 +143,13 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
...
       const render = () => {};
       const makeWidget = connectBreadcrumb(render);
-      const helper = jsHelper({}, '', {
+      const helper = jsHelper(createSearchClient(), '', {
         hierarchicalFacets: [
           {
             name: 'country',
             attributes: ['country', 'sub_country'],
             separator: ' > ',
+            // @ts-ignore
```
Haroenv, line 152: why did this one need to be ignored? value seems valid to me.

src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -116,17 +162,20 @@ export default function connectBreadcrumb(renderFn, unmountFn = noop) {
...
+        const facetValues = results.getFacetValues(
+          facetName,
+          {}
+        ) as SearchResults.HierarchicalFacet;
         const data = Array.isArray(facetValues.data) ? facetValues.data : [];
         const items = transformItems(shiftItemsValues(prepareItems(data)));
 
         renderFn(
           {
             canRefine: items.length > 0,
```
Haroenv, line 174: here's the other canRefine. Why would you say it's internal?

src/widgets/breadcrumb/breadcrumb.tsx
francoischalifour: Yes—we've got a [type for templates](https://github.com/algolia/instantsearch.js/blob/bae6ed2875b03a8a43c576085a19f2ceb43ba1f2/src/types/widget.ts#L199-L201).

src/types/connector.ts
nunomaduro: @francoischalifour Actually, if you return a item that has the shape ( extends ) it will work too.

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
nunomaduro: We are not in TS 3.9 yet. 👍

src/components/Breadcrumb/Breadcrumb.tsx
nunomaduro: @francoischalifour Component props are always equal to: WidgetOptions + ComponentSpecificProps?

src/widgets/breadcrumb/breadcrumb.tsx
nunomaduro: The `TTemplateData` is `any`?

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
nunomaduro: Actually, `rootPath`, as specified on the helper, can't be null. `null !== undefined`. Any ideas?

Haroenv: we should probably use `undefined` instead of `null` OR update the types then

src/components/Breadcrumb/Breadcrumb.tsx
francoischalifour: On top of my mind I don't know if it's always the case—but component props are for internal use anyway, so no need to over-abstract them.

nunomaduro: Ok, adding just the can refine.

nunomaduro:
Well feedback adressed, going to rebase now.

nunomaduro:
Rebased, pull request ready to have a final validation.

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
nunomaduro: Done.

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
```diff
@@ -173,12 +179,13 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
...
       const makeWidget = connectBreadcrumb(render);
-      const helper = jsHelper({}, '', {
+      const helper = jsHelper(createSearchClient(), '', {
         hierarchicalFacets: [
           {
             name: 'category',
             attributes: ['category', 'sub_category'],
             separator: ' > ',
+            // @ts-ignore
             rootPath: null,
```
francoischalifour, line 189: Can this be changed to `undefined`?

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
```diff
@@ -256,12 +265,13 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
...
       const render = () => {};
       const makeWidget = connectBreadcrumb(render);
-      const helper = jsHelper({}, '', {
+      const helper = jsHelper(createSearchClient(), '', {
         hierarchicalFacets: [
           {
             name: 'category',
             attributes: ['category', 'sub_category'],
             separator: ' > ',
+            // @ts-ignore
```
francoischalifour, line 274: This shouldn't be needed.

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
```diff
@@ -304,14 +315,16 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
...
-    widget.init({
-      helper,
-      state: helper.state,
-      createURL: () => '#',
-    });
+    widget.init!(
+      createInitOptions({
+        helper,
+        state: helper.state,
+        createURL: () => '#',
```
francoischalifour, line 325: This can be removed.

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
```diff
@@ -579,15 +607,19 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
...
-    widget.init({
-      helper,
-      state: helper.state,
-      createURL: state => state,
-    });
+    widget.init!(
+      createInitOptions({
+        helper,
+        state: helper.state,
+        createURL: state => (state as unknown) as string,
```
francoischalifour, line 620: You can return `JSON.stringify(state)` to keep type safety (in other places too).

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
nunomaduro: @francoischalifour Actually the `rootPath` do not exist on hierarchicalFacets type. 😢 

It's fine passing a ts ignore in this cases right? What do you propose?

nunomaduro:
```
    type HierarchicalFacet = {
      name: string;
      attributes: string[];
      separator: string;
    };
```

francoischalifour: You mean that it's mistyped or that it's not supposed to be there?

nunomaduro: I am not sure what an `HierarchicalFacet` is. But at the moment, the algoliasearch-helper don't contain the property `rootPath` on the `HierarchicalFacet`. Therefore, if we don't ts ignore on tests, typescript complains that property don't exist in HierarchicalFacet.

francoischalifour: I meant, are the tests passing without `rootPath`? Perhaps it was not meant to be provided from the start.

Haroenv: you  seem to have removed some of the other `rootPath: null`, but not this one. Is that on purpose?

PR #4380, opened 2020-04-09
chore(ts-migration): improves validating of types with api extractor

nunomaduro:
This pull request improves the current api extrator implementation, creating a temporary single public file API with:

```
export * from '../es/';
export * from '../es/connectors';
export * from '../es/middleware';
export * from '../es/types';
export * from '../es/widgets';
```

Previously, since we were validating file by file, the file `connectors/index.d.ts` was complaining that  'types' folder contains types that are not exported at `connectors/index.d.ts` ( the root level for api extrator ).

With this change, we are creating a temporary single file public api, and api extrator will no longer report those false positives.

@Haroenv We can't avoid the temporary folder in this case, as node don't have rights to write on the root system.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,6 +4,7 @@
 dist
 .cache
 node_modules/
+.temp/
 
 # TODO: don't ignore these examples once they're updated
 /examples/calendar-widget
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,10 +1,10 @@
 /cjs
 /es
+/scripts/build/.temp
 /website
 dist
 .cache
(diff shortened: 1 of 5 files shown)


nunomaduro:
Tests are failing as we are waiting for https://github.com/algolia/instantsearch.js/pull/4379.

Haroenv:
I've tried this out and made some changes, since it didn't work if the folder didn't already exist for me

nunomaduro:
Ho, actually forgot to commit the `.temp/.gitkeep`. Let me check the code you wrote.

PR #4381, opened 2020-04-10
chore(ts-migration): creates refine generic type

nunomaduro:
This pull request makes the refine type easier to reuse.

### DIFF
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -6,7 +6,7 @@ import {
   noop,
   warning,
 } from '../../lib/utils';
-import { Hits, Connector } from '../../types';
+import { Hits, Connector, Refine } from '../../types';
 
 const withUsage = createDocumentationMessageGenerator({
   name: 'autocomplete',
@@ -51,7 +51,7 @@ export type AutocompleteRendererOptions = {
   /**
    * Searches into the indices with the provided query.
    */
-  refine: (query: string) => void;
+  refine: Refine<string>;
 };
 
 export type AutocompleteConnector = Connector<
@@ -94,7 +94,7 @@ search.addWidgets([
     );
 
     type ConnectorState = {
-      refine?: (query: string) => void;
+      refine?: Refine<string>;
     };
 
     const connectorState: ConnectorState = {};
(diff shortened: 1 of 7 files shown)


src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -51,7 +51,7 @@ export type AutocompleteRendererOptions = {
   /**
    * Searches into the indices with the provided query.
    */
-  refine: (query: string) => void;
+  refine: Refine<string>;
```
Haroenv, line 54: the name "query" is now gone from autocomplete, is that correct?

```
refine(
    ^^^query: string
```

nunomaduro: Correct.

Haroenv: Tried to find a solution but couldn't :( https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBAZgdgHgCoDkCGBbCA+KBeKACgGcB7LAcQjggCcBLAY3SwC4pjgG4BzASny4AbqXoATAFASA9NKjAAFvWJRlcAOTAoQtABtxMuaEixEqTDnxEA2uawBddp279B20ZInHo9OJzRwjNAE8AjqGCAAShBgumhBWHDALBDq2FKMpH5aAB7svv6BwUQR0bHxEInJFgJ4uACucGIQML4QnjmE6gCSUADuygryCtBo9cCkmRixEMDQmfW6YlA89ELQpTFxCdTVWKrZEGjLpDAc5BBUNAzMFup8UrJQnQ9AA

nunomaduro: I think it's fine, 👍

nunomaduro:
Got it!

francoischalifour:
@Haroenv Can you explain your POV? I think this type is completely valid, we use it everywhere. If this is because of the argument name not displaying in IntelliSense, we've got the type _and_ the documentation now.

Haroenv:
the argument is called `query`, `page`, `value`, which I find more useful names than `value` in all cases. I didn't suggest we should immediately close this PR, just giving my pov in a way that prevented the PR to immediately be merged.

At the end of the day the extracted type didn't seem to bring that much value (to me), since it's a function which takes one argument & returns void, it's ok if that's repeated in every connector (for me)

PR #4383, opened 2020-04-10
chore(ts-migration): migrates configure and autocomplete related stories

nunomaduro:
This pull request migrates stories of connectors and widgets already migrated to typescript

### DIFF
--- a/stories/autocomplete.stories.ts
+++ b/stories/autocomplete.stories.ts
@@ -1,5 +1,7 @@
 import { storiesOf } from '@storybook/html';
 import { withHits } from '../.storybook/decorators';
+import { connectAutocomplete } from '../src/connectors';
+import { configure } from '../src/widgets';
 
 storiesOf('Basics|Autocomplete', module).add(
   'default',
@@ -8,7 +10,7 @@ storiesOf('Basics|Autocomplete', module).add(
 
     container.appendChild(instantSearchAutocomplete);
 
-    const customAutocomplete = instantsearch.connectors.connectAutocomplete(
+    const customAutocomplete = connectAutocomplete<{ container: HTMLElement }>(
       (renderOptions, isFirstRender) => {
         const {
           indices,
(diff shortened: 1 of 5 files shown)


stories/autocomplete.stories.ts
```diff
@@ -30,8 +32,10 @@ storiesOf('Basics|Autocomplete', module).add(
           widgetParams.container.appendChild(list);
         }
 
-        widgetParams.container.querySelector('input').value = currentRefinement;
-        widgetParams.container.querySelector('ul').innerHTML = indices
+        widgetParams.container.querySelector(
```
francoischalifour, line 35: You can type the selectors because `querySelector` is generic. Then you might not need the `!`.

```suggestion
        widgetParams.container.querySelector<HTMLInputElement>(
```

stories/autocomplete.stories.ts
```diff
@@ -8,7 +10,7 @@ storiesOf('Basics|Autocomplete', module).add(
 
     container.appendChild(instantSearchAutocomplete);
 
-    const customAutocomplete = instantsearch.connectors.connectAutocomplete(
+    const customAutocomplete = connectAutocomplete<{ container: HTMLElement }>(
```
Haroenv, line 13: how come container isn't inferred?

nunomaduro: @Haroenv There is no container on the connector, only on the widget. Here you have a custom widget, therefore you need to pass your widgets types.

stories/autocomplete.stories.ts
nunomaduro: I don't think so, the type returned in querySelector can only be truly known at run time, therefore it can be *indeed* null.

francoischalifour: Right—you can still type it as `HTMLInputElement` to have type safety for `value` though.

PR #4385, opened 2020-04-15
chore(ts-migration): migrates clear refinements connector, widgets an…

nunomaduro:
This pull request migrates the clear refinements connector, widgets, and stories content. Note: 

There is a couple of questions, that's why I have putted the **draft** state.

### DIFF
--- a/src/connectors/clear-refinements/connectClearRefinements.ts
+++ b/src/connectors/clear-refinements/connectClearRefinements.ts
@@ -14,25 +14,6 @@ const withUsage = createDocumentationMessageGenerator({
   connector: true,
 });
 
-export type ClearRefinementsConnectorParamsItem = {
-  /**
-   * Label to display in the option.
-   */
-  label: string;
-
-  /**
-   * Number of hits to display per page.
-   */
-  value: number;
-
-  /**
-   * The default hits per page on first search.
-   *
-   * @default false
-   */
-  default?: boolean;
-};
-
 export type ClearRefinementsConnectorParams = {
   /**
    * The attributes to include in the refinements to clear (all by default). Cannot be used with `excludedAttributes`.
@@ -47,7 +28,7 @@ export type ClearRefinementsConnectorParams = {
   /**
    * Function to transform the items passed to the templates.
    */
-  transformItems?: TransformItems<ClearRefinementsConnectorParamsItem>;
+  transformItems?: TransformItems<string>;
 };
 
 export type ClearRefinementsRendererOptions = {
--- a/src/connectors/clear-refinements/connectClearRefinements.ts
+++ b/src/connectors/clear-refinements/connectClearRefinements.ts
@@ -71,6 +71,7 @@ export type ClearRefinementsConnector = Connector<
   ClearRefinementsRendererOptions,
   ClearRefinementsConnectorParams
 >;
+
 const connectClearRefinements: ClearRefinementsConnector = function connectClearRefinements(
   renderFn,
   unmountFn = noop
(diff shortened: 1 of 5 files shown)


src/connectors/clear-refinements/connectClearRefinements.ts
```diff
@@ -7,73 +7,83 @@ import {
...
- *
- * // mount widget on the page
- * search.addWidgets([
- *   customClearRefinementsWidget({
- *     containerNode: $('#custom-clear-all-container'),
- *   })
- * ]);
- */
-export default function connectClearRefinements(renderFn, unmountFn = noop) {
+export type ClearRefinementsConnectorParamsItem = {
```
nunomaduro, line 17: I asked on slack, but I am quite sure that an item on clear refinements is a simple string.

src/connectors/clear-refinements/connectClearRefinements.ts
```diff
@@ -7,73 +7,83 @@ import {
...
+   * Function to transform the items passed to the templates.
+   */
+  transformItems?: TransformItems<ClearRefinementsConnectorParamsItem>;
+};
+
+export type ClearRefinementsRendererOptions = {
+  /**
+   * Triggers the clear of all the currently refined values.
+   */
+  refine: () => void;
```
nunomaduro, line 57: Can you confirm that there is no param here?

src/connectors/clear-refinements/connectClearRefinements.ts
```diff
@@ -7,73 +7,83 @@ import {
...
+
+  /**
+   * Indicates if search state is refined.
+   */
+  hasRefinements: boolean;
+
+  /**
+   * Creates a url for the next state when refinements are cleared.
+   */
+  createURL: CreateURL<void>;
```
nunomaduro, line 67: Can you confirm that there is no param here?

src/connectors/clear-refinements/connectClearRefinements.ts
nunomaduro: On the test for example, there are simple strings: https://github.com/algolia/instantsearch.js/pull/4385/files#diff-93bf43d5d7e3988ea2803945111d3d80L558.

nunomaduro: And same on docs: https://www.algolia.com/doc/api-reference/widgets/clear-refinements/js/#widget-param-transformitems.

Haroenv:
In the future we can probably avoid this connector and use currentRefinements instead, like in React

src/connectors/clear-refinements/connectClearRefinements.ts
francoischalifour: You're right, they're strings. In these cases, you can [try it in a sandbox](https://codesandbox.io/s/instantsearchjs-clear-refinements-qebvy?file=/src/app.js).

src/connectors/clear-refinements/connectClearRefinements.ts
francoischalifour: Yep

https://github.com/algolia/instantsearch.js/blob/bae6ed2875b03a8a43c576085a19f2ceb43ba1f2/src/connectors/clear-refinements/connectClearRefinements.js#L130-L141

src/connectors/clear-refinements/connectClearRefinements.ts
francoischalifour: Yep

https://github.com/algolia/instantsearch.js/blob/bae6ed2875b03a8a43c576085a19f2ceb43ba1f2/src/connectors/clear-refinements/connectClearRefinements.js#L143-L153

PR #4388, opened 2020-04-17
chore(ts-migration): migrates numeric menu and panel connector, widget, and stories

nunomaduro:
This pull request migrates 2 widgets, 1 connector and 2 stories:

- The numeric menu connector, widget, and stories;
- The panel widget, and stories;

### DIFF
--- a/.storybook/playgrounds/default.ts
+++ b/.storybook/playgrounds/default.ts
@@ -1,4 +1,5 @@
 import instantsearch from '../../src/index';
+import { panel, numericMenu } from '../../src/widgets';
 
 export const hitsItemTemplate = `
 <div
@@ -32,7 +33,7 @@ function instantSearchPlayground({
   const refinementList = document.createElement('div');
   leftPanel.appendChild(refinementList);
 
-  const brandList = instantsearch.widgets.panel({
+  const brandList = panel({
     templates: {
       header: 'Brands',
     },
@@ -45,18 +46,14 @@ function instantSearchPlayground({
     }),
   ]);
 
-  const numericMenu = document.createElement('div');
-  leftPanel.appendChild(numericMenu);
+  const numericMenuElement = document.createElement('div');
+  leftPanel.appendChild(numericMenuElement);
 
-  const priceMenu = instantsearch.widgets.panel({
-    templates: {
-      header: 'Price',
-    },
-  })(instantsearch.widgets.numericMenu);
+  const priceMenu = panel({ templates: { header: 'Price' } })(numericMenu);
 
   search.addWidgets([
     priceMenu({
-      container: numericMenu,
+      container: numericMenuElement,
       attribute: 'price',
       items: [
         { label: 'All' },
@@ -71,7 +68,7 @@ function instantSearchPlayground({
   const ratingMenu = document.createElement('div');
   leftPanel.appendChild(ratingMenu);
 
-  const ratingList = instantsearch.widgets.panel({
+  const ratingList = panel({
     templates: {
       header: 'Rating',
     },
--- a/.storybook/playgrounds/default.ts
+++ b/.storybook/playgrounds/default.ts
@@ -46,14 +46,14 @@ function instantSearchPlayground({
     }),
   ]);
 
-  const numericMenuElement = document.createElement('div');
-  leftPanel.appendChild(numericMenuElement);
+  const numericMenuContainer = document.createElement('div');
+  leftPanel.appendChild(numericMenuContainer);
 
   const priceMenu = panel({ templates: { header: 'Price' } })(numericMenu);
 
   search.addWidgets([
     priceMenu({
-      container: numericMenuElement,
+      container: numericMenuContainer,
       attribute: 'price',
       items: [
         { label: 'All' },
--- a/src/connectors/numeric-menu/__tests__/connectNumericMenu-test.ts
+++ b/src/connectors/numeric-menu/__tests__/connectNumericMenu-test.ts
@@ -494,7 +494,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/numeric-men
         ]),
         state: helper.state,
         helper,
-        createURL: () => '#',
       })
     );
 
@@ -516,7 +515,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/numeric-men
         ]),
         state: helper.state,
         helper,
-        createURL: () => '#',
       })
     );
 
--- a/src/connectors/numeric-menu/__tests__/connectNumericMenu-test.ts
+++ b/src/connectors/numeric-menu/__tests__/connectNumericMenu-test.ts
@@ -2,11 +2,24 @@ import jsHelper, {
   SearchResults,
   SearchParameters,
 } from 'algoliasearch-helper';
-import connectNumericMenu from '../connectNumericMenu';
+import connectNumericMenu, {
+  NumericMenuConnectorParamsItem,
+  NumericMenuRendererOptionsItem,
+} from '../connectNumericMenu';
+import { createSearchClient } from '../../../../test/mock/createSearchClient';
+import {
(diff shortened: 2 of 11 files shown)


src/widgets/panel/panel.tsx
```diff
@@ -96,27 +123,25 @@ const renderer = ({
...
-  container: HTMLElement | string;
-  [key: string]: any;
-};
-type PanelWidget = (
-  params?: PanelWidgetParams
-) => (
-  widgetFactory: WidgetFactory<{}, NestedWidgetOptions>
-) => (widgetOptions: NestedWidgetOptions) => Widget;
+export type PanelWidget = (
+  params: PanelWidgetOptions
```
eunjae-lee, line 127: I guess `PanelWidgetOptions` should be renamed to `PanelWidgetParams`?

.storybook/playgrounds/default.ts
```diff
@@ -45,18 +46,14 @@ function instantSearchPlayground({
     }),
   ]);
 
-  const numericMenu = document.createElement('div');
-  leftPanel.appendChild(numericMenu);
+  const numericMenuElement = document.createElement('div');
```
francoischalifour, line 49: In other stories we call these `numericMenuContainer`.

src/connectors/numeric-menu/__tests__/connectNumericMenu-test.ts
```diff
@@ -435,26 +472,31 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/numeric-men
...
-      createURL: () => '#',
-    });
+    widget.render!(
+      createRenderOptions({
+        results: new SearchResults(helper.state, [
+          createSingleSearchResponse(),
+        ]),
+        state: helper.state,
+        helper,
+        createURL: () => '#',
```
francoischalifour, line 497: You can remove `createURL` when they're empty. They're covered by `createRenderOptions`.

src/widgets/panel/panel.tsx
nunomaduro: I am not sure, because for types is always have been WidgetOptions.

Anyways, going to address of all of that together.

Just created this task for that: https://algolia.atlassian.net/browse/IFW-1260?atlOrigin=eyJpIjoiZjQ3ODg0ZDg0YmVmNDEzNmJjZTA1ZTAzYWRhNDEyMTEiLCJwIjoiaiJ9

src/widgets/numeric-menu/__tests__/numeric-menu-test.ts
```diff
@@ -55,23 +66,25 @@ describe('numericMenu()', () => {
...
 
     createURL = () => '#';
-    widget.init({ helper, instantSearchInstance: {} });
+    widget.init!({ helper, instantSearchInstance: {} });
   });
 
   it('calls twice render(<RefinementList props />, container)', () => {
-    widget.render({ state, results, createURL });
-    widget.render({ state, results, createURL });
+    widget.render!({ state, results, createURL });
```
francoischalifour, line 86: Any reasons the `createRenderOptions` and `createInitOptions` are not used below?

nunomaduro: @francoischalifour Nice catch.

nunomaduro: Fixed. Note that the snapshots got updated as we use a templateConfig as object instead of undefined.

src/widgets/numeric-menu/numeric-menu.tsx
```diff
@@ -47,77 +50,92 @@ const renderer = ({
...
+   * CSS class to add to each radio element (when using the default template).
+   */
+  radio?: string | string[];
+};
+
+export type NumericMenuTemplates = {
+  /**
+   * Item template, provided with `label` (the name in the configuration), `isRefined`, `url`, `value` (the setting for the filter) data properties.
+   */
+  item?: Template;
```
francoischalifour, line 99: We can type the template [as described in the documentation](https://www.algolia.com/doc/api-reference/widgets/numeric-menu/js/#widget-param-item) since it's generic.

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -4,83 +4,88 @@ import {
...
    * Higher bound of the option (<=)
    */
-  start: number;
+  start?: number;
+
   /**
    * Lower bound of the option (>=)
    */
-  end: number;
+  end?: number;
```
Haroenv, line 29: they  aren't both optional, at least one is required, this means it's more something like:

```ts
NumericMenuConnectorParamsItem = {
  label: string;
  start: string;
  end?: string;} | { label: string; start?: string; end: string }
```

src/widgets/panel/__tests__/panel-test.ts
```diff
@@ -25,7 +25,7 @@ beforeEach(() => {
 describe('Usage', () => {
   test('without arguments does not throw', () => {
     expect(() => {
-      panel();
+      panel({});
```
Haroenv, line 28: this test is to see that we don't destructure without default, giving an object changes it :)

src/widgets/panel/panel.tsx
```diff
@@ -162,10 +187,8 @@ const panel: PanelWidget = (widgetParams = {} as PanelWidgetParams) => {
     footer: cx(suit({ descendantName: 'footer' }), userCssClasses.footer),
   };
 
-  return (widgetFactory: WidgetFactory<{}, NestedWidgetOptions>) => (
-    widgetOptions = {} as NestedWidgetOptions
-  ): Widget => {
-    const { container } = widgetOptions;
+  return widgetFactory => widgetOptions => {
+    const { container } = widgetOptions || ({} as typeof widgetOptions);
```
Haroenv, line 191: this typeof trick is really cool

src/widgets/numeric-menu/numeric-menu.tsx
nunomaduro: @francoischalifour Can you rephrase what are you asking for? Sorry, didn't got it.

src/connectors/numeric-menu/connectNumericMenu.ts
nunomaduro: @Haroenv Not sure if we should go far on that one. As example, on API Clients we don't go that far on query rules.

src/widgets/panel/__tests__/panel-test.ts
nunomaduro: Fixed, thanks.

src/widgets/numeric-menu/numeric-menu.tsx
Haroenv: something roughly like this?

```
Template<{
  /**
   * the name of the attribute.
   */
  attribute: string;
  /**
   *  the label for the option.
   */
  label: string;
  /**
   * the encoded URL of the bounds object with a {start, end} form. This value can be used verbatim in the webpage and can be read by refine directly. If you want to inspect the value, you can do JSON.parse(window.decodeURI(value)) to get the object.
   */
  value: string;
  /**
   *  whether or not the refinement is selected.
   */
  isRefined: boolean;
  /**
   * the URL with the applied refinement.
   */
  url: string;
  /**
   * the CSS classes provided to the widget.
   */
  cssClasses: object;
}>
```

nunomaduro: @Haroenv Can you confirm that this type used in this generic, is no used elsewhere?

nunomaduro: @Haroenv Can you check the contents now?

nunomaduro:
Going to merge so I can move forward. Fell free to write down if you see anything that you don't agree. I can adjust later

PR #4391, opened 2020-04-22
chore(ts-migration): migrates infinite hits contents

nunomaduro:
This pull request migrates infinite hits contents.

I have a problem tho, I would like to pass a type to transform items:

```
    const widget = makeWidget({
      transformItems: <MYSPECIALTYPEHERE>(items) =>
        items.map(item => ({
          ...item,
          _highlightResult: {
            name: {
              value: item._highligh¨tResult.name.value.toUpperCase(),
            },
          },
        })),
      escapeHTML: true,
```

Going to try to solve this problem on top with @yannickcr

### DIFF
--- a/src/components/InfiniteHits/InfiniteHits.tsx
+++ b/src/components/InfiniteHits/InfiniteHits.tsx
@@ -27,7 +27,7 @@ type InfiniteHitsProps = {
   showMore: () => void;
   templateProps: {
     [key: string]: any;
-    templates: InfiniteHitsTemplates;
+    templates: Partial<InfiniteHitsTemplates>;
   };
   isFirstPage: boolean;
   isLastPage: boolean;
--- a/src/components/InfiniteHits/InfiniteHits.tsx
+++ b/src/components/InfiniteHits/InfiniteHits.tsx
@@ -27,7 +27,7 @@ type InfiniteHitsProps = {
   showMore: () => void;
   templateProps: {
     [key: string]: any;
-    templates: Partial<InfiniteHitsTemplates>;
+    templates: InfiniteHitsTemplates;
   };
   isFirstPage: boolean;
   isLastPage: boolean;
--- a/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
+++ b/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
@@ -376,7 +376,9 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     const renderFn = jest.fn();
     const makeWidget = connectInfiniteHits(renderFn);
     const widget = makeWidget({
-      transformItems: items => items.map(() => ({ name: 'transformed' })),
+      transformItems: items => {
+        return items.map(item => ({ ...item, name: 'transformed' }));
+      },
     });
 
     const helper = algoliasearchHelper({} as SearchClient, '', {});
@@ -418,9 +420,11 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
 
     const transformedHits = [
       {
+        objectID: '1',
         name: 'transformed',
       },
       {
+        objectID: '2',
         name: 'transformed',
       },
     ];
@@ -435,14 +439,12 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     const makeWidget = connectInfiniteHits(renderFn);
     const widget = makeWidget({
       transformItems: items =>
-        items.map(item => ({
-          ...item,
-          _highlightResult: {
-            name: {
-              value: item._highlightResult.name.value.toUpperCase(),
-            },
-          },
-        })),
+        items.map(item => {
+          // @ts-ignore
+          item._highlightResult.name.value = item._highlightResult.name.value.toUpperCase();
+
+          return item;
+        }),
       escapeHTML: true,
     });
 
--- a/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
+++ b/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
@@ -2,7 +2,7 @@ import algoliasearchHelper, {
   SearchResults,
   SearchParameters,
 } from 'algoliasearch-helper';
-import { SearchClient } from '../../../types';
+import { SearchClient, HitAttributeHighlightResult } from '../../../types';
 import { createInstantSearch } from '../../../../test/mock/createInstantSearch';
 import {
   createInitOptions,
@@ -440,8 +440,10 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     const widget = makeWidget({
       transformItems: items =>
         items.map(item => {
-          // @ts-ignore
-          item._highlightResult.name.value = item._highlightResult.name.value.toUpperCase();
+          const name = item._highlightResult!
+            .name as HitAttributeHighlightResult;
+
+          name.value = name.value.toUpperCase();
 
           return item;
         }),
--- a/src/connectors/infinite-hits/connectInfiniteHits.ts
+++ b/src/connectors/infinite-hits/connectInfiniteHits.ts
@@ -3,7 +3,7 @@ import {
   AlgoliaSearchHelper as Helper,
   SearchParameters,
 } from 'algoliasearch-helper';
-import { Hits, Connector, TransformItems } from '../../types';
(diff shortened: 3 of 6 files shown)


src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -12,16 +12,50 @@ import {
...
+   * @default `false`
+   */
+  showPrevious?: boolean;
+
+  /**
+   * Receives the items, and is called before displaying them.
+   * Useful for mapping over the items to transform, and remove or reorder them.
+   */
+  transformItems?: TransformItems<{
+    __hitIndex: number;
```
Haroenv, line 36: how come this isn't `{__hitIndex: number} & Hit` or similar?

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -376,7 +376,9 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     const renderFn = jest.fn();
     const makeWidget = connectInfiniteHits(renderFn);
     const widget = makeWidget({
-      transformItems: items => items.map(() => ({ name: 'transformed' })),
+      transformItems: items => {
+        return items.map(item => ({ ...item, name: 'transformed' }));
```
nunomaduro, line 380: Transformed items need to have the same shape.

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -435,14 +439,12 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
-          ...item,
-          _highlightResult: {
-            name: {
-              value: item._highlightResult.name.value.toUpperCase(),
-            },
-          },
-        })),
+        items.map(item => {
+          // @ts-ignore
+          item._highlightResult.name.value = item._highlightResult.name.value.toUpperCase();
```
nunomaduro, line 444: Someone knows where the name is coming from?

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -12,16 +12,52 @@ import {
...
+   * @default `false`
+   */
+  showPrevious?: boolean;
+
+  /**
+   * Receives the items, and is called before displaying them.
+   * Useful for mapping over the items to transform, and remove or reorder them.
+   */
+  transformItems?: TransformItems<
+    AlgoliaHit & {
```
nunomaduro, line 36: Is this `AlgoliaHit` or `Hit`?  The `Hit` also contains `__position`.

src/widgets/infinite-hits/infinite-hits.tsx
```diff
@@ -34,92 +34,84 @@ export type InfiniteHitsCSSClasses = {
...
-   * @default `false`
-   */
-  showPrevious?: boolean;
-  /**
-   * Receives the items, and is called before displaying them.
-   * Useful for mapping over the items to transform, and remove or reorder them.
+   * The template to use for each result.
    */
-  transformItems?: (items: any[]) => any[];
+  item?: Template<Hit>;
```
nunomaduro, line 94: Is this AlgoliaHit or Hit? The Hit also contains __position.

src/widgets/infinite-hits/infinite-hits.tsx
```diff
@@ -216,7 +208,12 @@ const infiniteHits: InfiniteHits = (
     connectInfiniteHits
   )(specializedRenderer, () => render(null, containerNode));
 
-  return makeInfiniteHits({ escapeHTML, transformItems, showPrevious });
+  return makeInfiniteHits({
+    container: containerNode,
```
nunomaduro, line 212: You agree that this was a mistake not having `container` right?

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
Haroenv: I don't understand your question, but from here, it's from  here,  no? https://github.com/algolia/instantsearch.js/blob/d5f21a4e9a0230e56f133347f57acd7f09bb34f8/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts#L462-L479

src/components/InfiniteHits/InfiniteHits.tsx
```diff
@@ -27,7 +27,7 @@ type InfiniteHitsProps = {
   showMore: () => void;
   templateProps: {
     [key: string]: any;
-    templates: InfiniteHitsTemplates;
+    templates: Partial<InfiniteHitsTemplates>;
```
francoischalifour, line 30: Why does this need to be partial now?

src/connectors/infinite-hits/connectInfiniteHits.ts
Haroenv: it's Hit, and it doesn't have __hitIndex I think: https://github.com/algolia/instantsearch.js/blob/d5f21a4e9a0230e56f133347f57acd7f09bb34f8/src/lib/utils/hits-absolute-position.ts#L10

src/widgets/infinite-hits/infinite-hits.tsx
Haroenv: Hit

src/widgets/infinite-hits/infinite-hits.tsx
Haroenv: The containerNode is already passed to `specializedRenderer`, makeWidget is the connector, which doesn't use container I think. It does seem a little inconsistent with the other widgets though

src/connectors/infinite-hits/connectInfiniteHits.ts
francoischalifour: We call `transformItems` after all the transformations, so this is with the internal hit containing the position and `queryId`:

https://github.com/algolia/instantsearch.js/blob/d8f6818a31d195ab7d818f5fdb5661256741a1b5/src/connectors/infinite-hits/connectInfiniteHits.ts#L192

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
nunomaduro: Got it, thanks! Let me check if I can avoid the ts ignore.

src/widgets/infinite-hits/infinite-hits.tsx
francoischalifour: See answer above. From reading the code, that's the internal hit—but you can try it in a test or sandbox.

src/widgets/infinite-hits/infinite-hits.tsx
francoischalifour: No, the connector doesn't need to be aware of the `container` (same as any other connector).

nunomaduro: @Haroenv @francoischalifour I see. Going to still pass to the container so we don't perform a TS ignore.

nunomaduro: I see. Going to pass the container anyways if sounds ok for you guys, because normally the connector returns a function that accepts the widgetParams. The widgetParams in the case requires a container.

src/components/InfiniteHits/InfiniteHits.tsx
nunomaduro: Nice catch.

src/widgets/infinite-hits/infinite-hits.tsx
francoischalifour: It means that there's an issue somewhere else then. Why did we not need to pass `container` in other connectors?

Connectors are not bound to any DOM, we use them for Vue InstantSearch, Angular InstantSearch and later, React InstantSearch, where DOM elements are not part of the API.

nunomaduro: Humm; let's zoom on this one?

src/widgets/infinite-hits/infinite-hits.tsx
```diff
@@ -212,11 +207,15 @@ const infiniteHits: InfiniteHits = (
     renderState: {},
   });
 
-  const makeInfiniteHits = withInsights(
+  const makeInfiniteHitsWidget = withInsights(
```
francoischalifour, line 210: We can call this back `makeInfiniteHits`.

PR #4392, opened 2020-04-22
fix(router): don't write an existing URL

Haroenv:
**Summary**

Every write puts a history entry, we never want to write the same as what's already written to the URL.

Demonstration of the bug being fixed in user land: https://codesandbox.io/s/vigilant-mayer-u3lyd?file=/src/app.js:2097-2279 (note that this searchFunction is very similar to the default one in Shopify)

The history router didn't have tests yet, so I added just one for this case, as not to lose time writing tests for existing cases.

**Result**

condition around pushing the URL, meaning it won't write if the URL is already the same. In theory a createURL could give a non-qualified URL, but that would make this condition a lot more complicated, so I'm ok with that caveat.

Also fixed the typescript of the class wrapper a bit

### DIFF
--- a/src/lib/routers/__tests__/history.test.ts
+++ b/src/lib/routers/__tests__/history.test.ts
@@ -26,6 +26,32 @@ describe('life cycle', () => {
     expect(pushState).toHaveBeenCalledTimes(1);
   });
 
+  it('does not write if already externally updated to desired URL', async () => {
+    const pushState = jest.spyOn(window.history, 'pushState');
+    const router = historyRouter({
+      writeDelay: 0,
+    });
+
+    const fakeState = { identifier: 'fake state' };
+
+    router.write({ some: 'state one' });
+
+    // external update before timeout passes
+    window.history.pushState(
+      fakeState,
+      '',
+      'http://localhost/?some=state%20two'
+    );
+
+    // this write isn't needed anymore
+    router.write({ some: 'state two' });
+    await wait(0);
+
+    expect(pushState).toHaveBeenCalledTimes(1);
+    // proves that InstantSearch' write did not happen
+    expect(history.state).toBe(fakeState);
+  });
+
   it('does not write the same url title twice', async () => {
     const title = jest.spyOn(window.document, 'title', 'set');
     const pushState = jest.spyOn(window.history, 'pushState');
--- a/src/lib/routers/__tests__/history.test.ts
+++ b/src/lib/routers/__tests__/history.test.ts
@@ -1,0 +1,22 @@
+import historyRouter from '../history';
+
+const wait = (ms = 0) => new Promise(res => setTimeout(res, ms));
+
+describe('life cycle', () => {
+  it('does not write the same url twice', async () => {
+    const router = historyRouter({
+      writeDelay: 0,
+    });
+
+    const spy = jest.spyOn(window.history, 'pushState');
+
+    router.write({ some: 'state' });
+    await wait(0);
+    router.write({ some: 'state' });
+    await wait(0);
+    router.write({ some: 'state' });
+    await wait(0);
+
+    expect(spy).toHaveBeenCalledTimes(1);
+  });
+});
--- a/src/lib/routers/__tests__/history.test.ts
+++ b/src/lib/routers/__tests__/history.test.ts
@@ -24,6 +24,17 @@ describe('life cycle', () => {
     await wait(0);
 
     expect(pushState).toHaveBeenCalledTimes(1);
+    expect(pushState.mock.calls).toMatchInlineSnapshot(`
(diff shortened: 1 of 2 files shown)


francoischalifour:
Why is it not implemented in the router middleware?

https://github.com/algolia/instantsearch.js/blob/bae6ed2875b03a8a43c576085a19f2ceb43ba1f2/src/middleware/createRouter.ts#L59

Haroenv:
in the router middleware you don't know what the URL to write is, right?

Haroenv:
e2e test is failing, but it seems to resolve just after the assertion, cc @yannickcr

francoischalifour:
@Haroenv You know the `uiState` and the route, which are the source of truth. Solving the problem at this level would solve it for each router implementation.

Haroenv:
Doing an identity function on a string is much simpler than on an object, no? Also you can't know if the _first_ write isn't needed, since that depends on whether the final URL is identical the the one we already have :)

src/lib/routers/history.ts
```diff
@@ -26,6 +26,13 @@ type BrowserHistoryProps = {
   parseURL: ParseURL;
 };
 
+type BrowserHistoryArgs = {
+  windowTitle?: (routeState: RouteState) => string;
+  writeDelay?: number;
+  createURL?: CreateURL;
+  parseURL?: ParseURL;
+};
+
```
yannickcr, line 35: Just above this you have `BrowserHistoryProps` with the almost same shape.
Since it is only used to define the type of the private properties on `BrowserHistory`, you could use `BrowserHistoryArgs` for that and completely get ride of `BrowserHistoryProps`.

You'll just have to mark some properties as required, ex:
```js
  /**
   * Parses the URL into a route state.
   * It should be symetrical to `createURL`.
   */
private readonly parseURL: Required<BrowserHistoryArgs>['parseURL'];
```

yannickcr:
> e2e test is failing, but it seems to resolve just after the assertion

I should have fixed this case in the last `instantsearch-e2e-tests` version, but it was not updated on the project yet (but renovate bot should have submited PR for now, no? :thinking: )

src/lib/routers/history.ts
```diff
@@ -128,9 +128,11 @@ class BrowserHistory implements Router {
     }
 
     this.writeTimer = window.setTimeout(() => {
-      setWindowTitle(title);
+      if (window.location.href !== url) {
```
francoischalifour, line 131: Why don't we wrap the whole inner function?

Haroenv: the timeout still needs to be cleared if the write isn't done :)

francoischalifour: So this would work, right?

```ts
class BrowserHistory implements Router {
  public write(routeState: RouteState): void {
    if (this.writeTimer) {
      window.clearTimeout(this.writeTimer);
    }

    const url = this.createURL(routeState);

    if (window.location.href !== url) {
      this.writeTimer = window.setTimeout(() => {
        const title = this.windowTitle && this.windowTitle(routeState);

        setWindowTitle(title);
        window.history.pushState(routeState, title || '', url);
      }, this.writeDelay);
    }

    this.writeTimer = undefined;
  }
}
```

Haroenv: Hmm, it could, but really you want to check _just before_ you are going to write. An external program could have written to the URL between write start & timeout end

francoischalifour: I see what you mean! Ideally that should be tested too then.

Haroenv:
I think lint isn't running is related to GitHub's downtime, it passes on my end

src/lib/routers/__tests__/history.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    const router = historyRouter({
+      writeDelay: 0,
+    });
+
+    router.write({ some: 'state' });
+    router.write({ some: 'second' });
+    router.write({ some: 'third' });
+    await wait(0);
+
+    expect(pushState).toHaveBeenCalledTimes(1);
```
francoischalifour, line 94: What is the expected history state (or what was `pushState` called with)? We can specify it here.

src/lib/routers/__tests__/history.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    router.write({ some: 'state' });
+    await wait(0);
+
+    router.write({ some: 'state' });
+    await wait(0);
+
+    router.write({ some: 'state' });
+    await wait(0);
+
+    expect(pushState).toHaveBeenCalledTimes(1);
```
francoischalifour, line 26: What is the expected history state (or what was `pushState` called with)? We can specify it here.

PR #4394, opened 2020-04-23
chore(vscode): exclude cjs, dist, es from file explorer

eunjae-lee:
**Summary**

This PR excludes cjs, dist, es folders from file explorer in VSCode.

image

We haven't put any editor-specific config in the repository, but I'm opening this PR to see what you think of it.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,7 @@
+{
+  "search.exclude": {
+    "cjs/**": true,
+    "dist/**": true,
+    "es/**": true
+  }
+}


Haroenv:
does this not follow gitignore already?

eunjae-lee:
> does this not follow gitignore already?

really?? it was not the case for me.
before this config, I saw files from es, cjs, ... but after this config I do not see any more.

eunjae-lee:
weird. I reverted that vscode settings, and now without the config it still works fine. My VSCode was buggy then? I'll close this for now.

Issue #4393, opened 2020-04-22
initialUiState doesn't work with routing


textarea:
**Describe the bug 🐛**
initialUiState setting doesn't work if there is routing setting (not the simple "routing: true" but with seo-friendly urls settings).

**To Reproduce 🔍**
As you can see from the live demo initialUiState query = 'stick' is not  working, if you comment line 31 "routing: searchRouting," initialUiState will works.
Live demo: https://codesandbox.io/s/flamboyant-cache-l6ne6?file=/src/app.js

**Expected behavior 💭**
initialUiState works with routing enabled

**Environment:**
 - OS: mac
 - Browser chrome
 - Version 81

Haroenv:
This is expected behaviour, but something we might consider to change in the future. The reason is that both initial Ui State & routing are providing UI state, so we didn't want to make a complicated merge function between both and deep merges. Thus routing takes precedence if it's present. Relevant code is here:

https://github.com/algolia/instantsearch.js/blob/bae6ed2875b03a8a43c576085a19f2ceb43ba1f2/src/middleware/createRouter.ts#L50-L53

I'll discuss this with the team to see what to do in this case.

For now a workaround would be to fall back to the configure widget instead of initial UI State, this will take precedence over routing.

textarea:
Thanks @Haroenv , 
my hope is to have routing synchronized with the initial filter, but with the configure widget it doesn't.

Haroenv:
An alternative that was just suggested by @francoischalifour is to use the router only to set the initial ui state, like this: https://codesandbox.io/s/interesting-sanderson-euw2n?file=/src/search-routing.js

textarea:
@Haroenv in your demo I don't see the initial "?query=stick" string in the url, only after others filters selection.
It is my same issue.

Haroenv:
So do you want any  time a user visits the URL it to be changed to another one, with the uiState added? You could do that with a custom widget, which has a "getWidgetState" which returns the default, although I wouldn't advice that, as it will not allow people to use the "forward" button in the browser anymore after pressing "back" first.

textarea:
@Haroenv you are right.

PR #4395, opened 2020-04-23
fix(router): merge results of intialUiState with RouteState

Haroenv:
**Summary**

Initial UI State can be used to set a default refinement, e.g. an item in a refinement list to be checked. This can be done in parallel to routing, and should merge per index.

related to #4393, but not fixing it, since they want intial UI State to be _written_ to the router too, which probably shouldn't be done, since that will cause a navigation as an effect of loading

also related to https://github.com/algolia/instantsearch.js/discussions/4455

**Result**

Slightly changed the code so that typescript didn't think the instance was "any"

Before this PR the second new test was failing


Haroenv:
The team doesn't agree whether it makes sense to merge (first-level) between uiState and routeState, and requires you to write custom routing to take intialUiState in account (`search._initialUiState` can be used in `read`)

PR #4390, opened 2020-04-22
chore(ts-migration): renames hits files

nunomaduro:
This pull request renames the files related to hits.

The pull request containing the changes comes later.

PR #4396, opened 2020-04-28
chore(ts-migration): migrates hits contents connector, widget, and stories.

nunomaduro:
This pull request migrates the hits connector, widget, and stories.

As usual, this pull request needs to merged first: https://github.com/algolia/instantsearch.js/pull/4390.

Going to validate myself this work, but you can help making a validation too 👍

### DIFF
--- a/src/connectors/hits/__tests__/connectHits-test.ts
+++ b/src/connectors/hits/__tests__/connectHits-test.ts
@@ -168,7 +168,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
         _highlightResult: {
           foobar: {
             value: `<script>${TAG_PLACEHOLDER.highlightPreTag}foobar${TAG_PLACEHOLDER.highlightPostTag}</script>`,
-            matchLevel: 'none',
+            matchLevel: 'partial',
             matchedWords: ['foobar'],
           },
         },
@@ -192,7 +192,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
         _highlightResult: {
           foobar: {
             value: '&lt;script&gt;<mark>foobar</mark>&lt;/script&gt;',
-            matchLevel: 'none',
+            matchLevel: 'partial',
             matchedWords: ['foobar'],
           },
         },
--- a/src/connectors/hits/__tests__/connectHits-test.ts
+++ b/src/connectors/hits/__tests__/connectHits-test.ts
@@ -4,6 +4,13 @@ import algoliasearchHelper, {
 } from 'algoliasearch-helper';
 import { TAG_PLACEHOLDER } from '../../../lib/escape-highlight';
 import connectHits from '../connectHits';
+import {
+  createInitOptions,
+  createRenderOptions,
+} from '../../../../test/mock/createWidget';
+import { createSearchClient } from '../../../../test/mock/createSearchClient';
+import { createSingleSearchResponse } from '../../../../test/mock/createAPIResponse';
+import { EscapedHits, HitAttributeHighlightResult } from '../../../types';
 
 jest.mock('../../../lib/utils/hits-absolute-position', () => ({
   // The real implementation creates a new array instance, which can cause bugs,
@@ -15,6 +22,7 @@ jest.mock('../../../lib/utils/hits-absolute-position', () => ({
 describe('connectHits', () => {
   it('throws without render function', () => {
     expect(() => {
+      // @ts-ignore
       connectHits()({});
     }).toThrowErrorMatchingInlineSnapshot(`
 "The render function is not valid (received type Undefined).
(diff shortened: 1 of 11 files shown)


src/connectors/hits/connectHits.ts
```diff
@@ -6,56 +6,51 @@ import {
...
+   * Whether to escape HTML tags from hits string values.
+   *
+   * @default true
+   */
+  escapeHTML?: boolean;
+
+  /**
+   * Function to transform the items passed to the templates.
+   */
+  transformItems?: TransformItems<Hit>;
```
nunomaduro, line 40: It's `Hit` because it will already contain `__queryID` and `__position`.

src/connectors/hits/connectHits.ts
```diff
@@ -6,56 +6,51 @@ import {
...
- *     containerNode: $('#custom-hits-container'),
- *   })
- * ]);
- */
-export default function connectHits(renderFn, unmountFn = noop) {
+export type HitsRendererOptions = {
+  /**
+   * The matched hits from Algolia API.
+   */
+  hits: Hits;
```
nunomaduro, line 21: It's `Hit` because it will already contain `__queryID` and `__position`.

src/connectors/hits/connectHits.ts
```diff
@@ -6,56 +6,51 @@ import {
...
+
+  /**
+   * The response from the Algolia API.
+   */
+  results?: SearchResults<AlgoliaHit>;
```
nunomaduro, line 26: It's `AlgoliaHit ` because it will NOT contain `__queryID` and `__position`.

src/lib/escape-highlight.ts
```diff
@@ -61,7 +63,7 @@ export default function escapeHits(hits: Hit[]): Hit[] {
     (hits as any).__escaped = true;
   }
 
-  return hits;
+  return (hits as unknown) as EscapedHits<THit>;
```
nunomaduro, line 66: As discussed before, I could do a type guard here. Keeping it simple for now.

src/types/connector.ts
```diff
@@ -33,7 +33,7 @@ export type RendererOptions<TWidgetParams> = {
   /**
    * The current insights client, if any.
    */
-  insights?: InsightsClient;
+  insights?: InsightsClientWrapper;
```
nunomaduro, line 36: Going to verify this now.. Not sure about this type here.

src/widgets/hits/hits.tsx
```diff
@@ -36,65 +51,83 @@ const renderer = ({ renderState, cssClasses, containerNode, templates }) => (
...
+  empty?: Template;
+
+  /**
+   * Template to use for each result. This template will receive an object containing a single record. The record will
+   * have a new property `__hitIndex` for the position of the record in the list of displayed hits.
+   *
+   * @default ''
+   */
+  item?: Template<
+    Hit & {
```
nunomaduro, line 97: Can you confirm this generic here?

src/connectors/hits/__tests__/connectHits-test.ts
```diff
@@ -145,37 +164,46 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
 
     const hits = [
       {
+        objectID: '1',
         _highlightResult: {
           foobar: {
             value: `<script>${TAG_PLACEHOLDER.highlightPreTag}foobar${TAG_PLACEHOLDER.highlightPostTag}</script>`,
+            matchLevel: 'none',
```
Haroenv, line 171:
```suggestion
            matchLevel: 'partial',
```

src/connectors/hits/connectHits.ts
Haroenv: is there a difference between `Hit` and `Array<Hit>`?

src/connectors/hits/connectHits.ts
Haroenv: is AlgoliaHit not the default for SearchResults?

src/widgets/hits/hits.tsx
```diff
@@ -36,65 +51,83 @@ const renderer = ({ renderState, cssClasses, containerNode, templates }) => (
...
+      __hitIndex: number;
```
Haroenv, line 98: I think it's not __hitIndex, but position, and thus part of Hit

src/widgets/hits/hits.tsx
```diff
@@ -36,65 +51,83 @@ const renderer = ({ renderState, cssClasses, containerNode, templates }) => (
       hits={receivedHits}
       results={results}
       templateProps={renderState.templateProps}
-      insights={insights}
+      insights={insights as InsightsClientWrapper}
```
Haroenv, line 54: how come this isn't a given, since it's a parameter?

src/connectors/hits/connectHits.ts
nunomaduro: It's on the helper, and no. The default is `any`. 🤷🏻‍♂️

src/widgets/hits/hits.tsx
nunomaduro: Actually the component used by this widget adds a __hitIndex: https://github.com/algolia/instantsearch.js/blob/ee6a9c657c97adebba9fb9404eae454c3996b86d/src/components/Hits/Hits.js#L34.

src/widgets/hits/hits.tsx
nunomaduro: It's not important ( as is something internal ). What happens here is: The given `insights` may be undefined. But in reality, is never undefined as the connector always give us a insights instance for the widget. WDYT?

PR #4402, opened 2020-04-29
chore(ts-migration): current refinements contents

nunomaduro:
This pull request adjusts the current refinements contents to the new conventions

### DIFF
--- a/src/components/CurrentRefinements/CurrentRefinements.tsx
+++ b/src/components/CurrentRefinements/CurrentRefinements.tsx
@@ -2,11 +2,14 @@
 
 import { h } from 'preact';
 import { isSpecialClick, capitalize } from '../../lib/utils';
-import { Item } from '../../connectors/current-refinements/connectCurrentRefinements';
+import {
+  CurrentRefinementsConnectorParamsItem,
+  CurrentRefinementsConnectorParamsRefinement,
+} from '../../connectors/current-refinements/connectCurrentRefinements';
 import { CurrentRefinementsComponentCSSClasses } from '../../widgets/current-refinements/current-refinements';
 
 type CurrentRefinementsProps = {
-  items: Item[];
+  items: CurrentRefinementsConnectorParamsItem[];
   cssClasses: CurrentRefinementsComponentCSSClasses;
 };
 
@@ -15,12 +18,7 @@ const createItemKey = ({
   value,
   type,
   operator,
-}: {
-  attribute: string;
-  value: string;
-  type: string;
-  operator?: string;
-}): string =>
+}: CurrentRefinementsConnectorParamsRefinement): string =>
   [attribute, type, value, operator]
     .map(key => key)
     .filter(Boolean)
--- a/src/components/CurrentRefinements/CurrentRefinements.tsx
+++ b/src/components/CurrentRefinements/CurrentRefinements.tsx
@@ -6,7 +6,11 @@ import {
   CurrentRefinementsConnectorParamsItem,
   CurrentRefinementsConnectorParamsRefinement,
 } from '../../connectors/current-refinements/connectCurrentRefinements';
-import { CurrentRefinementsComponentCSSClasses } from '../../widgets/current-refinements/current-refinements';
+import { CurrentRefinementsCSSClasses } from '../../widgets/current-refinements/current-refinements';
+
+export type CurrentRefinementsComponentCSSClasses = {
+  [TClassName in keyof CurrentRefinementsCSSClasses]: string;
+};
 
 type CurrentRefinementsProps = {
   items: CurrentRefinementsConnectorParamsItem[];
--- a/src/connectors/current-refinements/connectCurrentRefinements.ts
+++ b/src/connectors/current-refinements/connectCurrentRefinements.ts
@@ -87,7 +87,7 @@ export type CurrentRefinementsConnectorParamsItem = {
   refinements: CurrentRefinementsConnectorParamsRefinement[];
 
   /**
-   * The function that removes the refinement.
+   * Removes the given refinement and triggers a new search.
    */
   refine(refinement: CurrentRefinementsConnectorParamsRefinement): void;
 };
@@ -121,8 +121,14 @@ export type CurrentRefinementsRendererOptions = {
    */
   items: CurrentRefinementsConnectorParamsItem[];
 
+  /**
+   * Removes the given refinement and triggers a new search.
+   */
   refine(refinement: CurrentRefinementsConnectorParamsRefinement): void;
 
+  /**
+   * Generates a URL for the next state.
+   */
   createURL: CreateURL<CurrentRefinementsConnectorParamsRefinement>;
 };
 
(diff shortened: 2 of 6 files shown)


src/components/CurrentRefinements/CurrentRefinements.tsx
```diff
@@ -15,12 +18,7 @@ const createItemKey = ({
   value,
   type,
   operator,
-}: {
-  attribute: string;
-  value: string;
-  type: string;
-  operator?: string;
-}): string =>
+}: CurrentRefinementsConnectorParamsRefinement): string =>
```
nunomaduro, line 21: Please confirm this.

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -15,45 +15,81 @@ import {
...
+  /**
+   * The attribute on which the refinement is applied.
+   */
   attribute: string;
-  type: string;
+
+  /**
+   * The type of the refinement.
+   */
+  type: CurrentRefinementsConnectorParamsRefinementType;
```
nunomaduro, line 38: This was a string before, please confirm that is indeed one of CurrentRefinementsConnectorParamsRefinementType strings.

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -15,45 +15,81 @@ import {
...
+export type CurrentRefinementsConnectorParamsRefinementType =
+  | 'facet'
+  | 'exclude'
+  | 'disjunctive'
+  | 'hierarchical'
+  | 'numeric'
+  | 'query'
+  | 'tag';
+
+export type CurrentRefinementsConnectorParamsRefinement = {
```
nunomaduro, line 29: Previously we also exported a normalised type, I decided to remove it to make things simple.

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -15,45 +15,81 @@ import {
...
-type ConnectorNumericRefinement = {
-  value: number;
-} & ConnectorRefinement;
-
-export type Item = {
+export type CurrentRefinementsConnectorParamsItem = {
+  /**
+   * The index name.
+   *
+   * @internal
```
nunomaduro, line 70: Is this `indexName` public?

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -297,7 +348,7 @@ function normalizeRefinement(refinement: Refinement): ConnectorRefinement {
       )} ${refinement.name}`
     : refinement.name;
 
-  const normalizedRefinement: ConnectorRefinement = {
+  const normalizedRefinement: CurrentRefinementsConnectorParamsRefinement = {
```
nunomaduro, line 351: Previously we also exported a normalised type, I decided to remove it to make things simple.

src/connectors/current-refinements/connectCurrentRefinements.ts
francoischalifour: Will users have to cast the type as [we're doing in the tests](https://github.com/algolia/instantsearch.js/blob/453eed3d98f3ba3d8ffa345ad3b50bffb4cb0d1a/src/components/CurrentRefinements/__tests__/CurrentRefinements-test.tsx#L324)?

src/connectors/current-refinements/connectCurrentRefinements.ts
francoischalifour: [Yes](https://www.algolia.com/doc/api-reference/widgets/current-refinements/js/#connector-param-render-items).

src/connectors/current-refinements/connectCurrentRefinements.ts
francoischalifour: We use it internally to [create the React key](https://github.com/algolia/instantsearch.js/blob/56fcb8ccb185ce24d63ca8f5c8e5ae193836226b/src/components/CurrentRefinements/CurrentRefinements.tsx#L43). That's useful in a multi-index context—so I think we can make it public.

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -206,7 +257,8 @@ function getItems({
   return items.reduce<any[]>(
```
francoischalifour, line 257: You can pass `items.reduce<CurrentRefinementsConnectorParamsItem[]>` to have it fully typed below (in `filter` and `refine`).

src/widgets/current-refinements/current-refinements.tsx
```diff
@@ -4,86 +4,74 @@ import { h, render } from 'preact';
...
+   * CSS class to add to the delete element.
    */
-  transformItems?: (items: Item[]) => any;
+  delete?: string | string[];
 };
 
-type CurrentRefinementsRendererWidgetParams = {
-  container: HTMLElement;
-  cssClasses: CurrentRefinementsComponentCSSClasses;
+export type CurrentRefinementsComponentCSSClasses = {
```
francoischalifour, line 53: It'd make more sense if this lived in the component file instead (`CurrentRefinements.tsx`).

src/connectors/current-refinements/connectCurrentRefinements.ts
nunomaduro: Well, yes.

At least on the experimental yes, because we are not exporting every single type yet.

nunomaduro: Moving it to `string` meanwhile.

src/connectors/current-refinements/connectCurrentRefinements.ts
nunomaduro: Done.

src/widgets/current-refinements/current-refinements.tsx
nunomaduro: done.

nunomaduro:
@francoischalifour Feedback addressed.

src/connectors/current-refinements/connectCurrentRefinements.ts
yannickcr: It seems that the users have to do some casting at some point or they will get in that case: https://stackoverflow.com/questions/37978528/typescript-type-string-is-not-assignable-to-type since 

If we do not want to weaken the type, the simplest way for them to cast it is to use the `const` assertion ( https://devblogs.microsoft.com/typescript/announcing-typescript-3-4-rc/ ), this way they do not need to type to be exposed.

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -231,34 +267,34 @@ function getItems({
...
   state: SearchParameters,
-  refinement: ItemRefinement
+  refinement: CurrentRefinementsConnectorParamsRefinement
 ): SearchParameters {
   switch (refinement.type) {
     case 'facet':
       return state.removeFacetRefinement(
         refinement.attribute,
-        refinement.value
+        refinement.value as string
```
francoischalifour, line 276: Are these `as string` needed?

PR #4406, opened 2020-05-04
chore(ts-migration): improves build

nunomaduro:
This pull request is the last one of my mission. And it just reverts some work optimistic work done at the beginning that I thought we could deliver 100% this quarter. 

Some important notes:

- I have re-tested all 45% connectors/widgets migrated in this mission using a real instant-search project in typescript with the snippets from the docs.
- API Extractor shows tons of warnings while validating types as they are tons of missing exports, I guess we are going to address them once we have 100% connectors/widgets migrated.
- Running `tsc` with a project that requires `4.4.1-experimental-typescript.0` still shows 5 errors ( 7 prior to this mission ). Would be cool to address that soon.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -207,3 +207,24 @@ yarn run release:prepare
 
 The script will ask you a question about the next version. If it's wrong, you can say "No" and specify the version (e.g. "7.0.0-beta.0"). Then, it will open a pull request for that release. When the pull request is merged, CircleCI will publish it to npm with a `--tag beta` argument.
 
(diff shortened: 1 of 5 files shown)


eunjae-lee:
I think we already have the same script file in `master` branch.
https://github.com/algolia/instantsearch.js/blob/master/scripts/release/build-experimental-typescript.js
Is there anything different?

nunomaduro:
@eunjae-lee Thanks for asking. Yes it already exists. We are just revert them some work done in this branch.

eunjae-lee:
Got it! :)

PR #4411, opened 2020-05-13
chore(release): remove old release process

Haroenv:
**Summary**

Having the old release process still present is confusing

**Result**

Removed yarn script step & the script for `yarn publish`
Other scripts are still used for release-maintenance


package.json
```diff
@@ -42,7 +42,6 @@
     "test:e2e:saucelabs": "wdio wdio.saucelabs.conf.js",
     "test:size": "bundlesize",
     "test:argos": "argos upload functional-tests/screenshots --token $ARGOS_TOKEN || true",
-    "release": "./scripts/release/publish.js",
     "release:maintenance": "./scripts/release/publish-maintenance.sh",
     "release:view-changelog": "conventional-changelog -p angular",
```
eunjae-lee, line 46: We actually can remove these.
When we want to release the old versions like v3, we will checkout v3 and there we'll have publish-maintenance.sh file :)

eunjae-lee: scripts/release/bump-package-version.js
scripts/release/conventional-changelog.js
scripts/release/publish-maintenance.js
scripts/release/publish.js

eunjae-lee:
oh no, we need `build-experimental-typescript.js`. It's used in `ship.config.js`.

PR #4413, opened 2020-05-18
fix(infiniteHits): support cache

eunjae-lee:
**Summary**

This PR adds a cache layer to InfiniteHits.
Related to: https://github.com/algolia/react-instantsearch/pull/2921

`sessionStorage` is brought from the React InstantSearch.

**How to Test**
Example: https://infinitehits-cache-m86tvb8nd.now.sh/
Code: https://codesandbox.io/s/instantsearchjs-3pqv4?file=/src/app.js:469-535


Haroenv:
I think netlify doesn't show up because it's a PR which isn't to master 🤔

eunjae-lee:
Oh I even didn't notice netlify didn't show up.

eunjae-lee:
Yeah, if we had `chore/typescript-migration` here and required status checks, then Netlify would've showed up.

image

Preferably when `chore/typescript-migration` is merged to master, then we can change the target branch to `master` and the Netlify will show up.

image

eunjae-lee:
The scroll wasn't restored. I checked again, and it turns out it's the case only on CodeSandbox.
It works well both on my local machine and on Vercel: https://infinitehits-cache-m86tvb8nd.now.sh/
When I tested on CodeSandbox, when I press back button to come back to the InstantSearch page, there were many CodeSandbox-related things that were loaded. It delayed the load of IS, and resulted no scroll restoration.

So it's ready for review!

Haroenv:
lint failure seems to be in the base branch only

src/widgets/infinite-hits/infinite-hits.tsx
```diff
@@ -21,6 +21,7 @@ import {
   Hit,
   InsightsClientWrapper,
   Renderer,
+  InfiniteHitsCache,
```
francoischalifour, line 24: Would it make sense to define the `InfiniteHitsCache` type in the connector instead?

eunjae-lee: Ah yeah, now I see the pattern.

eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4413/commits/312ddd3b69a2a7849ad5db4316484e7dc2693520

eunjae-lee:
> Nice feature!
> 
> There's no tests of the `infiniteHits` widget using the `cache` option, but only a unit test for `createInfiniteHitsSessionStorageCache`, which is a specific implementation of the cache.
> 
> In order to test the widget in a closer way of a user implementation, can we add a higher-level test for `infiniteHits` with a manual implementation of `cache`?

@francoischalifour thanks for the feedback. What kind of test are you expecting? Tests on widget level? or e2e test?
I didn't particularly add widget-level tests, because anyway I have to mock either sessionStorage or localStorage, then it becomes almost the same as the in-memory cache implementation. 

Maybe I can write a test where I provide a cache to an InfiniteHits instance, and take the cache and create another InfiniteHits instance with the cache to see if it properly renders all the hits without search request. What do you think?

francoischalifour:
What you could do is an integration test that follows these steps:

1. Add `infiniteHits({ cache: customCache })` where the custom cache implements the documented interface
2. Search stuff...
3. Go back to a previous state
4. Make sure that the cache is read

Would that make sense?

eunjae-lee:
@francoischalifour you're talking about writing tests in https://github.com/algolia/instantsearch-e2e-tests

not in src/widgets/infinite-hits/__tests__/infinite-hits-test.ts

right?

francoischalifour:
No, I'm talking about this repo.

Example:

```js
const customCache = { /* ... */ };

search.addWidgets([infiniteHits({ cache: customCache }), pagination()]);

// Interact with the widgets...
// Assertions...
```

yannickcr:
I think @francoischalifour was thinking about an integration test closer to the real usage from a user POV.

Here's an example using testing-libary: 

```js
import { wait, getByLabelText } from '@testing-library/dom';

import instantsearch from '../../../';

import apiResults from './__fixtures__/api-results.json';

describe('infiniteHits', () => {
  let search;
  let searchClient;

  beforeEach(() => {
    searchClient = {
      search: jest.fn(() => Promise.resolve(apiResults)),
    };

    search = instantsearch({
      indexName: 'instant_search',
      searchClient,
    });
  });

  afterEach(() => {
    try {
      search.dispose();
    } catch (e) {
      // Fail if the search was never started, ignoring
    }
  });

  it('must hit cache when navigating to a previously visited page', async () => {
    const infiniteHitsContainer = document.createElement('div');
    const paginationContainer = document.createElement('div');

    const customCache = {
      read: jest.fn(() => {
        // Your custom read implementation
      }),
      write: jest.fn(() => {
        // Your custom write implementation
      }),
    };

    search.addWidgets([
      instantsearch.widgets.infiniteHits({
        container: infiniteHitsContainer,
        cache: customCache,
      }),
      instantsearch.widgets.pagination({
        container: paginationContainer,
        scrollTo: false,
      }),
    ]);

    search.start();

    await wait();

    // Assert that the client was called (page 1 was not cached)
    expect(searchClient.search).toBeCalledTimes(1);

    // Navigate to page 2
    getByLabelText(paginationContainer, 'Next').click();

    await wait();

    // Assert that the client was called (page 2 was not cached)
    expect(searchClient.search).toBeCalledTimes(2);

    // Navigate to page 1
    getByLabelText(paginationContainer, 'Previous').click();

    await wait();

    // Assert that the client was not called (page 1 was cached)
    expect(searchClient.search).toBeCalledTimes(2);
  });
});
```

eunjae-lee:
@yannickcr alright, thanks!
I'll try with that way.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -78,26 +138,30 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
...
         .overrideStateWithoutTriggeringChangeEvent({
           ...helper.state,
-          page: firstReceivedPage - 1,
+          page: getFirstReceivedPage() - 1,
         })
         .search();
     };
     const getShowMore = (helper: Helper): (() => void) => () => {
-      helper.setPage(lastReceivedPage + 1).search();
+      helper.setPage(getLastReceivedPage() + 1).search();
```
Haroenv, line 164: we are calculating this multiple times now, is that expected?

eunjae-lee:
Okay this was quite tough.
It was impossible to test the cache behavior of InfiniteHits with "pagination" widget, because whenever page changes, connectPagination refined the search. And that's not the behavior we want to test.

So I added two integration tests.
The first one is a basic test checking if `read` and `write` methods of the cache are being called correctly.

The second one is more similar to what I manually tested on the example website.
It creates a `instantsearch()` instance and add InfiniteHits. It "loads more" to the second page.
Now we have the hits of the two pages in the cache.

With that cache, we create another `instantsearch()` instance and see if the InfiniteHits widget correctly displays the whole hits from the cache.

What do you think? This took a lot and I'm out of idea 😅

src/connectors/infinite-hits/connectInfiniteHits.ts
eunjae-lee: When a new page is loaded, the `cachedHits` object will be filled with the hits.

```js
        if (cachedHits![page] === undefined) {
          cachedHits![page] = results.hits;
          cache.write({ state, hits: cachedHits });
        }
```

So `getShowMore` function needs to calculate it every time.

src/widgets/infinite-hits/__tests__/infinite-hits-integration-test.ts
```diff
@@ -0,0 +1,122 @@
...
+  let search;
+  let searchClient;
+  let container;
+
+  let cachedState: any;
+  let cachedHits: any;
+  let customCache;
+
+  beforeEach(() => {
+    searchClient = algoliasearch('latency', '6be0576ff61c053d5f9a3225e2a90f76');
```
Haroenv, line 17: a real index doesn't seem like that stable of a source for tests. Maybe creating an array with counters and using a fake client could be better?

eunjae-lee: It makes sense.
https://github.com/algolia/instantsearch.js/pull/4413/commits/64abc2c6463037e586a8ad9bf0ac666fd088101d
I think I did it correctly, right?

eunjae-lee:
@yannickcr @Haroenv I reverted the merge, and now we've got [this PR](https://github.com/algolia/instantsearch.js/pull/4430) to revert it.
But the problem is this PR is still `merged` status. I probably cannot merge this onto master.. any idea?

yannickcr:
Yes you'll have to re-create a PR, but you can re-use this branch for it.

fabien:
@eunjae-lee I'm having trouble with this feature using the latest instantsearch.js on Chrome on Mac (latest) - Safari works fine, even on iOS for me. History within the same page (going back) works as expected though.

- https://infinitehits-cache-m86tvb8nd.now.sh/ - doesn't work in Chrome
- https://infinite-hits-cache-react.now.sh/ - works in all browsers

I tried locally, linking to a local `detail.html` to check if it had anything to do with switching to an external domain (google in the original example). Routing is enabled, so I don't know what's different from the React version.

eunjae-lee:
@fabien We're relying on the browsers to store/restore the scroll position, which might fail due to many reasons. By the way this example(https://infinitehits-cache-m86tvb8nd.now.sh/) works fine in my Chrome (MacOS 10.15.7, Chrome 85.0.4183.102).

Anyway the way you store and restore the scroll position is a good workaround.
Let us know if you need any help!

fabien:
@eunjae-lee thanks for the confirmation, though strange it works for you (my setuP MacOS 10.13.6, Chrome 85.0.4183.121) - looks like I've resolved it as shown, all good now!

PR #4414, opened 2020-05-18
chore: add codesandbox ci

eunjae-lee:
**Summary**

This PR configures CodeSandbox CI app.

related to https://github.com/algolia/react-instantsearch/pull/2927

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,0 +1,5 @@
+{
+  "sandboxes": [
+    "github/algolia/create-instantsearch-app/tree/templates/instantsearch.js"
+  ]
+}
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,5 +1,3 @@
 {
-  "sandboxes": [
-    "github/algolia/create-instantsearch-app/tree/templates/instantsearch.js"
-  ]
+  "sandboxes": ["instantsearchjs-es-template-pcw1k"]
 }


eunjae-lee:
IIUC, it just replaces the dependency with the temporary one. I don't think they use it from script tag according to the sandbox created in the react-IS repo: https://codesandbox.io/s/eloquent-bush-wy34l

francoischalifour:
What @Haroenv means is that [the template uses a script tag](https://github.com/algolia/create-instantsearch-app/blob/a2c0687d16fbb659307e99facccd12fad19f0301/instantsearch.js/index.html#L48) which CodeSandbox might not be able to replace during deployment.

eunjae-lee:
Oops. Closed by mistake.

eunjae-lee:
> What @Haroenv means is that [the template uses a script tag](https://github.com/algolia/create-instantsearch-app/blob/a2c0687d16fbb659307e99facccd12fad19f0301/instantsearch.js/index.html#L48) which CodeSandbox might not be able to replace during deployment.

You're right. What do you think of adding instantsearch.js as a dependency in create-react-app template and `import` it there? Any blocker? If not, I'll send a PR.

Haroenv:
I think it's better to use a different example which imports

francoischalifour:
Yeah I think it's a bit too early to move to imports in the sandbox because we document the UMD usage everywhere in the documentation. We can create an InstantSearch.js ESM template.

eunjae-lee:
Okay sounds good. I guess I can just create a code sandbox which imports es module and use it as the base for the PRs.
Or, do you want to have it in a proper repo like doc-code-samples?

francoischalifour:
That's fine to me!

PR #4415, opened 2020-05-19
fix: ensure lastResults object exists

chrisbradleydev:
**Summary**

This error was surfaced on mobile Safari. Ensure `lastResults` is an object before evaluating object property.

**Result**

This solves the error demonstrated in the included screenshot.

Screen Shot 2020-05-19 at 4 33 45 PM

### DIFF
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -365,8 +365,8 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
           this._isSearchStalled = true;
           this._render(
             this.helper,
-            this.helper.lastResults,
-            this.helper.lastResults._state
+            this.helper && this.helper.lastResults,
+            this.helper.lastResults && this.helper.lastResults._state
           );
         }, this._stalledSearchDelay);
       }
--- a/src/lib/InstantSearch.js
+++ b/src/lib/InstantSearch.js
@@ -360,13 +360,17 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
     this.helper.search();
 
     this.helper.on('search', () => {
-      if (!this._isSearchStalled && !this._searchStalledTimer) {
+      if (
+        !this._isSearchStalled &&
+        !this._searchStalledTimer &&
+        this.helper.lastResults
+      ) {
         this._searchStalledTimer = setTimeout(() => {
           this._isSearchStalled = true;
           this._render(
             this.helper,
-            this.helper && this.helper.lastResults,
-            this.helper.lastResults && this.helper.lastResults._state
+            this.helper.lastResults,
+            this.helper.lastResults._state
           );
         }, this._stalledSearchDelay);
       }


Haroenv:
This is confusing, since `lastResults` gets set right before firing the `search` event. Do you have a way to reproduce this in a sandbox?

ctbradley:
@Haroenv I'll try and reproduce with CodeSandbox.

Haroenv:
I think I see how this can happen, it would be if the very first search is stalled, since this is the `search` event, not the `result` event

https://github.com/algolia/algoliasearch-helper-js/blob/67407a0dfd99402bc1a77bd005385633c3881624/src/algoliasearch.helper.js#L1195-L1198

Would need to check if we already have a test for this though, I feel like not having results woful be a straightforward way to write this test.

We probably want to render `null` or something for the results, and not with && like here, since that will return a different object

src/lib/InstantSearch.js
```diff
@@ -365,8 +365,8 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
           this._isSearchStalled = true;
           this._render(
             this.helper,
-            this.helper.lastResults,
-            this.helper.lastResults._state
+            this.helper && this.helper.lastResults,
```
Haroenv, line 368: I think this line here isn't needed, helper is always  going to be defined since we call methods on it

src/lib/InstantSearch.js
```diff
@@ -365,8 +365,8 @@ See: https://www.algolia.com/doc/guides/building-search-ui/widgets/create-your-o
...
+            this.helper.lastResults && this.helper.lastResults._state
```
Haroenv, line 369: however the state maybe could be local if there's no last results? Something like 

```suggestion
            this.helper.lastResults ? this.helper.lastResults._state : this.helper.state
```

ctbradley: Your suggestion produces this error when I try to navigate to another page.

Screen Shot 2020-05-20 at 9 15 59 AM

Haroenv: I think that error is related to the line above more than this one. Maybe we should find a way to create an empty SearchResults and use that to render when there's not already a lastResults. (results is null, not state). 

I'd love to try out what you have though, I have not seen those errors before, so I think something else is going on as well

ctbradley: I'm working on the sandbox now.

ctbradley: @Haroenv I haven't been able to reproduce this error in a sandbox. I setup the project to have the same exact implementation of Algolia / vue-instantsearch and I cannot surface this error. However I can assure you this error has happened thousands of times in our production application. How should I proceed to get this solved?

gwardwell: @Haroenv jumping in because @ctbradley is on my team and the error is happening in our product.

We've seen 7,757 occurrences of this error in our application. Replicating in a sandbox doesn't seem to surface the problem, but we also can't seem to identify a difference that would allow us to change our production application to solve the problem. Any advice you could give up on next steps would be fantastic.

yannickcr:
Even if the true reasons of the problem are undetermined, if this PR is fixing your problem and since the change is harmless, I'm for merging it.

We'll release a maintenance version of `instantsearch.js` v3 with this fix.

However, if you find more information on how is case occurs on you production application please share it with us :slightly_smiling_face:

Haroenv:
Thinking more about this, while I do like the initiative here, I think `lastResults` will still be `null` in problematic cases here. If this actually fixes your issue, so be it, but I don't think it's fully fixed yet

ctbradley:
@Haroenv I agree with your point. This measure is only a band-aid on the actual issue. Something is causing `lastResults` to be null. Unfortunately I was unable to reproduce this in a sandbox.

Haroenv:
in v4 we have this solved as `if (!this._isSearchStalled && !this._searchStalledTimer) {` is changed to `if (!this._isSearchStalled && !this._searchStalledTimer && this.helper.lastResults) {` and leaving the rest as-is. This will mean the render will no longer happen and can't cause errors

ctbradley:
Are you suggesting this solution for v3?

Haroenv:
yes, I'm suggesting this solution for the current branch on which you are

ctbradley:
Refactored with your suggestion.

devdreamsolution (approved):

Celtian:
@Haroenv Are you goint to merge into v3?

Haroenv:
Sorry, v3 has been released years ago, and this issue is fixed in v3. Sorry to @chrisbradleydev for losing track of this PR, but I don't think it still makes sense to release a v3 version now. Is there a reason you didn't yet update to v4? It's much smaller and has many more features, minimal breaking changes!

PR #4418, opened 2020-05-25
fix(rangeInput): fix a bug that floating number cannot be typed

eunjae-lee:
## Summary

This PR fixes a regression which disabled floating numbers in rangeInput widget.

**Before this PR**
```js
  onInput = name => event => {
    this.setState({
      [name]: Number(event.currentTarget.value),
    });
  };
```

When user wants to type "10.34", that callback will be fired with the following `event.currentTarget.value` values:

- `1`
- `10`
- `10.`
- `10.3`
- `10.34`

When it had "10.", `Number("10.")` returns `10` which makes the input value back to `10`. So people never can get to "10.3".

So I removed `Number()` function there (which was the case before the regression). And I used `Number()` when refining.

* [The working example from this branch](https://codesandbox.io/s/instantsearchjs-0d34s?file=/src/app.js:382-470)
* [The failing example from master(v4.5.0)](https://codesandbox.io/s/instantsearchjs-zg1ri?file=/src/app.js)

**Result**

User should be able to type floating numbers.

### DIFF
--- a/src/components/RangeInput/RangeInput.js
+++ b/src/components/RangeInput/RangeInput.js
@@ -24,14 +24,17 @@ class RangeInput extends Component {
 
   onInput = name => event => {
     this.setState({
-      [name]: Number(event.currentTarget.value),
+      [name]: event.currentTarget.value,
     });
   };
 
   onSubmit = event => {
     event.preventDefault();
 
-    this.props.refine([this.state.min, this.state.max]);
+    this.props.refine([
+      this.state.min && Number(this.state.min),
+      this.state.max && Number(this.state.max),
+    ]);
   };
 
   render() {


francoischalifour:
Nice! It would be valuable to add a test for this to avoid future regressions. You can draw inspiration from the [`SearchBox` events tests](https://github.com/algolia/instantsearch.js/blob/8195d6e92f20292229d46973b06ee55fe5bdea5d/src/components/SearchBox/__tests__/SearchBox-test.js#L186).

eunjae-lee:
@francoischalifour Yesterday I tried to add a test for that, but it was hard to simulate user input like in real world.
In real world, we type and it adds each character at a time to input, but with testing-library, it just triggered an event with a value like '10.34'.

Today I tried their [user-event](https://testing-library.com/docs/ecosystem-user-event) package which is supposed to give a better simulation.
I ran

```js
await userEvent.type(minInput, '10.34');
```

but it triggered the following 5 events.

- 1
- 10
- 10.
- 10.3
- 10.34

So this obviously cannot cover this regression. I think it could be only possible with real e2e test like selenium or cypress.

francoischalifour:
Is it not possible to tell the testing library to type `10.` and to expect that the input is effectively `10.`, and not `0`?

eunjae-lee:
It turns out it's not that easy 😅
https://github.com/testing-library/react-testing-library/issues/678

francoischalifour:
Right, let's merge this as-is then.

Issue #2182, opened 2017-06-05
Allow ToggleWidgetValues to accept multiple values


timkelty:
**Do you want to request a *feature* or report a *bug*?**
*feature*

**Feature: What is your use case for such a feature?**
I have an attribute I'd like to use a toggle widget for, but I'd like to have the "on" value be one of many options.

**Feature: What is your proposed API entry? The new option to add? What is the behavior?**
I'd like to use it like this:

```
search.addWidget(instantsearch.widgets.toggle({
  container: '#toggle',
  attributeName: `hasMedia`,
  on: [true, 'true', 1, '1'],
}));
```

An alternate implementation might allow you to pass a fn with an argument of the value of the specified attr, returning a bool:

```
search.addWidget(instantsearch.widgets.toggle({
  container: '#toggle',
  attributeName: `hasMedia`,
  on: (value) => {
    return value && value !== 'off';
  },
}));
```

**What is the version you are using? Always use the latest one before opening a bug issue.**
1.11.12/2.beta-5

iam4x:
I will discuss about implementation with @bobylito but I'm pretty sure it's a good thing that we can add into the v2 without a breaking changes for older implementations 👍

timkelty:
@iam4x that's what I was hoping, thanks!

bobylito:
Hi @timkelty how is the component supposed to figure out which value to select? Or will it add all the filters at once when `on`?

timkelty:
@bobylito right - so in the array example, any one of those values would be considered "on".

In the fn example, it would only be on if you returned `true`.

bobylito:
So if I understand correctly you have different values that are "on" in a single attribute of the dataset, right? Wouldn't it be easier to have a single value? Can you elaborate on the use case maybe?

timkelty:
@bobylito It would be simpler yes, and my use-case is actually kind of dumb...basically my indexer can run out of my database or directly from the request. If it populates from the database, certain values get "minimized", e.g. `true`/`false` to `1`/`0`.

_Anyway_ - while my use case isn't great, because I could just hack into my indexing stuff and fix that, I thought it could be useful other ways.

For example - Consider a "Has Images" toggle. To do this, you'd have to index a new `hasImages` bool attribute.

However, if you accept a fn as I suggest, you could do something like (with `images` an array of your images):

```
search.addWidget(instantsearch.widgets.toggle({
  container: '#toggle',
  attributeName: `images`,
  on: (value) => {
[...]

PR #4420, opened 2020-05-28
feat(connectToggleRefinement): add support for array values

Kocal:
Hi :wave: 

This PR is a proposal for #2182 which add support for multiples values on `on`/`off` for the toggle refinement connector.

See #2182 for more information.

**Result**

It allow us to use the toggle refinement connector with multiple values:
```js
connectToggleRefinement()({
  attribute: 'whatever',
  on: ['my', 'values'],
  off: ['something', 'else']
});
```

and the toggle refinement widget:
```js
search.addWidget(instantsearch.widgets.toggleRefinement({
  container: '#toggle',
  attribute: 'whatever',
  on: ['my', 'values'],
  off: ['something', 'else']
}));
```

On [the playground](https://codesandbox.io/s/instantsearchjs-y5pyp), I've used:
```js
  toggleRefinement({
    container: '#toggle',
    attribute: 'brand',
    on: ['Sony', 'Incipio'],
  }),
```

and you can see it toggle both values.

What do you think?
Thanks!

### DIFF
--- a/src/connectors/toggle-refinement/connectToggleRefinement.js
+++ b/src/connectors/toggle-refinement/connectToggleRefinement.js
@@ -270,7 +270,7 @@ export default function connectToggleRefinement(renderFn, unmountFn = noop) {
 
       getWidgetState(uiState, { searchParameters }) {
         const isRefined =
-          on && on.every(v => state.isDisjunctiveFacetRefined(attribute, v));
+          on && on.every(v => searchParameters.isDisjunctiveFacetRefined(attribute, v));
 
         if (!isRefined) {
           return uiState;
--- a/src/connectors/toggle-refinement/connectToggleRefinement.js
+++ b/src/connectors/toggle-refinement/connectToggleRefinement.js
@@ -269,10 +269,8 @@ export default function connectToggleRefinement(renderFn, unmountFn = noop) {
       },
 
       getWidgetState(uiState, { searchParameters }) {
-        const isRefined = searchParameters.isDisjunctiveFacetRefined(
-          attribute,
-          on
-        );
+        const isRefined =
+          on && on.every(v => state.isDisjunctiveFacetRefined(attribute, v));
 
         if (!isRefined) {
           return uiState;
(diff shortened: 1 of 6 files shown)


src/connectors/toggle-refinement/connectToggleRefinement.js
```diff
@@ -185,7 +206,9 @@ export default function connectToggleRefinement(renderFn, unmountFn = noop) {
       },
 
       render({ helper, results, state, instantSearchInstance }) {
-        const isRefined = helper.state.isDisjunctiveFacetRefined(attribute, on);
+        const isRefined =
```
Haroenv, line 209: similar to this, we also need to edit it in `getWidgetState`: ?instant_search%5BrefinementList%5D%5Bbrand%5D%5B0%5D=Sony&instant_search%5BrefinementList%5D%5Bbrand%5D%5B1%5D=Incipio

Haroenv:
Relevant sandbox: https://codesandbox.io/s/instantsearchjs-mp4zs?file=/src/app.js

src/connectors/toggle-refinement/connectToggleRefinement.js
Kocal: Oh, thank you!

eunjae-lee:
> Relevant sandbox: [codesandbox.io/s/instantsearchjs-mp4zs?file=/src/app.js](https://codesandbox.io/s/instantsearchjs-mp4zs?file=/src/app.js)

I clicked "branch" and "Metra", so got this url:

https://mp4zs.csb.app/?instant_search%5BrefinementList%5D%5Bbrand%5D%5B0%5D=Sony&instant_search%5BrefinementList%5D%5Bbrand%5D%5B1%5D=Incipio&instant_search%5BrefinementList%5D%5Bbrand%5D%5B2%5D=Metra&instant_search%5Btoggle%5D%5Bbrand%5D=true

but opening the url won't get Metra selected.

Haroenv:
That's probably since they're on the same attribute, which isn't something we officially support I think. If the widgets are mounted in the other order, the result would be correct. I don't think we can really find a solution for that, except if toggle refinement doesn't remove other refinements in `getWidgetSearchParameters`

eunjae-lee:
You're right. I changed the order of the widgets, and the result is correct. Since it's a "toggle" refinement, I guess the current behavior probably makes sense. We can ship it and later get feedback.

PR #4423, opened 2020-06-03
fix(pagination): isLastPage when there's no pages

Haroenv:
fixes https://github.com/algolia/vue-instantsearch/issues/785

**Summary**

This is discovered by @dotboris. When there are no pages, this.total === 0, so this.currentPage - 1 will be -1 instead of 0, and falsely indicate it's not the last page.

**Result**

isLastPage is true when there's no results.

same as https://github.com/algolia/instantsearch.js/pull/4422, but for **v3** (Vue InstantSearch is using that before our next major version)

### DIFF
--- a/src/connectors/pagination/Paginator.js
+++ b/src/connectors/pagination/Paginator.js
@@ -48,7 +48,7 @@ class Paginator {
   }
 
   isLastPage() {
-    return this.currentPage === this.total - 1;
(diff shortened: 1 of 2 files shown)


Haroenv:
e2e tests on maintenance are flaky still apparently. I think they can be ignored

PR #4428, opened 2020-06-08
chore: improvements to typescript migration

Haroenv:
**Summary**

- correct entry file
- type  createHTMLMarker
- utils: prevent circular import

**Result**

some less warnings

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "scripts/transforms"
   ],
   "dependencies": {
-    "@types/googlemaps": "3.39.6",
+    "@types/googlemaps": "^3.39.6",
     "algoliasearch-helper": "^3.1.1",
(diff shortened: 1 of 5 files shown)


Haroenv:
I added the dependency, but other @types dependencies aren't added as "dependency", not sure what the best is

package.json
```diff
@@ -52,6 +52,7 @@
     "scripts/transforms"
   ],
   "dependencies": {
+    "@types/googlemaps": "3.39.6",
```
francoischalifour, line 55: Could this be added to `devDependencies` like the other `@types`?

Either way, we do not pin `@types` deps in this project.

Haroenv: google maps is a dependency though, not dev dep, same with hogan.js, classnames, qs

PR #4433, opened 2020-06-12
refactor(index): remove unnecessary `page` UI state reset branch

francoischalifour:
In #4359, we introduced an unnecessary code branch that never gets executed (and is not covered by any tests).

We reset the `page` `localUiState` but then reassign the whole `localUiState`.


francoischalifour:
@Haroenv Nope, that's because they're flaky. I reran them.

Issue #4409, opened 2020-05-11
onStateChange does not receive page update from infinite hits


Haroenv:
## 🔍 Bug reproduction

**Steps to reproduce the behavior:**

1. Go to https://codesandbox.io/s/dark-river-27qdw?file=/src/app.js:229-242
2. Click on "show more"
4. See console.log onStateChange was called
5. see that page did not change

**Live reproduction:**

https://codesandbox.io/s/dark-river-27qdw?file=/src/app.js:229-242

## 💭 Expected behavior

page should change, like it does when you don't use `onStateChange`

## Environment

- Version: 4.4.0

giojkd:
Cannot to make onStateChange at all

`
        const search = instantsearch({
            indexName: 'myindex',
            searchClient,
            initialUiState: {
                myindex: {
                    query: keyword,
                    refinementList:presetFacets,
                    toggle:{
                        is_on_sale: onlyOnSale
                    }
                },
            },
            onStateChange({ uiState, setUiState }) {
                console.log(uiState);
                setUiState(uiState);
            },
        });`

Any idea ?

Haroenv:
can you open a new issue with reproduction in a sandbox please? I'm not sure what is meant to happen with just the InstantSearch constructor @giojkd

logan-jobzmall:
This is still happening unfortunately.

Haroenv:
@logan-jobzmall, do you have a reproduction? The original sandbox works as expected as far as I can tell

PR #4435, opened 2020-06-15
fix(infiniteHits): work with controlled mode

eunjae-lee:
## Summary

fixes https://github.com/algolia/instantsearch.js/issues/4409

This PR fixes some bugs around controlled mode.

### showPrevious + onStateChange
https://codesandbox.io/s/instantsearchjs-xr81o?file=/src/app.js
load some pages, refresh, and click "Show previous". It won't work correctly.

### onStateChange without pagination
https://codesandbox.io/s/instantsearchjs-x713i?file=/src/app.js
cannot load more page.

## Examples
- ✅ [showPrevious + onStateChange](https://codesandbox.io/s/instantsearchjs-pb4hc?file=/src/app.js)
- ✅ [showPrevious](https://codesandbox.io/s/instantsearchjs-ykjr9?file=/src/app.js)
- ✅ [onStateChange](https://codesandbox.io/s/instantsearchjs-737yh?file=/src/app.js)
- ✅ [normal](https://codesandbox.io/s/instantsearchjs-fhz1e)
- ✅ [onStateChange without pagination](https://codesandbox.io/s/instantsearchjs-yi2j1)

## Unresolved issue

With the example `showPrevious + onStateChange` and `showPrevious`,
- load some pages
- refresh
- click "show previous"
- then the `page` param in the url stays the same,
- but the pagination changes.

I've found out that this is not an issue from this PR, and exists in the stable version, too.
https://codesandbox.io/s/instantsearchjs-h20yo?file=/src/app.js

### DIFF
--- a/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
+++ b/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
@@ -191,6 +191,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     const helper = algoliasearchHelper({} as SearchClient, '', {});
     helper.setPage(1);
     helper.search = jest.fn();
+    (helper as any).searchWithoutTriggeringOnStateChange = jest.fn();
     helper.emit = jest.fn();
 
     widget.init!(
@@ -227,7 +228,10 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     showPrevious();
     expect(helper.state.page).toBe(0);
     expect(helper.emit).not.toHaveBeenCalled();
-    expect(helper.search).toHaveBeenCalledTimes(1);
+    expect(helper.search).toHaveBeenCalledTimes(0);
+    expect(
+      (helper as any).searchWithoutTriggeringOnStateChange
+    ).toHaveBeenCalledTimes(1);
 
     // the results should be prepended if there is an decrement in page
     const previousHits = [
--- a/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
+++ b/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
@@ -793,7 +793,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
   });
 
   describe('getWidgetState', () => {
-    test('returns the `uiState` empty without `showPrevious` option', () => {
+    test('returns the `uiState` with `page` when `showPrevious` not given', () => {
       const render = jest.fn();
       const makeWidget = connectInfiniteHits(render);
       const helper = algoliasearchHelper(createSearchClient(), 'indexName', {
@@ -809,10 +809,12 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
         }
       );
 
-      expect(actual).toEqual({});
+      expect(actual).toEqual({
+        page: 2,
+      });
     });
 
-    test('returns the `uiState` empty with `showPrevious` option on first page', () => {
+    test('returns the `uiState` with `page` when `showPrevious` given on first page', () => {
       const render = jest.fn();
       const makeWidget = connectInfiniteHits(render);
       const helper = algoliasearchHelper(createSearchClient(), 'indexName', {
@@ -830,7 +832,9 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
         }
       );
 
-      expect(actual).toEqual({});
+      expect(actual).toEqual({
+        page: 1,
+      });
     });
 
     test('returns the `uiState` containing `page` with `showPrevious` option', () => {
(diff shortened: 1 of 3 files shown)


eunjae-lee:
- ✅ [showPrevious + onStateChange](https://codesandbox.io/s/instantsearchjs-tj8iu?file=/src/app.js)
- ✅ [showPrevious](https://codesandbox.io/s/instantsearchjs-tk7ww?file=/src/app.js)

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -158,7 +157,7 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
           ...helper.state,
           page: getFirstReceivedPage() - 1,
         })
-        .search();
+        .searchWithoutTriggeringOnStateChange();
```
eunjae-lee, line 160: It searches without triggering `onStateChange` on InstantSearch instance, because this is meant to be silent and to not modify the state.

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -805,10 +809,12 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
         }
       );
 
-      expect(actual).toEqual({});
+      expect(actual).toEqual({
```
eunjae-lee, line 812: If `uiState` is empty, then `getWidgetSearchParameters` doesn't get the `page` and won't be able to refine the search correctly.

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -303,11 +302,10 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
       },
 
       getWidgetState(uiState, { searchParameters }) {
-        const page = searchParameters.page || 0;
-
-        if (!hasShowPrevious || !page) {
-          return uiState;
-        }
+        const page = Math.max(
```
eunjae-lee, line 305: It used to just return `uiState` (possibly without `page` in it) and let `getWidgetSearchParameters` to use `0` when `page` is missing.
```js
const page = uiState.page ? uiState.page - 1 : 0;
```

It can be incorrect, so now it always return `uiState` with `page`. And it gets the max number between the two, because when the current page is 4, user can click "showPrevious" which gives us smaller page, but we shouldn't change the page to that.

src/widgets/index/index.ts
```diff
@@ -365,6 +365,10 @@ const index = (props: IndexProps): Index => {
         return mainHelper.search();
       };
 
+      (helper as any).searchWithoutTriggeringOnStateChange = () => {
```
eunjae-lee, line 368: had to cast as any, because it doesn't exist in the type def.

eunjae-lee:
> There's however a bug when returning to the first page:

Nice catch!
I only thought about receiving previous page by "show previous" button.
I should fix it somehow.

> In a more generic way, the pagination doesn't work as expected now. If you select page 3, you've got 60 results. If you come back to page 2, you still get 60 results.

Yeah, this is a weird behavior when we use infiniteHits + pagination.
I guess, it would be also weird if we get rid of the last 20 results when coming back to page 2. (we could scroll, though)

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -305,10 +304,6 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
       getWidgetState(uiState, { searchParameters }) {
         const page = searchParameters.page || 0;
 
-        if (!hasShowPrevious || !page) {
```
francoischalifour, line 308: We used to return `uiState` when page is `0` because otherwise `page=1` shows up in the UI state and in the routing, whereas it's the default value. Therefore we should keep this branch with only the `!page` condition.

francoischalifour, line 308: Why did we need to remove the `!hasShowPrevious` condition?

eunjae-lee: If we don't turn on `showPrevious`, `getWidgetState` always returns the uiState in this branch without setting any `page` key. And if there is no pagination widget in the widget tree, then uiState never get to have `page`. So "show more" didn't work. (example: https://codesandbox.io/s/instantsearchjs-dmuu2?file=/src/app.js)

eunjae-lee: Thanks for the context. I didn't know that. I tested it and confirmed the behavior you said. So I put it back.

francoischalifour: We seem to want to return `uiState` when the page comes from `showPrevious` and that it's lower than the current `uiState.page`.

The condition would look like this `!hasShowPrevious && page <= uiState.page || !page`.

Wdyt?

eunjae-lee: Hey I tried your suggestion, but realized that in that function `uiState.page` is always undefined and that's why I removed that condition in the first place.

eunjae-lee:
Thanks for the review!

PR #4408, opened 2020-05-05
chore: migrates 45% connectors/widgets to TypeScript

nunomaduro:
Hey team,

This pull request aggregates all the stacked pull requests already tests/validated in isolation that migrate 45% connectors/widgets to TypeScript.

Remember, we are now targeting the **master**.

You can check the remaining state of this mission here: https://github.com/algolia/instantsearch.js/discussions/4407.

Please would be cool if you could checkout this branch locally, and test a typed built with your current projects with specific focus on: 

- breadcrumb
- clear-refinements
- configure
- confirure-related-items
- current-refinements
- hits
- hits-per-page
- infinive-hits
- panel
- autocomplete

PR #4436, opened 2020-06-15
fix(deps): move @types/googlemaps to devDependencies

yannickcr:
**Summary**

In https://github.com/algolia/instantsearch.js/pull/4408 `@types/googlemaps` was added as a dependency while it should have been added as a devDependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "scripts/transforms"
   ],
   "dependencies": {
-    "@types/googlemaps": "^3.39.6",
     "algoliasearch-helper": "^3.1.1",
     "classnames": "^2.2.5",
     "events": "^1.1.0",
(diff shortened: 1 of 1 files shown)


Haroenv:
Why? I think all the @types of dependencies should be dependencies or similar, no?

yannickcr:
Unless the package consumer needs the type when using `instantsearch.js` it should be in the devDependencies ([link](https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies)). I don't think this is the case for `@types/googlemaps`. Also all our other **@types** packages are in devDependencies.

Haroenv:
I think the other ones (classname & qs) should instead both be moved to dependencies instead, since consumers need to install it themselves or else will get errors, no? https://stackoverflow.com/a/46011417/3185307

yannickcr:
I'll make some additional tests to see what kind of errors we can currently encounter.

devdreamsolution (approved):

PR #4443, opened 2020-07-09
feat(connector): add `refineLater`

eunjae-lee:
**Summary**

**Result**

### DIFF
--- a/src/connectors/configure/connectConfigure.ts
+++ b/src/connectors/configure/connectConfigure.ts
@@ -29,7 +29,7 @@ export type ConfigureRendererOptions = {
    * Refine the given search parameters.
    */
   refine: Refine;
-  setSearchParametersWithoutSearch: Refine;
+  refineLater: Refine;
 };
 
 const withUsage = createDocumentationMessageGenerator({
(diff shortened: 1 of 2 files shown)


src/connectors/configure/connectConfigure.ts
```diff
@@ -116,6 +133,7 @@ const connectConfigure: ConfigureConnector = function connectConfigure(
         renderFn(
           {
             refine: connectorState.refine!,
+            setSearchParametersWithoutSearch: connectorState.setSearchParametersWithoutSearch!,
```
Haroenv, line 136: `refineWithoutApplying`?

eunjae-lee: It's not refining at the moment, so `refineLater` sounds more appropriate than `refineWithoutApplying`. And `draft` sounds like there should be `commit` or `save`, etc.
I struggled the name 😂 and so far I love `refineLater`.

francoischalifour: What about something like `batchRefine`?

`refineLater` doesn't give a clue as to when it's going to refine.

eunjae-lee: IMO, `batch` feels like there should be a dedicated trigger for that.
Or, what if we use terms like `schedule` or `defer`?
As in `scheduleRender`, ...

eunjae-lee:
I created this branch as a stepping stone for https://github.com/algolia/instantsearch.js/pull/4446
and I guess we can use helper directly there.
So I'll close this. If we need this functionality somewhere else, then we can bring this back.

PR #4444, opened 2020-07-09
chore: update buildCommand for CodeSandbox CI

eunjae-lee:
**Summary**

This PR updates the CodeSandbox CI configuration.
It now runs `yarn run build`, but if it runs only `yarn run build:es` it can reduce the build time.

**Result**

Same, but faster.


Haroenv:
sorry, only approve when the `abc` file is removed

eunjae-lee:
Yeah, 

image


this comment didn't show up before, so added a dummy comment to see if we still get the comments from the ci. Now it's gone.

PR #4446, opened 2020-07-15
feat(insights): introduce `insights` middleware (1/4)

eunjae-lee:
## Summary

This PR introduces `insights` middleware.
It is to automatically setup userToken across search calls and Insights api calls.
Once this middleware is added to `instantSearchInstance`, when user calls

```js
window.aa('setUserToken', userToken);
// or 
insightsClient('setUserToken', userToken);
```

the middleware will automatically apply the userToken to the searchParameters. Then the next search call will include this userToken.

**Stacked PRs**
1. [introduce `insights` middleware](https://github.com/algolia/instantsearch.js/pull/4446) ← current
2. [send events from hits and refinementList](https://github.com/algolia/instantsearch.js/pull/4456)
4. [add tests](https://github.com/algolia/instantsearch.js/pull/4464)

**examples**
- basic example: https://codesandbox.io/s/instantsearchjs-jetcz?file=/src/app.js
- with search-insights **umd**: https://codesandbox.io/s/instantsearchjs-zn81s?file=/src/app.js
- `insightsClient: false`: https://codesandbox.io/s/instantsearchjs-1igbb?file=/src/app.js
(_working example_: [esm](https://codesandbox.io/s/instantsearchjs-5qwf4?file=/src/app.js), [umd]())


src/lib/InstantSearch.ts
```diff
@@ -243,6 +244,39 @@ See ${createDocumentationLink({
...
+      ({ refineLater }, isFirstRendering) => {
+        if (!isFirstRendering) {
+          return;
+        }
+
+        (this.insightsClient as Function)(
+          'onSetUserToken',
+          userToken => {
+            console.log('setting userToken!', userToken);
+            refineLater({ userToken });
```
Haroenv, line 271: I actually feel lie this should be done with `helper.setState` or `setSearchParameter` instead of the overhead of configure (and therefore also avoids us creating this new refineLater option :)

eunjae-lee: Yeah it makes sense. Now that we have only one usage of `refineLater` here, we could just not add the option to the connector and later consider adding one if needed somewhere else.

src/lib/InstantSearch.ts
```diff
@@ -1,6 +1,7 @@
 import algoliasearchHelper, { AlgoliaSearchHelper } from 'algoliasearch-helper';
 import EventEmitter from 'events';
 import index, { Index, isIndexWidget } from '../widgets/index/index';
+import connectConfigure from '../connectors/configure/connectConfigure';
```
Haroenv, line 4:
```suggestion
```

src/lib/InstantSearch.ts
```diff
@@ -243,6 +245,40 @@ See ${createDocumentationLink({
...
+    const setUpdater = (callback: (userToken: string) => void) => {
+      (this.insightsClient as Function)('onSetUserToken', callback, {
+        immediate: true,
+      });
+    };
+
+    const customWidgetForUserToken = {
+      init({ helper }) {
+        setUpdater(userToken => {
+          const nextSearchParameters = mergeSearchParameters(
```
Haroenv, line 271: there's no need for this, you can do directly `helper.setQueryParameter('userToken', userToken)` (even without widget IMHO)

eunjae-lee: Ah right, we already have helper in the instance. I should use `this.helper` instead of `this.mainHelper`. Am I correct?

Haroenv: Actually now that I think of it, it might be best to do it in a (fake) widget after all, but in the `getWidgetSearchParameters` part (as well as updating) to make sure it gets propagated to multi-index etc.

eunjae-lee: I tried with `this.helper.setQueryParameter()` and it seems to work correctly with multi-index situations.

eunjae-lee:
https://codesandbox.io/s/instantsearchjs-051n8?file=/package.json


Kapture 2020-07-16 at 14 45 42

src/middleware/insights.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export type InsightsProps = {
+  insightsClient: InsightsClient;
+};
+
+export type CreateInsightsMiddleware = (props: InsightsProps) => Middleware;
+
+export const createInsightsMiddleware: CreateInsightsMiddleware = props => {
+  const { insightsClient } = props;
+
```
Haroenv, line 12:
```suggestion
  if (!insightsClient) {
    throw new Error(
      'passing insightsClient to instantsearch is required for insightsMiddleware',
    );
  }
```

src/lib/main.ts
```diff
@@ -39,5 +40,6 @@ instantsearch.createInfiniteHitsSessionStorageCache = createInfiniteHitsSessionS
 instantsearch.highlight = helpers.highlight;
 instantsearch.snippet = helpers.snippet;
 instantsearch.insights = helpers.insights;
+instantsearch.middleware = middleware;
```
eunjae-lee, line 43: middleware vs middlewares

src/middleware/insights.ts
```diff
@@ -0,0 +1,36 @@
...
+          userToken => {
+            instantSearchInstance.mainIndex
+              .getHelper()
+              .setQueryParameter('userToken', userToken);
+          },
+          { immediate: true }
+        );
+      },
+      unsubscribe() {
+        insightsClient('onUserTokenChange', () => {});
```
francoischalifour, line 36: We should pass `undefined` to prevent the callback from being called [according to the Search Insights implementation](https://github.com/algolia/search-insights.js/blob/ddca3535e657b7fed6adea7eed88eb4f11a5675d/lib/_cookieUtils.ts#L51-L53):

```suggestion
        insightsClient('onUserTokenChange', undefined);
```

eunjae-lee: Nice catch, and it makes me realize there is a bug in search-insights. (passing `undefined` there can make it throw an exception [here](https://github.com/algolia/search-insights.js/blob/ddca3535e657b7fed6adea7eed88eb4f11a5675d/lib/_cookieUtils.ts#L72))

src/types/insights.ts
```diff
@@ -10,11 +10,21 @@ export type InsightsClientPayload = {
...
-export type InsightsClient = (
+export type InsightsSendEvent = (
   method: InsightsClientMethod,
   payload: InsightsClientPayload
 ) => void;
 
+export type InsightsOnUserTokenChangeMethod = 'onUserTokenChange';
+
+export type InsightsOnUserTokenChange = (
+  method: InsightsOnUserTokenChangeMethod,
```
Haroenv, line 21:
```suggestion
export type InsightsOnUserTokenChange = (
  method: 'onUserTokenChange',
```

why is the method name exported separately?

eunjae-lee: Actually no reason. I'll put it directly into the function type def.

src/middleware/insights.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export type InsightsProps = {
+  insightsClient: InsightsClient;
+};
+
+export type CreateInsightsMiddleware = (props: InsightsProps) => Middleware;
+
+export const createInsightsMiddleware: CreateInsightsMiddleware = props => {
+  const { insightsClient } = props;
+  if (!insightsClient) {
```
Haroenv, line 13:
```suggestion
export const createInsightsMiddleware: CreateInsightsMiddleware = ({ }) => {
  const { insightsClient } = props;
  if (!insightsClient) {
```

src/middleware/insights.ts
```diff
@@ -0,0 +1,121 @@
...
+    throw new Error(
+      'passing insightsClient to instantsearch is required for insightsMiddleware'
+    );
+  }
+  return ({ instantSearchInstance }) => {
```
Haroenv, line 18:
```suggestion
export type InsightsProps = {};

export type CreateInsightsMiddleware = (props?: InsightsProps) => Middleware;

export const createInsightsMiddleware: CreateInsightsMiddleware = (props = {}) => {
  return ({ instantSearchInstance }) => {
    const insightsClient = instantSearchInstance.insightsClient;
    if (!insightsClient) {
      throw new Error(
        'passing insightsClient to instantsearch is required for insightsMiddleware'
      );
    }
```

I expected more something like this (still needs a test with InstantSearch with insightsClient & InstantSearch without insightsClient)

I left the props for RFC-37

src/lib/main.ts
francoischalifour: "middlewares" is not proper English.

src/middleware/insights.ts
eunjae-lee: Do you mean `(props = {})`?

src/middleware/insights.ts
eunjae-lee: Oh, before, I didn't read your message in the exception carefully.

As I commented in the RFC,

> My rationale behind this is to move insightsClient totally from instantsearch initialisation to the middleware (eventually).

I want to get rid of `insightsClient` from the `instantSearchInstance` in the end. What do you think?

Haroenv: Ah like that! I think it does make sense, but then the error message should be changed, so that this insightsClient becomes required, and it overrides the one set by the constructor (which should give a warning if they don’t match)

src/middleware/insights.ts
```diff
@@ -0,0 +1,121 @@
...
+      subscribe() {
+        updateQueryParameter(instantSearchInstance);
+        ensureInitialization({
+          instantSearchInstance,
+          insightsClient,
+          callback: () => {
+            setUserTokenFromQueueOrCookie(insightsClient, setUserToken);
+            watchUserToken(insightsClient, setUserToken);
+          },
+        });
```
francoischalifour, line 40: Why so many indirections? It makes the code harder to understand and to delete in the end. We better inline most of it—only `getAppIdAndApiKey` seems useful to extract.

eunjae-lee: @francoischalifour 
Okay, here, no offence at all, but genuinely curious on your perspective:

* Why do you think they're indirect?
* Maybe I have abstracted them in a poor way?
* Or, names aren't good enough to help you understand?

I splited the code into relevant chunks and named them after what they do. 

And I just tried inlined them after you comment, and I'm not sure if they're better.

* Do you think this inlined version is better than the abstracted one?
* Or, in your mind, did you have a better version than this screenshot?

image

src/middleware/insights.ts
eunjae-lee: just updated the error message: https://github.com/algolia/instantsearch.js/pull/4446/commits/20254cf18539b16ab1bff92907c40537ffd852d8

src/middleware/insights.ts
francoischalifour: In the current PR, you created functions just to name what a piece of code was doing. This is not really the point of functions—they're aimed at reducing complexity when the code they're hosting is supposed to be used at several places.

The code on the screenshot is way more straightforward. Later on, we'll feel confident removing a piece of code because it's inlined. Creating functions brings a mental overhead that creates friction when times comes to refactor/delete code (e.g., "this function had a name, we probably knew what we meant with this, let's not remove it").

From the outside, we (the team, the contributors, and yourself in a few days) understand the inlined code way better because it reads from top to bottom easily—with the functions, we jump from one statement to another, loosing track of what gets executed and when it gets executed.

I'd recommend reading "[Goodbye, Clean Code](https://overreacted.io/goodbye-clean-code/)".

eunjae-lee: Thanks for the explanation and the link. I'll give this change a try!

eunjae-lee:
```
insightsClient('setUserToken', 'def');

search.EXPERIMENTAL_use(createInsightsMiddleware({ insightsClient }));
```

In current implementation, if user sets userToken before creating the middleware, it will get ignored.
Does it make sense to you as long as it's documented? Or, should we allow user do that and somehow catch the token?

francoischalifour:
@eunjae-lee Is there a use case where the Insights' `userToken` can differ from InstantSearch's? As far as I know, there's none. We could display a dev warning saying that we detected that a token was previously set, but it's ignored because of the middleware (this warning can be displayed even if the `userToken` is the same, since it's useless to set it outside of the middleware).

eunjae-lee:
@francoischalifour You're talking about this, right?

```js
insightsClient('setUserToken', 'def');

search.EXPERIMENTAL_use(createInsightsMiddleware({ insightsClient }));
```

(To make sure we're on the same page,) user interacts with `insightsClient` to set userToken and the middleware gets updated and set the token into searchParameters. So

> Is there a use case where the Insights' userToken can differ from InstantSearch's?

The answer is "no". And I think it makes sense to add a warning.

src/lib/main.ts
eunjae-lee: https://en.wiktionary.org/wiki/middleware
image

I don't think I've seen `s` at the end of `middleware` before, though.
So you agree it's natural to have `middleware` although we have `routers, stateMappings, connectors, widgets`, right?

eunjae-lee:
```yarn run build:umd``` is failing at the moment.

```
src/index.ts → dist/instantsearch.development.js...
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
[!] Error: 'RouterProps' is not exported by src/middleware/createRouter.ts
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module
src/middleware/index.ts (15:23)
13: export type Middleware = (options: MiddlewareOptions) => MiddlewareDefinition;
14: 
15: export { createRouter, RouterProps } from './createRouter';
                           ^
16: export { createInsightsMiddleware } from './insights';
Error: 'RouterProps' is not exported by src/middleware/createRouter.ts
    at error (/Users/eunjaelee/workspace/instantsearch.js/node_modules/rollup/dist/rollup.js:5330:30)
    at Module.error (/Users/eunjaelee/workspace/instantsearch.js/node_modules/rollup/dist/rollup.js:9733:16)
    at handleMissingExport (/Users/eunjaelee/workspace/instantsearch.js/node_modules/rollup/dist/rollup.js:9634:28)
    at Module.getVariableForExportName (/Users/eunjaelee/workspace/instantsearch.js/node_modules/rollup/dist/rollup.js:9868:24)
    at NamespaceVariable.initialise (/Users/eunjaelee/workspace/instantsearch.js/node_modules/rollup/dist/rollup.js:4835:55)
    at Module.getOrCreateNamespace (/Users/eunjaelee/workspace/instantsearch.js/node_modules/rollup/dist/rollup.js:9810:36)
    at Module.traceVariable (/Users/eunjaelee/workspace/instantsearch.js/node_modules/rollup/dist/rollup.js:10068:36)
    at ModuleScope.findVariable (/Users/eunjaelee/workspace/instantsearch.js/node_modules/rollup/dist/rollup.js:8679:39)
    at Identifier$1.bind (/Users/eunjaelee/workspace/instantsearch.js/node_modules/rollup/dist/rollup.js:4353:40)
    at AssignmentExpression.bind (/Users/eunjaelee/workspace/instantsearch.js/node_modules/rollup/dist/rollup.js:3102:23)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Command failed: /usr/local/Cellar/yvm/3.6.7/versions/v1.16.0/bin/yarn.js build:umd
```

If I delete [this line](https://github.com/algolia/instantsearch.js/pull/4446/files#diff-557adfec4ab6f321a240e49e2a7c6c7dR43), the error goes away.
Not sure what's going on yet.

File: src/lib/main.ts (hunk 5 above)
Comment (Haroenv) [reply src/lib/main.ts:43]: I think it should be middlewares, at least for the object containing all possible middlewares. Interestingly my spell-checker only accepts singular "middleware" as correct, and not "middlewares"

File: src/middleware/insights.ts [hunk 11]
```diff
@@ -0,0 +1,114 @@
+import { InsightsClient, Middleware } from '../types';
+import { getInsightsAnonymousUserToken } from '../helpers';
+import { warning, noop } from '../lib/utils';
+
+export type InsightsProps = {
+  insightsClient: false | InsightsClient;
```
francoischalifour, line 6: Accepting `false` for the Insights client is confusing because it can't be `true`. Why don't we use `null` as we're doing with other options?

src/middleware/insights.ts
```diff
@@ -0,0 +1,114 @@
...
+        if (hasInsightsClient) {
+          // When `aa('init', { ... })` is called, it creates an anonymous user token in cookie.
+          // We can set it as userToken.
+          setUserTokenToSearch(getInsightsAnonymousUserToken());
+        }
+
+        if (Array.isArray((insightsClient as any).queue)) {
+          // Context: The umd build of search-insights is asynchronously loaded by the snippet.
+          //
+          // When user called `aa('setUserToken', 'my-user-token')` before `search-insights` is loaded,
```
francoischalifour, line 77:
```suggestion
          // When user calls `aa('setUserToken', 'my-user-token')` before `search-insights` is loaded,
```

src/middleware/insights.ts
Haroenv: null is a fair option, I suggested `false`, since that's what I did with Vue: https://github.com/algolia/vue-instantsearch/blob/master/src/mixins/widget.js#L47-L55

EDIT: it's actually `true` there, but `null` would make more sense (maybe a next major)

eunjae-lee: Yeah, `null` sounds good.
https://github.com/algolia/instantsearch.js/pull/4446/commits/b0b7add67b28f14da18f9e2d30c0b57649588567

src/middleware/insights.ts
```diff
@@ -0,0 +1,114 @@
...
+        });
+      },
+      unsubscribe() {
+        insightsClient('onUserTokenChange', undefined);
+      },
+    };
+  };
+};
+
+function getAppIdAndApiKey(searchClient) {
```
Haroenv, line 101: maybe add a unit test here. Since we run the tests with algoliasearch v3 and v4 separately, you should be able to cover both branches.

Other case we can add a dev dependency for `"algoliasearch-v3": "npm:algoliasearch@3"` and `import algoliasearch from 'algoliasearc-v3'`

src/types/insights.ts
```diff
@@ -1,20 +1,43 @@
...
 export type InsightsClientPayload = {
   eventName: string;
   queryID: string;
   index: string;
   objectIDs: string[];
   positions?: number[];
 };
-
-export type InsightsClient = (
+export type InsightsSendEvent = (
```
Haroenv, line 11:
```suggestion

export type InsightsSendEvent = (
```

src/types/insights.ts
```diff
@@ -1,20 +1,43 @@
 export type InsightsClientMethod =
   | 'clickedObjectIDsAfterSearch'
   | 'convertedObjectIDsAfterSearch';
-
 export type InsightsClientPayload = {
```
Haroenv, line 4:
```suggestion

export type InsightsClientPayload = {
```

src/middleware/index.ts
```diff
@@ -1,13 +1 @@
...
-  onStateChange(options: { uiState: UiState }): void;
-  subscribe(): void;
-  unsubscribe(): void;
-};
-
-export type Middleware = ({
-  instantSearchInstance: InstantSearch,
-}) => MiddlewareDefinition;
-
-export { createRouter, RouterProps } from './createRouter';
```
Haroenv, line 13: not sure if anyone uses this, but this changes the API

Haroenv (commented):
lgtm if open comments are resolved 👍

src/middleware/insights.ts
eunjae-lee: TIL: "npm:algoliasearch@3" syntax. Nice! https://github.com/algolia/instantsearch.js/pull/4446/commits/7f49f25ed23470295db24d7c6b0e6d06151ad4e2

src/lib/main.ts
eunjae-lee: image

https://www.wordhippo.com/what-is/the-plural-of/middleware.html

While `middleware` is more common, it's also safe to use `middlewares` if we want to explicitly indicate it's a collection.

src/middleware/index.ts
eunjae-lee: I added them back https://github.com/algolia/instantsearch.js/pull/4446/commits/ad0dafbf0e763d589d4da898e7b0431de022ef57
but not like before because of [this](https://github.com/algolia/instantsearch.js/pull/4463/files/c97960e3384f39db2677611b5a5f876e0f66f2a7#diff-f5dcfa3be457d40ab6df1ec1ef46c014)
(tl;dr -> rollup failed to resolve `RouterProps` type from `./createRouter`. It didn't happen before because this was not exported at `src/lib/main.ts`, but not it is.)

src/lib/main.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4446/commits/f34170d98118f82c2975bb1f6aad23ec0cc1f3b9

Haroenv (approved):
I think it's good now, assuming that nobody imported `instantsearch.js/es/middleware`, since that will be broken now. Since middleware is experimental, I think it's ok

src/middlewares/index.ts
```diff
@@ -0,0 +1,2 @@
+export { createInsightsMiddleware } from './insights';
```
francoischalifour, line 1: To be consistent with the file naming, we can rename the file `createInsightsMiddleware` (and perhaps `createRouterMiddleware`?).

src/middlewares/index.ts
```diff
@@ -0,0 +1,2 @@
...
+export * from './createRouter';
```
francoischalifour, line 2: We exports the types from the router middleware. Any opinions on aligning the two? (either exporting the types or not exporting the types)

src/types/index.ts
```diff
@@ -3,3 +3,4 @@ export * from './connector';
 export * from './widget';
 export * from './insights';
 export * from './algoliasearch';
+export * from './middleware';
```
francoischalifour, line 6: We can rename the types file to `middlewares` too.

eunjae-lee: This can stay as `middleware` since it's not a collection of middlewares but just a bunch of types related to middleware. Like the same way, in that folder we can find `connector` and `widget`. What do you think?

src/middlewares/index.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4446/commits/183b1ca2e8f7727ba397a1cb459975fd6ddfa9d9 good call!

src/middlewares/index.ts
eunjae-lee: I think it's good to export all of them https://github.com/algolia/instantsearch.js/pull/4446/commits/e7c7e5858aada1f4ee105764d524ca0bc54212ed

src/types/index.ts
francoischalifour: Yeah makes sense!

PR #4449, opened 2020-07-18
fix(connectToggleRefinement): fix onFacetValue/offFacetValue on render when using arrays for on/off

Kocal:
**Summary**

I was working on #4445 and I found a bug introduced by myself in #4420. :neutral_face: 

The bug is in `render` method, where `onFacetValue` and `offFacetValue` could be wrong when using arrays for `on` or `off`.

I've added a failing test: 
Capture d’écran de 2020-07-18 23-55-57

**Result**

`render` method now handle `onFacetValue` and `offFacetValue` correctly when using arrays for `on` and `off` options.

Thanks!

### DIFF
--- a/src/connectors/toggle-refinement/__tests__/connectToggleRefinement-test.js
+++ b/src/connectors/toggle-refinement/__tests__/connectToggleRefinement-test.js
@@ -152,6 +152,117 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/toggle-refi
     }
   });
 
+  it('Renders during init and render with array value', () => {
+    // test that the dummyRendering is called with the isFirstRendering
+    // flag set accordingly
+    const rendering = jest.fn();
+    const makeWidget = connectToggleRefinement(rendering);
+
+    const attribute = 'whatever';
+    const widget = makeWidget({
+      attribute,
+      on: ['a', 'b'],
+      off: ['c'],
+    });
+
+    const config = widget.getWidgetSearchParameters(new SearchParameters({}), {
+      uiState: {},
+    });
+    expect(config).toEqual(
+      new SearchParameters({
+        disjunctiveFacets: [attribute],
+        disjunctiveFacetsRefinements: {
+          [attribute]: ['c'],
+        },
+      })
+    );
+
+    const helper = jsHelper({}, '', config);
+    helper.search = jest.fn();
+
+    widget.init({
+      helper,
+      state: helper.state,
+      createURL: () => '#',
+    });
+
+    {
+      // should call the rendering once with isFirstRendering to true
+      expect(rendering).toHaveBeenCalledTimes(1);
+      const isFirstRendering =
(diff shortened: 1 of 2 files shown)


src/connectors/toggle-refinement/__tests__/connectToggleRefinement-test.js
```diff
@@ -152,6 +152,117 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/toggle-refi
...
+        rendering.mock.calls[rendering.mock.calls.length - 1][1];
+      expect(isFirstRendering).toBe(true);
+
+      // should provide good values for the first rendering
+      const { value, widgetParams } = rendering.mock.calls[
+        rendering.mock.calls.length - 1
+      ][0];
+      expect(value).toEqual({
+        name: 'whatever',
+        count: null,
```
Haroenv, line 202: since both of the `count` branches do a reduce starting with `0`, how do we get `null` here?

Kocal: Oh! Good catch, I forgot to handle this part https://github.com/algolia/instantsearch.js/pull/4449/files#diff-e59e0f69c70c301af28c0131f3a4fe9fR252-R257

Kocal: I don't know, I don't investigated yet.
However, we already have [`count: null`](https://github.com/algolia/instantsearch.js/blob/master/src/connectors/toggle-refinement/__tests__/connectToggleRefinement-test.js#L94) for a different - and not related - test.

Kocal: Maybe because of this https://github.com/algolia/instantsearch.js/blob/master/src/connectors/toggle-refinement/connectToggleRefinement.js#L221

Haroenv: interesting that it comes from render, I thought you're asserting right above that it was called only once

Kocal: What is the state of your comment? Is it fine or it's a bug? 👀

tristanbes:
nice catch !

src/connectors/toggle-refinement/__tests__/connectToggleRefinement-test.js
eunjae-lee: I think that `null` comes from `init` of the connector [here](https://github.com/algolia/instantsearch.js/blob/f570974615f009e18dc6a510fc5a4519df690347/src/connectors/toggle-refinement/connectToggleRefinement.js#L191). So it looks normal.

eunjae-lee:
CircleCI didn't run. I ran the tests and they all passed on my computer, so I'm merging it.

Kocal:
Thanks!

Is there a reason for CircleCI not running? It's not the first time I see this. :(

PR #4450, opened 2020-07-20
fix(middleware): rename EXPERIMENTAL_use to use

eunjae-lee:
**Summary**

This PR renames `EXPERIMENTAL_use` to `use`.

The definition of Middleware could change later, but I don't think the signature of use(...middleware) will change a lot.
And we're thinking of introducing a new middleware. So it's a good time to get rid of `EXPERIMENTAL_` prefix.

However for compatibility, we will keep the alias for a while.


eunjae-lee:
Putting it back until we make 'insights' middleware and proves this interface works.

src/lib/InstantSearch.ts
```diff
@@ -270,6 +270,11 @@ See ${createDocumentationLink({
     return this;
   }
 
+  // This is an alias to keep for a while
+  public EXPERIMENTAL_use(...middleware: Middleware[]): this {
+    return this.use(...middleware);
```
Haroenv, line 275: add a dev warning here IMO

src/lib/InstantSearch.ts
```diff
@@ -270,6 +270,11 @@ See ${createDocumentationLink({
     return this;
   }
 
+  // This is an alias to keep for a while
```
Haroenv, line 273:
```suggestion
  // @major we shipped with EXPERIMENTAL_use, but have changed that to just `use` now
```

src/lib/InstantSearch.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4450/commits/67c4a00e9bd85566a316b0194bc69adb2549dd86

Haroenv (approved):
is good, we can add a test about the experimental_use warning maybe?

eunjae-lee:
pinging for visibility cc @francoischalifour @yannickcr

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -252,6 +252,50 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
...
+
+    const middleware = () => ({
+      onStateChange: () => {},
+      subscribe: () => {},
+      unsubscribe: () => {},
+    });
+
+    expect(() => {
+      search.EXPERIMENTAL_use(middleware);
+    }).toWarnDev();
```
francoischalifour, line 275: We better assert the actual message because it has happened that other warnings were printed when calling a lifecycle method.

src/lib/InstantSearch.ts
```diff
@@ -277,6 +277,16 @@ See ${createDocumentationLink({
     return this;
   }
 
+  // @major we shipped with EXPERIMENTAL_use, but have changed that to just `use` now
+  public EXPERIMENTAL_use(...middleware: Middleware[]): this {
+    warning(
+      false,
+      'We recommend you to replace `EXPERIMENTAL_use` with `use`.'
```
francoischalifour, line 284: Let's explain _why_ to users.

```suggestion
      'The InstantSearch middleware API is now considered stable so we recommend replacing `EXPERIMENTAL_use` with `use` before upgrading to the next major version.'
```

yannickcr (approved):

Haroenv (changes requested):
adding a "request changes" for François' suggestions

src/lib/__tests__/InstantSearch-test.js
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4450/commits/5cd0d7983d94bc8442b566aec3ca0dcd27371816

src/lib/InstantSearch.ts
```diff
@@ -277,6 +277,16 @@ See ${createDocumentationLink({
     return this;
   }
 
+  // @major we shipped with EXPERIMENTAL_use, but have changed that to just `use` now
+  public EXPERIMENTAL_use(...middleware: Middleware[]): this {
+    warning(
+      false,
+      'The middleware API is now considered stable so we recommend replacing `EXPERIMENTAL_use` with `use` before upgrading to the next major version.'
```
Haroenv, line 284:
```suggestion
      'The middleware API is now considered stable, so we recommend replacing `EXPERIMENTAL_use` with `use` before upgrading to the next major version.'
```

Haroenv (approved):

francoischalifour (approved):

PR #4456, opened 2020-07-28
feat(insights): send events from hits and refinementList (2/4)

eunjae-lee:
## Summary

This PR improves `insights` middleware to send events.
This PR also
- exposes `sendEvent` to `connectRefinementList`
- exposes `sendEvent` and `bindEvent` to `connectHits`
- exposes `bindEvent` to `Hits` and its temlate
to demonstrate how in practice it could work.

[example](https://codesandbox.io/s/instantsearchjs-c9s30?file=/src/app.js) → It works when you check something in the `refinementList` and click any buttons in the `hits` and the custom hits widgets. (You can see the network panel of your browser to inspect the events sent to Algolia Insights)

**Stacked PRs**
1. [introduce `insights` middleware](https://github.com/algolia/instantsearch.js/pull/4446)
2. [send events from hits and refinementList](https://github.com/algolia/instantsearch.js/pull/4456) ← current
4. [add tests](https://github.com/algolia/instantsearch.js/pull/4464)

## Two cases

1. The facet related connectors will run `sendEvent('click', ...)` along with `refine` and expose `sendEvent` to component layer.
2. The hit related connectors will run `sendEvent('view', ...)` when rendering items and expose `sendEvent` to component layer. It will also expose `bindEvent` to templates which will be handled by insightsListener.

We expose `instantSearchInstance.sendEventToInsights()`:
- which is called in connectors
- which can be used by users when they manually want to send events

## Details about the design

<details>
<summary>How to let users sendEvent from `hits` and `infiniteHits`?</summary>

### `click` or `conversion` events from `connectHits`

We need to provide a way for users to send `click` or `conversion` events.

#### Case 1. custom widget

Users can create custom components by using connectors:

```js
let listener;
const renderFn = (renderOptions) => {
  const { hits, sendEvent } = renderOptions;

  const elem = /* ... */;
  if (listener) {
    elem.removeEventListener("click", listener);
  }
  listener = (event) => {
    const index = /* ... */;
    const hit = hits[index];
    sendEvent('conversion', hit, "Product Bookmarked");  // or 'click'
    // ↑ I've got feedback from Mathieu that we cannot use default event names for 'click' and 'conversion' in the context of hits.
    // It's because there can be many different types of clicks and conversions. Moreover the event names are used by Perso. So we should leave it to the users.
  }
  elem.addEventListener("click", listener);
};
const customWidget = connectHits(renderFn);
```

And they can use `sendEvent` exposed by the connector. So this it looks okay in this case.

#### Case 2. template

`hits` and `infiniteHits` widgets are normally used with the `item` template.

```js
templates: {
  item(hit) {
    const html = /* ... */;
    return html;
  },
},
```

In this case, even if we expose `sendEvent` as in `item(hit, sendEvent, eventName)`, it's not easy to use because at this stage we don't have DOM element to attach event listeners.

And even if we try to add event listeners outside the InstantSearch lifecycle,

```js
search.addWidgets([
  hits({
    container: "#hits",
    templates: {
      item(hit) {
        /* ... */
      },
    },
  }),
]);

document.querySelector("#hits").addEventListener("click", (event) => {
  // How do we get `hit` object here?
});
```

So, in the end, we go back to what we've been doing:

```js
templates: {
  item: (hit) => `
    <button ${instantsearch.insights("clickedObjectIDsAfterSearch", {
      eventName: "click-result",
      objectIDs: [hit.objectID],
    })}>
      Click event
    </button>
  `;
}
```

Similarly,

```js
templates: {
  item: (hit, bindEvent) => `
    <button ${bindEvent("click", hit, "Product Bookmarked")}>
      Click event
    </button>
  `;
}
```

`bindEvent` is similar to `sendEvent`, but instead of delivering it to `insights` middleware, it returns a string like the following:

```js
return `data-insights-event="${btoa(JSON.stringify({ widgetType, eventType, aaMethod, payload ))}"`
```

And we can extend [insightsListener](https://github.com/algolia/instantsearch.js/blob/master/src/lib/insights/listener.tsx#L26:L26) to handle this data- attribute.

According to this addition, I'm in favor of deprecating the old way of

```js
    <button ${instantsearch.insights("clickedObjectIDsAfterSearch", {
      eventName: "click-result",
      objectIDs: [hit.objectID],
    })}>
      Click event
    </button>
```
and things related to it (for example, `withInsights` and `withInsightsListener`).
</details>

<details>
<summary>Sending custom events</summary>

If users want to send raw custom event without any implication, they can call `search.sendEventToInsights(args)` and `args` will be sent to `insights` middleware if being used.
The minimum `args` for Algolia Insights is `{ aaMethod, payload }`. The full signature is `{ widgetType, eventType, aaMethod, payload }`.

</details>

<details>
<summary>Default event names</summary>

### Facet
We normally send only `click` event from these connectors: connectRefinementList, connectBreadcrumb, ...
And its default event name is `Filter Applied`, which makes sense.

### Hit
Here we could have all of three types of events:
- view
- click
- conversion

In case of `view`, we can safely use `Hits Viewed` as a default event name.

However in case of `click` or `conversion`, we cannot force any default event name because there can be different types of events and if we use a default name for all, it can affect the quality of personalization. So we do not have any default name and let users decide.
</details>


src/middleware/createRouter.ts
```diff
@@ -51,6 +51,8 @@ export const createRouter: RoutingManager = (props = {}) => {
     };
 
     return {
+      $$type: 'aism.router',
```
eunjae-lee, line 54: I've added `$$type` to the middleware to distinguish middlwares.
Not sure about the prefix `aism.` though.

src/middleware/index.ts
```diff
@@ -1,16 +1,19 @@
 import { UiState, InstantSearch } from '../types';
 
-export type MiddlewareDefinition = {
+export type MiddlewareDefinition<TExtraDefinition> = {
+  $$type: string;
   onStateChange(options: { uiState: UiState }): void;
   subscribe(): void;
   unsubscribe(): void;
-};
+} & TExtraDefinition;
```
eunjae-lee, line 8: `insights` middleware needs to have something extra. What do you think of the usage of generic here? Any better way to achieve it?

src/middleware/createRouter.ts
Haroenv: I think `ais` is good enough

src/middleware/insights.ts
```diff
@@ -95,6 +107,9 @@ aa('setUserToken', 'your-user-token');
       unsubscribe() {
         insightsClient('onUserTokenChange', undefined);
       },
+      sendEvent(event: InsightsEvent) {
```
Haroenv, line 110: I think this should be added to InstantSearch directly

```
instantsearch.middlewareGrabBag.insights.SendEvent = ...
```

or even directly:

```
instantsearch.$insights.sendEvent
```

Maybe that's less easily typeable though

src/lib/InstantSearch.ts
```diff
@@ -587,6 +588,20 @@ Feel free to give us feedback on GitHub: https://github.com/algolia/instantsearc
...
+  public sendEventToInsights(event: InsightsEvent) {
+    const insights = this.getInsightsMiddleware();
+    if (!insights) {
+      return;
+    }
+    insights.sendEvent(event);
+  }
+
+  private getInsightsMiddleware() {
+    return this.middleware.find(
```
Haroenv, line 601: This doesn't seem like the most efficient method, I think it should be added to an object instead of an array

src/lib/InstantSearch.ts
```diff
@@ -587,6 +588,20 @@ Feel free to give us feedback on GitHub: https://github.com/algolia/instantsearc
 
     this.mainHelper.clearCache().search();
   }
+
+  public sendEventToInsights(event: InsightsEvent) {
```
Haroenv, line 592: this probably should be called directly from the widgets, without a builtin method for insights in the InstantSearch API

src/middleware/insights.ts
eunjae-lee: I've changed it to `this._insightsMiddleware.sendEvent()`

src/lib/InstantSearch.ts
eunjae-lee: `instantSearchInstance.sendEventToInsights()` will be mostly called from connectors.
(For example, [here](https://github.com/algolia/instantsearch.js/blob/feat/send-events/src/connectors/refinement-list/connectRefinementList.js#L308-L312) -> [here](https://github.com/algolia/instantsearch.js/blob/feat/send-events/src/lib/utils/createSendEventForFacet.ts#L19:L19))

And we can guide users to use it when they want to send a custom event.
So I've exposed `sendEventToInsights()` in InstantSearch as a public method.
What do you think?

src/lib/InstantSearch.ts
```diff
@@ -256,6 +262,13 @@ See ${createDocumentationLink({
       const newMiddleware = fn({ instantSearchInstance: this });
       this.middleware.push(newMiddleware);
 
+      if (
+        !this._insightsMiddleware &&
+        isInsightsMiddlewareDefinition(newMiddleware)
+      ) {
+        this._insightsMiddleware = newMiddleware;
+      }
```
Haroenv, line 270: I think this code should be _inside_ the middleware

src/lib/InstantSearch.ts
Haroenv: It makes sense, this way the widgets can assume the middleware exists, even if it doesn't. 

We have a similar thing with createURL returning a default of `#`

I thought a little, but can't think of a way to share this default value efficiently. The downside I see here is that if someone wants to make their own middleware which is optional, they need to add some code to InstantSearch

Haroenv: Could we maybe do something like `instantsearchInstance.sendEvent?()` (which gets called in the widget directly, after being added to the instance by the middleware)

src/lib/insights/listener.tsx
```diff
@@ -23,19 +26,51 @@ const findInsightsTarget = (
...
+
+  if (typeof serializedPayload !== 'string') {
+    throw new Error(
+      'The insights middleware expects `data-insights-event` to be a base64-encoded JSON string.'
+    );
+  }
+
+  try {
+    const payload = JSON.parse(atob(serializedPayload));
+    return payload;
```
Haroenv, line 40:
```suggestion
    return JSON.parse(atob(serializedPayload));
```

src/lib/utils/createSendEventForFacet.ts
```diff
@@ -0,0 +1,47 @@
...
+        });
+      }
+    } else if (args.length === 1) {
+      instantSearchInstance.sendEventToInsights(args[0]);
+    } else {
+      throw new Error(`You need to pass two arguments: eventType, facetValue.
+(eventType = 'click')
+
+If you want to send a custom payload, you can pass one object: sendEvent(customPayload);
+`);
```
Haroenv, line 43: this error needs to be smaller in prod IMO

src/lib/utils/isFacetRefined.ts
```diff
@@ -0,0 +1,13 @@
+import { AlgoliaSearchHelper } from 'algoliasearch-helper';
+
+export default function isFacetRefined(
+  helper: AlgoliaSearchHelper,
+  facet: string,
+  value: string
+) {
+  if (helper.state.isConjunctiveFacet(facet)) {
```
Haroenv, line 8: I think this use case fits us good enough, but it means that if there are two refinement lists on the same attribute, one disjunctive, one conjunctive, the isFacetRefined will be wrong.

However, this should be checked, since maybe it's not possible at all to mix

src/middleware/index.ts
Haroenv: is this still needed?

src/middleware/insights.ts
Haroenv: This can be a function in scope of the middleware, which is being called by by a function which is added to the instance in subscribe:

```
// pseudo-code
insights() {
  const sendEvent = ...
  return {
    subscribe() {
      instantsearch.sendInsightsEvent = sendEvent
```

instantsearch can then either have a noop sendEvent by default, or the widgets can check if sendEvent exists first

src/lib/InstantSearch.ts
eunjae-lee: > The downside I see here is that if someone wants to make their own middleware which is optional, they need to add some code to InstantSearch

I agree. Potentially we could have `instantSearchInstant.middlewareBlahBlah` as a hub to access middlewares or things returned/added by middlewares. I think you left this idea as a comment somewhere in this PR or the other.

For now, I like your idea [here](https://github.com/algolia/instantsearch.js/pull/4456#discussion_r462103344). I'll try that.

src/middleware/insights.ts
eunjae-lee: I love this idea. It makes so many verbose things disappear. https://github.com/algolia/instantsearch.js/pull/4456/commits/19751711777f8621e483d5d4a33ada728f938ae6

src/middleware/index.ts
eunjae-lee: not anymore

src/lib/InstantSearch.ts
eunjae-lee: done https://github.com/algolia/instantsearch.js/pull/4456/commits/19751711777f8621e483d5d4a33ada728f938ae6

src/lib/utils/isFacetRefined.ts
eunjae-lee: image

okay it's not possible at all at least at the moment.

src/lib/utils/createSendEventForFacet.ts
Haroenv: We want insights probably not to be an error in production, but maybe an error in dev can make sense

eunjae-lee: Okay, it makes sense. I'll throw errors with __DEV__ condition.

eunjae-lee:
After discussing with Mathieu, 
in case of hit releated events, we cannot use a default event name for "click" and "conversion".
So the signature becomes

```js
sendEvent(eventType, hits, eventName);

sendEvent("conversion", hits, "Product Bookmarked");
```

I've added this in the summary of this PR.

eunjae-lee:
> ✅ for the API now, exact events still need to be validated IMO

@Haroenv Do you mean you want to see a sample or to see all the events (which we defined) to be implemented?

If the former, [this example](https://codesandbox.io/s/instantsearchjs-c9s30?file=/src/app.js) correctly sends events to Insights when you click any buttons in the hits or custom hits.

Haroenv:
only hits & refinementlist seem to have been implemented, not menu, hierarchical menu etc, is that expected?

eunjae-lee:
> only hits & refinementlist seem to have been implemented, not menu, hierarchical menu etc, is that expected?

It's intended. This PR seems to be big enough, so I was thinking of stacking another PR on top of this.

src/lib/utils/createSendEventForFacet.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (!isFacetRefined(helper, attribute, facetValue)) {
+        // send event only when the facet is being checked "ON"
+        instantSearchInstance.sendEventToInsights({
+          insightsMethod: 'clickedFilters',
+          widgetType,
+          eventType,
+          payload: {
+            eventName,
+            index: helper.getIndex(),
+            filters: [`${attribute}:${JSON.stringify(facetValue)}`],
```
eunjae-lee, line 33: I stringified `facetValue` because the value could contain special characters like double qoutes.
Is it enough? I think we should have similar logic somewhere in either InstantSearch, helper or client.
I haven't looked into them yet.
If someone already knows, can you point me to it?

eunjae-lee: It's not for the js client. The [doc](https://www.algolia.com/doc/api-reference/api-methods/clicked-filters/#method-param-filters) refers to `filters` in the api reference. So I thought I should follow the `filters` format which requires the value to be quoted when it's with a space or quotes, etc.

eunjae-lee:
unit tests and lint are fixed in [the next(3rd) PR](https://github.com/algolia/instantsearch.js/pull/4463)

src/types/middleware.ts
```diff
@@ -2,6 +2,7 @@ import InstantSearch from '../lib/InstantSearch';
 import { UiState } from './widget';
 
 export type MiddlewareDefinition = {
+  $$type: string;
```
francoischalifour, line 5: Since users are meant to create their own middleware, I think `$$type` should be optional.

src/connectors/hits/connectHits.ts
```diff
@@ -51,15 +65,26 @@ const connectHits: HitsConnector = function connectHits(
...
 
     return {
       $$type: 'ais.hits',
 
-      init({ instantSearchInstance }) {
+      init({ instantSearchInstance, helper }) {
+        sendEvent = createSendEventForHits({
+          instantSearchInstance,
+          helper,
+          widgetType: 'ais.hits',
```
francoischalifour, line 78: We can reuse `this.$$type`.

src/connectors/hits/connectHits.ts
```diff
@@ -51,15 +65,26 @@ const connectHits: HitsConnector = function connectHits(
...
       widgetParams || ({} as typeof widgetParams);
+    let sendEvent;
+    let bindEvent;
 
     return {
       $$type: 'ais.hits',
 
-      init({ instantSearchInstance }) {
+      init({ instantSearchInstance, helper }) {
+        sendEvent = createSendEventForHits({
```
francoischalifour, line 75: The `sendEvent` and `bindEvent` function references are lost between `init` and `render` because they're reassigned. This means that if users bind an event callback in `init` (when `isFirstRender` is true), the callback will be a noop even during `render` (when `isFirstRender` is false).

You can see how we usually fix this by searching for `connectorState` in the codebase.

src/lib/utils/createSendEventForHits.ts
```diff
@@ -0,0 +1,157 @@
...
+      eventType,
+      payload: {
+        eventName,
+        index,
+        queryID,
+        objectIDs,
+      },
+    };
+  } else if (__DEV__) {
+    throw new Error(`eventType("${eventType}") is not supported.
```
francoischalifour, line 107: Having different paths execution in dev and prod might lead to unexpected behaviors. Should we throw in all environments or display a warning instead?

src/lib/utils/createSendEventForHits.ts
```diff
@@ -0,0 +1,157 @@
...
+  const positions = hitsArray.map(hit => hit.__position);
+  const index = helper.getIndex();
+
+  if (eventType === 'view') {
+    return {
+      insightsMethod: 'viewedObjectIDs',
+      widgetType,
+      eventType,
+      payload: {
+        eventName: eventName || 'Hits Viewed',
```
francoischalifour, line 76: Why is this the only event name defaulted?

src/widgets/hits/hits.tsx
```diff
@@ -52,6 +53,10 @@ const renderer = ({
       results={results}
       templateProps={renderState.templateProps}
       insights={insights as InsightsClientWrapper}
+      sendEvent={(event: InsightsEvent) => {
```
francoischalifour, line 56: The function reference is different every time. Is it OK in our case?

src/middleware/insights.ts
```diff
@@ -49,6 +57,7 @@ aa('setUserToken', 'your-user-token');
     });
 
     return {
+      $$type: 'ais.insights',
```
francoischalifour, line 60: Why do we need `$$type` on middleware?

Haroenv: I think it was needed for a previous design where InstantSearch read from the middleware array for the sendEvent function, but now it's changed so the middleware augments the InstantSearch instance directly

src/connectors/hits/connectHits.ts
eunjae-lee: @francoischalifour Can you explain a more how they're reassigned?

Before that, **what** is reassigned? For example, in [`connectBreadcrumb.ts`](https://github.com/algolia/instantsearch.js/blob/feat/send-events/src/connectors/breadcrumb/connectBreadcrumb.ts#L90:L90), `connectorState.createURL` is assigned only once in `init`.

In the current branch, I compared `sendEvent` when isFirstRendering==true and ==false, and they were the same (`===`). Which part am I missing?

src/lib/utils/createSendEventForHits.ts
eunjae-lee: It's about analytics, so even if something's wrong, it shouldn't throw errors in production just because of analytics. However, I think it's good to throw in __DEV__, so that they can get notified and have a chance to fix things before deployment. 

In the line 107, it's not 

> It could go wrong, please be aware of xxx.

but, it's actually

> It won't send any events at all.

WDYT?

src/middleware/insights.ts
eunjae-lee: Right. It's not needed any more. So I've removed it. https://github.com/algolia/instantsearch.js/pull/4456/commits/77b77a9b96f127b92f573aa0ffe77263b1408fc5

src/lib/utils/createSendEventForHits.ts
eunjae-lee: I've added the following section in the PR description:
`Details about the design > Default event names`

src/connectors/hits/connectHits.ts
Haroenv: I think the difference here is that the function isn't also created in `render`, and thus indeed doesn't change reference

eunjae-lee: Had a call with François, and it's is good as is because `sendEvent` is not reassigned in `render` method.

Haroenv (approved):

src/widgets/hits/hits.tsx
eunjae-lee: It's okay because this `sendEvent` is used [in the Preact context](https://github.com/algolia/instantsearch.js/blob/feat/send-events/src/lib/insights/listener.tsx#L76:L76).

src/connectors/hits/connectHits.ts
```diff
@@ -24,6 +28,16 @@ export type HitsRendererOptions = {
    * The response from the Algolia API.
    */
   results?: SearchResults<AlgoliaHit>;
+
+  /**
+   * Send event to insights middleware
```
francoischalifour, line 33:
```suggestion
   * Sends an event to the Insights middleware.
```

src/connectors/hits/connectHits.ts
```diff
@@ -24,6 +28,16 @@ export type HitsRendererOptions = {
...
+   */
+  sendEvent: SendEventForHits;
+
+  /**
+   * Returns a string of data-insights-event attribute for insights middleware
```
francoischalifour, line 38:
```suggestion
   * Returns a string for the `data-insights-event` attribute for the Insights middleware
```

src/lib/insights/listener.tsx
```diff
@@ -23,19 +26,50 @@ const findInsightsTarget = (
...
+    throw new Error(
+      'The insights middleware was unable to parse `data-insights-event`.'
+    );
+  }
+};
+
 const insightsListener = (BaseComponent: any) => {
   function WithInsightsListener(props: WithInsightsListenerProps) {
     const handleClick = (event: MouseEvent): void => {
+      // old way, e.g. instantsearch.insights("clickedObjectIDsAfterSearch", { .. })
```
francoischalifour, line 50: We now allow people to use the two at the same time? Shouldn't we check the new way, and then the old way exclusively?

```js
if (props.sendEvent) {}
else {}
```

src/lib/utils/createSendEventForHits.ts
```diff
@@ -0,0 +1,157 @@
+import { AlgoliaSearchHelper } from 'algoliasearch-helper';
```
francoischalifour, line 1: Why are these files not in the Insights middleware folder?

src/lib/utils/createSendEventForHits.ts
francoischalifour: Yep it makes sense on a user POV. It just harder to test the two code paths and the flow is harder to follow.

src/connectors/hits/connectHits.ts
```diff
@@ -92,10 +120,14 @@ const connectHits: HitsConnector = function connectHits(
           typeof escapeHits
         >).__escaped = initialEscaped;
 
+        sendEvent('view', results.hits);
+
         renderFn(
           {
             hits: results.hits,
             results,
+            sendEvent,
```
francoischalifour, line 129: What happens when `sendEvent` is called without the Insights middleware?

francoischalifour:
At every render we now send `widgetCount + search = widgetCount + 1` network requests, which is linear and seems huge. We could achieve `1 + 1 = 2` network requests if we batch the calls to the Algolia `events` endpoint.

Would it be possible to do so?

src/connectors/hits/connectHits.ts
eunjae-lee: `sendEvent` eventually calls `instantSearchInstance.sendEventToInsights()` and it's `noop` by default. (https://github.com/algolia/instantsearch.js/blob/feat/send-events/src/lib/InstantSearch.ts#L245:L245)

src/lib/utils/createSendEventForHits.ts
eunjae-lee: These files are quite big enough by themselves including test cases (the test cases are added in #4464).
They are there because of the middlewares, but also they're not directly used by middlewares but by connectors. So I put them in separate files.

src/lib/insights/listener.tsx
eunjae-lee: Makes sense! https://github.com/algolia/instantsearch.js/pull/4456/commits/61ca509ced8082c6efd007d8ab04f0c65645bba6

Haroenv:
only the (infinite)hits widget sends view events, right?

francoischalifour:
@Haroenv If you refine an item in a RefinementList, it calls the `events` endpoint twice:

- RefinementList: filter clicked
- Hits: viewed

My point was that we should (at some point) batch these two in a single network request.

eunjae-lee:
> @Haroenv If you refine an item in a RefinementList, it calls the `events` endpoint twice:
> 
>     * RefinementList: filter clicked
> 
>     * Hits: viewed
> 
> 
> My point was that we should (at some point) batch these two in a single network request.

@francoischalifour I missed your feedback before.

(1) When a page is rendered, `view` event will be sent from hits/infiniteHits.
(2) And when user clicks on refinementList, then `click` will be sent.

There should be a time gap between (1) and (2).
(1) is programmatic and (2) is triggered by user action.

If there are a bunch of events sent programmatically, then we could batch them in a bucket of 1 second or something like that. For example, if a user has multiple hits or infiniteHits widgets on a single page, then multiple `view` events could be sent.

Let me know what you think. We can improve this later including more opinions :)

francoischalifour:
We could wait for the next tick before sending all events as a batch (which is what we're doing in the JavaScript search helper I believe).

That's not a blocker at all, just wanted everybody to be aware of that.

PR #4457, opened 2020-07-28
feat(core): introduce `getWidgetRenderState` (2/n)

francoischalifour:
> This is part 2 of the series "[Render State](https://github.com/algolia/instantsearch-rfcs/pull/38)".

## Related

- algolia/instantsearch-rfcs#38
- algolia/instantsearch.js#4454

## Description

This introduces the widget lifecycle hook `getWidgetRenderState` and implements it for `connectSearchBox`.

The `getWidgetRenderState` method allows to accumulate and store the global render state of the widgets tree. This render state has a similar structure to the UI state but stores the render params.

## Motivation

When you mount a widget in the InstantSearch tree, you may want to use its render params (e.g., `refine` function) in another widget to create richer and more interactive interfaces.

In InstantSearch.js, it requires using the JavaScript helper, which means that you get out of the InstantSearch API. In React InstantSearch, you can hack around it to [use a React context and provide the render params to other components](https://github.com/algolia/unified-instantsearch-ecommerce/blob/8a5adffd651e8b0df6101bfef817a0a7decaeda4/src/components/SearchBox/ConnectedSearchBox.js#L11-L13). Still, this is a workaround that goes against React's paradigm.

**This PR aims at providing an InstantSearch API layer that provides the render state to all widgets.**

### In an InstantSearch widget

The render state is available in the render params.

```tsx
const customPagination = connectPagination(
  ({ renderState, widgetParams }, isFirstRender) => {
    if (isFirstRender) {
      const button = document.createElement('button');
      button.textContent = 'Trigger search';
      button.addEventListener('click', () => {
        // Control the search box from the pagination widget
        renderState.instant_search.searchBox.refine('samsung');
      });
      widgetParams.container.appendChild(button);
    }
  }
);
```

### In a component outside of InstantSearch

The render state is available on the InstantSearch instance on `search.renderState` so that developers can interact with the search state from outside of InstantSearch widgets. This avoids nesting multiple widgets to use their render params. Using widgets also have the cons of relying on the network and the search results to be back before rendering (because we wait for results before rendering). This results in a UI flash, even if you don't want to display only search data coming from the network; this goes against the progressive enhancement paradigm.

A pattern that can emerge is to mount virtual widgets to make their render state available on the instance, then use the render state params in another website's element.

```tsx
function App({ search }) {
  function filterSamsungProducts() {
    search.renderState.instant_search.refinementList.brand.refine('Samsung');
  }

  return (
    <div>
      <button onClick={filterSamsungProducts}>Check Samsung products</button>
    </div>
  );
}
```

_The JSX syntax is used for convenience_.

### With React InstantSearch

> This PR doesn't deal with React InstantSearch because it's not based on InstantSearch.js, but this section explains how we can expose this Render State API in React as a next step.

The render state could be exposed using a React Context provider:

```tsx
function App() {
  return (
    <InstantSearch.RenderStateProvider>
      {(renderState) => (
        /* ... */
      )}
    </InstantSearch.RenderStateProvider>
  );
}
```

And using React hooks:

```tsx
function Search() {
  const renderState = useRenderState();

  return (
    /* ... */
  );
}
```

### With Vue InstantSearch

This is an example of a `renderState` mixin that exposes the API to Vue InstantSearch.

```html
<template>
  <ais-refinement-list>
    <template slot-scope="{ refine, items }">
      <button
        @click="
          refine(items[0].value);
          setQuery('');
        "
      >
        Refine first element and reset query
      </button>
    </template>
  </ais-refinement-list>
</template>

<script>
  export default {
    mixins: [renderState],
    methods: {
      setQuery(query) {
        return this.renderState.instant_search.searchBox.refine(query);
      },
    },
  };
</script>
```

## Detailed design

The render state has a similar structure to the UI state but stores the render params.

> **Note:** the render state is exclusively stored "per widget", as opposed to the UI state that delegates concepts like the search query to a `query` key, instead of a `searchBox` key.

### Storage

The render state is stored on the InstantSearch instance as an object.

```tsx
type RenderState = {
  [indexName: string]: {
    searchBox: {
      query: string;
      refine(query: string): void;
      // ...
    };
    // ...
  };
};

const renderState: RenderState = {
  instant_search: {
    searchBox: {
      query: 'iphone',
      refine: ƒ(query),
      // ...
    },
  },
};
```

### API addition

Each widget can now implement a new method called `getWidgetRenderState(indexRenderState, renderOptions): IndexRenderState`.

**Example**

```ts
const widget = {
  // ...
  getWidgetRenderState(renderState, { helper, searchMetadata }) {
    return {
      ...renderState,
      searchBox: {
        query: helper.state.query || '',
        refine: this._refine,
        clear: this._cachedClear,
        widgetParams,
        isSearchStalled: searchMetadata.isSearchStalled,
      },
    };
  },
};
```

Having an API for the render state (previously called render params) results in a few benefits:

- No direct behavioral change in how InstantSearch and its widgets work because the API is an addition
- Fewer inconsistencies between `init` and `render` returned render params (they would both use `getWidgetRenderState`)

Storing the render state on the InstantSearch instance unlocks APIs that we wanted to implement:

- **Access `widgetParams` from the InstantSearch instance.** This can be useful to track which widget options are used. (E.g., `renderState.instant_search.searchBox.widgetParams`)
- **Provide information to hide widgets with `panel`.** We now have the `getWidgetRenderState` API that can be used to compute a `canRefine` value, instead of finding the information with the JavaScript helper.

## How we teach this

- Create a new section about the render state in the documentation

## Next steps

- Implement the `getWidgetRenderState` method in all widgets

### DIFF
--- a/src/connectors/search-box/connectSearchBox.js
+++ b/src/connectors/search-box/connectSearchBox.js
@@ -72,8 +72,7 @@ export default function connectSearchBox(renderFn, unmountFn = noop) {
 
     function clear(helper) {
       return function() {
-        helper.setQuery('');
-        helper.search();
+        helper.setQuery('').search();
       };
     }
 
@@ -86,7 +85,8 @@ export default function connectSearchBox(renderFn, unmountFn = noop) {
         this._clear();
       },
 
-      init({ helper, instantSearchInstance }) {
+      init(initOptions) {
+        const { helper, renderState, instantSearchInstance } = initOptions;
         this._cachedClear = this._cachedClear.bind(this);
         this._clear = clear(helper);
 
@@ -107,27 +107,21 @@ export default function connectSearchBox(renderFn, unmountFn = noop) {
 
         renderFn(
           {
-            query: helper.state.query || '',
-            refine: this._refine,
-            clear: this._cachedClear,
-            widgetParams,
+            ...this.getWidgetRenderState(renderState, initOptions).searchBox,
             instantSearchInstance,
           },
           true
         );
       },
 
-      render({ helper, instantSearchInstance, searchMetadata }) {
+      render(renderOptions) {
+        const { helper, renderState, instantSearchInstance } = renderOptions;
         this._clear = clear(helper);
 
         renderFn(
           {
-            query: helper.state.query || '',
-            refine: this._refine,
-            clear: this._cachedClear,
-            widgetParams,
+            ...this.getWidgetRenderState(renderState, renderOptions).searchBox,
             instantSearchInstance,
-            isSearchStalled: searchMetadata.isSearchStalled,
           },
           false
         );
@@ -137,6 +131,19 @@ export default function connectSearchBox(renderFn, unmountFn = noop) {
         unmountFn();
 
         return state.setQueryParameter('query', undefined);
+      },
+
+      getWidgetRenderState(renderState, { helper, searchMetadata }) {
+        return {
+          ...renderState,
+          searchBox: {
+            query: helper.state.query || '',
+            refine: this._refine,
+            clear: this._cachedClear,
+            widgetParams,
+            isSearchStalled: searchMetadata.isSearchStalled,
+          },
+        };
       },
 
       getWidgetUiState(uiState, { searchParameters }) {
--- a/src/types/widget.ts
+++ b/src/types/widget.ts
@@ -7,28 +7,19 @@ import {
 } from 'algoliasearch-helper';
 import { InstantSearch } from './instantsearch';
 
-export type InitOptions = {
-  instantSearchInstance: InstantSearch;
-  parent: Index | null;
-  uiState: UiState;
-  state: SearchParameters;
-  helper: Helper;
-  templatesConfig: object;
-  createURL(state: SearchParameters): string;
-};
-
 export type ScopedResult = {
   indexId: string;
   results: SearchResults;
   helper: Helper;
 };
 
-export type RenderOptions = {
+type SharedRenderOptions = {
   instantSearchInstance: InstantSearch;
+  parent: Index | null;
   templatesConfig: object;
-  results: SearchResults;
   scopedResults: ScopedResult[];
   state: SearchParameters;
+  renderState: IndexRenderState;
   helper: Helper;
   searchMetadata: {
     isSearchStalled: boolean;
@@ -36,12 +27,21 @@ export type RenderOptions = {
   createURL(state: SearchParameters): string;
 };
 
+export type InitOptions = SharedRenderOptions & {
+  uiState: UiState;
+  results: undefined;
+};
+
+export type RenderOptions = SharedRenderOptions & {
+  results: SearchResults;
+};
+
 export type DisposeOptions = {
   helper: Helper;
   state: SearchParameters;
 };
 
-export type WidgetStateOptions = {
+export type WidgetUiStateOptions = {
   searchParameters: SearchParameters;
   helper: Helper;
 };
@@ -119,6 +119,31 @@ export type UiState = {
   [indexId: string]: IndexUiState;
 };
 
+export type RenderState = {
+  [indexId: string]: IndexRenderState;
+};
+
+export type IndexRenderState = Partial<{
+  searchBox: WidgetRenderState<
+    {
+      query: string;
+      refine(query: string): void;
+      clear(): void;
+      isSearchStalled: boolean;
+    },
+    {
+      queryHook?(query: string, refine: (query: string) => void);
+    }
+  >;
+}>;
+
+type WidgetRenderState<
+  TWidgetRenderState,
+  TWidgetParams
+> = TWidgetRenderState & {
+  widgetParams: TWidgetParams;
+};
+
 /**
  * Widgets are the building blocks of InstantSearch.js. Any valid widget must
  * have at least a `render` or a `init` function.
@@ -169,6 +194,13 @@ export type Widget = {
    */
   dispose?(options: DisposeOptions): SearchParameters | void;
   /**
+   * Returns the render params to pass to the render function.
+   */
+  getWidgetRenderState?(
+    renderState: IndexRenderState,
+    widgetRenderStateOptions: InitOptions | RenderOptions
+  ): IndexRenderState;
+  /**
    * This function is required for a widget to be taken in account for routing.
    * It will derive a uiState for this widget based on the existing uiState and
    * the search parameters applied.
@@ -177,7 +209,7 @@ export type Widget = {
    */
   getWidgetUiState?(
     uiState: IndexUiState,
-    widgetStateOptions: WidgetStateOptions
+    widgetUiStateOptions: WidgetUiStateOptions
   ): IndexUiState;
   /**
    * This function is required for a widget to be taken in account for routing.
@@ -189,7 +221,7 @@ export type Widget = {
    */
   getWidgetState?(
     uiState: IndexUiState,
-    widgetStateOptions: WidgetStateOptions
+    widgetStateOptions: WidgetUiStateOptions
   ): IndexUiState;
   /**
    * This function is required for a widget to behave correctly when a URL is
--- a/src/types/widget.ts
+++ b/src/types/widget.ts
@@ -29,7 +29,6 @@ type SharedRenderOptions = {
 
 export type InitOptions = SharedRenderOptions & {
   uiState: UiState;
-  results: undefined;
 };
 
 export type RenderOptions = SharedRenderOptions & {
--- a/src/widgets/index/index.ts
+++ b/src/widgets/index/index.ts
@@ -277,7 +277,6 @@ const index = (props: IndexProps): Index => {
                 renderState: localInstantSearchInstance!.renderState,
                 templatesConfig: localInstantSearchInstance!.templatesConfig,
                 createURL,
-                results: undefined,
                 scopedResults: [],
                 searchMetadata: {
                   isSearchStalled: localInstantSearchInstance!._isSearchStalled,
@@ -304,7 +303,6 @@ const index = (props: IndexProps): Index => {
               renderState: localInstantSearchInstance!.renderState,
               templatesConfig: localInstantSearchInstance!.templatesConfig,
               createURL,
-              results: undefined,
               scopedResults: [],
               searchMetadata: {
                 isSearchStalled: localInstantSearchInstance!._isSearchStalled,
@@ -482,7 +480,6 @@ const index = (props: IndexProps): Index => {
               renderState: instantSearchInstance.renderState,
               templatesConfig: instantSearchInstance.templatesConfig,
               createURL,
-              results: undefined,
               scopedResults: [],
               searchMetadata: {
                 isSearchStalled: instantSearchInstance._isSearchStalled,
@@ -514,7 +511,6 @@ const index = (props: IndexProps): Index => {
             templatesConfig: instantSearchInstance.templatesConfig,
             renderState: instantSearchInstance.renderState,
             createURL,
-            results: undefined,
             scopedResults: [],
             searchMetadata: {
               isSearchStalled: instantSearchInstance._isSearchStalled,
--- a/src/widgets/index/index.ts
+++ b/src/widgets/index/index.ts
@@ -264,6 +264,9 @@ const index = (props: IndexProps): Index => {
           _uiState: localUiState,
         });
 
+        // We compute the render state before calling `init` in a separate loop
+        // to construct the whole render state object that is then passed to
+        // `init`.
         widgets.forEach(widget => {
           if (widget.getWidgetRenderState) {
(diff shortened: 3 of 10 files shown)


src/connectors/search-box/connectSearchBox.js
```diff
@@ -139,6 +133,19 @@ export default function connectSearchBox(renderFn, unmountFn = noop) {
         return state.setQueryParameter('query', undefined);
       },
 
+      getWidgetRenderState(renderState, { helper, searchMetadata }) {
+        return {
+          ...renderState,
+          searchBox: {
```
Haroenv, line 139: I know we agreed on spreading for consistency, but now it seems a little odd to me. I wonder if we could have the key being calculated based on the $$type and return directly the object containing query? Not sure yet what's better, but this seems quite indirect

Haroenv: The reason I'm thinking this, is because the hook is get**Widget**RenderState, not `getCurrentRenderState`

src/types/widget.ts
```diff
@@ -7,41 +7,41 @@ import {
...
+  renderState: IndexRenderState;
   helper: Helper;
   searchMetadata: {
     isSearchStalled: boolean;
   };
   createURL(state: SearchParameters): string;
 };
 
+export type InitOptions = SharedRenderOptions & {
+  uiState: UiState;
```
Haroenv, line 31: should we call render with uiState too? Seems simpler

src/connectors/search-box/connectSearchBox.js
francoischalifour: > I wonder if we could have the key being calculated based on the $$type

That's what I was thinking initially but in the end it's way more shady. That's an implicit API.

> The reason I'm thinking this, is because the hook is get**Widget**RenderState, not `getCurrentRenderState`

We've got the same behavior for `getWidgetUiState` which returns the whole index UI state, not the widget UI state.

src/connectors/search-box/connectSearchBox.js
```diff
@@ -107,27 +107,21 @@ export default function connectSearchBox(renderFn, unmountFn = noop) {
...
         this._clear = clear(helper);
 
         renderFn(
           {
-            query: helper.state.query || '',
-            refine: this._refine,
-            clear: this._cachedClear,
-            widgetParams,
+            ...this.getWidgetRenderState(renderState, renderOptions).searchBox,
             instantSearchInstance,
```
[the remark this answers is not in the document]
francoischalifour: Nope, very few connectors call `renderFn` with all these information.

src/types/widget.ts
francoischalifour: I didn't because we made the mistake to call this `uiState` instead of `initialUiState`, which would be extremely misleading when passed to `render()`. You would expect this `uiState` to be up-to-date.

src/connectors/search-box/connectSearchBox.js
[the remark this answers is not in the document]
Haroenv: I feel like this iteration on all renderFn is a moment to standardise that, wdyt?

src/types/widget.ts
Haroenv: You're right, I forgot this was only initial

src/connectors/search-box/connectSearchBox.js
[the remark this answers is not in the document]
francoischalifour: Perhaps as a last set of PRs in the "Render State" scope, but let's not introduce too many changes right now.

src/types/widget.ts
```diff
@@ -7,41 +7,41 @@ import {
...
+  results: undefined;
```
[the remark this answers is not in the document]
francoischalifour: So that we have close to argument-parity in `init` and `render`. Do you think that's too explicit? (note that this is internal code)

src/connectors/search-box/connectSearchBox.js
```diff
@@ -139,6 +133,19 @@ export default function connectSearchBox(renderFn, unmountFn = noop) {
...
+          ...renderState,
+          searchBox: {
+            query: helper.state.query || '',
+            refine: this._refine,
+            clear: this._cachedClear,
+            widgetParams,
+            isSearchStalled: searchMetadata.isSearchStalled,
+          },
+        };
+      },
```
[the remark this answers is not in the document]
francoischalifour: I didn't want to rush that decision now because it might be too naive. There are probably things that we'll miss. E.g., inability for a widget to be responsible of multiple UI states—that could be the case if `infiniteHits` starts controlling another UI state than `page`. Besides, that'd be a breaking change.

src/widgets/index/index.ts
```diff
@@ -264,16 +264,51 @@ const index = (props: IndexProps): Index => {
           _uiState: localUiState,
         });
 
+        localWidgets.forEach(widget => {
```
yannickcr, line 267: Is it normal to do it on `localWidgets` while we execute the `init` method only on `widgets` just after ?

src/widgets/index/index.ts
```diff
@@ -434,6 +469,35 @@ const index = (props: IndexProps): Index => {
         helper!.lastResults = results;
       });
 
+      localWidgets.forEach(widget => {
```
yannickcr, line 472: You can merge this loop with the following one.

src/widgets/index/index.ts
```diff
@@ -264,16 +264,51 @@ const index = (props: IndexProps): Index => {
           _uiState: localUiState,
         });
 
+        localWidgets.forEach(widget => {
+          if (widget.getWidgetRenderState) {
+            const widgetRenderState = widget.getWidgetRenderState(
+              localInstantSearchInstance!.renderState[this.getIndexId()] || {},
+              {
```
yannickcr, line 271: You could store these parameters in variables then re-use them in the `init`, `render` and `getWidgetRenderState` calls.

It would simplify the code (no more repeated code) and make it more readable imho:

```js
const renderState = instantSearchInstance.renderState[this.getIndexId()] || {};

const initOptions = {
    uiState,
    helper: this.getHelper()!,
    parent: this,
    instantSearchInstance,
    state: helper!.state,
    renderState: instantSearchInstance.renderState,
    templatesConfig: instantSearchInstance.templatesConfig,
    createURL,
    results: undefined,
    scopedResults: [],
    searchMetadata: {
        isSearchStalled: instantSearchInstance._isSearchStalled,
    },
};

const widgetRenderState = widget.getWidgetRenderState(renderState, initOptions);
```

src/widgets/index/index.ts
```diff
@@ -569,3 +673,21 @@ const index = (props: IndexProps): Index => {
 };
 
 export default index;
+
+function storeRenderState({
```
yannickcr, line 677: I think it would be better to have it as a method in `index`, it would simplify the signature (no need to pass `instantSearchInstance` and `parent`).

src/types/widget.ts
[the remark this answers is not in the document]
yannickcr: Yeah I don't really see any advantage in doing that, I find it confusing for the reader.

src/connectors/search-box/connectSearchBox.js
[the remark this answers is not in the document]
yannickcr: The point of the discussion was that if we do the merge ourself it would not be a breaking change and it should not have any impact on the widgets that return a full UI State in all cases.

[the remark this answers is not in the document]
francoischalifour: To my first point, I don't think this is the right time to do this, and we'll need proper testing to decide if we want to do this at some point.

src/widgets/index/index.ts
francoischalifour: Good catch.

src/widgets/index/index.ts
francoischalifour: Nope, otherwise we won't get the full render state passed to the widget when they `render`. The goal of having this as a loop before rendering is to do a full pass to construct the render state.

src/widgets/index/index.ts
francoischalifour: The params values are not the same during `init` and `render` so we cannot share them between these two methods.

Reusing the same params between `getWidgetRenderState` and `init|render` might leads to bug because they could expose stale values.

src/widgets/index/index.ts
francoischalifour: We call it only twice (and will call it only one time once when remove the `init` method). Besides, it would make it available on the `index` widget scope, which we want to avoid.

src/widgets/index/index.ts
yannickcr: Ok, since this particularity is not explicit in the code could you add a comment to explain it?

src/widgets/index/index.ts
yannickcr: Not sharing between `init` and `render` but between `getWidgetRenderState`/`init` and then `getWidgetRenderState`/`render`.

For the stale value problem, the 2 calls are synchronous and the only thing that is updated between them is `renderState`, it could be passed separately from the option object to avoid that.

src/widgets/index/index.ts
yannickcr: If you do not want to expose it you can set it as a private function, like `createURL`.

src/connectors/search-box/connectSearchBox.js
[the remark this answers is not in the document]
yannickcr: It feels like a missed opportunity here but ok.

src/widgets/index/index.ts
eunjae-lee: There's a lot of properties going on in that function call, but I think similar code "twice" is still okay.

francoischalifour: Yeah, since our internal state is based on mutation, it's too dangerous to reuse the same reference in the two function calls. I just tried and the current tests fail. It's safer to duplicate the code here.

src/widgets/index/index.ts
francoischalifour: Right, but we still need to pass the `instantSearchInstance` reference from `init` and `render` to mutate it, so it won't behave the same if we rely on `localInstantSearchInstance`.

src/types/widget.ts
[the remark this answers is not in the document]
francoischalifour: e16eda5b85e2e29907bcf10ec240e5bfb256b08a

src/widgets/index/index.ts
francoischalifour: 7ebd868b3544439590abf57e04a10ffbbd50ff03

src/widgets/index/index.ts
francoischalifour: b84947b56f2238a4c5a3f131ce4c9205f94cae0c

src/widgets/index/index.ts
yannickcr: Still winning one parameter :smile:  , and it is more coherent with what we are already doing in this module.

src/widgets/index/index.ts
yannickcr: Ok, I find this very concerning that it is "dangerous" to do this. But this is more relevant to bigger problems on how we manage the state.

src/types/widget.ts
```diff
@@ -119,6 +118,31 @@ export type UiState = {
...
+  searchBox: WidgetRenderState<
+    {
+      query: string;
+      refine(query: string): void;
+      clear(): void;
+      isSearchStalled: boolean;
+    },
+    {
+      queryHook?(query: string, refine: (query: string) => void);
+    }
```
Haroenv, line 135: Will we need to repeat the types here? Could they be imported from the widgets maybe?

src/widgets/index/index.ts
```diff
@@ -491,13 +595,15 @@ const index = (props: IndexProps): Index => {
         // be delayed. The render is triggered for the complete tree but some parts do
         // not have results yet.
 
-        if (widget.render && derivedHelper!.lastResults) {
+        if (widget.render) {
```
Haroenv, line 598: This might be a breaking change, since it's possible  `render` is called without results, which might throw errors in the widgets themselves.

src/widgets/index/index.ts
Haroenv: I agree with François, we want to avoid exposing a public API we didn't think about, private isn't really private

src/types/widget.ts
francoischalifour: SearchBox is not migrated to TypeScript. We'll reuse `*RendererOptions` for other widgets here.

src/widgets/index/index.ts
francoischalifour: There's no behavior change, it's taken care of here now:

https://github.com/algolia/instantsearch.js/blob/a8d42cdfe3e767887fdb52172fc1adadc6ce000d/src/widgets/index/index.ts#L558-L560

PR #4459, opened 2020-07-30
fix(ts): correct dummy v4 client

Haroenv:
**Summary**

Client v3 also has `addAlgoliaAgent`, and TypeScript doesn't seem strict enough to infer that the signatures don't match, unless a "marker" key is added.

**Result**

- https://github.com/algolia/algoliasearch-helper-js/pull/789
- same, but here

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   ],
   "dependencies": {
     "@types/googlemaps": "^3.39.6",
-    "algoliasearch-helper": "^3.2.1",
+    "algoliasearch-helper": "^3.2.2",
     "classnames": "^2.2.5",
     "events": "^1.1.0",
     "hogan.js": "^3.0.2",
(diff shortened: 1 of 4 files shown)


src/types/algoliasearch.ts
```diff
@@ -38,6 +39,8 @@ export type MultiResponse<THit = any> = {
   results: Array<SearchResponse<THit>>;
 };
 
+// ok?
+
```
Haroenv, line 43:
```suggestion
```

src/types/algoliasearch.ts
```diff
@@ -20,6 +20,7 @@ import {
 
 type DummySearchClientV4 = {
   readonly addAlgoliaAgent: (segment: string, version?: string) => void;
```
Haroenv, line 22:
```suggestion
```

package.json
```diff
@@ -92,7 +92,7 @@
     "@wdio/selenium-standalone-service": "5.16.5",
     "@wdio/spec-reporter": "5.16.5",
     "@wdio/static-server-service": "5.16.5",
-    "algoliasearch": "4.1.0",
+    "algoliasearch": "3.35.1",
```
Haroenv, line 95: needs to be 4.3.1

PR #4463, opened 2020-07-31
feat(insights): implement sendEvent in all the connectors (3/4)

eunjae-lee:
## Summary

This PR implements `sendEvent` in the following connectors and updates the existing test cases accordingly:

- connectAutocomplete
- connectGeoSearch
- connectHierarchicalMenu
- connectBreadcrumb
- connectMenu
- connectNumericMenu
- connectRange
- connectRatingMenu
- connectToggleRefinement 

**Stacked PRs**
1. [introduce `insights` middleware](https://github.com/algolia/instantsearch.js/pull/4446)
2. [send events from hits and refinementList](https://github.com/algolia/instantsearch.js/pull/4456)
3. [implement sendEvent in all the connectors](https://github.com/algolia/instantsearch.js/pull/4463) ← current
4. [add tests](https://github.com/algolia/instantsearch.js/pull/4464)

## Result

Now the connectors send event when refining, and expose `sendEvent` at `init` and `render` method.

PR #4464, opened 2020-07-31
feat(insights): add tests (4/4)

eunjae-lee:
## Summary

This PR adds test cases for all the related stacked PRs.

- [x] test if `insights` middleware correct sets up userToken.
- [x] test if `insights` middleware send event via insightsClient.
- [x] test if `insights` middleware calls `onEvent` when it's given.
- [x] test if `sendEvent` correctly called when widgets refine.
- [x] test if `sendEvent` exposed to components works fine.
- [x] test if `createSendEvent***` helpers work fine.
- [x] test if `bindEvent` exposed to Hits and InfiniteHits works fine with its listener.

**Stacked PRs**
1. [introduce `insights` middleware](https://github.com/algolia/instantsearch.js/pull/4446)
2. [send events from hits and refinementList](https://github.com/algolia/instantsearch.js/pull/4456)
4. [add tests](https://github.com/algolia/instantsearch.js/pull/4464) ← current

## Next Steps

**1. Update the documentation**

**2. When this feature is released and become stable in the future, we need to deprecate**

- `insightsClient` property in instantSearchInstance
- public usage of `getInsightsAnonymousUserToken()` (now the middleware uses it internally)
- the old way of attaching events on hits

```js
           <button ${instantsearch.insights('clickedObjectIDsAfterSearch', {
             eventName: 'click-result',
             objectIDs: [hit.objectID],
           })}>
             Click event
           </button>
           <button ${instantsearch.insights('convertedObjectIDsAfterSearch', {
             eventName: 'click-result',
             objectIDs: [hit.objectID],
           })}>
             Conversion event
           </button>
```

and all the helpers and listeners around that function.


package.json
```diff
@@ -142,7 +142,7 @@
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "64 kB"
+      "maxSize": "66 kB"
```
Haroenv, line 145: is this still 66KB after the PRs I've made, or have you not yet rebased master?

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -566,4 +566,155 @@ search.addWidgets([
       );
     });
   });
+
+  describe('sendEvent', () => {
+    let firstIndexHits;
+    let secondIndexHits;
+    let instantSearchInstance;
+    let render;
+    beforeEach(() => {
```
Haroenv, line 575: this style is ok, but in other files / tests we have more used function like `createInitialisedWidget`, you could have a  similar `createRenderedWidget`?

src/connectors/hits/__tests__/connectHits-test.ts
```diff
@@ -550,4 +550,165 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
       expect(nextState.highlightPostTag).toBe('</mark>');
     });
   });
+
+  describe('sendEvent & bindEvent', () => {
```
Haroenv, line 554:
```suggestion
  describe('insights', () => {
```

src/connectors/hits/__tests__/connectHits-test.ts
```diff
@@ -550,4 +550,165 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
...
+        createRenderOptions({
+          results,
+          state: helper.state,
+          helper,
+        })
+      );
+    });
+
+    describe('sendEvent', () => {
+      it('sends view event when hits are rendered', () => {
```
Haroenv, line 601: can you add a test that if some other widget causes a rerender, without changing search parameters, that it doesn't get called twice?

src/lib/utils/__tests__/createSendEventForFacet-test.ts
```diff
@@ -0,0 +1,104 @@
...
+import { InstantSearch, SearchClient } from '../../../types';
+
+jest.mock('../isFacetRefined', () => jest.fn());
+
+import isFacetRefined from '../isFacetRefined';
+
+const instantSearchInstance = {} as InstantSearch;
+let sendEvent;
+let helper;
+beforeEach(() => {
```
Haroenv, line 12: would like to see this as a constructor instead of beforeEach too (easier to see where variables are from, and avoids mutation / global variables everywhere)

src/lib/utils/createSendEventForHits.ts
```diff
@@ -1,5 +1,5 @@
 import { InstantSearch, Hit } from '../../types';
-import { InsightsEvent } from '../../middleware/insights';
+import { InsightsEvent } from '../../middlewares/insights';
```
Haroenv, line 2: how come this change is in this PR and not a previous one?

src/middlewares/insights.ts
```diff
@@ -83,10 +83,16 @@ aa('setUserToken', 'your-user-token');
...
+          // ['setUserToken', 'my-user-token'] gets stored in `aa.queue`.
+          // Whenever `search-insights` is finally loaded, it will process the queue.
+          //
+          // But the reason why we handle it here is
+          // (1) At this point, even though `search-insights` is not loaded yet,
+          // we still want to read the token from the queue.
+          // Otherwise, the first search call will be fired without the token.
+          // (2) Or, user could use a customized client of `search-insights`,
+          // for example, by using `createInsightsClient` function.
+          // Then `processQueue` might not be called. But we still want to read the token from the queue.
```
Haroenv, line 95: in that case, there's no queue, right? I think (2) should be removed

test/mock/createInsightsClient.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (!instance[methodName]) {
+      throw new Error(`${methodName} doesn't exist in this mocked instance`);
+    }
+    instance[methodName](...args);
+  };
+}
+
+export function createInsightsUmdVersion() {
+  const globalObject: any = {};
+  globalObject.aa = (...args) => {
```
Haroenv, line 50: can this not be local scope in this mock?

test/mock/createInstantSearch.ts
```diff
@@ -7,7 +7,7 @@ import defer from '../../src/lib/utils/defer';
 export const createInstantSearch = (
   args: Partial<InstantSearch> = {}
 ): InstantSearch => {
-  const searchClient = createSearchClient();
+  const searchClient = args.client || createSearchClient();
```
Haroenv, line 10:
```suggestion
  const searchClient = args.searchClient || createSearchClient();
```

Haroenv: although it's also possible to do this with default in the destructuring, like in the following line

Haroenv (commented):
🆒

package.json
eunjae-lee: it was outdated https://github.com/algolia/instantsearch.js/pull/4464/commits/4c41441d628e30c130e66c71bd055b3824891d44

test/mock/createInsightsClient.ts
eunjae-lee: I guess it won't matter much, but do you see any reason for it not to be locally scoped?

src/middlewares/insights.ts
eunjae-lee: Ah you're correct!

src/lib/utils/createSendEventForHits.ts
eunjae-lee: I don't remember well but I guess it must be one of the typical mistakes that happens with stacked PRs.

test/mock/createInstantSearch.ts
eunjae-lee: but `searchClient` doesn't exist in `args`.
I've updated it to be destructured from `args` though: https://github.com/algolia/instantsearch.js/pull/4464/commits/8300a1db7678224ab1509013da817d8500e6cc2e

src/lib/utils/__tests__/createSendEventForFacet-test.ts
eunjae-lee: thanks for the reminder: https://github.com/algolia/instantsearch.js/pull/4464/commits/669501559bff1fda2a340c02658b87fe574053a9

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4464/commits/669501559bff1fda2a340c02658b87fe574053a9

src/connectors/hits/__tests__/connectHits-test.ts
eunjae-lee: @Haroenv Can you give me an idea? What kind of situation can it be?

Initially I thought of adding a `poweredBy` widget. But adding a new widget calls `scheduleSearch` in the end.

Is there a flow where InstantSearch doesn't trigger a search but just trigger a rerender of all widgets?

Haroenv: for example adding a widget should indeed not cause a new view event, since the parameters are the same, but since we do a render (do we?), it could send the view event again.

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -978,4 +978,177 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
       expect(actual.page).toEqual(2);
     });
   });
+
+  describe('sendEvent & bindEvent', () => {
```
Haroenv, line 982: maybe call this `insights` in all connectors? It's slightly inconsistent now

Haroenv (approved):
tests look fine to me now, would just make the describe titles consistent :)

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
eunjae-lee: good call. done https://github.com/algolia/instantsearch.js/pull/4464/commits/d228af5a6ef1b258712c7a4762babd648bbc56b8

src/connectors/hits/__tests__/connectHits-test.ts
eunjae-lee: @Haroenv Oops. I missed this and already merged this PR. I can address it somewhere else.

> it could send the view event again.

Then we should somehow compare the search parameters and send view events only when they have changed. Is it similar to what you have in your mind?

PR #4466, opened 2020-08-03
feat(autocomplete): implement `getWidgetRenderState`

francoischalifour:
This implements the `getWidgetRenderState` widget lifecycle hook in `autocomplete`.

⚠️ The following line throws a [TypeScript error](https://app.circleci.com/pipelines/github/algolia/instantsearch.js/4567/workflows/9954690f-500b-42af-8d85-4c986dc45b1e/jobs/19424/parallel-runs/0/steps/0-107) that I plan to fix in a later PR to move forward with this widget API (any help is welcome):

https://github.com/algolia/instantsearch.js/blob/3639991ce5f32199710d1bc9d685b88793465178/src/types/widget.ts#L147-L152

### DIFF
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -102,24 +102,36 @@ search.addWidgets([
     return {
       $$type: 'ais.autocomplete',
 
-      init({ instantSearchInstance, helper }) {
+      init(initOptions) {
+        const { helper, renderState, instantSearchInstance } = initOptions;
         connectorState.refine = (query: string) => {
           helper.setQuery(query).search();
         };
 
         renderFn(
           {
-            widgetParams,
-            currentRefinement: helper.state.query || '',
-            indices: [],
-            refine: connectorState.refine,
+            ...this.getWidgetRenderState!(renderState, initOptions)
+              .autocomplete!,
             instantSearchInstance,
           },
           true
         );
       },
 
-      render({ helper, scopedResults, instantSearchInstance }) {
+      render(renderOptions) {
+        const { renderState, instantSearchInstance } = renderOptions;
+
+        renderFn(
+          {
+            ...this.getWidgetRenderState!(renderState, renderOptions)
+              .autocomplete!,
+            instantSearchInstance,
+          },
+          false
+        );
+      },
+
+      getWidgetRenderState(renderState, { helper, scopedResults }) {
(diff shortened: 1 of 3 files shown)


src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -102,24 +102,36 @@ search.addWidgets([
...
         };
 
         renderFn(
           {
-            widgetParams,
-            currentRefinement: helper.state.query || '',
-            indices: [],
-            refine: connectorState.refine,
+            ...this.getWidgetRenderState!(renderState, initOptions)
+              .autocomplete!,
```
Haroenv, line 114: does this gWRS not always have `autocomplete` in it?

francoischalifour: It doesn't have `autocomplete` until you mount an `autocomplete` widget (just like how UI state works).

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -102,24 +102,36 @@ search.addWidgets([
...
+            ...this.getWidgetRenderState!(renderState, renderOptions)
+              .autocomplete!,
+            instantSearchInstance,
+          },
+          false
+        );
+      },
+
+      getWidgetRenderState(renderState, { helper, scopedResults }) {
         const indices = scopedResults.map(scopedResult => {
```
Haroenv, line 135: I think scopedResults might be undefined in init (or does it get called with [])?

src/connectors/autocomplete/connectAutocomplete.ts
Haroenv: Maybe we should try to solve this later, but I'm not suer how you'd write your own widget in TS if it's not added to RenderState.

A possibility is that addWidgets has a generic, which gets filled by the widget itself, and returns a new InstantSearch type, which does have the new widget. I'm not sure, since that could be too dynamic to work in ts though

src/connectors/autocomplete/connectAutocomplete.ts
francoischalifour: It's `[]` in init.

src/connectors/autocomplete/connectAutocomplete.ts
francoischalifour: Yep, we already had this discussion with UI state and TypeScript. Let's think about it later!

PR #4467, opened 2020-08-04
feat(breadcrumb): implement `getWidgetRenderState`

francoischalifour:
This implements the `getWidgetRenderState` widget lifecycle hook in `breadcrumb`.

### DIFF
--- a/src/connectors/breadcrumb/connectBreadcrumb.ts
+++ b/src/connectors/breadcrumb/connectBreadcrumb.ts
@@ -108,7 +108,9 @@ const connectBreadcrumb: BreadcrumbConnector = function connectBreadcrumb(
     return {
       $$type: 'ais.breadcrumb',
 
-      init({ createURL, helper, instantSearchInstance }) {
+      init(initOptions) {
+        const { helper, createURL } = initOptions;
+
         connectorState.createURL = facetValue => {
           if (!facetValue) {
             const breadcrumb = helper.getHierarchicalFacetBreadcrumb(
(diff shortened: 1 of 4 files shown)


src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -186,6 +175,38 @@ const connectBreadcrumb: BreadcrumbConnector = function connectBreadcrumb(
         unmountFn();
       },
 
+      getWidgetRenderState(renderState, { results, state }) {
```
francoischalifour, line 178: @yannickcr Bouncing back on https://github.com/algolia/instantsearch.js/pull/4457#discussion_r461592742 — this is why it's convenient to be explicit about `results: undefined` in TypeScript, otherwise we wouldn't be able to use `results` here.

stories/breadcrumb.stories.js
```diff
@@ -29,7 +29,7 @@ storiesOf('Metadata/Breadcrumb', module)
           virtualHierarchicalMenu(),
 
           breadcrumb({
-            container: breadcrumb,
+            container: breadcrumbDiv,
```
francoischalifour, line 32: This currently breaks in the [prod deployment](http://instantsearchjs.netlify.com/stories?path=/story/metadata-breadcrumb--default).

src/types/widget.ts
```diff
@@ -142,13 +147,18 @@ export type IndexRenderState = Partial<{
...
+  >;
 }>;
 
 type WidgetRenderState<
   TWidgetRenderState,
+  // @ts-ignore
   TWidgetParams
 > = TWidgetRenderState & {
-  widgetParams: TWidgetParams;
+  widgetParams: any; // @TODO type as TWidgetParams
```
francoischalifour, line 161: I'll take care of this in a future PR.

src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -186,6 +175,38 @@ const connectBreadcrumb: BreadcrumbConnector = function connectBreadcrumb(
...
+
+        const items = getItems();
+
+        return {
+          ...renderState,
+          breadcrumb: {
+            canRefine: items.length > 0,
+            createURL: connectorState.createURL,
+            items,
+            refine: connectorState.refine,
```
Haroenv, line 204: createURL & refine aren't cached like in clearRefinement, is that expected?

francoischalifour: I kept it like it was in this PR, but it's not expected. There are multiple connectors that don't keep the same reference. We should fix this and automate some tests run on all widgets.

PR #4468, opened 2020-08-04
feat(clearRefinements): implement `getWidgetRenderState`

francoischalifour:
This implements the `getWidgetRenderState` widget lifecycle hook in `clearRefinements`.

### DIFF
--- a/src/connectors/clear-refinements/connectClearRefinements.ts
+++ b/src/connectors/clear-refinements/connectClearRefinements.ts
@@ -85,20 +85,27 @@ const connectClearRefinements: ClearRefinementsConnector = function connectClear
     return {
       $$type: 'ais.clearRefinements',
 
-      init({ instantSearchInstance }) {
+      init(initOptions) {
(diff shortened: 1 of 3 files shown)


src/connectors/clear-refinements/connectClearRefinements.ts
```diff
@@ -154,6 +157,33 @@ const connectClearRefinements: ClearRefinementsConnector = function connectClear
       dispose() {
         unmountFn();
       },
+
+      getWidgetRenderState(renderState, { scopedResults }) {
+        const attributesToClear = scopedResults.reduce<
```
Haroenv, line 162: this gets calculated both here and in `refine` itself still. Should we call it in the beginning of `refine` and use its value, instead of computing it again instead?

francoischalifour: Addressed in bfbe04dcf11a0b3af882d30830bb933b7d9a82fc.

PR #4469, opened 2020-08-04
feat(configure): implement `getWidgetRenderState`

francoischalifour:
This implements the `getWidgetRenderState` widget lifecycle hook in `configure`.

### DIFF
--- a/src/connectors/configure/connectConfigure.ts
+++ b/src/connectors/configure/connectConfigure.ts
@@ -99,25 +99,28 @@ const connectConfigure: ConfigureConnector = function connectConfigure(
     return {
       $$type: 'ais.configure',
 
-      init({ instantSearchInstance, helper }) {
+      init(initOptions) {
+        const { renderState, helper, instantSearchInstance } = initOptions;
+
         connectorState.refine = refine(helper);
 
         renderFn(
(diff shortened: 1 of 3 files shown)


src/connectors/configure/connectConfigure.ts
```diff
@@ -129,6 +132,16 @@ const connectConfigure: ConfigureConnector = function connectConfigure(
         return getInitialSearchParameters(state, widgetParams);
       },
 
+      getWidgetRenderState(renderState) {
+        return {
+          ...renderState,
+          configure: {
```
Haroenv, line 138: you can have multiple configure widgets in the same tree right? How does this behave? Not sure how to get a key for the configure though. The refine will be the same for them, but widgetParams will be different.

francoischalifour: Good call—this is why accumulating is more powerful than naively merging render states at the index level.

I'd suggest taking care of shared render state in future PRs, wdyt?

Haroenv: fine for me if you write it down somewhere so we won't need to go through all PRs to find the unresolved questions afterwards :)

PR #4471, opened 2020-08-06
feat(hierarchicalMenu): implement `getWidgetRenderState`

francoischalifour:
This implements the `getWidgetRenderState` widget lifecycle hook in `hierarchicalMenu`.

### DIFF
--- a/src/connectors/hierarchical-menu/connectHierarchicalMenu.js
+++ b/src/connectors/hierarchical-menu/connectHierarchicalMenu.js
@@ -113,93 +113,45 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
         return this.isShowingMore ? showMoreLimit : limit;
       },
 
-      init({ helper, createURL, instantSearchInstance }) {
+      init(initOptions) {
+        const { helper, renderState, instantSearchInstance } = initOptions;
+
         this.cachedToggleShowMore = this.cachedToggleShowMore.bind(this);
         this._refine = function(facetValue) {
           helper.toggleRefinement(hierarchicalFacetName, facetValue).search();
         };
 
-        // Bind createURL to this specific attribute
-        function _createURL(facetValue) {
-          return createURL(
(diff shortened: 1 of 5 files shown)


src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -113,93 +113,45 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
...
-        // widget has requested more values (maxValuesPerFacet > getLimit()).
-        // Because this is used for making the search of facets unable or not, it is important
-        // to be conservative here.
-        const hasExhaustiveItems =
-          maxValuesPerFacetConfig > currentLimit
-            ? facetValues.length <= currentLimit
-            : facetValues.length < currentLimit;
+        const { renderState, instantSearchInstance } = renderOptions;
 
         this.toggleShowMore = this.createToggleShowMore(renderOptions);
```
Haroenv, line 148: what is this function used for? It seems to be using `cachedToggleShowMore` in the other occurrences

src/connectors/configure/connectConfigure.ts
```diff
@@ -99,25 +99,28 @@ const connectConfigure: ConfigureConnector = function connectConfigure(
     return {
       $$type: 'ais.configure',
 
-      init({ instantSearchInstance, helper }) {
+      init(initOptions) {
+        const { renderState, helper, instantSearchInstance } = initOptions;
```
Haroenv, line 103: this seems to be from a different PR?

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
francoischalifour: That's a proxy to `cachedToggleShowMore` that keeps the same function reference across renders. It was introduced a long time ago.

Haroenv: I see now, thanks!

PR #4476, opened 2020-08-14
fix(typescript): export correct types

Haroenv:
**Summary**

This PR fixes some miscellaneous typescript issues discovered in angular instantsearch

**Result**

- if JS client v3 is used, it's imperative that the ts-ignore comments stay present, since not the same exports are used. Solution is to double-annotate to keep the multi-line version in exposed to the d.ts (see e.g. https://github.com/microsoft/TypeScript/issues/39520)
- similar in places.js
- when updating TypeScript, I found that api-extractor can't deal with `import()` types, so those have to be transformed
- fix stricter behaviour of tranformItems in hitsPerPage (we send the user the "render" values, not the one they entered)
- SpeechRecognitionError is no longer exported in lib.d.ts, and is a generic Event now
- updated Typescript to 3.8.3. I've tried 3.9.2 as well, but that has a bug with `function()` (transforms to `(arg:any) => ;` instead of `(arg:any) => any;`). I didn't open a TS issue for it, since it's fixed in v4 which we can migrate to later

### DIFF
--- a/scripts/build/types.js
+++ b/scripts/build/types.js
@@ -6,7 +6,60 @@ const fs = require('fs');
 const path = require('path');
 const shell = require('shelljs');
 
+console.log(`Compiling definitions...`);
+
 shell.exec(`tsc -p tsconfig.declaration.json --outDir es/`);
+
+// replace block ts-ignore comments with line ones to support TS < 3.9
+shell.sed(
+  '-i',
+  /\*\* @ts-ignore/g,
+  '/ @ts-ignore',
+  path.join(__dirname, '../../es/**/*.d.ts')
+);
+
+// TypeScript uses `import()` types in very small files, but API extractor does not support that
+// We will hardcode the type definitions here, to avoid the build not passing.
+// "Error: The expression contains an import() type, which is not yet supported by API Extractor:"
+
+fs.writeFileSync(
+  path.join(
+    __dirname,
+    '../../es/connectors/infinite-hits/connectInfiniteHitsWithInsights.d.ts'
+  ),
+  `
+// edited by yarn build:types
+import { Connector } from '../../types';
+import {
+  InfiniteHitsRendererOptions,
+  InfiniteHitsConnectorParams,
+} from './connectInfiniteHits';
+
+declare const connectInfiniteHitsWithInsights: Connector<
+  InfiniteHitsRendererOptions,
+  InfiniteHitsConnectorParams
+>;
+export default connectInfiniteHitsWithInsights;
+`
+);
+
+fs.writeFileSync(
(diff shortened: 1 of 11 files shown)


src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -205,10 +205,10 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
         );
       },
 
-      // eslint-disable-next-line valid-jsdoc
       /**
-       * @param {Object} param0
-       * @param {import('algoliasearch-helper').SearchParameters} param0.state
+       * @param {Object} param0 cleanup arguments
+       * @param {any} param0.state current search parameters
+       * @returns {void}
```
Haroenv, line 211:
```suggestion
       * @returns {any}
```

scripts/build/types.js
```diff
@@ -6,8 +6,61 @@ const fs = require('fs');
...
+} from './connectHits';
+
+declare const connectHitsWithInsights: Connector<
+  HitsRendererOptions,
+  HitsConnectorParams
+>;
+export default connectHitsWithInsights;
+`
+);
+
```
yannickcr, line 63: This seems very fragile to me since any change in the source file will not be reflected in the declaration file if we forget to update this code accordingly.

I think it would be better to change the source file to explicitly import the needed types, like they made here https://github.com/phryneas/react-async/commit/70393af94ddd030de56bfcadcd17f00e6762dfa9 (source: https://github.com/microsoft/rushstack/issues/1050#issuecomment-531656759)

We would get something like this for `connectInfiniteHitsWithInsights.ts`:

```ts
import { withInsights } from '../../lib/insights';
import connectInfiniteHits, {
  InfiniteHitsRendererOptions,
  InfiniteHitsConnectorParams,
} from './connectInfiniteHits';
import { Connector } from '../../types';

/**
 * Due to https://github.com/microsoft/web-build-tools/issues/1050, we need
 * Connector<...> imported in this file, even though it is only used implicitly.
 * This _uses_ Connector<...> so it is not accidentally removed by someone.
 */
declare type ImportWorkaround = Connector<
  InfiniteHitsRendererOptions,
  InfiniteHitsConnectorParams
>;

const connectInfiniteHitsWithInsights = withInsights(connectInfiniteHits);

export default connectInfiniteHitsWithInsights;
```

Haroenv: That’s a good idea, I’ll try it

Haroenv:
At least in the playground code like this was working in v4, but not 3.9:

```js
/**
 * @param {function()} arg0
 */
function(arg0) {}
```

I couldn’t find an issue for it though, but since it’s fixed, there must have been one

PR #4480, opened 2020-08-18
fix(configureRelatedItems): support nested attributes

francoischalifour:
This brings support for nested attributes in the [`configureRelatedItems`](https://www.algolia.com/doc/api-reference/widgets/configure-related-items/js/) widget.

## Description

Previously, an attribute like `hierarchicalCategories.lvl0` was considered as non-existant because we were accessing it as `hit[hierarchicalCategories.lvl0]`. We now access it with our `getPropertyByPath` which correctly picks up the attribute.

## Next steps

Implement the same fix in React InstantSearch.

## Related

- [Slack message](https://algolia.slack.com/archives/C2XP6HBKQ/p1597715211006100)


francoischalifour:
@Haroenv Yep, I'll update the doc to add an example with a nested attribute.

PR #4483, opened 2020-08-19
fix(searchBox): pass "spellcheck" property correctly to input

Haroenv:
**Summary**

React and Preact differ slightly in how they pass attributes / properties. Therefore it seems like the correct value (although https://github.com/preactjs/preact/issues/485 says otherwise, that must have changed in preact X). This was causing `spellcheck` not to be applied.

see: https://github.com/preactjs/preact/blob/020d5e599b8b84745932f092a835f494b1dd2261/src/jsx.d.ts#L711

ref: https://discourse.algolia.com/t/how-to-add-spellcheck-property-to-search-box-widget/10741

**Result**

The `spellcheck` property is actually on the dom node for input.

I have checked, and the other properties on the input are present.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,7 +7,7 @@ module.exports = {
     'new-cap': [
       'error',
       {
-        capIsNewException: [
+        capIsNewExceptions: [
           'EXPERIMENTAL_use',
           'EXPERIMENTAL_connectConfigureRelatedItems',
           'EXPERIMENTAL_configureRelatedItems',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,25 +2,20 @@ module.exports = {
   extends: ['algolia', 'algolia/jest', 'algolia/react', 'algolia/typescript'],
   plugins: ['react-hooks'],
   rules: {
-    'no-param-reassign': 0,
-    'import/no-extraneous-dependencies': 0,
+    'no-param-reassign': 'off',
+    'import/no-extraneous-dependencies': 'off',
     'new-cap': [
       'error',
       {
-        capIsNewExceptions: [
-          'EXPERIMENTAL_use',
-          'EXPERIMENTAL_configureRelatedItems',
-          'EXPERIMENTAL_connectConfigureRelatedItems',
-        ],
+        capIsNewExceptionPattern: '^EXPERIMENTAL_*',
       },
     ],
-    'react/no-string-refs': 1,
-    'react/no-unknown-property': [2, { ignore: ['spellcheck'] }],
+    'react/no-string-refs': 'error',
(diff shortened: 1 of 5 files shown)


src/components/SearchBox/SearchBox.js
```diff
@@ -159,7 +159,7 @@ class SearchBox extends Component {
             autoComplete="off"
             autoCorrect="off"
             autoCapitalize="off"
-            spellCheck={false}
+            spellcheck={false}
```
francoischalifour, line 162: I had the same bug in the Preact export of DocSearch and the actual fix is to use:

```suggestion
            spellCheck="false"
```

Haroenv: interesting, my fix seems to work too, but I'll try your suggestion

src/components/SearchBox/__tests__/SearchBox-test.js
```diff
@@ -438,6 +438,20 @@ describe('SearchBox', () => {
       expect(mount(<SearchBox {...defaultProps} />)).toMatchSnapshot();
     });
 
+    test('sets autocorrect... properties', () => {
```
francoischalifour, line 441:
```suggestion
    test('sets search input properties', () => {
```

src/components/SearchBox/__tests__/SearchBox-test.js
```diff
@@ -438,6 +438,20 @@ describe('SearchBox', () => {
...
+      const res = render(
+        <SearchBox {...defaultProps} autofocus={true} query="sample query" />
+      );
+      const input = res.getByDisplayValue('sample query');
+
+      expect(input.getAttribute('autofocus')).toBe('true');
```
francoischalifour, line 447: We have helpers for DOM elements in this file. Did you try them?

```suggestion
      expect(input).toHaveAttribute('autofocus', 'true');
```

.eslintrc.js
```diff
@@ -2,24 +2,24 @@ module.exports = {
   extends: ['algolia', 'algolia/jest', 'algolia/react', 'algolia/typescript'],
   plugins: ['react-hooks'],
   rules: {
-    'no-param-reassign': 0,
-    'import/no-extraneous-dependencies': 0,
+    'no-param-reassign': 'off',
```
francoischalifour, line 5: Any particular reasons this file was changed?

Haroenv: I originally changed it for adding the `spellcheck` exception, and made the file consistent about 0/off then

src/components/SearchBox/__tests__/SearchBox-test.js
Haroenv: did not know about those, thanks

francoischalifour (approved):

PR #4487, opened 2020-08-24
fix(router): skip router write on duplicate entries

francoischalifour:
This fixes duplicate entries in the router that happen when `router.write` is called even though `uiState` hasn't changed.

## Explanation

We used to call the `write` method of the router at every state change. A state change happens at each UI interaction or **whenever a widget is mounted/unmounted**. The latter is problematic in component-based flavors of InstantSearch (Vue InstantSearch is the only one relying on InstantSearch 4 right now) because widgets are often added/removed.

The duplicate entry detection check was happening in the [history router itself](https://github.com/algolia/instantsearch.js/blob/3cf43c7187841cf961a0280307af1a5f7a4e8da7/src/lib/routers/history.ts#L131-L135), which meant that whenever users opt-out of this default router and implemented their own, they would get this issue.

This fix checks that `router.write` needs to be called in the router middleware itself. Since it's one level above the history router, or custom user implementation, it will get fixed no matter the router implementations. We check that the `routeState` is different from the last received to call `router.write`. We base the check on `routeState`, and not `uiState`, because users might plug their own logic in `stateToRoute` that returns a `routeState` that differs from `uiState`; the two do not necessarily change at the same time.

## Related

- #4392
- algolia/vue-instantsearch#841
- Closes algolia/vue-instantsearch#839


src/lib/__tests__/RoutingManager-test.ts
```diff
@@ -443,6 +437,84 @@ describe('RoutingManager', () => {
...
+      // Trigger change without UI state change
+      search.removeWidgets([fakeHitsPerPage]);
+
+      expect(router.write).toHaveBeenCalledTimes(1);
+
+      await runAllMicroTasks();
+
+      triggerChange = true;
+      // Trigger change without UI state change but with a route change
+      search.removeWidgets([fakeHitsPerPage]);
```
eunjae-lee, line 508: We removed `fakeHitsPerPage` before, and can we do it again here?

francoischalifour: It indeed makes more sense to remove another widget.

Addressed in a6b5dbd7efd8869a45ee884e4e1edf7fc51c732a.

src/lib/__tests__/RoutingManager-test.ts
```diff
@@ -443,6 +437,85 @@ describe('RoutingManager', () => {
...
+          router,
+          stateMapping,
+        },
+      });
+
+      const fakeSearchBox: any = createFakeSearchBox();
+      const fakeHitsPerPage1 = createFakeHitsPerPage();
+      const fakeHitsPerPage2 = createFakeHitsPerPage();
+
+      search.addWidgets([fakeSearchBox, fakeHitsPerPage1]);
```
eunjae-lee, line 484:
```suggestion
      search.addWidgets([fakeSearchBox, fakeHitsPerPage1, fakeHitsPerPage2]);
```

Haroenv (approved):
makes sense :)

PR #4489, opened 2020-08-25
chore: setup scheduled release

eunjae-lee:
**Summary**

This PR sets up scheduled release. And it additionally upgrades `ship.js` to the latest version and updated some configs due to the breaking changes from `ship.js`.

We can periodically release the newer version of the library. And I hope to get rid of unnecessary communication like:

> Can I release this today? Any blocker?

Instead, the pull request will be created automatically, and as reviewers, we can address any blocker in the pull request if there is any.

**Result**

Every Tuesday 09:00 UTC (11am Paris), `shipjs prepare` will run and create a release pull-request.
If there is a patch update and there is no `fix` commit, then it will skip.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,6 +124,21 @@ jobs:
           command: |
             yarn shipjs trigger
 
+  'prepare release':
+    <<: *defaults
+    steps:
+      - checkout
+      - run: *install_yarn_version
+      - restore_cache: *restore_yarn_cache
+      - run: *run_yarn_install
+      - save_cache: *save_yarn_cache
+      - run:
+          name: Prepare a pull request for next release
+          command: |
+            git config --global user.email "instantsearch-bot@algolia.com"
+            git config --global user.name "InstantSearch"
+            yarn run release --yes --no-browse
+
 workflows:
   version: 2
   ci:
@@ -135,3 +150,18 @@ workflows:
       - type check js (optional)
       - e2e tests
       - release if needed
+          filters:
+            branches:
+              only:
+                - master
+                - next
+  'scheduled release':
+    triggers:
+      - schedule:
+          cron: "0 9 * * 2"
+          filters:
+            branches:
+              only:
+                - master
+    jobs:
+      - prepare release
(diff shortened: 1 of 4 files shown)


ship.config.js
```diff
@@ -20,10 +19,9 @@ module.exports = {
   beforeCommitChanges: ({ exec }) => {
     exec('yarn doctoc');
   },
-  pullRequestTeamReviewer: ['instantsearch-for-websites'],
+  pullRequestTeamReviewers: ['instantsearch-for-websites'],
   buildCommand: ({ version }) =>
     `NODE_ENV=production VERSION=${version} yarn build`,
-  testCommandBeforeRelease: () => 'echo "No need to test again."',
```
eunjae-lee, line 26: The default behavior of the latest shipjs skips this test step, so we don't need this any more.

ship.config.js
```diff
@@ -33,10 +31,9 @@ module.exports = {
     }
   },
   slack: {
-    // disable slack notification for `prepared` and `releaseStart` lifecycle.
+    // disable slack notification for `prepared` lifecycle.
```
eunjae-lee, line 34: `releaseStart` lifecycle has been removed from shipjs.

ship.config.js
```diff
@@ -20,10 +19,9 @@ module.exports = {
   beforeCommitChanges: ({ exec }) => {
     exec('yarn doctoc');
   },
-  pullRequestTeamReviewer: ['instantsearch-for-websites'],
+  pullRequestTeamReviewers: ['instantsearch-for-websites'],
```
eunjae-lee, line 22: renamed in shipjs.

Haroenv (approved):

.circleci/config.yml
```diff
@@ -135,3 +150,18 @@ workflows:
       - type check js (optional)
       - e2e tests
       - release if needed
+          filters:
```
Haroenv, line 153: this probably needs to be part of `triggers` or a level lower?

eunjae-lee: https://github.com/algolia/shipjs/blob/main/.circleci/config.yml#L90-L96

It works like that.
I hate that I need to find an example every time I modify circleci config :(

Haroenv (approved):
as far as I can tell this is good

PR #4492, opened 2020-08-26
chore(refinementList): reuse templates from searchBox

Haroenv:
**Summary**

I had a look at the source map explorer, and found that the refinementList had a big defaultTemplates file.

Since the refinementList already uses searchBox, this doesn't add any new code if you're not using searchBox.

**Result**

Slightly smaller bundle size, but not so significant after gzip (.04KB)


eunjae-lee (approved):
Even though there's no big difference in bundle size, it's still good to have less svg in the code :)

PR #4496, opened 2020-08-28
feat(routing): introduce stateMappings.singleIndexQ for compatibility with v2

tnir:
**Summary**

In instantsearch.js v2, a developer was able to use `q` as a querystring parameter key by enabling `urlSync`.

This PR provides `q` querystring key instead of `query` with usage of `singleIndexQ`.
This feature helps developers update instantsearch.js from v2 to the latest (v4).

cf. singleIndex: https://www.algolia.com/doc/api-reference/widgets/single-index-state-mapping/js/

**Result**

Demo: http://715-update-instantsearch-to-4-7-1.178.62.207.141.nip.io/search/?q=helm

Demo code: https://gitlab.com/gitlab-org/gitlab-docs/-/merge_requests/1096

While documentation has not updated yet, it will be done when the feature is approved.

### DIFF
--- a/src/lib/stateMappings/__tests__/singleIndexQ-test.ts
+++ b/src/lib/stateMappings/__tests__/singleIndexQ-test.ts
@@ -1,0 +1,215 @@
+import singleIndexQStateMapping from '../singleIndexQ';
+
+describe('singleIndexQStateMapping', () => {
+  describe('stateToRoute', () => {
+    it('passes normal state through', () => {
+      const stateMapping = singleIndexQStateMapping('indexName');
+      const actual = stateMapping.stateToRoute({
+        indexName: {
+          query: 'zamboni',
+          refinementList: {
+            color: ['red'],
+          },
+        },
+      });
+
+      expect(actual).toEqual({
+        q: 'zamboni',
+        refinementList: {
+          color: ['red'],
+        },
+      });
+    });
+
+    it('removes configure', () => {
+      const stateMapping = singleIndexQStateMapping('indexName');
+      const actual = stateMapping.stateToRoute({
+        indexName: {
+          query: 'zamboni',
+          refinementList: {
+            color: ['red'],
+          },
+          configure: {
+            advancedSyntax: false,
+          },
+        },
+      });
+
+      expect(actual).toEqual({
+        q: 'zamboni',
+        refinementList: {
+          color: ['red'],
+        },
+      });
+    });
+
+    it('passes non-UiState through', () => {
+      const stateMapping = singleIndexQStateMapping('indexName');
+      const actual = stateMapping.stateToRoute({
+        indexName: {
+          query: 'zamboni',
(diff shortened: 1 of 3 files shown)


Haroenv:
Hi @tnir, thanks for your pull request! Is there a reason you prefer this being in the library, rather than in user land? v2 has been last released around two years ago, meaning that most people would already have migrated to this at the time v3 came out (if I'm not wrong). 

there are more changes than just `query` to `q` between v2's `urlSync`, so I'm not sure if this is fully correct. IIRC @damcou made a stateMapping which complied to both versions here: https://github.com/algolia/algoliasearch-magento-2/blob/53dadcc2d21251cc493ed35f6206f555df60fd03/view/frontend/web/internals/common.js#L543

Would love to hear your view on this!

tnir:
@Haroenv Thanks. Yes, many users already migrated to v4 (or v3 at least). I love `q` rather than `query`, which is the default v2 behavior. While it supplies `query`, `singleIndex` is so handy. The introduced stateMapping `singleIndexQ` would supply `q` in a light way.

Why `q`? Google (search) and Twitter search have used `q` so many users are familiar with `q`. Some developers would prefer using `q` without any coding of their own `stateMapping` to achieve end-user preference and/or a shorter user-friendly URL.

As becoming fat is not my favorite, I would like to know how I can resolve this kind of preference 😄

Haroenv:
We originally had a mapping where you provide `query: q` in urlSync, but found that it quickly lead to people needing more customisation, and at that point, they'd need to make changes to this statemapping, and it wouldn't fit anymore.

I wonder if there's more some way to make these kind of customisations easier:

https://codesandbox.io/s/exciting-field-zmsn1?file=/src/app.js

maybe some function that does the looping / mapping for you (since it does get a little verbose), but then what if people want to map the values?

Haroenv:
@algolia/instantsearch-for-websites what do you think?

tnir:
As you mentioned, customization gets harder if people select this `singleIndexQ`. But those who want to customize would choose `simple` stateMapping or implement their own from the scratch.

On the example you showed looks like generating `instant_search%5Bq%5D=15&instant_search%5Bundefined%5D%5Bbrand%5D%5B0%5D=HP` as a full query string. That is too verbose, and I expect `q=15&refinementList%5Bbrand%5D%5B0%5D=HP` or `q=15`.

francoischalifour:
Hi @tnir, thanks for the contribution!

We usually don't add an API if users don't show interest for it. We never received any requests regarding this compatibility layer for v2, I guess migrations guides are enough for most of them. Therefore we're not going to embed this in InstantSearch itself.

We however suggest to share this idea in [our Discourse forum](https://discourse.algolia.com) where the community shares their InstantSearch implementations and ideas. It could be interesting to see how many people engage with such a post to possibly add this feature in InstantSearch later!

PR #4509, opened 2020-09-10
fix(infiniteHits): compute `isLastPage` based on cached pages

francoischalifour:
This fixes a bug in the `infiniteHits` widget where the `isLastPage` value could conflict with the cached state.

## Reproduction

Assume you have an `infiniteHits` widget with a `pagination` widget (which can happen when you want better SEO).

1. Click page 3
2. Click page 1
3. Click "Show more results"
4. Notice you're on page 4 which doesn't exist

[See reproduction →](https://codesandbox.io/s/instantsearchjs-infinitehits-page-bug-pyd4v)

## Fix

We now don't rely only on the helper state but also on the cached state. If the last page has been cached in the `infiniteHits` cache, we consider `isLastPage` to be `true`.


eunjae-lee:
@francoischalifour you might want to check the failing e2e test.

Haroenv:
I ran the tests again @eunjae-lee and it passes now :)

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -291,7 +291,9 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
         }
 
         const isFirstPage = getFirstReceivedPage() === 0;
-        const isLastPage = results.nbPages <= results.page + 1;
+        const isLastPage =
+          results.nbPages <= results.page + 1 ||
```
Haroenv, line 295: is this case still needed, does `getLastReceivedPage` not cover the case?

eunjae-lee: Ah, that's true. The current page is always equal or lower than the last received page.

francoischalifour: That's right — a badly written test was failing and got me to keep that line.

Addressed in 1e76cb1e23f6d8153effd98a5f70b77f96f8deea.

Haroenv (approved):
thanks!

eunjae-lee (approved):

PR #4510, opened 2020-09-11
chore(storybook): fix breadcrumb example

Haroenv:
**Summary**

fix error in breadcrumb example

**Result**

changed breadcrumb to breadcrumbDiv

Changed it to TS too, since it would have broken the build


yannickcr (approved):

eunjae-lee (approved):
good call on changing it to ts

PR #4513, opened 2020-09-11
fix(hitsPerPage): update link to hitsPerPage widget

tnir:
**Summary**

Replaces a link to the docs on the old website with one on the new one.

**Result**

A developer would see v4 docs instead of v2.


francoischalifour (approved):
Thanks!

PR #4518, opened 2020-09-21
fix: provide both `getWidgetRenderState` and `getRenderState` in connectors

eunjae-lee:
**Summary**

This PR splits `getWidgetRenderState` into `getWidgetRenderState` and `getRenderState`.

### DIFF
--- a/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
+++ b/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
@@ -429,14 +429,14 @@ search.addWidgets([
     });
   });
 
-  describe('getWidgetRenderState', () => {
+  describe('getRenderState', () => {
     test('returns the render state', () => {
       const renderFn = jest.fn();
       const unmountFn = jest.fn();
       const createAutocomplete = connectAutocomplete(renderFn, unmountFn);
       const autocomplete = createAutocomplete({});
 
-      const renderState1 = autocomplete.getWidgetRenderState!(
+      const renderState1 = autocomplete.getRenderState!(
         {},
         createInitOptions()
       );
@@ -450,7 +450,7 @@ search.addWidgets([
 
       autocomplete.init!(createInitOptions());
 
-      const renderState2 = autocomplete.getWidgetRenderState!(
+      const renderState2 = autocomplete.getRenderState!(
         {},
         createRenderOptions()
       );
@@ -474,7 +474,7 @@ search.addWidgets([
 
       autocomplete.init!(createInitOptions());
 
-      const renderState = autocomplete.getWidgetRenderState!(
+      const renderState = autocomplete.getRenderState!(
         {},
         createRenderOptions({ helper })
       );
--- a/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
+++ b/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
@@ -436,10 +436,7 @@ search.addWidgets([
       const createAutocomplete = connectAutocomplete(renderFn, unmountFn);
       const autocomplete = createAutocomplete({});
 
-      const renderState1 = autocomplete.getRenderState!(
-        {},
-        createInitOptions()
-      );
+      const renderState1 = autocomplete.getRenderState({}, createInitOptions());
 
       expect(renderState1.autocomplete).toEqual({
         currentRefinement: '',
@@ -450,7 +447,7 @@ search.addWidgets([
 
       autocomplete.init!(createInitOptions());
 
-      const renderState2 = autocomplete.getRenderState!(
+      const renderState2 = autocomplete.getRenderState(
         {},
         createRenderOptions()
       );
@@ -474,7 +471,7 @@ search.addWidgets([
 
       autocomplete.init!(createInitOptions());
 
-      const renderState = autocomplete.getRenderState!(
+      const renderState = autocomplete.getRenderState(
         {},
         createRenderOptions({ helper })
       );
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -6,7 +6,7 @@ import {
   noop,
   warning,
 } from '../../lib/utils';
-import { Hits, Connector } from '../../types';
+import { Hits, Connector, AutocompleteWidgetRenderState } from '../../types';
 
 const withUsage = createDocumentationMessageGenerator({
   name: 'autocomplete',
@@ -110,7 +110,9 @@ search.addWidgets([
 
         renderFn(
           {
-            ...this.getWidgetRenderState!(initOptions),
+            ...this.getWidgetRenderState!<AutocompleteWidgetRenderState>(
+              initOptions
+            ),
             instantSearchInstance,
           },
           true
@@ -122,7 +124,9 @@ search.addWidgets([
 
         renderFn(
           {
-            ...this.getWidgetRenderState!(renderOptions),
+            ...this.getWidgetRenderState!<AutocompleteWidgetRenderState>(
+              renderOptions
+            ),
             instantSearchInstance,
           },
           false
@@ -132,7 +136,9 @@ search.addWidgets([
       getRenderState(renderState, renderOptions) {
         return {
           ...renderState,
-          autocomplete: this.getWidgetRenderState!(renderOptions),
+          autocomplete: this.getWidgetRenderState!<
+            AutocompleteWidgetRenderState
+          >(renderOptions),
         };
       },
 
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -6,7 +6,7 @@ import {
   noop,
   warning,
 } from '../../lib/utils';
-import { Hits, Connector, AutocompleteWidgetRenderState } from '../../types';
+import { Hits, Connector } from '../../types';
 
 const withUsage = createDocumentationMessageGenerator({
   name: 'autocomplete',
@@ -56,9 +56,12 @@ export type AutocompleteRendererOptions = {
 
 export type AutocompleteConnector = Connector<
   AutocompleteRendererOptions,
-  AutocompleteConnectorParams,
-  AutocompleteWidgetRenderState
+  AutocompleteConnectorParams
 >;
+
+type WidgetParams = {
+  escapeHTML?: boolean;
+};
 
 const connectAutocomplete: AutocompleteConnector = function connectAutocomplete(
   renderFn,
@@ -66,8 +69,8 @@ const connectAutocomplete: AutocompleteConnector = function connectAutocomplete(
 ) {
   checkRendering(renderFn, withUsage());
 
-  return widgetParams => {
-    const { escapeHTML = true } = widgetParams || ({} as typeof widgetParams);
+  return (widgetParams: WidgetParams) => {
+    const { escapeHTML = true } = widgetParams || ({} as WidgetParams);
 
     warning(
       !(widgetParams as any).indices,
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -133,10 +133,10 @@ search.addWidgets([
         );
       },
 
-      getRenderState(renderState, renderStateOptions) {
+      getRenderState(renderState, renderOptions) {
         return {
           ...renderState,
-          autocomplete: this.getWidgetRenderState!(renderStateOptions),
+          autocomplete: this.getWidgetRenderState!(renderOptions),
         };
       },
 
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -110,9 +110,9 @@ search.addWidgets([
 
         renderFn(
           {
-            ...(this.getWidgetRenderState!(
+            ...this.getWidgetRenderState!<AutocompleteWidgetRenderState>(
               initOptions
-            ) as AutocompleteWidgetRenderState),
+            ),
             instantSearchInstance,
           },
           true
@@ -124,9 +124,9 @@ search.addWidgets([
 
         renderFn(
           {
-            ...(this.getWidgetRenderState!(
+            ...this.getWidgetRenderState!<AutocompleteWidgetRenderState>(
               renderOptions
-            ) as AutocompleteWidgetRenderState),
+            ),
             instantSearchInstance,
           },
           false
@@ -136,7 +136,9 @@ search.addWidgets([
       getRenderState(renderState, renderOptions) {
         return {
           ...renderState,
-          autocomplete: this.getWidgetRenderState!(renderOptions),
+          autocomplete: this.getWidgetRenderState!<
+            AutocompleteWidgetRenderState
+          >(renderOptions),
         };
       },
 
--- a/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/src/connectors/autocomplete/connectAutocomplete.ts
@@ -59,18 +59,14 @@ export type AutocompleteConnector = Connector<
   AutocompleteConnectorParams
 >;
 
-type WidgetParams = {
-  escapeHTML?: boolean;
-};
-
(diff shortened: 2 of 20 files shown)


src/types/widget.ts
```diff
@@ -262,10 +276,23 @@ export type Widget = {
...
    * Returns the render params to pass to the render function.
    */
-  getWidgetRenderState?(
+  getRenderState?(
     renderState: IndexRenderState,
-    widgetRenderStateOptions: InitOptions | RenderOptions
+    renderOptions: InitOptions | RenderOptions
   ): IndexRenderState;
+  /**
+   * Returns the render params to pass to the render function.
```
yannickcr, line 284: Maybe be more precise about what this function returns, same for `getRenderState` so the differences between the two is clear for everyone.

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -103,35 +103,44 @@ search.addWidgets([
...
           helper.setQuery(query).search();
         };
 
         renderFn(
           {
-            ...this.getWidgetRenderState!(renderState, initOptions)
-              .autocomplete!,
+            ...(this.getWidgetRenderState!(
+              initOptions
+            ) as AutocompleteWidgetRenderState),
```
yannickcr, line 115: No way to type the function return type instead of casting it here ? (casting this way is not type safe)

francoischalifour: We can probably make `getWidgetRenderState` generic?

Haroenv: I tried to find solutions to make the whole widget generic based on the type passed to the connector, but I couldn't really make it work. Just having gwrs generic will be a good stop-gap

src/types/widget.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4518/commits/6adc8b9c45eeea54f9391a7ffbfe2d7db1b6b5cf

eunjae-lee:
What do you think about this usage of discriminated type union?
If you prefer adding a new generic into `Connector`, I can do that,
but with this way, we can now differentiate widgets with `$$type`.
@yannickcr @francoischalifour @Haroenv

yannickcr:
I'm more in favor of the discriminated type union, seems more clear to me.

With it we can also be more precise about the type for each widget (like making `getWidgetRenderState` as required in the "`AutocompleteWidget`" type so we can avoid using `!` everywhere in the widget itself).

src/types/widget.ts
```diff
@@ -328,7 +284,73 @@ export type Widget = {
...
+      getWidgetRenderState?: (
+        renderOptions: InitOptions | RenderOptions
+      ) => CurrentRefinementsWidgetRenderState;
+    }
+  | {
+      $$type: 'ais.hierarchicalMenu';
+      getWidgetRenderState?: (
+        renderOptions: InitOptions | RenderOptions
+      ) => HierarchicalMenuWidgetRenderState;
+    }
```
Haroenv, line 329: what if this part is a generic that the connector passes to widget? it can have a default of `{}` or something like that?

eunjae-lee: I think that's possible too. Then the definition of the connector will become something like:

```js
// from
export type AutocompleteConnector = Connector<
  AutocompleteRendererOptions,
  AutocompleteConnectorParams
>;

// to
type AutocompleteXYZ = {
  $$type: 'ais.autocomplete';
  getWidgetRenderState?: (
    renderOptions: InitOptions | RenderOptions
  ) => AutocompleteWidgetRenderState;
};

export type AutocompleteConnector = Connector<
  AutocompleteRendererOptions,
  AutocompleteConnectorParams,
  AutocompleteXYZ,
>;

// with

export type WidgetFactory<TConnectorParams, TWidgetParams, TXyz> = (
  /**
   * The params of the widget.
   */
  widgetParams: TConnectorParams & TWidgetParams
) => Widget & TXyz;
```

src/types/widget.ts
```diff
@@ -300,7 +284,73 @@ export type Widget = {
     state: SearchParameters,
     widgetSearchParametersOptions: WidgetSearchParametersOptions
   ): SearchParameters;
-};
+} & (
```
francoischalifour, line 287: With this strategy, custom widgets that implement `getWidgetRenderState` will throw a TypeScript error. We still need to add a generic `getWidgetRenderState` independent of the `$$type`.

francoischalifour, line 287: Generics were created to solve this specific problem. They're a solution to define an interface that we implement in our widgets, and that users can implement in their own widgets. Here we loose this API definition.

Why avoiding them?

eunjae-lee: That's a good point.

eunjae-lee: I'm not avoiding them, but was trying to compare them and I literally didn't know what's better and what has which drawbacks. I think https://github.com/algolia/instantsearch.js/pull/4518#discussion_r498073495 makes it enough to choose generics. Let me know if you have more thoughts on the drawbacks of this approach (discriminated type union)  in addition to that.

eunjae-lee:
I've changed the implementation to use generics. We have a bunch of type errors here, but they will be fixed in other PRs when we work on `getWidgetRenderState` and `getRenderState` in each connector.

Let me know what you think about the change here!

src/types/widget.ts
yannickcr: Yes, and I don't see why this is an issue :thinking:

eunjae-lee: IIUC, if a user is using TypeScript, and implement a custom widget with a $$type: let's say.. `ais.myCustomWidget`. Then the custom widget's type `Widget` won't include `getWidgetRenderState` and the user has no proper way to include it.

yannickcr: What I mean is that we can (must) add a generic `getWidgetRenderState` method to the base Widget type to avoid this and for the user to get proper typing in a custom widget.

Also users must not put a `$$type` key on a custom widget since this is for internal use only.

yannickcr: @francoischalifour I'm not sure what API definition we are loosing here. Also Generic have their own limitations and in our case we want to have stricter definition on our own widget than what we enforce for custom ones. This is where Generics find their limits here, forcing us to be less strict on our own widgets. I find discriminated type unions here to fit perfectly for what we want to accomplish while keeping the types more readable. For example here's what I have in mind: https://github.com/algolia/instantsearch.js/blob/311654f856a4b3625395a370e95ca606806efb6d/src/types/widget.ts#L237-L475

Haroenv: I think a generic for RenderState defaulting to any / an object / similar would make sense though, that way external users can still use very loose typescript if they want, but can also opt-in to the same safety as us. We won't be able to put custom widgets in uiState & render state without a separate generic on InstantSearch I guess

eunjae-lee: @yannickcr Can you help me understand, without generics, but with discriminated type unions, how to enable customers to create custom widgets and still let them implement `getWidgetRenderState` without type issue?

yannickcr: @eunjae-lee With discriminated type unions a custom widget created by the user will not have a `$$type` property, so it will match the `GenericWidget` type.

In that case if the user wants to implement `getWidgetRenderState` he will need to follow the type of `getWidgetRenderState` as implemented in `GenericWidget` (`getWidgetRenderState?: (renderOptions: InitOptions | RenderOptions) => any;` here but as @Haroenv we could force it to return an object)

Haroenv (commented):
I think this is the way to go, but let's have a meeting to align on this?

src/types/widget.ts
eunjae-lee: Oh I didn't see you used `interface` to extend the `GenericWidget`.

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -103,35 +107,40 @@ search.addWidgets([
...
+        const { helper, instantSearchInstance } = initOptions;
         connectorState.refine = (query: string) => {
           helper.setQuery(query).search();
         };
 
         renderFn(
           {
-            ...this.getWidgetRenderState!(renderState, initOptions)
-              .autocomplete!,
+            ...this.getWidgetRenderState!(initOptions),
```
Haroenv, line 117: these exclamation marks aren't needed anymore now right?

eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4518/commits/80037b6edc0a26095272799bef80245d7bfae328

eunjae-lee:
@francoischalifour  Hey can you give it a review by end of today? :)

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -429,14 +429,14 @@ search.addWidgets([
...
-  describe('getWidgetRenderState', () => {
+  describe('getRenderState', () => {
     test('returns the render state', () => {
       const renderFn = jest.fn();
       const unmountFn = jest.fn();
       const createAutocomplete = connectAutocomplete(renderFn, unmountFn);
       const autocomplete = createAutocomplete({});
 
-      const renderState1 = autocomplete.getWidgetRenderState!(
+      const renderState1 = autocomplete.getRenderState!(
```
Haroenv, line 439: also in the tests it can be removed

src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -59,14 +59,18 @@ export type AutocompleteConnector = Connector<
...
+
 const connectAutocomplete: AutocompleteConnector = function connectAutocomplete(
   renderFn,
   unmountFn = noop
 ) {
   checkRendering(renderFn, withUsage());
 
-  return widgetParams => {
-    const { escapeHTML = true } = widgetParams || ({} as typeof widgetParams);
+  return (widgetParams: WidgetParams) => {
```
Haroenv, line 72: isn't this `AutocompleteConnectorParams` already?

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
```diff
@@ -85,7 +85,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
 
       helper.toggleRefinement('category', 'Decoration');
 
-      const renderState1 = breadcrumb.getWidgetRenderState!(
+      const renderState1 = breadcrumb.getRenderState!(
```
Haroenv, line 88:
```suggestion
      const renderState1 = breadcrumb.getRenderState(
```

src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -73,7 +78,8 @@ export type BreadcrumbRendererOptions = {
 
 export type BreadcrumbConnector = Connector<
   BreadcrumbRendererOptions,
-  BreadcrumbConnectorParams
+  BreadcrumbConnectorParams,
+  BreadcrumbWidgetRenderState
```
Haroenv, line 82: I thought we were going to move calculate RenderState in the Connector type directly, so the arguments could stay as-is?

Haroenv (changes requested):

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
eunjae-lee: Hmm. To remove the exclamation mark from `getRenderState`, we should go like this:

image


What do you think?

src/connectors/autocomplete/connectAutocomplete.ts
eunjae-lee: was experimental. I've reverted it back to the original.

src/connectors/breadcrumb/connectBreadcrumb.ts
eunjae-lee: Oh sorry. I forgot to mention this.

I think we can focus on the typing around the widget itself and `connectAutocomplete` as an example.
Once it's all agreed and this PR is merged, then we can fix types in new PRs, one per connector.

eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4518/commits/d7e5206a393a3af6164e57d3691139f26acb1dbd

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4518/commits/2f2c18a75cdc3aeee0d85f2f13a8b04ff8cb75ba

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
Haroenv: yes, I think that makes sense

src/connectors/configure/connectConfigure.ts
```diff
@@ -132,13 +131,17 @@ const connectConfigure: ConfigureConnector = function connectConfigure(
...
 
-      getWidgetRenderState(renderState) {
+      getRenderState(renderState, renderOptions) {
         return {
           ...renderState,
-          configure: {
-            refine: connectorState.refine!,
-            widgetParams,
-          },
+          configure: this.getWidgetRenderState(renderOptions),
```
Haroenv, line 137: maybe leave a comment here that if we want to merge to not show only the last configure widget, it would happen here? 

I agree with your decision that getWidgetRenderState should not accept a previous renderState, and merging should happen in getRenderState

src/connectors/search-box/__tests__/connectSearchBox-test.js
```diff
@@ -307,7 +308,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/search-box/
     expect(helper.search).toHaveBeenCalledTimes(1);
   });
 
-  describe('getWidgetRenderState', () => {
+  describe('getRenderState', () => {
```
Haroenv, line 311: in most connectors we're missing tests for getWidgetRenderState now. I guess it could be done as a subsequent part of this PR as long as we keep a list of which tests are missing in jira

src/types/widget.ts
```diff
@@ -143,62 +143,76 @@ export type RenderState = {
...
-    },
-    {
-      queryHook?(query: string, refine: (query: string) => void);
-    }
-  >;
-  autocomplete: WidgetRenderState<
-    AutocompleteRendererOptions,
-    AutocompleteConnectorParams
-  >;
+  searchBox: SearchBoxWidgetRenderState;
```
Haroenv, line 206: what if we write WidgetRenderState<SearchBoxRendererOptions, SearchBoxConnectorParams> directly here? Then we don't have all those types that need to be exported. The TS ones can be directly imported, the JS ones written inline (IMHO)

Having types for each individual widget here feels like it clutters this file a bit IMO

Haroenv: If it definitely has to exist, I'd prefer a renderstate file (maybe a uiState file too, but out of scope of this PR, maybe make it stacked if we agree on it)

src/connectors/search-box/__tests__/connectSearchBox-test.js
eunjae-lee: Done DX-1355

src/types/widget.ts
eunjae-lee: Initially I splitted them because I had to re-use them in other places, but with the current way we don't have to. I'll merge them.

src/connectors/configure/connectConfigure.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4518/commits/6c233e750569231ef28b3316348ec571aea68295

src/types/widget.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4518/commits/ae03fd4cda36beb0a56c88fe64592238ba465dff

Haroenv (approved):
for me this looks good, but would appreciate a review from @yannickcr & @francoischalifour too please

eunjae-lee:
I'm merging now. Feel free to give feedback later when you want something to be different. This being not merged kind of blocks the subsequent tasks.

PR #4519, opened 2020-09-21
fix(insights): fix the regression that it didn't send events with instantsearch.insights()

eunjae-lee:
**Summary**

This PR fixes the regression introduced since 4.8.0 where InstantSearch didn't send Insights events implemented with `instantsearch.insights()` in `hits` or `infiniteHits` widgets.

**Result**

It sends events now.

**Test**
If you run this [example ->](https://codesandbox.io/s/throbbing-worker-5or8z?file=/src/app.js), it will correctly send events when you click buttons. However if you change the version of InstantSearch to 4.8.0, it won't any more.

### DIFF
--- a/src/lib/insights/listener.tsx
+++ b/src/lib/insights/listener.tsx
@@ -58,7 +58,9 @@ const insightsListener = (BaseComponent: any) => {
           const payload = parseInsightsEvent(targetWithEvent);
           props.sendEvent(payload);
         }
-      } else {
+      }
+
+      {
         // old way, e.g. instantsearch.insights("clickedObjectIDsAfterSearch", { .. })
         const insightsTarget = findInsightsTarget(
           event.target as HTMLElement | null,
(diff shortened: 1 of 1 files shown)


src/lib/insights/listener.tsx
```diff
@@ -58,7 +58,9 @@ const insightsListener = (BaseComponent: any) => {
           const payload = parseInsightsEvent(targetWithEvent);
           props.sendEvent(payload);
         }
-      } else {
+      }
+
+      {
```
eunjae-lee, line 63: `prop.sendEvent` exists, so it always followed the first if branch. We should run them both.

yannickcr (approved):

src/lib/insights/listener.tsx
Haroenv: if we should run them both, you should probably remove the curly brace here and on line 72

Haroenv (changes requested):

Haroenv (approved):

francoischalifour (commented):
Why don't we cover this regression with a test?

PR #4523, opened 2020-09-24
fix(refinementList): cap `maxFacetHits` to 100 for SFFV

### LINKED ISSUE #4521: Searching for facet can throw an error when a refinement list is expanded with more than 100 items.
## 🐛 Bug description

Searching for facet can throw an error when a refinement list is expanded with more than 100 items.

## 🔍 Bug reproduction

1. Go to https://codesandbox.io/s/admiring-haslett-ipfw1?file=/src/app.js
2. Click on "Show more" button on the refinement list
3. Type any key on the facet search bar

## The problematic configuration

```js
    container: '#brand-list',
    attribute: 'brand',
    searchable: true,
    limit: 10,
    showMore: true,
    showMoreLimit: 1000,
  }),
```

## 💭 Expected behavior

We should be able to search. Although we have `showMoreLimit: 1000`, facet search shouldn't use `1000` as is. We should cap it (probably to 100?).

## 🖥 Screenshots

image

## Environment

- OS: macOS 10.15.6
- Browser: FireFox
- Version: 80.0.1

francoischalifour:
We cap the [`maxFacetHits`](https://www.algolia.com/doc/api-reference/api-parameters/maxFacetHits/) value to 100 because the Algolia API doesn't support a greater number.

An Enterprise customer has faced this issue in Vue InstantSearch and it's challenging to fix user-land.

Closes #4521.

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,7 +146,7 @@
     },
     {
       "path": "./dist/instantsearch.development.js",
-      "maxSize": "150.45 kB"
+      "maxSize": "160 kB"
     }
   ]
 }
--- a/src/connectors/refinement-list/__tests__/connectRefinementList-test.js
+++ b/src/connectors/refinement-list/__tests__/connectRefinementList-test.js
@@ -1511,6 +1511,70 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
     });
   });
 
+  it('caps the search in facet values to 100 facet hits', () => {
+    const { makeWidget, rendering } = createWidgetFactory();
+    const widget = makeWidget({
+      attribute: 'category',
+      limit: 50,
+      showMoreLimit: 1000,
+    });
+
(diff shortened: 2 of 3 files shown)


Haroenv (approved):
looks good!

eunjae-lee:
seems to work fine!
https://codesandbox.io/s/nameless-https-rr83i?file=/src/app.js

eunjae-lee (changes requested):
looks good as long as bundlesize conf gets updated

francoischalifour:
@eunjae-lee I see that the dev bundlesize threshold was decreased. We originally put a high arbitrary number because it's counter productive to track the dev bundle size. It leads us to going for poorer DX.

I increased it back to a higher number than needed.

eunjae-lee (approved):

PR #4524, opened 2020-09-25
fix(insights): show deprecation warnings for old insights related properties and functions

eunjae-lee:
**Summary**

This PR deprecates and shows warning messages for the following things:
- `getInsightsAnonymousUserToken()`
- `InstantSearch({ insightsClient })` ← `insightsClient` prop
- `analytics` widget

### DIFF
--- a/src/helpers/get-insights-anonymous-user-token.ts
+++ b/src/helpers/get-insights-anonymous-user-token.ts
@@ -22,7 +22,7 @@ export function getInsightsAnonymousUserTokenInternal(): string | undefined {
 }
 
 /**
- * @deprecated This function will be still supported in 4.x releases, but not further. It is replaced by `insights` middleware. For more information, visit https://www.algolia.com/doc/guides/getting-insights-and-analytics/search-analytics/click-through-and-conversions/how-to/send-click-and-conversion-events-with-instantsearch/js/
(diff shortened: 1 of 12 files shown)


src/helpers/get-insights-anonymous-user-token.ts
```diff
@@ -15,6 +17,19 @@ function getCookie(name: string): string | undefined {
   return undefined;
 }
 
-export default function getInsightsAnonymousUserToken(): string | undefined {
+export function getInsightsAnonymousUserTokenInternal(): string | undefined {
```
eunjae-lee, line 20: This logic is used internally by `insights` middleware. So we keep this for us.

francoischalifour:
Did we make sure that the Insights middleware covers all cases that the `analytics` widget covers?

src/widgets/analytics/analytics.ts
```diff
@@ -76,6 +77,13 @@ const analytics: AnalyticsWidget = function analytics(widgetParams) {
     throw new Error(withUsage('The `pushFunction` option is required.'));
   }
 
+  warning(
+    false,
+    `\`analytics\` widget has been deprecated. It is still supported in 4.x releases, but not further. It is replaced by \`insights\` middleware.
```
Haroenv, line 82: we should show here exactly how to replace it IMO, the analytics widget gets fired on different moments than the insights one I think. Maybe we can look at how magento uses it, and migrate there to the insights middleware before deprecating? Maybe adding a FAQ or a guide in "upgrade guides"?

eunjae-lee: As soon as the doc PR is merged & deployed, I will add the link here.

Haroenv (approved):
looks good to me

PR #4527, opened 2020-09-30
feat(panel): spread widgetRenderState in the options in panel

eunjae-lee:
**Summary**

This PR spreads `widgetRenderState` in the options of `hidden`, `collapsed` and `templates` in `panel`.

**Example**

```js
const breadcrumbInPanel = panel<typeof breadcrumb>({
  collapsed({ canRefine }) {
    return canRefine === false;
  },
  templates: {
    header({ canRefine }) {
      return `Breadcrumb that can${canRefine ? '' : "'t "} refine`;
    },
    footer:
      'The panel collapses if it cannot refine. Click "Home". This panel will collapse and you will not see this footer anymore.',
  },
})(breadcrumb); 
```


src/widgets/panel/panel.tsx
```diff
@@ -81,13 +81,13 @@ export type PanelWidgetOptions = {
    * A function that is called on each render to determine if the
    * panel should be hidden based on the render options.
    */
-  hidden?(options: RenderOptions): boolean;
+  hidden?(options: RenderOptions & { widgetRenderState?: unknown }): boolean;
```
eunjae-lee, line 84: Not sure what we can do about this. Any idea?

stories/panel.stories.ts
```diff
@@ -129,4 +145,45 @@ storiesOf('Basics/Panel', module)
...
     })
+  )
+  .add(
+    'collapsed unless canRefine',
+    withHits(
+      ({ search, container }) => {
+        const breadcrumbInPanel = panel({
+          collapsed: ({ widgetRenderState }) => {
+            // TODO: We need better typing here. Not sure how, yet.
+            return widgetRenderState && !(widgetRenderState as any).canRefine;
```
Haroenv, line 156:
```suggestion
            return widgetRenderState && !(widgetRenderState as any).canRefine;
```

in which case is widgetRenderState undefined or null?

stories/panel.stories.ts
```diff
@@ -129,4 +145,45 @@ storiesOf('Basics/Panel', module)
...
       ]);
     })
+  )
+  .add(
+    'collapsed unless canRefine',
+    withHits(
+      ({ search, container }) => {
+        const breadcrumbInPanel = panel({
+          collapsed: ({ widgetRenderState }) => {
+            // TODO: We need better typing here. Not sure how, yet.
```
Haroenv, line 155: I don't think it's possible to get type values from the third call inside the first :(

I guess it's possible if panel was a single function call, but that's very different from the current implementation

stories/panel.stories.ts
eunjae-lee: when it's a custom widget provided by user https://github.com/algolia/instantsearch.js/blob/feat/widget-render-state-in-panel/src/types/widget.ts#L324:L324

src/widgets/panel/panel.tsx
Haroenv: this is an alternative where panel is generic and you need to repeat the connector type, that might be good enough:

https://www.typescriptlang.org/play?#code/C4TwDgpgBA6glgEwOYWAHgCoD4oF4oDeAUFFCsPMqgEoQB2CEATgMrACGwEAFAJQBcUDAG4iAX1FFQkKAGEA9nToQAxsHlNMOfHzw5K5LZIBmAVzpq4iqGHbKANpigQAHlwYBnOYuVqNaOxAsLG4SGzsIewAFdiZ2AFsPQWJSUgALREY6blikJMIw1KgAd0RyWgZmNk4IQVpgUyY6DHAINHrG5tatAG0AcnIDGnpGVg4uPoBdLFEisQEoQNmoMQAaIl4C0iZUTqhuFR9VdSZBDE3cHAIxcSIiQ7oPYCh4kAUlY41Bd98TtAIoId5PZBE8mHA6EgVtp9hccNwUmRUENgBVRtUuLpEdtdk1CID5MDBH0AEYQOwePorZZrcS8SS2BxoaQQeTGF5vI5+JghMLYqAZBBZHJMPKbfmkB4eYEQAB09nkSBFYuWpFpavWvG4rx+nyYvCIQA

stories/panel.stories.ts
Haroenv: good call, we can't force them to return it, but I wonder if we could default it to an empty object to prevent this condition? not sure if it's a good idea though

stories/panel.stories.ts
```diff
@@ -129,4 +145,45 @@ storiesOf('Basics/Panel', module)
         }),
       ]);
     })
+  )
+  .add(
+    'collapsed unless canRefine',
+    withHits(
+      ({ search, container }) => {
+        const breadcrumbInPanel = panel({
+          collapsed: ({ widgetRenderState }) => {
```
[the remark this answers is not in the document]
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4527/commits/ab6fd0a64a9ded34758677f018a38b756ed1089c

stories/panel.stories.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4527/commits/ab6fd0a64a9ded34758677f018a38b756ed1089c

Yeah I think it's good to provide an empty object by default, so that users don't have to deal with nullish check.

src/widgets/panel/panel.tsx
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4527/commits/de2cbef68588d7209893db7038bf5f616d740680
it works!

```
panel<BreadcrumbWidget>({ ... })(breadcrumb);
```

if I don't provide the generic, then `widgetRenderState` is `any`.

stories/panel.stories.ts
```diff
@@ -150,10 +152,9 @@ storiesOf('Basics/Panel', module)
     'collapsed unless canRefine',
     withHits(
       ({ search, container }) => {
-        const breadcrumbInPanel = panel({
+        const breadcrumbInPanel = panel<BreadcrumbWidget>({
```
Haroenv, line 155:
```suggestion
        const breadcrumbInPanel = panel<typeof breadcrumb>({
```

can we accept this, and panel could be using something with a few layers of ReturnType?

eunjae-lee: I just checked and `typeof breadcrumb` is `BreadcrumbWidget`. So you can already replace it without any problem, meaning we cannot use fewer layers of ReturnType even with this change. https://github.com/algolia/instantsearch.js/pull/4527/commits/d1da39bce910c1f3588cf5f7cd29a60e153ce330

src/widgets/panel/panel.tsx
```diff
@@ -76,18 +76,30 @@ export type PanelTemplates = {
   collapseButtonText?: Template<{ collapsed: boolean }>;
 };
 
-export type PanelWidgetOptions = {
+export type PanelWidgetOptions<TWidget extends WidgetFactory<any, any, any>> = {
```
Haroenv, line 79: not sure if this is the user api, but I think the widget should be optional, and we can do something like `extends WF | never = never` and then do a conditional on line 86 where if it's given we do the current code and if it extends never we do either an object or the union of all widget render states. What do you think?

src/widgets/panel/panel.tsx
```diff
@@ -248,7 +260,17 @@ const panel: PanelWidget = widgetParams => {
         return undefined;
       },
       render(...args) {
-        const [options] = args;
+        const [renderOptions] = args;
+
+        let widgetRenderState;
```
francoischalifour, line 265: We can simplify this branching logic:

```ts
const widgetRenderState = widget.getWidgetRenderState?.(renderOptions) || {};
```

src/widgets/panel/panel.tsx
```diff
@@ -76,18 +76,30 @@ export type PanelTemplates = {
...
+export type PanelWidgetOptions<TWidget extends WidgetFactory<any, any, any>> = {
   /**
    * A function that is called on each render to determine if the
    * panel should be hidden based on the render options.
    */
-  hidden?(options: RenderOptions): boolean;
+  hidden?(
+    options: RenderOptions & {
+      widgetRenderState: ReturnType<
+        Exclude<ReturnType<TWidget>['getWidgetRenderState'], undefined>
```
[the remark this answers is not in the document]
francoischalifour: Hmm I'm not sure what line 87 does then, it looks really complicated.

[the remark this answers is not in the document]
eunjae-lee: So, Let me break it down.

First of all, the usage would be

```js
panel<typeof breadcrumb>({ ... })(breadcrumb);
```

And

> `TWidget extends WidgetFactory`

- It means TWidget is a WidgetFactory (Yeah it's TWidget, but not widget, but a widget factory).
- `ReturnType<Twidget>` -> It returns a widget
- `ReturnType<TWidget>['getWidgetRenderState']` -> From the widget, I'm picking `getWidgetRenderState` function
- `getWidgetRenderState` can exist, but in case of custom widget, it might be `undefined`.
- `Exclude<ReturnType<TWidget>['getWidgetRenderState'], undefined>` -> However we assume it exists here, so picking the real type.
- `ReturnType<Exclude<ReturnType<TWidget>['getWidgetRenderState'], undefined>>` -> Now we've got the return of `getWidgetRenderState`, which is going to be one of SearchBox WidgetRenderState, Autocomplete WidgetRenderState, Breadcrumb WidgetRenderState, ...

Yeah that long journey is to get the real WidgetRenderState for it, but not sure how to clean it up.

src/widgets/panel/panel.tsx
eunjae-lee: I just played with the idea. The issue is, when TWidget is not given, it's either

1. user is using a custom widget
2. user is using one of the built-in widgets but just not providing the type.

Let's say it's 2 and a breadcrumb. Showing the union of all widget render states might help users, but also can lead them to a wrong type, because it will also show things for searchBox, autocomplete, ... it might be confusing.

And if it's 1, showing the union really doesn't help at all.

The current implementation, when no generic is given, will just type widgetRenderState as `any`. They didn't provide the type, so we're giving them back `any`. Sort of makes sense? 
What do you think?

src/widgets/panel/panel.tsx
```diff
@@ -248,7 +256,13 @@ const panel: PanelWidget = widgetParams => {
         return undefined;
       },
       render(...args) {
-        const [options] = args;
+        const [renderOptions] = args;
+
+        const options = {
+          ...renderOptions,
+          widgetRenderState: (widget.getWidgetRenderState?.(renderOptions) ||
+            {}) as any,
```
eunjae-lee, line 264: Hmmm here 

```
widget.getWidgetRenderState?.(renderOptions) || {}
```

The type of it is `unknown`, because literally we don't know it there. So without casting, it cannot be passed to `hidden` and `collapsed`:

```js
hidden: Boolean(hidden(options)),
collapsed: Boolean(collapsedFn(options)),
```

But once we cast and pass it there, later on the user-land, as long as they provide the type to `panel()`, they will get benefit.
What do you think? I feel like I've come to a dead-end.

src/widgets/panel/panel.tsx
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4527/commits/c89a118ba4805366bb008fed630b5bc2f290d90f

src/widgets/panel/panel.tsx
Haroenv: ah yeah that does make sense. Might be better even to type it with unknown :) I don't know if we can find a place to document when and why to add the generic though

src/widgets/panel/panel.tsx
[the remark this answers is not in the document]
Haroenv:
```ts
export type PanelRenderOptions<
  TWidget extends WidgetFactory<any, any, any>
> = RenderOptions & {
  widgetRenderState: ReturnType<TWidget>['getWidgetRenderState'] extends (
    renderOptions: any
  ) => infer TRenderState
    ? TRenderState
    : unknown;
};
```

I think this is a little simpler maybe?

src/widgets/panel/panel.tsx
Haroenv: I made this pr to help: https://github.com/algolia/instantsearch.js/pull/4558

src/widgets/panel/panel.tsx
eunjae-lee: image

https://github.com/algolia/instantsearch.js/pull/4527/commits/0fcc02dede5938c926c6aec766a904ef6df3be57
trying to understand this error message, but cannot.
Can you help me?

src/widgets/panel/panel.tsx
```diff
@@ -77,14 +79,16 @@ export type PanelTemplates<TWidget extends WidgetFactory<any, any, any>> = {
 };
 
 export type PanelRenderOptions<
-  TWidget extends WidgetFactory<any, any, any>
+  TWidget extends WidgetFactory<unknown, unknown, unknown>
```
Haroenv, line 82: we should have an intermediate type in this file `UnknownWidgetFactory` instead of repeating this so many times imo

eunjae-lee: I can do that, but can you help me with this? https://github.com/algolia/instantsearch.js/pull/4527#discussion_r515983857

src/widgets/panel/panel.tsx
Haroenv: unknown is a specific type, and the breadcrumb doesn't fit in unknown, but it does in any (as far as I can tell)

eunjae-lee: In my understanding, 

```ts
type Aa = unknown;
const a: Aa = '3'
```

this should work and it does [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAgghlAvFArgOwNZoPYHc0DcAUAMbZoDOwUcAXLAsgOQDMTRHQA).

> Anything is assignable to unknown

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type

but the error message in the screenshot doesn't say something cannot be assigned to `unknown`, but the other way around.

Do you think we can revert it back to `any`, so that it won't harm the user-land typing if we're not going to deal with this now?

Haroenv (approved):
I think this is good to go, but let's make sure we have @francoischalifour & @yannickcr also review this

src/widgets/panel/panel.tsx
```diff
@@ -59,40 +59,51 @@ export type PanelCSSClasses = {
   footer?: string | string[];
 };
 
-export type PanelTemplates = {
+export type UnknownWidgetFactory = WidgetFactory<any, any, any>;
```
francoischalifour, line 62: If we export this, we can move it to `src/types/widget.ts` and use it in both places.

src/widgets/panel/panel.tsx
```diff
@@ -103,14 +114,14 @@ export type PanelWidgetOptions = {
...
-const renderer = ({
+const renderer = <TWidget extends UnknownWidgetFactory>({
   containerNode,
   bodyContainerNode,
   cssClasses,
   templates,
 }) => ({ options, hidden, collapsible, collapsed }) => {
   render(
-    <Panel
+    <Panel<TWidget>
```
francoischalifour, line 124: Wow—since when is this syntax supported? Wasn't aware of it!

src/widgets/panel/panel.tsx
```diff
@@ -248,7 +259,14 @@ const panel: PanelWidget = widgetParams => {
         return undefined;
       },
       render(...args) {
-        const [options] = args;
+        const [renderOptions] = args;
+
+        const options = {
+          ...(widget.getWidgetRenderState
+            ? widget.getWidgetRenderState(renderOptions)
+            : {}),
```
francoischalifour, line 267: Didn't this work?

```suggestion
          ...widget.getWidgetRenderState?.(renderOptions) || {},
```

eunjae-lee: FYI https://github.com/algolia/instantsearch.js/pull/4527#discussion_r515244256

src/widgets/panel/panel.tsx
eunjae-lee: Yeah, Haroen introduced it to us in the stacked PR: https://github.com/algolia/instantsearch.js/pull/4558
and it's interesting!

src/widgets/panel/panel.tsx
eunjae-lee: For now, `UnknownWidgetFactory` is used only inside this file. I think it should be alright to put it here. However, I've wondering when to export a type and when to not? Do we export all the types that are related to public APIs?

francoischalifour: What I mean is that we use this also here so we might as well reuse this type: https://github.com/algolia/instantsearch.js/pull/4527/files#diff-6b1f4a847fe51829e361b8d076cfff7b761fb616d83cabdfea076a08524f5c82R20.

src/widgets/panel/panel.tsx
francoischalifour: All right next TS version upgrade will allow it then!

src/widgets/panel/panel.tsx
Haroenv: a long time already, TS 2.9: https://mariusschulz.com/blog/passing-generics-to-jsx-elements-in-typescript

francoischalifour: Damn, I wish I had known that sooner!

src/widgets/panel/panel.tsx
eunjae-lee: Alright. I forgot about the component. https://github.com/algolia/instantsearch.js/pull/4527/commits/44643f9383765feba496e3507cb5db033501f88f

francoischalifour (approved):
Good to merge once:

- PR title is updated (because we don't pass the `widgetRenderState` object anymore)
- Lint is fixed
- Bundle size is increased

Well done!

eunjae-lee:
>     * PR title is updated (because we don't pass the `widgetRenderState` object anymore)
Done.

>     * Lint is fixed
>     * Bundle size is increased

This is based on `feat/render-state`. The type errors and the bundle size issue comes from there. So I will fix it there.

PR #4532, opened 2020-10-12
feat(hitsPerPage): implement `getRenderState` and `getWidgetRenderState`

shortcuts:
The **test:watch** yelled at me to add the object at **L798** in **connectHitsPerPage-test.ts** so please if you have any advices or question about the code let me know


src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -190,6 +192,7 @@ You may want to add another entry to the \`items\` option with this value.`
 
         renderFn(
           {
+            ...this.getWidgetRenderState(initOptions),
             items: transformItems(normalizeItems(state)),
             refine: connectorState.setHitsPerPage,
             createURL: connectorState.createURLFactory(helper.state),
```
Haroenv, line 198: these functions aren't needed anymore, since they are now part of getWidgetRenderState

Haroenv (commented):
looks like a good start! I haven't read the tests yet, seems like they don't run for external PRs, so we need to see them manually

src/connectors/hits-per-page/connectHitsPerPage.ts
shortcuts: Ok, I was wondering after I splitted the functions, thanks

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -205,15 +193,12 @@ You may want to add another entry to the \`items\` option with this value.`
...
+        const { results, instantSearchInstance } = initOptions;
         const hasNoResults = results.nbHits === 0;
 
         renderFn(
           {
             ...this.getWidgetRenderState(initOptions),
-            items: transformItems(normalizeItems(state)),
-            refine: connectorState.setHitsPerPage,
-            createURL: connectorState.createURLFactory(state),
             hasNoResults,
```
Haroenv, line 196: hasNoResults is also already part of getWidgetRenderState 👍

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -205,15 +193,12 @@ You may want to add another entry to the \`items\` option with this value.`
       },
 
       render(initOptions) {
-        const { state, results, instantSearchInstance } = initOptions;
+        const { results, instantSearchInstance } = initOptions;
         const hasNoResults = results.nbHits === 0;
```
Haroenv, line 191: same for this

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -235,11 +220,20 @@ You may want to add another entry to the \`items\` option with this value.`
...
+            )
+          );
+        };
+
         return {
           items: transformItems(normalizeItems(state)),
           refine: connectorState.setHitsPerPage,
-          createURL: connectorState.createURLFactory,
+          createURL: connectorState.createURLFactory(state),
           hasNoResults: results?.nbHits === 0,
```
Haroenv, line 237: this should be `results ? results.nbHits === 0 : true`, since otherwise it's wrong in init :)

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -181,21 +181,9 @@ You may want to add another entry to the \`items\` option with this value.`
...
-          );
-        };
-
         renderFn(
           {
             ...this.getWidgetRenderState(initOptions),
-            items: transformItems(normalizeItems(state)),
-            refine: connectorState.setHitsPerPage,
-            createURL: connectorState.createURLFactory(helper.state),
             hasNoResults: true,
```
Haroenv, line 181: this can be removed

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -235,11 +220,20 @@ You may want to add another entry to the \`items\` option with this value.`
...
-      getWidgetRenderState({ state, results }) {
+      getWidgetRenderState({ state, results, createURL }) {
+        connectorState.createURLFactory = helperState => value => {
+          return createURL(
+            helperState.setQueryParameter(
+              'hitsPerPage',
+              !value && value !== 0 ? undefined : value
+            )
+          );
+        };
```
Haroenv, line 231: I think this function should probably stay in init for consistency with connectorState.setHitsPerPage, but I'm not sure how other connectors dealt with this yet

Haroenv (commented):
good job so far!

src/connectors/hits-per-page/connectHitsPerPage.ts
eunjae-lee: I agree. It can stay in `init`.

Haroenv: basically the problem is that in the test (similar to breadcrumb) it seems to call getWidgetRenderState before init has happened, which means it will throw, since connectorState is still empty object at that point

eunjae-lee:
@Haroenv @shortcuts 
What do you think about this approach? https://github.com/algolia/instantsearch.js/pull/4532/commits/3cc8e7fa5b501c47f13efb5e1f008aad6ffa482d

`getRenderState` is called before `init` (https://github.com/algolia/instantsearch.js/blob/feat/getWidgetRenderState-hitsPerPage/src/widgets/index/index.ts#L267:L270)
So do we rely on the fact and let `getWidgetRenderState` to implement `connectorState`? Or, do we implement `connectorState` even before `init`, but instead, let it have factory-type functions to accept things like helper or state later on?

I pushed my commit to show you for discussion, so we can drop it if we decide to go for the previous implementation.

shortcuts:
@eunjae-lee 

As I'm not that familiar with the logic behind the connectorState, your approach seems cleaner if we want to update widgets later on or if we have the same issue with an other widget, and it's not that different to other widgets.

Haroenv:
your approach is the way to go as far as I can see @eunjae-lee, thanks!

yannickcr (approved):

devdreamsolution (approved):

PR #4534, opened 2020-10-13
fix(infiniteHits): do not cache the cached hits inside the connector

### LINKED ISSUE #4531: InfiniteHits cache cannot get invalidated
## 🐛 Bug description

We have `cache` object passed to InfiniteHits, but we store the hits in `cachedHits`. So `cachedHits` becomes the second layer of cache, which cannot be invalidated. InfiniteHits always should reflect the status of `cache`.

## 🔍 Bug reproduction

1. Go to https://codesandbox.io/s/compassionate-goldberg-bmltl?file=/src/app.js
2. Click on "Clear Cache" button
3. The hits are not refreshed.

## 💭 Expected behavior

We should either get rid of `cachedHits` or make it up-to-date with `cache`.

## Environment

- OS: macOS 10.15.7
- Browser: FireFox
- Version: 81.0.1 (64-bit)

eunjae-lee:
**Summary**

This PR fixes an issue where `connectInfiniteHits` was caching the `cachedHits` internally. Even if the cache gets cleared, the connector cannot know it. It's an unnecessary double layer of caches.

fixes #4531 

**Result**

After the cache gets invalidated, the connect should read hits correctly from the results.

The following example works:
https://codesandbox.io/s/dreamy-forest-eofp6?file=/src/app.js

### DIFF
--- a/src/connectors/infinite-hits/connectInfiniteHits.ts
+++ b/src/connectors/infinite-hits/connectInfiniteHits.ts
@@ -157,18 +157,24 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
     let showMore: () => void;
     let sendEvent;
     let bindEvent;
-    const getFirstReceivedPage = (state: SearchParameters) => {
+    const getFirstReceivedPage = (
+      state: SearchParameters,
+      cachedHits: InfiniteHitsCachedHits
+    ) => {
       const { page = 0 } = state;
-      const pages = Object.keys(cache.read({ state }) || {}).map(Number);
+      const pages = Object.keys(cachedHits).map(Number);
       if (pages.length === 0) {
         return page;
       } else {
         return Math.min(page, ...pages);
       }
     };
-    const getLastReceivedPage = (state: SearchParameters) => {
+    const getLastReceivedPage = (
+      state: SearchParameters,
+      cachedHits: InfiniteHitsCachedHits
+    ) => {
       const { page = 0 } = state;
-      const pages = Object.keys(cache.read({ state }) || {}).map(Number);
+      const pages = Object.keys(cachedHits).map(Number);
       if (pages.length === 0) {
         return page;
       } else {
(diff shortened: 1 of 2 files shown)


src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -281,17 +285,14 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
...
         (results.hits as any).__escaped = initialEscaped;
 
-        if (cachedHits === undefined) {
-          cachedHits = cache.read({ state }) || {};
-        }
-
+        const cachedHits = cache.read({ state }) || {};
         if (cachedHits![page] === undefined) {
           cachedHits![page] = results.hits;
           cache.write({ state, hits: cachedHits });
```
Haroenv, line 291: I don't think these exclamation marks are still needed now

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -207,18 +203,20 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
           widgetType: this.$$type!,
         });
 
+        const cachedHits = cache.read({ state: helper.state }) || {};
+        const pageNumbers = Object.keys(cachedHits).map(Number);
+        firstReceivedPage = Math.min(...pageNumbers);
```
Haroenv, line 208: moving this to a global will be harder for getWidgetRenderState I think, not sure what the alternative is though.

Could it be stored in the cache too maybe? (note that we would need to have some kind of migration or something like that for active caches, although that won't be an issue, since we use sessionStorage)

Haroenv: we should put in the release notes of the version containing this that if you use localStorage will required to be cleaned on the first visit of this new page if we go for saving it in cache), not sure what's best

Haroenv (commented):
overall this looks all right!

src/connectors/infinite-hits/connectInfiniteHits.ts
Haroenv:
```suggestion
        if (cachedHits[page] === undefined) {
          cachedHits[page] = results.hits;
          cache.write({ state, hits: cachedHits });
```

src/connectors/infinite-hits/connectInfiniteHits.ts
eunjae-lee: You're talking about `firstReceivedPage` and `lastReceivedPage`, right?
https://github.com/algolia/instantsearch.js/pull/4534/commits/83cf619f49fd85a906f4971da382a1135127f447

what if we do not store them, and calculate every time when needed?

src/connectors/infinite-hits/connectInfiniteHits.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4534/commits/2bd1aa21c85208a527314d5c413c312be7fe6cee

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -153,42 +153,41 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
...
     let sendEvent;
     let bindEvent;
-
-    const getFirstReceivedPage = () =>
-      Math.min(...Object.keys(cachedHits || {}).map(Number));
-    const getLastReceivedPage = () =>
-      Math.max(...Object.keys(cachedHits || {}).map(Number));
+    const getFirstReceivedPage = (state: SearchParameters) => {
+      const { page = 0 } = state;
+      const pages = Object.keys(cache.read({ state }) || {}).map(Number);
```
Haroenv, line 162: we probably don't want to read the cache multiple times, this function should accept the result of cache IMO

eunjae-lee: Yeah, we could do that, but I initially didn't do it because `getShowPrevious` and `getShowMore` are calling `getFirstReceivedPage` and `getLastReceivedPage` and they don't have access to the result of cache.

Now that we kind of agree on _"do not store firstReceivedPage or lastReceivedPage in the upper scope"_ part, I've changed the implementation to provide `cachedHits` to here and there. https://github.com/algolia/instantsearch.js/pull/4534/commits/1e12727ca2a4a94a5459d6e66ba389a71f0a12e2

What do you think?

Haroenv: seems all right like this for me

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -234,32 +247,7 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
...
-        );
-        currentState.numericRefinements = filterEmptyRefinements(
-          currentState.numericRefinements
-        );
-
-        if (!isEqual(currentState, prevState)) {
-          cachedHits = cache.read({ state }) || {};
-          prevState = currentState;
-        }
+        const { page = 0 } = state;
```
yannickcr, line 250: I think the above comment is no longer needed.

Haroenv (approved):

src/connectors/infinite-hits/connectInfiniteHits.ts
eunjae-lee: good catch!

yannickcr (approved):

PR #4535, opened 2020-10-13
feat(infinite-hits): implement `getRenderState` and `getWidgetRenderState`

yannickcr:
This implements the `getRenderState` and `getWidgetRenderState` widget lifecycle hooks in `infiniteHits`.


Haroenv:
let's make sure this works with #4534 if we rebase :)

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -178,107 +176,116 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
...
+            cache.write({ state, hits: cachedHits });
+          }
+
+          hits = extractHitsFromCachedHits(cachedHits!);
+        }
+
+        return {
+          hits,
+          results,
+          showPrevious: getShowPrevious(helper),
```
Haroenv, line 282: these references are no longer stable now I think

eunjae-lee: what do you mean by no longer stable? could you explain more?

Haroenv: if you do addEventListener on an early render, the function reference isn't the same as another one, meaning that removeEventListener wouldn't work I think (although I always get confused about this)

yannickcr: Understood, that's very error prone. I updated the tests to check that we get the same function at the first and second renders.

Haroenv:
@yannickcr could you rebase now the InstantSearch clearing fix is merged on base?

devdreamsolution (approved):

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -968,6 +968,134 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
+      const hits = [
+        { objectID: '1', name: 'name 1' },
+        { objectID: '2', name: 'name 2' },
+      ];
+
+      const results = new SearchResults(helper.state, [
+        createSingleSearchResponse({ hits, queryID: 'theQueryID' }),
+      ]);
+
+      const renderState2 = infiniteHitsWidget.getRenderState!(
```
Haroenv, line 1007: this can be two separate tests if the second uses `getInitializedWidget`

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -968,6 +968,134 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
+      const hits = [
+        { objectID: '1', name: 'name 1' },
+        { objectID: '2', name: 'name 2' },
+      ];
+
+      const results = new SearchResults(helper.state, [
+        createSingleSearchResponse({ hits, queryID: 'theQueryID' }),
+      ]);
+
+      const renderState2 = infiniteHitsWidget.getWidgetRenderState!(
```
Haroenv, line 1070: same here, one test for init, one for render will be clearer if we ever break anything on it

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -207,92 +207,110 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
...
 
-        results.hits = transformItems(results.hits);
-
-        // Make sure the escaped tag stays after mapping over the hits.
-        // This prevents the hits from being double-escaped if there are multiple
-        // hits widgets mounted on the page.
-        (results.hits as any).__escaped = initialEscaped;
+        const widgetRenderState = this.getWidgetRenderState(renderOptions);
 
         const cachedHits = cache.read({ state }) || {};
```
Haroenv, line 225: the cache read is still here, but not the cache write, is that expected?

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -207,92 +207,110 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
...
+          results.hits = addQueryID(results.hits, results.queryID);
+
+          results.hits = transformItems(results.hits);
+
+          // Make sure the escaped tag stays after mapping over the hits.
+          // This prevents the hits from being double-escaped if there are multiple
+          // hits widgets mounted on the page.
+          (results.hits as any).__escaped = initialEscaped;
+
+          if (cachedHits[page] === undefined) {
```
Haroenv, line 288: the cache read should probably move from render to before here, wdyt?

src/connectors/infinite-hits/connectInfiniteHits.ts
Haroenv: maybe we can expose it from getWidgetRenderState as well, something like "currentPage", to use for sendEvent?

It makes sense not to compute the cache read multiple times

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
yannickcr: I don't really see the advantage of doing this, we will still have 2 renders to do in the second test anyway. Also since there is a relation between the 2 renders (the functions we got from the second render must match the ones we received in the first), I think it is clearer to keep everything in the same test.

src/connectors/infinite-hits/connectInfiniteHits.ts
yannickcr: It is expected, we need it for the `sendEvent` call that must stay in `render`.
One alternative way I thought of was to return `cachedHits` from `getWidgetRenderState` but I don't know if it is better or not.

src/connectors/infinite-hits/connectInfiniteHits.ts
yannickcr: I did not understood what you meant by that, can you explain more precisely?

(all this side effects like update of the `results` object and `cache.write` should not be in `getWidgetRenderState` which should be a pure getter but I did not wanted to mess with the logic here for now)

eunjae-lee: As @Haroenv and @francoischalifour discussed in another PR, when we get rid of `init` in the future, it's natural that `getWidgetRenderState` is the place where side effects occur.

Although `getWidgetRenderState` has side effects, I think it's okay as long as it's idempotent. In this case, it's okay to call `getWidgetRenderState` multiple times as long as it reads the cache and writes the same one (not changing the cache). And, of course, `getWidgetRenderState` shouldn't trigger `sendEvent` inside it.

To sum up, I think it's good to have `cache.read` inside `getWidgetRenderState`. And we add another property to `InfiniteHitsWidgetRenderState`, which is something like `currentPageHits`. If so, inside `render` we can get `currentPageHits` from the render state and call `sendEvent`.

What do you think?

src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -245,6 +247,7 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
 
       getWidgetRenderState({ results, helper, state, instantSearchInstance }) {
         let isFirstPage;
+        let currentPageHits;
```
eunjae-lee, line 250: We'd better make this `[]` by default, or to make the type optional, because `getWidgetRenderState` here returns `undefind` as `currentPageHits` when there's no results.

eunjae-lee (approved):

PR #4536, opened 2020-10-13
feat(range): implement `getRenderState` and `getWidgetRenderState`

yannickcr:
This implements the `getRenderState` and `getWidgetRenderState` widget lifecycle hooks in `range`.

### DIFF
--- a/src/connectors/range/connectRange.js
+++ b/src/connectors/range/connectRange.js
@@ -191,32 +191,35 @@ export default function connectRange(renderFn, unmountFn = noop) {
         };
       },
 
-      init({ helper, instantSearchInstance }) {
-        const stats = {};
-        const currentRange = this._getCurrentRange(stats);
-        const start = this._getCurrentRefinement(helper);
-
+      init(initOptions) {
         renderFn(
           {
-            // On first render pass an empty range
-            // to be able to bypass the validation
-            // related to it
-            refine: this._refine(helper, {}),
-            format: rangeFormatter,
-            range: currentRange,
-            widgetParams: {
-              ...widgetParams,
-              precision,
-            },
-            start,
-            instantSearchInstance,
+            ...this.getWidgetRenderState(initOptions),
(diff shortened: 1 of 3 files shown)


src/connectors/range/connectRange.js
```diff
@@ -226,20 +229,27 @@ export default function connectRange(renderFn, unmountFn = noop) {
...
-            },
-            start,
-            instantSearchInstance,
+        let refine;
+
+        if (!results) {
+          // On first render pass an empty range
+          // to be able to bypass the validation
+          // related to it
+          refine = this._refine(helper, {});
```
Haroenv, line 238: don't we have access to the current refinement already? as far as I can tell the validation gets done only once you call refine, and that would be the same issue as empty object, right? Or do I misunderstand it?

src/connectors/range/connectRange.js
```diff
@@ -191,32 +191,35 @@ export default function connectRange(renderFn, unmountFn = noop) {
...
+            instantSearchInstance: renderOptions.instantSearchInstance,
+          },
+          false
+        );
+      },
+
+      getRenderState(renderState, renderOptions) {
+        return {
+          ...renderState,
+          range: this.getWidgetRenderState(renderOptions),
```
Haroenv, line 217: range should be scoped per attribute

src/connectors/range/connectRange.js
yannickcr: I have no idea here, I just took back the existing logic.

src/types/widget.ts
```diff
@@ -205,6 +205,7 @@ export type IndexRenderState = Partial<{
     >;
   };
   hits: WidgetRenderState<HitsRendererOptions, HitsConnectorParams>;
+  range: any; // @TODO type when connectRange is migrated to TypeScript
```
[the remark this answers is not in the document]
yannickcr: I added some typings, hope they are correct.

Haroenv (approved):
I think this is good 👍

PR #4537, opened 2020-10-13
fix(types): fix type errors

eunjae-lee:
**Summary**

This PR fixes a bunch of type errors in `feat/render-state` branch. This PR is not fixing all the type errors, because there will be separate PRs for connectors and corresponding widgets.

**TODO**
- [x] fix a type error in `breadcrumb` widget (https://github.com/algolia/instantsearch.js/blob/fix/type-errors/src/widgets/breadcrumb/breadcrumb.tsx#L121:L121)
- [x] fix `places` widget (https://github.com/algolia/instantsearch.js/blob/fix/type-errors/src/widgets/places/places.ts#L34:L34)

### DIFF
--- a/src/widgets/analytics/analytics.ts
+++ b/src/widgets/analytics/analytics.ts
@@ -242,8 +242,11 @@ const analytics: AnalyticsWidget = function analytics(widgetParams) {
       }
     },
 
-    getRenderState(renderState) {
-      return renderState;
+    getRenderState(renderState, renderOptions) {
+      return {
+        ...renderState,
+        analytics: this.getWidgetRenderState(renderOptions),
+      };
     },
 
     getWidgetRenderState() {
--- a/src/widgets/analytics/analytics.ts
+++ b/src/widgets/analytics/analytics.ts
@@ -241,6 +241,16 @@ const analytics: AnalyticsWidget = function analytics(widgetParams) {
         window.removeEventListener('beforeunload', onUnload);
       }
     },
+
+    getRenderState(renderState) {
+      return renderState;
+    },
+
+    getWidgetRenderState() {
+      return {
+        widgetParams,
+      };
+    },
   };
 };
 
--- a/src/widgets/analytics/analytics.ts
+++ b/src/widgets/analytics/analytics.ts
@@ -61,7 +61,7 @@ export type AnalyticsWidgetParams = {
 
 const withUsage = createDocumentationMessageGenerator({ name: 'analytics' });
 
-export type AnalyticsWidget = WidgetFactory<{}, AnalyticsWidgetParams>;
+export type AnalyticsWidget = WidgetFactory<{}, {}, AnalyticsWidgetParams>;
 
 const analytics: AnalyticsWidget = function analytics(widgetParams) {
   const {
--- a/src/widgets/analytics/analytics.ts
+++ b/src/widgets/analytics/analytics.ts
@@ -1,6 +1,6 @@
 import { SearchParameters, SearchResults } from 'algoliasearch-helper';
 import { createDocumentationMessageGenerator } from '../../lib/utils';
-import { WidgetFactory } from '../../types';
+import { Widget } from '../../types';
 
 export type AnalyticsWidgetParamsPushFunction = (
   /**
@@ -61,7 +61,7 @@ export type AnalyticsWidgetParams = {
 
 const withUsage = createDocumentationMessageGenerator({ name: 'analytics' });
 
-export type AnalyticsWidget = WidgetFactory<{}, {}, AnalyticsWidgetParams>;
+export type AnalyticsWidget = (widgetParams: AnalyticsWidgetParams) => Widget;
 
 const analytics: AnalyticsWidget = function analytics(widgetParams) {
   const {
@@ -240,16 +240,6 @@ const analytics: AnalyticsWidget = function analytics(widgetParams) {
         document.removeEventListener('click', onClick);
         window.removeEventListener('beforeunload', onUnload);
       }
-    },
-
-    getRenderState(renderState) {
-      return renderState;
-    },
-
-    getWidgetRenderState() {
-      return {
-        widgetParams,
-      };
     },
   };
 };
--- a/src/widgets/analytics/analytics.ts
+++ b/src/widgets/analytics/analytics.ts
@@ -1,6 +1,6 @@
 import { SearchParameters, SearchResults } from 'algoliasearch-helper';
 import { createDocumentationMessageGenerator } from '../../lib/utils';
-import { Widget } from '../../types';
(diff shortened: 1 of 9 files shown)


src/widgets/configure/configure.ts
```diff
@@ -1,15 +1,25 @@
...
  * to enable when the widget mounts.
  */
-export type ConfigureWidgetOptions = PlainSearchParameters;
+export type ConfigureWidgetParams = PlainSearchParameters;
 
-export type ConfigureWidget = WidgetFactory<{}, ConfigureWidgetOptions>;
+export type ConfigureWidget = (
+  widgetParams: ConfigureConnectorParams['searchParameters']
+) => Widget<{
+  renderState: WidgetRenderState<
```
Haroenv, line 18: why is this different to the others?

src/widgets/places/places.ts
```diff
@@ -133,6 +133,16 @@ const placesWidget: WidgetFactory<{}, PlacesWidgetOptions> = (
         .setQueryParameter('aroundLatLngViaIP', false)
         .setQueryParameter('aroundLatLng', position || undefined);
     },
+
+    getRenderState(renderState) {
```
Haroenv, line 137: this is all right, but looking at this file, I can see it's setting parameters in init, isn't that duplicate with getWidgetSearchParameters? can probably just be done in a separate PR

src/widgets/configure/configure.ts
eunjae-lee: I forgot to mention it. If it was

```js
export type ConfigureWidget = WidgetFactory<
  ConfigureRendererOptions,
  ConfigureConnectorParams,
  ConfigureWidgetParams
>;
```

(just like others)

then users would've done something like

```js
configure({ searchParameters: { ... });
```

but inside the widget, we are actually passing `widgetParams` to `searchParameters`

```js
const configure: ConfigureWidget = function configure(widgetParams) {
  // This is a renderless widget that falls back to the connector's
  // noop render and unmount functions.
  const makeWidget = connectConfigure(noop);

  return makeWidget({ searchParameters: widgetParams });
};
```

So `widgetParams` in configure widget is a kind of irregular.

src/widgets/places/places.ts
eunjae-lee: Hmm. If I understand correctly, when `placesAutocomplete.on('change'` happens in `init`, it should store the information somewhere else and let `getWidgetSearchParameters` read it and return the updated search parameters, right?
But how does 'change' event listener of placesAutocomplete trigger a re-render?

Haroenv: Let's leave it out of this PR and think about it another time

src/widgets/analytics/analytics.ts
```diff
@@ -241,6 +241,16 @@ const analytics: AnalyticsWidget = function analytics(widgetParams) {
...
+
+    getRenderState(renderState) {
+      return renderState;
+    },
+
+    getWidgetRenderState() {
+      return {
+        widgetParams,
+      };
+    },
```
Haroenv, line 253: since getWidgetRenderState is optional, I think it could make sense for this widget not to have it, and leave it underpowered in favour for insights middleware

Haroenv (approved):
this is good, although maybe analytics & places should not have the life cycles at all, since they only implement them a little & the life cycles are optional

src/widgets/analytics/analytics.ts
yannickcr: Same for the `places` widget, no ?

eunjae-lee: As long as we type those widgets like

```js
export type AnalyticsWidget = WidgetFactory<{}, {}, AnalyticsWidgetParams>;
```

TS will complain if there's no getWidgetRenderState() or getRenderState().

1. We can provide a dummy functions as-is
or
2. We can update their types to something like `() => Widget` (not providing `renderState` as a generic to Widget)

What do you think? @Haroenv @yannickcr

yannickcr: Mmm, I'm not sure, but that's weird we have to do this for some optional methods :thinking: 

Question: How a user must type a custom widget ? And will he get the same issue ?

Haroenv: that's because the WidgetRender params should be the last in the WidgetFactory type so you can type it as `none` here

eunjae-lee: Not sure if I follow you @Haroenv 
If user is going to use `WidgetFactory` type instead of `Widget`, I don't see a way to not implement `getWidgetRenderState()` and `getRenderState`() without a TS error.

Haroenv: I see what's going on now, since this is WidgetFactory, and not Widget, however if you put `never` for the `TRendererOptions` generic argument, it should mean you can't have the life cycles (as expected)

eunjae-lee: putting `never` asked for 

```ts
getWidgetRenderState: (
  renderOptions
) => never;
```

I guess on the user land, they don't need to use `WidgetFactory`, which is not so beneficial to users:

```ts
export type WidgetFactory<TRendererOptions, TConnectorParams, TWidgetParams> = (
  widgetParams: TConnectorParams & TWidgetParams
) => Widget<{
  renderState: WidgetRenderState<TRendererOptions, TWidgetParams>;
}>;
```

I've updated the implementation just to use `Widget`.
What do you think?

Haroenv (approved):
👌

Haroenv (changes requested):
sorry for being inconsistent, but I think with #4541 it makes sense to add params / getWidgetRenderState to these widgets too. Sorry!

eunjae-lee:
> sorry for being inconsistent, but I think with #4541 it makes sense to add params / getWidgetRenderState to these widgets too. Sorry!

No worries! It makes sense :)

src/widgets/analytics/analytics.ts
```diff
@@ -241,6 +241,16 @@ const analytics: AnalyticsWidget = function analytics(widgetParams) {
         window.removeEventListener('beforeunload', onUnload);
       }
     },
+
+    getRenderState(renderState) {
+      return renderState;
```
Haroenv, line 246: you need to add the local getWidgetRenderState call here :)

eunjae-lee: you're right https://github.com/algolia/instantsearch.js/pull/4537/commits/949db21de0cdff9e843bfddb38b5c2d9c697a169

Haroenv (approved):
complicated PR, but good to go now I think

eunjae-lee:
Yeah didn't see this coming. I should've tackled one thing at a time :(

PR #4549, opened 2020-10-27
feat(refinement-list): implement `getRenderState` and `getWidgetRenderState`

yannickcr:
This implements the `getRenderState` and `getWidgetRenderState` widget lifecycle hooks in `refinementList`.

### DIFF
--- a/src/connectors/refinement-list/__tests__/connectRefinementList-test.js
+++ b/src/connectors/refinement-list/__tests__/connectRefinementList-test.js
@@ -5,6 +5,11 @@ import jsHelper, {
 import { TAG_PLACEHOLDER } from '../../../lib/escape-highlight';
 import connectRefinementList from '../connectRefinementList';
 import { createInstantSearch } from '../../../../test/mock/createInstantSearch';
+import {
+  createInitOptions,
+  createRenderOptions,
+} from '../../../../test/mock/createWidget';
+import { createSingleSearchResponse } from '../../../../test/mock/createAPIResponse';
 
 describe('connectRefinementList', () => {
   const createWidgetFactory = () => {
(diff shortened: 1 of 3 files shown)


src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -300,73 +300,80 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
...
+
+          const maxValuesPerFacetConfig = state.maxValuesPerFacet;
+          const currentLimit = this.getLimit();
+          // If the limit is the max number of facet retrieved it is impossible to know
+          // if the facets are exhaustive. The only moment we are sure it is exhaustive
+          // is when it is strictly under the number requested unless we know that another
+          // widget has requested more values (maxValuesPerFacet > getLimit()).
+          // Because this is used for making the search of facets unable or not, it is important
+          // to be conservative here.
+          hasExhaustiveItems =
```
Haroenv, line 356: where's this variable used now?

Haroenv: looks like L210 and L193, but I think it might be better if we put this in a function or something? Would be a bit more recalculating, but not sure if that's a big deal

Haroenv: it's also passed to renderFn in showMore, but not in getWidgetRenderState. We should make that consistent and expose it here + use getWidgetRenderState in the renderFn in showMore

Haroenv (changes requested):
let's use the getWidgetRenderState in showMore rendering too

yannickcr:
Here's my attempt to remove the `render` function alltogether so we can use `getWidgetRenderState` everywhere https://github.com/algolia/instantsearch.js/pull/4549/commits/74ab771d35aacf882ebe185defd960510419cd6e

The only thing that bother me is the need to pass a `isFromSearch` flag in `renderOptions` (an alternative way is to set it in the upper scope but I'm not sure it is better, thought ?)

(The tests pass with some small changes, hope I did not break anything, this file is kind of a mess and a rewrite wouldn't be too much)

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -154,126 +154,62 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
...
-          isShowingMore,
-          canToggleShowMore,
-          toggleShowMore,
-          hasExhaustiveItems,
-        },
-        isFirstSearch
-      );
-    };
-
     /* eslint-disable max-params */
```
Haroenv, line 163: is this disable still useful? if it's not public api, could we make it take an object as arguments instead?

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -154,126 +154,62 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
...
+              tags
+            )
+            .then(results => {
+              renderFn({
+                ...this.getWidgetRenderState({
+                  results,
+                  state,
+                  createURL,
+                  instantSearchInstance,
+                  isFromSearch: true,
```
Haroenv, line 206: I'm not sure, but maybe isFromSearch could be part of a second argument to getWidgetRenderState; something like "localRenderState"?

devdreamsolution (approved):

src/connectors/refinement-list/connectRefinementList.js
yannickcr: Yep, it is not useful anymore, I removed it.

src/connectors/refinement-list/connectRefinementList.js
yannickcr: I moved it to the upper scope instead, tell me if it is ok for you.

src/connectors/refinement-list/__tests__/connectRefinementList-test.js
```diff
@@ -2168,6 +2173,211 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
...
+      const renderState2 = refinementListWidget.getRenderState(
+        {},
+        createRenderOptions({
+          helper,
+          state: helper.state,
+          results,
+        })
+      );
+
+      expect(renderState2.refinementList).toEqual({
```
Haroenv, line 2236: could we split these tests up in init / render separately?

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -155,137 +155,93 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
...
-              isFromSearch: true,
-              isFirstSearch: false,
+    let toggleShowMore;
+
+    const createSearchForFacetValues = function(helper) {
+      return (
+        state,
+        createURL,
+        instantSearchInstance,
+        isShowingMore
```
Haroenv, line 172: would maybe be simpler if these arguments were `renderOptions` so we can forward easily to getWidgetRenderState

src/connectors/refinement-list/connectRefinementList.js
```diff
@@ -155,137 +155,85 @@ export default function connectRefinementList(renderFn, unmountFn = noop) {
...
+              renderFn(
+                {
+                  ...this.getWidgetRenderState({
+                    ...renderOptions,
+                    results: lastResultsFromMainSearch,
+                  }),
+                  items: normalizedFacetValues,
+                  canToggleShowMore,
+                  canRefine: true,
+                  instantSearchInstance,
```
Haroenv, line 229: this is an interesting pattern, but it makes sense, since getWidgetRenderState can't know about the state at this point. 

What if the widget renders again a second time when it's in search for facet values mode, won't it be overridden by the main items?

yannickcr: I updated an existing test to check for the renderingOptions between renders https://github.com/algolia/instantsearch.js/pull/4549/commits/6ebeb4c76cd3c64d9a57460be0e50b53dab82fac

I ran the test on the previous implementation too, and got the same results (minus `toggleShowMore` that was `undefined` in `renderingOptions3` in the previous implementation, but it was a bug, now fixed).

Haroenv: I don't think this tests an external render

yannickcr: What do you mean by an external render ?

eunjae-lee (commented):
I think the PR looks good overall, but let's wait for https://github.com/algolia/instantsearch.js/pull/4549/files#r526950706

src/connectors/refinement-list/connectRefinementList.js
Haroenv: a render that comes from another widget, for example a new search in search box triggering a render to refinementList. I want to make sure that the searched items still show, and it doesn't fall back to the items from hits

Haroenv: I actually checked this now in the current version, and it behaves the same (query gets reset and no longer in search mode). It seems like that behaviour is on purpose, so this issue is to discuss separately!

Haroenv (approved):

src/connectors/refinement-list/connectRefinementList.js
yannickcr: Ok, lets see that later on then.

eunjae-lee (approved):

PR #4551, opened 2020-10-28
feat(poweredBy): getWidgetRenderState

Haroenv:
DX-206


eunjae-lee:
It's too bad that it's not considered as "renamed and modified". Once this merged, will the file history of "connectPoweredBy.ts" start only from here?

shortcuts:
> It's too bad that it's not considered as "renamed and modified". Once this merged, will the file history of "connectPoweredBy.ts" start only from here?

That's what I was wondering, I don't see any name changes on my side. The commit history will start from the new file.

eunjae-lee:
Okay, I just did [a test](https://github.com/algolia/instantsearch.js/pull/4552/files).
If you rename js to ts first, and commit them.
And in a separate commit, make actual changes there.
Then git will be able to track them correctly.

Haroenv:
making a separate commit won't help here since we're in a batch PR which will be squashed at the end unfortunately @eunjae-lee

eunjae-lee:
> making a separate commit won't help here since we're in a batch PR which will be squashed at the end unfortunately @eunjae-lee

ehh right.

eunjae-lee:
Then can you split the commits in this PR just to help us review the changes even though it will be squashed at the end?
It's quite difficult to figure out the changes @Haroenv

Haroenv:
in https://github.com/algolia/instantsearch.js/pull/4551/commits/0435d689bf8764f855a0cf01029846862a455613#diff-57893d658ec377d112dce456273d95a8dbdd6968bef10c1c65eb6d1b4272fcf1 you can see most of the changes to the connector, the test was only additive

eunjae-lee (approved):
Looks good to me!

PR #4556, opened 2020-10-30
chore(yarn): clean lockfile

Haroenv:
**Summary**

I used https://github.com/atlassian/yarn-deduplicate and ran yarn install again afterwards (https://github.com/atlassian/yarn-deduplicate/issues/5)

**Result**

Basically what this does is remove a bunch of overlappign versions of dependencies, which will result in a faster install & smaller node modules


yannickcr:
Nice !
However, it seems that the bundle size increased from 64.36KB to 64.98KB.
I don't think this was indented, what could explain this change ?

francoischalifour:
Perhaps the Babel upgrades changed a few transpilation outputs?

Haroenv:
the bundlesize change was because of uglify, they must have released something weird in 3.7.x that wasn't yet behaving like this in 3.6.0. I didn't want to do a deep investigation, so I forced the older version

francoischalifour (approved):

PR #4558, opened 2020-10-30
feat(panel): smoother accessing of widgetRenderState

Haroenv:
this kinda works @eunjae-lee, although some parts still seem to be inferred as any. The public API works though

I also made a small change: not using widgetRenderState as key, but spreading the renderer. This way the usage is so much smoother (and you can use it for the templates too, which comes up from time to time)

### DIFF
--- a/src/widgets/panel/panel.tsx
+++ b/src/widgets/panel/panel.tsx
@@ -111,14 +111,14 @@ export type PanelWidgetOptions<TWidget extends WidgetFactory<any, any, any>> = {
 const withUsage = createDocumentationMessageGenerator({ name: 'panel' });
 const suit = component('Panel');
 
-const renderer = ({
+const renderer = <TWidget extends WidgetFactory<any, any, any>>({
   containerNode,
   bodyContainerNode,
   cssClasses,
   templates,
 }) => ({ options, hidden, collapsible, collapsed }) => {
   render(
-    <Panel
+    <Panel<TWidget>
       cssClasses={cssClasses}
       hidden={hidden}
       collapsible={collapsible}
@@ -222,7 +222,7 @@ const panel: PanelWidget = widgetParams => {
         </svg>`,
     };
 
-    const renderPanel = renderer({
+    const renderPanel = renderer<typeof widgetFactory>({
       containerNode: getContainerNode(container),
       bodyContainerNode,
       cssClasses,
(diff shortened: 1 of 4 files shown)


src/widgets/panel/panel.tsx
```diff
@@ -259,9 +259,8 @@ const panel: PanelWidget = widgetParams => {
         const [renderOptions] = args;
 
         const options = {
+          ...(widget.getWidgetRenderState?.(renderOptions) || {}),
```
Haroenv, line 262: possible since I reuse the type for TWidget for widgetFactory

stories/panel.stories.ts
```diff
@@ -151,11 +151,13 @@ storiesOf('Basics/Panel', module)
     withHits(
       ({ search, container }) => {
         const breadcrumbInPanel = panel<typeof breadcrumb>({
-          collapsed({ widgetRenderState }) {
-            return widgetRenderState.canRefine;
+          collapsed({ canRefine }) {
+            return canRefine === false;
```
Haroenv, line 155: your example used to be wrong

stories/panel.stories.ts
```diff
@@ -151,11 +151,13 @@ storiesOf('Basics/Panel', module)
     withHits(
       ({ search, container }) => {
         const breadcrumbInPanel = panel<typeof breadcrumb>({
-          collapsed({ widgetRenderState }) {
-            return widgetRenderState.canRefine;
+          collapsed({ canRefine }) {
```
francoischalifour, line 154: This usage is much better and eventually what we want to enable in next major version so if there can't be collision with the IS instance (which is also passed and spread) that's great!

src/widgets/panel/panel.tsx
```diff
@@ -78,11 +78,11 @@ export type PanelTemplates = {
...
 export type PanelRenderOptions<
   TWidget extends WidgetFactory<any, any, any>
-> = RenderOptions & {
-  widgetRenderState: ReturnType<
-    Exclude<ReturnType<TWidget>['getWidgetRenderState'], undefined>
-  >;
-};
+> = RenderOptions & ReturnType<TWidget>['getWidgetRenderState'] extends (
+  renderOptions: any
+) => infer TRenderState
```
eunjae-lee, line 83: didn't know about `infer`. looks nice 👍

eunjae-lee (approved):
looks nice!

stories/panel.stories.ts
eunjae-lee: Yeah, I didn't realize there cannot be collision. This looks so much better.

Haroenv: as far as I can tell there's no collision here, so this doesn't even need a major :)

eunjae-lee: Yeah, it makes sense there is no collision :)

PR #4561, opened 2020-11-03
test(renderState): add tests for getWidgetRenderState at connectAutocomplete

eunjae-lee:
**Summary**

This PR adds the missing tests for `getWidgetRenderState` at `connectAutocomplete`.

DX-1356

### DIFF
--- a/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
+++ b/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
@@ -532,9 +532,20 @@ search.addWidgets([
         createRenderOptions({ helper })
       );
 
+      const hits = [];
+      // @ts-ignore-next-line
+      hits.__escaped = true;
+
       expect(renderState).toEqual({
         currentRefinement: 'query',
-        indices: expect.any(Array),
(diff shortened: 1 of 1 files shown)


src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -485,6 +485,62 @@ search.addWidgets([
...
+
+      autocomplete.init!(createInitOptions());
+
+      const renderState = autocomplete.getWidgetRenderState(
+        createRenderOptions({ helper })
+      );
+
+      expect(renderState).toEqual({
+        currentRefinement: 'query',
+        indices: expect.any(Array),
```
Haroenv, line 537: I think we should assert what's actually in this indices array. In your case it's probably an empty array?

eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4561/commits/924eabc7c18a81d9c6e156c5cd50c96ac5877617 what do you think?

Haroenv (approved):

PR #4564, opened 2020-11-03
feat(getWidgetRenderState): support connectGeoSearch

Haroenv:
I didn't migrate the connector to typescript, so had to write some types hardcoded in the widget file

### DIFF
--- a/src/connectors/geo-search/connectGeoSearch.js
+++ b/src/connectors/geo-search/connectGeoSearch.js
@@ -172,86 +172,97 @@ const connectGeoSearch = (renderFn, unmountFn = noop) => {
 
     let sendEvent;
 
-    const init = initArgs => {
-      const { state, helper, instantSearchInstance } = initArgs;
-      const isFirstRendering = true;
+    return {
+      $$type,
 
-      sendEvent = createSendEventForHits({
-        instantSearchInstance,
-        index: helper.getIndex(),
-        widgetType: $$type,
-      });
+      init(initArgs) {
+        const { helper, instantSearchInstance } = initArgs;
+        const isFirstRendering = true;
 
-      widgetState.internalToggleRefineOnMapMove = createInternalToggleRefinementOnMapMove(
-        noop,
-        initArgs
-      );
-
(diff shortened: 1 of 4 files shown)


src/connectors/geo-search/connectGeoSearch.js
```diff
@@ -172,86 +172,97 @@ const connectGeoSearch = (renderFn, unmountFn = noop) => {
...
-          hasMapMoveSinceLastRefine,
-          widgetParams,
+    return {
+      $$type,
+
+      init(initArgs) {
+        const { helper, instantSearchInstance } = initArgs;
+        const isFirstRendering = true;
+
+        sendEvent = createSendEventForHits({
```
eunjae-lee, line 182: In `index` widget, it runs `getRenderState` of all widgets before `init`. It means it will initially have `sendEvent: undefined`. I guess we're okay with that, since we call `init` right after that?

Haroenv: I don't see how sendEvent could be defined before init since you need access to the InstantSearch. What did we do for the other sendEvent connectors @eunjae-lee ?

eunjae-lee: I think your approach here makes sense. Before init, sendEvent means nothing.
I'm just reminded that I should change my previous implementations to this.

What I did before is to assign them inside getWidgetRenderState:

```js
getWidgetRenderState() {
  if (!sendEvent) {
    sendEvent = ...
  }

  return {
    ...
  }
}
```

We just need to make sure when we tests the return of `getWidgetRenderState` **BEFORE** init will have undefined sendEvent.

Haroenv: I prefer that implementation, pushed the change

eunjae-lee: hmm that's safer and especially matches the type, though. Okay, then let's go with it.

eunjae-lee (approved):
looks good 👍

shortcuts (approved):

PR #4565, opened 2020-11-03
feat(stats): implement `getRenderState` and `getWidgetRenderState`

shortcuts:
This implements the `getRenderState` and `getWidgetRenderState` widget lifecycle hooks in `stats`.

### DIFF
--- a/src/connectors/stats/__tests__/connectStats-test.js
+++ b/src/connectors/stats/__tests__/connectStats-test.js
@@ -1,5 +1,11 @@
 import jsHelper from 'algoliasearch-helper';
 const SearchResults = jsHelper.SearchResults;
+import { createSearchClient } from '../../../../test/mock/createSearchClient';
+import { createSingleSearchResponse } from '../../../../test/mock/createAPIResponse';
+import {
+  createInitOptions,
+  createRenderOptions,
+} from '../../../../test/mock/createWidget';
 
 import connectStats from '../connectStats';
 
@@ -30,6 +36,110 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
           dispose: expect.any(Function),
         })
       );
+    });
+  });
+
+  describe('getRenderState', () => {
+    test('returns the render state', () => {
+      const renderFn = jest.fn();
+      const unmountFn = jest.fn();
+      const createStats = connectStats(renderFn, unmountFn);
+      const stats = createStats();
+      const helper = jsHelper(createSearchClient(), 'indexName', {
+        index: 'indexName',
+      });
+      helper.search = jest.fn();
+
+      const renderState1 = stats.getRenderState(
+        { stats: {} },
+        createInitOptions({ helper })
+      );
+
+      expect(renderState1.stats).toEqual({
+        hitsPerPage: undefined,
+        nbHits: 0,
+        nbPages: 0,
+        page: 0,
+        processingTimeMS: -1,
+        query: '',
+        widgetParams: {},
+      });
+
+      stats.init(createInitOptions({ helper, state: helper.state }));
+
+      const results = new SearchResults(helper.state, [
+        createSingleSearchResponse(),
+      ]);
+
+      const renderState2 = stats.getRenderState(
+        { stats: {} },
+        createRenderOptions({
+          helper,
+          state: helper.state,
+          results,
+        })
+      );
+
+      expect(renderState2.stats).toEqual({
+        hitsPerPage: results.hitsPerPage,
+        nbHits: results.nbHits,
+        nbPages: results.nbPages,
+        page: results.page,
+        processingTimeMS: results.processingTimeMS,
+        query: results.query,
+        widgetParams: {},
+      });
+    });
+  });
+
+  describe('getWidgetRenderState', () => {
+    test('returns the widget render state', () => {
+      const renderFn = jest.fn();
+      const unmountFn = jest.fn();
+      const createStats = connectStats(renderFn, unmountFn);
+      const stats = createStats();
+      const helper = jsHelper(createSearchClient(), 'indexName', {
+        index: 'indexName',
+      });
+      helper.search = jest.fn();
+
+      const renderState1 = stats.getWidgetRenderState(
+        createInitOptions({ helper })
+      );
+
+      expect(renderState1).toEqual({
+        hitsPerPage: undefined,
+        nbHits: 0,
+        nbPages: 0,
+        page: 0,
+        processingTimeMS: -1,
+        query: '',
+        widgetParams: {},
+      });
+
+      stats.init(createInitOptions({ helper, state: helper.state }));
+
+      const results = new SearchResults(helper.state, [
+        createSingleSearchResponse(),
+      ]);
+
+      const renderState2 = stats.getWidgetRenderState(
+        createRenderOptions({
+          helper,
+          state: helper.state,
+          results,
+        })
+      );
+
+      expect(renderState2).toEqual({
+        hitsPerPage: results.hitsPerPage,
+        nbHits: results.nbHits,
+        nbPages: results.nbPages,
+        page: results.page,
+        processingTimeMS: results.processingTimeMS,
+        query: results.query,
+        widgetParams: {},
+      });
     });
   });
 
--- a/src/connectors/stats/__tests__/connectStats-test.js
+++ b/src/connectors/stats/__tests__/connectStats-test.js
@@ -120,7 +120,17 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
       stats.init(createInitOptions({ helper, state: helper.state }));
 
       const results = new SearchResults(helper.state, [
-        createSingleSearchResponse(),
+        createSingleSearchResponse({
+          hits: [
+            { brand: 'samsung', objectID: '1' },
+            { brand: 'apple', objectID: '2' },
+            { brand: 'sony', objectID: '3' },
+            { brand: 'benq', objectID: '4' },
+            { brand: 'dyson', objectID: '5' },
+          ],
+          hitsPerPage: 3,
+          query: 'apple',
+        }),
       ]);
 
       const renderState2 = stats.getWidgetRenderState(
--- a/src/connectors/stats/__tests__/connectStats-test.js
+++ b/src/connectors/stats/__tests__/connectStats-test.js
@@ -81,12 +81,12 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
       );
 
       expect(renderState2.stats).toEqual({
-        hitsPerPage: results.hitsPerPage,
-        nbHits: results.nbHits,
-        nbPages: results.nbPages,
-        page: results.page,
-        processingTimeMS: results.processingTimeMS,
-        query: results.query,
+        hitsPerPage: 20,
+        nbHits: 0,
+        nbPages: 0,
+        page: 0,
+        processingTimeMS: 0,
+        query: '',
         widgetParams: {},
       });
     });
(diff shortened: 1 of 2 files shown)


src/connectors/stats/__tests__/connectStats-test.js
```diff
@@ -33,6 +39,110 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
...
+        { stats: {} },
+        createInitOptions({ helper })
+      );
+
+      expect(renderState1.stats).toEqual({
+        hitsPerPage: undefined,
+        nbHits: 0,
+        nbPages: 0,
+        page: 0,
+        processingTimeMS: -1,
```
Haroenv, line 63: is this the current behaviour already for init? I wonder if it shouldn't be 0 instead; -1ms should never render I think

Haroenv: if it's already the case, that would be a small breaking change / fix, but normally we don't render on init anyway

src/connectors/stats/__tests__/connectStats-test.js
```diff
@@ -33,6 +39,110 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
...
+
+      const renderState2 = stats.getWidgetRenderState(
+        createRenderOptions({
+          helper,
+          state: helper.state,
+          results,
+        })
+      );
+
+      expect(renderState2).toEqual({
```
Haroenv, line 134: can this be two different tests?

src/connectors/stats/connectStats.js
```diff
@@ -88,5 +80,24 @@ export default function connectStats(renderFn, unmountFn = noop) {
...
+    getRenderState(renderState, renderOptions) {
+      return {
+        ...renderState,
+        stats: this.getWidgetRenderState(renderOptions),
+      };
+    },
+
+    getWidgetRenderState({ results, helper }) {
+      return {
+        hitsPerPage: results ? results.hitsPerPage : helper.state.hitsPerPage,
```
Haroenv, line 93: since it's the same condition for all, I think it might be better like this:

```
if (!results) { return { nbHits: 0...}}
return { nbHits: results.nbHits }
```

that will save a bit of computation

src/connectors/stats/connectStats.js
```diff
@@ -88,5 +80,24 @@ export default function connectStats(renderFn, unmountFn = noop) {
...
+        nbHits: results ? results.nbHits : 0,
+        nbPages: results ? results.nbPages : 0,
+        page: results ? results.page : helper.state.page || 0,
+        processingTimeMS: results ? results.processingTimeMS : -1,
+        query: results ? results.query : helper.state.query || '',
+        widgetParams,
```
shortcuts, line 99: I wasn't too sure about the multiple ternaries, should I just go for a simple `if:else` statement?

Haroenv: I think that will be cleaner & slightly faster

src/connectors/stats/__tests__/connectStats-test.js
shortcuts: `processingTimeMS` at `init` was already returning `-1`.
I think it make sense because returning `0` would mean that there was results, no?

src/connectors/stats/__tests__/connectStats-test.js
```diff
@@ -120,7 +120,17 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
...
+          hits: [
+            { brand: 'samsung', objectID: '1' },
+            { brand: 'apple', objectID: '2' },
+            { brand: 'sony', objectID: '3' },
+            { brand: 'benq', objectID: '4' },
+            { brand: 'dyson', objectID: '5' },
+          ],
+          hitsPerPage: 3,
+          query: 'apple',
+        }),
```
shortcuts, line 133: @Haroenv I've updated the `singleSearchResponse` so it's different than the `getRenderState` one, not sure if that's what you meant.

Haroenv: what I meant is that these two asserts should be in their own `it/test` function, since they don't test the same case (init/render). Sorry for being unclear!

shortcuts: Oh I got it now! Thanks, I'll update that :D

shortcuts: Let me know if that's what you had in mind https://github.com/algolia/instantsearch.js/pull/4565/commits/f4b71de2481934f52838d6b7434f55b384175adf

src/connectors/stats/__tests__/connectStats-test.js
```diff
@@ -119,6 +119,55 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
...
+      const renderFn = jest.fn();
+      const unmountFn = jest.fn();
+      const createStats = connectStats(renderFn, unmountFn);
+      const stats = createStats();
+      const helper = jsHelper(createSearchClient(), 'indexName', {
+        index: 'indexName',
+      });
+      helper.search = jest.fn();
+
+      const renderState1 = stats.getWidgetRenderState(
```
Haroenv, line 155: these variables don't need 1 or 2 added to them anymore now 👍

shortcuts: @Haroenv Let's hope this time this is what you meant 😂 
https://github.com/algolia/instantsearch.js/pull/4565/commits/5e07d5746c5bd60eeae6c4ce12fc8e975831e42e

src/connectors/stats/connectStats.js
```diff
@@ -88,5 +80,36 @@ export default function connectStats(renderFn, unmountFn = noop) {
...
+
+    getRenderState(renderState, renderOptions) {
+      return {
+        ...renderState,
+        stats: this.getWidgetRenderState(renderOptions),
+      };
+    },
+
+    getWidgetRenderState({ results, helper }) {
+      if (results) {
```
francoischalifour, line 92: Usually we try to reverse the flow and to remove unnecessary nesting:

```ts
// Let's get rid of this special flow
if (!results) {
  return /* ... */
}

// Let's treat the common case
return /* ... */
```

It's easier to follow the flow of the lifecycle.

shortcuts: Fixed in https://github.com/algolia/instantsearch.js/pull/4565/commits/6d23d58b8408171da02dd73ff76dc2d68460fced :D

src/connectors/stats/__tests__/connectStats-test.js
```diff
@@ -33,6 +39,155 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
...
+        { stats: {} },
+        createRenderOptions({
+          helper,
+          state: helper.state,
+          results,
+        })
+      );
+
+      expect(renderState2.stats).toEqual({
+        hitsPerPage: results.hitsPerPage,
```
francoischalifour, line 84: Here we're just asserting that we wrote the right code but not the right feature. We should use actual values instead of variables because we can't understand if the right values are passed here.

src/connectors/stats/__tests__/connectStats-test.js
```diff
@@ -33,6 +39,155 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
...
+
+      const renderState2 = stats.getWidgetRenderState(
+        createRenderOptions({
+          helper,
+          state: helper.state,
+          results,
+        })
+      );
+
+      expect(renderState2).toEqual({
```
francoischalifour, line 134: Same as above.

src/connectors/stats/__tests__/connectStats-test.js
shortcuts: It makes sense, as there's no queries/results I can just return default values.

Should it also be the case with results?

francoischalifour: Yes—I would create test results in the `createSingleSearchResponse()` call to make sure that the right values are returned from `getRenderState`.

src/connectors/stats/__tests__/connectStats-test.js
shortcuts: Updated in https://github.com/algolia/instantsearch.js/pull/4565/commits/ab81329b6e51822e82ea5e7a3b4a144329d7dbb3

src/connectors/stats/__tests__/connectStats-test.js
shortcuts: Updated in https://github.com/algolia/instantsearch.js/pull/4565/commits/ab81329b6e51822e82ea5e7a3b4a144329d7dbb3

src/connectors/stats/__tests__/connectStats-test.js
```diff
@@ -33,6 +39,155 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
...
+        page: 0,
+        processingTimeMS: 0,
+        query: '',
+        widgetParams: {},
+      });
+    });
+  });
+
+  describe('getWidgetRenderState', () => {
+    test('returns the widget render state', () => {
```
Haroenv, line 96:
```suggestion
    test('returns the widget render state without results', () => {
```

src/connectors/stats/__tests__/connectStats-test.js
```diff
@@ -33,6 +39,155 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
...
+        hitsPerPage: undefined,
+        nbHits: 0,
+        nbPages: 0,
+        page: 0,
+        processingTimeMS: -1,
+        query: '',
+        widgetParams: {},
+      });
+
+      stats.init(createInitOptions({ helper, state: helper.state }));
```
Haroenv, line 120: from here on this should be a new test called "... with empty result"

src/connectors/stats/__tests__/connectStats-test.js
```diff
@@ -33,6 +39,155 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
...
+    });
+  });
+
+  describe('getWidgetRenderState', () => {
+    test('returns the widget render state', () => {
+      const renderFn = jest.fn();
+      const unmountFn = jest.fn();
+      const createStats = connectStats(renderFn, unmountFn);
+      const stats = createStats();
+      const helper = jsHelper(createSearchClient(), 'indexName', {
```
Haroenv, line 101: in the first assertion the helper doesn't need to be created manually, createInitOptions will do it already

src/connectors/stats/__tests__/connectStats-test.js
```diff
@@ -33,6 +39,155 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
...
+        widgetParams: {},
+      });
+
+      stats.init(createInitOptions({ helper, state: helper.state }));
+
+      const results = new SearchResults(helper.state, [
+        createSingleSearchResponse(),
+      ]);
+
+      const renderState2 = stats.getWidgetRenderState(
```
Haroenv, line 126: once you've switched the tests, there's no need for ...2 anymore

Haroenv (commented):
just some small suggestions for the test left

devdreamsolution (approved):

src/connectors/stats/__tests__/connectStats-test.js
shortcuts: https://github.com/algolia/instantsearch.js/pull/4565/commits/176400bd268d80421aeffe79c4351c42c69bfdd6 should cover every use cases

Haroenv (approved):
perfect!

PR #4568, opened 2020-11-03
feat(sortBy): implement `getRenderState` and `getWidgetRenderState`

shortcuts:
This implements the `getRenderState` and `getWidgetRenderState` widget lifecycle hooks in `sortBy`.

### DIFF
--- a/src/connectors/sort-by/__tests__/connectSortBy-test.js
+++ b/src/connectors/sort-by/__tests__/connectSortBy-test.js
@@ -11,6 +11,7 @@ import {
   createInitOptions,
   createRenderOptions,
 } from '../../../../test/mock/createWidget';
+import { createSingleSearchResponse } from '../../../../test/mock/createAPIResponse';
 
 describe('connectSortBy', () => {
   describe('Usage', () => {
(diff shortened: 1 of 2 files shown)


src/connectors/sort-by/__tests__/connectSortBy-test.js
```diff
@@ -257,6 +258,169 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/sort-by/js/
...
+          items: [
+            { label: 'default', value: 'index_default' },
+            { label: 'asc', value: 'index_asc' },
+            { label: 'desc', value: 'index_desc' },
+          ],
+        },
+      });
+
+      sortBy.init(createInitOptions({ helper }));
+      sortBy.getWidgetRenderState({ helper }).refine('index_desc');
```
eunjae-lee, line 307: We used to have this pattern in tests:

```js
      const { refine } = renderFn.mock.calls[renderFn.mock.calls.length - 1][0];
      refine('index_desc');
```

It reads what's rendered from `init` call and get `refine` from there.
However, getting `refine` from `getWidgetRenderState()` also looks okay to me.

Just leaving this comment for Clément's information, and to see if others think differently.

eunjae-lee (approved):
Looks good to me, but would like to get @yannickcr's opinion too

src/connectors/sort-by/__tests__/connectSortBy-test.js
shortcuts: Thanks for the heads up!

You're right, I think I saw it in an other connector but now that we have `getWidgetRenderState` it might be easier to understand (at least for me 🤔 ).

Haroenv: yes, I think using getWidgetRenderState is the way to go for new tests I think. It's okay if older tests still use `refine` from the calls, since the implementation should be the same anyway 👍

francoischalifour: Since render state is now public API it makes sense to use it—and it's such a relief!

src/connectors/sort-by/connectSortBy.js
```diff
@@ -103,7 +103,8 @@ export default function connectSortBy(renderFn, unmountFn = noop) {
     return {
       $$type: 'ais.sortBy',
 
-      init({ helper, instantSearchInstance, parent }) {
+      init(initOptions) {
+        const { helper, instantSearchInstance, parent } = initOptions;
         const currentIndex = helper.state.index;
```
francoischalifour, line 108: We should get the current index from `getWidgetRenderState` otherwise we can get inconsistencies. We compute it twice at different spots here and we don't prevent the logic from differing.

PR #4570, opened 2020-11-03
fix(hits): provide `sendEvent` and `bindEvent` to `init`

eunjae-lee:
**Summary**

This PR fixes two minor issues:

1. In `connectHits`, it didn't return `sendEvent` and `bindEvent` in some condition. Now it does.
2. Casting an empty object as `typeof widgetParams` in `connectPoweredBy`.

Both discovered by `yarn type-check`.

### DIFF
--- a/src/connectors/hits/connectHits.ts
+++ b/src/connectors/hits/connectHits.ts
@@ -115,6 +115,8 @@ const connectHits: HitsConnector = function connectHits(
         if (!results) {
           return {
             hits: [],
+            sendEvent,
+            bindEvent,
             results: undefined,
             widgetParams,
           };
--- a/src/connectors/powered-by/connectPoweredBy.ts
+++ b/src/connectors/powered-by/connectPoweredBy.ts
@@ -47,7 +47,7 @@ const connectPoweredBy: PoweredByConnector = function connectPoweredBy(
     'utm_campaign=poweredby';
 
   return widgetParams => {
-    const { url = defaultUrl } = widgetParams || {};
+    const { url = defaultUrl } = widgetParams || ({} as typeof widgetParams);
 
     return {
       $$type: 'ais.poweredBy',
--- a/src/connectors/hits/__tests__/connectHits-test.ts
+++ b/src/connectors/hits/__tests__/connectHits-test.ts
@@ -461,6 +461,8 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
 
       expect(renderState1.hits).toEqual({
         hits: [],
+        sendEvent: expect.any(Function),
+        bindEvent: expect.any(Function),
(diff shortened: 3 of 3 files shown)


src/connectors/powered-by/connectPoweredBy.ts
```diff
@@ -47,7 +47,7 @@ const connectPoweredBy: PoweredByConnector = function connectPoweredBy(
     'utm_campaign=poweredby';
 
   return widgetParams => {
-    const { url = defaultUrl } = widgetParams || {};
+    const { url = defaultUrl } = widgetParams || ({} as typeof widgetParams);
```
Haroenv, line 50: how come this isn't PoweredByConnectorParams or something? makes sense either way though, since all keys are optional

Haroenv (approved):

src/connectors/powered-by/connectPoweredBy.ts
eunjae-lee: It's actually supposed to be `PoweredByWidgetParams` type, but 

```
export type PoweredByConnector = Connector<
  PoweredByRendererOptions,
  PoweredByConnectorParams
>;
```

We don't explicitly have it at the connector definition.

src/connectors/hits/connectHits.ts
```diff
@@ -115,6 +115,8 @@ const connectHits: HitsConnector = function connectHits(
         if (!results) {
           return {
             hits: [],
+            sendEvent,
```
francoischalifour, line 118: How come this isn't covered by a test? That's public API.

eunjae-lee: I guess it wasn't tested because at `init` phase, we cannot send events.

francoischalifour: It's still important to test it because **users bind events during `init`**. If it's not passed, the API is not usable because the reference to `sendEvent` doesn't exist.

francoischalifour (changes requested):
See https://github.com/algolia/instantsearch.js/pull/4570#discussion_r517253103.

src/connectors/hits/connectHits.ts
eunjae-lee: Yeah right. It makes sense.

eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4570/commits/4bff938aa7144695c1c90ec86e3c3990cad3c8a5

francoischalifour (approved):
Since it's now a fix, we can rename the PR before merging.

Haroenv (approved):

devdreamsolution (approved):

shortcuts (approved):

PR #4571, opened 2020-11-04
fix(sortBy): get `currentIndex` from `getWidgetRenderState`

shortcuts:
This prevent double computation while getting the `currentIndex` in `init`

### DIFF
--- a/src/connectors/sort-by/connectSortBy.js
+++ b/src/connectors/sort-by/connectSortBy.js
@@ -105,16 +105,18 @@ export default function connectSortBy(renderFn, unmountFn = noop) {
 
       init(initOptions) {
         const { helper, instantSearchInstance, parent } = initOptions;
-        const currentIndex = helper.state.index;
-        const isCurrentIndexInItems = find(
-          items,
-          item => item.value === currentIndex
-        );
 
         this.initialIndex = parent.getIndexName();
         this.setIndex = indexName => {
           helper.setIndex(indexName).search();
         };
+
+        const widgetRenderState = this.getWidgetRenderState(initOptions);
+        const currentIndex = widgetRenderState.currentRefinement;
+        const isCurrentIndexInItems = find(
+          items,
+          item => item.value === currentIndex
+        );
 
         warning(
           isCurrentIndexInItems,
@@ -123,7 +125,7 @@ export default function connectSortBy(renderFn, unmountFn = noop) {
 
         renderFn(
           {
-            ...this.getWidgetRenderState(initOptions),
+            ...widgetRenderState,
             instantSearchInstance,
           },
           true


Haroenv:
I know you suggested this @francoischalifour, but this is pretty much only for debugging, and it seems to make the code more complicated; I think I prefer the previous version

devdreamsolution (approved):

francoischalifour:
@Haroenv Why would it be for debugging?

The whole point is to not duplicate the logic that we have about "current index" (remember that this logic changed when we introduced Federated Search). In the next big refactor, we won't have to think about this logic a different places, it's all contained in `getWidgetRenderState`.

Now we don't need to worry how to get the "current index" value because we use the API that we expose to users to retrieve this value, _and this is exactly what Render State is about_.

And if you're concerned about the function being called at the top of the function, you shouldn't. The whole point of `getWidgetRenderState` is that it can be called whenever because it doesn't trigger side effects.

francoischalifour (approved):

Haroenv:
I mean that the function call seemed a bit like overkill, since `currentIndex = helper.state.index`, and the variable is only used for a dev warning

eunjae-lee:
I don't have a strong opinion in either way, but it took me a little to understand why

```
const currentIndex = widgetRenderState.currentRefinement; 
```

shortcuts:
@Haroenv @francoischalifour @eunjae-lee  From a maintenance perspective it make sense to avoid multiple compute point by using the `getWidgetRenderState` function imo. For now we only have two, but if/as we keep adding features, it will be easier.

I could add a little comment line above it to explain? or revert to the old solution for now?

francoischalifour:
It lowers room for inconsistency, which is what we aim for in this codebase.

I'm wondering if you're surprised because you're used to seeing this lifecycle function called only this way?

```ts
renderFn(
  {
    ...this.getWidgetRenderState(initOptions),
    instantSearchInstance,
  },
  true
);
```

It shouldn't be the only way we call it. It's a bit like if you would ban using `props` in a React component somewhere else than in its rendering `return` expression.

You could also see it that way:

`props` → `getProps` (because we don't have the reactive system of React) → `getWidgetRenderState` (rename to our concept)

Haroenv:
I was just confused because the computation isn't a computation but just a getter from the helper, so not complicated at all

shortcuts:
So, should I merge it?
@yannickcr any thoughts on it?
(ref: https://github.com/algolia/instantsearch.js/pull/4568#discussion_r517251970)

francoischalifour:
> I was just confused because the computation isn't a computation but just a getter from the helper, so not complicated at all

@haroenv Yeah but this logic can change anytime (again, it changed during our Federated Search refactor)—it wasn't retrieve from the helper before. The goal is to think of it as an API, you don't have to know that it comes from the helper.

Haroenv (approved):

PR #4574, opened 2020-11-06
feat(connectPagination): add getWidgetRenderState & refactor to TS

Haroenv:
**Summary**

I found it much easier to do both of these at the same time, so unfortunately the diff is removed by git since the files are too different

**Result**

- make connectPagination TS
- make Paginator internal class TS
- fix some issues in the pagination widget test to no longer use internals
- fixes the types in the playground & index test


Haroenv:
the diff is broken here, but you can copy both sides in vs code & compare

eunjae-lee (approved):
Thanks for the tip. It was indeed much easier to compare on VSCode.
I like that we don't use internal and inconsistent methods like `widget.refine()` or `widget.getMaxPage()` for tests any more, and now we get them from `getWidgetRenderState()`.

src/connectors/pagination/connectPagination.ts
```diff
@@ -0,0 +1,187 @@
...
+
+      getWidgetSearchParameters(searchParameters, { uiState }) {
+        const page = uiState.page ? uiState.page - 1 : 0;
+
+        return searchParameters.setQueryParameter('page', page);
+      },
+
+      getWidgetRenderState({ results, helper, createURL }) {
+        if (!connectorState.refine) {
+          connectorState.refine = page => {
```
francoischalifour, line 147: Conceptually `getWidgetRenderState` shouldn't trigger side effects when you call it—I'm wondering if those assignments should be in `init`?

francoischalifour, line 147: It would be nice to have such a syntax:

```ts
connectorState.refine ??= page => { /* ... */ }
```

Haroenv: If I do that, then the tests which ran getRenderState before init (in index) seemed to fail. In some other cases we had moved it to getWidgetRenderState too. I'm not sure either what's the best

Haroenv: https://tc39.es/proposal-logical-assignment/#sec-assignment-operators

eunjae-lee: I just tried and it requires `@babel/plugin-proposal-logical-assignment-operators` to be added. I can create a separate PR with this if you want.

Haroenv: I think we can decide on that later

francoischalifour: > getRenderState before init (in index) seemed to fail

Which makes sense because `init` is supposed to be called before `getRenderState` right now.

If we think long-term, when `init` doesn't exist anymore, it makes sense to manipulate `refine` _only_ in `getWidgetRenderState` (because it'll become contained in that method)—so I think it's fine to have this as it is now. But let's all be aware of this unexpected side effect.

Haroenv: that's what I thought too, if init doesn't exist anymore, getWidgetRenderState should be in control of the initialisation of functions which should only be set once.

francoischalifour (approved):

PR #4575, opened 2020-11-09
fix(insights): do not throw when sending event right after creating insights middleware

eunjae-lee:
## Summary

There is a bug where an error is thrown when sending events right after creating `insights` middleware.

## Details

It happens only when:

- The UMD version of `search-insights` is used.
- The `search-insights` is not loaded yet at the time of creation of `insights` middleware.

In `insights` middleware, the way it intializes `insightsClient` is asynchronous. If it happens too late (normally it happens when the library itself hasn't loaded yet), sending event will throw an error saying "the insightsClient is not intialized yet".

So, in this PR, we change to initialize `insightsClient` synchronously without checking if it was already initialized before.

## A behavioral change

Before this PR, userToken that was set before creating insights middleware was ignored. However since this PR, that has to  change, because although user has a code like the following:

```js
aa('init', { ... });
aa('setUserToken', 'token-a');

search.use(createInsightsMiddleware({ ... }));
```

`insights` middleware calls `aa('init', ...)` anyway. It means `aa(init', ...)` will override the token with an anonymous token. Of course, we can update `search-insights` not to override already given token with an anonymous one, but there is no guarantee that user upgrades search-insights correctly. So we cannot rely on it.

Since `insights` middleware initializes `insightsClient`, we guide not to call `aa('init')` by themselves, but we cannot stop it. And if they set userToken like above, we shouldn't lose the token (although before this PR, we used to ignore the token before creating the middleware).

This behavioral change can be considered as a breaking change, but we can also consider it as a "fix" because it is correct to prefer a userToken that was explicitly set by user over an anonymous token.

## Alternatives

There are a couple of other options requiring changes in `search-insights` side, but it's risky to have a change in InstantSearch that depends on a change in `search-insights` because we cannot guarantee users will upgrade them both.

## Future task

This doesn't concern another issue we've recently discussed about, which is to provide an option to disable cookie at all. It will be dealt separately.

## Result

1. When UMD search-insights is not loaded yet, it doesn't throw when sending events. It stores it in the queue and send it later correctly.
2. If user sets a token before creating `insights` middleware, it takes it into account. (It used to not.)

### DIFF
--- a/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -79,6 +79,22 @@ describe('insights', () => {
       });
     });
 
+    it('does not throw when an event is sent right after the creation', () => {
+      const { insightsClient, instantSearchInstance } = createTestEnvironment();
+      createInsightsMiddleware({
+        insightsClient,
+      })({ instantSearchInstance });
+
+      expect(() => {
+        insightsClient('viewedObjectIDs', {
+          userToken: 'my-user-token',
+          index: instantSearchInstance.indexName,
+          eventName: 'Products Viewed',
+          objectIDs: ['obj-id0', 'obj-id1'],
+        });
+      }).not.toThrow();
+    });
+
     it('warns dev if userToken is set before creating the middleware', () => {
       const { insightsClient, instantSearchInstance } = createTestEnvironment();
       insightsClient('setUserToken', 'abc');
--- a/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -6,6 +6,7 @@ import {
   createAlgoliaAnalytics,
   createInsightsClient,
   createInsightsUmdVersion,
+  AlgoliaAnalytics,
   ANONYMOUS_TOKEN,
 } from '../../../test/mock/createInsightsClient';
 import { warning } from '../../lib/utils';
@@ -36,11 +37,11 @@ describe('insights', () => {
     };
   };
 
-  const createUmdTestEnvironment = () => {
+  const createUmdTestEnvironment = (algoliaAnalytics?: AlgoliaAnalytics) => {
     const {
       insightsClient,
       libraryLoadedAndProcessQueue,
-    } = createInsightsUmdVersion();
+    } = createInsightsUmdVersion(algoliaAnalytics);
     const instantSearchInstance = createInstantSearch({
       client: algoliasearch('myAppId', 'myApiKey'),
     });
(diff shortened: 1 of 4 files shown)


src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -36,6 +36,30 @@ describe('insights', () => {
     };
   };
 
+  const createUmdTestEnvironment = () => {
```
eunjae-lee, line 39: This has been moved here from other `describe` scope with no change.

src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -236,6 +251,7 @@ aa('setUserToken', 'your-user-token');`);
           insightsClient,
         })({ instantSearchInstance });
         middleware.subscribe();
+        libraryLoadedAndProcessQueue();
```
eunjae-lee, line 254: This was missing. We need this because the library will eventually be loaded.

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -76,28 +85,19 @@ aa('setUserToken', 'your-user-token');
...
           // We can set it as userToken.
-          setUserTokenToSearch(getInsightsAnonymousUserTokenInternal());
+          setUserTokenToSearch(anonymousUserToken);
+        }
+
+        if (queuedUserToken) {
+          insightsClient('setUserToken', queuedUserToken);
         }
 
-        if (Array.isArray((insightsClient as any).queue)) {
```
eunjae-lee, line 85: This part has been move to the line 44. And there it doesn't call `setUserTokenToSearch()`. It instead sets a variable there, and later on, at the line 95, it reads the variable and sets it.

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -38,26 +38,35 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = props => {
...
+      //
+      // When user calls `aa('setUserToken', 'my-user-token')` before `search-insights` is loaded,
+      // ['setUserToken', 'my-user-token'] gets stored in `aa.queue`.
+      // Whenever `search-insights` is finally loaded, it will process the queue.
+      //
+      // But here's the reason why we handle it here:
+      // At this point, even though `search-insights` is not loaded yet,
+      // we still want to read the token from the queue.
+      // Otherwise, the first search call will be fired without the token.
+      (insightsClient as any).queue.forEach(([method, firstArgument]) => {
```
francoischalifour, line 55: Why using `forEach` if we override the variable every time? Using `find` would make more sense:

```ts
find(insightsClient.queue.slice().reverse(), /* ... */)
```

- `slice` to not mutate the queue
- `reverse` to find the latest occurrence

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -38,26 +38,35 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = props => {
     _insightsClient === null ? (noop as InsightsClient) : _insightsClient;
```
francoischalifour, line 38: If we type it correctly here, we shouldn't have to cast `insightsClient` every time.

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -76,28 +85,19 @@ aa('setUserToken', 'your-user-token');
...
-        if (hasInsightsClient) {
+        const anonymousUserToken = getInsightsAnonymousUserTokenInternal();
+        if (hasInsightsClient && anonymousUserToken) {
           // When `aa('init', { ... })` is called, it creates an anonymous user token in cookie.
           // We can set it as userToken.
-          setUserTokenToSearch(getInsightsAnonymousUserTokenInternal());
+          setUserTokenToSearch(anonymousUserToken);
+        }
+
+        if (queuedUserToken) {
```
francoischalifour, line 95: Why can this condition and the one below be satisfied at the same time? It makes the flow harder to understand.

If that what it means, I'd suggest changing to this:

```ts
// We consider the `userToken` coming from a `init` call to have a higher
// importance than the one coming from the queue.
if (userTokenBeforeInit) {
  insightsClient('setUserToken', userTokenBeforeInit);
} else if (queuedUserToken) {
  insightsClient('setUserToken', queuedUserToken);
}
```

test/mock/createInsightsClient.ts
```diff
@@ -26,13 +26,23 @@ export function createAlgoliaAnalytics() {
       callback(values._userToken);
     }
   };
+  const sendEvent = () => {
```
francoischalifour, line 29: Why do we implement another whole logic in its mock?

src/middlewares/createInsightsMiddleware.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4575/commits/fa7ea26d1c4c4d2a49960cd89a731d50a6fea9f2

src/middlewares/createInsightsMiddleware.ts
eunjae-lee: Do you mean regarding `(insightsClient as any).queue` right?
https://github.com/algolia/instantsearch.js/pull/4575/commits/5ddedeec8806d997372d27de340f1194cce2e174

src/middlewares/createInsightsMiddleware.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4575/commits/8b10f8940ac1b6e94ab8cb3757688207f57353fd

test/mock/createInsightsClient.ts
eunjae-lee: Good point. I thought about it again, and I think the new test is better.
https://github.com/algolia/instantsearch.js/pull/4575/commits/0add3dcab5e1f18e57e86fdcc6f00ed6bdc7e909

francoischalifour: We probably shouldn't even mock any logic, but just spy for methods called. That's coming from before so it shouldn't block that PR though.

src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -79,22 +104,46 @@ describe('insights', () => {
...
+      const middleware = createInsightsMiddleware({
+        insightsClient,
+      })({ instantSearchInstance });
+      middleware.subscribe();
 
-insightsClient('setUserToken', 'your-user-token');
-// or
-aa('setUserToken', 'your-user-token');`);
+      // It tries to send an event.
+      instantSearchInstance.sendEventToInsights({
```
francoischalifour, line 121: We keep testing internals here, that's fragile. Again, that's how it was before these changes.

francoischalifour (approved):

test/mock/createInsightsClient.ts
eunjae-lee: At some point, I began to wonder if I just should've used the real search-insights client for the tests, because there are some implementations on the search-insights side that needs to be tested with the insights middleware.

src/middlewares/__tests__/createInsightsMiddleware.ts
eunjae-lee: Although the method is public, in this context, we're indeed testing the internals. I'll keep it that way for now.

PR #4577, opened 2020-11-10
chore(ts): update typing of widgetParams

Haroenv:
**Summary**

fix typing of widgetParams in the context of connectors & getWidgetRenderState 

**Result**

wdigetParams in the context of getWidgetRenderState is different in the two contexts:
- in connectors it's connectorparams & twidgetparams (they pass all arguments from widgetParams through to widgetParams in getWidgetRenderState)
- in widgets it's actually only connectorparams (they don't pass all arguments through to the connector, and set defaults)

### DIFF
--- a/src/types/connector.ts
+++ b/src/types/connector.ts
@@ -64,13 +64,19 @@ export type Connector<TRendererOptions, TConnectorParams> = <TWidgetParams>(
   /**
    * The render function.
    */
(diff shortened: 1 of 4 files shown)


eunjae-lee (approved):
It looks correct to me. I've tested it and it seems to be working fine.

francoischalifour (approved):

PR #4579, opened 2020-11-17
fix(renderState): initialize connectorState before init

eunjae-lee:
**Summary**

This PR intializes `connectorState` before `init`. Since `refine` and `createURL` was initialized in `init`, it threw an exception when `getRenderState` was called inside `index` widget.

It's fixed with a regression test.

### DIFF
--- a/src/connectors/numeric-menu/__tests__/connectNumericMenu-test.ts
+++ b/src/connectors/numeric-menu/__tests__/connectNumericMenu-test.ts
@@ -1131,6 +1131,68 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/numeric-men
   });
 
   describe('getWidgetRenderState', () => {
+    it('returns the widget render state before init', () => {
+      const rendering = jest.fn();
+      const makeWidget = connectNumericMenu(rendering);
+      const widget = makeWidget({
+        attribute: 'numerics',
+        items: [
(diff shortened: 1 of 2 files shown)


src/connectors/numeric-menu/__tests__/connectNumericMenu-test.ts
```diff
@@ -1131,6 +1131,68 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/numeric-men
...
+        attribute: 'numerics',
+        items: [
+          { label: 'below 10', end: 10 },
+          { label: '10 - 20', start: 10, end: 20 },
+          { label: 'more than 20', start: 20 },
+        ],
+      });
+
+      const helper = jsHelper(createSearchClient(), '');
+      helper.search = () => helper;
```
shortcuts, line 1147: I think it's not necessary to initialize `helper.search()` since calling `createSearchClient()` will do it, I might be wrong tho

eunjae-lee: you're right! https://github.com/algolia/instantsearch.js/pull/4579/commits/47e83d61c41975a1d57a17bfc978cda92b9cf9d5

yannickcr (approved):

PR #4580, opened 2020-11-17
fix(insights): use internal `find` util method

francoischalifour:
## Description

In #4575 we introduced a regression in IE because we used the native `find` method instead of our internal util.

## Related

- https://algolia.slack.com/archives/C2XP6HBKQ/p1605624265187700
- https://secure.helpscout.net/conversation/1340807801/309760?folderId=3647171

### DIFF
--- a/src/middlewares/createInsightsMiddleware.ts
+++ b/src/middlewares/createInsightsMiddleware.ts
@@ -1,6 +1,6 @@
 import { InsightsClient, InsightsClientMethod, Middleware } from '../types';
 import { getInsightsAnonymousUserTokenInternal } from '../helpers';
-import { warning, noop, getAppIdAndApiKey } from '../lib/utils';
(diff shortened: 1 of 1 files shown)


yannickcr (approved):

shortcuts (approved):

Haroenv:
I'll think if we can add a browser compat test, since the e2e test clearly doesn't catch this

eunjae-lee:
I'm wondering how the e2e tests were successful in IE env in the first place.

francoischalifour:
Wondering too...

Haroenv:
the insights middleware must not have been used on the e2e test @eunjae-lee

eunjae-lee:
> the insights middleware must not have been used on the e2e test @eunjae-lee

that must be it

PR #4581, opened 2020-11-17
feat(answers): add `EXPERIMENTAL_answers` widget

eunjae-lee:
## Summary

This PR adds `answers` widget.

* Example: https://codesandbox.io/s/instantsearchjs-forked-dj2bm?file=/src/app.js:375-1115

### TODO
- [x] Fix type errors
- [x] Test highlighting
- [x] Write tests
- [x] Put comments on the properties
- [x] Do we need a connector or not?
- [x] Fix type errors
- [x] Debounce
- [x] Concurrent-safe
- [x] The loader is flickering too much (probably can be solved by debouncing)
- [x] Remove `stories/algoliasearch.umd.js`

## Next
In the subsequent PR
- [ ] Applying `findAnswers` from helper 
- [ ] Applying facets
- [ ] Use index name from parent index widget
- [ ] loading state (`'idle' | 'loading' | 'stalled'`)


eunjae-lee:
1. 
I see the type errors regarding highlight, which I think is coming from algoliasearch package.

2.
I need to apply facets to `/answers` call but don't know how to get search parameter. If I understand correctly, first I need to find which index this answers widget belongs to, and call `theIndex.getWidgetSearchParameter(...)`. Is it correct?

Haroenv:
> I see the type errors regarding highlight, which I think is coming from algoliasearch package.

I fixed these in https://github.com/algolia/instantsearch.js/pull/4587, either we merge that quickly or split it up

eunjae-lee:
> I fixed these in #4587, either we merge that quickly or split it up

Oh nice. I can wait for its merge :)

eunjae-lee:
This PR is ready for review. I've decided to support "facets" in a separate PR.
@Haroenv @shortcuts

package.json
```diff
@@ -143,7 +143,7 @@
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "65.25 kB"
+      "maxSize": "65.80 kB"
```
shortcuts, line 146: I just upgraded it to `65.5 kB` in my PR (https://github.com/algolia/instantsearch.js/pull/4592), should I remove it?

eunjae-lee: @shortcuts you can keep it there because we don't know which PR gets merged first :)

shortcuts: Right but if I override your limit it might fail your tests, unless your bundle size is < 65.5 kB

eunjae-lee: If it fails, I will increase it again :)

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,179 @@
...
+   */
+  nbHits?: number;
+};
+
+export type AnswersConnector = Connector<
+  AnswersRendererOptions,
+  AnswersConnectorParams
+>;
+
+const connectAnswers: AnswersConnector = function connectAnswers(
```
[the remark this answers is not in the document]
eunjae-lee: I guess this is a better place: https://github.com/algolia/instantsearch.js/pull/4581/commits/feb1656d16b6172838da37ef5248b43ad74cdd9c

src/types/algoliasearch.ts
```diff
@@ -19,11 +19,12 @@ import {
   SearchResponse as SearchResponseV4,
   // no comma, TS is particular about which nodes expose comments
   // eslint-disable-next-line prettier/prettier
-  SearchForFacetValuesResponse as SearchForFacetValuesResponseV4
+  SearchForFacetValuesResponse as SearchForFacetValuesResponseV4,
```
Haroenv, line 22:
```suggestion
  SearchForFacetValuesResponse as SearchForFacetValuesResponseV4
```

src/types/algoliasearch.ts
```diff
@@ -19,11 +19,12 @@ import {
...
   /** @ts-ignore */
   // @ts-ignore
   // eslint-disable-next-line import/no-unresolved
 } from '@algolia/client-search';
+export { FindAnswersResponse } from '@algolia/client-search';
```
Haroenv, line 27: this one can be consolidated with the other imports

src/types/algoliasearch.ts
```diff
@@ -39,6 +40,8 @@ export type SearchClient = {
   search: DefaultSearchClient['search'];
   searchForFacetValues: DefaultSearchClient['searchForFacetValues'];
   addAlgoliaAgent?: DefaultSearchClient['addAlgoliaAgent'];
+  initIndex?: DefaultSearchClient['initIndex'];
+  transporter?: SearchClientV4['transporter'];
```
Haroenv, line 44: what do we use transporter for?

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,179 @@
...
+      $$type: 'ais.answers',
+
+      init(initOptions) {
+        const { instantSearchInstance } = initOptions;
+        // eslint-disable-next-line no-warning-comments
+        // FIXME: remove this customization once the engine accepts url encoded query params
+        if (instantSearchInstance.client.transporter) {
+          instantSearchInstance.client.transporter.userAgent.value =
+            'answers-test';
+        }
```
Haroenv, line 87: I'm not sure, but I wonder if it should be user land instead?

src/types/algoliasearch.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581/commits/3c3e592a75e7ed7e8bbd7ce1ac1e17535addf013

src/types/algoliasearch.ts
eunjae-lee: It's for

```
         if (instantSearchInstance.client.transporter) {
          instantSearchInstance.client.transporter.userAgent.value =
            'answers-test';
        } 
```

src/connectors/answers/connectAnswers.ts
eunjae-lee: I'm not too sure, either, but if we guide people to do it user-land, that code might stick forever, whereas this can be removed in the later version when the engine fixes the bug.

src/widgets/answers/answers.tsx
```diff
@@ -0,0 +1,168 @@
...
+  } = widgetParams || ({} as typeof widgetParams);
+
+  if (!container) {
+    throw new Error(withUsage('The `container` option is required.'));
+  }
+
+  const containerNode = getContainerNode(container);
+  const cssClasses = {
+    root: cx(suit(), userCssClasses.root),
+    emptyRoot: cx(suit({ modifierName: 'empty' }), userCssClasses.emptyRoot),
```
Haroenv, line 143: why is it called emptyRoot when the generated class name is empty? Could you in parallel also open a PR on specs so we can discuss final DOM before it leaves experimental?

src/types/algoliasearch.ts
Haroenv: I think we maybe should instead @ts-ignore there, since the change there is only while it's experimental

src/types/algoliasearch.ts
```diff
@@ -25,6 +26,8 @@ import {
   // eslint-disable-next-line import/no-unresolved
 } from '@algolia/client-search';
 
+export { FindAnswersResponse };
```
Haroenv, line 29: is this `any` on v3 or an error?

src/connectors/answers/__tests__/connectAnswers-test.ts
```diff
@@ -0,0 +1,436 @@
...
+  };
+
+  it('is a widget', () => {
+    const render = jest.fn();
+    const unmount = jest.fn();
+
+    const makeWidget = connectAnswers(render, unmount);
+    const widget = makeWidget({
+      queryLanguages: ['en'],
+      attributesForPrediction: ['description'],
```
Haroenv, line 76: do we have an error message if any of these parameters are missed, or do we rely on the engine erroring?

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,179 @@
...
+   * Attributes to use for predictions.
+   * If empty, we use all `searchableAttributes` to find answers.
+   * All your `attributesForPrediction` must be part of your `searchableAttributes`.
+   */
+  attributesForPrediction?: string[];
+
+  /**
+   * The languages in the query. Currently only supports `en`.
+   */
+  queryLanguages: string[];
```
Haroenv, line 39:
```suggestion
  queryLanguages: Array<'en'>;
```

maybe this is good for the InstantSearch widget?

src/connectors/answers/connectAnswers.ts
Haroenv: I wonder if the user agent can be overridden from the requestOptions, have you tried that? You will set the userAgent for the InstantSearch session like this as well

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,179 @@
...
+
+        // call /answers API
+        runConcurrentSafePromise(
+          answersIndex.findAnswers(query, queryLanguages, {
+            nbHits,
+            attributesForPrediction,
+          })
+        ).then(result => {
+          lastAnswersResult = result;
+          lastAnswersResult.hits = addAbsolutePosition<typeof result.hits[0]>(
```
Haroenv, line 136: what about queryID and __escaped?

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,179 @@
...
+      );
+    }
+
+    const runConcurrentSafePromise = createConcurrentSafePromise<
+      FindAnswersResponse<Hit>
+    >();
+
+    let lastAnswersResult: Partial<FindAnswersResponse<Hit>>;
+    let isLoading = false;
+    const debouncedRenderFn = debounce(renderFn, 200);
```
Haroenv, line 75: I think the debounce timeout should be a parameter; I can imagine tweaking it around to find the best value

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,179 @@
...
+          {
+            ...this.getWidgetRenderState(renderOptions),
+            instantSearchInstance: renderOptions.instantSearchInstance,
+          },
+          false
+        );
+
+        // call /answers API
+        runConcurrentSafePromise(
+          answersIndex.findAnswers(query, queryLanguages, {
```
Haroenv, line 130: why is it the render that's debounced, and not the findAnswers itself?

src/widgets/answers/answers.tsx
eunjae-lee: What do you mean by
> when the generated class name is empty

image

image

src/connectors/answers/connectAnswers.ts
eunjae-lee: Makes sense https://github.com/algolia/instantsearch.js/pull/4581/commits/725347141e15818dc4fa19ddc38d9252593d03ce

src/connectors/answers/connectAnswers.ts
eunjae-lee: Yeah it works! 
https://github.com/algolia/instantsearch.js/pull/4581/commits/94e607360a002f8a7c18a7259e79522c3acc95f5

src/types/algoliasearch.ts
eunjae-lee: I created an empty repo with v3 to see how it goes, but VSCode didn't tell me anything.
To make it explicit: https://github.com/algolia/instantsearch.js/pull/4581/commits/f0ccf1c775c01d0886e9f4e199b80ebb2bdf68d6

src/connectors/answers/__tests__/connectAnswers-test.ts
eunjae-lee: Yeah we throws an error, but the test was missing: https://github.com/algolia/instantsearch.js/pull/4581/commits/92e9a996f7f1ba4fda1b73070f2963952edf0bbf

src/connectors/answers/connectAnswers.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581/commits/87d96f7f8f08a4cf72da02b193bf10ca1c8f2ee3

src/connectors/answers/connectAnswers.ts
eunjae-lee: When I received multiple responses in a short time window, the multiple renderings made it look flashy.

Haroenv: I mean that not only you should debounce the rendering, but also the sending of the requests (since the request time can fluctuate)

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,217 @@
+import escapeHits from '../../lib/escape-highlight';
+import {
+  checkRendering,
+  createDocumentationMessageGenerator,
+  createConcurrentSafePromise,
+  addQueryID,
+  debounce,
+  addAbsolutePosition,
+  noop,
+} from '../../lib/utils';
```
shortcuts, line 10:
```suggestion
import {
  checkRendering,
  createDocumentationMessageGenerator,
  createConcurrentSafePromise,
  addQueryID,
  debounce,
  addAbsolutePosition,
  noop,
  escapeHits,
} from '../../lib/utils';
```

After merging master, this will be needed

src/connectors/answers/__tests__/connectAnswers-test.ts
```diff
@@ -0,0 +1,447 @@
...
+  debounce: jest.fn(callback => (...args) => setTimeout(callback(...args), 0)),
+}));
+import { debounce } from '../../../lib/utils/debounce';
+import connectAnswers from '../connectAnswers';
+
+const wait = (time = 10) => {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve();
+    }, time);
```
shortcuts, line 19:
```suggestion
    setTimeout(resolve, time);
```

This should avoid the `missing args` warning on `resolve()`

src/connectors/answers/__tests__/connectAnswers-test.ts
```diff
@@ -0,0 +1,447 @@
...
+          })
+        )
+      ).toEqual({
+        hits: [{ title: '', __position: 1 }],
+        isLoading: false,
+        widgetParams: {
+          queryLanguages: ['en'],
+          attributesForPrediction: ['description'],
+        },
+      });
```
shortcuts, line 444: Not sure why but this test throw `Received: serializes to the same string` even though it should work (`.toContainEqual` show that they are equals). This is still an opened issue but this workaround fixes it https://github.com/facebook/jest/issues/8475#issuecomment-537830532

Haroenv: I think it might be happening because the hits array has the same items, but not the same properties (like __escaped), is that possible?

Haroenv: it could also be due to the circular dependency fixed in #4618

shortcuts: I ran the tests after merging master and it still occurs

shortcuts: > I think it might be happening because the hits array has the same items, but not the same properties (like __escaped), is that possible?

`__escaped: true` is returned when logging `widget.getWidgetRenderState`

eunjae-lee: It was because of the changes in the connector. I fixed it :) 
https://github.com/algolia/instantsearch.js/pull/4581/commits/51507ec25400f155fc63e16956ffb7e488f5439f

src/widgets/answers/answers.tsx
Haroenv: maybe rootEmpty? what do we call this in the other widgets?

eunjae-lee: We've got `emptyRoot`s in hits and infinite-hits.

src/connectors/answers/connectAnswers.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581/commits/f446f18d5669d5c99312a2a348a0046acd457612
here you go

src/widgets/__tests__/index.tests.ts
```diff
@@ -2,6 +2,14 @@ import { PlacesInstance } from 'places.js';
 import * as widgets from '../';
 import { Widget } from '../../types';
 
+/**
```
eunjae-lee, line 5: @Haroenv WDYT?

src/components/Answers/Answers.tsx
```diff
@@ -0,0 +1,75 @@
...
+            templateKey="item"
+            rootTagName="li"
+            rootProps={{ className: cssClasses.item }}
+            key={hit.objectID}
+            data={{
+              ...hit,
+              __hitIndex: position,
+            }}
+          />
+        ))}
```
shortcuts, line 69:
```suggestion
    {isLoading ? (
      <Template
        {...templateProps}
        templateKey="loader"
        rootProps={{ className: cssClasses.loader }}
      />
    ) : (
      <ul className={cssClasses.list}>
        {hits.map((hit, position) => (
          <Template
            {...templateProps}
            templateKey="item"
            rootTagName="li"
            rootProps={{ className: cssClasses.item }}
            key={hit.objectID}
            data={{
              ...hit,
              __hitIndex: position,
            }}
          />
        ))}
      </ul>
    )}
```

src/connectors/answers/__tests__/connectAnswers-test.ts
```diff
@@ -0,0 +1,471 @@
...
+
+const defaultRenderDebounceTime = 10;
+const defaultSearchDebounceTime = 10;
+
+// 10(render debounce) + 10(search debounce) + 10(just to make sure)
+const wait = (time = 30) => {
+  return new Promise(resolve => {
+    setTimeout(resolve, time);
+  });
+};
```
shortcuts, line 19: We could move this function to `test/utils` as it's now used/declared in multiple files (`createMetadataMiddleware.ts`, `history.test.ts`, and this one)

src/lib/utils/debounce.ts
```diff
@@ -0,0 +1,39 @@
...
+export function debounce(func: Function, wait: number): Function {
+  let lastTimeout;
+  return function(...args) {
+    // @ts-ignore-next-line
+    const that = this;
+    return new Promise<void>(resolve => {
+      clearTimeout(lastTimeout);
+      lastTimeout = setTimeout(() => {
+        lastTimeout = null;
+        func.apply(that, args);
```
shortcuts, line 11: As you are using arrow functions, I think you can get rid of the `const that = this` logic if you spread the args. wdyt @Haroenv @eunjae-lee 
```suggestion
        func(...args);
```

src/lib/utils/debounce.ts
```diff
@@ -0,0 +1,39 @@
+// Debounce the function call to the trailing edge.
```
Haroenv, line 1: I would still need to test this in detail, but I think this is the typed variant of the current code:


```ts
type Func = (...args: any[]) => any;

type DebouncedFunction<TFunction extends Func> = (
  this: ThisParameterType<TFunction>,
  ...args: Parameters<TFunction>
) => Promise<ReturnType<TFunction>>;

// Debounce the function call to the trailing edge.
export function debounce<TFunction extends Func>(
  func: TFunction,
  wait: number
): DebouncedFunction<TFunction> {
  let lastTimeout: ReturnType<typeof setTimeout> | null = null;
  return function(...args) {
    const that = this;
    return new Promise(resolve => {
      clearTimeout(lastTimeout!);
      lastTimeout = setTimeout(() => {
        lastTimeout = null;
        resolve(func.apply(that, args));
      }, wait);
    });
  };
}

// Debounced function will return a promise which will be either resolved or rejected.
export function debounceAsync<TFunction extends Func>(
  func: TFunction,
  wait: number
): DebouncedFunction<TFunction> {
  let lastTimeout: ReturnType<typeof setTimeout> | null = null;
  let lastReject: (reason?: any) => void | undefined;
  return function(...args) {
    const that = this;
    return new Promise((resolve, reject) => {
      if (typeof lastReject === 'function') {
        lastReject();
      }
      lastReject = reject;
      clearTimeout(lastTimeout!);
      lastTimeout = setTimeout(() => {
        lastTimeout = null;
        Promise.resolve(func.apply(that, args))
          .then(resolve)
          .catch(reject);
      }, wait);
    });
  };
}
```

src/connectors/answers/__tests__/connectAnswers-test.ts
eunjae-lee: Good point. Thanks for finding out those :)  https://github.com/algolia/instantsearch.js/pull/4581/commits/ea8a2f1f027ded9611bfd4ce7890787da7b86280

src/lib/utils/debounce.ts
eunjae-lee: It works in some cases that way.
https://codesandbox.io/s/serene-shape-k7upd?file=/src/index.js
If you see the examples in the link above, the example 2 won't work without `that`.
Of course, it's not that important in our case, though.

src/lib/utils/debounce.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581/commits/3df3e8102bf72e310cd1bfbe4aa3823c30d24dbe

Thanks! It seems to work fine.
I made a little change to the implementation of `debounce` because it doesn't have to return a Promise. It was mistakenly left there where I started to write debounceAsync.

src/lib/utils/debounce.ts
shortcuts: Oh ok! I didn't know, but isn't this issue related to how we create the object and not how `debounce` works? I know it's not the same but if we use `b.data` in the example 2, it will work

.eslintrc.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
           'EXPERIMENTAL_use',
```
Haroenv, line 11:
```
"capIsNewExceptionPattern": "^EXPERIMENTAL_"
```

https://eslint.org/docs/rules/new-cap#capisnewexceptionpattern

this will be nicer, since now we have older items that are still listed here, but no longer experimental

src/widgets/answers/answers.tsx
```diff
@@ -0,0 +1,175 @@
...
+  AnswersWidgetParams
+>;
+
+const answersWidget: AnswersWidget = widgetParams => {
+  const {
+    container,
+    attributesForPrediction,
+    queryLanguages,
+    nbHits,
+    searchDebounceTime,
```
Haroenv, line 132: ...extraParameters isn't exposed here, is that on purpose?

src/widgets/__tests__/index.tests.ts
Haroenv: useful, although, I'm not sure this is the right place, what about adding "updating this file" and putting it in contributing.md?

src/types/widget.ts
```diff
@@ -431,8 +437,8 @@ export type Widget<
     | 'ais.sortBy'
     | 'ais.stats'
     | 'ais.toggleRefinement'
-    | 'ais.voiceSearch';
-
+    | 'ais.voiceSearch'
+    | 'ais.answers';
```
Haroenv, line 441: the list is alphabetical

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,241 @@
...
+      getRenderState(renderState, renderOptions) {
+        return {
+          ...renderState,
+          answers: this.getWidgetRenderState(renderOptions),
+        };
+      },
+
+      getWidgetRenderState() {
+        return {
+          hits: (lastResult && lastResult.hits) || [],
```
Haroenv, line 223: where's refine?

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,241 @@
...
+    let debouncedRefine;
+
+    return {
+      $$type: 'ais.answers',
+
+      init(initOptions) {
+        const { state, instantSearchInstance } = initOptions;
+        const answersIndex = instantSearchInstance.client!.initIndex!(
+          state.index
+        );
```
Haroenv, line 113: this can be done in widget creation, before init

.storybook/static/answers.css
```diff
@@ -0,0 +1,136 @@
+.my-Answers .ais-Answers-loader {
+  display: none;
+}
+
+.my-Answers.ais-Answers--empty .ais-Answers-loader {
```
Haroenv, line 5: is this still needed with the pr to specs merged?

src/connectors/answers/connectAnswers.ts
eunjae-lee: @Haroenv do you mean like in line 105? We need an access to `instantSearchInstance` and I don't see any way before init.

src/connectors/answers/connectAnswers.ts
eunjae-lee: I didn't put it because I considered Answers widget like Hits widget. So the answers is refined by SearchBox, not by itself.

src/types/algoliasearch.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581/commits/6040086eb0c56a92540a16e1ba1384cce4068c66

.eslintrc.js
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581/commits/bf6d953fe87ce3a3fff66a6d20756d0fa5150788

src/widgets/answers/answers.tsx
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581/commits/6a972eb19f3a252e41b40b4ec26f2407a4a83534

src/types/widget.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581/commits/a9d46b18431465c35effd68a1ad1c965ed7c89a1

src/lib/utils/debounce.ts
Haroenv: since our functions don't use `this`, I think it might not even be needed though

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,254 @@
...
+            isLoading = false;
+            debouncedRender(
+              {
+                ...this.getWidgetRenderState(renderOptions),
+                instantSearchInstance: renderOptions.instantSearchInstance,
+              },
+              false
+            );
+          })
+          .catch(_error => {});
```
Haroenv, line 224: are we sure all errors should be caught? not sure how to expose it to the user though

src/types/widget.ts
```diff
@@ -432,7 +439,6 @@ export type Widget<
     | 'ais.stats'
     | 'ais.toggleRefinement'
     | 'ais.voiceSearch';
-
```
Haroenv, line 435:
```suggestion

```

src/connectors/answers/connectAnswers.ts
eunjae-lee: Good catch. It was initially there because of debounced function rejecting the promise.

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    return {
+      $$type: 'ais.answers',
+
+      init(initOptions) {
+        const { state, instantSearchInstance } = initOptions;
+        const answersIndex = instantSearchInstance.client!.initIndex!(
+          state.index
+        );
+        if (!answersIndex.findAnswers) {
```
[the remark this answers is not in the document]
eunjae-lee: This is really tough. On v3, there's no v4, so we cannot add a type guard like `hasFindAnswers(index): index is IndexV4`. In the point view of v3, it's checking something not existing at all. Any idea?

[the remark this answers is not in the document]
eunjae-lee: @francoischalifour @yannickcr any idea? or at least a way to disable type checking on this file?

src/lib/utils/debounce.ts
eunjae-lee: okay, since it's not used yet, let's remove it. https://github.com/algolia/instantsearch.js/pull/4581/commits/81cba95bccc361888319ce1fa38b079e953dbf1a

src/connectors/answers/connectAnswers.ts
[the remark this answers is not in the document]
Haroenv: I feel like there could be some way to check (in parallel to the type checking) the version of the client (if you import it from the node module) and return nothing / an error in the case of v3

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,251 @@
...
+});
+
+export type AnswersRendererOptions = {
+  /**
+   * The matched hits from Algolia API.
+   */
+  hits: Hits;
+
+  /**
+   * whether it's still loading the results from the /answers API
```
francoischalifour, line 31:
```suggestion
   * Whether it's still loading the results from the Answers API.
```

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,251 @@
...
+   * Attributes to use for predictions.
+   * If empty, we use all `searchableAttributes` to find answers.
+   * All your `attributesForPrediction` must be part of your `searchableAttributes`.
+   */
+  attributesForPrediction?: string[];
+
+  /**
+   * The languages in the query. Currently only supports `en`.
+   */
+  queryLanguages: Array<'en'>;
```
francoischalifour, line 47:
```suggestion
  queryLanguages: ['en'];
```

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  /**
+   * Maximum number of answers to retrieve from the Answers Engine.
+   * Cannot be greater than 1000.
+   */
```
francoischalifour, line 52:
```suggestion
   * @default 1
   */
```

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,251 @@
...
+      queryLanguages,
+      attributesForPrediction,
+      nbHits = 1,
+      renderDebounceTime = 100,
+      searchDebounceTime = 100,
+      escapeHTML = true,
+      extraParameters = {},
+    } = widgetParams || ({} as typeof widgetParams);
+
+    if (!queryLanguages || queryLanguages.length === 0) {
```
francoischalifour, line 103: Do we want to default `queryLanguages` to `['en']`? (Especially because this is the only supported value.)

src/connectors/answers/connectAnswers.ts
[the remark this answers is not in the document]
francoischalifour: What happens if we cast `answersIndex` to `IndexV4` here? (via `unknown` first might be required, then to `IndexV4`)

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,251 @@
...
+      getRenderState(renderState, renderOptions) {
+        return {
+          ...renderState,
+          answers: this.getWidgetRenderState(renderOptions),
+        };
+      },
+
+      getWidgetRenderState() {
+        return {
+          hits: (lastResult && lastResult.hits) || [],
```
francoischalifour, line 233:
```suggestion
          hits: lastResult?.hits || [],
```

src/widgets/answers/answers.tsx
```diff
@@ -0,0 +1,179 @@
...
+      cssClasses={cssClasses}
+      hits={hits}
+      isLoading={isLoading}
+      templateProps={renderState.templateProps}
+    />,
+    containerNode
+  );
+};
+
+export type AnswersTemplates = {
```
francoischalifour, line 51: We can use the `Partial` type to make these types more composable and reusable.

src/widgets/answers/answers.tsx
```diff
@@ -0,0 +1,179 @@
...
+   */
+  loader?: Template;
+
+  /**
+   * Template to use for each result. This template will receive an object containing a single record.
+   */
+  item?: Template<Hit>;
+};
+
+export type AnswersCSSClasses = {
```
francoischalifour, line 71: We can use the `Partial` type to make these types more composable and reusable.

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,251 @@
...
+    let isLoading = false;
+    const debouncedRender = debounce(renderFn, renderDebounceTime);
+    let debouncedRefine;
+
+    return {
+      $$type: 'ais.answers',
+
+      init(initOptions) {
+        const { state, instantSearchInstance } = initOptions;
+        const answersIndex = instantSearchInstance.client!.initIndex!(
```
francoischalifour, line 123: Just checking but is it a use case to use another search client (i.e. another application  ID) for Answers?

src/connectors/answers/connectAnswers.ts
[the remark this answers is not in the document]
yannickcr: I think you should be able to create the type guard using the types already defined in https://github.com/algolia/instantsearch.js/blob/master/src/types/algoliasearch.ts

Something like:
```js
export function isSearchClientV4(
  searchClient: SearchClientV3 | SearchClientV4
): searchClient is SearchClientV4 {
  return (searchClient as SearchClientV4).transporter !== undefined;
}
```

src/connectors/answers/connectAnswers.ts
Haroenv: it was an explicit choice from the answers team to make users pass queryLanguages, so later when other languages are added, the language isn't wrong

src/connectors/answers/connectAnswers.ts
Haroenv: at the moment there would be use cases for that as well in multi-index, but we decided that you should use a custom client to merge the requests. I think that for now at least we should do the same with answers

francoischalifour: Then we should do the same with React InstantSearch, which [does accept a `searchClient` prop](https://github.com/algolia/react-instantsearch/blob/0228c51c2d65e477fbce6471e04a85597f5bfda6/packages/react-instantsearch-dom/src/hooks/useAnswers.js#L13) right now.

src/widgets/answers/answers.tsx
francoischalifour: Yes, like so:

```suggestion
export type AnswersTemplates = Partial<{ /* ... */ }>
```

src/connectors/answers/connectAnswers.ts
eunjae-lee: > Just checking but is it a use case to use another search client (i.e. another application ID) for Answers?

I initially misunderstood your question. I'm not so sure about another application ID. We do not have any case using different application IDs under a single instantSearchInstance. For multi-index, we should support <index> widget, though.

The reason why we're accepting `searchClient` in React InstantSearch is because I don't have access to the `client` like I do in IS.js. Not sure if we need to require another prop on the vanilla JS to make it look the same as React IS.

By the way, the next step for Answers widget is to pass all the facets from the current widget tree down to Answers API call, so I think it makes more sense to keep things under the same application ID.

francoischalifour: > The reason why we're accepting `searchClient` in React InstantSearch is because I don't have access to the `client` like I do in IS.js. Not sure if we need to require another prop on the vanilla JS to make it look the same as React IS.

I don't think that this current limitation should have such an impact on our API then. Getting the search client in the React `Answers` widget should be possible.

> By the way, the next step for Answers widget is to pass all the facets from the current widget tree down to Answers API call, so I think it makes more sense to keep things under the same application ID.

That makes sense!

eunjae-lee: > Getting the search client in the React Answers widget should be possible.
Not sure if I understood you correctly.
Do you mean it should be possible without explicitly accepting the search client?

francoischalifour: I mean that we should be able to get the search client via the React Context with some modifications. A technical limitation that we can solve shouldn't impact the user API. But we can discuss this on the React InstantSearch PR.

eunjae-lee: @francoischalifour I agree. Let's continue on the React IS PR.

src/widgets/answers/answers.tsx
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581/commits/e8275509afe232d9c6d740a71b2114000896ce73

src/widgets/answers/answers.tsx
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4581/commits/e8275509afe232d9c6d740a71b2114000896ce73

src/types/algoliasearch.ts
```diff
@@ -16,15 +16,27 @@ import algoliasearch, {
...
+  FindAnswersResponse as FindAnswersResponseV4,
   SearchResponse as SearchResponseV4,
+  FindAnswersOptions as FindAnswersOptionsV4,
   // no comma, TS is particular about which nodes expose comments
   // eslint-disable-next-line prettier/prettier
-  SearchForFacetValuesResponse as SearchForFacetValuesResponseV4
+  SearchForFacetValuesResponse as SearchForFacetValuesResponseV4,
   /** @ts-ignore */
   // @ts-ignore
   // eslint-disable-next-line import/no-unresolved
```
shortcuts, line 27: These lines don't anything for me anymore, is that normal?

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,258 @@
...
+  FindAnswersOptions,
+  FindAnswersResponse,
+} from '../../types';
+
+type IndexWithAnswers = {
+  readonly findAnswers: any;
+};
+
+function hasFindAnswersMethod(
+  answersIndex: IndexWithAnswers | any
```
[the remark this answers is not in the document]
Haroenv:
```suggestion
  answersIndex: object
```

since it's a type guard, it should accept as broad as possible, in this case an object (not any) but asserts it has answers

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,258 @@
...
+import {
+  Connector,
+  Hits,
+  Hit,
+  FindAnswersOptions,
+  FindAnswersResponse,
+} from '../../types';
+
+type IndexWithAnswers = {
+  readonly findAnswers: any;
```
Haroenv, line 20:
```suggestion
  readonly findAnswers: ReturnType<SearchClient['initIndex']>['findAnswers'];
```

(but with some Required) should do the trick I think

src/connectors/answers/connectAnswers.ts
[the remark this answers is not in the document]
Haroenv: maybe it could even accept `ReturnType<SearchClient['initIndex']>` as argument, and only assert that findAnswers exists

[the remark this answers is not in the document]
eunjae-lee:
```
type IndexWithAnswers = {
  readonly findAnswers: any;
} & ReturnType<Exclude<SearchClient['initIndex'], undefined>>;

function hasFindAnswersMethod(
  answersIndex: ReturnType<Exclude<SearchClient['initIndex'], undefined>>
): answersIndex is IndexWithAnswers {
  return typeof (answersIndex as IndexWithAnswers).findAnswers === 'function';
} 
```

I tried this but got this

image

Anything related to "answers" will fail in client v3 because it doesn't have `findAnswers` method and related types. The current version is what @yannickcr and I got so far, but let's see next time if it can be improved.

src/types/algoliasearch.ts
eunjae-lee: If you run these two lines, it will fail without the line:
https://github.com/algolia/instantsearch.js/blob/feat/answers/.circleci/config.yml#L70:L71

src/components/Answers/__tests__/Answers-test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    },
+  },
+};
+
+describe('Answers', () => {
+  describe('Rendering', () => {
+    it('renders without anything', () => {
+      const { container } = render(<Answers {...defaultProps} />);
+      expect(container.querySelector('.root')!.classList).toContain('empty');
+      expect(container).toMatchSnapshot();
```
francoischalifour, line 32: How about inlining snapshots? It's easy to lose context and miss parts when reviewing this kind of snapshots.

src/components/Answers/__tests__/Answers-test.tsx
```diff
@@ -0,0 +1,92 @@
...
+        },
+      };
+      const { container } = render(
+        <Answers
+          {...props}
+          isLoading={false}
+          hits={[{ objectID: '1', title: 'hello!', __position: 1 }]}
+        />
+      );
+      expect(container.querySelector('.list')!.textContent).toEqual(
```
francoischalifour, line 87: We can use `toHaveTextContent`.

src/components/Answers/__tests__/Answers-test.tsx
```diff
@@ -0,0 +1,92 @@
...
+      item: 'item',
+    },
+  },
+};
+
+describe('Answers', () => {
+  describe('Rendering', () => {
+    it('renders without anything', () => {
+      const { container } = render(<Answers {...defaultProps} />);
+      expect(container.querySelector('.root')!.classList).toContain('empty');
```
francoischalifour, line 31: We can use `toHaveClass`.

eunjae-lee: thanks for the suggestion https://github.com/algolia/instantsearch.js/pull/4581/commits/4f68c9e5503356e88d05edc1da21e19ccde27411 for all the three comments you recently left.

src/types/algoliasearch.ts
Haroenv: the issue doesn't happen when typescript is ran in this project, but with the built .d.ts files

src/widgets/answers/answers.tsx
```diff
@@ -0,0 +1,179 @@
...
+   */
+  loader: Template;
+
+  /**
+   * Template to use for each result. This template will receive an object containing a single record.
+   */
+  item: Template<Hit>;
+}>;
+
+export type AnswersCSSClasses = Partial<{
```
Haroenv, line 71: the partial should be used in the widget arguments, while the type should have everything required

eunjae-lee: that's correct 👍🏼 https://github.com/algolia/instantsearch.js/pull/4581/commits/04a9da7305a6c52809e857f2bce8017c9d473063

src/connectors/answers/connectAnswers.ts
```diff
@@ -0,0 +1,258 @@
...
+      queryLanguages,
+      attributesForPrediction,
+      nbHits = 1,
+      renderDebounceTime = 100,
+      searchDebounceTime = 100,
+      escapeHTML = true,
+      extraParameters = {},
+    } = widgetParams || ({} as typeof widgetParams);
+
+    // @ts-ignore checking the wrong parameter
```
Haroenv, line 114:
```suggestion
    // @ts-ignore checking for the wrong value
```

Haroenv (approved):

PR #4582, opened 2020-11-18
feat(configure): getRenderState for multiple configure widgets

Haroenv:
DX-222

### DIFF
--- a/src/connectors/configure/connectConfigure.ts
+++ b/src/connectors/configure/connectConfigure.ts
@@ -29,8 +29,6 @@ export type ConfigureRendererOptions = {
    * Refine the given search parameters.
    */
   refine: Refine;
-
-  _identifier: number;
 };
 
 const withUsage = createDocumentationMessageGenerator({
@@ -60,8 +58,6 @@ export type ConfigureConnector = Connector<
   ConfigureRendererOptions,
   ConfigureConnectorParams
 >;
-
-let uniqueIdentifier = 0;
 
 const connectConfigure: ConfigureConnector = function connectConfigure(
   renderFn = noop,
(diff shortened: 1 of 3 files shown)


Haroenv:
All in all, I feel like this explores all the options, meaning that if someone really requires this, they would need to fork connectConfigure, and add a feature very similar to this PR, but that we shouldn't merge it as-is and rather keep the last added configure widget as the only renderState

francoischalifour:
The way that I saw this was that we don't need identifiers. At the end of the day, with the Configure render state, you want to know what are _all_ the search parameters that it applied.

So I think it's fine to merge all the Configure render state properties in there—with a priority to the last props if duplicated, just like how `searchParameters` are sent to Algolia.

Wdyt?

Haroenv:
ah yes good idea, we can merge all searchParameters. Thanks for the idea @francoischalifour

src/connectors/configure/connectConfigure.ts
```diff
@@ -132,13 +132,23 @@ const connectConfigure: ConfigureConnector = function connectConfigure(
...
-          // If we want to merge widgetRenderState of multiple configure widgets,
-          // we should modify this part.
-          configure: this.getWidgetRenderState(renderOptions),
+          configure: {
+            ...widgetRenderState,
+            widgetParams: {
+              ...widgetRenderState.widgetParams,
+              searchParameters: mergeSearchParameters(
+                new algoliasearchHelper.SearchParameters(
+                  renderState.configure?.widgetParams.searchParameters
```
shortcuts, line 144: Not sure if it's needed but we could add a fallback (`|| {}`) in case configure is `undefined`?

Haroenv: it's not needed, since undefined is a valid input for search parameters :)

eunjae-lee (approved):
It looks good to me.

shortcuts (approved):

PR #4583, opened 2020-11-18
fix(renderState): initialize sendEvent before init

yannickcr:
**Summary**

Similar to https://github.com/algolia/instantsearch.js/pull/4579, this PR moves the `sendEvent` initialization into `getRenderState` so it is properly initialized if it gets called before `init`.

The `init` calls were also removed from the `getRenderState` and `getWidgetRenderState` tests and the tests were splitted (with and without results).


eunjae-lee (approved):
Looks good to me

shortcuts (approved):

PR #4584, opened 2020-11-18
ci(polyfills): enfore no unwanted polyfills are used

Haroenv:
**Summary**

We want to avoid polyfills to be needed for eg. Array.find, since it's not supported on IE11 and avoid issues like https://github.com/algolia/instantsearch.js/pull/4580

**Result**

adds a new babel plugin for umd compilation only which is normally meant for injecting polyfills, but we use to error when a polyfill would be needed

In theory we could write our own polyfill provider ([docs](https://github.com/babel/babel-polyfills/blob/main/docs/polyfill-provider.md)), but that seems more complicated that using the existing es-shims one.

Thanks @ljharb for helping me out on the babel slack with understanding this!

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -68,7 +68,7 @@ module.exports = api => {
             // this is actually only called on arrays
             'String.prototype.includes',
           ];
-          if (defaultShouldInject && exclude.indexOf(name) === -1) {
+          if (defaultShouldInject && !exclude.includes(name)) {
             throw new Error(
               `Usage of a builtin which isn't allowed to be polyfilled: ${name}`
             );
--- a/babel.config.js
+++ b/babel.config.js
@@ -4,6 +4,7 @@ const wrapWarningWithDevCheck = require('./scripts/babel/wrap-warning-with-dev-c
 
 const isCJS = process.env.BABEL_ENV === 'cjs';
 const isES = process.env.BABEL_ENV === 'es';
+const isUMD = process.env.BABEL_ENV === 'umd';
 
 const clean = x => x.filter(Boolean);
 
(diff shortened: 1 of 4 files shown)


src/lib/utils/find.ts
```diff
@@ -6,14 +6,18 @@
...
+  for (let i = 0; i < items.length; i++) {
+    value = items[i];
+    // inlined for performance: if (Call(predicate, thisArg, [value, i, list])) {
+    if (predicate(value, i, items)) {
+      return value;
+    }
   }
 
-  return items.find(predicate, thisArg);
+  return undefined;
```
Haroenv, line 20: I could write a more complicated error check in the babel config so that it could for example allow Array.find exactly twice, but that seemed much more complicated than rewriting this to an as-efficient version as Array.find

Haroenv:
here's a demo failing test: https://app.circleci.com/pipelines/github/algolia/instantsearch.js/5087/workflows/7830b6b4-09b0-4225-8919-05d3b6d3fe34/jobs/22653 (from this PR: https://github.com/algolia/instantsearch.js/pull/4585)

src/lib/utils/find.ts
eunjae-lee: Makes sense

eunjae-lee (approved):
It's really handy 👍

babel.config.js
```diff
@@ -44,6 +45,38 @@ module.exports = api => {
...
+            'Array.prototype.values',
+            'Function.prototype.name',
+
+            // we require polyfills for this already
+            'Array.prototype.includes',
+            // false positive (babel doesn't know types)
+            // this is actually only called on arrays
+            'String.prototype.includes',
+          ];
+          if (defaultShouldInject && exclude.indexOf(name) === -1) {
```
francoischalifour, line 71: Perhaps use `.includes()` here?

francoischalifour (approved):
Really handy!

PR #4573, opened 2020-11-05
feat(widgets): add getWidgetRenderState and getRenderState

Haroenv:
**Summary**

this new hook allows us to read the render state from every connector to be able to refine from outside of a widget and from the arguments in panel for a wrapped widget

**Result**

- all widgets have getWidgetRenderState, getRenderState
- you can access `renderState` from the InstantSearch instance
- you can access the render state from hidden, collapsed and templates in the panel widget

PR #4587, opened 2020-11-20
fix(range): consistently convert min & max to numbers

Haroenv:
Basically we had two methods to turn a number (or string) into a precision:

1. Math.floor(min * pow) / pow
2. Number(Number(v).toFixed(precision))

these behave different for precision 0 & a number over .5, let's say 1.5 turns into 1 for method 1, and into 2 for method 2. This is a bug (not caught by most of our tests, notably the ones i had to change) and causes the range to change to a number it's not supposed to be. Pay special attention to the new `toPrecision` utility, and the places it was used.

To efficiently compare this, you will have to compare some of the test files in vscode, since they changed too much after transforming to TypeScript

At the same time I also converted to TypeScript, but not the components themselves, since that was too much work.

typescript for tests is expected to fail until https://github.com/algolia/algoliasearch-client-javascript/pull/1229 is released (since now certain parts aren't random objects, but rather search responses)

The version for the helper is bumped for better typescript support as well.


Haroenv:
let's merge this as a separate commit to #4573 though, since it fixes a bug separate from the transition to getWidgetRenderState

src/connectors/range/__tests__/connectRange-test.ts
```diff
@@ -0,0 +1,2151 @@
...
+            helper,
+            attribute: 'price',
+          }),
+        })
+      );
+
+      // @ts-ignore
+      refine(['10.50', '490.50']);
+
+      expect(helper.getNumericRefinement(attribute, '>=')).toEqual([10]);
```
eunjae-lee, line 770: it'd be nice to have a little comment about floor for min and ceil for max.

eunjae-lee (approved):
Indeed the changes are big, but it wasn't that difficult to compare them in VSCode, especially because the changes were very consistent.

Looks good to me.

src/connectors/range/__tests__/connectRange-test.ts
Haroenv: 36e87c07

src/connectors/range/connectRange.ts
```diff
@@ -14,68 +22,128 @@ const withUsage = createDocumentationMessageGenerator(
...
- * previously set bound or to set an infinite bound.
- * @property {{min: number, max: number}} range Results bounds without the current range filter.
- * @property {Array<number, number>} start Current numeric bounds of the search.
- * @property {{from: function, to: function}} formatter Transform for the rendering `from` and/or `to` values.
- * Both functions take a `number` as input and should output a `string`.
- * @property {Object} widgetParams All original `CustomRangeWidgetOptions` forwarded to the `renderFn`.
- */
+export type Min = number | undefined;
+export type Max = number | undefined;
+export type Start = [Min, Max];
```
francoischalifour, line 27: Since we export these types, can we prefix them with `Range`?

francoischalifour, line 27: Would `RangeBoundaries` make more sense?

```suggestion
export type RangeBoundaries = [Min, Max];
```

src/connectors/range/connectRange.ts
```diff
@@ -110,40 +181,39 @@ export default function connectRange(renderFn, unmountFn = noop) {
       const isValidNewNextMin = isFiniteNumber(newNextMin);
       const isValidMinCurrentRange = isFiniteNumber(currentRangeMin);
       const isGreaterThanCurrentRange =
-        isValidMinCurrentRange && currentRangeMin <= newNextMin;
+        isValidMinCurrentRange && currentRangeMin! <= newNextMin!;
```
francoischalifour, line 184: If this repo tooling supports it, we can start using [`assert`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) functions so that the TS compiler understands that for this code path, those variables exist.

src/connectors/range/connectRange.ts
```diff
@@ -342,10 +420,7 @@ export default function connectRange(renderFn, unmountFn = noop) {
         // can not use setQueryParameters || removeNumericRefinement, because
         // they both keep the old value. This isn't immutable, but it is fine
         // since it's already a copy.
-        stateWithoutDisjunctive.numericRefinements = {
-          ...state.numericRefinements,
-          [attribute]: undefined,
-        };
+        delete stateWithoutDisjunctive.numericRefinements[attribute];
```
francoischalifour, line 423: Any particular reasons for this change?

src/connectors/range/connectRange.ts
Haroenv: isFiniteNumber already does that, but since we saved it to variables, it doesn't work in all cases. I didn't want to risk making too many changes here (since the variables were chosen to make the final equation not too complex). Not sure what you prefer

src/connectors/range/connectRange.ts
Haroenv: undefined isn't a valid value according to the type, since `x?: y` isn't the same as `x: y | undefined`

src/connectors/range/connectRange.ts
Haroenv: I prefer boundaries, but the property in the renderState which is called `start`, so I went with the same

Haroenv: are you suggesting this?

```
export type RangeMin = number | undefined;
export type RangeMax = number | undefined;
export type RangeStart = [RangeMin, RangeMax];
export type RangeRange = {
  min: RangeMin;
  max: RangeMax;
};
```

or keep RangeRange as Range?

francoischalifour: Yes, and keeping `Range`.

francoischalifour: It's not only used for `start` so it's fine for the name to differ: https://github.com/algolia/instantsearch.js/blob/0844f036cc51938a2ac6b3fccd6ff5ee948dc93a/src/connectors/range/connectRange.ts#L40

src/connectors/range/connectRange.ts
francoischalifour: Fine—we can introduce this pattern in future PRs. The main advantage is that we don't need to use `!` at every use, but only assert that the variable exists in the scope (it'll also be handy in our `index` widget where the lifecycle mutates the "class variables").

src/connectors/range/connectRange.ts
francoischalifour: Right, we might as well change the type in the Helper then?

src/connectors/range/connectRange.ts
Haroenv: changed it to need less !

src/connectors/range/connectRange.ts
Haroenv: the type in the helper is `numericRefinements: { [facet: string]: SearchParameters.OperatorList }`, do you think that should be `numericRefinements: { [facet: string]: SearchParameters.OperatorList | undefined }`? I feel like we'd need to change a couple more types for that to be consistent. I'd rather keep the delete, it's not a performance footgun anymore nowadays

Haroenv: with https://github.com/algolia/algoliasearch-helper-js/pull/801 we can use this for dispose instead (much cleaner IMO)

```js
        return state
          .removeDisjunctiveFacet(attribute)
          .removeNumericRefinement(attribute);
```

Haroenv: fixed in https://github.com/algolia/instantsearch.js/pull/4587/commits/93dd0e5edcfc5048c44b94e235a3b0412eff75fa

francoischalifour (approved):

PR #4588, opened 2020-11-23
feat(highlight): accept array for attribute

Haroenv:
React InstantSearch already supports this form for attribute, and it gets documented in https://github.com/algolia/doc/pull/5141

In a future major we can drop the string form for a safer access to the values (especially if eg. the value contains dots)

### DIFF
--- a/src/helpers/__tests__/snippet-test.ts
+++ b/src/helpers/__tests__/snippet-test.ts
@@ -1,4 +1,7 @@
-import highlight from '../highlight';
+import snippet from '../snippet';
+
+const NONE = 'none' as const;
+const FULL = 'full' as const;
 
 /* eslint-disable @typescript-eslint/camelcase */
 const hit = {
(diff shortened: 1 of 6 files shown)


eunjae-lee (commented):
In React InstantSearch, we support for finding value through arrays in string attribute like:

```
animal.cats[1].name
```

However, the string attribute is going to go away, so we don't need to care.

With this PR, the array attribute already works:

```js
['animal', 'cats', 1, 'name']
```

but we should fix the type to allow number as a part of path. What do you think?

Haroenv:
I was thinking of that, but didn't want to bloat the PR. Number should indeed be allowed too, even though it gets coerced to a string. Today it already works as `animal.cats.1.name`, but that's not very discoverable.

eunjae-lee:
> I was thinking of that, but didn't want to bloat the PR. Number should indeed be allowed too, even though it gets coerced to a string. Today it already works as `animal.cats.1.name`, but that's not very discoverable.

Ah right. My bad. Not `animal.cats[1].name` but `animal.cats.1.name`.

What I wanted to ask you was,

```
function getPropertyByPath(
  const parts = path.split('.');
  object: object | undefined,
  path: string | string[]
)
```

What if the type of `path` is `string | Array<string | number>`. Does it make sense?

eunjae-lee (approved):

src/helpers/__tests__/snippet-test.ts
```diff
@@ -122,4 +123,33 @@ describe('snippet', () => {
...
+        attribute: 'categories.1',
+        hit,
+      })
+    ).toMatchInlineSnapshot(`"Streaming Media Players"`);
+  });
+
+  test('with array attribute as array', () => {
+    expect(
+      snippet({
+        attribute: ['categories', 1],
```
francoischalifour, line 150: Since it's a path it makes sense to _only_ accept strings:

```
attribute: ['categories', '1'],
```

It'll make the type more simple.

src/helpers/highlight.ts
```diff
@@ -4,7 +4,8 @@ import { TAG_REPLACEMENT } from '../lib/escape-highlight';
 import { component } from '../lib/suit';
 
 export type HighlightOptions = {
-  attribute: string;
+  // @MAJOR only accept array of paths here
+  attribute: string | Array<string | number>;
```
francoischalifour, line 8: Since it's a path it makes sense to _only_ accept strings:

```
attribute: ['categories', '1'],
```

It'll make the type less permissive.

src/helpers/highlight.ts
```diff
@@ -4,7 +4,8 @@ import { TAG_REPLACEMENT } from '../lib/escape-highlight';
 import { component } from '../lib/suit';
 
 export type HighlightOptions = {
-  attribute: string;
+  // @MAJOR only accept array of paths here
```
francoischalifour, line 7: `string` and `string[]` probably. You don't want to force an array since a string covers 95% use cases. Besides, it'll mean renaming the option to `attributePath`.

src/helpers/__tests__/snippet-test.ts
Haroenv: either version works, since it gets coerced anyway when you do a key check, but I'm ok reverting my previous commit

src/helpers/highlight.ts
Haroenv: so you're suggesting no change in a future major then? if we'd move to the array form only in the docs, doesn't it also make sense to in time move to the array form only here?

francoischalifour: I meant the `string` type would be for attributes like `label`, and wouldn't support nested attributes (e.g., `lvl0.label`). We'd dev warn about it and guide toward the `string[]` type.

Haroenv: I see, good point!

Haroenv: Updated in 19514d4

src/lib/utils/getPropertyByPath.ts
```diff
@@ -1,5 +1,8 @@
-function getPropertyByPath(object: object, path: string): any {
-  const parts = path.split('.');
+function getPropertyByPath(
+  object: object | undefined,
+  path: string | string[]
+): any {
+  const parts = Array.isArray(path) ? path : path.split('.');
```
francoischalifour, line 5: Do we want to start warning already for string paths?

francoischalifour (approved):

src/lib/utils/getPropertyByPath.ts
Haroenv: adding a subsequent PR after we've updated the docs

PR #4592, opened 2020-11-26
feat(reverseHighlight/reverseSnippet): Implements reverseHighlight and reverseSnippet

shortcuts:
This implements `reverseHighlight` and `reverseSnippet`.


.storybook/static/storybook.css
```diff
@@ -258,3 +258,10 @@ body {
   background: #f7f7f7;
   color: black;
 }
+
+/* tmp class during dev phase for ReverseHighlight and reverseSnippet component */
+.ais-ReverseHighlight-highlighted,
```
francoischalifour, line 263: To give more context to the team here:

Designs for reverse highlighting are not setting background color like for regular highlighting. This would be awkward because by definition, multiple isolated parts can match.

image

Designs usually only consist of a font color and/or font weight.

image

It therefore might make sense to support a new CSS class in our [InstantSearch Specs](https://github.com/algolia/instantsearch-specs).

.storybook/static/storybook.css
```diff
@@ -258,3 +258,10 @@ body {
...
+.ais-ReverseSnippet-highlighted {
+  background: transparent;
```
francoischalifour, line 265:
```suggestion
  background: none;
```

src/helpers/reverseHighlight.ts
```diff
@@ -0,0 +1,42 @@
...
+import { component } from '../lib/suit';
+
+export type ReverseHighlightOptions = {
+  // @MAJOR string should no longer be allowed to be a path, only array can be a path
+  attribute: string | string[];
+  highlightedTagName?: string;
+  hit: Partial<Hit>;
+  cssClasses?: {
+    highlighted?: string;
+  };
```
francoischalifour, line 13: This makes types easier to compose.

```suggestion
  cssClasses?: Partial<{
    highlighted: string;
  }>;
```

src/lib/utils/__tests__/getReversedHighlight-test.ts
```diff
@@ -0,0 +1,39 @@
...
+      expect(getReversedHighlight('')).toEqual('');
+    });
+
+    test('with space only string', () => {
+      expect(getReversedHighlight('         ')).toEqual('         ');
+    });
+
+    test('with one match', () => {
+      expect(getReversedHighlight('this is a <mark>match</mark>')).toEqual(
+        '<mark>this</mark> <mark>is</mark> <mark>a</mark> match'
```
francoischalifour, line 15: The algorithm in Autocomplete outputs the following format, which results in less markup:

```html
<mark>cell </mark>phone<mark> accessories disc </mark>ch<mark>argers opp</mark>
```

Not sure which is best.

src/lib/utils/getReversedHighlight.ts
```diff
@@ -0,0 +1,33 @@
+import unescape from './unescape';
+
+const getReversedHighlight = (attribute: string) => {
```
francoischalifour, line 3: I think that the Autocomplete way of doing it (inspired by the way React InstantSearch does it) is more API-based, and makes more sense rather than using Regex.

1. We first parse the attribute to have the highlighted parts. This step would be valuable in InstantSearch.js to then expose the highlighted parts to Vue InstantSearch and Angular InstantSearch (so that do don't rely on unsafe XSS APIs): https://github.com/algolia/autocomplete.js/blob/547719b1e59f76474ad68e2b65dbefb3d64772be/packages/autocomplete-preset-algolia/src/highlight/parseAttribute.ts#L18
2. We reverse these matches to get the reverse highlighted matches _with the same API_: https://github.com/algolia/autocomplete.js/blob/next/packages/autocomplete-preset-algolia/src/highlight/reverseHighlightedParts.ts#L3
3. We expose the API to our renderers: https://github.com/algolia/autocomplete.js/blob/next/packages/autocomplete-preset-algolia/src/highlight/parseAlgoliaHitReverseHighlight.ts#L11
4. Our renderers use this API to speak the right language given the framework (here, vanilla JavaScript): https://github.com/algolia/autocomplete.js/blob/547719b1e59f76474ad68e2b65dbefb3d64772be/packages/autocomplete-js/src/highlight.ts#L39-L54

src/lib/utils/__tests__/getReversedHighlight-test.ts
Haroenv: I wonder if this could be better:

```
___pre____post___pre___post___

swap tags

___post____pre___post___pre___

add pre in the beginning and post in the end

pre___post____pre___post___pre___post

optional: remove empty tags in beginning and end, if like this

prepost____pre___post___prepost

to this

____pre___post___

```

src/lib/utils/getReversedHighlight.ts
Haroenv: it would indeed be interesting to get an array of parts (regardless of which attribute it comes from), and using the highlight pre and post tag to split it. Then you can map over them to turn them into highlighted or non-highlighted parts (then joining up again for plain html)

src/lib/utils/__tests__/getReversedHighlight-test.ts
shortcuts: > The algorithm in Autocomplete outputs the following format, which results in less markup:
> 
> ```
> <mark>cell </mark>phone<mark> accessories disc </mark>ch<mark>argers opp</mark>
> ```
> 
> Not sure which is best.

The output is like that because I've followed the actual `highlight` schema which avoid adding characters other than alphanumerical in `<mark>`. If it's not necessary, it's indeed better this way and would reduce the complexity of the function

src/lib/utils/getReversedHighlight.ts
shortcuts: It was actually the first method I was doing but I thought it was a bit complex/overkill compared to simply reversing the highlighted parts.

I'll give it an other try :D

shortcuts: I think I didn't get what it was at first place because now that it's almost done, I feel dumb thinking that it was overkill, doing tests and I'll push it ASAP!

shortcuts: @francoischalifour in this method, you don't escape the highlight on non alphanumeric characters, right? 

Should I implement it to match `highlight` behaviour?

francoischalifour: We escape these characters by default: https://github.com/algolia/autocomplete.js/blob/547719b1e59f76474ad68e2b65dbefb3d64772be/packages/autocomplete-preset-algolia/src/highlight/parseAttribute.ts#L5-L11

The way we do it in Autocomplete might differ a bit from InstantSearch, but we shouldn't change how it's done in IS.js now since it'll probably increase the scope of this feature.

shortcuts:
I ran into a problem with **reverseHighlighting** and **non alphanumeric characters** where I didn't find which logic to follow. I expose below the 3 method I came across so I can get your thoughts on that.

Could you please let me know what you think/which way is the best for you?

Sorry for the low quality post

1. The **actual behavior** that autocomplete follows, which leaves any non alphanumeric character **bold**.
actual_behavior

2. The **sibling behavior**, where the non alphanumeric character will take the highlighted value of the **previous** and **next** element.
sibling_behavior

3. The **highlight behavior**, where it just escape any non alphanumeric characters like the current **highlight** method.
highlight_behavior

cc @algolia/instantsearch-for-websites

Haroenv:
Thanks for trying out all those options @shortcuts, I'm a fan of the method #2's output, and I think it would be better for highlight itself too.

src/lib/utils/reverseHighlightedParts.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  // #1 // Actual behavior, same as Autocomplete
+  //   return parts.map(part => ({ ...part, isHighlighted: !part.isHighlighted }));
+
+  // #2 // Sibiling matching behavior, match the highlight of the siblings
+  return parts.map((part, i) => ({
+    ...part,
+    isHighlighted: !getHighlightFromSiblings(parts, i),
+  }));
```
shortcuts, line 16: @Haroenv 

I've split the logic in `getHighlightFromSiblings` so it returns either the matching siblings `isHighlighted` value (if they match) or the value of the current part.

This way we can use it in `reverseHighlightedParts` or `highlight`

src/lib/utils/reverseHighlightedParts.ts
```diff
@@ -0,0 +1,39 @@
+import { HighlightedParts } from '../../types';
```
shortcuts, line 1: The logic of the 3 methods are here but I did my implementation with the #2, in case we don't agree on it.

I'll remove it once the PR move out of draft.

francoischalifour:
@shortcuts The "sibling behavior" version is interesting. The only place I find it a bit weird is:

> (**Model)**

[...] where the last `)` is bold. It makes me wonder is never highlighting non-alphanumeric characters would make sense.

Have you seen this pattern anywhere else?

shortcuts:
@francoischalifour I've never seen this pattern used before and this behavior is the only reason why I have a doubt about it.

I could think of a workaround for the parenthesis where if something inside is highlighted, we remove the highlight on the opening/closing one but that would leave some characters bold like 

| Before      | After |
| ----------- | ----------- |
| <pre>(Mod<b>elExample)</b></pre>      | <pre>(Mod<b>elExample</b>)</pre>       |

The thing is that it would add even more computation on top of what's already done

Haroenv:
I think the sibling behaviour in that case is actually correct, since you are matching on the rest of the sentence, including that closing parenthesis. I think (Mode**lExample)** looks more "pay attention from here onwards", while (Mode**lExample**) looks like "pay only attention to 'lExample'", which will be odd if there's other highlighted text after this word.

Google's autocomplete seems to try following a sibling behaviour, although it often sees separators as typos and won't highlight the match

shortcuts:
You're right, they are also doing it.

I didn't find a good "long sentence in parentheses" to do more tests but that's already a good point.

Screenshot 2020-12-07 at 10 05 56

francoischalifour (approved):
Looks good—great job!

Since the bundlesize integration is disabled right now, what's the bundle size increase of this feature?

src/lib/utils/unescape.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * This implementation is taken from Lodash implementation.
+ * See: https://github.com/lodash/lodash/blob/4.17.11-npm/unescape.js
+ */
+
+// Used to map HTML entities to characters.
+const htmlEscapes = {
```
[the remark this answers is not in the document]
shortcuts: Wouldn't it add a bit of computation as the `key:value` are swapped?

shortcuts:
Here's what I got for the bundle size
```zsh
 FAIL  ./dist/instantsearch.production.min.js: 65.29KB > maxSize 65.25KB (gzip)
 PASS  ./dist/instantsearch.development.js: 151.83KB < maxSize 160KB (gzip)
```

src/lib/utils/unescape.ts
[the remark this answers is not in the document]
francoischalifour: Ah yep didn't realize that!

francoischalifour:
You can increase it to 65.5KB then.

src/lib/utils/unescape.ts
```diff
@@ -0,0 +1,29 @@
...
+  '&amp;': '&',
+  '&lt;': '<',
+  '&gt;': '>',
+  '&quot;': '"',
+  '&#39;': "'",
+};
+
+// Used to match HTML entities and HTML characters.
+const regexEscapedHtml = /&(amp|quot|lt|gt|#39);/g;
```
francoischalifour, line 16: Can this be constructed from `htmlEscapes`?

shortcuts: I'm not sure on how to do it without adding a loop to get all the keys

src/lib/utils/__tests__/concatHighlightedParts-test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const oneMatch: HighlightedParts[] = [
+  { isHighlighted: false, value: 'Amazon' },
+  {
+    isHighlighted: true,
+    value: ' - Fire HD8 - 8&quot; - Tablet - 16GB - Wi-Fi - Black',
+  },
+];
+
+const multipleMatches: HighlightedParts[] = [
```
Haroenv, line 12: I would put these arrays inside the relevant test, not top-level

src/lib/utils/__tests__/getHighlightedParts-test.ts
```diff
@@ -0,0 +1,37 @@
+import getHighlightedParts from '../getHighlightedParts';
+
+const oneMatch =
+  '<mark>Amazon</mark> - Fire HD8 - 8&quot; - Tablet - 16GB - Wi-Fi - Black';
+
+const multipleMatches =
```
Haroenv, line 6: same here, I'd prefer the input to be in the test

src/lib/utils/__tests__/getHighlightFromSiblings-test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const oneMatch: HighlightedParts[] = [
+  { isHighlighted: true, value: 'Amazon' },
+  {
+    isHighlighted: false,
+    value: ' - Fire HD8 - 8&quot; - Tablet - 16GB - Wi-Fi - Black',
+  },
+];
+
+const multipleMatches: HighlightedParts[] = [
```
Haroenv, line 12: same

src/lib/utils/__tests__/reverseHighlightedParts-test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const oneMatch: HighlightedParts[] = [
+  { isHighlighted: false, value: 'Amazon' },
+  {
+    isHighlighted: true,
+    value: ' - Fire HD8 - 8&quot; - Tablet - 16GB - Wi-Fi - Black',
+  },
+];
+
+const multipleMatches: HighlightedParts[] = [
```
Haroenv, line 12: same

src/lib/utils/getHighlightFromSiblings.ts
```diff
@@ -0,0 +1,20 @@
+import unescape from './unescape';
+import { HighlightedParts } from '../../types';
+
+const getHighlightFromSiblings = (parts: HighlightedParts[], i: number) => {
+  const current = parts[i];
+  const isAlphanumeric = new RegExp(/\w/gi);
```
Haroenv, line 6: this regex can be constructed outside of the function so it doesn't need to be recreated

Haroenv: be careful as it's a global regex though, that sometimes has odd behaviour since regex is stateful

src/lib/utils/getHighlightFromSiblings.ts
```diff
@@ -0,0 +1,20 @@
+import unescape from './unescape';
+import { HighlightedParts } from '../../types';
+
+const getHighlightFromSiblings = (parts: HighlightedParts[], i: number) => {
```
Haroenv, line 4: this could be `export default getHighlightFromSiblings(parts: HighlightedParts[], i: number) {` which has a bit less syntax. Not really worried about it though

Haroenv (approved):

src/lib/utils/getHighlightFromSiblings.ts
shortcuts: Isn't it better to do named export? I know that we don't do it in instantsearch.js but just wondering

src/lib/utils/getHighlightFromSiblings.ts
```diff
@@ -1,9 +1,13 @@
 import unescape from './unescape';
 import { HighlightedParts } from '../../types';
 
-const getHighlightFromSiblings = (parts: HighlightedParts[], i: number) => {
+const isAlphanumeric = new RegExp(/\w/i);
```
Haroenv, line 4: now this should be called `hasAlphanumeric` I think, since it's only checking the presence, not whether it's fully alphanumeric 👍

shortcuts: Yep that make sense 🤔

Haroenv (approved):
looks good!

PR #4593, opened 2020-11-26
fix: remove a warning about insights that is not relevant anymore

eunjae-lee:
**Summary**

This PR removes a warning about insights that is not relevant anymore.
The warning message was useful when user added `search-insights` but forgot to pass to `insightsClient` prop to `instantsearch()` call, but now that we don't encourage that prop anymore in favor of insights middleware, this warning message misleads those who use insights middleware.


src/lib/InstantSearch.ts
```diff
@@ -192,17 +192,6 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
...
     );
 
-    warning(
-      Boolean(insightsClient) || !hasDetectedInsightsClient(),
-      withUsage(`InstantSearch detected the Insights client in the global scope.
-To connect InstantSearch to the Insights client, make sure to specify the \`insightsClient\` option:
-
-const search = instantsearch({
-  /* ... */
-  insightsClient: window.aa,
```
Haroenv, line 202: shouldn't instead we change this to suggest using the insights middleware instead?

eunjae-lee: but at that point, we do not have any guarantee whether or not user will ever call `search.use(...)`.

francoischalifour (approved):

PR #4595, opened 2020-12-01
fix(configure): ignore invalid userToken

eunjae-lee:
**Summary**

This PR fixes `configure` widget to ignore invalid userToken.

Although we have fixed nullish user token issues in `insights` middleware, if users still put `null` into `configure` widget, it was passing the `null` as a string to search api. We ignore it from now on.

### DIFF
--- a/src/connectors/configure/__tests__/connectConfigure-test.ts
+++ b/src/connectors/configure/__tests__/connectConfigure-test.ts
@@ -100,6 +100,35 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/configure/j
     );
   });
 
+  it('should not apply `null` as userToken', () => {
+    const makeWidget = connectConfigure(noop);
+    const widget = makeWidget({
+      searchParameters: {
+        // @ts-ignore-next-line
+        userToken: null,
+      },
+    });
+    expect(
+      widget.getWidgetSearchParameters!(new SearchParameters({}), {
+        uiState: {},
+      })
+    ).toEqual(new SearchParameters({}));
+  });
+
+  it('should not apply an empty string as userToken', () => {
+    const makeWidget = connectConfigure(noop);
+    const widget = makeWidget({
+      searchParameters: {
+        userToken: '',
+      },
+    });
+    expect(
+      widget.getWidgetSearchParameters!(new SearchParameters({}), {
+        uiState: {},
+      })
+    ).toEqual(new SearchParameters({}));
+  });
+
   it('should apply searchParameters with a higher priority', () => {
     const makeWidget = connectConfigure(noop);
     const widget = makeWidget({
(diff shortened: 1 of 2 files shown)


src/connectors/configure/connectConfigure.ts
```diff
@@ -78,18 +94,23 @@ const connectConfigure: ConfigureConnector = function connectConfigure(
 
     function refine(helper: AlgoliaSearchHelper): Refine {
       return (searchParameters: PlainSearchParameters) => {
+        const newSearchParameters = excludeInvalidUserToken(searchParameters);
```
Haroenv, line 97: hmm, I think instead this should be in a dev check only, we can't validate all search parameters in configure. There's other invalid combinations, but those get rejected by the API.

eunjae-lee: Many customers especially on ecommerce don't use bundlers, and they don't get dev warnings. And this must be something more than a warning, because it will definitely screw the analysis. However, I don't think it's something the API should "reject", because on their end, they just receive a string `"null"` which they just cannot reject or ignore. It's on our side to send a correct one, IMO.

Haroenv: I'm not sure this is the right place for this check though, can't you send userToken with different search parameter methods as well?

eunjae-lee: > send userToken with different search parameter methods

can you explain a little more, especially 'different search parameter methods' part?

Haroenv: you could set it with the helper too, or with configure related items, or with voice search. I'm not sure if this is the right place

eunjae-lee: Ah I've got your point. It makes sense. To block the invalid token fundamentally, we should do it in helper. What do you think?

Haroenv: I guess so, but aren't there other invalid values than null & ""? Won't this become a long list of values eventually?

eunjae-lee: I just checked and
* insights API: throws an error with invalid token
* search API: doesn't throw

So I guess we should not send any invalid user token to search API because there doesn't seem to be a guard. The rule is simple, though: `[a-zA-Z0-9_-]{1,64}` (https://www.algolia.com/doc/rest-api/insights/#method-param-usertoken)

yannickcr: Like @Haroenv I think we should do this check in the helper (if we agree that it is the helper responsibility to validate the parameters) and throw an exception in it in case of an invalid token (with an helpful error message to ease debugging)

src/connectors/configure/__tests__/connectConfigure-test.ts
```diff
@@ -196,6 +225,52 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/configure/j
...
+    // set a valid userToken
+    refine({ userToken: 'abc' });
+    expect(getUserToken()).toEqual('abc');
+
+    // @ts-ignore-next-line
+    refine({ userToken: null });
+    expect(getUserToken()).toEqual('abc');
+
+    refine({ userToken: '' });
+    expect(getUserToken()).toEqual('abc');
```
yannickcr, line 271: That behavior is very disturbing, as a user I would rather prefer to have an error when using an invalid userToken than a silent error with no explanation why my token was not updated.

src/connectors/configure/connectConfigure.ts
eunjae-lee: @yannickcr which one do you think is proper? exception vs warning
I feel like it's a bit too much to throw an exception.
My rationale is 

> The user token is wrong? okay, this request won't be analyzed. But let's still provide the search result.

yannickcr: Does sending an invalid token to the search API has any impact ? If no then a warning should suffice imho.

Haroenv: the problem is that if we throw an error, it will break their search interface, for something inconsequential to the actual search.

yannickcr: Yes, that's why I said a warning should be enough if an invalid userToken has no impact :wink:

eunjae-lee:
closing in favor of https://github.com/algolia/algoliasearch-helper-js/pull/802

PR #4596, opened 2020-12-03
chore(router): deduplicate uiState setting code

Haroenv:
**Summary**

Since we have added the router middleware, InstantSearch has gained the setUiState function which does practically the same (except it also warns for invalid state, which is useful for routing!).

**Result**

There's a couple existing tests on this behaviour that fail if the onUpdate code is removed

Bundle size should be slightly smaller, but likely not significantly

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -522,9 +522,10 @@ See ${createDocumentationLink({
     }
   }
 
-  public setUiState: (
-    uiState: UiState | ((uiState: UiState) => UiState)
-  ) => void = uiState => {
+  public setUiState(
+    uiState: UiState | ((uiState: UiState) => UiState),
+    suppressWarning = false
+  ): void {
     if (!this.mainHelper) {
       throw new Error(
         withUsage('The `start` method needs to be called before `setUiState`.')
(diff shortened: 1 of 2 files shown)


src/lib/InstantSearch.ts
```diff
@@ -522,17 +522,18 @@ See ${createDocumentationLink({
     }
   }
 
-  public setUiState: (
-    uiState: UiState | ((uiState: UiState) => UiState)
-  ) => void = uiState => {
+  public setUiState(
+    uiState: UiState | ((uiState: UiState) => UiState),
+    suppressWarning = false
```
Haroenv, line 527: I'm not sure whether we should just make it non-experimental now and remove the warning completely

yannickcr: `setUiState` is 9 month old, if until now we did not get any change to do to this API then I think it is safe to remove this warning and set it as stable.

yannickcr (approved):

src/lib/InstantSearch.ts
Haroenv: made a subsequent PR, we can discuss there: https://github.com/algolia/instantsearch.js/pull/4597

francoischalifour (approved):
Nice!

Good to me once we delete the warning logic.

Haroenv:
merging as two separate commits 👍

PR #4603, opened 2020-12-08
feat(index): expose createURL

Haroenv:
**Summary**

This is useful for recreating what an index widget does when it renders, see ssr in Vue, as well as poc/telemetry2

I think in the future we might want a function to retrieve the parameters given to render / init.

**Result**

index exposes the scoped `createURL` function to be used in render / init / getWidgetRenderState etc.

### DIFF
--- a/src/widgets/index/__tests__/index-test.ts
+++ b/src/widgets/index/__tests__/index-test.ts
@@ -14,6 +14,7 @@ import { runAllMicroTasks } from '../../../../test/utils/runAllMicroTasks';
 import { Widget, InstantSearch } from '../../../types';
 import index from '../index';
 import { warning } from '../../../lib/utils';
+import { refinementList } from '../..';
 
 describe('index', () => {
   const createSearchBox = (args: Partial<Widget> = {}): Widget =>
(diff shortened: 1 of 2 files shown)


Haroenv:
just realised the tests didn't get committed, let's add something now

src/widgets/index/__tests__/index-test.ts
```diff
@@ -651,6 +652,103 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index-widge
...
+      const instance = index({ indexName: 'indexName' });
+      const searchBox = createSearchBox();
+      const pagination = createPagination();
+
+      instance.addWidgets([searchBox, pagination]);
+
+      instance.init(
+        createInitOptions({
+          instantSearchInstance: createInstantSearch({
+            // @ts-ignore
```
shortcuts, line 678: Is it still needed as it's called at L679?

shortcuts (approved):
LGTM! 🥇

src/widgets/index/__tests__/index-test.ts
Haroenv: yes, this function is meant to return a string, but an object is easier to assert on

PR #4604, opened 2020-12-09
feat(core): expose metadata of widgets

Haroenv:
this adds a new middleware (enabled by default) which calls `getWidgetRenderState` on all widgets to get the widgetParams


src/widgets/places/places.ts
```diff
@@ -4,7 +4,7 @@ import {
   PlacesInstance,
   // no comma, TS is particular about which nodes expose comments
   // eslint-disable-next-line prettier/prettier
-  ReconfigurableOptions
+  ReconfigurableOptions,
```
Haroenv, line 7:
```suggestion
  ReconfigurableOptions
```

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,86 @@
...
+
+          const initOptions = {
+            instantSearchInstance,
+            parent,
+            results: parent.getResults()!,
+            scopedResults: resolveScopedResultsFromIndex(parent),
+            state: parent.getResults()!._state,
+            helper: parent.getHelper()!,
+            // @TODO: https://github.com/algolia/instantsearch.js/pull/4603
+            // createURL: parent.createURL,
```
Haroenv, line 51: merge only after https://github.com/algolia/instantsearch.js/pull/4603

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,86 @@
...
+  type: string;
+  params: string[];
+  officialWidget: boolean;
+};
+
+const ALGOLIA_CRAWLER_USER_AGENT = /Algolia Crawler\/[0-9]+.[0-9]+.[0-9]+/;
+
+export const createTelemetryMiddleware = (): Middleware => {
+  const isTelemetryEnabled =
+    typeof window !== undefined &&
```
eunjae-lee, line 14:
```suggestion
    typeof window !== 'undefined' &&
```

src/types/widget.ts
```diff
@@ -394,6 +398,12 @@ export type Widget<
     | 'ais.stats'
     | 'ais.toggleRefinement'
     | 'ais.voiceSearch';
+
+  /**
+   * private marker on widgets to differentiate a widget from a connector
+   */
+  $$officialWidget?: boolean;
```
eunjae-lee, line 405: Maybe `$$official` or `$$isOfficial`?

src/lib/InstantSearch.ts
```diff
@@ -250,6 +250,8 @@ See ${createDocumentationLink({
       const routerOptions = typeof routing === 'boolean' ? undefined : routing;
       this.use(createRouterMiddleware(routerOptions));
     }
+
+    this.use(createTelemetryMiddleware());
```
eunjae-lee, line 254: What do you think about this?: https://github.com/algolia/instantsearch-rfcs/pull/39#issuecomment-721798880

Haroenv: We can add an option, but I'm not sure if it's needed

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,84 @@
...
+        setTimeout(() => {
+          const widgets = instantSearchInstance.mainIndex.getWidgets();
+
+          const parent = instantSearchInstance.mainIndex;
+
+          const initOptions = {
+            instantSearchInstance,
+            parent,
+            // @TODO: https://github.com/algolia/instantsearch.js/pull/4609
+            // scopedResults: parent.getScopedResults(),
```
Haroenv, line 46: still to merge

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,81 @@
...
+            renderState: instantSearchInstance.renderState,
+            templatesConfig: instantSearchInstance.templatesConfig,
+            searchMetadata: {
+              isSearchStalled: instantSearchInstance._isSearchStalled,
+            },
+          };
+
+          widgets.forEach(widget => {
+            const widgetParams = widget.getWidgetRenderState
+              ? (widget.getWidgetRenderState(initOptions) as any).widgetParams
```
eunjae-lee, line 57: I think technically `widgetParams` returned from `getWidgetRenderState` can be nullish. For example, user implemented the function but returns null in some cases.

src/widgets/panel/panel.tsx
```diff
@@ -252,6 +252,7 @@ const panel: PanelWidget = widgetParams => {
 
     return {
       ...widget,
+      $$props: widgetOptions,
```
eunjae-lee, line 255: Can you explain what it is? I can't find `$$props` being used elsewhere.

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,81 @@
...
+  const payloadContainer = document.createElement('meta');
+  const refNode = document.querySelector('head')!;
+  payloadContainer.name = 'instantsearch:widgets';
+  refNode.appendChild(payloadContainer);
+
+  return ({ instantSearchInstance }) => {
+    return {
+      onStateChange() {},
+      subscribe() {
+        setTimeout(() => {
```
eunjae-lee, line 35: I don't remember correctly. Could you explain again about the Intermittent issue? IIRC, it was something about `getScopedResults` not having any result yet sometimes. Is `setTimeout` still the valid way to ensure it? I think you said you were going to fix it but not sure if it's applied here or not.

src/middlewares/createTelemetryMiddleware.ts
Haroenv: ah that's true, and would cause a nasty error

src/widgets/panel/panel.tsx
Haroenv: leftover from the old option, to remove, thanks!

src/middlewares/createTelemetryMiddleware.ts
Haroenv: it was about the `state` coming from the `results` at the time, but I changed it to come from the helper instead (like the index widget); and that works as expected!

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -1,11 +1,70 @@
...
+    );
+
+    payload.widgets.push({
+      type: widget.$$type || 'custom.widget',
+      params,
+      officialWidget: Boolean(widget.$$officialWidget),
+    });
+
+    if (widget.$$type === 'ais.index') {
+      extractPayload(
```
eunjae-lee, line 59: This seems to be gathering payloads in a flat array. Won't we lose information about the structure of nested indices?

Haroenv: yes, we lose the information for now, but flat information will be much easier to search; as well as indexName not being anonymous

eunjae-lee: Ah right. I've forgotten about it being stored in Algolia. And I didn't think of indexName. Then it makes sense to me to have it this way.

eunjae-lee (approved):
It looks good to me. It'd be good to have a snapshot test on the telemetry payload, so that we can have a quick glance on how it looks like, but not a blocker to merge the PR.

src/types/widget.ts
Haroenv: I don't have a strong opinion

shortcuts: `$$isOfficial` seems appropriate

shortcuts (approved):

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,110 @@
...
+  const refNode = document.querySelector('head')!;
+  payloadContainer.name = 'instantsearch:widgets';
+  refNode.appendChild(payloadContainer);
+
+  return ({ instantSearchInstance }) => {
+    return {
+      onStateChange() {},
+      subscribe() {
+        setTimeout(() => {
+          extractPayload(
```
yannickcr, line 95: IMHO it would be clearer for `extractPayload` to return the list of widget instead of updating the `payload` parameter as a side effect.

```js
payload.widgets = extractPayload(
    instantSearchInstance.mainIndex.getWidgets(),
    instantSearchInstance
);
```

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,110 @@
...
+    searchMetadata: {
+      isSearchStalled: instantSearchInstance._isSearchStalled,
+    },
+  };
+
+  widgets.forEach(widget => {
+    let widgetParams = {};
+
+    if (widget.getWidgetRenderState) {
+      const renderState = widget.getWidgetRenderState(initOptions);
```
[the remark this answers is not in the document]
Haroenv: I actually had a look, and while we are missing the parameters that are not sent to the connector (visual-only ones), we still retrieve all the values without their defaults, since those are only assigned on the destructuring on the connector. I think that's a fair tradeoff that doesn't require rewriting all widgets. I will however after this do a pass on all widgets, and make sure they don't set any defaults that are also set in the connectors

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,110 @@
...
+        widgetParams = renderState.widgetParams;
+      }
+    }
+
+    // since we destructure in all widgets, the parameters with defaults are set to "undefined"
+    const params = Object.keys(widgetParams).filter(
+      key => widgetParams[key] !== undefined
+    );
+
+    payload.widgets.push({
```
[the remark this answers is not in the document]
Haroenv: In the original proposal, the payload had to travel over the network, by not merging each type of widget, we have slightly more information (maybe someone uses 1, maybe someone uses 100 refinement lists, that sounds like interesting info to me)

src/middlewares/createTelemetryMiddleware.ts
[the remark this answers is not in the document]
Haroenv: I think in a second pass, we actually could send the visual-only parameters to the connector as well, since it's impossible for the parameters to clash significantly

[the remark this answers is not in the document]
yannickcr: I think this is a mandatory step, in the current PR we are missing a lot of information (see https://github.com/algolia/instantsearch.js/blob/2431b966e771d8d9386d919eb469c79b27b5b24a/src/widgets/pagination/pagination.js for example, only `totalPages` and `padding` are sent to the connector, we are missing `templates`, `showFirst`, `showLast`, etc.)

Also in most widget the parameters are send after destructuring, so with their default values.

src/middlewares/createTelemetryMiddleware.ts
Haroenv: that would require a flatMap or similar over the widgets (because of the index widget having more children), and thus seems like more code than currently. We can see payload as an opaque structure that seems logical to mutate as a whole

src/middlewares/createTelemetryMiddleware.ts
Haroenv: I see these options:

1. ignore non-connector parameters for now
2. pass the original widgetParams to makeWidget as $$widgetParams
3. pass each of the widget params to the connector (even though it does nothing)

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+  });
+}
+
+export function createTelemetryMiddleware(): Middleware {
+  const isTelemetryEnabled =
+    typeof window !== 'undefined' &&
+    window.navigator.userAgent.indexOf('Algolia Crawler') > -1;
+
+  if (!isTelemetryEnabled) {
```
francoischalifour, line 73: Can we rather not attach the middleware when this condition is met?

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  const payload: TelemetryPayload = {
+    widgets: [],
+  };
+
+  const payloadContainer = document.createElement('meta');
+  const refNode = document.querySelector('head')!;
+  payloadContainer.name = 'instantsearch:widgets';
+  refNode.appendChild(payloadContainer);
```
francoischalifour, line 88: DOM mutations should happen in `subscribe`.

src/middlewares/createTelemetryMiddleware.ts
francoischalifour: I'm not sure to understand. Can you explain it with a comment if appropriate? (Otherwise, explain it here with more context?)

src/types/widget.ts
```diff
@@ -363,7 +363,11 @@ export type WidgetRenderState<
 export type Widget<
   TWidgetOptions extends { renderState: unknown } = { renderState: unknown }
 > = {
+  /**
+   * identifier for official widgets
```
francoischalifour, line 367: We capitalize our JS doc everywhere else.

src/types/widget.ts
```diff
@@ -394,6 +398,12 @@ export type Widget<
     | 'ais.stats'
     | 'ais.toggleRefinement'
     | 'ais.voiceSearch';
+
+  /**
+   * private marker on widgets to differentiate a widget from a connector
+   */
+  $$official?: boolean;
```
francoischalifour, line 405: Why can't we base this on the `ais` namespace? (`$$type.startsWith('ais.')`).

It'd be nice not having to extend our APIs for this.

src/middlewares/createTelemetryMiddleware.ts
Haroenv: at an earlier time the widgets were getWidgetRenderState'ed using "results.getState", but that would fail if the results aren't yet available. Now I changed it to only request getWidgetRenderState using info that's given at init,and thus can't fail anymore @francoischalifour

src/middlewares/createTelemetryMiddleware.ts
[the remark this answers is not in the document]
Haroenv: > so with their default values

I've checked, and that's not really true. They're destructured, but the default value is applied in the connector

src/middlewares/createTelemetryMiddleware.ts
Haroenv: I wanted to keep the cod related to telemetry in this file, rather than adding conditions in InstantSearch

src/types/widget.ts
Haroenv: the goal is to differentiate between a widget and a connector. Do you see another way?

src/middlewares/createTelemetryMiddleware.ts
francoischalifour: We can export a `isTelemetryEnabled` function from this file and use it in `InstantSearch.ts` then.

src/types/widget.ts
Haroenv: going with $widgetType in a separate PR

src/lib/InstantSearch.ts
Haroenv: For now we aren't adding an option

src/middlewares/__tests__/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,149 @@
...
+  }))(window.navigator.userAgent)
+);
+
+const wait = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
+const defaultUserAgent =
+  'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1 Safari/605.1.15';
+const algoliaUserAgent = 'Algolia Crawler 5.3.2';
+
+describe('telemetry', () => {
```
francoischalifour, line 40:
```suggestion
describe('createTelemetryMiddleware', () => {
```

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,104 @@
...
+      }
+    }
+
+    // since we destructure in all widgets, the parameters with defaults are set to "undefined"
+    const params = Object.keys(widgetParams).filter(
+      key => widgetParams[key] !== undefined
+    );
+
+    payload.widgets.push({
+      type: widget.$$type || 'custom.widget',
```
francoischalifour, line 52: Why not default to `undefined`?

src/middlewares/createTelemetryMiddleware.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export function isTelemetryEnabled() {
+  return (
+    typeof window !== 'undefined' &&
+    window.navigator.userAgent.indexOf('Algolia Crawler') > -1
+  );
+}
+
+export function createTelemetryMiddleware(): Middleware {
```
francoischalifour, line 73: Is this name relevant given that Telemetry is not used in InstantSearch but in the crawler task?

francoischalifour (changes requested):
We don't test that the Telemetry middleware is added to the InstantSearch instance when the crawler agent is present.

src/middlewares/createTelemetryMiddleware.ts
Haroenv: you can't filter on undefined

francoischalifour: That would be the responsibility of our post-process script to assign a value to filter on with Algolia. We don't need to apply the constraint from the final system in InstantSearch itself.

src/types/widget.ts
Haroenv: https://github.com/algolia/instantsearch.js/pull/4624

src/middlewares/createMetadataMiddleware.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+export function isMetadataEnabled() {
+  return (
+    typeof window !== 'undefined' &&
+    window.navigator.userAgent.indexOf('Algolia Crawler') > -1
+  );
+}
+
+export function createMetadataMiddleware(): Middleware {
```
francoischalifour, line 73: "metadata" is not very descriptive. How about `createMetadataDomMiddleware`?

src/middlewares/createMetadataMiddleware.ts
```diff
@@ -0,0 +1,104 @@
+import { InstantSearch, Middleware, Widget } from '../types';
+import { Index } from '../widgets/index/index';
+
+type WidgetMetaData = {
+  type?: string;
```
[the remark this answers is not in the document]
Haroenv: it gets serialised to JSON, so there's not really a difference between those two

src/middlewares/createMetadataMiddleware.ts
Haroenv: I can't find a better description, dom seems redundant to me. It's creating a `meta` tag (short for metadata) and filling it. Since there's no other metadata in InstantSearch, this seems correct to me

PR #4605, opened 2020-12-09
fix: warn about invalid userToken

eunjae-lee:
**Summary**

This PR updates `algoliasearch-helper` to 3.3.4 to warn users about invalid userToken.

related: https://github.com/algolia/algoliasearch-helper-js/pull/802

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2871,10 +2871,10 @@ ajv@^6.1.0, ajv@^6.10.0, ajv@^6.10.2, ajv@^6.5.5:
     json-schema-traverse "^0.4.1"
     uri-js "^4.2.2"
 
-algoliasearch-helper@^3.3.3:
-  version "3.3.3"
-  resolved "https://registry.yarnpkg.com/algoliasearch-helper/-/algoliasearch-helper-3.3.3.tgz#8f1338790c3b478a8223f14ee447b9bdab4ab68b"
-  integrity sha512-1MKryf/yLQK9qgCaHtM+OBmG+R3qD6wxN8NnZstlCB8LijCZjoX1mgdema3+cBaa/zfmsD2q6/aP9kUKQmH4DQ==
+algoliasearch-helper@^3.3.4:
+  version "3.3.4"
+  resolved "https://registry.npmjs.org/algoliasearch-helper/-/algoliasearch-helper-3.3.4.tgz#4a3c56d42a2a81589d5722b73653b2deaf3e7064"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -2919,27 +2919,6 @@ algoliasearch@4.8.2:
     "@algolia/requester-node-http" "4.8.2"
     "@algolia/transporter" "4.8.2"
 
-algoliasearch@^3.35.1:
```
Haroenv, line 2922: unfortunately you need to keep this entry in the lock file, otherwise algoliasearch-v3 won't build anymore in storybook as far as I can tell (it doesn't deal properly with the aliased module + places.js having a dependency

eunjae-lee: Not sure why it got deleted.

yarn.lock
```diff
@@ -2871,14 +2871,14 @@ ajv@^6.1.0, ajv@^6.10.0, ajv@^6.10.2, ajv@^6.5.5:
...
-  integrity sha512-1MKryf/yLQK9qgCaHtM+OBmG+R3qD6wxN8NnZstlCB8LijCZjoX1mgdema3+cBaa/zfmsD2q6/aP9kUKQmH4DQ==
+algoliasearch-helper@^3.3.4:
+  version "3.3.4"
+  resolved "https://registry.npmjs.org/algoliasearch-helper/-/algoliasearch-helper-3.3.4.tgz#4a3c56d42a2a81589d5722b73653b2deaf3e7064"
+  integrity sha512-1Ts2XcgGdjGlDrp3v6zbY8VW+X9+jJ5rBmtPBmXOQLd4b5t/LpJlaBdxoAnlMfVFjywP7KSAdmyFUNNYVHDyRQ==
   dependencies:
     events "^1.1.1"
 
-"algoliasearch-v3@npm:algoliasearch@3.35.1":
+"algoliasearch-v3@npm:algoliasearch@3.35.1", algoliasearch@^3.35.1:
```
eunjae-lee, line 2881: @Haroenv This line has changed too. Do you think I should revert it?

Haroenv: yes, you should revert it too.

yarn.lock
Haroenv: Yarn thinks this is the right resolution (since it deduplicates), but some other tools try to use the node_modules folder name, which is `algoliasearch-v3` if it gets deduplicated

Haroenv (approved):

PR #4609, opened 2020-12-18
feat(index): expose scoped results getter

Haroenv:
**Summary**

like #4603, scoped results is also useful to be exposed for Vue InstantSearch & telemetry

**Result**

new function on index widgets: `getScopedResults` which is used in render for all widgets already.


eunjae-lee (approved):
looks good to me!

shortcuts (approved):

PR #4612, opened 2020-12-23
refactor: rename all references to widgetOptions as widgetParams

Haroenv:
The two terms were used interchangeably, mostly in comments and in function arguments. There's now only `TWidgetOptions` in the Widget generic, which isn't widgetParams

BREAKING CHANGE: if you're using experimental-typescript and importing a type of the form `...WidgetOptions` (eg. replace `HitsWidgetOptions` with `HitsWidgetParams`)


francoischalifour (approved):

eunjae-lee (approved):
Looks good to me

PR #4613, opened 2020-12-24
chore(configure): fix type

Haroenv:
**Summary**

the "widgetParams" passed to the connector / renderState are the ones that are passed to the connector, not the ones that are passed to the widget.

**Result**

this error only showed up in TS 4.1, but is still wrong in the older version of TS used here


src/widgets/configure/configure.ts
```diff
@@ -10,14 +9,14 @@ import { noop } from '../../lib/utils';
  * A list of [search parameters](https://www.algolia.com/doc/api-reference/search-api-parameters/)
  * to enable when the widget mounts.
  */
-export type ConfigureWidgetParams = PlainSearchParameters;
+export type ConfigureWidgetParams = ConfigureConnectorParams['searchParameters'];
```
shortcuts, line 12: Isn't it the same? As`ConfigureConnectorParams['searchParameters'] === PlainSearchParameters`, the naming makes more sense anyway tho

shortcuts (approved):

src/widgets/configure/configure.ts
Haroenv: Yes, it's the same, but if we add any comment to search parameters option, it will be visible here automatically

PR #4618, opened 2021-01-05
chore(escape-highlight): fix circular dependency

shortcuts:
**Summary**

There was a circular dependency in `escape-highlight.ts`

```
src/lib/utils/index.ts -> src/lib/utils/concatHighlightedParts.ts -> src/lib/escape-highlight.ts -> src/lib/utils/index.ts
```


src/lib/escape-highlight.ts
```diff
@@ -1,5 +1,6 @@
+import escape from '../lib/utils/escape';
```
Haroenv, line 1: instead escape-highlight should be put inside `utils` and the import from `./escape` I think (then all references to escape-highlight need to be moved too)

Haroenv (changes requested):
good idea, but I think this file should be in utils actually

src/lib/escape-highlight.ts
shortcuts: Everything should be fixed in https://github.com/algolia/instantsearch.js/pull/4618/commits/313d37f0f6540100857f884b395121e0dd2d73cf

shortcuts: False alarm, some files didn't throw errors during tests

shortcuts: Now fixed in https://github.com/algolia/instantsearch.js/pull/4618/commits/25110df6511569ee41bc6bae0d25087a68786efa

src/lib/utils/escape-highlight.ts
```diff
@@ -1,6 +1,6 @@
-import escape from '../lib/utils/escape';
-import isPlainObject from '../lib/utils/isPlainObject';
-import { Hit, FacetHit, EscapedHits } from '../types';
+import escape from './escape';
+import isPlainObject from './isPlainObject';
+import { Hit, FacetHit, EscapedHits } from '../../types';
```
Haroenv, line 3: does this file have tests? if so, they should also be moved 👍

Haroenv (commented):
I think this is a good change, but just realised moving a file is a potential breaking change, so would like a second opinion

src/lib/utils/escape-highlight.ts
shortcuts: I've moved it in the `utils/__tests__/` folder in https://github.com/algolia/instantsearch.js/pull/4618/commits/25110df6511569ee41bc6bae0d25087a68786efa

Haroenv (approved):
from my point of view, this is an acceptable breaking change, but a second opinion is appreciated

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -10,7 +10,7 @@ import {
 import { createSearchClient } from '../../../../test/mock/createSearchClient';
 import { createSingleSearchResponse } from '../../../../test/mock/createAPIResponse';
 import connectAutocomplete from '../connectAutocomplete';
-import { TAG_PLACEHOLDER } from '../../../lib/escape-highlight';
+import { TAG_PLACEHOLDER } from '../../../lib/utils/';
```
Haroenv, line 13:
```suggestion
import { TAG_PLACEHOLDER } from '../../../lib/utils';
```

src/connectors/hits/__tests__/connectHits-test.ts
```diff
@@ -2,7 +2,7 @@ import algoliasearchHelper, {
   SearchParameters,
   SearchResults,
 } from 'algoliasearch-helper';
-import { TAG_PLACEHOLDER } from '../../../lib/escape-highlight';
+import { TAG_PLACEHOLDER } from '../../../lib/utils/';
```
Haroenv, line 5:
```suggestion
import { TAG_PLACEHOLDER } from '../../../lib/utils';
```

src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -9,7 +9,7 @@ import {
   createRenderOptions,
 } from '../../../../test/mock/createWidget';
 import { createSingleSearchResponse } from '../../../../test/mock/createAPIResponse';
-import { TAG_PLACEHOLDER } from '../../../lib/escape-highlight';
+import { TAG_PLACEHOLDER } from '../../../lib/utils/';
```
Haroenv, line 12:
```suggestion
import { TAG_PLACEHOLDER } from '../../../lib/utils';
```

src/lib/utils/index.ts
```diff
@@ -31,6 +31,12 @@ export { default as mergeSearchParameters } from './mergeSearchParameters';
 export { default as resolveSearchParameters } from './resolveSearchParameters';
 export { default as toArray } from './toArray';
 export { warning, deprecate } from './logger';
+export {
+  default as escapeHits,
```
Haroenv, line 35:
```suggestion
  escapeHits,
```

we should make it a regular export, not a default one, since we're touching the file anyway now

shortcuts: Good idea! All done in https://github.com/algolia/instantsearch.js/pull/4618/commits/89280b5ba96293392bd5ee4ed6268ac5c6b04a0c

Haroenv (approved):

eunjae-lee:
> I think this is a good change, but just realised moving a file is a potential breaking change, so would like a second opinion

How come is it a breaking change? Do you mean in a case where user is directly importing `src/lib/escape-highlight.ts`? If that's what you mean, then I agree that it's acceptable and the PR seems good to me.

eunjae-lee (approved):

PR #4619, opened 2021-01-07
fix(insights): don't quote values

Haroenv:
**Summary**

The values for personalisation are meant to be identical to those of the engine, otherwise they show up twice

**Result**

Values to send to insights are no longer stringified


marchelbling (approved):
👍

francoischalifour (approved):

eunjae-lee (approved):

PR #4624, opened 2021-01-12
feat(widgets): annotate widget instances with $$widgetType

Haroenv:
Adds a new $$widgetType to all widgets, so they can be differentiated from connectors, as well as differentiating between widgets that share a connector.

This differentiation hasn't been done for rangeInput & rangeSlider yet

### DIFF
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "65.65 kB"
+      "maxSize": "65.75 kB"
     },
     {
       "path": "./dist/instantsearch.development.js",
--- a/package.json
+++ b/package.json
@@ -143,7 +143,7 @@
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "65.5 kB"
+      "maxSize": "65.65 kB"
     },
     {
       "path": "./dist/instantsearch.development.js",
--- a/src/lib/utils/typedObject.ts
+++ b/src/lib/utils/typedObject.ts
@@ -1,0 +1,20 @@
+// https://stackoverflow.com/a/65117465/3185307
+
+export const keys = Object.keys as <TObject extends {}>(
+  yourObject: TObject
+) => Array<keyof TObject>;
+
+export const values = Object.values as <TObject extends {}>(
+  yourObject: TObject
+) => Array<TObject[keyof TObject]>;
+
+type Entries<TObject> = {
+  [TKey in keyof TObject]: [TKey, TObject[TKey]];
+}[keyof TObject];
+export const entries = Object.entries as <TObject extends {}>(
+  yourObject: TObject
+) => Array<Entries<TObject>>;
+
+export const fromEntries = Object.fromEntries as <TKey extends string, TValue>(
+  yourObjectEntries: Array<[TKey, TValue]>
+) => Record<TKey, TValue>;
--- a/src/lib/utils/typedObject.ts
+++ b/src/lib/utils/typedObject.ts
@@ -1,5 +1,7 @@
-// https://stackoverflow.com/a/65117465/3185307
-
+/**
+ * A typed version of Object.keys, to use when looping over a static object
+ * inspired from https://stackoverflow.com/a/65117465/3185307
+ */
 export const keys = Object.keys as <TObject extends {}>(
   yourObject: TObject
 ) => Array<keyof TObject>;
--- a/src/lib/utils/typedObject.ts
+++ b/src/lib/utils/typedObject.ts
@@ -3,18 +3,3 @@
 export const keys = Object.keys as <TObject extends {}>(
   yourObject: TObject
 ) => Array<keyof TObject>;
-
-export const values = Object.values as <TObject extends {}>(
-  yourObject: TObject
-) => Array<TObject[keyof TObject]>;
-
-type Entries<TObject> = {
-  [TKey in keyof TObject]: [TKey, TObject[TKey]];
-}[keyof TObject];
-export const entries = Object.entries as <TObject extends {}>(
-  yourObject: TObject
-) => Array<Entries<TObject>>;
-
-export const fromEntries = Object.fromEntries as <TKey extends string, TValue>(
-  yourObjectEntries: Array<[TKey, TValue]>
-) => Record<TKey, TValue>;
--- a/src/middlewares/createMetadataMiddleware.ts
+++ b/src/middlewares/createMetadataMiddleware.ts
@@ -3,6 +3,7 @@ import { Index } from '../widgets/index/index';
 
 type WidgetMetaData = {
   type: string | undefined;
+  widgetType: string | undefined;
   params: string[];
 };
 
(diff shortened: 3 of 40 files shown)


src/lib/utils/typedObject.ts
```diff
@@ -0,0 +1,5 @@
+// https://stackoverflow.com/a/65117465/3185307
```
yannickcr, line 1: Could you add a more descriptive comment than just the link ? It would help to understand the purpose of this snippet without having to open the link.

Thinking about something like:

```js
/*
 * Typed `Object.keys` to return `Array<keyof TObject>` instead of `Array<string>`
 * source: https://stackoverflow.com/a/65117465/3185307
 */
```

src/middlewares/createMetadataMiddleware.ts
```diff
@@ -50,6 +51,7 @@ function extractPayload(
 
     payload.widgets.push({
       type: widget.$$type,
+      widgetType: widget.$$widgetType,
```
yannickcr, line 54: If I understand correctly `widget.$$type` is the connector type and `widget.$$widgetType` is the actual widget type, is that right ?

If that the case maybe it would be clearer to rename `type` to `connectorType` in the payload ?

Haroenv: widgets without a connector (index, analytics) technically set type too, which is why I didn't rename it to connectorType :)

package.json
```diff
@@ -143,7 +143,7 @@
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "65.5 kB"
+      "maxSize": "65.65 kB"
```
francoischalifour, line 146: We're used to increment with a 0.25KB step.

```suggestion
      "maxSize": "65.75 kB"
```

src/types/widget.ts
```diff
@@ -386,6 +386,41 @@ export type Widget<
     | 'ais.places'
     | 'ais.poweredBy'
     | 'ais.queryRules'
+    // @MAJOR: remove individual types for rangeSlider & rangeInput
```
francoischalifour, line 389: Why is it a breaking change to change these to `ais.range`?

https://github.com/algolia/instantsearch.js/blob/27b959cb3997a79a528c1fdd73a84ce1bb9a6f9f/src/widgets/range-slider/range-slider.js#L153

src/types/widget.ts
```diff
@@ -386,6 +386,41 @@ export type Widget<
...
+    | 'ais.infiniteHits'
+    | 'ais.menu'
+    | 'ais.menuSelect'
+    | 'ais.numericMenu'
+    | 'ais.pagination'
+    | 'ais.places'
+    | 'ais.poweredBy'
     | 'ais.queryRuleCustomData'
     | 'ais.queryRuleContext'
     | 'ais.range'
```
francoischalifour, line 426: `ais.range` is not a `$$widgetType`, is it?

src/widgets/__tests__/index.tests.ts
```diff
@@ -0,0 +1,172 @@
+import { PlacesInstance } from 'places.js';
+import * as widgets from '../';
+import { Widget } from '../../types';
+
+// @TODO: move to typedObject once this no longer needs to be polyfilled
```
francoischalifour, line 5: Not sure this needs to be a TODO since it depends on the language support and not on us. Wdyt?

src/widgets/__tests__/index.tests.ts
```diff
@@ -0,0 +1,172 @@
...
+  [TKey in keyof TObject]: [TKey, TObject[TKey]];
+}[keyof TObject];
+const entries = Object.entries as <TObject extends {}>(
+  yourObject: TObject
+) => Array<Entries<TObject>>;
+
+type Widgets = typeof widgets;
+type WidgetNames = keyof typeof widgets;
+
+function initiateAllWidgets(): Array<[WidgetNames, Widget]> {
```
francoischalifour, line 16: This has a maintenance cost and it'll be hard to detect if we miss to update this test when we introduce a new widget.

src/widgets/__tests__/index.tests.ts
```diff
@@ -0,0 +1,172 @@
...
+          'ais.',
+        ])
+      );
+    });
+
+    test('is the same as name', () => {
+      const widgetInstances = initiateAllWidgets();
+
+      widgetInstances.forEach(([name, widget]) => {
+        if (name === 'panel') {
```
francoischalifour, line 162: These automated tests are hard to read, to maintain and they quickly become conditional. We already cover most of those in each `{widget}-test.ts` files. Why changing the strategy for `$$widgetType`?

src/widgets/__tests__/index.tests.ts
Haroenv: if we miss to update this, the test will fail

src/widgets/__tests__/index.tests.ts
Haroenv: it's true, but some things should be true for all widgets regardless of what the name is. I found that this gave me more confidence that it is implemented correctly, for example that no typo happened between the test & implementation (which would likely also exist in the test, if it's written manually)

src/widgets/__tests__/index.tests.ts
Haroenv: what I meant is that this utility can't be in typedObject, since our polyfill test will fail if source files use object.entries, which is why I put the utility only in the test. I can remove the comment if you don't think it's important

eunjae-lee: Then you might change the comment from "TODO" to "It's not in the source but in this test because .."

src/widgets/__tests__/index.tests.ts
eunjae-lee: Whenever we adds a widget, we should check this, but might forget to do so (which will be caught by tests). I think it's acceptable to do this test for all widgets in this one place in the uniformed way.

src/middlewares/createMetadataMiddleware.ts
eunjae-lee: What if we clean up the types for widgets without a connector? Is it a good time to do so? No strong opinion, but genuinely asking.

src/types/widget.ts
Haroenv: not a breaking change, you're right, rather a todo. It required a significant change in the check ui state code to test both  type and widgetType, which is why I'm not doing it here yet. I can change it to another comment if you prefer?

francoischalifour: Yes, that's less misleading.

src/widgets/__tests__/index.tests.ts
francoischalifour: It's going to be a pain to maintain but as you wish.

eunjae-lee (approved):
Looks good to me!

src/types/widget.ts
Haroenv:
```suggestion
    // @TODO: remove individual types for rangeSlider & rangeInput once updating checkIndexUiState
```

PR #4627, opened 2021-01-15
fix(range): don't go out of bounds with min or max given

### LINKED ISSUE #4626: connectRange does not trigger an update when both range bounds are set to undefined
## 🐛 Bug description

I'm working on a `connectRange()` implementation with a custom slider library (noUiSlider) and have run into the issue that `connectRange` will not trigger an update when I call `refine([undefined, undefined])`. I have tracked down the code that ultimately causes this behaviour, which is [this `if`-statement][issue-source] in the `connectRange` source code.

I have also found that this issue **only occurs** when I have an override for the `onStateChange` method on the `InstantSearchOptions` object passed to the `instantsearch()` factory function. If I remove this override everything works as expected.
 
## 🔍 Bug reproduction

1. Create and register a `connnectRange()` widget
2. Have an override of `onStateChange()` in the configuration passed to the `instantsearch()` factory function, exactly as [the documentation][onstatechange-doc] shows:

    ```typescript
    {
        onStateChange({ uiState, setUiState })
        {
            setUiState(uiState)
        }
    }
    ```
3. Restrict one bound of the range (UI updates as expected)
4. Reset that bound to match the corresponding range min/max value. (UI does not update)

**Live reproduction:**

In this [live example][live-example] perform the following steps:
1. click down the max-value of the price number input, and watch the results count go down
2. click back up the max-value, then the results count should go back up, but it doesn't. Nor does the hits list update as expected.
3. click on any pagination link, and watch the results count update now. 

    _Because of the observation in step 3, it looks like it that the bounds are updated internally, because if I have any other interaction afterwards that triggers an update of the search results, the reset bounds do get noticed and my UI updates using the `undefined` bounds._

4. Comment (disable) the `onStateChange` override
5. Perform steps 1 to 3 again, and see everything work as expected.

## 💭 Expected behavior

The [refine() documentation][refine-docblock] states:

> You can use `undefined` to remove a previously set bound or to set an infinite bound.

This makes me expect that when the state is for instance `[10, undefined]` and then I call `refine([undefined, undefined])` that the lower bound `10` is being removed, and then a new search is triggered, regardless of other non-related options such as the `onStateChange` method.

## Environment

- OS: macOS
- Browser: Chrome - Version 87.0.4280.88 (Official Build) (x86_64)
- instantsearch.js version: `4.10.0-experimental-typescript.0`

[live-example]: https://codesandbox.io/s/agitated-bhaskara-s78pq
[onstatechange-doc]: https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/#widget-param-onstatechange
[refine-docblock]: https://github.com/algolia/instantsearch.js/blob/v4.11.0/src/connectors/range/connectRange.ts#L38
[issue-source]: https://github.com/algolia/instantsearch.js/blob/v4.11.0/src/connectors/range/connectRange.ts#L329

eunjae-lee:
**Summary**

fixes #4626

This PR fixes the bug where `connectRange` set the wrong range when `min` or `max` was given.

If you look at lines like [this](https://github.com/algolia/instantsearch.js/blob/fix/range/src/connectors/range/connectRange.ts#L152:L152), `connectRange` assumes there is only one value per operator (It's getting the first item from the array). However, in `getWidgetSearchParameters`, it adds refinements for min and max bound if given. After that, it again adds the bound from `uiState`.

It's reproduced in controlled mode (by calling `setUiState`).

**Result**

It should fixes the reproduction from #4626
The fixed example: https://codesandbox.io/s/confident-pasteur-ocls6?file=/src/app.js


eunjae-lee:
Wait, if the one from `uiState` is out of the bound, then it should keep things in the bound. I will add a new test and update the PR next week.

Haroenv:
This is indeed correct, but I'm not sure which one of the two should be taken as the source of truth. In theory there's nothing wrong with an out-of-bounds refinement, but then more of the connector is wrong.

eunjae-lee:
> This is indeed correct, but I'm not sure which one of the two should be taken as the source of truth. In theory there's nothing wrong with an out-of-bounds refinement, but then more of the connector is wrong.

Hmm I thought the min, max bound given to the widget should be the source of truth. In the latest change https://github.com/algolia/instantsearch.js/pull/4627/commits/404d29d95b22a765bb196dc649093a39d4dd124e, the out-of-bound refinements are capped with the bound.
What do you think?

Haroenv:
this makes sense I think, although you'll need to change the initial test to make sense

eunjae-lee:
> this makes sense I think, although you'll need to change the initial test to make sense

can you tell me what about the initial test needs to change?

yannickcr (approved):

src/connectors/range/__tests__/connectRange-test.ts
```diff
@@ -1941,6 +1942,54 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/range-input
       });
     });
 
+    test('returns the `SearchParamters` with only one min and one max', () => {
```
Haroenv, line 1945:
```suggestion
    test('returns the `SearchParameters` with only one min and one max', () => {
```

Haroenv: the title doesn't seem to be what it's testing

Haroenv (commented):
change is good, but I don't know what the first test is meant to assert

eunjae-lee:
> change is good, but I don't know what the first test is meant to assert

It might be unclear to you.

```js
      const widget = makeWidget({
        attribute: 'price',
        min: 0,
        max: 500,
      });

      const actual = widget.getWidgetSearchParameters!(helper.state, {
        uiState: {
          range: {
            price: '0:400',
          },
        },
      });

      console.log(actual.numericRefinements.price['<=']);
```

Before this PR, it used to print `[400, 500]`, instead of `[400]`.

src/connectors/range/__tests__/connectRange-test.ts
```diff
@@ -1941,6 +1942,54 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/range-input
       });
     });
 
+    test('returns the `SearchParameters` with only one min and one max', () => {
```
shortcuts, line 1945: Considering https://github.com/algolia/instantsearch.js/pull/4627#issuecomment-770747015, the description should maybe be something like `'returns the SearchParameters with the correct price range'` or something more specific

shortcuts (approved):

src/connectors/range/__tests__/connectRange-test.ts
eunjae-lee: sounds good https://github.com/algolia/instantsearch.js/pull/4627/commits/b9bc2f33163a4bbcad72ff861e7cd649147f1881

Haroenv (approved):

PR #4629, opened 2021-01-20
Create compressed-size.yml

Haroenv:
**Summary**

**Result**


Haroenv:
doesn't support node versions, and it seems bundlesize works again??

PR #4630, opened 2021-01-20
refactor(PoweredBy): very slightly smaller image

Haroenv:
**Summary**

I've tried out a couple different approaches, but couldn't significantly lower the size. 

**Result**

These small changes don't show up visually for me, but still are a little smaller


eunjae-lee (approved):
Kapture 2021-01-21 at 10 05 13

I see no difference either

shortcuts (approved):

PR #4634, opened 2021-01-25
fix(index): only set listeners on init once

Haroenv:
**Summary**

Without this, if init is called multiple times (like in Vue InstantSearch), multiple event listeners are attached, as well as the helper being recreated. This causes the references from the first init call to no longer being correct later in the instance.

Since we save the references in the very first call of getWidgetRenderState in most widgets (`if (!refine) refine =`), the helper reference needs to stay stable.

**Result**

This fixes https://github.com/algolia/vue-instantsearch/issues/911 by avoiding to have a different helper instance when init gets called multiple times.

### DIFF
--- a/src/widgets/index/index.ts
+++ b/src/widgets/index/index.ts
@@ -372,6 +372,12 @@ const index = (props: IndexProps): Index => {
     },
 
     init({ instantSearchInstance, parent, uiState }: IndexInitOptions) {
+      if (helper !== null) {
+        // helper is already initialized, therefore we do not need to set up
+        // any listeners
+        return;
+      }
+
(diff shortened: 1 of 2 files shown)


src/widgets/index/index.ts
```diff
@@ -416,7 +422,7 @@ const index = (props: IndexProps): Index => {
         return mainHelper.search();
       };
 
-      (helper as any).searchWithoutTriggeringOnStateChange = () => {
+      helper.searchWithoutTriggeringOnStateChange = () => {
```
Haroenv, line 425: this has been added to the helper types a while ago :)

francoischalifour (approved):

PR #4638, opened 2021-02-03
feat(queryRuleContext): allow to make refinements based on query

Haroenv:
**Summary**

this use case came up in a support ticket, they wanted to add a query rule when the query _did not_ match something. The engine only has positive matches, so negative matches need to be done frontend.

**Result**

Before that use case was implemented like this: https://codesandbox.io/s/dynamically-deactivate-rule-forked-wouf8?file=/src/app.js 

With this PR it can be implemented like this: https://codesandbox.io/s/dynamically-deactivate-rule-forked-h9w20?file=/src/app.js

(still a bit longer than I wanted, but that's because you can't transform inside trackedFilters, and enabling it would be a breaking change)

### DIFF
--- a/src/connectors/query-rules/connectQueryRules.ts
+++ b/src/connectors/query-rules/connectQueryRules.ts
@@ -74,7 +74,8 @@ function getRuleContextsFromTrackedFilters({
         // An empty object is technically not a `SearchResults` but `getRefinements`
         // only accesses properties, meaning it will not throw with an empty object.
         helper.lastResults || ({} as SearchResults),
-        sharedHelperState
+        sharedHelperState,
+        true
       )
         .filter(
           (refinement: InternalRefinement) => refinement.attribute === facetName
--- a/src/connectors/clear-refinements/connectClearRefinements.ts
+++ b/src/connectors/clear-refinements/connectClearRefinements.ts
@@ -195,7 +195,7 @@ function getAttributesToClear({
   excludedAttributes,
   transformItems,
 }): AttributesToClear {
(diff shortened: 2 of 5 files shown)


src/connectors/query-rules/connectQueryRules.ts
```diff
@@ -74,7 +74,8 @@ function getRuleContextsFromTrackedFilters({
         // An empty object is technically not a `SearchResults` but `getRefinements`
         // only accesses properties, meaning it will not throw with an empty object.
         helper.lastResults || ({} as SearchResults),
-        sharedHelperState
+        sharedHelperState,
+        true
```
eunjae-lee, line 78: `getRefinements` is not added in this PR, but what if we rename the third parameter to something like `includeQuery`?
https://github.com/algolia/instantsearch.js/blob/feat/rule-filters-query/src/lib/utils/getRefinements.ts#L96:L96

So far there were only two cases calling `getRefinements` to see which refinements to "clear". I guess that's how it ended up having that parameter name.

Of course, not a blocker of the PR.

eunjae-lee (approved):
Looks good to me!

src/connectors/query-rules/connectQueryRules.ts
Haroenv: yes, the parameter is called like that because of the old "clearsQuery" option for current refinements, but it's indeed confusing now that we don't use it anymore. Let's change it!

francoischalifour (approved):

PR #4640, opened 2021-02-03
Update babel from 7.8.3 to 7.8.7 to support Node.js 14 as development environment

tnir:
**Summary**

Updates babel from 7.8.3 to 7.8.7 to support Node.js 13.x+ (including 14.15+ LTS) as development environment.

**Result**

Closes #4639

PR #4647, opened 2021-02-09
feat(build): avoid overly strict babel helpers

### LINKED ISSUE #4639: Build fails on Node.js 14.15
## 🐛 Bug description

Build fails on Node.js 14.15

## 🔍 Bug reproduction

```
$ git clone ...
$ yarn
[...]
$ yarn build
yarn run v1.22.10
$ yarn run build:cjs && yarn run build:es && yarn run build:umd
$ rm -rf cjs && BABEL_ENV=cjs babel src --extensions '.js,.ts,.tsx' --out-dir cjs/ --ignore 'src/index.es.ts','**/__tests__','**/__mocks__' --quiet
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /workspace/instantsearch.js/node_modules/@babel/helper-compilation-targets/package.json
    at throwExportsNotFound (internal/modules/esm/resolve.js:290:9)
    at packageExportsResolve (internal/modules/esm/resolve.js:513:3)
    at resolveExports (internal/modules/cjs/loader.js:432:36)
    at Function.Module._findPath (internal/modules/cjs/loader.js:472:31)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:867:27)
    at Function.Module._load (internal/modules/cjs/loader.js:725:27)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/workspace/instantsearch.js/node_modules/@babel/preset-env/lib/debug.js:8:33)
    at Module._compile (internal/modules/cjs/loader.js:1063:30) {
  code: 'ERR_PACKAGE_PATH_NOT_EXPORTED'
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

```
$ node --version
v14.15.4
$ npm --version
7.5.2
```

Haroenv:
**Summary**

Make the bundlesize slightly smaller by optimising the babel options

**Result**

Since we don't use the cases presented in those plugins as far as I can see (spreading non-iterables, symbols...)

Impact:

- 1.4KB~ for everything loose
- .6KB~ for only the specified transforms

Closes #4639, #4640, #4644


babel.config.js
```diff
@@ -88,6 +90,14 @@ module.exports = api => {
         {
           modules,
           targets,
+          loose: true,
+          exclude: [
+            // not needed, we don't have Symbols
+            '@babel/plugin-transform-typeof-symbol',
+            // we are passing an option
+            '@babel/plugin-transform-classes',
+            '@babel/proposal-object-rest-spread',
```
Haroenv, line 99: do we pass loose: true for all plugins, or specifically for these two?

I'm in favour of passing loose: true generally

package.json
```diff
@@ -143,7 +143,7 @@
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "66 kB"
+      "maxSize": "66.15 kB"
```
Haroenv, line 146:
```suggestion
      "maxSize": "65 kB"
```

if we go for loose

Haroenv, line 146:
```suggestion
      "maxSize": "66 kB"
```

If we go for only the specified plugins

Haroenv:
need to investigate which of these Jest was relying on

babel.config.js
```diff
@@ -88,6 +90,14 @@ module.exports = api => {
         {
           modules,
           targets,
+          loose: !isTest,
```
francoischalifour, line 93: Why do we use a different `loose` value in test environment? That seems error prone.

francoischalifour (commented):
Overall, I'm not sure using `loose` (which is risky and unintuitive in some cases) is worth this small bundle size gain.

Haroenv:
do you think we should update babel with the non-loose (bigger) output as in #4644 then @francoischalifour ?

francoischalifour:
Yes—it's `0.15 kB` bigger but with an updated and stable toolchain.

PR #4648, opened 2021-02-10
feat(smartSort): add widget

eunjae-lee:
## Summary

This PR adds the `smartSort` widget.

When "Smart Sort" is enabled, the engine will not provide all the results, but a subset of relevant hits.
For example, when you search for "iPhone" and sort by low price, you probably get iPhone cables, because they're cheap.
However in the context of Smart Sort, you will get real iPhone products although the total number of hits will be much less than the former.

### How to enable Smart Sort?
1. You can configure your index to have `relevancyStrictness` value on Algolia dashboard

**How do we know the results is smartly sorted?**
The search response will contain `appliedRelevancyStrictness` property (optional). If it exists and greater than 00, it means it's smartly sorted. If it's 0 or doesn't exist, then it's traditional sort.

### What does this widget do?

```js
smartSort({ container: "#smart-sort" });
```

The widget will display a button saying either "See all results" or "See relevant results". It depends if the current results are smartly sorted or not.

If a user clicks the button because they want to see all the results regardless of smart sort, the widget will attach `relevancyStrictness=0` to the search parameter.

If clicked again, it will go back to the previous value (given by widgetParams, either a value or undefined).


src/components/SmartSort/SmartSort.tsx
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const SmartSort = ({
+  cssClasses,
+  templates,
+  isSmartSorted,
+  relevancyStrictness,
+  refine,
+}: SmartSortProps) => (
+  <button
```
Haroenv, line 25: shouldn't it be: 

```
container
  template (text)
  button
    template (button)
```

(let's add this widget to specs too)

src/connectors/smart-sort/connectSmartSort.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      getWidgetRenderState({ results, helper }) {
+        const { appliedRelevancyStrictness } = results || {};
+
+        return {
+          isSmartSorted:
+            typeof appliedRelevancyStrictness !== 'undefined' &&
+            appliedRelevancyStrictness > 0 &&
+            appliedRelevancyStrictness <= 100,
+          refine: relevancyStrictness => {
```
Haroenv, line 90: this function reference isn't stable, meaning it won't be the same between renders. You can set it once like in search box and other widgets

src/connectors/smart-sort/connectSmartSort.ts
```diff
@@ -0,0 +1,126 @@
...
+            helper
+              .setQueryParameter('relevancyStrictness', relevancyStrictness)
+              .search();
+          },
+          widgetParams,
+        };
+      },
+
+      getWidgetSearchParameters(state, { uiState }) {
+        return mergeSearchParameters(
```
Haroenv, line 100: no need to merge search parameters here, it won't work with relevancyStrictness

you can do setQueryParameter / getQueryParameter, no?

src/connectors/smart-sort/connectSmartSort.ts
```diff
@@ -0,0 +1,126 @@
...
+          new algoliasearchHelper.SearchParameters({
+            relevancyStrictness: widgetParams.relevancyStrictness,
```
Haroenv, line 102: is this meant to be the default strictness? 

I think it's meant to toggle between:

1. undefined (100 set by the engine)
2. user-provided value (default to 0)

eunjae-lee: I see two cases:

1. smartSort({ }); // no explicit relevancyStrictness
-> Clicking "See all results" set 0 to suppress it.
-> Clicking "See relevant results" set undefined to fallback to the value in the index configuration.

2. smartSort({ relevancyStrictness: xx });
-> Clicking "See all results" set 0 to suppress it.
-> Clicking "See relevant results" go back to the value in the widgetParams.

What do you think?

src/connectors/smart-sort/connectSmartSort.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4648/commits/09af204c52a8d328b63a9e09e8ed42872ae9eb19

src/connectors/smart-sort/connectSmartSort.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4648/commits/858f44ba4c3f5e4c350c38729031b0c7a25bf223

makes sense?

src/components/SmartSort/SmartSort.tsx
eunjae-lee: good suggestion 👍🏼 
https://github.com/algolia/instantsearch.js/pull/4648/commits/3bd77d6216129ef2c4c66054f461f8cc6fb9695e

https://github.com/algolia/instantsearch.js/pull/4648/commits/901fa5d5788bfdec3ae9c33ef1c1bc5992c119d9

src/connectors/smart-sort/connectSmartSort.ts
```diff
@@ -0,0 +1,130 @@
...
+          widgetParams,
+        };
+      },
+
+      getWidgetSearchParameters(state, { uiState }) {
+        return state.setQueryParameter(
+          'relevancyStrictness',
+          uiState.smartSort?.relevancyStrictness ??
+            state.relevancyStrictness ??
+            widgetParams.relevancyStrictness
```
Haroenv, line 111: I think by applying the relevancyStrictness by default, we are forcing relevancyStrictness on non-smart sort indices, does that have any undefined behaviour? 

I find it a little odd that this widget would apply the parameter when it shouldn't display.

Possibly this is because we're missing a condition:

normal index: display nothing
smart index (appliedRelevancyStrictness is set):
- appliedRelevancyStrictness === 0: isSmartSorted: false
- appliedRelevancyStrictness !== 0: isSmartSorted: true

what do you think?

src/connectors/smart-sort/connectSmartSort.ts
```diff
@@ -0,0 +1,130 @@
+import { Connector } from '../../types';
+import { createDocumentationMessageGenerator, noop } from '../../lib/utils';
+
+export type SmartSortConnectorParams = {
+  relevancyStrictness?: number;
```
Haroenv, line 5: I still think that having this default strictness could pose problems for detecting which state we are in. I think it should stay being set in the settings only

src/connectors/smart-sort/connectSmartSort.ts
eunjae-lee: I had a sync and we're good to remove `relevancyStrictness` from the widget parameter to the component altogether. I had a wrong understanding that it was a part of the spec.

eunjae-lee: In the updated version, I didn't remove the ability to apply relevancyStrictness altogether. I just removed the widget parameter. The component still gets `refine` function and call either `refine(0)` or `refine(undefined)` to toggle the status. https://github.com/algolia/instantsearch.js/pull/4648/commits/2f3409f316f62b53052b3aeeb1e3ee18a67c7573

src/connectors/smart-sort/connectSmartSort.ts
```diff
@@ -98,6 +82,7 @@ const connectSmartSort: SmartSortConnector = function connectSmartSort(
             typeof appliedRelevancyStrictness !== 'undefined' &&
             appliedRelevancyStrictness > 0 &&
             appliedRelevancyStrictness <= 100,
+          isVirtualReplica: typeof appliedRelevancyStrictness !== 'undefined',
```
Haroenv, line 85:
```suggestion
          isVirtualReplica:appliedRelevancyStrictness !== undefined,
```

you can do an assertion directly against undefined

src/connectors/smart-sort/connectSmartSort.ts
```diff
@@ -98,6 +82,7 @@ const connectSmartSort: SmartSortConnector = function connectSmartSort(
             typeof appliedRelevancyStrictness !== 'undefined' &&
             appliedRelevancyStrictness > 0 &&
             appliedRelevancyStrictness <= 100,
```
Haroenv:
```suggestion
            appliedRelevancyStrictness !== undefined &&
            appliedRelevancyStrictness > 0,
```

I don't see why this could be > 100

stories/smart-sort.stories.ts
```diff
@@ -0,0 +1,40 @@
...
+    }, searchOptions)
+  )
+  .add(
+    'explicit value',
+    withHits(({ search, container }) => {
+      search.addWidgets([
+        smartSort({
+          container,
+          templates: {
+            button: ({ isSmartSorted }) =>
```
Haroenv, line 34: no error when the text template isn't defined?

Haroenv: actually that makes sense, although I'd still use it in the demo here

stories/smart-sort.stories.ts
```diff
@@ -0,0 +1,40 @@
+import algoliasearch from 'algoliasearch/lite';
+import { storiesOf } from '@storybook/html';
+import { withHits } from '../.storybook/decorators';
+import smartSort from '../src/widgets/smart-sort/smart-sort';
+
+const searchOptions = {
+  appId: 'C7RIRJRYR9',
+  apiKey: '77af6d5ffb27caa5ff4937099fcb92e8',
+  indexName: 'test_Bestbuy_vr_price_asc',
+  searchClient: algoliasearch('C7RIRJRYR9', '77af6d5ffb27caa5ff4937099fcb92e8'),
```
Haroenv, line 10: searchClient gets created from appId, apiKey options iirc, no need to create the search client again since there's no special arguments to algoliasearch anymore

src/connectors/smart-sort/connectSmartSort.ts
eunjae-lee: AFAIK being greater than 100 is invalid. So I just made it explicit about it. Should we loosen it?

stories/smart-sort.stories.ts
eunjae-lee: You're right. I used to have some argument to override the host to a specific one, but forgot to delete this https://github.com/algolia/instantsearch.js/pull/4648/commits/f681b8778dc1f9f90c8c3e0c1fc17425c7b28269

src/connectors/smart-sort/connectSmartSort.ts
Haroenv: it's the applied relevancy strictness, the engine would throw the error when you save the settings, so it's impossible to receive this situation, no?

src/connectors/smart-sort/connectSmartSort.ts
```diff
@@ -0,0 +1,111 @@
...
+          };
+        }
+
+        const { appliedRelevancyStrictness } = results || {};
+
+        return {
+          isSmartSorted:
+            typeof appliedRelevancyStrictness !== 'undefined' &&
+            appliedRelevancyStrictness > 0 &&
+            appliedRelevancyStrictness <= 100,
```
Haroenv, line 84:
```suggestion
            typeof appliedRelevancyStrictness !== 'undefined' &&
            appliedRelevancyStrictness > 0,
```

I still think we don't need the <=100

Haroenv (commented):
🔥

Haroenv (approved):

shortcuts (approved):
🚀

src/widgets/smart-sort/smart-sort.tsx
```diff
@@ -0,0 +1,108 @@
...
+import defaultTemplates from './defaultTemplates';
+
+export type SmartSortCSSClasses = Partial<{
+  root: string;
+  text: string;
+  button: string;
+}>;
+
+export type SmartSortTemplates = Partial<{
+  text: string | (({ isSmartSorted }: { isSmartSorted: boolean }) => string);
```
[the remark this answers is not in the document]
eunjae-lee: Good catch https://github.com/algolia/instantsearch.js/pull/4648/commits/d54831c2c134251427a3837d2dc16c2967e2d8a9

PR #4649, opened 2021-02-11
feat(stats): apply nbSortedHits

eunjae-lee:
**Summary**

This PR updates `stats` widget to reflect `nbSortedHits` when `0 < appliedRelevancyStrictness <= 100`.

**Result**

- When smart sorted: `150 relevant results sorted out of 1,234 found in 42ms`
- Normally: `1 result found in 42ms`

storybook: https://deploy-preview-4649--instantsearchjs.netlify.app/stories/?path=/story/metadata-stats--with-sorted-hits

related: #4648


src/components/Stats/Stats.js
```diff
@@ -21,11 +23,14 @@ const Stats = ({
       rootTagName="span"
       rootProps={{ className: cssClasses.text }}
       data={{
+        hasManySortedHits: nbSortedHits > 1,
         hasManyResults: nbHits > 1,
         hasNoResults: nbHits === 0,
         hasOneResult: nbHits === 1,
```
Haroenv, line 29: we might need these conditions for smart sort too

src/widgets/stats/stats.js
```diff
@@ -42,6 +44,11 @@ const renderer = ({ containerNode, cssClasses, renderState, templates }) => (
       cssClasses={cssClasses}
       hitsPerPage={hitsPerPage}
       nbHits={nbHits}
+      nbSortedHits={nbSortedHits}
+      isSmartSorted={
+        typeof appliedRelevancyStrictness !== 'undefined' &&
+        appliedRelevancyStrictness > 0
+      }
```
Haroenv, line 51: this condition should be in the connector, and probably be more like `typeof helper.state.relevancyStrictness !== 'undefined'`, wdyt?

eunjae-lee: If it's `0`, it's not smartly sorted. There are two cases when we get `0` as `relevancyStrictness`

1. customers set it like that on dashboard (meaning they don't want smart sort on the index)
2. a user clicked "see all results" button from <SmartSort /> and InstantSearch attached `relevancyStrictness=0` to the search parameter to suppress the smart setting (even it was configured on dashboard).

Does it make sense?

Haroenv: I see, that means that you should probably remove the option to set a value for relevancyStrictness in the widget, so it always toggles between 0 and undefined, wdyt? 

still think this condition should be in the connector

eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4649/commits/bf2d1694d19f9eb4d928dbd9cbb83cb201de3200

src/components/Stats/Stats.js
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4649/commits/9ad4b886c6933bafad751fb4efa9aae7d7ed8588

src/connectors/stats/connectStats.js
```diff
@@ -108,6 +110,10 @@ export default function connectStats(renderFn, unmountFn = noop) {
       return {
         hitsPerPage: results.hitsPerPage,
         nbHits: results.nbHits,
+        nbSortedHits: results.nbSortedHits,
+        isSmartSorted:
+          typeof results.appliedRelevancyStrictness !== 'undefined' &&
```
Haroenv, line 115: if it's undefined it's not bigger than 0, so this first condition actually isn't needed I think

eunjae-lee: makes sense https://github.com/algolia/instantsearch.js/pull/4649/commits/a3b67610d49f025e8e032a2c7961cd32cb0883b1

eunjae-lee: I just remembered why I did it so. Even though it's in JS, but when in TS, it gives an error saying it can be undefined so I had to guard it too. So I just reverted the change to get back the undefined check

Haroenv (approved):
looks good, but let's release this once the feature is ready to be released stable (together with doc etc.) wdyt?

also update bundlesize

eunjae-lee:
> looks good, but let's release this once the feature is ready to be released stable (together with doc etc.) wdyt?
> 
> also update bundlesize

Yeah I forgot to mention it. I will merge this once `smartSort` is merged onto `master`.
And the bundlesize will be updated from that commit.

yannickcr (approved):

src/widgets/stats/defaultTemplates.js
```diff
@@ -1,5 +1,15 @@
 export default {
```
shortcuts, line 1: I don't really know how the templates work, but I think we should handle the case where `nbHits === nbSortedHits`, to avoid this kind of sentences `10,000 relevant results sorted out of 10,000 found in 1ms`

shortcuts (approved):

src/widgets/stats/defaultTemplates.js
Haroenv: yes, that's an important note! isSmartSorted should be false in that case IMO

shortcuts: I just handled it [like that](https://github.com/algolia/react-instantsearch/pull/3010/files#diff-7f595657982de401bfc86c6843f6e39d019e23ae53ff282e017303121a455a1bR63) in RIS, should I fix it before merging `smartSort`?

Haroenv: I think so, but @eunjae-lee and you have been closer to the implementation of this feature, wdyt?

shortcuts: If we think `isSmartSorted` reflect what the engine says, it should stay like that. But it make sense to consider that if the number of results are the same we consider it has `not smart sorted`. 🤔

shortcuts: Added here: https://github.com/algolia/react-instantsearch/pull/3009/commits/9fa94a9e4274839ae05026581a57d3611a323365, wdyt @eunjae-lee ?

Haroenv (changes requested):
would like a consistent value of isSmartSorted between the two :)

src/widgets/stats/defaultTemplates.js
eunjae-lee: I think `isSmartSorted` should stay as is. With `0 < appliedRelevancyStrictness`, it's indeed smartly sorted from the engine. It's just the result happen to be the same as normal sort. Giving another meaning to `isSmartSort` could be confusing (like "why is it false? I set the value on dashboard?")

However it'd be a good addition in detail to render proper text. So I'm thinking of the change only in stats. What do you think? @Haroenv @shortcuts

shortcuts: > Giving another meaning to isSmartSort could be confusing (like "why is it false? I set the value on dashboard?")

It could be confusing from a developer POV indeed, but I think it make sense to consider that if the hits are the same as the sorted hits, then it's not smart sorted. As the boolean isn't `isSmartSortConfigured` for example, I think we can take this liberty IMO

eunjae-lee: When it's configured but happens to have the same number of hits, if we follow your logic, we should display a button saying "See relevant results" because isSmartSort === false. And when clicking the button will `refine(undefined)`, which will do nothing (because it's already undefined). That's confusing, no?

shortcuts: True, this would add a lot of conditions to add to handle it!

shortcuts: Should we disable the button on the default template when the hits are matching then? Or do we let the user handle this case?

eunjae-lee: It will likely happen when there's empty query, etc. IMO we can release the smartSort widget as-is, and update only the default template in stats widget. If we get better understanding and learn patterns, we could add some logic to smartSort, but I feel it's too early to assume conditional behaviors.

shortcuts: Ok, I'll revert it to [`nbHits !== nbSortedHits`](https://github.com/algolia/react-instantsearch/pull/3010/files#diff-7f595657982de401bfc86c6843f6e39d019e23ae53ff282e017303121a455a1bR63) then!

eunjae-lee: but let's hear what @Haroenv thinks

eunjae-lee: @Haroenv not sure if I understand you correctly.
My suggestion was

With appliedRelevancyStrictness > 0, when nbSortedHits === nbHits:
1. Display the stat as `1000 hits in 10ms` (without smart sort mentioned)
2. Display the button as `See all results` (still as if it's a smart sort)

Which part do you disagree? Do you think we should display the button as `See relevant results` or not display the button at all?

Ant-hem: Chiming there! I don't have a *strong opinion* on that topic but I can help by giving you some context and pointers on the feature.

The idea behind the banner was to be "transparent" with the end-users as the algorithm can be perceived as a "black-box". That's why I am not really disturbed by the following:

> `10,000 relevant results sorted out of 10,000 found in 1ms`

That's just means all the available hits were selected and then sorted by the Algorithm. I don't see anything weird with it, just the "transparency" we initially wanted.

On top on that, relying on `nbHits` can be dangerous as this value can sometimes, be interpolated by the engine and thus being volatile and incorrect. See `exhaustiveNbHits` field in the response.

To finish, one rule of thumb is `relevancyStrictness=0` means that the algorithm is disabled and that no results selection/deletion is being done. It's *almost* (not exactly) equivalent to a hard sort. When `relevancyStrictness> 0`  the algorithm is enabled and it does select/remove results.

Hope that helps!

Haroenv: I think the button should be displayed, so for the context of SmartSort, isSmartSorted is true as long as appliedRelevancyStrictness is not 0, but in the stats widget I think it doesn't make sense to show "x relevant results shown out of x in yms", but it should be "x results found in yms", which correlates with isSmartSorted: false

Haroenv (approved):

src/widgets/stats/defaultTemplates.js
eunjae-lee: @Haroenv  agree with what you said, except that two `isSmartSorted` variables in connectSmartSort and connectStats will have different definitions (but the same name), so confusing.
So I renamed `isSmartSorted` to `areHitsSorted` in the context of stats, and added the condition to check `nbSortedHits !== nbHits`.
https://github.com/algolia/instantsearch.js/pull/4649/commits/3c24187e10bcf4907077203cb9233fa229b383a7

PR #4654, opened 2021-02-16
feat(ts): update to typescript 4

Haroenv:
- replace all ts-ignore by ts-expect-error
- remove ignores that were no longer needed
- make ignored imports work more efficiently
- update typescript-eslint (because it was needed)
- fix all errors marked by typescript-eslint
- algoliasearch v3 interop without eslint disables

this isn't a breaking change, since it only changes typescript, but worth considering


src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -244,7 +244,7 @@ function getRefinementsItems({
   includedAttributes,
   excludedAttributes,
 }: {
-  results: SearchResults;
+  results: SearchResults | Record<string, never>;
```
eunjae-lee, line 247: what is it for?

src/helpers/__tests__/highlight-test.ts
```diff
@@ -3,7 +3,6 @@ import highlight from '../highlight';
 const NONE = 'none' as const;
 const FULL = 'full' as const;
 
-/* eslint-disable @typescript-eslint/camelcase */
```
eunjae-lee, line 6: Can you explain which config allowed it not having this comment?

src/types/algoliasearch.ts
```diff
@@ -1,39 +1,27 @@
...
+import algoliasearch from 'algoliasearch';
+import * as AlgoliaSearch from 'algoliasearch';
+/** @ts-ignore */
+import * as ClientSearch from '@algolia/client-search';
+
+/** @ts-ignore */
+type SearchResponseV3<TObject> = AlgoliaSearch.Response<TObject>;
+/** @ts-ignore */
+type SearchResponseV4<TObject> = ClientSearch.SearchResponse<TObject>;
+
```
eunjae-lee, line 14: these are so much better!

eunjae-lee (approved):
looks good

src/helpers/__tests__/highlight-test.ts
Haroenv: the camelcase lint rule no longer exists

src/connectors/current-refinements/connectCurrentRefinements.ts
Haroenv: I changed the signature of getRefinementsItem to that it can be called with an empty object too (like how it's done in a couple places, like query rules and current refinements before the results are in)

src/lib/utils/getPropertyByPath.ts
```diff
@@ -1,4 +1,5 @@
 function getPropertyByPath(
+  // eslint-disable-next-line @typescript-eslint/ban-types
   object: object | undefined,
```
yannickcr, line 3: We could use `Record<string, any>` instead of `object` here.

src/widgets/query-rule-custom-data/__tests__/query-rule-custom-data-test.ts
```diff
@@ -142,7 +142,8 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/query-rule-
           default: expect.any(Function),
         });
         expect(
-          (templates.default as Function)({
+          // @ts-expect-error (TS did not understand the jest expectation narrowed the type)
+          templates.default({
```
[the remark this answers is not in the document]
Haroenv: yes, however the eslint config warns against that now, and it's only a test, so I preferred this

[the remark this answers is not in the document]
yannickcr: Oh? What warning did you get? I had none in my testing :thinking:

yannickcr (approved):
:+1:

PR #4655, opened 2021-02-17
fix(insights): don't reset page

Haroenv:
**Summary**

helper.setQueryParameter resets the page, unlike helper.state.setQueryParameter

Also added a new test for the correct state of the page in both cases of setQueryParameter.

see https://github.com/algolia/vue-instantsearch/issues/888 (doesn't fix that issue, but this issue is uncovered there)

**Result**

using insights middleware no longer resets the page

### DIFF
--- a/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -207,7 +207,7 @@ describe('insights', () => {
       expect(getUserToken()).toEqual('abc');
     });
 
-    it('applies userToken without resetting the page', () => {
+    it('applies userToken before subscribe() without resetting the page', () => {
       const {
         insightsClient,
         instantSearchInstance,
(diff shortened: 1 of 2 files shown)


src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -193,6 +207,23 @@ describe('insights', () => {
...
+      const {
+        insightsClient,
+        instantSearchInstance,
+        getUserToken,
+        helper,
+      } = createTestEnvironment();
+      const middleware = createInsightsMiddleware({
+        insightsClient,
+      })({ instantSearchInstance });
+      insightsClient('setUserToken', 'abc');
```
eunjae-lee, line 220: Not a big difference but what if this line is below `middleware.subscribe()`?

```js
      helper.setPage(100);
      middleware.subscribe();
      expect(helper.state.page).toBe(100); 
```

is already proved correct in the previous test, so it'd make sense to set user token after those, and see if it still doesn't change the page.

eunjae-lee (approved):
Looks good to me 🙌

shortcuts (approved):

PR #4659, opened 2021-02-19
poc: dynamic widgets

Haroenv:


stories/stack.stories.ts
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    const categoryContainer = stack.createContainer('categories');
+
+    categoryContainer.apply(
+      menu({
+        container: categoryContainer.container,
+        attribute: 'categories',
+      })
+    );
```
eunjae-lee, line 101: I don't remember which API we have decided on in the last RFC meeting, but what if

```js
stack.add(container => menu({ container, attribute: 'categories' }));
```

in `add()` we can create a container and set the object to `localWidgets` altogether.

Haroenv: how would you know what the attribute is in that case?

Haroenv: I've tried a similar option in https://github.com/algolia/instantsearch.js/pull/4659/commits/b785c1cb740ca421c36fd904144b84819ccde99c#diff-e1fb3ad37886947412250842890ce946afafddf6239f99be67cd710141bf6275

stories/stack-4.stories.ts
```diff
@@ -76,8 +77,9 @@ const connectDynamicWidgets: DynamicWidgetsConnector = function connectDynamicWi
       init(initOptions) {
         widgets.forEach(widget => {
           const attribute = getAttribute(widget, initOptions);
-          localWidgets.set(attribute, { widget, isMounted: false });
+          localWidgets.set(attribute, { widget, isMounted: true });
         });
+        initOptions.parent!.addWidgets(widgets);
```
eunjae-lee: Nice little addition for a good impact :)

Haroenv:
implementation in https://github.com/algolia/instantsearch.js/pull/4687

PR #4667, opened 2021-03-03
feat(insights): add hits and attributes to InsightsEvent

eunjae-lee:
**Summary**

This PR adds `hits` and `attributes` to InsightsEvent.

## Detail

Previously we had this:
```
export type InsightsEvent = {
  insightsMethod?: InsightsClientMethod;
  payload: any;
  widgetType: string;
  eventType: string;
};
```

For example,

- insightsMethod: "viewedObjectIDs"
- widgetType: "ais.hits"
- eventType: "view"

`window.aa(insightsMethod, payload);` just worked, which means `payload` is a tailored object for Algolia Insights.

When users want to send events to third party trackers, they might need more information other than what is in the `payload`.

In case of `Hits Viewed`, `payload` contains `eventName`, `index` and `objectIDs`. When they need to send events to, for example, Segment, they need other data like position, product name, etc.

So we're passing the whole `hits` object inside the `InsightsEvent`.

In case of facet widgets, we're adding `attribute` to `InsightsEvent`. It's to easily filter out events like:

```
onEvent: (event, aa) {
  if (event.attribute === 'price') {
    // do not track events for `price` attribute
  }
  aa(event.insightsMethod, event.payload);
}
```


src/lib/utils/createSendEventForHits.ts
```diff
@@ -60,7 +60,8 @@ const buildPayload: BuildPayload = ({
       return null;
     }
   }
-  const hitsArray = Array.isArray(hits) ? hits : [hits];
+  // `Array.from(hits)` is to remove `hits.__escaped`
+  const hitsArray = Array.isArray(hits) ? Array.from(hits) : [hits];
```
Haroenv, line 64: Array.from isn't compatible with IE

eunjae-lee: ah right. https://github.com/algolia/instantsearch.js/pull/4667/commits/d3868b519395a21ca95d9b239206d8265ba0669e

Haroenv: you should instead remove the property if you don't want it to show up, or map over it

eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4667/commits/3b541f54c5ea3d8a41f6f5e9b3fac18a6a7f8530 WDTY?

Haroenv (approved):
I see why it's not in payload, since that would cause the default implementation to need to strip them, but I kinda feel like it fits better under payload or metadata than as top-level key

Haroenv:
(looks like some tests have failed due to the delete instead of spreading)

eunjae-lee:
> I see why it's not in payload, since that would cause the default implementation to need to strip them, but I kinda feel like it fits better under payload or metadata than as top-level key

you're right. `attribute` could probably be there as a top-level key, but `hits` doesn't really fit much. Do you think it makes sense to have a key like `metadata` and have `attribute` or `hits` under it?

PR #4670, opened 2021-03-05
fix(bindEvent): escape payload correctly

eunjae-lee:
## Summary

This PR updates the way we serialize & deserialize payloads for `bindEvent`. It's internal and users never got to deserialize by themselves, so it's not a breaking change.

## Details

Since ["add hits and attributes to InsightsEvent" PR](https://github.com/algolia/instantsearch.js/pull/4667), `bindEvent` method now serializes `hits` as well. And `hits` can contain special characters unlike the payloads that we used to have in insights middleware.

For example, 

```
btoa('™')
> Uncaught DOMException: String contains an invalid character
```

but
```
btoa(encodeURIComponent('™'))
> "JUUyJTg0JUEy"

decodeURIComponent(atob("JUUyJTg0JUEy"))
> "™"
```

## A minor concern

It's actually not about this PR, but about the previous ["add hits and attributes to InsightsEvent" PR](https://github.com/algolia/instantsearch.js/pull/4667).
Now that we expose `hits` to `InsightsEvent`, its payload is much bigger than before. It's going to bloat the DOM output. 
When they have templates like

```js
<button
  type="button"
  ${bindEvent('click', hit, 'Favorite')}
>
 ...
</button>
```

If they don't pass the whole `hit` object, and pass only a subset of it, it will reduce the size. What do you think?

### DIFF
--- a/src/connectors/hits/__tests__/connectHits-test.ts
+++ b/src/connectors/hits/__tests__/connectHits-test.ts
@@ -2,7 +2,7 @@ import algoliasearchHelper, {
   SearchParameters,
   SearchResults,
 } from 'algoliasearch-helper';
-import { TAG_PLACEHOLDER } from '../../../lib/utils';
+import { TAG_PLACEHOLDER, deserializePayload } from '../../../lib/utils';
 import connectHits from '../connectHits';
 import {
   createInitOptions,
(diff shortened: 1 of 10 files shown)


Haroenv (approved):
reasonable fix!

eunjae-lee:
> reasonable fix!

You also think the `A minor concern` part is okay?

Haroenv:
it's probably not really an issue IMO, although a subset was already set with objectIDs, which you said was insufficient for segment

eunjae-lee:
> it's probably not really an issue IMO, although a subset was already set with objectIDs, which you said was insufficient for segment

What I meant by a subset was to pass an object made from `hit` with some properties omitted like, for example, `image_url, description, ...` that won't be used in the `insights` middleware.

It could bloat the html size and might be an issue in case of SSR. Even if so, they can omit some properties to avoid the issue.

PR #4671, opened 2021-03-08
chore(deps): update Answers type (bump algoliasearch to 4.8.6)

eunjae-lee:
**Summary**

This PR updates the dependency `algoliasearch` to 4.8.6, which updates the type for Answers widget.

related: https://github.com/algolia/algoliasearch-client-javascript/pull/1258

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,109 +2,109 @@
 # yarn lockfile v1
 
 
-"@algolia/cache-browser-local-storage@4.8.5":
-  version "4.8.5"
-  resolved "https://registry.yarnpkg.com/@algolia/cache-browser-local-storage/-/cache-browser-local-storage-4.8.5.tgz#3eb10758c794d3cc8fc4e9f09e339d5b9589dc9c"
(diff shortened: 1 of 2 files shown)


Haroenv (approved):
not wrong, but won't really have client impact :)

eunjae-lee:
> not wrong, but won't really have client impact :)

yeah just a little change in type def.

shortcuts (approved):

yarn.lock
```diff
@@ -3382,7 +3382,8 @@ algoliasearch-helper@^3.4.4:
   dependencies:
     events "^1.1.1"
 
-"algoliasearch-v3@npm:algoliasearch@3.35.1", algoliasearch@^3.35.1:
+"algoliasearch-v3@npm:algoliasearch@3.35.1":
+  name algoliasearch-v3
```
Haroenv: hmm, this change still will make places fail, I think yarn of a certain version always tries to break the lock file to have only one entry for ones that don't share a name

eunjae-lee: Yeah I reverted that part. It's gonna keep bugging us until we figure out which version of yarn won't do this, or another way to avoid this.

PR #4673, opened 2021-03-09
feat(middleware): accept partial methods

eunjae-lee:
**Summary**

This PR allows middlewares to have partial methods. A middleware had to contain all of `subscribe, unsubscribe, onStateChange` methods, which isn't necessary.

**Result**

It works the same as before except that the following is possible now:

```js
search.use(() => ({
  onStateChange({ uiState }) {
    // ...
  }
});
```


src/lib/InstantSearch.ts
```diff
@@ -276,7 +276,7 @@ See ${createDocumentationLink({
     // middleware so they're notified of changes.
     if (this.started) {
       newMiddlewareList.forEach(m => {
-        m.subscribe();
+        m.subscribe?.();
```
Haroenv, line 279: what about setting default noops in https://github.com/algolia/instantsearch.js/pull/4673/files#diff-3c46288a69cb2289a7369cff0f629366e99041bc5f60c586cdc29322130f9f84R272 instead? That way less code is conditional

eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4673/commits/02a01bec27e595c2dc8811a5acf75a34c78e9d02

I have a question.

1. `PartialMiddleware` and `Middleware`
vs
2. `Middleware` and `InternalMiddleware`

Is it okay to have the type of the public one having `Partial` or whatever prefix? Or should we rename it to `Middleware` and rename the internal one to something else?

Haroenv (approved):
I think this makes sense, but I remember that @francoischalifour originally wanted middlewares to not have optional keys. I wonder if this has changed?

src/lib/InstantSearch.ts
Haroenv: Middleware should probably be the public one (thus with optionals) I think

shortcuts (approved):
Make sense!

francoischalifour:
> I think this makes sense, but I remember that @francoischalifour originally wanted middlewares to not have optional keys. I wonder if this has changed?

The original reason was that going from optional to required is a breaking change, while going the opposite isn't.

Besides, I tend to avoid fully optional structures—now a middleware can be an empty object.

eunjae-lee:
> The original reason was that going from optional to required is a breaking change, while going the opposite isn't.
→ makes sense

> Besides, I tend to avoid fully optional structures—now a middleware can be an empty object.
→ do we want to avoid it?

eunjae-lee:
@francoischalifour

francoischalifour:
> → do we want to avoid it?

That's not easy to answer. It'll add overhead to both our code and the middleware usage with conditional requirements. We have the same problem with the widget definition where nothing is required, but you still have to provide either `init` or `render`. You can't type-check that. That's why I don't think that this PR is necessary personally.

eunjae-lee:
> That's not easy to answer. It'll add overhead to both our code and the middleware usage with conditional requirements. We have the same problem with the widget definition where nothing is required, but you still have to provide either `init` or `render`. You can't type-check that. That's why I don't think that this PR is necessary personally.

This PR adds `noop` upfront, so I don't see it as an overhead. We can show a warning message if it's an empty object. And if we want, we can do it like [this](https://www.typescriptlang.org/play?#code/C4TwDgpgBAsglgEwQGwgdwIYCcIBEIBmcAdnMHAPbFQC8UA3gFBRRUDKwGwEAwgBYZiAcwgAKCmHJUAzgC4ogkAEp5ANwqIA3MyjSArgCNpAYyxwDYlVHVade4vqOnzltRoTaAvtsahIUAAVscgxkeCRkTBx8IlIpajogrBDkAB5wlHRsPEISMkpiAD4oADIdUSYWFnZObn5BEXkM1Cic2PyqAG0AchquXgFhCG6AXW0WTyUoAB8oCp0WRxMzCybETNaYvPiepecLUfGoSZm5yqr7PZWINYis6Ny4gp7Lw2WXUZ1JxkZjGWAFPIkilmptHh0EgxvD8-g4AQYgcE4KFQdktk8qLQGDori5RFN6MdGJ4Yf8oMZEclkWF1pE0eD4ljzq8nNd8QwiSTfmSEJSQbSwe1GXRzn06oMROzCSSubDpACboEkSiBfShQUmTi3vtLBzPAAaOwObVsgnHQ3VYgcfr1IZSzlAA), although it's not the cleanest.

Haroenv:
https://www.typescriptlang.org/play?#code/C4TwDgpgBAsglgEwQGwgdwIYCcIBEIBmcAdnMHAPbFQC8UA3gFBRRUDKwGwEAwgBYZiAcwgAKCmHJUAzgC4ogkAEp5ANwqIA3MyjSArgCNpAYyxwDYlVHVade4vqOnzltRoTaAvtsahIUAEFgABkIDGlgAHliCAAeABUAGigAVVoGAG0AaSgSKABrCBAKAih4gF15AAU4Y3yE5KyAPihPFroq7HIMZASWgDJUjMLi0pTyxl9waE6sbuR4JGRMHHwiUilqOiDQ8KiY2MWUdGw8QhIySmImyeMZYAVqrrgeo+XTtYvN9PpvW-uoAYnnMXgtECgVmd1pcqD8dI4TGYLKIlAxWoxPP8HA9jMD5m9IZ8Nlc4Sx7AjnMjUfR0ZjGHdsVAEHjQQSPudibC6EwWOxONx+IIRCi0Zi6QyIlAICzXuD3qsOTCtgx4YZES4RTTPIk7A41ZTLKKdbziBwuLwBMJDVqMUA is another option to allow partial but not empty objects (source: https://stackoverflow.com/a/48244432/3185307)

eunjae-lee:
> https://www.typescriptlang.org/play?#code/C4TwDgpgBAsglgEwQGwgdwIYCcIBEIBmcAdnMHAPbFQC8UA3gFBRRUDKwGwEAwgBYZiAcwgAKCmHJUAzgC4ogkAEp5ANwqIA3MyjSArgCNpAYyxwDYlVHVade4vqOnzltRoTaAvtsahIUAEFgABkIDGlgAHliCAAeABUAGigAVVoGAG0AaSgSKABrCBAKAih4gF15AAU4Y3yE5KyAPihPFroq7HIMZASWgDJUjMLi0pTyxl9waE6sbuR4JGRMHHwiUilqOiDQ8KiY2MWUdGw8QhIySmImyeMZYAVqrrgeo+XTtYvN9PpvW-uoAYnnMXgtECgVmd1pcqD8dI4TGYLKIlAxWoxPP8HA9jMD5m9IZ8Nlc4Sx7AjnMjUfR0ZjGHdsVAEHjQQSPudibC6EwWOxONx+IIRCi0Zi6QyIlAICzXuD3qsOTCtgx4YZES4RTTPIk7A41ZTLKKdbziBwuLwBMJDVqMUA is another option to allow partial but not empty objects (source: https://stackoverflow.com/a/48244432/3185307)

That's a nice way!

Haroenv (approved):
I personally think with the "at least one", the downside of empty objects is alleviated, as well as it being instantiated with noops, so no new conditions need to be added. However we need to have a consensus on this as a team

eunjae-lee:
> I personally think with the "at least one", the downside of empty objects is alleviated, as well as it being instantiated with noops, so no new conditions need to be added. However we need to have a consensus on this as a team

What do you mean by "no new conditions"?

Haroenv:
conditions when using middleware, as it was needed in the initial version of the pr

PR #4677, opened 2021-03-16
test(connectNumericMenu): avoid implicit any

Haroenv:
We need to declare the type to jest.fn() so that the return type of getInitializedWidget doesn't get tainted and becomes any[]

This may need to be done in other widgets too


Haroenv:
the other exclamation marks are added because previously `widget` was any

shortcuts (approved):
Good catch! :D

PR #4678, opened 2021-03-17
chore: manage deprecated code

eunjae-lee:
**Summary**

This PR introduces `__KEEP_DEPRECATION__` to manage deprecated code better.
It's not meant to be merged quickly. It's more like a PoC for discussion. The namings aren't fixed either.

It's hard to manage the deprecated code, and get rid of them when it's time to do so.
In this PR, there is a new global variable `__KEEP_DEPRECATION__` which is `true` by default.
We can wrap the deprecated code with `if (__KEEP_DEPRECATION__) { ... }`. Running `yarn build` will transform it into `if (true) { ... }`.

If we create a branch like `next` for the next major version, there we can run `KEEP_DEPRECATION=false yarn build` to make it `if (false) { ... }` and get the dead code eliminated in the build.

Without this kind of way, we will have too much difference between `master` and `next` because we'd want to remove the deprecated code there. With this, we can still keep the same code but with the flag turned on, and safely test the new paths. Also we can actually remove the code once `next` is merged back to `master` or a little later.

---

Or, we could introduce a variable like `__NEXT_VERSION__` and wrap the deprecated code with `if (!__NEXT_VERSION__) { ... }`. This way, we can also write the code that will be introduced only in the next version.

What do you think?

- Edit1: By the way, I'm not suggesting to change the current ones we already wrote. It will be too much.
- Edit2: It cannot cover everything. We will still have imports, dependencies, etc that cannot be conditionally done with this.

### DIFF
--- a/src/lib/insights/listener.tsx
+++ b/src/lib/insights/listener.tsx
@@ -1,7 +1,7 @@
 /** @jsx h */
 
 import { h } from 'preact';
-import { deserializePayload } from '../utils';
+import { deserializePayload, deprecated } from '../utils';
 import { readDataAttributes, hasDataAttributes } from '../../helpers/insights';
 import { InsightsClientWrapper } from '../../types';
 import { InsightsEvent } from '../../middlewares/createInsightsMiddleware';
@@ -63,7 +63,7 @@ const insightsListener = (BaseComponent: any) => {
         }
       }
 
-      if (__KEEP_DEPRECATION__) {
+      deprecated(() => {
         // old way, e.g. instantsearch.insights("clickedObjectIDsAfterSearch", { .. })
         const insightsTarget = findInsightsTarget(
           event.target as HTMLElement | null,
@@ -74,7 +74,7 @@ const insightsListener = (BaseComponent: any) => {
           const { method, payload } = readDataAttributes(insightsTarget);
           props.insights(method, payload);
         }
-      }
+      });
     };
 
     return (
--- a/src/lib/insights/listener.tsx
+++ b/src/lib/insights/listener.tsx
@@ -63,15 +63,17 @@ const insightsListener = (BaseComponent: any) => {
         }
       }
 
-      // old way, e.g. instantsearch.insights("clickedObjectIDsAfterSearch", { .. })
-      const insightsTarget = findInsightsTarget(
-        event.target as HTMLElement | null,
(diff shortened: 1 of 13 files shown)


yarn.lock
```diff
@@ -3409,29 +3409,7 @@ algoliasearch-helper@^3.4.4:
   dependencies:
     events "^1.1.1"
 
-"algoliasearch-v3@npm:algoliasearch@3.35.1":
```
Haroenv, line 3412: this needs to be reverted again 📦

src/lib/insights/listener.tsx
```diff
@@ -63,15 +63,17 @@ const insightsListener = (BaseComponent: any) => {
...
-      // old way, e.g. instantsearch.insights("clickedObjectIDsAfterSearch", { .. })
-      const insightsTarget = findInsightsTarget(
-        event.target as HTMLElement | null,
-        event.currentTarget as HTMLElement | null,
-        element => hasDataAttributes(element)
-      );
-      if (insightsTarget) {
-        const { method, payload } = readDataAttributes(insightsTarget);
-        props.insights(method, payload);
+      if (__KEEP_DEPRECATION__) {
```
Haroenv, line 66: What do you see as the advantages of this vs. a comment // @MAJOR? 

I see how we can then expose a build without deprecated code, but will that difference be significant enough to justify the change?

eunjae-lee: In my opinion, it will enable us to test(unit, e2e) with different builds.

By the way, I'm not suggesting to change the current ones we already wrote. It will be too much.

Haroenv: I wonder actually if it might not be better to wrap this code in a conditional similar to `warn`, but specifically for deprecated code. That way we force deprecated code that's still called to have a warning, and could strip out all deprecated code in a similar fashion as code after a warning

eunjae-lee: that's a nicer way

major: > What do you see as the advantages of this vs. a comment // @major?

Thank you for tagging me, but I'm not sure I can be much help here. I wish you the best, though! 🤗

eunjae-lee:
I'm not 100% convinced, so closing this for now.

PR #4679, opened 2021-03-17
fix(createURL): correctly remove page in state

Haroenv:
**Summary**

createURL didn't reset the page because it uses SearchParameter methods vs. helper methods which get the page reset automatically.

**Result**

fixes https://github.com/algolia/vue-instantsearch/issues/932

all createURLs now reset the page as expected

### DIFF
--- a/src/connectors/breadcrumb/connectBreadcrumb.ts
+++ b/src/connectors/breadcrumb/connectBreadcrumb.ts
@@ -167,23 +167,21 @@ const connectBreadcrumb: BreadcrumbConnector = function connectBreadcrumb(
         if (!connectorState.createURL) {
           connectorState.createURL = facetValue => {
             if (!facetValue) {
-              const breadcrumb = helper.getHierarchicalFacetBreadcrumb(
+              const breadcrumb = state.getHierarchicalFacetBreadcrumb(
                 hierarchicalFacetName
               );
               if (breadcrumb.length > 0) {
                 return createURL(
-                  helper.state.toggleFacetRefinement(
-                    hierarchicalFacetName,
-                    breadcrumb[0]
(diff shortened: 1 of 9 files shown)


Haroenv:
In the future we should rely on getWidgetSearchParameters IMO, although at the moment that doesn't reset the page either (otherwise it doesn't work when pagination or infinite hits isn't the last widget). Maybe `isPageReset` could be useful somehow?

jonathan-bird:
@Haroenv That was pretty much my approach too when I created my hacky workaround but thinking about it more, you want pretty much every facet in the URL except pagination so I think a "reset pagination" type method is the best way to handle it.

eunjae-lee:
> In the future we should rely on getWidgetSearchParameters IMO, although at the moment that doesn't reset the page either (otherwise it doesn't work when pagination or infinite hits isn't the last widget). Maybe `isPageReset` could be useful somehow? Or call `this.getWidgetSearchParameters().resetPage()` in refine and createURL, which could make sense too?

While this PR looks good to me, could you rephrase what you said?

Haroenv:
If you check the implementations of getWidgetSearchParameters, you'll see that none of them reset the page. That's because the actual page reset logic is done in the index widget's change listener on the helper.

That means that even if we use getWidgetSearchParameters for createURL and refine, we still would have to reset the page after retrieving the state.

It could be quite interesting longer-term to use a createURL based on UIState alone though, as well as relying on uiState alone for refine. I think it should be possible in theory, but will have breaking changes for non-builtin widgets relying only on the helper state. Maybe in a major version?

eunjae-lee (approved):
Thanks for the explanation. This PR looks good to me.

PR #4680, opened 2021-03-17
fix(utils): circular dependency in createSendEventForHits

shortcuts:
**Summary**

Fix circular dependency in createSendEventForHits

```
src/lib/utils/index.ts -> src/lib/utils/createSendEventForHits.ts -> src/lib/utils/index.ts
```

### DIFF
--- a/src/lib/utils/createSendEventForHits.ts
+++ b/src/lib/utils/createSendEventForHits.ts
@@ -1,5 +1,5 @@
 import { InstantSearch, Hit, Hits, EscapedHits } from '../../types';
-import { serializePayload } from '../../lib/utils';
+import { serializePayload } from '../../lib/utils/serializer';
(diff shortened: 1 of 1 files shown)


Haroenv (approved):
indeed the only one left. Do you think we could add a test for this?

Haroenv:
(careful when merging that the commit message is correct)

shortcuts:
> indeed the only one left. Do you think we could add a test for this?

@Haroenv As our build warns us, couldn't we make it throw if there's one?

Haroenv:
We could use `onwarn` to throw certain warnings as errors? https://rollupjs.org/guide/en/#onwarn

shortcuts:
> We could use `onwarn` to throw certain warnings as errors? https://rollupjs.org/guide/en/#onwarn

Done in https://github.com/algolia/instantsearch.js/pull/4682

PR #4681, opened 2021-03-18
feat(ts): convert stats, connectStats

shortcuts:
**Summary**

- Convert `stats`, `connectStats` to TypeScript
- Replace snapshots to inlined ones


src/components/Stats/Stats.tsx
```diff
@@ -0,0 +1,53 @@
...
+  processingTimeMS: number;
+  query: string;
+};
+
+const Stats = ({
+  nbHits,
+  nbSortedHits,
+  cssClasses,
+  templateProps,
+  ...rest
```
shortcuts, line 29: I've used `...rest` to avoid re-writing all the props, should it stay as before?

src/components/Stats/__tests__/Stats-test.tsx
```diff
@@ -14,7 +16,12 @@ describe('Stats', () => {
 
   it('should render <Template data= />', () => {
     const wrapper = mount(
-      <Stats {...getProps()} templateProps={{ templates: defaultTemplates }} />
+      (
+        <Stats
+          {...getProps()}
+          templateProps={{ templates: defaultTemplates }}
+        />
+      ) as ReactElementLike
```
shortcuts, line 24: Casting it as `ReactElementLike` because we return an `Element`, not sure that this is the best way to do it though.

francoischalifour: We can convert the test to the Testing Library. We want to move away from Enzyme.

Haroenv: Since we want to migrate later, it's good that you first made the existing tests passing. I don't have a big opinion on whether it stays like this now or gets migrated

src/widgets/stats/stats.tsx
```diff
@@ -0,0 +1,165 @@
...
+   * CSS class to add to the text span element.
+   */
+  text: string | string[];
+};
+
+export type StatsTemplates = {
+  /**
+   * Text template, provided with `hasManyResults`, `hasNoResults`, `hasOneResult`, `hitsPerPage`, `nbHits`, `nbSortedHits`, `nbPages`, `page`, `processingTimeMS`, `query`.
+   */
+  text: Template;
```
Haroenv, line 37:
```suggestion
  text: Template<{hasManuResults.....>;
```

src/widgets/stats/defaultTemplates.ts
```diff
@@ -0,0 +1 @@
+export { defaultTemplates as default } from './stats';
```
Haroenv, line 1: what about importing it in the widget file instead?

src/components/Stats/Stats.tsx
```diff
@@ -0,0 +1,53 @@
...
+}: StatsProps) => (
+  <div className={cx(cssClasses.root)}>
```
Haroenv, line 31: is cx to allow arrays for the class name here? smart fix!

src/connectors/stats/__tests__/connectStats-test.ts
```diff
@@ -65,13 +66,13 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/stats/js/#c
       const renderFn = jest.fn();
       const unmountFn = jest.fn();
       const createStats = connectStats(renderFn, unmountFn);
-      const stats = createStats();
+      const stats = createStats({});
```
Haroenv, line 69: we should still have a test somewhere instantiating createStats() without argument, is there one?

src/connectors/stats/connectStats.ts
```diff
@@ -3,58 +3,71 @@ import {
...
- *   })
- * ]);
  */
-export default function connectStats(renderFn, unmountFn = noop) {
+
+export type StatsRendererOptions = {
+  /**
+   * The maximum number of hits per page returned by Algolia.
+   */
+  hitsPerPage: number | undefined;
```
Haroenv, line 22:
```suggestion
  hitsPerPage?: number;
```

src/connectors/stats/connectStats.ts
```diff
@@ -3,58 +3,71 @@ import {
...
+  /**
+   * The number of hits in the result set.
+   */
+  nbHits: number;
+  /**
+   * The number of sorted hits in the result set (when using Relevant sort).
+   */
+  nbSortedHits: number | undefined;
```
Haroenv, line 30:
```suggestion
  nbSortedHits?: number;
```

src/connectors/stats/connectStats.ts
```diff
@@ -3,58 +3,71 @@ import {
...
+   */
+  processingTimeMS: number;
+  /**
+   * The query used for the current search.
+   */
+  query: string;
+  /**
+   * All original `StatsConnectorParams` forwarded to the `renderFn`.
+   */
+  widgetParams: StatsConnectorParams;
```
Haroenv, line 54:
```suggestion
```

(this gets added by Connector type)

src/connectors/stats/connectStats.ts
```diff
@@ -3,58 +3,71 @@ import {
...
+>;
+
+const connectStats: StatsConnector = function connectStats(
+  renderFn,
+  unmountFn = noop
+) {
   checkRendering(renderFn, withUsage());
 
-  return (widgetParams = {}) => ({
+  return widgetParams => ({
```
Haroenv, line 70: we should still allow empty widgetParams, otherwise it's a breaking change

src/types/algoliasearch.ts
```diff
@@ -38,7 +38,10 @@ export type SearchResponse<
   THit
 > = DefaultSearchClient extends DummySearchClientV4
   ? SearchResponseV4<THit>
-  : SearchResponseV3<THit> & { appliedRelevancyStrictness?: number };
+  : SearchResponseV3<THit> & {
+      appliedRelevancyStrictness?: number;
+      nbSortedHits?: number;
+    };
```
Haroenv, line 44: maybe put this interface in its own type called SearchResponsePatchV3 or something?

src/widgets/stats/__tests__/stats-test.ts
```diff
@@ -0,0 +1,164 @@
...
+import { castToJestMock } from '../../../../test/utils/castToJestMock';
+
+jest.mock('preact', () => {
+  const module = jest.requireActual('preact');
+
+  return {
+    ...module,
+    render: jest.fn(),
+  };
+});
```
Haroenv, line 12:
```suggestion
import { render as preactRender } from 'preact';
import stats from '../stats';
import { castToJestMock } from '../../../../test/utils/castToJestMock';

const render = castToJestMock(preactRender);
jest.mock('preact', () => {
  const module = jest.requireActual('preact');

  module.render = jest.fn();

  return module;
});
```

(then remove the calls to castToJestMock in the actual tests)

This is more consistent with the other tests

Haroenv (commented):
good changes! just make sure it can still be called without widgetParams

src/components/Stats/Stats.tsx
shortcuts: Yes!

src/widgets/stats/defaultTemplates.ts
shortcuts: Do you mean the other way around? We do it this way in the latest `refinementList` update, but the other way in other files.

Which one should be the one we stick to?

src/widgets/stats/stats.tsx
shortcuts: Fixed in https://github.com/algolia/instantsearch.js/pull/4681/commits/d9217028858b5fdd2bce87a72b38438d373d0f3f

src/types/algoliasearch.ts
shortcuts: as this is only related to virtual replica indices with the `relevant sort` option, is `SearchWithRelevantSortResponseV3` good?

src/widgets/stats/defaultTemplates.ts
Haroenv: good point, it's the same for me actually, in refinement list I moved them inside since I did some more logic, but not sure what's best

src/types/algoliasearch.ts
Haroenv: maybe RelevantSortResponse

shortcuts: Added in https://github.com/algolia/instantsearch.js/pull/4681/commits/96060a6f2a852eb904ebed98e98400305ba2da58

src/connectors/stats/connectStats.ts
Haroenv: I added a test in 349dcc22 that shows it isn't needed :)

src/connectors/stats/__tests__/connectStats-test.ts
Haroenv: done in 349dcc22

Haroenv (approved):

PR #4682, opened 2021-03-18
chore(ci): make build fail when circular dependency is detected

shortcuts:
**Summary**

The `build` command should fail when a circular dependency is detected, other warnings will be treated as before.

Related: https://github.com/algolia/instantsearch.js/pull/4680#issuecomment-801794688
Resources: [rollup doc](https://rollupjs.org/guide/en/#onwarn)

This PR should be merged after https://github.com/algolia/instantsearch.js/pull/4680 so that it can prove that it works

### DIFF
--- a/scripts/rollup/config.js
+++ b/scripts/rollup/config.js
@@ -49,6 +49,12 @@ const createConfiguration = ({ mode, filename }) => ({
     banner: license,
     sourcemap: true,
   },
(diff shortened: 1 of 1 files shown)


eunjae-lee:
Amazing!

francoischalifour:
That's a good addition—thanks!

PR #4685, opened 2021-03-18
feat(hierarchicalMenu): implement canRefine

Haroenv:
**Summary**

implements canRefine in connectHierarchicalMenu as true when more than 0 items are possible available (a single item can still refine, even if the value of that is small, this way it's consistent with breadcrumb)

DX-1305

**Result**

hierarchicalMenu now has a canRefine

### DIFF
--- a/src/connectors/hierarchical-menu/connectHierarchicalMenu.js
+++ b/src/connectors/hierarchical-menu/connectHierarchicalMenu.js
@@ -232,6 +232,7 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
         return {
           items,
           refine: this._refine,
+          canRefine: items.length > 0,
           createURL: _createURL,
           sendEvent,
           widgetParams,
--- a/src/connectors/hierarchical-menu/__tests__/connectHierarchicalMenu-test.js
+++ b/src/connectors/hierarchical-menu/__tests__/connectHierarchicalMenu-test.js
@@ -472,6 +472,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hierarchica
         category: {
           items: [],
           refine: expect.any(Function),
+          canRefine: false,
           createURL: expect.any(Function),
           sendEvent: expect.any(Function),
           widgetParams: { attributes: ['category', 'subCategory'] },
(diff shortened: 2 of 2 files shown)


eunjae-lee (approved):

shortcuts (approved):

src/connectors/hierarchical-menu/connectHierarchicalMenu.js
```diff
@@ -232,6 +232,7 @@ export default function connectHierarchicalMenu(renderFn, unmountFn = noop) {
         return {
           items,
           refine: this._refine,
+          canRefine: items.length > 0,
```
Haroenv, line 235: should be > 1, and possibly same for breadcrumb?

eunjae-lee: Even if there is only one item, it doesn't mean it's already selected. So, I think canRefine should be true. What do you think?

Haroenv: yes, but it won't make a difference wether it's selected or not. If all items are of a single category, selecting it won't change the results.

eunjae-lee: Yes, that's correct. This and [that](https://github.com/algolia/instantsearch.js/pull/4691) make sense, but it will deviate from the implementation of Vue InstantSearch. Should we deviate now and later introduce a breaking change in Vue InstantSearch when it starts using the canRefine from InstantSearch.js?

Haroenv: Yes, I think it's better to have a small breaking change in Vue InstantSearch later when we use canRefine there :)

eunjae-lee: Thinking again, another point to consider is, when this widget is wrapped with `panel`, it could collapse or hide the widget when there is only one item if developers use `canRefine` as the condition. Would you expect it to be hidden with only item? or with zero item?

Haroenv: I would expect it to be hidden when there's no difference possible anymore in the results, therefore 1 item would do that :)

PR #4687, opened 2021-03-19
feat(dynamicWidgets): implementation

Haroenv:
**Summary**

Adds a new widget _dynamicWidgets_ that can be used to conditionally render other widgets. This condition is based on the search results. At the moment there isn't yet a default way to enforce facet ordering in the Algolia engine, thus the data available to `transformItems` is an empty array. This will change once the Algolia engine adds support for facet ordering and this widget will move out of experimental mode.

**Result**

```js
search.addWidgets([
  dynamicWidgets({
    container: '#sidebar',
    transformItems(attributes, { results }) {
      // add a condition based on the results, eg. if you add the ordering via a query rule:
      return results.userData[0].facetOrdering;
    },
    widgets: [
      (container) => refinementList({ container, attribute: 'brand' }),
      (container) => menu({ container, attribute: 'categories' }),
      (container) =>
        panel({ templates: { header: 'rating' } })(ratingMenu)({
          container,
          attribute: 'rating',
        }),
    ],
  }),
]);
```


src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -0,0 +1,106 @@
...
+        );
+      },
+      dispose() {
+        unmountFn();
+      },
+      getRenderState(renderState) {
+        // @TODO: decide whether it makes sense to return anything,
+        // knowing that you could have multiple instances of dynamic
+        // and they are hard/impossible to distinguish at this point
+        return renderState;
```
Haroenv, line 90: I think it makes sense not to return any renderState here

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -0,0 +1,106 @@
...
+      },
+      getWidgetRenderState({ results }) {
+        if (!results) {
+          return { attributesToRender: [], widgetParams };
+        }
+
+        // @TODO: retrieve the facet order out of the results:
+        // results.facetOrder.map(facet => facet.attribute)
+        const attributesToRender = transformItems([], { results });
```
Haroenv, line 99: this will be done after merge

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -0,0 +1,106 @@
...
+            });
+          }
+        });
+
+        parent!.addWidgets(widgetsToMount);
+        // make sure this only happens after the regular render, otherwise it
+        // happens too quick, since render is "deferred" for the next microtask,
+        // so this needs to be a whole task later
+        setTimeout(() => parent!.removeWidgets(widgetsToUnmount), 0);
+
```
Haroenv, line 74: this code could potentially be part of the widget instead of the connector since vue (and react) can rely on the host's ability to mount and unmount, but that would mean the widget itself becomes slightly more complicated, since it would have both rendering logic and widget mounting logic

src/lib/utils/getWidgetAttribute.ts
```diff
@@ -0,0 +1,37 @@
...
+        : widgetParams.attributes[0];
+
+    if (typeof attribute !== 'string') throw new Error();
+
+    return attribute;
+  } catch (e) {
+    throw new Error(
+      `Could not find the attribute of the widget:
+
+${JSON.stringify(widget)}
```
Haroenv, line 32: for custom widgets without type this will serialise as `{}`, do we see an alternative?

src/widgets/dynamic-widgets/dynamic-widgets.ts
```diff
@@ -0,0 +1,66 @@
+import connectDynamicWidgets, {
+  DynamicWidgetsConnectorParams,
+  DynamicWidgetsRendererOptions,
+} from '../../connectors/dynamic-widgets/connectDynamicWidgets';
+import { getWidgetAttribute } from '../../lib/utils';
+import { Widget, WidgetFactory } from '../../types';
+
+export type DynamicWidgetsWidgetParams = {
+  container: HTMLElement;
```
Haroenv, line 9:
```suggestion
  container: HTMLElement | string;
```

src/widgets/dynamic-widgets/dynamic-widgets.ts
```diff
@@ -0,0 +1,66 @@
...
+      attributesToRender.forEach(attribute => {
+        if (!containers.has(attribute)) {
+          return;
+        }
+        const container = containers.get(attribute)!;
+        rootContainer.appendChild(container);
+      });
+    },
+    () => {
+      rootContainer.innerText = '';
```
Haroenv, line 37: different way of emptying the container?

src/widgets/dynamic-widgets/dynamic-widgets.ts
```diff
@@ -0,0 +1,66 @@
...
+    transformItems,
+    widgets: connectorWidgets,
+  });
+
+  return {
+    ...widget,
+    init(initOptions) {
+      widgets.forEach(cb => {
+        const container = document.createElement('div');
+        container.className = 'ais-DynamicWidgets-widget';
```
Haroenv, line 51: use suit

src/widgets/dynamic-widgets/dynamic-widgets.ts
```diff
@@ -0,0 +1,66 @@
...
+  const connectorWidgets: Widget[] = [];
+
+  const makeWidget = connectDynamicWidgets(
+    ({ attributesToRender }) => {
+      attributesToRender.forEach(attribute => {
+        if (!containers.has(attribute)) {
+          return;
+        }
+        const container = containers.get(attribute)!;
+        rootContainer.appendChild(container);
```
Haroenv, line 33: root container needs to be made _inside_ container

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -0,0 +1,106 @@
...
+        renderFn(
+          {
+            ...renderState,
+            instantSearchInstance: renderOptions.instantSearchInstance,
+          },
+          false
+        );
+      },
+      dispose() {
+        unmountFn();
```
Haroenv, line 84: mounted widgets need to be unmounted

src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
```diff
@@ -0,0 +1,581 @@
...
+              createMultiSearchResponse({
+                userData: [{ MOCK_facetOrder: ['test2', 'test1'] }],
+              }).results
+            ),
+          })
+        );
+
+        await runAllMacroTasks();
+
+        expect(parent.getWidgets()).toMatchInlineSnapshot(`
```
eunjae-lee, line 362: why isn't the order test2 → test1 ?

eunjae-lee (changes requested):

src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
Haroenv: the widgets themselves don't have a useful concept of ordering, therefore the only important thing is that the right ones are mounted more than their order.

src/widgets/__tests__/index.test.ts
```diff
@@ -121,6 +121,13 @@ function initiateAllWidgets(): Array<[WidgetNames, Widget]> {
           attributes: ['attr1', 'attr2'],
         });
       }
+      case 'EXPERIMENTAL_dynamicWidgets': {
+        const EXPERIMENTAL_dynamicWidgets = widget as Widgets['EXPERIMENTAL_dynamicWidgets'];
+        return EXPERIMENTAL_dynamicWidgets({
+          container,
```
shortcuts, line 127:
```suggestion
          container,
          transformItems() {
            return [];
          }
```

as it's now required

eunjae-lee (approved):

src/lib/utils/getWidgetAttribute.ts
shortcuts: We could just throw `Could not find the attribute of the custom widget.`, wdyt?

shortcuts (approved):

PR #4688, opened 2021-03-22
feat(numericMenu): implement canRefine

eunjae-lee:
**Summary**

implements canRefine in connectNumericMenu as true when there are hits

DX-1306

**Result**

numericMenu now has a canRefine

### DIFF
--- a/src/connectors/numeric-menu/connectNumericMenu.ts
+++ b/src/connectors/numeric-menu/connectNumericMenu.ts
@@ -324,11 +324,13 @@ const connectNumericMenu: NumericMenuConnector = function connectNumericMenu(
           });
         }
 
+        const facetValues = transformItems(prepareItems(state));
+
         return {
           createURL: connectorState.createURL!(state),
-          items: transformItems(prepareItems(state)),
+          items: facetValues,
           hasNoResults: results ? results.nbHits === 0 : true,
-          canRefine: Boolean(results && results.nbHits > 0),
+          canRefine: facetValues.length > 0,
           refine: connectorState.refine!,
           sendEvent: connectorState.sendEvent!,
           widgetParams,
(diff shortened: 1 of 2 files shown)


src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -323,6 +328,7 @@ const connectNumericMenu: NumericMenuConnector = function connectNumericMenu(
           createURL: connectorState.createURL!(state),
           items: transformItems(prepareItems(state)),
           hasNoResults: results ? results.nbHits === 0 : true,
+          canRefine: Boolean(results && results.nbHits > 0),
```
Haroenv, line 331: maybe `items.every(({count}) => count > 0)`?

eunjae-lee: every vs some?

Haroenv: both every and some shortcut as soon as the condition is fulfilled or not fulfilled so it's not that big of a difference, the condition just needs to be inverted

Haroenv (changes requested):

src/connectors/numeric-menu/connectNumericMenu.ts
eunjae-lee: I'm looking at it again, and there is no count available: https://github.com/algolia/instantsearch.js/pull/4688/files#diff-9dc3db9af2e7feeb52390473e9a560153d8b66bf8a7b358d1041f52b74ff610cR35-R53

since it's an array made by user, we cannot know how many hits each item contains ahead.

so, what about just `items.length > 0`?

eunjae-lee: Actually that was intended to check the items after `transformItems`, but we shouldn't need to cover whatever happens in the userland. If we do it here, then we should do it in sortBy, too.
So I think I should just close this PR. What do you think?

Haroenv:
makes sense, since we don't have the info of the count, there should not be a canRefine

PR #4689, opened 2021-03-22
feat(toggleRefinement): implement canRefine

eunjae-lee:
**Summary**

implements canRefine in connectToggleRefinement as true when there value.count exists.

DX-1309

**Result**

toggleRefinement now has a canRefine

### DIFF
--- a/src/connectors/toggle-refinement/connectToggleRefinement.js
+++ b/src/connectors/toggle-refinement/connectToggleRefinement.js
@@ -61,6 +61,7 @@ const createSendEvent = ({ instantSearchInstance, attribute, on, helper }) => (
  * @typedef {Object} ToggleRenderingOptions
  * @property {ToggleValue} value The current toggle value.
  * @property {function():string} createURL Creates an URL for the next state.
+ * @property {boolean} canRefine Indicates if search state can be refined.
  * @property {function(value)} refine Updates to the next state by applying the toggle refinement.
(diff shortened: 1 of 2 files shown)


src/connectors/toggle-refinement/connectToggleRefinement.js
```diff
@@ -306,6 +307,7 @@ export default function connectToggleRefinement(renderFn, unmountFn = noop) {
             createURL,
           }),
           sendEvent,
+          canRefine: Boolean(results ? nextRefinement.count : null),
```
Haroenv, line 310: I'm not sure what condition it should be here. Maybe to be false, but current refinement and next refinement should have a 0 count? I'm not sure!

eunjae-lee: This condition is actually coming from Vue InstantSearch

```js
const mapStateToCanRefine = state => Boolean(state.value && state.value.count);
```

and `state.value` is made in InstantSearch.js like this:

```js
          value: {
            name: attribute,
            isRefined,
            count: results ? nextRefinement.count : null,
            onFacetValue,
            offFacetValue,
          },
```

I'm not too sure either, but it'd be good to align between InstantSearch.js and Vue InstantSearch first, and then later change to something else if needed from the InstantSearch.js and propagate to Vue InstantSearch. What do you think?

Haroenv (approved):
good enough for now

PR #4691, opened 2021-03-22
feat(ratingMenu): implement canRefine

eunjae-lee:
**Summary**

implements canRefine in connectRatingMenu as true when nbHits > 0.

DX-1308

**Result**

ratingMenu now has a canRefine

### DIFF
--- a/src/connectors/rating-menu/connectRatingMenu.js
+++ b/src/connectors/rating-menu/connectRatingMenu.js
@@ -66,6 +66,7 @@ const createSendEvent = ({
  * @property {StarRatingItems[]} items Possible star ratings the user can apply.
  * @property {function(string): string} createURL Creates an URL for the next
  * state (takes the item value as parameter). Takes the value of an item as parameter.
+ * @property {boolean} canRefine Indicates if search state can be refined.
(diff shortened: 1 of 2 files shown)


src/connectors/rating-menu/connectRatingMenu.js
```diff
@@ -327,6 +328,7 @@ ${
         return {
           items: facetValues,
           hasNoResults: results ? results.nbHits === 0 : true,
+          canRefine: Boolean(results && results.nbHits > 0),
```
Haroenv, line 331: instead it should be facetValues.length > 1 I think

tkrugg: I think I'm missing somehting. 
Isn't `canRefine` supposed to be on the item level?
My understanding was: `canRefine` tells you whether you can click (i.e. call refine) on a specific item.

Haroenv: it's meant for if you can refine the whole widget, every item can be refined normally

Haroenv (approved):

PR #4692, opened 2021-03-22
feat(sortBy): implement canRefine

eunjae-lee:
**Summary**

implements canRefine in connectSortBy as true when nbHits > 0.

DX-1314

**Result**

sortBy now has a canRefine

### DIFF
--- a/src/connectors/sort-by/connectSortBy.js
+++ b/src/connectors/sort-by/connectSortBy.js
@@ -27,6 +27,7 @@ const withUsage = createDocumentationMessageGenerator({
  * @typedef {Object} SortByRenderingOptions
  * @property {string} currentRefinement The currently selected index.
  * @property {SortByItem[]} options All the available indices
+ * @property {boolean} canRefine Indicates if search state can be refined.
  * @property {function(string)} refine Switches indices and triggers a new search.
  * @property {boolean} hasNoResults `true` if the last search contains no result.
  * @property {Object} widgetParams All original `CustomSortByWidgetParams` forwarded to the `renderFn`.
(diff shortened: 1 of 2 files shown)


Haroenv (approved):

src/connectors/sort-by/connectSortBy.js
```diff
@@ -164,6 +165,7 @@ export default function connectSortBy(renderFn, unmountFn = noop) {
         return {
           currentRefinement: helper.state.index,
           options: transformItems(items),
+          canRefine: Boolean(results && results.nbHits > 0),
```
Haroenv, line 168: actually, no, this is wrong. You can still change index if this index has no refinements, so canRefine is probably always true

Haroenv (changes requested):

src/connectors/sort-by/connectSortBy.js
eunjae-lee: You're right. If it's always going to be true, then I think we shouldn't add it just for the consistency. What do you think?

Haroenv: I'm not sure, I think we should probably leave it out if there's no reasonable for it not to refine, same as DX-1312

tkrugg: But you're right, I don't since it's worth adding at all.

One use case for `canRefine === false` I think might be when we have only one index and it's already selected, and no indices at all. Both have workaround maybe `canRefine` superfluous IMHO.

Haroenv:
as discussed, sortby can always refine, therefore it makes no sense to have the boolean

PR #4693, opened 2021-03-22
feat(relevantSort): implement canRefine

eunjae-lee:
**Summary**

implements canRefine in connectRelevantSort as true when nbHits > 0 and it's a virtual replica.

**Result**

relevantSort now has a canRefine

### DIFF
--- a/src/connectors/relevant-sort/connectRelevantSort.ts
+++ b/src/connectors/relevant-sort/connectRelevantSort.ts
@@ -100,7 +100,7 @@ const connectRelevantSort: RelevantSortConnector = function connectRelevantSort(
             typeof appliedRelevancyStrictness !== 'undefined' &&
             appliedRelevancyStrictness > 0,
           isVirtualReplica,
-          canRefine: isVirtualReplica && results?.nbHits! > 0,
+          canRefine: isVirtualReplica,
           refine: connectorState.refine,
           widgetParams,
         };
--- a/src/connectors/relevant-sort/connectRelevantSort.ts
+++ b/src/connectors/relevant-sort/connectRelevantSort.ts
@@ -6,8 +6,24 @@ export type RelevantSortConnectorParams = Record<string, unknown>;
 type Refine = (relevancyStrictness: number) => void;
 
 export type RelevantSortRendererOptions = {
+  /**
+   * Indicates if it has appliedRelevancyStrictness greater than zero
+   */
   isRelevantSorted: boolean;
+
+  /**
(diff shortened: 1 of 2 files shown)


src/connectors/relevant-sort/connectRelevantSort.ts
```diff
@@ -77,11 +93,14 @@ const connectRelevantSort: RelevantSortConnector = function connectRelevantSort(
...
 
+        const isVirtualReplica = appliedRelevancyStrictness !== undefined;
+
         return {
           isRelevantSorted:
             typeof appliedRelevancyStrictness !== 'undefined' &&
             appliedRelevancyStrictness > 0,
-          isVirtualReplica: appliedRelevancyStrictness !== undefined,
+          isVirtualReplica,
+          canRefine: isVirtualReplica && results?.nbHits! > 0,
```
eunjae-lee, line 103: Even if `nbSortedHits === 0`, there can be a possibility to see different results with different `relevancyStrictness`. So we're checking it with `nbHits`, not with `nbSortedHits`.

Haroenv: I wonder if just `isVirtualReplica` isn't enough here, can't you in theory still swap between virtual and regular view of something with 0 hits? On the other hand, having no hits is probably a good condition to hide it

Haroenv (approved):

src/connectors/relevant-sort/connectRelevantSort.ts
eunjae-lee: Umm right. We can swap between virtual and regular, so even if it's not virtual, as long as nbHits > 0, it should be able to refine. So I'm thinking of just `results?.nbHits! > 0`. Makes sense to you?

Haroenv: actually thought about it again, and I think the current behaviour is correct. Can someone confirm? @francoischalifour @shortcuts

shortcuts: > so even if it's not virtual, as long as nbHits > 0, it should be able to refine.

But refine rely on the `relevancyStrictness`, so we should check if it's a virtual replica anyway, no?

I think the current implementation make sense

eunjae-lee: Oh right. `isVirtualReplica` is true when it's 0, too. So it seems alright to me.

shortcuts (approved):

Haroenv (approved):

PR #4694, opened 2021-03-23
feat(metadata): expose client's algolia agent

Haroenv:
**Summary**

The user agent can be used to inspect whether the client is using InstantSearch.js via an integration like Vue InstantSearch or Angular InstantSearch

**Result**

metadata now exposes the algolia agent

### DIFF
--- a/src/middlewares/__tests__/createMetadataMiddleware.ts
+++ b/src/middlewares/__tests__/createMetadataMiddleware.ts
@@ -1,3 +1,5 @@
+import algoliasearch from 'algoliasearch';
+import algoliasearchV3 from 'algoliasearch-v3';
 import { createMetadataMiddleware } from '..';
 import { createSearchClient } from '../../../test/mock/createSearchClient';
 import { wait } from '../../../test/utils/wait';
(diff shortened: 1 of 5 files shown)


tsconfig.v3.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig",
+  "exclude": [
+    "examples",
+    "es",
+    // this test has specific code for v3 and v4, so already checked in the v4 test
```
eunjae-lee, line 6: This comment doesn't break anything?

src/middlewares/__tests__/createMetadataMiddleware.ts
```diff
@@ -110,45 +112,220 @@ describe('createMetadataMiddleware', () => {
...
+        });
+
+        search.start();
+
+        await wait(100);
+
+        expect(document.head).toMatchInlineSnapshot(`
+          <head>
+            <meta
+              content="{\\"widgets\\":[],\\"ua\\":\\"Algolia for JavaScript (4.8.6); Node.js (12.14.1); instantsearch.js (4.17.0); JS Helper (3.4.4)\\"}"
```
eunjae-lee, line 177: Can we check these with regex with only actual number on the major?
Every time we update the version of any of them, this test will fail.

tsconfig.v3.json
Haroenv: no, tsconfig allows comments

src/middlewares/__tests__/createMetadataMiddleware.ts
Haroenv: solved in afc78a6

eunjae-lee (approved):
Looks good to me.

PR #4699, opened 2021-03-25
fix(setUiState): make sure previous ui state is stored

Haroenv:
currently if you call setUiState, the localUiState in index widgets doesn't get updated to prevent middlewares being called multiple times, but that means if you call setUiState multiple times, the local state doesn't change

### DIFF
--- a/src/lib/__tests__/InstantSearch-test.js
+++ b/src/lib/__tests__/InstantSearch-test.js
@@ -1615,7 +1615,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
     expect(searchClient.search).toHaveBeenCalledTimes(2);
   });
 
-  test('notifies all middleware', async () => {
+  test('notifies all middlewares', async () => {
     const searchClient = createSearchClient();
     const search = new InstantSearch({
       indexName: 'indexName',
--- a/src/lib/__tests__/InstantSearch-test.js
+++ b/src/lib/__tests__/InstantSearch-test.js
@@ -1434,7 +1434,7 @@ describe('use', () => {
     const middleware = jest.fn(() => middlewareSpy);
 
     search.addWidgets([searchBox]);
-    search.EXPERIMENTAL_use(middleware);
+    search.use(middleware);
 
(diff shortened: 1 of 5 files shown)


src/lib/__tests__/InstantSearch-test.js
```diff
@@ -1633,6 +1633,58 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
...
+    search.EXPERIMENTAL_use(middleware);
+    search.start();
+    expect(onMiddlewareStateChange).toHaveBeenCalledTimes(0);
+
+    search.setUiState({
+      indexName: {},
+      test: {},
+    });
+
+    expect(onMiddlewareStateChange).toHaveBeenCalledTimes(2);
```
Haroenv, line 1666: called once per index instead of only once (before), I don't see an easy way to change that tbh

eunjae-lee: ah, it was called twice because there was an index widget in the tree. but I fear calling onMiddlewareStateChange once per index might lead to a weird interim state especially for routing. just a hunch/concern, not a concrete observation, though.

Haroenv: solved by deferring the state change, which causes the onStateChange only to be called once

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -1603,12 +1615,51 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
...
+    const onMiddlewareStateChange = jest.fn();
+    const middleware = () => {
+      return {
+        subscribe() {},
+        unsubscribe() {},
+        onStateChange: onMiddlewareStateChange,
+      };
+    };
+
+    search.EXPERIMENTAL_use(middleware);
```
eunjae-lee, line 1633:
```suggestion
    search.use(middleware);
```

You can just use `use` now :)

eunjae-lee (approved):
Looks good to me.

francoischalifour (approved):

PR #4700, opened 2021-03-29
feat(ts): convert sortBy, connectSortBy

shortcuts:
**Summary**

Convert `sortBy`, `connectSortBy` to TypeScript


src/widgets/sort-by/__tests__/sort-by-test.ts
```diff
@@ -106,11 +133,4 @@ describe('sortBy()', () => {
       },
     ]);
   });
-
-  it('sets the underlying index', () => {
-    widget.setIndex('index-b');
-
-    expect(helper.setIndex).toHaveBeenCalledTimes(1);
-    expect(helper.search).toHaveBeenCalledTimes(1);
-  });
```
shortcuts, line 115: Should this be possible? As we don't expose `setIndex`, I feel like this test isn't needed

Haroenv: it should be calling `getWidgetRenderState().refine` for an equivalent test

Haroenv: however it makes sense that this is tested in the connector alone

src/connectors/sort-by/connectSortBy.ts
```diff
@@ -0,0 +1,189 @@
...
+            instantSearchInstance,
+          },
+          false
+        );
+      },
+
+      dispose({ state }) {
+        unmountFn();
+
+        return state.setIndex(connectorState.initialIndex || '');
```
Haroenv, line 135: we probably prefer throwing instead of setting to an empty string, or even `if initialIndex setIndex`

shortcuts: Wouldn't it be a breaking change? For now, we accept it to be `undefined` if I understand correctly

Haroenv: in the previous implementation setIndex was called with undefined, which I think would already not be the behaviour you want (it would make subsequent calls fail):

```

> const helper = require('algoliasearch-helper')
undefined
> const h = helper({search: console.log}, 'test')
undefined
> h.search()
[ { indexName: 'test', params: { facets: [], tagFilters: '' } } ]
> h.setIndex().search()
[ { indexName: undefined, params: { facets: [], tagFilters: '' } } ]
```

I'm in favour of either keeping it broken in the same was as before (`initialIndex!`), or only calling setIndex if we know the previous index, otherwise leaving it as-is

shortcuts: Warns and only call `setIndex` when the previous index is defined: https://github.com/algolia/instantsearch.js/pull/4700/commits/f725991120710cd96605fc8a9c0db9fd1d04a3ea

src/connectors/sort-by/connectSortBy.ts
```diff
@@ -132,7 +132,14 @@ const connectSortBy: SortByConnector = function connectSortBy(
...
 
-        return state.setIndex(connectorState.initialIndex || '');
+        warning(
+          connectorState.initialIndex !== undefined,
+          "The index name can't be undefined when unmounting."
+        );
+
+        return connectorState.initialIndex
+          ? state.setIndex(connectorState.initialIndex)
+          : undefined;
```
Haroenv, line 142:
```suggestion
          : state;
```

src/connectors/sort-by/connectSortBy.ts
```diff
@@ -132,7 +132,14 @@ const connectSortBy: SortByConnector = function connectSortBy(
       dispose({ state }) {
         unmountFn();
 
-        return state.setIndex(connectorState.initialIndex || '');
+        warning(
+          connectorState.initialIndex !== undefined,
+          "The index name can't be undefined when unmounting."
+        );
```
Haroenv, line 138: I don't think this really needs a warning, since there's no practical way of it happening :)

shortcuts: Should it need a test then? As it can't really happen

Haroenv (approved):

PR #4702, opened 2021-03-29
feat(ts): convert RefinementList component to TypeScript

yannickcr:
**Summary**

Converts the` RefinementList` component to TypeScript


src/components/RefinementList/__tests__/RefinementListItem-test.tsx
```diff
@@ -2,10 +2,27 @@
 
 import { h } from 'preact';
 import { shallow } from 'enzyme';
+import { ReactElementLike } from 'prop-types';
 import RefinementListItem from '../RefinementListItem';
 
+type Props = {
```
Haroenv, line 8: if the props were exported from the component, you could use them here instead of repeating

src/components/RefinementList/__tests__/RefinementListItem-test.tsx
```diff
@@ -2,10 +2,27 @@
...
+  templateKey: string;
+  templateProps: {
+    template: string;
+  };
+  subItems: h.JSX.Element;
+};
+
 describe('RefinementListItem', () => {
-  const props = {
+  const props: Props = {
```
Haroenv, line 25: or I think the type here likely could be omitted. What doesn't work if it's removed?

src/connectors/refinement-list/connectRefinementList.ts
```diff
@@ -51,6 +51,10 @@ export type RefinementListItem = {
    * Indicates if the list item is refined.
    */
   isRefined: boolean;
+  /**
+   * List item childrens.
+   */
+  data?: RefinementListItem[];
```
Haroenv, line 57: only hierarchical menu has `data`, not refinementList

src/widgets/refinement-list/refinement-list.tsx
```diff
@@ -24,7 +24,7 @@ const withUsage = createDocumentationMessageGenerator({
 const suit = component('RefinementList');
 const searchBoxSuit = component('SearchBox');
 
-type RefinementListOwnTemplates = {
+export type RefinementListOwnTemplates = {
```
Haroenv, line 27: what is this exported for? shouldn't most use cases use `RefinementListTemplates`, which is slightly lower

src/components/RefinementList/RefinementList.tsx
```diff
@@ -1,37 +1,113 @@
...
 import RefinementListItem from './RefinementListItem';
 import SearchBox from '../SearchBox/SearchBox';
+import { RefinementListItem as TRefinementListItem } from '../../connectors/refinement-list/connectRefinementList';
+import { RefinementListOwnTemplates } from '../../widgets/refinement-list/refinement-list';
+import { RatingMenuTemplates } from '../../widgets/rating-menu/rating-menu';
+import { MenuTemplates } from '../../widgets/menu/menu';
+import { SearchBoxTemplates } from '../../widgets/search-box/types';
+import { CreateURL } from '../../types';
+
+export type RefinementListCSSClasses = {
```
Haroenv, line 17: these are the same as in the refinement list widget, except marked as optional. I think they should be defined here, however I don't think it's optional

src/components/RefinementList/RefinementList.tsx
```diff
@@ -1,37 +1,113 @@
...
+export type RefinementListProps = {
+  createURL: CreateURL<string>;
+  cssClasses?: RefinementListCSSClasses;
+  depth?: number;
+  facetValues?: TRefinementListItem[];
+  attribute?: string;
+  templateProps?:
+    | PreparedTemplateProps<RefinementListOwnTemplates>
+    | PreparedTemplateProps<RatingMenuTemplates>
+    | PreparedTemplateProps<MenuTemplates>;
```
Haroenv, line 57: could the props & component be generic so that the templates type is passed as usage, and we don't need to expose RefinementListOwnTemplates

src/components/RefinementList/RefinementList.tsx
```diff
@@ -1,37 +1,113 @@
...
+  toggleShowMore?: () => void;
+  isShowingMore?: boolean;
+  hasExhaustiveItems?: boolean;
+  canToggleShowMore?: boolean;
+  searchIsAlwaysActive?: boolean;
+  className?: string;
+  children?: h.JSX.Element;
+};
+
+const defaultProps = Object.freeze({
```
Haroenv, line 73: why do we need to freeze here (same question on the other PR), can't we assume that nothing will modify it anyway by marking `as const`? I think Object.freeze isn't ie11 compatible

src/components/RefinementList/RefinementList.tsx
```diff
@@ -1,37 +1,113 @@
...
+
+  private searchBox = createRef<SearchBox>();
+
+  public constructor(props: RefinementListPropsWithDefaultProps) {
     super(props);
     this.handleItemClick = this.handleItemClick.bind(this);
   }
 
-  shouldComponentUpdate(nextProps, nextState) {
-    const isStateDifferent = this.state !== nextState;
```
Haroenv, line 18: why is this changed?

src/components/RefinementList/RefinementList.tsx
```diff
@@ -1,37 +1,113 @@
...
       this.props.facetValues,
       nextProps.facetValues
     );
 
-    return isStateDifferent || areFacetValuesDifferent;
+    return areFacetValuesDifferent;
   }
 
-  refine(facetValueToRefine, isRefined) {
-    this.props.toggleRefinement(facetValueToRefine, isRefined);
```
Haroenv, line 28: is this second argument not used?

src/components/RefinementList/RefinementList.tsx
```diff
@@ -63,18 +139,28 @@ class RefinementList extends Component {
...
+        [this.props.cssClasses.selectedItem]: facetValue.isRefined,
+      },
+      this.props.cssClasses.disabledItem && {
+        [this.props.cssClasses.disabledItem]: !facetValue.count,
+      },
+      this.props.cssClasses.parentItem && {
+        [this.props.cssClasses.parentItem]:
+          facetValue.data && facetValue.data.length > 0,
+      }
+    );
```
Haroenv, line 154:
```suggestion
    const refinementListItemClassName = cx(this.props.cssClasses.item, {
      [this.props.cssClasses.selectedItem]: facetValue.isRefined,
      [this.props.cssClasses.disabledItem]: !facetValue.count,
      [this.props.cssClasses.parentItem]: hasChildren,
    });
```

this previous version seems much smaller, why did you choose this?

src/components/RefinementList/RefinementList.tsx
```diff
@@ -115,59 +216,65 @@ class RefinementList extends Component {
...
-      this.refine(facetValueToRefine, isRefined);
+      this.refine(facetValueToRefine);
       return;
     }
 
     let parent = originalEvent.target;
 
     while (parent !== originalEvent.currentTarget) {
       if (
+        parent &&
```
Haroenv, line 227: parent can't be undefined

src/components/RefinementList/RefinementList.tsx
```diff
@@ -115,59 +216,65 @@ class RefinementList extends Component {
...
       [`${this.props.cssClasses.depth}${this.props.depth}`]: this.props
         .cssClasses.depth,
     });
 
-    const showMoreButtonClassName = cx(this.props.cssClasses.showMore, {
-      [this.props.cssClasses.disabledShowMore]: !(
-        this.props.showMore === true && this.props.canToggleShowMore
-      ),
-    });
+    const showMoreButtonClassName = cx(
```
Haroenv, line 270: same question here

src/components/RefinementList/RefinementList.tsx
```diff
@@ -191,18 +298,22 @@ class RefinementList extends Component {
 
     const templates = this.props.searchBoxTemplateProps
       ? this.props.searchBoxTemplateProps.templates
-      : {};
+      : undefined;
 
     const searchBox = this.props.searchFacetValues && (
       <div className={this.props.cssClasses.searchBox}>
         <SearchBox
-          ref={searchBoxRef => (this.searchBox = searchBoxRef)}
```
Haroenv, line 199: cool change, I didn't know it was possible to create a ref in a class component

Haroenv (commented):
very impressive!

src/components/RefinementList/__tests__/RefinementListItem-test.tsx
yannickcr: Everything works, but in `render(askedProps)` `askedProps` is set as `any` and I tried to avoid implicit `any` in the files I converted.

So if I need to type `askedProps` at least also type `props` to keep consistency.

src/connectors/refinement-list/connectRefinementList.ts
yannickcr: But hierarchical menu also uses the `RefinementList` component so the `RefinementListItem` received can either have a data property or not.

What I could do is to split this type in 2 (eg. `RefinementListItem` and `HierarchicalRefinementListItem`) and update the `RefinementList` component to accept the 2 types, so the `data` property does not leak into `connectRefinementList`. I'll do this in the HierachicalMenu PR since it needs to be coverted to TS first.

src/widgets/refinement-list/refinement-list.tsx
yannickcr: It is used to type `templateProps` here https://github.com/algolia/instantsearch.js/pull/4702/files#diff-000e299be1cea0f8ffc1aae9af4be59c88914c2d4c8fb5c0d80975ff85c2448fR55

src/components/RefinementList/RefinementList.tsx
yannickcr: This check was useless since the component has no internal state :wink:

src/components/RefinementList/RefinementList.tsx
yannickcr: Yes, this type is not perfect. Like `templateProps` it could be different depending of which widgets is calling the component (could be `RefinementListCSSClasses` but also `HierarchicalMenuCSSClasses`, `RatingMenuCSSClasses` or `MenuCSSClasses`).

I'll see if I can make the component generic to avoid this issue (same for `templateProps`)

src/components/RefinementList/RefinementList.tsx
yannickcr: Exactly. I'll see if making the component generic solves this issue, but since some classes are not always defined depending of the widget calling the component I think we will still have to do this at some point.

src/connectors/refinement-list/connectRefinementList.ts
Haroenv: I think there should be a type in the refinement list component that indeed declares the item with data optional, and a separate item type in both connect refinement list & connect hierarchical menu

src/components/RefinementList/RefinementList.tsx
Haroenv: ah yes good point, they aren't exactly the same. Let's opt for repeating here then so that we can have a type with only the classes we actually want to expose in each widget (searchable doesn't exist on menu for example)

src/components/RefinementList/RefinementList.tsx
Haroenv: let's go with `as const` here, since we own the object there's not a real risk of it being modified

src/components/RefinementList/RefinementList.tsx
Haroenv: considering it worked before, we can probably make them required to keep the logic simpler here (since it already works now)

yannickcr:
@Haroenv Here's the attempt to set RefinementList as a Generic (to pass templateProps and cssClasses) https://github.com/algolia/instantsearch.js/pull/4702/commits/232974f3ac03ccb3d76852fdb56c53062854b81e

Let me know what you think :slightly_smiling_face:

src/components/RefinementList/RefinementList.tsx
```diff
@@ -130,44 +223,43 @@ class RefinementList extends Component {
...
       this.props.toggleRefinement(actualValue);
     }
   }
 
-  render() {
-    // Adding `-lvl0` classes
-    const cssClassList = cx(this.props.cssClasses.list, {
-      [`${this.props.cssClasses.depth}${this.props.depth}`]: this.props
-        .cssClasses.depth,
-    });
```
yannickcr, line 164: `cssClasses.depth` is never set by the widget, so if I'm right this code always default to `cssClasses.list`.

(I also removed the related test, that was working since we were passing the `depth` class directly without using the widget)

Haroenv: also not with hierarchical menu?

src/widgets/rating-menu/rating-menu.tsx
```diff
@@ -78,6 +78,12 @@ export type RatingMenuCSSClasses = {
   count: string | string[];
 };
 
+type RatingMenuRendererCSSClasses = Required<
```
Haroenv, line 81: I'm not sure I fully understand the need of these types

src/connectors/refinement-list/connectRefinementList.ts
yannickcr: Done in https://github.com/algolia/instantsearch.js/pull/4711 (split it in 2 types `RefinementListItem` and `HierarchicalMenuItem`)

src/components/RefinementList/RefinementList.tsx
yannickcr: Nothing in hierarchical menu either https://github.com/algolia/instantsearch.js/blob/8e3c406c8f29bcae56d2f82f07cbd087043346fe/src/widgets/hierarchical-menu/hierarchical-menu.js#L197-L225

src/widgets/refinement-list/refinement-list.tsx
yannickcr: (it is no longer exported since `RefinementList` is now set as a Generic)

src/components/RefinementList/RefinementList.tsx
yannickcr: Done in https://github.com/algolia/instantsearch.js/commit/232974f3ac03ccb3d76852fdb56c53062854b81e

src/components/RefinementList/RefinementList.tsx
yannickcr: Making the component a Generic solved the issue, we can now keep the original syntax.

src/widgets/rating-menu/rating-menu.tsx
yannickcr: This type is used to type `cssClasses` in the renderer by derivating it from `RatingMenuCSSClasses` (the difference being that for each property we only accept `string` and not `string[]`)

PS: the Required is actually not useful here since all properties are already required in `RatingMenuCSSClasses`, I'll fix that.

src/components/RefinementList/RefinementList.tsx
yannickcr: Updated in https://github.com/algolia/instantsearch.js/commit/232974f3ac03ccb3d76852fdb56c53062854b81e, the CSSClasses type is now passed as a parameter to the generic.

Haroenv (approved):
this looks good, but since it's a big pr, let's make sure it's reviewed by someone else too

eunjae-lee (approved):
Massive work. Looks good to me. Thanks Yannick!

Haroenv:
something here broke the RatingMenu component, will investigate!

PR #4703, opened 2021-03-29
feat(ts): convert Template component to TypeScript

yannickcr:
**Summary**

Converts the `Template` component to TypeScript


src/components/Template/Template.tsx
```diff
@@ -1,23 +1,65 @@
...
+  data?: Record<string, any>;
+  rootProps?: Record<string, any>;
+  rootTagName?: string;
+  templateKey?: string;
+  templates?: Record<
+    string,
+    | string
+    | ((data: any, bindEvent: (...args: any[]) => string) => string)
+    | undefined
+  >;
```
Haroenv, line 16:
```suggestion
  templates?: Record<
    string,
    Template<TData>
  >;
```

and passing Data to the props too?

src/components/Template/Template.tsx
```diff
@@ -1,23 +1,65 @@
...
+    compileOptions?: {
+      asString?: boolean;
+      sectionTags?: Array<{
+        o?: string;
+        c?: string;
+      }>;
+      delimiters?: string;
+      disableLambda?: boolean;
+    };
+  };
```
Haroenv, line 32: good one, this should be put in the type in `prepareTemplateProps` too

src/components/Template/Template.tsx
```diff
@@ -1,23 +1,65 @@
...
+
+const defaultProps = Object.freeze({
+  data: {},
+  rootTagName: 'div',
+  useCustomCompileOptions: {},
+  templates: {},
+  templatesConfig: {},
+});
+
+class Template extends Component<TemplateProps> {
```
Haroenv, line 45: template should be generic with TData

src/components/Template/Template.tsx
Haroenv: (Template should come from /types/widget)

src/components/Template/Template.tsx
```diff
@@ -1,23 +1,65 @@
...
 
-  render() {
-    const RootTagName = this.props.rootTagName;
-    const useCustomCompileOptions = this.props.useCustomCompileOptions[
-      this.props.templateKey
-    ];
+  public render() {
+    const RootTagName = (this.props.rootTagName as unknown) as (
+      props: h.JSX.HTMLAttributes
+    ) => h.JSX.Element;
```
Haroenv, line 59: isn't it a string though? I'm pretty sure you should be allowed to start a string jsx element, no?

src/components/Template/Template.tsx
```diff
@@ -1,23 +1,65 @@
...
+    const useCustomCompileOptions =
+      this.props.templateKey &&
```
Haroenv, line 61: I think templateKey can be made required

src/components/Template/__tests__/Template-test.tsx
```diff
@@ -30,16 +31,16 @@ describe('Template', () => {
       templates: { test: 'it configures compilation <%options%>' },
       data: { options: 'delimiters' },
       useCustomCompileOptions: { test: true },
-      templatesConfig: { compileOptions: { delimiters: '<% %>' } },
+      templatesConfig: { helpers: {}, compileOptions: { delimiters: '<% %>' } },
     });
-    const wrapper = mount(<Template {...props} />);
+    const wrapper = mount((<Template {...props} />) as ReactElementLike);
```
Haroenv, line 36: why doe it need to be cast? Especially the one from prop-types is confusing to me

src/components/Template/Template.tsx
yannickcr: All string are not allowed (it must match an HTML Element tag name), but I found a better solution to avoid casting:

```ts
rootTagName?: keyof h.JSX.IntrinsicElements;
```

src/components/Template/__tests__/Template-test.tsx
yannickcr: I took the same approach as in https://github.com/algolia/instantsearch.js/pull/4681.

(Enzyme is waiting for a `ReactElement` but preact returns an `Element`, so we need to cast it)

src/components/Template/Template.tsx
yannickcr: Tried it but I was not able to pass it without breaking everything :disappointed: 

I took another approach and used `PreparedTemplateProps` to create part of the props (but as a result `TTemplateData` is still set to `any`). We can work together on this to find a better solution if you want.

Haroenv: we can iterate on this another time since the component is internal, possibly add a todo here?

src/connectors/rating-menu/connectRatingMenu.ts
```diff
@@ -16,7 +23,16 @@ const $$type = 'ais.ratingMenu';
...
+type SendEvent = (...args: [InsightsEvent] | [string, string, string?]) => void;
+
+type CreateSendEvent = (createSendEventArgs: {
+  instantSearchInstance: InstantSearch;
+  helper: AlgoliaSearchHelper;
+  getRefinedStar: () => number | number[] | undefined;
+  attribute: string;
+}) => SendEvent;
+
+const createSendEvent: CreateSendEvent = ({
```
Haroenv, line 35: these changes possibly need to be rebased, I thought this was merged separately

yannickcr: Done (mistake on my side, this PR was intended to have `feat/ts-ratingmenu` as base and not `master`)

Haroenv (approved):

PR #4707, opened 2021-03-30
test(connectors): avoid implicit any

shortcuts:
**Summary**

In some tests, widgets or functions doesn't return the right type.

Similar to https://github.com/algolia/instantsearch.js/pull/4677


Haroenv (approved):
looks good

PR #4708, opened 2021-03-30
feat(middleware): add unuse method

eunjae-lee:
**Summary**

This PR adds `unuse` method to unsubscribe and remove middleware that was added to InstantSearch.

So far, `this.middleware` in InstantSearch instance used to be `MiddlewareDefinition[]`.
Since this PR, it's now `Array<{ middleware: Middleware, middlewareDefinition: MiddlewareDefinition }>`.
(By the way, `type Middleware = (options: MiddlewareOptions) => MiddlewareDefinition;`)

The reason is, when we add a middleware, we pass a `middleware: Middleware`. However, we invokes the function and stores the middlewareDefinition. And here we loses the `middleware`. Later when we want to unuse the middleware, we have no way to match between the middleware and its middlewareDefinition.
To keep the connection, we're storing both of them. It doesn't chage any public API.

**Result**

Now we can unuse (unsubscribe) middlewares.


src/lib/InstantSearch.ts
```diff
@@ -285,6 +288,21 @@ See ${createDocumentationLink({
     return this;
   }
 
+  /**
+   * Removes a middleware from the InstantSearch lifecycle.
+   */
+  public unuse(...middlewareToUnuse: Middleware[]): this {
+    this.middleware
+      .filter(m => middlewareToUnuse.includes(m.middleware))
```
Haroenv, line 296: use `indexOf` here, `includes` doesn't exist on ie11

src/lib/InstantSearch.ts
```diff
@@ -263,14 +266,14 @@ See ${createDocumentationLink({
...
-   * minor versions.
    */
   public use(...middleware: Middleware[]): this {
     const newMiddlewareList = middleware.map(fn => {
       const newMiddleware = fn({ instantSearchInstance: this });
-      this.middleware.push(newMiddleware);
+      this.middleware.push({
+        middleware: fn,
+        middlewareDefinition: newMiddleware,
+      });
```
Haroenv, line 276:
```suggestion
      this.middleware.push([fn,newMiddleware]);
```

this will prevent the name `middlewareDefinition` in the output. Will only have a tiny bundlesize impact though, so not sure if worth it.

Haroenv (commented):
implementation is a bit confusing since we need reference to the creator function and its return value as well, so I don't see a significantly different way

src/lib/InstantSearch.ts
eunjae-lee: Yeah but we already use `includes` in InstantSearch.js and addressed in the guide: https://www.algolia.com/doc/guides/building-search-ui/installation/js/#browser-support

src/lib/InstantSearch.ts
eunjae-lee: I tried it and `68.06` → `68.05`, so 0.01kb diff. 

```js
  public unuse(...middlewareToUnuse: Middleware[]): this {
    this.middleware
      .filter(m => middlewareToUnuse.includes(m[0]))
      .forEach(m => m[1].unsubscribe());

    this.middleware = this.middleware.filter(
      m => !middlewareToUnuse.includes(m[0])
    );

    return this;
  }
```

it looks a little bit confusing with [0] and [1].

We can use different names for the object instead of `middleware & middlewareDefinition` if you have any idea, like `fn & middleware`?

src/lib/InstantSearch.ts
Haroenv: I didn't realise we required it to be polyfilled now, lgtm!

src/lib/InstantSearch.ts
Haroenv: what about `creator` and `instance`?

Haroenv: since it's internal, no big deal, but let's go with creator

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -1576,6 +1577,62 @@ describe('use', () => {
...
+    const middleware1 = jest.fn(() => middlewareSpy1);
+    const middlewareSpy2 = {
+      onStateChange: jest.fn(),
+      subscribe: jest.fn(),
+      unsubscribe: jest.fn(),
+    };
+    const middleware2 = jest.fn(() => middlewareSpy2);
+
+    search.addWidgets([
+      searchBoxOriginal({ container: document.createElement('div') }),
```
Haroenv, line 1602:
```suggestion
      connectSearchBox(() => {})(),
```

this is consistent with the other tests

Haroenv (approved):
not perfect, but lgtm with the test changed

Haroenv (approved):

PR #4711, opened 2021-04-01
feat(ts): convert hierarchical-menu to TypeScript

yannickcr:
**Summary**

Converts the `hierarchical-menu` to TypeScript


src/components/RefinementList/RefinementList.tsx
```diff
@@ -23,7 +24,7 @@ export type RefinementListProps<
   createURL: CreateURL<string>;
   cssClasses: TCSSClasses;
   depth?: number;
-  facetValues?: TRefinementListItem[];
+  facetValues?: TRefinementListItem[] | HierarchicalMenuItem[];
```
Haroenv, line 27: why does one have a T prefix and not the other?

src/widgets/hierarchical-menu/hierarchical-menu.tsx
```diff
@@ -3,26 +3,188 @@
...
 
+type HierarchicalMenuTemplates = {
+  /**
+   * Item template, provided with `name`, `count`, `isRefined`, `url` data properties.
+   */
+  item: Template;
+  /**
+   * Template used for the show more text, provided with `isShowingMore` data property.
+   */
+  showMoreText: Template;
```
Haroenv, line 39: data generics are missing here

src/widgets/hierarchical-menu/hierarchical-menu.tsx
```diff
@@ -3,26 +3,188 @@
...
+   * - Parent lvl0
+   *
+   * With `showParentLevel` set to `false`:
+   * - Parent lvl0
+   *   - **lvl1**
+   *     - **lvl2**
+   * - Parent lvl0
+   * - Parent lvl0
+   */
+  showMoreLimit?: number;
```
Haroenv, line 152: the description here is for `showParentLevel`, but the item is `showMoreLimit`

src/widgets/hierarchical-menu/hierarchical-menu.tsx
```diff
@@ -3,26 +3,188 @@
...
+  /**
+   * How to sort refinements. Possible values: `count|isRefined|name:asc|name:desc`.
+   * You can also use a sort function that behaves like the standard Javascript [compareFunction](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Syntax).
+   */
+  sortBy?:
+    | Array<'count' | 'isRefined' | 'name:asc' | 'name:desc'>
+    | (() => void);
```
Haroenv, line 159: SortBy helper type exists

Haroenv (commented):
overall pretty good, just some stuff that used to be wrong already that I noticed now :)

src/components/RefinementList/RefinementList.tsx
yannickcr: I has to rename the `RefinementListItem` type to `TRefinementListItem` since there is already the `RefinementListItem` component in this file. I did not rename the `HierarchicalMenuItem` type since there is no name conflict with it. I can apply the same naming to it to be more coherent.

src/components/RefinementList/RefinementList.tsx
```diff
@@ -310,7 +325,9 @@ class RefinementList<
     const facetValues = this.props.facetValues &&
       this.props.facetValues.length > 0 && (
         <ul className={this.props.cssClasses.list}>
-          {this.props.facetValues.map(this._generateFacetItem, this)}
+          {(this.props.facetValues as Array<
+            TRefinementListItem | HierarchicalMenuItem
+          >).map(this._generateFacetItem, this)}
```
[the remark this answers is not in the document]
yannickcr: There is an issue in TypeScript, when you try to iterate on an array with an union type you get the error "This expression is not callable." https://github.com/microsoft/TypeScript/issues/36390

You have to manually cast it to avoid the error (https://github.com/microsoft/TypeScript/issues/36390#issuecomment-641718624)

According to this https://github.com/microsoft/TypeScript/commit/b217f22e798c781f55d17da72ed099a9dee5c650 it should be fixed in v4.2.3 but we are using the v4.1.5 in the project.

src/connectors/hierarchical-menu/connectHierarchicalMenu.ts
```diff
@@ -5,45 +5,120 @@ import {
...
+   */
+  count: number;
+  /**
+   * Indicates if the refinement is applied.
+   */
+  isRefined: boolean;
+  /**
+   * n+1 level of items, same structure HierarchicalMenuItem
+   */
+  data?: HierarchicalMenuItem[] | null;
```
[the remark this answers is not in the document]
yannickcr: I was not sure so I left it optional, will fix it.

src/components/RefinementList/RefinementList.tsx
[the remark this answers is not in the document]
Haroenv:
```suggestion
          {
            // @ts-expect-error https://github.com/microsoft/TypeScript/commit/b217f22e798c781f55d17da72ed099a9dee5c650
            this.props.facetValues.map(this._generateFacetItem, this)
          }
```

something like this would fit best I think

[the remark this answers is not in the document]
Haroenv: or updating in a parallel PR already now

Haroenv (approved):
thanks again for this @yannickcr

eunjae-lee (approved):
Looks good to me!

PR #4712, opened 2021-04-01
chore(test): use real search-insights to test the insights middleware

eunjae-lee:
**Summary**

This PR adds the real `search-insights` library as a devDependency and uses it to test the `insights` middleware.
Before, we've been using a fake insightsClient which is just a shell with some mocked methods.
However the fake insightsClient mimiking the behavior of search-insights could lead to false positive situations.
Before we end up implementing all the logic into the fake insightsClient, we'd better use the real library.

**Result**

It works the same as before.


src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -350,7 +349,7 @@ describe('insights', () => {
           insightsClient,
         })({ instantSearchInstance });
         middleware.subscribe();
-        expect(getUserToken()).toEqual(ANONYMOUS_TOKEN);
+        expect(getUserToken()).toEqual('token-from-queue-before-init');
       });
```
eunjae-lee, line 353: This is technically a breaking change, but the previous behavior was wrong. If user calls `init` twice, the second `init` shouldn't override the previous usertoken with anonymous usertoken (ref: https://github.com/algolia/search-insights.js/pull/251)

package.json
```diff
@@ -136,6 +136,7 @@
     "rollup-plugin-replace": "2.2.0",
     "rollup-plugin-uglify": "6.0.4",
     "scriptjs": "2.5.9",
+    "search-insights": "https://pkg.csb.dev/algolia/search-insights.js/commit/0b3c26c5/search-insights",
```
eunjae-lee, line 139: The library is not released, so we have this temporary for now. We shouldn't merge this PR yet.

test/mock/createInsightsClient.ts
```diff
@@ -1,85 +1,58 @@
-export const ANONYMOUS_TOKEN = 'anonymous-user-id-1';
+import {
+  AlgoliaAnalytics,
+  getFunctionalInterface,
+  getRequesterForBrowser,
+  processQueue,
+} from 'search-insights/index-browser.cjs';
```
eunjae-lee, line 6: The updated `search-insights` now exports these functions, which enables our tests.

Haroenv: is createInsightsClient not enough for this test? it allows a custom requester to be set which here can be a fake one

eunjae-lee: You're right. We could use a custom requester here. Do you think we should copy&paste `getFunctionalInterface` and `processQueue` in InstantSearch.js instead?

Haroenv: good question, is there no way to cause those methods to be called from the outside somehow?

eunjae-lee: I don't see any way. They're [used internally](https://github.com/algolia/search-insights.js/blob/chore/wip/lib/entry-umd.ts#L6:L14). And normally people won't need `getFunctionalInterface` or `processQueue` unless they have a very specific usecase (which I can't think of). However here, we need them to control the flow in our test cases like

1. the queue is ready
2. test some stuff
3. let's say the library is loaded now. let's process the queue

Haroenv: I wonder if it would be best to manually set a queue similar to the browser script, and after a condition call aa = createInsightsClient() (which then immediately flushes, like in real life)?

eunjae-lee: We're already doing the queuing part here https://github.com/algolia/instantsearch.js/blob/fix/insights-test/test/mock/createInsightsClient.ts#L23:L29
Then I should just copy&paste `getFunctionalInterface` and `processQueue` over here, since we haven't seen the real usecase where people need access to those functions. Especially their implementations won't likely change in the future, so it looks okay to copy them.

eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4712/commits/d0e0fe5111682eeeba0d72e23ca130d2149482a8

@Haroenv what do you think? The only thing we need from `search-insights` is now `AlgoliaAnalytics`. I'll update the PR on that side.

Haroenv: possibly, I'm just not fully sure what exactly the needs are here, personally I'd go for using just the things that are already exported

eunjae-lee: https://github.com/algolia/instantsearch.js/blob/fix/insights-test/src/middlewares/__tests__/createInsightsMiddleware.ts#L157:L159

In the test cases we need to test things that happen before the umd library is loaded, and after loaded. To control its timing, I had to make `libraryLoadedAndProcessQueue` function which is basically calling `processQueue`. Does it make sense? Let me know if you don't understand any part or have different opinion on this.

test/mock/createInsightsClient.ts
```diff
@@ -1,85 +1,89 @@
...
+  return analytics;
+}
+
+function getFunctionalInterface(instance: AlgoliaAnalytics) {
+  return (functionName: string, ...functionArguments: any[]) => {
+    instance[functionName](...functionArguments);
+  };
+}
+
+function processQueue(instance: AlgoliaAnalytics, globalObject: any) {
```
tkrugg, line 61: In the same spirit of now having to rewrite insights logic here, shouldn't we get some of theses fonctions from the lib ? https://github.com/algolia/search-insights.js/blob/6c58e24abdc8cb7cb334b067a743619c28469db4/lib/_processQueue.ts

tkrugg (commented):
this is great 👍 
Left one comment, I'm thinking maybe there are more part of the library we can use off the source instead of mocking them here: eg processQueue, getFunctionInterface

test/mock/createInsightsClient.ts
tkrugg: `processQueue` and `getFunctionalInterface` are private APIs but actually exported (for internal use). I think I would just use them instead of copy pasting them.

eunjae-lee:
rel: https://github.com/algolia/search-insights.js/pull/252#issuecomment-814915802

I'm trying to import directly from the source. I need to do something with the jest configuration. Applying many different suggestions online is not working well 😅 I might try more.

Haroenv (approved):
this is perfect!

Haroenv:
ts seems to fail because of things in node_modules. possibly fixable like this? https://github.com/microsoft/TypeScript/issues/17042#issuecomment-317589758 (considering we likely don't want to check here), or is it possible to do skipLibCheck for imports from within tests only?

eunjae-lee:
> ts seems to fail because of things in node_modules. possibly fixable like this? [microsoft/TypeScript#17042 (comment)](https://github.com/microsoft/TypeScript/issues/17042#issuecomment-317589758) (considering we likely don't want to check here), or is it possible to do skipLibCheck for imports from within tests only?

it's probably about jest + ts + esm combination.

https://github.com/facebook/jest/issues/9860
https://github.com/facebook/jest/issues/9395

eunjae-lee:
I've tried

```js
// jest.config.js

  transform: {
    '<rootDir>/scripts/jest/setupTests.ts': 'ts-jest',
    'node_modules/search-insights/.+\\.(j|t)sx?$': 'ts-jest',
  },
  transformIgnorePatterns: [
    '<rootDir>/scripts/jest/setupTests.ts',
    'node_modules/search-insights/.+\\.(j|t)sx?$',
  ],
```

```sh
yarn test --testPathPattern src/middlewares/__tests__/createInsightsMiddleware.ts
```

and got 

```
    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /Users/eunjaelee/workspace/instantsearch.js/scripts/jest/setupTests.ts:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import Enzyme from 'enzyme';
                                                                                             ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1059:14)
          at Array.forEach (<anonymous>)
```

I'm stuck with setupTests.ts` file at the moment.

Haroenv:
seeing the ts fail, I wonder if skipLibFiles should be true

tsconfig.json
```diff
@@ -19,5 +19,10 @@
     "allowSyntheticDefaultImports": true,
     "skipLibCheck": true
   },
-  "exclude": ["examples", "es"]
+  "exclude": [
+    "examples",
+    "es",
+    "src/middlewares/__tests__/createInsightsMiddleware.ts",
+    "test/mock/createInsightsClient.ts"
```
eunjae-lee, line 26: This test file imports `createInsightsClient` from `test/mock/createInsightsClient.ts` and it imports it from `node_modules/search-insights`.

Unless we exclude these two files, `tsc` will type-check `search-insights` and fail.
I tried to avoid it without having to exclude the test file here, but

- https://github.com/Microsoft/TypeScript/issues/14293
- https://github.com/microsoft/TypeScript/issues/36017
- https://www.reddit.com/r/typescript/comments/747xz7/suppress_errors_in_node_modules/

> "exclude" only prevents items from being included by "include"; it doesn't prevent them from being included via import statements or <reference>s. If you import levelup, and levelup imports leveldown, then excluding leveldown or node_modules won't have an effect.

couldn't find a way.

Haroenv: is skipLibCheck really not an option?

Haroenv: or fixing the TS of search-insights not to have those issues?

eunjae-lee: > skipLibCheck

It's already turned on in our tsconfig but it doesn't affect here. (tried to pass it via CLI too)

> or fixing the TS of search-insights not to have those issues?

We could try. But, running `tsc` in `search-insights` already tells me everything's fine probably because we have different tsconfig and different version of tsc.

The possible next step would be to figure out the way to modify either the source or tsconfig of search-insights so that it works fine in InstantSearch.js without having to exclude these.

eunjae-lee: To be clear, what I meant by "The possible next step" was in a separate PR. I'd like to get this PR merged with no great amount of further investment at the moment. We can revisit and do the "next step" later. cc @Haroenv @tkrugg

Haroenv: unfortunately I don't think we can merge a PR with so many failing tests

eunjae-lee: Oh, I didn't notice that. I'll take a look.

package.json
```diff
@@ -136,6 +136,7 @@
     "rollup-plugin-replace": "2.2.0",
     "rollup-plugin-uglify": "6.0.4",
     "scriptjs": "2.5.9",
+    "search-insights": "https://pkg.csb.dev/algolia/search-insights.js/commit/761a3f68/search-insights",
```
Haroenv, line 139: lgtm pending this :)

eunjae-lee: updated https://github.com/algolia/instantsearch.js/pull/4712/commits/2ad3cc07fda94ebe7a5662a3097228ddae9e6758

tkrugg (approved):

Haroenv (approved):

PR #4724, opened 2021-04-06
fix(insights): do not throw when userToken is not given

eunjae-lee:
**Summary**

This PR prevents `insights` middleware from throwing when `userToken` is not given yet.
_(The previous description wasn't clear and I was confused myself too.)_

Adding the `insights` middleware to InstantSearch and assigning a userToken could happen at the same time, or not. For example, although InstantSearch is loaded, the user is not logged in, so there may not be a userToken to set (let's assume this customer is not using anonymous user token from cookie).

Without this PR, we can guide like

> Add the middleware only after you have a user token. Otherwise, the middleware will throw errors without the token. Oh, and when a user logs out and you want to call `aa('setUserToken', '')` to remove the user token, you need to unuse the middleware. Otherwise it will throw errors because there is no user token.

But after this PR, I'd like to guide like

> Add the insights middleware once. Whenever you have a user token, call `aa('setUserToken', 'something');`. If you want to remove the userToken, then do so. The middleware will stop sending events.

---

~This PR fixes the `insights` middleware not to throw when `userToken` is not given yet.
It's coming from the behavior of `search-insights`, but we might still allow customers to use the `insights` middleware without setting userToken. For example, they might want to leverage the `insights` middleware to send events to third-party trackers, but not necessarily to Algolia directly. In that case `userToken` for Algolia is not needed, but they still want the middleware to work.~

~Currently this PR is based on https://github.com/algolia/instantsearch.js/pull/4712, but I will rebase it on master once it's merged. So this PR shouldn't be merged until then.~

### DIFF
--- a/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -8,6 +8,7 @@ import {
 } from '../../../test/mock/createInsightsClient';
 import { warning } from '../../lib/utils';
 import { SearchClient } from '../../types';
+import { runAllMicroTasks } from '../../../test/utils/runAllMicroTasks';
 
 describe('insights', () => {
   const createTestEnvironment = () => {
@@ -103,9 +104,8 @@ describe('insights', () => {
       });
     });
 
-    it('does not throw when an event is sent right after the creation in UMD', () => {
+    it('does not throw without userToken in UMD (1)', () => {
       const {
-        analytics,
         insightsClient,
         libraryLoadedAndProcessQueue,
         instantSearchInstance,
@@ -113,6 +113,7 @@ describe('insights', () => {
 
       const middleware = createInsightsMiddleware({
         insightsClient,
+        insightsInitParams: { useCookie: false },
       })({ instantSearchInstance });
       middleware.subscribe();
 
@@ -127,22 +128,72 @@ describe('insights', () => {
         },
         widgetType: 'ais.hits',
       });
-      expect(analytics.viewedObjectIDs).toHaveBeenCalledTimes(0);
 
-      // But, the library hasn't been loaded yet, so the event stays in the queue.
-      expect(insightsClient.queue[insightsClient.queue.length - 1]).toEqual([
-        'viewedObjectIDs',
-        { eventName: 'Hits Viewed', index: '', objectIDs: ['1', '2'] },
-      ]);
+      // When the library is loaded later, it consumes the queue and sends the event.
+      expect(() => {
+        libraryLoadedAndProcessQueue();
+      }).not.toThrow(
+        "Before calling any methods on the analytics, you first need to call 'setUserToken' function or include 'userToken' in the event payload."
+      );
+    });
+
+    it('does not throw without userToken in UMD (2)', () => {
+      const {
+        insightsClient,
+        libraryLoadedAndProcessQueue,
+        instantSearchInstance,
+      } = createUmdTestEnvironment();
+
+      const middleware = createInsightsMiddleware({
+        insightsClient,
+        insightsInitParams: { useCookie: false },
+      })({ instantSearchInstance });
+      middleware.subscribe();
 
       // When the library is loaded later, it consumes the queue and sends the event.
       libraryLoadedAndProcessQueue();
-      expect(analytics.viewedObjectIDs).toHaveBeenCalledTimes(1);
-      expect(analytics.viewedObjectIDs).toHaveBeenCalledWith({
-        eventName: 'Hits Viewed',
-        index: '',
-        objectIDs: ['1', '2'],
-      });
+
+      expect(() => {
+        // It tries to send an event.
+        instantSearchInstance.sendEventToInsights({
+          eventType: 'view',
+          insightsMethod: 'viewedObjectIDs',
+          payload: {
+            eventName: 'Hits Viewed',
+            index: '',
+            objectIDs: ['1', '2'],
+          },
+          widgetType: 'ais.hits',
+        });
+      }).not.toThrow(
+        "Before calling any methods on the analytics, you first need to call 'setUserToken' function or include 'userToken' in the event payload."
+      );
+    });
+
+    it('does not throw without userToken in CJS', () => {
+      const { insightsClient, instantSearchInstance } = createTestEnvironment();
+
+      const middleware = createInsightsMiddleware({
+        insightsClient,
+        insightsInitParams: { useCookie: false },
+      })({ instantSearchInstance });
+      middleware.subscribe();
+
+      expect(() => {
+        // It tries to send an event.
+        instantSearchInstance.sendEventToInsights({
+          eventType: 'view',
+          insightsMethod: 'viewedObjectIDs',
+          payload: {
+            eventName: 'Hits Viewed',
+            index: '',
+            objectIDs: ['1', '2'],
+          },
+          widgetType: 'ais.hits',
+        });
+      }).not.toThrow(
+        "Before calling any methods on the analytics, you first need to call 'setUserToken' function or include 'userToken' in the event payload."
+      );
     });
 
     it('applies clickAnalytics', () => {
--- a/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -194,6 +194,31 @@ describe('insights', () => {
       );
     });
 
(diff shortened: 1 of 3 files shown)


francoischalifour:
> For example, they might want to leverage the `insights` middleware to send events to third-party trackers, but not necessarily to Algolia directly. In that case `userToken` for Algolia is not needed, but they still want the middleware to work.

Seeing the code, this means that they'll still see a warning when using the middleware with a non-Algolia analytics integration. Should we not rather explain (e.g., with a guide) how to create a middleware to plug their own analytics solution?

eunjae-lee:
> Seeing the code, this means that they'll still see a warning when using the middleware with a non-Algolia analytics integration. Should we not rather explain (e.g., with a guide) how to create a middleware to plug their own analytics solution?

If they use the `insights` middleware for non-Algolia analytics integration, they must've had `onEvent` hook and go controlled mode. Then they don't end up going into that branch: https://github.com/algolia/instantsearch.js/pull/4724/files#diff-d934c60fefaa23618441fb40ed23c85dde8006e7661144b0434add680a996f62R125-R127

eunjae-lee:
↑ @francoischalifour @tkrugg

francoischalifour:
If they do provide `onEvent`, it will never call `insightsClient` (https://github.com/algolia/instantsearch.js/pull/4724/files#diff-d934c60fefaa23618441fb40ed23c85dde8006e7661144b0434add680a996f62L127) and therefore not throw without `userToken`, right?

I'm not sure to understand why this PR exactly.

eunjae-lee:
@francoischalifour you're right!
You've made me realized that users will have no exception at all when

- they didn't set userToken
- & they provide `onEvent`
- & they don't send events with `insightsClient`

Thanks for pointing it out! Now we don't need this.

eunjae-lee:
pushed a commit but it's not recognized this PR, although the commit is in the remote :(

eunjae-lee:
I've reopened this PR and updated the description above @francoischalifour @tkrugg

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -126,7 +127,14 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = props => {
...
             onEvent(event, _insightsClient);
           } else if (event.insightsMethod) {
-            insightsClient(event.insightsMethod, event.payload);
+            if (hasUserToken()) {
+              insightsClient(event.insightsMethod, event.payload);
+            } else {
+              warning(
+                false,
+                'Cannot send event to Algolia Insights because `userToken` is not set.'
+              );
```
francoischalifour, line 136: Why don't we test that this warning is correctly displayed?

src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -103,16 +103,16 @@ describe('insights', () => {
       });
     });
 
-    it('does not throw when an event is sent right after the creation in UMD', () => {
+    it('does not throw without userToken in UMD (1)', () => {
```
francoischalifour, line 106: What are these `(1)` and `(2)` suffixes for? Can we name the tests based on the scenario?

eunjae-lee: yup https://github.com/algolia/instantsearch.js/pull/4724/commits/3b7f5674611358ab0b75391036adae959e83a03d

src/middlewares/createInsightsMiddleware.ts
eunjae-lee: added https://github.com/algolia/instantsearch.js/pull/4724/commits/8c489dd5aebe325d3d10a88b974b1b5f53db52d0

src/types/insights.ts
```diff
@@ -46,7 +46,8 @@ export type InsightsClient = InsightsSendEvent &
   InsightsOnUserTokenChange &
   InsightsGetUserToken &
   InsightsInit &
-  InsightsSetUserToken & {
+  InsightsSetUserToken &
+  InsightsGetUserToken & {
```
francoischalifour, line 50: Isn't this a duplicate with line 47?

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -126,7 +127,14 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = props => {
           if (onEvent) {
             onEvent(event, _insightsClient);
           } else if (event.insightsMethod) {
-            insightsClient(event.insightsMethod, event.payload);
+            if (hasUserToken()) {
+              insightsClient(event.insightsMethod, event.payload);
+            } else {
+              warning(
+                false,
+                'Cannot send event to Algolia Insights because `userToken` is not set.'
```
francoischalifour, line 135: Can we add an actionable doc link that teaches users how to solve this?

src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -127,22 +127,96 @@ describe('insights', () => {
...
-      expect(analytics.viewedObjectIDs).toHaveBeenCalledTimes(1);
-      expect(analytics.viewedObjectIDs).toHaveBeenCalledWith({
-        eventName: 'Hits Viewed',
-        index: '',
-        objectIDs: ['1', '2'],
-      });
+
+      expect(() => {
+        // It tries to send an event.
+        instantSearchInstance.sendEventToInsights({
```
francoischalifour, line 156: It's out of scope for this PR but we're testing with internals here. It would be less fragile to actually mount a widget and interact with it to trigger an event.

src/middlewares/createInsightsMiddleware.ts
eunjae-lee: https://www.algolia.com/doc/guides/building-search-ui/going-further/send-insights-events/js/#setting-the-usertoken

This might be the doc, but it's for js and this code will be used in Vue InstantSearch, too. What do you think? It's okay to just add the js version of the doc?

src/types/insights.ts
eunjae-lee: Good catch https://github.com/algolia/instantsearch.js/pull/4724/commits/ad13acb2b833d045ba3dbc724f01c2637061511b

src/middlewares/__tests__/createInsightsMiddleware.ts
eunjae-lee: Yes, I agree. I'll convert some unit tests to integration tests in the near future.

src/middlewares/createInsightsMiddleware.ts
eunjae-lee: https://github.com/algolia/instantsearch.js/pull/4724/commits/b0245fa51ca7434ac2093464061bf66401e2931d

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -132,7 +132,11 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = props => {
             } else {
               warning(
                 false,
-                'Cannot send event to Algolia Insights because `userToken` is not set.'
+                `
+Cannot send event to Algolia Insights because \`userToken\` is not set.
+
+See https://www.algolia.com/doc/guides/building-search-ui/going-further/send-insights-events/js/#setting-the-usertoken
```
francoischalifour, line 138: That's [how we display our links for documentation](https://github.com/algolia/instantsearch.js/blob/1a939c0279160a5339ec8e86a668f1e46fc3b27d/src/lib/utils/documentation.ts#L28):

```suggestion
See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-further/send-insights-events/js/#setting-the-usertoken
```

src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -215,7 +215,9 @@ describe('insights', () => {
           widgetType: 'ais.hits',
         })
       ).toWarnDev(
-        '[InstantSearch.js]: Cannot send event to Algolia Insights because `userToken` is not set.'
+        `[InstantSearch.js]: Cannot send event to Algolia Insights because \`userToken\` is not set.
+
+See https://www.algolia.com/doc/guides/building-search-ui/going-further/send-insights-events/js/#setting-the-usertoken`
```
francoischalifour, line 220:
```suggestion
See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-further/send-insights-events/js/#setting-the-usertoken`
```

francoischalifour (approved):
Good to merge after this last change!

PR #4726, opened 2021-04-08
chore(ci): consistent modern yarn version

Haroenv:
**Summary**

So we have these packages in our package.json as devDependencies:

    "algoliasearch": "4.8.6",
    "algoliasearch-v3": "npm:algoliasearch@3.35.1",

This works fine for most use cases, when we import the package as "algoliasearch-v3" it resolves as expected to v3. However we also have another package depending itself on 3.35.1 in devDeps too:

    "places.js": "1.17.1",

It has 3.35.1 as dependency:

    "algoliasearch": "^3.35.1",

And thus imports it as "algoliasearch".

This worked fine and i have this in yarn.lock:

    "algoliasearch-v3@npm:algoliasearch@3.35.1":
      name algoliasearch-v3
      version "3.35.1"
      ...

    algoliasearch@^3.35.1:
      version "3.35.1"
      ...

    algoliasearch@4.8.6:
      version "4.8.6"
      ...

However now when i add any package (or yarn --force), it changes yarn.lock to this:

    "algoliasearch-v3@npm:algoliasearch@3.35.1", algoliasearch@^3.35.1:
      version "3.35.1"
      ...

    algoliasearch@4.8.6:
      version "4.8.6"
      ...

This seems fine, although the node_modules folder created for that is node_modules/algoliasearch-v3 alone, inside node_modules/places.js there's no other node_modules for the algoliasearch with the right name.

This can be tried out in the master branch by running yarn (right node_modules, with a algoliasearch inside places.js), and then failing by running yarn --force. when you're on yarn 1.13 like the netlify setup.

**Result**

having a new enough yarn version automatically used in all tools will fix this issue. It can be tested by verifying that the netlify build will pass in this branch (unlike for example https://github.com/algolia/instantsearch.js/pull/4720)


eunjae-lee (approved):
This is awesome!

PR #4728, opened 2021-04-08
fix(ratingMenu): use url in default template

Haroenv:
**Summary**

this is the cause of the from time to time failing e2e test, basically sometimes the event.preventDefault didn't apply, which causes the page to refresh to / instead of the right url.

The actual url is `url`, and not in `href`, so this means that before now it always was wrong.

The docs are correct and use `url`: https://www.algolia.com/doc/api-reference/widgets/rating-menu/js/#templates

**Result**

url in ratingMenu now is actually correct, e2e tests will pass


francoischalifour (approved):
Good catch

PR #4732, opened 2021-04-13
refactor(types): split up widget.ts

Haroenv:
**Summary**

the types in Widget.ts aren't all directly related to widgets, which makes it very hard to read

**Result**

more smaller files

### DIFF
--- a/src/types/widget.ts
+++ b/src/types/widget.ts
@@ -3,101 +3,10 @@ import {
   AlgoliaSearchHelper as Helper,
   SearchParameters,
   SearchResults,
-  PlainSearchParameters,
 } from 'algoliasearch-helper';
 import { InstantSearch } from './instantsearch';
-import { Hit, GeoLoc } from './results';
-import { BindEventForHits, SendEventForHits } from '../lib/utils';
-import {
-  AutocompleteRendererOptions,
-  AutocompleteConnectorParams,
-} from '../connectors/autocomplete/connectAutocomplete';
-import {
-  BreadcrumbRendererOptions,
-  BreadcrumbConnectorParams,
-} from '../connectors/breadcrumb/connectBreadcrumb';
-import {
-  ClearRefinementsRendererOptions,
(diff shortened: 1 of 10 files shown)


src/types/widget.ts
```diff
@@ -473,7 +172,9 @@ export type Widget<
    */
   getWidgetSearchParameters?(
     state: SearchParameters,
-    widgetSearchParametersOptions: WidgetSearchParametersOptions
+    widgetSearchParametersOptions: {
```
eunjae-lee, line 175: Any reason you removed `WidgetSearchParametersOptions` type and access it via `Parameters` later?

eunjae-lee (approved):
Awesome. Thanks for cleaning it up!

src/types/widget.ts
Haroenv: the type later (PR to come) depends on the generic passed to Widget, so the WidgetSearchParametersOptions would need the generic too. I might revisit this part once its migrated fully!

francoischalifour (commented):
A convention that I see more often is to name the type files like the type.

Example: `UiState.ts`

Haroenv:
What if it exports more than one item @francoischalifour? for example `instantsearch` exports `InstantSearch` & `InstantSearchOptions` etc. Case-only name changes are also a little tricky on git

francoischalifour:
@Haroenv We could either export `InstantSearchOptions` from `InstantSearch.ts` (since it's closely related), or create another `InstantSearchOptions.ts` file.

Haroenv:
In https://github.com/algolia/instantsearch.js/pull/4734 I'm adding different exports in most of those files (IndexUiState / UiState in ui-state for example). Do you think those should be in their own file or something IndexUiState should be exported from UiState.ts?

francoischalifour:
I think it's OK for `IndexUiState` to be exported from `UiState.ts`.

tkrugg (approved):

PR #4734, opened 2021-04-15
refactor(types): new WidgetDescription setup

Haroenv:
**Summary**

sets up Widget, Connector and WidgetFactory to accept a `WidgetDescription` generically instead of the TRenderState now

**Result**

note: CI will be failing until the final PR here (panel, #4738)

### DIFF
--- a/src/types/connector.ts
+++ b/src/types/connector.ts
@@ -1,14 +1,14 @@
 import { SearchResults } from 'algoliasearch-helper';
 import { InstantSearch } from './instantsearch';
 import { InsightsClient } from './insights';
-import { WidgetFactory } from './widget-factory';
 import { Hits } from './results';
+import { Widget, WidgetDescription } from './widget';
 
 /**
  * The base renderer options. All render functions receive
  * the options below plus the specific options per connector.
  */
-export type RendererOptions<TWidgetParams> = {
+export type RenderState<TWidgetParams> = {
   /**
    * The original widget params. Useful as you may
    * need them while using the render function.
@@ -40,11 +40,11 @@ export type RendererOptions<TWidgetParams> = {
 /**
  * The render function.
  */
-export type Renderer<TRenderOptions, TWidgetParams> = (
+export type Renderer<TRenderState, TWidgetParams> = (
   /**
    * The base render options plus the specific options of the widget.
    */
-  renderOptions: TRenderOptions & RendererOptions<TWidgetParams>,
+  renderState: TRenderState & RenderState<TWidgetParams>,
 
   /**
    * If is the first run.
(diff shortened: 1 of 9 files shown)


src/types/connector.ts
```diff
@@ -1,14 +1,14 @@
...
-import { WidgetFactory } from './widget-factory';
 import { Hits } from './results';
+import { Widget, WidgetDescription } from './widget';
 
 /**
  * The base renderer options. All render functions receive
  * the options below plus the specific options per connector.
  */
-export type RendererOptions<TWidgetParams> = {
+export type RenderState<TWidgetParams> = {
```
eunjae-lee, line 11: Makes so much sense!

src/types/widget.ts
```diff
@@ -110,102 +111,167 @@ export type BuiltinWidgetTypes =
...
+   * Called once before the first search.
    */
-  init?(options: InitOptions): void;
+  init?: (options: InitOptions) => void;
   /**
-   * Called after each search response has been received
+   * Called after each search response has been received.
    */
-  render?(options: RenderOptions): void;
+  render?: (options: RenderOptions) => void;
```
eunjae-lee, line 139: A widget must have at least one of [init, render]. What if we extract this part and use `AtLeastOne` type util?

eunjae-lee (commented):
Looks great! Thanks Haroen.

src/types/widget.ts
Haroenv: I thought about that, but in practice that would make the type more complicated for few gain, since $$type is required now, and it doesn't really make the usage simpler (will still need to use `init!()`, since you can't know which one of the two is actually implemented). Instead I think we should in a major version make render required both here and in addWidgets;

eunjae-lee: Oh yeah, it makes sense!

eunjae-lee (approved):
Looks good to me!

src/types/render-state.ts
```diff
@@ -1,223 +1,68 @@
...
+import { RelevantSortWidgetDescription } from '../connectors/relevant-sort/connectRelevantSort';
+import { SearchBoxWidgetDescription } from '../connectors/search-box/connectSearchBox';
+import { SortByWidgetDescription } from '../connectors/sort-by/connectSortBy';
+import { StatsWidgetDescription } from '../connectors/stats/connectStats';
+import { ToggleRefinementWidgetDescription } from '../connectors/toggle-refinement/types';
+import { VoiceSearchWidgetDescription } from '../connectors/voice-search/connectVoiceSearch';
+import { AnalyticsWidgetDescription } from '../widgets/analytics/analytics';
+import { PlacesWidgetDescription } from '../widgets/places/places';
+
+type ConnectorRenderStates = AnswersWidgetDescription['indexRenderState'] &
```
tkrugg, line 29: shouldn't we use the Expand helper here?

src/types/widget.ts
```diff
@@ -110,102 +111,167 @@ export type BuiltinWidgetTypes =
...
+
+type RequiredWidgetType<TWidgetDescription extends WidgetDescription> = {
+  /**
+   * Identifier for widgets.
+   */
+  $$widgetType: TWidgetDescription['$$widgetType'];
+};
+
+type WidgetType<
+  TWidgetDescription extends WidgetDescription
```
tkrugg, line 155: why not use WidgetDescription directly here?

tkrugg (approved):

src/types/render-state.ts
Haroenv: there's no conditionals, so intellisense automatically expanded the type already :)

src/types/widget.ts
Haroenv: because the actual value of the description is needed, so it needs to be a generic :)

PR #4735, opened 2021-04-15
refactor(ts): migrate connectors to new Widget type

Haroenv:
**Summary**

migrates all the connectors to the new type from #4734 

**Result**

fixes some smaller things in the tests, plus implementation of refinementList (removed dead code)

- connectToggleRefinement: nest `getRenderState` per attribute
- connectToggleRefinement: remove `state` state from getWidgetRenderState
- connectRefinementList: remove dead code in getWidgetRenderState
- connectRelevantSort: remove "relevantSort" nesting, since there's only one property

These are technically breaking changes, but small enough to just be listed in the changeling IMO

### DIFF
--- a/src/connectors/answers/connectAnswers.ts
+++ b/src/connectors/answers/connectAnswers.ts
@@ -14,6 +14,7 @@ import {
   Hit,
   FindAnswersOptions,
   FindAnswersResponse,
+  WidgetRenderState,
 } from '../../types';
 
 type IndexWithAnswers = {
@@ -31,7 +32,7 @@ const withUsage = createDocumentationMessageGenerator({
   connector: true,
 });
 
-export type AnswersRendererOptions = {
+export type AnswersRenderState = {
   /**
    * The matched hits from Algolia API.
    */
@@ -89,8 +90,16 @@ export type AnswersConnectorParams = {
   extraParameters?: FindAnswersOptions;
 };
 
+export type AnswersWidgetDescription = {
+  $$type: 'ais.answers';
+  renderState: AnswersRenderState;
+  indexRenderState: {
+    answers: WidgetRenderState<AnswersRenderState, AnswersConnectorParams>;
+  };
+};
+
 export type AnswersConnector = Connector<
-  AnswersRendererOptions,
+  AnswersWidgetDescription,
   AnswersConnectorParams
 >;
 
--- a/src/connectors/refinement-list/connectRefinementList.ts
+++ b/src/connectors/refinement-list/connectRefinementList.ts
@@ -1,8 +1,4 @@
-import {
-  AlgoliaSearchHelper,
-  SearchForFacetValues,
-  SearchResults,
-} from 'algoliasearch-helper';
+import { AlgoliaSearchHelper, SearchResults } from 'algoliasearch-helper';
 import {
   escapeFacets,
   TAG_PLACEHOLDER,
@@ -20,9 +16,9 @@ import {
   RenderOptions,
   Widget,
   InitOptions,
-  WidgetRenderState,
   FacetHit,
   CreateURL,
+  WidgetRenderState,
 } from '../../types';
 
 const withUsage = createDocumentationMessageGenerator({
@@ -51,10 +47,6 @@ export type RefinementListItem = {
    * Indicates if the list item is refined.
    */
   isRefined: boolean;
-  /**
-   * List item childrens.
-   */
-  data?: RefinementListItem[];
 };
 
 export type RefinementListConnectorParams = {
@@ -94,7 +86,7 @@ export type RefinementListConnectorParams = {
   transformItems?: TransformItems<RefinementListItem>;
 };
 
-export type RefinementListRendererOptions = {
+export type RefinementListRenderState = {
   /**
    * The list of filtering values returned from Algolia API.
    */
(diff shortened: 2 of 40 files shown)


src/connectors/answers/connectAnswers.ts
```diff
@@ -89,8 +90,16 @@ export type AnswersConnectorParams = {
   extraParameters?: FindAnswersOptions;
 };
 
+export type AnswersWidgetDescription = {
+  $$type: 'ais.answers';
+  renderState: AnswersRenderState;
+  indexRenderState: {
+    answers: WidgetRenderState<AnswersRenderState, AnswersConnectorParams>;
```
Haroenv, line 97:
```suggestion
    answers: AnswersRenderState;
```

I wish it could just be this, and then augmented with the parameters when you use it (eg. in Widget['getRenderState']), but I can't get it to work, so for now, the index render state only types the connector params, not the connector & widget params. I think this is a reasonable tradeoff for now, considering the use cases for reading the widget params from index render state are likely slim (and can use an `as any`)

src/connectors/relevant-sort/__tests__/connectRelevantSort-test.ts
```diff
@@ -277,11 +277,11 @@ describe('connectRelevantSort', () => {
 
       expect(
         widget.getWidgetUiState!(
-          { relevantSort: { relevancyStrictness: 25 } },
+          { relevantSort: 25 },
           { helper, searchParameters: helper.state }
         )
       ).toEqual({
-        relevantSort: { relevancyStrictness: undefined },
+        relevantSort: 25,
```
Haroenv, line 284: I don't understand why it was 25 and not undefined before, this seems like a fix

src/connectors/relevant-sort/connectRelevantSort.ts
```diff
@@ -109,17 +123,15 @@ const connectRelevantSort: RelevantSortConnector = function connectRelevantSort(
...
 
       getWidgetUiState(uiState, { searchParameters }) {
         return {
           ...uiState,
-          relevantSort: {
-            ...uiState.relevantSort,
-            relevancyStrictness: searchParameters.relevancyStrictness,
-          },
+          relevantSort:
+            searchParameters.relevancyStrictness || uiState.relevantSort,
```
Haroenv, line 134: uiState change is technically a breaking change, but small enough to just mention in changelog

src/connectors/toggle-refinement/connectToggleRefinement.js
```diff
@@ -218,7 +218,10 @@ export default function connectToggleRefinement(renderFn, unmountFn = noop) {
       getRenderState(renderState, renderOptions) {
         return {
           ...renderState,
-          toggleRefinement: this.getWidgetRenderState(renderOptions),
+          toggleRefinement: {
+            ...renderState.toggleRefinement,
+            [attribute]: this.getWidgetRenderState(renderOptions),
+          },
```
Haroenv, line 224: uiState change is technically a breaking change, but small enough to just mention in changelog

src/connectors/refinement-list/connectRefinementList.ts
```diff
@@ -385,25 +384,13 @@ const connectRefinementList: RefinementListConnector = function connectRefinemen
         }
 
         if (results) {
-          if (!isFromSearch) {
```
Haroenv, line 388: this code was unreachable, isFromSearch is never passed to getWidgetRenderState

src/types/widget.ts
```diff
@@ -155,7 +155,12 @@ type WidgetType<
   TWidgetDescription extends WidgetDescription
 > = TWidgetDescription extends RequiredKeys<WidgetDescription, '$$widgetType'>
   ? RequiredWidgetType<TWidgetDescription>
-  : Partial<RequiredWidgetType<TWidgetDescription>>;
+  : {
+      /**
+       * Identifier for widgets.
+       */
+      $$widgetType?: string;
+    };
```
Haroenv, line 163: previous implementation sometimes caused infinite loops

src/connectors/refinement-list/connectRefinementList.ts
```diff
@@ -301,20 +304,17 @@ const connectRefinementList: RefinementListConnector = function connectRefinemen
...
-
               renderFn(
                 {
                   ...widget.getWidgetRenderState({
                     ...renderOptions,
                     results: lastResultsFromMainSearch,
                   }),
                   items: normalizedFacetValues,
-                  canToggleShowMore,
+                  canToggleShowMore: false,
```
eunjae-lee, line 314: How do you know it's `false`?

src/connectors/relevant-sort/__tests__/connectRelevantSort-test.ts
eunjae-lee: When the helper state doesn't include appliedRelevancyStrictness at all, then the widget shouldn't, either. I guess that's what this test is about.

src/connectors/answers/connectAnswers.ts
eunjae-lee: Looks good to me. We could improve this part later if we find a solution.

eunjae-lee (commented):
Thanks for the work!

src/connectors/refinement-list/connectRefinementList.ts
Haroenv: show more doesn't display when in search for facet values, it can't work (would need to do a new search). If you call the method today the query will disappear

src/connectors/relevant-sort/__tests__/connectRelevantSort-test.ts
Haroenv: but if the helper state doesn't have relevancy strictness, but the ui state (route) does have it, doesn't that apply the strictness in getWidgetSearchParameters?

src/connectors/refinement-list/connectRefinementList.ts
eunjae-lee: Ah it makes sense.

src/connectors/relevant-sort/connectRelevantSort.ts
shortcuts: `|| uiState.relevantSort`

Does this mean we can now set the `relevantSort` from the frontend?

shortcuts (approved):

PR #4736, opened 2021-04-15
refactor(ts): update index with new type

Haroenv:
**Summary**

updates the index widget

**Result**

because the types don't match, Index no longer is a Widget, which has impact in addWidget / removeWidget / getWidget

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -339,7 +339,7 @@ See ${createDocumentationLink({
    * Widgets can be added either before or after InstantSearch has started.
    * @param widgets The array of widgets to add to InstantSearch.
    */
-  public addWidgets(widgets: Widget[]) {
+  public addWidgets(widgets: Array<Widget | Index>) {
     if (!Array.isArray(widgets)) {
       throw new Error(
         withUsage(
@@ -374,7 +374,7 @@ See ${createDocumentationLink({
    *
    * The widget must implement a `dispose()` method to clear its state.
    */
-  public removeWidget(widget: Widget) {
+  public removeWidget(widget: Widget | Index) {
     warning(
       false,
       'removeWidget will still be supported in 4.x releases, but not further. It is replaced by `removeWidgets([widget])`'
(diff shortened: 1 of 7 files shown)


src/lib/InstantSearch.ts
```diff
@@ -339,7 +339,7 @@ See ${createDocumentationLink({
    * Widgets can be added either before or after InstantSearch has started.
    * @param widgets The array of widgets to add to InstantSearch.
    */
-  public addWidgets(widgets: Widget[]) {
+  public addWidgets(widgets: Array<Widget | Index>) {
```
eunjae-lee, line 342: This gives me the impression that `index` is not a widget at all, whereas in fact it's a sort of special widget.

```suggestion
  public addWidgets(widgets: Array<Widget | IndexWidget>) {
```

I think it aligns better with other widgets like `HitsWidget, RefinementListWidget, ...` and also with `isIndexWidget()` function.

eunjae-lee (approved):
Looks clear to me.

src/lib/InstantSearch.ts
Haroenv: makes sense, let's do this while it's not yet a breaking change

src/widgets/index/index.ts
```diff
@@ -641,7 +659,11 @@ const index = (props: IndexProps): Index => {
           // `dispose` because the index is removed. We can't call `removeWidgets`
           // because we want to keep the widgets on the instance, to allow idempotent
           // operations on `add` & `remove`.
-          widget.dispose({ helper: helper!, state: helper!.state });
+          widget.dispose({
+            helper: helper!,
+            state: helper!.state,
+            parent: this,
+          });
```
Haroenv, line 666:
```suggestion
          widget.dispose({ helper: helper!, state: helper!.state });
```

src/widgets/index/index.ts
```diff
@@ -351,7 +365,11 @@ const index = (props: IndexProps): Index => {
       if (localInstantSearchInstance && Boolean(widgets.length)) {
         const nextState = widgets.reduce((state, widget) => {
           // the `dispose` method exists at this point we already assert it
-          const next = widget.dispose!({ helper: helper!, state });
+          const next = widget.dispose!({
+            helper: helper!,
+            state,
+            parent: this,
+          });
```
Haroenv, line 372:
```suggestion
          const next = widget.dispose!({ helper: helper!, state });
```

shortcuts (approved):

PR #4737, opened 2021-04-15
refactor(ts): migrate widgets

Haroenv:
**Summary**

migrates the widgets to the types in #4735 

**Result**


src/widgets/breadcrumb/breadcrumb.tsx
```diff
@@ -113,7 +113,7 @@ export type BreadcrumbWidgetParams = {
 };
 
 export type BreadcrumbWidget = WidgetFactory<
-  BreadcrumbRendererOptions,
+  BreadcrumbWidgetDescription & { $$widgetType: 'ais.breadcrumb' },
```
tkrugg, line 116: is the `{ $$widgetType: 'ais.breadcrumb' }` part need?
It seems it's already inside [`BreadcrumbWidgetDescription`](https://github.com/algolia/instantsearch.js/blob/70715e09d078e556552aab357d5ff3cf92154232/src/connectors/breadcrumb/connectBreadcrumb.ts#L80)

src/widgets/menu-select/menu-select.tsx
```diff
@@ -85,7 +86,7 @@ const renderer = ({
     refine,
     items,
     instantSearchInstance,
-  }: MenuRendererOptions & RendererOptions<MenuConnectorParams>,
+  }: MenuRenderState & RendererOptions<MenuConnectorParams>,
```
tkrugg, line 89: I was expecting to see here something like other renderers:

`Renderer< MenuRenderState, Partial<MenuWidgetParams>>`

What's the reason for not following the same pattern?

src/widgets/menu/menu.tsx
```diff
@@ -119,7 +120,7 @@ const renderer = ({
     isShowingMore,
     toggleShowMore,
     canToggleShowMore,
-  }: MenuRendererOptions & RendererOptions<MenuConnectorParams>,
+  }: MenuRenderState & RendererOptions<MenuConnectorParams>,
```
tkrugg, line 123: Same question about pattern here: https://github.com/algolia/instantsearch.js/pull/4737/files#r622174727

> I was expecting to see here something like other renderers:
> `Renderer< MenuRenderState, Partial<MenuWidgetParams>>`
> What's the reason for not following the same pattern?

src/widgets/breadcrumb/breadcrumb.tsx
Haroenv: yes, otherwise it's a connector, not a widget :)

src/widgets/menu-select/menu-select.tsx
Haroenv: essentially it's the same, but this required less changes. Since this is purely internal, we can mutualise later without breaking change

src/widgets/voice-search/voice-search.tsx
```diff
@@ -61,13 +62,13 @@ type VoiceSearchRendererWidgetParams = {
 } & VoiceSearchWidgetParams;
 
 type VoiceSearch = WidgetFactory<
-  VoiceSearchRendererOptions,
+  VoiceSearchWidgetDescription & { $$type: 'ais.voiceSearch' },
```
shortcuts, line 65:
```suggestion
  VoiceSearchWidgetDescription & { $$widgetType: 'ais.voiceSearch' },
```

src/widgets/hits/hits.tsx
```diff
@@ -94,16 +95,11 @@ export type HitsTemplates = {
...
    *
    * @default ''
    */
-  item?: TemplateWithBindEvent<
-    Hit & {
-      __hitIndex: number;
-    }
-  >;
+  item?: TemplateWithBindEvent<Hit>;
 };
```
eunjae-lee, line 103: we need `__hitIndex` back

eunjae-lee (changes requested):

src/widgets/menu-select/menu-select.tsx
eunjae-lee: I'm fine either way.

tkrugg (commented):
left a question on __hitIndex but otherwise 👍

eunjae-lee (approved):

shortcuts (approved):
Looks clean to me!

PR #4741, opened 2021-04-20
chore(ts): more detailed type for escapeRefinement

Haroenv:
**Summary**

escapeRefinement works for both numbers and strings, and indeed returns a number or a string, however in the case of string, it can't return a number

**Result**

makes type more correct, updates eslint rules to allow overloading

### DIFF
--- a/src/lib/utils/escapeRefinement.ts
+++ b/src/lib/utils/escapeRefinement.ts
@@ -1,3 +1,5 @@
+function escapeRefinement(value: number): string | number;
+function escapeRefinement(value: string): string;
 function escapeRefinement(value: string | number): string | number {
   if (typeof value === 'number' && value < 0) {
     value = String(value).replace(/^-/, '\\-');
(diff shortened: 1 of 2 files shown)


eunjae-lee (approved):
Looks good to me.

src/lib/utils/escapeRefinement.ts
```diff
@@ -1,3 +1,5 @@
+function escapeRefinement(value: number): string | number;
+function escapeRefinement(value: string): string;
```
tkrugg, line 2: 1. I think I would have changed the function so it always returns a string just for simplicity.
2. isn't the return type on the function declaration `function escapeRefinement(value: string | number): string | number` is superfluous ?

tkrugg (approved):
left a few questions, nothing major! looks good

src/lib/utils/escapeRefinement.ts
Haroenv: that's a good point, if this function is ever touched again, it could return a string in all cases

the last declaration isn't superfluous, you need a "overarching" function if you're doing overrides

PR #4742, opened 2021-04-20
feat(ts): correct type for widget templates

Haroenv:
**Summary**

makes the types of ...Template consistent

DX-1788

**Result**

- ratingMenu


Haroenv:
superseded by #4785

PR #4743, opened 2021-04-20
feat(ts): migrate toggleRefinement & connectToggleRefinement

Haroenv:
**Summary**

converts toggleRefinement and connectToggleRefinement

**Result**

toggleRefinement and connectToggleRefinement are converted to typescript

(note typescript will fail until this is rebased onto master, I didn't do it now because it's 9 PRs deep)


src/components/ToggleRefinement/ToggleRefinement.tsx
```diff
@@ -0,0 +1,50 @@
...
+  templateProps,
+}: ToggleRefinementProps) => (
+  <div className={cssClasses.root}>
+    <label className={cssClasses.label}>
+      <input
+        className={cssClasses.checkbox}
+        type="checkbox"
+        checked={currentRefinement.isRefined}
+        onChange={event =>
+          refine({ isRefined: !(event.target as HTMLInputElement).checked })
```
eunjae-lee, line 35: This can be a slight breaking change since previously the `refine` didn't look like this, but like [this](https://github.com/algolia/instantsearch.js/blob/master/src/widgets/toggle-refinement/toggle-refinement.js#L40:L40).
I do prefer this change of wrapping isRefined in an object consistently.
However it can be a breaking change to those who create custom components using the connector.
What do you think?

Oh, no. it's not a breaking change. Custom components will use the connector, not the widget or the component file. So it's only internal change 👍🏻

eunjae-lee (approved):

src/connectors/toggle-refinement/connectToggleRefinement.ts
```diff
@@ -73,95 +146,66 @@ const createSendEvent = ({ instantSearchInstance, attribute, on, helper }) => (
...
-      // Checking
+    const toggleRefinementFactory = (helper: AlgoliaSearchHelper) => (
+      {
+        isRefined,
+      }: {
+        isRefined: boolean;
+      } = { isRefined: false }
+    ) => {
       if (!isRefined) {
         sendEvent('click', isRefined);
```
tkrugg, line 179: I'm not super familiar with this but isn't we supposed to pass the `attribute:{{value}}`  here ?

tkrugg (approved):
I'm not the most familiar with the sendEvent logic. i'm not sure on that part.
Otherwise 👍

src/connectors/toggle-refinement/connectToggleRefinement.ts
Haroenv: let's not change it in this PR :)

PR #4744, opened 2021-04-20
fix(insights): use getUserToken method instead of _get

eunjae-lee:
**Summary**

This PR replaces `_get` with `getUserToken`. At the time of writing the insights middleware, I wasn't aware of `getUserToken` method. 

**Result**

It works the same.

### DIFF
--- a/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -86,25 +86,6 @@ describe('insights', () => {
   });
 
   describe('initialize', () => {
-    it('initialize insightsClient', () => {
-      const { insightsClient, instantSearchInstance } = createTestEnvironment();
-      expect.assertions(3);
-
-      insightsClient('setUserToken', 'abc');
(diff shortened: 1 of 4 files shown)


src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -86,25 +86,6 @@ describe('insights', () => {
   });
 
   describe('initialize', () => {
-    it('initialize insightsClient', () => {
```
eunjae-lee, line 89: This test is redundant. We do not need _hasCredentials anywhere in search-insights or InstantSearch.js. (The test may have been added to see the possibility of avoiding duplicated init calls, but it's taken care in another PR.)

Haroenv (approved):
do we need to check that the insights client is at least version 1.2.0? I'm not sure if it's possible, since there's no method to get the version. since that version is pretty old, I think it's ok to use it here

EDIT: question irrelevant, getUserToken is older than _get

shortcuts (approved):

PR #4748, opened 2021-04-27
chore(insights): add user agent

eunjae-lee:
**Summary**

This PR adds `insights-middleware` to the user agent in ~both **search call** and~ **insights call** when the `insights` middleware is subscribed.

There can be another way to check the usage of the `insights` middleware other than this, but this seems to be quick and harmless.

If we already had a dozen public middlewares and wanted to track which middlewares are used, this method wouldn't be appropriate. But it's not the case, so the user agent seems okay.

### DIFF
--- a/src/middlewares/createInsightsMiddleware.ts
+++ b/src/middlewares/createInsightsMiddleware.ts
@@ -88,6 +88,12 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = props => {
     return {
       onStateChange() {},
       subscribe() {
+        if (
+          typeof instantSearchInstance.client.addAlgoliaAgent === 'function'
+        ) {
+          instantSearchInstance.client.addAlgoliaAgent(`insights-middleware`);
+        }
+
         // At the time this middleware is subscribed, `mainIndex.init()` is already called.
         // It means `mainIndex.getHelper()` exists.
         const helper = instantSearchInstance.mainIndex.getHelper()!;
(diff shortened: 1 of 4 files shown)


src/middlewares/createInsightsMiddleware.ts
```diff
@@ -88,6 +88,13 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = props => {
     return {
       onStateChange() {},
       subscribe() {
+        if (
+          typeof instantSearchInstance.client.addAlgoliaAgent === 'function'
+        ) {
+          instantSearchInstance.client.addAlgoliaAgent(`insights-middleware`);
+        }
```
Haroenv, line 95:
```suggestion
```

this doesn't effect how the search behaves, so I don't think it should be added

eunjae-lee: If we add it to insights call, we can know

> What percentage of insights calls are done by the insights middleware?

If we add it to search call, we can know

> What percentage of InstantSearch users are using the insights middleware?

That's why I added to both.

francoischalifour: Can't we rather add a similar `addAlgoliaAgent` method to the Insights client?

Haroenv: that's already done @francoischalifour, check the lower lines :)

> What percentage of InstantSearch users are using the insights middleware?

I feel like that should rather be part of the metadata middleware

eunjae-lee: > this doesn't effect how the search behaves

> I feel like that should rather be part of the metadata middleware

+ Without it, I think we could probably combine data to figure this out.

So let's remove the user agent from the search call for now.
https://github.com/algolia/instantsearch.js/pull/4748/commits/ec4ecc006677827d7efd00f9cb64d19ef478b43e

Haroenv (approved):

PR #4746, opened 2021-04-22
chore(test): updates jest

Haroenv:
**Summary**

also avoids some of the extra `console.warn` calls that we weren't testing, since now that takes up a lot of space in the output

**Result**

1. the updates from the latest jest versions
2. less annoying warn messages in the console (more can be done later)
3. suppress or fix the new lint rules

PR #4750, opened 2021-04-28
feat(core): add getUiState function

Haroenv:
**Summary**

This is useful for if you want to read the ui state externally from InstantSearch, like for example in https://github.com/algolia/autocomplete/pull/550

**Result**

a new `getUiState` function on the InstantSearch instance that can be used to retrieve the latest ui state, separately from the router.


francoischalifour:
That'll be useful for external integrations to read the InstantSearch internal state! 🙂

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -101,6 +103,9 @@ See: https://www.algolia.com/doc/guides/building-search-ui/going-further/backend
   });
 
   it('throws if insightsClient is not a function', () => {
+    const warn = jest.spyOn(global.console, 'warn');
```
francoischalifour, line 106: Why do we need these changes here?

src/lib/__tests__/InstantSearch-test.js
```diff
@@ -2017,6 +2041,176 @@ See https://www.algolia.com/doc/guides/building-search-ui/widgets/customize-an-e
   });
 });
 
+describe('getUiState', () => {
+  test('retrieves empty ui state without widgets', () => {
```
francoischalifour, line 2045:
```suggestion
  test('retrieves empty UI state without widgets', () => {
```

francoischalifour (approved):

src/lib/__tests__/InstantSearch-test.js
Haroenv: a continuation of #4746, the warnings now take up a lot of place, so I removed them from this file.

PR #4751, opened 2021-05-03
chore(ts): move tests and stories to ts

Haroenv:
**Summary**

some files were not transformed yet when their relevant files were

**Result**

files with minimal changes are changed to typescript


src/widgets/refinement-list/__tests__/refinement-list-test.ts
```diff
@@ -88,6 +90,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
         };
 
         renderWidget({ cssClasses });
+        // @ts-expect-error
```
francoischalifour, line 93: Why do we expect an error? Should we cast instead?

yarn.lock
```diff
@@ -6804,10 +6804,10 @@ eslint-plugin-jest@24.3.5:
   dependencies:
     "@typescript-eslint/experimental-utils" "^4.0.1"
 
-eslint-plugin-prettier@3.3.1:
-  version "3.3.1"
-  resolved "https://registry.yarnpkg.com/eslint-plugin-prettier/-/eslint-plugin-prettier-3.3.1.tgz#7079cfa2497078905011e6f82e8dd8453d1371b7"
-  integrity sha512-Rq3jkcFY8RYeQLgk2cCwuc0P7SEFwDravPhsJZOQ5N4YI4DSg50NyqJ/9gdZHzQlHf8MvafSesbNJCcP/FF6pQ==
+eslint-plugin-prettier@3.1.2:
```
francoischalifour, line 6807: Why did we downgrade this?

Haroenv: this is the one in package.json, I must have resolved a conflict wrong

src/widgets/refinement-list/__tests__/refinement-list-test.ts
Haroenv: yes, casting would be more correct, maybe with a todo comment? I mostly wanted quick wins in this PR. Can spend the extra minutes finding the best solution

francoischalifour (approved):

PR #4765, opened 2021-05-31
feat(ts): convert pagination widget and component

shortcuts:
**Summary**

Converts `pagination` widget and `Pagination` component to TypeScript.

**Ticket**
[DX-1704](https://algolia.atlassian.net/browse/DX-1704)


src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+      link: this.props.cssClasses.link,
+    };
+
+    if (isDisabled) {
+      cssClasses.item = cx(cssClasses.item, this.props.cssClasses.disabledItem);
+    } else if (isSelected) {
+      cssClasses.item = cx(cssClasses.item, this.props.cssClasses.selectedItem);
+    }
+
+    const url = !isDisabled ? createURL(pageNumber) : '#';
```
Haroenv, line 72: is createURL no longer optional?

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+  public render() {
+    return (
+      <div
+        className={cx(this.props.cssClasses.root, {
+          [this.props.cssClasses.noRefinementRoot]: this.props.nbPages! <= 1,
+        })}
+      >
+        <ul className={this.props.cssClasses.list}>
+          {this.props.showFirst && this.firstPageLink()}
```
Haroenv, line 163: much cleaner without passing props here, thanks!

src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,174 @@
...
+          {this.props.showPrevious && this.previousPageLink()}
+          {this.pages()}
+          {this.props.showNext && this.nextPageLink()}
+          {this.props.showLast && this.lastPageLink()}
+        </ul>
+      </div>
+    );
+  }
+}
+
```
Haroenv, line 173: defaultProps have been removed, does that make a difference in usage anywhere (a prop that isn't given or something?)

Haroenv (commented):
looks great, just a couple small questions!

src/components/Pagination/Pagination.tsx
shortcuts: it's defined here: https://github.com/algolia/instantsearch.js/pull/4765/files/5fa4d4c3ead6aeda1745bdb83792262539164fab#diff-d28dbf12c34fc4ebf83f4564d72917e83e74efd40d6056d2889fb087efc80537R43-R50

We also use this syntax in other components such as `SearchBox`, do you think we should use `ComponentName.defaultProps`?

src/components/Pagination/Pagination.tsx
shortcuts: Oops I mixed both types, it should be optional in the `Pagination` but not in `pageLink`, right?

src/components/Pagination/Pagination.tsx
Haroenv: I missed that bit, perfectly logical to me!

src/components/Pagination/Pagination.tsx
Haroenv: I checked and it's actually defined in all cases, so the check that was here before indeed isn't needed

Haroenv (approved):

PR #4766, opened 2021-05-31
feat(ts): convert rangeInput widget and component

shortcuts:
**Summary**

Converts `rangeInput` widget and `RangeInput` component to TypeScript.


src/components/RangeInput/RangeInput.tsx
```diff
@@ -1,49 +1,62 @@
...
-  componentWillReceiveProps(nextProps) {
+  public componentWillReceiveProps(nextProps: RangeInputProps) {
     this.setState({
       min: nextProps.values.min,
       max: nextProps.values.max,
     });
   }
 
-  onInput = name => event => {
+  private onInput = (key: string) => (event: Event) => {
```
Haroenv, line 41:
```suggestion
  private onInput = (key: string) => (event: InputChangeEvent) => {
```

not sure what the right name is, but I think there's a more precise event that can avoid the cast

src/components/RangeInput/RangeInput.tsx
```diff
@@ -1,49 +1,62 @@
...
+    this.props.refine([this.state.min, this.state.max]);
   };
 
-  render() {
+  public render() {
     const { min: minValue, max: maxValue } = this.state;
     const { min, max, step, cssClasses, templateProps } = this.props;
-    const isDisabled = min >= max;
-
+    const isDisabled = min && max ? min >= max : false;
```
Haroenv, line 58: I checked, and this behaves the same, undefined on either side of a comparison will cause false, so this is correct!

src/widgets/range-input/__tests__/range-input-test.ts
```diff
@@ -94,8 +99,9 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/range-input
       attribute,
     });
 
-    widget.init({ helper, instantSearchInstance });
-    widget.render({ results, helper });
+    widget.init!(createInitOptions({ helper, instantSearchInstance }));
+    // @ts-expect-error
```
Haroenv, line 103: can you add which error is expected?

src/widgets/range-input/range-input.tsx
```diff
@@ -0,0 +1,229 @@
...
+    />,
+    containerNode
+  );
+};
+
+export type RangeInputWidget = WidgetFactory<
+  Omit<RangeWidgetDescription, '$$type'> & {
+    $$widgetType: 'ais.rangeInput';
+    $$type: 'ais.rangeInput';
+  },
```
Haroenv, line 157: good call here, in a later version we can leave $$type as range, it originally wasn't changed since the impact on https://github.com/algolia/instantsearch.js/blob/master/src/lib/utils/checkIndexUiState.ts was too big to use widgetType instead of type.

Haroenv (commented):
all good, just one question about the added expect-error

src/components/RangeInput/RangeInput.tsx
shortcuts: I think the event you think of is the `ChangeEvent` but it's in React, I don't see any other event to do it, let me know if you found the name!

src/widgets/range-input/__tests__/range-input-test.ts
shortcuts: Sure, added here: https://github.com/algolia/instantsearch.js/pull/4766/commits/bfaa2f34f463c5d56ada3f8fda37d7225010d10f

Haroenv (approved):
looks good!

src/components/RangeInput/RangeInput.tsx
shortcuts: Since we use this implementation in our codebase (e.g. [SearchBox](https://github.com/algolia/instantsearch.js/blob/master/src/components/SearchBox/SearchBox.tsx#L78) and it's as correct as the `preact` implementation, I think we should stick to it!

PR #4769, opened 2021-06-01
test(utils): move to individual files and migrate to TS

Haroenv:
**Summary**

moves the util tests to individual files and migrates them to typescript

**Result**

same tests, some small fixes, index file removed


francoischalifour (approved):
Nice

PR #4770, opened 2021-06-02
refactor(ts): migrate rangeSlider 

Haroenv:
**Summary**

migrates rangeSlider and the Slider component

**Result**

migrated slider and its component.

Changes some minor things in the implementation of Rheostat to fit better with preact/ie11/typescript


src/components/Slider/Rheostat.tsx
```diff
@@ -351,21 +415,16 @@ class Rheostat extends Component {
     });
   }
 
-  startMouseSlide(ev) {
+  private startMouseSlide(ev: MouseEvent) {
     this.setStartSlide(ev, ev.clientX, ev.clientY);
 
-    if (typeof document.addEventListener === 'function') {
```
tkrugg, line 357: I had no idea we had code like this 😄

tkrugg (approved):

src/components/Slider/Rheostat.tsx
Haroenv: this is because it was migrated from Rheostat that supported way older browsers

PR #4767, opened 2021-06-01
refactor(ts): migrate components

Haroenv:
**Summary**

refactors the following components to TS:
- Hits DX-2005
- Selector DX-2006
- ClearRefinements DX-2007

**Result**

mostly renamed files, some slights changes wrt. correct imports

PR #4774, opened 2021-06-04
fix(ts): make `CSSClasses` types consistent

shortcuts:
**Summary**

In some cases, we were exporting our `CSSClasses` types as required for our widgets, while they are in reality optional and partial.

To avoid our users to provide extra logic over our types and to keep it consistent with our usage, the types are now:
- Exported as optional in the widget.
- Export as required in the component.

**Result**

- The exported types are now correct.
- When a component is provided, its `CSSClasses` type will be in the same file, named `FootComponentCSSClasses`

**Next step**

- Find the correct type for `RefinementListComponentCSSClasses`
- Provide the same update to the `templates` types

**Ticket**

[DX-1644](https://algolia.atlassian.net/browse/DX-1644)


src/components/Answers/Answers.tsx
```diff
@@ -3,22 +3,22 @@
...
-  header: string;
-  loader: string;
-  list: string;
-  item: string;
-};
+export type AnswerComponentCSSClasses = Required<
+  {
+    [TClassName in keyof AnswersCSSClasses]: string;
+  }
+>;
```
Haroenv, line 16: maybe a helper for this in /types? 

```ts
type ComponentCSSClasses<TCSSClasses> = Required<{ [className in keyof TCSSClasses]: string; }>;
```

shortcuts: Added here, looks even cleaner! https://github.com/algolia/instantsearch.js/pull/4774/commits/58fc6225cdd9fe01f5521e1ec7d80d36efb7d32f

src/components/Panel/Panel.tsx
```diff
@@ -5,16 +5,14 @@ import { useState, useEffect, useRef } from 'preact/hooks';
...
+  RenderOptions,
+  UnknownWidgetFactory,
+} from '../../types';
 
-type PanelComponentCSSClasses = Required<
-  Omit<
-    {
-      [TKey in keyof PanelCSSClasses]: string;
-    },
-    // `collapseIcon` is only used in the default templates of the widget
```
Haroenv, line 15: maybe put this comment back?

src/widgets/menu/menu.tsx
```diff
@@ -25,43 +30,43 @@ export type MenuCSSClasses = {
   /**
    * CSS class to add to the root element.
    */
-  root: string;
+  root?: string;
```
Haroenv, line 33:
```suggestion
  root?: string | string[];
```

double-check if there are others in this (wrong) style please

src/widgets/menu/menu.tsx
```diff
@@ -84,6 +89,8 @@ export type MenuTemplates = {
   }>;
 };
 
+export type MenuComponentCSSClasses = ComponentCSSClasses<MenuCSSClasses>;
```
Haroenv, line 92: should this type not be defined in the component file?

Haroenv: actually it should maybe import the one from RefinementList or something, wdyt?

src/widgets/numeric-menu/numeric-menu.tsx
```diff
@@ -135,7 +133,7 @@ export type NumericMenuTemplates = {
     /**
      * The CSS classes provided to the widget.
      */
-    cssClasses: NumericMenuRendererCSSClasses;
+    cssClasses: NumericMenuComponentCSSClasses;
```
Haroenv, line 136: alternatively, like here there's indeed a use for the classes specifically passed by this widget, so it's also all right for the version you have today

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
```diff
@@ -13,10 +13,12 @@ import connectQueryRules, {
...
   QueryRulesWidgetDescription,
 } from '../../connectors/query-rules/connectQueryRules';
-import CustomData from '../../components/QueryRuleCustomData/QueryRuleCustomData';
+import CustomData, {
+  QueryRuleCustomDataComponentCSSClasses,
+} from '../../components/QueryRuleCustomData/QueryRuleCustomData';
 
 export type QueryRuleCustomDataCSSClasses = {
-  root: string;
+  root?: string;
```
Haroenv, line 21:
```suggestion
  root?: string | string[];
```

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
```diff
@@ -54,7 +56,11 @@ const renderer: Renderer<
   const { container, cssClasses, templates } = widgetParams;
 
   render(
-    <CustomData cssClasses={cssClasses} templates={templates} items={items} />,
+    <CustomData
+      cssClasses={cssClasses as QueryRuleCustomDataComponentCSSClasses}
```
Haroenv, line 60: unlike the others, it seems that this component doesn't use cx on the class names, causing you to cast, that's probably wrong?

src/widgets/relevant-sort/relevant-sort.tsx
```diff
@@ -59,7 +61,7 @@ const renderer: Renderer<
 
   render(
     <RelevantSort
-      cssClasses={cssClasses}
+      cssClasses={cssClasses as RelevantSortComponentCSSClasses}
```
Haroenv, line 64: how come this weirdness isn't needed in the other components? is it because the component only accepts component classes and TS doesn't know in the renderer that the classes are converted before they're passed? Maybe we need to specify in the WidgetParams of renderer that the classes are componentClasses?

src/widgets/menu/menu.tsx
Haroenv: looking at the similar code in numeric menu, the current version is best imo

src/widgets/menu/menu.tsx
shortcuts: This seems to be all of them: https://github.com/algolia/instantsearch.js/pull/4774/commits/0f2a8aad54e7185faabfb22e2c8b2bfca3bc971a

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
shortcuts: Fixed in https://github.com/algolia/instantsearch.js/pull/4774/commits/18d2b5ffb91771aecc5905a46d632306769383b4

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
shortcuts: Does this seems better? https://github.com/algolia/instantsearch.js/pull/4774/commits/18d2b5ffb91771aecc5905a46d632306769383b4

src/widgets/relevant-sort/relevant-sort.tsx
shortcuts: See https://github.com/algolia/instantsearch.js/pull/4774#discussion_r647230512

src/widgets/query-rule-custom-data/query-rule-custom-data.tsx
Haroenv: yes, slightly bigger, but consistent now

Haroenv:
can we leave this PR open until the remaining components are merged? 

#4767
#4770 
#4771

shortcuts:
> can we leave this PR open until the remaining components are merged?

Yes, this make sense!

Haroenv (approved):
I think this is good!

PR #4776, opened 2021-06-08
feat(dynamicWidgets): add default attributesToRender & transformItems

Haroenv:
**Summary**

Retrieves the default attributesToRender from the result, and gives a default to transformItems now

**Result**

- [x] retrieve attributesToRender from the result
- [x] add a default for transformItems (identity)
- [x] update helper & client with new TS (https://github.com/algolia/algoliasearch-client-javascript/pull/1273, https://github.com/algolia/algoliasearch-helper-js/pull/823)


src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
```diff
@@ -39,17 +43,12 @@ describe('connectDynamicWidgets', () => {
       `);
     });
 
-    it('correct usage', () => {
+    it('minimal usage', () => {
```
francoischalifour, line 46: I don't think this test makes sense anymore if we don't take into account the first implem. It's like if we were testing that adding each option doesn't throw.

src/widgets/dynamic-widgets/__tests__/dynamic-widgets-test.ts
```diff
@@ -51,21 +51,16 @@ describe('dynamicWidgets()', () => {
       `);
     });
 
-    test('transformItems is required', () => {
+    test('transformItems is not required', () => {
```
francoischalifour, line 54: Same for that test.

francoischalifour (approved):

src/widgets/dynamic-widgets/__tests__/dynamic-widgets-test.ts
Haroenv: good point, not needed anymore for both

eunjae-lee (approved):
Looks great without the temporary code any more!

PR #4778, opened 2021-06-11
fix(clearRefinements): do not throw when widgetParams is not given

eunjae-lee:
**Summary**

This PR prevents connectClearRefinement from throwing an error when widgetParams is not given.
When it's used with the official clearRefinement widget, then normally at least `container` is there, but if it's used in a custom widget or in Vue InstantSearch, there could be no widgetParams at all.

### DIFF
--- a/src/connectors/clear-refinements/connectClearRefinements.ts
+++ b/src/connectors/clear-refinements/connectClearRefinements.ts
@@ -95,8 +95,9 @@ const connectClearRefinements: ClearRefinementsConnector = function connectClear
     } = widgetParams || {};
 
     if (
-      (widgetParams || {}).includedAttributes &&
-      (widgetParams || {}).excludedAttributes
+      widgetParams &&
+      widgetParams.includedAttributes &&
+      widgetParams.excludedAttributes
     ) {
       throw new Error(
         withUsage(
--- a/src/connectors/clear-refinements/connectClearRefinements.ts
+++ b/src/connectors/clear-refinements/connectClearRefinements.ts
@@ -94,7 +94,10 @@ const connectClearRefinements: ClearRefinementsConnector = function connectClear
       transformItems = (items => items) as TransformItems<string>,
     } = widgetParams || {};
 
-    if (widgetParams.includedAttributes && widgetParams.excludedAttributes) {
+    if (
+      (widgetParams || {}).includedAttributes &&
+      (widgetParams || {}).excludedAttributes
+    ) {
       throw new Error(
         withUsage(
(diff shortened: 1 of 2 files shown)


src/connectors/clear-refinements/connectClearRefinements.ts
```diff
@@ -94,7 +94,10 @@ const connectClearRefinements: ClearRefinementsConnector = function connectClear
       transformItems = (items => items) as TransformItems<string>,
     } = widgetParams || {};
 
-    if (widgetParams.includedAttributes && widgetParams.excludedAttributes) {
+    if (
+      (widgetParams || {}).includedAttributes &&
+      (widgetParams || {}).excludedAttributes
+    ) {
```
Haroenv, line 100:
```suggestion
    if (includedAttributes && excludedAttributes) {
```

why not like this?

eunjae-lee:
```
    const {
            includedAttributes = [],
            excludedAttributes = ['query'],
            ...
```

Yeah, it's because by default, they're filled with stuff.

Haroenv:
```suggestion
    if (
      widgetParams &&
      widgetParams.includedAttributes &&
      widgetParams.excludedAttributes
    ) {
```

This way is shorter :)

eunjae-lee, line 100:
```suggestion
    if (
      widgetParams?.includedAttributes &&
      widgetParams?.excludedAttributes
    ) {
```

shorter? :)

Haroenv: this is still doing four conditionals instead of my suggested 3

eunjae-lee: You mean at the JavaScript engine level? Not sure how they evaluate that expression, but let's go with yours :)

Haroenv (approved):

Haroenv:
> Vue InstantSearch

Doesn't Vue InstantSearch always at least give an empty object since it's calculated from props? This PR still makes sense, but I'm not sure whether Vue InstantSearch is relevant

eunjae-lee:
> > Vue InstantSearch
> 
> Doesn't Vue InstantSearch always at least give an empty object since it's calculated from props? This PR still makes sense, but I'm not sure whether Vue InstantSearch is relevant

Our official widgets provide an empty object, but if it's a custom widget and the one who made the widget didn't provide an empty object, it will fail, which was the case in the e-commerce demo in Vue InstantSearch.

PR #4779, opened 2021-06-11
ci(ts): remove test for ts-in-js

Haroenv:
**Summary**

Since all output files are now TypeScript, we no longer need to allow JavaScript in our configuration!

(Before this PR the test still failed, but that was due to a small inconcistency in the geo search renderer, that we aren't migrating ATM)

**Result**

One less test on CI, and no more configuration for checkJS, allowJS in typescript. This means CI is now finally fully green again

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,18 +99,6 @@ jobs:
             yarn run website:build
             yarn run test:e2e:saucelabs
 
-  'type check js (optional)':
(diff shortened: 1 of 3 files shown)


francoischalifour (approved):
That feels good!

eunjae-lee:
Happy to see the green check mark on our commits 😍

PR #4784, opened 2021-06-16
feat(facets): apply result from facet ordering

Haroenv:
passes facetOrderign to refinementList, menu, hierarchicalMenu which will read facet ordering from the results if available and no sortBy is given, but fall back to sortBy if no facetOrdering is available.

The option facetOrdering defaults to `true` if no sortBy is given, to make it apply out of the box.

references:
- NLP-110
- [RFC 45](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/flexible-facet-values.md)
- https://github.com/algolia/algoliasearch-helper-js/pull/822


src/connectors/hierarchical-menu/connectHierarchicalMenu.ts
```diff
@@ -81,6 +83,11 @@ export type HierarchicalMenuConnectorParams = {
    * You can also use a sort function that behaves like the standard Javascript [compareFunction](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Syntax).
    */
   sortBy?: SortBy<HierarchicalMenuItem>;
+  /**
+   * Apply the sorting of facet values defined in settings
+   * Defaults to `true` if sortBy is not given
+   */
+  facetOrdering?: boolean;
```
tkrugg, line 90: this one is called `facetOrdering` when what it does is actually order the facet values (not just the facets). 
Could this confused with dynamic widgets? do you to name it like this to keep it an implementation detail?

src/connectors/hierarchical-menu/connectHierarchicalMenu.ts
```diff
@@ -174,7 +181,8 @@ const connectHierarchicalMenu: HierarchicalMenuConnector = function connectHiera
       limit = 10,
       showMore = false,
       showMoreLimit = 20,
-      sortBy = ['name:asc'],
+      sortBy = DEFAULT_SORT,
+      facetOrdering = sortBy === DEFAULT_SORT,
```
tkrugg, line 185: Here the `sortBy === DEFAULT_SORT` portion is evaluated when `facetOrdering` has not yet been declared into the scope, but it assumes its sibling `sortBy` has been.
For example, someone refactoring would think it's good a idea to have theses attributes alpha sorted. It's an object literal, `{a, b}` or `{b, a}`, it shouldn't matter.

But then this doesn't work:

```js
{ 
   facetOrdering = sortBy === DEFAULT_SORT,
   ...
   sortBy = DEFAULT_SORT
}
```

So I think it's easy to mess up. The kind of things that would go into a _ES6 the bad part_ 😄
WDYT?

src/widgets/hierarchical-menu/hierarchical-menu.tsx
```diff
@@ -358,6 +359,7 @@ const hierarchicalMenu: HierarchicalMenuWidget = function hierarchicalMenu(
       showMore,
       showMoreLimit,
       sortBy,
+      facetOrdering,
       transformItems,
     }),
     $$widgetType: 'ais.hierarchicalMenu',
```
tkrugg, line 365: I noticed something that looks like a quirk, it's probably not related to this change, but i'm hoping you can explain it.




I think it maybe because it fallbacks to count sorting, but the whole behaviour feels quirky, doesn't it?

Screenshot at Jun 18 18-25-48


```js
instantsearch.widgets.hierarchicalMenu({
     container,
      attributes: ['hierarchicalCategories.lvl0', 'hierarchicalCategories.lvl1', 'hierarchicalCategories.lvl2'],
}),
       
```

tkrugg (commented):
Nice! so many applications for this feature!
I left 1 comment and 1 question :)

src/connectors/hierarchical-menu/connectHierarchicalMenu.ts
```diff
@@ -174,7 +181,8 @@ const connectHierarchicalMenu: HierarchicalMenuConnector = function connectHiera
...
       transformItems = (items => items) as TransformItems<HierarchicalMenuItem>,
```
tkrugg, line 186: In the RFC it was discussed we might want to output a warning to raise awareness on the overlap, has that been implemented anywhere?

src/connectors/hierarchical-menu/connectHierarchicalMenu.ts
Haroenv: True, but not sure what the alternative is? A separate declaration?

src/connectors/hierarchical-menu/connectHierarchicalMenu.ts
Haroenv: The name of the feature is called facet ordering, both for values and for facets. It's a good point that it can be confused, but I'm not sure what the alternative name can be

src/connectors/hierarchical-menu/connectHierarchicalMenu.ts
Haroenv: In the conclusion of the warning we went with: since this is enabled by default, we can't give a warning if facet ordering would apply but there's also facet ordering (since sortBy was the only way to disable). At this point however an option "facetOrdering" has been added, so a warning is possible again.

src/widgets/hierarchical-menu/hierarchical-menu.tsx
Haroenv: This happens because the count swaps to 1 at the end (not sure what causes that, but likely some engine bug), for a hierarchical facet `alpha` makes much more sense as the sortRemainingBy. We'll make sure to make that clear in the guides!

src/connectors/hierarchical-menu/connectHierarchicalMenu.ts
tkrugg: yes that's why I had in mind. I know it's not as pretty
```js
{
   sortBy = DEFAULT_SORT,
   facetOrdering
}

if (facetOrdering === undefined) {
    facetOrdering = (sortBy === DEFAULT_SORT)
}

```

Haroenv: not really a fan of that, since it requires two different declarations and to make facetOrdering a `let` or rename the variable for a condition that works totally fine. If facetOrdering is moved on top of sortBy, eslint will warn about "use before define" :)

tkrugg: Yes and typescript won't compile it so I guess it's fine after all.

src/widgets/hierarchical-menu/__tests__/hierarchical-menu-test.ts
```diff
@@ -133,18 +133,33 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hierarchica
...
       expect(results.getFacetValues).toHaveBeenCalledTimes(1);
       expect(results.getFacetValues).toHaveBeenCalledWith('hello', {
+        facetOrdering: true,
         sortBy: ['name:asc'],
       });
     });
 
     it('has a sortBy option', () => {
-      widget = hierarchicalMenu({ ...options, sortBy: ['name:asc'] });
+      widget = hierarchicalMenu({ ...options, sortBy: ['name:desc'] });
```
tkrugg, line 142: Is there a reason for needing this test to change to `name:desc`.

src/widgets/hierarchical-menu/__tests__/hierarchical-menu-test.ts
```diff
@@ -133,18 +133,33 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hierarchica
...
 
       expect(results.getFacetValues).toHaveBeenCalledTimes(1);
       expect(results.getFacetValues).toHaveBeenCalledWith('hello', {
+        facetOrdering: false,
+        sortBy: ['name:desc'],
+      });
+    });
+
+    it('has a facetOrdering option', () => {
+      widget = hierarchicalMenu({ ...options, facetOrdering: false });
```
tkrugg, line 155: I don't think we're testing the (trivial) case when it's true.

src/connectors/hierarchical-menu/__tests__/connectHierarchicalMenu-test.ts
```diff
@@ -691,6 +691,175 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hierarchica
...
+          data: null,
+          exhaustive: true,
+          isRefined: false,
+          label: 'Outdoor',
+          value: 'Outdoor',
+        },
+      ];
+
+      test.each`
+        ordered  | facetOrdering | sortBy          | expected
```
tkrugg, line 768: I had a hard time understanding what `ordered` meant here. 
I think I know now it means _"the engine returned something in `renderingContent.facetOrdering.values[attribute]`"_. Could we call it something that would make it obivous it's the order dictated by the API response?

I think any of the following would remove the confusion: 
- `responseHasFacetOrdering` 
- `renderingContentFacetOrdering`  
- `responseFacetOrdering`

src/widgets/hierarchical-menu/__tests__/hierarchical-menu-test.ts
Haroenv: so that it's not the same as the default value

src/connectors/hierarchical-menu/connectHierarchicalMenu.ts
tkrugg: Not directly related to naming.
I understand why e can consider this not to be a breaking change when doing simply`refinementList({ atribute: "brand" })`, as the user has to explicitly set `facetOrdering` in their dashboard for this behaviour to change.

I'm struggling to understand how that's better and more explicit than `refinementList({ sortBy: 'auto' })`.

I understand the following facts:
- `sortBy` wording is slightly confusing for this use case, given there possibly hidden elements via `sortRemaining: "hidden"`, **but so is `facetOrdering`**. 
-  `sortBy` currently has a DEFAULT_SORT we can't change to be `auto` which means users will have to enable it in two places (in the dashboard, and in the widget). But we can make this a default for next majors and use a dev warning in v4. 

What I see is that it's much harder to explain/document:
- the difference between `sortBy` and `facetOrdering` isn't obvious from the outside: both words are two close, 
- the behaviour is tricky: they overlap, they cancel one another (per the test cases, there 12 situations that can determine the behaviour) 
- `sortBy` and `facetOrdering` are here to stay, but they already feel like legacy. 

On the contrary, something like `sortBy: "auto"` is easy to explain, because there are just 3 cases:
   * undefined -> we follow DEFAULT SORTING
   * defined to something like : ["isRefined", "count:asc"] -> follows the order you set
   * define to`"auto"` -> follows the order your set in the dashboard

src/widgets/hierarchical-menu/__tests__/hierarchical-menu-test.ts
tkrugg: oh that makes sense.

src/connectors/hierarchical-menu/connectHierarchicalMenu.ts
Haroenv: I actually think the right order is:

1. once facet ordering is common, de-emphasise sortBy
2. explain to do all sorting post-ordering in transformItems
3. remove sortBy in a major version

In React InstantSearch for example there's no sortBy, and thus we _need_ a new property `facetOrdering` (https://github.com/algolia/react-instantsearch/pull/3067) to determine to use it or not use it. The logic there is slightly simpler, as it's true by default, as you can't override sortBy.

We first need to write the guides to emphasise not using sortBy in the frontend anymore, as that should be set by facetOrdering, and possibly override in transformItems.

eunjae-lee: My reasoning here (facetOrdering: boolean) is, let's say there's an attribute `brand`. A customer can set the facet ordering on dashboard, and it will be automatically applied to their website (as long as they have up-to-date IS). Let's say they have this widget with `brand` in 5 places across their website. They can disable all these on the dashboard. If they want to disable that facet ordering in only one place, they can apply `facetOrdering: false` to that specific widget.

If we go with `sortBy: 'auto'`, to enable the behavior, they have to go over all the code and put that attribute to 5 places. If they want to disable this at one place, they can modify sortBy to something else. If they want to disable all these facet ordering on `brand`, then they cannot do it on dashboard. They have to modify their code altogether. I feel like this can be a big downside of this approach. After all, this new facet ordering feature, if I understand correctly, is designed to give power to business users so that they can make changes on the dashboard without having to change code every single time they want something.

Haroenv: As a final option, we can have the following api:

```
const { sortBy = 'auto' } = widgetParams

// later
getFacetValues(attribute, { sortBy: sortBy === 'auto' ? defaultOrder : sortBy, facetOrdering: sortBy === 'auto'})
```

Some considerations around making it a value for sortBy

1. facetOrdering does more than auto sorting
2. auto doesn't have the reference to facetOrdering, so is confusing alone (name could be different, but not sure what fits, as it's a string)
3. you can't pass a custom fallback sortBy, although that's likely not a huge issue
4. what about React InstantSearch? does it keep a `facetOrdering` prop?
5. using sortBy is one less prop, although it's already very variadic

seafoox: I'd personally suggest we design the interface based on 
* what makes more sense for the user, not how we had to do it in the API side because of internal constraints of ours. Assuming that this might evolve in the futur. I'm referring to the use of a new attribute `facetOrdering` instead of using the existing one facets one and using an `array`
* we keep instant search widgets as loosely coupled to Algolia as possible, so this way it's easier to use in a project that doesn't use Algolia: I'm thinking about people using Algolia with a 3rd party platform - ultimately people who use their service could be potential customer of Algolia tomorrow.

Because of that I personally kite like the approach suggest by @tkrugg of using `sortBy: "auto"`. Wondering if instead of `auto` using `api`, or `auto_from_api` could also make sense this way it's even clearer that it comes from the API.

eunjae-lee: @seafoox 

> I'm referring to the use of a new attribute facetOrdering

The way this PR is proposing, is not having to put `facetOrdering` at all by default. When there are multiple widgets for a same attribute, and you only want to disable facet ordering on a specific widget, then you pass `facetOrdering: false` to the widget. Otherwise, it's `true` by default. Leveraging `sortBy`, customers have to put `sortBy: 'auto'` to every single widget to enable facet ordering whereas with `facetOrdering: true` by default, they don't have to do anything except for configuring on dashboard.

> we keep instant search widgets as loosely coupled to Algolia as possible, so this way it's easier to use in a project that doesn't use Algolia

I don't know if it's what we should pursue. We do not implement features in InstantSearch with other search backend services in mind. If that's what we need to do it, then we need to re-assess our approach, but until then, it's better for us to make things as smooth as possible for Algolia customers.

Haroenv: note that we could still default to "auto", if we consider that we don't need to provide a sortBy fallback (it can still be the default_sortby) that can't be overridden, as in my last code snippet

Haroenv: We discussed and are going with:

* no facetOrdering option
* facetOrdering is used if you don't provide sortBy (or undefined)

This still gives a default facet ordering, but avoids the two parameters. In React InstantSearch we are keeping facetOrdering as an option, as it has no sortBy to work with

src/connectors/refinement-list/__tests__/connectRefinementList-test.ts
```diff
@@ -2589,6 +2589,119 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
...
+          value: 'Microsoft',
+        },
+      ];
+
+      test.each`
+        facetOrderingInResult | sortBy           | expected
+        ${true}               | ${undefined}     | ${resultsViaFacetOrdering}
+        ${false}              | ${undefined}     | ${resultsViaSortBy}
+        ${true}               | ${['isRefined']} | ${resultsViaSortBy}
+        ${false}              | ${['isRefined']} | ${resultsViaSortBy}
```
tkrugg, line 2646: I think it's be worth picking a `sortBy` input that yields a different result than the default value (same for `connectMenu-test.ts` and `connectHierarhicalMenu-test.ts`). Just having `resultsViaSortByDefaultValue` and `resultsViaSortByIsRefined`. 

The rational: right now, I could (accidentally) change the implementation to ignore `sortBy` prop and this test would still pass. 

Another test is still going to break, so I don't consider this a blocker, just a safer testing strategy. Not a blocker for this PR.

tkrugg (approved):
Great work! 👍

PR #4790, opened 2021-06-23
fix(mainHelper): allow a mainHelper to be set before start

Haroenv:
**Summary**

In Vue InstantSearch server side rendering, we set a mainHelper https://github.com/algolia/vue-instantsearch/blob/118e0a39b8a047930520220da757ff0175cec12d/src/util/createServerRootMixin.js#L234-L235 before the instance is started.

This worked fine with existing fixes done, except for in routing, because every place that changes the state uses helper from the index, which is derived of the helper that's created before start. Routing and middleware however uses the helper which gets created in start, and thus don't trigger a search on the helper that actually searches.

To avoid these issues, we prevent recreation of the main helper if it's already set (by Vue InstantSearch for example). Whenever we will add the proper API for server side rendering in InstantSearch, we can rethink this part and probably remove it on a major.

**Result**

mainHelper doesn't get recreated if there's already one present

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -425,7 +425,10 @@ See ${createDocumentationLink({
     // This Helper is used for the queries, we don't care about its state. The
     // states are managed at the `index` level. We use this Helper to create
     // DerivedHelper scoped into the `index` widgets.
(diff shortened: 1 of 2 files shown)


eunjae-lee (approved):
Looks good to me 🙌

PR #4785, opened 2021-06-16
fix(ts): make template types consistent

shortcuts:
**Summary**

Follow up of https://github.com/algolia/instantsearch.js/pull/4774

In some cases, we were exporting our Templates types as required for our widgets, while they are in reality optional and partial.

To avoid our users to provide extra logic over our types and to keep it consistent with our usage, the types are now:
- Exported as optional in the widget.
- Export as required in the component.

**Result**

The exported types are now correct.

**Ticket**

[DX-1778](https://algolia.atlassian.net/browse/DX-1778)

PR #4791, opened 2021-06-23
refactor(ts): remove implicit any

Haroenv:
- adds missing @types devDependency
- adds missing annotations 
- rewrite the types for the answers connector


Haroenv:
Actually not ready for review, I'll wait until #4785 is fully merged, in case there's some conflicts, I'd rather them be in this PR

shortcuts (commented):
Cool PR! Just a few questions, I'll do an other review after they are answered :D

src/widgets/refinement-list/refinement-list.tsx
```diff
@@ -162,6 +162,10 @@ type RefinementListSearchableTemplates = {
 
 export type RefinementListTemplates = RefinementListOwnTemplates &
   RefinementListSearchableTemplates;
+export type RefinementListComponentTemplates = Required<
+  RefinementListOwnTemplates
+> &
+  RefinementListSearchableTemplates;
```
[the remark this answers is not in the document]
shortcuts: ^ this might be auto-fixed after the master merge?

src/components/Slider/Rheostat.tsx
```diff
@@ -291,7 +308,7 @@ class Rheostat extends Component<Props, State> {
     if (!this.props.snapPoints!.length) return value;
 
     return this.props.snapPoints!.reduce((snapTo, snap) =>
-      Math.abs(snapTo - value) < Math.abs(snap - value) ? snapTo : snap
+      Math.abs(snapTo! - value) < Math.abs(snap - value) ? snapTo : snap
```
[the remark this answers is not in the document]
Haroenv: yes, for the time being there's no checks, so clearly certain things that we mark as optional get defaulted at a certain step, but putting the ! here indicates the possible problem in the future, I didn't want to change code in this pr

src/widgets/stats/stats.tsx
```diff
@@ -88,6 +92,13 @@ const renderer = ({
   cssClasses,
   containerNode,
   templates,
+}: {
```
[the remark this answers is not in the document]
Haroenv: the remaining ones don't use prepareTemplateProps now that master is merged :)

src/widgets/refinement-list/refinement-list.tsx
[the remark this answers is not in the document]
Haroenv: it's in the file (same for menu, hierarchical menu, rating menu) because the refinement list component is shared

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -21,13 +18,11 @@ export type BreadcrumbComponentCSSClasses = ComponentCSSClasses<
 export type BreadcrumbComponentTemplates = Required<BreadcrumbTemplates>;
 
 export type BreadcrumbProps = {
-  items: BreadcrumbItem[];
+  items: BreadcrumbConnectorParamsItem[];
```
tkrugg, line 21: Open question: wouldn't it be better to have a specific type here. Using the connector type leaks the internal detail this `BreadcrumpProps.items` is the same as `BreadcrumbConnectorParamsItem`, when the widget user doesn't really have to know about it.

src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -21,13 +18,11 @@ export type BreadcrumbComponentCSSClasses = ComponentCSSClasses<
...
   cssClasses: BreadcrumbComponentCSSClasses;
-  templateProps: {
-    templates: BreadcrumbComponentTemplates;
-  };
-  createURL(value: string | undefined): string;
-  refine(value: string | undefined): void;
+  templateProps: PreparedTemplateProps<BreadcrumbComponentTemplates>;
+  createURL(value?: string | null): string;
```
tkrugg, line 24: Not sure why we need to handle`null` here. I don't know in which case is this value `null` ?

src/components/Slider/Slider.tsx
```diff
@@ -1,43 +1,45 @@
...
-    | boolean
-    | {
-        format(value: number): string;
-      };
-  cssClasses: {
-    root: string;
-    disabledRoot: string;
-  };
+  step?: number;
+  tooltips?: RangeSliderWidgetParams['tooltips'];
```
tkrugg, line 26: are we happy using the widget type here?
Currently widgets depends on component, not the other way around.

src/components/Slider/Slider.tsx
```diff
@@ -54,18 +56,31 @@ class Slider extends Component<SliderProps> {
...
+  private createHandleComponent = (
+    tooltips: RangeSliderWidgetParams['tooltips']
+  ) => (props: HandleProps) => {
     // display only two decimals after comma,
     // and apply `tooltips.format()` if any
     const roundedValue =
-      Math.round(parseFloat(props['aria-valuenow']) * 100) / 100;
+      Math.round(
+        // have to cast as a string, as the value given to the prop is a number, but becomes a string when read
+        parseFloat((props['aria-valuenow'] as unknown) as string) * 100
```
tkrugg, line 80: isn't `'aria-valuenow'` optional? I think casting as unknown hide the fact this value can be undefined but I'm not yet sure how we can avoid that.

src/connectors/numeric-menu/connectNumericMenu.ts
```diff
@@ -127,9 +128,15 @@ export type NumericMenuConnector = Connector<
...
-) => {
+const createSendEvent = ({
+  instantSearchInstance,
+  helper,
+  attribute,
+}: {
+  instantSearchInstance: InstantSearch;
+  helper: AlgoliaSearchHelper;
+  attribute: string;
+}) => (...args: [InsightsEvent] | [string, string, string?]) => {
```
tkrugg, line 139: TIL !

tkrugg (approved):

src/components/Slider/Slider.tsx
Haroenv: There's other components where that's the case. It's true it's a little coupled at the moment, but all components for example rely on the widget for the css classes (instead of the other way round), so we can fix that some other time (add a lint rule to avoid depending too circular), fixing this one instance is insignificant.

src/components/Slider/Slider.tsx
Haroenv:
```suggestion
        parseFloat((props['aria-valuenow'] as unknown) as (string | undefined)) * 100
```

is this better?

src/components/Breadcrumb/Breadcrumb.tsx
Haroenv: This is the component, so it's not public anyway :) But I import flow this can better be solved globally

src/components/Slider/Slider.tsx
Haroenv: parseFloat doesn't accept undefined, so it will be for some other time when we figure out why it even needs to be parsed as a float if it's already a number

src/components/Breadcrumb/Breadcrumb.tsx
Haroenv: https://github.com/algolia/instantsearch.js/blob/c57d7bbdd9451299b3af48ce6f894b012c24b9d5/src/connectors/breadcrumb/connectBreadcrumb.ts#L270

this is what makes it null, and I don't see how to avoid it (last item?)

tkrugg: or right! that's when the value has been selected! 👍  thanks for clarifying.

PR #4794, opened 2021-06-23
fix(relevantSort): export `RelevantSortWidgetParams` type

francoischalifour:
This type cannot be imported from other projects, as opposed to other widgets.


Haroenv (approved):
can we add some rule with API Extractor (https://github.com/algolia/instantsearch.js/blob/master/scripts/build/types.js#L50) to detect these issues?

francoischalifour:
@Haroenv Yes, good idea. [Added to the backlog](https://algolia.atlassian.net/browse/DX-2230?atlOrigin=eyJpIjoiNDhmNzZmYjQyMjBiNDA1Mzg0MDdjOGIyNTdiZWU1NjQiLCJwIjoiaiJ9).

PR #4796, opened 2021-06-25
fix(onStateChange): propagate change to middleware

### LINKED ISSUE #4795: URL routing is not working while a custom onStateChange function is defined
## 🐛 Bug description

When using the [custom `onStateChange()` implementation](https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/#widget-param-onstatechange) together with [URL routing](https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/#widget-param-routing) the router is not called upon UI state changes.

## 🔍 Bug reproduction

1. Enable basic URL routing
2. Observe the URL being updated upon state changes
3. Now add a custom `onStateChange()` function
4. Observe the URL no longer being updated upon state changes

### Live reproduction

**Without custom `onStateChange()` everything works as expected:**

https://codesandbox.io/s/inspiring-carson-tc2sv?file=/src/app.js:215-291

**Enable the custom `onStateChange()` and the router stops working:**

https://codesandbox.io/s/inspiring-carson-tc2sv?file=/src/app.js

## 💭 Expected behavior

The router is expected to work identically, based on the router configuration, when a custom `onStateChange` function is either defined or not.

## Environment

- OS: macOS
- Browser: Chrome
- Version: `4.24.0-experimental-typescript.0`

Haroenv:
This wrong behaviour was originally introduced in https://github.com/algolia/instantsearch.js/pull/4699 by removing onInternalStateChange, since I assumed the helper state change event would call the middleware, but it doesn't happen in all cases.

This means that now onInternalStateChange gets called one more time in certain situations, but that's not really an issue, as it's deferred anyway.

I've checked and the integration test written by @tkrugg indeed starts failing on #4699 

Fixes #4795

### DIFF
--- a/src/lib/__tests__/InstantSearch-integration-test.ts
+++ b/src/lib/__tests__/InstantSearch-integration-test.ts
@@ -1,8 +1,4 @@
-import {
-  getByText,
-  fireEvent,
-  getByPlaceholderText,
-} from '@testing-library/dom';
+import { getByText, fireEvent } from '@testing-library/dom';
 import instantsearch from '../../index.es';
 import { configure, searchBox } from '../../widgets';
 import { connectConfigure } from '../../connectors';
@@ -52,7 +48,7 @@ describe('configure', () => {
 });
 
 describe('middleware', () => {
-  it("runs middleware's onStateChange uiState has changed", async () => {
+  it("runs middlewares' onStateChange uiState has changed", async () => {
     const container = document.createElement('div');
     const search = instantsearch({
       indexName: 'instant_search',
@@ -76,7 +72,7 @@ describe('middleware', () => {
 
     search.start();
 
-    fireEvent.input(getByPlaceholderText(container, 'search'), {
+    fireEvent.input(container.querySelector('input')!, {
       target: { value: 'q' },
     });
 
@@ -84,7 +80,7 @@ describe('middleware', () => {
     expect(middlewareDefinition.onStateChange).toHaveBeenCalledTimes(1);
   });
 
-  it("runs middleware's onStateChange uiState has changed and onStateChange was provided", async () => {
+  it("runs middlewares' onStateChange uiState has changed and onStateChange was provided", async () => {
     const container = document.createElement('div');
     const search = instantsearch({
       indexName: 'instant_search',
@@ -111,7 +107,7 @@ describe('middleware', () => {
 
     search.start();
 
-    fireEvent.input(getByPlaceholderText(container, 'search'), {
+    fireEvent.input(container.querySelector('input')!, {
       target: { value: 'q' },
     });
 
--- a/src/lib/__tests__/InstantSearch-integration-test.ts
+++ b/src/lib/__tests__/InstantSearch-integration-test.ts
@@ -48,7 +48,7 @@ describe('configure', () => {
 });
 
 describe('middleware', () => {
-  it("runs middlewares' onStateChange uiState has changed", async () => {
+  it("runs middlewares' onStateChange when uiState changes", async () => {
     const container = document.createElement('div');
     const search = instantsearch({
       indexName: 'instant_search',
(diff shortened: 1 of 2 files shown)


tkrugg (commented):
It looks good, thanks for fixing this so fast.
The regression was that `onInternalStateChange` was no longer called in `setUiState`. We should have a test checking this specifically (middleware.onStateChange not being called was just a side effect of the regression).

Haroenv:
that's more of a side-effect of what actually was the wrong behaviour (middleware weren't called), the onInternalStateChange could change without breaking the behaviour @tkrugg

tkrugg:
> that's more of a side-effect of what actually was the wrong behaviour (middleware weren't called), the onInternalStateChange could change without breaking the behaviour @tkrugg, we shouldn't have a test on the solution, but on the problem it solves

I think we agree on the idea but disagree on what we call side-effect and behaviour. 
I'm really speaking from the `setUiState` point of view (because that's where I think we should add a test).
The behaviour of this public function is "change uiState and notify everybody with uiState has changed".
The regression happened because the second part of this responsibility was removed, and there was nothing that could have prevented it. Such a test would have made it very clear.

middleware being notified *is* the side-effect. If at some point we decide middleware will be notified by another mean, setUiState test doesn't have to know about it, it's not its responsibility. I don't think we can ever go wrong by having a test clearly checking the behaviour of a public function is preserved, and here its behaviour is not "it should call middleware", it's "it should notify others of UI State update". That's why I think we're missing a test.

Haroenv:
There's actually already two tests here, but they passed with the bug already, do you have any idea what's missing in the condition @tkrugg? https://github.com/algolia/instantsearch.js/blob/e8329aecb386755a039cf10850e394d0d71f29f4/src/lib/__tests__/InstantSearch-test.tsx#L1769-L1856

eunjae-lee:
The new 2 test cases make sense to me. Which scenario do you think we should test more? @tkrugg

eunjae-lee:
> I was thinking one that would check onInternalStateUpdate is called when setUIState is, because that's the "notify everybody state has changed" part of the implem.

yeah, but also the implementation can change while the end result stays the same.

The contract we have is 

> when `setUiState` is called(A), the middlewares' onStateChange gets called(B).

It's not important what happens between (A) and (B). We trigger (A) and see if (B) is achieved. By testing this way, we have a room to refactor , even rename or do whatever with the internal implementations.

src/lib/__tests__/InstantSearch-integration-test.ts
```diff
@@ -44,3 +46,72 @@ describe('configure', () => {
     });
   });
 });
+
+describe('middleware', () => {
+  it("runs middlewares' onStateChange uiState has changed", async () => {
```
francoischalifour, line 51: I think that's what you meant?

```suggestion
  it("runs middlewares' onStateChange when uiState changes", async () => {
```

src/lib/__tests__/InstantSearch-integration-test.ts
```diff
@@ -44,3 +46,72 @@ describe('configure', () => {
...
+
+    fireEvent.input(container.querySelector('input')!, {
+      target: { value: 'q' },
+    });
+
+    await runAllMicroTasks();
+    expect(middlewareDefinition.onStateChange).toHaveBeenCalledTimes(1);
+  });
+
+  it("runs middlewares' onStateChange uiState has changed and onStateChange was provided", async () => {
```
francoischalifour, line 83: I don't understand this test name. Can we rephrase it?

francoischalifour (approved):

src/lib/__tests__/InstantSearch-integration-test.ts
Haroenv:
```suggestion
  it("runs middlewares' onStateChange when uiState changes (with root onStateChange)", async () => {
```

Hard to phrase it correctly, but this is the bug fixed in this PR. Is the suggestion better @francoischalifour ?

francoischalifour: Better yes. How about this?

```suggestion
  it("runs middlewares' onStateChange when uiState changes with user-provided onStateChange param", async () => {
```

PR #4797, opened 2021-06-25
chore: update google maps types

jpoehnelt:
https://www.npmjs.com/package/@types/googlemaps is deprecated in favor of https://www.npmjs.com/package/@types/google.maps

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "scripts/transforms"
   ],
   "dependencies": {
-    "@types/googlemaps": "^3.39.6",
     "@types/hogan.js": "^3.0.0",
     "algoliasearch-helper": "^3.4.5",
     "classnames": "^2.2.5",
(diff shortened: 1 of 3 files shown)


Haroenv:
You moved it from dependencies to devDependencies. At the moment we haven't yet decided whether type dependencies are dependencies or something else, do you have a preference @jpoehnelt? If it's in devDependencies, people won't know which version of the type dependencies we use.

Thanks for your PR!

jpoehnelt:
> people won't know which version of the type dependencies we use

I don't see how this matters given options like `skipLibCheck`.

package.json
```diff
@@ -52,7 +52,6 @@
     "scripts/transforms"
   ],
   "dependencies": {
-    "@types/googlemaps": "^3.39.6",
     "@types/hogan.js": "^3.0.0",
```
Haroenv, line 55:
```suggestion
    "@types/google.maps": "^3.45.3",
    "@types/hogan.js": "^3.0.0",
```

package.json
```diff
@@ -80,6 +79,7 @@
     "@testing-library/preact": "1.0.2",
     "@types/classnames": "^2.2.7",
     "@types/enzyme": "^3.1.15",
+    "@types/google.maps": "^3.45.3",
```
Haroenv, line 82:
```suggestion
```

Haroenv (commented):
let's go with keeping it as a dependency, maybe changing it still before typescript is released as stable. Thanks!

Haroenv (approved):
thanks @jpoehnelt !

PR #4798, opened 2021-06-28
fix(sortBy): do not write the default state

Haroenv:
**Summary**

Without this check, if a route state is read (eg. in ssr) with the default value, it will be persisted in the URL, as the refinement is the default refinement, and the ui state was kept as-is, instead of being cleared.

**Result**

fixes https://github.com/algolia/vue-instantsearch/issues/1005


eunjae-lee:
I think I understand the issue partially. How can I reproduce the issue on my own?

https://user-images.githubusercontent.com/499898/123647677-95e47180-d828-11eb-8e62-b4a4fc560226.mp4

I don't understand the flow of the default state added to state.

Haroenv:
The default state gets added in Vue InstantSearch in server side rendering, not 100% sure how, but I can tell that it happens only then (see sandbox in linked issue) @eunjae-lee

tkrugg (approved):

PR #4801, opened 2021-06-29
chore(dependencies): remove prop-types

Haroenv:
**Summary**

prop-types is no longer used, so it can be removed!

**Result**

- remove prop-types
- remove babel setup for prop-types
- remove prop-type usage in tests (for casting with enzyme, replaced that with a custom wrapper for enzyme.mount/shallow)

This PR is split in three commits for:

1. moving prop-types to devDep
2. removing usage as a cast-target for enzyme
3. removing prop-types completely


eunjae-lee (approved):
Amazing. Thanks for the work!

PR #4803, opened 2021-07-05
feat(ts): use `Partial` exports

shortcuts:
**Summary**

We decided to convert all the exported types that are completely optional to `Partial` since it's less error prone.

**Results**

Fully optional types are now `Partial`s.


Haroenv (approved):
let's go, this looks good to me

PR #4806, opened 2021-07-09
chore(ts): include search-insights

Haroenv:
**Summary**

InsightsClient can be imported more directly and hopefully have the ignored files no longer ignored

**Result**

typescript failures in node_modules still somehow, even though skipLibCheck is set to true??

I can try ignoring the two files again, but at least still have the less rewritten types?

### DIFF
--- a/src/types/insights.ts
+++ b/src/types/insights.ts
@@ -1,3 +1,5 @@
+import { InsightsClient as OriginalInsightsClient } from "search-insights/lib/types";
+
 export type InsightsClientMethod =
   | 'viewedObjectIDs'
   | 'clickedFilters'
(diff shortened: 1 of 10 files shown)


src/types/insights.ts
```diff
@@ -1,3 +1,5 @@
+import { InsightsClient as OriginalInsightsClient } from "search-insights/lib/types";
+
 export type InsightsClientMethod =
```
Haroenv, line 3: this can probably also be imported from the types of search-insights

eunjae-lee:
oh I hope we get it fixed this time 🙏🏻
let me know if you need any help!

Haroenv:
I can't get this to work while search-insights still has so many TS errors, not sure why it doesn't consider it as node_modules and leaves it alone 😢

PR #4808, opened 2021-07-12
fix(routing): prevent history.pushState on InstantSearch.dispose call

tkrugg:
**Summary**

This change is aiming a fixing an issue we've had with routing on [Angular IS](https://github.com/algolia/angular-instantsearch/issues?q=is%3Aissue+back+routing) and [Vue IS](https://github.com/algolia/vue-instantsearch/issues?q=is%3Aissue+back+button+routing+), and essentially in any context where [dispose](https://github.com/algolia/instantsearch.js/blob/master/src/lib/InstantSearch.ts#L506) is called. On Angular/Vue InstantSearch, dispose is casually called when we navigate from route A to B. During this transition, all the widgets of route A are unmounted -- InstantSearch included, hence the [dispose call](https://github.com/algolia/angular-instantsearch/blob/master/src/instantsearch/instantsearch.ts#L247).

In the [following Vanilla example](https://codesandbox.io/s/gallant-buck-mvcpv?file=/src/app.js), you can see that `window.history.pushState` is being when the InstantSearch instance is disposed. This is the result of 2 things `dispose` does:
1. it removes the widgets, which triggers the router middleware's [onStateChange](https://github.com/algolia/instantsearch.js/blob/master/src/middlewares/createRouterMiddleware.ts#L49), roughly:
  
   https://github.com/algolia/instantsearch.js/blob/ddf60adea7cf35f93338d6a70466ccfdbb71749e/src/lib/InstantSearch.ts#L511 
   ```
   InstantSearch.dispose() -> removeWidgets -> ... -> RouterMiddleware.onStateChange -> window.history.pushState
   ```
 
 
 

2. middleware unsubscribe -> triggers routers's dispose.
https://github.com/algolia/instantsearch.js/blob/ddf60adea7cf35f93338d6a70466ccfdbb71749e/src/lib/InstantSearch.ts#L526-L528

   ```
   InstantSearch.dispose() -> RouterMiddleware.unsubscribe() -> router.dipose()- -> window.history.pushState
   ```

**Result**

There are probably a few different ways to solve the issue, here are 2:

**Solution A**: (implemented in this PR)
   1. On the router level: neutralize `route.write` call after `router.dispose` was called 
   2. On the router level: enable `router.write` to use `window.history.replaceState` when called from dispose.

**Solution B** 
1. On the middleware level: prevent middleware's `onStateChange` to be called if middleware has been disposed. 
2. On the router level: enable `router.write` to use `window.history.replaceState` when called from dispose. (same as A)

### DIFF
--- a/src/lib/__tests__/RoutingManager-test.ts
+++ b/src/lib/__tests__/RoutingManager-test.ts
@@ -4,8 +4,10 @@ import qs from 'qs';
 import { createSearchClient } from '../../../test/mock/createSearchClient';
 import { createWidget } from '../../../test/mock/createWidget';
 import { runAllMicroTasks } from '../../../test/utils/runAllMicroTasks';
+import { wait } from '../../../test/utils/wait';
 import { Router, Widget, UiState, StateMapping, RouteState } from '../../types';
 import historyRouter from '../routers/history';
+import simple from '../stateMappings/simple';
 import instantsearch from '../main';
 
 const createFakeRouter = (args: Partial<Router> = {}): Router => ({
@@ -536,6 +538,51 @@ describe('RoutingManager', () => {
           triggerChange: true,
         },
       });
+    });
+    test('cleans URL on dispose', async () => {
+      const searchClient = createSearchClient();
+      const writeDelay = 400;
+
+      const search = instantsearch({
+        indexName: 'indexName',
+        searchClient,
+        routing: {
+          router: historyRouter({
+            writeDelay,
+          }),
+          stateMapping: simple(),
+        },
+      });
+
+      const fakeSearchBox: any = createFakeSearchBox();
+      search.addWidgets([fakeSearchBox]);
+
+      search.start();
+
+      await runAllMicroTasks();
+
+      // Wait for ${writeDelay} to pass then check URL has been initialized
+      await wait(1.5 * writeDelay);
+      expect(window.location.search).toEqual(``);
+      expect(window.history.length).toEqual(1);
+
+      // Trigger an update - push a change
+      fakeSearchBox.refine('Apple');
+
+      // Wait for ${writeDelay} to pass then check URL has been updated
+      await wait(1.5 * writeDelay);
+      expect(window.location.search).toEqual(
+        `?${encodeURI('indexName[query]=Apple')}`
+      );
+      expect(window.history.length).toEqual(2);
+
+      // Trigger a dispose
+      search.dispose();
+
+      // Wait for ${writeDelay} to pass then check URL has been cleaned
+      await wait(1.5 * writeDelay);
+      expect(window.location.search).toEqual('');
+      expect(window.history.length).toEqual(2);
     });
   });
 
--- a/src/lib/routers/history.ts
+++ b/src/lib/routers/history.ts
@@ -119,7 +119,7 @@ class BrowserHistory implements Router {
   /**
    * Pushes a search state into the URL.
    */
-  public write(routeState: RouteState): void {
+  public write(routeState: RouteState, replace: boolean = false): void {
     const url = this.createURL(routeState);
     const title = this.windowTitle && this.windowTitle(routeState);
 
@@ -129,7 +129,13 @@ class BrowserHistory implements Router {
 
     this.writeTimer = window.setTimeout(() => {
       setWindowTitle(title);
-      window.history.pushState(routeState, title || '', url);
+
+      if (replace) {
+        window.history.replaceState(routeState, title || '', url);
+      } else {
+        window.history.pushState(routeState, title || '', url);
+      }
+
       this.writeTimer = undefined;
     }, this.writeDelay);
   }
@@ -187,7 +193,7 @@ class BrowserHistory implements Router {
       window.clearTimeout(this.writeTimer);
     }
 
-    this.write({});
+    this.write({}, true);
   }
 }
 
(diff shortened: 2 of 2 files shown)


src/lib/routers/history.ts
```diff
@@ -119,7 +120,8 @@ class BrowserHistory implements Router {
   /**
    * Pushes a search state into the URL.
    */
-  public write(routeState: RouteState): void {
+  public write(routeState: RouteState, replace: boolean = false): void {
```
tkrugg, line 123: Correct me if I'm wrong, adding an optional parameter should not be a breaking change for someone relying on this type.

src/lib/routers/history.ts
```diff
@@ -129,7 +131,13 @@ class BrowserHistory implements Router {
 
     this.writeTimer = window.setTimeout(() => {
       setWindowTitle(title);
-      window.history.pushState(routeState, title || '', url);
+
+      if (replace) {
+        window.history.replaceState(routeState, title || '', url);
```
Haroenv, line 136: this means the last search state will be removed in regular use cases where you dispose the InstantSearch:

```
mount: ?
query: ?query=123 --> deleted at dispose
dispose: ?
```

Haroenv: Actually in SPA implementation, this breaks in the same way. What should be an option is *what to do* for writing after dispose happens,

1. vanilla, where dispose is on the same page: write on dispose
2. angular, vue etc. with Vue Router, where dispose happens together with a page change: no write after disposed

src/lib/routers/history.ts
```diff
@@ -119,7 +120,8 @@ class BrowserHistory implements Router {
   /**
    * Pushes a search state into the URL.
    */
-  public write(routeState: RouteState): void {
+  public write(routeState: RouteState, replace: boolean = false): void {
+    if (this._disposed) return;
```
Haroenv, line 124: if you don't write when the dispose has happened, will the replaceState even ever happen?

Haroenv (changes requested):

src/lib/routers/history.ts
Haroenv: yes, that's not an issue

src/lib/routers/history.ts
tkrugg: yes, because the `_dispose` is set after the final `write` call.
You can tell it runs by the looking [at the test](https://github.com/algolia/instantsearch.js/pull/4808/files#diff-b6b548e96fccd4f7c292cf53d63f1ae7314c42cad4c51e3b354f985d438b3203R580-R586): the URL is indeed updated, the history stack didn't change length.

src/lib/__tests__/RoutingManager-test.ts
```diff
@@ -537,6 +539,51 @@ describe('RoutingManager', () => {
...
+
+      // Wait for ${writeDelay} to pass then check URL has been updated
+      await wait(1.5 * writeDelay);
+      expect(window.location.search).toEqual(
+        `?${encodeURI('indexName[query]=Apple')}`
+      );
+      expect(window.history.length).toEqual(2);
+
+      // Trigger a dispose
+      search.dispose();
```
Haroenv, line 580: you're not testing here if the widgets get removed before the search

tkrugg: To me I do because that's what `search.dispose()` does at first (remove widgets).
Is this the flow you have in mind?

```ts
...
const fakeSearchBox: any = createFakeSearchBox();
search.addWidgets([fakeSearchBox]);

search.start();
...
fakeSearchBox.refine('Apple');
... // check history
search.removeWidgets([fakeSearchBox])
... // check history
search.dispose()
.. // check history

```

Haroenv: I know it's in Vue InstantSearch / Angular InstantSearch, but it would be interesting to see how that behaves, what actually gets unmounted when

tkrugg: This test is supposed to be close to what happens in [Angular IS](https://github.com/algolia/angular-instantsearch/blob/master/src/instantsearch/instantsearch.ts#L247) and [Vue IS](https://github.com/algolia/vue-instantsearch/blob/513017a47d38ccddd0557c01ce732e6e5e7648d6/src/util/createInstantSearchComponent.js#L71-L78) when a route unmounts. 

Maybe we can try to be even closer to what's happening in Angular IS and Vue IS by simulating the cascading unmounting that occurs on route change. I'll do some research and update here.

src/lib/routers/history.ts
tkrugg: > 1. vanilla, where dispose is on the same page: write on dispose

do you mean write with `replaceState`? I don't see many uses cases for dispose in vanilla not pushing a new state on dispose as it was a navigation event we may want to go back to sounds fair to me.

> 2. angular, vue etc. with Vue Router, where dispose happens together with a page change: no write after disposed

So that would mean exposing an option removing the final `write({ })` altogether?
Sounds fair to me. 

```js
history({ preventWriteOnDispose: true })
```


```js
dispose() {
    ...
    if (!preventWriteOnDispose) {
        write({ })
    }
   _diposed = true;
},

write(routeState) {
   if (_diposed) return // we shouldn't be able to write after dispose
   ...
}
```

We'd still need to make sure we're no longer able to make writes after `dipose` call.
We can either do that here or on the middleware level (ensuring `onStateChange` has no effect after `unsubscribe` has been called). This sounds like a more legit change.

tkrugg:
Closing now but I'll reopen this later when I get time to more testing

PR #4811, opened 2021-07-15
ci(types): update dependencies & test on CI build:types

Haroenv:
**Summary**

build:types failed silently on master when a release is done

**Result**

1. add build:types to main CI
2. update dependencies to make build pass
3. exit with right error code if build:types fails

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,11 +36,14 @@ jobs:
       - run: *run_yarn_install
       - save_cache: *save_yarn_cache
       - run:
-          name: Build & Test packages size
-          command: |
-            yarn run build
-            yarn run test:size
-
+          name: Build library
+          command: yarn run build
+      - run:
+          name: Test packages size
+          command: yarn run test:size
+      - run:
+          name: Type Export
+          command: yarn run build:types
   lint:
     <<: *defaults
     steps:
@@ -55,9 +58,6 @@ jobs:
       - run:
           name: Type Checking
           command: yarn run type-check
-      - run:
-          name: Type Export
-          command: yarn run build:types
 
   'type check algoliasearch v3':
     <<: *defaults
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,6 +55,9 @@ jobs:
       - run:
           name: Type Checking
           command: yarn run type-check
+      - run:
+          name: Type Export
+          command: yarn run build:types
 
   'type check algoliasearch v3':
     <<: *defaults
--- a/scripts/build/types.js
+++ b/scripts/build/types.js
@@ -61,5 +61,6 @@ if (!result.succeeded) {
       ` and ${result.warningCount} warnings`
   );
 
+  // eslint-disable-next-line no-process-exit
   process.exit(1);
 }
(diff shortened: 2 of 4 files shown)


scripts/build/types.js
```diff
@@ -61,5 +61,6 @@ if (!result.succeeded) {
       ` and ${result.warningCount} warnings`
   );
 
-  process.exitCode = 1;
+  // eslint-disable-next-line no-process-exit
+  process.exit(1);
```
tkrugg, line 65: I don't think this is it. I've tried running this file and it does return code 1.
the issue comes from [here](https://github.com/algolia/instantsearch.js/blob/4ff0392767bfc7a2398b649b2f9362ddb9a8cc75/scripts/release/build-experimental-typescript.js#L23), where the code 1 returned by shell.exec is ignored.

Per [exec](https://www.npmjs.com/package/shelljs#examples) docs, 

```js
// Run external tool synchronously
if (shell.exec('git commit -am "Auto-commit"').code !== 0) {
  shell.echo('Error: Git commit failed');
  shell.exit(1);
}
```
this would effectively make sure that `scripts/release/build-experimental-typescript.js` return code 1 if any of the exec inside do.

Haroenv, line 65:
```suggestion
  process.exitCode = 1;
```

Haroenv: all right, no change is needed then, as the test is now ran separately too, so won't fail only on publish

.circleci/config.yml
```diff
@@ -55,6 +55,9 @@ jobs:
       - run:
           name: Type Checking
           command: yarn run type-check
+      - run:
+          name: Type Export
+          command: yarn run build:types
```
tkrugg, line 60: Does it make sense to add this to the linting step? I think I would have expected this in the build step

scripts/build/types.js
tkrugg: I did that change here: https://github.com/algolia/instantsearch.js/pull/4812

.circleci/config.yml
Haroenv: build is to test the bundle size though, maybe it should be a fully separate step for typescript to both test the build and export?

.circleci/config.yml
```diff
@@ -55,6 +55,9 @@ jobs:
       - run:
           name: Type Checking
           command: yarn run type-check
```
francoischalifour, line 57: Is this step still required?

We seem to be calling `tsc` in `build:types`: https://github.com/algolia/instantsearch.js/blob/364d8f8291252900f3136acfa56f73960029c81b/scripts/build/types.js#L11

.circleci/config.yml
francoischalifour: We could run `build` and `build:types` in the job, and run `test:size` in another job after the `build` job? We do something similar in the Autocomplete CI: https://github.com/algolia/autocomplete/blob/822775c5754ebbd7036658b550e77b9cc966acdb/.circleci/config.yml#L202-L204

.circleci/config.yml
Haroenv: it's with the declaration config, which isn't the same as far as I can tell

.circleci/config.yml
Haroenv: Changed it to build, it makes more sense that way indeed

francoischalifour (approved):

PR #4812, opened 2021-07-15
fix(build): ensure build fails when types building fails

tkrugg:
**Summary**
At the moment failing build types are not being surfaced to circle ci. See for example this [build](https://app.circleci.com/pipelines/github/algolia/instantsearch.js/6416/workflows/3398a5d9-002b-4220-8263-6d45d98cf33d/jobs/30500), step `Trigger a release if the latest commit is a release commit` should have failed but it didn't.

This is because [`yarn built:types`  is run via `shell.exec`](https://github.com/algolia/instantsearch.js/blob/4ff0392767bfc7a2398b649b2f9362ddb9a8cc75/scripts/release/build-experimental-typescript.js#L23) and the return code of 1 is ignored, instead of passed to calling process.

Per [exec](https://www.npmjs.com/package/shelljs#examples) docs, it's the caller's reponsibility to read the exit code and exit or continue.

```js
// Run external tool synchronously
if (shell.exec('git commit -am "Auto-commit"').code !== 0) {
  shell.echo('Error: Git commit failed');
  shell.exit(1);
}
```

This change would effectively make sure that `scripts/release/build-experimental-typescript.js` return code 1 if any of the exec inside do.

NB: `exec` is also called [in other areas](https://github.com/algolia/instantsearch.js/blob/296e9c459a68f8bdfaf82973c6cf5cfcca874c06/ship.config.js#L27) but in this case, it's not from `shelljs` but the one in [shipjs-lib](https://github.com/algolia/shipjs/blob/main/packages/shipjs-lib/src/lib/shell/exec.js) which reads and throws when code is 1.

### DIFF
--- a/scripts/release/build-experimental-typescript.js
+++ b/scripts/release/build-experimental-typescript.js
@@ -19,11 +19,11 @@ fs.writeFileSync(
   `export default '${newVersion}';\n`
 );
 
-if (
-  shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`).code !== 0
-) {
-  shell.exit(1);
-}
-if (shell.exec('yarn build:types').code !== 0) {
+const results = [
+  shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`),
+  shell.exec('yarn build:types'),
+];
+
+if (results.some(result.code !== 0)) {
   shell.exit(1);
 }
--- a/scripts/release/build-experimental-typescript.js
+++ b/scripts/release/build-experimental-typescript.js
@@ -24,6 +24,6 @@ const results = [
   shell.exec('yarn build:types'),
 ];
 
-if (results.some(result.code !== 0)) {
+if (results.some(({ code }) => code !== 0)) {
   shell.exit(1);
 }
--- a/scripts/release/build-experimental-typescript.js
+++ b/scripts/release/build-experimental-typescript.js
@@ -19,5 +19,11 @@ fs.writeFileSync(
   `export default '${newVersion}';\n`
 );
 
-shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`);
-shell.exec('yarn build:types');
+if (
+  shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`).code !== 0
+) {
+  shell.exit(1);
+}
+if (shell.exec('yarn build:types').code !== 0) {
+  shell.exit(1);
+}


scripts/release/build-experimental-typescript.js
```diff
@@ -19,5 +19,11 @@ fs.writeFileSync(
...
-shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`);
-shell.exec('yarn build:types');
+if (
+  shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`).code !== 0
+) {
+  shell.exit(1);
+}
+if (shell.exec('yarn build:types').code !== 0) {
+  shell.exit(1);
+}
```
Haroenv, line 29:
```suggestion
shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`, {
  fatal: true,
});
shell.exec('yarn build:types', {
  fatal: true,
});
```

https://github.com/shelljs/shelljs#execcommand--options--callback

Haroenv (commented):
is this the same but a simpler option?

scripts/release/build-experimental-typescript.js
tkrugg: You're right it's better and I tried it but I couldn't get it to work.
So I looked a the issues and it turns out it's a know issue to be [fixed in 0.9 (not yet released)](https://github.com/shelljs/shelljs/issues/981#issuecomment-626840798).
The workaround apparently is to set it globally, which I didn't want because shelljs is used by other stuff like ship.js and I wasn't sure of the impact there.

Haroenv: wouldn't we want this to be set globally anyway?

tkrugg: I don't think that would be a good idea. As of now, there are two projects depending on it, there might be more in the future (if we introduce new dependencies or some dependencies introduce it).
[Shipjs seems to rely on the fact that it can ignore errors](https://github.com/algolia/shipjs/blob/96d6ff22214849b7c0c665b3d104b00ded43b38a/packages/shipjs-lib/src/lib/git/getCommitBodies.js#L5), so there is a risk of enabling it globally break the release scripts. Nor safe now, or the future.


```sh
yarn why shelljs
yarn why v1.22.10
[1/4] 🤔  Why do we have the module "shelljs"...?
[2/4] 🚚  Initialising dependency graph...
warning Resolution field "algoliasearch@3.35.0" is incompatible with requested version "algoliasearch@^3.35.1"
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "shelljs@0.8.3"
info Has been hoisted to "shelljs"
info This module exists because it's specified in "devDependencies".
info Disk size without dependencies: "316KB"
info Disk size with unique dependencies: "452KB"
info Disk size with transitive dependencies: "1.43MB"
info Number of shared dependencies: 11
=> Found "shipjs-lib#shelljs@0.8.4"
info This module exists because "shipjs#shipjs-lib" depends on it.
info Disk size without dependencies: "316KB"
info Disk size with unique dependencies: "452KB"
info Disk size with transitive dependencies: "1.43MB"
info Number of shared dependencies: 11
=> Found "@storybook/core#shelljs@0.8.4"
info This module exists because "@storybook#html#@storybook#core" depends on it.
info Disk size without dependencies: "348KB"
info Disk size with unique dependencies: "484KB"
info Disk size with transitive dependencies: "1.46MB"
info Number of shared dependencies: 11
✨  Done in 1.08s.
```

Haroenv: I don't understand what you're afraid of, those scripts don't happen at the same time or the same state no? Where are you getting that the settings are persisted outside of the script?

Haroenv:
```suggestion
const results = [
  shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`),
  shell.exec('yarn build:types'),
];

if (results.some(result.code !== 0)) {
  shell.exit(1);
}
```

ok what about this instead then? still checking the return code manually, but leaves visual room for the script

scripts/release/build-experimental-typescript.js
```diff
@@ -19,5 +19,11 @@ fs.writeFileSync(
...
 );
 
-shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`);
-shell.exec('yarn build:types');
+const results = [
+  shell.exec(`NODE_ENV=production VERSION=${newVersion} yarn build`),
+  shell.exec('yarn build:types'),
+];
+
+if (results.some(result.code !== 0)) {
```
tkrugg, line 27:
```suggestion
if (results.some(({ code }) => code !== 0)) {
```

Haroenv (approved):

PR #4814, opened 2021-07-20
feat(types): allow typed access to properties added to entry

Haroenv:
**Summary**

The helpers in the esm entry file were attached to the "prototype" of instantsearch instead of exported, which makes them non-treeshakeable. Removing it now would be a breaking change, but that form can at least already be encouraged.

**Result**

- move the code in main.ts to index.ts for consistency with esm entry file
- deprecate things attached to index.es.ts (for /helpers https://github.com/algolia/instantsearch.js/discussions/4749)


src/lib/utils/logger.ts
```diff
@@ -1,10 +1,5 @@
 import noop from './noop';
 
-type Deprecate<TCallback = (...args: any[]) => any> = (
```
Haroenv, line 3: this previous form caused the generic to already be filled when deprecate gets declared, which is why it must be inline

src/index.es.ts
```diff
@@ -10,18 +9,79 @@ import {
...
+  "import { reverseHighlight } from 'instantsearch.js/es/'"
+);
+instantsearch.snippet = deprecate(
+  snippet,
+  "import { snippet } from 'instantsearch.js/es/'"
+);
+instantsearch.reverseSnippet = deprecate(
+  reverseSnippet,
+  "import { reverseSnippet } from 'instantsearch.js/es/'"
+);
```
eunjae-lee, line 74:
```suggestion
instantsearch.createInfiniteHitsSessionStorageCache = deprecate(
  createInfiniteHitsSessionStorageCache,
  "import { createInfiniteHitsSessionStorageCache } from 'instantsearch.js/es'"
);
instantsearch.highlight = deprecate(
  highlight,
  "import { highlight } from 'instantsearch.js/es'"
);
instantsearch.reverseHighlight = deprecate(
  reverseHighlight,
  "import { reverseHighlight } from 'instantsearch.js/es'"
);
instantsearch.snippet = deprecate(
  snippet,
  "import { snippet } from 'instantsearch.js/es'"
);
instantsearch.reverseSnippet = deprecate(
  reverseSnippet,
  "import { reverseSnippet } from 'instantsearch.js/es'"
);
```

src/index.es.ts
```diff
@@ -10,18 +9,79 @@ import {
   getInsightsAnonymousUserToken,
 } from './helpers';
 import { createInfiniteHitsSessionStorageCache } from './lib/infiniteHitsCache';
+import { deprecate } from './lib/utils';
 
-const instantsearch = (options: InstantSearchOptions): InstantSearch =>
+interface InstantSearchModule {
```
eunjae-lee, line 14: Any reason you went for `interface` instead of `type` here?

eunjae-lee (approved):
This is so much better!

src/index.es.ts
```diff
@@ -10,18 +9,79 @@ import {
...
+ * Two parameters are required to get you started with InstantSearch.js:
+ *  - `indexName`: the main index that you will use for your new search UI
+ *  - `searchClient`: the search client to plug to InstantSearch.js
+ *
+ * The [search client provided by Algolia](https://github.com/algolia/algoliasearch-client-javascript)
+ * needs an `appId` and an `apiKey`. Those parameters can be found in your
+ * [Algolia dashboard](https://www.algolia.com/api-keys).
+ *
+ * If you want to get up and running quickly with InstantSearch.js, have a
+ * look at the [getting started](getting-started.html).
```
francoischalifour, line 48:
```suggestion
 * look at the [getting started](https://www.algolia.com/doc/guides/building-search-ui/getting-started/js/).
```

src/index.es.ts
```diff
@@ -10,18 +9,79 @@ import {
...
+
+/**
+ * InstantSearch is the main component of InstantSearch.js. This object
+ * manages the widget and lets you add new ones.
+ *
+ * Two parameters are required to get you started with InstantSearch.js:
+ *  - `indexName`: the main index that you will use for your new search UI
+ *  - `searchClient`: the search client to plug to InstantSearch.js
+ *
+ * The [search client provided by Algolia](https://github.com/algolia/algoliasearch-client-javascript)
```
francoischalifour, line 43:
```suggestion
 * The [search client provided by Algolia](algolia.com/doc/api-client/getting-started/what-is-the-api-client/javascript/)
```

src/__tests__/index-test.ts
```diff
@@ -19,7 +19,22 @@ describe('instantsearch()', () => {
     });
   });
 
-  it('includes the highlight helper function', () => {
-    expect(instantsearch.highlight).toBeInstanceOf(Function);
+  it('includes the helper functions', () => {
```
francoischalifour, line 22:
```suggestion
  it('includes the API and the helper functions', () => {
```

src/index.es.ts
```diff
@@ -10,18 +9,79 @@ import {
...
+  /** @deprecated */
+  reverseHighlight: typeof reverseHighlight;
+  /** @deprecated */
+  snippet: typeof snippet;
+  /** @deprecated */
+  reverseSnippet: typeof reverseSnippet;
+  /** @deprecated */
+  insights: typeof insights;
+  /** @deprecated */
+  getInsightsAnonymousUserToken: typeof getInsightsAnonymousUserToken;
```
francoischalifour, line 32: We deprecate these two but don't export them.

src/lib/utils/logger.ts
francoischalifour: I don't get this, what do you mean?

Haroenv:
```
let deprecate: Deprecate = ...
```

deprecate is no longer generic there, and I didn't find a way to make a let generic (only a function, type or class)

src/index.es.ts
Haroenv: since they are already deprecated alltogether, I thought I shouldn't add them to  a new entry point, wdyt?

src/index.es.ts
Haroenv: initially I thought callable + properties was only possible on interface, but it looks like type is possible too. Let's go with type for consistency

src/index.ts
```diff
@@ -1,3 +1,66 @@
...
+ * Two parameters are required to get you started with InstantSearch.js:
+ *  - `indexName`: the main index that you will use for your new search UI
+ *  - `searchClient`: the search client to plug to InstantSearch.js
+ *
+ * The [search client provided by Algolia](https://github.com/algolia/algoliasearch-client-javascript)
+ * needs an `appId` and an `apiKey`. Those parameters can be found in your
+ * [Algolia dashboard](https://www.algolia.com/api-keys).
+ *
+ * If you want to get up and running quickly with InstantSearch.js, have a
+ * look at the [getting started](getting-started.html).
```
Haroenv, line 45:
```suggestion
 * look at the [getting started](https://www.algolia.com/doc/guides/building-search-ui/getting-started/js/).
```

src/index.ts
```diff
@@ -1,3 +1,66 @@
...
+
+/**
+ * InstantSearch is the main component of InstantSearch.js. This object
+ * manages the widget and lets you add new ones.
+ *
+ * Two parameters are required to get you started with InstantSearch.js:
+ *  - `indexName`: the main index that you will use for your new search UI
+ *  - `searchClient`: the search client to plug to InstantSearch.js
+ *
+ * The [search client provided by Algolia](https://github.com/algolia/algoliasearch-client-javascript)
```
Haroenv, line 40:
```suggestion
 * The [search client provided by Algolia](algolia.com/doc/api-client/getting-started/what-is-the-api-client/javascript/)
```

src/index.es.ts
francoischalifour: Ah right, I forgot they were deprecated!

src/index.es.ts
Haroenv, line 14:
```suggestion
type InstantSearchModule {
```

src/index.ts
```diff
@@ -1,3 +1,66 @@
...
+import * as connectors from './connectors/index';
+import * as widgets from './widgets/index';
+import * as helpers from './helpers/index';
+import * as middlewares from './middlewares/index';
+
+import * as routers from './lib/routers/index';
+import * as stateMappings from './lib/stateMappings/index';
+import { createInfiniteHitsSessionStorageCache } from './lib/infiniteHitsCache/index';
+
+interface InstantSearchModule {
```
Haroenv, line 13:
```suggestion
type InstantSearchModule {
```

src/index.es.ts
```diff
@@ -10,18 +9,71 @@ import {
...
 import { createInfiniteHitsSessionStorageCache } from './lib/infiniteHitsCache';
+import { deprecate } from './lib/utils';
 
-const instantsearch = (options: InstantSearchOptions): InstantSearch =>
+type InstantSearchModule = {
+  (options: InstantSearchOptions): InstantSearch;
+  version: string;
+
+  // @major remove these in favour of the exports
+  /** @deprecated */
```
Haroenv, line 19: @francoischalifour suggested *not* deprecating these yet, as we aren't sure where the export should best be from. At the moment it's /es/helpers, but that could change? 

Personally I'm more of the opinion to deprecate them with /es/helpers as a workaround, still leaving the option open for a more thorough review of what's exported where in a major version

wdyt @tkrugg / @shortcuts

src/lib/utils/logger.ts
Haroenv: https://github.com/microsoft/TypeScript/issues/17574

src/index.es.ts
shortcuts: As long as we have clear warnings etc., I don't see any issue deprecating them

shortcuts (approved):
Really nice!

francoischalifour:
The thing is we'll likely update/clean our exports during the React InstantSearch migration, so I'm not sure that's the right time for this change (also definitely an improvement) because it will make it harder to not introduce breaking changes.

Haroenv:
Changing exports is always a breaking change @francoischalifour, right? I know for sure that we don't want to continue exposing the helpers as a property on the es entry point (it's not documented), the deprecation message mentions /es/helpers, as I don't think it's useful to discourage use without explaining what a right way of using it is. 

This PR mainly focuses on having the right types for something that already is the case, whether it's typed or not, removing those entries is a breaking change :)

francoischalifour:
Sorry I got confused and forgot that the PR changed from updating the exports to only updating typings—all good then!

francoischalifour (approved):

PR #4816, opened 2021-07-28
feat(ts): allow custom ui state and route state in routing

Haroenv:
## InstantSearch generic:

```ts
import algoliasearch from 'algoliasearch/lite';
import instantsearch from 'instantsearch.js/es';
import { history } from 'instantsearch.js/es/lib/routers';
import { StateMapping, UiState } from 'instantsearch.js/es/types';

type SwagIndexUiState = { swag: boolean };
type SwagUiState = { [indexName: string]: SwagIndexUiState };

// can also be written inline, no need for the type then
const stateMapping: StateMapping<SwagUiState, SwagUiState> = {
  stateToRoute(uiState) {
    return Object.keys(uiState).reduce(
      (state, indexId) => ({
        ...state,
        [indexId]: { swag: uiState[indexId].swag },
      }),
      {}
    );
  },

  routeToState(routeState = {}) {
    return Object.keys(routeState).reduce(
      (state, indexId) => ({
        ...state,
        [indexId]: routeState[indexId],
      }),
      {}
    );
  },
};

instantsearch<SwagUiState, SwagUiState>({
  indexName: '',
  searchClient: algoliasearch('', ''),
  routing: {
    router: history(),
    stateMapping,
  },
});
```

## generic middleware

```ts
import instantsearch from 'instantsearch.js/es'
import { history } from 'instantsearch.js/es/lib/routers';
import { createRouterMiddleware } from 'instantsearch.js/es/middlewares';
import { StateMapping, UiState } from 'instantsearch.js/es/types';

type SwagIndexUiState = { swag: boolean };
type SwagUiState = { [indexName: string]: SwagIndexUiState };

const stateMapping: StateMapping<SwagUiState, SwagUiState> = {
  stateToRoute(uiState) {
    return Object.keys(uiState).reduce(
      (state, indexId) => ({
        ...state,
        [indexId]: { swag: uiState[indexId].swag },
      }),
      {}
    );
  },

  routeToState(routeState = {}) {
    return Object.keys(routeState).reduce(
      (state, indexId) => ({
        ...state,
        [indexId]: routeState[indexId],
      }),
      {}
    );
  },
};

search.use(
  createRouterMiddleware<SwagUiState, SwagUiState>({
    router: history(),
    stateMapping,
  })
);
search.addWidgets([instantsearch.widgets.hits({ container })]);
```

## references

fixes DX-2298

### DIFF
--- a/src/index.es.ts
+++ b/src/index.es.ts
@@ -12,10 +12,10 @@ import {
 import { createInfiniteHitsSessionStorageCache } from './lib/infiniteHitsCache';
 
 const instantsearch = <
-  TUiState extends UiState = UiState,
+  TUiState = Record<string, unknown>,
   TRouteState = TUiState
 >(
-  options: InstantSearchOptions<TUiState, TRouteState>
+  options: InstantSearchOptions<UiState & TUiState, TRouteState>
 ) => new InstantSearch(options);
 
 instantsearch.version = version;
--- a/src/index.es.ts
+++ b/src/index.es.ts
@@ -16,7 +16,7 @@ const instantsearch = <
   TRouteState = TUiState
 >(
   options: InstantSearchOptions<TUiState, TRouteState>
-): InstantSearch => new InstantSearch(options);
+) => new InstantSearch(options);
 
 instantsearch.version = version;
 instantsearch.snippet = snippet;
--- a/src/index.es.ts
+++ b/src/index.es.ts
@@ -1,4 +1,4 @@
-import { InstantSearchOptions } from './types';
+import { InstantSearchOptions, UiState } from './types';
 import InstantSearch from './lib/InstantSearch';
 import version from './lib/version';
 import {
@@ -11,8 +11,12 @@ import {
 } from './helpers';
 import { createInfiniteHitsSessionStorageCache } from './lib/infiniteHitsCache';
 
-const instantsearch = (options: InstantSearchOptions): InstantSearch =>
-  new InstantSearch(options);
+const instantsearch = <
+  TUiState extends UiState = UiState,
+  TRouteState = TUiState
+>(
+  options: InstantSearchOptions<TUiState, TRouteState>
+): InstantSearch => new InstantSearch(options);
 
 instantsearch.version = version;
 instantsearch.snippet = snippet;
--- a/src/lib/__tests__/RoutingManager-test.ts
+++ b/src/lib/__tests__/RoutingManager-test.ts
@@ -4,7 +4,13 @@ import qs from 'qs';
 import { createSearchClient } from '../../../test/mock/createSearchClient';
 import { createWidget } from '../../../test/mock/createWidget';
 import { runAllMicroTasks } from '../../../test/utils/runAllMicroTasks';
-import { Router, Widget, UiState, StateMapping, RouteState } from '../../types';
+import {
+  Router,
+  Widget,
+  UiState,
+  StateMapping,
+  IndexUiState,
+} from '../../types';
 import historyRouter from '../routers/history';
 import instantsearch from '../main';
 
@@ -383,7 +389,7 @@ describe('RoutingManager', () => {
       const router = createFakeRouter({
         onUpdate(fn) {
           history.subscribe(state => {
-            fn(state);
+            fn(state as UiState);
           });
         },
         write: jest.fn(state => {
@@ -475,7 +481,7 @@ describe('RoutingManager', () => {
       const router = createFakeRouter({
         onUpdate(fn) {
           history.subscribe(state => {
-            fn(state);
+            fn(state as UiState);
           });
         },
         write: jest.fn(state => {
@@ -549,10 +555,10 @@ describe('RoutingManager', () => {
       const searchClient = createSearchClient();
       const stateMapping = createFakeStateMapping({});
       const router = historyRouter({
-        windowTitle(routeState: RouteState) {
+        windowTitle(routeState) {
           return `Searching for "${routeState.query}"`;
         },
-      } as any);
+      });
 
       const search = instantsearch({
         indexName: 'instant_search',
@@ -596,7 +602,7 @@ describe('RoutingManager', () => {
         url: createFakeUrlWithRefinements({ length: 22 }),
       });
 
-      const router = historyRouter();
+      const router = historyRouter<IndexUiState>();
       // @ts-expect-error: This method is considered private but we still use it
       // in the test after the TypeScript migration.
       // In a next refactor, we can consider changing this test implementation.
@@ -605,7 +611,7 @@ describe('RoutingManager', () => {
         location: window.location,
       });
 
-      expect(parsedUrl.refinementList.brand).toBeInstanceOf(Array);
+      expect(parsedUrl.refinementList!.brand).toBeInstanceOf(Array);
       expect(parsedUrl).toMatchInlineSnapshot(`
         Object {
           "refinementList": Object {
@@ -643,7 +649,7 @@ describe('RoutingManager', () => {
         url: createFakeUrlWithRefinements({ length: 100 }),
       });
 
-      const router = historyRouter();
+      const router = historyRouter<IndexUiState>();
       // @ts-expect-error: This method is considered private but we still use it
       // in the test after the TypeScript migration.
       // In a next refactor, we can consider changing this test implementation.
@@ -652,13 +658,13 @@ describe('RoutingManager', () => {
         location: window.location,
       });
 
-      expect(parsedUrl.refinementList.brand).toBeInstanceOf(Array);
+      expect(parsedUrl.refinementList!.brand).toBeInstanceOf(Array);
     });
   });
 
   describe('createURL', () => {
     it('returns an URL for a `routeState` with refinements', () => {
-      const router = historyRouter();
+      const router = historyRouter<IndexUiState>();
       const actual = router.createURL({
         query: 'iPhone',
         page: 5,
--- a/src/lib/__tests__/RoutingManager-test.ts
+++ b/src/lib/__tests__/RoutingManager-test.ts
@@ -41,32 +41,30 @@ const createFakeStateMapping = (
   ...args,
 });
 
-type Entry = Record<string, unknown>;
-
-type HistoryState = {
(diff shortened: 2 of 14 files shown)


francoischalifour (commented):
Can we let users not have to provide `UiState`, but we augment it on our side?

src/index.es.ts
```diff
@@ -11,8 +11,12 @@ import {
...
 import { createInfiniteHitsSessionStorageCache } from './lib/infiniteHitsCache';
 
-const instantsearch = (options: InstantSearchOptions): InstantSearch =>
-  new InstantSearch(options);
+const instantsearch = <
+  TUiState extends UiState = UiState,
+  TRouteState = TUiState
+>(
+  options: InstantSearchOptions<TUiState, TRouteState>
+): InstantSearch => new InstantSearch(options);
```
Haroenv, line 19:
```suggestion
const instantsearch = <
  TUiState = Record<string, unknown>,
  TRouteState = TUiState
>(
  options: InstantSearchOptions<UiState & TUiState, TRouteState>
): InstantSearch => new InstantSearch(options);
```

src/lib/main.ts
```diff
@@ -28,8 +28,12 @@ import { InstantSearchOptions } from '../types';
...
  * @param {InstantSearchOptions} options The options
  */
-const instantsearch = (options: InstantSearchOptions) =>
-  new InstantSearch(options);
+const instantsearch = <
+  TUiState extends UiState = UiState,
+  TRouteState = TUiState
+>(
+  options: InstantSearchOptions<TUiState, TRouteState>
+) => new InstantSearch(options);
```
Haroenv, line 36:
```suggestion
const instantsearch = <
  TUiState = Record<string, unknown>,
  TRouteState = TUiState
>(
  options: InstantSearchOptions<UiState & TUiState, TRouteState>
) => new InstantSearch(options);
```

Haroenv (commented):
is this what you expect @francoischalifour ?

src/index.es.ts
```diff
@@ -11,8 +11,12 @@ import {
 } from './helpers';
 import { createInfiniteHitsSessionStorageCache } from './lib/infiniteHitsCache';
 
-const instantsearch = (options: InstantSearchOptions): InstantSearch =>
-  new InstantSearch(options);
+const instantsearch = <
+  TUiState extends UiState = UiState,
+  TRouteState = TUiState
+>(
+  options: InstantSearchOptions<TUiState, TRouteState>
```
Haroenv, line 18:
```suggestion
const instantsearch = <
  TUiState = Record<string, unknown>,
  TRouteState = TUiState
>(
  options: InstantSearchOptions<UiState & TUiState, TRouteState>
```

src/index.es.ts
Haroenv, line 19:
```suggestion
) => new InstantSearch(options);
```

this is inferred and otherwise loses the generic

src/lib/main.ts
francoischalifour: Yes!

src/index.es.ts
```diff
@@ -11,8 +11,12 @@ import {
 } from './helpers';
 import { createInfiniteHitsSessionStorageCache } from './lib/infiniteHitsCache';
 
-const instantsearch = (options: InstantSearchOptions): InstantSearch =>
-  new InstantSearch(options);
+const instantsearch = <
+  TUiState = Record<string, unknown>,
+  TRouteState = TUiState
+>(
+  options: InstantSearchOptions<UiState & TUiState, TRouteState>
```
Haroenv, line 18: I was just thinking, should the argument maybe be TIndexUiState (so you can & all custom widgets together). The downside I see is that would mean the argument TRouteState likely still would be a full ui state (key per index)

francoischalifour: How would we get the full UI state typing if users only provide `IndexUiState`?

Haroenv: It would be possible like this: `InstantSearchOptions<UiState & { [indexName: string]: TIndexUiState] }, TRouteState>`

francoischalifour: And what if you have multiple indices that don't share the same index UI state?

Haroenv: Good point, let's keep it as suggested in the PR then! UiState is essentially always optional, but I see why you'd want that

src/lib/stateMappings/simple.ts
```diff
@@ -1,32 +1,36 @@
...
     stateToRoute(uiState) {
-      return Object.keys(uiState).reduce<RouteState>(
+      return Object.keys(uiState).reduce(
         (state, indexId) => ({
           ...state,
           [indexId]: getIndexStateWithoutConfigure(uiState[indexId]),
         }),
-        {}
+        {} as TUiState
       );
```
francoischalifour, line 24: Would this work to avoid casting the initial value?

```ts
Object.keys(uiState).reduce<TUiState>(
  (state, indexId) => ({
    ...state,
    [indexId]: getIndexStateWithoutConfigure(uiState[indexId]),
  }),
  {}
);
```

src/middlewares/createRouterMiddleware.ts
```diff
@@ -5,45 +5,59 @@ import {
...
+  TUiState extends UiState = UiState,
+  TRouteState = TUiState
+>(
+  props: RouterProps<TUiState, TRouteState> = {}
+): InternalMiddleware<TUiState> => {
   const {
-    router = historyRouter(),
-    stateMapping = simpleStateMapping(),
+    router = historyRouter<TRouteState>(),
+    // technically this is wrong, as without stateMapping parameter given, the routeState *must* be UiState
```
francoischalifour, line 28: Can you explain a bit more?

src/lib/__tests__/RoutingManager-test.ts
```diff
@@ -383,7 +389,7 @@ describe('RoutingManager', () => {
       const router = createFakeRouter({
         onUpdate(fn) {
           history.subscribe(state => {
-            fn(state);
+            fn(state as UiState);
```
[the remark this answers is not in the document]
Haroenv: it's actually history which has the wrong type (Entry)

src/lib/stateMappings/simple.ts
Haroenv: no, as TUiState doesn't necessarily allows empty object (but we assume it does),

```
Argument of type '{}' is not assignable to parameter of type 'TUiState'.
  '{}' is assignable to the constraint of type 'TUiState', but 'TUiState' could be instantiated with a different subtype of constraint 'UiState'
```

src/middlewares/createRouterMiddleware.ts
Haroenv: simpleStateMapping is a `StateMapping<UiState, UiState>`, and `createRouterMiddleware` can be called with a different generic than UiState, so the types don't really match. The only thing I can think of otherwise is only conditionally allowing stateMapping to be conditional, but I couldn't find how

Haroenv: I tried things with branding, but still could not get anything working as soon as simpleStatemapping is actually used (as it uses UiState (`const stateMapping: StateMapping<UiState, TRouteState> | StateMapping<TRouteState, TRouteState>` becomes a union, and even though in that case both types _are_ the same, typescript doesn't seem to be able to know that)

<details>
<summary>I tried</summary>

```ts

export type RouterProps<
  TUiState extends UiState = UiState,
  TRouteState = UiState & { ___default: true }
> = TUiState extends { ___default: true }
  ? {
      router?: Router<TRouteState>;
      stateMapping?: never;
    }
  : {
      router?: Router<TRouteState>;
      stateMapping: StateMapping<TUiState, TRouteState>;
    };

export const createRouterMiddleware = <
  TUiState extends UiState = UiState,
  TRouteState = UiState & { ___default: true }
>(
  props: RouterProps<TUiState, TRouteState> = {} as RouterProps<
    TUiState,
    TRouteState
  >
): InternalMiddleware<TUiState> => {
  const {
    router = historyRouter<TRouteState>(),
    stateMapping = simpleStateMapping<TRouteState>(),
  } = props;
```

and variations of it

</details>

src/middlewares/createRouterMiddleware.ts
```diff
@@ -5,45 +5,64 @@ import {
...
+  props: RouterProps<TUiState, TRouteState> = {}
+): InternalMiddleware<TUiState> => {
   const {
-    router = historyRouter(),
-    stateMapping = simpleStateMapping(),
+    router = historyRouter<TRouteState>(),
+    // We have to cast simpleStateMapping as a StateMapping<TUiState, TRouteState>.
+    // this is needed because simpleStateMapping is StateMapping<TUiState, TUiState>.
+    // While it's only used when UiState and RouteState are the same, unfortunately
+    // TypeScript still considers them separate types.
```
Haroenv, line 33: new comment @francoischalifour

src/types/middleware.ts
```diff
@@ -2,8 +2,8 @@ import InstantSearch from '../lib/InstantSearch';
 import { UiState } from './ui-state';
 import { AtLeastOne } from './utils';
 
-export type MiddlewareDefinition = {
-  onStateChange(options: { uiState: UiState }): void;
+export type MiddlewareDefinition<TUiState extends UiState = UiState> = {
+  onStateChange(options: { uiState: TUiState }): void;
```
francoischalifour, line 6: Isn't it supposed to be this here too?

```suggestion
export type MiddlewareDefinition<TUiState = Record<string, unknown>> = {
  onStateChange(options: { uiState: UiState & TUiState }): void;
```

Haroenv: I was thinking every place except the root only needs to deal with UiState and likely no extensions (not really user code), so all except the entry point expects the consumer to make the mix

francoischalifour: Right, that looks fair like that!

francoischalifour: Right, that looks fair like that!

francoischalifour (approved):

Haroenv:
Right, that looks fair like that!

francoischalifour:
Right, that looks fair like that!

PR #4817, opened 2021-07-30
chore(ts): avoid more ts-ignores

Haroenv:
ts-expect-error will error if it's no longer needed, so it's better in almost all cases. There's some moments where we need to keep ts-ignore though:
- things that fail only with algoliasearch v3 or v4
- optional packages (places)
- code that doesn't always error

At the same time updated the algoliasearch-helper to a version that includes `showParentLevel` in the typings

part of DX-2299

### DIFF
--- a/src/widgets/answers/__tests__/answers-test.ts
+++ b/src/widgets/answers/__tests__/answers-test.ts
@@ -8,12 +8,13 @@ import { runAllMicroTasks } from '../../../../test/utils/runAllMicroTasks';
 import { wait } from '../../../../test/utils/wait';
 import answers from '../answers';
 import searchBox from '../../search-box/search-box';
+import { createInitOptions } from '../../../../test/mock/createWidget';
 
 describe('answers', () => {
   describe('Usage', () => {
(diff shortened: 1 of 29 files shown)


src/widgets/answers/__tests__/answers-test.ts
```diff
@@ -47,18 +48,20 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/answers/js/
         indexName: 'instant_search',
         searchClient,
       });
+      search.helper = helper;
```
francoischalifour, line 51: Why did we have to do this?

francoischalifour (approved):

src/widgets/answers/__tests__/answers-test.ts
Haroenv: createInitOptions gets the state from the InstantSearch instance (before it was undefined), in theory we could also have grabbed it from the helper argument, but that isn't such a big difference (usually they will be the same)

shortcuts:
That's a satisfying PR

PR #4820, opened 2021-08-02
fix(types): fixing hits and results types in connectHits and connectInfiniteHits

tkrugg:
####  `connectHits`

The types `Hit` and `AlgoliaHit` are mostly the same.

https://github.com/algolia/instantsearch.js/blob/21b6c53d4b1a3f3dd3a2375a5cfe0a31bef6ded6/src/types/results.ts#L60-L63

As of today, the hits exposed via results indeed [contain __position and __queryID](https://github.com/algolia/instantsearch.js/blob/21b6c53d4b1a3f3dd3a2375a5cfe0a31bef6ded6/src/connectors/hits/connectHits.ts#L158).

####  `connectInfiniteHits`

While trying use `InfiniteHitsRenderState` in Angular InstantSearch I noticed how `hits` and `results` while being [part of the state expored by`connectInfiniteHits`](https://github.com/algolia/instantsearch.js/blob/master/src/connectors/infinite-hits/connectInfiniteHits.ts#L331-L342) are not actually visible on the type.

### DIFF
--- a/src/connectors/hits/connectHits.ts
+++ b/src/connectors/hits/connectHits.ts
@@ -16,7 +16,6 @@ import {
   Connector,
   Hits,
   Hit,
-  AlgoliaHit,
   WidgetRenderState,
 } from '../../types';
 import { SearchResults } from 'algoliasearch-helper';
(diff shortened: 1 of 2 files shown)


Haroenv (approved):
looks like this was an oversight!

PR #4829, opened 2021-08-13
fix(ts): correct entry point

Haroenv:
The entry point used to be:

```ts
// this is wrong!!!
export { default } from './index.es';
export * from './connectors';
export * from './helpers';
export * from './types';
export * from './widgets';
export * from './lib/routers';
```

While really the entry point should be the same as index.es.ts, as otherwise the js bundle doesn't match the declaration.

This commit fixes that by no longer generating an entry point, and renaming index.es.d.ts to index.d.ts in the built files.

In the same commit I also move the declaration-related configuration files to the subfolder to make the root a little cleaner

### DIFF
--- a/scripts/build/api-extractor.json
+++ b/scripts/build/api-extractor.json
@@ -1,6 +1,12 @@
 {
   "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
 
+  "projectFolder": "../../",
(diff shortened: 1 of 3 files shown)


eunjae-lee (approved):
I didn't get what you meant at first, but looking at the current build output, this PR looks good to me!

image
(↑ before this PR)

PR #4830, opened 2021-08-13
refactor(tests): avoid global jsdom declaration

Haroenv:
**Summary**

we added jsdom as a global in our files, but barely used it.

**Result**

declare jsdom explicitly in the file that's using it for jsdom.redeclare


package.json
```diff
@@ -82,6 +82,7 @@
     "@types/enzyme": "^3.1.15",
     "@types/jest": "^26.0.22",
     "@types/jest-diff": "^24.3.0",
+    "@types/jsdom": "16.2.13",
```
francoischalifour, line 85: We usually don't pin the type deps.

francoischalifour (approved):

package.json
Haroenv: ah yes good point

PR #4832, opened 2021-08-16
feat(typescript): expose types at regular build

### LINKED ISSUE #4340: Type definitions for Typescript
Hi! How far are you with type checking the project? Cause I am currently trying to integrate instantsearch for a project utilizing a typescript front-end framework and getting a lot of type errors when building the frontend.

Haroenv:
**Summary**

The typescript migration is complete

This PR moves away from -experimental and to stable.

**Result**

InstantSearch is now shipped with type definitions built-in!

Note that, like every type change, this could be a breaking change, as manually defined definitions will now merge with the ones provided by InstantSearch.

In almost all cases manual definitions or changes made are no longer needed, if you do however see any problems, please open an issue or pull request with reproduction.

If you were using -experiemental-typescript tag up to now, that can be changed to a regular InstantSearch version from this release onwards.

fixes #4340


scripts/typescript/api-extractor.json
```diff
@@ -32,10 +34,6 @@
         "addToApiReportFile": true
       },
 
-      "ae-internal-missing-underscore": {
```
Haroenv, line 35: this didn't actually error anymore since a while already

.circleci/config.yml
```diff
@@ -37,13 +37,15 @@ jobs:
...
-          command: yarn run build
+          command: yarn run build:umd
       - run:
           name: Test packages size
           command: yarn run test:size
       - run:
           name: Type Export
-          command: yarn run build:types
+          command: |
+            yarn run build:es
```
Haroenv, line 47: this way the cjs isn't needlessly built, but could also run those two build steps together

francoischalifour: We don't catch if the CJS build fails anymore. Is this a problem?

francoischalifour (approved):
Huge milestone! 🎉

.circleci/config.yml
Haroenv: didn't think of that, I'll add it back

.circleci/config.yml
```diff
@@ -37,7 +37,10 @@ jobs:
       - save_cache: *save_yarn_cache
       - run:
           name: Build library
-          command: yarn run build
+          command: |
+            yarn run build:umd
+            yarn run build:cjs
+            yarn run build:es
```
eunjae-lee, line 43: nice touch

eunjae-lee (approved):
Let's go 🚀

PR #4834, opened 2021-08-17
feat(ts): export types from entry point

Haroenv:
**Summary**

Types are useful outside of InstantSearch for for example Vue InstantSearch, React InstantSearch, so we want to expose this from the root

**Result**

This was changed in https://github.com/algolia/instantsearch.js/pull/4829, as i thought it was a mistake (like most of the other entries in that fake file), but /types is still useful


francoischalifour (approved):
Thanks!

PR #4837, opened 2021-08-17
fix(types): export all types as "type" to avoid exporting in .js

Haroenv:
**Summary**

This issue is caused by `export * from './types'` added in https://github.com/algolia/instantsearch.js/pull/4834, which causes everything in /types to be exposed, but unfortunately because `export from` in the InstantSearch file mixed a type and a value, and thus exported it.

This is done by enabling [consistent-type-imports](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/consistent-type-imports.md) in eslint to make sure it's correct in all cases

references:
- code that throws: https://unpkg.com/browse/instantsearch.js@4.27.1/es/types/instantsearch.js
- vite reproduction: https://github.com/eunjae-lee/vue-instantsearch-with-vue3-and-vite

**Result**

all type imports are actually `import type` or `export type` now


Haroenv:
looks like `h` gets considered as a type-only use, even though it's used by jsx, I couldn't find any open issues about it :(

eunjae-lee (approved):
Looks good and works fine on my mac

PR #4839, opened 2021-08-19
fix(svg): remove xmlns

Haroenv:
**Summary**

xml namespace is not required for svg elements (and has never been https://stackoverflow.com/a/18468348/3185307), only for svg documents.

**Result**

- This saves a probably insignificant amount of bundlesize
- xmlns nowhere left


eunjae-lee (approved):

Haroenv:
bundlesize saved is .04 kb by the way :D

PR #4840, opened 2021-08-19
refactor(sendEventForHits): remove __escaped more efficiently

Haroenv:
**Summary**

Mapping over the hits isn't heavy, but it's not required, as .slice also removes the __escaped key on hits

**Result**

replaced .map(identity) with .slice() to clone the hits


eunjae-lee (approved):
makes sense!

PR #4841, opened 2021-08-19
fix(sendEvent): split > 20 objects in multiple calls

Haroenv:
**Summary**

the backend api doesn't allow > 20 hits per event, so we split the sendEvent calls up into multiple if > 20 DX-2379

**Result**

edited sendEvent and bindEvent to both split up the vents in arrays where it makes sense, and calling sendEvent multiple times.

We choose to split at this level and not inside the events listener, as you can't know what method is used, and you'd need a lot of conditional code to handle all chunking

### DIFF
--- a/src/connectors/hits/__tests__/connectHits-test.ts
+++ b/src/connectors/hits/__tests__/connectHits-test.ts
@@ -846,26 +846,28 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
           expect(payload.startsWith('data-insights-event=')).toBe(true);
           expect(
             deserializePayload(payload.substr('data-insights-event='.length))
-          ).toEqual({
-            eventType: 'click',
-            hits: [
-              {
-                __position: 0,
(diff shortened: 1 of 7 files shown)


eunjae-lee (approved):
For me and others viewing this in the future, there are some constraints that force us to call sendEvent multiple times:

1. `insightsClient()` takes only a single event (`search-insights.js` has to support it first.)
2. `onEvent` hook in the insights middleware receives a single event.

I guess we should've seen this coming and had plural events in the code from the beginning.

JessycaFrederick:
I'm still experiencing this problem. I'm using InstantSearch.js v 4.43.1. Please let me know what other information you need.

JessycaFrederick:
Thanks @Haroenv ... it turns out I wasn't using the latest version of InstantSearch.js and I was confused by the naming conventions. Also, I copied the script from the docs which were out of date. I believe support has taken care of it and I'm no longer getting the errors.

PR #4843, opened 2021-08-23
test(jest): opt-in to shorter snapshot format

Haroenv:
**Summary**

Changing from snapshot to pure expect is useful sometimes if you know the assertion will be pretty hard to write, but before this PR requires a lot of editing. Since we use this shorter format for snapshots, no more "Array" or "Object" will be printed in snapshots.

**Result**

- no more `Object` or `Array` in snapshots 🙌 
- latest jest (setImmediate no longer exists in browser tasks, causing removal of runAllMicroTasks)

This PR is easier to review as their individual commits


src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -38,6 +37,7 @@ type PaginationWidgetInstance = Widget<
 >;
 
 jest.useFakeTimers();
+jest.setTimeout(10);
```
eunjae-lee, line 40: What is this for?

src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -417,7 +417,7 @@ See https://www.algolia.com/doc/api-reference/widgets/configure/js/`);
 
     search.start();
 
-    await runAllMicroTasks();
+    await Promise.resolve();
```
eunjae-lee, line 420: What if we use `wait(0)` here too?
Or does it make sense to bring `runAllMicroTasks` back and replace its implementation with `wait()`?

src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -1047,7 +1047,8 @@ describe('dispose', () => {
...
+    await Promise.resolve();
```
eunjae-lee, line 1051: Calling it once didn't pass the test?

Haroenv: yes, there's two micro-tasks that need to happen, this implementation makes that clearer than wait(0) IMO

src/lib/__tests__/InstantSearch-test.tsx
Haroenv: we definitely can use `wait(0)`, but I'm not sure if it really means the same, this code happens before a timeout runs

src/lib/__tests__/InstantSearch-test.tsx
Haroenv: testing, that should be removed!

src/lib/__tests__/InstantSearch-test.tsx
eunjae-lee: oh I see, but I feel like this is testing too much detail here because what we want to know is if `onRender` is called after rendering happens. We do not want to test if onRender is called two micro-tasks after start().

So I'm wondering what if we 

* `const runAllMicroTasks = () => wait(0)` and use it normally
* use `await Promise.resolve()` when we need to test specific timing issue?

src/lib/__tests__/InstantSearch-test.tsx
eunjae-lee: While I understand what you mean, I fear this subtleness that might not make any real life difference could complicate writing tests.

Haroenv: makes sense, I'll check whether those tests can be done with wait too

src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -1220,20 +1216,15 @@ describe('scheduleStalledRender', () => {
...
     // Trigger a new search
     search.mainHelper!.search();
 
     // Reaches the delay
-    jest.runOnlyPendingTimers();
-
-    // Wait for the `render`
-    await Promise.resolve();
-    await Promise.resolve();
+    await wait(search._stalledSearchDelay);
```
eunjae-lee, line 1227: nice touch

eunjae-lee (approved):

PR #4844, opened 2021-08-23
feat(ts): expose built files in umd

Haroenv:
**Summary**

Exposes the entry point for umd for each of the umd entry points.

This is achieved using the dtsRollup feature of api-extractor, which isn't needed for the es build, as local files work perfectly fine in that case.

Make sure to check the generated files when reviewing this PR

**Result**

A situation like https://github.com/algolia/create-instantsearch-app/pull/464 is done easier by simply declaring what's import from the umd

### DIFF
--- a/scripts/typescript/extract.js
+++ b/scripts/typescript/extract.js
@@ -19,12 +19,6 @@ shell.sed(
   path.join(__dirname, '../../es/**/*.d.ts')
 );
 
-// expose only the es entry point, not the umd entry point
-shell.mv(
-  path.join(__dirname, '../../es/index.es.d.ts'),
-  path.join(__dirname, '../../es/index.d.ts')
-);
-
 console.log();
 console.log(`Validating definitions...`);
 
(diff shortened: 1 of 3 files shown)


Haroenv:
https://codesandbox.io/s/instantsearch-js-forked-34p6s?file=/src/app.js proves that it works 🙌

scripts/typescript/extract.js
```diff
@@ -47,3 +41,24 @@ if (!result.succeeded) {
...
+  path.join(__dirname, '../../dist/instantsearch.production.d.ts'),
+  path.join(__dirname, '../../dist/instantsearch.production.min.d.ts')
+);
+shell.cp(
+  path.join(__dirname, '../../dist/instantsearch.production.d.ts'),
+  path.join(__dirname, '../../dist/instantsearch.development.d.ts')
+);
+shell.cp(
+  path.join(__dirname, '../../dist/instantsearch.production.d.ts'),
+  path.join(__dirname, '../../dist/instantsearch.development.min.d.ts')
```
eunjae-lee, line 63: in which case do people need development.d.ts file?

Haroenv: all the files are the same, but typescript works by matching a .d.ts file to a .js file, so if you're importing /dist/development.min.js then the .d.ts is needed for that to work

eunjae-lee: Thanks for explaining!

eunjae-lee (approved):

francoischalifour (approved):

PR #4845, opened 2021-08-24
feat(panel): render templates on init with render state

Haroenv:
**Summary**

Before this PR the initial render happens *before* widget init. This doesn't have a huge effect, although it got rendered with just an empty object. This makes things needlessly dynamic (more than the types were saying even, because the Template isn't super strict), and would make a template like `header({ widgetParams }) { return widgetParams.attribute }` throw, even though with this PR it is possible without conditionals or flashing.

Under very strict conditions this could be construed as a breakign change, although it's closer to a fix, therefore I have classified it as a new feature.

**Result**

panel's templates always get called with render state, but also with init options. Before this PR it got called with either results or an empty object

### DIFF
--- a/src/widgets/panel/panel.tsx
+++ b/src/widgets/panel/panel.tsx
@@ -11,7 +11,13 @@ import {
 import { component } from '../../lib/suit';
 import type { PanelComponentCSSClasses } from '../../components/Panel/Panel';
 import Panel from '../../components/Panel/Panel';
-import type { Template, RenderOptions, WidgetFactory } from '../../types';
+import type {
+  Template,
+  RenderOptions,
+  WidgetFactory,
+  InitOptions,
+  Widget,
+} from '../../types';
 
 export type PanelCSSClasses = Partial<{
   /**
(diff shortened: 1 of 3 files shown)


src/widgets/panel/panel.tsx
```diff
@@ -265,32 +283,37 @@ const panel: PanelWidget = (panelWidgetParams) => {
...
+          options,
+          hidden: true,
+          collapsible,
+          collapsed: false,
+        });
+
+        if (typeof widget.init === 'function') {
+          widget.init.call(this, ...args);
+        }
       },
```
Haroenv, line 317: this is the main change, the rest is mostly types and tests

francoischalifour (approved):
I think we initially wanted to display the header and footer before the widget's `init` method is called, but I don't really see why. Having consistent data passed to the templates is more important imo. Besides, the `panel` widget is not a side effect anymore.

Haroenv:
I think originally it might have made sense together with the `hidden` function, as that one only gets called with results, but still it was inconsistent as it either got called with empty object, either with the results, the templates become less conditional with this PR :)

PR #4847, opened 2021-08-24
feat(dynamicWidgets): add fallbackWidget

Haroenv:
**Summary**

adds a `fallbackWidget` key to dynamicWidgets to allow usage when you don't know in advance what the attributes will be

**Result**

- new key: "fallbackWidget" that gets used and instantiated when the attribute is slated for mounting
- no longer automatically adding widgets on init

To investigate: does this break routing?
To do: update tests if it turns out the mounting wasn't needed

### DIFF
--- a/src/connectors/dynamic-widgets/connectDynamicWidgets.ts
+++ b/src/connectors/dynamic-widgets/connectDynamicWidgets.ts
@@ -18,6 +18,7 @@ export type DynamicWidgetsRenderState = {
 
 export type DynamicWidgetsConnectorParams = {
   widgets: Widget[];
+  fallbackWidget?(args: { attribute: string }): Widget;
   transformItems?(
     items: string[],
     metadata: { results: SearchResults }
@@ -42,22 +43,18 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
     checkRendering(renderFn, withUsage());
 
     return (widgetParams) => {
-      const { widgets, transformItems = (items) => items } = widgetParams;
+      const {
+        widgets,
+        transformItems = (items) => items,
+        fallbackWidget,
+      } = widgetParams;
 
       if (
-        !widgets ||
-        !Array.isArray(widgets) ||
-        widgets.some((widget) => typeof widget !== 'object')
-      ) {
-        throw new Error(
-          withUsage('The `widgets` option expects an array of widgets.')
-        );
-      }
-
-      if (
-        !widgets ||
-        !Array.isArray(widgets) ||
-        widgets.some((widget) => typeof widget !== 'object')
+        !(
+          widgets &&
+          Array.isArray(widgets) &&
+          widgets.every((widget) => typeof widget === 'object')
+        )
       ) {
         throw new Error(
           withUsage('The `widgets` option expects an array of widgets.')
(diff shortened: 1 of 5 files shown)


stories/dynamic-widgets.stories.ts
```diff
@@ -14,20 +14,26 @@ storiesOf('Basics/DynamicWidgets', module).add(
     search.addWidgets([
       instantsearch.widgets.EXPERIMENTAL_dynamicWidgets({
         container: dynamicWidgetsContainer,
+        fallbackWidget: ({ attribute, container }) =>
+          instantsearch.widgets.panel<
+            typeof instantsearch.widgets.refinementList
+          >({
```
francoischalifour, line 20: We should be able to use the `RefinementListWidget` type.

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -72,9 +78,8 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
         init(initOptions) {
           widgets.forEach((widget) => {
             const attribute = getWidgetAttribute(widget, initOptions);
-            localWidgets.set(attribute, { widget, isMounted: true });
+            localWidgets.set(attribute, { widget, isMounted: false });
           });
-          initOptions.parent!.addWidgets(widgets);
```
francoischalifour, line 77: What is this replaced with?

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -18,6 +18,7 @@ export type DynamicWidgetsRenderState = {
 
 export type DynamicWidgetsConnectorParams = {
   widgets: Widget[];
+  fallbackWidget?(attribute: string): Widget;
```
francoischalifour, line 21: In React InstantSearch we pass it in an object:

```suggestion
  fallbackWidget?(params: { attribute: string }): Widget;
```

stories/dynamic-widgets.stories.ts
Haroenv: yes, it's the same

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
Haroenv: I was experimenting whether it was actually needed, either something was wrong with my local environment that kept the refinement, but they no longer seem to be writing an empty url initially without this line. I need to check more

Haroenv:
It's working, but I don't understand how, can anyone explain? https://codesandbox.io/s/instantsearch-js-forked-uf6ky?file=/src/app.js

eunjae-lee:
> It's working, but I don't understand how, can anyone explain? https://codesandbox.io/s/instantsearch-js-forked-uf6ky?file=/src/app.js

Which part do you not understand? Tbh, I don't understand all the flow perfectly bit by bit, but I can look into detail if you give me a pointer.

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
Haroenv: We found in #4849 that the write which happens on initial mount (before the widgets are added) was caused by the insights middleware, and not normal behaviour

eunjae-lee (approved):
This looks good to me 🎉

PR #4849, opened 2021-08-26
fix(middleware): subscribe middleware before initializing main index

eunjae-lee:
## Summary

This PR changes the implementation of `instantSearchInstance.start()` so that it subscribes middlewares **BEFORE** `mainIndex.init()` (It was the opposite.)

### What it solves

The router middleware used to apply the router state into `_initialUiState` at its creation, not in `subscribe()`. It was problematic when `instantSearchInstance` is re-used and have different router state that is not being taken into account. So this change allows the router middleware to modify initialUiState at `subscribe()`.

This PR also updates the implementation of insights middleware. It used to rely on the helper to set and get user token. Now at `subscribe()`, the main index is not initialized, which means there is no helper yet. So it now adds a dummy configure widget to apply `clickAnalytics` and `userToken`.

---

<details>
<summary>previous description</summary>
This PR fixes the router middleware to set correct initial ui state when instantSearchInstance is re-used.

### When does this happen?

In Vue InstantSearch, when `ais-instant-search` component gets unmounted, it calls `dispose()`.
https://github.com/algolia/vue-instantsearch/blob/2d6879be7a16bbf7fc6465891180304782d492e7/src/util/createInstantSearchComponent.js#L72:L74

Later when the component gets mounted again, it will call `instantSearchInstance.start()`.

### Why is it broken?

https://github.com/algolia/instantsearch.js/blob/0e151a9552092807ecbc6993f3f6193fef621f44/src/middlewares/createRouterMiddleware.ts#L60:L63

Inside `createRouterMiddleware`, we read the url from the router at creation of router middleware, whereas it was supposed to be at `subscribe()`. Thus, it reads only once, and when it's re-used it will stay outdated.

---
**More info:**

When the router middleware is created, it updates `instantSearchInstance._initialUiState`. At this point, `instantSearchInstance.start()` hasn't been called yet. When `start()` is called, it initialize the mainIndex with `mainIndex.init(...)` and even though corresponding widgets don't exist in the widget tree, it takes initialUiState into account and keep it for a while.

```js
   indexName: 'hello',
   searchClient,
   initialUiState: {
    hello: {
      query: 'world'
    }
   }
})

search.start()
```

Even though there is no `searchBox` widget to hold `query` inside the uiState, it will hold `query` for a while.

* If `searchBox` is added, then it will trigger a new search with the query (Before then, the search param doesn't include it.)
* Before `searchBox` is added, if `search.getUiState()` is called, it will [re-evaluate it](https://github.com/algolia/instantsearch.js/blob/0e151a9552092807ecbc6993f3f6193fef621f44/src/lib/InstantSearch.ts#L612) and `query` disappears.

This seems inconsistent, considering we normally eliminate states from uiState if there is no corresponding widget (no matter we like the behavior or not.)

---

#### Fix 1. Use `setUiState` instead of updating initialUiState.

Instead of manipulating the private instance variable `_initialUiState`, leveraging the public API `setUiState` seems solid. However, this unnecessarily triggers uiState change to `instantsearch({ onStateChange })` and to middlewares. It also could cause one more search API calls (empty one + another one with the new state.) If we call setUiState from the router middleware when only there is something to set, then it's less problematic.

Another problem with this approach is, if corresponding widgets are not in the widget tree, the result of `setUiState` will go away, whereas what's written in `initialUiState` remains regardless.

#### Fix 2. Call `middleware.subscribe()` before `mainIndex.init()`

Then it can get to update `_initialUiState` before it's passed to `mainIndex.init()`, but this doesn't seem right. It's better to have middlewares started after we initialize the main index and settle down with states. Also, this change will break the insights middleware and possibly custom middlewares.

#### Fix 3. Re-create middlewares when restarted 👈

When the instance gets reused and restarted, we'd probably want to re-create the middlewares. The definition of middleware in IS.js is "a function that returns an object." We could re-run the functions to replace the middleware objects. This is a non-breaking change, and possibly will work. I will try it out and will keep this PR updated.
</details>


src/lib/InstantSearch.ts
```diff
@@ -151,6 +151,7 @@ class InstantSearch<
   public mainHelper: AlgoliaSearchHelper | null;
   public mainIndex: IndexWidget;
   public started: boolean;
+  public disposed: boolean;
```
eunjae-lee, line 154: I'd like to have `public state: 'none' | 'started' | 'disposed'`, but I'm not sure how much `started` being used. At least, it's not documented https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/.

Haroenv: started is used now, by Vue InstantSearch for example, but could we not introduce `state` instead of `disposed` now? (we can still set both state and started and deprecate `started`)

eunjae-lee: That sounds more future-proof.

src/lib/InstantSearch.ts
```diff
@@ -480,6 +482,20 @@ See ${createDocumentationLink({
       });
     });
 
+    // if the instance is being re-started,
+    if (this.disposed) {
+      this.disposed = false;
+      this.middleware = this.middleware.map(({ creator }) => ({
```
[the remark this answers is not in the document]
eunjae-lee: it's similar to what we have in `use()`: https://github.com/algolia/instantsearch.js/blob/25f41b2b3dcc71eb5b1c5598d98808d5cf2b023f/src/lib/InstantSearch.ts#L278:L302

but `use()` can be called either before or after `start()`.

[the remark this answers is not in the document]
Haroenv: maybe .use could be called somehow?

test/mock/createInstantSearch.ts
```diff
@@ -17,6 +17,7 @@ export const createInstantSearch = (
     mainHelper,
     client,
     started: false,
+    disposed: false,
```
Haroenv, line 20:
```suggestion
    state: 'none',
```

src/lib/InstantSearch.ts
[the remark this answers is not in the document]
eunjae-lee: yes,
```js
      this.middleware = this.middleware.map(({ creator }) => ({
        creator,
        instance: {
          subscribe: noop,
          unsubscribe: noop,
          onStateChange: noop,
          ...creator({ instantSearchInstance: this }),
        },
      }));
```

can become

```js
const middlewaresToUse = this.middleware.map(({ creator }) => creator);
this.middleware = [];
this.use(...middlewares)
```

I didn't use `use()` function in this implementation because `use()` includes irrelevant code like calling `subscribe()` if the instantSearchInstance is already started. The second version looks short, but I feel like it creates unnecessary link to the function.

What do you think? The current version is too verbose?

src/lib/InstantSearch.ts
eunjae-lee: edit: 'none' → 'initial'

src/lib/InstantSearch.ts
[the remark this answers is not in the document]
Haroenv: I actually don't really understand why the middleware needs to be recreated, should the state not be set at `subscribe` (instead at function body now for routing middleware) so that it doesn't become stateful on creation, but on subscribe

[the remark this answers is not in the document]
eunjae-lee: Ah okay, about that part, initially I tried to make `subscribe()` be the place where it sets the state from the URL (Fix 1 or Fix 2 in the PR description). And it didn't work, and this was the final try. But reading your comment, I've become to not like making middlewares stateful on creation. Fix 1 and 2 are breaking the current behaviors. Do you have any other idea? If not, I should probably close this for now, and revisit it later for next major(, although as we've recently said, it's hard to prioritize these small/invisible tasks when working on major.)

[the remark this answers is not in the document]
Haroenv: why did moving the code to subscribe not work? I feel like that's the correct solution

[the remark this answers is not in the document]
eunjae-lee: > Subscribe is called on start & restart right?

yes. And moving `_initialUiState = ....` to `subscribe()` didn't work because 

Inside `search.start()`,

```js
mainIndex.init({
  // ...
  uiState: this._initialUiState,
});

(each middleware).subscribe();
```

It's already too late.

src/lib/InstantSearch.ts
```diff
@@ -480,6 +483,20 @@ See ${createDocumentationLink({
...
+          unsubscribe: noop,
+          onStateChange: noop,
+          ...creator({ instantSearchInstance: this }),
+        },
+      }));
+    }
+
     this.mainHelper = mainHelper;
 
     this.mainIndex.init({
```
Haroenv, line 502: what if we move this to after the middleware subscribe happens? technically breaking behaviour, but I don't think it's something we defined as api (which order middleware happen in). Then we can move the initialUiState patch to before the init

eunjae-lee: That's suggested in fix 2, and let me give it another try. It got me many failed test cases here and there, but I didn't looked at them closely. It could be trivial, or not.

Haroenv: I guess the subscribe was expected to be called with a set ui state, which causes the tests to fail, although I'm not sure if that's really needed

src/middlewares/createRouterMiddleware.ts
```diff
@@ -78,6 +76,11 @@ export const createRouterMiddleware = <
       },
 
       subscribe() {
+        instantSearchInstance._initialUiState = {
+          ...initialUiState,
```
Haroenv, line 79: this can continue to be read from the InstantSearch instance right?

eunjae-lee: But if subscribe() gets called multiple times, it could accidentally accumulate states.

For example, 

1. /search?hitsPerPage=3&query=abc

2. /search?query=def

but at the second visit, it will still get `hitsPerPage=3`, which is wrong.

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -88,25 +90,29 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = (props) => {
...
+    let widget: ReturnType<typeof createWidget>;
+
     return {
       onStateChange() {},
       subscribe() {
         insightsClient('addAlgoliaAgent', 'insights-middleware');
 
-        // At the time this middleware is subscribed, `mainIndex.init()` is already called.
-        // It means `mainIndex.getHelper()` exists.
-        const helper = instantSearchInstance.mainIndex.getHelper()!;
```
eunjae-lee, line 98: The insights middleware used to rely on the helper to set/get user token. However, now that we get this `subscribe()` called before `mainIndex.init()` we no longer have access to it at subscribe(). We could setTimeout(fn, 0) but it will make the first query being fired without user token.

Anyway I don't like touching helper directly here, so changed the implementation to create a custom configure widget within the middleware to manipulate the search parameters.

src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -1540,8 +1540,8 @@ describe('use', () => {
       .invocationCallOrder[0];
     const middlewareSubscribeCallOrder =
       middlewareSpy.subscribe.mock.invocationCallOrder[0];
-    // Checks that `mainIndex.init` was called before subscribing the middleware.
-    expect(widgetsInitCallOrder).toBeLessThan(middlewareSubscribeCallOrder);
+    // Checks that `mainIndex.init` was called after subscribing the middleware.
+    expect(widgetsInitCallOrder).toBeGreaterThan(middlewareSubscribeCallOrder);
```
eunjae-lee, line 1544: This is technically a breaking change, but I don't think it will break anything in real life.

eunjae-lee:
The tests are broken now. Don't review it yet. I'll fix them.

src/middlewares/createRouterMiddleware.ts
```diff
@@ -57,10 +58,6 @@ export const createRouterMiddleware = <
     // casting to UiState here to keep createURL unaware of custom UiState
     // (as long as it's an object, it's ok)
     instantSearchInstance._createURL = topLevelCreateURL as CreateURL<UiState>;
-    instantSearchInstance._initialUiState = {
-      ...instantSearchInstance._initialUiState,
-      ...stateMapping.routeToState(router.read()),
-    };
```
Haroenv, line 61: in that case the initial ui state can be captured here instead of making it an argument @eunjae-lee

eunjae-lee: yeah good point!

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -6,6 +7,7 @@ import type {
 } from '../types';
 import { getInsightsAnonymousUserTokenInternal } from '../helpers';
 import { warning, noop, getAppIdAndApiKey, find } from '../lib/utils';
+import connectConfigure from '../connectors/configure/connectConfigure';
```
Haroenv, line 10: what about the configure widget instead of connector? It doesn't have any warnings or validation that would make the impact bigger

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -88,25 +90,29 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = (props) => {
...
       subscribe() {
         insightsClient('addAlgoliaAgent', 'insights-middleware');
 
-        // At the time this middleware is subscribed, `mainIndex.init()` is already called.
-        // It means `mainIndex.getHelper()` exists.
-        const helper = instantSearchInstance.mainIndex.getHelper()!;
+        let searchParameters: Partial<PlainSearchParameters> = {
+          clickAnalytics: true,
+        };
+        widget = createWidget({ searchParameters });
```
Haroenv, line 104: I'd create the arguments dynamically, but not the widget twice, that way you can add the widget with the known arguments *after* the if (or is that not possible?)

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -132,7 +138,13 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = (props) => {
           if (onEvent) {
             onEvent(event, _insightsClient);
           } else if (event.insightsMethod) {
-            if (hasUserToken()) {
+            // At this point, instantSearchInstance must be started and
+            // it means there is a configure widget (added above).
+            const hasUserToken = Boolean(
+              instantSearchInstance.renderState[instantSearchInstance.indexName]
+                .configure!.widgetParams.searchParameters.userToken
```
Haroenv, line 145: if the setUserTokenToSearch never was called, isn't configure undefined?

src/middlewares/createInsightsMiddleware.ts
eunjae-lee:
```diff
        let searchParameters: Partial<PlainSearchParameters> = {
          clickAnalytics: true,
        };
-       widget = createWidget({ searchParameters });
-       instantSearchInstance.addWidgets([widget]);

        const setUserTokenToSearch = (userToken?: string) => {
          instantSearchInstance.removeWidgets([widget]);
          searchParameters = {
            ...searchParameters,
            userToken,
          };
          widget = createWidget({ searchParameters });
          instantSearchInstance.addWidgets([widget]);
        };

        const anonymousUserToken = getInsightsAnonymousUserTokenInternal();
        if (hasInsightsClient && anonymousUserToken) {
          // When `aa('init', { ... })` is called, it creates an anonymous user token in cookie.
          // We can set it as userToken.
          setUserTokenToSearch(anonymousUserToken);
        } else {
+         widget = createWidget({ searchParameters });
+         instantSearchInstance.addWidgets([widget]);
        }
```

something like this?

src/middlewares/createInsightsMiddleware.ts
eunjae-lee: We add it in the beginning for `clickAnalytics`
https://github.com/algolia/instantsearch.js/blob/6fba77c7167ee039ff6d3685616e7dfe64c81e8e/src/middlewares/createInsightsMiddleware.ts#L104:L106

src/middlewares/createInsightsMiddleware.ts
eunjae-lee: Not sure which warnings or validation you mean. My initial thought on using the connector was to avoid any widget within middlewares. The connctors, middlewares, utils, etc can be shared across flavors, but we won't share widgets (although configure widget is renderless.)

src/middlewares/createInsightsMiddleware.ts
Haroenv: I guess I'm not really sure what the flow should be. In that diff, aren't you adding the widget twice?

Haroenv: Yep, the widget is added multiple times now. Instead maybe create two widgets, one for userToken and one for clickAnalytics?

eunjae-lee: In the diff or not, it's not adding the widget twice, unless I misunderstand something. Inside `setUserTokenToSearch` function, it removes the existing widget and then accumulate searchParameters with new userToken and add the new widget again.

Haroenv: 🤔 I missed the removing. It still seems much more complicated than needed, could it not just be two widgets, as configure already merges?

eunjae-lee: Okay, now we're on the similar page. In which part do you feel complicated? That conditional part? Or adding/removing widget?

If the former, indeed I can put two widgets, one for clickAnalytics and another one for userToken. I just want to make sure, we both are aware that although we do that, we still need to remove/add the widget for userToken when we want to update it.

Haroenv: adding and removing the widget seems much more complicated than the previous code which used the helper. I feel like the widget's refine could make sense to use (make sure to check it doesn't erase parameters of another configure widget)

Haroenv: I almost feel as if there should be a lifecycle event *before* the index init happened (.use essentially) for routing, but one after init has happened to use the helper

eunjae-lee: Yeah, you're right. I did try using `refine` function from the widget, but that's given after `render()` function of the widget is called, which is, again, too late because instantSearchInstance is already started and a API call is fired.

I thought of having additional lifecycle method like `beforeInit` so that the router middleware can manipulate initialUIState. But I had a hard time thinking of another good use case to justify the addition of the API.

What do you think? Should we hold this and have a short meeting when François comes back?

src/middlewares/createInsightsMiddleware.ts
```diff
@@ -91,27 +91,30 @@ export const createInsightsMiddleware: CreateInsightsMiddleware = (props) => {
...
-          searchParameters = {
-            ...searchParameters,
-            userToken,
-          };
-          widget = createWidget({ searchParameters });
-          instantSearchInstance.addWidgets([widget]);
+          if (configureUserToken) {
+            instantSearchInstance.renderState[
+              instantSearchInstance.indexName
+            ].configure!.refine({ userToken });
```
Haroenv, line 110: did you double check whether this left other parameters in place? I think it does, but not 100% sure

eunjae-lee: yup, it's done by connectConfigure
https://github.com/algolia/instantsearch.js/blob/321b9521b9bbd3d6fb0dc861ac87e15ec966ba3c/src/connectors/configure/connectConfigure.ts#L101:L104

Haroenv (commented):
overall I think this is a decent solution and doesn't require too many changes, although the insights is a little dubious if it's the best solution to use configure?

src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -1,28 +1,33 @@
...
-    };
-    // @ts-expect-error
-    instantSearchInstance.mainIndex = {
-      getHelper: () => helper,
-    };
+    instantSearchInstance.start();
+
+    const helper = instantSearchInstance.mainIndex.getHelper();
+
+    const getUserToken = () => (helper.state as any).userToken;
```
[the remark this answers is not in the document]
eunjae-lee: It exists only in search client v4:
https://github.com/algolia/algoliasearch-helper-js/blob/2053c7987206af62399f07225b3cbee84e9d5141/index.d.ts#L31:L36
https://github.com/algolia/instantsearch.js/pull/4849/commits/4066caf5d950d2f47675451af62ed90d76e78275

eunjae-lee:
Rerunning e2e twice made it pass... Anyway reading the test case, it seems fine.

src/lib/__tests__/InstantSearch-test.tsx
francoischalifour: We've got a customer report saying that their app broke. I think that's because of this commit, because `instantSearchInstance.renderState[instantSearchInstance.indexName]` is undefined in `createInsightsMiddleware`.

Any middleware that relies on `instantSearchInstance` might break now, because there's no certainty that the instance has started anymore.

Could we rather have a `setInitialUiState` method which updates `initialUiState` without these PR changes?

eunjae-lee: We actually fixed that regression at https://github.com/algolia/instantsearch.js/pull/4897

I tried to fix the issue without introducing new API, but I let you and the team decide whether to revert #4849 and #4897 and introduce `setInitialUiState`.

PR #4901, opened 2021-09-08
fix(insightsMiddleware): throw an error when credentials can't be extracted

Haroenv:
**Summary**

When api or appId is missing, an error gets thrown by search insights, but that doesn't explain *why* that error gets thrown.

Ideally this would also clarify what the next step is (usually making sure to spread the previous search client, or leaving all properties in place, but i'm not too sure how to word that.

This doesn't add a new error where there wasn't already one, as the search-insights library threw an error in this case already.

**Result**

An error that explains when the search client doesn't match insights' expectation

### DIFF
--- a/src/lib/utils/getAppIdAndApiKey.ts
+++ b/src/lib/utils/getAppIdAndApiKey.ts
@@ -1,7 +1,5 @@
 // typed as any, since it accepts the _real_ js clients, not the interface we otherwise expect
-export function getAppIdAndApiKey(
-  searchClient: any
-): [string | undefined, string | undefined] {
+export function getAppIdAndApiKey(searchClient: any) {
   if (searchClient.transporter) {
     // searchClient v4
     const { headers, queryParameters } = searchClient.transporter;
--- a/src/lib/utils/getAppIdAndApiKey.ts
+++ b/src/lib/utils/getAppIdAndApiKey.ts
@@ -1,5 +1,7 @@
 // typed as any, since it accepts the _real_ js clients, not the interface we otherwise expect
-export function getAppIdAndApiKey(searchClient: any) {
+export function getAppIdAndApiKey(
+  searchClient: any
+): [string | undefined, string | undefined] {
   if (searchClient.transporter) {
     // searchClient v4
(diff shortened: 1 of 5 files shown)


src/lib/utils/getAppIdAndApiKey.ts
```diff
@@ -1,5 +1,7 @@
 // typed as any, since it accepts the _real_ js clients, not the interface we otherwise expect
-export function getAppIdAndApiKey(searchClient: any) {
+export function getAppIdAndApiKey(
+  searchClient: any
```
Haroenv, line 3: I looked into typing this, but as it accepts either v3 or v4, it would need some ts-ignore's like algoliasearch.ts and that doesn't seem worth it

src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -114,6 +117,19 @@ describe('insights', () => {
...
+    it('throws when search client does not have credentials', () => {
+      const { insightsClient, instantSearchInstance } = createTestEnvironment({
+        searchClient: createSearchClient(),
+      });
+      expect(() =>
+        createInsightsMiddleware({
+          insightsClient,
+        })({ instantSearchInstance })
+      ).toThrowErrorMatchingInlineSnapshot(
+        `"[insights middleware]: could not extract Algolia credentials from searchClient"`
```
eunjae-lee, line 129: Seems reasonable, but I'm curious in which case searchClient didn't include the credentials? In that case, InstantSearch.js itself won't be functioning, no?

Haroenv: if you do the following:

```
const client = algoliasearch()
const searchClient = {
  search: client.search // or intercept
}
```

In that case only the "required" methods are present, as well as with a "backend search" without an algolia client the insights-thrown error is confusing

eunjae-lee: Ah okay, that makes sense.

eunjae-lee (approved):

PR #4907, opened 2021-09-14
fix(infinite/hits): stop saving the transformed results in cache

### LINKED ISSUE #4819: Hits widget transformItems is called twice, receives already-transformed items
## 🐛 Bug description

It appears that the [transformItems](https://www.algolia.com/doc/api-reference/widgets/hits/js/#widget-param-transformitems) function of the hits widget is being called twice for a single search invocation, and is receiving the already-transformed items on the second run.

## 🔍 Bug reproduction

1. Add a `transformItems` callback to the hits widget that returns transformed items.
2. Add a `console.log` notification somewhere inside the transformation function.
3. Reload the page and look for notifications in the console.

**Live reproduction:**

https://codesandbox.io/s/fervent-frost-bbzb5?file=/src/app.js
https://codesandbox.io/s/stupefied-frog-c2krg?file=/src/instantsearch.ts

Including two examples here, one based on `create-instantsearch-app` and another based on the instantsearch sandbox from the main documentation. The latter already included a `transformItems` call, so required fewer additions on my part.

## 💭 Expected behavior

`transformItems` is called only once for a single search invocation, and is never passed already-transformed items.

## Environment

- OS: Windows
- Browser: Firefox, Edge, Chrome
- Version: 4.25.2

## Additional context

I've run into this bug while upgrading from an older 4.3.1 install to the latest 4.25.2 release. However, experimenting via codesandbox.io reveals that the bug was likely introduced in the 4.9 branch, as 4.9.2 is the earliest release I tested that allowed me to replicate it.

It particularly affects us because our transformed items cannot be safely transformed a second time, leading to an unexpected runtime error rather than it silently continuing.

Possibly related to [this issue](https://github.com/algolia/vue-instantsearch/issues/707) in the vue-instantsearch repo?

Haroenv:
**Summary**

Since we introduced getWidgetRenderState, transformItems gets called two times. On its own this isn't problematic, however if the transformItems "mutates" a key, it will receive next the "mutated" hits instead of the ones that come from the results

One thing which is debatable about this PR is which parts should be saved to `results.hits`. I opted for just the escaped hits, but i wonder if there's any use cases for considering the query id and position as part of the hits directly.

However, I feel like if we want this behaviour, it shouldn't be hits and infiniteHits mutating results, but rather something global that gets called right after the results get returned. In fact, I also think it makes more sense to escape hits there than in hits.

I don't think removing the escapeHTML option is possible here without breaking change though, so if we want to do that, we should mark it as a possibility for a major in this PR

**Result**

- transformItems only gets `hits` that come directly computed from a result, not from a previous transformItems
- the result of transformItems does not get stored on result.hits (could be considered a breaking change, but I don't think so)

FX-11
fixes #4819

### DIFF
--- a/src/connectors/answers/__tests__/connectAnswers-test.ts
+++ b/src/connectors/answers/__tests__/connectAnswers-test.ts
@@ -240,7 +240,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/answers/js/
 
     // render with hits
     const expectedHits = [{ title: '', objectID: 'a', __position: 1 }];
-    (expectedHits as any).__escaped = true;
     expect(renderFn).toHaveBeenCalledTimes(3);
     expect(renderFn).toHaveBeenNthCalledWith(
       3,
@@ -303,7 +302,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/answers/js/
     // wait for debounce
     await wait(30);
     const expectedHits = [{ title: '', objectID: 'a', __position: 1 }];
-    (expectedHits as any).__escaped = true;
     expect(renderFn).toHaveBeenCalledTimes(4);
     expect(renderFn).toHaveBeenNthCalledWith(
       4,
@@ -425,7 +423,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/answers/js/
 
       await wait(30);
       const expectedHits = [{ title: '', objectID: 'a', __position: 1 }];
-      (expectedHits as any).__escaped = true;
       expect(
         widget.getWidgetRenderState(
           createRenderOptions({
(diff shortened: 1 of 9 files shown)


francoischalifour (commented):
We also [mutate the hits in `connectAnswers`](https://github.com/algolia/instantsearch.js/blob/506dceb9853e4677c52c9cb0b26d6f1f662b7e13/src/connectors/answers/connectAnswers.ts#L215-L217). We should opt for the same strategy everywhere.

Haroenv:
Ah yes, I forgot about answers, it's a little different but can also be changed. Note that in answers it's not the global result that's mutated though @francoischalifour

francoischalifour (approved):

PR #4908, opened 2021-09-15
fix(es): add warning to typescript declaration of keys to be imported from helpers

Haroenv:
**Summary**

We deprecated the properties added to instantsearch in #4814, but the deprecation does not mention the solution

**Result**

The same solution as at runtime is given by typescript, importing from ./helpers

### DIFF
--- a/src/index.es.ts
+++ b/src/index.es.ts
@@ -20,19 +20,19 @@ type InstantSearchModule = {
   version: string;
 
   // @major remove these in favour of the exports
-  /** @deprecated */
+  /** @deprecated import { createInfiniteHitsSessionStorageCache } from 'instantsearch.js/es/helpers */
   createInfiniteHitsSessionStorageCache: typeof createInfiniteHitsSessionStorageCache;
-  /** @deprecated */
+  /** @deprecated import { highlight } from 'instantsearch.js/es/helpers */
   highlight: typeof highlight;
-  /** @deprecated */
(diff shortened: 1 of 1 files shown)


src/index.es.ts
```diff
@@ -20,19 +20,19 @@ type InstantSearchModule = {
...
+  /** @deprecated import { reverseHighlight } from 'instantsearch.js/es/helpers */
   reverseHighlight: typeof reverseHighlight;
-  /** @deprecated */
+  /** @deprecated import { snippet } from 'instantsearch.js/es/helpers */
   snippet: typeof snippet;
-  /** @deprecated */
+  /** @deprecated import { reverseSnippet } from 'instantsearch.js/es/helpers */
   reverseSnippet: typeof reverseSnippet;
-  /** @deprecated */
+  /** @deprecated use createInsightsMiddleware */
```
francoischalifour, line 33: We can add `@link https://www.algolia.com/doc/api-reference/widgets/insights/js/`

src/index.es.ts
```diff
@@ -20,19 +20,19 @@ type InstantSearchModule = {
   version: string;
 
   // @major remove these in favour of the exports
-  /** @deprecated */
+  /** @deprecated import { createInfiniteHitsSessionStorageCache } from 'instantsearch.js/es/helpers */
```
francoischalifour, line 23: There are missing `'` after all the imports.

sarahdayan (approved):

PR #4913, opened 2021-09-30
refactor(widgetParams): enforce contract in a convenient way

Haroenv:
**Summary**

the code of getWidgetAttribute was error-prone as it hides errors. In an effort to simplify the code, I changed the type of `widgetParams`

**Result**

 - unrelated errors aren't caught in dynamic widgets
 - slightly change the return of getWidgetRenderState


francoischalifour (commented):
That looks better to me, although we need to fix other types as the CI shows.

PR #4917, opened 2021-10-06
feat(dynamicWidgets): request wildcard facets by default

Haroenv:
**Summary**

make sure only one network request happens on mount of dynamic widgets, and no extra request when the requested widgets change

**Result**

use https://github.com/algolia/algoliasearch-helper-js/pull/874 if the wildcard is requested (with a new option)


seafoox:
Can you please provide an example for each of the case
- Default behaviour with * used
- Sequential behaviour where we wait the response from the first query in order to know what facet to retrieve value for.

I think understand that people would need to use wildcardFacets=true|false, but would like to confirm my understanding.

Thanks,

PR #4920, opened 2021-10-11
docs(dynamicWidgets): implement tsdoc for parameters

Haroenv:
**Summary**

Adds tsdoc documentation for the widget parameters of dynamic widgets

**Result**

No undocumented parameters


Haroenv:
@sarahdayan we can add link tags, but in a separate PR, as it would be inconsistent with the other documentation

PR #4930, opened 2021-10-15
fix(types): export correct types from search-insights

eunjae-lee:
**Summary**

This PR exports the correct types from `search-insights` package instead of having own types within the repo.

This fixes https://github.com/algolia/search-insights.js/issues/330

**Result**

It works the same, but likely no (or less) type errors.


Haroenv:
are the files which are in the *ignore* field of the tsconfig still needed to be there now that it's listed as a dependency @eunjae-lee ?

PR #4933, opened 2021-10-19
fix(router): skip history push on browser back and forward actions

### LINKED ISSUE #4740: Incorrect behavior on back action of browser history
## 🐛 Bug description

Incorrect behavior of the browser history on back action: User should click twice to get to previous page in browser history.

## 🔍 Bug reproduction

1. Go to `https://instantsearchjs.netlify.app/examples/e-commerce/`
2. Scroll down to pagination and click on second page.
3. After that, click on third page
4. Now try to click on browser's back button - it navigates to previous state (second page) correctly.
5. Try to click on browser's back button again - now it doesn't works correctly. In the result you will be still on the second page instead of first one.

## 💭 Expected behavior

Using browser's back button I expects previous state on each click.

## Environment

- OS: Linux / macOS
- Browser: any
- Version: instantsearch.js 4.20

## Additional context

Look at the flow in `window.addEventListener('popstate', this._onPopState);`. The issue causes because in `AlgoliaSearchHelper.prototype.setState` new instance of SearchParameters is created, so it will be always true for `if (state !== this.state) {` in `AlgoliaSearchHelper.prototype._change`. After that pushState is called.

pliusnin:
**Summary**

Issue was reported here https://github.com/algolia/instantsearch.js/issues/4740
The main issue is that after popstate event is fired, router.write method is called where pushState confuse the browser history. So next time you try to go back wrong state will be used. Fix will prevent pushState on popstate event.

**Result**

Closes #4740

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -66,13 +66,13 @@ module.exports = {
       files: ['*.ts', '*.tsx'],
       // this is the same files as ignored in tsconfig.json
       excludedFiles: [
-        "examples/**/*",
-        "es",
+        'examples/**/*',
+        'es',
         // these two files are temporarily excluded because
         // they import files from node_modules/search-insights directly
         // and it causes the type-checking to fail.
-        "src/middlewares/__tests__/createInsightsMiddleware.ts",
-        "test/mock/createInsightsClient.ts"
+        'src/middlewares/__tests__/createInsightsMiddleware.ts',
+        'test/mock/createInsightsClient.ts',
       ],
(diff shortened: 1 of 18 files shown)


Haroenv:
I think the idea of this fix is correct, thanks!  however it might be more correct to avoid the write in routing manager if it came from the callback of onUpdate, what do you think?

pliusnin:
I've double checked and I think we have to keep the check on history router level as it's the issue related on this, not on Router Manager.

Issue #4916, opened 2021-10-04
Type conflict between algoliasearch client and its usage in instantSearch


bidoubiwa:
## 🐛 Bug description

related to: #4911 

The InstantSearch client returns upon search with the geo widget the `insideBoundingBox` field containing a string. 

See the following test: 
 
https://github.com/algolia/instantsearch.js/blob/6e6293046de82f90336bc857efa9d0f5fcb8e62b/src/widgets/geo-search/__tests__/geo-search-test.ts#L1717-L1721

Nonetheless the declared type dictates that it should return an array of array of numbers: 

see [SearchOptions](https://github.com/algolia/algoliasearch-client-javascript/blob/00ad9246a9f7822b97c96b14d5627862c15050f3/packages/client-search/src/types/SearchOptions.ts#L295) types. 

```ts
readonly insideBoundingBox?: ReadonlyArray<readonly number[]>;
```

## 🔍 Bug reproduction

**Steps to reproduce the behavior:**

1. Create a custom searchClient for instantSearch
2. Try to adapt `insideBoundingBox`
3. TypeError

```
Argument of type 'readonly (readonly number[])[] | undefined' is not assignable to parameter of type 'string'.
```

## 💭 Expected behavior

They should be of same type

## Environment
-  "algoliasearch": "^4.10.5",
-   "instantsearch.js": "^4.30.1",

Haroenv:
yes, it's really the client type that is a little more restrictive than needed. Does having a ts-ignore have an impact on any site though?

The correct type is ```ReadonlyArray<readonly number[]> | `${number},${number},${number},${number}`;```, and then the ts-ignore could be ignored, when the algoliasearch requirement is set to the version in which that fix is done

bidoubiwa:
```
Does having a ts-ignore have an impact on any site though?
```

It creates a bit of magic code/technical debt as you'll have to remember that somewhere in your code lies a `ts-ignore` that may create bugs if instantSearch decides to change the type of `insindeBoundingBox`.

It is the reason I think why you use TypeScript, to avoid manipulating variables in another type.  

> The correct type is ReadonlyArray<readonly number[]> | `${number},${number},${number},${number}`;, and then the ts-ignore could be ignored, when the algoliasearch requirement is set to the version in which that fix is done

Thats cool 😊 Is the fix planned for a near release?

Haroenv:
[...]

PR #4938, opened 2021-10-20
chore(version): bump algoliasearch version to v0.11.0

bidoubiwa:
**Summary**

fixes: #4916 

This PR updates [`algoliasearch`](https://github.com/algolia/algoliasearch-client-javascript) and `@algolia/client-search` to the latest release: [`v0.11.0`](https://github.com/algolia/algoliasearch-client-javascript/releases/tag/4.11.0).

This new version introduces a fix on the type of `insideBoundingBox` (see https://github.com/algolia/algoliasearch-client-javascript/pull/1310). `insideBoundingBox` used to only accept a `ReadonlyArray<readonly number[]>;`, wich was conflicting with `instantsearch.js` using it as a `string` #4916 .

The fix introduced in the new @algolia/client-search packages expanded the type of `insideBoundingBox` to also accept a string ([this line](https://github.com/algolia/algoliasearch-client-javascript/blob/dfa7254b1d56b7c339dde39fbb371db369d865fe/packages/client-search/src/types/SearchOptions.ts#L295)). 

```ts
  readonly insideBoundingBox?: ReadonlyArray<readonly number[]> | string;
```

In this PR, the packages are updated and the code is made up-to-date with the introduced changes. 

- Typescript ignore comments to avoid typescript error are removed
- Typescript workaround to enforce `string` type is removed

**Result**

Screenshot 2021-10-19 at 23 11 43

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,7 +95,7 @@
     "@wdio/selenium-standalone-service": "5.16.5",
     "@wdio/spec-reporter": "5.16.5",
     "@wdio/static-server-service": "5.16.5",
-    "algoliasearch": "4.10.3",
+    "algoliasearch": "4.11.0",
     "algoliasearch-v3": "npm:algoliasearch@3.35.1",
     "babel-eslint": "10.0.3",
     "babel-jest": "27.1.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,56 +2,43 @@
 # yarn lockfile v1
 
 
-"@algolia/cache-browser-local-storage@4.10.3":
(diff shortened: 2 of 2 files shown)


Haroenv:
I made a "local" version of this PR and saw that CI is now failing for algoliasearch v3. I think we could either ts-ignore, keep the casting or keep everything as-is (except updating of course). What do you think @bidoubiwa? 

The algoliasearch (v3) types no longer exist on DefinitelyTyped, so I don't think we can add the string type there 🤔

PR #4950, opened 2021-11-01
Add missing peer dependency from algoliasearch-helper

yacinehmito:
The dependency algoliasearch-helper [declares algoliasearch as a peer dependency](https://github.com/algolia/algoliasearch-helper-js/blob/develop/package.json#L103) [1], but instantsearch.js does not declare it as a dependency nor a peer dependency.
This can cause undefined issues during peer dependency resolution, and triggers warnings when using yarn v2+.

The commit adds algoliasearch as a peer dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -144,6 +144,9 @@
     "typescript": "4.3.5",
     "webpack": "4.41.5"
   },
+  "peerDependencies": {
+    "algoliasearch": ">= 3.1 < 5"
+  },
   "resolutions": {
     "places.js/algoliasearch": "3.35.0"
   },


Haroenv:
With typescript in mind it should probably indeed be a peer dependency, even though it's technically optional still

tkrugg:
I guess we could add it now. 
Most users already have `algoliasearch` installed and won't even see the warning. The range `>= 3.1 < 5` is large enough that we remain compatible with both v3 & v4.
@Haroenv can you please outline the downsides of merging this change?

Haroenv:
With npm v7 it breaks their build if they don't have it installed @tkrugg, so adding is a breaking change

Haroenv (approved):
I misunderstood in which cases this errors. As we have the peer dependency in the helper, this doesn't consist of a breaking change, but just a fix

yacinehmito:
Thank you 😄

PR #4958, opened 2021-11-17
chore(examples/media): add mentions in footer

dhayab:
**Summary**

This PR adds a footer in the new media example, informing that the dataset comes from Algolia blog, and providing them a direct link to this resource.

CleanShot 2021-11-17 at 10 43 29@2x

**[Preview](https://deploy-preview-4958--instantsearchjs.netlify.app/examples/media/)**

### DIFF
--- a/examples/media/src/app.css
+++ b/examples/media/src/app.css
@@ -596,7 +596,7 @@ body.filtering {
 
 .mentions {
   padding: 0.5rem 1rem;
-  font-size: 0.75rem;
+  font-size: 0.875rem;
   line-height: 1.25rem;
   color: #dadada;
   background-color: rgba(0, 0, 0, 0.75);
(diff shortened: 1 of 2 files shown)


Haroenv (approved):

examples/media/src/app.css
```diff
@@ -591,6 +575,49 @@ body.filtering {
...
+  width: 2.5rem;
+}
+
+.filtering .back-to-top {
+  display: none;
+}
+
+.mentions {
+  padding: 0.5rem 1rem;
+  font-size: 0.75rem;
```
seafoox, line 599: Could we make the text slightly bigger? I was thinking maybe the same size as the summary content in each hit.

seafoox (approved):
Thanks for adding it. Looks good to me.

PR #4960, opened 2021-11-24
feat: rely on `state` in `getWidgetRenderState`

francoischalifour:
## Description

This changes how **we provide the widgets' render state from the helper state to the retrieved state** (`renderOptions.helper.state` vs. `renderOptions.state`).

This is required to support React InstantSearch Hooks SSR[^1] to avoid flashes between SSR and CSR[^2] when hydrating.

## Explanation

As opposed to Vue InstantSearch SSR, React InstantSearch Hooks SSR works in a single data flow pass. We rely on a single helper (the one created by InstantSearch.js), to provide the search state and the search results to the widgets.

By design, InstantSearch.js computes the search parameters from the mounted widgets (e.g., `searchBox` controls the `query` search parameter). As soon as you start setting a search parameter that is not yet controlled by a widget (this happens when a child index inherits from its parent index `searchBox` state), InstantSearch isn't able to take control back on that search parameter. This is for this reason that we cannot override the helper state (aka search parameters) before widgets are mounted, and _then_ rely on the widgets to update the helper state again.

Because of this constraint, once we override the initial helper state on SSR, any interactions on the browser won't update the search parameters, leaving the UI stuck.

This PR modifies the way we return the widgets' render state from `getWidgetRenderState` to rely directly on `state`, and not `helper.state`. This lets us provide a computed initial state from `results._state` without controlling the search parameters yet, and display the correct UI state on screen without messing with InstantSearch internals. And therefore lets InstantSearch correctly controls the search parameters after user interactions.

## Impact

This change is only internal and shouldn't impact any user implementations.

[^1]: SSR: Server-Side Rendering
[^2]: CSR: Client-Side Rendering

### DIFF
--- a/src/connectors/infinite-hits/connectInfiniteHits.ts
+++ b/src/connectors/infinite-hits/connectInfiniteHits.ts
@@ -303,8 +303,8 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
             widgetType: this.$$type,
           });
           isFirstPage =
-            helper.state.page === undefined ||
-            getFirstReceivedPage(helper.state, cachedHits) === 0;
+            state.page === undefined ||
+            getFirstReceivedPage(state, cachedHits) === 0;
         } else {
           const { page = 0 } = state;
 
--- a/src/connectors/search-box/connectSearchBox.ts
+++ b/src/connectors/search-box/connectSearchBox.ts
@@ -138,10 +138,10 @@ const connectSearchBox: SearchBoxConnector = function connectSearchBox(
         };
       },
 
-      getWidgetRenderState({ helper, searchMetadata }) {
+      getWidgetRenderState({ helper, searchMetadata, state }) {
         if (!_refine) {
           const setQueryAndSearch = (query: string) => {
-            if (query !== helper.state.query) {
+            if (query !== state.query) {
               helper.setQuery(query).search();
             }
           };
@@ -159,7 +159,7 @@ const connectSearchBox: SearchBoxConnector = function connectSearchBox(
         _clear = clear(helper);
 
         return {
-          query: helper.state.query || '',
+          query: state.query || '',
           refine: _refine,
           clear: _cachedClear,
           widgetParams,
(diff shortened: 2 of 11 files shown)


src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -303,8 +303,8 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
             widgetType: this.$$type,
           });
           isFirstPage =
-            helper.state.page === undefined ||
```
Haroenv, line 306: I wonder if in the index widget, could we make the getter of helper.state warn? Or is that still used internally?

francoischalifour: There are still valid use cases for using `helper.state` in callbacks like `createURL()` or `refine()` to manipulate the latest helper state:

https://github.com/algolia/instantsearch.js/blob/7747c50c037b1a73cf74e9e82986a0c303822e5a/src/connectors/breadcrumb/connectBreadcrumb.ts#L204-L206

An ESLint rule should be able to catch that maybe? Not really trivial though. Ultimately we want the values from the main paths that are returned from `getWidgetRenderState` to be computed with `state`, not with `helper.state`, the rest is fine.

Haroenv (approved):
this change lgtm, as even if we fix this issue in React InstantSearch server side rendering later, state should be consistent with helper.state.

In a future version I'd like to see state being the same as helper.state though, and maybe even only one of the two exposed

src/connectors/search-box/connectSearchBox.ts
```diff
@@ -138,10 +138,10 @@ const connectSearchBox: SearchBoxConnector = function connectSearchBox(
         };
       },
 
-      getWidgetRenderState({ helper, searchMetadata }) {
+      getWidgetRenderState({ helper, searchMetadata, state }) {
         if (!_refine) {
           const setQueryAndSearch = (query: string) => {
-            if (query !== helper.state.query) {
+            if (query !== state.query) {
```
Haroenv, line 144: this is what caused the bug (state is stale in this callback), but the part of code isn't actually needed, we can still set the query and search without condition. Fixing in #4990

PR #4964, opened 2021-11-29
test(snapshots): automatically register widget serializer

Haroenv:
**Summary**

This makes it so that we can add any test that snapshots a widget and have the readable output.

No actual tests are changed, as all previous locations that used the widgetSnapshotSerializer manually set up the serializer

**Result**

all tests can snapshot widgets


Haroenv:
ok these tests are unexpectedly changing, so I'll close this PR for now, seems like Jest misbehaving

PR #4968, opened 2021-12-01
feat(dynamicWidgets): send facets * and maxValuesPerFacet by default

Haroenv:
**Summary**

The default additional search parameters of `{ facets: ['*'], maxValuesPerFacet: 20 }` prevents an extra network request when widgets mount/unmount, as both queries will be identical, and thus cached.

If a refinement is set, this still results in two network requests, as the refinements can't be applied before it's known if the widget should be mounted.

**Result**

When dynamic widgets is used without an explicit configure widget with `facets: ['*']` and `maxValuesPerFacet: highestLimit` an extra network request happens on load, which has a negative impact on performance. With the new `additionalParameters` default value, widgets mounting or unmounting no longer require a separate network request

Two new parameters are added to `connectDynamicWidgets` and `dynamicWidgets`:

```js
dynamicWidgets({
  // ... other parameters
  // overridden with the default value
  facets: ['*'],
  maxValuesPerFacet: 20,
});
```

The default behaviour can also be avoided, and the previous default (non-matching parameters, which does two network requests) can be done like this:

```js
dynamicWidgets({
  // ... other parameters
  // overridden default to ensure two requests are sent,
  // and that not all facets are requested
  facets: [],
});
```


src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
```diff
@@ -625,4 +625,14 @@ describe('connectDynamicWidgets', () => {
...
   });
+
+  describe('getWidgetSearchParameters', () => {
+    test.todo('it adds default facets and maxValuesPerFacet');
+
+    test.todo('it allows override of all parameters at once');
+
+    test.todo('it allows override with unrelated parameters');
+
+    test.todo('warns when a widget sets a higher limit');
```
Haroenv, line 636: still to do

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -156,7 +174,19 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
 
           unmountFn();
         },
+        getWidgetSearchParameters(state) {
+          return state.setQueryParameters(additionalParameters);
+        },
         getRenderState(renderState, renderOptions) {
+          // todo: should this warn when additionalParameters.maxValuesPerFacet is undefined?
```
Haroenv, line 181: should it warn when you explicitly leave out `maxValuesPerFacet` or is it worth warning about that behaviour (there's no point in on purpose requesting a lower number than what you're setting in limit)

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -156,7 +174,19 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
...
+          warning(
+            additionalParameters.maxValuesPerFacet
+              ? additionalParameters.maxValuesPerFacet >=
+                  renderOptions.state.maxValuesPerFacet!
+              : true,
+            `The maxValuesPerFacet set by dynamic widgets (${additionalParameters.maxValuesPerFacet}) is smaller than one of the limits set by a widget (${renderOptions.state.maxValuesPerFacet}). This causes a mismatch in query parameters and thus an extra network request when that widget is mounted.`
```
Haroenv, line 187: is this warning what we want

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -61,6 +75,10 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
     return (widgetParams) => {
       const {
         widgets,
+        additionalParameters = {
```
Haroenv, line 78: is this the right name/api? do we need two specific parameters instead?

sarahdayan:
Reviewing 🤓

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -180,6 +212,17 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
             );
           }
 
+          warning(
+            maxValuesPerFacet >= (state.maxValuesPerFacet || 0),
+            `The maxValuesPerFacet set by dynamic widgets (${maxValuesPerFacet}) is smaller than one of the limits set by a widget (${state.maxValuesPerFacet}). This causes a mismatch in query parameters and thus an extra network request when that widget is mounted.`
```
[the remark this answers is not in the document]
Haroenv: yes, it's allowed to be equal, then no warning is given, I wrote "smaller" as that seemed simpler to me

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -54,13 +71,17 @@ export type DynamicWidgetsConnector = Connector<
   DynamicWidgetsConnectorParams
 >;
 
+const MAX_WILDCARD_FACETS = 20;
```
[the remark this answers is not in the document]
Haroenv: those are unrelated

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -180,6 +212,17 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
...
 
+          warning(
+            maxValuesPerFacet >= (state.maxValuesPerFacet || 0),
+            `The maxValuesPerFacet set by dynamic widgets (${maxValuesPerFacet}) is smaller than one of the limits set by a widget (${state.maxValuesPerFacet}). This causes a mismatch in query parameters and thus an extra network request when that widget is mounted.`
+          );
+
+          warning(
+            attributesToRender.length <= MAX_WILDCARD_FACETS ||
+              widgetParams.facets !== undefined,
+            `More than ${MAX_WILDCARD_FACETS} facets are requested to be displayed without explicitly setting which facets to retrieve. This could have a performance impact. Set "facets" to [] to do two smaller network requests, or explicitly to ['*'] to avoid this warning.`
```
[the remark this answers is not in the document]
Haroenv: no, warning is only given when the condition is false, so when it's not smaller or equal (bigger)

src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
```diff
@@ -590,6 +590,52 @@ describe('connectDynamicWidgets', () => {
         widgetParams,
       });
     });
+
+    it('warns when facets is unset and more than 20 items are returned from attributesToDisplay', () => {
+      const renderFn = jest.fn();
+      const dynamicWidgets = connectDynamicWidgets(renderFn)({
```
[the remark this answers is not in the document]
Haroenv: consistent with the other tests, but indeed doesn't need to be a mock

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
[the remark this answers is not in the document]
sarahdayan: Right, I read the condition in the wrong order.

src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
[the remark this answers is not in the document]
sarahdayan: Got it. Not blocking but I guess we could start doing these adjustments even if it creates inconsistency, otherwise we'll likely never do it. The immediate advantage would be more readable tests.

Up to you.

sarahdayan (approved):

tkrugg (commented):
The changes look OK there are two things I'm noticing while playing with it.
When you set the facets to something, you have a double request: on the storybook, when I do the following:
```js
    instantsearch.widgets.dynamicWidgets({
          facets: ["categories","brand","rating", ... // add all facets


```
There is an extra HTTP query firing.
This might be that the resulting facet gets redundant attributes (I see `categories` twice for example)  and we may not be hitting the cache in that case

Haroenv:
@tkrugg I addressed this by not allowing values other than [] and [*], as passing specific facets wasn't a main use case, and has other possible solutions (virtual widgets, specific parameters on configure) that we can document later if people want that

src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
```diff
@@ -590,11 +627,54 @@ describe('connectDynamicWidgets', () => {
         widgetParams,
       });
     });
+
+    it('warns when facets is unset and more than 20 items are returned from attributesToDisplay', () => {
+      const dynamicWidgets = connectDynamicWidgets(() => {})({
+        transformItems() {
+          return Array.from({ length: 21 }, (_, i) => `test${i}`);
```
tkrugg, line 634: `Array.from(Array(21), ...)`  would probably be more readable.

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -39,6 +40,22 @@ export type DynamicWidgetsConnectorParams = {
...
     metadata: { results: SearchResults }
   ): string[];
+
+  /**
+   * To prevent unneeded extra network requests when widgets mount or unmount,
+   * we request all facet values.
+   *
+   * @default ['*']
+   */
+  facets?: ['*'] | never[];
```
tkrugg, line 50: I don't think you actually need the  `never`. `[]` types it like you'd expect.

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -180,6 +229,17 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
             );
           }
 
+          warning(
+            maxValuesPerFacet >= (state.maxValuesPerFacet || 0),
+            `The maxValuesPerFacet set by dynamic widgets (${maxValuesPerFacet}) is smaller than one of the limits set by a widget (${state.maxValuesPerFacet}). This causes a mismatch in query parameters and thus an extra network request when that widget is mounted.`
```
tkrugg, line 234: I think this message could be helpful if it:
- stated the name of the widget -> I know it probably adds a lot more complexity.
- said something like: look for any child widget (`menu` , `hierarhicalMenu` or `refinementList`) using `limit` and `showMoreLimit` and make sure to keep `maxValuesPerFacet` superior.

tkrugg (approved):

src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
Haroenv: why? that's what we are doing in other cases too, see eg. the range helper :)

PR #4971, opened 2021-12-08
feat(build): expose /es as a real es module

Haroenv:
**Summary**

- [x] add a package.json with {type: module} to the root of /es
- [x] add (fully resolved) .js extension to all imports via babel
- [x] setup storybook for dealing with the new paths
- [x] setup eslint for dealing with the new paths
- [x] add tests for module exports
- [x] update preact to a later version which supports module exports

**Result**

FX-534

instantsearch.js is now a real es module, but the paths or package for root isn't changed, as most imports are set manually

For users this shouldn't change anything, as the only difference will be that /es is marked as {type:module} and all of the imports are fully specified. If they're on older stacks that should not have an impact at all, but obviously needs to be researched


babel.config.js
```diff
@@ -25,7 +25,9 @@ module.exports = (api) => {
   ];
 
   const buildPlugins = clean([
-    '@babel/plugin-proposal-class-properties',
+    ['@babel/plugin-proposal-class-properties', { loose: true }],
+    ['@babel/plugin-proposal-private-methods', { loose: true }],
+    ['@babel/plugin-proposal-private-property-in-object', { loose: true }],
```
Haroenv, line 30: I'd rather do this for storybook alone, so looking into if it can have its own babel config

yarn.lock
```diff
@@ -143,11 +143,23 @@
   dependencies:
     "@babel/highlight" "^7.14.5"
 
+"@babel/code-frame@^7.16.0":
```
Haroenv, line 146: undo these changes

Haroenv:
This has now been tried out in sufficient environments to convince me there's no breaking change in marking /es as a module:

- Next.js (via React InstantSearch hooks)
- Vue InstantSearch via Vue CLI and Vue 2
- Vue InstantSearch via Vue CLI and Vue 3
- Create React App (via React InstantSearch hooks)

package.json
```diff
@@ -59,7 +60,7 @@
     "classnames": "^2.2.5",
     "@algolia/events": "^4.0.1",
     "hogan.js": "^3.0.2",
-    "preact": "^10.0.0",
+    "preact": "^10.5.14",
```
francoischalifour, line 63: Why is [v10.5.14](https://github.com/preactjs/preact/releases/tag/10.5.14) required now?

francoischalifour (approved):
Great!

package.json
Haroenv: it was the latest version that vscode suggested, I didn't check whether it was actually the latest, but it required at least https://github.com/preactjs/preact/pull/3320

francoischalifour: This PR was released in [v10.6.0](https://github.com/preactjs/preact/releases/tag/10.6.0).

Haroenv: will update to the latest!

tkrugg (approved):

PR #4975, opened 2021-12-13
feat(DynamicWidgets): throw when transformItems returns a non array

tkrugg:
`transformItems` implementations can get complicated and if you're not
cautious, it's very easy to have it return something other than an
array.

When this happens, the errors you get are not insightful on where the
mistake is.

```
TypeError: Cannot read property 'forEach' of undefined
```

This change helps the user identify this type of mistakes.
It adds a check on `transformItems` output to make sure it returns an
array.

closes FX-608

### DIFF
--- a/src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
+++ b/src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
@@ -182,11 +182,12 @@ describe('connectDynamicWidgets', () => {
         const renderFn = jest.fn();
         const widgetParams = {
           transformItems() {
-            // @ts-expect-error
             return null;
           },
           widgets: [],
         };
+
+        // @ts-expect-error
         const dynamicWidgets = connectDynamicWidgets(renderFn)(widgetParams);
 
         expect(() => {
--- a/src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
+++ b/src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
@@ -177,6 +177,27 @@ describe('connectDynamicWidgets', () => {
           false
         );
       });
+
+      it('fails when transformItems returns anything else than an array', () => {
+        const renderFn = jest.fn();
+        const widgetParams = {
+          transformItems() {
+            return null as any;
+          },
+          widgets: [],
+        };
+        const dynamicWidgets = connectDynamicWidgets(renderFn)(
+          widgetParams
+        );
+
+        expect(() => {
+          dynamicWidgets.render!(createRenderOptions());
+        }).toThrowErrorMatchingInlineSnapshot(`
+          "The \`transformItems\` option expects a function that returns an Array.
+
+          See documentation: https://www.algolia.com/doc/api-reference/widgets/dynamic-widgets/js/#connector"
+        `);
+      });
     });
 
     describe('widgets', () => {
--- a/src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
+++ b/src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
@@ -178,11 +178,12 @@ describe('connectDynamicWidgets', () => {
         );
       });
 
-      it('fails when transformItems returns anything else than an array', () => {
+      it('throws when transformItems returns anything else than an array', () => {
         const renderFn = jest.fn();
         const widgetParams = {
           transformItems() {
-            return null as any;
+            // @ts-expect-error
+            return null;
           },
           widgets: [],
         };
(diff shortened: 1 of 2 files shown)


src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -170,8 +170,21 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
...
+            { results }
+          );
+
+          if (!Array.isArray(transformedAttributesToRender)) {
+            throw new Error(
+              withUsage(
+                'The `transformItems` option expects a function that returns an Array.'
+              )
+            );
+          }
```
Haroenv, line 184: in the future this should be an invariant (like warning, but can have a slimmed down error message in production), but as we don't have that tooling yet, this is fine.

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -170,8 +170,21 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
           const attributesToRender =
             results.renderingContent?.facetOrdering?.facets?.order ?? [];
 
+          const transformedAttributesToRender = transformItems(
+            attributesToRender,
+            { results }
+          );
```
Haroenv, line 176: I'm ok with the existing suggestion as well

```suggestion
          const attributesToRender = transformItems(
            results.renderingContent?.facetOrdering?.facets?.order ?? [],
            { results }
          );
```


then a bit later: `attributesToRender` in the return

Haroenv (approved):
in theory you can also check that `transformedAttributesToRender.every(attribute => typeof attribute === 'string')`, but that seems overkill

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
tkrugg: good for me!

src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
```diff
@@ -177,6 +177,27 @@ describe('connectDynamicWidgets', () => {
...
+      it('fails when transformItems returns anything else than an array', () => {
+        const renderFn = jest.fn();
+        const widgetParams = {
+          transformItems() {
+            return null as any;
+          },
+          widgets: [],
+        };
+        const dynamicWidgets = connectDynamicWidgets(renderFn)(
+          widgetParams
```
Haroenv, line 190: lint requires this to be on previous line

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -167,11 +167,21 @@ const connectDynamicWidgets: DynamicWidgetsConnector =
...
+            throw new Error(
+              withUsage(
+                'The `transformItems` option expects a function that returns an Array.'
+              )
+            );
+          }
 
           return {
-            attributesToRender: transformItems(attributesToRender, { results }),
+            attributesToRender: transformedAttributesToRender,
```
Haroenv, line 184:
```suggestion
            attributesToRender,
```

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
francoischalifour: For future reference, here's the `invariant` we use in React InstantSearch: https://github.com/algolia/react-instantsearch/blob/acf33c13d5b4fdcce66fb961fb3393d8bfdd658f/packages/react-instantsearch-hooks/src/utils/invariant.ts

src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
```diff
@@ -177,6 +177,25 @@ describe('connectDynamicWidgets', () => {
           false
         );
       });
+
+      it('fails when transformItems returns anything else than an array', () => {
```
francoischalifour, line 181:
```suggestion
      it('throws when transformItems returns anything else than an array', () => {
```

src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
```diff
@@ -177,6 +177,25 @@ describe('connectDynamicWidgets', () => {
...
+        const renderFn = jest.fn();
+        const widgetParams = {
+          transformItems() {
+            return null as any;
```
francoischalifour, line 185: I think this better communicates that the code is wrong TypeScript-wise?

```suggestion
            // @ts-ignore
            return null;
```

francoischalifour (approved):

src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
Haroenv:
```suggestion
            // @ts-expect-error
            return null;
```

even better like this

tkrugg: it needs to be on the next line then.
Fixed it here: https://github.com/algolia/instantsearch.js/pull/4975/commits/1b7efdc1ef01498ed91867e71b88c07f55b17444

PR #4980, opened 2022-01-03
fix(connectBreadcrumb): returns an empty array if no hierarchicalFacets exist

FabienMotte:
**Summary**

- [x] returns an empty array of items if no `hierarchicalFacets` exist
- [x] refine to `null` throws an error if no `hierarchicalFacets` exist

**Result**

This PR fixes an error we had implementing the hook `useBreadcrumb` in React InstantSearch Hooks.

It also fixes an error when refine to `null` is called (usually used when you click on the "Home" link to clear the current refinement) when no `hierarchicalFacets` exist.

### DIFF
--- a/src/connectors/breadcrumb/connectBreadcrumb.ts
+++ b/src/connectors/breadcrumb/connectBreadcrumb.ts
@@ -132,7 +132,9 @@ const connectBreadcrumb: BreadcrumbConnector = function connectBreadcrumb(
         const breadcrumb = state.getHierarchicalFacetBreadcrumb(
           hierarchicalFacetName
         );
-        if (breadcrumb.length > 0) {
+        if (breadcrumb.length === 0) {
+          return state;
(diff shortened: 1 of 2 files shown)


Haroenv (approved):

src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -182,7 +184,7 @@ const connectBreadcrumb: BreadcrumbConnector = function connectBreadcrumb(
 
       getWidgetRenderState({ helper, createURL, results, state }) {
         function getItems() {
-          if (!results) {
+          if (!results || state.hierarchicalFacets.length === 0) {
```
francoischalifour, line 187: We can add a comment that explains that this expression is required for flavors that render immediately with empty results, without relying on `init()` (like RISHooks).

PR #4981, opened 2022-01-04
refactor(pagination): move from class-component to functional

Haroenv:
**Summary**

refactors the pagination component to not use a class anymore

This also simplifies the wrapping and removes implementation details from a test

**Result**

slightly better test, no class used in Pagination

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -1,6 +1,6 @@
 /** @jsx h */
 
-import { Component, h } from 'preact';
+import { h } from 'preact';
 import cx from 'classnames';
 
 import PaginationLink from './PaginationLink';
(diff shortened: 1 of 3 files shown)


src/components/Pagination/Pagination.tsx
```diff
@@ -32,145 +32,106 @@ export type PaginationProps = {
...
+  return (
+    <div
+      className={cx(props.cssClasses.root, {
+        [props.cssClasses.noRefinementRoot]: props.nbPages! <= 1,
+      })}
+    >
+      <ul className={props.cssClasses.list}>
+        {props.showFirst && (
+          <PaginationLink
+            key="first"
```
sarahdayan, line 75: Why do we need a key here since we're not in a list? Same question for previous, next and last.

src/components/Pagination/Pagination.tsx
```diff
@@ -32,145 +32,106 @@ export type PaginationProps = {
...
+            isDisabled={props.isFirstPage}
+            label={props.templates.previous}
+            pageNumber={props.currentPage - 1}
+            createURL={props.createURL}
+            cssClasses={props.cssClasses}
+            handleClick={handleClick}
+          />
+        )}
+
+        {pages}
```
sarahdayan, line 101: Not blocking, but since we're not using `pages` anywhere else, it could be inlined so the code is visually in the same order as the visual representation.

sarahdayan (approved):
Clean job!

src/components/Pagination/Pagination.tsx
Haroenv: we actually are in a list: `[first, previous, ...pages, next, last]`

src/components/Pagination/Pagination.tsx
Haroenv: sure, this was just closer to the previous implementation, I prefer it inline too

src/components/Pagination/Pagination.tsx
sarahdayan: This element is not part of an array, is it? It should be stable.

src/components/Pagination/Pagination.tsx
```diff
@@ -32,145 +32,104 @@ export type PaginationProps = {
...
-  label: string;
-  ariaLabel: string;
-  pageNumber: number;
-  additionalClassName: string | null;
-  isDisabled?: boolean;
-  isSelected?: boolean;
-  createURL(value: number): string;
-};
-
 const defaultProps = {
```
francoischalifour, line 35: This default props pattern doesn't work well with TypeScript, it lies about some fields being optional, while they can't be. We could default the props line 41 when destructuring.

src/components/Pagination/Pagination.tsx
francoischalifour: It is not needed here, as @sarahdayan explained.

src/components/Pagination/Pagination.tsx
```diff
@@ -32,145 +32,104 @@ export type PaginationProps = {
...
+    <div
+      className={cx(props.cssClasses.root, {
+        [props.cssClasses.noRefinementRoot]: props.nbPages! <= 1,
+      })}
+    >
+      <ul className={props.cssClasses.list}>
+        {props.showFirst && (
+          <PaginationLink
+            key="first"
+            ariaLabel={'First'}
```
francoischalifour, line 62: This is the syntax we use everywhere else. (Same applies in other places.)

```suggestion
            ariaLabel="First"
```

src/components/Pagination/PaginationLink.tsx
```diff
@@ -1,36 +1,48 @@
...
-  ariaLabel: string;
-  cssClasses: PaginationLinkComponentCSSClasses;
-  handleClick(pageNumber: number, event: MouseEvent): void;
-  isDisabled: boolean;
+type PageLinkProps = {
   label: string;
+  ariaLabel: string;
   pageNumber: number;
-  url?: string;
+  additionalClassName: string | null;
```
francoischalifour, line 12: Why introduce a new API when we can use the native `className` prop?

src/components/Pagination/PaginationLink.tsx
```diff
@@ -1,36 +1,48 @@
...
+}: PageLinkProps) {
+  const classes = {
+    item: cx(
+      cssClasses.item,
+      additionalClassName,
+      isDisabled && cssClasses.disabledItem,
+      isSelected && cssClasses.selectedItem
+    ),
+    link: cssClasses.link,
+  };
```
francoischalifour, line 39: We could reduce nesting by having two variables: `itemClassName` and `linkClassName`. The object as a whole doesn't hold any specific meaning.

PR #4987, opened 2022-01-06
chore(tests): recommend testing-library

Haroenv:
**Summary**

This will serve as a marker when editing tests that it might be time to refactor from enzyme to testing-library.

Reasons for this are:

- less buggy (enzyme for example doesn't distinguish between className passed to a component and a dom element)
- doesn't need hacks to work with preact

**Result**

existing tests using enzyme's mount/shallow are marked as deprecated, but obviously still work and don't fail the linting

### DIFF
--- a/test/utils/enzyme.ts
+++ b/test/utils/enzyme.ts
@@ -2,10 +2,16 @@ import type { ShallowWrapper, ReactWrapper } from 'enzyme';
 import { mount, shallow } from 'enzyme';
 import type { VNode } from 'preact';
 
+/**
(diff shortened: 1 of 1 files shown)


francoischalifour (commented):
We don't use these functions anymore so we can even remove them?

Haroenv:
we still use those functions, `import.*mount` still has 7 results in tests for me @francoischalifour

francoischalifour:
@Haroenv Indeed, I was searching for `preactMount` but we export them with a different name.

francoischalifour (approved):

PR #4988, opened 2022-01-06
refactor(hits): remove unneeded defaultProps

Haroenv:
**Summary**

Removed defaultProps from Hits component as the props are always present (similar finding as in #4986)

Also made defaultProps forbidden via eslint (for function components, classes can come when they're refactored to functions)

**Result**

- defaultProps is used in no further function components
- eslint will error if defaultProps is used by a function component


FabienMotte (approved):
Looks good to me

.eslintrc.js
```diff
@@ -52,6 +52,13 @@ module.exports = {
       },
     ],
     '@typescript-eslint/consistent-type-imports': 'error',
+    'no-restricted-syntax': [
+      'error',
+      {
+        selector: 'AssignmentExpression [name=defaultProps]',
+        message: 'defaultProps are not allowed, use function defaults instead.',
+      },
+    ],
```
francoischalifour, line 61: Wondering if this needed now that we don't have any occurrences since it's not a common pattern?

francoischalifour (approved):

.eslintrc.js
Haroenv: I accidentally used it in new code yesterday, so it's better to have it for a bit, especially when there's more class-to-function refactors to come where it could be forgotten to migrate

PR #4989, opened 2022-01-06
feat(routing): new router option to prevent write on dispose

FabienMotte:
**Summary**

This PR adds an option called `writeOnDispose` to the history router. It allows to whether call the `write` method on `dispose` or not. By default, the option is set to `true`.

This option allows to prevent `write` on `dispose`, so `window.history.pushState` is not called when the InstantSearch instance is disposed (when you go to another page for example).

https://github.com/algolia/instantsearch.js/blob/45433de9e21801f8b0c6d0859cd4e6aacb666104/src/lib/routers/history.ts#L198-L202

Also, in the router middleware, `write` method in `onStateChange` is prevented if the middleware gets unsubscribed.

https://github.com/algolia/instantsearch.js/blob/45433de9e21801f8b0c6d0859cd4e6aacb666104/src/middlewares/createRouterMiddleware.ts#L71-L78

For more context, you can read the related PR: https://github.com/algolia/instantsearch.js/pull/4808

**Result**

As a consequence, the browser "Go back" button should work as expected when `writeOnDispose` is set to `false`.

#### `writeOnDispose=false`

https://user-images.githubusercontent.com/662153/148411239-94f2966e-81cd-4b2c-9f31-c548ffecc69e.mov

#### `writeOnDispose=true` (default behavior)

https://user-images.githubusercontent.com/662153/148411122-1a7e9d4b-0fad-4e5c-a55d-5818cc751763.mov


src/lib/routers/history.ts
```diff
@@ -78,6 +79,13 @@ class BrowserHistory<TRouteState> implements Router<TRouteState> {
    */
   private shouldPushState: boolean = true;
 
+  /**
+   * Indicates if `write` should be called on `dispose`.
```
Haroenv, line 83:
```suggestion
   * Indicates if `write` should be called on `dispose`.
   * If another process (eg. frontend SPA router) is in charge of navigating
   * away from the search page, set this option to `false`.
```

src/lib/routers/history.ts
```diff
@@ -78,6 +79,13 @@ class BrowserHistory<TRouteState> implements Router<TRouteState> {
...
+   *
+   * @default true
+   */
+  private writeOnDispose: boolean = true;
```
Haroenv, line 87: I'd like to find a better name, but can't think of anything

- cleanup: true/false
- cleanupOnDispose: true/false
- singlepage: false/true (not really a fan)
- externalRouter: false/true (not really a fan)

src/middlewares/createRouterMiddleware.ts
```diff
@@ -62,13 +62,16 @@ export const createRouterMiddleware = <
 
     const initialUiState = instantSearchInstance._initialUiState;
 
+    let unsubscribed = false;
```
Haroenv, line 65: unsubscribed needs to be set to true on `subscribe`

Haroenv: we'd probably also need a test with the flow:

start
change (url changes)
dispose
change (url doesn't change)
start
change (url changes)

Haroenv (commented):
this was a long time coming, and I'm happy we'll finally have a better solution than "copy the router" for this feature

src/lib/routers/history.ts
sarahdayan: @Haroenv What don't you like about it? What meaning should it convey that you think is missing right now?

src/lib/routers/history.ts
```diff
@@ -190,7 +200,11 @@ class BrowserHistory<TRouteState> implements Router<TRouteState> {
       clearTimeout(this.writeTimer);
     }
 
-    this.write({} as TRouteState);
+    if (this.writeOnDispose) {
+      this.write({} as TRouteState);
+    } else {
+      this.shouldPushState = false;
```
sarahdayan, line 206: What does this prevent? I've ran the tests without this line and nothing breaks.

We likely need to test this behavior if this is actually needed.

src/lib/routers/history.ts
sarahdayan:
```suggestion
   * Indicates if `write` should be called on `dispose.
   * When using other client-side routing utilities to navigate between pages
   * (e.g., a front-end SPA routing library), set this option to `false`.
```

src/lib/__tests__/routing-test.ts
```diff
@@ -0,0 +1,152 @@
...
+      expect(window.location.search).toEqual(
+        `?${encodeURI('indexName[query]=Apple')}`
+      );
+      expect(pushState).toHaveBeenCalledTimes(1);
+
+      // Trigger a dispose
+      search.dispose();
+
+      // Wait for ${writeDelay} to pass then check URL has been cleaned
+      await wait(1.5 * writeDelay);
```
sarahdayan, line 50: I'm assuming the 1.5 factor is to ensure the `writeDelay` did elapse and avoid false positives. Since we run this 10 times from what I've counted, we're adding 6 seconds to this test suite.

Could we reduce `writeDelay` to avoid slowing down the test suite too much?

src/lib/__tests__/routing-test.ts
```diff
@@ -0,0 +1,152 @@
...
+          }),
+        },
+      });
+
+      search.addWidgets([connectSearchBox(() => {})({})]);
+
+      search.start();
+
+      // Wait for ${writeDelay} to pass then check URL has been initialized
+      await wait(1.5 * writeDelay);
```
sarahdayan, line 32: I'm assuming the 1.5 factor is to ensure the `writeDelay` did elapse and avoid false positives. Since we run this 10 times from what I've counted, we're adding 6 seconds to this test suite.

Could we reduce `writeDelay` to avoid slowing down the test suite too much?

src/lib/routers/history.ts
Haroenv: for me it seems an implementation detail, I'd rather have the option explain why you'd want to do that in the first place

src/lib/__tests__/routing-test.ts
Haroenv: fake timers should also fix that

src/lib/__tests__/routing-test.ts
```diff
@@ -0,0 +1,152 @@
...
+            writeDelay,
+          }),
+        },
+      });
+
+      search.addWidgets([connectSearchBox(() => {})({})]);
+
+      search.start();
+
+      // Wait for ${writeDelay} to pass then check URL has been initialized
```
sarahdayan, line 31: I'm not sure all the comments clarify anything, to me they're describing what the code already says.

Why do you think we need cooments here?

src/lib/__tests__/routing-test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+describe('routing', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('writeOnDispose=true', () => {
+    test('cleans URL on dispose', async () => {
+      const pushState = jest.spyOn(window.history, 'pushState');
+      const writeDelay = 400;
```
sarahdayan, line 15: Do we need a `writeDelay` in every test? I understand that we want to test this behavior, but having it for every test (10x in this file) slows down the CI. Wondering if this is necessary for any other reason.

src/lib/__tests__/routing-test.ts
```diff
@@ -0,0 +1,152 @@
...
+      );
+      expect(pushState).toHaveBeenCalledTimes(1);
+
+      // Trigger a dispose
+      search.dispose();
+
+      // Wait for ${writeDelay} to pass then check URL has been cleaned
+      await wait(1.5 * writeDelay);
+      expect(window.location.search).toEqual('');
+      expect(pushState).toHaveBeenCalledTimes(2);
```
sarahdayan, line 94: This whole part is the same as the previous `test`. Why do we need it again? Can't we either add the last part to the previous test, or just set things up without asserting and waiting to get up to this point?

src/lib/__tests__/routing-test.ts
sarahdayan: Yep, I assumed this was tried already.

src/middlewares/createRouterMiddleware.ts
FabienMotte: > unsubscribed needs to be set to true on `subscribe`

You mean, `unsubscribed = false;` on `subscribe`?

src/lib/__tests__/routing-test.ts
FabienMotte: @sarahdayan I've reduce the `writeDelay` to 10ms, which bring the total waiting time to 600ms, do you think that's acceptable?

src/middlewares/createRouterMiddleware.ts
Haroenv: yes, I meant false indeed, although we probably want to make the logic opposite to avoid negations (using subscribed = false initially, and checking if !subscribed )

src/lib/__tests__/routing-test.ts
FabienMotte: Sometimes we need to check if the URL was initialized/update/cleaned (after a dispose), that's why this comment is for, but I guess I can remove the beginning "Wait for ${writeDelay} to pass then" which is common for every sentence

src/middlewares/createRouterMiddleware.ts
FabienMotte: Thanks, I'll make the update. Also, do we want to prevent the calls to `subscribe`/`unsubscribe` based on the `subscribed` boolean?

```
subscribe() {
  if(subscribed) return;
  subscribed = true;
   // ...
}
```

Haroenv: subscribe will not be called if already subscribed, because it would mean the middleware was added twice, or InstantSearch started twice, both aren't possible

FabienMotte: Thanks for the clarification!

src/lib/__tests__/routing-test.ts
FabienMotte: Good point, I'll remove the duplicated assertions, thank you!

src/lib/routers/history.ts
FabienMotte: Should we add a paragraph in the docs to explain this new option (and keep the name as is it)?

francoischalifour:
Why shouldn't this be the default behavior?

If we find scenarios where this behavior as a default causes problems, I feel like we can find a sensible default without introducing an option.

Haroenv:
In regular InstantSearch.js, it probably should default to the current behaviour, as on dispose, the InstantSearch is no longer valid, and thus should return the URL to its previous state.

Maybe that shouldn't be the default, but changing it is breaking

francoischalifour:
I think that keeping the URL on `dispose()` in all flavors may be a good strategy, because:

1. the next UI state change will restore the URL
2. if users _need_ to clean the URL, they can manipulate the URL themselves (no API), or call a method on the router (different API: a method and not an option)

I'm wondering if (2) even happens. How many users call `dispose()` in InstantSearch.js? Is it a problem for them that the URL stays unchanged? Do they even notice?

I'm not sure whether this slight behavior change is a breaking change. Do you think that it could break users' apps? tests?

Haroenv:
I'm not sure if anyone expects that behaviour on purpose, but InstantSearch has always cleaned/cleared the url on dispose. People might have a ui where InstantSearch is conditional, and when it unmounts, while still on the same page, would expect to get back to a clean URL?

Again, not sure if this matters, but I'd edge on the side of caution

src/lib/__tests__/routing-test.ts
sarahdayan: Did you try fake timers? Not sure they work in that situation but worth trying.

Otherwise, 600ms sounds reasonable.

FabienMotte: Yes, but every test exceed the max timeout, it seems like the same issue I have: https://github.com/nock/nock/issues/2200

FabienMotte:
Superseded by: https://github.com/algolia/instantsearch.js/pull/5004

PR #4990, opened 2022-01-06
fix(searchbox): make sure setting query to the initial doesn't cause a stale state

### LINKED ISSUE #4979: searchBox reset button is broken
## 🐛 Bug description

As of v4.34.0, `searchBox`'s reset button is broken: it clears the search input but does not fire a new search. Downgrading to v4.33.2 fixes the issue.

## 🔍 Bug reproduction

1. Perform a search, preferably for something with no results (e.g. "whendidtimebegin") so that the issue can be clearly seen.
2. Click the reset button (the "X" on the right side of the search input).
3. Note that the search results are not refreshed.

**Live reproduction:**

The issue is reproducible via Algolia's official Storybook: https://instantsearchjs.netlify.app/stories/?path=/story/basics-searchbox--default

## 💭 Expected behavior

When `searchBox`'s reset button is clicked, the search input is cleared and search results are refreshed.

## 🖥 Screenshots

N/A, as it's easily reproducible via the Storybook.

## Environment

- OS: macOS
- Browser: Chrome, Safari (browser doesn't seem to be the issue)
- Version: Chrome 96.0.4664.110, Safari 15.1 (17612.2.9.1.20)

Haroenv:
**Summary**

I don't understand yet how this is working, but somehow if `_clear` is defined only once, refine('') starts actually clearing, even if that function doesn't use `_clear` itself

**Result**

fixes #4979

### DIFF
--- a/src/connectors/search-box/connectSearchBox.ts
+++ b/src/connectors/search-box/connectSearchBox.ts
@@ -1,4 +1,3 @@
-import type { AlgoliaSearchHelper } from 'algoliasearch-helper';
 import {
   checkRendering,
   createDocumentationMessageGenerator,
@@ -71,6 +70,9 @@ export type SearchBoxConnector = Connector<
   SearchBoxConnectorParams
 >;
 
+const defaultQueryHook: SearchBoxConnectorParams['queryHook'] = (query, hook) =>
+  hook(query);
+
 /**
  * **SearchBox** connector provides the logic to build a widget that will let the user search for a query.
  *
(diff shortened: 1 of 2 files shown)


src/connectors/search-box/connectSearchBox.ts
```diff
@@ -140,28 +133,19 @@ const connectSearchBox: SearchBoxConnector = function connectSearchBox(
 
       getWidgetRenderState({ helper, searchMetadata, state }) {
         if (!_refine) {
-          const setQueryAndSearch = (query: string) => {
-            if (query !== state.query) {
```
Haroenv, line 144: removing this conditional is what really fixes the bug, but the rest of this PR is cleanup

Haroenv: I've tested React InstantSearch hooks, Angular InstantSearch, Vue InstantSearch and vanilla and I see no cases where refine gets called with the same query as already was the case, so this previous optimisation isn't needed as far as I can tell

francoischalifour: For reference, this branch was introduced in 81c2e80dd7ebb1bdac284a579b898c436babc03b.

francoischalifour (approved):

PR #4996, opened 2022-01-24
fix(helpers): display warning if attribute is cannot be highlighted/snippeted

dhayab:
**Summary**

This PR shows a helpful warning in dev environment when an attribute cannot be highlighted or snippeted.
It's an adaptation from a similar message in the custom highlight implementation in Angular.

**Result**

image

### DIFF
--- a/src/helpers/highlight.ts
+++ b/src/helpers/highlight.ts
@@ -1,6 +1,6 @@
 import type { Hit } from '../types';
 import { component } from '../lib/suit';
-import { getPropertyByPath, TAG_REPLACEMENT } from '../lib/utils';
+import { getPropertyByPath, TAG_REPLACEMENT, warning } from '../lib/utils';
 
 export type HighlightOptions = {
   // @MAJOR string should no longer be allowed to be a path, only array can be a path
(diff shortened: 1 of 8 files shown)


src/helpers/highlight.ts
```diff
@@ -20,8 +20,22 @@ export default function highlight({
...
+    attribute
+  );
+
+  // @MAJOR fallback to attribute value if highlight is not found
+  warning(
+    highlightAttributeResult,
+    `Could not enable highlight for "${attribute}", will display an empty string.
+Please check whether this attribute exists and is either searchable or specified in \`attributesToHighlight\`.
+
+See: https://www.algolia.com/doc/guides/building-search-ui/ui-and-ux-patterns/highlighting-snippeting/js/
```
dhayab, line 34: I see other examples using an "alg.li" URL shortener, but I didn't find documentation on how to access it. If you have an idea on how to do it, I'll happily update the links.

Haroenv (approved):
this is great! I think it's technically a feat, not a fix, but then we'd have a minor version if this is the only thing we'd release 🤔

src/helpers/highlight.ts
Haroenv: I just created https://alg.li/highlighting (it's in short.cm, you can DM if you want access too, or check Okta)

Haroenv:
```suggestion
See: https://alg.li/highlighting
```

dhayab:
Yeah, since I want to use this as the basis for another task in Angular IS, I didn't want to go for a minor version update, but I'll follow what you think is best in this case.

src/helpers/highlight.ts
dhayab: Replaced all occurrences in 432be6e.

PR #4997, opened 2022-01-24
feat(dynamicWidgets): prevent calling of "render" when widgets are not yet added

Haroenv:
**Summary**

**Result**


Haroenv:
This would also solve infinite hits rendering with wrong results and messing up the cache

PR #5000, opened 2022-01-26
feat(connectors): expose search results to `transformItems` when available

dhayab:
**Summary**

This PR adds a secondary parameter to `transformItems` on all connectors that support it. It currently exposes the search results from Algolia.

**Result**

Search results can now be used in transformItems for various applications. For example, you can show the number of hits in the last segment of a breadcrumb widget:

```js
// ...

instantsearch.widgets.breadcrumb({
  container: breadcrumbDiv,
  attributes: [
    'hierarchicalCategories.lvl0',
    'hierarchicalCategories.lvl1',
    'hierarchicalCategories.lvl2',
  ],
  transformItems: (items, { results }) => {
    const lastCrumb = items.pop();
    return [
      ...items,
      {
        ...lastCrumb,
        label: `${lastCrumb.label} (${results.nbHits} hits)`,
      },
    ];
  },
}),

// ...
```

### DIFF
--- a/src/connectors/dynamic-widgets/connectDynamicWidgets.ts
+++ b/src/connectors/dynamic-widgets/connectDynamicWidgets.ts
@@ -1,3 +1,5 @@
+import type { SearchResults } from 'algoliasearch-helper';
+
 import {
   checkRendering,
   createDocumentationMessageGenerator,
@@ -35,7 +37,7 @@ export type DynamicWidgetsConnectorParams = {
    * Function to transform the items to render.
    * The function also exposes the full search response.
    */
-  transformItems?: TransformItems<string>;
+  transformItems?: TransformItems<string, { results: SearchResults }>;
 
   /**
    * To prevent unneeded extra network requests when widgets mount or unmount,
--- a/src/connectors/dynamic-widgets/connectDynamicWidgets.ts
+++ b/src/connectors/dynamic-widgets/connectDynamicWidgets.ts
@@ -1,5 +1,3 @@
-import type { SearchResults } from 'algoliasearch-helper';
-
 import {
   checkRendering,
   createDocumentationMessageGenerator,
@@ -7,7 +5,12 @@ import {
   noop,
   warning,
 } from '../../lib/utils';
-import type { Connector, TransformItems, Widget } from '../../types';
+import type {
+  Connector,
+  TransformItems,
+  TransformItemsMetadata,
+  Widget,
+} from '../../types';
 
 const withUsage = createDocumentationMessageGenerator({
   name: 'dynamic-widgets',
@@ -37,7 +40,12 @@ export type DynamicWidgetsConnectorParams = {
    * Function to transform the items to render.
    * The function also exposes the full search response.
    */
-  transformItems?: TransformItems<string, { results: SearchResults }>;
+  transformItems?: TransformItems<
+    string,
+    Omit<TransformItemsMetadata, 'results'> & {
+      results: NonNullable<TransformItemsMetadata['results']>;
+    }
+  >;
 
   /**
    * To prevent unneeded extra network requests when widgets mount or unmount,
--- a/src/connectors/dynamic-widgets/connectDynamicWidgets.ts
+++ b/src/connectors/dynamic-widgets/connectDynamicWidgets.ts
@@ -1,4 +1,3 @@
-import type { SearchResults } from 'algoliasearch-helper';
 import {
   checkRendering,
   createDocumentationMessageGenerator,
@@ -6,7 +5,7 @@ import {
   noop,
   warning,
 } from '../../lib/utils';
-import type { Connector, Widget } from '../../types';
+import type { Connector, TransformItems, Widget } from '../../types';
 
 const withUsage = createDocumentationMessageGenerator({
   name: 'dynamic-widgets',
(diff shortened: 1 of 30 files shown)


src/widgets/dynamic-widgets/__tests__/dynamic-widgets-test.ts
```diff
@@ -401,7 +401,7 @@ describe('dynamicWidgets()', () => {
         dynamicWidgets({
           container: rootContainer,
           transformItems(_items, { results }) {
-            return results.userData[0].MOCK_facetOrder;
+            return results!.userData[0].MOCK_facetOrder;
```
Haroenv, line 404: turning something non-optional to optional probably is a breaking change, especially as it isn't called with undefined results. Maybe allow the whole metadata parameter to be generic instead of only the results?

src/types/widget.ts
```diff
@@ -270,7 +270,12 @@ export type Widget<
 /**
  * Transforms the given items.
  */
-export type TransformItems<TItem> = (items: TItem[]) => TItem[];
+export type TransformItems<TItem, TResults = SearchResults<TItem>> = (
+  items: TItem[],
+  metadata: {
+    results: TResults | undefined;
+  }
+) => TItem[];
```
Haroenv, line 278:
```suggestion
export type TransformItems<TItem, TMetaData = {
    results?: SearchResults;
  }> = (
  items: TItem[],
  metadata: TMetadata
) => TItem[];
```

we shouldn't link item to results, as item is only the hits inside the hits/infinitehits component, where you can pass the generic explicitly by overriding TMetadata. What do you think?

src/connectors/refinement-list/__tests__/connectRefinementList-test.ts
```diff
@@ -520,6 +520,35 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
...
+      attribute: 'category',
+      transformItems,
+    });
+
+    const helper = jsHelper(createSearchClient(), '');
+
+    widget.init!(createInitOptions({ helper, state: helper.state }));
+    widget.render!(
+      createRenderOptions({
+        results: new SearchResults(helper.state, [
```
Haroenv, line 536: if results is saved in a variable before rendering, it can checked in the called with

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
```diff
@@ -36,10 +40,12 @@ export type DynamicWidgetsConnectorParams = {
...
    */
-  transformItems?(
-    items: string[],
-    metadata: { results: SearchResults }
-  ): string[];
+  transformItems?: TransformItems<
+    string,
+    Omit<TransformItemsMetadata, 'results'> & {
+      results: NonNullable<TransformItemsMetadata['results']>;
+    }
```
Haroenv, line 47: maybe we want to have an "override" internal type helper later, so it could be `type Override<TOriginal, TOverride> => Omit<TOriginal, keyof TOverride> & TOverride`, but not sure if we have the use case in other points already (possibly search for Omit in the code base, maybe index widget?)

Haroenv (commented):
👍

src/connectors/refinement-list/__tests__/connectRefinementList-test.ts
Haroenv: (same for all other added tests)

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
dhayab: This is indeed a pattern used in a few files. I'll create a task in Jira with the current list I have so we can dedicate time for it.

src/connectors/refinement-list/__tests__/connectRefinementList-test.ts
dhayab: Done in 7f31fe1c.

src/connectors/dynamic-widgets/connectDynamicWidgets.ts
dhayab: Done: [FX-780](https://algolia.atlassian.net/browse/FX-780).

Haroenv (approved):

sarahdayan (approved):
Good improvement!

Two notes:
- Let's make sure to document the new signature of `transformItems` in the documentation (although I assume you might want to first update InstantSearch.js in RISH, VIS and AIS then do a batch).
- Now that we have an object, it might be interesting to remove `items` as the first parameter in the next major and move it to the metadata object. Should we deprecate it now?

Haroenv:
> Now that we have an object, it might be interesting to remove items as the first parameter in the next major and move it to the metadata object. Should we deprecate it now?

I don't think that is needed. In most cases you want do do something like `items => items.map`, changing that to `({items}) => items.map` doesn't seem more friendly. Also I'm not sure how putting it in the second argument would help deprecation, as code would still need to change if we remove the first argument in a major

dhayab:
I created a follow-up task for documentation, after updating IS.js in all flavours. For the second note, I tend to agree with @Haroenv, as the method directly deals with `items`, it makes sense to separate it from other arguments.

francoischalifour:
I agree with Haroen too, that's a common pattern in functional programming, which also supports the identity function, and easy manipulation with Ramda for example.

sarahdayan:
@Haroenv @francoischalifour I meant have a single object argument, not keeping two arguments.

```js
({ items, results }) => items.map(fn).concat(results.nbHits)
```

PR #5002, opened 2022-01-27
fix(routing): initialize uistate early

Haroenv:
fixes https://github.com/algolia/vue-instantsearch/issues/1066
fixes https://github.com/algolia/vue-instantsearch/issues/1062
FX-349

see: https://github.com/algolia/instantsearch.js/pull/4849

I

**Summary**

n Vue InstantSearch SSR, the mainIndex gets init'ed *before* it does in regular InstantSearch (see https://codesandbox.io/s/beautiful-mestorf-ijc75?file), and at that point the ui state should already be initialized (before start).

**Result**

initial ui state gets set before subscribe of the middleware (before start), as was the case before #4849

### DIFF
--- a/src/middlewares/createRouterMiddleware.ts
+++ b/src/middlewares/createRouterMiddleware.ts
@@ -62,6 +62,11 @@ export const createRouterMiddleware = <
 
     const initialUiState = instantSearchInstance._initialUiState;
 
+    instantSearchInstance._initialUiState = {
+      ...initialUiState,
+      ...stateMapping.routeToState(router.read()),
+    };
+
     return {
       onStateChange({ uiState }) {
         const routeState = stateMapping.stateToRoute(uiState);
--- a/src/middlewares/__tests__/createRouterMiddleware.ts
+++ b/src/middlewares/__tests__/createRouterMiddleware.ts
@@ -4,7 +4,44 @@ import instantsearch from '../../index.es';
 import { searchBox } from '../../widgets';
 
 describe('router', () => {
-  it('sets initial ui state after reading URL', async () => {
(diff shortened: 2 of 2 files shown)


francoischalifour:
Couldn't we fix that in Vue InstantSearch? We now have the `instantSearchInstance._initialUiState` assignation both in the middleware creation and in `subscribe()`.

src/middlewares/createRouterMiddleware.ts
```diff
@@ -62,6 +62,11 @@ export const createRouterMiddleware = <
 
     const initialUiState = instantSearchInstance._initialUiState;
 
+    instantSearchInstance._initialUiState = {
+      ...initialUiState,
+      ...stateMapping.routeToState(router.read()),
+    };
```
dhayab, line 68: Is it possible to reuse the same code here and in `subscribe()`?

dhayab (approved):

src/middlewares/createRouterMiddleware.ts
francoischalifour, line 68: If ever we go for this solution, we need to document why this happens here.

But again, this doesn't seem too right to me. Perhaps we can manually call `subscribe()` where it's needed in VIS.

francoischalifour (changes requested):

src/middlewares/createRouterMiddleware.ts
Haroenv: That's an interesting idea

Haroenv:
This fix was only needed for Vue InstantSearch 3, so I need to create a branch there first to be able to release a hotfix

PR #5009, opened 2022-02-08
fix(metadata): ensure safe user agent detection

sarahdayan:
## Summary

This ensures user agent detection doesn't break in non-browser environments.

In React Native, `window.navigator` is [polyfilled with a custom `product` property](https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/Libraries/Core/setUpNavigator.js), and doesn't have other properties traditionally exposed in the browser.

This change ensures this code path doesn't yield a runtime error regardless of the environment.

### DIFF
--- a/src/middlewares/__tests__/createMetadataMiddleware.ts
+++ b/src/middlewares/__tests__/createMetadataMiddleware.ts
@@ -13,8 +13,11 @@ declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace NodeJS {
     interface Global {
-      navigator: {
-        userAgent: string;
+      navigator?: {
+        userAgent?: string;
+        // Deprecated property, but still used in React Native
+        // https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/Libraries/Core/setUpNavigator.js
+        product?: string;
       };
       window: Window;
     }
@@ -22,6 +25,7 @@ declare global {
 }
 
 const { window } = global;
+
 Object.defineProperty(
   window.navigator,
   'userAgent',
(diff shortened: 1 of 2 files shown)


src/middlewares/__tests__/createMetadataMiddleware.ts
```diff
@@ -13,28 +13,45 @@ declare global {
...
 Object.defineProperty(
   window.navigator,
   'userAgent',
   ((value) => ({
     get() {
       return value;
     },
-    set(v: string) {
-      value = v;
+    set(newValue) {
```
sarahdayan, line 36: Removed the typing as it doesn't have any effect. The above `namespace` does the type inference.

src/middlewares/__tests__/createMetadataMiddleware.ts
```diff
@@ -46,14 +63,14 @@ describe('createMetadataMiddleware', () => {
...
-      global.navigator.userAgent = defaultUserAgent;
+      global.navigator!.userAgent = defaultUserAgent;
 
       expect(isMetadataEnabled()).toBe(false);
+
+      global.window = window;
     });
 
     it("does not enable when there's no window", () => {
-      global.navigator.userAgent = algoliaUserAgent;
```
sarahdayan, line 55: I believe we don't need it in this test.

src/middlewares/__tests__/createMetadataMiddleware.ts
sarahdayan: Well, our linter doesn't like it apparently 😅

sarahdayan: Ignored the error, as it's not needed.

src/middlewares/__tests__/createMetadataMiddleware.ts
```diff
@@ -13,28 +13,49 @@ declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace NodeJS {
     interface Global {
-      navigator: {
-        userAgent: string;
+      navigator?: {
+        userAgent?: string;
+        // Deprecated property, but still used in React Native
+        // https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/Libraries/Core/setUpNavigator.js
+        product?: string;
```
Haroenv, line 20:
```suggestion
```

we don't use it, so no need to test it

sarahdayan: This is useful to test the React Native use case accurately. How would you do it?

PR #5012, opened 2022-02-11
test(jest): migrate to new defaults

Haroenv:
**Summary**

- use the jest-circus runner, which had no impact except slightly speeding up the build
- use the "node" test environment by default, and changing the tests that need it to jsdom

**Result**

Both of these changes had an impact to make it so that the complete test run (cached) went from ~20 to ~15 seconds on my machine.

Read more on the motivation of these changes here: https://jestjs.io/blog/2021/05/25/jest-27

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,7 +2,8 @@
 
 module.exports = {
   rootDir: process.cwd(),
-  testEnvironment: 'jest-environment-jsdom-global',
+  testRunner: 'jest-circus',
+  testEnvironment: 'node',
   setupFilesAfterEnv: ['./scripts/jest/setupTests.ts'],
   testPathIgnorePatterns: [
     '<rootDir>/node_modules/',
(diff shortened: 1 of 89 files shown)


sarahdayan (approved):
A -25% drop in CI time is always welcome 👏 

Now that annotations are needed when you need a JSDOM environment, can we maybe [update the contributing guide](https://github.com/algolia/instantsearch.js/blob/master/CONTRIBUTING.md#tests)?

It's easy to miss out unless you don't chek other test specs, and the error you'd get might not be easy to grasp unless you know the problem. If contributors can grep it in the repo and immediately find a solution, it's much better DX.

Haroenv:
There's actually a very helpful error given by jest if you try to use window/document etc. in Jest that suggests using jsdom, so I don't think I'd look into the contributing guide, but add the environment automatically. The error looks like this:

Screenshot 2022-02-03 at 13 18 45

PR #5014, opened 2022-02-17
fix(insights): send view events after rendering

### LINKED ISSUE #5013: Sending events to Insights middleware too early in connectHit
## 🔍 Bug reproduction

1. Go to https://codesandbox.io/s/laughing-kare-d7e0bo?file=/src/app.js
2. Search for `e`
3. Check the console log
4. When `onEvent` of Insights middleware is triggered, DOM is not updated yet, while `hits` passed from `connectHits` are up-to-date.

**Live reproduction:**

https://codesandbox.io/s/laughing-kare-d7e0bo?file=/src/app.js

## 💭 Expected behavior

https://github.com/algolia/instantsearch.js/blob/9209954594b048e9db3a91490002ae75522bceb2/src/connectors/hits/connectHits.ts#L107

↑ When you look at the code, the connector sends event before rendering. Although it passes the correct `hits` along with the event, but the DOM isn't up-to-date.

I think it can simply be like the following.

```diff
-       renderState.sendEvent('view', renderState.hits);

        renderFn(
          {
            ...renderState,
            instantSearchInstance: renderOptions.instantSearchInstance,
          },
          false
        );

+       renderState.sendEvent('view', renderState.hits);
```

https://codesandbox.io/s/objective-haze-25e8fi?file=/src/app.js

↑ this is a modified version, where `onEvent` is called after rendering.

## Context

This was discovered in the context of Google Tag Manager. On the GTM doc, we guide to use Insights middleware as "Hits Viewed" hook. When it's triggered, it notifies GTM so that it can track events. However, at that time, the DOM isn't up-to-date, so it went wrong.

eunjae-lee:
**Summary**

This PR fixes the order of sending view events and rendering in connectors.

fixes #5013

**Result**

At the time of view events being sent, the DOM is already updated.

### DIFF
--- a/src/connectors/hits/__tests__/connectHits-test.ts
+++ b/src/connectors/hits/__tests__/connectHits-test.ts
@@ -781,6 +781,55 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
           });
         });
 
+        it('sends view event after hits are rendered', () => {
+          const renderFn = jest.fn();
+          const makeWidget = connectHits(renderFn);
+          const widget = makeWidget({});
+
+          const helper = algoliasearchHelper(createSearchClient(), '', {});
+          helper.search = jest.fn();
+
+          const initOptions = createInitOptions({
+            helper,
+            state: helper.state,
+          });
+          const instantSearchInstance = initOptions.instantSearchInstance;
+          instantSearchInstance.sendEventToInsights = jest.fn();
+          widget.init!(initOptions);
+
+          const hits = [
+            {
+              __position: 0,
+              __queryID: 'test-query-id',
+              fake: 'data',
+              objectID: '1',
+            },
+          ];
+
+          const results = new SearchResults(helper.state, [
+            createSingleSearchResponse({ hits }),
+          ]);
+
+          widget.render!(
+            createRenderOptions({
+              results,
+              state: helper.state,
+              helper,
+            })
+          );
+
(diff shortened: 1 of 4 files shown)


src/connectors/hits/__tests__/connectHits-test.ts
```diff
@@ -781,6 +781,55 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
...
+          const lastInvocationCallOrder = (mockFn: jest.Mock): number =>
+            mockFn.mock.invocationCallOrder[
+              mockFn.mock.invocationCallOrder.length - 1
+            ];
+
+          expect(lastInvocationCallOrder(renderFn)).toBeLessThan(
+            lastInvocationCallOrder(
+              instantSearchInstance.sendEventToInsights as jest.Mock
+            )
+          );
```
eunjae-lee, line 830: you think we need a test for ordering like this? @Haroenv

Haroenv: I wonder if there's a way to write this kind of test generically to make sure view event is *always* after render. What do you think?

Haroenv (commented):
the approach works well. As it's easy to forget, the test makes sense to me. 

We need to also do this for infinite hits (maybe geo, autocomplete?)

Haroenv (approved):

dhayab (approved):

PR #5023, opened 2022-03-03
feat(infiniteHits): avoid caching artificial results

Haroenv:
**Summary**

Only cache the hits in infiniteHits if it's a "real" set of results

**Result**

in React InstantSearch Hooks, where we call getWidgetRenderState with fake results (empty hits), we need to ensure that those don't get cached.

> An alternative is to skip caching for `transformedHits.length === 0`, as those should not have an effect on the page, but this could have unintended effects

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "@types/google.maps": "^3.45.3",
     "@types/hogan.js": "^3.0.0",
     "@types/qs": "^6.5.3",
-    "algoliasearch-helper": "^3.7.0",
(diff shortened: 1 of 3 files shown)


sarahdayan:
I'd rather skip caching when unnecessary rather than add an internal flag. This way we don't introduce a new stateful concept and solve the problem logically. wdyt?

Haroenv:
This flag can be useful for suspense / avoiding to show the results later too, if we want to invest in it

francoischalifour (approved):

PR #5028, opened 2022-03-19
feat: prototype HTML templating

sarahdayan:
⚠ This PR is **NOT** meant to be either reviewed or merged. **Please ignore it.** ⚠

This PR was opened so that CodeSandbox picks it up and we can generate an online example based on the built dependencies. It will be closed once https://github.com/algolia/instantsearch-rfcs/pull/70 is closed.

### DIFF
--- a/stories/hits.stories.ts
+++ b/stories/hits.stories.ts
@@ -258,4 +258,34 @@ storiesOf('Results/Hits', module)
         insightsClient: fakeInsightsClient,
       }
     )
+  )
+  .add(
+    'with html function',
+    withHits(({ search, container, instantsearch }) => {
+      search.addWidgets([
+        instantsearch.widgets.hits({
+          container,
+          templates: {
+            // @ts-ignore
+            item(hit, _, html) {
+              return html`<img height="50" src="${hit.image}" />
+                <h2 style="font-size: 1rem">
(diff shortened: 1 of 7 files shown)


stories/hits.stories.ts
```diff
@@ -258,4 +258,34 @@ storiesOf('Results/Hits', module)
...
+          container,
+          templates: {
+            // @ts-ignore
+            item(hit, _, html) {
+              return html`<img height="50" src="${hit.image}" />
+                <h2 style="font-size: 1rem">
+                  ${instantsearch.highlight({
+                    attribute: 'name',
+                    hit,
+                    jsx: true,
```
Haroenv, line 276: there should be another way to differentiate between the two methods. Maybe `instantsearch.js/es/components` or something (to differentiate it from a hogan helper) 🤔

sarahdayan: This PR [isn't meant for review](https://github.com/algolia/instantsearch.js/pull/5028#issue-1174247571) 🙂

sarahdayan: We can discuss this further in an RFC, but I think the final implementation will depend on whether we want to introduce this in v4 or v5. Obviously we don't want a flag like that, and we likely don't want to make the package heavier with things some users don't need.

PR #5034, opened 2022-03-25
fix(currentRefinements): more detailed type for item.type

Haroenv:
**Summary**

Adds a more detailed type for the `item.type`, as it's a specific set of strings, not any string

**Result**

slightly more precise type, only impact for people using the connector. I wouldn't class this as a breaking change, even if it could in theory break someone's build.


FabienMotte (approved):

dhayab (approved):
Fine with me, users would already have been warned in the console if they used an unlisted value.

francoischalifour (approved):

PR #5039, opened 2022-03-30
feat(refinements): escape facet values starting with "-"

Haroenv:
**Summary**

If a facet starts with `-`, the filter will be equivalent to "NOT VALUE" instead of "-VALUE". This is fixed by using the "value" key of each facet instead of the name, via https://github.com/algolia/algoliasearch-helper-js/pull/889

This refactored some, but not all of the tests (as all would be too much work) to follow a more realistic code path.

FX-1135

**Result**

facet values that start with a "-" will no longer automatically negate the rest of that facet, but will instead filter on the literal value.


package.json
```diff
@@ -56,7 +56,7 @@
     "@types/google.maps": "^3.45.3",
     "@types/hogan.js": "^3.0.0",
     "@types/qs": "^6.5.3",
-    "algoliasearch-helper": "^3.7.4",
+    "algoliasearch-helper": "https://pkg.csb.dev/algolia/algoliasearch-helper-js/commit/c8b61403/algoliasearch-helper",
```
Haroenv, line 59: PR shouldn't be merged before this is released on the helper

src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
```diff
@@ -331,7 +331,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
       const makeWidget = connectBreadcrumb(render);
       const helper = algoliasearchHelper(createSearchClient(), '');
       const widget = makeWidget({
-        attributes: ['category', 'sub_category'],
+        attributes: ['category', 'subCategory'],
```
Haroenv, line 334: the result + configuration wasn't in sync, but it didn't break any of the tests

package.json
Haroenv: see https://github.com/algolia/algoliasearch-helper-js/pull/889

src/lib/utils/escape-highlight.ts
```diff
@@ -72,6 +72,8 @@ export function escapeHits<THit extends Hit>(
 export function escapeFacets(facetHits: FacetHit[]): FacetHit[] {
   return facetHits.map((h) => ({
     ...h,
+    // TODO: write a test for this and decide if it's the right place
+    value: h.value.replace(/^-/, '\\-'),
```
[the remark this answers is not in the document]
Haroenv: This is done in the helper in https://github.com/algolia/algoliasearch-helper-js/pull/900. Will be reflected once released on the helper side

sarahdayan (approved):
Looks solide (pending decision on the attribute's final name).

FabienMotte (approved):
Nice work! 👏

sarahdayan:
@Haroenv Should it be `escapedValue` now?

sarahdayan (approved):

PR #5041, opened 2022-04-08
fix(currentRefinements): correctly show and allow for refining escaped values

Haroenv:
**Summary**

Escaped values, like `-10` in toggleRefinement or `"-anything"` in refinementList/menu/hierarchicalMenu show up correctly (unescaped/non-escaped), however clicking on the refinement item didn't do anything, as it was also trying to remove the non-existing non-escaped value.

**Result**

This is fixed by introducing an escaped and non-escaped version of each relevant filter type and using that for the actual value of the refinement

I've also slightly refactored getRefinements and its tests, as they still showed some pre-jest signs.

You can test this with any index including "negative" refinements


.storybook/decorators/withHits.ts
```diff
@@ -26,94 +26,98 @@ type SearchOptions = Omit<
...
-
-  const urlLogger = action('Routing state');
-  const search = instantsearch({
-    indexName,
-    searchClient: algoliasearch(appId, apiKey),
-    routing: {
-      router: {
-        write: (routeState: object) => {
-          urlLogger(JSON.stringify(routeState, null, 2));
+export const withHits =
```
Haroenv, line 29: this file just hasn't been reformatted with prettier in a while I think, not sure how that happened, but only change is moving searchClient outside so it can be monkeypatched easier for debugging

Haroenv:
lint is failing, as it requires https://github.com/algolia/algoliasearch-helper-js/pull/903, please update when you merge and release that :)

src/widgets/toggle-refinement/__tests__/toggle-refinement-test.ts
```diff
@@ -38,499 +38,347 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/toggle-refi
...
+                  })
+                ),
+              });
+            },
+          }),
         });
-      });
-    });
-
-    describe('custom off value', () => {
```
Haroenv, line 459: these tests still exist in the connector, but don't need to exist in the widget, IMO

.storybook/decorators/withHits.ts
dhayab: With whitespace diff off its more manageable to read 👍 .

dhayab (approved):

FabienMotte (approved):

PR #5045, opened 2022-04-22
fix(routing): prevent writing the same URL twice

Haroenv:
**Summary**

We've given this from time to time as workarounds for customer issues, where due to state desync or other external influence the URL is already changed to the one InstantSearch would navigate to.

The specific state I test and am fixing here however is with `stateMapping` (described in https://github.com/algolia/instantsearch.js/discussions/5044) removing a piece of the state that still changes, and thus would cause duplicate writes of the same URL before this PR.

**Result**

The same URL will not be written twice subsequently

I had to change the test for replaceState a bit, as the cleaned URL being the same as the replaceState navigated URL being the same was hiding the issue we have that test for.

### DIFF
--- a/src/lib/__tests__/routing/spa-replace-state-test.ts
+++ b/src/lib/__tests__/routing/spa-replace-state-test.ts
@@ -22,8 +22,8 @@ describe('routing using `replaceState`', () => {
     // 2. Refine: '/?indexName[query]=Apple'
     // 3. Dispose: does not yet write
     // 4. Route change (with `replaceState`): '/about?external=true', replaces state 2
-    // 3. Dispose: writes '/about' (this is a bug, and should be fixed when we have a way to prevent it)
-    // 5. Back: '/about?external=true'
+    // 5. Dispose: writes '/about' (this is a bug, and should be fixed when we have a way to prevent it)
+    // 6. Back: '/about?external=true'
 
     const pushState = jest.spyOn(window.history, 'pushState');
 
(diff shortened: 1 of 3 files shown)


src/lib/__tests__/routing/spa-replace-state-test.ts
```diff
@@ -20,9 +20,10 @@ describe('routing using `replaceState`', () => {
     // -- Flow
     // 1. Initial: '/'
     // 2. Refine: '/?indexName[query]=Apple'
-    // 3. Dispose: '/about'
-    // 4. Route change (with `replaceState`): '/about'
-    // 5. Back: '/about'
+    // 3. Dispose: does not yet write
+    // 4. Route change (with `replaceState`): '/about?external=true', replaces state 2
+    // 3. Dispose: writes '/about' (this is a bug, and should be fixed when we have a way to prevent it)
+    // 5. Back: '/about?external=true'
```
FabienMotte, line 26:
```suggestion
    // 5. Dispose: writes '/about' (this is a bug, and should be fixed when we have a way to prevent it)
    // 6. Back: '/about?external=true'
```

src/lib/routers/history.ts
```diff
@@ -220,6 +210,30 @@ class BrowserHistory<TRouteState> implements Router<TRouteState> {
...
+      // OR
+      // - the last write was from InstantSearch.js
+      // (unlike a SPA, where it would have last written)
+      const lastPushWasByISAfterDispose = !(
+        this.isDisposed &&
+        this.latestAcknowledgedHistory !== window.history.length
+      );
+
+      return (
+        // When the the last state change was through popstate, the IS.js state changes,
```
FabienMotte, line 227:
```suggestion
        // When the last state change was through popState, the IS.js state changes,
```

FabienMotte (changes requested):
Nice! Thanks for refactoring the `write` condition into a proper method

FabienMotte (approved):

PR #5055, opened 2022-05-28
Use arrow functions in connector types to fix `@typescript-eslint/no-unbound-method` errors

zacharyliu:
**Summary**

Currently, the `RenderState` types for connectors inconsistently declare functions as either arrow functions or regular functions:

https://github.com/algolia/instantsearch.js/blob/3faca014aad08145c3b4cc66a5e841da3a0f64b8/src/connectors/search-box/connectSearchBox.ts#L41
https://github.com/algolia/instantsearch.js/blob/74d98f1fcd0d3560c63e4e4dc6dfaa848eb7f5bb/src/connectors/refinement-list/connectRefinementList.ts#L111

In user code, the latter leads to spurious lint errors from the [`@typescript-eslint/no-unbound-method` lint rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/unbound-method.md) when trying to assign the property to a standalone variable.

For example, destructuring the return value of the `useRefinementList` hook [as shown in the Algolia docs](https://www.algolia.com/doc/api-reference/widgets/refinement-list/react-hooks/#hook) results in this lint error:

image

I did a rudimentary regex search of the connector functions to find other affected type declarations, but may have missed some.

**Result**

Lint error goes away after building and `link`ing this branch.

### DIFF
--- a/src/connectors/current-refinements/connectCurrentRefinements.ts
+++ b/src/connectors/current-refinements/connectCurrentRefinements.ts
@@ -90,7 +90,7 @@ export type CurrentRefinementsConnectorParamsItem = {
   /**
    * Removes the given refinement and triggers a new search.
    */
-  refine(refinement: CurrentRefinementsConnectorParamsRefinement): void;
+  refine: (refinement: CurrentRefinementsConnectorParamsRefinement) => void;
 };
 
 export type CurrentRefinementsConnectorParams = {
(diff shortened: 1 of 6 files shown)


Haroenv:
Is it expected that you see eslint errors for library code? Shouldn't node_modules be ignored?

Haroenv:
If it makes sense to enforce a specific style, we should enable https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/method-signature-style.md

Haroenv:
I'm happy to accept a PR that changes all to arrow functions, and then also add the no-unbound-method rule here. Would you be interested in continuing?

Haroenv:
I'll close this for now, but happy to collaborate if you want to revisit this

PR #5058, opened 2022-06-07
fix(insights): don't send view event if search is stalled

Haroenv:
**Summary**

The insights view event should only be sent if the hits are different from the ones previously shown, not if the widget renders because of a stalled render.

Alternatively we could memoize-one the call to sendEventToInsights call (https://github.com/algolia/instantsearch.js/blob/99f6fe1dc51e4815e5b9efcfb30e3e2f3127e763/src/lib/utils/createSendEventForHits.ts#L168) or at the beginning of sendEventToInsights in the middleware (https://github.com/algolia/instantsearch.js/blob/55313e4ea4105b777f3f102e9f48a7e440496d25/src/middlewares/createInsightsMiddleware.ts#L144), but that could have negative side-effects, when for example the same hit is clicked twice, it should still send two events

**Result**

in a stalled search there should only be one view event sent.

A sandbox that's fixed with this PR is https://codesandbox.io/s/dreamy-cherry-h8sws3?file=/src/app.js

### DIFF
--- a/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
+++ b/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
@@ -8,11 +8,16 @@ import {
   createDisposeOptions,
 } from '../../../../test/mock/createWidget';
 import { createSearchClient } from '../../../../test/mock/createSearchClient';
-import { createSingleSearchResponse } from '../../../../test/mock/createAPIResponse';
+import {
+  createMultiSearchResponse,
+  createSingleSearchResponse,
+} from '../../../../test/mock/createAPIResponse';
 import type { AutocompleteRenderState } from '../connectAutocomplete';
 import connectAutocomplete from '../connectAutocomplete';
 import { TAG_PLACEHOLDER } from '../../../lib/utils';
 import type { SearchClient } from '../../../types';
+import { wait } from '../../../../test/utils/wait';
+import instantsearch from '../../../index.es';
 
 describe('connectAutocomplete', () => {
   const getInitializedWidget = (config = {}) => {
@@ -811,6 +816,58 @@ search.addWidgets([
         },
         widgetType: 'ais.autocomplete',
       });
+    });
+
+    it('does not send view event when hits are stalled rendered', async () => {
+      const renderFn = jest.fn();
+      const makeWidget = connectAutocomplete(renderFn);
+      const widget = makeWidget({});
+
+      const hits = [
+        {
+          objectID: '1',
+          fake: 'data',
+          __queryID: 'test-query-id',
+          __position: 0,
+        },
+        {
+          objectID: '2',
+          sample: 'infos',
+          __queryID: 'test-query-id',
+          __position: 1,
+        },
+      ];
+
+      const instantSearchInstance = instantsearch({
+        searchClient: createSearchClient({
+          search() {
+            return Promise.resolve(
+              createMultiSearchResponse(createSingleSearchResponse({ hits }))
+            );
+          },
+        }),
+        indexName: 'indexName',
+      });
+      instantSearchInstance.sendEventToInsights = jest.fn();
+      instantSearchInstance.start();
+
+      instantSearchInstance.addWidgets([widget]);
+
+      await wait(0);
+
+      expect(instantSearchInstance.sendEventToInsights).toHaveBeenCalledTimes(
+        1
+      );
+
+      instantSearchInstance.mainHelper!.hasPendingRequests = () => true;
+      instantSearchInstance._isSearchStalled = true;
+      instantSearchInstance.scheduleRender();
+
+      await wait(0);
+
+      expect(instantSearchInstance.sendEventToInsights).toHaveBeenCalledTimes(
+        1
+      );
     });
 
     it('sends click event', () => {
--- a/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
+++ b/src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
@@ -838,14 +838,17 @@ search.addWidgets([
         },
       ];
 
+      const searchClient = createSearchClient({
+        search() {
+          return Promise.resolve(
+            createMultiSearchResponse(createSingleSearchResponse({ hits }))
+          );
+        },
+      });
+
       const instantSearchInstance = instantsearch({
-        searchClient: createSearchClient({
-          search() {
-            return Promise.resolve(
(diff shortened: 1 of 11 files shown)


src/connectors/hits/connectHits.ts
```diff
@@ -113,7 +113,9 @@ const connectHits: HitsConnector = function connectHits(
           false
         );
 
-        renderState.sendEvent('view', renderState.hits);
+        if (!renderOptions.searchMetadata.isSearchStalled) {
+          renderState.sendEvent('view', renderState.hits);
+        }
```
francoischalifour, line 118: Could we rather move all the conditionals to [`createSendEventForHits()`](https://github.com/algolia/instantsearch.js/blob/0a517609de103eef4f8edfefe6e28a1d79a14209/src/lib/utils/createSendEventForHits.ts#L150)?

Haroenv: Sure, that's possible

Haroenv: Would it make sense here? I'm not convinced that a function responsible for transformation should also take global state in account: 

https://github.com/algolia/instantsearch.js/blob/99f6fe1dc51e4815e5b9efcfb30e3e2f3127e763/src/lib/utils/createSendEventForHits.ts#L91-L104

francoischalifour: I thought about having a short circuit before this line: https://github.com/algolia/instantsearch.js/blob/99f6fe1dc51e4815e5b9efcfb30e3e2f3127e763/src/lib/utils/createSendEventForHits.ts#L159

But I don't have the complete knowledge of this part.

> Also note that the sendEvent function can be called by end users. I'm not sure if we then also should prevent forwarding the event if the search is currently stalled

I think the question is: should it send an event when the search is stalled?

I don't think it should because (IIUC):

1. a render displays hits: send view event
2. user searches
3. a render displays hits again because search is stalled: ~send view event~

And so I don't think this (3) makes sense.

Haroenv: if we do this condition only for `view` event I agree, but if it should be all events that get shortcutted I don't think so, because you could be in a stalled state, but click on a buy button, that's a genuine action which should count right?

sarahdayan: I think it makes sense to have guards per event type centralized in `createSendEventForHits`. For now we can have only one for `view` since we have knowledge about when it should and should not be sent, we don't need to guard for all events.

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -813,6 +818,58 @@ search.addWidgets([
...
+
+      await wait(0);
+
+      expect(instantSearchInstance.sendEventToInsights).toHaveBeenCalledTimes(
+        1
+      );
+
+      instantSearchInstance.mainHelper!.hasPendingRequests = () => true;
+      instantSearchInstance._isSearchStalled = true;
+      instantSearchInstance.scheduleRender();
```
francoischalifour, line 864: Could we make a higher-level test that makes use of [`stalledSearchDelay`](https://www.algolia.com/doc/api-reference/widgets/instantsearch/js/#widget-param-stalledsearchdelay) and a search client mock that resolves after this time?

The test would be easier to understand, and more resilient over time.

francoischalifour: Yes, that could also work.

What bothers me in this test is that it interacts with deep internals and doesn't completely reflect the user scenario.

Haroenv: solved with a "dead" search client in https://github.com/algolia/instantsearch.js/pull/5058/commits/7131b4a1b67ee37b0317e4531fdfec40ee121793

Thanks for the idea!

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -859,11 +859,10 @@ search.addWidgets([
...
 
-      instantSearchInstance.mainHelper!.hasPendingRequests = () => true;
-      instantSearchInstance._isSearchStalled = true;
-      instantSearchInstance.scheduleRender();
-
-      await wait(0);
+      // this client never resolves, thus search is stalled
+      instantSearchInstance.client.search = () => new Promise(() => {});
+      instantSearchInstance.scheduleSearch();
+      await wait(400);
```
francoischalifour, line 865: Are these fake timers? Otherwise we can decrease this amount with `stalledSearchDelay`.

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
```diff
@@ -859,11 +859,10 @@ search.addWidgets([
         1
       );
 
-      instantSearchInstance.mainHelper!.hasPendingRequests = () => true;
-      instantSearchInstance._isSearchStalled = true;
-      instantSearchInstance.scheduleRender();
-
-      await wait(0);
+      // this client never resolves, thus search is stalled
+      instantSearchInstance.client.search = () => new Promise(() => {});
```
francoischalifour, line 863: I think this would work and wouldn't require to access the `instantSearchInstance.client.search` internal.

```js
const searchRef = {
  current: () => {
    return Promise.resolve(
      createMultiSearchResponse(createSingleSearchResponse({ hits }))
    );
  },
};

const instantSearchInstance = instantsearch({
  searchClient: createSearchClient({
    search() {
      return searchRef.current();
    },
  }),
  indexName: 'indexName',
});

// ...

searchRef.current = () => new Promise(() => {});
```

src/connectors/autocomplete/__tests__/connectAutocomplete-test.ts
Haroenv: had to add a margin, probably due to timers being "minimal", and them allowing to overtake eachother

francoischalifour (approved):

src/lib/utils/createSendEventForHits.ts
```diff
@@ -184,6 +190,7 @@ export function createBindEventForHits({
       index,
       methodName: 'bindEvent',
       args,
+      isSearchStalled: false,
```
Haroenv, line 193: we actually should use the isSearchStalled here, not just false

Haroenv, line 193: this is false, because if the search is stalled, you still want to bind events (for later)

PR #5060, opened 2022-06-08
fix(types): make all usages of UiState in InstantSearch generic

Haroenv:
**Summary**

This allows someone with custom widgets to still use setUiState etc. without type errors

**Result**

all places `UiState` is used in the InstantSearch instance are now generic and use TUiState.

There are still places left in individual widgets, middleware which aren't generic, but that can be solved in a different PR

A sandbox showing the fixed problem: https://codesandbox.io/s/instantsearch-js-forked-m8nm7s?file=/src/app.ts:1152-1262

### DIFF
--- a/src/lib/__tests__/InstantSearch-test.tsx
+++ b/src/lib/__tests__/InstantSearch-test.tsx
@@ -21,7 +21,7 @@ import {
 } from '../../../test/mock/createWidget';
 import { castToJestMock } from '../../../test/utils/castToJestMock';
 import type { IndexWidget } from '../../widgets/index/index';
-import type { Widget } from '../../types';
+import type { UiState, Widget } from '../../types';
 import type {
   PaginationConnectorParams,
   PaginationWidgetDescription,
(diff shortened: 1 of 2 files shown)


src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -2038,7 +2038,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
 
   test('with function form sets indices state', async () => {
     const searchClient = createSearchClient();
-    const search = new InstantSearch({
+    const search = new InstantSearch<UiState>({
```
Haroenv, line 2041: This is needed because initialUiState seems to "fill in" the generic. That makes sense, but in this case it makes things required that aren't actually required. Not sure how to solve that though! This already was the case before the PR, but became visible due to usage of TUiState in setUiState

Haroenv: https://github.com/algolia/instantsearch.js/pull/5061

FabienMotte (approved):

PR #5061, opened 2022-06-08
fix(types): avoid inferring UiState type from initialUiState

Haroenv:
**Summary**

if you pass initialUiState, that should be asserted by the generic UiState, not automatically accepted. Otherwise you get use cases like the "with function form sets indices state" test where setUiState thinks query is required, just because it has the default given

The way this works is by using `NoInfer` from https://github.com/Microsoft/TypeScript/issues/14829#issuecomment-504042546 which seems to be quite relied upon

This is a follow-up on #5060

**Result**

ui state is not inferred from initialUiState, but rather enforced by it. This means however that custom widgets are no longer "automatically detected from initialUiState", which I don't think is a breaking change, as it's only a type change

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -41,6 +41,11 @@ function defaultCreateURL() {
   return '#';
 }
 
+// this purposely breaks typescript's type inference to ensure it's not used
+// as it's used for a default parameter for example
+// source: https://github.com/Microsoft/TypeScript/issues/14829#issuecomment-504042546
+type NoInfer<T> = [T][T extends any ? 0 : never];
+
 /**
  * Global options for an InstantSearch instance.
  */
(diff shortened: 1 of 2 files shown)


src/lib/InstantSearch.ts
```diff
@@ -41,6 +41,11 @@ function defaultCreateURL() {
   return '#';
 }
 
+// this purposely breaks typescript's type inference to ensure it's not used
+// as it's used for a default parameter for example
+// source: https://github.com/Microsoft/TypeScript/issues/14829#issuecomment-504042546
+type NoInfer<T> = [T][T extends any ? 0 : never];
```
francoischalifour, line 47: Did you see this simplified version https://github.com/Microsoft/TypeScript/issues/14829#issuecomment-980401795 ?

Haroenv:
I measured it by looking at the 20/30 PR references in that typescript issue, which used this method, but also the others in the thread that don't work for this specific use case

FabienMotte (approved):

PR #5073, opened 2022-07-04
fix(ratingMenu): don't warn if results are artificial

Haroenv:
**Summary**

If the results are artificial (react instantsearch hooks), `maxValuesPerFacet` can evaluate to 0, as it's not necessarily set by another widget, thus isn't in the result.

Before this fix there's a useless warning if you use useConnector(connectRatingMenu)

**Result**

using useRatingMenu in React InstantSearch Hooks doesn't warn on first use


Haroenv:
I did a search and there's no other cases where there's a warning based on the results (the only other one was in infinite hits, and already handled)

dhayab (approved):

PR #5075, opened 2022-07-07
fix(errors): rethrow error as error if it's an object

Haroenv:
**Summary**

If you're using the v4 client, the errors (unfortunately) are fake errors, see https://github.com/algolia/algoliasearch-client-javascript/blob/3aec216a9704f818b4faaa8b97d988ec2794f9da/packages/transporter/src/errors/createRetryError.ts for example. When those errors get thrown, you get `"Uncaught, unspecified \"error\" event. ([object Object])"` still.

cfr errors like https://github.com/algolia/react-instantsearch/issues/3534 which have more clear errors (usually wrong api keys, network issues)

**Result**

no more "object Object" errors for search client errors

### DIFF
--- a/src/lib/InstantSearch.ts
+++ b/src/lib/InstantSearch.ts
@@ -493,10 +493,10 @@ See ${createDocumentationLink({
       if (!(error instanceof Error)) {
         // typescript lies here, error is in some cases { name: string, message: string }
         const err = error as Record<string, any>;
-        error = new Error(err.message);
-        Object.keys(err).forEach((key) => {
-          (error as any)[key] = err[key];
-        });
+        error = Object.keys(err).reduce((acc, key) => {
+          (acc as any)[key] = err[key];
+          return acc;
+        }, new Error(err.message));
       }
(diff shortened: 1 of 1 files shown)


src/lib/InstantSearch.ts
```diff
@@ -490,6 +490,14 @@ See ${createDocumentationLink({
     // Only the "main" Helper emits the `error` event vs the one for `search`
     // and `results` that are also emitted on the derived one.
     mainHelper.on('error', ({ error }) => {
+      if (!(error instanceof Error)) {
+        // typescript lies here, error is in some cases { name: string, message: string }
+        const err = error as Record<string, any>;
+        error = new Error(err.message);
+        Object.keys(err).forEach((key) => {
+          (error as any)[key] = err[key];
+        });
```
dhayab, line 499: (nitpicky) alternative:

```suggestion
        error = Object.entries(error).reduce((acc, [key, value]) => {
          acc[key] = value;
          return acc;
        }, new Error(error.message));
```

dhayab (approved):

PR #5079, opened 2022-07-25
chore: introduce VDOM highlighting components

sarahdayan:
## Summary

This PR introduces VDOM highlighting components in InstantSearch.js ([FX-1512](https://algolia.atlassian.net/browse/FX-1512)). They will be exposed via the Template API later on.

## Review

These new components depend on a common implementation from [`ui-components/highlight-vdom`](https://github.com/algolia/ui-components/tree/master/packages/highlight-vdom) and are wrapped  internally to add the InstantSearch-specific class names. In the future, we may want to move those in a separate `instantsearch-vdom` package to avoid repeating the same code in every VDOM-based InstantSearch flavor.

The final components (with the logic to parse hits and pass parts to the UI component) which will later be exposed in templates live in `src/helpers/components/`, near their existing string-based counterparts.

For now, this addition shouldn't impact the library, and doesn't require a release. It can be merged without waiting for other PRs.


src/helpers/components/Highlight.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export function Highlight<THit extends Hit<BaseHit>>({
+  hit,
+  attribute,
+  cssClasses,
+  ...props
+}: HighlightProps<THit>) {
+  const property =
+    getPropertyByPath(hit._highlightResult, attribute as string) || [];
+  const properties = Array.isArray(property) ? property : [property];
```
Haroenv, line 30: not sure if now is the moment, but on Angular InstantSearch (I think) we fall back to the attribute value instead of the highlighted one, which means that if something is sometimes a number, you don't need to add a conditional to the code. Maybe this could be done with an option, not by default? maybe later?

Haroenv (approved):
looks well-researched, and as far as I can tell this will work

src/helpers/components/Highlight.tsx
sarahdayan: Yes I saw the comment in the existing components and hesitated to implement because I didn't know how much we've discussed and agreed to this pattern and considered alternatives. Also, this would be a different behavior from the existing components (although not breaking) so I'm not sure we should do it now.

PR #5080, opened 2022-07-25
chore: introduce `html` module

sarahdayan:
## Summary

This PR introduces an `html` module in InstantSearch.js ([FX-1512](https://algolia.atlassian.net/browse/FX-1512)).

It will be exposed via the Template API later on.

## Review

For now, this addition shouldn't impact the library, and doesn't require a release. It can be merged without waiting for other PRs.


Haroenv (approved):

Haroenv:
as we're already using preact, and there's a htm/preact entry point, are they the same / is there a reason to have this file separately? maybe a bundle size impact?

src/lib/html.ts
```diff
@@ -0,0 +1,4 @@
+import htm from 'htm';
+import { createElement } from 'preact';
+
+export const html = htm.bind(createElement);
```
francoischalifour, line 4: There's a highly [optimized import for htm + preact](https://github.com/developit/htm#usage):

```js
import { html } from 'htm/preact';
```

Haroenv: there's actually no special code in that: https://github.com/developit/htm/blob/master/src/integrations/preact/index.mjs#L14-L19

but I agree with the question of whether it's needed

sarahdayan: Yes indeed, and there's no impact on bundle size from what I'm seeing. Also, the library doesn't bundle Preact so there shouldn't be conflicts. I'll remove this module and only keep the dependency.

I don't know how this version is more optimized though, because the [implementation is the same](https://github.com/developit/htm/blob/master/src/integrations/preact/index.mjs). But anyways 🤷‍♀️

sarahdayan: Ah ben I guess we agree 😂

package.json
```diff
@@ -62,6 +62,7 @@
     "algoliasearch-helper": "^3.10.0",
     "classnames": "^2.2.5",
     "hogan.js": "^3.0.2",
+    "htm": "^3.1.1",
```
francoischalifour, line 65: We could loosen the dep range so that the risk of getting duplicated is lower?

```suggestion
    "htm": "^3.0.0",
```

sarahdayan: Good idea! Addressed in 610ef47b.

francoischalifour (approved):

PR #5081, opened 2022-07-25
feat(html): introduce `html` templating

sarahdayan:
## Summary

This PR introduces `html` templating in InstantSearch.js.

This allows InstantSearch.js users to pass templates as HTML strings safely, with easier access to exposed data, and with an easier API to set up events.

[**Try this live**](https://codesandbox.io/s/instantsearch-js-forked-xfoouv?file=/src/app.js) | [**Read RFC**](https://github.com/algolia/instantsearch-rfcs/blob/master/accepted/html-templating-instantsearch.md)

>The current sandbox needs to import `@babel/runtime` for now, this will be fixed.

```js
instantsearch.widgets.hits({
  // ...
  templates: {
    item(hit, { html, components, sendEvent }) {
      return html`<article>
        <h2>${components.Highlight({ hit, attribute: 'name' })}</h2>
        <p>${components.Snippet({ hit, attribute: 'description' })}</p>
        <button
          onClick="${() => sendEvent('conversion', hit, 'Product Added')}"
        >
          Add to cart
        </button>
      </article>`;
    },
  },
});
```

**Note:** For now, the feature doesn't work in [`pagination`](https://www.algolia.com/doc/api-reference/widgets/pagination/js/) and [`geoSearch`](https://www.algolia.com/doc/api-reference/widgets/geo-search/js/). `pagination` will be handled separately, and `geoSearch` will be tackled later (at most when we deprecate Hogan.js) depending on complexity and demand.

## Tests

Testing this feature involves testing templates in ~20 widgets. That's a lot of files, so I did it in a [separate PR](https://github.com/algolia/instantsearch.js/pull/5087).


src/widgets/geo-search/geo-search.ts
```diff
@@ -269,7 +269,7 @@ const geoSearch: GeoSearchWidget = (widgetParams) => {
         templateKey: 'HTMLMarker',
         templates,
         data: item,
-      }),
+      }) as string,
```
sarahdayan, line 272: For now, `geoSearch` templates will only support strings. We'll assess separately how to bring `html` to this widget.

src/widgets/hits/__tests__/defaultTemplates-test.ts
```diff
@@ -28,7 +29,7 @@ describe('hits defaultTemplates', () => {
 
     expect(
       typeof defaultTemplates.item === 'function' &&
-        defaultTemplates.item(item, () => '')
+        defaultTemplates.item(item, {} as TemplateParams)
```
sarahdayan, line 32: This cast is necessary to make the test pass in its current state. We might want to drop it anyway later on and test default templates at the widget level instead of unit testing this implementation detail.

src/widgets/infinite-hits/__tests__/defaultTemplates-test.ts
```diff
@@ -28,7 +29,7 @@ describe('hits defaultTemplates', () => {
 
     expect(
       typeof defaultTemplates.item === 'function' &&
-        defaultTemplates.item(item, () => '')
+        defaultTemplates.item(item, {} as TemplateParams)
```
sarahdayan, line 32: This cast is necessary to make the test pass in its current state. We might want to drop it anyway later on and test default templates at the widget level instead of unit testing this implementation detail.

package.json
```diff
@@ -157,7 +157,7 @@
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "70.25 kB"
+      "maxSize": "71.75 kB"
```
sarahdayan, line 160: This is the cost of `htm` + the new code to support it. When we remove Hogan.js later on, this should significantly decrease.

src/types/templates.ts
```diff
@@ -1,12 +1,31 @@
...
+
+export type TemplateParams = BindEventForHits & {
+  html: typeof html;
+  components: {
+    Highlight: typeof Highlight;
+    ReverseHighlight: typeof ReverseHighlight;
+    Snippet: typeof Snippet;
+    ReverseSnippet: typeof ReverseSnippet;
+  };
+  sendEvent?: SendEventForHits;
```
Haroenv, line 23: should maybe be changed later only, but sendEvent can be for facets too I think (unless we don't expose that?)

src/widgets/geo-search/geo-search.ts
Haroenv: don't we need to warn or error when a jsx is returned then?

Haroenv (commented):
Should the default template of hits and infinite hits be changed here too, or do you want to them all in a separate PR?

src/components/Template/Template.tsx
```diff
@@ -59,6 +62,10 @@ class Template extends Component<TemplateProps> {
       return null;
     }
 
+    if (typeof content !== 'string') {
```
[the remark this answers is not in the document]
sarahdayan: Why do you think this is necessary to be specific? By contract, it should either be JSX or a string. Do you want to further validate for non-TypeScript users?

src/types/templates.ts
sarahdayan: For now we don't pass either `bindEvent` or `sendEvent` to any other component than `<Hits>` and `<InfiniteHits>`. I agree this can be useful for other widgets though, but we can do it separately.

src/widgets/geo-search/geo-search.ts
sarahdayan: Yes, I planned on tackling it separately.

sarahdayan:
>Should the default template of hits and infinite hits be changed here too, or do you want to them all in a separate PR?

Why do you want to chan ge them? They return a string, but they don't use Hogan.js.

src/components/Template/Template.tsx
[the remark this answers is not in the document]
francoischalifour: I agree with @Haroenv. It makes more sense to validate the data that we want to pass, rather than to invalidate the data that we don't want.

src/lib/utils/__tests__/createHelpers-tests.ts
```diff
@@ -34,7 +34,7 @@ describe('insights hogan helper', () => {
     })}>Add to cart</button>
     `;
 
-    expect(output.trim()).toEqual(expected.trim());
+    expect((output as string).trim()).toEqual(expected.trim());
```
francoischalifour, line 37: If TS isn't able to infer the `string` type, we better cast it at the variable declaration line 15.

src/widgets/geo-search/geo-search.ts
francoischalifour: What's different in `geoSearch()` to not support JSX?

dhayab (commented):
The implementation follows the RFC and tests pass. Good for me once the other comments are answered 👍.

src/components/Template/Template.tsx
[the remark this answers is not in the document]
sarahdayan: Addressed in e58b11fd. `VNode` types don't carry much that it makes sense to further check IMO, it will add to the bundle for no gain.

src/components/Template/Template.tsx
```diff
@@ -19,7 +20,8 @@ type TemplateProps = {
   rootProps?: Record<string, any>;
   rootTagName?: keyof JSX.IntrinsicElements;
   templateKey: string;
-  bindEvent?: (...args: any[]) => string;
+  bindEvent?: BindEventForHits;
+  sendEvent?: SendEventForHits;
```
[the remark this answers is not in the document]
sarahdayan: Only in connectors (`connectHierarchicalMenu`, `connectMenu`, `connectNumericMenu`, `connectRefinementList`) as far as I an tell, not in widgets.

src/lib/utils/__tests__/createHelpers-tests.ts
sarahdayan: Addressed in f4ca49a9.

src/components/Template/Template.tsx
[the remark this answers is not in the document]
Haroenv: if we ever add "top-level" templates that would allow override of the refine button, we likely should expose sendEvent for those too

Haroenv (approved):

dhayab (approved):

src/widgets/geo-search/geo-search.ts
sarahdayan: `geoSearch` is implemented in #5092.

PR #5085, opened 2022-08-11
fix(connectNumericMenu + connectRange): stop sending invalid clickedFilters event

dhayab:
JIRA: [FX-1633](https://algolia.atlassian.net/browse/FX-1633)

**Summary**

The [Clicked Filters](https://www.algolia.com/doc/api-reference/api-methods/clicked-filters/) method of Insights does not support numeric filters, but the `connectNumericMenu` and `connectRange` connectors still send this event automatically when refined. This generates errors on the dashboard's events debugger (see [CR-1588](https://algolia.atlassian.net/browse/CR-1588)).

This PR removes this call.

**Result**

The `clickedFilters` event is not sent anymore during the lifecycle of the widget using the underlying `connectNumericMenu` and `connectRange` connectors.


aymeric-giraudet (approved):

dhayab:
@aymeric-giraudet @sarahdayan can I have another review for this PR please? Since last time, I noticed `connectNumericMenu` also sends an invalid insights event so I removed the code there as well.

src/connectors/numeric-menu/__tests__/connectNumericMenu-test.ts
```diff
@@ -1127,66 +1127,6 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/numeric-men
     });
   });
 
-  describe('insights', () => {
```
sarahdayan, line 1130: Could we maybe add a test asserting that no event is sent when adding a facet? This will make the behavior explicit and not look like an omission for a newcomer, or when we look back this part of the code in 6 months.

sarahdayan (commented):
Looks like this code removal dropped the bundle size of ~4KB. We should probably adjust the threshold so that future additions' footprint is properly gauged.

src/connectors/numeric-menu/__tests__/connectNumericMenu-test.ts
dhayab: Good idea, thank you for thinking of this!

sarahdayan (approved):

Haroenv:
Just an update if someone reads this PR again, this is still the case today: <a href="https://www.algolia.com/doc/api-reference/api-methods/clicked-filters/#about-this-method"><q>You can’t send these events for numeric filters.</q></a>

PR #5087, opened 2022-08-17
test(html): test `html` templating

sarahdayan:
## Summary

This PR tests `html` templating in widgets.

Inline snapshots don't work in loops, so we can't use the `test.each` API without using separate snapshots. I reckon both the `test.each` format and the separate snapshots create a lot of mental overhead and bring little benefits, so I went for separate tests even if it means we can't ensure each templating solution produces the same output. This sounds acceptable for now.

I found that some template types were wrong while testing, so I fixed some of them. It will take a more extensive job to verify everything, we should do it separately and use the opportunity to refactor and eliminate the problem altogether.

## Next steps

After discussing it with @Haroenv we agreed that the current test specs for widgets aren't satisfactory anymore because they don't test widgets like a real user would do. Additionally, they're not exhaustive. This PR introduces templating tests in new test specs (using the `.test.{ts,tsx}` suffix instead of the legacy `-test.{ts,tsx}` one) where we'll then be able to progressively migrate existing tests while refactoring them.


package.json
```diff
@@ -63,7 +63,7 @@
     "classnames": "^2.2.5",
     "hogan.js": "^3.0.2",
     "htm": "^3.0.0",
-    "preact": "^10.6.0",
+    "preact": "^10.10.0",
```
sarahdayan, line 66: This fixes a TypeScript overlook to allow `defaultChecked` on `HTMLInputElement`s.

package.json
```diff
@@ -81,7 +81,7 @@
     "@storybook/html": "5.3.9",
     "@storybook/theming": "5.3.9",
     "@testing-library/dom": "7.9.0",
-    "@testing-library/jest-dom": "4.1.0",
+    "@testing-library/jest-dom": "4.2.0",
```
sarahdayan, line 84: This allows using the `toBeChecked` matcher.

.eslintrc.js
```diff
@@ -7,7 +7,8 @@ module.exports = {
     'new-cap': [
       'error',
       {
-        capIsNewExceptionPattern: '(\\.|^)EXPERIMENTAL_.+',
+        capIsNewExceptionPattern:
```
sarahdayan, line 10: This lets us have capitalized components (e.g., as exposed on `hits` and `infiniteHits`).

.eslintrc.js
```diff
@@ -67,6 +68,7 @@ module.exports = {
         'valid-jsdoc': 'off',
         'no-redeclare': 'off',
         '@typescript-eslint/no-redeclare': ['error'],
+        'react/prop-types': 'off',
```
sarahdayan, line 71: There's no point to prop types in TypeScript files.

aymeric-giraudet (approved):

src/widgets/hits/__tests__/hits.test.tsx
```diff
@@ -0,0 +1,572 @@
...
+
+    type CustomHit = { name: string; description: string };
+
+    function createMockedSearchClient(
+      subset: Partial<SearchResponse<CustomHit>> = {}
+    ) {
+      return createSearchClient({
+        search: jest.fn((requests) => {
+          return Promise.resolve(
+            createMultiSearchResponse(
```
dhayab, line 509: Regarding the failing type jobs, it seems to originate from the return type of this method. I understand the underlying issue regarding the expected `TObject` type but could not find how to ensure its compatibility with `Hit<CustomHit>`.

In the context of those test cases, we can remove the type inference with `createMultiSearchResponse<any>()` so it stops bugging us, while still maintaining type safety for each single search response below. Wdyt?

sarahdayan: Yes I agree. The mocks from React InstantSearch Hooks are better typed so instead of trying to fix it here, we can put those in common later on and work arund TypeScript for now.

dhayab (approved):

src/widgets/panel/__tests__/panel.test.tsx
```diff
@@ -0,0 +1,399 @@
...
+              return html`<a href="#"
+                >More brands ${results && `(${results.nbHits} results)`}</a
+              >`;
+            },
+            collapseButtonText({ collapsed }, { html }) {
+              return html`<span>${collapsed ? 'More' : 'Less'}</span>`;
+            },
+          },
+          collapsed: () => true,
+        })(refinementList)({ container, attribute: 'brand' }),
```
Haroenv, line 50: maybe this should be tested with a smaller widget so the snapshot is more focused?

src/widgets/search-box/search-box.tsx
```diff
@@ -28,15 +28,15 @@ export type SearchBoxTemplates = Partial<{
   /**
    * Template used for displaying the submit button. Can accept a function or a Hogan string.
    */
-  submit: Template;
+  submit: Template<{ cssClasses: SearchBoxComponentCSSClasses }>;
```
Haroenv, line 31: would it save any errors if Template required the generic? are there use cases for generic-less template left?

Haroenv (commented):
👍

src/widgets/hierarchical-menu/__tests__/hierarchical-menu.test.tsx
```diff
@@ -0,0 +1,255 @@
...
+            href="#"
+            style="font-weight: bold;"
+          >
+            <span>
+              Video Games
+               (
+              505
+              )
+            </span>
+          </a>
```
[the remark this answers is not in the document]
sarahdayan: Yes, I tried in Storybook and it renders correctly.

It looks like Jest adds whitespace when the content is in a variable. When I inline `100` I don't get whitespaces, but when I interpolate it I do have them.

src/widgets/panel/__tests__/panel.test.tsx
sarahdayan: Yes this is better, I switched for `stats` in 8ec8259e.

Haroenv (approved):

PR #5091, opened 2022-08-23
feat(connectors): deprecate `hasNoResults` in favor of `canRefine`

aymeric-giraudet:
**Summary**

[FX-740](https://algolia.atlassian.net/browse/FX-740)

We want to encourage users to use `canRefine` instead of `hasNoResults`, and to provide them with more meaningful and easier ways to show users if a widget should be interacted with.

**Result**

- *hits-per-page*: `canRefine` is `true` when `hasNoResults` is `false` and vice-versa, no need to check if `items > 0` as it's done beforehand
- *numeric-menu*: `canRefine` can be true even if `hasNoResults` is true, in the case where a refinement has been applied since we can refine to "All"
- *rating-menu*: `canRefine` can be true even if `hasNoResults` is true when a rating is selected
- *sort-by*: `canRefine` is `true` when there are results and at least 1 item

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,7 +157,7 @@
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "70.25 kB"
+      "maxSize": "70.75 kB"
     },
     {
       "path": "./dist/instantsearch.development.js",
--- a/package.json
+++ b/package.json
@@ -161,7 +161,7 @@
     },
     {
       "path": "./dist/instantsearch.development.js",
-      "maxSize": "151.50 kB"
+      "maxSize": "152.00 kB"
     }
   ]
 }
--- a/package.json
+++ b/package.json
@@ -157,7 +157,7 @@
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "70.75 kB"
+      "maxSize": "70.50 kB"
     },
     {
       "path": "./dist/instantsearch.development.js",
--- a/src/connectors/hits-per-page/connectHitsPerPage.ts
+++ b/src/connectors/hits-per-page/connectHitsPerPage.ts
@@ -91,8 +91,14 @@ export type HitsPerPageRenderState = {
 
   /**
    * Indicates whether or not the search has results.
+   * @deprecated Use `canRefine` instead.
    */
   hasNoResults: boolean;
+
+  /**
+   * Indicates if search state can be refined.
+   */
+  canRefine: boolean;
 };
 
 export type HitsPerPageWidgetDescription = {
@@ -259,11 +265,14 @@ You may want to add another entry to the \`items\` option with this value.`
       },
 
       getWidgetRenderState({ state, results, createURL, helper }) {
+        const hasNoResults = results ? results.nbHits === 0 : true;
+
         return {
           items: transformItems(normalizeItems(state), { results }),
           refine: connectorState.getRefine(helper),
           createURL: connectorState.createURLFactory({ state, createURL }),
-          hasNoResults: results ? results.nbHits === 0 : true,
+          hasNoResults,
+          canRefine: !hasNoResults,
           widgetParams,
         };
       },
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -157,7 +157,7 @@
   "bundlesize": [
     {
       "path": "./dist/instantsearch.production.min.js",
-      "maxSize": "70.25 kB"
+      "maxSize": "70.75 kB"
```
aymeric-giraudet, line 160: Could leave a smaller footprint if I could use `find` for the array operations, I could also use `filter` but I don't think it's great cause it iterates on every item

src/connectors/hits-per-page/connectHitsPerPage.ts
```diff
@@ -259,11 +265,14 @@ You may want to add another entry to the \`items\` option with this value.`
...
       getWidgetRenderState({ state, results, createURL, helper }) {
+        const hasNoResults = results ? results.nbHits === 0 : true;
+
         return {
           items: transformItems(normalizeItems(state), { results }),
           refine: connectorState.getRefine(helper),
           createURL: connectorState.createURLFactory({ state, createURL }),
-          hasNoResults: results ? results.nbHits === 0 : true,
+          hasNoResults,
+          canRefine: !hasNoResults,
```
dhayab, line 275: Since we're deprecating `hasNoResults` and planning on remove it, wdyt of reversing the logic so `hasNoResults` is the inverse of the computed value for `canRefine`? I understand it might not be easy for the other connectors since there are additional checks, but here it could work.

src/connectors/rating-menu/connectRatingMenu.ts
```diff
@@ -392,10 +397,12 @@ const connectRatingMenu: RatingMenuConnector = function connectRatingMenu(
...
         facetValues = facetValues.reverse();
 
+        const hasNoResults = results ? results.nbHits === 0 : true;
+
         return {
           items: facetValues,
-          hasNoResults: results ? results.nbHits === 0 : true,
-          canRefine: facetValues.length > 0,
+          hasNoResults,
+          canRefine: !hasNoResults || refinementIsApplied,
```
dhayab, line 405: Good addition 👍

dhayab (approved):
Looks good to me 👏

src/connectors/hits-per-page/connectHitsPerPage.ts
aymeric-giraudet: Makes sense, thanks ! :)

src/connectors/rating-menu/connectRatingMenu.ts
aymeric-giraudet: Thanks ! However after looking a bit more I think I'll also look at facet counts

src/connectors/rating-menu/connectRatingMenu.ts
```diff
@@ -392,10 +399,12 @@ const connectRatingMenu: RatingMenuConnector = function connectRatingMenu(
...
         facetValues = facetValues.reverse();
 
+        const hasNoResults = results ? results.nbHits === 0 : true;
+
         return {
           items: facetValues,
-          hasNoResults: results ? results.nbHits === 0 : true,
-          canRefine: facetValues.length > 0,
+          hasNoResults,
+          canRefine: (!hasNoResults || refinementIsApplied) && totalCount > 0,
```
dhayab, line 407: Can you walk me through the idea behind the addition of `totalCount > 0`?

From what I understand now, this conflicts with `(!hasNoResults || refinementIsApplied)` because if `hasNoResults` is truthy that means that there are no results (hits), which also means the count of facet values will amount to 0.

aymeric-giraudet: Let's say the user first clicks on 4 stars and above cause they want to see good movies, they can type a query which will yield no results but still have a total facet value which is more than 0, [like in this example](https://owgvsc.csb.app/?movies%5Bquery%5D=starship%20troopers&movies%5BratingMenu%5D%5Brating%5D=4) (sorry it does not look obvious it's selected though)
Now if the user types some more stuff and the total facet count is 0 we know this query won't ever yield any result, and `canRefine` will be false as opposed to the previous implem.

dhayab: Thanks Aymeric, that's clearer with your example 👍 .

PR #5092, opened 2022-08-24
refactor(html): migrate default templates to VDOM

sarahdayan:
## Summary

This PR migrates the default templates from Hogan.js to VDOM, and adds missing tests for them.

This is necessary before deprecating Hogan.js, to avoid triggering a warning when using default templates.

## Next steps

This PR revealed a number of issues with our current tests and some of our types. I limited the changes and fixes to keep the PR focused and legible, but we should tackle it to avoid slowing down future features.

I filed it as [technical debt](https://algolia.atlassian.net/l/cp/1t8Hz40B) so we can keep track of it.


src/lib/utils/cx.ts
```diff
@@ -0,0 +1,3 @@
+export function cx(cssClasses?: string | string[]) {
+  return Array.isArray(cssClasses) ? cssClasses.join(' ') : cssClasses;
+}
```
sarahdayan, line 3: This is only useful until we get rid of `string[]` as a possible type for `cssClasses`.

src/components/MenuSelect/__tests__/__snapshots__/MenuSelect-test.tsx.snap
```diff
@@ -74,7 +74,7 @@ exports[`MenuSelect should render <MenuSelect /> with items 1`] = `
           className="option"
           dangerouslySetInnerHTML={
             {
-              "__html": "foo ()",
+              "__html": "foo (0)",
```
sarahdayan, line 77: The previous snapshot was likely a bug.

src/components/Stats/__tests__/Stats-test.tsx
```diff
@@ -39,7 +39,7 @@ describe('Stats', () => {
               className="text"
               dangerouslySetInnerHTML={
                 {
-                  "__html": "results found in 42ms",
+                  "__html": "1,234 results found in 42ms",
```
sarahdayan, line 42: The previous snapshot was likely a bug.

src/widgets/breadcrumb/__tests__/__snapshots__/breadcrumb-test.ts.snap
```diff
@@ -28,8 +28,8 @@ exports[`breadcrumb() render renders transformed items correctly 1`] = `
   "refine": [Function],
   "templateProps": {
     "templates": {
-      "home": "Home",
-      "separator": ">",
+      "home": [Function],
+      "separator": [Function],
```
sarahdayan, line 32: These snapshots should go away, but I'd rather do this in a different PR. Applies to all other snapshots of this type.

aymeric-giraudet (approved):
Great job migrating all these ! 🎉

src/widgets/rating-menu/defaultTemplates.tsx
```diff
@@ -0,0 +1,86 @@
...
+        cssClasses={cssClasses}
+      >
+        {stars.map((isFull, index) => (
+          <svg
+            key={index}
+            className={cx(
+              cssClasses.starIcon?.concat(
+                isFull ? 'cssClasses.fullStarIcon' : 'cssClasses.emptyStarIcon'
+              )
+            )}
```
dhayab, line 61:
```suggestion
            className={cx(
              cssClasses.starIcon,
              isFull ? cssClasses.fullStarIcon : cssClasses.emptyStarIcon,
            )}
```

dhayab (commented):
Looks great. I just have a single comment below.

src/widgets/rating-menu/defaultTemplates.tsx
sarahdayan: Doesn't work, this `cx` is a temporary one to support the fact that `cssClasses` can be `string[]` (which we should drop in v5). This `cx` doesn't take variadic arguments, only a single one that's a `string` or a `string[]`.

dhayab: Ah I see! In this case if we're sure `cssClasses.starIcon` is an array that's fine. I updated the suggestion to return the values of `cssClasses.fullStarIcon` and `cssClasses.emptyStarIcon` in the ternary.

```suggestion
            className={cx(
              cssClasses.starIcon?.concat(
                isFull ? cssClasses.fullStarIcon : cssClasses.emptyStarIcon
              )
            )}
```

src/widgets/refinement-list/defaultTemplates.tsx
```diff
@@ -0,0 +1,40 @@
...
+          type="checkbox"
+          className={cx(cssClasses.checkbox)}
+          value={value}
+          defaultChecked={isRefined}
+        />
+        <span
+          className={cx(cssClasses.labelText)}
+          dangerouslySetInnerHTML={
+            isFromSearch ? { __html: highlighted } : undefined
+          }
```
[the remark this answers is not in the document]
Haroenv: unless you think that would be too big of a possible breaking change (as it used {{{ before), in which case maybe an  `@MAJOR or @TODO` comment?

src/widgets/stats/stats.tsx
```diff
@@ -74,19 +78,50 @@ export type StatsWidget = WidgetFactory<
...
+    hasOneSortedResults,
+    hasManySortedResults,
+    hasNoResults,
+    hasOneResult,
+    hasManyResults,
+    nbSortedHits,
+    nbHits,
+    processingTimeMS,
+  }) {
+    let sentence = '';
```
Haroenv, line 93: I'd like this more if it's extracted in a function so that instead of reassigning the variable, it can be different return statements, what do you think?

Haroenv (commented):
overall just a small couple of nits left, great work!

src/widgets/infinite-hits/defaultTemplates.ts
```diff
@@ -1,9 +1,15 @@
 import type { InfiniteHitsComponentTemplates } from '../../components/InfiniteHits/InfiniteHits';
 
 const defaultTemplates: InfiniteHitsComponentTemplates = {
-  empty: 'No results',
-  showPreviousText: 'Show previous results',
-  showMoreText: 'Show more results',
```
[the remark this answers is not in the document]
sarahdayan: Is it worth it? We're aiming to deprecate string-based templates anyway, should we spend time and add code for this?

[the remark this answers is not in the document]
Haroenv: I was wondering whether simple templates should also be forced to use a function, it seems heavy handed for something like "show more results", which for those people would always be a string. We can of course add the functionality back after the breaking change has happened though

src/widgets/rating-menu/defaultTemplates.tsx
sarahdayan: Damn, I missed this 🤦‍♀️ Fixed in [ac6464f](https://github.com/algolia/instantsearch.js/pull/5092/commits/ac6464ff631592a03a96d9e16fab7b8aa6091961).

src/widgets/infinite-hits/defaultTemplates.ts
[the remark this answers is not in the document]
sarahdayan: It could be worth having a separate discussion around templates in general in our libraries, yes.

In Autocomplete, all templates are functions. In Recommend there's no Templates API, but each component is a function. Whatever we pick, we should aim for a consistent experience.

Templates as strings are more straightforward, I agree, but accepting multiple types also means we have to handle more use cases internally, leading to more code and more opportunities to miss edge cases. I'm not sure this is worth it, especially since unless you use IE11, you can have really terse code with fat arrow functions.

src/widgets/hierarchical-menu/defaultTemplates.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { cx } from '../../lib/utils';
+import type { HierarchicalMenuComponentTemplates } from './hierarchical-menu';
+
+const defaultTemplates: HierarchicalMenuComponentTemplates = {
+  item({ url, label, count, cssClasses }) {
+    return (
+      <a className={cx(cssClasses.link)} href={url}>
+        <span className={cx(cssClasses.label)}>{label}</span>
+        <span className={cx(cssClasses.count)}>
+          {formatNumber(String(count))}
```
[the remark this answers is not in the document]
sarahdayan: That was a hasty refactor, I fixed it in 0d1888cf.

src/widgets/stats/stats.tsx
sarahdayan: Does something like [this](https://github.com/algolia/instantsearch.js/pull/5092/commits/e2389c1c46f0bf00756886b581da8dbdeee77ddd) works better for you?

Haroenv (approved):

src/widgets/refinement-list/defaultTemplates.tsx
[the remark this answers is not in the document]
sarahdayan: If we want to use the `<Highlight>` component we need to adapt it because SFFV highlighting uses a different DOM (different class names and didn't have a wrapping tag for non highlighted parts).

I think we can wait for the next major and avoid extra code for now. I added an annotation in 4531f2ad.

Haroenv (approved):

dhayab (approved):

Haroenv (approved):

PR #5095, opened 2022-08-30
feat(html): deprecate Hogan.js and string-based templates

sarahdayan:
## Summary

This PR deprecates Hogan.js and string-based templates by emitting a warning in the console (in dev mode) when using string-based templates.


src/components/Template/Template.tsx
```diff
@@ -38,6 +39,17 @@ class Template extends Component<TemplateProps> {
   }
 
   public render() {
+    warning(
+      Object.keys(this.props.templates).every(
+        (key) => typeof this.props.templates[key] === 'function'
+      ),
```
sarahdayan, line 45: Using `Object.keys` instead of `Object.values` for IE11 support.

src/components/Template/__tests__/Template-test.tsx
```diff
@@ -58,6 +65,27 @@ describe('Template', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('warns when using string-based templates', () => {
```
Haroenv, line 68: this needs a test for the geosearch logic IMO

src/components/RefinementList/__tests__/__snapshots__/RefinementListItem-test.tsx.snap
```diff
@@ -6,6 +6,7 @@ exports[`RefinementListItem renders an item 1`] = `
   onClick={[Function]}
 >
   <Template
+    $$widgetType=""
```
Haroenv, line 9:
```suggestion
```

src/components/GeoSearchControls/GeoSearchControls.tsx
```diff
@@ -55,6 +55,7 @@ const GeoSearchControls = ({
               >
                 <Template
                   {...templateProps}
+                  $$widgetType="ais.geoSearch"
```
Haroenv, line 58: instead of this exception, is there some way we could make it work as expected? maybe with a placeholder element we preact.render into?

src/widgets/geo-search/__tests__/geo-search.test.tsx
```diff
@@ -0,0 +1,362 @@
...
+
+        await wait(0);
+      }).not.toWarnDev();
+
+      await wait(0);
+    });
+
+    type CustomHit = { name: string; description: string };
+
+    function createMockedSearchClient(
```
Haroenv, line 53: this could be top-level

sarahdayan: I nested it in the `describe` block for all other widget tests, partly because this prevented having to make a more generic function right away.

Would it be worth moving them all as we write more tests in those files instead?

src/components/GeoSearchControls/GeoSearchControls.tsx
sarahdayan: You mean make `html` work in `geoSearch` right now instead of waiting, right?

src/widgets/geo-search/__tests__/geo-search.test.tsx
Haroenv: Personally I prefer having all the setup functions top-level as it's impossible to accidentally read over tests that way. I'm ok with doing this later too though

src/components/GeoSearchControls/GeoSearchControls.tsx
Haroenv: yes, it seems a little odd to have this single exception, without externally clear reason

src/widgets/geo-search/__tests__/geo-search.test.tsx
sarahdayan: I'll do this in the final PR, once all stacked ones are merged.

src/components/GeoSearchControls/GeoSearchControls.tsx
sarahdayan: This was mostly because of time constraints. It isn't meant to last.

Haroenv (changes requested):
my existing comments are still open

- Either we make geoSearch work and these workarounds can go away
- Either they don't go away and we make sure there's no `$$widgetType=""` and geoSearch behaviour is tested

Haroenv (approved):

Haroenv:
toggle needs a couple type changes because you added name now by the way

Haroenv (approved):

Issue #5100, opened 2022-09-07
Inserting an InsightsMiddleware before widgets results in urlSynced-state being cleared on reload


PLNech:
## 🐛 Bug description

When we `createInsightsMiddleware` and insert that component before other widgets in an `<InstantSearch>` page, any state that is in urlSync gets erased on page load (such as pressing Enter or navigating from another page).
CC @Haroenv with whom I discovered the root cause of the UI issue.

## 🔍 Bug reproduction

1. Create an insights middleware that sets some stuff
2. Insert it in your `<InstantSearch routing={true}>` page _before other widgets_ (such as `<RefinementList>`)
3. Click a filter to set some `urlSync`-ed parameters in the URL
4. Click on your URL and press enter
5. See the page reloading: the url is first kept, then erased during a second painting of the page

## :curly_loop: Workaround

We realized that changing the position of the middleware in the DOM, from:
```jsx
<InstantSearch>
  <InsightsMiddleware />
  <RefinementList />
  <Hits />
</InstantSearch>
```
to
```jsx
<InstantSearch>
  <RefinementList />
  <Hits />
  <InsightsMiddleware />
</InstantSearch>
```
is enough to workaround the race condition between the middleware and the other widgets. :thinking: 

**Live reproduction:**

https://codesandbox.io/s/nameless-forest-chr25g?file=/App.tsx

## 💭 Expected behavior

I'd expect the middleware to not clear the other widgets' state, _regardless of where it is inserted_.

Haroenv:
A smaller reproduction is:

```ts

describe('integration', () => {
  test('does not write to the URL on load', async () => {
    const router = history({
      writeDelay: 0,
    });
    const routerWriteSpy = jest.spyOn(router, 'write');
    const search = instantsearch({
      indexName: 'instant_search',
      searchClient: createSearchClient({
        // @ts-expect-error these types are not part of "SearchClient", but a real client exposes it
        applicationID: 'abc',
        apiKey: 'def',
      }),
      routing: {
        router,
      },
    });

    search.start();

    search.use(
      createInsightsMiddleware({
        insightsClient(eventName, ...args) {
          if (eventName === 'onUserTokenChange') {
            const [cb] = args;
            cb('token');
          }
        },
      })
    );

    await wait(1);

    expect(routerWriteSpy).toHaveBeenCalledTimes(0);
  });
});
```

essentially what happens is there are two parts in the insights middleware which add widgets:

[...]

PR #5101, opened 2022-09-09
fix(insights): make sure change in userToken can't reset the search parameters

Haroenv:
**Summary**

previously we relied on a configure widget to set the userToken and clickAnalytics parameters. Unfortunately when the insights middleware is the first parameter, and you have a userToken set via cookie, the widget is removed and added again (setUserTokenAndSearch is called twice), and thus a change event is triggered before the other widgets are even mounted.

**Result**

Helper methods are used to prevent a change event being called by setting or updating userToken. It still triggers a new search as expected

fixes https://github.com/algolia/instantsearch.js/issues/5100


src/types/middleware.ts
```diff
@@ -5,6 +5,7 @@ import type { AtLeastOne } from './utils';
 export type MiddlewareDefinition<TUiState extends UiState = UiState> = {
   onStateChange(options: { uiState: TUiState }): void;
   subscribe(): void;
+  started(): void;
```
aymeric-giraudet, line 8: So this means user-made custom middlewares now need to implement `started()` as well ?
Could it be optional ?

Haroenv: It's actually optional in the public definition, but our own middleware use the type without optional keys @aymeric-giraudet

aymeric-giraudet (approved):
All good then ! 😄

FabienMotte (approved):

src/types/middleware.ts
Haroenv: see https://github.com/algolia/instantsearch.js/blob/6e2bf7b5cb0f56395938b17d7ef44f986ea45ca1/src/types/middleware.ts#L22 :)

PR #5104, opened 2022-09-12
fix(setUiState): call onStateChange handler

### LINKED ISSUE #5103: Calling `setUiState` directly does not trigger `onStateChange` when defined.
## 🐛 Bug description

When calling `setUiState` directly, if `onStateChange` is defined, it does not get called. The state gets properly updated, but the handler is not triggered, even though a state change occurred.

## 🔍 Bug reproduction

https://codesandbox.io/s/bold-sunset-k7giog?file=/src/app.js

- Type something or filter using the facets to set the state to something other than the default value.
- Click the added button which will reset the state using `setUiState` manually.
- Note that the `onStateChange` function does not get called (nothing in the console).

## 💭 Expected behavior

When calling `setUiState` directly (e.g. via a button click or change), the `onStateChange` handler should get called if it's defined.

## Environment

- OS Mac Monterey 12.5.1
- Browser: Chrome
- Version: 4.44

## Known Workaround

Using a middleware `.onStateChange` will get called.

Haroenv:
**Summary**

When calling setUiState, the onStateChange function should also be called, so that it can be used to keep track of state changes from any source. However, we need to make sure that onStateChange itself doesn't cause infinite loops of course!

**Result**

- onStateChange is also called for setUiState
- a new (private) argument to setUiState is added to allow avoiding calling onStateChange

This could be considered a breaking change, as before this fix onStateChange wasn't called when you are using InstantSearch.setUiState, but I don't see how that could cause issues. Middleware's onStateChange already was called before this PR

fixes #5103


src/lib/InstantSearch.ts
```diff
@@ -616,8 +616,14 @@ See ${createDocumentationLink({
...
 
+  /**
+   * Set the UI state and trigger a search.
+   * @param uiState The next UI state or a function computing it from the current state
+   * @param fromStateChange private parameter used to know if the method is called from a state change
+   */
   public setUiState(
-    uiState: TUiState | ((previousUiState: TUiState) => TUiState)
+    uiState: TUiState | ((previousUiState: TUiState) => TUiState),
+    callOnStateChange: boolean = true
```
Haroenv, line 626: I hoped there would not be a need for this parameter, but the only alternative I saw was adding a private setUiState function too, and that seemed overkill (knowing that's callable by users too

src/lib/InstantSearch.ts
```diff
@@ -633,32 +639,27 @@ See ${createDocumentationLink({
         ? uiState(this.mainIndex.getWidgetUiState({}) as TUiState)
         : uiState;
 
-    const setIndexHelperState = (indexWidget: IndexWidget) => {
```
Haroenv, line 636: this function is extracted to an utility, as once uiState became an argument, it was no longer needed to redeclare it on every call. This should in theory make changing state/searching a little faster (probably imperceptible)

src/lib/InstantSearch.ts
```diff
@@ -633,32 +639,27 @@ See ${createDocumentationLink({
...
+              : finalUiState,
+            this.mainIndex
+          );
+
+          this.scheduleSearch();
+          this.onInternalStateChange();
+        },
+      });
+    } else {
+      setIndexHelperState(nextUiState, this.mainIndex);
```
Haroenv, line 658: this could be in the same branch if we had a default to onStateChange (same with the if inside helper.search override in index), should we do that in a major version, now or never?

cdhawke (commented):
Looks good to me, i'll let someone else approve though

FabienMotte (approved):
LGTM too! Thanks for the review meeting

PR #5110, opened 2022-09-20
Circleci/trigger dependents

Haroenv:
**Summary**

**Result**


Haroenv:
the code to synchronise the branches is too error-prone, as well as requiring a build or waiting for the codesandbox step to finish. This code can be used as a base later if needed

Issue #5109, opened 2022-09-14
I am using latest version of angular-instant search and this version of instant search is causing issue in application startup


Sharma1Krishna:
## 🔍 Bug reproduction

**Steps to reproduce the behavior:**

1. Go to `...`
2. Click on `...`
3. Scroll down to `...`

**Live reproduction:**

https://codesandbox.io/s/github/algolia/create-instantsearch-app/tree/templates/instantsearch.js

## Environment

- OS: [e.g. Windows / Linux / macOS / iOS / Android]
- Browser: [e.g. Chrome, Safari]
- Version: [e.g. 22]

Haroenv:
Hey @Sharma1Krishna, what's the "an issue"? Do you have a sandbox that reproduces what happens?

Sharma1Krishna:
Let me create one, I was using local earlier

Sharma1Krishna:
<img width="623" alt="image" src="https://user-images.githubusercontent.com/48248163/190099096-b62d96a4-42bc-4281-ae2b-111e57331b32.png">

Sharma1Krishna:
<img width="623" alt="image" src="https://user-images.githubusercontent.com/48248163/190099249-a57ef933-ea5f-41d0-8ecf-7fb0fed6c2b3.png">

I am getting this error whenever I am starting the application

Sharma1Krishna:
It was working earlier but it stopped when https://socket.dev/npm/package/instantsearch.js/overview/4.46.0 this issue got published

Sharma1Krishna:
you can try creating a new application in your local with angular version 13 with these two libraries

Sharma1Krishna:
Because of this issue, our whole team is blocked for the development of algolia-related stories, please help us asap.

Sharma1Krishna:
I tried lowering the version of these libraries and it is working at

"algoliasearch": "4.12.0",
    "angular-instantsearch": "4.1.0",
    
    and need to add one dev dependency 
    "@types/algoliasearch": "^4.0.0",
    
    but because this is lower version of algolia libraries , it is not compatible(a lot of existing functionality is breaking as we have also developed some customized version of algolia widgets and we are using custom connector as well) with our existing development which were based on latest version of these two libraries

Haroenv:
@types/algoliasearch doesn't need to exist, algoliasearch, angular-instantsearch and instantsearch.js come with their builtin definitions

Sharma1Krishna:
[...]

PR #5111, opened 2022-09-21
fix(imports): split out templating from ./utils

Haroenv:
**Summary**

Put everything related to `helpers` out of the `utils` folder, as not every bundler tree shakes correctly (metro for example)

**Result**

This should fix https://github.com/algolia/react-instantsearch/issues/3632 as well, as renderTemplate will no longer be in the imported path for connectors (react instantsearch hooks), and thus no errors can be in it either


aymeric-giraudet:
Great ! It might fix https://github.com/algolia/instantsearch.js/issues/5109 as well

Haroenv:
Not sure, I didn't find anything related to `react` in the bundle @aymeric-giraudet

aymeric-giraudet (approved):

PR #5123, opened 2022-09-29
feat: preserve shared state on unmount with a future flag

Haroenv:
This PR introduces a functional change in how UI state is handled when widgets are removed from InstantSearch.

Currently, when a widget is removed, its `dispose()` method returns a cleaned state that is used to compute the new shared UI state. This means that if there are multiple widgets targeting the same attribute, when one of them is removed, the widget that is left has its UI state reset as a side-effect.

Through the introduction of future flags in InstantSearch, we are able to explicitly switch to a new dispose mode, which doesn't use the cleaned state returned by removed widgets, to allow multiple instances of widgets to keep the refinements active.

It's easy to opt into this new dispose method:

```js
instantsearch({
  // ...
  future: {
    preserveSharedStateOnUnmount: true,
  },
});
```

Having this option enabled with a future flag allows us to easily release it without breaking existing implementations, allowing users to adopt it at their own rhythm.

This is made available for InstantSearch.js, React InstantSearch and Vue InstantSearch.

FX-2621

<details>
<summary>Initial PR description</summary>
we have ui state and widgets reading this ui state. This means that when you dispose a widget, you don't need to call that function if you start with a clean helper state instead of persisting the state.

This needs to be done in a major version, as people _might_ rely on state without ui state (i'm not sure if they do) and that would be removed if they dispose a widget.

This allows again for classic react-instantsearch-core style virtual widgets for the modal and dynamic widget use case

This could be introduced non-breaking as a different "mode" InstantSearch behaves in, that could be swapped the default value later
</details>


src/widgets/index/index.ts
```diff
@@ -347,28 +347,28 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
...
       if (localInstantSearchInstance && Boolean(widgets.length)) {
-        const nextState = widgets.reduce((state, widget) => {
-          // the `dispose` method exists at this point we already assert it
-          const next = widget.dispose!({
+        widgets.forEach((widget) => {
+          // doing nothing with the return value of dispose, this is handled by the next part
+          widget.dispose!({
             helper: helper!,
-            state,
+            state: helper!.state,
```
Haroenv, line 354: if we're actually doing this, dispose should no longer accept helper, state, nor should it return search parameters, but I didn't do that in this PR out of simplicity.

src/widgets/index/index.ts
```diff
@@ -450,7 +450,10 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
       };
 
       derivedHelper = mainHelper.derive(() =>
-        mergeSearchParameters(...resolveSearchParameters(this))
+        mergeSearchParameters(
+          mainHelper.state,
```
Haroenv, line 454: this part here is to ensure that "statically set parameters" on `mainHelper` persist (parameters on `helper` don't, as that actually is the helper of the mainIndex, confusing naming, I know).

If we're doing such a refactor, we probably should simply get rid of `helper`, and replace it with `mainHelper` only (which then gets name swapped)

examples/react/getting-started/src/App.tsx
```diff
@@ -40,17 +49,24 @@ export function App() {
           searchClient={searchClient}
           indexName="instant_search"
           insights={true}
+          disposeMode={disposeMode}
```
Haroenv, line 52: we probably need to set up an updater for this parameter, and in vue too (after it's accepted)

dhayab: I was thinking instead of handling prop update, we could warn (like we do with `searchClient`), we probably don't want users to mess with this at runtime, wdyt?

Haroenv: yes, however we could also just update the flag likely (until we have an InstantSearch options update method that is also used in constructor 👍)

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -235,8 +252,16 @@ Use \`InstantSearch.status === "stalled"\` instead.`
...
+      future = {
+        ...INSTANTSEARCH_FUTURE_DEFAULTS,
+        ...(options.future || {}),
+      },
     } = options;
 
+    if (indexName === null) {
+      throw new Error(withUsage('The `indexName` option is required.'));
+    }
+
```
Haroenv, line 264:
```suggestion
```

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -235,6 +252,10 @@ Use \`InstantSearch.status === "stalled"\` instead.`
       searchClient = null,
       insightsClient = null,
       onStateChange = null,
+      future = {
+        ...INSTANTSEARCH_FUTURE_DEFAULTS,
+        ...(options.future || {}),
+      },
     } = options;
```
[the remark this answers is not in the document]
dhayab: Yes we do! I'd like to do it with a link to the API reference. I don't think we solved the issue of displaying per-flavour API reference links, have we?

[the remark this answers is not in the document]
Haroenv: I guess just the JS reference and people can switch flavours is the best we can do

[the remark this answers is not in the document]
dhayab: I kind of see a future where RIS/VIS would register themselves to the InstantSearch.js instance so we could generate flavoured doc URL, but it may also be overkill just for this.

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -304,6 +306,21 @@ See ${createDocumentationLink({
       })}`
     );
 
+    safelyRunOnBrowser(() => {
+      if (options.future?.preserveSharedStateOnUnmount === undefined) {
```
Haroenv, line 310: why not `warn` or `warning`? they already take a boolean that can be filled with `options.future?.preserveSharedStateOnUnmount !== undefined`

tests/common/widgets/current-refinements/options.ts
```diff
@@ -146,7 +146,7 @@ export function createOptionsTests(
                 <span
                   class="ais-CurrentRefinements-categoryLabel"
                 >
-                  Samsung
+                  Apple
```
Haroenv, line 149: do you know what caused this change?

dhayab: I don't know exactly why, but it comes from https://github.com/algolia/instantsearch/pull/5123/files#diff-6af3d106c335b51f0e85d49f808b248a9704cb12dab1472606b6e9a383d981b1R508-R511.

I haven't been able to dig into where the refinements value order change, but during setup the order is as expected, and it changes as soon as the index is initialized.

A potential guess is that before, `mergeSearchParameters` only received one argument, and this short-circuited the reduce operation inside because it doesn't have an init value). Now we always have 2+ parameters with the addition of `mainHelper.state`, so we go through the reduce steps that merge all the facets type.

Ultimately, this order is the correct one, as it seems current refinements wants to respect refinements values order (except numeric refinements): https://github.com/algolia/instantsearch/blob/master/packages/instantsearch.js/src/connectors/current-refinements/connectCurrentRefinements.ts#L315.

Haroenv: sounds good! maybe let's split out this part of the PR in a separate commit then, so it shows up differently in the changelog

dhayab: Changes are now in https://github.com/algolia/instantsearch/pull/5875

packages/instantsearch.js/src/lib/InstantSearch.ts
dhayab: A warning might signify to users that not adopting the change will break their implementation, so I preferred a lighter informational message.

.codesandbox/ci.json
```diff
@@ -4,6 +4,7 @@
     "github/algolia/instantsearch/tree/templates/react-instantsearch",
     "/examples/react/default-theme",
     "/examples/react/next-app-router",
+    "/examples/react/getting-started",
```
Haroenv, line 7: should we replace `default-theme` with `getting-started` or just revert this change?

packages/instantsearch.js/src/lib/InstantSearch.ts
Haroenv: I'm mainly thinking of the __DEV__ wrapping for warn. Not sure if info or warn makes much difference though, it's a valid point

Haroenv (commented):
Not sure what the right solution is for warn/info but otherwise this lgtm

packages/instantsearch.js/src/lib/InstantSearch.ts
dhayab: Oh right, I didn't think about this and agree it should be scoped to dev mode. I'll add this.

.codesandbox/ci.json
dhayab: I'll revert this and we can decide in another PR.

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -307,7 +307,10 @@ See ${createDocumentationLink({
     );
 
     safelyRunOnBrowser(() => {
-      if (options.future?.preserveSharedStateOnUnmount === undefined) {
+      if (
+        __DEV__ &&
+        options.future?.preserveSharedStateOnUnmount === undefined
+      ) {
```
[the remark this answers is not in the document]
dhayab: I originally wanted to prevent this message from adding too much noise in server logs (ie Next), but I thought about it again now and we at least want them to notice the message at least once so that they explicitly set the flag. I'll remove it.

Haroenv (commented):
✔️

packages/instantsearch.js/src/widgets/index/__tests__/index-test.ts
sarahdayan, line 1: Why not use this opportunity to create an `tests/common/widgets/index/options.ts` file and test this new behavior in it? No need to migrate the rest of our tests, we can do it separately, but at least we could start "the right way" and test as a user, without manipulating the helper.

Also, we should add a test for the default case (`false`, but not passed by the user).

dhayab: Yes, [I've pushed the CTS here](https://github.com/algolia/instantsearch/commit/159f07868897a3876701d40331ade12ecfe05f0f) if you want to take a look, basically we saw lifecycle issues with React and Vue in the testing environment for which we could not find solutions for now.

Thanks for the suggestion on the missing test case, I've added it in 58ddb2a.

sarahdayan (approved):

PR #5124, opened 2022-09-29
fix(hierarchicalMenu): pass correct attribute name to Insights

sarahdayan:
When clicking on facets in a `hierarchicalMenu`, the passed `attribute` is always the first one passed in the [`attributes`](https://www.algolia.com/doc/api-reference/widgets/hierarchical-menu/js/#widget-param-attributes) option, no matter what level it actually is.

https://user-images.githubusercontent.com/5370675/193083634-ba433032-135d-4535-abb9-5cf9fbd8a58f.mov

This fixes it by inferring the level based on the `facetName` by splitting on `separator`.

FX-1801


src/lib/utils/createSendEventForFacet.ts
```diff
@@ -12,19 +12,23 @@ type CustomSendEventForFacet = (customPayload: any) => void;
...
+  helper: AlgoliaSearchHelper;
+  attribute: string | ((facetValue: string) => string);
+  widgetType: string;
+};
+
 export function createSendEventForFacet({
   instantSearchInstance,
   helper,
-  attribute,
+  attribute: attr,
```
sarahdayan, line 25: Used `attr` (instead of `_attribute` or else) so that l. 30 remains on a single line.

Haroenv (approved):

dhayab (approved):

PR #5126, opened 2022-09-29
build: upgrade to Node.js 14

sarahdayan:
This PR upgrades InstantSearch.js to use Node.js 14 and the dependencies that broke otherwise.

It also lints the project (I only fixed the errors, there are warnings remaining) with the new rules from dependency upgrades.

I upgraded to Node.js 14 (and not higher) for now so we can move forward with dependency updates for other PRs. Later on, we should upgrade to Node.js 18 and update all our dependencies.


src/types/widget.ts
```diff
@@ -118,9 +118,6 @@ export type BuiltinWidgetTypes =
   | 'ais.toggleRefinement'
   | 'ais.voiceSearch';
 
-// In this case we don't want to allow any record, as it would no longer warn
-// for unexpected parameters
-// eslint-disable-next-line @typescript-eslint/ban-types
```
Haroenv, line 123: do you know why this no longer errors for `object` usage?

Haroenv (approved):

src/types/widget.ts
sarahdayan: It looks like `object` is considered safe now, and `Object` is banned (see [rule details](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/ban-types.md#rule-details)).

dhayab (approved):

PR #5127, opened 2022-09-30
fix(status): send render event when loading starts

Haroenv:
Without this it's impossible to use `loading` for anything. I opted to only emit the event, as calling render on the widgets would  require changes in all the tests.

If we render (fully), you still don't get updated of `loading` initially because `render` only gets called when there are results:

https://github.com/algolia/instantsearch.js/blob/4d43bd8292446101742a10d1da9f042f780d8d45/src/widgets/index/index.ts#L573-L575

To fix all of these inconsistencies, we should join `init` and `render` in a single function (always calling render, even if it's the first render, also when there's no results), but this would be a significant breaking change, so shouldn't be done lightly.

Without this PR, `useInstantSearch` in React InstantSearch Hooks doesn't get updated when the status changes to `loading`, as that doesn't emit render (yet)


src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -1289,9 +1289,16 @@ describe('scheduleRender', () => {
...
+      renderSpy({
+        widgetRenderTimes: castToJestMock(widget.render!).mock.calls.length,
+      });
     });
+
+    await wait(0);
+
+    expect(renderSpy).toHaveBeenCalledTimes(2);
+    expect(renderSpy).toHaveBeenNthCalledWith(1, { widgetRenderTimes: 0 });
+    expect(renderSpy).toHaveBeenNthCalledWith(2, { widgetRenderTimes: 1 });
```
sarahdayan, line 1301: This isn't obvious what the behavior is when just reading those lines, can you add comments?

src/lib/__tests__/status.test.ts
```diff
@@ -185,7 +185,7 @@ describe('status', () => {
   });
 
   test('lets users render on error with the `render` event', async () => {
-    expect.assertions(4);
+    // expect.assertions(4);
```
sarahdayan, line 188: Why is this commented?

sarahdayan (approved):

PR #5132, opened 2022-10-05
feat(menu): Searchable menu widget

### LINKED ISSUE #2893: RFC: Menu with searchable behaviour
# Summary

Just like `refinementList`, there are some cases in which you'd want to search within the options, even if there's only one option that can be chosen

# Basic example

```js
search.addWidget(
  instantsearch.widgets.menu({
    container: "#menu",
    attribute: "country",
    searchable: true,
  })
);
```

# Motivation

When a list of items, where only one can be choses, is really long, it gets complicated to show all options. For that reason we added search for facet values on refinement lists. This behaviour also applies to Menu.

# Detailed design

1. add `searchForItems` (fn) to the connector
2. implement it the same as for refinementList ([link](https://github.com/algolia/instantsearch.js/blob/88497e564d9a5322340afc7e8d0344e1aa8db674/src/connectors/refinement-list/connectRefinementList.js#L190-L198))
3. add `searchable` option to widget
4. add the searchbox etc. to the widget as necessary
5. implement in Vue InstantSearch and Angular InstantSearch

# Drawbacks

Why should we *not* do this? Please consider:

- code size increase
  - should not be too big, because it's code which is very close to the refinement list code
- timing
  - we can wait until Vue InstantSearch has moved to connectors in beta so we don't need to wait for this to be implemented

# Alternatives

Using a refinement list and allowing selection of multiple values, or hacking around in that function to make sure only one refinement is possible. This doesn't seem like that good of an alternative to be honest.

# How we teach this

Update the documentation, maybe a guide for searchable lists in general like React InstantSearch

cdhawke:
**Summary**

Adds searchable props to the `menu` widget similar to `refinement-list`. This change adds additional properties to `connectMenu` and `menu`, as well as adding the relevant search templates. This will be used by the new merchandising console with the upgrade to `react-instantsearch-hooks`. They currently use the searchable `menu` component in `react-instantsearch-dom`. 

**Result**

Stories:
/story/refinements-menu--with-search
/story/refinements-menu--with-search-and-show-more

Screenshot: 
image

fixes #2893


package.json
```diff
@@ -59,7 +59,7 @@
     "@types/google.maps": "^3.45.3",
     "@types/hogan.js": "^3.0.0",
     "@types/qs": "^6.5.3",
-    "algoliasearch-helper": "^3.11.1",
+    "algoliasearch-helper": "https://pkg.csb.dev/algolia/algoliasearch-helper-js/commit/377e8f23/algoliasearch-helper",
```
Haroenv, line 62: need to remember to update after approved, before merge

src/widgets/menu/defaultTemplates.tsx
```diff
@@ -7,17 +7,40 @@ import { cx } from '../../lib/utils';
...
+    count,
+    highlighted,
+    label,
+    isFromSearch,
+    searchable,
+  }) {
     return (
       <a className={cx(cssClasses.link)} href={url}>
-        <span className={cx(cssClasses.label)}>{label}</span>
+        {searchable ? (
```
Haroenv, line 21: note that this is different to the template for refinementList, to avoid a breaking change. In a major version it could be changed to always use `highlighted` (maybe this should be a comment)

src/widgets/menu/defaultTemplates.tsx
```diff
@@ -7,17 +7,40 @@ import { cx } from '../../lib/utils';
...
+          <span
+            className={cx(cssClasses.label)}
+            dangerouslySetInnerHTML={
+              isFromSearch ? { __html: highlighted } : undefined
+            }
```
Haroenv, line 26: there should be a todo here about using the highlight component, like in refinementList

src/widgets/menu/menu.tsx
```diff
@@ -81,18 +99,48 @@ export type MenuTemplates = Partial<{
...
     isShowingMore: boolean;
   }>;
+  /**
+   * Templates to use for search for facet values when there are no results.
+   */
+  searchableNoResults: Template;
 }>;
 
-export type MenuComponentCSSClasses = ComponentCSSClasses<MenuCSSClasses>;
+export type MenuComponentCSSClasses = RefinementListComponentCSSClasses;
```
Haroenv, line 124:
```suggestion
export type MenuComponentCSSClasses = ComponentCSSClasses<MenuCSSClasses>;
```

src/widgets/menu/menu.tsx
```diff
@@ -2,6 +2,7 @@
 
 import { h, render } from 'preact';
 import cx from 'classnames';
+import type { RefinementListComponentCSSClasses } from '../../components/RefinementList/RefinementList';
```
Haroenv, line 5:
```suggestion
```

package.json
cdhawke: Yes, depends on https://github.com/algolia/algoliasearch-helper-js/pull/917

src/widgets/menu/defaultTemplates.tsx
```diff
@@ -7,17 +7,43 @@ import { cx } from '../../lib/utils';
...
+    count,
+    highlighted,
+    label,
+    isFromSearch,
+    searchable,
+  }) {
     return (
       <a className={cx(cssClasses.link)} href={url}>
-        <span className={cx(cssClasses.label)}>{label}</span>
+        {/* @MAJOR: use the `<Highlight>` component instead */}
```
Haroenv, line 21:
```suggestion
        {/* @TODO: use the `<Highlight>` component instead */}
```

no big deal, but this can be done without breaking change

Haroenv (commented):
in theory I see nothing wrong with this PR, but need to do some deeper testing

src/connectors/menu/connectMenu.ts
```diff
@@ -90,6 +106,14 @@ export type MenuRenderState = {
    * True if refinement can be applied.
    */
   canRefine: boolean;
+  /**
+   * Searches for values inside the list.
+   */
+  searchForItems(query: string): void;
+  /**
+   * `true` if the values are from an index search.
```
sarahdayan, line 114: We'll need to also update this in `connectRefinementList` (unrelated to this PR).

```suggestion
   * Whether the values are from an index search or not.
```

src/connectors/menu/connectMenu.ts
```diff
@@ -135,11 +159,12 @@ export type MenuConnector = Connector<
...
  * **Menu** connector provides the logic to build a widget that will give the user the ability to choose a single value for a specific facet. The typical usage of menu is for navigation in categories.
  *
- * This connector provides a `toggleShowMore()` function to display more or less items and a `refine()`
- * function to select an item. While selecting a new element, the `refine` will also unselect the
- * one that is currently selected.
- *
  * **Requirement:** the attribute passed as `attribute` must be present in "attributes for faceting" on the Algolia dashboard or configured as attributesForFaceting via a set settings call to the Algolia API.
+ *
+ *  * This connector provides:
+ * - a `refine()` function to select an item. While selecting a new element, the `refine` will also unselect the one that is currently selected.
```
sarahdayan, line 165:
```suggestion
 * - a `refine()` function to select an item. Selecting a new element also unselects the one that is currently selected.
```

src/connectors/menu/connectMenu.ts
```diff
@@ -135,11 +159,12 @@ export type MenuConnector = Connector<
...
+ * - a `toggleShowMore()` function to display more or less items
```
sarahdayan, line 166:
```suggestion
 * - a `toggleShowMore()` function to display more or less items.
```

src/connectors/menu/connectMenu.ts
```diff
@@ -198,6 +223,98 @@ const connectMenu: MenuConnector = function connectMenu(
...
+      value,
+      label,
+      highlighted: label,
+    });
+
+    let lastResultsFromMainSearch: SearchResults;
+    let lastItemsFromMainSearch: MenuItem[] = [];
+    let hasExhaustiveItems = true;
+
+    const createSearchForFacetValues = function (
```
sarahdayan, line 246:
```suggestion
    function createSearchForFacetValues(
```

src/widgets/menu/defaultTemplates.tsx
sarahdayan: Instead of the big ternary, could we have the following in l.34?

```js
{!isFromSearch && searchable ? highlighted : label}
```

sarahdayan (commented):
Great job!

Seeing the necessary code to bring this feature to `menu`, I hope we can look into abstracting the `searchable` feature so we can maintain it in one place. This would be the object of another PR of course.

Since we're here, I think we need to test the feature in the widget in [menu-select.test.tsx](https://github.com/algolia/instantsearch.js/blob/02138a1ea176438ecf23efed252cae2ebd651dc9/src/widgets/menu-select/__tests__/menu-select.test.tsx#L1). These tests should be as close as possible to how users interact with the widget.

src/widgets/menu/defaultTemplates.tsx
Haroenv: that would use `label` when it's searchable, but not searching (so it would be inconsistent with refinement list). Another option however could be:

```
{searchable ? !isFromSearch && highlighted : label}
```

but that's essentially the same as the existing code, but with the ternary two times

Haroenv:
While this is the right approach, the team has decided we haven't seen enough signal to invest into this code, and would rather invest in maintaining search in refinementList, as it can be used in combination with setUiState to emulate a Menu.

Thanks a lot for your PR though!

cdhawke:
@Haroenv  Is it worth adding some documentation or an example or something to the instantsearch-hooks docs for people migrating?

Haroenv:
Good point, I'll put it in the migration guide

PR #5134, opened 2022-10-06
feat(routing): include repeated indexId in URL correctly

Haroenv:
Before this PR the deepest nested indexId takes "complete control" of its indexId, and thus widgets mounted deeper didn't get "registered"

Added a new test to the existing getUiState describe, as well as an e2e-style "routing has correct url" file that can be used for other cases later.

CR-2067

FX-1821

One thing to consider is whether this is a breaking change. Before this change, URLs would have only "the last index with that id" (which was wrong), but now it'll include the correct ui state


dhayab (approved):
That works. As a sidenote, we do promote having different `indexId` to prevent issues with routing, but maybe this information is not visible enough: https://www.algolia.com/doc/api-reference/widgets/index-widget/js/#widget-param-indexid.

Haroenv:
Yep, but having separate indexIds also causes separate queries / refinements, maybe not always what you wanted

PR #5136, opened 2022-10-11
fix(currentRefinements): reset page number on refine

aymeric-giraudet:
### [FX-1834](https://algolia.atlassian.net/browse/FX-1834)

**Summary**

All refinement widgets but `currentRefinements` reset the pagination. 
This can result in users seeing "No results" when clearing a refinement narrows the results.

**Result**

Added test to make sure we reset pagination when calling `refine` on `currentRefinements`


src/connectors/current-refinements/__tests__/connectCurrentRefinements-test.ts
```diff
@@ -333,6 +333,28 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/current-ref
         widgetParams: {},
       });
     });
+
+    test('refine() resets the page', () => {
```
dhayab, line 337:
```suggestion
    test('resets the page on refine()', () => {
```

dhayab (approved):
Good 👍 !

Haroenv:
How could clearing a refinement narrow the results?

src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -316,6 +316,7 @@ function clearRefinementFromState(
   state: SearchParameters,
   refinement: CurrentRefinementsConnectorParamsRefinement
 ): SearchParameters {
+  state = state.resetPage();
```
Haroenv, line 319: doesn't each individual method already reset the page? eg. [removeDisjunctiveFacetRefinement](https://community.algolia.com/algoliasearch-helper-js/reference.html#AlgoliaSearchHelper#removeDisjunctiveFacetRefinement)

aymeric-giraudet: It does when [called on the helper](https://github.com/algolia/algoliasearch-helper-js/blob/d9ebb01382ec83e0c579c393c3e9df83a9261699/src/algoliasearch.helper.js#L658-L665) but [does not on a `SearchParameters` instance](https://github.com/algolia/algoliasearch-helper-js/blob/fc183ec2910d4ba33067b3487f3a70e940fd24d3/src/SearchParameters/index.js#L846-L856) 🤔

aymeric-giraudet: And I guess we do it with `SearchParameters` because `clearRefinementFromState` is also used for `createURL` ?

Haroenv: ah yes of course, makes sense!

Haroenv (approved):

PR #5149, opened 2022-10-25
fix(insights): check before usage of `document`

### LINKED ISSUE #5148: Algolia Analytics with react native
Hey ,
When I tried use middleware with react native I got this error 
Screenshot 2022-10-12 at 5 43 32 PM
unnamed

Haroenv:
**Summary**

search-insights is compatible with react-native or other environments without document, but the insights plugin wasn't. This PR fixes that.

I've searched and outside of widgets there's no other occurrences of `document`, but didn't yet find a way to make sure it's a failing test if document is used. We can do that after InstantSearch would be split probably.

**Result**

fixes #5148

FX-1926

createInsightsMiddleware supports environments without `document` or `document.cookie`

### DIFF
--- a/src/helpers/get-insights-anonymous-user-token.ts
+++ b/src/helpers/get-insights-anonymous-user-token.ts
@@ -3,7 +3,9 @@ import { warning } from '../lib/utils';
 export const ANONYMOUS_TOKEN_COOKIE_KEY = '_ALGOLIA';
 
 function getCookie(name: string): string | undefined {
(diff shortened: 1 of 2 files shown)


FabienMotte (approved):

aymeric-giraudet (approved):

Alyfaraj:
Hey..any update about new version for this fix ?

Haroenv:
It will be released Monday @Alyfaraj

PR #5157, opened 2022-11-08
fix(answers): mark as deprecated

Haroenv:
While the answers feature still works, it no longer has any difference from a regular search, and therefore no longer needs its own specialised widget.

Even though it's marked as experimental, I still don't think it's a good idea to delete without major version, as it's used in the downstream packages


PLNech (approved):
Bye Answers; Long live Semantic search! :grin:

aymeric-giraudet (approved):

FabienMotte (approved):

PR #5158, opened 2022-11-10
fix: shouldWrite return false if isDisposed

erictaylor:
**Summary**

Updates the "router"'s `createURL` method to take into account any existing search params. This way `shouldWrite` check can be valid and not potentially write a new history push in SPAs where another router has pushed new history changes with search params.

A quick demo of the problem and how this PRs change fixes the issue:
https://www.loom.com/share/09539f87f691407cb7692ba21fa17537

**Result**

By taking into account the existing search params of the URL when creating it to be used in `shouldWrite`, allows `shouldWrite` to return correct expected boolean when another history event has pushed to new url with search params.


Haroenv:
Thanks for your PR, but this indeed would cause a state that's no longer needed to continue to be written

PR #5159, opened 2022-11-10
fix undesirable push when outside history push with search params

erictaylor:
**Summary**

In some SPAs instantsearch.js router will inappropriately write an undesirable new history state. The source from what I found is that the `createURL` method does not take into account existing search params which ultimately leads the `shouldWrite` method not return the correct boolean value.

By taking into account existing search params in `createURL` we fix these undesireable `window.history.pushState` calls coming from the library.

A demo of the bug and the fix can be viewed here:
https://www.loom.com/share/45df116aab104f7086c5c45932cba0dd

**Result**

The Loom linked above demos the corrected behavior from this fix. Additionally unit tests have been updated to reflect this corrected behavior as well.


erictaylor:
#5167 succeeds this PR. This PR was auto closed when attempting to fix PR with monorepo changes that occurred on repo.

PR #5162, opened 2022-11-16
refactor: migrate to a monorepo

Haroenv:
This PR is obviously very big, but it doesn't work on its own unfortunately, everything needs to be moved for it to work.

1. a monorepo is set up with Lerna (latest version, so using nx)
2. source code of instantsearch.js is moved into its own package (same readme on both levels for now)
3. scripts are either kept in the same place, moved into instantsearch folder
4. mocks and testutils that will be useful for other packages later are moved into its own packages
5. examples are made part of the workspace, got updated dependencies and use the local instantsearch.js
6. examples are moved in a subfolder (in src and website) and have a redirect set up
7. ci has been rewritten to be able to use cache and workspace
8. modify the "algoliasearch v3" test to deal with multiple dependencies

This depends on two pull requests that can be merged or released either before or after this PR:
- e2e tests: https://github.com/algolia/instantsearch-e2e-tests/pull/38 relies on a branch release
- shipjs: https://github.com/algolia/shipjs/pull/984 relies on an alpha version


sarahdayan (approved):
Based on our peer review yesterday, LGTM 👍

Fantastic work ✨

aymeric-giraudet (approved):

PR #5167, opened 2022-11-22
Fix undesired history push when history changed with search params

erictaylor:
**Summary**

In some SPAs instantsearch.js router will inappropriately write an undesirable new history state. The source from what I found is that the `createURL` method does not take into account existing search params which ultimately leads the `shouldWrite` method not return the correct boolean value.

By taking into account existing search params in `createURL` we fix these undesireable `window.history.pushState` calls coming from the library.

A demo of the bug and the fix can be viewed here:
https://www.loom.com/share/45df116aab104f7086c5c45932cba0dd

**Result**

The Loom linked above demos the corrected behavior from this fix. Additionally unit tests have been updated to reflect this corrected behavior as well.


Haroenv (changes requested):
Sorry for the delay here, and thanks for rebasing, unfortunately I don't think this is the true solution for the problem 😢 

If we take for example https://codesandbox.io/s/instantsearch-js-forked-075yfp?file=/src/app.js (where state mapping picks the "instant_search" key into the query parameters). Here you see that if you add an item to the URL (typing something) and then remove it again so it shouldn't be written anymore (emptying the query) you'll see that it's still in the URL.

This is working without stateMapping as the "RouteState" object is controlled completely and can write anything you want, ignoring the rest of the URL, but that's not always the case.

For reference, I'm sure you know but `createURL` is customisable, so maybe in your case you don't modify the routeState or keep it similarly nested, you could use this solution on your app already.

Haroenv:
@erictaylor, do you have an idea around the limitations I mentioned? Thanks!

erictaylor:
On holiday leave currently. Will respond after the holiday.

Haroenv:
Thanks again for your Pull Request, but unfortunately this causes a removed parameter to still be in the URL. I'd love to find a way to fix this, but for now the only thing that works is taking full control of the search parameters. You can't assume that the parameters before InstantSearch mounts need to always stay.

PR #5171, opened 2022-11-24
chore(deps): deduplicate lockfile

Haroenv:
**Summary**

deduplicate installed packages in the repo (mainly using the https://github.com/scinos/yarn-deduplicate tool).
I couldn't find a tool that highlighted duplicates, but this is already an improvement, especially in preparation of more packages being added later.

**Result**

Highlights:
- only one version of babel in the whole project
- saves ~100MB of node_modules (950MB -> 850MB)

There's probably even more simplifications possible, but these I saw easily possible.

### DIFF
--- a/examples/js/calendar-widget/package.json
+++ b/examples/js/calendar-widget/package.json
@@ -12,7 +12,7 @@
     "instantsearch.js": "4.49.1"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


FabienMotte (approved):
Nice! 👏

PR #5175, opened 2022-11-30
test(is-module): test from built/resolved package

Haroenv:
**Summary**

imports from the built package (not local file imports) to have a more "real-life" scenario

**Result**

Also introduces the CJS test as present in react instantsearch hooks for more consistency


FabienMotte (approved):

FabienMotte:
Looks like bundlesize is stuck, not sure why

PR #5177, opened 2022-12-08
fix: simple state mapping `routeToState` function throw an error when state is `null`

FabienMotte:
#### Context

This PR fixes an issue related to the routing when using React InstantSearch Hooks with React Router.

[FX-1925](https://algolia.atlassian.net/browse/FX-1925)

#### Error reproduction

Using the [default state mapping function](https://github.com/algolia/instantsearch.js/blob/master/packages/instantsearch.js/src/lib/stateMappings/simple.ts), an error is thrown only on back navigation from a refined state, when reaching the initial / unrefined state:

```text
defaultStateMapping.ts:10 Uncaught TypeError: Cannot destructure property 'configure' of 'uiState' as it is null.
    at getIndexStateWithoutConfigure (defaultStateMapping.ts:10:11)
    at defaultStateMapping.ts:36:24
    at Array.reduce (<anonymous>)
    at Object.routeToState (defaultStateMapping.ts:32:38)
    at createRouterMiddleware.js:45:57
    at _onPopState (history.js:140:11)
```

CodeSandbox error reproduction: https://codesandbox.io/s/rish-with-react-router-simple-state-mapping-error-k6xqkn

Video reproduction of the error:

https://user-images.githubusercontent.com/662153/206485747-ef4b2045-c5ac-470b-9799-45063a3d7216.mov

#### Cause of the issue

When you navigate to the `/search` route, React Router [push some history state data](https://github.com/remix-run/react-router/blob/b67538b1172c3dcf07a2e431bbe7c2eb32b06eaf/packages/router/history.ts#L591) along with the new URL:
```js
{ usr: null, key: '4ykrr3bq' }
```

In the simple state mapping function, in the `routeToState` function, we reduce the history state assuming the key is the `indexId` and the value is the state. But using React Router, the first key is `usr`, so the corresponding value `null` is passed to the `getIndexStateWithoutConfigure` function.

https://github.com/algolia/instantsearch.js/blob/8c68f37364f4199b9be76fa7ddd3fc1aea596d6a/packages/instantsearch.js/src/lib/stateMappings/simple.ts#L27-L35

This function then tries to deconstruct the `configure` object from the `uiState` but since it's `null`, it throws an error.

https://github.com/algolia/instantsearch.js/blob/8c68f37364f4199b9be76fa7ddd3fc1aea596d6a/packages/instantsearch.js/src/lib/stateMappings/simple.ts#L3-L8

### Result

We always read the state from the URL because the state of the history can be incorect in some cases (e.g. using React Router) so no more error is thrown.


Haroenv:
It’s an interesting solution, and we probably should do this, but it makes me think again that the branch we have in onPopState of history router should probably ignore the history.state but always .read instead. It’s not so expensive and always going to be right (unlike history state that can be messed with)

Haroenv (changes requested):
this could be useful, but I think we should fix the read function instead. Are you taking that up @FabienMotte ?

FabienMotte:
> this could be useful, but I think we should fix the read function instead. Are you taking that up @FabienMotte ?

I updated the fix, test and the PR description (results section). WDYT?

Haroenv (approved):

PR #5179, opened 2022-12-12
chore(instantsearch.css): migrate `algolia/instantsearch-specs` to monorepo

sarahdayan:
## Summary

This PR migrates the [`algolia/instantsearch-specs`](https://github.com/algolia/instantsearch-specs) project to the InstantSearch monorepo.

## Notable changes

- The package is renamed `instantsearch.css` to match the npm package
- The `specs`site is moved to `website` (see [preview](https://deploy-preview-5179--instantsearchjs.netlify.app/specs/))
- The README is rewritten to follow the same style as the `instantsearch.js` one
- Commits are rewritten to reference pull requests from [`algolia/instantsearch-specs`](https://github.com/algolia/instantsearch-specs)

## Next steps

- Install and set up [Stylelint](https://stylelint.io/) to enforce quality rules and facilitate contributions from feature teams (in another PR to avoid cluttering this one with linting updates)
- Ensure that we're able to properly infer next versions _per package_ (e.g., `instantsearch-specs` recently got a [breaking change](https://github.com/algolia/instantsearch-specs/pull/125) but wasn't released yet)
- Decide how to consolidate all different sites (stories, examples, specs, etc., likely once all packages are migrated)
- Harmonize and/or consolidate READMEs, `.gitignore`, `.npmignore`, etc. (likely once all packages are migrated)

FX-2049


packages/instantsearch.css/netlify.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+  publish = "dist/"
```
Haroenv, line 2: does this work?

packages/instantsearch.css/astro.config.mjs
```diff
@@ -0,0 +1,5 @@
+import { defineConfig } from "astro/config";
```
Haroenv, line 1: could be done in a separate PR, but instantsearch.css should be only the css files, and the specs should be its own package (or folder in website)

Haroenv:
one thing that would still be interesting to do is change all the references to issues and pull requests in commits could be renamed from `#xxx` to `algolia/instantsearch-specs#xxx`

packages/instantsearch.css/netlify.toml
sarahdayan: The PR is still a WIP.

packages/instantsearch.css/astro.config.mjs
sarahdayan: The PR is still a WIP.

website/src/specs/src/main.js
```diff
@@ -0,0 +1,95 @@
...
+// Hover list of classes to highligh them
+
+function highlightMatches(item) {
+  const queue = [document.getElementById('js-pre')];
+  const word = item.getAttribute('data-value').replace(/\./g, '');
+  let curr;
+
+  while ((curr = queue.pop())) {
+    if (curr.textContent.match(word)) {
+      // eslint-disable-next-line no-loop-func
```
sarahdayan, line 35: Properly fixing this ESLint violation would mean refactoring this code, which isn't worth the effort right now.

Let's decide what to do with all the different sites, then we can clean up this code and test it.

website/src/specs/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "specs",
```
Haroenv, line 2:
```suggestion
  "name": "@instantsearch/specs-website",
```

not important, as this package isn't installed, but best to choose names that don't mirror public packages maybe?

website/src/specs/yarn.lock
```diff
@@ -0,0 +1,3468 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Haroenv, line 1: this isn't in the workspaces, so it gets its own yarn.lock and won't automatically be installed. To fix, add to root package.json#workspaces and remove yarn lock

website/src/specs/src/styles/index.scss
```diff
@@ -0,0 +1,336 @@
+// Quick and dirty layout for the PoC :]
+
```
dhayab, line 2: We can probably remove this now.

```suggestion
```

website/src/specs/yarn.lock
sarahdayan: Yes it's much better and requires less clumsy scripts 👍

Astro can [only use stylesheets from the public directory (or absolute links) when using `<link>` tags](https://docs.astro.build/en/guides/styling/#load-a-static-stylesheet-via-link-tags) so I'm using [`prepare`](https://docs.npmjs.com/cli/v9/using-npm/scripts#prepare-and-prepublish) scripts to build the library, then copy the themes in `specs/public/`.

dhayab (approved):
I love how the monorepo is shaping up! Tested locally, DX for specs website authoring is great and building works fine for all packages.

FabienMotte (approved):
Well done 👏

specs/package.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "license": "MIT",
+  "author": "Algolia <support@algolia.com>",
+  "scripts": {
+    "build": "astro build",
+    "check": "astro check && tsc",
+    "dev": "astro dev",
+    "preview": "astro preview",
+    "start": "astro dev",
+    "prepare": "rm -rf public && mkdir public && cp -r ../node_modules/instantsearch.css/themes public/themes"
```
Haroenv, line 18: wonder if this can be done with node_modules directly without the prepare script, but can be done after merge (ore never lol)

Haroenv (approved):

specs/src/main.js
```diff
@@ -0,0 +1,95 @@
...
+  changeThemeSelect.value =
+    changeThemeSelect.options[localStorageSelectChoice].value;
+  changeThemeSelect.addEventListener('change', (e) => {
+    customThemeStyleSheet.setAttribute(
+      'href',
+      changeThemeSelect.options[e.target.options.selectedIndex].value
+    );
+    localStorage.setItem('selectChoice', e.target.options.selectedIndex);
+  });
+}
```
Haroenv, line 24: the style is persisted in localStorage, and it displays it correctly in select, but I think the beginning of this block is wrong, and still chooses the default value.

comment is non-blocking as this is already the case since move to astro, probably earlier too

specs/package.json
sarahdayan: In Astro you need `<link>` tags to reference either an absolute URL or a relative from the `public/` directory, so it's difficult. Imports would fix this but it's difficult to manage switching stylesheets with static imports and the side-effect nature of CSS.

PR #5182, opened 2022-12-13
chore: migrate `create-instantsearch-app` to monorepo

aymeric-giraudet:
**Summary**

Didn't have to change any code in CISA itself, just some changes to how we run tests in the monorepo, some configs to adapt as well.

**Result**

- [X] Tests, lint and type checking work and are ran from the root
- [X] Updated all templates to reference `instantsearch.js` and proper new path of `create-instantsearch-app`
- [x] Created `templates` branch and updated GitHub action
- [X] Updated contributing guide with how to replace all #xxx in reference to original repo name


packages/instantsearch.js/tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
```
Haroenv, line 1: what tool is this one needed for?

packages/create-instantsearch-app/scripts/release-templates.js
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env node
+
+/*
+ * This script releases compiled templates to the branch `templates`.
+ * This branch is used for CodeSandbox.
+ * Example: https://codesandbox.io/s/github/algolia/create-instantsearch-app/tree/templates/instantsearch.js
```
Haroenv, line 6:
```suggestion
 * Example: https://codesandbox.io/s/github/algolia/instantsearch.js/tree/templates/instantsearch.js
```

packages/instantsearch.js/jest.config.js
```diff
@@ -0,0 +1,5 @@
+/* eslint-disable import/no-commonjs */
+module.exports = {
```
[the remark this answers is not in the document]
aymeric-giraudet: Removed it :)

packages/instantsearch.js/tsconfig.json
aymeric-giraudet: Just forgot to remove this while I was experimenting haha

.github/workflows/release-templates.yml
```diff
@@ -0,0 +1,39 @@
+name: release-templates
+
+on:
+  push:
+    tags:
+      - '*'
```
Haroenv, line 6: will this happen multiple times per release you think?

Haroenv (approved):

.github/workflows/release-templates.yml
aymeric-giraudet: Just made a test there : https://github.com/aymeric-giraudet/test-trigger-tags/actions
It will run for each tag even if they refer to the same commit indeed :/ 
So if we release some packages simultaneously then yeah, we probably have to narrow it down to certain packages then.
Which do you think it should be ? I guess `instantsearch.js@*`, `react-instantsearch-hooks@*` `vue-instantsearch@*` ?

Haroenv: just `instantsearch.js@*` and Create InstantSearch App too I think?

packages/create-instantsearch-app/docs/custom-templates.md
```diff
@@ -0,0 +1,161 @@
+# Creating a custom template
```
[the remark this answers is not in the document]
aymeric-giraudet: It has more lines than what's in the README though :/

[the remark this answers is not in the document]
Haroenv: I think having it in docs is fine, the readme links to it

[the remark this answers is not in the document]
sarahdayan: In this case we'll need to do a final pass to harmonize docs because we'll end up having them everywhere. At the very least we'll need an index.

PR #5183, opened 2022-12-14
chore(bundling): don't require "version" env variable

Haroenv:
Fixes the build script in prod mode for InstantSearch.js


aymeric-giraudet (approved):

package.json
```diff
@@ -86,7 +86,6 @@
     "rollup-plugin-node-resolve": "5.2.0",
     "rollup-plugin-replace": "2.2.0",
     "rollup-plugin-uglify": "6.0.4",
-    "semver": "6.3.0",
```
Haroenv, line 89: something else still depends on it, so not removed from lockfile

PR #5194, opened 2022-12-15
feat(monorepo): introduce react-based InstantSearch flavours to the codebase

Haroenv:
what i did:
- change configs (jest, babel, eslint, typescript) to be the same between packages
  - do some related linting/typing/test changes in InstantSearch.js
- choose best between custom babel plugins and test utils
- update extension-resolver babel plugin to work for local "absolute" packages
- enable e2e tests for react and react-hooks (new)
- rewrite commits with correct issue/pr links
- add script to ensure versions stay synced in React InstantSearch 

what i didn't do:
- consolidate between tests/mock and tests/mocks (there's an overlap, but required too much changes in tests for now - dependencies, types, etc.)
- import from `@instantsearch/testutils` (similarly complicated babel, tests ...  changes. Can also be done later)
- fix build output of react-instantsearch pacakge (already the case before migration)


.codesandbox/ci.json
```diff
@@ -1,6 +1,10 @@
 {
-  "sandboxes": ["instantsearchjs-es-template-pcw1k"],
-  "buildCommand": "build --no-private",
-  "packages": ["packages/instantsearch.js"],
+  "sandboxes": [
+    "instantsearchjs-es-template-pcw1k",
+    "github/algolia/create-instantsearch-app/tree/templates/react-instantsearch",
```
Haroenv, line 4:
```suggestion
    "github/algolia/instantsearch.js/tree/templates/react-instantsearch",
```

.github/ISSUE_TEMPLATE/Bug_report_Hooks.md
```diff
@@ -0,0 +1,42 @@
+---
```
Haroenv, line 1: to be removed

jest.config.js
```diff
@@ -25,7 +27,13 @@ const config = {
...
     'jest-watch-typeahead/testname',
   ],
-  transformIgnorePatterns: ['node_modules/(?!(search-insights)/)'],
+  transformIgnorePatterns: [
+    'node_modules/(?!(search-insights|instantsearch.js)/)',
+  ],
+  moduleNameMapper: {
+    '^react-instantsearch-(.*)$':
+      '<rootDir>/packages/react-instantsearch-$1/src/',
+  },
```
Haroenv, line 36: add instantsearch.js to modulenamemapper

.circleci/config.yml
```diff
@@ -101,6 +101,14 @@ jobs:
             - packages/instantsearch.js/es
             - packages/instantsearch.js/cjs
             - packages/instantsearch.js/dist
+            - packages/react-instantsearch/dist
```
sarahdayan, line 104: Unrelated to this PR, but `yarn run build --ignore='example-*'` in l. 91 could be changed into `yarn run build:ci` just like we have a `test:ci` script.

It's easier to follow when you're trying to debug CI scripts locally.

CONTRIBUTING.md
```diff
@@ -125,34 +125,44 @@ Go to <http://localhost:6006> for the widget playground.
...
   ▸ connectors/       << The source of all the connectors driving the widgets' logic
   ▸ helpers/          << The source of the method helpers
   ▸ lib/              << The core of the library (InstantSearch, routers, etc.)
   ▸ middleware/       << The source of the middleware
   ▸ types/            << The TypeScript declarations
   ▸ widgets/          << The source of the widgets
+▸ .storybook/                      << The storybook configuration
+▸ stories/                         << The widget stories
+▸ test/                            << Utilities for the tests
 ```
```
sarahdayan, line 166: Should this remain? We probably won't have it for all packages.

sarahdayan (commented):
How do we ensure `*-hooks` packages use parallel versioning?

CONTRIBUTING.md
Haroenv: It's equal for me, i guess it doesn't bring much value, but on the other hand, we want to make it as approachable as possible?

Haroenv:
> How do we ensure *-hooks packages use parallel versioning?

As they depend on each other, the version number should be chosen in parallel, although I believe it's possible for a feature in only hooks-web for example requiring a manual intervention for hooks-web. Not sure if you imagine a solution without manual work?

sarahdayan:
> As they depend on each other, the version number should be chosen in parallel

A change in `react-instantsearch-hooks-server` wouldn't result in an automatic bump in the others because no package depends on it.

> Not sure if you imagine a solution without manual work?

We could write a script that checks all `package.json` of the Hooks packages and throws if the `version`s aren't the same?

CONTRIBUTING.md
sarahdayan: If we keep it, then let's add it for React InstantSearch packages?

package.json
```diff
@@ -11,9 +12,10 @@
   ],
   "scripts": {
     "build": "lerna run build",
+    "build:ci": "lerna run build --ignore='example-*'",
     "doctoc": "doctoc --no-title --maxlevel 3 README.md CONTRIBUTING.md packages/*/README.md",
     "website:storybook": "lerna run storybook:build",
```
dhayab, line 17: Ran into this while testing locally. It appears that if the output directory (`website/stories`) doesn't exist before running the script, it will sometimes generate only one storybook, with the other one failing ([see log here](https://app.netlify.com/sites/instantsearchjs/deploys/63a1c74333215b000b787069#L624)).

Having an existing directory first seems to do the trick.

```suggestion
    "website:storybook": "mkdir -p website/stories && lerna run storybook:build",
```

Haroenv: Hm, I tried deleting the folder and it worked for me, I guess it's a race condition where it thinks it should create the folder but actually shouldn't.

Also odd there's an error but netlify didn't fail?

tests/versions/index.js
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env node
+/* eslint-disable no-process-exit, no-console */
+
+// This file does not use any dependencies, so that it can be ran before installing
+
+// It checks whether the versions of packages that should be versioned synchronously
+// are actually in sync. We need this as long as Lerna doesn't have a mixed mode.
+// see: https://github.com/lerna/lerna/issues/1159
```
Haroenv, line 8: Here's the script to check version numbers are sync @sarahdayan

sarahdayan: Beautiful!

sarahdayan (approved):
2fQ1Gq3KOpvNs4NTmu

dhayab (approved):
Amazing 👏

FabienMotte (approved):
Congrats 👏

PR #5200, opened 2022-12-16
style(instantsearch.css): install and set up Stylelint

sarahdayan:
## Summary

This PR introduces Stylelint, a popular and full-fetaured CSS linter (think of it as ESLint for CSS). Now that `instantsearch.css` is in the monorepo and will be updated by feature teams, it will be useful to have it to verify that our CSS follows agreed-upon quality guidelines.

### Configuration

The `.stylelintrc.json` file describes the current configurations, plugons, and rules we use. I installed the recommended configurations from the Stylelint docs, and added some arbitrary ones that we already had experimented with in Unified InstantSearch (see [config](https://github.com/algolia/unified-instantsearch-ecommerce/blob/master/.stylelintrc.json)).

- `stylelint-no-unsupported-browser-features`: ensures we don't write CSS that's incompatible with our target browsers
- `stylelint-order`: enforces order of declarations within a ruleset, similar to [`eslint-plugin-import-order`](https://www.npmjs.com/package/eslint-plugin-import-order)

### Next steps

I purposefully didn't lint in the same PR because that's going to generate a lot of noise that will be hard to review. We should proceed methodically to keep the diffs readable and avoid regressions. We can do this later, as Stylelint isn't currently run in the CI.

To lint locally, run `yarn workspace instantsearch.css lint`.

We can also fine tune our rules later on as we reduce the existing violations and better understand where problems remain.


packages/instantsearch.css/.stylelintrc.json
```diff
@@ -0,0 +1,25 @@
...
+    "selector-pseudo-element-no-unknown": [
+      true,
+      {
+        "ignorePseudoElements": ["input-placeholder"]
+      }
+    ],
+    "plugin/no-unsupported-browser-features": [
+      true,
+      {
+        "browsers": ["defaults"],
```
sarahdayan, line 20: We don't have one now, but later on we can use a `.browserlistrc` file to share between Stylelint, Babel, etc.

packages/instantsearch.css/package.json
```diff
@@ -21,12 +21,22 @@
     "styles:autoprefixer": "postcss themes/ --use autoprefixer --replace --no-map",
     "styles:minify": "sass src/themes/algolia.scss:themes/algolia-min.css src/themes/satellite.scss:themes/satellite-min.css src/themes/reset.scss:themes/reset-min.css --no-source-map --style compressed",
     "styles:build": "sass src/themes:themes --no-source-map",
-    "prepare": "yarn build"
+    "prepare": "yarn build",
+    "lint": "stylelint \"**/*.scss\""
```
sarahdayan, line 25: When we have taken care of all the blocking violations, we can add `&& lerna run lint` to the `lint` command of our main `package.json`.

Haroenv (commented):
how do we ensure that the tests will "at some point"  pass if it's not enforced on CI?

sarahdayan:
> how do we ensure that the tests will "at some point" pass if it's not enforced on CI?

https://github.com/algolia/instantsearch.js/pull/5200#discussion_r1050860504

Until then we can run the command I provided in the description to iteratively fix existing issues. Then, we can activate the check in the CI.

packages/instantsearch.css/.stylelintrc.json
dhayab: We should explicitly declare Internet Explorer 11 support as `defaults` doesn't include it.

```suggestion
        "browsers": ["defaults", "ie 11"],
```

dhayab (approved):

FabienMotte (approved):

Haroenv (approved):
I'd like to see this added to CI (but not required) so improvement can be seen soon

aymeric-giraudet (approved):

sarahdayan:
@Haroenv Your wish is my command ✨ 6959e3002

package.json
```diff
@@ -14,7 +14,7 @@
     "doctoc": "doctoc --no-title --maxlevel 3 README.md CONTRIBUTING.md packages/*/README.md",
     "website:storybook": "lerna run storybook:build",
     "website:examples": "lerna run website:examples",
-    "lint": "eslint --ext .js,.ts,.tsx .",
+    "lint": "eslint --ext .js,.ts,.tsx . && lerna run lint",
```
Haroenv, line 17: this is correct, but Create InstantSearch App lint should probably be removed, I have another stashed change for that

Haroenv: https://github.com/algolia/instantsearch.js/pull/5217

sarahdayan: I approved your PR, I'll let you merge it then rebase this one so it passes.

PR #5202, opened 2022-12-16
chore(instantsearch.css): remove unused class

sarahdayan:
## Summary

This removes the `.ais-ToggleRefinement-count` CSS class from our themes and examples. This class was removed a while ago from the `toggleRefinement` markup but somehow wasn't removed from the specs and themes.

This is technically a breaking change to remove those from the themes because users could have manually added an element with `class="ais-ToggleRefinement-count"` and rely on the styles. However, since `instantsearch.css` already received a breaking change and hasn't been released, this should be fine to embark this change with it.

Related: https://github.com/algolia/instantsearch-specs/issues/93

### DIFF
--- a/examples/js/e-commerce/src/app.css
+++ b/examples/js/e-commerce/src/app.css
@@ -318,10 +318,6 @@ mark {
   content: 'Yes';
 }
 
-.ais-ToggleRefinement-count {
-  display: none;
-}
-
 /* RatingMenu */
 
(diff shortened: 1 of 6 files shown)


aymeric-giraudet (approved):

FabienMotte (approved):

Haroenv (changes requested):
Vue InstantSearch has a count on toggleRefinement still

sarahdayan:
Ah, damn! Closing this then.

PR #5224, opened 2022-12-21
chore(dependencies): deduplicate devDependencies

Haroenv:
We have some duplicate dependencies since the migrations. This PR decreases that from a total of 3.5 to 3.1 GB, as it's fairly hard to find out what's duplicate and what will have the most impact.

What I've done:
- harmonise `@babel/core`
- harmonise `prop-types`
- moved webpack dependency for example building to its own "package"
- harmonise webpack versions using resolutions


scripts/bundlesize.config.json
```diff
@@ -0,0 +1,68 @@
+{
```
sarahdayan, line 1: This isn't a script, why move it in `scripts`? If it was at the root we wouldn't need to specify the path in `package.json`.

Haroenv: I didn't want a very big root directory, but I guess that can be fixed after

sarahdayan (approved):
Looks legit 👍

FabienMotte (approved):

PR #5229, opened 2022-12-21
items can be null in getWidgetRenderState

Yonn-Trimoreau:
I've installed your latest Magento module and found that in certain situations (which I could not clearly identify), this variable items can be null, causing an exception when trying to filter on a sub category.

When checking the nullity of this variable as I've done in the PR, the issue is fixed.

Here is a screenshot of the JS error when trying to filter "Lave-linge" at this url https://www.gacd.fr/amenagement/cuisine-electromenager/gros-electromenager.html

(I will fix this issue by patching in some way so the error won't be visible for long)

image

### DIFF
--- a/packages/instantsearch.js/src/connectors/menu/connectMenu.ts
+++ b/packages/instantsearch.js/src/connectors/menu/connectMenu.ts
@@ -316,7 +316,7 @@ const connectMenu: MenuConnector = function connectMenu(
           createURL: _createURL,
           refine: _refine,
           sendEvent,
-          canRefine: items.length > 0,
+          canRefine: items ? (items.length > 0) : false,
           widgetParams,
           isShowingMore,
           toggleShowMore: cachedToggleShowMore,


Haroenv:
I believe the magento plugin may be outdated in its InstantSearch version, did you isolate this issue already? It would be interesting to see in which case `items` can be null

Yonn-Trimoreau:
I have the latest Magento 2 module installed (3.9.1), and the only file I have access to in this module is the "algoliaBundle.min.js" which is, you guessed it, minified, so debugging is quite hard, and I don't know how to find the non minified algoliaBundle.js to proceed debugging.

I only know that the file mentions a version 4.5.0 as a header, and I found this bit of code by searching in the whole Algolia's GitHub organisation by text (took me some time already).

And I've manually changed this bit in my local version of the minified file and this works (I clearly don't know how and why).

Haroenv:
I'm trying to reproduce the bug, you mentioned that you patched it on the website, but I don't see the error and it still looks like the code from before the patch is applied:

Screenshot 2022-12-22 at 10 03 45

Thanks!

Haroenv:
Thanks again for your PR. Because we can't reproduce this, and I don't see how it can happen (transformItems return null maybe? but then it shouldn't work anyway because of other reasons) I'll close this PR. Feel free to open again with a test or continue with the patch on your website. 

Thanks!

PR #5230, opened 2022-12-22
ci(shipjs): also push tags of non-published packages 

Haroenv:
**Summary**

These tags get used to determine the next version number, and as we can't get rid of version number (yarn wouldn't include the package into the dependencies otherwise), that's required to be correct.

This also fixes a little error in one example package accidentally set to public

**Result**

every package will get a git tag, to make sure the next version number will be correct.

### DIFF
--- a/ship.config.js
+++ b/ship.config.js
@@ -21,6 +21,8 @@ module.exports = {
     return {
       // This version number is used for shouldPrepare.
       // for it to be useful, we need to chang shipjs to accept an array of nextVersions
+      // @TODO: update this to packages.map(package => package.version) once it's supported in shipjs
+      // see https://github.com/algolia/shipjs/issues/986
       nextVersion: packages[0].version,
     };
   },
--- a/ship.config.js
+++ b/ship.config.js
@@ -1,17 +1,13 @@
 const shell = require('shelljs');
 
 const packages = JSON.parse(
-  shell.exec('yarn run --silent lerna list --toposort --json --no-private', {
+  shell.exec('yarn run --silent lerna list --toposort --json', {
     silent: true,
   })
 );
 
 module.exports = {
-  shouldPrepare: ({ releaseType, commitNumbersPerType }) => {
-    const { fix = 0 } = commitNumbersPerType;
-    if (releaseType === 'patch' && fix === 0) {
-      return false;
-    }
+  shouldPrepare: () => {
     return true;
   },
   getTagName: () =>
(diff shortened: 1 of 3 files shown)


ship.config.js
```diff
@@ -23,7 +19,8 @@ module.exports = {
     );
 
     return {
-      // This is used for shouldPrepare
+      // This version number is used for shouldPrepare.
+      // for it to be useful, we need to chang shipjs to accept an array of nextVersions
```
sarahdayan, line 23: Should we add a `@TODO`?

ship.config.js
```diff
@@ -1,17 +1,13 @@
...
   })
 );
 
 module.exports = {
-  shouldPrepare: ({ releaseType, commitNumbersPerType }) => {
-    const { fix = 0 } = commitNumbersPerType;
-    if (releaseType === 'patch' && fix === 0) {
-      return false;
-    }
+  shouldPrepare: () => {
```
sarahdayan, line 10: Why always return `true`?

Haroenv: Because the package returned from version possibly doesn't have any relevant pushes, while other packages would have stuff to release. Therefore I think it's best if we get the pr every two weeks and choose to close if it's irrelevant

ship.config.js
Haroenv: the todo is more in shipjs than here, so not sure if it's here it should be marked? can add it if you think it's useful of course

sarahdayan: Then that would be two of them:
- One in Ship.js for the change
- One here to express that we need to make a change once we update to Ship.js x.y.z

Can be done separately ofc.

ship.config.js
sarahdayan: I think we need to document this process so we're all fully aware of it and can manage it. For now it's blurry to me how things would go and what are the different possible scenarios.

Also, we need to better handle this for when we have feature teams contribute, we can't be the only ones on duty of merging release PRs in the long run.

ship.config.js
Haroenv:
```suggestion
      // This version number is used for shouldPrepare.
      // for it to be useful, we need to chang shipjs to accept an array of nextVersions
      // @TODO: update this to packages.map(package => package.version) once it's supported in shipjs
      // see https://github.com/algolia/shipjs/issues/986
```

ship.config.js
Haroenv: Is what's described in https://github.com/algolia/shipjs/issues/986 sufficient? For now if a release is prepared, but none of the packages requires that update, shipjs will still prepare the release, possibly doing a useless release (not that big of a deal) that we can manually cancel if we don't think it's needed to release just yet, similarly as already happens from time to time today when the only changes are minor things that don't need to be released

sarahdayan: Got It. I created a [Jira ticket](https://algolia.atlassian.net/browse/FX-2162) so we handle this quickly, I'm not fond of such a manual process to last too long.

sarahdayan (approved):

PR #5232, opened 2022-12-22
test(jest): avoid flaky modules test

Haroenv:
The modules-* test about dependency detection of react-dom in react-instantsearch-hooks is flaky with regular jest workers, as I believe sometimes the jest mock of a module becomes global and still relevant for the next test.

Ideally we wouldn't need to do such tests when react-dom is exclusively user-provided, but for now hopefully running the tests in band will avoid that test being flaky.

test plan: run the unit tests a couple of times on CI and see it pass consistently


Haroenv:
too bad, this doesn't work :(

PR #5368, opened 2022-12-30
chore(examples): correct peer dependencies

Haroenv:
Small changes in examples:

- babel file will skip the constant elements transform in both webpack and parcel (both behave wrong)
- correct peer dependencies for parcel deps

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -10,7 +10,6 @@ const clean = (x) => x.filter(Boolean);
 
 module.exports = (api) => {
   const isTest = api.env('test');
-  const isWebpack = api.env('webpack');
   const modules = isTest || isCJS ? 'commonjs' : false;
   const targets = {};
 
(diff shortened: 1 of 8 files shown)


babel.config.js
```diff
@@ -27,7 +26,8 @@ module.exports = (api) => {
 
   const buildPlugins = clean([
     '@babel/plugin-proposal-class-properties',
-    !isWebpack && '@babel/plugin-transform-react-constant-elements',
+    (isCJS || isES || isUMD || isRollup) &&
+      '@babel/plugin-transform-react-constant-elements',
```
dhayab, line 30: This seems to cover all conditions. When is this plugin skipped?

Haroenv: it's skipped for webpack and parcel :)

dhayab (approved):

PR #5369, opened 2022-12-30
refactor(examples): replace react-scripts with parcel

Haroenv:
the main reason is that react-scripts version here is quite outdated, and causing some duplications and useless packages in our lockfile/node_modules.

- Some rewrites in the babel config as parcel warned for it incrementally
- kept babel-preset-react-app for react storybook, but moved it to the relevant package requiring it.


FabienMotte (approved):

Haroenv:
Something about this PR makes the E2E tests consistently fail on chrome and Firefox for React InstantSearch Hooks (not on IE), but I can't find what the cause would be, it loads fine on my side 🤔

PR #5399, opened 2022-12-30
chore(deps): deduplicate dependencies by using caret

Haroenv:
While we decided before that tools should be using exact dependencies, this actually is unsustainable, as manually going through each update and finding what's the most correct takes ages, and thus causes duplication when something new gets added.

This PR does the following:
- update all devDeps to ^ (deps for examples)
- run yarn-deduplicate to remove no longer needed versions

This brings node_modules from 3.2 to 2.9 GB, so I think that would be worth it.

More optimisations can be done later by selectively updating dependencies to make them match even more other dependencies.


Haroenv:
I'll redo this once the other PR is done instead of rebasing

PR #5402, opened 2023-01-02
docs(storybook): move dependencies to ensure right binary

Haroenv:
**Summary**

Otherwise vue storybook tries to build react etc.

**Result**

react storybook works now


dhayab (commented):
👍

PR #5403, opened 2023-01-02
fix(vue-instantsearch): compute initial state from helper instead of results

dhayab:
**Summary**

Currently, for SSR, we compute `initialResults` on the server using the helper results. This is used for initial rendering as well as for hydration. Because nested indices inherit from their parent, they are set with a merged state which prevent correct inheritance behavior later on, when interacting with InstantSearch client-side.

This PR instead relies on the index's helper "request" state, which isn't merged and allows expected behavior.

[CR-2516](https://algolia.atlassian.net/browse/CR-2516)
fixes https://github.com/algolia/instantsearch/issues/5383

### DIFF
--- a/examples/vue/nuxt/package.json
+++ b/examples/vue/nuxt/package.json
@@ -13,6 +13,7 @@
   "dependencies": {
     "algoliasearch": "4.14.3",
     "cross-env": "^5.2.0",
+    "css-loader": "^4.3.0",
     "nuxt": "^2.4.5",
     "vue-instantsearch": "4.7.0",
     "vue-server-renderer": "2.7.14"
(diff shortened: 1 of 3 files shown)


examples/vue/nuxt/package.json
```diff
@@ -13,6 +13,7 @@
   "dependencies": {
     "algoliasearch": "4.14.3",
     "cross-env": "^5.2.0",
+    "css-loader": "^4.3.0",
```
dhayab, line 16: Without this line, the example doesn't run (probably due to it being in a monorepo now). I set the version to one that is already in yarn.lock.

Haroenv (approved):
I believe this option was discussed in https://github.com/algolia/vue-instantsearch/pull/764 too, but it's correct like this. Thanks!

PR #5408, opened 2023-01-03
fix(url): add widgets early always (WIP)

Haroenv:
**Summary**

see: https://github.com/algolia/instantsearch/issues/5236

This should fix it in theory, although i can only reproduce this in real-life situations so far. Tests aren't yet adjusted.

As shouldSsr is set to true in connectors, it probably also be set to true in useConnector, and therefore probably just set to true always. It is slightly more expensive as the widget is checked in the list first.

sandbox showing this: https://codesandbox.io/s/silly-sanne-8q6tqt

TO DO:

- [ ] check whether this is the correct solution
- [ ] update tests
- [ ] consider whether shouldSsr should be true on useIndex
- [ ] consider whether shouldSsr logic is needed
- [ ] consider whether "widget already added" logic is needed

**Result**

As widgets get added before the createURL for example is calculated, this means all widgets are already added when anything is rendered.

### DIFF
--- a/packages/react-instantsearch-hooks/src/hooks/useConnector.ts
+++ b/packages/react-instantsearch-hooks/src/hooks/useConnector.ts
@@ -4,7 +4,6 @@ import { dequal } from '../lib/dequal';
 import { getIndexSearchResults } from '../lib/getIndexSearchResults';
 import { useIndexContext } from '../lib/useIndexContext';
 import { useInstantSearchContext } from '../lib/useInstantSearchContext';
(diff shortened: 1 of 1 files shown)


Haroenv:
😢 this doesn't work when the widgets aren't "in the right order", see for example <https://8q6tqt.sse.codesandbox.io/?instant_search%5Bquery%5D=something>. Because the searchbox isn't yet added when the pagination renders, its createURL doesn't have the query in it because of that

PR #5411, opened 2023-01-04
ci: run e2e job on bigger environment

dhayab:
This PR uses a larger docker environment to run the e2e tests job. While it still takes a while to run (which could be attributed in part to our saucelabs opensource plan), the tests seem to run more reliably (no retries) and the CPU usage doesn't hit 100%.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -213,6 +213,7 @@ jobs:
 
   e2e tests:
     <<: *defaults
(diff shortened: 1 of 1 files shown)


Haroenv (approved):
if there's even a hint of it retrying less, we should do this (as retrying is super slow). Numbers are pretty much the same as master though if I understand it correctly

PR #5415, opened 2023-01-06
ci: use environment variables from context

dhayab:
This PR updates the CircleCI config file so that jobs retrieve relevant environment variables from context. Doing this will allow us to use the same tokens in similar projects.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,8 @@ workflows:
   version: 2
   ci:
     jobs:
-      - build
+      - build:
+          context: fx-libraries
       - test metadata
       - lint:
           requires:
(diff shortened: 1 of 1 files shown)


aymeric-giraudet (approved):

.circleci/config.yml
```diff
@@ -36,7 +36,8 @@ workflows:
   version: 2
   ci:
     jobs:
-      - build
+      - build:
+          context: fx-libraries
```
Haroenv, line 40: every job will need the context, otherwise it can't clone I think

dhayab: It seems the r/o ssh key circleci uses to clone the repo does not exist in the github repo settings. I'm looking into it.

Haroenv (approved):

PR #5426, opened 2023-01-11
ci(eslint): report eslint tests via junit output

Haroenv:
**Summary**

enables junit reporting for global eslint.

**Result**

lint failures will show in circleci "tests" tab


Haroenv:
Hmm, we shouldn't do this while we still have warnings, as it's impossible to see the difference between an error and a warning :(

PR #5425, opened 2023-01-10
feat(react-instantsearch-hooks): Add useStats hook

cdhawke:
**Summary**

Add a hook for `useStats` to `react-instantsearch-hooks`.

The goal is to be able to use use this hook to provide a widget, as well as replace the 'not implemented' documentation: https://www.algolia.com/doc/api-reference/widgets/stats/react-hooks/#about

**Result**

Review added test for basic functionality / expectation.

**Next Steps**
- Add a widget to `react-instantsearch-hooks`
- Update examples
- Update documentation

PR #5427, opened 2023-01-11
feat(react-instantsearch-hooks-web): Add stats widget and ui component

cdhawke:
**Summary**

Similar in motivation to #5425 - add a reusable widget for providing stats.

Next steps:
- ~~Update examples~~ I went ahead and just updated the `example-react-instantsearch-hooks-default-theme` w/ the widget.
- Update documentation

**Results**

image

### DIFF
--- a/examples/react-hooks/default-theme/src/App.tsx
+++ b/examples/react-hooks/default-theme/src/App.tsx
@@ -21,6 +21,7 @@ import {
   SearchBox,
   SortBy,
   ToggleRefinement,
+  Stats,
 } from 'react-instantsearch-hooks-web';
 
 import {
@@ -108,6 +109,9 @@ export function App() {
           />
 
           <SearchBox placeholder="Search" autoFocus />
+          <div className="Stats">
+            <Stats />
+          </div>
 
           <div className="Search-header">
             <PoweredBy />
--- a/examples/react-hooks/default-theme/src/App.tsx
+++ b/examples/react-hooks/default-theme/src/App.tsx
@@ -21,7 +21,6 @@ import {
   SearchBox,
   SortBy,
   ToggleRefinement,
-  Stats,
 } from 'react-instantsearch-hooks-web';
 
 import {
@@ -109,9 +108,6 @@ export function App() {
           />
 
           <SearchBox placeholder="Search" autoFocus />
-          <div className="Stats">
-            <Stats />
-          </div>
 
           <div className="Search-header">
             <PoweredBy />
--- a/packages/react-instantsearch-hooks-web/src/ui/Stats.tsx
+++ b/packages/react-instantsearch-hooks-web/src/ui/Stats.tsx
@@ -2,12 +2,11 @@ import React from 'react';
 
 import { cx } from './lib/cx';
 
-export type StatsTranslationOptions = {
-  nbHits: number;
-  processingTimeMS: number;
-  nbSortedHits?: number;
-  areHitsSorted?: boolean;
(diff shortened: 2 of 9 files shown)


packages/react-instantsearch-hooks-web/src/ui/Stats.tsx
```diff
@@ -0,0 +1,54 @@
...
+      {...props}
+      className={cx('ais-Stats', classNames.root, props.className)}
+    >
+      <span className="ais-Stats-text">
+        {translations.stats(
+          nbHits,
+          processingTimeMS,
+          nbSortedHits,
+          areHitsSorted
+        )}
```
Haroenv, line 50: I'm just doubting whether this may be a lot of parameters for one function. Maybe it should be put in an object, although we'd need to check what's consistent with the other translations.

examples/react-hooks/default-theme/src/App.tsx
```diff
@@ -108,6 +109,9 @@ export function App() {
           />
 
           <SearchBox placeholder="Search" autoFocus />
+          <div className="Stats">
+            <Stats />
```
Haroenv, line 113: while this is great to test it out, we probably should revert it until released, as these examples are linked from eg. codesandbox. If we add something new that's not yet released, it will break.

We probably should change the type of links we use (latest released tag maybe for example), but for now master causes this issue

Haroenv (commented):
overall code looks great! just need to double check translation name & parameters

packages/react-instantsearch-hooks-web/src/ui/Stats.tsx
cdhawke: Can do - I only did it this way because it's the way `-dom` did it.

packages/react-instantsearch-hooks-web/src/widgets/Stats.tsx
```diff
@@ -0,0 +1,47 @@
...
+      : `${options.nbHits.toLocaleString()} results found in ${options.processingTimeMS.toLocaleString()}ms`;
+
+  const uiProps: UiProps = {
+    nbHits,
+    nbSortedHits,
+    processingTimeMS,
+    areHitsSorted,
+    translations: {
+      stats: statsTranslation,
+      ...(props.translations || {}),
```
Haroenv, line 42: I think with this implementation you override the translations twice if they're passed by the user. In other implementations it's like this:

```suggestion
export function Stats({ translations, ...props }: StatsProps) {
  const { nbHits, nbSortedHits, processingTimeMS, areHitsSorted } = useStats(
    undefined,
    { $$widgetType: 'ais.stats' }
  );

  const statsTranslation = (options: StatsTranslationOptions): string =>
    options.areHitsSorted
      ? `${options.nbSortedHits!.toLocaleString()} relevant results sorted out of ${options.nbHits.toLocaleString()} found in ${options.processingTimeMS.toLocaleString()}ms`
      : `${options.nbHits.toLocaleString()} results found in ${options.processingTimeMS.toLocaleString()}ms`;

  const uiProps: UiProps = {
    nbHits,
    nbSortedHits,
    processingTimeMS,
    areHitsSorted,
    translations: {
      stats: statsTranslation,
      ...translations,
```

Haroenv (commented):
👍 small suggestion, otherwise lgtm

dhayab (approved):
Thanks for your contribution!

Haroenv (approved):

PR #5429, opened 2023-01-12
WIP: optimistic rendering

Haroenv:
**Summary**

**Result**

### DIFF
--- a/examples/react-hooks/default-theme/src/App.tsx
+++ b/examples/react-hooks/default-theme/src/App.tsx
@@ -37,6 +37,11 @@ const searchClient = algoliasearch(
   'latency',
   '6be0576ff61c053d5f9a3225e2a90f76'
 );
+const search = searchClient.search;
+searchClient.search = async (queries) => {
+  await new Promise((resolve) => setTimeout(resolve, 1000));
+  return search(queries);
+};
 
 type HitProps = {
   hit: AlgoliaHit<{
@@ -54,7 +59,10 @@ function Hit({ hit }: HitProps) {
   );
 }
 
+const attributes = ['brand', 'price', 'categories'];
+
 export function App() {
+  const [count, setCount] = React.useState(1);
   return (
     <InstantSearch
       searchClient={searchClient}
@@ -65,38 +73,21 @@ export function App() {
 
       <div className="Container">
         <div>
-          <DynamicWidgets>
-            <Panel header="Brands">
-              <RefinementList
-                attribute="brand"
-                searchable={true}
-                searchablePlaceholder="Search brands"
-                showMore={true}
-              />
+          <button onClick={() => setCount(count + 1)}>add</button>
+          {Array.from({ length: count }).map((_, i) => (
+            <Panel header={attributes[i]} key={i}>
+              <RefinementList attribute={attributes[i]} />
             </Panel>
-            <Panel header="Categories">
-              <Menu attribute="categories" showMore={true} />
-            </Panel>
-            <Panel header="Hierarchy">
-              <HierarchicalMenu
-                attributes={[
-                  'hierarchicalCategories.lvl0',
-                  'hierarchicalCategories.lvl1',
-                  'hierarchicalCategories.lvl2',
-                ]}
-                showMore={true}
-              />
-            </Panel>
-            <Panel header="Price">
-              <RangeInput attribute="price" />
-            </Panel>
-            <Panel header="Free Shipping">
-              <ToggleRefinement
-                attribute="free_shipping"
-                label="Free shipping"
-              />
-            </Panel>
-          </DynamicWidgets>
+          ))}
+          <Panel header="hierarchy">
+            <HierarchicalMenu
+              attributes={[
+                'hierarchicalCategories.lvl0',
+                'hierarchicalCategories.lvl1',
+                'hierarchicalCategories.lvl2',
+              ]}
+            />
+          </Panel>
         </div>
         <div className="Search">
           <Breadcrumb
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -487,6 +487,7 @@ See ${createDocumentationLink({
       // these guarantees about the render event, thus emitting it once more
       // isn't a breaking change.
       this.emit('render');
+      this.scheduleRender();
 
       // This solution allows us to keep the exact same API for the users but
       // under the hood, we have a different implementation. It should be
--- a/packages/instantsearch.js/src/lib/__tests__/status.test.ts
+++ b/packages/instantsearch.js/src/lib/__tests__/status.test.ts
@@ -185,7 +185,6 @@ describe('status', () => {
   });
 
   test('lets users render on error with the `render` event', async () => {
-    // expect.assertions(4);
     const search = instantsearch({
       indexName: 'indexName',
       searchClient: createSearchClient({
--- a/packages/instantsearch.js/src/lib/utils/render-args.ts
+++ b/packages/instantsearch.js/src/lib/utils/render-args.ts
@@ -1,5 +1,6 @@
 import type { InstantSearch, UiState } from '../../types';
 import type { IndexWidget } from '../../widgets/index/index';
+import algoliasearchHelper from 'algoliasearch-helper';
 
 export function createInitArgs(
   instantSearchInstance: InstantSearch,
@@ -29,7 +30,12 @@ export function createRenderArgs(
   instantSearchInstance: InstantSearch,
   parent: IndexWidget
 ) {
-  const results = parent.getResults()!;
+  // Always make results out of the current state, this enables optimistic UI
+  // TODO: does this need to be conditional
+  const results = new algoliasearchHelper.SearchResults(
+    parent.getHelper()!.state,
+    parent.getResults()!._rawResults
+  );
 
   return {
     helper: parent.getHelper()!,
--- a/packages/instantsearch.js/src/widgets/index/index.ts
+++ b/packages/instantsearch.js/src/widgets/index/index.ts
@@ -493,16 +493,23 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
       });
 
       derivedHelper.on('result', ({ results }) => {
-        // The index does not render the results it schedules a new render
-        // to let all the other indices emit their own results. It allows us to
-        // run the render process in one pass.
-        instantSearchInstance.scheduleRender();
+        // TODO: does this need to be conditional
+        const actualResults = new algoliasearchHelper.SearchResults(
+          this.getHelper()!.state,
+          results._rawResults
+        );
 
         // the derived helper is the one which actually searches, but the helper
         // which is exposed e.g. via instance.helper, doesn't search, and thus
         // does not have access to lastResults, which it used to in pre-federated
         // search behavior.
-        helper!.lastResults = results;
+        helper!.lastResults = actualResults;
+        derivedHelper!.lastResults = actualResults;
+
+        // The index does not render the results it schedules a new render
+        // to let all the other indices emit their own results. It allows us to
+        // run the render process in one pass.
+        instantSearchInstance.scheduleRender();
       });
 
       // We compute the render state before calling `init` in a separate loop


examples/react-hooks/default-theme/src/App.tsx
```diff
@@ -37,6 +37,11 @@ const searchClient = algoliasearch(
   'latency',
   '6be0576ff61c053d5f9a3225e2a90f76'
 );
+const search = searchClient.search;
+searchClient.search = async (queries) => {
```
Haroenv, line 41: todo: revert this example before merge

packages/instantsearch.js/src/lib/__tests__/status.test.ts
```diff
@@ -185,7 +185,6 @@ describe('status', () => {
   });
 
   test('lets users render on error with the `render` event', async () => {
-    // expect.assertions(4);
```
Haroenv, line 188: this should have been removed in #5127 already, but I didn't see the review then

packages/instantsearch.js/src/lib/utils/createSendEventForHits.ts
```diff
@@ -163,7 +163,7 @@ export function createSendEventForHits({
       index,
       methodName: 'sendEvent',
       args,
-      isSearchStalled: instantSearchInstance.status === 'stalled',
+      invalidStatus: instantSearchInstance.status !== 'idle',
```
Haroenv, line 166: error also shouldn't send event

packages/instantsearch.js/src/lib/utils/render-args.ts
```diff
@@ -1,5 +1,6 @@
 import type { InstantSearch, UiState } from '../../types';
 import type { IndexWidget } from '../../widgets/index/index';
+import algoliasearchHelper from 'algoliasearch-helper';
```
Haroenv, line 3:
```suggestion
```

packages/instantsearch.js/src/lib/utils/render-args.ts
```diff
@@ -30,6 +31,10 @@ export function createRenderArgs(
   parent: IndexWidget
 ) {
   const results = parent.getResults()!;
+  // To make the UI optimistic, we will always render using the current state,
+  // but the previous results. This means a change will be visible immediately,
+  // regardless of the status of the network request.
+  results._state = parent.getHelper()!.state;
```
Haroenv, line 37: this is part A of the implementation

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -481,12 +481,7 @@ See ${createDocumentationLink({
 
     mainHelper.search = () => {
       this.status = 'loading';
-      // @MAJOR: use scheduleRender here
-      // For now, widgets don't expect to be rendered at the start of `loading`,
-      // so it would be a breaking change to add an extra render. We don't have
-      // these guarantees about the render event, thus emitting it once more
-      // isn't a breaking change.
-      this.emit('render');
+      this.scheduleRender(false);
```
Haroenv, line 484: this is part B of the implementation

packages/instantsearch.js/src/lib/utils/render-args.ts
Haroenv, line 37: instead of patching in render args, we could also patch in `on('results')` or similar, although when I tried that, that was slightly too late (a "result" render would be able to cause a temporary render with result state instead of eager state).

I'm not the biggest fan of doing this patching here, so if another place makes sense, that would be nice.

It can't be the constructor of search results, as that should still be based on the associated _state, as otherwise we're reading from the wrong query (eg previous query had a disjunctive refinement, current state doesn't have that, you don't want to read from first query, but still from second query, only after results are instantiated you patch _state).

packages/instantsearch.js/src/lib/utils/render-args.ts
```diff
@@ -30,6 +30,10 @@ export function createRenderArgs(
...
```
Haroenv, line 36: alternative that doesn't cause the change in getServerState (but is likely slower due to the constructor being more work to instantiate multiple times per render)

```suggestion
  const parentResults = parent.getResults()!;
  const results = new algoliasearchHelper.SearchResults(
    parentResults._state,
    parentResults._rawResults
  );
  // To make the UI optimistic, we will always render using the current state,
  // but the previous results. This means a change will be visible immediately,
  // regardless of the status of the network request.
  results._state = parent.getHelper()!.state;
```

probably a couple other alternatives that don't mutate possible too, but they'll all be higher impact probably

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -230,7 +230,15 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
...
+      if (!derivedHelper) return null;
+      if (!derivedHelper.lastResults) return null;
+
+      // To make the UI optimistic, we will always render using the current state,
+      // but the previous results. This means a change will be visible immediately,
+      // regardless of the status of the network request.
+      derivedHelper.lastResults._state = helper!.state;
+
+      return derivedHelper.lastResults;
     },
```
Haroenv, line 242: this is part A of the implementation

packages/react-instantsearch-hooks/src/components/__tests__/InstantSearchSSRProvider.test.tsx
```diff
@@ -539,10 +543,15 @@ describe('InstantSearchSSRProvider', () => {
...
+    expect(searchClient.search).toHaveBeenCalledTimes(0);
+    expect(appleRefinement).not.toBeChecked();
+
     userEvent.click(appleRefinement);
 
     await waitFor(() => {
-      expect(appleRefinement).toBeChecked();
+      expect(searchClient.search).toHaveBeenCalledTimes(1);
+      // possible bug in testing-library, this is checked in real life in the same situation
+      // expect(appleRefinement).toBeChecked();
```
Haroenv, line 554: I don't understand why the bottom assertion fails. It already is the case before this PR actually, as the assertion was written wrong (before this PR it already was checked, now the local state is trusted, and thus not checked).

Because the assertion was originally written wrong, I'm kind of doubting whether https://github.com/algolia/react-instantsearch/issues/3530 was even asserted by this test.

Haroenv: This test was never correctly passing (searchClient wasn't correct) and the behaviour changed in this PR. It's sufficiently covered under the expanded "renders refinements from local widget state" test now.

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -230,7 +230,15 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
...
 
     getResults() {
-      return derivedHelper && derivedHelper.lastResults;
+      if (!derivedHelper) return null;
+      if (!derivedHelper.lastResults) return null;
```
dhayab, line 234: Nitpick:

```suggestion
      if (!derivedHelper?.lastResults) return null;
```

dhayab (approved):
👏

packages/instantsearch.js/src/widgets/index/index.ts
Haroenv: sure, that seems equivalent

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -365,9 +365,10 @@ describe('useInstantSearch', () => {
 
     test('turns to loading and error when searching', async () => {
       const searchClient = createSearchClient({});
-      searchClient.search.mockImplementation(() =>
-        Promise.reject(new Error('API_ERROR'))
-      );
+      searchClient.search.mockImplementation(async () => {
+        await wait(100);
```
sarahdayan, line 369: Why do we need to wait?

packages/vue-instantsearch/src/util/__tests__/createServerRootMixin.test.js
```diff
@@ -1149,18 +1150,13 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/instantsear
...
-          ],
-        })
-      );
+      // renders with local state, not the one from results
+      expect(renderArgs.state).toEqual(localState);
+      results._state = localState;
+      expect(renderArgs.results).toEqual(results);
+      expect(renderArgs.scopedResults).toHaveLength(1);
+      expect(renderArgs.scopedResults[0].indexId).toEqual('lol');
+      expect(renderArgs.scopedResults[0].results).toEqual(results);
```
sarahdayan, line 1159: Why change to multiple assertions? What didn't work with `expect.objectContaining`?

packages/instantsearch.js/src/lib/utils/createSendEventForHits.ts
```diff
@@ -186,7 +186,7 @@ export function createBindEventForHits({
       index,
       methodName: 'bindEvent',
       args,
-      isSearchStalled: false,
+      invalidStatus: false,
```
sarahdayan, line 189: Why is the value different in `bindEvent` and `sendEvent`?

packages/instantsearch.js/src/lib/utils/createSendEventForHits.ts
```diff
@@ -31,13 +31,13 @@ const buildPayloads = ({
   widgetType,
   methodName,
   args,
-  isSearchStalled,
+  invalidStatus,
```
sarahdayan, line 34: Right now, this flag seems to be more accurately called `shouldSendViewEvent`.

Why don't we instead pass the InstantSearch instance and keep the logic here?

sarahdayan (commented):
In addition to your existing tests, I think each flavor should have the same suite of user-oriented tests (testing the UI, not checking what methods are called) where we slow down the search request, click a refinement, and assess whether the checkbox is checked or not (before this PR it should not, with this PR it should).

Ideally, we do this for every impacted widget (all refinement widgets, pagination, etc.) This could be in a `optimistic-ui` test suite.

Later on, when we do a common test suite as in Predict, we could factorize that into one.

packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -335,7 +335,11 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
             { results }
           );
 
-          if (cachedHits[page] === undefined && !results.__isArtificial) {
+          if (
+            cachedHits[page] === undefined &&
+            !results.__isArtificial &&
+            instantSearchInstance.status === 'idle'
+          ) {
```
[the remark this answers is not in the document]
Haroenv: this is already covered by existing tests (which failed without this condition), but I can add it

packages/react-instantsearch-hooks/src/hooks/__tests__/useInstantSearch.test.tsx
Haroenv: I think Jest's (or node's) implementation of timers and promises made it so that the error state happened immediately and skipped the loading state, which I wanted to assert in this test. Any time <16 ms was flaky so I added a bit more margin

packages/vue-instantsearch/src/util/__tests__/createServerRootMixin.test.js
Haroenv: the error message was completely unreadable when the test failed, two massive objects without any diff at all. I had to manually diff to see where the error even was.

packages/instantsearch.js/src/lib/utils/createSendEventForHits.ts
Haroenv: That's possible too, I didn't want to touch the code too much, but happy to change if you think it should be refactored

packages/instantsearch.js/src/lib/utils/createSendEventForHits.ts
Haroenv: I believe that's a bug introduced in https://github.com/algolia/instantsearch/commit/1686dfb096cfce062e268feda7956e3b160bf2da

Haroenv: actually, no it's not a bug. BindEvent always should get its info, it's sendEvent(view) which should be skipped if the status is not right. The logic is in here, but on further thought, it probably should be in sendEvent directly

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -1359,10 +1363,15 @@ describe('scheduleStalledRender', () => {
     search.mainHelper!.search();
     search.mainHelper!.search();
 
+    await wait(0);
+
+    // search starts
```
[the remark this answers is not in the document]
Haroenv: It's more asserting "has been called another time since the last assert (they are 1,2,3 in order). Do you know another way of describing that?

packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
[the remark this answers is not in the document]
sarahdayan: If that's already tested, no need, and it can be covered in more user-oruented tests anyway.

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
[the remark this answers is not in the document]
sarahdayan: Ah yeah, it always gets me because it's mutable—I usually clear mocks in between to add clarity.

One way we can clarify is maybe this:

```suggestion
    // search starts, `widget.render` is immediately called once more
```

packages/instantsearch.js/src/lib/utils/createSendEventForHits.ts
sarahdayan: It's debt we don't have to clean up later (aka never) so since we have good code coverage, I think we should use the opportunity.

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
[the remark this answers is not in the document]
Haroenv: Resetting mock makes sense, then I can just do hasBeenCalledTimes(1) multiple times

packages/instantsearch.js/src/lib/utils/createSendEventForHits.ts
Haroenv: I've made a draft PR here: https://github.com/algolia/instantsearch/pull/5439

packages/instantsearch.js/src/components/Pagination/Pagination.tsx
```diff
@@ -81,7 +81,7 @@ function Pagination(props: PaginationProps) {
         {props.pages.map((pageNumber) => (
           <PaginationLink
             key={pageNumber}
-            ariaLabel={`${pageNumber + 1}`}
+            ariaLabel={`Page ${pageNumber + 1}`}
```
Haroenv, line 84: other aria-labels aren't configurable, so changing this one to add text isn't a breaking change

packages/vue-instantsearch/src/components/Pagination.vue
```diff
@@ -88,6 +88,7 @@
             <a
               :class="suit('link')"
               :href="state.createURL(page)"
+              :aria-label="`Page ${page + 1}`"
```
Haroenv, line 91: other aria-labels aren't configurable, so adding this one to add text isn't a breaking change

packages/react-instantsearch-hooks/src/components/__tests__/InstantSearchSSRProvider.test.tsx
```diff
@@ -437,112 +466,4 @@ describe('InstantSearchSSRProvider', () => {
       ]);
     });
   });
-
-  // Fixes https://github.com/algolia/react-instantsearch/issues/3530
-  test('renders initial refinement and allows to refine them', async () => {
```
Haroenv, line 442: mentioned in a different comment too, but this test wasn't actually asserting what it was meant to (toBeChecked at the bottom should be .not.toBeChecked if we actually fixed the issue, but we never did). The similar assertion about it being interactive is added to the other test earlier.

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -230,7 +230,16 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
...
     getResults() {
-      return derivedHelper && derivedHelper.lastResults;
+      if (!derivedHelper?.lastResults) return null;
+
+      // To make the UI optimistic, we patch the state to display to the current
+      // one instead of the one associated with the latest results.
+      // This means user-driven UI changes (e.g., checked checkbox) are reflected
+      // immediately instead of waiting for Algolia to respond, regardless of
+      // the status of the network request.
+      derivedHelper.lastResults._state = helper!.state;
```
Haroenv, line 240: root of the implementation

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+        },
+        attributes,
+      };
+      const env = await setup(options);
+      const { act = fakeAct } = env;
+
+      // wait for initial results
+      await act(async () => {
+        await wait(margin + delay);
+        await wait(0);
```
dhayab, line 47: Could you explain or comment why we wait a second time after most of the interactions?

dhayab (commented):
Great set up for the common test suite!

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+  createSearchClient,
+  createMultiSearchResponse,
+  createSingleSearchResponse,
+} from '@instantsearch/mocks';
+import type { HierarchicalMenuSetup } from '.';
+import { fakeAct } from '../common';
+import userEvent from '@testing-library/user-event';
+
+export function createOptimisticUiTests(setup: HierarchicalMenuSetup) {
+  // @TODO: after helper is updated with https://github.com/algolia/algoliasearch-helper-js/pull/925, enable this test
```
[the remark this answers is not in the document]
Haroenv: I'll enable it in a separate PR

packages/instantsearch.js/src/components/Pagination/Pagination.tsx
Haroenv: The tests would fail or need to be skipped between the tests. Happy to revert in this PR if needed

tests/common-unit/hierarchical-menu/optimistic-ui.ts
Haroenv: I don't fully understand, but I believe it's related to jest timers being quite strict in ordering. If there's no two timers waited for, a timer added _during_ the initial wait will still be triggered after.

packages/instantsearch.js/src/components/Pagination/Pagination.tsx
dhayab: Ah that's right, I saw you use those new aria label strings in tests. It's fine for me.

tests/common-unit/hierarchical-menu/index.ts
```diff
@@ -0,0 +1,10 @@
+import type { TestSetup } from '../common';
+import { createOptimisticUiTests } from './optimistic-ui';
+
+export type HierarchicalMenuSetup = TestSetup<{ attributes: string[] }>;
```
sarahdayan, line 4: Could it be worth allowing to pass all widget parameters right now? If later we need them, the diff will be smaller for existing tests that implement the common test suite.

Same for the other tests.

tests/common-unit/hierarchical-menu/index.ts
```diff
@@ -0,0 +1,10 @@
+import type { TestSetup } from '../common';
```
sarahdayan, line 1: I wouldn't categorize this as `common-unit`, this distinction is ambiguous IMO. Maybe just `tests/common/widgets/...`?

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+  describe.skip('optimistic UI', () => {
+    test('is checked immediately with a slow network', async () => {
```
sarahdayan, line 14:
```suggestion
    test('checks the clicked refinement immediately regardless of network latency', async () => {
```

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+              );
+            }),
+          }),
+        },
+        attributes,
+      };
+      const env = await setup(options);
+      const { act = fakeAct } = env;
+
+      // wait for initial results
```
sarahdayan, line 44:
```suggestion
      // Wait for initial results to populate widgets with data
```

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+
+      // select item
+      {
+        const firstItem = env.container.querySelector(
+          '.ais-HierarchicalMenu-link'
+        )!;
+        await act(async () => {
+          userEvent.click(firstItem);
+          await wait(0);
+          await wait(0);
```
sarahdayan, line 68: Why do we need this?

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+      {
+        expect(
+          env.container.querySelectorAll('.ais-HierarchicalMenu-item')
+        ).toHaveLength(2);
+        expect(
+          env.container.querySelectorAll('.ais-HierarchicalMenu-item--selected')
+        ).toHaveLength(0);
+      }
+
+      // select item
```
sarahdayan, line 60:
```suggestion
      // Select a refinement
```

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+      {
+        const firstItem = env.container.querySelector(
+          '.ais-HierarchicalMenu-link'
+        )!;
```
sarahdayan, line 64: Could we use Testing Library's queries so we can select by label? This way we could also easily assert whether the right refinement is selected.

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+        await act(async () => {
+          userEvent.click(firstItem);
+          await wait(0);
+          await wait(0);
+        });
+
+        // immediately after interaction
```
sarahdayan, line 71:
```suggestion
        // UI has changed immediately after the user interaction
```

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+        expect(
+          env.container.querySelectorAll('.ais-HierarchicalMenu-item--selected')
+        ).toHaveLength(1);
+      }
+
+      // after result comes in
```
sarahdayan, line 77:
```suggestion
      // Wait for new results to come in
```

tests/common-unit/hierarchical-menu/optimistic-ui.ts
sarahdayan, line 47: Why must we have this?

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+        await act(async () => {
+          await wait(delay + margin);
+        });
+
+        expect(
+          env.container.querySelectorAll('.ais-HierarchicalMenu-item--selected')
+        ).toHaveLength(1);
+      }
+
+      // unselect item
```
sarahdayan, line 88:
```suggestion
      // Unselect the selected refinement
```

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+        const firstItem = env.container.querySelector(
+          '.ais-HierarchicalMenu-link'
+        )!;
+        await act(async () => {
+          userEvent.click(firstItem);
+          await wait(0);
+          await wait(0);
+        });
+
+        // immediately after interaction
```
sarahdayan, line 99:
```suggestion
        // UI has changed immediately after the user interaction
```

tests/common-unit/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,118 @@
...
+        expect(
+          env.container.querySelectorAll('.ais-HierarchicalMenu-item--selected')
+        ).toHaveLength(0);
+      }
+
+      // after result comes in
```
sarahdayan, line 105:
```suggestion
      // Wait for new results to come in
```

tests/common-unit/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "@instantsearch/tests",
+  "version": "1.0.3",
+  "description": "common unit tests for all InstantSearch flavours"
```
sarahdayan, line 4:
```suggestion
  "description": "Common tests for all InstantSearch flavors."
```

sarahdayan (commented):
My comments for the `hierarchical-menu` common suite apply to others when relevant.

tests/common-unit/hierarchical-menu/index.ts
Haroenv: Ah yes, I forgot that we have the type for the widgetParams. Totally makes sense to add

tests/common-unit/hierarchical-menu/index.ts
Haroenv: tests/common/widgets makes sense. Should the name of the package also be `@instantsearch/tests/widgets` or still from `@instantsearch/tests` directly?

tests/common-unit/hierarchical-menu/optimistic-ui.ts
Haroenv: I don't know why it didn't work when I added a larger margin, but when there's two different calls, we get to wait until the render happened (it's debounced). I also tried out controlled client, but it needed the same double wait, so it didn't seem worth it to me.

tests/common-unit/hierarchical-menu/index.ts
sarahdayan: I think `@instantsearch/tests` works, because we can later write common tests for other pieces (e.g., connector usage) or full apps.

tests/common-unit/hierarchical-menu/optimistic-ui.ts
sarahdayan: Sure, we can investigate this later.

tests/common/widgets/hierarchical-menu/optimistic-ui.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      // Select a refinement
+      {
+        const firstItem = screen.getByRole('link', {
+          name: 'Apple 200',
+        });
+        await act(async () => {
+          userEvent.click(firstItem);
+          await wait(0);
+          await wait(0);
```
aymeric-giraudet, line 72: Why do these actions need to be awaited twice ?

aymeric-giraudet (approved):
Nice PR ! Great job on abstracting the common test suites :D

tests/common/widgets/hierarchical-menu/optimistic-ui.ts
aymeric-giraudet: Nevermind I saw Dhaya asked the same up there :D

tests/common-unit/hierarchical-menu/optimistic-ui.ts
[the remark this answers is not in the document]
Haroenv: as it wasn't yet merged, I did this in same PR anyway

sarahdayan (approved):
Gorgeous work on this topic, and I'm excited we have a common test suite now. Can't wait to move most our tests there!

FabienMotte (approved):
Really nice work on the common test suite 👏

PR #5432, opened 2023-01-13
feat: handling routing in Next.js

aymeric-giraudet:
### [FX-1976](https://algolia.atlassian.net/browse/FX-1976)

- Added 3 new options to `history` router
- Added a check and warning in `react-instantsearch-hooks` if Next.js is detected without the new router
- New package for the router, along with e2e tests

About the e2e tests : 

I decided to put them inside of the package itself rather than in `tests/e2e` because it's very specific. For now it's only local because I'm not sure how to approach CI with those, I don't think we should run them for any commit.
Also I edited the `next` example so I could test more but if you think it should be a separate example like `next-routing` I will do it !

[FX-1976]: https://algolia.atlassian.net/browse/FX-1976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react-instantsearch-hooks-next-router/index.ts
```diff
@@ -0,0 +1,80 @@
+import { history } from 'instantsearch.js/es/lib/routers';
+import nextRouterSingleton from 'next/router';
+
+import type { Router, UiState } from 'instantsearch.js';
+import type { BrowserHistoryArgs } from 'instantsearch.js/es/lib/routers/history';
+import type { Router as NextRouter } from 'next/router';
+
+export function createInstantSearchNextRouter<TRouteState = UiState>(
+  url: string | undefined,
```
Haroenv, line 9: maybe that should be called `serverURL`?

packages/react-instantsearch-hooks-next-router/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "exports": {
+    "types": "./dist/index.d.ts",
+    "require": "./dist/index.cjs",
+    "default": "./dist/index.modern.mjs"
+  },
+  "scripts": {
+    "build": "microbundle index.ts"
+  },
+  "devDependencies": {
+    "microbundle": "0.15.1"
```
Haroenv, line 20: we should probably not mix bundlers and use rollup here too (or switch bundlers to micro bundle another time)

packages/react-instantsearch-hooks-next-router/index.ts
```diff
@@ -0,0 +1,80 @@
...
+      let initialPathname: string;
+
+      if (typeof window !== 'undefined') {
+        initialPathname = nextRouterSingleton.pathname;
+
+        if (nextRouterSingleton.router?._bps) {
+          previousBeforePopState = nextRouterSingleton.router._bps;
+        }
+
+        nextRouterSingleton.beforePopState(() => {
```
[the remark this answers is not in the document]
aymeric-giraudet: It's not considered as an event by Next.js like `routeChangeComplete` and so on :/

packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
```diff
@@ -179,3 +182,23 @@ function addAlgoliaAgents(
...
+  routing: UseInstantSearchApiProps<TUiState, TRouteState>['routing']
+) {
+  if (!routing || typeof window === 'undefined' || !('__NEXT_DATA__' in window))
+    return;
+
+  // @ts-expect-error: _isNextRouter is only set on the Next.js router
+  const isUsingNextRouter = routing !== true && routing?.router?._isNextRouter;
+
+  warn(
+    isUsingNextRouter,
```
Haroenv, line 196: I believe if this whole condition is inlined, it will not be in the prod bundle, and thus also not needed to increase bundlesize

tsconfig.json
```diff
@@ -19,6 +19,7 @@
     "**/es",
     "**/dist",
     "tests/e2e",
+    "**/__tests__/e2e",
```
Haroenv, line 22: why do they fail TS?

yarn.lock
```diff
@@ -1081,7 +1081,15 @@
...
-"@babel/plugin-transform-flow-strip-types@^7.0.0", "@babel/plugin-transform-flow-strip-types@^7.14.5", "@babel/plugin-transform-flow-strip-types@^7.16.0":
+"@babel/plugin-transform-flow-strip-types@^7.0.0", "@babel/plugin-transform-flow-strip-types@^7.14.5":
+  version "7.14.5"
+  resolved "https://registry.yarnpkg.com/@babel/plugin-transform-flow-strip-types/-/plugin-transform-flow-strip-types-7.14.5.tgz#0dc9c1d11dcdc873417903d6df4bed019ef0f85e"
+  integrity sha512-KhcolBKfXbvjwI3TV7r7TkYm8oNXHNBqGOy6JDVwtecFaRoKYsUUqJdS10q0YDKW1c6aZQgO+Ys3LfGkox8pXA==
+  dependencies:
+    "@babel/helper-plugin-utils" "^7.14.5"
+    "@babel/plugin-syntax-flow" "^7.14.5"
+
+"@babel/plugin-transform-flow-strip-types@^7.16.0":
```
Haroenv, line 1092: you can probably run `npx yarn-deduplicate`, as I think these are compatible

packages/react-instantsearch-hooks-router-nextjs/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "react-instantsearch-hooks-router-nextjs",
+  "version": "6.39.2",
```
Haroenv, line 3: if we're using sync versioning, this should be in the metadata version test too

examples/react-hooks/next/pages/search/[category].tsx
```diff
@@ -0,0 +1,163 @@
...
+        routing={{
+          router: createInstantSearchRouterNext({
+            singletonRouter,
+            serverUrl: url,
+            routerOptions: {
+              createURL({ routeState, location, qsModule }) {
+                const { category, ...queryParameters } = routeState;
+                const { origin, pathname } = location;
+
+                return `${origin}${pathname}${qsModule.stringify(
```
Comment (Haroenv) [inline examples/react-hooks/next/pages/search/[category].tsx:88]: it's good for a demo, but do we want to encourage the main next demo have two different InstantSearch instances?

examples/react-hooks/next/package.json
```diff
@@ -11,11 +11,12 @@
   "dependencies": {
     "algoliasearch": "4.14.3",
     "instantsearch.css": "8.0.0",
-    "next": "12.1.6",
+    "next": "13.1.1",
```
Haroenv, line 14: does it work with older next too?

examples/react-hooks/next/components/Menu.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import Link from 'next/link';
+
+export default function Menu() {
+  return (
+    <div>
+      Menu:
+      <ul style={{ display: 'flex', gap: '1em', listStyle: 'none' }}>
+        <li>
+          <Link href={`/search/`}>Home</Link>
```
Haroenv, line 10: the right way to make this would be `useMenu` + Portal, no? Again, this is good for what people asked for in the original question, but I don't think for the main demo

examples/react-hooks/next/pages/search/[category].tsx
Comment (aymeric-giraudet) [reply examples/react-hooks/next/pages/search/[category].tsx:88]: Yeah I guess I should make a separate demo called `next-routing` maybe ?

examples/react-hooks/next/package.json
aymeric-giraudet: Yes I tested each major from 9 to 13

examples/react-hooks/next/components/Menu.tsx
aymeric-giraudet: Yeah I'll make an easier dynamic routing example in a separate demo for routing showcase

tsconfig.json
Haroenv: ah, they have their own tsconfig? I believe something is possible with nested configs, but I'm not fully sure. Could you put an ambient file at that module root and see if that declares the types as expected?

examples/react-hooks/next/components/Menu.tsx
aymeric-giraudet: Okay I made an example with navigation to another route without InstantSearch, and to show that Next.js links work I just also put a `Link` to show that navigation links outside of InstantSearch do update the state as well

tsconfig.json
aymeric-giraudet: Done ! Added XML reference tags in a `.d.ts` file

packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
aymeric-giraudet: I still wrapped the whole thing into an IIFE for readability.
It still is a external function too cause I don't wanna pollute the hook itself

packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
```diff
@@ -31,11 +32,17 @@ export function useInstantSearchApi<TUiState extends UiState, TRouteState>(
...
   const initialResults = serverState?.initialResults;
-  const searchRef = useRef<InstantSearch<TUiState, TRouteState> | null>(null);
   const prevPropsRef = useRef(props);
 
+  let searchRef = useRef<InstantSearch<TUiState, TRouteState> | null>(null);
+  // To prevent starting two instances with React Strict Mode + SSR, we use
+  // a ref from a parent.
+  if (serverState) {
+    searchRef = serverState.ssrSearchRef;
+  }
```
aymeric-giraudet, line 44: Okay so basically with `InstantSearchSSRProvider`, in Strict Mode it creates two instances of InstantSearch and starts both without disposing the first.
We do have to start at first render else there's a hydration mismatch.

I spent quite a lot of time looking for a solution within this file but it doesn't seem possible so I added a `useRef` in the provider which we use instead.
The only drawback I can think of is it's no longer possible to have two instances of `InstantSearch` within the same `InstantSearchSSRProvider` context which I guess wouldn't make sense anyway

packages/react-instantsearch-hooks-router-nextjs/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "dist"
+  ],
+  "scripts": {
+    "clean": "rm -rf dist",
+    "build": "yarn build:cjs && yarn build:es && yarn build:types",
+    "build:cjs": "BABEL_ENV=cjs babel src --root-mode upward --extensions '.js,.ts,.tsx' --out-dir dist/cjs --ignore '**/__tests__/**/*','**/__mocks__/**/*' --quiet && ../../scripts/prepare-cjs.sh",
+    "build:es": "BABEL_ENV=es babel src --root-mode upward --extensions '.js,.ts,.tsx' --out-dir dist/es --ignore '**/__tests__/**/*','**/__mocks__/**/*' --quiet",
+    "build:types": "tsc -p ./tsconfig.declaration.json --outDir ./dist/es",
+    "test:exports": "node ./__tests__/module/is-es-module.mjs && node ./__tests__/module/is-cjs-module.cjs",
+    "test:start-server": "yarn workspace example-react-instantsearch-hooks-next-routing-example build && yarn workspace example-react-instantsearch-hooks-next-routing-example start",
```
Haroenv, line 47: can you not use the static server of wdio like done in the "common" e2e tests?

packages/react-instantsearch-hooks/src/components/InstantSearchSSRProvider.tsx
```diff
@@ -18,10 +18,14 @@ export type InstantSearchSSRProviderProps =
...
  */
-export function InstantSearchSSRProvider({
-  children,
-  ...props
-}: InstantSearchSSRProviderProps) {
+export function InstantSearchSSRProvider<
+  TUiState extends UiState,
+  TRouteState = TUiState
+>({ children, ...props }: InstantSearchSSRProviderProps) {
+  const ssrSearchRef = React.useRef<InstantSearch<UiState, TRouteState> | null>(
```
Haroenv, line 25: is the `notify` callback as used for getRenderState not sufficient?

packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
```diff
@@ -179,3 +188,31 @@ function addAlgoliaAgents(
     searchClient.addAlgoliaAgent!(userAgent!);
   });
 }
+
+function warnNextRouter<TUiState extends UiState, TRouteState>(
+  routing: UseInstantSearchApiProps<TUiState, TRouteState>['routing']
+) {
+  warn(
+    (function () {
```
Haroenv, line 196: alternatively to the inline, it can completely be put in `if(__DEV__)` manually

Haroenv (commented):
👍

packages/react-instantsearch-hooks-router-nextjs/package.json
aymeric-giraudet: I would have to make Next output static files so I would have to get ride of `getServerSideProps` and stuff :/

packages/react-instantsearch-hooks/src/components/InstantSearchSSRProvider.tsx
aymeric-giraudet: The problem is client-side only, after `getServerState` did its thing so that's why I only handle it with `InstantSearchSSRContext`.
Do you mean we should handle it for both `getServerState` rendering and Next.js SSR ? I put the presence of `InstantSearchSSRContext` to be the condition to use the ref from this context so I guess it should work.

aymeric-giraudet: Ah sorry you meant to merge the two contexts and just use `notifyServer` instead ?
I'll see if it's possible yeah

Haroenv (approved):

dhayab (approved):

PR #5438, opened 2023-01-18
feat(rendering): revert search state on error

Haroenv:
**Summary**

going a step further than #5429, we want to also when there's an error revert to the last known good state.

**Result**

This goes back to the previous (non-optimistic) behaviour when an error happens, with a caveat that the state will be fully consistent. That means UiState and the URL also update back to the previous state

### DIFF
--- a/packages/instantsearch.js/src/widgets/index/index.ts
+++ b/packages/instantsearch.js/src/widgets/index/index.ts
@@ -212,6 +212,7 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
   let localParent: IndexWidget | null = null;
   let helper: Helper | null = null;
   let derivedHelper: DerivedHelper | null = null;
+  let lastValidSearchParameters: SearchParameters | null = null;
 
   return {
     $$type: 'ais.index',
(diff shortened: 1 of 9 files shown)


packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -584,6 +586,15 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
...
 
+      // we can't attach a listener to the error event of search, as the error
+      // then would no longer be thrown for global handlers.
+      if (
+        instantSearchInstance.status === 'error' &&
+        !instantSearchInstance.mainHelper!.hasPendingRequests()
+      ) {
+        helper!.setState(lastValidSearchParameters!);
+      }
+
```
Haroenv, line 597: I wonder if this can be put somewhere else? I thought in InstantSearch error handler, but that doesn't know local state etc, so would require a dedicated method on index widget.

I guess this could be problematic if render is called another time after error, with error again, it will again reset. (not sure if that behaviour is right?)

examples/react-hooks/default-theme/src/App.tsx
```diff
@@ -32,10 +33,22 @@ import {
 import { Tab, Tabs } from './components/layout';
 
 import './App.css';
+import { override } from './override-method';
 
-const searchClient = algoliasearch(
-  'latency',
-  '6be0576ff61c053d5f9a3225e2a90f76'
+let shouldThrowError = false;
+const searchClient = override(
```
Haroenv, line 39: oh, these changes need to be reverted

dhayab (approved):
Good for me!

PR #5452, opened 2023-01-25
wip: insights prototypes

dhayab:
> **Warning**
> Do not merge


Haroenv:
I think everything here is now in #5490, so this branch can be removed (all changes are still readable here if we need to reference it)

PR #5457, opened 2023-01-26
fix(insights): pass correct sendEvent to html template

Haroenv:
**Summary**

There was an existing prop called `sendEvent`, which we passed through to the component when we introduced html templates. However, this sendEvent was actually sendEventToInsights, which is of course problematic, as it doesn't match the documentation, nor does it add the necessary conversion from a simple "three argument" form to "object form".

This PR changes the prop that existed from `sendEventToInsights` to `sendEvent` which comes from the connector.

discovered in https://github.com/algolia/instantsearch/discussions/5447#discussioncomment-4779169

**Result**

`sendEvent` in html templates behaves as it should, and uses `sendEvent` from the connector.

This PR is fairly urgent, as the documentation has been updated to use `sendEvent` instead of `bindEvent` (which didn't have this bug)

**Reviewer notes**

Hide whitespace, as that's the only thing changed in the snapshots because they were indented wrong :)

### DIFF
--- a/packages/instantsearch.js/src/lib/insights/listener.tsx
+++ b/packages/instantsearch.js/src/lib/insights/listener.tsx
@@ -9,7 +9,7 @@ import type { InsightsEvent } from '../../middlewares/createInsightsMiddleware';
 type WithInsightsListenerProps = {
   [key: string]: unknown;
   insights: InsightsClient;
-  sendEventToInsights?: (event: InsightsEvent) => void;
+  sendEvent?: (event: InsightsEvent) => void;
 };
 
 const findInsightsTarget = (
@@ -48,7 +48,7 @@ const parseInsightsEvent = (element: HTMLElement): InsightsEvent[] => {
 const insightsListener = (BaseComponent: any) => {
   function WithInsightsListener(props: WithInsightsListenerProps) {
     const handleClick = (event: MouseEvent): void => {
-      if (props.sendEventToInsights) {
+      if (props.sendEvent) {
         // new way with insights middleware
         const targetWithEvent = findInsightsTarget(
           event.target as HTMLElement | null,
@@ -58,7 +58,7 @@ const insightsListener = (BaseComponent: any) => {
         if (targetWithEvent) {
           const payload = parseInsightsEvent(targetWithEvent);
 
-          payload.forEach((single) => props.sendEventToInsights!(single));
+          payload.forEach((single) => props.sendEvent!(single));
         }
       }
 
--- a/packages/instantsearch.js/src/lib/insights/listener.tsx
+++ b/packages/instantsearch.js/src/lib/insights/listener.tsx
@@ -9,7 +9,7 @@ import type { InsightsEvent } from '../../middlewares/createInsightsMiddleware';
 type WithInsightsListenerProps = {
   [key: string]: unknown;
   insights: InsightsClient;
-  sendEvent?: (event: InsightsEvent) => void;
+  sendEventToInsights?: (event: InsightsEvent) => void;
 };
 
 const findInsightsTarget = (
@@ -48,7 +48,7 @@ const parseInsightsEvent = (element: HTMLElement): InsightsEvent[] => {
 const insightsListener = (BaseComponent: any) => {
   function WithInsightsListener(props: WithInsightsListenerProps) {
     const handleClick = (event: MouseEvent): void => {
-      if (props.sendEvent) {
+      if (props.sendEventToInsights) {
         // new way with insights middleware
         const targetWithEvent = findInsightsTarget(
           event.target as HTMLElement | null,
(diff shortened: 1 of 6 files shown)


packages/instantsearch.js/src/widgets/hits/__tests__/hits-test.ts
```diff
@@ -82,170 +82,172 @@ describe('hits()', () => {
 
     expect(render).toHaveBeenCalledTimes(2);
     expect(firstRender.props).toMatchInlineSnapshot(`
-{
```
Haroenv, line 85: this diff is a little weird because it's properly indented now, but the only changes are adding `sendEventToInsights` (and changing `sendEvent` reference, but that doesn't show in the snapshot)

packages/instantsearch.js/src/lib/insights/listener.tsx
```diff
@@ -58,7 +58,7 @@ const insightsListener = (BaseComponent: any) => {
         if (targetWithEvent) {
           const payload = parseInsightsEvent(targetWithEvent);
 
-          payload.forEach((single) => props.sendEvent!(single));
+          payload.forEach((single) => props.sendEventToInsights!(single));
```
dhayab, line 61: Would it work if we just rely on the fixed `sendEvent` prop that correctly references the method from the connector?

Haroenv: Let me check, it should indeed work as well!

Haroenv: yep that worked. Implemented in 3aa05880e

dhayab (approved):

packages/instantsearch.js/src/widgets/hits/__tests__/hits-integration-test.ts
```diff
@@ -122,7 +122,101 @@ describe('hits', () => {
       );
     });
 
-    it('sends click event', async () => {
+    it('sendEvent sends click event', async () => {
```
sarahdayan, line 125:
```suggestion
    it('sends `click` event with `sendEvent`', async () => {
```

packages/instantsearch.js/src/widgets/hits/__tests__/hits-integration-test.ts
```diff
@@ -122,7 +122,101 @@ describe('hits', () => {
...
+                ${item.title}
+              </button>
+            `,
+          },
+        }),
+      ]);
+      search.start();
+      await wait(0);
+
+      expect(onEvent).toHaveBeenCalledTimes(1); // view event by render
```
sarahdayan, line 148: Could we also test `toHaveBeenCalledWith` to test that this is a `view` event? This way we can remove the comment.

packages/instantsearch.js/src/widgets/hits/__tests__/hits-integration-test.ts
```diff
@@ -122,7 +122,101 @@ describe('hits', () => {
...
+      fireEvent.click(getByText(container, 'title 1'));
+      expect(onEvent).toHaveBeenCalledTimes(2);
+      expect(onEvent.mock.calls[onEvent.mock.calls.length - 1][0]).toEqual({
```
sarahdayan, line 151: Can we clear the mock before so we remove the view event and we can simplify this assertion with `toHaveBeenCalledWith`?

packages/instantsearch.js/src/widgets/hits/__tests__/hits-integration-test.ts
```diff
@@ -122,7 +122,101 @@ describe('hits', () => {
...
+          eventName: 'Item Clicked',
+          index: 'instant_search',
+          objectIDs: ['object-id0'],
+          positions: [1],
+        },
+        widgetType: 'ais.hits',
+      });
+    });
+
+    it('sendEvent sends conversion event', async () => {
```
sarahdayan, line 172:
```suggestion
    it('sends `conversion` event with `sendEvent`', async () => {
```

packages/instantsearch.js/src/widgets/hits/__tests__/hits-integration-test.ts
```diff
@@ -122,7 +122,101 @@ describe('hits', () => {
...
+        payload: {
+          eventName: 'Product Ordered',
+          index: 'instant_search',
+          objectIDs: ['object-id1'],
+        },
+        widgetType: 'ais.hits',
+      });
+    });
+
+    it('bindEvent sends click event', async () => {
```
sarahdayan, line 219:
```suggestion
    it('sends`click` event with `bindEvent`', async () => {
```

sarahdayan (approved):
Comments are optional, feel free to apply or not and merge afterwards.

packages/instantsearch.js/src/widgets/hits/__tests__/hits-integration-test.ts
Haroenv: That's already tested in the first test in this describe. It would be a little noisy IMO

PR #5460, opened 2023-01-27
fix(events): prevent sending duplicate view events

### LINKED ISSUE #5442: Insight track function called multiple times
### Insights track called multiple times

Insights track function is called as many times as the useHits hook or Hit component are used in the InstantSearch component.

### 🔍 Steps to reproduce

Go to the live reproduction
See the multiple logs in the console
Remove one component using the useHits hook
See the number of logs decreasing

### Live reproduction

https://codesandbox.io/s/relaxed-james-x92y0k?file=/src/App.tsx

### 💭 Expected behavior

the 'view Hits' event is supposed to be called juste one time, regardless of how many times you used the useHits hook.

### Package version

algoliasearch 4.14.3, instantsearch.js 4.43.1, react-instantsearch-hooks-web 6.38.1, search-insights 2.2.3-hooks-web

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

Haroenv:
**Summary**

Quick and dirty solution where we keep track of which hits have been sent and skip duplicates.

Technically this is a memory leak, but i don't know what condition to use to clear the cache?

fixes #5442

**Result**

duplicate view events don't get sent multiple times

### DIFF
--- a/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
@@ -150,6 +150,8 @@ export function createInsightsMiddleware<
           immediate: true,
         });
 
+        const viewEventCache: Record<string, boolean> = {};
+
         instantSearchInstance.sendEventToInsights = (event: InsightsEvent) => {
           if (onEvent) {
             onEvent(event, _insightsClient);
@@ -158,6 +160,14 @@ export function createInsightsMiddleware<
               (helper.state as PlainSearchParameters).userToken
             );
             if (hasUserToken) {
+              if (event.insightsMethod === 'viewedObjectIDs') {
+                const hits = JSON.stringify(event.payload.hits);
+                if (viewEventCache[hits]) {
+                  return;
+                } else {
+                  viewEventCache[hits] = true;
+                }
+              }
               insightsClient(event.insightsMethod, event.payload);
             } else {
               warning(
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -150,6 +150,8 @@ export function createInsightsMiddleware<
           immediate: true,
         });
 
+        const viewEventCache: Record<string, boolean> = {};
+
```
Haroenv, line 154: maybe it could be cleared on the next promise / timer?

tkrugg (commented):
Hey team, I'm a bit confused on this PR cause I thought the risk of double events was more on clicks 🤔

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
tkrugg: I think you can use the queryID as a key, it's probably more efficient than stringifying the whole payload. Wdyt?

Haroenv: ah yes queryID would be better indeed!

Haroenv: Unfortunately view isn't sent with queryID

Haroenv:
> Hey team, I'm a bit confused on this PR cause I thought the risk of double events was more on clicks 🤔

I made this PR based on the issue described where if you have multiple hits widgets/connectors, you get multiple insights view events, which seems wrong / not needed?

Haroenv:
This solution is too hardcoded on view, and there's another approach with waiting a tick before triggering sendEvent, which would be useful for sendEvent too. I'll close this for now.

PR #5461, opened 2023-01-30
fix(infiniteHits): read cache correctly when search is loading

Haroenv:
**Summary**

The cache is based on search parameters (minus page), so when the state updates, we want to still use the previous state, as the previous hits are still valid.

One thing that could change this design is if the cache wasn't purely based on the parameters, and somehow was persisted too, so we could ask for the "last result", but that's not possible without major change in the cache.

Alternatives I tried before settling on this solution:
- exposing previousState to all calls of `getWidgetRenderState`
  - possible, but a large diff, and needs to be handled in all widgets, types, useConnector
- skipping render if status is `loading`
  - this is not sufficient, as React InstantSearch Hooks calls `getWidgetRenderState` itself, and thus gets to the mismatched cache case as well

**Result**

infiniteHits never renders empty hits on external interaction.

fixes https://algolia.atlassian.net/browse/CR-2769 (mostly, the customer has another bug that would show up when you have no results, and then results again, but that wasn't related to optimistic UI)

### DIFF
--- a/packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
+++ b/packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
@@ -292,9 +292,16 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
         };
       },
 
-      getWidgetRenderState({ results, helper, state, instantSearchInstance }) {
+      getWidgetRenderState({ results, helper, parent, instantSearchInstance }) {
         let isFirstPage: boolean;
         let currentPageHits: Array<Hit<THit>> = [];
+        /**
+         * We bail out of optimistic UI here, as the cache is based on search
+         * parameters, and we don't want to invalidate the cache when the search
+         * is loading.
+         */
+        const state = parent.getPreviousState()!;
+
         const cachedHits = cache.read({ state }) || {};
 
         if (!results) {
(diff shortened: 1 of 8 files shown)


dhayab (approved):
Great fix!

packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -292,9 +292,22 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
...
+        instantSearchInstance,
+      }) {
         let isFirstPage: boolean;
         let currentPageHits: Array<Hit<THit>> = [];
+        /**
+         * We bail out of optimistic UI here, as the cache is based on search
+         * parameters, and we don't want to invalidate the cache when the search
+         * is loading.
+         */
+        const state = parent.getPreviousState() || existingState;
```
dhayab, line 309: Are there any other cases / widgets where we would want to bail out from optimistic UI? I can only think of the searchbox, since it has its own behavior linked to focus state.

Haroenv: searchbox is indeed already handled, and no other widgets keep track of the previous state / results

aymeric-giraudet (approved):

PR #5462, opened 2023-01-30
chore(jest): use local InstantSearch.js

Haroenv:
**Summary**

mapping all files (widgets, connectors, utils) allows us to run `jest --watch` without ever needing to build to see updates

FX-2617

**Result**

you can run `yarn jest --watch` without needing to build InstantSearch.js, even for Vue InstantSearch or React InstantSearch Hooks

Fix the babel extension resolver plugin to not assume that the package name is verbatim in the path

> Note: this PR should be rebased, not squashed, as the two commits work separately

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -40,6 +40,8 @@ const config = {
     '^react-instantsearch-(.*[^v6])$':
       '<rootDir>/packages/react-instantsearch-$1/src/',
     '^instantsearch.js$': '<rootDir>/packages/instantsearch.js/src/',
+    '^instantsearch.js/es(.*)$': '<rootDir>/packages/instantsearch.js/src$1',
+    '^instantsearch.js/(.*)$': '<rootDir>/packages/instantsearch.js/$1',
   },
   modulePathIgnorePatterns: [
     '<rootDir>/packages/create-instantsearch-app/src/templates',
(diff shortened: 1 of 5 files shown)


Haroenv:
bleh, I didn't expect this to have an impact on the babel tests. I can either use a different package that we won't have a moduleNameMapper for, or use Jest Projects to isolate the babel tests out?

jest.config.js
```diff
@@ -37,6 +37,8 @@ const config = {
     '^react-instantsearch-(.*)$':
       '<rootDir>/packages/react-instantsearch-$1/src/',
     '^instantsearch.js$': '<rootDir>/packages/instantsearch.js/src/',
+    '^instantsearch.js/es(.*)$': '<rootDir>/packages/instantsearch.js/src$1',
```
Haroenv, line 40: something here causes jest to resolve wrong in the ReactDOMServer tests only it seems, not sure why.

Haroenv: solved since we no longer have React InstantSearch dom :D

tests/babel-testfixture/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@instantsearch/babel-testfixture",
+  "description": "A package to use as a test in extension-resolver",
```
dhayab, line 3:
```suggestion
  "private": true,
  "description": "A package to use as a test in extension-resolver",
```

dhayab (approved):

PR #5463, opened 2023-01-30
fix(types): expose InstantSearchStatus

Haroenv:
**Summary**

this isn't needed now, but i needed in a different iteration of a PR. It's probably interesting if it's exposed

**Result**

InstantSearchStatus type is exposed from InstantSearch.js root (so you don't have to do `InstantSearch['status']`


Haroenv (commented):
I'm not 100% sold we need this type exposed per se, but it could be useful and I already had the diff stashed

FabienMotte (approved):

PR #5469, opened 2023-01-31
fix(HitsPerPage): compute selected from isRefined

Haroenv:
**Summary**

fixes https://github.com/algolia/instantsearch/issues/5397

Slight improvement in the HitsPerPage + SortBy tests, as it was relying on Vue state instead of the DOM

I wanted both to be exactly the same, but SortBy doesn't have isRefined on the items apparently.

Workaround in the example to make HitsPerPage responsive to the URL is also removed.

**Result**

This makes HitsPerPage responsive to the URL, as expected like other widgets.

### DIFF
--- a/examples/vue/e-commerce/src/App.vue
+++ b/examples/vue/e-commerce/src/App.vue
@@ -259,19 +259,15 @@
                 {
                   label: '16 hits per page',
                   value: 16,
-                  default:
-                    getSelectedHitsPerPageValue() === 16 ||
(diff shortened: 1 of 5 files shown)


sarahdayan (approved):
Nice, can't wait to make all these widget and component tests common across flavors.

Haroenv:
Yes, I doubted to write a common test here, but I was juggling too many other things to write it just yet

PR #5478, opened 2023-02-14
fix(templates): cd into proper build dir

aymeric-giraudet:
**Summary**

We didn't `cd` into the proper build folder so this action was failing : https://github.com/algolia/instantsearch/actions/runs/4173320532/jobs/7225507147


dhayab (approved):
👍

Haroenv (approved):

PR #5482, opened 2023-02-16
fix(breadcrumb): guard against undefined facets

Haroenv:
**Summary**

If a breadcrumb widget gets "its" hierarchical menu added dynamically, there will be a render where getFacetValues returns `undefined`.

I checked, and all other cases of `getFacetValues` guard already for its result possibly being undefined.

Also added the common test for optimistic UI for breadcrumb, as it's a related fix.

**Result**

No error when hierarchical menu for a breadcrumb gets added dynamically

### DIFF
--- a/packages/instantsearch.js/src/__tests__/common.test.tsx
+++ b/packages/instantsearch.js/src/__tests__/common.test.tsx
@@ -3,6 +3,7 @@
  */
 import {
   createHierarchicalMenuTests,
+  createBreadcrumbTests,
   createRefinementListTests,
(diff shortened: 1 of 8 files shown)


dhayab (approved):

dhayab:
@Haroenv should this guard be also applied to the hierarchical menu as well?

Haroenv:
> @Haroenv should this guard be also applied to the hierarchical menu as well?

Hierarchical menu, menu, refinement list etc. already have this guard

PR #5484, opened 2023-02-16
feat(insights): automatically load search-insights if not passed

Haroenv:
**Summary**

When the insights middleware is instantiated, it will attempt to extract or load a search-insights client.

**Result**

- insights middleware makes `insightsClient` optional
- first it checks if there's an unlinked existing insights client
- then it loads search-insights from jsDelivr
- if there's an error in this loading, nothing happens

FX-2244

### DIFF
--- a/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -14,6 +14,7 @@ import { history } from '../../lib/routers';
 import { wait } from '@instantsearch/testutils/wait';
 import type { JSDOM } from 'jsdom';
 import type { PlainSearchParameters } from 'algoliasearch-helper';
+import { fireEvent } from '@testing-library/dom';
 
 declare const jsdom: JSDOM;
 
@@ -120,7 +121,11 @@ describe('insights', () => {
 
   describe('insightsClient', () => {
     it('does nothing when insightsClient is passed', () => {
-      createInsightsMiddleware({ insightsClient: () => {} });
+      const { instantSearchInstance } = createTestEnvironment();
+
+      instantSearchInstance.use(
+        createInsightsMiddleware({ insightsClient: () => {} })
+      );
 
       expect(document.body).toMatchInlineSnapshot(`<body />`);
       expect((window as any).AlgoliaAnalyticsObject).toBe(undefined);
@@ -128,7 +133,11 @@ describe('insights', () => {
     });
 
     it('does nothing when insightsClient is null', () => {
-      createInsightsMiddleware({ insightsClient: null });
+      const { instantSearchInstance } = createTestEnvironment();
+
+      instantSearchInstance.use(
+        createInsightsMiddleware({ insightsClient: null })
+      );
 
       expect(document.body).toMatchInlineSnapshot(`<body />`);
       expect((window as any).AlgoliaAnalyticsObject).toBe(undefined);
@@ -140,7 +149,9 @@ describe('insights', () => {
       const aa = () => {};
       (window as any).aa = aa;
 
-      createInsightsMiddleware();
+      const { instantSearchInstance } = createTestEnvironment();
+
+      instantSearchInstance.use(createInsightsMiddleware());
 
       expect(document.body).toMatchInlineSnapshot(`<body />`);
       expect((window as any).AlgoliaAnalyticsObject).toBe('aa');
@@ -148,7 +159,9 @@ describe('insights', () => {
     });
 
     it('loads the script when insightsClient is not passed', () => {
-      createInsightsMiddleware();
+      const { instantSearchInstance } = createTestEnvironment();
+
+      instantSearchInstance.use(createInsightsMiddleware());
 
       expect(document.body).toMatchInlineSnapshot(`
         <body>
@@ -161,9 +174,37 @@ describe('insights', () => {
       expect((window as any).aa).toEqual(expect.any(Function));
     });
 
-    it.todo('notifies when the script fails to be added');
+    it('notifies when the script fails to be added', () => {
+      const { instantSearchInstance } = createTestEnvironment();
+      const createElement = document.createElement;
+      document.createElement = () => {
+        throw new Error('error');
+      };
 
-    it.todo('notifies when the script fails to load');
+      instantSearchInstance.on('error', (error) =>
+        expect(error).toMatchInlineSnapshot(
+          `[Error: [insights middleware]: could not load search-insights.js. Please load it manually following https://alg.li/insights-init]`
+        )
+      );
+
+      instantSearchInstance.use(createInsightsMiddleware());
+
+      document.createElement = createElement;
+    });
+
+    it('notifies when the script fails to load', () => {
+      const { instantSearchInstance } = createTestEnvironment();
+
+      instantSearchInstance.on('error', (error) =>
+        expect(error).toMatchInlineSnapshot(
+          `[Error: [insights middleware]: could not load search-insights.js. Please load it manually following https://alg.li/insights-init]`
+        )
+      );
+
+      instantSearchInstance.use(createInsightsMiddleware());
+
+      fireEvent(document.querySelector('script')!, new ErrorEvent('error'));
+    });
   });
 
   describe('initialize', () => {
--- a/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -97,16 +97,16 @@ describe('insights', () => {
 
   beforeEach(() => {
     warning.cache = {};
+
+    (window as any).AlgoliaAnalyticsObject = undefined;
+    (window as any).aa = undefined;
+
+    document.body.innerHTML = '';
   });
 
   describe('usage', () => {
-    it('throws when insightsClient is not given', () => {
-      expect(() =>
-        // @ts-expect-error
-        createInsightsMiddleware()
-      ).toThrowErrorMatchingInlineSnapshot(
-        `"The \`insightsClient\` option is required if you want userToken to be automatically set in search calls. If you don't want this behaviour, set it to \`null\`."`
-      );
+    it('passes when insightsClient is not given', () => {
+      expect(() => createInsightsMiddleware()).not.toThrow();
     });
 
     it('passes with insightsClient: null', () => {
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -33,32 +42,55 @@ export type InsightsProps<
...
   if (_insightsClient !== null && !_insightsClient) {
-    if (__DEV__) {
-      throw new Error(
-        "The `insightsClient` option is required if you want userToken to be automatically set in search calls. If you don't want this behaviour, set it to `null`."
-      );
-    } else {
-      throw new Error(
-        'The `insightsClient` option is required. To disable, set it to `null`.'
-      );
-    }
```
Haroenv, line 55: effect of this PR, this error no longer happens, even when the auto-load fails, as it could well be something the user has no awareness of.

This means unfortunately you could silently send no events at all.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -33,32 +42,55 @@ export type InsightsProps<
...
+          const script = document.createElement('script');
+          script.async = true;
+          script.src = ALGOLIA_INSIGHTS_SRC;
+          script.onerror = () => {
+            // @TODO: it would be useful to track errors
+          };
+          document.body.appendChild(script);
+        } catch (e) {
+          // @TODO: it would be useful to track errors
+        }
```
Haroenv, line 89: alternative that could be interesting, so it can be at least seen in the console

```suggestion
          script.onerror = () => {
            throw new Error('.... insights failed to load ...');
          };
          document.body.appendChild(script);
        } catch (e) {
          throw new Error('.... insights failed to load ...');
        }
```

Haroenv: with some reorganisation, the middleware could likely also be found to be removed, but not sure if that's right to do?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
dhayab: Starting from the version we release this:
- if customers instantiate the middleware with a nooped insights client → we do nothing, it won't send events (but this is expected)
- if customers instantiate the middleware with an insights client → we do nothing, it will send events
- if customers instantiate the middleware without an insights client → we download it for them, it will send events
- if customers don't instantiate the middleware → we instantiate the middleware and download the client, it will send events

Am I missing a scenario?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
dhayab: > alternative that could be interesting, so it can be at least seen in the console

What would we recommend the customer to do in this case? Redirect them to the installation guide part around Insights? In dev environment it's unlikely they would have CSP headers set so I don't know what is the guidance we can provide them by showing an error.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -33,32 +42,55 @@ export type InsightsProps<
...
+          };
+
+        insightsClient = window[pointer];
+
+        try {
+          const script = document.createElement('script');
+          script.async = true;
+          script.src = ALGOLIA_INSIGHTS_SRC;
+          script.onerror = () => {
+            // @TODO: it would be useful to track errors
```
dhayab, line 84: Going back to a comment made during discovery, would it be interesting if we add an Insights (Error) (or similar) string in the Algolia user agent when an error occurs?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: Yes, that's correct. We used to imply somewhat that people should add an insights client or manually pass `null` or a `noop`. As `undefined` (most likely argument) now loads automatically, we don't error in that case anymore

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: Yes, there's quite a few different reasons why it could fail, so the error message wouldn't be easy to write

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -118,6 +118,54 @@ describe('insights', () => {
...
+          />
+        </body>
+      `);
+      expect((window as any).AlgoliaAnalyticsObject).toBe('aa');
+      expect((window as any).aa).toEqual(expect.any(Function));
+    });
+
+    it.todo('notifies when the script fails to be added');
+
+    it.todo('notifies when the script fails to load');
```
sarahdayan, line 166: We could at least warn in dev mode.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: One way we could keep this working is if we're able to differentiate the user passing the middleware vs. us passing the middleware. We could have a an internal `createInsightsMiddleware` (not exported) that we use and allows undefined client, and the one users use that still doesn't.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -33,32 +42,55 @@ export type InsightsProps<
...
+        insightsClient = window[pointer];
+      }
+
+      if (!insightsClient) {
+        window.AlgoliaAnalyticsObject = pointer;
+        window[pointer] =
+          window[pointer] ||
+          function (...args: any[]) {
+            (window[pointer].queue = window[pointer].queue || []).push(args);
+          };
```
sarahdayan, line 75: Should we test the queue now that we have it implemented on our end?

Maybe:
- The assignment then swap as the library is loaded
- The processing of events sent before loading

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: I would go with warnings, not errors, as we shouldn't break the app for not being able to load Insights, no?

A warning in dev mode with a message like this:

>InstantSearch wasn't able to automatically load the Insights client. This can happen for a number of reasons; please check our troubleshooting guide for more info: https://algolia.com/doc/...

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: but then, would you only auto-load for the default implementation? it means that if someone wants to do a small change (eg filter out a certain widget, or pass initParams), they need to load insights themselves.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: an error in an async context doesn't break the page, but warning could also work.

Do we know what to suggest to people though?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: that would mean we'd have to mock or somehow wait on the code from jsDelivr right? As that's the code which actually processes the queue.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: Isn't it already the case?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: dev-only has a downside that the most likely situation it would fail in is production (due to stricter CSP setup than localhost usually). That's why I was more thinking about errors

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: That's my point. We could preserve the existing behavior of _not_ accepting that the user passed an undefined `insightsClient`. There's no reason for them to do so if they're manually setting up the middleware, right?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: The use cases we know about are:
- CSP, they need to accept jsDelivr or manually load Insights + set up the middleware
- jsDelivr is down, they can manually load Insights + set up the middleware for 100% success

And we could still send warnings/errors in production indeed. Maybe errors are better since it could be reported by Sentry or Bugsnag if they use it?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: Yes. That could be an opportunity to use something like [MSW](https://mswjs.io/docs/) to mock the service being queried rather than write complicated mocks ourselves.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: Could be interesting, but it adds something to the search client, but it technically has nothing to do with it

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: to make this work, we'll need to find a way to pass `runScripts: "dangerously"` to JSDOM, I couldn't find a way to do that on a cursory glance yet

Haroenv: Let's do this in a follow-up PR. Could be interesting indeed, but I think it may require quite a lot of unrelated changes in configuration etc.

sarahdayan, line 75: It's a few more lines, but I find this a lot more readable than with all the short-circuit assignments. wdyt?

I tried it locally and the bundle size difference for `instantsearch.js/dist/instantsearch.production.min.js` is 0.01KB.

```suggestion
        if (!window[pointer]) {
          window[pointer] = (...args: any[]) => {
            if (!window[pointer].queue) {
              window[pointer].queue = [];
            }

            window[pointer].queue.push(args);
          };
        }
```

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
dhayab: That's true, I'll resolve this in favor of [this comment](https://github.com/algolia/instantsearch/pull/5484#discussion_r1112121564).

dhayab (approved):

PR #5486, opened 2023-02-20
chore(toWarnDev): provide the warning in the match

Haroenv:
This allows for easier debugging when using `.not.toWarnDev` (it will mention what the actual warning is)

### DIFF
--- a/tests/utils/matchers/toWarnDev.ts
+++ b/tests/utils/matchers/toWarnDev.ts
@@ -5,9 +5,9 @@ import { diff } from 'jest-diff';
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace jest {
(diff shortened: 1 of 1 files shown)


sarahdayan (commented):
Can you add a [test](https://github.com/algolia/instantsearch/blob/3159afe57472fe2b669dceb5f1ee638b658f7f52/tests/utils/matchers/__tests__/toWarnDev-test.ts)?

Haroenv:
I don't think the assertion message is testable without going too much into implementation details. None of the other messages are tested either @sarahdayan. It's what's shown to the user when the assertion passes (but it's not meant to, so using `.not.`, but assertion messages can't be asserted on

sarahdayan (approved):

PR #5488, opened 2023-02-20
feat(insights): automatically add insights middleware

Haroenv:
**Summary**

Automatically add the insights middleware

**Result**

[FX-2248](https://algolia.atlassian.net/browse/FX-2248)

- add the insights middleware automatically
- identify internal vs. user-provided middleware
- make sure adding or removing the automatic insights middleware doesn't cause a search
- when a user adds a different insights middleware, it takes precedence
- update a few tests that now have `clickAnalytics: true`

[FX-2248]: https://algolia.atlassian.net/browse/FX-2248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aymeric-giraudet (approved):

aymeric-giraudet:
Nice ! 🙌

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -297,12 +298,17 @@ See ${createDocumentationLink({
...
     if (routing) {
-      const routerOptions = typeof routing === 'boolean' ? undefined : routing;
+      const routerOptions = typeof routing === 'boolean' ? {} : routing;
+      routerOptions.$$internal = true;
       this.use(createRouterMiddleware(routerOptions));
     }
 
+    // This is the default middleware,
+    // any user-provided middleware will be added later and override this one.
+    this.use(createInsightsMiddleware({ $$internal: true }));
```
Haroenv, line 308: something still needs to be fixed here, the script is always loaded from the internal middleware, even if users add a different middleware later (even right after). Not too sure how to fix that yet

Haroenv: solved through https://github.com/algolia/instantsearch/pull/5493

packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -153,6 +153,11 @@ function getStateWithoutPage(state: PlainSearchParameters) {
   return rest;
 }
 
+function normalizeState(state: PlainSearchParameters) {
+  const { clickAnalytics, userToken, ...rest } = state || {};
+  return rest;
+}
```
Haroenv, line 159: if infiniteHits is added before start, it's reading from the cache before the insights middleware is started, and thus it doesn't have clickAnalytics + userToken yet

Haroenv: to solve this, we could change some things about the insights middleware to make sure we have the search parameters right away, even before init, but that's currently not possible yet (helper is null)

PR #5493, opened 2023-02-21
feat(insights): option to disable automatic insights middleware

Haroenv:
instead of passing your own middleware, where timing could be tricky, to disable the internal middleware you pass `insights: false` to InstantSearch. Note that the internal middleware will still be removed when a different middleware is passed to allow easy customisation

FX-2250

### DIFF
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -136,7 +136,7 @@ export type InstantSearchOptions<
   /**
    * Automatically enables the Insights middleware and loads the Insights library
    * if not already loaded.
-   * 
+   *
    * The Insights middleware sends view and click events automatically, and lets
    * you set up your own events.
    *
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -134,8 +134,11 @@ export type InstantSearchOptions<
   routing?: RouterProps<TUiState, TRouteState> | boolean;
 
   /**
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -133,6 +133,14 @@ export type InstantSearchOptions<
    */
   routing?: RouterProps<TUiState, TRouteState> | boolean;
 
+  /**
+   * Enables the insights middleware. This middleware will send view and click events,
+   * as well as allowing to set up your own events.
+   *
+   * @default true
+   */
```
sarahdayan, line 141:
```suggestion
  /**
   * Automatically enables the Insights middleware and loads the Insights library
   * if not already loaded.
   * 
   * The Insights middleware sends view and click events automatically, and lets
   * you set up your own events.
   *
   * @default true
   */
```

sarahdayan (approved):

packages/instantsearch.js/src/lib/InstantSearch.ts
sarahdayan: Clarity police 👮🏻

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -133,6 +133,17 @@ export type InstantSearchOptions<
    */
   routing?: RouterProps<TUiState, TRouteState> | boolean;
 
+  /**
+   * Automatically enables the Insights middleware and loads the Insights library
+   * if not already loaded.
+   *
```
FabienMotte, line 139:
```suggestion
   *
```

FabienMotte (approved):

PR #5495, opened 2023-02-23
fix(infinitehits/hits): correctly deprecate __hitIndex

Haroenv:
**Summary**

Add a real warning and deprecation for __hitIndex (the position within the visible page)

**Result**

JSDoc comments only work when in a docblock, so that's changed.

Also added a proper warning so string templates see this too.


aymeric-giraudet (approved):

FabienMotte (approved):

Haroenv:
Hmm, I need to find a way that doesn't warn on the default template

PR #5500, opened 2023-02-23
fix(analytics): generate fallback userToken

Haroenv:
**Summary**

**Result**


packages/instantsearch.js/.storybook/playgrounds/default.ts
```diff
@@ -129,13 +129,13 @@ const instantSearchPlayground: Playground = function instantSearchPlayground({
...
   ]);
 
-  const insights = instantsearch.middlewares.createInsightsMiddleware({
-    insightsClient: null,
-    onEvent: props => {
-      console.log('insights onEvent', props);
-    },
-  });
-  search.use(insights);
+  // const insights = instantsearch.middlewares.createInsightsMiddleware({
```
Haroenv, line 132: to revert

sarahdayan:
With the RFC on decoupling (or using v3) we'll likely not need this. Should we close @Haroenv?

PR #5510, opened 2023-02-27
fix(insights): pass credentials in payload

Haroenv:
**Summary**

If something else controls insights client too, it could be initialised with the wrong parameters. To ensure the right app is targeted, we pass appId and apiKey in the payload as well.

This requires https://github.com/algolia/search-insights.js/pull/398 or similar to have any effect though

**Result**

appId/apiKey is in event payload

### DIFF
--- a/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
@@ -226,6 +226,10 @@ export function createInsightsMiddleware<
         });
 
         instantSearchInstance.sendEventToInsights = (event: InsightsEvent) => {
+          // Pass credentials in event payload, in case init was called with different credentials
+          event.payload.appId = event.payload.appId || appId;
+          event.payload.apiKey = event.payload.apiKey || apiKey;
+
           if (onEvent) {
             onEvent(event, _insightsClient as TInsightsClient);
           } else if (event.insightsMethod) {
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -215,6 +215,10 @@ export function createInsightsMiddleware<
         });
 
         instantSearchInstance.sendEventToInsights = (event: InsightsEvent) => {
+          // Pass credentials in event payload, in case init was called with different credentials
+          event.payload.appId = event.payload.appId || appId;
+          event.payload.apiKey = event.payload.apiKey || apiKey;
+
```
dhayab, line 221: With the latest direction (https://github.com/algolia/search-insights.js/pull/398, will be backported to v2 once it's settled), credentials will probably be added as custom headers through a third argument of `insights()`.

Haroenv: that's a little annoying for `onEvent` as it takes only one argument now 🤔

dhayab: A customer-defined onEvent won't need to handle custom credentials, no? Or am I assuming too much?

Haroenv: I was just thinking of the default use case of "I'm sending to another service, but still using algolia as well". They'd have to know the second arg of insightsClient too when they're in the "multiple apps" situation too

Haroenv: alternatively what we could do is `onEvent(event, Object.assign(insightsClient, credentials), insightsClient)` as that's backwards compatible still

Haroenv: I'm mainly seeing this as a possible issue as we document `onEvent(event, aa) { aa(event); ... }`, which would be forced to be different, causing people to possibly lose credentials. That might not be too much of an issue when we still init, but I'm on the fence

sarahdayan:
Closed in favor of #5554.

PR #5512, opened 2023-02-27
fix(userToken): prevent search waterfall

Haroenv:
**Summary**

This makes sure when the same userToken is set multiple times, it doesn't cause multiple search requests, as could be the case if something would call setUserToken without condition.

Discovered in #5500, but should be merged separately.

**Result**

if userToken is set multiple times, we make sure to not search if it's already the same as before.

### DIFF
--- a/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -110,6 +110,8 @@ describe('insights', () => {
     (window as any).aa = undefined;
 
     document.body.innerHTML = '';
+
+    document.cookie = '_ALGOLIA=;';
   });
 
   describe('usage', () => {
@@ -625,8 +627,7 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
     });
 
     it('searches once per unique userToken', async () => {
-      const { insightsClient, instantSearchInstance } =
-        createTestEnvironment();
+      const { insightsClient, instantSearchInstance } = createTestEnvironment();
 
       instantSearchInstance.addWidgets([connectSearchBox(() => ({}))({})]);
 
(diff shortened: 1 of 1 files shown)


packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -621,6 +624,31 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
       ).toEqual('def');
     });
 
+    it('searches once per unique userToken', async () => {
+      const { insightsClient, instantSearchInstance, getUserToken } =
```
FabienMotte, line 628:
```suggestion
      const { insightsClient, instantSearchInstance } =
```

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -186,12 +186,21 @@ export function createInsightsMiddleware<
         }
 
         const setUserTokenToSearch = (userToken?: string) => {
+          const existingToken = (helper.state as PlainSearchParameters)
```
sarahdayan, line 189: This can be computed after the early return you added in line 192.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -186,12 +186,21 @@ export function createInsightsMiddleware<
...
+            .userToken;
+
+          if (!userToken) {
+            return;
+          }
```
sarahdayan, line 194: Is this already tested?

Haroenv: it's in the new test, right?

Haroenv: tested in cc0277530

FabienMotte (approved):

sarahdayan (approved):

PR #5521, opened 2023-03-02
ci: run e2e in a scheduled pipeline

dhayab:
**Summary**

This PR stops running e2e tests on every branch and instead moves the job into a scheduled pipeline. It's currently set to run daily at 9PM CET. In case of failure a notification is sent to Slack.

It also moves the existing weekly release scheduled workflow to a scheduled pipeline as well, managed through Circle CI.

**Result**

CI pipelines are now completed in less time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,8 @@
 version: 2.1
 
+orbs:
+  slack: circleci/slack@4.12.1
+
 references:
   workspace_root: &workspace_root ~/instantsearch
   attach_workspace: &attach_workspace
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -51,37 +57,44 @@ workflows:
...
           requires:
             - build
-      - examples:
-          requires:
-            - build
-      - e2e tests:
-          context: fx-libraries
-          requires:
-            - examples
       - release if needed:
```
Haroenv, line 60: should the release possibly run e2e as well? I can see the pros and cons of both

.circleci/config.yml
```diff
@@ -230,6 +243,12 @@ jobs:
           command: yarn run test:e2e:saucelabs
       - store_test_results:
           path: junit/wdio/
+      - slack/notify:
+          channel: $SLACK_E2E_NOTIF_CHANNEL
+          event: fail
+          # See: https://app.slack.com/block-kit-builder/
+          custom: |
+            {"blocks":[{"type":"header","text":{"type":"plain_text","text":":warning: End-to-end tests failed","emoji":true}},{"type":"section","fields":[{"type":"mrkdwn","text":"*Project*: $CIRCLE_PROJECT_REPONAME"}]},{"type":"actions","elements":[{"type":"button","text":{"type":"plain_text","text":"CircleCI"},"url":"$CIRCLE_BUILD_URL"},{"type":"button","text":{"type":"plain_text","text":"SauceLabs"},"url":"https://app.saucelabs.com/dashboard/tests/vdc"}]}]}
```
Haroenv, line 251: can this json be formatted?

Haroenv (commented):
👍

.circleci/config.yml
dhayab: Making it a prerequisite to `release if needed` would work if we filter the job to be run on tags, but since we're now pushing multiple tags at once, it would run unnecessarily multiple times. Alternatively, we could in the end run e2e tests on master pipelines as well as in a daily schedule. It would still be better than having it run for all branches. Wdyt?

Haroenv: I think running it on master may be the most "correct"

Haroenv (approved):

FabienMotte (approved):
Awesome! 👏

PR #5523, opened 2023-03-02
fix(insights): call `init` only if it has not been called before

FabienMotte:
**Summary**

[FX-2272](https://algolia.atlassian.net/browse/FX-2272)

If `init` has already been called outside on `search-insights`, calling `init` in the insights middleware would clear the previous options and we want to avoid that behavior.

**Result**

We make sure `init` is called only if it has not been called before in the `sendEventToInsights` method:

- First, we try to send the event.
- If it fails, we catch the error, meaning that `search-insights` is not initialized yet and we `init` before sending the event.
- Then, if the `search-insights` queue length has changed in the meantime, it means that `search-insights` is not yet loaded.
- If `init` has not already been added to the queue, we add it and make sure to put it in the first position in the queue.


Haroenv:
let's make this pr to feat/automatic-insights maybe?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -230,7 +224,42 @@ export function createInsightsMiddleware<
...
+            try {
+              // We first try to send the event.
+              insightsClient(event.insightsMethod, event.payload);
+            } catch (e) {
+              // We catch if `search-insights` is not initialized yet
+              // and we `init` before sending the event.
+              insightsClient('init', {
+                appId,
+                apiKey,
+                useCookie: true,
```
Haroenv, line 239: maybe these should be factored inside insightsInitParams?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -230,7 +224,42 @@ export function createInsightsMiddleware<
           if (onEvent) {
             onEvent(event, _insightsClient as TInsightsClient);
           } else if (event.insightsMethod) {
-            insightsClient(event.insightsMethod, event.payload);
+            // We keep the queue length for later.
```
Haroenv, line 227:
```suggestion
            // We keep the queue length to check whether the client is already loaded
```

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -230,7 +224,42 @@ export function createInsightsMiddleware<
...
+              });
+              insightsClient(event.insightsMethod, event.payload);
+            }
+
+            if (
+              // If the `search-insights` queue length has changed, it means
+              // that `search-insights` is not yet loaded.
+              (insightsClient.queue?.length || 0) > queueLength &&
+              // We check if `init` has already been added to the queue.
+              insightsClient.queue?.every((el) => el[0] !== 'init')
```
Haroenv, line 250:
```suggestion
              insightsClient.queue?.every(([eventType]) => eventType !== 'init')
```

can we destructure?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv, line 239: once that's merged/stabilised, we should also add the "patch" argument

Haroenv (commented):
I think this is the right approach

Haroenv (commented):
I just thought of a flaw on this approach. It means that we don't apply the userToken that gets created by `useCookie`/`setUserToken` _before_ the first event

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
FabienMotte: Good idea 👍

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
FabienMotte: Is that what you were thinking? https://github.com/algolia/instantsearch/pull/5523/commits/531ce681ba3b4336f916b61f4f66e4859bf8d707

FabienMotte:
> I just thought of a flaw on this approach. It means that we don't apply the userToken that gets created by `useCookie`/`setUserToken` _before_ the first event

I'm not sure to understand how it can be an issue?

Haroenv:
I'd like to see what happens when a customer has `setUserToken` (but not init), does that still get updated in the helper state? Essentially: see the tests that are breaking now

PR #5524, opened 2023-03-02
fix(events): deduplicate equal events

Haroenv:
- adds a temporary cache for events
- only the first event of a type is sent (fixing #5442)
- if there's a new result, we send again
- if multiple events happen at the same time (default, and custom), they're only deduplicated when they have the same even type
- a double click or multiple clicks over time still get deduplicated, we could solve this by clearing the cache after a certain timeout _as well_

FX-2246

see also #5460, a previous approach

draft because tests aren't updated yet


Haroenv:
While interesting for view, we'll not go this approach.

PR #5527, opened 2023-03-03
fix(insights): user-defined click overrides 

Haroenv:
in sendEvent, we mark internal events (click and conversion), and prevent them being sent if there was a custom click right before.

Note that this doesn't yet work with bindEvent, as it's higher in the bubbling tree

FX-2246


packages/instantsearch.js/src/widgets/infinite-hits/__tests__/infinite-hits-integration-test.ts
```diff
@@ -486,9 +486,9 @@ describe('infiniteHits', () => {
 
       fireEvent.click(getByText(container, 'title 2'));
 
-      // The default `click` one + the custom one
-      expect(onEvent).toHaveBeenCalledTimes(2);
-      expect(onEvent.mock.calls[onEvent.mock.calls.length - 1][0]).toEqual({
+      // The custom one only
```
sarahdayan, line 489: Shouldn't it send the built-in `click` event as well? We're not dropping it when it's a `conversion` event.

Haroenv: It's with `bindEvent`, so we don't know (easily) what kind of event was sent. Do you think it's needed to check inside the default event handler what kind of event is bound?

sarahdayan: Can we spend a bit more time to dig into how we can make it happen?

I understand we might be technically limited, but this differs from the scenario we agreed on in the [RAPID](https://docs.google.com/document/d/1HiPY7lK6sT6HNA9K0PP-El9yFSq-ZM74V6YCnlchH1k/edit#heading=h.jk50s25qbuc3). If we eventually realize it's not possible, it should be known in the team so it's not considered as a bug, and probably documented.

Haroenv: In later commits I've made this work by moving bindEvent up the tree + sharing the "dedupe" logic between sendEvent + bindEvent calls :) This simplifies the logic significantly

packages/instantsearch.js/src/lib/insights/listener.tsx
```diff
@@ -43,7 +43,7 @@ export const createInsightsEventHandler =
     }
   };
 
-function findInsightsTarget(
+export function findInsightsTarget(
```
sarahdayan, line 46: You can unexport it.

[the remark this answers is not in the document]
Haroenv:
```suggestion
function findInsightsTarget(
```

tests/common/widgets/hits/insights.ts
dhayab, line 1: This and the infinite hits insights test look really similar. Do you think we could go one step further and rely on a reusable test suite for both?

dhayab (approved):

tests/common/widgets/hits/insights.ts
Haroenv: I tried it, but didn't get to a suitable shape for the tests. The common tests shape will become clearer when we have more of them I think

PR #5528, opened 2023-03-06
refactor(bindEvent): move listener inside components

Haroenv:
**Summary**

This was needlessly abstracted with a HOC, not allowing us to use this same function inside onClick of an individual hit, which could help with deduplicating sendEvent + bindEvent

I didn't remove the `withInsightsListener` HOC, as it could (in theory) be used externally.

**Result**

- no empty div around hits/infiniteHits
- Hits/InfiniteHits components own their event life cycle completely
- better segmentation of v1 "insights helper" API


packages/instantsearch.js/src/lib/insights/listener.tsx
```diff
@@ -2,23 +2,58 @@
...
+        element.hasAttribute('data-insights-method') &&
+        element.hasAttribute('data-insights-payload')
+    );
+    if (insightsThroughFunction) {
+      const { method, payload } = readDataAttributes(insightsThroughFunction);
+      insights!(method, payload);
+      hasSentAnEvent = true;
+    }
+
+    return hasSentAnEvent;
```
FabienMotte, line 49: I'm not sure to understand why we need to return a boolean from the handler?

Haroenv: it's not really needed for this PR indeed, it was with #5527 in mind. I'll revert here

FabienMotte (approved):
Nice! 👏

PR #5529, opened 2023-03-06
fix(insights): only `init` if `version` allows it

FabienMotte:
**Summary**

[FX-2283](https://algolia.atlassian.net/browse/FX-2283)

Alternative to #5523 

If `init` has already been called outside on `search-insights`, calling `init` in the insights middleware would clear the previous options and we want to avoid that behavior.

**Result**

- `init` is not called when the default Insights middleware is used.
- `init` is called with `partial: true` option to update options without clearing them when a user pass a custom middleware.
- `init` is called when `insightsInitParams` are passed (even if it's the default Insights middleware).

[FX-2283]: https://algolia.atlassian.net/browse/FX-2283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -136,12 +136,26 @@ export function createInsightsMiddleware<
...
-    insightsClient('init', {
-      appId,
-      apiKey,
-      useCookie: true,
-      ...insightsInitParams,
-    });
+
+    const version = insightsClient.version;
+    let canInit = false;
+    if (typeof version !== 'undefined') {
```
Haroenv, line 142: sorry, I think this is a breaking change, as the current version of the plugin doesn't check. Maybe if we add && $internal?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -151,7 +165,7 @@ export function createInsightsMiddleware<
       $$internal,
       onStateChange() {},
       subscribe() {
-        if (!needsToLoadInsightsClient) return;
+        if (!needsToLoadInsightsClient || !canInit) return;
```
Haroenv, line 168: I think we should just let it have errors, not skip everything

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
FabienMotte: Indeed, I forgot to based the condition on the `$$internal` flag, thanks!

Haroenv (commented):
this should be good once those versions are released

FabienMotte:
@Haroenv Thanks! Should I close #5523 then?

Haroenv:
Yes, I think #5523 is a dead end. We may need to do it later still if there's an option coming from the response, but that doesn't block this PR I think.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -136,12 +136,28 @@ export function createInsightsMiddleware<
...
-      apiKey,
-      useCookie: true,
-      ...insightsInitParams,
-    });
+
+    const version = insightsClient.version;
+    let canInit = !$$internal; // User provided insights middleware.
+
+    // We only check the `version` for the default insights middleware.
+    if ($$internal && typeof version !== 'undefined') {
```
Haroenv, line 144: maybe this means now that if it's internal, we just _don't_ init, as the version which will allow partial init is the same one that doesn't need init anymore? We'll need to update tests of course, as for now this is still an opt-out branch

FabienMotte: Yes it makes sense to me, `init({ partial: true })` should be the same as no `init` now.

Haroenv: as well as the useCookie still defaulting to true, it makes sense to keep this branch as-is now, but later we'll be able to remove likely

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -82,7 +82,7 @@ export function createInsightsMiddleware<
             window[pointer].queue.push(args);
           };
         }
-
+        window[pointer].version = VERSION;
```
Haroenv, line 85: this should be in the previous branch, as otherwise it may override a different version already said?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -82,7 +82,7 @@ export function createInsightsMiddleware<
...
         insightsClient = window[pointer];
         insightsClient.shouldAddScript = true;
```
Haroenv, line 87: maybe this statement should also be inside the !window[pointer] branch?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -170,6 +189,8 @@ export function createInsightsMiddleware<
         }
       },
       started() {
+        if (!canInit) return;
```
Haroenv, line 192: is it better to have a middleware that does nothing (like this), or to just make it work without canInit, continuing to assume you have to call `init` yourselves.

I'd just remove this early return, and likely everything works fine, right?

Haroenv (changes requested):
I'm not convinced we should return if we can't init (as it realistically should be optional or done by the user)

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: did this require a new test, or did the existing test already cover the branch?

tests/mocks/createInsightsClient.ts
```diff
@@ -16,22 +24,22 @@ export function createInsights<TVersion extends string | undefined = '2.4.0'>({
       requestFn: jest.fn(),
     })
   );
+  const mockedInsightsClient = jest.fn(
+    getFunctionalInterface(analytics)
+  ) as unknown as jest.Mock<InsightsClient>;
```
FabienMotte, line 29: We had an error in the `__tests__/createInsightsMiddleware.ts` file:
```
Type 'Mock<void, [method: keyof InsightsMethodMap, options?: InitParams | undefined] | [method: keyof InsightsMethodMap, callback: (version: string) => void] | ... 12 more ... | [method: ...]>' is not assignable to type 'ProvidedInsightsClient'.
```
Not sure if there is a better way to fix this.

Haroenv: you can use the `castToJestMock` utility function we have that does the same

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -138,12 +139,17 @@ export function createInsightsMiddleware<
...
     });
-    insightsClient('init', {
-      appId,
-      apiKey,
-      useCookie: true,
-      ...insightsInitParams,
-    });
+
+    // Only `init` if user provided the `insights` middleware.
+    if (!$$internal) {
```
Haroenv, line 144:
```suggestion
    if (!$$internal || insightsInitParams) {
```

this would require another test, but you can pass the init params to the insights option, so in that case we should still init.

Haroenv (changes requested):
👍 I think this is the way to go!

Haroenv (approved):

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
FabienMotte: Good catch, thanks Haroen, I added a test and updated the PR description.

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -581,6 +569,74 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
         ]
       `);
     });
+
+    it('does not call `init` when default middleware is used', () => {
```
sarahdayan, line 573: This assumes the "default" middleware always uses a version that's able not to `init`, which might not be the case if Search Insights is already present but using an older v2 version, no?

Haroenv: good point, maybe it should be "shouldAddScript" as the condition instead?

FabienMotte: Should we update the condition to something like this:
```
if (!$$internal || insightsInitParams || !isModernInsightsClient(insightsClient)) {
  // ...
}
```

sarahdayan: @FabienMotte Would `isModernInsightsClient` use the `version`?

FabienMotte: Yes, the function already exists: https://github.com/algolia/instantsearch/blob/52f54bfb408c165f07184252f3737853b2c98a6c/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts#L315

sarahdayan: Yeah, so the only scenarios when `init` should be called are **when the user provides `insightsInitParams`** _or_ **when the Insights version doesn't support not calling `init`**.

But the fact that it's internal or not doesn't count here, am I correct?

sarahdayan: I'd also clarify the condition, because `isModernInsightsClient` doesn't tell us what it actually supports.

FabienMotte: WDYT of `supportsOptionalInitMethod(insightsClient)`?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -246,7 +247,7 @@ export function createInsightsMiddleware<
         let insightsClientWithLocalCredentials =
           insightsClient as InsightsClientWithLocalCredentials;
 
-        if (isModernInsightsClient(insightsClient)) {
+        if (supportsOptionalInitMethod(insightsClient)) {
```
Haroenv, line 250: I liked modern more, as it is optional init + credentials through arguments. It's fine like this too of course

sarahdayan: Either let's convey this with an explicit name, or use a comment, wdyt?

Haroenv: I guess technically "supportsOptionalInitMethod" covers both usages of the function, but I wonder if it should be "supportsCredentialsInPayload" or similar?

FabienMotte: The function is called in both places, the first one being to check if the optional `init` method is supported and the other one is to check if the credentials are supported as extra params.

So maybe we should keep a generic function name like `isModernInsightsClient()` that encapsulates both behavior and add comments at both places of use, WDYT?

Haroenv: I think `modern` makes sense, but a jsdoc comment to the function saying it has: "partial init" and "credentials via event" maybe?

FabienMotte: Done, thanks!

Haroenv (approved):

dhayab (approved):
LGTM. I'm stealing the comment description of `isModernInsightsClient()` for my PR 😛!

FabienMotte:
Thanks a lot @dhayab!

PR #5532, opened 2023-03-07
fix(DynamicWidgets): prevent non-stable fallbackComponent

### LINKED ISSUE #5530: createInstantSearchRouterNext without SSR clears state when filters are applied
### 🐛 Current behavior

Clicking refinements briefly shows the filtered items, and then reverts the selection when using createInstantSearchRouterNext in client-side only mode.

### 🔍 Steps to reproduce

1. Clone https://github.com/jmotes-lapero/algolia-issues-repro
2. `npm install`
3. `npm run dev`
4. Try to apply facet filters

### Live reproduction

https://codesandbox.io/s/github/jmotes-lapero/algolia-issues-repro

### 💭 Expected behavior

Checkboxes should persist and search results filtered by the new facet value.

### Package version

react-instantsearch-hooks-router-nextjs 6.40.1

### Operating system

Ubuntu Linux

### Browser

Chrome

### Code of Conduct

Haroenv:
All our documentation makes sure to use stable references, although it could also make sure to clarify that is a requirement as well.

**Summary**

When you re-render and use a fallbackComponent which isn't stable, this causes that widget to unmount and remount (as a different widget). We now prevent that by making sure (and warning when it isn't) that fallbackComponent is stable.

Note that this could be considered breaking if someone was _only_ rendering when the widget actually changed (some other state possibly), but I'd consider that code already broken personally, as a render happening when the widget doesn't change would erase the state.

fixes #5530

**Result**

- add a warning when the fallbackComponent isn't stable
- only ever use the first value of fallbackComponent

### DIFF
--- a/packages/react-instantsearch-hooks/src/components/DynamicWidgets.tsx
+++ b/packages/react-instantsearch-hooks/src/components/DynamicWidgets.tsx
@@ -5,8 +5,9 @@ import { invariant } from '../lib/invariant';
 
 import type { DynamicWidgetsConnectorParams } from 'instantsearch.js/es/connectors/dynamic-widgets/connectDynamicWidgets';
 import type { ReactElement, ComponentType, ReactNode } from 'react';
+import { warn } from '../lib/warn';
 
-function FallbackComponent() {
+function DefaultFallbackComponent() {
   return null;
 }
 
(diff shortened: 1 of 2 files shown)


aymeric-giraudet (approved):

packages/react-instantsearch-hooks/src/components/DynamicWidgets.tsx
```diff
@@ -26,9 +27,16 @@ export type DynamicWidgetsProps = Omit<
...
-  fallbackComponent: Fallback = FallbackComponent,
+  fallbackComponent: Fallback = DefaultFallbackComponent,
   ...props
 }: DynamicWidgetsProps) {
+  const FallbackComponent = React.useRef(Fallback);
+
+  warn(
+    Fallback === FallbackComponent.current,
+    'The `fallbackComponent` prop of `DynamicWidgets` changed between renders. Please provide a stable reference.'
+  );
```
dhayab, line 38: Would adding a link to the doc and update it to show an example of how to achieve that help?

I'm thinking of this part: https://www.algolia.com/doc/api-reference/widgets/dynamic-facets/react-hooks/#widget-param-fallbackcomponent.

Haroenv: is 0f6c9fe clear?

dhayab (approved):

PR #5535, opened 2023-03-07
test(jest): runInBand

Haroenv:
**Summary**

This should make the "modules" tests no longer flaky, as the mocks happen one by one, not simultaneous by using jest's runInBand option

**Result**

unit tests will no longer fail because of "modules-none" etc. tests of React InstantSearch Hooks checking for presence of react


Haroenv:
I just remembered I've tried this in https://github.com/algolia/instantsearch/pull/5232 already, and then it didn't work, maybe different now?

PR #5546, opened 2023-03-14
feat(current-refinements): provide indexId of refinements in transformItems

dhayab:
**Summary**

The `currentRefinements` widget provides a way to transform/filter refined items to be returned, which is useful in multi-indices context, to potentially remove refinements from another index.

This works well when indices are different, but in the context of multi-indices targeting the same index with different configuration, it is impossible to easily determine which refinements are from where. 

This PR adds a `indexId` parameter to the items provided in `transformItems`, which solves this issue.


Haroenv (approved):

Haroenv (changes requested):
Well I guess typescript needs to pass first, but code looks good

Haroenv (approved):

PR #5547, opened 2023-03-15
fix(searchbox): add aria-hidden to svg icons

pat-sullivan:
**Summary**
Added `aria-hidden="true"` attributes to the following SVG icons used in the repo: 
- ais-SearchBox-resetIcon
- ais-SearchBox-loadingIcon
- ais-SearchBox-submitIcon

This is in response to a Shopify customer support ticket saying these icons were flagged (for missing title attributes) during an accessibility audit. Since these icons are nested below buttons with title attributes, an aria-hidden attribute seemed the better option.


packages/instantsearch.js/src/widgets/search-box/__tests__/search-box.test.tsx
```diff
@@ -65,6 +65,7 @@ describe('searchBox', () => {
           height="10"
           viewBox="0 0 40 40"
           width="10"
+          aria-hidden="true"
```
Haroenv, line 68: is there a forgotten commit? I can see all the test updates, but the code didn't change 🤔

Haroenv (changes requested):
Thanks @pat-sullivan, this looks like a great start! we should indeed have these images aria-hidden, as they are purely visual and have a parent with a title

packages/instantsearch.js/src/widgets/search-box/__tests__/search-box.test.tsx
Haroenv: react-instantsearch-hooks-web/ui is updated, but not React InstantSearch dom / components, InstantSearch.js / components, Vue InstantSearch / components :)

pat-sullivan: You're right 😅 I did a search for exact match classnames, but it looks like some are dynamically built. Thanks for catching. Will update soon.

packages/vue-instantsearch/src/components/__tests__/__snapshots__/RefinementList.js.snap
```diff
@@ -46,6 +47,7 @@ exports[`allows search bar classes override when it's searchable 1`] = `
              role="img"
              viewbox="0 0 20 20"
              width="1em"
+             aria-hidden="true"
```
Haroenv, line 50: these Vue snapshots need aria-hidden as the first property (it's alphabetically sorted)

pat-sullivan: @Haroenv Thanks for pointing that out! I've made the necessary adjustments. However I can't seem to get past that flaky unit test on `packages/react-instantsearch-hooks-server/src/__tests__/modules-none.test.tsx`

Haroenv (approved):
thanks! This looks good to me

PR #5549, opened 2023-03-15
feat(hits/infinite-hits): send a default click event and expose sendEvent to slot

Haroenv:
it's like FX-2271, but for Vue.

I recommend you review this PR with "ignore whitespace" enabled, it's quite confusing otherwise

### DIFF
--- a/packages/vue-instantsearch/src/util/vue-compat/index-vue3.js
+++ b/packages/vue-instantsearch/src/util/vue-compat/index-vue3.js
@@ -9,13 +9,34 @@ export { Vue, Vue2, isVue2, isVue3 };
 
 export function renderCompat(fn) {
   function h(tag, props, children) {
-    if (typeof props === 'object' && (props.attrs || props.props)) {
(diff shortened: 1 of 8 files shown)


packages/vue-instantsearch/src/util/vue-compat/index-vue3.js
```diff
@@ -9,13 +9,34 @@ export { Vue, Vue2, isVue2, isVue3 };
...
       delete flatProps.props;
+      delete flatProps.scopedSlots;
+      delete flatProps.on;
 
-      return Vue.h(tag, flatProps, children);
+      return Vue.h(
+        tag,
+        flatProps,
+        props.scopedSlots
+          ? { default: () => children, ...props.scopedSlots }
```
Haroenv, line 37: This needs to be in a transpilable form, Vue InstantSearch doesn’t have object spread.

dhayab (approved):

PR #5550, opened 2023-03-16
fix(svg): don't style width/height in attributes with unit

Haroenv:
Browsers warn for a width/height attribute which isn't a number, as it can only be a "unitless value" (px). Despite that it does seem to work.

Components involved:
- SearchBox in Vue
- Panel in JS
- Refresh in hooks example

### DIFF
--- a/packages/vue-instantsearch/src/components/SearchInput.vue
+++ b/packages/vue-instantsearch/src/components/SearchInput.vue
@@ -68,8 +68,7 @@
           aria-hidden="true"
           role="img"
           xmlns="http://www.w3.org/2000/svg"
-          width="1em"
-          height="1em"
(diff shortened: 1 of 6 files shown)


packages/vue-instantsearch/src/components/SearchInput.vue
```diff
@@ -68,8 +68,7 @@
           aria-hidden="true"
           role="img"
           xmlns="http://www.w3.org/2000/svg"
-          width="1em"
-          height="1em"
+          style="width: 1em; height: 1em"
```
dhayab, line 71: We could reuse the unitless dimensions from InstantSearch.js reset icon template.

```suggestion
          width="10"
          height="10"
```

dhayab (approved):

PR #5551, opened 2023-03-16
fix(insights): keep track of script loading in global

Haroenv:
**Summary**

This is needed for example in StrictMode with React InstantSearch Hooks, where there are two InstantSearch instances, but the first one never gets started.

Before this PR, the "should load script" was linked to "should add window.aa", but that doesn't necessarily correspond if `start` is never called on that instance.

**Result**

insights client loads in strict mode

### DIFF
--- a/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -185,6 +185,49 @@ describe('insights', () => {
       expect((window as any).aa).toEqual(expect.any(Function));
     });
 
+    it('loads script, even when globals are set up by a different instance', () => {
+      const { instantSearchInstance: instantSearchInstance1 } =
+        createTestEnvironment({
+          started: false,
+        });
+      const { instantSearchInstance: instantSearchInstance2 } =
+        createTestEnvironment({
+          started: false,
+        });
+
+      expect((window as any).AlgoliaAnalyticsObject).toBe(undefined);
+
+      // middleware is added to first instance
+      instantSearchInstance1.use(createInsightsMiddleware());
+
+      // it sets up globals
+      expect(document.body).toMatchInlineSnapshot(`<body />`);
+      expect((window as any).AlgoliaAnalyticsObject).toBe('aa');
+      expect((window as any).aa).toEqual(expect.any(Function));
+
+      // middleware is set up on second instance
+      instantSearchInstance2.use(createInsightsMiddleware());
+
+      // globals stay as-is
+      expect(document.body).toMatchInlineSnapshot(`<body />`);
+      expect((window as any).AlgoliaAnalyticsObject).toBe('aa');
+      expect((window as any).aa).toEqual(expect.any(Function));
+
+      // only second instance starts
+      instantSearchInstance2.start();
+
+      // which finally loads search-insights
+      expect(document.body).toMatchInlineSnapshot(`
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -185,6 +185,49 @@ describe('insights', () => {
       expect((window as any).aa).toEqual(expect.any(Function));
     });
 
+    it('loads script, even when globals are set up by a different instance', () => {
```
sarahdayan, line 188: It's clear to us now, but in the future this test doesn't tell me much about when it can happen.

Could we have a test in the RISH packages, using strict mode instead?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -61,9 +61,9 @@ export function createInsightsMiddleware<
     $$internal = false,
   } = props;
 
-  let insightsClient: InsightsClient = _insightsClient || noop;
+  let insightsClient: InsightsClient & { needsToLoadInsightsClient?: boolean } =
```
sarahdayan, line 64: Shorter and less redundant, wdyt?

```suggestion
  let insightsClient: InsightsClient & { shouldLoad?: boolean } =
```

Haroenv: I think that's a little unclear, as load could mean many things just like that, but committed `shouldAddScript` now, makes sense

packages/react-instantsearch-hooks/src/__tests__/integration.test.tsx
sarahdayan, line 1: I would go per use case instead of "type of test" and call this test file `insights.test.tsx` instead (just like we already have `compat.test.tsx`.

packages/react-instantsearch-hooks/src/__tests__/integration.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import { createSearchClient } from '@instantsearch/mocks';
+import { render } from '@testing-library/react';
+import React, { StrictMode } from 'react';
+
+import { InstantSearch } from '..';
+
+describe('insights', () => {
+  test('loads insights client in StrictMode', () => {
```
sarahdayan, line 12:
```suggestion
  test('automatically loads the Insights client in `StrictMode`', () => {
```

packages/react-instantsearch-hooks/src/__tests__/integration.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const searchClient = createSearchClient();
+
+    render(
+      <StrictMode>
+        <InstantSearch searchClient={searchClient} indexName="123" />
+      </StrictMode>
+    );
+
+    // sets global
```
sarahdayan, line 21:
```suggestion
    // Sets the global
```

packages/react-instantsearch-hooks/src/__tests__/integration.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    expect((window as any).aa).toEqual(expect.any(Function));
+    // loads script
```
sarahdayan, line 23:
```suggestion
    // Injects the script
```

sarahdayan (approved):

PR #5553, opened 2023-03-17
feat: warn if same attribute is used for different faceting type

aymeric-giraudet:
**Summary**

### [FX-2231](https://algolia.atlassian.net/browse/FX-2231)

**Result**

Added warnings in each connector that adds facets to search parameters if the same attribute is used for a different kind of facet, in which case the search parameters will not be mutated.
If we have a `Menu` and `RefinementList` using the same attribute, currently none of them would be displayed as they would have no facets to choose from. With these changes, only the first one will work and a warning for the second one will be displayed.

I guess this could have been handled directly by the helper, but doing it in InstantSearch makes it so that we can tell which connector has this problem, allowing users to debug more easily.


packages/instantsearch.js/src/connectors/toggle-refinement/connectToggleRefinement.ts
```diff
@@ -425,6 +426,16 @@ const connectToggleRefinement: ToggleRefinementConnector =
         },
 
         getWidgetSearchParameters(searchParameters, { uiState }) {
+          if (searchParameters.isHierarchicalFacet(attribute)) {
```
Haroenv, line 429: conjunctive probably too?

packages/instantsearch.js/src/connectors/range/connectRange.ts
```diff
@@ -417,6 +418,16 @@ const connectRange: RangeConnector = function connectRange(
       },
 
       getWidgetSearchParameters(searchParameters, { uiState }) {
+        if (searchParameters.isHierarchicalFacet(attribute)) {
```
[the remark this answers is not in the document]
aymeric-giraudet: Me neither but better be safe than sorry maybe ? Although it does take some space in certain bundles :/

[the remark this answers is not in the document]
Haroenv: I think range we can avoid the warning, as it's really numerical, so you wouldn't combine it with anything

packages/instantsearch.js/src/connectors/toggle-refinement/connectToggleRefinement.ts
```diff
@@ -425,6 +426,19 @@ const connectToggleRefinement: ToggleRefinementConnector =
         },
 
         getWidgetSearchParameters(searchParameters, { uiState }) {
+          if (
+            searchParameters.isHierarchicalFacet(attribute) ||
+            searchParameters.isConjunctiveFacet(attribute)
+          ) {
+            warning(
+              false,
+              `ToggleRefinement: Attribute "${attribute}" is already used by another widget applying hierarchical or conjunctive faceting.
```
Haroenv, line 435:
```suggestion
              `ToggleRefinement: Attribute "${attribute}" is already used by another widget.
```

maybe clear enough?

Haroenv (approved):
I think it's good like this! possibly we could have a single generic warning that just says "a different widget", but the specificity doesn't hurt

packages/instantsearch.js/src/connectors/toggle-refinement/connectToggleRefinement.ts
```diff
@@ -425,6 +426,19 @@ const connectToggleRefinement: ToggleRefinementConnector =
         },
 
         getWidgetSearchParameters(searchParameters, { uiState }) {
+          if (
+            searchParameters.isHierarchicalFacet(attribute) ||
+            searchParameters.isConjunctiveFacet(attribute)
+          ) {
```
dhayab, line 432: Nitpick, but why do you have a condition as a wrapper instead of putting the negated assertion in the first parameter of `warning()`?

dhayab (approved):
Good for me! I'll just leave a question below.

packages/instantsearch.js/src/connectors/toggle-refinement/connectToggleRefinement.ts
dhayab: 🤦  That explains it, thank you!

PR #5557, opened 2023-03-21
feat(insights): only send automatic events based on condition

Haroenv:
FX-2294

still WIP, as i didn't update any test, and we don't know the exact key, but this should be fine

In this solution, we are assuming it is ok:

- to load search-insights
- to set clickAnalytics: true
- to load the insights middleware automatically
- not remove insights middleware when it "becomes useless"

but it's not ok: 
- to send events without agreement of the developer

This introduces a hypothetical `renderingContent.analytics` boolean set on the dashboard, which we only honor if `insights` is not touched by the developer (aka left to undefined).

### DIFF
--- a/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
@@ -44,6 +44,16 @@ export type InsightsProps<
    * @internal indicator for the default insights middleware
    */
   $$internal?: boolean;
+  /**
+   * If this is true, don't send events, unless the search response contains:
+   *
+   * {
+   *   "renderingContent": {
+   *     "analytics": true
+   *  }
+   * }
+   */
+  verifyEventPermission?: boolean;
 };
 
 const ALGOLIA_INSIGHTS_SRC =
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -238,6 +249,24 @@ export function createInsightsMiddleware<
...
+              helper.lastResults?.renderingContent as
+                | Record<string, unknown>
+                | undefined
+            )?.analytics
+          ) {
+            // @TODO: does this need to be a warning?
+            warning(
+              false,
+              'Cannot send event to Algolia Insights because it is disabled through the dashboard.'
+            );
```
Haroenv, line 265: not sure if a warning is needed

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -44,6 +44,16 @@ export type InsightsProps<
...
+  /**
+   * If this is true, don't send events, unless the search response contains:
+   *
+   * {
+   *   "renderingContent": {
+   *     "analytics": true
+   *  }
+   * }
+   */
+  verifyEventPermission?: boolean;
```
Haroenv, line 56: is this the right name for the option? (it's essentially to detect "insights is not set to true by the customer")

PR #5558, opened 2023-03-21
fix(insights): send extra parameters only for applicable versions

Haroenv:
still to do: update tests

**Summary**

**Result**

### DIFF
--- a/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -177,7 +177,7 @@ describe('insights', () => {
       expect(document.body).toMatchInlineSnapshot(`
         <body>
           <script
-            src="https://cdn.jsdelivr.net/npm/search-insights@2.3.0/dist/search-insights.min.js"
+            src="https://cdn.jsdelivr.net/npm/search-insights@2.4.0/dist/search-insights.min.js"
           />
         </body>
       `);
@@ -220,7 +220,7 @@ describe('insights', () => {
       expect(document.body).toMatchInlineSnapshot(`
         <body>
           <script
-            src="https://cdn.jsdelivr.net/npm/search-insights@2.3.0/dist/search-insights.min.js"
+            src="https://cdn.jsdelivr.net/npm/search-insights@2.4.0/dist/search-insights.min.js"
           />
         </body>
       `);
--- a/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
@@ -9,7 +9,7 @@ import {
 
 import type {
   InsightsClient,
-  InsightsClientMethod,
+  InsightsMethodMap,
   InternalMiddleware,
   Hit,
 } from '../types';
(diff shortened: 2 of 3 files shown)


packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -46,8 +57,13 @@ export type InsightsProps<
...
 };
 
-const ALGOLIA_INSIGHTS_SRC =
-  'https://cdn.jsdelivr.net/npm/search-insights@2.3.0/dist/search-insights.min.js';
+const ALGOLIA_INSIGHTS_VERSION = '2.4.0';
+const ALGOLIA_INSIGHTS_SRC = `https://cdn.jsdelivr.net/npm/search-insights@${ALGOLIA_INSIGHTS_VERSION}/dist/search-insights.min.js`;
+
+export type InsightsClientWithGlobals = InsightsClient & {
+  shouldAddScript?: boolean;
+  version?: string;
```
dhayab, line 65: `version` is available in `InsightsClient` if you update `search-insights` to v2.4.0.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -18,9 +18,20 @@ import type {
...
+  | 'clickedObjectIDsAfterSearch'
+  | 'clickedObjectIDs'
+  | 'clickedFilters'
+  | 'convertedObjectIDsAfterSearch'
+  | 'convertedObjectIDs'
+  | 'convertedFilters'
+  | 'viewedObjectIDs'
+  | 'viewedFilters';
+
+export type InsightsEvent<TMethod extends InsightsMethod = InsightsMethod> = {
```
dhayab, line 32: Maybe those can be moved to **src/types/insights.ts** ?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: Makes sense, I'll leave it as optional for now and update the dependency in a separate PR

packages/instantsearch.js/src/lib/utils/createSendEventForHits.ts
```diff
@@ -138,7 +138,7 @@ export function _buildEventPayloadsForHits({
           widgetType,
           eventType,
           payload: {
-            eventName,
+            eventName: eventName || 'Hit Converted',
```
Haroenv, line 141: alternatively eventName could be required in the type when it's click or convert, but not sure if that can easily be done

dhayab: We could maybe have `sendEventForHits` satisfy both a custom payload signature and `type, hit, eventName`? For now though, having those defaults works for me.

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -872,7 +872,7 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
         eventType: 'click',
         payload: {
           hello: 'world',
-        },
+        } as any,
```
dhayab, line 875: Can you walk me through why this needs to be typed as any?

Haroenv: because the payload is now correctly typed, and these aren't correct payloads

packages/instantsearch.js/src/lib/utils/createSendEventForHits.ts
Haroenv: Yes, I prefer defaults too, as typescript types aren't enforced at runtime, meaning there will be more valid events.

dhayab (approved):

PR #5559, opened 2023-03-22
fix(insights): only load script if we write window.aa

Haroenv:
**Summary**

We were attempting to load search-insights, even if window.aa had potentially already been set up by something different.

**Result**

Only when window[window.AlgoliaAnalyticsObject] is set by InstantSearch it is safe to load search insights, otherwise it will not be used.

### DIFF
--- a/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
@@ -89,7 +89,8 @@ export function createInsightsMiddleware<
     });
   }
   // if still no insightsClient was found, we use a noop
-  const insightsClient: InsightsClientWithGlobals = potentialInsightsClient || noop;
+  const insightsClient: InsightsClientWithGlobals =
+    potentialInsightsClient || noop;
 
   return ({ instantSearchInstance }) => {
     // remove existing default insights middleware


packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -228,6 +228,39 @@ describe('insights', () => {
...
+      const anyWindow = window as any;
+      anyWindow.AlgoliaAnalyticsObject = {
+        type: 'not a string',
+      };
+      instantSearchInstance.use(createInsightsMiddleware());
+
+      expect(document.body).toMatchInlineSnapshot(`
+        <body>
+          <script
+            src="https://cdn.jsdelivr.net/npm/search-insights@2.3.0/dist/search-insights.min.js"
```
Haroenv, line 243:
```suggestion
            src="https://cdn.jsdelivr.net/npm/search-insights@2.4.0/dist/search-insights.min.js"
```

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -81,13 +81,16 @@ export function createInsightsMiddleware<
...
 
-        insightsClient = window[pointer];
-        insightsClient.shouldAddScript = true;
+        potentialInsightsClient = window[pointer];
       }
     });
   }
+  // if still no insightsClient was found, we use a noop
+  const insightsClient: InsightsClient & { shouldAddScript?: boolean } =
+    potentialInsightsClient || noop;
```
Haroenv, line 93:
```suggestion
  const insightsClient: InsightsClientWithGlobals = potentialInsightsClient || noop;
```

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -61,17 +61,17 @@ export function createInsightsMiddleware<
...
     safelyRunOnBrowser(({ window }: { window: any }) => {
       const pointer = window.AlgoliaAnalyticsObject || 'aa';
 
       if (typeof pointer === 'string') {
-        insightsClient = window[pointer];
+        potentialInsightsClient = window[pointer];
       }
 
-      if (!insightsClient) {
+      if (!potentialInsightsClient) {
```
Haroenv, line 74: before this change, if window.AlgoliaAnalyticsObject wasn't a string, `insightsClient` still was `noop`, and thus not null

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -81,13 +81,16 @@ export function createInsightsMiddleware<
             window[pointer].queue.push(args);
           };
           window[pointer].version = ALGOLIA_INSIGHTS_VERSION;
+          window[pointer].shouldAddScript = true;
         }
 
-        insightsClient = window[pointer];
-        insightsClient.shouldAddScript = true;
```
Haroenv, line 87: we should only add the script if we wrote window.aa, as otherwise nothing will happen with the script

aymeric-giraudet (approved):

PR #5560, opened 2023-03-22
feat(insights): only automatically add middleware based on condition

Haroenv:
similar to #5557, but for the whole middleware.

FX-2292

(this obviously didn't update the tests, as the effect is widespread, mainly no userToken for the first search)

Another thing to figure out is that `clickAnalytics` isn't set to true, so that first request doesn't have a queryID. Maybe that should be fixed backend (to also send one if not asked)?

### DIFF
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -228,13 +228,20 @@ Use \`InstantSearch.status === "stalled"\` instead.`
       numberLocale,
       initialUiState = {} as TUiState,
       routing = null,
-      insights = true,
+      insights,
       searchFunction,
       stalledSearchDelay = 200,
       searchClient = null,
       insightsClient = null,
       onStateChange = null,
     } = options;
+
+    if (insights === undefined) {
+      console.log(
+        "insights argument isn't passed to instantsearch, so we will attempt to load the middleware if the results request it."
+      );
+      (this as any)._shouldVerifyInsights = true;
+    }
 
     if (indexName === null) {
       throw new Error(withUsage('The `indexName` option is required.'));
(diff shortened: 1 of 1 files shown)


packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -621,6 +628,20 @@ See ${createDocumentationLink({
...
 
+    // Search results are only notified to the derived helper of the main index
+    // because the main helper is not used to search. This listener is used to
+    // decide whether we start insights middleware automatically.
+    if (
+      (this as any)._shouldVerifyInsights &&
+      this.middleware.every((m) => m.instance.$$type !== 'ais.insights')
+    ) {
+      this.mainHelper.derivedHelpers[0].once('result', ({ results }) => {
+        if (results.renderingContent?.analytics) {
```
Haroenv, line 639: this condition needs to be checked

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -228,14 +228,21 @@ Use \`InstantSearch.status === "stalled"\` instead.`
...
       searchClient = null,
       insightsClient = null,
       onStateChange = null,
     } = options;
 
+    if (insights === undefined) {
+      console.log(
+        "insights argument isn't passed to instantsearch, so we will attempt to load the middleware if the results request it."
+      );
+      (this as any)._shouldVerifyInsights = true;
```
Haroenv, line 243: if so, we can see if clickAnalytics gets forced to true here anyway

packages/instantsearch.js/src/lib/InstantSearch.ts
Haroenv, line 639: and clickAnalytics to be disabled if this is false

packages/instantsearch.js/src/lib/InstantSearch.ts
Haroenv: this logic can also be moved in `start` as helper is available there.

Haroenv (commented):
comments to explain how we can have clickAnalytics true for first query anyway

PR #5563, opened 2023-03-23
feat(useInstantSearchApi): warn when an unstable search client is detected

sarahdayan:
One of the most common mistakes customers make is to provide a new instance of the search client at each re-render, often by instantiating the client inside their component. While this can sometimes be wanted (e.g., to update an ephemeral API key), this is more frequently an oversight which causes InstantSearch to dispatch more search requests than necessary.

This change adds a dev warning that lets you know of an unstable search client. I'll also update the linked docs to add a note on client stability.

If we're happy to move forward with this change, I'll do it for Vue InstantSearch as well.

### DIFF
--- a/packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
+++ b/packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
@@ -93,6 +93,11 @@ export function useInstantSearchApi<TUiState extends UiState, TRouteState>(
     }
 
     if (prevProps.searchClient !== props.searchClient) {
+      warn(
+        false,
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
```diff
@@ -93,6 +93,11 @@ export function useInstantSearchApi<TUiState extends UiState, TRouteState>(
     }
 
     if (prevProps.searchClient !== props.searchClient) {
+      warn(
+        false,
```
Haroenv, line 97: there should be some way to pass `yesIKnowMySearchClientIsntStable` to InstantSearch to disable the warning IMO

Haroenv (approved):
ok, as it's dev

PR #5572, opened 2023-04-01
feat(ais-instant-search): warn when an unstable search client is detected

sarahdayan:
This ports the change made in https://github.com/algolia/instantsearch/pull/5563 in Vue InstantSearch.

I'm still unsure this is necessary in Vue InstantSearch, as Vue keeps track of object references between updates. Happy to drop it if we think it doesn't apply here.


Haroenv (approved):
in most cases people will already have stable references, but this will be useful if it's written inside the template

PR #5578, opened 2023-04-04
feat(insights): require optin for insights option

Haroenv:
**Summary**

The `insights` option now defaults to `false`, instead of true.

**Result**

Insights is no longer fully automatic.

FX-2300


FabienMotte (approved):

FabienMotte:
Looks like some snapshots are not updated

sarahdayan (approved):
LGTM once tests are fixed.

Haroenv:
Ah, I had the wrong local build of InstantSearch.js, so the Vue InstantSearch and React InstantSearch Hooks tests still had the old values. Updating the tests, then merging

PR #5579, opened 2023-04-05
POC: serverless server side rendering for InstantSearch.js

Haroenv:
**Summary**

In the context of Hack Day, we imagined how server side rendering could be set up for InstantSearch.js, without customers running any server side rendering-specific code in their server.

**Result**

A POC with some entry points:

- consumer-server.js which fetches the API and renders html templates in the page
- instantsearch-code.js / instantsearch-code.html, the InstantSearch code set up by the customer (in the dashboard)
- instantsearch-server.js our API, which uses JSDOM to render instantsearch-code.html and evaluate instantsearch-code.js, and finally return the html templates to the consumer

### DIFF
--- a/examples/js/server-rendering/instantsearch-server.js
+++ b/examples/js/server-rendering/instantsearch-server.js
@@ -1,0 +1,113 @@
+import http from 'node:http';
+import { JSDOM, ResourceLoader } from 'jsdom';
+import { readFile } from 'node:fs/promises';
+
+import {
+  getInitialResults,
+  waitForResults,
+} from 'instantsearch.js/es/lib/server.js';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import instantsearch from 'instantsearch.js/dist/instantsearch.development.js';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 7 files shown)


examples/js/server-rendering/instantsearch-server.js
```diff
@@ -0,0 +1,113 @@
...
+
+    window.setTimeout = setTimeout;
+    window.setInterval = setInterval;
+
+    Object.assign(global, window, {
+      window,
+      // @TODO: how to add all global window properties?
+      HTMLElement: window.HTMLElement,
+      XMLHttpRequest: window.XMLHttpRequest,
+    });
```
Haroenv, line 56: in a real use case, we wouldn't add these all in current global scope, but as exposed variables to the isolate which runs the instantsearch code.

examples/js/server-rendering/instantsearch-server.js
```diff
@@ -0,0 +1,113 @@
...
+  .listen(3000);
+
+function createDOM(code) {
+  window.search = undefined;
+
+  // eslint-disable-next-line no-eval
+  eval(code);
+
+  return { search: window.search };
+}
```
Haroenv, line 113: this should also be in an isolated part of the code

PR #5580, opened 2023-04-05
feat(insights): annotate events with algoliaSource

Haroenv:
**Summary**

We annotate the events sent to insights with both "intstantsearch" and "instantsearch-internal", to identify the kind of events sent.

**Result**

- algoliaSource is an array
- default it always has "instantsearch"
- if it's an "internal" event, it also has "instantsearch-internal"

To be able to make this work, all internal events were changed to "xxx:internal", meaning the "skip click if two internal ones in a row" feature in createSendEventForHits was refined to click only (view for example will be sent multiple times in a tick if >20 hitsPerPage)

FX-2287

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -2,11 +2,11 @@
   "files": [
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "74 kB"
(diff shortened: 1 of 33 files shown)


FabienMotte (approved):
Looks good to me, just need to bump some sizes in the bundlesize config.

PR #5588, opened 2023-04-11
fix(server): error when multiple InstantSearch instances in getServerState

Haroenv:
**Summary**

In server environments, the InstantSearch instance gets passed from InstantSearchSSRProvider to InstantSearch, meaning that all child InstantSearch components have the _same_ instance, which can cause many issues.

As this already is working badly, I don't think erroring would be a breaking change, but possibly people are already relying on the current behaviour?

I considered finding a place to have this error in useInstantSearch (as it's the real root cause), I couldn't find anything that still worked with strict mode only creating one InstantSearch instance for both renders.

**Result**

fixes https://github.com/algolia/instantsearch/issues/5585

An error gets thrown if there are multiple InstantSearch components in a single call to getWidgetState

### DIFF
--- a/packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
+++ b/packages/react-instantsearch-hooks-server/src/__tests__/getServerState.test.tsx
@@ -132,6 +132,29 @@ describe('getServerState', () => {
     );
   });
 
+  test('throws with multiple <InstantSearch> components', async () => {
+    function App({ serverState }: { serverState?: InstantSearchServerState }) {
+      return (
+        <InstantSearchSSRProvider {...serverState}>
+          <InstantSearch
+            searchClient={createSearchClient({})}
+            indexName="index"
+          />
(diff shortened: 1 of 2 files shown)


FabienMotte (approved):
Looks good to me having the error here

aymeric-giraudet (approved):
Alternatively we could [also check here in this condition](https://github.com/algolia/instantsearch/blob/1f81b61cf0e78aeb7022d5e9eca0d49de727dc64/packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts#L43) whether the upper searchRef is already defined in which case we know it's nested so we could just use a local ref.
But as you've mentioned in the issue they should probably just use multiple indices so idk

Haroenv:
@aymeric-giraudet I checked that, but in strict mode the second render (or really any subsequent render) has the ref already defined, so you can't use it to detect the behaviour.

PR #5590, opened 2023-04-12
feat(instantsearch): make root indexName optional

Haroenv:
requires https://github.com/algolia/algoliasearch-helper-js/pull/938

Essentially this allows us to create neat non-inheriting indices like this with only three queries sent:

```jsx
const App = (
  <InstantSearch>
    <Index indexName="1" indexId="1a" />
    <Index indexName="1" indexId="1b" />
    <Index indexName="2" />
  </InstantSearch>
);
```

Combined with https://github.com/algolia/react-instantsearch/pull/3690 again, you could also create

```jsx
// 🚨 this isn't possible yet
const App = (
  <InstantSearch>
    <Index indexName="1">
      {/* other widgets, causing no results */}
      <NoResults>
        <Index indexName="1" parentIndexId={null}>
          {/* display stuff from a separate query,
        without interference of inheritance */}
        </Index>
      </NoResults>
    </Index>
  </InstantSearch>
);
```

**Summary**

Allow for more flexible use cases to avoid inheritance of indices, like https://github.com/algolia/instantsearch/issues/5585#issuecomment-1502860567, without custom code skipping the root index.

This does mean that if you forget the indexName, there's a new warning, and no rendering.

**Result**

- IndexName of InstantSearch is optional
- No query gets done for the root index if indexName isn't given

### DIFF
--- a/packages/instantsearch.js/package.json
+++ b/packages/instantsearch.js/package.json
@@ -33,7 +33,7 @@
     "@types/google.maps": "^3.45.3",
     "@types/hogan.js": "^3.0.0",
     "@types/qs": "^6.5.3",
-    "algoliasearch-helper": "^3.11.3",
+    "algoliasearch-helper": "https://pkg.csb.dev/algolia/algoliasearch-helper-js/commit/6ca09d86/algoliasearch-helper",
     "hogan.js": "^3.0.2",
     "htm": "^3.0.0",
     "preact": "^10.10.0",
--- a/packages/instantsearch.js/package.json
+++ b/packages/instantsearch.js/package.json
@@ -33,7 +33,7 @@
     "@types/google.maps": "^3.45.3",
     "@types/hogan.js": "^3.0.0",
     "@types/qs": "^6.5.3",
-    "algoliasearch-helper": "https://pkg.csb.dev/algolia/algoliasearch-helper-js/commit/6ca09d86/algoliasearch-helper",
+    "algoliasearch-helper": "^3.13.0",
     "hogan.js": "^3.0.2",
     "htm": "^3.0.0",
     "preact": "^10.10.0",
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -17,6 +17,7 @@ import {
   noop,
   warning,
   setIndexHelperState,
+  isIndexWidget,
 } from './utils';
 import version from './version';
 
@@ -62,7 +63,7 @@ export type InstantSearchOptions<
   /**
    * The name of the main index
    */
-  indexName: string;
+  indexName?: string;
 
   /**
    * The search client to plug to InstantSearch.js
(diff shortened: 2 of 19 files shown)


packages/instantsearch.js/package.json
```diff
@@ -33,7 +33,7 @@
     "@types/google.maps": "^3.45.3",
     "@types/hogan.js": "^3.0.0",
     "@types/qs": "^6.5.3",
-    "algoliasearch-helper": "^3.11.3",
+    "algoliasearch-helper": "https://pkg.csb.dev/algolia/algoliasearch-helper-js/commit/6ca09d86/algoliasearch-helper",
```
Haroenv, line 36: this and the other references will need to be changed once that version is released

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -62,7 +63,7 @@ export type InstantSearchOptions<
   /**
    * The name of the main index
```
Haroenv, line 64:
```suggestion
   * The name of the main index. If no indexName is provided, you have to manually add an index widget.
```

yarn.lock
```diff
@@ -8521,10 +8521,10 @@ ajv@^8.0.1:
     require-from-string "^2.0.2"
     uri-js "^4.2.2"
 
-algoliasearch-helper@3.11.3, algoliasearch-helper@^3.11.3:
-  version "3.11.3"
-  resolved "https://registry.yarnpkg.com/algoliasearch-helper/-/algoliasearch-helper-3.11.3.tgz#6e7af8afe6f9a9e55186abffb7b6cf7ca8de3301"
-  integrity sha512-TbaEvLwiuGygHQIB8y+OsJKQQ40+JKUua5B91X66tMUHyyhbNHvqyr0lqd3wCoyKx7WybyQrC0WJvzoIeh24Aw==
+"algoliasearch-helper@https://pkg.csb.dev/algolia/algoliasearch-helper-js/commit/6ca09d86/algoliasearch-helper":
```
Haroenv, line 8524: should be changed to the released version

packages/instantsearch.js/src/lib/InstantSearch.ts
dhayab, line 1: > This does mean that if you forget the indexName, there's no warning, no error, and no rendering. Maybe there should be a warning or error if there are no child widgets that will be rendered (not sure if it's possible though)?

If the goal is to prevent firing requests to Algolia with no index at all, would it be possible to have a `noIndexNameWarning` flag set to `true` initially that gets changed to `false`:
- if the index name is set in InstantSearch parameters
- when addWidgets is called with a child index

This might be a bit overengineered though. Wdyt?

Haroenv: I forgot to update the description, there is a warning now.

FabienMotte (approved):

dhayab (approved):

PR #5594, opened 2023-04-13
feat(widgets): introduce parentIndexId

Haroenv:
**Summary**

This allows you to pop out of the index hierarchy, to create more advanced nesting, for example for "no results" pages which query again, or for query suggestions.

This is not needed for vanilla InstantSearch, as the widget isn't in charge of being added, addWidgets is called by the user.

Now the parentIndexId logic is added to useWidget/widgetMixin, meaning it's valid for all widgets.

DynamicWidgets needs to expose index context/provider again to pretend to the child widgets they are part of the parentIndex as requested.

**Result**

parentIndexId as a prop to the React InstantSearch Hooks Index widget.

This enables the following UI/index combination

```jsx
const App = () => (
  <InstantSearch>
    <Index indexName="index1">
      <Index indexName="index2" parentIndexId={null}>
        <Index indexName="index3" />
      </Index>
      <Index indexName="index4" />
    </Index>
  </InstantSearch>
);

// output:
const mainIndex = {
  id: '',
  widgets: [
    {
      id: 'index1',
      widgets: [{ id: 'index4' }],
    },
    {
      id: 'index2',
      widgets: [{ id: 'index3' }],
    },
  ],
};
```

### DIFF
--- a/packages/react-instantsearch-hooks/src/hooks/__tests__/useConnector.test.tsx
+++ b/packages/react-instantsearch-hooks/src/hooks/__tests__/useConnector.test.tsx
@@ -9,6 +9,8 @@ import {
 import {
   createInstantSearchTestWrapper,
   createInstantSearchSpy,
+  castToJestMock,
+  widgetSnapshotSerializer,
 } from '@instantsearch/testutils';
 import { render, waitFor } from '@testing-library/react';
 import { renderHook } from '@testing-library/react-hooks';
@@ -29,6 +31,8 @@ import type {
   HitsConnectorParams,
   HitsWidgetDescription,
 } from 'instantsearch.js/es/connectors/hits/connectHits';
+
+expect.addSnapshotSerializer(widgetSnapshotSerializer);
 
 type CustomSearchBoxWidgetDescription = {
   $$type: 'test.searchBox';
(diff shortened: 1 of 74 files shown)


packages/vue-instantsearch/src/components/DynamicWidgets.js
```diff
@@ -65,6 +65,13 @@ export default {
       type: Number,
       default: undefined,
     },
+    // TODO: this only adds the dynamic widgets to the parent index, not the
+    // widgets that it renders. The parentIndexId should be forwarded.
+    parentIndexId: {
+      type: String,
+      required: false,
+      default: undefined,
+    },
```
Haroenv, line 74: this should be figured out if we want to have parentIndexId on every widget

packages/vue-instantsearch/src/mixins/widget.js
```diff
@@ -65,23 +63,50 @@ Read more on using connectors: https://alg.li/vue-custom`
...
-        this.getParentIndex().addWidgets([this.widget]);
+        this.parentIndex.addWidgets([this.widget]);
       },
       deep: true,
     },
   },
+  computed: {
+    parentIndex() {
+      const mainIndex = this.instantSearchInstance.mainIndex;
+      const parentIndexId = this.parentIndexId;
```
Haroenv, line 86: to debate whether the prop is called parentIndexId or parentIndex?

packages/vue-instantsearch/src/components/DynamicWidgets.js
Haroenv: maybe we could write to the vnode before rendering so it has the extra parentIndexId prop, but that seems weird.

packages/vue-instantsearch/src/components/DynamicWidgets.js
```diff
@@ -65,6 +65,18 @@ export default {
...
+  },
+  provide() {
+    return {
+      // To ensure child widgets are registered in the correct parent.
+      // DynamicWidgets can only have widget children, so this doesn't get propagated further.
+      $_ais_getParentIndex: () => this.parentIndex,
```
Haroenv, line 78: this still has to be tested

packages/vue-instantsearch/src/mixins/widget.js
```diff
@@ -65,23 +63,50 @@ Read more on using connectors: https://alg.li/vue-custom`
...
+            foundIndex = currentIndex;
+          }
+        });
+
+        if (!foundIndex) {
+          throw new Error(`Couldn't find index ${parentIndexId}`);
+        }
+
+        return foundIndex;
+      }
```
Haroenv, line 105: this has to be tested

packages/react-instantsearch-hooks/src/hooks/__tests__/useConnector.test.tsx
```diff
@@ -659,4 +663,66 @@ describe('useConnector', () => {
...
+          <StrictMode>
+            <InstantSearchSpy searchClient={searchClient} indexName="root">
+              <Index indexName="1">
+                <Index indexName="2" parentIndexId={null}>
+                  <CustomWidget attribute="2-natural" />
+                  <CustomWidget parentIndexId="1" attribute="1-moved" />
+                </Index>
+                <CustomWidget attribute="1-natural" />
+                {/* @TODO: parentIndexId 2 isn't found? */}
+                {/* <CustomWidget parentIndexId="2" attribute="2-moved" /> */}
```
Haroenv, line 683: we get an error that index 2 can't be found. It seems like CustomWidget gets rendered when only its parents are created, not yet the other components. Maybe that's specific StrictMode behaviour, but it would be discounting the feature of parentIndexId with a different value than `null` quite a bit.

Haroenv:
For now this isn't workable, as you can't delay until all sibling indices are mounted, meaning you still need to mount the widgets in order, giving it not many use cases.

PR #5595, opened 2023-04-13
fix(removeWidgets): defer search by a full tick

Haroenv:
In React InstantSearch Hooks, the removeWidget cleanup is deferred by a timeout, as it ensures the widget doesn't get recreated in strict mode needlessly.

However, this meant that each one of the removeWidget calls happened in their own context, as the order was

timeout (remove)
promise (defer)
timeout (remove)
promise (defer)

and thus was never deduplicated.

Still to figure out in this PR whether the scheduleSearch using a timer is the way to go, or if `scheduleSearch` in index itself should be in a setTimeout (did not yet verify whether that would work too).

I'd also like to know if the order in which different timers are fired is consistent enough for this to work always. We may need to reimplement a queue ourselves.

### DIFF
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -640,6 +640,15 @@ See ${createDocumentationLink({
     }
   });
 
+  public scheduleSearch2 = defer(
+    () => {
+      if (this.started) {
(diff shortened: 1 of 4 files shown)


Haroenv:
The impact of this is too large, I instead created #5602 which is scoped to React InstantSearch Hooks only.

PR #5604, opened 2023-04-18
test(DynamicWidgets): clearer assertions for widgets

Haroenv:
When the assertions for objectContaining fail, they give a whole lot of information that isn't actually useful, as we only want to know the type.

I therefore changed it to using the snapshot serialiser we have for widgets, and the diffs are much neater when there's the wrong widgets mounted.


aymeric-giraudet (approved):
Nice !

PR #5608, opened 2023-04-24
refactor(sendEventForHits): remove indirection

Haroenv:
**Summary**

At some point earlier there was more information than "payloads", but that was refactored out, meaning that the payloads indirection can be removed

**Result**

simpler createSendEventForHits function (and no hidden breaking change)


dhayab (approved):
Great 👏

Issue #790, opened 2016-01-22
urlSync bug with base attribute


alexandremeunier:
Hello,

There is a bug happening when `urlSync: true` and the html page has a `base` attribute which is different from the current URL. It will instantly replace the URL with the `base` url, to which is appended the instantsearch.js state as a query string

Demo: http://url-syncbug.surge.sh/test

The base attribute is: 
```html
<base href="http://url-syncbug.surge.sh">
```

The page URL is instantly changed to http://url-syncbug.surge.sh/?q=rake&hPP=6&idx=Version_development&p=0&is_v=1

One solution could be to replace the entire URI path instead of just the query params part

vvo:
Working on this, got a fix (use the full url). The underlying issue is that as soon as you use a base href,  when you push relative things inside pushState or location.href then they are resolved to the base href tag.

Which can either:
- result in a full page load (when using #)
- result in a security issue (when using pushState)

GabrielDelepine:
This is my fix (really really durty)

Be carefull, it's durty.

...

Did I told you it's durty ? Yes ?

```javascript
        instantsearch.widgets.durtyFixPagination = function() {
            return {
                render: function render(options) {

                    setTimeout(function() {
                        var matches = document.querySelectorAll("#pagination a[href]");
                        var domain = (qpConfig2.WEBAPPMOBILE ? qpConfig2.URLS.MOBILE : qpConfig2.URLS.WWW);
                        var urlBegin = domain + '/?';
                        angular.forEach(matches, function(a) {
                            if (urlBegin === a.href.substr(0, urlBegin.length)) {
                                a.href = domain + qpRoute.getCurrentRoute().translatedUrl + a.href.substr(urlBegin.length - 1);
                            }
                        });
                        console.info('pagination fix done');
                    }, 800);

                }
            };
        };
```

vvo:
@GabrielDelepine this should already be fixed in latest version, wonder why did you had to fix it? Are you on latest version?

GabrielDelepine:
Hi @vvo 

The fix is in the master branch. I am waiting for the release ```1.4.3```.

vvo:
@GabrielDelepine The fix was done and is available in 1.4.2, that's why I am wondering. See https://github.com/algolia/instantsearch.js/commit/e58aadcd9aae77165b802b46a7d3de5864569e1e

vvo:
Please open a new issue if you still have problems with urlSync and base tag

PR #5613, opened 2023-04-27
feat(history): warn when `createURL` does not return an absolute URL

aymeric-giraudet:
**Summary**

There are two reasons we want users to create absolute URLs with `createURL` : 

- This issue which is actually mentioned in comments : #790 
- In [shouldWrite](https://github.com/algolia/instantsearch/blob/a8b5c1e5bbd6afac39fce523f7d7c2ec02f51153/packages/instantsearch.js/src/lib/routers/history.ts#L263) we expect it to match `window.location.href`.

At first I was thinking about ignoring what's before the `pathname` in `shouldWrite` but as there are other reasons we want URLs to be absolute, I think adding a warning might be what's best.

**Result**

Added a test to check that it's outputted.

### DIFF
--- a/packages/instantsearch.js/src/lib/routers/history.ts
+++ b/packages/instantsearch.js/src/lib/routers/history.ts
@@ -213,16 +213,18 @@ class BrowserHistory<TRouteState> implements Router<TRouteState> {
       location: this.getLocation(),
     });
 
-    try {
-      // We just want to check if the URL is valid.
-      // eslint-disable-next-line no-new
-      new URL(url);
-    } catch (e) {
-      warning(
-        false,
-        `The URL returned by the \`createURL\` function is invalid.
-Please make sure it returns an absolute URL to avoid issues, e.g: \`https://algolia.com/search?query=iphone\`.`
-      );
+    if (__DEV__) {
+      try {
+        // We just want to check if the URL is valid.
+        // eslint-disable-next-line no-new
+        new URL(url);
+      } catch (e) {
+        warning(
+          false,
+          `The URL returned by the \`createURL\` function is invalid.
+  Please make sure it returns an absolute URL to avoid issues, e.g: \`https://algolia.com/search?query=iphone\`.`
+        );
+      }
     }
 
     return url;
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/lib/routers/history.ts
```diff
@@ -207,11 +207,25 @@ class BrowserHistory<TRouteState> implements Router<TRouteState> {
...
    */
   public createURL(routeState: TRouteState): string {
-    return this._createURL({
+    const url = this._createURL({
       qsModule: qs,
       routeState,
       location: this.getLocation(),
     });
+
+    try {
```
aymeric-giraudet, line 216: I could extract this into a function so that it's only executed in dev, or wrap it in `if(__DEV__)`, not sure what's best :/

Haroenv (approved):

packages/instantsearch.js/src/lib/routers/history.ts
```diff
@@ -207,11 +207,27 @@ class BrowserHistory<TRouteState> implements Router<TRouteState> {
    * See: https://github.com/algolia/instantsearch.js/issues/790
    */
   public createURL(routeState: TRouteState): string {
-    return this._createURL({
+    const url = this._createURL({
```
Haroenv, line 210: I can see the risk (and little perf impact), but personally I'd also be happy with `new URL(this._createURL(), location.href)` (and making sure everything `URL` as well as `string`

aymeric-giraudet: I tried with this : 
```javascript
public createURL(routeState: TRouteState): string {
    const location = this.getLocation();
    return new URL(
      this._createURL({
        qsModule: qs,
        routeState,
        location,
      }),
      location.href
    ).href;
  }
}
```
And it seems to work well, even with SSR provided their `getLocation` does return a URL.

However I just realized that `URL` surprisingly [does not work in IE11](https://caniuse.com/?search=URL) 🤦

Haroenv: We're lucky we're only using it in tests then! I indeed forgot URL is not compatible with IE. In a next major we could do that to simplify things then (use and accept URL everywhere, cast to URL in createURL)

PR #5620, opened 2023-05-04
fix(infinite-hits): do not write cache with incomplete state caused by dynamic widgets

dhayab:
**Summary**

In an InstantSearch implementation with dynamic widgets and a custom infinite hits cache, the latter gets overwritten incorrectly because facets are not initially known (until their relevant widgets are mounted by dynamic widgets).

This PR detects this context and prevents writing an incomplete state in the cache.

FX-1647

### DIFF
--- a/packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
+++ b/packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
@@ -9,6 +9,7 @@ import {
   noop,
   createSendEventForHits,
   createBindEventForHits,
+  walkIndex,
 } from '../../lib/utils';
 
 import type { SendEventForHits, BindEventForHits } from '../../lib/utils';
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -355,10 +356,31 @@ const connectInfiniteHits: InfiniteHitsConnector = function connectInfiniteHits(
...
+                .some(({ $$type }) => $$type === 'ais.dynamicWidgets')
+            ) {
+              hasDynamicWidgets = true;
+            }
+          });
+
+          const hasNoFacets =
+            !results?.disjunctiveFacets?.length &&
+            !results?.facets?.length &&
+            !results?.hierarchicalFacets?.length;
```
aymeric-giraudet, line 377: Since you're in the `else` block of `if(!results)`, this should suffice :
```suggestion
            !results.disjunctiveFacets?.length &&
            !results.facets?.length &&
            !results.hierarchicalFacets?.length;
```

aymeric-giraudet (approved):

Haroenv:
@mrzdevcore do you have a test that shows the situation you are faced? Happy to have a look on how to keep all situations working

mrzdevcore:
Here is a minimal sandbox to test the issue, 

[https://codesandbox.io/s/bold-dewdney-fsbm7y?file=/src/app.js:946-1006](https://codesandbox.io/s/bold-dewdney-fsbm7y?file=/src/app.js:946-1006)

Here you can check that there are 13 results shown by stats but infiniteHits display empty results

PR #5639, opened 2023-05-23
feat(templates): add "insights" flag if supported

Haroenv:
**Summary**

To promote the usage of insights, it should be default true in new implementations. This PR adds that for Create InstantSearch App.

**Result**

If you install Create InstantSearch App with a insights-supporting version of InstantSearch, insights will be set to true in the options


dhayab (approved):
Great idea!

PR #5642, opened 2023-05-23
fix(dependencies): update metalsmith packages

Haroenv:
**Summary**

light run of updating dependencies of Create InstantSearch App 

**Result**

metalsmith and its related packages are updated to the latest available versions.


packages/create-instantsearch-app/package.json
```diff
@@ -41,9 +41,9 @@
     "latest-semver": "2.0.0",
     "load-json-file": "6.2.0",
     "lodash.camelcase": "4.3.0",
-    "metalsmith": "2.3.0",
-    "metalsmith-ignore": "1.0.0",
-    "metalsmith-in-place": "4.4.1",
+    "metalsmith": "2.5.1",
+    "@metalsmith/remove": "1.3.0",
+    "@metalsmith/in-place": "4.6.0",
```
dhayab, line 46: Will those versions match those in the helper?

Haroenv: the helper already used different versions for the site, but mostly the same, yes

aymeric-giraudet (approved):

PR #5644, opened 2023-05-23
fix(js): use html templates

Haroenv:
**Summary**

This avoids throwing a warning for the default example

**Result**

Modern html templates used in the InstantSearch.js template of Create InstantSearch App


packages/create-instantsearch-app/src/templates/InstantSearch.js/src/app.js.hbs
```diff
@@ -18,12 +18,12 @@ search.addWidgets([
     container: '#hits',
     {{#if attributesToDisplay}}
     templates: {
-      item: `
+      item: (hit, { html, components }) => `
```
dhayab, line 21:
```suggestion
      item: (hit, { html, components }) => html`
```

PR #5648, opened 2023-05-25
chore(major): leave major comments on how to improve highlighting

Haroenv:
**Summary**

A problem we currently have (see https://github.com/algolia/instantsearch/issues/5237) is that escapeHTML is one of the slower parts of InstantSearch.

With `html``/jsx rendering without dangerousInnerHtml, we no longer need to escape by default, but for that to work, escaping and not escaping should both leave the "placeholders" in the final output.

Current workaround for the performance impact: pass `escapeHTML={false}` to every Hits/infiniteHits etc. and `<Configure highlightPreTag={TAG_REPLACEMENT.highlightPreTag} highlightPostTag={TAG_REPLACEMENT.highlightPostTag} />`

**Result**

`@MAJOR` comments on how to avoid the need for escapeHTML for hits. Facets can also be done at the same time.

### DIFF
--- a/packages/instantsearch.js/src/connectors/answers/connectAnswers.ts
+++ b/packages/instantsearch.js/src/connectors/answers/connectAnswers.ts
@@ -118,6 +118,7 @@ const connectAnswers: AnswersConnector = function connectAnswers(
       nbHits = 1,
       renderDebounceTime = 100,
       searchDebounceTime = 100,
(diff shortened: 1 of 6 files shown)


aymeric-giraudet (approved):
Nice anticipation !
Wasn't aware of this issue, I'll make sure to mention it for the migration of RIS to RISH :)

PR #5668, opened 2023-06-06
Revert "fix(infinite-hits): do not write cache with incomplete state caused by dynamic widgets"

mrzdevcore:
Reverts algolia/instantsearch#5620

The commit introduce a regression in our case when there is no facets to display, connectInfiniteHits return an empty hits array despite the fact that there are results.

### DIFF
--- a/packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
+++ b/packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
@@ -17,6 +17,7 @@ import {
   createRenderOptions,
 } from '../../../../test/createWidget';
 import instantsearch from '../../../index.es';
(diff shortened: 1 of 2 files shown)


mrzdevcore:
Here is a minimal sandbox to test the issue,

[https://codesandbox.io/s/bold-dewdney-fsbm7y?file=/src/app.js:946-1006](https://codesandbox.io/s/bold-dewdney-fsbm7y?file=/src/app.js:946-1006)

Here you can check that there are 13 results shown by stats but infiniteHits display empty results

Haroenv:
Thanks for your PR, I've found a way that covers both your edge case (the new test) as well as the previous test. Does that work in your application as well @mrzdevcore ?

Haroenv:
So that CI will run, I've made this into a "project-owned" PR, but you'll still get co-authorship :) Thanks for your work!

PR #5669, opened 2023-06-07
fix(infiniteHits): correctly avoid writing to cache when dynamicWidgets has no facets in the results

Haroenv:
**Summary**

In #5620 we added a logic to ensure cache does not get written when we expect dynamic widgets to be mounted in the future. This check was too strict and prevented writing when they're _were_ widgets, but they had no results.

I have a couple questions still that maybe should be solved:
- is this check still too strict, what if no widgets get mounted? shouldn't we also write to the cache then?
- what if there are static refinement widgets mounted outside of dynamicWidgets? Does that break this check?

I believe both of these questions can be solved after this PR

**Result**

fixup upon https://github.com/algolia/instantsearch/pull/5620
replaces https://github.com/algolia/instantsearch/pull/5668
FX-2397

Read the condition for deciding if there's any widgets mounted from the state, not from the results.

### DIFF
--- a/packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
+++ b/packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
@@ -17,6 +17,7 @@ import {
   createRenderOptions,
 } from '../../../../test/createWidget';
 import instantsearch from '../../../index.es';
+import { createInfiniteHitsSessionStorageCache } from '../../../lib/infiniteHitsCache';
 import { TAG_PLACEHOLDER, deserializePayload } from '../../../lib/utils';
 import connectInfiniteHits from '../connectInfiniteHits';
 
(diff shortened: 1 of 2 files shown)


dhayab (approved):
Thanks for looking into this, this works better indeed!

In the end I'm wondering if we could ideally, reliably identify the render sequence caused by dynamic widgets instead of inferring it from the facets.

If not, we could maybe keep the detection pattern we have but only prevent cache writing (L387), and not impacting L386. This could  be enough to take care of the 2 questions you mentionned in the PR description.

Haroenv:
Yes, separating the rendering from writing the cache probably would be better. I'll do it in a follow-up so we can bisect if needed.

Haroenv:
I also think that storing the state of "hey I'm dynamic widgets and will mount widgets soon" could be stored somewhere and used instead of this reverse engineering

mrzdevcore:
It's working fine. Thanks

PR #5673, opened 2023-06-12
fix(createServerRootMixin): forward i18n on  option

mrzdevcore:
Issue during SSR, i18n is not defined inside the cloned component instance

This fix solve the issue by adding $i18n in the options of the cloned component.

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -46,7 +46,7 @@
     },
     {
       "path": "packages/vue-instantsearch/vue3/umd/index.js",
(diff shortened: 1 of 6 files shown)


Haroenv (commented):
could you put this after line#33 in that same file? there are other properties copied in a similar way. 

A test would also be very useful!

Haroenv:
Thanks for your PR, this will be useful :)

mrzdevcore:
@Haroenv I have added for i18n for both vue3 and vue2.

PR #5674, opened 2023-06-13
docs: add guidelines for widget and common tests

dhayab:
**Summary**

This PR adds a section on the contributing guidelines, documenting how to write tests for InstantSearch widgets.

In addition, it adds a common test suite for Breadcrumb's base behavior as an example that contributors can use for their own needs.

FX-2066

> **Note**
> Because this PR has 2 scopes, instead of squashing, after review I can rewrite the history to keep scoped commits and merge both of them into the main branch.


CONTRIBUTING.md
```diff
@@ -165,9 +166,162 @@ This monorepo has as goal to be used for all InstantSearch flavors and tools. To
...
+await waitFor(() => {
+  // Test the elements that did change, making it clear what the consequences of
+  // the interaction are.
+  expect(showMoreButton).toHaveTextContent('Show less');
+  expect(listItems).toHaveLength(20);
+});
+```
+
+> **Note**
+>Sometimes, a second snapshot is clearer and shorter than a series of isolated assertions. Use your judgment to determine what will make the test the tersest and easiest to understand.
```
Haroenv, line 294:
```suggestion
>Sometimes, a second (focused) snapshot is clearer and shorter than a series of isolated assertions. Use your judgment to determine what will make the test the tersest and easiest to understand.
```

not sure if this clarification is needed, but could be useful

tests/common/README.md
```diff
@@ -0,0 +1,63 @@
+# Common test suites
+
+To ensure InstantSearch widgets behave similarly in JavaScript, React and Vue, we needed framework agnostic tests suites to assert behaviors that needs to be consistent across flavors.
```
Haroenv, line 3:
```suggestion
To ensure InstantSearch widgets behave similarly in JavaScript, React and Vue, we have introduced framework agnostic tests suites to assert behaviors that needs to be consistent across flavors.
```

or 

```suggestion
To ensure InstantSearch widgets behave similarly in JavaScript, React and Vue, we need framework agnostic tests suites to assert behaviors that needs to be consistent across flavors.
```

tests/common/widgets/breadcrumb/options.ts
```diff
@@ -0,0 +1,298 @@
...
+  }),
+});
+
+/**
+ * Removes rendered DOM discrepancies between flavors when
+ * they don't affect the behavior of the widget
+ * @param element A DOM element to normalize
+ * @returns A normalized DOM element
+ */
+function normalizeSnapshot(element: Element | null): Element | null {
```
Haroenv, line 43: not sure if it's needed, but this is "more consistent" I think

```suggestion
function normalizeSnapshot<TElement extends Element | null>(
  element: TElement
): TElement {
```

That way the element doesn't become optional if it was already asserted

tests/common/widgets/breadcrumb/options.ts
```diff
@@ -0,0 +1,298 @@
...
+    .replace(/<\/li> <li/g, '</li><li')
+    // Vue renders a useless extra space between last list item element and closing list tag
+    .replace(/<\/li> <\/ul>/g, '</li></ul>');
+
+  return normalizedElement;
+}
+
+export function createOptionsTests(setup: BreadcrumbSetup, act: Act) {
+  describe('options', () => {
+    test('renders with props', async () => {
```
Haroenv, line 60:
```suggestion
    test('renders with required props', async () => {
```

tests/common/widgets/breadcrumb/options.ts
```diff
@@ -0,0 +1,298 @@
...
+        </div>
+    `);
+
+      await act(async () => {
+        userEvent.click(screen.getAllByText('Cameras & Camcorders')[0]);
+
+        await wait(0);
+      });
+
+      // expect(searchClient.search).toHaveBeenCalledTimes(2);
```
Haroenv, line 284: why is this commented? did it not pass?

Haroenv (commented):
really good, and such clear writing!!

dhayab:
> really good, and such clear writing!!

Thanks! All the credit goes to @sarahdayan, I mostly refined the branch she provided me.

tests/common/widgets/breadcrumb/options.ts
dhayab: In the end it caused issues with TypeScript. I opted for a more simplistic approach, as there's no reason for the `element` to be null when we want to normalize for snapshot. See https://github.com/algolia/instantsearch/pull/5674/commits/635991e942c11f87099e50d1fe28ede68094fa83.

tests/common/widgets/breadcrumb/options.ts
```diff
@@ -0,0 +1,294 @@
...
+function normalizeSnapshot(element: Element): Element {
+  const normalizedElement = element.cloneNode() as Element;
+  normalizedElement.innerHTML = element.innerHTML
+    // Vue renders a useless extra space between list item elements
+    .replace(/<\/li> <li/g, '</li><li')
+    // Vue renders a useless extra space between last list item element and closing list tag
+    .replace(/<\/li> <\/ul>/g, '</li></ul>');
+
+  return normalizedElement;
+}
```
sarahdayan, line 52: Could this be abstracted into a custom Jest matcher?

dhayab: Oh! I like this idea, I'll try something!

Haroenv: I like it being explicit in the test, as we don't want to indiscriminately replace the spacing, right? Otherwise I believe we could also edit the html snapshot serializer to remove explicitly authored spaces in ul/li

dhayab: I kept the definition of the "normalize function" in the common test suite itself, but abstracted the rest in an extension of the `toMatchInlineSnapshot` matcher: https://github.com/algolia/instantsearch/pull/5674/commits/bf6e2b582bb5655e3d3ea5da6b73a7cb752dbaa6.

Haroenv (approved):

PR #5677, opened 2023-06-16
feat(algoliaAgent): track Next.js version as Algolia agent

FabienMotte:
**Summary**

This PR tracks the Next.js version by adding it as an Algolia agent when a Next.js app is detected.
To get the Next.js version:
- on the browser: `window.next.version` (e.g. `12.1.6`).
- in SSR: `process.env.NEXT_RUNTIME` (`nodejs` or `edge`).

### DIFF
--- a/packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
+++ b/packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
@@ -278,6 +278,6 @@ function getNextVersion() {
   return (
     (typeof window !== 'undefined' &&
       ((window as any).next?.version as string | undefined)) ||
-    (typeof process !== 'undefined' && process.env?.NEXT_RUNTIME)
+    (typeof process !== 'undefined' ? process.env?.NEXT_RUNTIME : undefined)
   );
 }
--- a/packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
+++ b/packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
@@ -271,12 +271,13 @@ You can ignore this warning if you are using a custom router that suits your nee
 
 /**
  * Gets the version of Next.js if it is available in the `window` object,
- * otherwise it returns the version from the npm package dependencies (in SSR).
+ * otherwise it returns the NEXT_RUNTIME environment variable (in SSR),
+ * which is either `nodejs` or `edge`.
  */
 function getNextVersion() {
   return (
     (typeof window !== 'undefined' &&
       ((window as any).next?.version as string | undefined)) ||
-    process.env.npm_package_dependencies_next
+    process.env?.NEXT_RUNTIME
   );
 }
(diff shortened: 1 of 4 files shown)


Haroenv (changes requested):
`npm_package_dependencies_next` is removed in newer versions of npm unfortunately, I think an incomplete picture probably is worse than no picture

packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
```diff
@@ -262,3 +268,15 @@ You can ignore this warning if you are using a custom router that suits your nee
...
+
+/**
+ * Gets the version of Next.js if it is available in the `window` object,
+ * otherwise it returns the version from the npm package dependencies (in SSR).
+ */
+function getNextVersion() {
+  return (
+    (typeof window !== 'undefined' &&
+      ((window as any).next?.version as string | undefined)) ||
+    process.env.npm_package_dependencies_next
```
Haroenv, line 280: process.env isn't save to access either I think, would need `?.`

FabienMotte:
> 
Could we rely on `process.env.NEXT_RUNTIME` instead (only available from v12.1.6)?

packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
aymeric-giraudet: Is `process` even safe ? I guess it gets transformed by most bundlers but if using UMD this fails :/

packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
```diff
@@ -177,6 +180,7 @@ export function useInstantSearchApi<TUiState extends UiState, TRouteState>(
   const cleanupTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const store = useSyncExternalStore<InstantSearch<TUiState, TRouteState>>(
     useCallback(() => {
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
```
Haroenv, line 183: weird, why was this needed now?

Haroenv (commented):
code and test looks correct, just not sure about the eslint comments

packages/react-instantsearch-hooks/src/lib/useInstantSearchApi.ts
FabienMotte: Yeah I tried to align my TS version on the CI one but I still have the eslint error, I'm not sure why

Haroenv: I guess as it doesn't happen on CI, let's leave it as-is :)

FabienMotte: Thanks!

Haroenv (approved):

PR #5681, opened 2023-06-19
fix(insights): delay usertoken updates if received after init

dhayab:
**Summary**

When the Insights middleware receives a new `userToken`, it immediately applies it to the state. This causes an issue when dynamic widgets are initialized, because a new search will occur at the same time, but the second one won't be caught by the search client cache, due to the userToken now being part of the search request.

~~This PR detects when a rerender is expected, and in consequence delays the application of the userToken until the next tick, so that the search request can be returned from cache.~~

This PR always delays the application of an updated userToken until the next tick, if received after init, so that the search request can be returned from cache.

This doesn't apply to userToken retrieved from cookies, or set as initial insights param, as those are retrieved before init.

FX-2400

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -30,7 +30,7 @@
     },
     {
       "path": "packages/react-instantsearch-hooks-web/dist/umd/ReactInstantSearchHooksDOM.min.js",
-      "maxSize": "53.50 kB"
+      "maxSize": "54 kB"
     },
     {
       "path": "packages/react-instantsearch-dom/dist/umd/ReactInstantSearchDOM.min.js",
--- a/examples/react-hooks/default-theme/package.json
+++ b/examples/react-hooks/default-theme/package.json
@@ -18,6 +18,6 @@
     "@parcel/packager-raw-url": "2.8.0",
     "@parcel/transformer-webmanifest": "2.8.0",
     "parcel": "2.8.0",
-    "typescript": "4.9.4"
+    "typescript": "5.1.3"
   }
 }
--- a/examples/react-hooks/e-commerce/package.json
+++ b/examples/react-hooks/e-commerce/package.json
@@ -21,6 +21,6 @@
     "@parcel/packager-raw-url": "2.8.0",
     "@parcel/transformer-webmanifest": "2.8.0",
     "parcel": "2.8.0",
-    "typescript": "4.9.4"
+    "typescript": "5.1.3"
   }
 }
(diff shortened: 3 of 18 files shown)


packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -203,6 +203,7 @@ class InstantSearch<
    * The error gets cleared when the next valid search response is rendered.
    */
   public error: Error | undefined = undefined;
+  public willRerender: boolean = false;
```
Haroenv, line 206: I'm fine with this being a global variable, but I think this name doesn't sufficiently explain what/when it's used. A couple things on my mind:

- I don't want anyone outside of of InstantSearch depending on this, so it should be prefixed with `_`
- I'm not sure `re-render` is what we want to convey, maybe it's "unstable widgets", "pending render", "render requested" or something more like that?

Haroenv (changes requested):
I'm good with the code, but the name should be rethought, maybe in-person

packages/instantsearch.js/src/lib/InstantSearch.ts
dhayab: Totally makes sense. I'll schedule a quick sync so we can figure out the best way to approach this.

dhayab: Following our sync, I switched the approach to always delay setting the usertoken if it is received after init.

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -1108,4 +1124,79 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
...
+
+    expect(searchClient.search).toHaveBeenCalledTimes(2);
+    expect(
+      searchClient.search.mock.calls.map(([[call]]) => call.params.userToken)
+    ).toMatchInlineSnapshot(`
+      [
+        undefined,
+        undefined,
+      ]
+    `);
```
Haroenv, line 1182: maybe this should be `searchClient.search.mock.calls[0][0].params.userToken).toBeUndefined()` and (`[1]`)? it's somewhat clear, but possibly a bit much logic for a test

Haroenv (approved):
glad this logic works, needs much less changes :)

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
dhayab: This is clearer for me as well. I've updated the test.

aymeric-giraudet (approved):
Nice solution !

PR #5682, opened 2023-06-19
test: update guidelines and clean up tests

sarahdayan:
This is a followup of the work we've done with @dhayab in #5674.

- Slightly updates the README in the common test suite
- Removes leftover tests for breadcrumb we no longer needed
- Puts more tests in common for breadcrumb

We can probably start migrating many tests to the common test suites and remove a lot of other tests. This will make the inconsistencies stand out and we can start better testing our widgets and plan for changes in future majors.


packages/instantsearch.js/src/widgets/breadcrumb/__tests__/breadcrumb-test.ts
```diff
@@ -1,22 +0,0 @@
-/**
- * @jest-environment jsdom
- */
-
-import breadcrumb from '../breadcrumb';
-
-describe('breadcrumb()', () => {
```
Haroenv, line 7: should there be some kind of indication in the breadcrumb widget to why it doesn't have any tests? as for now this would be the only widget without tests. Maybe a markdown that points to the common tests?

packages/vue-instantsearch/src/__tests__/common.test.js
```diff
@@ -78,8 +78,8 @@ createHierarchicalMenuTests(async ({ instantSearchOptions, widgetParams }) => {
 });
 
 createBreadcrumbTests(async ({ instantSearchOptions, widgetParams }) => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const { transformItems, ...hierarchicalWidgetParams } = widgetParams;
```
Haroenv: maybe a clarifying comment can be useful to why we don't support transformItems here?

tests/common/widgets/breadcrumb/options.ts
```diff
@@ -296,5 +291,169 @@ export function createOptionsTests(setup: BreadcrumbSetup, act: Act) {
...
+              </a>
+            </li>
+            <li
+              class="ais-Breadcrumb-item ais-Breadcrumb-item--selected"
+            >
+              <span
+                aria-hidden="true"
+                class="ais-Breadcrumb-separator"
+              >
+                &gt;
```
Haroenv, line 373: maybe an extra assertion explicitly on the `&gt;` could be useful, as the separator in the widget isn't taken in account for the rendering

Haroenv (commented):
🆒

tests/common/widgets/breadcrumb/options.ts
sarahdayan: Good idea, addressed in 0bb5937bb.

packages/vue-instantsearch/src/__tests__/common.test.js
sarahdayan: Addressed in f843104ff.

packages/instantsearch.js/src/widgets/breadcrumb/__tests__/breadcrumb-test.ts
sarahdayan: There is `packages/instantsearch.js/src/widgets/breadcrumb/__tests__/breadcrumb.test.tsx`, I moved the remaining tests to it.

Haroenv (approved):

dhayab (approved):

PR #5692, opened 2023-06-29
fix(router): do not update ui state if there are no widgets

dhayab:
**Summary**

This PR prevents the UI state from being updated when there are no widgets mounted. This can happen intermittently with Next.js when the route is changed to an InstantSearch-enabled page.

See [CR-3772](https://algolia.atlassian.net/browse/CR-3475).

**Result**

When using routing, UI state is not updated on subscribe if widgets are not yet mounted.

[CR-3772]: https://algolia.atlassian.net/browse/CR-3772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/middlewares/__tests__/createRouterMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/__tests__/createRouterMiddleware.ts
@@ -8,6 +8,8 @@ import { wait } from '@instantsearch/testutils/wait';
 import instantsearch from '../../index.es';
 import { searchBox } from '../../widgets';
 
+import type { Router } from '../../index.es';
+
 describe('router', () => {
   it('sets initial ui state after reading URL', async () => {
     const searchClient = createSearchClient();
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/middlewares/__tests__/createRouterMiddleware.ts
```diff
@@ -102,4 +104,41 @@ describe('router', () => {
...
+      indexName: 'my-index',
+      searchClient,
+      routing: {
+        router,
+      },
+    });
+
+    const setUiStateSpy = jest.spyOn(search, 'setUiState');
+
+    search.mainIndex.getHelper = jest.fn(() => search.mainHelper);
```
dhayab, line 136: I couldn't figure out why the main index widget would not return a correct helper through this method in the test suite, so this makes it work, but it's not ideal.

Haroenv: When this line is commented out, the test still passes for me though

Haroenv, line 136:
```suggestion
```

dhayab: You're right! I cleaned my repo and tested again, it works. I don't know what might have had caused this earlier. I'll remove the line thanks!

Haroenv (approved):

aymeric-giraudet (approved):
Great !

PR #5696, opened 2023-07-03
fix(url): base createURL on UiState instead of SearchParameters

### LINKED ISSUE #5236: `createURL` always returns root/empty url on first render
## 🐛 Bug description

`createURL` produces incorrect URLs on the server (it's always root URL). 
The same applies to standard components from `react-instantsearch-hooks-web`: `Pagination`, etc.

## 🔍 Bug reproduction

1. Fork [[SSR example with Remix](https://codesandbox.io/s/github/algolia/doc-code-samples/tree/master/react-instantsearch-hooks/remix?file=/app/routes/index.tsx)
2. Edit `app/root.tsx` and comment out `<Scripts />`
3. Refresh the page and inspect the `href`s of pagination links (`.ais-Pagination-link`)

**Live reproduction:**

https://codesandbox.io/s/zen-cache-7cyfjo?file=/app/root.tsx

## 💭 Expected behavior

`createURL` should produce correct URLs when server rendering, without running any code on the client.

## Environment

- React InstantSearch Hooks version: 6.38.1
- React version: 17.0.2
- Browser: Chrome
- OS: MacOS

Haroenv:
**Summary**

This allows the URL to make sense (be filled) before the widget is mounted (like in React InstantSearch Hooks for the initial render)

**Result**

FX-2416: overall task
FX-2438: setup to allow createURL to use UiState, ensuring routing uses UiState on initial mount
FX-2439: refinementList
FX-2440: pagination
FX-2441: breadcrumb
FX-2442: clearRefinements (skipped)
FX-2443: currentRefinements (skipped)
FX-2444: hierarchicalMenu
FX-2445: hitsPerPage
FX-2446: menu
FX-2447: numericMenu
FX-2448: ratingMenu
FX-2449: toggleRefinement
FX-2450: tests

Both clearRefinements and currentRefinements have not been solved as it would require the whole widget to work with uiState instead of helper state (which it currently does). This should be done at a later stage, as it's unlikely to be fixable without translating manually between every possible widget's ui state and search state.

fixes #5236


sarahdayan:
That's a huge PR, is there any way to use stacked PRs instead (e.g., one for overall setup, and one per component/one for all components and their tests)?

Haroenv:
> That's a huge PR, is there any way to use stacked PRs instead (e.g., one for overall setup, and one per component/one for all components and their tests)?

You can review each commit individually :)

packages/instantsearch.js/src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -233,7 +238,42 @@ const connectBreadcrumb: BreadcrumbConnector = function connectBreadcrumb(
         };
       },
 
-      getWidgetSearchParameters(searchParameters) {
+      getWidgetUiState(uiState, { searchParameters }) {
```
Haroenv, line 241: somewhat hidden feature here: breadcrumb is now usable without hierarchicalMenu

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -277,7 +279,14 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
...
+      nextState: SearchParameters | ((state: IndexUiState) => IndexUiState)
+    ) {
+      if (typeof nextState === 'function') {
+        return localInstantSearchInstance!._createURL({
+          [indexId]: nextState(localUiState),
+        });
+      }
       return localInstantSearchInstance!._createURL({
         [indexId]: getLocalWidgetsUiState(localWidgets, {
           searchParameters: nextState,
```
Haroenv, line 292: this is the real meat of the PR. Because you pass a createURL with only uiState, we don't need to calculate the UiState based on the localWidgets (which aren't yet mounted) and can pass the UiState directly to _createURL

packages/instantsearch.js/src/__tests__/common.test.tsx
```diff
@@ -25,11 +43,30 @@ import {
...
+  }>((renderOptions) => {
+    renderOptions.widgetParams.container.innerHTML = `
+        <a
+          data-testid="HierarchicalMenu-link"
+          href="${renderOptions.createURL('value')}"
+        >
+          LINK
+        </a>
+      `;
+  });
```
sarahdayan, line 62: I'm still confused on why we need to create a dedicated component to test this. Can't this be tested directly by checking that the links generated by the built-in widgets work properly?

packages/instantsearch.js/src/connectors/breadcrumb/connectBreadcrumb.ts
sarahdayan: That's pretty cool! Do we actually test that?

packages/instantsearch.js/src/__tests__/common.test.tsx
Haroenv: most widgets don't have links, which is why for consistency all widgets are tested with a custom component.

sarahdayan: Then maybe this shouldn't go in the setup for widgets, and instead we need a separate common test suite for connectors/Hooks.

Haroenv: Could be interesting, although now the routing tests also use the widgets that get mounted alongside it. How would you see it simplified?

sarahdayan: If widgets are creating links with `createURL` then I'd test the links generated by the widget (in the `routing` suite as you're doing now).

If they're not, but their connector expose `createURL`, then I'd add `tests/common/connectors` for connector tests with a `routing` test suite as the one for widgets.

Then in packages (e.g., `instantsearch.js`) I'd execute the `create*` functions twice, once with the setup for widgets, once with the setup for connectors.

sarahdayan (approved):

packages/instantsearch.js/src/__tests__/common.test.tsx
Haroenv: https://algolia.atlassian.net/browse/FX-2468

PR #5746, opened 2023-07-11
test(common): split widget/connector tests

Haroenv:
- remove custom widget in widget tests
- remove usage of the original widget for refinement in "connector"/"custom widget" tests
- split tests into multiple files (so it can be read more easily)

The part hardest to transfer was moving from the refinement widget to a button in a custom widget, i don't think I made any mistakes, but that could be checked the most thoroughly

I was thinking on some kind of meta-test in the future to ensure we implement all tests in all flavours, but i'm not too sure what the way we could enforce that could be (looping over the test creators or something with typescript could be used, but all seem fragile)

FX-2468

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "type-check": "tsc && lerna run type-check",
     "type-check:v3": "tsc --project tsconfig.v3.json",
     "test": "jest && lerna run test",
-    "test:ci": "./script/retry.sh 3 'jest --maxWorkers=4 --ci' && lerna run test --concurrency=1",
+    "test:ci": "./scripts/retry.sh 3 'jest --maxWorkers=4 --ci' && lerna run test --concurrency=1",
     "test:size": "bundlesize",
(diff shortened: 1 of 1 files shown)


packages/vue-instantsearch/src/__tests__/common-connectors.test.js
```diff
@@ -0,0 +1,355 @@
...
+        ),
+      },
+      document.body.appendChild(document.createElement('div'))
+    );
+
+    await nextTick();
+  }
+);
+
+function createCustomWidget({
```
dhayab, line 300: well done!

dhayab (approved):
Wow, really nice restructuring 👏 !

> The part hardest to transfer was moving from the refinement widget to a button in a custom widget, i don't think I made any mistakes, but that could be checked the most thoroughly

I looked at all the places in connector tests, nothint seemed out of place.

PR #5751, opened 2023-07-12
test(common): validate that all tests are implemented

Haroenv:
The common test suite tests are now moved into an object of "setups" and ran separately.

This allows for a couple things that are nice
- you can easily skip running a test or run just a single one by putting a condition in the loop
- TypeScript enforces all tests are implemented
- jest enforces all tests are implemented

### DIFF
--- a/packages/react-instantsearch-hooks-web/src/__tests__/common-connector.test.tsx
+++ b/packages/react-instantsearch-hooks-web/src/__tests__/common-connector.test.tsx
@@ -1,18 +1,7 @@
 /**
  * @jest-environment jsdom
  */
-import {
-  createRefinementListConnectorTests,
-  createHierarchicalMenuConnectorTests,
-  createBreadcrumbConnectorTests,
-  createMenuConnectorTests,
-  createPaginationConnectorTests,
-  createHitsPerPageConnectorTests,
-  createNumericMenuConnectorTests,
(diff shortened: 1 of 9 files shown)


packages/react-instantsearch-hooks-web/src/__tests__/common-connector.test.tsx
```diff
@@ -311,5 +305,25 @@ createToggleRefinementConnectorTests(
...
+        routing: true,
+      },
+    },
+  };
+
+  Object.keys(testSuites).forEach((testName) => {
+    // @ts-ignore (typescript is only referentially typed)
+    // https://github.com/microsoft/TypeScript/issues/38520
+    testSuites[testName](setups[testName], act, testOptions[testName]);
+  });
```
Haroenv, line 328: ideally this would pass typescript, but I don't find a way yet

Haroenv: (note that it's not related to this PR specifically, as the ts-ignore was already here on master

dhayab (approved):
This looks great. Do we need to explain this structure in the README of the common tests package?

Haroenv:
the CTS readme explains only how to write the tests, which hasn't changed, so I think it's ok :)

PR #5756, opened 2023-07-17
feat(react-instantsearch-hooks web): rename `<Stats>` translation

sarahdayan:
The `<Stats>` widget was introduced in https://github.com/algolia/instantsearch/pull/5427 but the translation name isn't consistent with the convention we agreed on in the [Translations RFC](https://github.com/algolia/instantsearch-rfcs/blob/master/implemented/translations-api-rish.md#naming).

This changes the name for something more consistent and conventional.

The change is breaking for anyone already using the `<Stats>` widget.

### DIFF
--- a/packages/react-instantsearch-hooks-web/src/widgets/Stats.tsx
+++ b/packages/react-instantsearch-hooks-web/src/widgets/Stats.tsx
@@ -8,6 +8,7 @@ import type {
   StatsTranslationOptions,
 } from '../ui/Stats';
 import type { UseStatsProps } from 'react-instantsearch-hooks';
+import { warn } from '../../../react-instantsearch-hooks/src/lib/warn';
 
 type UiProps = Pick<
   StatsUiComponentProps,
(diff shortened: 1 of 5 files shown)


aymeric-giraudet:
It could also be added to the specs there : https://github.com/algolia/instantsearch/blob/master/specs/src/pages/widgets/stats.md

Just something like : 
```yml
translations:
  - name: rootElementText
    default: ...
    description: ...
```

dhayab (approved):
Thanks for fixing this!

I'm wondering, as it's breaking for current users of the widget, it might be safer to keep `translations.stats` as a fallback and, if its presence is detected:
- display a deprecation warning on console
- reassign to `translations.rootElementText` 

Wdyt?

Haroenv (changes requested):

Haroenv (approved):

packages/react-instantsearch-hooks-web/src/widgets/Stats.tsx
```diff
@@ -8,6 +8,7 @@ import type {
   StatsTranslationOptions,
 } from '../ui/Stats';
 import type { UseStatsProps } from 'react-instantsearch-hooks';
+import { warn } from '../../../react-instantsearch-hooks/src/lib/warn';
```
Haroenv, line 11:
```suggestion
import { warn } from 'react-instantsearch-hooks/src/lib/warn';
```

this needs to be exported to be usable I believe (otherwise use `/es/` path but that's fragile)

Haroenv (changes requested):

dhayab (approved):

aymeric-giraudet (approved):

packages/react-instantsearch-hooks-web/src/widgets/__tests__/Stats.test.tsx
```diff
@@ -168,6 +168,45 @@ describe('Stats', () => {
     });
   });
 
+  test('renders with deprecated translations (with a deprecation warning)', async () => {
+    const client = createMockedSearchClient();
+    let result: ReturnType<typeof render> | undefined = undefined;
+
+    expect(async () => {
```
Haroenv, line 175:
```suggestion
    await expect(async () => {
```


I think this is what the lint warning is for?

Haroenv (approved):

PR #5762, opened 2023-07-18
fix(version): ensure version file/export matches package

Haroenv:
**Summary**

Same as done for the other flavours, missed during helper migration from the original release script

**Result**

Helper has and keeps the same version number


dhayab (approved):
👏

sarahdayan (approved):

PR #5764, opened 2023-07-19
feat(caching): sort filters and queries

Haroenv:
This allows a better cache hit rate, because when widgets mount and unmount, it's possible that while the filters are the same, they are requested in a different order.

This PR fixes that on two approaches:

- sorting the filters (alphabetically by value per type), which has no impact on the SearchResponse
- sorting the queries (alphabetically by facet name per type), which also needs to be taken in account for the SearchResponse (so it is implemented in SearchParameters)

fixes [CR-3813](https://algolia.atlassian.net/browse/CR-3813)

[CR-3813]: https://algolia.atlassian.net/browse/CR-3813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/algoliasearch-helper/src/requestBuilder.js
+++ b/packages/algoliasearch-helper/src/requestBuilder.js
@@ -273,85 +273,93 @@ var requestBuilder = {
     var facetFilters = [];
 
     var facetsRefinements = state.facetsRefinements || {};
-    Object.keys(facetsRefinements).forEach(function (facetName) {
-      var facetValues = facetsRefinements[facetName] || [];
-      facetValues.forEach(function (facetValue) {
-        facetFilters.push(facetName + ':' + facetValue);
+    Object.keys(facetsRefinements)
+      .sort()
+      .forEach(function (facetName) {
+        var facetValues = facetsRefinements[facetName] || [];
+        facetValues.sort().forEach(function (facetValue) {
+          facetFilters.push(facetName + ':' + facetValue);
+        });
       });
-    });
 
     var facetsExcludes = state.facetsExcludes || {};
-    Object.keys(facetsExcludes).forEach(function (facetName) {
-      var facetValues = facetsExcludes[facetName] || [];
-      facetValues.forEach(function (facetValue) {
-        facetFilters.push(facetName + ':-' + facetValue);
+    Object.keys(facetsExcludes)
+      .sort()
+      .forEach(function (facetName) {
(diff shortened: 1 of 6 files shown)


packages/algoliasearch-helper/test/spec/hierarchical-facets/with-a-disjunctive-facet.js
```diff
@@ -29,7 +29,7 @@ test('hierarchical facets: combined with a disjunctive facet', function () {
 
   helper.setQuery('a').search();
 
-  var disjunctiveFacetsValuesQuery = client.search.mock.calls[0][0][1];
+  var disjunctiveFacetsValuesQuery = client.search.mock.calls[0][0][2];
```
dhayab, line 32: Was that wrong before?

packages/algoliasearch-helper/src/requestBuilder.js
```diff
@@ -4,9 +4,7 @@ var merge = require('./functions/merge');
 
 function sortObject(obj) {
   return Object.keys(obj)
-    .sort(function (a, b) {
-      return a.localeCompare(b);
-    })
```
dhayab, line 9: I'm guessing you removed it because it didn't bring any value as we just need consistency in the sorting and not ensuring numbers are correctly sorted, or was there other concerns?

packages/algoliasearch-helper/test/spec/hierarchical-facets/with-a-disjunctive-facet.js
Haroenv: the order of queries has changed (is now ordered by attribute)

packages/algoliasearch-helper/src/requestBuilder.js
Haroenv: localeCompare isn't about numbers, but about things like æ being sorted at the end of the alphabet in some languages. We don't really need that, and everything behaves better if we use the default sort (it's also much faster)

dhayab: thanks for the clarification!

dhayab (approved):
👏

sarahdayan (approved):

PR #5765, opened 2023-07-19
fix(instantsearch): keep algoliasearch-helper as external dependency during build

dhayab:
**Summary**

Since the integration of the algoliasearch-helper to the monorepo, it is locally linked to other packages that depend on it (most notably, instantsearch.js). This created an issue during the building of the UMD version of the package, where babel doesn't transpile the helper dependency because now it is locally linked thanks to lerna.

This PR adds a babel configuration option that forces the library to be transpiled.

**Result**

InstantSearch.js's UMD package now works correctly and doesn't throw an error at runtime.

[CR-3961](https://algolia.atlassian.net/browse/CR-3961)  
Closes https://github.com/algolia/instantsearch/issues/5763

[CR-3961]: https://algolia.atlassian.net/browse/CR-3961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -10,11 +10,11 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "74.50 kB"
+      "maxSize": "84.25 kB"
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "165 kB"
+      "maxSize": "175.5 kB"
     },
     {
       "path": "packages/react-instantsearch/dist/umd/Core.min.js",
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -10,11 +10,11 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "84.25 kB"
+      "maxSize": "74.50 kB"
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "175.5 kB"
+      "maxSize": "165 kB"
     },
     {
       "path": "packages/react-instantsearch/dist/umd/Core.min.js",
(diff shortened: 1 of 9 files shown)


bundlesize.config.json
dhayab, line 1: I find the 10kb bump to be slightly worrying, wdyt?

dhayab:
I'm working on a follow-up PR that will run an UMD e-commerce example on e2e, so that we can detect issues at runtime with this specific use-case.

bundlesize.config.json
Haroenv: that's a huge difference, what's in the diff?

Haroenv: first notable change I see is _AsyncGenerator, that shouldn't be needed

Haroenv: I'm also seeing "_rollupPluginBabelHelpers" in this version that's not in last released one

packages/instantsearch.js/scripts/rollup/rollup.config.js
```diff
@@ -25,6 +25,7 @@ const plugins = [
     rootMode: 'upward',
     exclude: /node_modules/,
     extensions: ['.js', '.ts', '.tsx'],
+    sourceType: 'unambiguous',
```
Haroenv, line 28: I also wonder if the helper can be considered external? maybe that helps (although I'm not sure)

Haroenv, line 28:
```suggestion
    exclude: /node_modules|algoliasearch-helper/,
    extensions: ['.js', '.ts', '.tsx'],
```

this works for me @dhayab, it leaves commonjs in place for the commonjs plugin, and doesn't add a bunch of babel plugins, essentially still using the helper as an external dependency.


I'm surprised to why the change is only done in InstantSearch.js, probably needed for React InstantSearch Hooks, React InstantSearch Core and Vue InstantSearch too right?

dhayab: Awesome, that seems to work indeed! As for React InstantSearch and Vue, we might have the same issue, it's probably that it's less common to have UMD implementations of those, maybe?

Haroenv (approved):

PR #5769, opened 2023-07-20
tests(e2e): add test for umd e-commerce website

dhayab:
This PR runs an e2e test on a UMD variant of the InstantSearch.js e-commerce example.

Successful run on Saucelabs here: https://app.saucelabs.com/tests/6666f810bd7842428a5c45a9aadb430b (works also for FF and IE).


examples/js/e-commerce-umd/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "example-instantsearch-ecommerce-umd",
+  "version": "1.22.0",
+  "private": true,
+  "scripts": {
+    "start": "BABEL_ENV=parcel parcel index.html --port 3000",
+    "build": "BABEL_ENV=parcel parcel build index.html --public-url .",
+    "website:examples": "BABEL_ENV=parcel parcel build index.html --public-url . --dist-dir=../../../website/examples/js-umd/e-commerce"
```
Haroenv, line 8: technically this means it's also accessible from our main docs, but that's ok, we can document how it works that way

Haroenv (approved):

examples/js/e-commerce-umd/index.html
```diff
@@ -0,0 +1,125 @@
...
+            stroke-linecap="round"
+            stroke-linejoin="round"
+          />
+        </svg>
+
+        Filters
+      </button>
+    </aside>
+
+    <script src="/packages/instantsearch.js/dist/instantsearch.production.min.js"></script>
```
Haroenv, line 121: what is this URL relative to?

dhayab: To the root of the monorepo: https://parceljs.org/features/dependency-resolution/#absolute-specifiers.

PR #5773, opened 2023-07-24
chore(lint): order all imports

Haroenv:
**Summary**

We had these rules on parts of the project already, now everywhere

Only places I noted that seemed interesting to take in account are:
- tests, where order of import may matter
- comments on import (eslint, typescript), that may not have been moved correctly

**Result**

Import order eslint rules are now consistent across the whole monorepo


dhayab (approved):
LGTM 👏

PR #5778, opened 2023-07-25
fix: add a future warning when the package name changes

Haroenv:
**Summary**

A warning on import of the packages to let people know to change packages.

**Result**

A warning on import of the root entry point of React InstantSearch Hooks, React InstantSearch Hooks web, React InstantSearch Hooks server and React InstantSearch Hooks router nextjs with the next name

The warning is dated for 9 August, to give us a day margin to definitely have released, happy to move it a little later as well if we prefer.

Note that if you're using React InstantSearch Hooks web, you'll get the warning for both React InstantSearch Hooks and React InstantSearch Hooks web, I didn't find how to avoid it

FX-2391

### DIFF
--- a/packages/react-instantsearch-hooks-router-nextjs/src/index.ts
+++ b/packages/react-instantsearch-hooks-router-nextjs/src/index.ts
@@ -1,8 +1,14 @@
 import history from 'instantsearch.js/es/lib/routers/history';
+import { warn } from 'react-instantsearch-hooks';
 
 import type { Router, UiState } from 'instantsearch.js';
 import type { BrowserHistoryArgs } from 'instantsearch.js/es/lib/routers/history';
 import type { Router as NextRouter, SingletonRouter } from 'next/router';
+
+warn(
+  Date.now() < new Date('2023-08-09').getTime(),
+  'The package `react-instantsearch-hooks-router-nextjs` is replaced by `react-instantsearch-router-nextjs`. The API is the same, but the package name has changed. Please update your dependencies.'
+);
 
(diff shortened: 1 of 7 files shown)


packages/react-instantsearch-hooks-router-nextjs/src/index.ts
```diff
@@ -1,9 +1,15 @@
 import history from 'instantsearch.js/es/lib/routers/history';
+import { warn } from 'react-instantsearch-hooks';
 
 import type { Router, UiState } from 'instantsearch.js';
 import type { BrowserHistoryArgs } from 'instantsearch.js/es/lib/routers/history';
 import type { Router as NextRouter, SingletonRouter } from 'next/router';
 
+warn(
+  Date.now() < new Date('2023-08-09').getTime(),
+  'The package `react-instantsearch-hooks-router-nextjs` is replaced by `react-instantsearch-router-nextjs`. The API is the same, but the package name has changed. Please update your dependencies.'
```
aymeric-giraudet, line 10: We could also add a link to the upgrade guide which will document codemods and other stuff

Haroenv: https://www.algolia.com/doc/guides/building-search-ui/upgrade-guides/react/ would be the link then, or a specific heading?

aymeric-giraudet: I think this is good enough personally

aymeric-giraudet (approved):

PR #5779, opened 2023-07-25
chore(examples): handle linting issues in e-commerce-umd example

dhayab:
**Summary**

This PR updates some files to handle linting rules that were introduced recently.

Additionally the example was merged in between a release cycle, so it missed on updating its dependency version to instantsearch.js, this is fixed now.

### DIFF
--- a/examples/js/e-commerce-umd/src/routing.ts
+++ b/examples/js/e-commerce-umd/src/routing.ts
@@ -1,11 +1,12 @@
 /* eslint complexity: off */
 
-import type { UiState } from 'instantsearch.js';
 import {
   getFallbackHitsPerPageRoutingValue,
   getFallbackSortByRoutingValue,
   getFallbackRatingsRoutingValue,
 } from './widgets';
+
+import type { UiState } from 'instantsearch.js';
 
 type RouteState = {
   query?: string;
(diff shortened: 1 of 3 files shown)


examples/js/e-commerce-umd/src/routing.ts
```diff
@@ -1,12 +1,13 @@
 /* eslint complexity: off */
 
-import type { UiState } from 'instantsearch.js';
 import {
   getFallbackHitsPerPageRoutingValue,
   getFallbackSortByRoutingValue,
   getFallbackRatingsRoutingValue,
 } from './widgets';
 
+import type { UiState } from 'instantsearch.js';
```
aymeric-giraudet, line 9: Does this come after local imports because it's a type import ?

Haroenv (approved):

aymeric-giraudet (approved):

examples/js/e-commerce-umd/src/routing.ts
dhayab: Yep, here's the order:

```js
groups: [
  'builtin',
  'external',
  'parent',
  'sibling',
  'index',
  'type',
],
```

PR #5780, opened 2023-07-26
feat(react-instantsearch-v7): remove deprecated APIs

sarahdayan:
This removes the deprecated translation for the `<Stats>` widget.

I also removed the `warn` export from `react-instantsearch-hooks` since we don't want users to start using it (unlikely, but better avoid it if we can). If we need it again, let's move these utils in a core package.

https://algolia.atlassian.net/browse/FX-2503


packages/react-instantsearch-hooks-web/src/widgets/__tests__/Stats.test.tsx
```diff
@@ -168,7 +168,7 @@ describe('Stats', () => {
     });
   });
 
-  test('renders with deprecated translations (with a deprecation warning)', async () => {
+  test.skip('renders with deprecated translations (with a deprecation warning)', async () => {
```
Haroenv, line 171: why not removed?

packages/react-instantsearch-hooks/src/index.ts
```diff
@@ -35,4 +35,3 @@ export * from './connectors/useStats';
 export * from './connectors/useToggleRefinement';
 export * from './hooks/useConnector';
 export * from './hooks/useInstantSearch';
-export * from './lib/warn';
```
Haroenv, line 38: other usages of warn need to be removed too

packages/react-instantsearch-hooks-web/src/widgets/__tests__/Stats.test.tsx
sarahdayan: Whoops, accidentally left it in.

packages/react-instantsearch-hooks/src/index.ts
sarahdayan: Ah, I didn't notice it was used by other code sections after I exported it. Reintroducing it so it doesn't break the build, we'll move it later.

Haroenv (dismissed):
use and the warnings on import also need to be removed in v7 :)

packages/react-instantsearch-hooks-web/src/widgets/__tests__/Stats.test.tsx
Haroenv: it should be deleted, no?

dhayab (approved):

dhayab:
We could wait for https://github.com/algolia/instantsearch/pull/5781 to be merged, to integrate the removal of `use`.

Haroenv (approved):
this is good, but use and the import warnings also need to be removed

sarahdayan:
@Haroenv Waiting for #5781 to be merged to remove `use`.

dhayab (approved):

Haroenv (approved):

PR #5789, opened 2023-07-31
fix: use npm registry to retrieve library versions

dhayab:
**Summary**

We currently rely on an Algolia index to retrieve package versions for libraries. This index currently has issues and is not up to date. We've identified that this will cause problem in our CI pipeline when we release the new version of React InstantSearch, as we replace the `react-instantsearch` package in a major, which will require setting version constraint on our CISA CLI.

**Result**

We now directly query the NPM registry to get versions for libraries, and there's an in-memory cache to optimize for multiple queries for the same package in the same CISA CLI session.

### DIFF
--- a/packages/create-instantsearch-app/src/utils/fetchLibraryVersions.js
+++ b/packages/create-instantsearch-app/src/utils/fetchLibraryVersions.js
@@ -1,0 +1,31 @@
+const https = require('https');
+
+function fetchCachedLibraryVersions() {
+  const cache = new Map();
+  return function fetchLibraryVersions(libraryName) {
+    if (cache.has(libraryName)) {
+      return Promise.resolve(cache.get(libraryName));
+    }
+
+    return new Promise((resolve) => {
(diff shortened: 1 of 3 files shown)


packages/create-instantsearch-app/src/utils/fetchLibraryVersions.js
```diff
@@ -0,0 +1,31 @@
...
+
+function fetchCachedLibraryVersions() {
+  const cache = new Map();
+  return function fetchLibraryVersions(libraryName) {
+    if (cache.has(libraryName)) {
+      return Promise.resolve(cache.get(libraryName));
+    }
+
+    return new Promise((resolve) => {
+      https.get(`https://registry.npmjs.org/${libraryName}`, (res) => {
```
Haroenv, line 11: reject the promise on error too?

Haroenv (commented):
👍

Haroenv (approved):

PR #5793, opened 2023-08-01
ci: run router nextjs e2e in ci

dhayab:
**Summary**

This PR adds a CI job that runs e2e tests that target `react-instantsearch-router-nextjs` in Saucelabs.

It is designed to be run on master and on a daily schedule, like `e2e tests`. It is run separately, to benefit from parallelization.

While technically not reliant on the `examples` job, I still left it as a dependency just to keep the e2e jobs together (but it's a subjective choice, i'm open to discussing this).

[**See successful job →**](https://app.circleci.com/pipelines/github/algolia/instantsearch/9178/workflows/8b526154-10cf-46e8-8e37-3abfdfd98ac9)


aymeric-giraudet (approved):

Haroenv (approved):

packages/react-instantsearch-router-nextjs/package.json
```diff
@@ -56,6 +57,7 @@
     "@wdio/cli": "5.16.9",
     "@wdio/jasmine-framework": "5.16.5",
     "@wdio/local-runner": "5.16.9",
+    "@wdio/sauce-service": "5.16.5",
```
Haroenv, line 60: should we use this for the other e2e tests later too?

Haroenv: oh I misread which package.json this was on, I thought it was the root one @dhayab

dhayab: nw, while looking into it I found we have 2 different versions of sauce for no particular reason, so i'll remove the extra version.

dhayab:
Reasked for a review for the latest commit https://github.com/algolia/instantsearch/pull/5793/commits/80ddd123585396010c05b069f431af07cf39b8b4

Haroenv (approved):

PR #5800, opened 2023-08-07
feat: add more info to instantsearch-codemods `package.json`

aymeric-giraudet:
**Summary**

Just added a bit more stuff to the file to make it more in line with the other packages.
Not sure if I should change the version already though

### DIFF
--- a/packages/instantsearch-codemods/package.json
+++ b/packages/instantsearch-codemods/package.json
@@ -1,9 +1,24 @@
 {
   "name": "instantsearch-codemods",
+  "description": "Codemods for Algolia InstantSearch libraries",
   "version": "0.0.0",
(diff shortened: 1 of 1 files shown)


packages/instantsearch-codemods/package.json
```diff
@@ -1,9 +1,24 @@
 {
   "name": "instantsearch-codemods",
+  "description": "Codemods for Algolia InstantSearch libraries",
   "version": "0.0.0",
```
Haroenv, line 4: this version is fine, we'll see what the release process bumps it to and then update if needed, 1.0.0 sounds good, but it's a codemod so the version doesn't matter much

Haroenv (approved):

PR #5802, opened 2023-08-07
fix(codemods): add empty transforms object so that script does not fail

aymeric-giraudet:
**Summary**

I'll probably have to raise an issue to the maintainers of `@codeshift/cli` too 😅


Haroenv (approved):

Haroenv:
It somewhat implies that we should have put only one transform per file (for easier testing too), but I think it's fine like this as well.

PR #5807, opened 2023-08-10
fix(use): compatibility for modern babel

Haroenv:
**Summary**

`use` now properly gets renamed

this is used by codeshift somehow (found by trying it out in the doc-code-samples), not sure why our tests are using an old version of babel (i see `return require('../parser/babel5Compat')(options);` in node_modules)

**Result**

Unfortunately no new test for this: https://github.com/facebook/jscodeshift/issues/510

### DIFF
--- a/packages/instantsearch-codemods/src/rish-to-ris.ts
+++ b/packages/instantsearch-codemods/src/rish-to-ris.ts
@@ -45,11 +45,15 @@ export default function transformer(
           path.value.id.type === 'ObjectPattern' &&
           (path.value.id.properties.find((prop) => {
             return (
-              prop.type === 'Property' &&
+              // Babel 5/6 in JSCodeshift, Babel 7 in codeshift
+              // https://github.com/babel/babel/blob/main/.github/CHANGELOG-v6.md#600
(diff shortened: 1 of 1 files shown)


Haroenv:
I wonder if it has something to do with TypeScript/babel parser maybe?

packages/instantsearch-codemods/src/rish-to-ris.ts
```diff
@@ -45,12 +45,16 @@ export default function transformer(
...
+              // Babel 5/6 in JSCodeshift, Babel 7 in codeshift
+              // https://github.com/babel/babel/blob/main/.github/CHANGELOG-v6.md#600
+              (prop.type === 'Property' || prop.type === 'ObjectProperty') &&
               prop.key.type === 'Identifier' &&
               prop.key.name === 'use'
             );
           }) as ObjectProperty);
 
+        console.log(path.value.id.properties[0].type);
+
```
Haroenv, line 57:
```suggestion
```

dhayab (approved):
Oh that explains the doc-code-samples thing. Thanks for finding out and fixing it!

PR #5818, opened 2023-08-21
test(hits-per-page): add common test suite

sarahdayan:
This reworks the tests for the `hits-per-page` widget to move them in the common test suite, reduce flavor-specific tests, and remove unrealistic tests that manipulate internals.

[FX-2513](https://algolia.atlassian.net/browse/FX-2513)

[FX-2513]: https://algolia.atlassian.net/browse/FX-2513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aymeric-giraudet (approved):

tests/common/widgets/hits-per-page/options.ts
```diff
@@ -0,0 +1,175 @@
...
+      const searchClient = createSearchClient({});
+
+      await setup({
+        instantSearchOptions: {
+          indexName: 'indexName',
+          searchClient,
+        },
+        widgetParams: {
+          items: [
+            { label: '10', value: 10, default: true },
```
Haroenv, line 24: we should also have a test where default isn't the first item

tests/common/widgets/hits-per-page/options.ts
```diff
@@ -0,0 +1,175 @@
...
+            { label: '20', value: 20 },
+            { label: '30', value: 30 },
+          ],
+        },
+      });
+
+      await act(async () => {
+        await wait(0);
+      });
+
```
Haroenv, line 34: I wonder if this should be combined with a hits widget so the length can be checked (although that will mostly rely on the search client mock don't think it's worth it)

packages/react-instantsearch/src/ui/__tests__/HitsPerPage.test.tsx
```diff
@@ -1,149 +0,0 @@
-/**
- * @jest-environment jsdom
- */
-
-import { render } from '@testing-library/react';
```
Haroenv, line 5: I think this test has some value, no? In the future if we reuse the ui component for all flavours, this will be public api and thus should be tested. Maybe this file should be kept

packages/instantsearch.js/src/widgets/hits-per-page/__tests__/hits-per-page.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+          }),
+        ]);
+      }).toThrowErrorMatchingInlineSnapshot(`
+"The \`container\` option is required.
+
+See documentation: https://www.algolia.com/doc/api-reference/widgets/hits-per-page/js/"
+`);
+    });
+
+    test('warns with an item without a `value`', () => {
```
Haroenv, line 34: is this warning only in the InstantSearch.js widget? maybe should be moved to connector

packages/instantsearch.js/src/widgets/hits-per-page/__tests__/hits-per-page.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+ */
+/** @jsx h */
+import { createSearchClient } from '@instantsearch/mocks';
+
+import instantsearch from '../../../index.es';
+import hitsPerPage from '../hits-per-page';
+
+describe('hitsPerPage', () => {
+  describe('options', () => {
+    test('throws without a `container`', () => {
```
Haroenv, line 12: css classes test is gone

Haroenv: (probably could be made common with a mapping from one of the class names to the other two , like done in the other existing "options" test)

Haroenv (commented):
looks good! I think just a little bit too eager with removing tests possibly

packages/instantsearch.js/src/widgets/hits-per-page/__tests__/hits-per-page.test.tsx
sarahdayan: In what other options tests do you see the mapping? There's only breadcrumb for now and I don't see it.

For now I added it in the flavor-specific suite (5d6cbb565).

packages/react-instantsearch/src/ui/__tests__/HitsPerPage.test.tsx
sarahdayan: Technically, we should be the only ones using them. Even if they're exposed from the package, they remain an internal component that shouldn't be used for other purposes, and would already be tested via widget tests. The only thing that could "suffer" is code coverage.

I see the point for facilitating versioning though. Let's maybe then only keep the RIS ones and add a comment on all UI tests to specify we're only keeping them from the perspective of migrating UI components as shared APIs. WDYT?

tests/common/widgets/hits-per-page/options.ts
sarahdayan: Done in 0a4413eee.

packages/instantsearch.js/src/widgets/hits-per-page/__tests__/hits-per-page.test.tsx
sarahdayan: Yes, makes sense, I did it in 9d0ab94ab.

Later on we should migrate all connector tests to connector common test suites. Won't do it here because it would exceed the allocated time for the task, but that should be the next step once we're done with widgets.

packages/react-instantsearch/src/ui/__tests__/HitsPerPage.test.tsx
Haroenv: I think that makes sense. I see the argument for not having ui tests as well as they overlap though. Happy either way

Haroenv (approved):
good! we can do the css classes common later (too much mapping now). I think for now it's probably better to revert the removal of ui component tests (as you didn't remove the ui component test for js either)

sarahdayan:
> (as you didn't remove the ui component test for js either)

That's only because there were no UI tests for this component in InstantSearch.js, but I removed everything UI-wise.

Restored the file in e4c0d2af3.

Haroenv:
There are UI component tests: https://github.com/algolia/instantsearch/tree/master/packages/instantsearch.js/src/components/RefinementList/__tests__

PR #5825, opened 2023-08-25
test(menu): add common test suite

dhayab:
This reworks the tests for the **menu** widget to move them in the common test suite and reduce flavor-specific tests.

FX-2515


tests/utils/normalizeSnapshot.ts
```diff
@@ -13,5 +13,15 @@ export function normalizeSnapshot(html: string) {
...
+      // Vue renders extra whitespace between span elements
+      .replace(/<\/span> <span/g, '</span><span')
+      // Vue renders an empty comment for falsy v-if predicates
+      .replace(/(\s+)?<!---->/g, '') // Vue 2
+      .replace(/(\s+)?<!--v-if-->/g, '') // Vue 3
+      // Vue renders extra whitespace after list elements
+      .replace(/<\/ul> </g, '</ul><')
+      // InstantSearch.js <Template> adds a div wrapper around menu items
+      // @MAJOR: Remove this once <Template> stops defaulting to div root tags
+      .replace(/<div>\s*(<a[\s\S]*?<\/a>)\s*<\/div>/g, '$1')
```
dhayab, line 25: Wondering if I should scope this string replace to only the common menu test runs.

Currently it doesn't affect other tests relying on this normalize function, but it could create issues in the future with other widgets?

packages/vue-instantsearch/src/components/__tests__/Menu.js
```diff
@@ -45,291 +45,6 @@ const defaultProps = {
...
-      },
-    });
-
-    await wrapper.find('.ais-Menu-showMore').trigger('click');
-
-    expect(toggleShowMore).toHaveBeenCalledTimes(1);
-  });
-});
-
 it('exposes send-event method for insights middleware', async () => {
```
Haroenv, line 48: why is this test not in CTS insights (probably already covered)?

tests/utils/normalizeSnapshot.ts
Haroenv: Agreed, this seems to broad, and should be for the menu test alone

Haroenv (commented):
👍

packages/vue-instantsearch/src/components/__tests__/Menu.js
sarahdayan, line 1: We don't need to test the UI component in `vue-instantsearch` since it's an internal API (I've removed it in my own PRs, see https://github.com/algolia/instantsearch/pull/5818/files#diff-2f48a1aa4aa91a005e6af6cc50cd5badac335fda7d45007d94b7f911da389846). We can keep the ones in `react-instantsearch` for when we move the UI components as generic JSX ones in their own packages, and remove the ones we have in `vue-instantsearch` and `instantsearch.js`.

packages/vue-instantsearch/src/components/__tests__/__snapshots__/Menu.js.snap
sarahdayan, line 1: Same comment as the previous one, no need to keep this snapshot.

tests/utils/normalizeSnapshot.ts
sarahdayan: You can scope it to `<Menu>` with the following regex:

```txt
/<div>\s*(<a class="ais-Menu-link".+?><\/a>)\s*<\/div>/g
```

sarahdayan (changes requested):

tests/utils/normalizeSnapshot.ts
dhayab: I did both, I better scoped it to only match `ais-Menu-link` and moved it into the options test suite for the menu widget. See https://github.com/algolia/instantsearch/pull/5825/commits/6e06ebc71dfb52836b445cc874e639c63d2b37e7.

tests/common/widgets/menu/options.ts
```diff
@@ -0,0 +1,1009 @@
...
+  describe('options', () => {
+    const searchClient = createAlgoliaSearchClient({
+      search: jest.fn((requests) => {
+        return Promise.resolve(
+          createMultiSearchResponse(
+            ...requests.map(() =>
+              createSingleSearchResponse({
+                facets: {
+                  brand: {
+                    'Insignia™': 746,
```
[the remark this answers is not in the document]
dhayab: Done in https://github.com/algolia/instantsearch/pull/5825/commits/6e06ebc71dfb52836b445cc874e639c63d2b37e7.

Haroenv (approved):

packages/vue-instantsearch/src/components/__tests__/Menu.js
dhayab: I think for the scope of this PR, leaving the insights tests only might not warrant the introduction of another testing library, but totally agree on the idea of having tests that are designed to follow more closely what users are expected to do.

sarahdayan (approved):

PR #5827, opened 2023-08-30
feat(server): add experimental support for React Server Components and Next.js app directory

aymeric-giraudet:
Took a lot of inspiration from https://github.com/apollographql/apollo-client-nextjs

However I found that if we call `use` only server-side without a `<Suspense>` boundary, we can completely render the page on the server without needing JS to swap the `<template>` tags like it does in their Client Components examples. We'll still have to see with Meta DevRel if it's okay to do 😅

The way I see it, it could be packaged as: 
- Either a stable or beta version of React InstantSearch which supports RSC, basically just the code calling `use` in `useWidget` and probably a new context.
- A package for Next.js which would expose a component based on `<InstantSearch>`, which would also take care of injecting the initial results by directly importing the context from `next/navigation`, and retrieving them when client-side.

Users would simply have to use something like `<NextInstantSearch>` instead of `<InstantSearch>`, with the exact same options.


examples/react/next-app-router/app/page.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <Highlight hit={hit} attribute="name" className="Hit-label" />
+      <span className="Hit-price">${hit.price}</span>
+    </>
+  );
+}
+
+export default function SearchPage() {
+  return (
+    <InstantSearchWrapper ServerInsertedHTMLContext={ServerInsertedHTMLContext}>
+      <InstantSearch searchClient={client} indexName="instant_search">
```
aymeric-giraudet, line 37: Might simply be `<NextInstantSearch searchClient={client} indexName="instant_search">` if we package it separately.

packages/react-instantsearch-core/src/lib/useInstantSearchApi.tsx
```diff
@@ -50,15 +55,31 @@ export type InternalInstantSearch<
...
-  const initialResults = serverState?.initialResults;
+  const { promiseRef, insertHTML } = useRSCContext();
+  let initialResults =
+    serverState?.initialResults ||
+    (typeof window !== 'undefined'
+      ? window[InstantSearchInitialResults]
+      : undefined);
   const prevPropsRef = useRef(props);
 
+  if (Array.isArray(initialResults)) {
```
aymeric-giraudet, line 79: Not sure why their `insertHTML` function inserts it twice, one in `<head>` and one in `<body>`.
Also from the docs they say it should be in `layout.js` but seems like it works in a `page.js` file ?

examples/react/next-app-router/app/layout.tsx
```diff
@@ -0,0 +1,25 @@
...
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  return (
+    <html lang="en">
+      <head>
+        <meta
+          name="google-site-verification"
+          content="AMB0fkgxwBoeRwRG93xQ7IUkmteaFHss6Jz9BBvq7EU"
```
Haroenv, line 19: probably don't want to keep this?

packages/react-instantsearch-core/src/components/InstantSearch.tsx
```diff
@@ -33,7 +36,19 @@ export function InstantSearch<
...
     </InstantSearchContext.Provider>
   );
 }
+
+function TriggerSearch() {
+  const instantsearch = useInstantSearchContext();
+  const { promiseRef } = useRSCContext();
+
+  if (promiseRef.current?.status === 'pending') {
+    instantsearch.mainHelper?.searchOnlyWithDerivedHelpers();
```
Haroenv, line 50: this is a promise, could we add it to the chain or does it not make sense?

packages/react-instantsearch-core/src/lib/useInstantSearchApi.tsx
```diff
@@ -120,6 +141,26 @@ export function useInstantSearchApi<TUiState extends UiState, TRouteState>(
       serverContext.notifyServer({ search });
     }
 
+    if (promiseRef?.current === null && typeof window === 'undefined') {
+      promiseRef.current = wrapPromiseWithState(
+        new Promise((resolve) => {
+          search.once('render', () => {
```
Haroenv, line 147: what ensures that at this point _all_ widgets have been mounted and not just the first one?

Haroenv: something on the TriggerSearch "widget" maybe could be used

packages/react-instantsearch-core/src/lib/useWidget.ts
```diff
@@ -83,7 +88,11 @@ export function useWidget<TWidget extends Widget | IndexWidget, TProps>({
     };
   }, [parentIndex, widget, shouldAddWidgetEarly, search, props]);
 
-  if (shouldAddWidgetEarly) {
+  if (shouldAddWidgetEarly && promiseRef.current?.status === 'pending') {
     parentIndex.addWidgets([widget]);
   }
+
+  if (typeof window === 'undefined' && promiseRef.current) {
```
Haroenv, line 95: in the future we'll need to rename promiseRef, as it's not too clear what it's goal is

Haroenv (commented):
wow, this is super promising! left some comments but overall I think this approach may work. Maybe just a second thought of how next-focused this is (with the "insert html context" that maybe could be more generic?)

examples/react/next-app-router/app/page.tsx
aymeric-giraudet: Some of it does, as injecting `initialResults` uses a Next.js-specific API.
Most of the changes will be directly in `react-instantsearch`, with tiny wrappers for frameworks that just inject initial results with their framework's API.

aymeric-giraudet: Or they could indeed just pass their own `insertHTML` function, but I think it's nice to just have to replace `<InstantSearch>` with no prop change

examples/react/next-app-router/app/page.tsx
```diff
@@ -0,0 +1,50 @@
+'use client';
```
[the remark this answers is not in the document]
aymeric-giraudet: Whichever component that has `<InstantSearch>` at its root yes, but they're free to have a server component in `page.js` which also contains their search component

packages/react-instantsearch-core/src/lib/useInstantSearchApi.tsx
aymeric-giraudet: It's appended after `<InstantSearch>`'s children so they should all have been added at this point, might not with portals or something though

examples/react/next-app-router/app/page.tsx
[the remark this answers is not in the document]
Haroenv: Do you get a clear error if it's accidentally a server component? if not we should give some warnign/error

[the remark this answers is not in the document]
aymeric-giraudet: They would as `<InstantSearch>` calls hooks, but probably not the clearest.
We'll have to test a lot of scenarios yes 😅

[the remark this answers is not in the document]
dhayab: The stack trace might not be clear enough, but the message explicitly say "Please add 'use client'" and the stack trace finishes at the `page.tsx` (or whatever page in which they add InstantSearch).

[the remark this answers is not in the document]
aymeric-giraudet: Tried adding `'use client'` in `react-instantsearch/index.ts` but it throws an error that you can't do `export *` in a client boundary.
Anyway I've seen that Apollo and other libs have users place `'use client'` themselves, and I think it's probably better that they explicitly add their boundaries themselves rather than it being hidden in the library which can be confusing.

buraktamturk:
I had a huge success using this branch on almost production ready website.

There is only thing to mention right now on getLocation method passed to the historyRouter is kind of problematic.
Image a website with two pages /categories/first-category and /categories/second-category, both pages use InstantSearch Next SSR and both of the pages use the same index (with different filter query sent to Algolia). Navigation between pages using next/link causes the following errors:

1) Navigation from /categories/first-category?indexName[page]=2 does not reset UI state on /categories/second-category. On the second URL 90% of times page is missing on the URL however the pagination is always at the 2nd. (even when you set key={pathName} on NextInstantSearchSSR).
2) Navigation from any page to /categories/first-category, or from /categories/first-category to /categories/first-category causes the URL of the page to reset to the previous page (i.e. categories/first-category page displays but url reverts to / or second-category displays but url reverts to first-category) on half of the times (on iOS Safari, it is all of the times). Despite the URL on the browser bar is wrong, the correct page renders.

I went for some debugging, I figured out that the issue is the next app router page change and window.location url change might happen out of order - i.e. the new component renders before window.location is actually updated.

So the following code was the key part of the solution. Instead of relying on window.location, relying on the path and query string came from nextjs router hooks on getLocation function solves the issues I mentioned here.

```javascript
function SomeComponent() {
    const pathName = usePathname();
    const searchParams = useSearchParams();

    // storing them as a reference as I have no idea if the reference to the getLocation can update dynamically
    const pathNameRef = useRef();
    pathNameRef.current = pathName;
    const queryStringRef = useRef();
    queryStringRef.current = searchParams.toString();
    
    const routing = useMemo(() => ({
        router: {
            getLocation: () => {
                if (typeof window === 'undefined') {
                    const url = `${
                        headers().get('x-forwarded-proto') || 'http'
                    }://${headers().get('host')}${pathNameRef.current}?${searchParams}`;
                    return new URL(url);
                }

                return new URL(`${window.location.protocol}//${window.location.host}${pathNameRef.current}?${queryStringRef.current}`);
            },
       }
    }), [ pathNameRef, queryStringRef ]);
    
    return <NextInstantSearchSSR
          key={pathName} // initially I tought this pathName here would solve the issues, it won't
          searchClient={searchClient}
          indexName="Products"
          routing={routing}
      >
        {/* Some other components*/ }
    </NextInstantSearchSSR>
}
```

aymeric-giraudet:
Wow thanks a ton for trying the PR and finding a fix for this @buraktamturk ! 
I'll integrate this into the lib and make sure to credit you as co-author :)

packages/react-instantsearch-core/src/lib/__use.ts
```diff
@@ -0,0 +1,32 @@
...
+type Use = <T>(promise: Promise<T>) => T;
+// Prevent webpack from complaining about our feature detection of the
+// use property of the React namespace, which is expected not
+// to exist when using current stable versions, and that's fine.
+const useKey = 'use' as keyof typeof React;
+const realHook = React[useKey] as Use | undefined;
+
+// This is named with two underscores to allow this hook to evade typical rules of
+// hooks (i.e. it can be used conditionally)
+export const __use =
```
Haroenv, line 18: we can't get rid of the fallback without a breaking change, as older react versions (that don't use server side rendering for example) don't have use

Haroenv (changes requested):
some small cleanups to do still :)

examples/react/next/pages/index.tsx
```diff
@@ -88,7 +88,7 @@ export const getServerSideProps: GetServerSideProps<HomePageProps> =
     const protocol = req.headers.referer?.split('://')[0] || 'https';
     const url = `${protocol}://${req.headers.host}${req.url}`;
     const serverState = await getServerState(<HomePage url={url} />, {
-      renderToString,
+      renderToString: renderToStaticNodeStream,
```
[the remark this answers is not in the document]
aymeric-giraudet: Ah I forgot to remove this, it's not needed I was just trying to fix stuff

packages/react-instantsearch-core/src/lib/wrapPromiseWithState.ts
```diff
@@ -0,0 +1,60 @@
+// Taken from https://github.com/apollographql/apollo-client/blob/ba3e7d9fa7d46e4c636148bbf01552833db0ceda/src/utilities/promises/decoration.ts
```
Haroenv, line 1: cc @mattcarrollcode, it would be nice if React exposes this so it doesn't need to be reimplemented :)

packages/react-instantsearch-core/src/lib/__use.ts
aymeric-giraudet: It won't get called unless it's wrapped by `InstantSearchRSCContext` though

packages/react-instantsearch-nextjs/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "test:exports": "node ./__tests__/module/is-es-module.mjs && node ./__tests__/module/is-cjs-module.cjs"
+  },
+  "devDependencies": {
+    "instantsearch.js": "4.56.11",
+    "next": "13.4.19",
+    "react-instantsearch-core": "7.0.3"
+  },
+  "peerDependencies": {
+    "next": ">= 13.4 && < 14",
+    "react-instantsearch": ">= 7.0.3 && < 8"
```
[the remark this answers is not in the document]
aymeric-giraudet: Shouldn't it be `7.1.0` as it's a new feature ?

packages/react-instantsearch-core/src/lib/__use.ts
Haroenv: in that case we probably should directly use React.use (not import as that probably errors in some cases with old react as it's outside of the conditional?)

packages/react-instantsearch-core/src/lib/useInstantSearchApi.ts
```diff
@@ -56,9 +57,13 @@ export function useInstantSearchApi<TUiState extends UiState, TRouteState>(
   const forceUpdate = useForceUpdate();
   const serverContext = useInstantSearchServerContext<TUiState, TRouteState>();
   const serverState = useInstantSearchSSRContext<TUiState, TRouteState>();
+  const waitingForResultsRef = useRSCContext();
   const initialResults = serverState?.initialResults;
   const prevPropsRef = useRef(props);
 
+  const shouldRenderAtOnce =
+    serverContext || initialResults || waitingForResultsRef;
```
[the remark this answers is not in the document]
aymeric-giraudet: I kept `ref` to make it obvious the value is in `current`, I feel like it would be lost if I remove it

packages/react-instantsearch-core/src/lib/__use.ts
aymeric-giraudet: Can't do `React.use` either as it's not in the types, but I kept the `React['use' as keyof typeof React]` and it works well

packages/react-instantsearch-core/src/lib/use.ts
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
+
+type Use = <T>(promise: Promise<T>) => T;
+const useKey = 'use' as keyof typeof React;
+
+// @TODO: Remove this file and import directly from React when available.
+export const use = React[useKey] as Use;
```
Haroenv, line 7: why is this not available yet now?

Haroenv (approved):

packages/react-instantsearch-core/src/lib/use.ts
aymeric-giraudet: Cause `use` is still a canary feature therefore it's not available in typings

packages/react-instantsearch-nextjs/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "types": "./dist/es/index.d.ts",
+    "require": "./dist/cjs/index.js",
+    "default": "./dist/es/index.js"
+  },
+  "sideEffects": false,
+  "license": "MIT",
+  "homepage": "https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/react/",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/algolia/instantsearch.js"
```
dhayab, line 20: nitpick 
```suggestion
    "url": "https://github.com/algolia/instantsearch"
```

packages/react-instantsearch-nextjs/README.md
```diff
@@ -0,0 +1,147 @@
...
+React InstantSearch is [MIT licensed](../../LICENSE).
+
+<!-- Links -->
+
+[contributing-bugreport]: https://github.com/algolia/instantsearch.js/issues/new?template=BUG_REPORT.yml&labels=triage,Library%3A%20React+InstantSearch
+[contributing-featurerequest]: https://github.com/algolia/instantsearch.js/discussions/new?category=ideas&labels=triage,Library%3A%20React+InstantSearch&title=Feature%20request%3A%20
+[contributing-newissue]: https://github.com/algolia/instantsearch.js/issues/new?labels=triage,Library%3A%20React+InstantSearch
+[contributing-label-easy]: https://github.com/algolia/instantsearch.js/issues?q=is%3Aopen+is%3Aissue+label%3A%22Difficulty%3A+Easy%22+label%3A%22Library%3A%20React+InstantSearch%22
+[contributing-label-bug]: https://github.com/algolia/instantsearch.js/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22+label%3A%22Library%3A%20React+InstantSearch%22
+[contributing-label-chore]: https://github.com/algolia/instantsearch.js/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Chore%22+label%3A%22Library%3A%20React+InstantSearch%22
```
dhayab, line 147: Nitpick to align with a recent PR:

```suggestion
Please read [our contribution process](https://github.com/algolia/instantsearch/blob/master/CONTRIBUTING.md) to learn more.

## License

React InstantSearch is [MIT licensed](../../LICENSE).

<!-- Links -->

[contributing-bugreport]: https://github.com/algolia/instantsearch/issues/new?template=BUG_REPORT.yml&labels=triage,Library%3A%20React+InstantSearch
[contributing-featurerequest]: https://github.com/algolia/instantsearch/discussions/new?category=ideas&labels=triage,Library%3A%20React+InstantSearch&title=Feature%20request%3A%20
[contributing-newissue]: https://github.com/algolia/instantsearch/issues/new?labels=triage,Library%3A%20React+InstantSearch
[contributing-label-easy]: https://github.com/algolia/instantsearch/issues?q=is%3Aopen+is%3Aissue+label%3A%22Difficulty%3A+Easy%22+label%3A%22Library%3A%20React+InstantSearch%22
[contributing-label-bug]: https://github.com/algolia/instantsearch/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Bug%22+label%3A%22Library%3A%20React+InstantSearch%22
[contributing-label-chore]: https://github.com/algolia/instantsearch/issues?q=is%3Aissue+is%3Aopen+label%3A%22Type%3A+Chore%22+label%3A%22Library%3A%20React+InstantSearch%22
```

dhayab (approved):

PR #5830, opened 2023-08-31
feat(`react-instantsearch-ssr-nextjs`): turn Next.js RSC/SSR solution into its own package

sarahdayan:


examples/react/next-app-router/package.json
```diff
@@ -15,7 +15,8 @@
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "react-instantsearch": "7.0.1",
-    "react-instantsearch-router-nextjs": "7.0.1"
+    "react-instantsearch-router-nextjs": "7.0.1",
+    "react-instantsearch-ssr-nextjs": "7.0.1"
```
Haroenv, line 19: do you think we could have instead one package `react-instantsearch-nextjs` that has both?

aymeric-giraudet: The router only works for `pages` while the other only works for `app` though 😅

packages/react-instantsearch-ssr-nextjs/src/NextInstantSearchSSR.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export function NextInstantSearchSSR<
+  TUiState extends UiState = UiState,
+  TRouteState = TUiState
+>({
+  children,
+  ...instantSearchProps
+}: NextInstantSearchSSRProps<TUiState, TRouteState>) {
+  const promiseRef = useRef<PromiseWithState<void> | null>(null);
+  const insertHTML =
```
aymeric-giraudet, line 32: I think we should keep `insertHTML` out of `react-instantsearch-core`, it's Next-specific and other RSC implems would probably use something different for passing data from server to client, so the RSC context should only have a ref to a promise.
What we can do is before calling `searchOnlyWithDerivedHelpers` we do `promiseRef.current.then(/* insertHtml logic */)`, but we might have to make sure that it properly inserts the `<script>` tags before serving the response.

aymeric-giraudet: If it does not properly wait, have a component _before_ `children` that reassigns `promiseRef` to `promiseRef.then(/* insertHTML */)`

sarahdayan: Here it's in the separate `react-instantsearch-ssr-nextjs`, not in core.

aymeric-giraudet: This is in `react-instantsearch-core` : https://github.com/algolia/instantsearch/pull/5830/files#diff-d5ddf6e0b8b366a56a1b63f942409045d3394790e12650e1be7ca8e0d43a6f25
As well as the logic for inserting a `<script>` tag and getting `initialResults`, this should all be in the new package as it's Next.js specific.
`react-instantsearch-core` should only bring RSC support with `use`, hydration however is a Next.js specific concern

aymeric-giraudet: I think we should try to put the most possible in the new package

examples/react/next-app-router/package.json
dhayab: As long as it's experimental I'd keep it as a separate package. But I dig the idea of having a single package for nextjs specific things in the future.

packages/react-instantsearch-ssr-nextjs/src/NextInstantSearchSSR.tsx
dhayab: Because of dependencies in both `useWidget` and `useInstantSearchApi`, we need to colocate `useRSCContext` and `InstantSearchRSCContext` in react-instantsearch-core as well, or else we'd have circular deps issues I think.

aymeric-giraudet: It's fine for the context to be in `react-instantsearch-core`, it has to be there anyway.
I just think `insertHTML` can be done solely in this package though. I'll try after I'm done with my RFC :)

packages/react-instantsearch-ssr-nextjs/src/NextInstantSearchSSR.tsx
```diff
@@ -29,27 +40,93 @@ export function NextInstantSearchSSR<
   ...instantSearchProps
 }: NextInstantSearchSSRProps<TUiState, TRouteState>) {
   const promiseRef = useRef<PromiseWithState<void> | null>(null);
+  const isServerSide = typeof window === 'undefined';
+
+  let initialResults;
+  if (!isServerSide) {
+    initialResults = window[InstantSearchInitialResults]?.pop();
+  }
```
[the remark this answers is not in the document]
aymeric-giraudet: We use the `isServerSide` value for determining whether to render `<InitializePromise>` and `<TriggerSearch>` though :/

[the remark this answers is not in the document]
aymeric-giraudet: I guess I can check if `initialResults` is defined instead tho

PR #5837, opened 2023-09-06
fix(vue-instantsearch): prevent click on navigation items when pressing modifiers

sarahdayan:
This replicates the same behavior as in InstantSearch.js and React InstantSearch—whenever the end user is pressing a modifier key or the middle mouse button, we prevent from refining.

The behavior will be tested in https://github.com/algolia/instantsearch/pull/5836 which will be rebased onto it, so I skipped tests here to avoid doing the same work twice. Let me know if you think differently.


packages/vue-instantsearch/src/components/Pagination.vue
```diff
@@ -135,14 +135,14 @@
             name="last"
             :createURL="() => state.createURL(state.nbPages - 1)"
             :is-last-page="state.isLastPage"
-            :refine="() => refine(state.nbPages - 1)"
+            :refine="() => refine(state.nbPages - 1, $event)"
```
Haroenv, line 138: does this work?

packages/vue-instantsearch/src/components/Pagination.vue
```diff
@@ -107,14 +107,14 @@
...
-            :refine="() => refine(state.currentRefinement + 1)"
+            :refine="() => refine(state.currentRefinement + 1, $event)"
           >
             <template v-if="!state.isLastPage">
               <a
                 :class="suit('link')"
                 aria-label="Next"
                 :href="state.createURL(state.currentRefinement + 1)"
-                @click.prevent="refine(state.currentRefinement + 1)"
+                @click.prevent="refine(state.currentRefinement + 1, $event)"
```
Haroenv, line 117: can we use .exact? https://vuejs.org/guide/essentials/event-handling.html#key-modifiers

sarahdayan: Yes, it's actually much better! I forgot about this Vue magic.

Haroenv (approved):

packages/vue-instantsearch/src/components/Pagination.vue
sarahdayan: That was actually unnecessary.

PR #5838, opened 2023-09-07
 test(current-refinements): add common test suite

dhayab:
This PR moves relevant tests for the Current Refinements widget into a common test suite, and cleans up superfluous flavor-specific tests.

FX-2509


tests/common/widgets/current-refinements/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+type WidgetParams = Parameters<CurrentRefinementsWidget>[0];
+export type CurrentRefinementsWidgetSetup = TestSetup<{
+  widgetParams: Omit<WidgetParams, 'container'>;
+  testParams?: {
+    /**
+     * Sets up an InstantSearch instance wrapped in a form element with the provided `onSubmit` event handler.
+     */
+    formWrapperSubmitHandler?: jest.Mock<any, any>;
+  };
```
dhayab, line 16: For a test I needed to have InstantSearch inside of a `<form>` with mocked submit event handler. This new `testParams` can be defined during `setup()` to send additional information to the flavor-specific implementation of the test suite.

packages/instantsearch.js/src/components/CurrentRefinements/CurrentRefinements.tsx
```diff
@@ -72,6 +72,7 @@ const CurrentRefinements = ({
 
               <button
                 className={cssClasses.delete}
+                type="button"
```
dhayab, line 75: This should be non breaking change, we have a test that ensures we do not submit a parent form when deleting a current refinement.

packages/react-instantsearch/src/ui/CurrentRefinements.tsx
```diff
@@ -97,7 +96,7 @@ export function CurrentRefinements({
             <span
               className={cx('ais-CurrentRefinements-label', classNames.label)}
             >
-              {item.label}:
+              {capitalize(item.label)}:
```
dhayab, line 99: I pondered the idea of doing the reverse, but both IS.js and Vue have capitalization for the label. I don't think this is a breaking change.

packages/vue-instantsearch/src/components/CurrentRefinements.vue
```diff
@@ -33,13 +33,11 @@
...
-                    {{ refinement.label }}
-                  </template>
-                </span>
-                <button
+                  <template v-else>{{ refinement.label }}</template> </span
+                ><button
                   :class="suit('delete')"
-                  @click="item.refine(refinement)"
+                  type="button"
+                  @click.left.exact="item.refine(refinement)"
```
dhayab, line 40: This aligns with other flavours preventing refinement when a modifier key is pressed at the time of clicking on the button. Don't think it's a breaking change.

tests/common/widgets/current-refinements/options.ts
```diff
@@ -0,0 +1,406 @@
...
+import type { CurrentRefinementsWidgetSetup } from '.';
+import type { TestOptions } from '../../common';
+
+function normalizeSnapshot(html: string) {
+  // React InstantSearch adds a noRefinement modifier on the list.
+  // @MAJOR: Remove this once React InstantSearch aligns with spec.
+  return commonNormalizeSnapshot(html).replace(
+    / ais-CurrentRefinements-list--noRefinement/g,
+    ''
+  );
```
dhayab, line 18: I think it would be a breaking change, removing this additional class name from React InstantSearch, so I added it to the snapshot normalization process.

tests/utils/normalizeSnapshot.ts
```diff
@@ -6,7 +6,10 @@ export function normalizeSnapshot(html: string) {
       // Vue renders extra whitespace inside option item elements
       .replace(/>\s+?(\w+?)\s+?<\/option>/gs, '>$1</option>')
       // Vue renders extra whitespace inside button elements
-      .replace(/>\s+(.+?)\s+<\/button>/gs, '>$1</button>')
+      .replace(
+        /(<button[^>]*>)\s+(.+?)\s+(<\/button>)/gs,
+        (_, open, content, close) => `${open}${content.trim()}${close}`
+      )
```
dhayab, line 12: The previous expression was too greedy and would sometime target non-matching button tags, this one only works in the scope of a single button element.

packages/vue-instantsearch/src/components/CurrentRefinements.vue
Haroenv: just to double check before we commit this, does this transpile and work correctly in both vue 2 and vue 3?

packages/vue-instantsearch/src/components/CurrentRefinements.vue
```diff
@@ -17,7 +17,7 @@
             :item="item"
             :createURL="state.createURL"
           >
-            <span :class="suit('label')">{{ capitalize(item.label) }}: </span>
+            <span :class="suit('label')">{{ capitalize(item.label) }}:</span>
```
Haroenv, line 20: will the refinement now not look like `attribute:value` instead of `attribute: value`?

packages/vue-instantsearch/src/__tests__/common-widgets.test.js
```diff
@@ -362,6 +362,45 @@ const testSetups = {
...
+          h(AisRefinementList, { props: { attribute } })
+        ),
+        h(AisCurrentRefinements, { props: widgetParams }),
+        h(GlobalErrorSwallower),
+      ]);
+
+    mountApp(
+      {
+        render: renderCompat((h) =>
+          testParams?.formWrapperSubmitHandler
```
Haroenv, line 388: it would be good if the submit handler was defined for every test in a widget (the others can just have it as a noop)

packages/vue-instantsearch/src/__tests__/common-widgets.test.js
```diff
@@ -362,6 +362,45 @@ const testSetups = {
...
+    testParams,
+  }) {
+    const refinementListAttributes = Object.keys(
+      instantSearchOptions.initialUiState?.indexName.refinementList || {}
+    );
+
+    const search = (h) =>
+      h(AisInstantSearch, { props: instantSearchOptions }, [
+        ...refinementListAttributes.map((attribute) =>
+          h(AisRefinementList, { props: { attribute } })
```
Haroenv, line 379: we're no longer testing the output of current refinements with widgets other than refinement list

packages/react-instantsearch/src/__tests__/common-widgets.test.tsx
```diff
@@ -242,6 +242,32 @@ const testSetups: TestSetupsMap<TestSuites> = {
...
+    widgetParams,
+    testParams,
+  }) {
+    const refinementListAttributes = Object.keys(
+      instantSearchOptions.initialUiState?.indexName?.refinementList || {}
+    );
+    const search = (
+      <InstantSearch {...instantSearchOptions}>
+        {refinementListAttributes.map((attribute) => (
+          <RefinementList key={attribute} attribute={attribute} />
```
Haroenv, line 256: we need to test all widget and refinement types

packages/react-instantsearch/src/ui/CurrentRefinements.tsx
Haroenv: I think this is ok, but would also be ok with leaving the difference and later in a major converging on lowercase

Haroenv (changes requested):

tests/common/widgets/current-refinements/options.ts
Haroenv: could we also add the --noRefinement class in the other flavours instead? or do you think it's useless (it's not very useful with :empty selector)

tests/common/widgets/current-refinements/index.ts
Haroenv: I just tried it out, and you can attach a "submit" event handler to body, but only when the button is in a form:

```js
// works:
document.body.append(document.createElement('form'))
document.querySelector('form').append(document.createElement('button'), { textContent: 'click me' }))
document.body.addEventListener('submit', console.log)

// does not work
document.body.append(document.createElement('button'), { textContent: 'click me' }))
document.body.addEventListener('submit', console.log)
```

This means we can change the setup to always put the CR in a form, and in the test attach a listener to body's submit event

packages/vue-instantsearch/src/components/CurrentRefinements.vue
dhayab: Yes, this mirrors what we have for JS / React, and (in our demos) the spacing is set with the default CSS rules from algolia/satellite.

packages/vue-instantsearch/src/components/CurrentRefinements.vue
dhayab: Tests pass (both `unit tests` and `vue v3`) and I just checked with Vue 3 [in a sandbox](https://codesandbox.io/s/suspicious-goldwasser-mlwms3?file=/src/App.vue) it works as expected.

tests/common/widgets/current-refinements/options.ts
dhayab: I thought about it, but same as you, I really can't see any use for this. I'd rather wait and remove it next major.

packages/vue-instantsearch/src/__tests__/common-widgets.test.js
dhayab: Good call, I'll put a test with the full suite of facets / refinements in the common test suite.

packages/vue-instantsearch/src/components/CurrentRefinements.vue
Haroenv: I think I'd prefer js & react to conform to Vue here, allowing for better copy/paste too for example

packages/vue-instantsearch/src/__tests__/common-widgets.test.js
Haroenv: yes, could be in the shared widget tests

dhayab: Done in https://github.com/algolia/instantsearch/pull/5838/commits/3a1c4a480cf93a1a4d0f44a62a2981f4eefa0d62. Although, I haven't found a way to add `tag` and `exclude` refinement types without access to the helper.

packages/react-instantsearch/src/__tests__/common-widgets.test.tsx
dhayab: I'll close this comment thread and we can continue the conversation at https://github.com/algolia/instantsearch/pull/5838#discussion_r1319655476.

tests/common/widgets/current-refinements/index.ts
dhayab: The form is now created by the options test and the InstantSearch implementations use it as their root container: https://github.com/algolia/instantsearch/pull/5838/commits/c519cbb63c2729b12d1a95f9f18e980ea56b308d.

tests/common/widgets/current-refinements/options.ts
```diff
@@ -0,0 +1,406 @@
...
+
+        await wait(0);
+      });
+
+      expect(
+        document.querySelectorAll('.ais-CurrentRefinements-item')
+      ).toHaveLength(1);
+    });
+
+    it('inclusively restricts what refinements to display', async () => {
```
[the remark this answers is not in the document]
dhayab: Done in https://github.com/algolia/instantsearch/pull/5838/commits/00aff00b60e047507063b8f0b0d0adc0cecb4572

packages/vue-instantsearch/src/components/CurrentRefinements.vue
dhayab: You're right. I've updated this in https://github.com/algolia/instantsearch/pull/5838/commits/4941b50c2a9d34a70a238a50aa2e9b9cd1b4788d.

tests/common/widgets/current-refinements/options.ts
```diff
@@ -24,9 +24,14 @@ export function createOptionsTests(
 ) {
   describe('options', () => {
     const searchClient = createAlgoliaSearchClient({});
+    const onSubmitListener = jest.fn();
 
     beforeEach(() => {
       searchClient.search.mockClear();
+
+      const form = document.createElement('form')!;
```
Haroenv, line 32: sorry, that's not what I meant. The form can be created in the setup of the tests

tests/common/widgets/current-refinements/options.ts
```diff
@@ -24,9 +24,14 @@ export function createOptionsTests(
...
+      form.addEventListener('submit', onSubmitListener);
```
Haroenv, line 33:
```suggestion
      body.addEventListener('submit', onSubmitListener);
```

and like this we won't need to make anything conditional

Haroenv (changes requested):

packages/vue-instantsearch/src/__tests__/common-widgets.test.js
Haroenv: do the tag and exclude refinement tests still exist in the connector? if so that's good enough imo

tests/common/widgets/current-refinements/options.ts
dhayab: I'm afraid this lead to false positive in the test of "not calling the event handler". Also, wouldn't the buttons need to be inside the form for the body onSubmit handler to catch the event? In this case it would still require the root container of the widgets to be set as the form element.

Haroenv: I see what you mean, but it could also be false positive now if in the setup the button isn't put in the form but directly in the body. We can solve both by adding a test before the assertion that the layout is `body > form > button`

Haroenv (approved):

PR #5841, opened 2023-09-08
fix(create-instantsearch-app): fix option label for `attributesForFaceting`

sarahdayan:
The label for `attributesForFaceting` was wrong, it said "Attributes to display" instead of "Attributes for faceting".

### DIFF
--- a/packages/create-instantsearch-app/src/cli/index.js
+++ b/packages/create-instantsearch-app/src/cli/index.js
@@ -54,24 +54,22 @@ program
   .option('--config <config>', 'The configuration file to get the options from')
   .option('--no-installation', 'Ignore dependency installation')
   .option('--no-interactive', 'Ask no interactive questions')
-  .action(dest => {
+  .action((dest) => {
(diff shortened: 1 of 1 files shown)


packages/create-instantsearch-app/src/cli/index.js
```diff
@@ -155,17 +157,17 @@ const getQuestions = ({ appName }) => ({
...
-      filter: (attributes) => attributes.filter(Boolean),
+      filter: attributes => attributes.filter(Boolean),
       when: ({ appId, apiKey, indexName }) =>
         attributesToDisplay.length === 0 && appId && apiKey && indexName,
     },
     {
       type: 'checkbox',
       name: 'attributesForFaceting',
-      message: 'Attributes to display',
+      message: 'Attributes for faceting',
```
sarahdayan, line 167: Prettier generated a larger diff than expected but this is the only change that matters.

dhayab (approved):

Haroenv (approved):

PR #5843, opened 2023-09-12
chore(docs): replace algolia/instantsearch.js with algolia/instantsearch

Haroenv:
**Summary**

leftover from the monorepo migration, also some changes in the readmes to fix links more precise for vue and react

**Result**

- readme files have right links
- all algolia/instantsearch.js links replaced

One thing I couldn't find is where we configure conventional-changelog to add algolia/instantsearch.js URLs, because that still seems to be the case recently?

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -1,7 +1,7 @@
 {
   "sandboxes": [
     "instantsearchjs-es-template-pcw1k",
-    "github/algolia/instantsearch.js/tree/templates/react-instantsearch",
+    "github/algolia/instantsearch/tree/templates/react-instantsearch",
     "/examples/react/default-theme",
     "/examples/vue/default-theme"
   ],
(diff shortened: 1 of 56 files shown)


dhayab:
> One thing I couldn't find is where we configure conventional-changelog to add algolia/instantsearch.js URLs, because that still seems to be the case recently?

This is probably computed from the remote url set when cloning the repository. When preparing releases locally, this can be solved by recloning from `git@github.com:algolia/instantsearch`, but for scheduled releases I'm not sure where to specify this in CircleCI. I'll check it out.

Haroenv:
Ah, you might be right that it's related to the local remote when generating, as I didn't update mine yet, and likely we didn't update the remote in circleci

dhayab:
I noticed in https://github.com/algolia/instantsearch/pull/5844 that the changelog for CISA was using the correct URL, but not the one for Vue InstantSearch. The only difference I see that could matter was that in their respective `package.json` files, CISA has "algolia/instantsearch" in the repository field, whereas Vue has "https://github.com/algolia/instantsearch.js".

So I guess your changes will fix changelog generation next time!

dhayab (approved):

PR #5848, opened 2023-09-13
fix(insights): use init method params type from Search Insights

francoischalifour:
This fixes a typing issue in the Insights middleware (and the `insights` prop of InstantSearch) where the init params type was out of sync with Search Insights. Specifically, the `userToken` param was missing.

In my case, I do need to provide `userToken` at this level because it comes from my app. If I don't provide it, then InstantSearch sends automatic events with a random user token, which is different than the one for my queries.

This is now accepted by the typings:

```tsx
<InstantSearch
  searchClient={searchClient}
  indexName={indexName}
  insights={{
    insightsInitParams: {
      userToken,
    },
  }}
>
  {/* ... */}
</InstantSearch>
```

The `InitParams` type isn't exported by Search Insights, so I inferred the init params from the `InsightsMethodMap` type.

### DIFF
--- a/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
@@ -28,13 +28,7 @@ export type InsightsProps<
   TInsightsClient extends ProvidedInsightsClient = ProvidedInsightsClient
 > = {
   insightsClient?: TInsightsClient;
-  insightsInitParams?: {
-    userHasOptedOut?: boolean;
-    useCookie?: boolean;
-    anonymousUserToken?: boolean;
-    cookieDuration?: number;
-    region?: 'de' | 'us';
-  };
+  insightsInitParams?: Omit<InsightsMethodMap['init'][0], 'appId' | 'apiKey'>;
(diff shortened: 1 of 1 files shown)


packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -28,13 +28,7 @@ export type InsightsProps<
...
 > = {
   insightsClient?: TInsightsClient;
-  insightsInitParams?: {
-    userHasOptedOut?: boolean;
-    useCookie?: boolean;
-    anonymousUserToken?: boolean;
-    cookieDuration?: number;
-    region?: 'de' | 'us';
-  };
+  insightsInitParams?: Omit<InsightsMethodMap['init'][0], 'appId' | 'apiKey'>;
```
Haroenv, line 31:
```suggestion
  insightsInitParams?: Partial<InsightsMethodMap['init'][0]>;
```

it seems to allow override of all params in theory

Haroenv (changes requested):
Thanks! Can be more precise type

Haroenv (approved):

PR #5850, opened 2023-09-13
fix(prettier): consistent version

Haroenv:
- both in Create InstantSearch App and globally
- also updated to avoid another duplicate version
- prettified everything again (except .hbs files because prettier seems to think they're always html)


yarn.lock
```diff
@@ -8554,6 +8554,13 @@ ajv@^8.0.1:
     require-from-string "^2.0.2"
     uri-js "^4.2.2"
 
+algoliasearch-helper@3.14.0:
+  version "3.14.0"
+  resolved "https://registry.npmjs.org/algoliasearch-helper/-/algoliasearch-helper-3.14.0.tgz#2409c2591952719ab6fba1de77b3bbe5094ab85e"
+  integrity sha512-gXDXzsSS0YANn5dHr71CUXOo84cN4azhHKUbg71vAWnH+1JBiR4jf7to3t3JHXknXkbV0F7f055vUSBKrltHLQ==
+  dependencies:
+    "@algolia/events" "^4.0.1"
+
```
dhayab, line 8563: This is caused by our dependency on `react-instantsearch-core@6` which itself requires this precise version of the helper. I think we can probably (in another PR) remove the only test that requires the old library ([/packages/react-instantsearch-core/src/__tests__/compat.test.tsx](https://github.com/algolia/instantsearch/blob/master/packages/react-instantsearch-core/src/__tests__/compat.test.tsx)), because we can basically assert the same with any container which is the correct `<InstantSearch>`, and the message itself is tested in `useInstantSearch.test.ts`.

Haroenv: ah I thought that was on purpose as the last version of React InstantSearch to exist will always have 3.14.0. I don't think it's a big deal though (but we could have an override of the helper version as long as it's the same major still).

dhayab (commented):
Shipjs (with `yarn release -D`) keeps on reverting linting in the modified changelog files. Perhaps we could exclude these from prettier?

dhayab (approved):

PR #5854, opened 2023-09-18
feat(instantsearch): hydrate search client

### LINKED ISSUE #5839: Unwanted client-side query after SSR with Remix/Shopify-Hydrogen
### 🐛 Current behavior

Using the latest Remix (or any version above 1.17.0) and [Algolia's official SSR example for Remix](https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react/#with-remix), if we add a few more routes, navigating between the routes will trigger additional/unwanted client-side query. The page should be SSRed as the `serverState` is already in the loader, so the client-side query is not needed. 

If you compare the `queryId` from the `serverState` with the client-side query, the `queryIDs` are different which will cause problems if `ClickAnalytics` is enabled.

After some research, we found that the issue is impacting all Remix apps that use 1.17.0 and above. [Remix 1.17.0 ](https://github.com/remix-run/remix/releases/tag/remix%401.17.0)uses the updated React-router that now supports React's Transaction API. For the transition-enabled router, it seems to cause unwanted fetches when the InstantSeach component hydrates after SSR

### 🔍 Steps to reproduce

1. Go to https://7lylvl-3000.csb.app/
2. Click the navigations on the top left corner
3. You should see both remix loader data and the Algolia query in the network tab in the dev tool. 
4. The Algolia queries in below screenshot are unwanted

image

### Live reproduction

https://codesandbox.io/p/sandbox/shy-leftpad-7lylvl

### 💭 Expected behavior

https://codesandbox.io/p/sandbox/broken-resonance-dwtqf2

Here is another minimum setup that uses Algolia's official Remix SSR example, but the remix version is 1.16.1 which doesn't uses the `startTransition` enabled router. If you click around the navigation, you can only see Remix's loader data fetches which contain the `serverState` that Algolia needs to render the page

https://dwtqf2-3000.csb.app/

image

### Package version

"react-instantsearch": "7.0.2"

### Operating system

macOs 13.4.1

### Browser

Chrome 116.0.5845.140

### Code of Conduct

### LINKED ISSUE #5864: Extra client-side call being made when navigating to Algolia page (Next 13, pages dir)
### 🐛 Current behavior

When using the [boilerplate code](https://github.com/algolia/instantsearch/tree/master/examples/react/next), navigating from a simple page to a page using InstantSearch, unwanted Algolia requests appear in the network tab in the browser:

Scherm­afbeelding 2023-09-22 om 16 51 13

**In my older projects where I'm using NextJS 12, this issue does not occur. The page simply loads without a client-side call to algolia.**

### 🔍 Steps to reproduce

1. From the homepage, click on the "Go to algolia page" link
2. Notice how a new Algolia request appears in the network tab, as soon as the Algolia page loads

### Live reproduction

https://codesandbox.io/p/sandbox/focused-hoover-z58yzs?file=%2Fpages%2Fsearch.tsx%3A21%2C10-21%2C15

### 💭 Expected behavior

No client-side algolia call when switching between pages.

### Package version

instantsearch 7.1.0, next 13.5.1

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

dhayab:
This draft PR shows a proof of concept of hydrating the search client with initial results obtained during server-side rendering, so that the initial client-side request hits the cache.

The code is adapted from the implementation that worked in React InstantSearch v6: https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-core/src/core/createInstantSearchManager.js#L329.

**results**:

fixes #5864
fixes #5839
FX-2647


packages/react-instantsearch-core/src/lib/useInstantSearchApi.ts
```diff
@@ -106,6 +107,8 @@ export function useInstantSearchApi<TUiState extends UiState, TRouteState>(
       search._initialResults = initialResults || {};
     }
 
+    hydrateSearchClient(props.searchClient, initialResults);
```
Haroenv, line 110: personally I would see this as part of the _initialResults code in InstantSearch, not in React InstantSearch

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -594,6 +595,8 @@ See ${createDocumentationLink({
     });
 
     if (this._initialResults) {
+      hydrateSearchClient(this.client, this._initialResults);
+
       const originalScheduleSearch = this.scheduleSearch;
```
Haroenv, line 600: do we need to do all this orchestration still if the client has a cache? I guess for clients without cache?

Haroenv (commented):
👍 with TS this works for me, thanks!

packages/instantsearch.js/src/lib/InstantSearch.ts
dhayab: I think it's fine to keep this for when cache is disabled indeed.

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
```diff
@@ -0,0 +1,132 @@
...
+                  indexName: request.index,
+                  params: request.params,
+                }))
+              ),
+            []
+          ),
+        ],
+      },
+      {
+        results: Object.values(results).reduce(
```
Haroenv, line 69: Object.values isn't in the polyfill allowlist, either allow it or change the function used (using Object.keys for example)

Haroenv (approved):

PR #5859, opened 2023-09-20
fix: problem in dev with Next 13.5 + routing changes

aymeric-giraudet:
**Summary**

- Bumped Next to `13.5`
- `initialResults` simply get injected as a single object rather than an array which we call `pop` on, it's not needed anymore and fails in dev mode as the page mounts twice
- Fixed typing in `routing.router` option, should be partial
- Fixed a small thing in multipage routing

### DIFF
--- a/packages/react-instantsearch-nextjs/package.json
+++ b/packages/react-instantsearch-nextjs/package.json
@@ -49,7 +49,7 @@
   },
   "devDependencies": {
     "instantsearch.js": "4.56.11",
-    "next": "13.4.19",
+    "next": "13.5.1",
     "react-instantsearch-core": "7.0.3"
   },
   "peerDependencies": {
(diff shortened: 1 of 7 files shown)


packages/react-instantsearch-nextjs/package.json
```diff
@@ -49,7 +49,7 @@
   },
   "devDependencies": {
     "instantsearch.js": "4.56.11",
-    "next": "13.4.19",
+    "next": "13.5.1",
     "react-instantsearch-core": "7.0.3"
   },
   "peerDependencies": {
```
Haroenv, line 55: do we not need to bump peer too?

Haroenv (commented):
👍

packages/react-instantsearch-nextjs/package.json
aymeric-giraudet: Not necessarily but it's best we keep one version of Next in `node_modules`

aymeric-giraudet: I could have kept the examples on `13.4` but `13.5` builds faster so that's nicer for CI :D

dhayab (approved):

Haroenv (approved):

PR #5860, opened 2023-09-20
fix: typo fixed in next app router docs

thedevdavid:
**Summary**

Fixed 2 typos in the new next package README file

### DIFF
--- a/packages/react-instantsearch-nextjs/README.md
+++ b/packages/react-instantsearch-nextjs/README.md
@@ -75,10 +75,12 @@ const searchClient = algoliasearch('YourApplicationID', 'YourSearchOnlyAPIKey');
 export function Search() {
   return (
 -   <InstantSearch indexName="YourIndexName" searchClient={searchClient}>
-+   <InstantSearchSSRNext indexName="YourIndexName" searchClient={searchClient}>
++   <InstantSearchNext indexName="YourIndexName" searchClient={searchClient}>
       <SearchBox />
       {/* other widgets */}
-    </InstantSearch>
+
+-   </InstantSearch>
++   </InstantSearchNext>
   );
 }
 ```


Haroenv (approved):
Thanks! we had both of these options as a previous name for the component

Haroenv:
Does the new package work well for you? is there any feedback you have on how it works?

thedevdavid:
> Does the new package work well for you? is there any feedback you have on how it works?

At first, it works quite well I'd say. I managed to integrate it into a new Next.js 13.5 app, and Instantsearch works well with Apollo Client & Next-Auth in the mix. The integration was a bit confusing though because the docs suggested rendering a `<Search />` component into a page instead of the usual provider pattern on a layout.

I made a `<SearchProvider>` component instead, everything looks okay. Next is to figure out the SEO-friendly routing. That's going to be a bit complex because my search records have dozens of filtering options, sooo we'll definitely see how it works.

aymeric-giraudet:
Hey @thedevdavid, we made a component to replace `<InstantSearch>` as we need to inject a component in its children that waits for results to resume rendering of widgets (thanks to the new `use` function in React)

Do you use InstantSearch in multiple pages ? Perhaps even somewhere in the layout, like for an autocomplete searchbox in a header ?
What's in your `<SearchProvider>` exactly ? Asking these questions to understand better so we can give better documentation / features :D

thedevdavid:
> Do you use InstantSearch in multiple pages ? Perhaps even somewhere in the layout, like for an autocomplete searchbox in a header ? What's in your `<SearchProvider>` exactly ? Asking these questions to understand better so we can give better documentation / features :D

Sure. Thanks for listening! I did set up very vague and dirty example of what I want to achieve: https://codesandbox.io/p/sandbox/nostalgic-drake-p83fp6

The use case is that I want to have control over my layout and where to render widgets, refinements, results, autocomplete, etc. and I have searchbars on 2 routes. One on the search page, and another (an autocomplete) on the landing.

How would you solve this issue?

thedevdavid:
@aymeric-giraudet , @Haroenv also 1 more use-case:
Building a custom modal/portal for refinements and filters. Example:
image

Haroenv:
For the modal use case, https://github.com/algolia/instantsearch/pull/5123 will be useful for you, as it won't remove the parameters for widgets that are still mounted somewhere else.

thedevdavid:
Sounds good! Do you have any ETA on that?

Haroenv:
should be soon, it's planned

PR #5861, opened 2023-09-21
fix(InstantSearchNext): do not scroll on push

aymeric-giraudet:
**Summary**

`router.push` scrolls to the top by default, so I made this PR to disable that as we don't do this is in other flavors.
We could make it into an option but I think it's best to leave it like that, it probably makes more sense to scroll to the top depending on the widget, for example Pagination, whereas it wouldn't make sense with InfiniteHits.

Also included a small fix to the README file

### DIFF
--- a/packages/react-instantsearch-nextjs/README.md
+++ b/packages/react-instantsearch-nextjs/README.md
@@ -58,7 +58,7 @@ export function Search() {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


Haroenv (approved):
good! scroll to top should be a separate feature from routing

PR #5868, opened 2023-09-25
test(searchbox): add common test suite

aymeric-giraudet:
**Summary**

### [FX-2520](https://algolia.atlassian.net/browse/FX-2520)

**Result**

Had to update a bunch of snapshots as I tried to make flavors output the same markup

[FX-2520]: https://algolia.atlassian.net/browse/FX-2520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/utils/matchers/vue.ts
```diff
@@ -41,7 +41,7 @@ const toHaveBooleanAttribute =
     if ((isVue2 && value === attribute) || (isVue3 && value === '')) {
       return {
         pass: true,
-        message: () => '',
+        message: () => ``,
```
dhayab, line 44: Is that a prettier output?

packages/react-instantsearch/src/widgets/__tests__/SearchBox.test.tsx
```diff
@@ -151,115 +24,6 @@ describe('SearchBox', () => {
     });
   });
```
dhayab, line 25: This could be moved in the common test suite by taking into account case adjustments for `autoFocus` in IS.js and Vue, in their package's common test implementation. Wdyt?

packages/vue-instantsearch/src/components/SearchInput.vue
```diff
@@ -66,15 +62,13 @@
       <slot name="reset-icon">
         <svg
           aria-hidden="true"
-          role="img"
-          xmlns="http://www.w3.org/2000/svg"
-          style="width: 1em; height: 1em"
+          height="10"
           viewBox="0 0 20 20"
           :class="suit('resetIcon')"
+          width="10"
```
dhayab, line 68: Does it still render correctly without the xmlns attribute?

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,436 @@
...
+      await act(async () => {
+        await wait(0);
+      });
+
+      userEvent.type(screen.getByRole('searchbox'), 'iPhone');
+
+      expect(state.indexName?.query).not.toBeDefined();
+      expect(screen.getByRole('searchbox')).toHaveValue('iPhone');
+    });
+  });
```
dhayab, line 435: What's the difference between this test and the specific one in Vue?

dhayab (commented):
Overall looks good, I just have some questions before I approve it.

packages/vue-instantsearch/src/components/SearchInput.vue
aymeric-giraudet: It's required only in `.svg` files but optional when embedded in HTML, and it already worked in React and vanilla so that's why I removed it

tests/common/widgets/search-box/options.ts
aymeric-giraudet: There is no `searchAsYouType` option in Vue

packages/react-instantsearch/src/widgets/__tests__/SearchBox.test.tsx
aymeric-giraudet: Problem is they are handled differently, in IS.js it adds the `autofocus` attribute but for some reason jsdom does not focus it so we have to test the attribute rather than `toHaveFocus`.
For React, [it doesn't get added to the list of attributes](https://github.com/facebook/react/blob/34de2986dfeddda31cb76d298c6d59b271d976fa/packages/react-dom-bindings/src/client/ReactDOMComponent.js#L636-L641) but it gets polyfilled so here we can only test with `toHaveFocus` 🫠

aymeric-giraudet: I could do something like:
```js
try {
  expect(screen.getByRole('searchbox').toHaveAttribute('autofocus', 'true');
catch {
  expect(screen.getByRole('searchbox').toHaveFocus();
}
```

dhayab (approved):

packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
```diff
@@ -191,7 +192,7 @@ class SearchBox extends Component<
             rootProps={{
               className: cssClasses.submit,
               type: 'submit',
-              title: 'Submit the search query.',
+              title: 'Submit the search query',
```
sarahdayan, line 195: If we're really strict about it, this is breaking. A CSS rule targeting the element by `title` would no longer work.

packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
```diff
@@ -204,7 +205,7 @@ class SearchBox extends Component<
             rootProps={{
               className: cssClasses.reset,
               type: 'reset',
-              title: 'Clear the search query.',
+              title: 'Clear the search query',
```
sarahdayan, line 208: If we're really strict about it, this is breaking. A CSS rule targeting the element by `title` would no longer work.

packages/instantsearch.js/src/components/SearchBox/__tests__/SearchBox-test.tsx
sarahdayan, line 1: I don't think we need those component tests anymore. In the other common test suite PRs we've only kept the ones in React InstantSearch for when we create an `instantsearch-vdom` package in the future. The ones in InstantSearch.js should go away as they only test an internal API that's covered by our widget tests.

packages/instantsearch.js/src/components/SearchBox/__tests__/__snapshots__/SearchBox-test.tsx.snap
sarahdayan, line 1: As per my other comment, I think we can get rid of this.

packages/instantsearch.js/src/widgets/search-box/__tests__/search-box.test.tsx
sarahdayan, line 1: In this file you could merge the tests from `search-box-test.ts`. You can check the previously merged common test suite PRs to see what's usually migrated — what should be enough to have is as "options" `describe` block with the following tests:

- "throws without a `container`"
- "adds custom CSS classes"

packages/vue-instantsearch/src/components/SearchInput.vue
sarahdayan, line 1: Removing attributes from the markup is technically breaking, as users may be targeting them with CSS or even custom JavaScript.

sarahdayan (changes requested):

packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
dhayab: I don't think the prop changes on this PR should block us, as long as we provide consistent class names that are consistent and well documented, for customers to rely on.

packages/instantsearch.js/src/widgets/search-box/__tests__/search-box.test.tsx
dhayab: I've added the relevant tests to the correct test file and deleted the older one.

packages/vue-instantsearch/src/components/SearchInput.vue
dhayab: I don't think the prop changes on this PR should block us, as long as we provide consistent class names that are consistent and well documented, for customers to rely on.

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,461 @@
...
+import type { TestOptions } from '../../common';
+import type { InstantSearch, UiState } from 'instantsearch.js';
+
+function normalizeSnapshot(html: string) {
+  return (
+    commonNormalizeSnapshot(html)
+      // Preact and React transform some SVG attributes differently
+      .replace(/fillRule/g, 'fill-rule')
+      .replace(/strokeWidth/g, 'stroke-width')
+      .replace(/strokeOpacity/g, 'stroke-opacity')
```
Haroenv, line 24: do both work? I thought that possibly the attributes need to be camelCase

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,461 @@
...
+                      stroke-opacity=".5"
+                    />
+                    <path
+                      d="M36 18c0-9.94-8.06-18-18-18"
+                    >
+                      <animatetransform
+                        attributeName="transform"
+                        dur="1s"
+                        from="0 18 18"
+                        repeatCount="indefinite"
```
Haroenv, line 151: this snapshot is now quite inconsistent, stroke-opacity in kebab-case, but repeatCount, attributeName in camelCase (also I thought the tag was animateTransform, not animatetransform).

Of course most likely this all works fine in html, but I'd still like to see our html output & tests to use only a single style.

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,461 @@
...
+      await act(async () => {
+        userEvent.type(screen.getByRole('searchbox'), 'iPhone');
+
+        await wait(0);
+      });
+
+      expect(
+        screen.getByRole('button', {
+          name: /clear the search query/i,
+        })
```
Haroenv, line 219: we're not using the same selector in both cases, maybe use the class for both?

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,461 @@
...
+      await act(async () => {
+        userEvent.type(screen.getByRole('searchbox'), 'a');
+
+        await wait(0);
+      });
+
+      expect(
+        document.querySelector('.ais-SearchBox-loadingIndicator')
+      ).toBeVisible();
+    });
```
Haroenv, line 261: then a test that after a timeout the indicator goes away again?

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,461 @@
...
+        await wait(0);
+      });
+
+      userEvent.type(screen.getByRole('searchbox'), 'iPhone');
+
+      expect(state.indexName.query).toBe('iPhone');
+      expect(screen.getByRole('searchbox')).toHaveValue('iPhone');
+    });
+
+    test('refines when hitting the Enter key', async () => {
```
Haroenv, line 307: this test also passes when searchAsYouType: false isn't implemented. It should maybe be in a describe so that can be tested and skipped in vue

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,461 @@
...
+      await act(async () => {
+        await wait(0);
+      });
+
+      userEvent.type(screen.getByRole('searchbox'), '{enter}');
+
+      expect(state.indexName.query).toBe('iPhone');
+    });
+
+    test('refines when clicking on the submit button', async () => {
```
Haroenv, line 335: same here, the state is also allowed to be updated if searchAsYouType is false

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,461 @@
...
+            },
+          },
+        },
+        widgetParams: {},
+      });
+
+      await act(async () => {
+        await wait(0);
+      });
+
```
Haroenv, line 388: we should assert that the query _is_ set

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,461 @@
...
+      userEvent.click(
+        screen.getByRole('button', {
+          name: /clear the search query/i,
+        })
+      );
+
+      expect(state.indexName.query).not.toBeDefined();
```
Haroenv, line 395: here and in the other tests which test the state, we also need to test that the input value is the one you expect

(I know the dashboard had a bug until last week where clear didn't actually clear the input value, which is easy to accidentally be in because of the state alias)

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,461 @@
...
+          onStateChange: ({ setUiState }) => (localSetUiState = setUiState),
+        },
+        widgetParams: {},
+      });
+
+      await act(async () => {
+        await wait(0);
+      });
+
+      userEvent.type(screen.getByRole('searchbox'), 'i');
```
Haroenv, line 415: we can assert the ui state too

tests/common/widgets/search-box/options.ts
```diff
@@ -0,0 +1,461 @@
...
+        userEvent.tab();
+
+        await wait(0);
+      });
+
+      expect(screen.getByRole('searchbox')).toHaveValue('iPhone');
+    });
+  });
+
+  skippableDescribe('searchAsYouType option', skippedTests, () => {
```
Haroenv, line 437: the earlier seachAsYouType: false tests should be in here

Haroenv (changes requested):
great implementation, but the tests could be a little more precise

tests/common/widgets/search-box/options.ts
dhayab: To me the inconsistency is initially on React's side (screenshot from getting started): 
CleanShot 2023-11-02 at 11 52 30@2x

I believe nor JS neither Vue's DOM outputs are better in this case, we just need to choose one source of truth and make adjustments for the others.

tests/common/widgets/search-box/options.ts
dhayab: JS (Preact) has the first format, React and Vue the second.

tests/common/widgets/search-box/options.ts
dhayab: There's an issue with Vue in a testing environment where the searchbox still reports the previous value, not sure how to fix it. Do you have an idea?

dhayab: Found it: we took over `onStateChange` and did not propagate the new ui state ourselves.

tests/common/widgets/search-box/options.ts
Haroenv: Yes, but the camelcase name is actually react-specific and only `stroke-opacity` works. https://codepen.io/Haroenv/pen/poGEqpz

We should fix the preact/js output to be kebab-case as it should be, and possibly it should be removed if it doesn't even work?

Haroenv (approved):
as this already existed before this PR, I'm ok with the inconsistent/wrong svg output, but we should make sure that all html is actually valid (separately of this PR I guess)

tests/common/widgets/search-box/options.ts
dhayab: Fixed in 7789ccc59

Haroenv: the snapshot "cleaning" also should be removed, right @dhayab?

dhayab: Definitely 😅 on it.

Haroenv (approved):

sarahdayan (approved):
Approving this because this does what was expected and we need to move forward, but I insist that prop changes can be dangerous. Here it's unlikely that it would break anything (I don't imagine many customers style using the `title` attribute) but no matter how well we document, we could break an implementation from one minor to the next.

Anyway, let's go 🚀

PR #5873, opened 2023-09-29
feat(useInstantSearch): expose renderState

Haroenv:
**Summary**

The renderState is useful for interacting with a different widget that you know / assume is also mounted. This is good for example for refining within a hit component without adding a second refinement widget (which could have side effects if hits is conditional, despite that impact being made smaller through #5123).

Can also be useful for a reusable widget that allows the user to pass their own options to e.g. refinementList, but still expects it to be mounted by the user.

An example use case:

```js
function Hits(props) {
  const { renderState } = useInstantSearch();
  const { hits } = useHits();

  return (
    <ol>
      {hits.map((hit) => (
        <li key={hit.objectID}>
          <button
            type="button"
            onClick={() => {
              indexRenderState.refinementList.keyword.refine(hit.keyword);
            }}
          >
            {hit.keyword}
            {indexRenderState.refinementList.keyword.items.find(
              (item) => item.label === hit.keyword
            )
              ? ' (refined)'
              : ''}
          </button>
        </li>
      ))}
    </ol>
  );
}
```

**Result**

[RFC-71](https://github.com/algolia/instantsearch-rfcs/blob/master/implemented/use-instantsearch.md) is implemented more completely

### DIFF
--- a/packages/react-instantsearch-core/src/lib/useSearchState.ts
+++ b/packages/react-instantsearch-core/src/lib/useSearchState.ts
@@ -10,6 +10,8 @@ export type SearchStateApi<TUiState extends UiState> = {
   setUiState: InstantSearch<TUiState>['setUiState'];
   indexUiState: TUiState[string];
   setIndexUiState: IndexWidget<TUiState>['setIndexUiState'];
+  renderState: InstantSearch<TUiState>['renderState'];
+  indexRenderState: InstantSearch<TUiState>['renderState'][string];
 };
 
 export function useSearchState<
@@ -20,6 +22,8 @@ export function useSearchState<
   const indexId = searchIndex.getIndexId();
   const [uiState, setLocalUiState] = useState(() => search.getUiState());
   const indexUiState = uiState[indexId] as TUiState[string];
+  const [renderState, setRenderState] = useState(() => search.renderState);
+  const indexRenderState = renderState[indexId] || {};
 
   const setUiState = useCallback<SearchStateApi<TUiState>['setUiState']>(
     (nextUiState) => {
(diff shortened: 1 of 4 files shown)


aymeric-giraudet (approved):

packages/react-instantsearch-core/src/lib/useSearchState.ts
```diff
@@ -20,6 +22,8 @@ export function useSearchState<
   const indexId = searchIndex.getIndexId();
   const [uiState, setLocalUiState] = useState(() => search.getUiState());
   const indexUiState = uiState[indexId] as TUiState[string];
+  const [renderState, setRenderState] = useState(() => search.renderState);
+  const indexRenderState = renderState[indexId] || {};
```
francoischalifour, line 26: Is there a reason why we have this `{}` fallback for the index render state but it's not needed for the index UI state?

francoischalifour (approved):
I'm wondering if at some point we want to provide render state as a generic to InstantSearch, like UI state and Route state. Currently I don't think there's a way to have a type-safe render state with custom widgets.

packages/react-instantsearch-core/src/lib/useSearchState.ts
Haroenv: getUiState uses the mounted index widgets (all) as its start value, but renderState starts with an empty object

Haroenv:
> Currently I don't think there's a way to have a type-safe render state with custom widgets

You're right, I think it should be reworked, and instead of passing UiState it should be a list of widgets and we can infer all other things from it (uiState, renderState), but not easy to do that backwards compatible / correctly

PR #5874, opened 2023-09-29
test(menu-select): common test suite

aymeric-giraudet:
**Summary**

There weren't many logic tests to remove, probably because they're normally tested in the connector (so `connectMenu`)
Also obviously had to skip testing for React as it doesn't have `MenuSelect`


tests/common/widgets/menu-select/index.ts
```diff
@@ -0,0 +1,29 @@
...
+import type { MenuSelectWidget } from 'instantsearch.js/es/widgets/menu-select/menu-select';
+
+type WidgetParams = Parameters<MenuSelectWidget>[0];
+export type MenuSelectWidgetSetup = TestSetup<
+  {
+    widgetParams: Omit<WidgetParams, 'container'>;
+  },
+  {
+    flavor: string;
+  }
```
dhayab, line 15: We don't need this here, do we?

packages/instantsearch.js/src/widgets/menu-select/__tests__/menu-select.test.tsx
dhayab, line 1: Can you add a test for `cssClasses`?

aymeric-giraudet: Can't find other CTS that tests this, when I change it it doesn't seem to be taken into account for vue

dhayab: Not in CTS, in the JS widget tests. You can look at hits or current refinements for example. We should have done it for all CTS migrated widgets but it seems we forgot a few on the way when I searched for `adds custom CSS classes` 😓 .

aymeric-giraudet: Ah yes, done 👍

Haroenv (approved):

dhayab (approved):

PR #5878, opened 2023-10-03
feat(insights): allow sending additional data in events payloads

dhayab:
This PR adds the ability for users to specify additional data when sending events.

The main use case is currently to handle `addToCart` and `purchase` events, but any payload data can be send with any supported event type.

FX-2651

### DIFF
--- a/examples/js/getting-started/index.html
+++ b/examples/js/getting-started/index.html
@@ -36,13 +36,15 @@
     <div class="container">
       <div class="search-panel">
         <div class="search-panel__filters">
-          <div id="brand-list"></div>
+          <div id="category"></div>
         </div>
 
         <div class="search-panel__results">
           <div id="searchbox"></div>
-          <div id="hits"></div>
-          <div id="pagination"></div>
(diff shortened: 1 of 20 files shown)


examples/js/getting-started/index.html
```diff
@@ -36,13 +36,15 @@ <h1 class="header-title">
...
         </div>
 
         <div class="search-panel__results">
           <div id="searchbox"></div>
-          <div id="hits"></div>
-          <div id="pagination"></div>
+          <h2>sendEvent</h2>
+          <div id="hits-sendevent"></div>
+          <h2>bindEvent</h2>
+          <div id="hits-bindevent"></div>
```
Haroenv, line 47: to revert?

Haroenv (commented):
test & code look correct, example to revert :) (I wonder if we could get a good flow of still showing the example for the PR but not merging it, maybe using code sandbox, but that doesn't work for local development)

Haroenv (approved):

PR #5883, opened 2023-10-04
feat(instantsearch): allow the insights middleware to be added in answer to a server setting

dhayab:
**Summary**

When `insights` is undefined, the Insights middleware can be initialized in answer to a server setting, received with the initial results.

FX-2643

### DIFF
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -682,7 +682,10 @@ See documentation: ${createDocumentationLink({
           .some(({ results }) => results?._automaticInsights);
         if (hasAutomaticInsights) {
           this.use(
-            createInsightsMiddleware({ $$internal: true, $$automatic: true })
+            createInsightsMiddleware({
+              $$internal: true,
+              $$clickAnalytics: false,
+            })
           );
         }
       });
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -677,10 +677,10 @@ See documentation: ${createDocumentationLink({
     // Any user-provided middleware will be added later and override this one.
     if (typeof this._insights === 'undefined') {
       mainHelper.derivedHelpers[0].once('result', () => {
-        const hasQueryID = this.mainIndex
+        const hasAutomaticInsights = this.mainIndex
           .getScopedResults()
-          .some(({ results }) => results?.queryID !== undefined);
-        if (hasQueryID) {
+          .some(({ results }) => results?._automaticInsights === true);
+        if (hasAutomaticInsights) {
           this.use(
             createInsightsMiddleware({ $$internal: true, $$automatic: true })
           );
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -679,7 +679,7 @@ See documentation: ${createDocumentationLink({
       mainHelper.derivedHelpers[0].once('result', () => {
         const hasAutomaticInsights = this.mainIndex
           .getScopedResults()
-          .some(({ results }) => results?._automaticInsights === true);
+          .some(({ results }) => results?._automaticInsights);
         if (hasAutomaticInsights) {
           this.use(
             createInsightsMiddleware({ $$internal: true, $$automatic: true })
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -207,6 +207,7 @@ class InstantSearch<
   public _createURL: CreateURL<TUiState>;
   public _searchFunction?: InstantSearchOptions['searchFunction'];
   public _mainHelperSearch?: AlgoliaSearchHelper['search'];
+  public _insights: InstantSearchOptions['insights'];
   public middleware: Array<{
     creator: Middleware<TUiState>;
     instance: MiddlewareDefinition<TUiState>;
@@ -342,6 +343,8 @@ See documentation: ${createDocumentationLink({
     this._initialUiState = initialUiState as TUiState;
     this._initialResults = null;
 
+    this._insights = insights;
+
     if (searchFunction) {
       warning(
         false,
@@ -358,8 +361,9 @@ See documentation: ${createDocumentationLink({
       this.use(createRouterMiddleware(routerOptions));
     }
 
-    // This is the default middleware,
-    // any user-provided middleware will be added later and override this one.
+    // This is the default Insights middleware,
+    // added when `insights` is set to true by the user.
+    // Any user-provided middleware will be added later and override this one.
     if (insights) {
       const insightsOptions = typeof insights === 'boolean' ? {} : insights;
       insightsOptions.$$internal = true;
@@ -667,6 +671,22 @@ See documentation: ${createDocumentationLink({
     this.middleware.forEach(({ instance }) => {
       instance.started();
     });
+
+    // This is the automatic Insights middleware,
+    // added when `insights` is unset and the initial results possess `queryID`.
+    // Any user-provided middleware will be added later and override this one.
+    if (typeof this._insights === 'undefined') {
+      mainHelper.derivedHelpers[0].once('result', () => {
+        const hasQueryID = this.mainIndex
+          .getScopedResults()
+          .some(({ results }) => results?.queryID !== undefined);
+        if (hasQueryID) {
+          this.use(
+            createInsightsMiddleware({ $$internal: true, $$automatic: true })
+          );
+        }
+      });
+    }
   }
 
   /**
--- a/packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
+++ b/packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
@@ -500,11 +500,11 @@ See https://www.algolia.com/doc/api-reference/widgets/configure/js/`);
       });
     const mapMiddlewares = (middlewares: InstantSearch['middleware']) =>
       middlewares.map(
-        // @ts-ignore: $$automatic is only applicable to insights middleware
-        ({ instance: { $$type, $$internal, $$automatic } }) => ({
+        // @ts-ignore: $$clickAnalytics is only applicable to insights middleware
+        ({ instance: { $$type, $$internal, $$clickAnalytics } }) => ({
           $$type,
           $$internal,
-          $$automatic,
+          $$clickAnalytics,
         })
       );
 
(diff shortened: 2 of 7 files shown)


packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -667,6 +671,22 @@ See documentation: ${createDocumentationLink({
...
       instance.started();
     });
+
+    // This is the automatic Insights middleware,
+    // added when `insights` is unset and the initial results possess `queryID`.
+    // Any user-provided middleware will be added later and override this one.
+    if (typeof this._insights === 'undefined') {
+      mainHelper.derivedHelpers[0].once('result', () => {
+        const hasQueryID = this.mainIndex
+          .getScopedResults()
```
Haroenv, line 681: does this also work when no root indexName is given? i believe no result happens then, but to check. If it doesn't call the callback in that case, we need to listen to all derivedHelpers, not just the first one.

(it reminds me that i'm not 100% sure whether https://github.com/algolia/instantsearch/blob/bed6c5fd003da7ba85618cd5f2fb4ff1e60c50f9/packages/instantsearch.js/src/lib/server.ts#L17-L19 works correctly too)

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -667,6 +671,22 @@ See documentation: ${createDocumentationLink({
...
+          .some(({ results }) => results?.queryID !== undefined);
```
Haroenv, line 682: we could need to check that clickAnalytics: true wasn't in the parameters

(and rules / secured api key, but that's not readable easily, maybe via `params`)

Haroenv (commented):
implementation seems correct, will approach once we're 100% sure this is the way we want to detect it

packages/instantsearch.js/src/lib/InstantSearch.ts
dhayab: This is not needed anymore, as we have an explicit option coming from the engine now.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -54,6 +58,7 @@ export function createInsightsMiddleware<
     insightsInitParams,
     onEvent,
     $$internal = false,
+    $$automatic = false,
```
sarahdayan, line 61: This flag is your internal parameter to decide to set `clickAnalytics=true`, which you don't want to do when you know the engine already did it, correct?

If yes, I think it deserves a rename because "automatic" refers to the behavior of auto-sending events, it's not exclusive to Dashboard opt-in IMO.

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -667,6 +671,22 @@ See documentation: ${createDocumentationLink({
...
     });
+
+    // This is the automatic Insights middleware,
+    // added when `insights` is unset and the initial results possess `queryID`.
+    // Any user-provided middleware will be added later and override this one.
+    if (typeof this._insights === 'undefined') {
+      mainHelper.derivedHelpers[0].once('result', () => {
+        const hasAutomaticInsights = this.mainIndex
+          .getScopedResults()
+          .some(({ results }) => results?._automaticInsights === true);
```
sarahdayan, line 682: Since it's either `true` or `undefined`, this should suffice.

```suggestion
          .some(({ results }) => results?._automaticInsights);
```

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -519,39 +577,31 @@ See https://www.algolia.com/doc/api-reference/widgets/configure/js/`);
...
-        }))
-      ).toEqual([
+      expect(mapMiddlewares(search.middleware)).toEqual([
         {
           $$type: 'ais.insights',
           $$internal: true,
+          $$automatic: false,
         },
       ]);
     });
```
sarahdayan, line 588: You can add a test with `insights: true` + `_automaticInsights: true` to verify the middleware is added only once.

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -474,43 +476,99 @@ See https://www.algolia.com/doc/api-reference/widgets/configure/js/`);
   });
 
   describe('insights middleware', () => {
-    test('does not add insights middleware by default', () => {
-      const search = new InstantSearch({
-        searchClient: createSearchClient(),
-        indexName: 'test',
+    const createSearchClientWithInsights = () =>
```
sarahdayan, line 479: It's long, but it's 100% unambiguous.

```suggestion
    const createSearchClientWithAutomaticInsightsOptedIn = () =>
```

sarahdayan (approved):
It's already good and works well as is, I left comments for improvement.

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -474,43 +476,99 @@ See https://www.algolia.com/doc/api-reference/widgets/configure/js/`);
...
-        }))
-      ).toEqual([]);
+          $$automatic,
+        })
+      );
+
+    test('insights: undefined does not add the insights middleware if `_automaticInsights: true` is not found in initial response', async () => {
+      const search = new InstantSearch({
+        indexName: 'indexName',
+        searchClient: createSearchClientWithInsights(),
```
dhayab, line 514:
```suggestion
        searchClient: createSearchClientWithAutomaticInsightsOptedIn(),
```

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -474,43 +476,99 @@ See https://www.algolia.com/doc/api-reference/widgets/configure/js/`);
...
+      ]);
+      search.start();
+
+      await wait(0);
+
+      expect(mapMiddlewares(search.middleware)).toEqual([]);
+    });
+
+    test('insights: undefined adds the insights middleware if `_automaticInsights: true` is found in at least one index in initial response', async () => {
+      const searchClient = createSearchClientWithInsights();
```
dhayab, line 529:
```suggestion
      const searchClient = createSearchClientWithAutomaticInsightsOptedIn();
```

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -474,43 +476,99 @@ See https://www.algolia.com/doc/api-reference/widgets/configure/js/`);
...
+          $$internal: true,
+          $$automatic: true,
+        },
+      ]);
     });
 
     test('insights: true adds only one insights middleware', () => {
       const search = new InstantSearch({
-        searchClient: createSearchClient(),
+        searchClient: createSearchClientWithInsights(),
```
dhayab, line 555:
```suggestion
        searchClient: createSearchClientWithAutomaticInsightsOptedIn(),
```

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -474,43 +476,99 @@ See https://www.algolia.com/doc/api-reference/widgets/configure/js/`);
...
           $$internal: true,
+          $$automatic: false,
         },
       ]);
     });
 
     test('insights: options adds only one insights middleware', () => {
       const search = new InstantSearch({
-        searchClient: createSearchClient(),
+        searchClient: createSearchClientWithInsights(),
```
dhayab, line 571:
```suggestion
        searchClient: createSearchClientWithAutomaticInsightsOptedIn(),
```

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -519,39 +577,31 @@ See https://www.algolia.com/doc/api-reference/widgets/configure/js/`);
...
+          $$automatic: false,
         },
       ]);
     });
 
     test('insights: options passes options to middleware', () => {
       const insightsClient = Object.assign(jest.fn(), { version: '2.6.0' });
       const search = new InstantSearch({
-        searchClient: createSearchClient(),
+        searchClient: createSearchClientWithInsights(),
```
dhayab, line 592:
```suggestion
        searchClient: createSearchClientWithAutomaticInsightsOptedIn(),
```

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
```diff
@@ -581,17 +631,27 @@ See https://www.algolia.com/doc/api-reference/widgets/configure/js/`);
...
-          $$type,
-          $$internal,
-        }))
-      ).toEqual([]);
+      expect(mapMiddlewares(search.middleware)).toEqual([]);
+    });
+
+    test('insights: false does not add Insights middleware even if `_automaticInsights: true` is set', async () => {
+      const search = new InstantSearch({
+        searchClient: createSearchClientWithInsights(),
```
dhayab, line 644:
```suggestion
        searchClient: createSearchClientWithAutomaticInsightsOptedIn(),
```

packages/instantsearch.js/src/lib/__tests__/InstantSearch-test.tsx
dhayab: Done in [e10f7ea](https://github.com/algolia/instantsearch/pull/5883/commits/e10f7eaab73557f461f36f23ef84b0d9558acbba)

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
dhayab: I renamed it `$$clickAnalytics` in [e10f7ea](https://github.com/algolia/instantsearch/pull/5883/commits/e10f7eaab73557f461f36f23ef84b0d9558acbba).

sarahdayan (approved):

JasonBerry (approved):

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -667,6 +671,25 @@ See documentation: ${createDocumentationLink({
     this.middleware.forEach(({ instance }) => {
       instance.started();
     });
+
+    // This is the automatic Insights middleware,
+    // added when `insights` is unset and the initial results possess `queryID`.
+    // Any user-provided middleware will be added later and override this one.
+    if (typeof this._insights === 'undefined') {
+      mainHelper.derivedHelpers[0].once('result', () => {
```
Haroenv, line 679: it's good, `result` still gets called, but with `null` on the first derived helper. This code works :)

Haroenv (approved):

PR #5884, opened 2023-10-05
fix(accessibility): CR-4249 instantearch accessibility fixes

paynerichards:
**Summary**

This PR addresses accessibility concerns brought up in CR-429 by adding ARIA labels to slider handles and sortBy menu, and expanding on pagination labels

Also: CR-1201 adding aria-label to SearchBox

**Result**

Targeted areas have accessible labels


packages/instantsearch.js/src/components/Pagination/Pagination.tsx
```diff
@@ -113,7 +113,7 @@ function Pagination(props: PaginationProps) {
 
         {props.showLast && (
           <PaginationLink
-            ariaLabel="Last"
+            ariaLabel={`Last Page of ${props.nbPages}`}
```
Haroenv, line 116: this sentence reads a little weird, has that been tested?

packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
```diff
@@ -183,6 +183,7 @@ class SearchBox extends Component<
             onInput={this.onInput}
             onBlur={this.onBlur}
             onFocus={this.onFocus}
+            aria-label="Search for results"
```
Haroenv, line 186: Searchbox is also used to search for filters, probably this should be a prop used by SearchBox widget + RefinementList widget

Haroenv, line 186: is this text a guideline somewhere? personally I'd expect just "search", because what else than results can you search for?

packages/instantsearch.js/src/components/Slider/Slider.tsx
```diff
@@ -92,8 +92,13 @@ class Slider extends Component<SliderProps> {
         props['data-handle-key'] === 1 && 'rheostat-handle-upper'
       );
 
+      const ariaLabel =
+        props['data-handle-key'] === 0
+          ? 'Min Filter Handle'
+          : 'Max Filter Handle';
```
Haroenv, line 98: should these be abbreviations?

packages/instantsearch.js/src/widgets/sort-by/sort-by.tsx
```diff
@@ -94,6 +94,7 @@ const renderer =
           currentValue={currentRefinement}
           options={options}
           setValue={refine}
+          ariaLabel="Sort results by"
```
Haroenv, line 97: technically this might not necessarily fit all content allowed in the values, but good enough I think

packages/react-instantsearch/src/ui/__tests__/Pagination.test.tsx
```diff
@@ -32,10 +32,10 @@ describe('Pagination', () => {
         nextPageItemText: '›',
         lastPageItemText: '››',
         pageItemText: ({ currentPage }) => String(currentPage),
-        firstPageItemAriaLabel: 'First',
-        previousPageItemAriaLabel: 'Previous',
-        nextPageItemAriaLabel: 'Next',
-        lastPageItemAriaLabel: 'Last',
+        firstPageItemAriaLabel: 'First Page',
```
Haroenv, line 35: why not other changes in React InstantSearch?

Haroenv (commented):
🪨 thanks for your contribution! I have a couple questions:

packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
paynerichards: I'm happy to change this to just "Search", we had initially said "Search for products" ([APEX-713](https://algolia.atlassian.net/browse/APEX-713)), but I figured this component would be used for more than just products so went more general

packages/react-instantsearch/src/ui/__tests__/Pagination.test.tsx
paynerichards: reverting this, as i hadn't caught that it was in wrong package. good catch! thank you

packages/instantsearch.js/src/components/Pagination/Pagination.tsx
dhayab: I see a variant in the [US Web Design System](https://designsystem.digital.gov/components/pagination/#accessibility), which I think would work better:

> Voice “last page” on the last page’s link in a bounded set. Use aria-label="last page, page [#]" to voice that the last element in the navigation is the last page in the paginated section.

dhayab (commented):
Current changes look good to me, I'll let you handle Haroen's remark and you can run `yarn jest -u` to update the test snapshots, before we can move on with this contribution.

dhayab:
@paynerichards it seems there are unresolved snapshot mismatchs, probably in /tests/common/widgets/pagination/option.ts. Is it solved when rerunning a snapshot update command with Jest?

paynerichards:
@Haroenv @dhayab, do i need to go in to add all the same changes in vue/react packages?

Haroenv:
Yes, the test and behaviour is common for all flavours, so the change should be done everywhere, thanks! If you can also do it in /specs, that would be ideal.

Sorry for missing that in earlier reviews

Haroenv (approved):

dhayab (approved):
👏

PR #5902, opened 2023-10-23
chore(insights): rename $$clickAnalytics to $$automatic 

Haroenv:
**Summary**

When the only use case is to disable click analytics parameter, that makes sense, but if we want to use that for a flag in the algoliaSource too, i think the name should be something like $$automatic

**Result**

`$$clickAnalytics` parameter changed to `$$automatic`

### DIFF
--- a/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
@@ -189,6 +189,8 @@ export function createInsightsMiddleware<
           clickAnalytics: helper.state.clickAnalytics,
         };
 
+        // We don't want to force clickAnalytics when the insights is enabled from the search response.
+        // This means we don't enable insights for indices that don't opt in
         if (!$$automatic) {
           helper.overrideStateWithoutTriggeringChangeEvent({
             ...helper.state,
--- a/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
@@ -58,7 +58,7 @@ export function createInsightsMiddleware<
     insightsInitParams,
     onEvent,
     $$internal = false,
-    $$automatic = true,
+    $$automatic = false,
   } = props;
 
   let potentialInsightsClient: ProvidedInsightsClient = _insightsClient;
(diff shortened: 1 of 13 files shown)


dhayab (commented):
As Sarah mentioned in [her comment](https://github.com/algolia/instantsearch/pull/5883/files/cb030066b1b54887773db3f3c564420235c252c9#r1361951141), the addition of `clickAnalytics` to the search parameters merits an explicit trigger variable.

This is a bit YAGNI for now, but I'm thinking one way we could have both, is to keep your changes that clearly state the mdidleware is added from an automatic toggle in the dashboard and add a new `const shouldAddClickAnalytics = !$automatic` that we can use in our if condition there. This way we can still add more conditions in the future for adding this search parameters while not inferring it is solely driven by the automatic option.

Haroenv:
I see what you mean, but I'm not sure if 

```js
const shouldAddClickAnalytics = !$$automatic;
if (shouldAddClickAnalytics) {
  setState({ clickAnalytics: true})
}
```

is really clearer than

```js
if (!$$automatic) {
  setState({ clickAnalytics: true})
}
```

dhayab (approved):
Re https://github.com/algolia/instantsearch/pull/5902#issuecomment-1777049643, it's not that important as it's internal, it will just require future contributors to dig a bit more to understand why automatic affects `clickAnalytics`. Either way I'm fine with this as is, since for now, it's the only thing adding or not this search parameter.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -189,7 +189,7 @@ export function createInsightsMiddleware<
           clickAnalytics: helper.state.clickAnalytics,
         };
 
-        if ($$clickAnalytics) {
+        if (!$$automatic) {
```
Haroenv, line 192:
```suggestion
        // We don't want to force clickAnalytics when the insights is enabled from the search response.
        // This means we don't enable insights for indices that don't opt in
        if (!$$automatic) {
```

Haroenv: this could be an alternative @dhayab

dhayab: Good for me

PR #5904, opened 2023-10-24
feat(templates): add "future" flags

Haroenv:
**Summary**

adds `future` to the templates with `preserveSharedStateOnUnmount` set to true

**Result**

These give a warning, so we want to encourage people to use these flags as much as possible.

### DIFF
--- a/packages/create-instantsearch-app/src/templates/React InstantSearch/src/App.tsx.hbs
+++ b/packages/create-instantsearch-app/src/templates/React InstantSearch/src/App.tsx.hbs
@@ -31,6 +31,8 @@ const searchClient = algoliasearch(
   '{{apiKey}}'
 );
 
+const future = { preserveSharedStateOnUnmount: true };
+
 export function App() {
   return (
     <div>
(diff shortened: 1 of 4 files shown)


packages/create-instantsearch-app/src/templates/React InstantSearch/src/App.tsx.hbs
```diff
@@ -31,6 +31,8 @@ const searchClient = algoliasearch(
   '{{apiKey}}'
 );
 
+const future = { preserveSharedStateOnUnmount: true };
+
```
dhayab, line 35: Why outside of the app? It shouldn't create issues if inlined, no?

Haroenv: inline objects can be annoying if we track the changes (which we don't yet do). Can put inline though

dhayab: It's fine as is. Good for me!

dhayab (approved):

PR #5906, opened 2023-10-25
feat: prompt users to enable support for events

dhayab:
**Summary**

This PR updates create-instantsearch-app with a new question asked during the interactive process, to get users consent on enabling events in InstantSearch templates.

This question is only added when the template has a library version that has the `insights` prop.

FX-2660

**Result**

CleanShot 2023-10-25 at 18 15 10@2x

### DIFF
--- a/packages/create-instantsearch-app/src/templates/InstantSearch.js/src/app.js.hbs
+++ b/packages/create-instantsearch-app/src/templates/InstantSearch.js/src/app.js.hbs
@@ -15,7 +15,7 @@ const search = instantsearch({
   indexName: '{{indexName}}',
   searchClient,
   future: { preserveSharedStateOnUnmount: true },
-  {{#if flags.insights}}insights: true,{{/if}}
+  {{#if enableInsights}}insights: true,{{/if}}
 });
 
 {{#if flags.autocomplete}}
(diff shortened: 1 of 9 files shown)


packages/create-instantsearch-app/src/templates/InstantSearch.js/src/app.js.hbs
```diff
@@ -15,7 +15,7 @@ const search = instantsearch({
   indexName: '{{indexName}}',
   searchClient,
   future: { preserveSharedStateOnUnmount: true },
-  {{#if flags.insights}}insights: true,{{/if}}
+  {{#if enableInsights}}insights: true,{{/if}}
```
Haroenv, line 18: We can transform the flag so this still is based on flags.insights, no? I think that would be more beautiful in template

Haroenv (approved):
LGTM, but I think better if the template still uses flags.

packages/create-instantsearch-app/src/templates/InstantSearch.js/src/app.js.hbs
dhayab: I think this changes the meaning from flags.insights = "supports the insights prop" to flags.insights = "enable insights", which is not great, even if it's internal.

I'd prefer leaving the `flags.insights` prop untouched for this reason.

dhayab: Oh okay, I see how the meaning is changed in `postProcessAnswers` for autocomplete. I'll copy this behaviour then!

Haroenv (approved):

PR #5908, opened 2023-10-26
refactor(core): move options diff to core

Haroenv:
**Summary**

We deal with updating of parameters of InstantSearch in both React InstantSearch and Vue InstantSearch, in different ways. This PR aims to fix that, by moving the updating to core.

**Result**

`search.update(props: Partial<Options>)`

Still open questions:
- how to do `insights` (does middleware need update option too?) (do we handle anything other than init props)
- how to warn/error for routing/onStateChange (or do we handle it?)
- test that it works properly of course
- should some of the code in constructor be moved to update?

### DIFF
--- a/packages/instantsearch.js/src/lib/InstantSearch.ts
+++ b/packages/instantsearch.js/src/lib/InstantSearch.ts
@@ -844,15 +844,12 @@ See documentation: ${createDocumentationLink({
 
   /**
    * Update the parameters passed to the InstantSearch constructor.
-   * @returns true if anything has been updated
    */
   public update(
     props: Partial<InstantSearchOptions<TUiState, TRouteState>>
-  ): boolean {
-    let hasUpdated = false;
+  ): void {
     if (this.indexName !== props.indexName) {
       this.helper!.setIndex(props.indexName || '').search();
-      hasUpdated = true;
     }
 
     if (this.client !== props.searchClient) {
@@ -881,12 +878,10 @@ See documentation: ${createDocumentationLink({
       // ]);
 
       this.mainHelper!.setClient(props.searchClient).search();
-      hasUpdated = true;
     }
 
     if (this.onStateChange !== props.onStateChange) {
       this.onStateChange = props.onStateChange;
-      hasUpdated = true;
     }
 
     if (this._searchFunction !== props.searchFunction) {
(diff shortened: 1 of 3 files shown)


packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -840,6 +842,100 @@ See documentation: ${createDocumentationLink({
...
+    //     // TODO: update existing middleware somehow (or maybe remount it)
+    //   } else if (options.insights && !existing) {
+    //     const insightsOptions =
+    //       typeof options.insights === 'boolean' ? {} : options.insights;
+    //     insightsOptions.$$internal = true;
+    //     this.use(createInsightsMiddleware(insightsOptions));
+    //   } else if (!options.insights && existing) {
+    //     this.unuse(existing.creator);
+    //   }
+    // }
```
dhayab, line 927: I'd love for middlewares to remove some of the load from this method. Maybe we could iterate over middlewares, call their `update()` method, and get a returned value (boolean? object?) which could tell `InstantSearch.update()` to unuse them.

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -840,6 +842,100 @@ See documentation: ${createDocumentationLink({
...
+
+    // TODO: validate other options (onStateChange, routing, give a warning for all non-updated values?)
+
+    // Updating the `routing` prop is not supported because InstantSearch.js
+    // doesn't let us change it. This might not be a problem though, because `routing`
+    // shouldn't need to be dynamic.
+    // If we find scenarios where `routing` needs to change, we can always expose
+    // it privately on the InstantSearch instance. Another way would be to
+    // manually inject the routing middleware in this library, and not rely
+    // on the provided `routing` prop.
```
dhayab, line 937: Still can't see why routing would need to change during the lifecycle of an InstantSearch app 😅.

Haroenv: yeah I'd rather ignore or show an error than handling it

packages/instantsearch.js/src/lib/InstantSearch.ts
Haroenv: in this case probably `existing.update(options.insights)` and handling that somehow in the middleware indeed was best, wasn't sure whether the api was worth the coals

PR #5912, opened 2023-10-27
fix(createURL): ensure refined value gets removed

Haroenv:
**Summary**

If there's only one refinement left, we used to keep `uiState` purposely as-is, although actually it needs to be cleared from the refinements.

This also has an impact (positive, removing stray refinements) on dispose with the new disposeMode likely, but to investigate.

This also needs to be done and tested for all other refinement widgets.

We possibly also want to remove the `refinementList` object if it no longer has values (unless routing already deals with empty objects fine)

This behaviour became obvious after https://github.com/algolia/instantsearch/pull/5696 when routing became based on uiState instead of searchParameters

**Result**

getWidgetUiState removes no-longer-valid values from uiState, instead of leaving the object as-is if there are no refinements.

Original reason for that code was to avoid recreating the object if nothing has changed, but that's not correct. Not sure if this has a perf impact, but I think it's unavoidable

Sandbox that shows the issue for refinementList: https://codesandbox.io/s/algolia-kickstart-userefinementlist-72kryv?file=/src/index.js

FX-2667


packages/instantsearch.js/src/connectors/refinement-list/connectRefinementList.ts
```diff
@@ -433,14 +433,29 @@ const connectRefinementList: RefinementListConnector =
...
+                    .toggleFacetRefinement(attribute, facetValue)
+                    .disjunctiveFacetsRefinements,
+                  fv: facetValue,
+                  nextui: this.getWidgetUiState(uiState, {
+                    searchParameters: state
+                      .resetPage()
+                      .toggleFacetRefinement(attribute, facetValue),
+                    helper,
+                  }).refinementList,
+                });
```
Haroenv, line 451:
```suggestion
```

packages/instantsearch.js/src/__tests__/common-connectors.test.tsx
```diff
@@ -109,13 +109,22 @@ const testSetups: TestSetupsMap<TestSuites> = {
         <button data-testid="RefinementList-refine">
           refine
         </button>
+        <button data-testid="RefinementList-refine-value">
```
Haroenv, line 112:
```suggestion
      <button data-testid="RefinementList-refine-value">
```

packages/instantsearch.js/src/__tests__/common-connectors.test.tsx
```diff
@@ -109,13 +109,22 @@ const testSetups: TestSetupsMap<TestSuites> = {
         <button data-testid="RefinementList-refine">
           refine
         </button>
+      <button data-testid="RefinementList-refine-value">
```
Haroenv, line 112: not sure if this is the best way to test it, maybe the value we're refining / making a createURL is configurable when you click on it (through an input for example)

packages/instantsearch.js/src/connectors/refinement-list/connectRefinementList.ts
```diff
@@ -475,10 +475,6 @@ const connectRefinementList: RefinementListConnector =
               ? searchParameters.getDisjunctiveRefinements(attribute)
               : searchParameters.getConjunctiveRefinements(attribute);
 
-          if (!values.length) {
-            return uiState;
-          }
```
Haroenv, line 480: there's similar code as this in every refinement widget, all needs to be replaced

packages/instantsearch.js/src/connectors/refinement-list/connectRefinementList.ts
```diff
@@ -562,4 +559,27 @@ As this is not supported, please make sure to only use this attribute with one o
     };
   };
 
+function removeEmptyRefinementsFromUiState(indexUiState: IndexUiState) {
```
Haroenv, line 562: does removing this function break any tests?

dhayab: Yes, it adds empty values in widget's ui states, which are not expected by tests, and can in some instance show with the default router. I've added this method in each connector for now, but I'd like to see if I can clean these empty refinements from a single location instead.

Haroenv: maybe it can be done in the index widget after getUiState has been called, so you only need to iterate once, instead of for every widget (although getWidgetUiState is called in multiple places of course)

dhayab: It wasn't feasible to remove empty attributes at a later point in the index widget, as:
- some parts of the call directly use results from connectors' `getWidgetUiState()`
- the predicates used to determine wether attributes are considered empty differs between connectors

aymeric-giraudet (approved):
🎉

sarahdayan (approved):
Looks correct!

Optional, but we could probably factorize all the `removeEmptyRefinementsFromUiState` functions into one with a variable `key` (and maybe some more logic to pass for the reduce) and see if this has an impact on the increased bundle size.

dhayab:
> Looks correct!
> 
> Optional, but we could probably factorize all the `removeEmptyRefinementsFromUiState` functions into one with a variable `key` (and maybe some more logic to pass for the reduce) and see if this has an impact on the increased bundle size.

Tried it initially but I got so many frustrations with types I went back to this state. Ideally the refactored function would accept the connector indexUiState as a generic type which extends a base type, and a predicate argument that returns whether the value is empty or not. But the type intersections for [IndexUiState](https://github.com/algolia/instantsearch/blob/master/packages/instantsearch.js/src/types/ui-state.ts#L20-L40) are difficult to work with.

PR #5913, opened 2023-10-30
test: `hydrateSearchClient` to make sure it caches results

aymeric-giraudet:
**Summary**

#### [FX-2647](https://algolia.atlassian.net/browse/FX-2647)

I started making e2e tests with [wdio-intercept-service](https://webdriver.io/docs/wdio-intercept-service/) but due to limitations with selenium, there's no way to intercept requests made when the page loads reliably, only requests resulting from user actions can be tested :/

We would have to use Cypress or Puppeteer to test it then.

So this just tests the implementation itself, which is still better than nothing 😅
Also changed `Object.values` to `Object.keys` for it to build properly.

[FX-2647]: https://algolia.atlassian.net/browse/FX-2647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/lib/utils/__tests__/hydrateSearchClient-test.ts
+++ b/packages/instantsearch.js/src/lib/utils/__tests__/hydrateSearchClient-test.ts
@@ -1,0 +1,74 @@
+import { hydrateSearchClient } from '../hydrateSearchClient';
+
+import type { SearchClient, InitialResults } from '../../../types';
+
+describe('hydrateSearchClient', () => {
+  let client: SearchClient & {
+    _cacheHydrated?: boolean;
+    _useCache?: boolean;
+    cache?: Record<string, string>;
+  };
+  const initialResults = {
+    instant_search: {
+      results: [{ index: 'instant_search', params: 'params', nbHits: 1000 }],
+      state: {},
+      rawResults: [{ index: 'instant_search', params: 'params', nbHits: 1000 }],
+    },
+  } as unknown as InitialResults;
+
+  it('should not hydrate the client if no results are provided', () => {
+    client = {} as unknown as SearchClient;
+    hydrateSearchClient(client, undefined);
+
+    expect(client._cacheHydrated).not.toBeDefined();
+  });
+
+  it('should not hydrate the client if the cache is disabled', () => {
+    client = { _useCache: false } as unknown as SearchClient;
+
+    hydrateSearchClient(client, initialResults);
+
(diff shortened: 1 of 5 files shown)


packages/instantsearch.js/src/lib/utils/__tests__/hydrateSearchClient-test.ts
```diff
@@ -0,0 +1,74 @@
+import { hydrateSearchClient } from '../hydrateSearchClient';
+
+import type { SearchClient, InitialResults } from '../../../types';
+
+describe('hydrateSearchClient', () => {
```
Haroenv, line 5: let's add another test for all those cases to see that no actual network request happens? could spy on XMLHttpRequest probably?

bundlesize.config.json
```diff
@@ -10,7 +10,7 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "74.50 kB"
+      "maxSize": "75 kB"
```
Haroenv, line 13: is it this change causing the increase or something from the base branch? if so the commit probably could be moved (no big deal if it's squashed later of course)

aymeric-giraudet: Yeah it's caused by the base branch, I'll put it there once we merge this then

packages/instantsearch.js/src/lib/utils/__tests__/hydrateSearchClient-test.ts
aymeric-giraudet: I added a test in `InstantSearchSSRProvider` which is used for both Next.js implems, where I instantiate a search client with a mocked requester and it works well

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
```diff
@@ -53,10 +53,10 @@ export function hydrateSearchClient(
       {
         method: 'search',
         args: [
-          Object.values(results).reduce(
-            (acc, result) =>
+          Object.keys(results).reduce(
```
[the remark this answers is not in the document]
aymeric-giraudet: I needed the build to work when trying e2e which is why it's here
I think it's fine to keep it here, emphasises the version with `Object.keys` has been tested as well

PR #5927, opened 2023-11-13
test(sortBy): add common test suite

sarahdayan:
This reworks the tests for the `sortBy` widget to move them in the common test suite and reduce flavor-specific tests.

https://algolia.atlassian.net/browse/FX-2521


packages/react-instantsearch/src/ui/SortBy.tsx
```diff
@@ -42,6 +42,7 @@ export function SortBy({
         className={cx('ais-SortBy-select', classNames.select)}
         onChange={(event) => onChange(event.target.value)}
         value={value}
+        aria-label="Sort results by"
```
sarahdayan, line 45: This harmonizes the UI component with the Vue one. This seems useful since we don't provide a `<label>` element to label the `<select>`.

More info: https://accessibilityinsights.io/info-examples/web/select-name/

tests/utils/normalizeSnapshot.ts
```diff
@@ -4,7 +4,7 @@ export function normalizeSnapshot(html: string) {
       // Vue renders a useless extra space between list item elements
       .replace(/<\/li> <li/g, '</li><li')
       // Vue renders extra whitespace inside option item elements
-      .replace(/>\s+?(\w+?)\s+?<\/option>/gs, '>$1</option>')
+      .replace(/>\s+([\w() ]+?)\s+<\/option>/gs, '>$1</option>')
```
sarahdayan, line 7: This allows parentheses and spaces in option labels.

packages/vue-instantsearch/src/components/SortBy.vue
```diff
@@ -10,6 +10,7 @@
       <select
         :class="suit('select')"
         @change="state.refine($event.currentTarget.value)"
+        aria-label="Sort results by"
```
sarahdayan, line 13: Same as in https://github.com/algolia/instantsearch/pull/5927/files#r1390850941

packages/react-instantsearch/src/ui/SortBy.tsx
dhayab: We don't have translatable aria labels on our libraries for now, so I'm fine if we keep it static for now. We could maybe have a dedicated task listing all untranslatable aria attributes across the library to add support for them at once.

dhayab (approved):

packages/react-instantsearch/src/ui/SortBy.tsx
sarahdayan: We actually do have some translatable aria labels, for example for [`<Pagination>`](https://www.algolia.com/doc/api-reference/widgets/pagination/react/#widget-param-translations). But I agree this can be done separately.

aymeric-giraudet (approved):

PR #5937, opened 2023-11-20
poc: add multi-client with Recommend

sarahdayan:
This is a POC to generate a CodeSandbox. **Please don't review.**


packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -261,20 +265,30 @@ Use \`InstantSearch.status === "stalled"\` instead.`
       },
     } = options;
 
-    if (searchClient === null) {
-      throw new Error(withUsage('The `searchClient` option is required.'));
+    if (searchClient === null && client === null) {
```
Haroenv, line 268:
```suggestion
    if (searchClient === null || client === null) {
```

Haroenv: exactly, that's what I suggest :)

sarahdayan: You're suggesting that we throw if any of those terms is `false`, but we only need to do it if they both are :)

PR #5943, opened 2023-11-23
poc: warn added widgets require missing clients

dhayab:
**Summary**

**Result**

### DIFF
--- a/packages/instantsearch.js/src/connectors/frequently-bought-together/connectFrequentlyBoughtTogether.ts
+++ b/packages/instantsearch.js/src/connectors/frequently-bought-together/connectFrequentlyBoughtTogether.ts
@@ -1,12 +1,13 @@
 import { getFrequentlyBoughtTogether } from '@algolia/recommend-core';
+
 import {
-  SendEventForHits,
   checkRendering,
   createDocumentationMessageGenerator,
   createSendEventForHits,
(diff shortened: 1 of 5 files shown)


packages/instantsearch.js/src/connectors/frequently-bought-together/connectFrequentlyBoughtTogether.ts
```diff
@@ -52,6 +53,8 @@ const connectFrequentlyBoughtTogether: FrequentlyBoughtTogetherConnector =
       return {
         $$type: 'ais.frequentlyBoughtTogether',
 
+        requires: ['recommendClient'],
```
Haroenv, line 56: not related to the api, but personally I think what we should do is replace getWidgetSearchParameters with

```
getWidgetParameters() {
  return {
    search: ...,
    recommend: ...,
  }
}
```

(we can detect that the same in index of course, but it seems less hardcoded)

Haroenv: main difference is that you don't redo the work twice, once for checking and once for actually setting the parameters. If we have getWidgetParameters, you just translate the parameters, then check if anything set parameters for the reco client.

PR #5946, opened 2023-11-29
fix(cache-hydration): normalize params returned from response

aymeric-giraudet:
**Summary**

There are discrepancies between how different versions of the search engine serialize query parameters.
This makes it so this caching solution does not work with all engines.

**Result**

We can normalize the params returned from the request so that it matches the same format as the one used by the search client.

### DIFF
--- a/packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
+++ b/packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
@@ -62,7 +62,7 @@ export function hydrateSearchClient(
         params: serializeQueryParameters(request.params!),
       }));
 
-      return (client as ClientV3_4).transporter.responsesCache.get(
+      return (client as unknown as ClientV3_4).transporter.responsesCache.get(
         {
           method: 'search',
           args: [requestsWithSerializedParams, ...methodArgs],
@@ -73,7 +73,7 @@ export function hydrateSearchClient(
       );
     };
 
-    (client as ClientV3_4).transporter.responsesCache.set(
+    (client as unknown as ClientV3_4).transporter.responsesCache.set(
       {
         method: 'search',
         args: cachedRequest,
--- a/packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
+++ b/packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
@@ -1,10 +1,18 @@
-// @ts-nocheck (types to be fixed during actual implementation)
-import qs from 'qs';
+import type {
+  SearchClient,
+  InitialResults,
+  ClientV3_4,
+  SearchOptions,
+  SearchResponse,
+} from '../../types';
 
-import type { InitialResults, SearchClient } from '../../types';
+type ClientWithCache = SearchClient & { cache: Record<string, string> };
 
 export function hydrateSearchClient(
-  client: SearchClient,
+  client: SearchClient & {
+    _cacheHydrated?: boolean;
+    _useCache?: boolean;
+  },
   results?: InitialResults
 ) {
   if (!results) {
@@ -16,11 +24,26 @@ export function hydrateSearchClient(
   // - Third party clients (detected by the `addAlgoliaAgent` function missing)
 
   if (
-    (!client.transporter || client._cacheHydrated) &&
+    (!('transporter' in client) || client._cacheHydrated) &&
     (!client._useCache || typeof client.addAlgoliaAgent !== 'function')
   ) {
     return;
   }
+
+  const cachedRequest = Object.keys(results).map((key) =>
+    results[key].results.map((result) => ({
+      indexName: result.index,
+      // We normalize the params received from the server as they can
+      // be serialized differently depending on the engine.
+      params: serializeQueryParameters(
+        deserializeQueryParameters(result.params)
+      ),
+    }))
+  );
+  const cachedResults = Object.keys(results).reduce<Array<SearchResponse<any>>>(
+    (acc, key) => acc.concat(results[key].results),
+    []
+  );
 
   // Algoliasearch API Client >= v4
   // To hydrate the client we need to populate the cache with the data from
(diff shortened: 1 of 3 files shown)


packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
```diff
@@ -58,7 +60,9 @@ export function hydrateSearchClient(
               acc.concat(
                 results[key].results.map((request) => ({
```
Haroenv, line 61: name is wrong, should be result

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
```diff
@@ -1,4 +1,6 @@
 // @ts-nocheck (types to be fixed during actual implementation)
+import qs from 'qs';
```
Haroenv, line 2: not sure if the parsing is complex, `=` is encoded so we probably can do `split('&').reduce((acc, curr) => {const [key, val] = curr.split('='); acc[key] = val; return acc;}, {})` for the parse? 

that avoids leaking qs through more of the code base

Haroenv: if we had the browser support, it could use URLSearchParams or FormData, but we can't for now, but that's no big loss I guess

aymeric-giraudet: Well we would have to handle arrays as well as keys without `=` since that's what the other engine does.
I'll look into the other solution which to put the request in `initialResults`, at this point I think it's less trouble than writing our own parsing function which would have to be thoroughly tested

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
```diff
@@ -56,9 +58,11 @@ export function hydrateSearchClient(
...
             (acc, key) =>
               acc.concat(
-                results[key].results.map((request) => ({
-                  indexName: request.index,
-                  params: request.params,
+                results[key].results.map((result) => ({
+                  indexName: result.index,
+                  // We normalize the params received from the server as they can
+                  // be serialized differently depending on the engine.
+                  params: serializeQueryParameters(qs.parse(result.params)),
```
Haroenv, line 65: while this helps in some cases (different normalisation), this isn't actually what the client does:

https://github.com/algolia/algoliasearch-client-javascript/blob/075b9fa2eeb83b12700d7c9c402f0520ab04681c/packages/transporter/src/createTransporter.ts#L88-L95

it uses the params from the request to create the key, and therefore no normalisation is needed. Obviously it's annoying that we need to re-implement createMappedRequestOptions instead, but I think it's a more correct solution.

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
```diff
@@ -85,9 +89,9 @@ export function hydrateSearchClient(
       requests: Object.keys(results).reduce(
         (acc, key) =>
           acc.concat(
-            results[key].rawResults.map((request) => ({
-              indexName: request.index,
-              params: request.params,
+            results[key].rawResults.map((result) => ({
+              indexName: result.index,
+              params: serializeQueryParameters(qs.parse(result.params)),
```
Haroenv, line 94: it might or might not be right, but in v3 this is how the cache key is computed, also from the parameters, not from the result: 

https://github.com/algolia/algoliasearch-client-javascript/blob/5c1ce1308749aa0807940456393aa23ae46fb9eb/src/AlgoliaSearchCore.js#L232-L240

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
aymeric-giraudet: The problem is we don't have access to the request, but yeah the shape shouldn't be different

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
aymeric-giraudet:
```js
query.split('&').reduce((acc, curr) => {const [key, val] = curr.split('='); acc[key] = val ? decodeURIComponent(val) : '' ; return acc;}, {})
```

This seems to work fine actually, we don't need to parse the arrays

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
```diff
@@ -80,40 +90,36 @@ export function hydrateSearchClient(
...
+        results: Object.keys(results).map((key) => results[key].results),
       }),
     };
   }
 }
 
+function deserializeQueryParameters(parameters: string) {
+  return parameters.split('&').reduce<Record<string, any>>((acc, parameter) => {
+    const [key, value] = parameter.split('=');
+    acc[key] = value ? decodeURIComponent(value) : '';
```
Haroenv, line 110: don't think it's worth deviating from the design, but instead of decode/encode uri component, we could simplify it and leave params uri encoded

Haroenv (approved):
Still would like us to look into making sure we pass the `params` from server -> client as well to avoid this duplicate work, but that can be done later

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
aymeric-giraudet: Actually I can't as it's also used in the mocked `client.search` which has decoded params, or I'd have to put a boolean or something. Don't know if it's worth it.

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
Haroenv: Yes, the format is `encodeURI(param)=encodeURI(JSON.stringify(val))` I believe

dhayab (approved):

PR #5949, opened 2023-11-29
fix(nextjs): prevent onUpdate/onStateChange on own write

Haroenv:
**Summary**

Essentially, because we use router.push to write the URLs, we can't distinguish from a third-party URL/route update, which obviously should be taken in account for InstantSearch compared to a push we do ourselves in push/write, of which InstantSearch is already aware.

Impact this would be preventing is double `onStateChange` calls when you're using next routing (CR-5006)

As far as i can tell react-instantsearch-nextjs doesn't have this issue as it doesn't listen to the route change event (is that an oversight?)

**Result**

no duplicate `onStateChange` when you're using React InstantSearch router nextjs


packages/react-instantsearch-router-nextjs/__tests__/e2e/onStateChange.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      const output = await $('output#onStateChange');
+      expect(await output.getText()).toBe('1');
+    });
+  });
+
+  describe('InstantSearch', () => {
+    it('works when not on a i18n route', async () => {
+      await browser.url('/');
+
+      console.log(document.body.innerHTML);
```
Haroenv, line 38:
```suggestion
```

packages/react-instantsearch-router-nextjs/__tests__/e2e/onStateChange.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
```
Haroenv, line 39:
```suggestion

```

packages/react-instantsearch-router-nextjs/__tests__/e2e/onStateChange.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      const output = await $('output#onStateChange');
+      expect(await output.getText()).toBe('1');
+    });
+  });
+
+  describe('InstantSearch', () => {
+    it('works when not on a i18n route', async () => {
+      await browser.url('/');
+      const refinementLink = await $('.ais-RefinementList-labelText=Apple');
```
Haroenv, line 37:
```suggestion
      await browser.url('/');

      const refinementLink = await $('.ais-RefinementList-labelText=Apple');
```

packages/react-instantsearch-router-nextjs/src/index.ts
```diff
@@ -161,6 +167,7 @@ export function createInstantSearchRouterNext<TRouteState = UiState>(
       singletonRouter.push(url, undefined, {
         shallow: true,
       });
+      // lastPushFromThis = true;
```
Haroenv, line 170:
```suggestion
      lastPushFromThis = true;
```

examples/react/next-routing/pages/index.tsx
```diff
@@ -69,7 +70,13 @@ export default function HomePage({ serverState, url }: HomePageProps) {
           }),
         }}
         insights={true}
+        // only for tests
+        onStateChange={({ uiState, setUiState }) => {
+          setOnStateChangeCalled((times) => times + 1);
+          setUiState(uiState);
+        }}
       >
+        <output id="onStateChange">{onStateChangeCalled}</output>
```
Haroenv, line 79: I think we probably should either have this in a test.tsx file or a separate example, what do you think?

Haroenv (commented):
can't approve because I made the original PR, but LGTM @aymeric-giraudet

dhayab (approved):

PR #5952, opened 2023-12-05
poc: common JSX components

aymeric-giraudet:
Only tested with React and Vue 2 for now, the examples still work well.

One problem for now with Vue 2 is that we can't use `DefaultHitComponent`, as you can't call `h` with a function component. Maybe Vue 3 will work.

Next is to try with Preact (expecting TS errors) and Vue 3.


packages/vue-instantsearch/src/components/Hits.js
```diff
@@ -0,0 +1,57 @@
...
+import { connectHitsWithInsights } from 'instantsearch.js/es/connectors';
+
+import { createSuitMixin } from '../mixins/suit';
+import { createWidgetMixin } from '../mixins/widget';
+import { renderCompat } from '../util/vue-compat';
+
+const augmentH = (baseH) => (tag, propsWithClassName, children) => {
+  const { className, ...props } = propsWithClassName;
+  return baseH(tag, Object.assign(props, { class: className }), [children]);
+};
```
aymeric-giraudet, line 11: This is to map `className` to `class` and children has to be an array for Vue's `h` function

packages/vue-instantsearch/src/components/Hits.js
```diff
@@ -0,0 +1,57 @@
...
+        transformItems: this.transformItems,
+      };
+    },
+  },
+  render: renderCompat(function (baseH) {
+    if (!this.state) return null;
+
+    const h = augmentH(baseH);
+
+    return createHits({ createElement: h })({
```
aymeric-giraudet, line 52: Call to `createHits` could be externalized so that we don't call it every render

packages/vue-instantsearch/src/components/Hits.js
Haroenv: vue 2 and 3 are quite different, we have renderCompat, but I wonder if we should just have separate files for vue 2 and vue 3 once we simplify the rendering?

packages/vue-instantsearch/src/components/Hits.js
Haroenv: how can it be externalised? if I remember it correctly in vue 2 you only have access to `h` in `render`'s arguments

PR #5956, opened 2023-12-07
fix(routing): ensure UiState is cleaned up allowing "undefined" for values

Haroenv:
**Summary**

Since #5912, we ensure to clean the UiState from empty values, however this was done in a slightly inefficient (looping over values) and unsafe (no checking for undefined) way, as the types implied that RouteState/UiState always is completely given (if you're using TypeScript that's actually the case, you'd get an error returning `undefined` for refinementList[attribute] for example).

**Result**

- Simplified the UiState cleaning
- ensure `undefined` doesn't cause issues
- add new test to each `connectors/*/routing` where relevant (pushed the tests separately to show they fail with current implementation: https://app.circleci.com/pipelines/github/algolia/instantsearch/9831/workflows/cba7210e-6955-4459-8c82-307d77319955/jobs/47848/tests)

fixes https://github.com/algolia/instantsearch/issues/5954

### DIFF
--- a/tests/common/common.ts
+++ b/tests/common/common.ts
@@ -69,7 +69,7 @@ export function runTestSuites<
   testSetups: TestSetupsMap<TTestSuites>;
   testOptions: TestOptionsMap<TTestSuites>;
 }) {
-  test.skip('has all the tests', () => {
+  test('has all the tests', () => {
     expect(Object.keys(testSetups).sort()).toEqual(
       Object.keys(testSuites).sort()
     );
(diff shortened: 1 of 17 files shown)


tests/common/common.ts
```diff
@@ -66,7 +69,7 @@ export function runTestSuites<
   testSetups: TestSetupsMap<TTestSuites>;
   testOptions: TestOptionsMap<TTestSuites>;
 }) {
-  test('has all the tests', () => {
+  test.skip('has all the tests', () => {
```
Haroenv, line 72:
```suggestion
  test('has all the tests', () => {
```

dhayab (approved):
🙇

PR #5959, opened 2023-12-11
chore(node): update version

Haroenv:
**Summary**

Update node for local development.

This required a small patch for `uglify-js@2` as `fs.writeFileSync` didn't accept `Buffer` anymore directly. The patch needs to be inside `algoliasearch-helper` package, as we also have v3 globally installed.

**Result**

Node is on the latest stable version.


dhayab (approved):
> This required an update to webpack and need to pass NODE_OPTIONS to whatever is using it (would be interesting if we could migrate away from it, in Vue it could be Vite instead, for Storybook maybe we move it to -specs?)

I agree, it would be nice to do a pass on the monorepo to refresh it with either up to date versions of existing libraries or which are deprecated and need to be replaced.

PR #5960, opened 2023-12-11
test(versions): add "insights" test

Haroenv:
**Summary**

Ensure that search-insights is always on the latest version on CI

requires #5958 of course to pass CI

**Result**

We are warned in CI if there's a newer search-insights version available

example failing run: https://app.circleci.com/pipelines/github/algolia/instantsearch/9868/workflows/c1d28ba1-bcfd-4c1e-920c-b91e33ccf86a/jobs/47989
example passing run: https://app.circleci.com/pipelines/github/algolia/instantsearch/9872/workflows/0c60a622-0070-4dd9-a45b-d9a0e68a91bc/jobs/48022

(note: you can ignore netlify failure, it ran correctly here: https://app.netlify.com/sites/instantsearchjs/deploys/65784298cf0a014abe7cfd9e, seems like just the GitHub status didn't work)

### DIFF
--- a/tests/versions/index.mjs
+++ b/tests/versions/index.mjs
@@ -1,7 +1,11 @@
 #!/usr/bin/env node
-/* eslint-disable no-process-exit, no-console, import/no-commonjs */
-const fs = require('fs');
-const path = require('path');
+/* eslint-disable no-process-exit, no-console*/
+import fs from 'fs';
+import { createRequire } from 'module';
+import path from 'path';
+
+const require = createRequire(import.meta.url);
+const dirname = path.dirname(new URL(import.meta.url).pathname);
 
 // This file does not use any dependencies, so that it can be ran before installing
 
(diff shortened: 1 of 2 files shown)


dhayab:
We could go one step further and set up a weekly/monthly GitHub Action that updates the Insights version in all relevant source files + creates a PR (for example with https://github.com/marketplace/actions/create-pull-request).

Haroenv:
Yes, that could be combined

tests/versions/index.mjs
```diff
@@ -102,6 +104,50 @@ let hasError = false;
...
+          dirname,
+          '../../packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts'
+        ),
+        'utf-8'
+      )
+      .match(/ALGOLIA_INSIGHTS_VERSION = '(.*)';/);
+
+    if (latest !== dependency || latest !== middleware) {
+      console.error(`search-insights is not up to date.`);
+      hasError = true;
```
dhayab, line 138: I'm not comfortable with failing our pipelines because we don't have the latest version of the Insights client. For example, if the latest is a new major, it might require more work to update on our end, which we might not be able to allocate right away.

An alternative, in addition to a "dependabot-like" scheduled PR, could be to still pass the CI but publish a comment on the PR (or add a check like Bundlesize with a warning).

Haroenv: That's fair, I'll look into doing this in a PR instead

PR #5961, opened 2023-12-11
chore(examples): migrate vue examples to vite

Haroenv:
**Summary**

Migrate the "regular" Vue CLI examples to use Vite instead. I didn't touch the server side rendering or Nuxt example, so this doesn't yet remove any dependency.

**Result**

- "regular" Vue examples use Vite, no longer vue-cli, and therefore are no longer built with Webpack.
- "specs" Astro site is updated (so same esbuild version is used)
- *not changed*: Nuxt example, Vue server side rendering example (both still using webpack)

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -17,5 +17,5 @@
     "packages/instantsearch.css",
     "packages/instantsearch-ui-components"
   ],
-  "node": "16"
+  "node": "20"
 }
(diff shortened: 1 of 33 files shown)


Haroenv:
Unfortunately this is blocked as Astro 4 requires node 18.14+ and the latest version configurable on codesandbox CI is 18.12 😢 (cc @compuives)

aymeric-giraudet (approved):

PR #5963, opened 2023-12-12
fix(searchbox): only refine if composition ended when using an IME

dhayab:
**Summary**

Characters entered using an IME trigger search requests although the character composition is not done. This should only happen if the character has been fully composed.

This PR stops refining on `input` when composing with an IME and instead rely on the `compositionend` event in that case.

**Result**

Characters entered using an IME do not trigger search requests until relevant.

CR-5106

### DIFF
--- a/packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
+++ b/packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
@@ -87,12 +87,17 @@ class SearchBox extends Component<
     const { searchAsYouType, refine, onChange } = this.props;
     const query = (event.target as HTMLInputElement).value;
 
-    if (searchAsYouType) {
-      refine(query);
-    }
-    this.setState({ query });
+    if (
+      event.type === 'compositionend' ||
+      !(event as KeyboardEvent).isComposing
+    ) {
+      if (searchAsYouType) {
+        refine(query);
+      }
+      this.setState({ query });
 
-    onChange(event);
+      onChange(event);
+    }
   };
 
   public componentWillReceiveProps(nextProps: SearchBoxPropsWithDefaultProps) {
@@ -184,6 +189,9 @@ class SearchBox extends Component<
             spellCheck="false"
             maxLength={512}
             onInput={this.onInput}
+            // see: https://github.com/preactjs/preact/issues/1978
+            // eslint-disable-next-line react/no-unknown-property
+            oncompositionend={this.onInput}
             onBlur={this.onBlur}
             onFocus={this.onFocus}
             aria-label={ariaLabel}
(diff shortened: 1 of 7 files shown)


Haroenv:
just checking, does IE have composition support? I sas FF sometimes triggers in wrong order too: https://github.com/w3c/uievents/issues/202#issue-316461024

dhayab:
> just checking, does IE have composition support? I sas FF sometimes triggers in wrong order too: [w3c/uievents#202 (comment)](https://github.com/w3c/uievents/issues/202#issue-316461024)

IE11 doesn't implement `event.isComposing` so the current behaviour will stay (search requests will be triggered during composition). There should be no issue with the order of events, as the input value will be the same, so it is caught by the search client cache.

packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
```diff
@@ -184,6 +189,9 @@ class SearchBox extends Component<
             spellCheck="false"
             maxLength={512}
             onInput={this.onInput}
+            // see: https://github.com/preactjs/preact/issues/1978
+            // eslint-disable-next-line react/no-unknown-property
+            oncompositionend={this.onInput}
```
Haroenv, line 194: not relevant now, but I wonder how we'll deal with these differences with only one JSX source

tests/common/widgets/search-box/options.ts
```diff
@@ -307,6 +307,72 @@ export function createOptionsTests(
...
+                value: query,
+              },
+            });
+          }
+        });
+      });
+
+      expect(screen.getByRole('searchbox')).toHaveValue(character);
+
+      expect(searchClient.search).toHaveBeenCalledTimes(2);
```
Haroenv, line 364: reset the mock after setup maybe so it's clear it's only one?

Haroenv (approved):
thanks!

Haroenv (approved):

tests/common/widgets/search-box/options.ts
```diff
@@ -307,6 +307,74 @@ export function createOptionsTests(
...
+      // see:
+      // - https://en.wikipedia.org/wiki/Stroke_count_method
+      // - https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionend_event
+      const character = '木';
+      const strokes = ['一', '丨', '丿', '丶', character];
+
+      await act(async () => {
+        await wait(0);
+
+        searchClient.search.mockClear();
```
Haroenv, line 331:
```suggestion
        castToJestMock(searchClient.search).mockClear();
```

also why not right after setup + wait 0, before starting to compose?

```
await act(async () => {
  await wait(0);
  castToJestMock(searchClient.search).mockClear();
});

// ... character etc
```

dhayab: I need to wrap the events inside of `act()` so I'm reusing the one that waits for initialization.

aymeric-giraudet (approved):

PR #5966, opened 2023-12-18
feat(history): provide option to not clean url on dispose

dhayab:
**Summary**

Currently, the InstantSearch history router cleans up the URL from refinements it added during its lifecycle when it is disposed. This can be useful when InstantSearch is disposed but no route change has occurred, but the heuristics we use to detect whether we should write to the URL or not can sometimes conflict with third-party routers (ie: Next) when route changes occur.

This PR provides a option that users can use to turn off cleaning the URL on dispose. For example, with the Next.js Router:

```jsx
<InstantSearch
  routing={{
    router: createInstantSearchRouterNext({
      singletonRouter,
      serverUrl: url,
      routerOptions: {
        cleanUrlOnDispose: false,
      },
    }),
  }}
>
  // ...
</InstantSearch>
```

[CR-5095](https://algolia.atlassian.net/browse/CR-5095)

[CR-5095]: https://algolia.atlassian.net/browse/CR-5095?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/lib/routers/history.ts
+++ b/packages/instantsearch.js/src/lib/routers/history.ts
@@ -27,6 +27,14 @@ export type BrowserHistoryArgs<TRouteState> = {
   start?: (onUpdate: () => void) => void;
   dispose?: () => void;
   push?: (url: string) => void;
+  /**
+   * Whether the URL should be cleaned up when the router is disposed.
+   * This can be useful when closing a modal containing InstantSearch, to
+   * remove active refinements from the URL.
+   * @default true
+   */
+  // @MAJOR: Switch the default to `false` in the next major version.
+  cleanUrlOnDispose?: boolean;
 };
 
 const setWindowTitle = (title?: string): void => {
@@ -97,10 +105,9 @@ class BrowserHistory<TRouteState> implements Router<TRouteState> {
   private latestAcknowledgedHistory: number = 0;
 
   private _start?: (onUpdate: () => void) => void;
-
   private _dispose?: () => void;
-
   private _push?: (url: string) => void;
+  private _cleanUrlOnDispose: boolean;
 
   /**
    * Initializes a new storage provider that syncs the search state to the URL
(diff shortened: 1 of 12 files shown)


packages/instantsearch.js/src/lib/routers/history.ts
```diff
@@ -125,6 +133,7 @@ class BrowserHistory<TRouteState> implements Router<TRouteState> {
     this._start = start;
     this._dispose = dispose;
     this._push = push;
+    this._cleanUrlOnDispose = cleanUrlOnDispose;
```
Haroenv, line 136: in the constructor we can warn if the value isn't set (in dev) to make people more aware of the change they should do. also all examples then should have the option set to show the right example.

Haroenv (approved):
can go a little further in deprecating existing behaviour, but code and tests lgtm

packages/instantsearch.js/src/lib/routers/history.ts
dhayab: There's now an informational message in dev (similar to what we have for `preserveSharedStateOnUnmount`).

On updating examples, there's still the particular case of `routing: true` which I've left as-is. This means they will still write on dispose, and show a message on the console. I've chosen not to update them as I think explicitly adding the router with the option clashes with the simplicity of this approach, and users will still have to learn about the change in behaviour in migration notes / codemods for next major version.

francoischalifour:
Any reasons why this behavior is not synced with `preserveSharedStateOnUnmount`?

Haroenv:
We didn't consider it, `preserveSharedStateOnUnmount` is about the state when one of multiple widgets unmounts, while this is when 1 of 1 InstantSearch components unmounts. I think they are different?

francoischalifour:
Without digging into the technical parts (and as a user) I expected the two to work together but there might only be a slight overlap indeed.

This feature also seems to relate to the [unmount timeout logic](https://github.com/algolia/instantsearch/blob/305ed204721d219ea925e99349bcbf54bc068a48/packages/react-instantsearch-core/src/lib/useInstantSearchApi.ts#L230) we have, and this might be an opportunity to get rid of this timeout. (Which is causing issues with HMR btw.)

PR #5919, opened 2023-11-07
test(toggleRefinement): add common test suite

sarahdayan:
This reworks the tests for the `toggleRefinement` widget to move them in the common test suite and reduce flavor-specific tests.

https://algolia.atlassian.net/browse/FX-2523

PR #5975, opened 2023-12-28
test(toggle): add explicit label test

Haroenv:
**Summary**

- uses the `label` prop, casted to a template for the JS test
- fixes the type for the template in JS
- ensure the count in Vue (outlier) is tested and normalised
- remove "required" from Vue label prop (as it isn't actually required anymore since #5919)

**Result**

ToggleRefinement label is tested

### DIFF
--- a/tests/common/widgets/toggle-refinement/options.ts
+++ b/tests/common/widgets/toggle-refinement/options.ts
@@ -18,6 +18,14 @@ function normalizeSnapshot(html: string) {
       .replace(/name="\w+"/g, '')
       .replace(/value="\w+"/g, '')
       .replace(/<input.+?>(\s+)/gs, (match) => match.trim())
+      // Vue InstantSearch adds the count after the label if available.
+      // @MAJOR: Remove this once Vue InstantSearch aligns with spec.
+      .replace(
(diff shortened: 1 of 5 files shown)


tests/common/widgets/toggle-refinement/options.ts
```diff
@@ -18,6 +18,14 @@ function normalizeSnapshot(html: string) {
...
       .replace(/value="\w+"/g, '')
       .replace(/<input.+?>(\s+)/gs, (match) => match.trim())
+      // Vue InstantSearch adds the count after the label if available.
+      // @MAJOR: Remove this once Vue InstantSearch aligns with spec.
+      .replace(
+        /<span\n?\s*class="ais-ToggleRefinement-count"\n?\s*>\n?.*\s*<\/span>/g,
+        ''
+      )
+      // React renders `checked` as `checked=""` through an attribute, other frameworks use the property.
+      .replace(/checked=""/g, '')
```
dhayab, line 28: Would it make sense to move this to the common normalizer?

dhayab (approved):

tests/common/widgets/toggle-refinement/options.ts
Haroenv: not sure, you need to be aware to use the `toBeChecked` or similar check in the assertion, so I think it may be logical to leave that in the test itself

PR #5976, opened 2023-12-28
chore(connectSearchBox): don't use `searchMetadata`

Haroenv:
**Summary**

searchMetadata is deprecated, so we shouldn't read it anymore. Further here we should probably also deprecate `isSearchStalled` in the render method, but not sure how to `stalled` in React then

I do wonder if all widgets should have access to `instantSearchInstance` in the render function in React after all to be able to access these global information without also using `useInstantSearch`, but that's a different conversation

**Result**

`isSearchStalled` doesn't use `searchMetadata` anymore in search box connector


dhayab (approved):
Good to me!

As for exposing the instant search instance in the render function, i'm not opposed to the idea, but I guess we should wait to see if there's a better way to do it in the context of our current unification work.

PR #5978, opened 2023-12-28
fix(rsc): wait for results in `useSearchResults`

aymeric-giraudet:
**Summary**

Getting results from `useInstantSearch` does not work in server-side RSC because we don't wait for the results like we do in `useWidget`.

**Result**

This will result in re-mounting the hook once results are received, so that users can have access to results and display things related to it in SSR.


Haroenv (commented):
Thanks! Can we add a test or something though?

PR #5991, opened 2024-01-10
fix(hydration): generate cache with search parameters from server-side request

### LINKED ISSUE #5964: deserializeQueryParameters - TypeError: Cannot read properties of undefined (reading 'split')
## Unhandled Runtime Error

image

## Here is the line of the error code 👇

image

## Localhost console debug

Did a quick debug and found that the `result` item from `deserializeQueryParameters` has no `params` key. 

image

### 🔍 Steps to reproduce

I have updated the libraries and this error appeared

algoliasearch: 4.20.0 > 4.21.1 
instantsearch.js: 4.58.0 > 4.62.0
react-instantsearch: 7.2.0 > 7.4.1
react-instantsearch-core: 7.2.0 > 7.4.1
react-instantsearch-router-nextjs: 7.2.0 > 7.4.1

### Live reproduction

-

### 💭 Expected behavior

Working code

### Package version

react-instantsearch: 7.4.1, react-instantsearch-core: 7.4.1, react-instantsearch-router-nextjs: 7.4.1, instantsearch.js: 4.62.0, algoliasearch: 4.21.1

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

dhayab:
**Summary**

In an SSR context, the search request performed server-side is cached in the client-side search client to prevent sending an unnecessary duplicate request when the user interacts with refinements.

This is done by manually recomposing the cached results during hydration and adding it to the search client's cache mechanism. The key of a cached request uses the search parameters from the response, normally returned by Algolia, but customers can limit what fields are returned by setting the `responseFIelds` option. If `params` is omitted, than the app throws an error during hydration.

This PR changes slightly how the request key is crafted, by instead extracting the request search parameters from the server-side search request, and using it instead of the response search parameters.

**Result**

Hydration now works correctly and the server-side request is effectively cached whether fields are filtered using `responseFields` or not.

> [!NOTE]
> The issue of cache miss in implementations containing multiple indices is still unsolved and can be handled by FX-2720.
> Furthermore, if in those multiple indices there are multiples of the same `indexName` + `indexId`, the current implementation of `getInitialResults()` uses only the last instance for params and results, which might not be desirable.

Fixes #5964 
FX-2702

### DIFF
--- a/packages/instantsearch.js/src/lib/__tests__/server.test.ts
+++ b/packages/instantsearch.js/src/lib/__tests__/server.test.ts
@@ -277,8 +277,7 @@ describe('getInitialResults', () => {
     search.start();
 
     const requestParams = await waitForResults(search);
-
-    expect(getInitialResults(search.mainIndex, requestParams)).toEqual({
+    const expectedInitialResults = {
       indexName: expect.objectContaining({
         requestParams: expect.objectContaining({
           query: 'apple',
@@ -295,6 +294,16 @@ describe('getInitialResults', () => {
           hitsPerPage: 2,
         }),
       }),
-    });
+    };
+
+    expect(getInitialResults(search.mainIndex, requestParams)).toEqual(
+      expectedInitialResults
+    );
+
+    // Multiple calls to `getInitialResults()` with the same requestParams
+    // return the same results
+    expect(getInitialResults(search.mainIndex, requestParams)).toEqual(
+      expectedInitialResults
+    );
   });
 });
--- a/packages/instantsearch.js/src/lib/__tests__/server.test.ts
+++ b/packages/instantsearch.js/src/lib/__tests__/server.test.ts
@@ -25,7 +25,10 @@ describe('waitForResults', () => {
 
     searches[0].resolver();
 
-    await expect(output).resolves.toBeUndefined();
+    await expect(output).resolves.toEqual([
+      expect.objectContaining({ indexName: 'indexName' }),
+      expect.objectContaining({ indexName: 'indexName2' }),
+    ]);
   });
 
   test('throws on a search client error', async () => {
@@ -237,6 +240,45 @@ describe('getInitialResults', () => {
           },
         ],
       },
+    });
+  });
+
+  test('returns the current results with request params if specified', async () => {
+    const search = instantsearch({
+      indexName: 'indexName',
+      searchClient: createSearchClient(),
+      initialUiState: {
+        indexName: {
+          query: 'apple',
+        },
+        indexName2: {
+          query: 'samsung',
+        },
+      },
+    });
+
+    search.addWidgets([
+      connectSearchBox(() => {})({}),
+      index({ indexName: 'indexName2' }).addWidgets([
+        connectSearchBox(() => {})({}),
+      ]),
+    ]);
+
+    search.start();
+
+    const requestParams = await waitForResults(search);
+
+    expect(getInitialResults(search.mainIndex, requestParams)).toEqual({
+      indexName: expect.objectContaining({
+        requestParams: expect.objectContaining({
+          query: 'apple',
+        }),
+      }),
+      indexName2: expect.objectContaining({
+        requestParams: expect.objectContaining({
+          query: 'samsung',
+        }),
+      }),
     });
   });
 });
--- a/packages/instantsearch.js/src/lib/__tests__/server.test.ts
+++ b/packages/instantsearch.js/src/lib/__tests__/server.test.ts
@@ -272,11 +272,50 @@ describe('getInitialResults', () => {
       index({ indexName: 'indexName2', indexId: 'indexId' }).addWidgets([
         connectConfigure(() => {})({ searchParameters: { hitsPerPage: 2 } }),
       ]),
+      index({ indexName: 'indexName2', indexId: 'indexId' }).addWidgets([
+        connectConfigure(() => {})({ searchParameters: { hitsPerPage: 3 } }),
+      ]),
     ]);
 
     search.start();
 
     const requestParams = await waitForResults(search);
+
+    // Request params for the same index name + index id are not deduplicated,
+    // so we should have data for 4 indices (main index + 3 index widgets)
+    expect(requestParams).toHaveLength(4);
+    expect(requestParams).toMatchInlineSnapshot(`
+      [
+        {
+          "facets": [],
+          "query": "apple",
+          "tagFilters": "",
+        },
+        {
+          "facets": [],
+          "query": "samsung",
+          "tagFilters": "",
+        },
+        {
+          "facets": [],
+          "hitsPerPage": 2,
+          "query": "apple",
+          "tagFilters": "",
+        },
+        {
+          "facets": [],
+          "hitsPerPage": 3,
+          "query": "apple",
+          "tagFilters": "",
+        },
+      ]
+    `);
+
+    // `getInitialResults()` generates a dictionary of initial results
+    // keyed by index id, so indexName2/indexId should be deduplicated...
+    expect(Object.entries(getInitialResults(search.mainIndex))).toHaveLength(3);
+
+    // ...and only the latest duplicate params are in the returned results
     const expectedInitialResults = {
       indexName: expect.objectContaining({
         requestParams: expect.objectContaining({
@@ -291,7 +330,7 @@ describe('getInitialResults', () => {
       indexId: expect.objectContaining({
         requestParams: expect.objectContaining({
           query: 'apple',
-          hitsPerPage: 2,
+          hitsPerPage: 3,
         }),
       }),
     };
--- a/packages/instantsearch.js/src/lib/__tests__/server.test.ts
+++ b/packages/instantsearch.js/src/lib/__tests__/server.test.ts
@@ -3,7 +3,7 @@ import {
   createSearchClient,
 } from '@instantsearch/mocks';
 
-import { connectSearchBox } from '../../connectors';
+import { connectConfigure, connectSearchBox } from '../../connectors';
 import instantsearch from '../../index.es';
 import { index } from '../../widgets';
 import { getInitialResults, waitForResults } from '../server';
@@ -14,8 +14,15 @@ describe('waitForResults', () => {
     const search = instantsearch({
       indexName: 'indexName',
       searchClient,
+      initialUiState: {
+        indexName: {
+          query: 'apple',
+        },
+      },
     }).addWidgets([
-      index({ indexName: 'indexName2' }),
+      index({ indexName: 'indexName2' }).addWidgets([
+        connectConfigure(() => {})({ searchParameters: { hitsPerPage: 2 } }),
+      ]),
       connectSearchBox(() => {})({}),
     ]);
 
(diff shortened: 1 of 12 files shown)


packages/instantsearch.js/src/lib/server.ts
```diff
@@ -37,17 +55,28 @@ export function waitForResults(search: InstantSearch): Promise<void> {
...
     if (searchResults) {
-      initialResults[widget.getIndexId()] = {
+      const indexId = widget.getIndexId();
+      initialResults[indexId] = {
         // We convert the Helper state to a plain object to pass parsable data
         // structures from server to client.
         state: { ...searchResults._state },
         results: searchResults._rawResults,
+        requestParams: requestParams?.find(
+          ({ indexName }) => indexName === indexId
```
Haroenv, line 78: this should use the widget's index name, not index id

dhayab: Ah I see. I'll update the sample code to use multiple of the same index with different ids and check if I can match request + results in the hydration step.

dhayab: I simplified the implementation as extracted request params and initial results are ordered identically.

packages/instantsearch.js/src/lib/server.ts
```diff
@@ -37,17 +57,27 @@ export function waitForResults(search: InstantSearch): Promise<void> {
...
     if (searchResults) {
-      initialResults[widget.getIndexId()] = {
+      const indexId = widget.getIndexId();
+      const requestParams = requestParamsList?.shift();
+      initialResults[indexId] = {
         // We convert the Helper state to a plain object to pass parsable data
         // structures from server to client.
         state: { ...searchResults._state },
         results: searchResults._rawResults,
+        ...(requestParams && { requestParams }),
```
Haroenv, line 80:
```suggestion
       requestParams,
```

What's the difference with this? undefined if it's not available?

packages/react-instantsearch-nextjs/src/InitializePromise.tsx
```diff
@@ -26,7 +38,7 @@ export function InitializePromise() {
 
   const injectInitialResults = () => {
     let inserted = false;
-    const results = getInitialResults(search.mainIndex);
+    const results = getInitialResults(search.mainIndex, requestParamsList);
```
Haroenv, line 41: I guess I don't really understand this code right now, but how come this can't be read from waitForResults? does it need to be set before the results are known or something?

packages/react-instantsearch-nextjs/src/InitializePromise.tsx
```diff
@@ -17,6 +19,16 @@ export function InitializePromise() {
       throw new Error('Missing ServerInsertedHTMLContext');
     });
 
+  // Extract search parameters from the search client to use them
+  // later during hydration.
+  let requestParamsList: Array<SearchOptions | undefined>;
```
Haroenv, line 24: params is always defined, no? why isn't this:

```suggestion
  let requestParamsList: SearchOptions[];
```

packages/instantsearch.js/src/lib/server.ts
```diff
@@ -37,17 +57,27 @@ export function waitForResults(search: InstantSearch): Promise<void> {
...
+   * returned by `waitForResults()`.
+   */
+  requestParamsList?: RequestParams[]
+): InitialResults {
   const initialResults: InitialResults = {};
 
   walkIndex(rootIndex, (widget) => {
     const searchResults = widget.getResults();
     if (searchResults) {
-      initialResults[widget.getIndexId()] = {
```
Haroenv, line 46: I'm fine with keeping the indexId variable, but just making sure that it's not doing anything now

packages/instantsearch.js/src/lib/server.ts
```diff
@@ -37,17 +57,27 @@ export function waitForResults(search: InstantSearch): Promise<void> {
...
+      const indexId = widget.getIndexId();
+      const requestParams = requestParamsList?.shift();
```
Haroenv, line 74: should we keep track of the index and use the item at index `i` instead of mutating? Not sure if it's a good idea that `getInitialResults` effectively can only be called once (also let's add a test for calling it multiple times)

packages/instantsearch.js/src/lib/server.ts
```diff
@@ -37,17 +57,27 @@ export function waitForResults(search: InstantSearch): Promise<void> {
 /**
  * Walks the InstantSearch root index to construct the initial results.
  */
-export function getInitialResults(rootIndex: IndexWidget): InitialResults {
+export function getInitialResults(
+  rootIndex: IndexWidget,
+  /**
+   * Search parameters sent to the search client, usually
```
Haroenv, line 63: usually?

packages/instantsearch.js/src/lib/server.ts
```diff
@@ -1,21 +1,41 @@
...
 
-import type { IndexWidget, InitialResults, InstantSearch } from '../types';
+import type {
+  IndexWidget,
+  InitialResults,
+  InstantSearch,
+  SearchOptions,
+} from '../types';
+
+type RequestParams = SearchOptions | undefined;
```
Haroenv, line 10: same question, how can this be undefined?

packages/instantsearch.js/src/lib/__tests__/server.test.ts
```diff
@@ -239,4 +249,52 @@ describe('getInitialResults', () => {
       },
     });
   });
+
+  test('returns the current results with request params if specified', async () => {
```
Haroenv, line 253: we need another test with two indices with the same indexName and indexId that aren't adjacent too I think. I wonder if the deduplication could cause issues

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
```diff
@@ -34,16 +34,24 @@ export function hydrateSearchClient(
     return;
   }
 
-  const cachedRequest = Object.keys(results).map((key) =>
-    results[key].results.map((result) => ({
-      indexName: result.index,
+  const cachedRequest = Object.keys(results).map((key) => {
+    const { state, requestParams, results: serverResults } = results[key];
+    return serverResults.map((result) => ({
+      indexName: result.index || state.index,
```
Haroenv, line 40:
```suggestion
      indexName: state.index || result.index,
```

Haroenv: this should be good if we ever add a feature in the backend that redirects indices (I believe I've heard of it, not sure if it ever came to fruition)

packages/instantsearch.js/src/lib/utils/hydrateSearchClient.ts
```diff
@@ -34,16 +34,24 @@ export function hydrateSearchClient(
...
       // We normalize the params received from the server as they can
       // be serialized differently depending on the engine.
-      params: serializeQueryParameters(
-        deserializeQueryParameters(result.params)
-      ),
-    }))
-  );
+      // We use search parameters from the server request to craft the cache
+      // if possible, and fallback to those from results if not.
+      ...(requestParams || result.params
```
Haroenv, line 45: can we early return if `!(requestParams || result.params)` so we don't need the conditional in this map?

Haroenv (commented):
Thanks for tackling this! I think it can do with a couple more tests of more advanced features though

packages/instantsearch.js/src/lib/server.ts
dhayab: Yes, if it's `undefined` it will throw because Next won't serialize it as JSON.

The 2 options are either omitting the property or returning `null`. I chose the first option in this case.

packages/react-instantsearch-nextjs/src/InitializePromise.tsx
dhayab: It's similar to what's in server.ts for the pages router, just that because we have access to the search instance outside of both `waitForResults()` and `injectInitialResults()`, it allows me to put the code that extracts the queries params before and directly reference them instead of passing them as a return value of promises.

packages/instantsearch.js/src/lib/server.ts
dhayab: I can revert that you're right.

packages/react-instantsearch-nextjs/src/InitializePromise.tsx
dhayab: `params` is typed as an optional parameter

packages/instantsearch.js/src/lib/server.ts
dhayab: There's a follow-up below 😅

> Search parameters sent to the search client, usually returned by `waitForResults()`.

packages/react-instantsearch-nextjs/src/InitializePromise.tsx
Haroenv: my question was more about the potential desynchronising of the search client wrapper in the future if we do any changes to that. Not sure if it matters

packages/react-instantsearch-nextjs/src/InitializePromise.tsx
Haroenv: right, but when is it actually undefined? isn't it always there? is that to avoid a BC on internal API?

packages/instantsearch.js/src/lib/server.ts
Haroenv: I mean when is it not returned from waitForResults?

dhayab: Ah I get it, currently, never. I'll remove this word.

packages/instantsearch.js/src/lib/__tests__/server.test.ts
Haroenv: if you walk through the indices, you should be able to "skip back" to a previous result if you store the "request ids" in order right?


pseudo code for how that could work in the hydration of index widgets phase:

```js
let currentIndex = 0
let seenIndexIds: Record<IndexId, number> = {};
walkIndex(({indexName, indexId}) => {
  currentIndex = seenIndexIds[indexId] ?? currentIndex + 1;
  seenIndexIds[indexId] = currentIndex
})
```

dhayab: I looked at it and it's actually working as expected: there are as many `requestParams` as there are `walkIndex()` iterations.

The reason there's a mismatch later on is actually unrelated, `initialResults` is an object that is keyed by index ids, so the same index ids in later iterations overwrite previous ones.

I don't know if we should be more restrictive about having multiple identical index ids, but at least I can it knowingly and comment on the context.

Haroenv: Hmm that makes sense. I guess we can separately make sure initialResults contains the combined results / uiState, and not accidentally only those of one instance

packages/react-instantsearch-nextjs/src/InitializePromise.tsx
dhayab: This type is optional because you could send manual requests with the client, where queries don't require `params` to be set (https://www.algolia.com/doc/api-reference/api-methods/multiple-queries/#method-param-query-object).

In practice with InstantSearch, the helper always build the requests queries with params.

Haroenv: Fine to be extra wary I guess :)

Haroenv (approved):
I think this is good to go, however, in the future we need a good look at FX-2720 in detail (and probably also in the scope of FX-2713, maybe even warn or error for that use case as it may be broken)

Not fully sure if the code duplication around param extraction between React InstantSearch and InstantSearch.js is good long-term but ok I think

PR #5993, opened 2024-01-11
fix(RatingMenu/RefinementList): handle clicks in svg

Haroenv:
Handle the clicks in the RefinementList component (used by RatingMenu) when the clicked element isn't a `HTMLElement`.

The fixed behaviour is fairly finicky to reproduce, you need to click exactly on the "star image", if you click between them or on the text, no issue existed before this PR either.

The removed check was introduced in https://github.com/algolia/instantsearch/pull/4702, when translating the component to TypeScript.

This implements a barebones CTS for RatingMenu, including being fully skipped for React InstantSearch, as the widget doesn't exist.

### DIFF
--- a/packages/instantsearch.js/src/__tests__/common-widgets.test.tsx
+++ b/packages/instantsearch.js/src/__tests__/common-widgets.test.tsx
@@ -23,6 +23,7 @@ import {
   toggleRefinement,
   sortBy,
   stats,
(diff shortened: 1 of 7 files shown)


dhayab (approved):
I remember being frustrated once or twice with this, thanks for handling it!

PR #5995, opened 2024-01-12
feat(cisa): can now display images by picking an attribute

aymeric-giraudet:
**Summary**

### [FX-2610](https://algolia.atlassian.net/browse/FX-2610)

**Result**

- New `imageAttribute` option, takes a single string
  - From the CLI it can either be a root key, or something like `images.default` or `images[0]`
  - For the sake of simplicity, in interactive mode we only let them choose from root attributes with values that are strings without whitespace. Attribute keys containing `image` or `img` will be on top of the list.
  - We remove `imageAttribute` from `attributesToDisplay` if it was specified in CLI, although it wouldn't make sense as `attributesToDisplay` should be in `_highlightResults`
  - Templates have been updated to add a simple `img` tag with corresponding `src`, and `alt` is `attributesToDisplay[0]` as it's also mapped to a `h1` tag.
- Fixed a bug in the React InstantSearch template which was setting a `fallback` prop to `DynamicWidgets` instead of `fallbackComponent`

[FX-2610]: https://algolia.atlassian.net/browse/FX-2610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/create-instantsearch-app/src/cli/index.js
+++ b/packages/create-instantsearch-app/src/cli/index.js
@@ -24,6 +24,7 @@ const getAnswersDefaultValues = require('./getAnswersDefaultValues');
 const getAttributesFromIndex = require('./getAttributesFromIndex');
 const getConfiguration = require('./getConfiguration');
 const getFacetsFromIndex = require('./getFacetsFromIndex');
+const getPotentialImageAttributes = require('./getPotentialImageAttributes');
 const isQuestionAsked = require('./isQuestionAsked');
 const postProcessAnswers = require('./postProcessAnswers');
 
@@ -43,6 +44,10 @@ program
     '--attributes-to-display <attributesToDisplay>',
     'The attributes of your index to display in hits',
     splitArray
+  )
+  .option(
+    '--image-attribute <imageAttribute>',
+    'The attribute for image display in hits'
   )
   .option(
     '--attributes-for-faceting <attributesForFaceting>',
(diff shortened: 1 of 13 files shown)


packages/create-instantsearch-app/src/templates/JavaScript Client/src/app.js.hbs
```diff
@@ -17,6 +17,9 @@ function renderHits(query) {
               `<li class="ais-hits--item">
                 <article>
                 {{#if attributesToDisplay}}
+                  {{#if imageAttribute}}
+                  <img src="${ hit.{{imageAttribute}} }" alt="${ hit.{{attributesToDisplay.[0]}} }" />
```
aymeric-giraudet, line 21: I have to put whitespace between curly braces here and at other places else handlebars compilation will fail.
It's fine as the generated code will be formatted anyway

packages/create-instantsearch-app/src/cli/index.js
```diff
@@ -159,6 +164,36 @@ const getQuestions = ({ appName }) => ({
...
+      )}`,
+      pageSize: 10,
+      choices: async (answers) => [
+        {
+          name: 'None',
+          value: undefined,
+        },
+        new inquirer.Separator(),
+        new inquirer.Separator('From your index'),
+        ...(await getPotentialImageAttributes(answers)),
```
Haroenv, line 182: does inquirer have a "text input" as fallback possible or not?

packages/create-instantsearch-app/src/templates/JavaScript Client/src/app.js.hbs
Haroenv: I wonder if we could use a different character than `{`, but not sure if it's worth the confusing

packages/create-instantsearch-app/src/templates/React InstantSearch/src/App.tsx.hbs
```diff
@@ -54,7 +54,7 @@ export function App() {
           <div className="search-panel">
             <div className="search-panel__filters">
               {{#if flags.dynamicWidgets}}
-              <DynamicWidgets fallback={RefinementList}>
+              <DynamicWidgets fallbackComponent={RefinementList}>
```
Haroenv, line 57: unrelated, but thanks for that fix!

Haroenv (approved):

packages/create-instantsearch-app/src/cli/index.js
aymeric-giraudet: Not from what I've seen, we would need to have a `manual` option that would trigger another step but I'm not sure it's worth the trouble

dhayab (approved):
Really nice!

PR #5997, opened 2024-01-13
feat(createInsightsMiddleware): support `authenticatedUserToken`

sarahdayan:
This supports the new `authenticatedUserToken` from Search Insights.

When provided, it takes precedence over the `userToken`. When not, or when unset, the `userToken` is used.

> [!NOTE]  
> The code and tests are getting a bit more complicated than they could, but I didn't want to make a refactor in this PR to avoid complexifying the diffs. I'll do it in a separate one.

https://algolia.atlassian.net/browse/FX-2707

### DIFF
--- a/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
@@ -112,9 +112,13 @@ export function createInsightsMiddleware<
     );
 
     let queuedUserToken: string | undefined = undefined;
+    let queuedAuthenticatedUserToken: string | undefined = undefined;
     let userTokenBeforeInit: string | undefined = undefined;
+    let authenticatedUserTokenBeforeInit: string | undefined = undefined;
 
-    if (Array.isArray(insightsClient.queue)) {
+    const { queue } = insightsClient;
+
+    if (Array.isArray(queue)) {
       // Context: The umd build of search-insights is asynchronously loaded by the snippet.
       //
       // When user calls `aa('setUserToken', 'my-user-token')` before `search-insights` is loaded,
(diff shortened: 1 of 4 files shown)


packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -241,21 +254,64 @@ export function createInsightsMiddleware<
...
+          if (authenticatedUserToken) {
+            insightsClient('setAuthenticatedUserToken', authenticatedUserToken);
+          }
+        }
+
+        // We consider the `userToken` or `authenticatedUserToken` before an
+        // `init` call of higher importance than one from the queue.
+        const tokenBeforeInit =
+          authenticatedUserTokenBeforeInit || userTokenBeforeInit;
+        const queuedToken = queuedAuthenticatedUserToken || queuedUserToken;
```
sarahdayan, line 276: This all could be simpler but it can be tackled in a separate refactoring PR.

benhinchley (approved):

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -241,21 +254,64 @@ export function createInsightsMiddleware<
...
+          'onAuthenticatedUserTokenChange',
+          (authenticatedUserToken) => {
+            // If we're unsetting the `authenticatedUserToken`, we revert to the `userToken`
+            if (!authenticatedUserToken) {
+              insightsClient('getUserToken', null, (_, userToken) => {
+                setUserTokenToSearch(userToken);
+              });
+            }
+
+            setUserTokenToSearch(authenticatedUserToken);
```
dhayab, line 308: Could there be a race condition here where reverting to the `userToken` would be overridden by an undefined `authenticatedUserToken`?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -241,21 +254,64 @@ export function createInsightsMiddleware<
...
-        if (userTokenBeforeInit) {
-          setUserTokenToSearch(userTokenBeforeInit, true);
-          insightsClient('setUserToken', userTokenBeforeInit);
-        } else if (queuedUserToken) {
-          setUserTokenToSearch(queuedUserToken, true);
-          insightsClient('setUserToken', queuedUserToken);
+        function setUserToken(
+          token: string | number,
+          userToken?: string | number,
+          authenticatedUserToken?: string | number
```
Haroenv, line 260: you mentioned the refactor, but I'm not sure if it's a good idea to have two adjacent optional params. Shouldn't the first userToken be required (but allowed to be undefined)?

sarahdayan: Isn't it the same? If not passed, the `userToken ` will be `undefined`.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
dhayab: Ah yes, that's it.

dhayab (approved):

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: Yes, but there's no real use case for passing `authenticatedUserToken` if `userToken` isn't passed (you can't do it). If the signature is required (maybe even for both) we won't have weird edge cases if ever we call setUserToken without parameters

Haroenv (approved):
refactor can be later

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: Yep I think I see what you mean. This function is meant to be internal (and temporary) just to avoid repetition and further inflating bundle size. I can rewrite it more cleanly later.

PR #6004, opened 2024-01-17
fix(perf): remove usage of SearchParameters.clearRefinements

Haroenv:
This functions is much slower than individually clearing the specified refinements. In a test with DynamicWidgets + 118 RefinementLists the initial mount goes from 550ms to 350ms in my tests.

I've looked around to see if there's good tools for these improvements, to ensure we don't regress, but i couldn't find much (perf is mainly focused on apps).

Unfortunately I can't really deprecate the method as we still use it in a non-hot path inside the helper itself to create the right queries in queryBuilder. We'll just have to rely on "seeing other examples" to avoid using parameters.clearRefinements.

As widgets can't be mixed across the same attribute, there's no actual risk in this change as far as I can tell.

before|after
---|---
Screenshot 2024-01-17 at 11 25 07 | Screenshot 2024-01-17 at 11 25 16

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -10,7 +10,7 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "75.75 kB"
+      "maxSize": "76 kB"
     },
     {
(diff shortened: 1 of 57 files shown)


dhayab (approved):
This is great!

PR #6007, opened 2024-01-17
fix(requestBuilder): sort facet refinements in a non-mutating manner

Haroenv:
**Summary**

introduced in https://github.com/algolia/instantsearch/pull/5764, but wasn't an issue until merge/setQueryParameters doesn't make a new copy constantly of the parameters.

Essentially what's going on:
- when you build the requests, it's sorting the facet values
- this then gets updated in the parameters as well as it's the same object
- with this fix, the sorting is done non-mutating (similar to .toSorted) to avoid the sorting of the parameters for cache to have a potential impact on the UI

**Result**

tests that fail in #6006 are now passing. This commit can be reviewed separately, but really needs to be merged before.


dhayab (approved):
Good for me

PR #6009, opened 2024-01-18
fix(searchbox): ignore composition events with option

### LINKED ISSUE #5986: Samsung Note 20 Chrome has erratic behavior
### 🐛 Current behavior

When using Chrome on the Note 20 Ultra using the demo link provided in documentation it doesn't let you see results until you press enter. There's also an issue when pressing enter and changing the search query that it causes old results to display.

### 🔍 Steps to reproduce

Load chrome on a Note 20 (possibly other Samsung devices)
Pull up reproduction site
Start typing and you'll notice nothing displays
Hit enter and results appear
Delete results and type something else to see old results instead of new

### Live reproduction

https://codesandbox.io/p/sandbox/github/algolia/instantsearch/tree/master/examples/js/getting-started?file=%2Fsrc%2Fapp.js

### 💭 Expected behavior

The results to display properly as the data is typed how it performs in other mobile devices.

### Package version

4.63.0

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

dhayab:
**Summary**

We adjusted how `SearchBox` handles input events in #5963 to prevent sending unnecessary search requests during a [composition session](https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/isComposing), mostly occuring with people using [input method editors](https://developer.mozilla.org/en-US/docs/Glossary/Input_method_editor) to enter their search query.

After release, we received #5986 which was directly linked to the previous adjustment. During our tests, multiple Android target devices initiate composition sessions when typing, if a "predictive text" feature is enabled on a virtual keyboard with a latin character set. This breaks the  "search as you type" experience for no reason, as latin characters can be used as-is to refine search queries. This behaviour is not consistent with desktop usage and mobile usage on iOS.

After careful consideration, it seems there is no reliable way to correctly filter-out events within composition sessions in relevant character sets. Instead, this PR sets this behaviour behind a option on the `Searchbox` widget.

For example, in InstantSearch.js:
```js
searchBox({
  container: '#searchbox',
  ignoreCompositionEvents: true,
})
```

**Result**

- by default, all input events are handled by the `SearchBox` widget
- with `ignoreCompositionEvents: true`, input events within a composition session are ignored and won't trigger search requests

FX-2721
Fixes #5986

### DIFF
--- a/packages/react-instantsearch/src/widgets/SearchBox.tsx
+++ b/packages/react-instantsearch/src/widgets/SearchBox.tsx
@@ -28,12 +28,19 @@ export type SearchBoxProps = Omit<
      * @default true
      */
     searchAsYouType?: boolean;
+    /**
+     * Whether to update the search state in the middle of a
+     * composition session.
+     * @default false
+     */
+    ignoreCompositionEvents?: boolean;
     translations?: Partial<UiProps['translations']>;
   };
 
 export function SearchBox({
   queryHook,
   searchAsYouType = true,
+  ignoreCompositionEvents = false,
   translations,
   ...props
 }: SearchBoxProps) {
@@ -44,10 +51,10 @@ export function SearchBox({
   const [inputValue, setInputValue] = useState(query);
   const inputRef = useRef<HTMLInputElement>(null);
 
-  function setQuery(newQuery: string, compositionComplete = true) {
+  function setQuery(newQuery: string, isComposing = false) {
     setInputValue(newQuery);
 
-    if (searchAsYouType && compositionComplete) {
+    if (searchAsYouType && !(ignoreCompositionEvents && isComposing)) {
       refine(newQuery);
     }
   }
@@ -63,11 +70,10 @@ export function SearchBox({
   function onChange(
     event: Parameters<NonNullable<SearchBoxUiComponentProps['onChange']>>[0]
   ) {
-    const compositionComplete =
-      event.type === 'compositionend' ||
-      !(event.nativeEvent as KeyboardEvent).isComposing;
-
-    setQuery(event.currentTarget.value, compositionComplete);
+    setQuery(
+      event.currentTarget.value,
+      (event.nativeEvent as KeyboardEvent).isComposing
+    );
   }
 
   function onSubmit(event: React.FormEvent<HTMLFormElement>) {
--- a/packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
+++ b/packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
@@ -27,6 +27,7 @@ type SearchBoxProps = {
   refine?: (value: string) => void;
   autofocus?: boolean;
   searchAsYouType?: boolean;
+  ignoreCompositionEvents?: boolean;
   isSearchStalled?: boolean;
   disabled?: boolean;
   ariaLabel?: string;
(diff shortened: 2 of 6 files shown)


packages/react-instantsearch/src/widgets/SearchBox.tsx
```diff
@@ -44,10 +51,10 @@ export function SearchBox({
   const [inputValue, setInputValue] = useState(query);
   const inputRef = useRef<HTMLInputElement>(null);
 
-  function setQuery(newQuery: string, compositionComplete = true) {
+  function setQuery(newQuery: string, isComposing = false) {
     setInputValue(newQuery);
 
-    if (searchAsYouType && compositionComplete) {
+    if (searchAsYouType && !(ignoreCompositionEvents && isComposing)) {
```
aymeric-giraudet, line 57: Won't it actually refine when `ignoreCompositionEvents` is `false` here, no matter what the value of `isComposing` is ?
I feel like this is reversed ?

aymeric-giraudet: I feel like this should be `searchAsYouType && (ignoreCompositionEvents || !isComposing)`

packages/react-instantsearch/src/widgets/SearchBox.tsx
```diff
@@ -28,12 +28,19 @@ export type SearchBoxProps = Omit<
      * @default true
      */
     searchAsYouType?: boolean;
+    /**
+     * Whether to update the search state in the middle of a
+     * composition session.
+     * @default false
+     */
+    ignoreCompositionEvents?: boolean;
     translations?: Partial<UiProps['translations']>;
```
aymeric-giraudet, line 37: This describes what happens when this is `false` right ? I think this should be more clear as I got confused

packages/react-instantsearch/src/widgets/SearchBox.tsx
```diff
@@ -28,12 +28,19 @@ export type SearchBoxProps = Omit<
      * @default true
      */
     searchAsYouType?: boolean;
+    /**
+     * Whether to update the search state in the middle of a
+     * composition session.
+     * @default false
+     */
+    ignoreCompositionEvents?: boolean;
```
aymeric-giraudet, line 36: This describes what happens when this is `false` right ? I think this should be more clear as I got confused

aymeric-giraudet: I understood "ignoring composition events" as "I want to send a request no matter if it's composing or not" which seems to actually be the contrary

tats-u:
We should watch `compositionstart` to toggle the trigger event between `change` and `compositionend` without depending on the `isComposing` property in `KeyboardEvent`.

packages/react-instantsearch/src/widgets/SearchBox.tsx
dhayab: I get what you mean. It's hard not to get too verbose with the impact this option has. I had other names:
- `skipCompositionEvents` (but I guess the problem is similar)
- `updateQueryWithCompositionEvents` (really long but more meaningful?)

Do you have other ideas?

In the end, this option will mostly be surfaced through documentation and guides, so maybe we can go with an acceptable name instead of a more descriptive but really long one.

aymeric-giraudet: I prefer the second one as it's positive but we would have to change the logic :/
I guess we can keep it as is then, makes sense as it would probably be rarely set

aymeric-giraudet (approved):

PR #6024, opened 2024-01-30
feat(instantsearch-ui-components): initialize the package

sarahdayan:
## Summary

This initializes the `instantsearch-ui-components` package with the bare minimum.

https://algolia.atlassian.net/browse/FX-2744

## Notes

- I only export the ES module since this is designed for our usage only, at least for now.
- Not sure what version we want to start with, so I started with `0.1.0` ([why?](https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase)). Again, it's an internal package so we should be okay with breaking things as we need to, while sending the signal for anyone still willing to use it in their app that it's not stable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ jobs:
             - packages/react-instantsearch-nextjs/dist
             - packages/vue-instantsearch/vue2
             - packages/vue-instantsearch/vue3
-
+            - packages/instantsearch-ui-components/dist
   test metadata:
     <<: *defaults
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,6 +167,7 @@ jobs:
             - packages/vue-instantsearch/vue2
             - packages/vue-instantsearch/vue3
             - packages/instantsearch-ui-components/dist
+
   test metadata:
     <<: *defaults
     steps:
--- a/packages/instantsearch-ui-components/package.json
+++ b/packages/instantsearch-ui-components/package.json
@@ -9,7 +9,6 @@
   "type": "module",
   "exports": {
     "types": "./dist/es/index.d.ts",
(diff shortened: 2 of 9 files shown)


aymeric-giraudet (approved):
I think we might still need `cjs` for stuff like Remix, we'll test later and see how it goes

.circleci/config.yml
```diff
@@ -166,7 +166,7 @@ jobs:
             - packages/react-instantsearch-nextjs/dist
             - packages/vue-instantsearch/vue2
             - packages/vue-instantsearch/vue3
-
+            - packages/instantsearch-ui-components/dist
```
Haroenv, line 169:
```suggestion
            - packages/instantsearch-ui-components/dist

```

nbd but there was a separator here before

packages/instantsearch-ui-components/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "version": "0.1.0",
+  "description": "Common UI components for InstantSearch.",
+  "source": "src/index.ts",
+  "types": "dist/es/index.d.ts",
+  "main": "dist/es/index.js",
+  "module": "dist/es/index.js",
+  "type": "module",
+  "exports": {
+    "types": "./dist/es/index.d.ts",
+    "require": "./dist/cjs/index.js",
```
Haroenv, line 12: if we're not building cjs, we don't have this file right?

Haroenv:
```suggestion
```

Haroenv (approved):
two small changes should be done, good otherwise :)

packages/instantsearch-ui-components/package.json
sarahdayan: Yup, forgot to remove it!

PR #6029, opened 2024-02-01
feat(instantsearch-ui-components): introduce `Highlight` component

sarahdayan:
## Summary

This moves the `Highlight` component from `@algolia/ui-components` to the monorepo.

https://algolia.atlassian.net/browse/FX-2745

### DIFF
--- a/packages/instantsearch-ui-components/src/lib/cx.ts
+++ b/packages/instantsearch-ui-components/src/lib/cx.ts
@@ -1,0 +1,5 @@
+export function cx(
+  ...classNames: Array<string | number | boolean | undefined | null>
+) {
+  return classNames.filter(Boolean).join(' ');
+}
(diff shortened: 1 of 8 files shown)


packages/instantsearch-ui-components/src/lib/cx.ts
sarahdayan, line 1: Later on we can move this util into a `instantsearch-core` package. In the meantime we can centralize it here since all flavors (despite Angular) will use this package.

Haroenv: it makes sense in UI for me, that's where it is used directly

packages/instantsearch-ui-components/src/types/Renderer.ts
```diff
@@ -0,0 +1,76 @@
...
+declare global {
+  namespace JSX {
+    interface Element {}
+    interface IntrinsicElements {}
+  }
+}
+
+// For safety when there's no @types/react or preact, we don't directly use
+// JSX.IntrinsicElements
+type IntrinsicElements = keyof JSX.IntrinsicElements extends never
```
Haroenv, line 14: in the future we probably want to get rid of these types and make everything a bit more generic / less focused on react etc. For now I think it's generic enough as Vue isn't in typescript anyway

Haroenv (approved):

dhayab (approved):

PR #6036, opened 2024-02-05
fix(insights): use up-to-date index name when sending events

dhayab:
**Summary**

When enabling insights, the event payloads include the index name from which the data comes from. Currently, this index name is provided as a string and can be incorrect when the index changes, due to a `sortBy()` for example.

This PR implements changes to always retrieve the index name during the payload creation so it always matches the current state of InstantSearch.

**Result**

Events sent always include the correct index name.

CR-5442

### DIFF
--- a/packages/instantsearch.js/src/lib/utils/__tests__/createSendEventForHits-test.ts
+++ b/packages/instantsearch.js/src/lib/utils/__tests__/createSendEventForHits-test.ts
@@ -541,6 +541,8 @@ describe('createSendEventForHits', () => {
       })
     );
 
+    // This is what happens when a user changes the active index
+    // using sortBy() for example.
     helper.setIndex('index2');
 
     sendEvent('view', hits, 'Products Viewed');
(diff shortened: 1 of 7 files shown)


packages/instantsearch.js/src/lib/utils/__tests__/createSendEventForHits-test.ts
```diff
@@ -524,6 +528,29 @@ describe('createSendEventForHits', () => {
...
+      });
+
+    sendEvent('view', hits, 'Products Viewed');
+    expect(instantSearchInstance.sendEventToInsights).toHaveBeenLastCalledWith(
+      expect.objectContaining({
+        payload: expect.objectContaining({ index: 'index1' }),
+      })
+    );
+
+    helper.setIndex('index2');
```
Haroenv, line 544: This is fine, but a sortBy could be interesting too (or just a comment saying sortBy does this behind the hood), in case it doesn’t exactly call setIndex in a future

Haroenv (approved):

packages/instantsearch.js/src/lib/utils/__tests__/createSendEventForHits-test.ts
dhayab: I'll opt for a comment here if it's okay.

PR #6037, opened 2024-02-05
feat(instantsearch-ui-components): update build

sarahdayan:
## Summary

This modifies the build for `instantsearch-ui-components`:

- Includes runtime helpers as in the original `@algolia/ui-components`. This notably avoids issues with inlined helpers that currently cause our polyfill checker to fail.
- Includes a CJS bundle, which is necessary for the CJS bundles of each flavor.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -121,6 +121,19 @@ module.exports = (api) => {
           ],
         ],
       },
+      {
+        test: 'packages/instantsearch-ui-components',
+        plugins: [
+          [
+            '@babel/plugin-transform-runtime',
+            {
+              corejs: false,
+              helpers: true,
+              regenerator: true,
+            },
+          ],
+        ],
+      },
     ],
     // jsx is transpiled, so the comment should no longer be present in the final files
     shouldPrintComment: (value) => value !== '* @jsx h ',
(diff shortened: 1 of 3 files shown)


babel.config.js
```diff
@@ -121,6 +121,19 @@ module.exports = (api) => {
...
       },
+      {
+        test: 'packages/instantsearch-ui-components',
+        plugins: [
+          [
+            '@babel/plugin-transform-runtime',
+            {
+              corejs: false,
+              helpers: true,
+              regenerator: true,
```
Haroenv, line 132: I think we want false, no? We don’t use async/await and we don’t want regenerator in the bundle.

packages/instantsearch.js/scripts/rollup/rollup.config.js
```diff
@@ -23,6 +23,7 @@ const plugins = [
   }),
   babel({
     rootMode: 'upward',
+    runtimeHelpers: true,
```
Haroenv, line 26: Does this change the build, or is it only for node modules?

Haroenv (approved):

babel.config.js
sarahdayan: This is correct, it was in `@algolia/ui-components` which I copy/pasted as-is, but it isn't necessary. Removed in [9054a2b](https://github.com/algolia/instantsearch/pull/6037/commits/9054a2ba53a47b2b3ff29b70b647b175fbd242d1).

Haroenv (approved):

PR #6038, opened 2024-02-06
feat(react-instantsearch): replace `Highlight` and `cx`

sarahdayan:
## Summary

This replaces `Highlight` and `cx` in React InstantSearch.

https://algolia.atlassian.net/browse/FX-2746

## Notes

- The PR contains more changes than expected because of the `cx` replacement. That can be skimmed through.


dhayab (approved):

packages/react-instantsearch/src/ui/InternalHighlight.tsx
```diff
@@ -1,101 +1,9 @@
...
-              <span className={classNames.separator}>{separator}</span>
-            )}
-          </Fragment>
-        );
-      })}
-    </span>
-  );
-}
+export const InternalHighlight = createHighlightComponent({
+  createElement: createElement as Pragma,
```
Haroenv, line 7: does createElement not match Pragma? is Pragma typed wrong?

Haroenv (approved):

packages/react-instantsearch/src/ui/InternalHighlight.tsx
sarahdayan: It's a complex topic, because Preact and React types aren't fully compatible. I picked the types from `@algolia/ui-components` which were initially meant to be as generic as possible, but we never really spent time making sure they are.

IMO that's too big of a topic to be settled in this PR. I'd love to spend some time reinforcing our UI types so they're really generic, but type casting could do the job until we have time to spend on it, as there's sufficient overlap for TypeScript to be happy.

Haroenv: LGTM, in a future version we can indeed make Pragma more compatible / generic

PR #6041, opened 2024-02-06
feat(templates): allow Fragment as root

aymeric-giraudet:
**Summary**

[FX-2735](https://algolia.atlassian.net/browse/FX-2735)

**Result**

Settled with a `fragment` string after all, allowing a component is trickier to type because of `defaultProps`, and that's technically not what we want.

One problem is that since we still have Hogan templates and Fragment does not allow `dangerouslySetInnerHTML`, we need a workaround for still supporting them. 
I came up with a component that mounts an empty `div` and immediately sets `outerHTML` with the template's output. Of course this will no longer be necessary once we drop Hogan support.

[FX-2735]: https://algolia.atlassian.net/browse/FX-2735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/instantsearch.js/src/components/Template/__tests__/Template-test.tsx
```diff
@@ -56,6 +56,26 @@ describe('Template', () => {
...
   });
 
+  it('can have Fragment as rootTagName with string template', () => {
+    const props = getProps({
+      rootTagName: 'fragment',
+      templates: { test: '<span>test</span>' },
+    });
+    const wrapper = render(<Template {...props} />);
+
+    expect(wrapper.container).toMatchSnapshot();
```
Haroenv, line 66: good to have a regular snapshot as the other tests have that too and this file will eventually be simplified when Hogan is no longer there.

packages/instantsearch.js/src/components/Template/Template.tsx
```diff
@@ -21,7 +33,7 @@ const defaultProps = {
 export type TemplateProps = {
   data?: Record<string, any>;
   rootProps?: Record<string, any>;
-  rootTagName?: keyof JSX.IntrinsicElements;
+  rootTagName: keyof JSX.IntrinsicElements | 'fragment';
```
Haroenv, line 36:
```suggestion
  rootTagName: keyof JSX.IntrinsicElements | Fragment;
```

I can get this working if I change the defaultProps FYI. I don' think it matters much, but in case you want it:

```diff
diff --git a/packages/instantsearch.js/src/components/Template/Template.tsx b/packages/instantsearch.js/src/components/Template/Template.tsx
index 796dfd127..70f2531d9 100644
--- a/packages/instantsearch.js/src/components/Template/Template.tsx
+++ b/packages/instantsearch.js/src/components/Template/Template.tsx
@@ -22,9 +22,11 @@ class RawHtml extends Component<{ content: string }> {
   }
 }
 
+type TagName = keyof JSX.IntrinsicElements | typeof Fragment;
+
 const defaultProps = {
   data: {},
-  rootTagName: 'div',
+  rootTagName: 'div' as TagName,
   useCustomCompileOptions: {},
   templates: {},
   templatesConfig: {},
@@ -33,7 +35,7 @@ const defaultProps = {
 export type TemplateProps = {
   data?: Record<string, any>;
   rootProps?: Record<string, any>;
-  rootTagName: keyof JSX.IntrinsicElements | 'fragment';
+  rootTagName: TagName;
   templateKey: string;
   bindEvent?: BindEventForHits;
   sendEvent?: SendEventForHits;
@@ -69,8 +71,7 @@ See: https://www.algolia.com/doc/guides/building-search-ui/upgrade-guides/js/#up
       );
     }
 
-    const RootTagName =
-      this.props.rootTagName === 'fragment' ? Fragment : this.props.rootTagName;
+    const RootTagName = this.props.rootTagName;
 
     const useCustomCompileOptions =
       this.props.useCustomCompileOptions[this.props.templateKey];
diff --git a/packages/instantsearch.js/src/components/Template/__tests__/Template-test.tsx b/packages/instantsearch.js/src/components/Template/__tests__/Template-test.tsx
index ef3b246c5..f6c4603f0 100644
--- a/packages/instantsearch.js/src/components/Template/__tests__/Template-test.tsx
+++ b/packages/instantsearch.js/src/components/Template/__tests__/Template-test.tsx
@@ -5,7 +5,7 @@
 
 import { mount, shallow } from '@instantsearch/testutils/enzyme';
 import { render } from '@testing-library/preact';
-import { h } from 'preact';
+import { Fragment, h } from 'preact';
 
 import { warning } from '../../../lib/utils';
 import Template from '../Template';
@@ -58,7 +58,7 @@ describe('Template', () => {
 
   it('can have Fragment as rootTagName with string template', () => {
     const props = getProps({
-      rootTagName: 'fragment',
+      rootTagName: Fragment,
       templates: { test: '<span>test</span>' },
     });
     const wrapper = render(<Template {...props} />);
@@ -68,7 +68,7 @@ describe('Template', () => {
 
   it('can have Fragment as rootTagName with Preact template', () => {
     const props = getProps({
-      rootTagName: 'fragment',
+      rootTagName: Fragment,
       templates: { test: () => <span>test</span> },
     });
     const wrapper = render(<Template {...props} />);
```

Haroenv (approved):

packages/instantsearch.js/src/components/Template/Template.tsx
```diff
@@ -85,6 +98,11 @@ See: https://www.algolia.com/doc/guides/building-search-ui/upgrade-guides/js/#up
       return <RootTagName {...this.props.rootProps}>{content}</RootTagName>;
     }
 
+    // This is to handle Hogan templates with Fragment as rootTagName
+    if (RootTagName === Fragment) {
+      return <RawHtml content={content} />;
+    }
```
dhayab, line 104: Should this be also conditioned by `nonFunctionTemplates` (which would need to be hoisted one level up in that case)? I think we could just let `RootTagName` be `Fragment` with html templating.

Haroenv: Isn't that because Fragment can't have dangerousInnerHTML as in the description? It's already conditioned to be only strings though, as right before it's listed as not being an object.

Simple strings now fall into this case but that's probably ok?

dhayab: Ah well that's it, and it was already mentioned in the description. Thanks for the clarification!

packages/instantsearch.js/src/components/Template/Template.tsx
aymeric-giraudet: Ah yes I could have just narrowed it indeed thanks ! Was thinking of more radical solutions haha
But ultimately when we drop support for Hogan the whole `Template` component should go away, so it's probably best not to have to import `Fragment` in files that later won't need it

dhayab (approved):

packages/instantsearch.js/src/components/Template/Template.tsx
aymeric-giraudet: Ah yes cause we allow functions to return strings as well, although those strings should not contain HTML.

packages/instantsearch.js/src/components/Template/Template.tsx
```diff
@@ -10,6 +10,28 @@ import type { BindEventForHits, SendEventForHits } from '../../lib/utils';
...
+class RawHtml extends Component<{ content: string }> {
+  ref = createRef();
+  nodes: Element[] = [];
+
+  componentDidMount() {
+    const fragment = new DocumentFragment();
+    const root = document.createElement('div');
+    root.innerHTML = this.props.content;
+    this.nodes = [...root.children];
+    fragment.append(...this.nodes);
```
Haroenv, line 22: appendNode is IE11 compatible if I remember it correctly, not append

packages/instantsearch.js/src/components/Template/Template.tsx
```diff
@@ -10,6 +10,28 @@ import type { BindEventForHits, SendEventForHits } from '../../lib/utils';
...
+    this.ref.current.replaceWith(fragment);
+  }
+
```
Haroenv, line 25: do we not need a componentWill/HaveReceiveProps in case content changes?

Haroenv (changes requested):
don't know if IE11 is compatible

packages/instantsearch.js/src/components/Template/Template.tsx
aymeric-giraudet: There is but we still need to unmount anyway, not sure if it's worth spending so much time on something caused by tech debt.
Only win would be slightly better performance for those who still use string templates which we discourage 🤷‍♂️

packages/instantsearch.js/src/components/Template/Template.tsx
aymeric-giraudet: Ah yes it's `appendChild`, I'll need to use `forEach` too

Would be nice to settle on IE11 for once, as SFCC did already

packages/instantsearch.js/src/components/Template/Template.tsx
Haroenv: makes sense, was just not sure if it's a bug or if it properly updates :)

PR #6042, opened 2024-02-07
feat(ui-components): migrate `Hits` component

dhayab:
This PR introduces `createHits()`, available from `instantsearch-ui-components`. This method can be used by the Hits widget from each flavor to setup a UI component that renders appropriately.

## Design considerations

The UI component expects from widgets the following props:

```tsx
export type HitsProps<THit> = React.ComponentProps<'div'> & {
  hits: THit[];
  itemComponent: (props: {
    hit: THit;
    index: number;
    className: string;
    onClick: () => void;
    onAuxClick: () => void;
  }) => JSX.Element;
  sendEvent: SendEventForHits;
  classNames?: Partial<HitsClassNames>;
  emptyComponent?: (props: { className: string }) => JSX.Element;
};
```

**`itemComponent`**

Used by the component as a direct child of an `<ol>`. Widgets use this to wrap the item template / hit component / item slot.

**`classNames`**

The class list is from IS.js/RISH. Vue's component maps its specific structure to the more common one.

**`emptyComponent`**

When provided, the component renders this instead of returning an empty list if there are no hits. Only IS.js has this currently.

FX-2729


packages/instantsearch-ui-components/src/components/Hits.tsx
```diff
@@ -1,9 +1,10 @@
 /** @jsx createElement */
 import { cx } from '../lib';
 
+// These will need to be imported from a shared package in the future
```
Haroenv, line 4:
```mermaid
flowchart TD
  types --> instantsearch-core
  connectors --> instantsearch-core
  index --> instantsearch-core
  instantsearch --> instantsearch-core
  routers? --> instantsearch-core
  components --> js
  components --> react
  components --> vue
  instantsearch-core --> components
  instantsearch-core --> js
  instantsearch-core --> react
  instantsearch-core --> vue
```

This is probably the right order indeed. We can start with moving only types, then move connectors + other things after in a non-breaking way (although we need to make sure to re-export the types from the place they are right now, otherwise it's a BC)

dhayab: I've pushed an alternative with alternative simplified types that don't require to be imported from instantsearch.js. In Hits' case I don't see an issue, as they're not exposed to contributors.

Haroenv: Yes, as we're not calling it particularly much, it's ok, "components" accepts a lot, but we have precise types in the usage 🐐

packages/instantsearch-ui-components/src/components/__tests__/Hits.test.tsx
```diff
@@ -41,85 +45,27 @@ describe('Hits', () => {
...
-              </strong>
-            </li>
-          </ol>
-        </div>
-      </div>
-    `);
-  });
-
-  test('passes sendEvent to hitComponent', () => {
+  test('renders with custom div props', () => {
```
sarahdayan, line 61: Just for the sake of consistent test names, no biggie.

```suggestion
  test('forwards `div` props to the root element', () => {
```

packages/instantsearch.js/src/widgets/hits/__tests__/hits.test.tsx
```diff
@@ -741,6 +741,41 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/"
       `);
     });
 
+    test('provides __hitIndex in the list to each item', async () => {
```
sarahdayan, line 744: Why is this important to have this test?

tests/common/widgets/hits/options.ts
```diff
@@ -126,6 +130,87 @@ export function createOptionsTests(
       `
       );
     });
+
+    skippableDescribe('instantsearch.js', skippedTests, () => {
```
sarahdayan, line 134: Right now we have two ways of writing these flavor-specific tests:
1. Have dedicated test files per flavor (e.g., `packages/instantsearch.js/src/widgets/hits/__tests__/hits.test.tsx`)
2. Use `skippableDescribe`

Should we pick one approach?

packages/instantsearch-ui-components/src/components/Hits.tsx
```diff
@@ -0,0 +1,95 @@
...
+};
+
+export function createHits({ createElement }: Renderer) {
+  return <THit extends Hit>({
+    classNames = {},
+    hits,
+    itemComponent: ItemComponent,
+    sendEvent,
+    emptyComponent: EmptyComponent,
+    ...props
```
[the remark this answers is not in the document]
Haroenv: I think I tried finding an eslint rule and it didn't work. Will try again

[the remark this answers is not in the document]
Haroenv: What's missing is this test: https://github.com/algolia/ui-components/pull/29/files we don't have them here I believe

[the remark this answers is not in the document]
dhayab: I see, I'll destructure in the function body now, and I can add these tests in a separate PR.

packages/instantsearch-ui-components/src/components/Hits.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }: HitsProps<THit>) => {
+    if (hits.length === 0 && EmptyComponent) {
+      return (
+        <EmptyComponent
+          className={cx(
+            'ais-Hits',
```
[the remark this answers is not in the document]
dhayab: It made sense for me to have `cx` compute the class list at the UI component level, the alternative being having the same logic for each flavor.

packages/instantsearch.js/src/widgets/hits/__tests__/hits.test.tsx
dhayab: This ensures we keep track of an InstantSearch.js specific change in the hit data provided to the item template. It previously was tested in https://github.com/algolia/instantsearch/blob/5edf56bcadaccbc6d91f465ad710385ce869b3de/packages/instantsearch.js/src/components/Hits/__tests__/Hits-test.tsx#L186-L213 and I moved this to here as I removed obsolete test files.

tests/common/widgets/hits/options.ts
dhayab: I wasn't sure I would still need flavor specific widget tests files before, so I tried this with skippableDescribe, but forgot to get back to those later. I'll move the one for InstantSearch.js into its flavor test file and keep the skippable for the others as it's the more common one (to help remind us we'll need to harmonize the behavior in the future major of IS.js).

packages/instantsearch-ui-components/src/components/Hits.tsx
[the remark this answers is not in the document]
Haroenv: https://github.com/algolia/instantsearch/pull/6054 🎁

Haroenv (commented):
👍

packages/instantsearch-ui-components/src/components/Hits.tsx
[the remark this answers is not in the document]
sarahdayan: I'm fine with having the base AIS class names in the UI component, I'll change it for `Highlight`.

bundlesize.config.json
```diff
@@ -10,27 +10,27 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "76.25 kB"
+      "maxSize": "76.5 kB"
```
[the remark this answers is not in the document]
dhayab: The Hits UI component has to let widgets provide it with flavor specific components props. It contributes to not quite reducing the impact on this change for now. It will surely reduce the bundle size when those differences are no more.

To illustrate this, here is the main diff for React InstantSearch's UMD bundle (not minified):

CleanShot 2024-02-14 at 17 13 16@2x
CleanShot 2024-02-14 at 17 12 04@2x

packages/instantsearch-ui-components/src/components/Hits.tsx
```diff
@@ -0,0 +1,95 @@
...
+   * Class names to apply to each item element
+   */
+  item: string;
+};
+
+export function createHits({ createElement }: Renderer) {
+  return <THit extends Hit>({
+    classNames = {},
+    hits,
+    itemComponent: ItemComponent,
```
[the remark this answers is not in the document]
dhayab: I wanted to prevent confusing it with what we understand as hitComponent. This one wraps it with the outer `<li>`.

packages/instantsearch.js/src/connectors/hits/__tests__/connectHits-test.ts
```diff
@@ -493,8 +493,8 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
     it('returns the render state', () => {
       const renderFn = jest.fn();
       const unmountFn = jest.fn();
-      const createHits = connectHits(renderFn, unmountFn);
-      const hitsWidget = createHits({});
+      const createHitsComponent = connectHits(renderFn, unmountFn);
+      const hitsWidget = createHitsComponent({});
```
Haroenv, line 497:
```suggestion
      const createHits = connectHits(renderFn, unmountFn);
      const hitsWidget = createHits({});
```

packages/instantsearch.js/src/connectors/hits/__tests__/connectHits-test.ts
```diff
@@ -549,8 +549,8 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/hits/js/#co
     it('returns the widget render state', () => {
       const renderFn = jest.fn();
       const unmountFn = jest.fn();
-      const createHits = connectHits(renderFn, unmountFn);
-      const hitsWidget = createHits({});
+      const createHitsComponent = connectHits(renderFn, unmountFn);
+      const hitsWidget = createHitsComponent({});
```
Haroenv, line 553:
```suggestion
      const createHits = connectHits(renderFn, unmountFn);
      const hitsWidget = createHits({});
```

packages/instantsearch.js/src/connectors/hits/__tests__/connectHits-test.ts
Haroenv: this change was automated by accident

packages/vue-instantsearch/src/components/Hits.js
```diff
@@ -0,0 +1,113 @@
...
+
+    return h(createHitsComponent({ createElement: h }), {
+      hits: this.state.hits,
+      itemComponent,
+      sendEvent: this.state.sendEvent,
+      classNames: this.classNames && {
+        root: this.classNames['ais-Hits'],
+        list: this.classNames['ais-Hits-list'],
+        item: this.classNames['ais-Hits-item'],
+      },
```
Haroenv, line 110: good aliasing here, but we definitely should have only one shape in next major :)

Haroenv (approved):

PR #6046, opened 2024-02-12
docs(examples): use htm in examples templates

sarahdayan:
## Summary

This updates all examples to use function templates, with `html` when necessary.

We're still using string HTML and deprecated APIs when using connectors (e.g., in `example-instantsearch-media`), which we can do another pass on later on as we might be able to use templates and simplify the examples instead of pulling Preact directly.


examples/js/media/src/widgets/Dates.ts
```diff
@@ -1,8 +1,9 @@
 import { numericMenu, panel } from 'instantsearch.js/es/widgets';
 
 import { DATE_RANGES } from '../utils';
+import { PanelTemplates } from 'instantsearch.js/es/widgets/panel/panel';
```
Haroenv, line 4:
```suggestion
import { numericMenu, panel } from 'instantsearch.js/es/widgets';
import { PanelTemplates } from 'instantsearch.js/es/widgets/panel/panel';

import { DATE_RANGES } from '../utils';
```

examples/js/e-commerce-umd/src/templates/panel.ts
```diff
@@ -1,12 +1,16 @@
...
-  <path fill="#21243d" d="M0 6h13v1H0z" fill-rule="evenodd"/>
-</svg>
-  `;
+import { PanelTemplates } from 'instantsearch.js/es/widgets/panel/panel';
+
+export const collapseButtonText: PanelTemplates<any>['collapseButtonText'] = (
+  { collapsed },
+  { html }
+) => {
+  return html`<svg xmlns="http://www.w3.org/2000/svg" width="13" height="13">
```
Haroenv, line 7: maybe cleaner with the conditional outside as it was, but fine either way

examples/js/e-commerce-umd/src/widgets/ResultsNumberMobile.ts
```diff
@@ -3,6 +3,8 @@ const { stats } = window.instantsearch.widgets;
 export const resultsNumberMobile = stats({
   container: '[data-widget="results-number-mobile"]',
   templates: {
-    text: '<strong>{{#helpers.formatNumber}}{{nbHits}}{{/helpers.formatNumber}}</strong> results',
+    text({ nbHits }, { html }) {
+      return html` <strong>${nbHits.toLocaleString()}</strong> results`;
```
Haroenv, line 7:
```suggestion
      return html`<strong>${nbHits.toLocaleString()}</strong> results`;
```

examples/js/e-commerce/src/templates/panel.ts
```diff
@@ -1,12 +1,16 @@
...
-  <path fill="#21243d" d="M0 6h13v1H0z" fill-rule="evenodd"/>
-</svg>
-  `;
+import { PanelTemplates } from 'instantsearch.js/es/widgets/panel/panel';
+
+export const collapseButtonText: PanelTemplates<any>['collapseButtonText'] = (
+  { collapsed },
+  { html }
+) => {
+  return html`<svg xmlns="http://www.w3.org/2000/svg" width="13" height="13">
```
Haroenv, line 7: same, cleaner with two full templates IMO

examples/js/media/src/widgets/Dates.ts
```diff
@@ -1,8 +1,9 @@
...
 
-const createDatesList = (header: string) =>
+const createDatesList = (header: PanelTemplates<any>['header']) =>
```
Haroenv, line 6:
```suggestion
const createDatesList = (header: PanelTemplates<typeof numericMenu>['header']) =>
```

not sure if that's correct, but I believe the type can be found

examples/js/media/src/widgets/Dates.ts
```diff
@@ -14,7 +15,7 @@ export const createDates = ({
   header,
 }: {
   container: string;
-  header: string;
+  header: PanelTemplates<any>['header'];
```
Haroenv, line 18: same here?

Haroenv (approved):
great catch!

Haroenv (approved):

examples/js/e-commerce-umd/src/widgets/Brands.ts
```diff
@@ -3,7 +3,9 @@ import { collapseButtonText } from '../templates/panel';
 const { panel, refinementList } = window.instantsearch.widgets;
 const brandRefinementList = panel({
```
Haroenv, line 4:
```suggestion
const brandRefinementList = panel<typeof refinementList>({
```

I think it needs something similar to this for the types to work? not sure though

PR #6054, opened 2024-02-14
test(types): ensure correct signature

Haroenv:
**Summary**

Add a test and fix the signature of createHitsComponent. The test ensures:

- export of ui component is a function
- function has correct name
- type of the return is a function
- return type has only one parameter, which is just a name

**Result**

- No reproduction of https://github.com/algolia/ui-components/pull/29 in this code base
- test based on https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#traversing-the-ast-with-a-little-linter (I think this could be eslint as well, but that was harder to understand. We can migrate later with basically the same logic, or continue using this TS test/linter as it is based on the output, not the input)

### DIFF
--- a/packages/instantsearch-ui-components/__tests__/types.test.ts
+++ b/packages/instantsearch-ui-components/__tests__/types.test.ts
@@ -1,0 +1,117 @@
+import fs from 'fs';
+import path from 'path';
+
+import ts from 'typescript';
+
+/**
+ * Simple linting using the TypeScript compiler API.
+ */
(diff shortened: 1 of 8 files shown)


dhayab (approved):
This is great! Should we hold on until Hits is merged to resolve conflicts or is it okay to merge this into the feature branch right away (but still resolving current conflicts)?

Haroenv:
Resolved the commits, good to merge in feature branch IMO

PR #6055, opened 2024-02-15
chore: deduplicate Search client dependencies

sarahdayan:
## Summary

This deduplicates the Search client dependencies so they all point to the latest version.


Haroenv (approved):
(we really should migrate devDependencies to ^ some time, with the lock file I think it's better for deduplicating info)

Haroenv (changes requested):
actually it seems types have failed with latest, probably something to do with https://github.com/algolia/algoliasearch-client-javascript/pull/1460

sarahdayan:
> actually it seems types have failed with latest, probably something to do with [algolia/algoliasearch-client-javascript#1460](https://github.com/algolia/algoliasearch-client-javascript/pull/1460)

This is because of the response type of `response.results` but it only affects test files so I think we can type cast here.

packages/instantsearch.js/src/connectors/dynamic-widgets/__tests__/connectDynamicWidgets-test.ts
```diff
@@ -21,6 +21,7 @@ import connectHierarchicalMenu from '../../hierarchical-menu/connectHierarchical
 import connectRefinementList from '../../refinement-list/connectRefinementList';
 
 import type { DynamicWidgetsConnectorParams } from '../connectDynamicWidgets';
+import type { SearchResponse } from '@algolia/client-search';
```
Haroenv, line 24: you want to import from `../types/algoliasearch` probably

Haroenv (approved):

dhayab (approved):

PR #6059, opened 2024-02-16
fix(helper): correctly set isRefined for hierarchical facet values with trailing spaces

dhayab:
**Summary**

Hierarchical facet values that have a trailing space are not correctly flagged with `isRefined: true` during the generation of the hierarchical tree. This is because its path part is compared to its name, which is trimmed by another part of the code.

This PR ensures the part is also trimmed when comparing it to the name.

**Result**

Hierarchical facet values that are refined are marked as so by the returned data from `getFacetValues()`.

Derived from CR-5241


aymeric-giraudet (approved):

Haroenv:
How does this happen? can it happen for non-hierarchical facets too? seems a little odd for the place of the normalisation of the facet value to be.

dhayab:
> How does this happen? can it happen for non-hierarchical facets too? seems a little odd for the place of the normalisation of the facet value to be.

This happens because the deep equality check fails, as the current refinement part is compared against a reference that has been trimmed here: https://github.com/algolia/instantsearch/blob/master/packages/algoliasearch-helper/src/SearchResults/generate-hierarchical-tree.js#L192.

Conjunctive / disjunctive facets go through another method to set their `isRefined` flag which works as expected.

I can't see anywhere else to normalise the variable for this check, this is very specific to the hierarchical menu generation.

Haroenv:
Thanks, but I meant: how do you get an untrimmed facet value?

dhayab:
Ah! This was discovered in a customer's dataset, one of their facet values had a trailing space.

dhayab:
@Haroenv are you satisfied with the changes, can I go ahead and merge it?

Haroenv (approved):
if there's a valid reason for this to exist, why not, but I would overall suggest to fix indexing first

PR #6061, opened 2024-02-16
fix(examples): remove unused package

syed-yawar:
**Summary**

Remove `react-instantsearch-router-nextjs` from **package.json** to avoid confusion for users and developers, as it's not utilized in this context.

**Result**
This package usage in the example won't cause confusion for users and developers.


Haroenv (approved):
thanks! this was indeed copied wrong from the other example

PR #6069, opened 2024-02-21
chore(examples): add react types dev dependencies

dhayab:
**Summary**

This PR adds the `@types/react` dev dependency on relevant React InstantSearch examples so Codesandbox properly interprets JSX markup in the editor.


Haroenv (approved):
we don't need react-dom or others?

dhayab:
I do remember in the past (React 17 and before) having to add `@types/react-dom`, but JSX elements are directly available from `@types/react@18`.

PR #6071, opened 2024-02-22
fix(build): do not use babel runtime helpers in ui components

### LINKED ISSUE #6060: Rollup build error after upgrade to 4.65.0
### 🐛 Current behavior

After npm upgrade revved instantsearch.js from 4.64.2 to 4.65.0

Using:
```
import algoliasearch from 'algoliasearch/lite';
import { connectStats } from 'instantsearch.js/es/connectors';
import { searchBox, hits, stats, index } from 'instantsearch.js/es/widgets';
```

Build fails with error:

```
[vite]: Rollup failed to resolve import "@babel/runtime/helpers/extends" from "/home/forge/example.com/node_modules/instantsearch-ui-components/dist/es/components/Highlight.js".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
error during build:
Error: [vite]: Rollup failed to resolve import "@babel/runtime/helpers/extends" from "/home/forge/example.com/node_modules/instantsearch-ui-components/dist/es/components/Highlight.js".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
    at viteWarn (file:///home/forge/example.com/node_modules/vite/dist/node/chunks/dep-52909643.js:48216:27)
    at onRollupWarning (file:///home/forge/example.com/node_modules/vite/dist/node/chunks/dep-52909643.js:48248:9)
    at onwarn (file:///home/forge/example.com/node_modules/vite/dist/node/chunks/dep-52909643.js:47976:13)
    at file:///home/forge/example.com/node_modules/rollup/dist/es/shared/node-entry.js:24276:13
    at Object.logger [as onLog] (file:///home/forge/example.com/node_modules/rollup/dist/es/shared/node-entry.js:25950:9)
    at ModuleLoader.handleInvalidResolvedId (file:///home/forge/example.com/node_modules/rollup/dist/es/shared/node-entry.js:24862:26)
    at file:///home/forge/example.com/node_modules/rollup/dist/es/shared/node-entry.js:24822:26
```

### 🔍 Steps to reproduce

npm run build

### Live reproduction

server build

### 💭 Expected behavior

build success

### Package version

instantsearch.js 4.65.0

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

dhayab:
**Summary**

The **instantsearch-ui-components** package is currently transformed during build to rely on babel runtime helpers (ie:  `extends` and `objectWithoutProperties`).

This is problematic as we don't set **@babel/runtime** as a dependency in instantsearch.js, which causes issues either during build with module bundlers, or at runtime with UMD.

This PR stops relying on this package and insteads reverts to inline these helper methods in each relevant component file.

> [!NOTE]
> A side-effect of this, is because `_objectWithoutProperties` calls `Object.assign()`, it needs to be excluded from the polyfill check we run. But this should pose no problem, we already recommend using polyfills for IE11 [in our documentation](https://www.algolia.com/doc/guides/building-search-ui/installation/js/#supported-browsers).

**Result**

InstantSearch.js does not trigger an error during build because of a missing **@babel/runtime** dependency.

Fixes #6060

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -75,6 +75,7 @@ module.exports = (api) => {
 
             // we require polyfills for this already
             'Array.prototype.includes',
+            'Object.assign',
 
             // false positive (babel doesn't know types)
             // this is actually only called on arrays
@@ -117,19 +118,6 @@ module.exports = (api) => {
               helpers: true,
               regenerator: false,
               useESModules: isES || isRollup,
-            },
-          ],
-        ],
-      },
-      {
-        test: 'packages/instantsearch-ui-components',
-        plugins: [
-          [
-            '@babel/plugin-transform-runtime',
-            {
-              corejs: false,
-              helpers: true,
-              regenerator: false,
             },
           ],
         ],


babel.config.js
```diff
@@ -75,6 +75,7 @@ module.exports = (api) => {
 
             // we require polyfills for this already
             'Array.prototype.includes',
+            'Object.assign',
```
Haroenv, line 78: we should still warn for this though, using Object.assign non-polyfilled is problematic (but we catch it with IE11 e2e tests I guess)

Haroenv (approved):
overall approve, but really object.assign should not be allowed in the polyfill check

dhayab:
Closing this in favor of https://github.com/algolia/instantsearch/pull/6072

PR #6076, opened 2024-02-26
chore(examples): use local polyfills when possible

dhayab:
**Summary**

For our e2e tests we add polyfills in e-commerce examples, to include unsupported methods in IE11. This PR replaces remote polyfill links with local scripts.

> [!IMPORTANT]
> With this PR, IE11 testing is only performed on InstantSearch.js (`js` and `js-umd`). This change is temporary.

→ [Passing CI](https://app.circleci.com/pipelines/github/algolia/instantsearch/10297/workflows/ac0689f9-16c5-43a9-983d-e6bc21d0f4c8)


examples/polyfills.js
```diff
@@ -0,0 +1,5529 @@
...
+ * - Element.prototype.before, License: CC0
+ * - Element.prototype.prepend, License: CC0
+ * - Element.prototype.remove, License: CC0
+ * - Element.prototype.replaceWith, License: CC0
+*/
+
+(function(self, undefined) {
+  if (typeof Object.entries !== 'undefined') {
+    return;
+  }
```
dhayab, line 91: This simple early return prevents executing the polyfills on supported browsers (I assume all browsers we support but IE11 handle `Object.entries`).

examples/vue/e-commerce/public/index.html
```diff
@@ -20,7 +20,7 @@
       href="https://cdn.jsdelivr.net/npm/instantsearch.css@7.3.1/themes/reset-min.css"
     />
 
-    <script src="https://polyfill.io/v3/polyfill.min.js?features=default%2CArray.prototype.find%2CArray.prototype.includes%2CPromise%2CObject.assign%2CObject.entries"></script>
+    <script src="https://polyfill.io/v3/polyfill.min.js?version=3.111.0&features=default%2CArray.prototype.find%2CArray.prototype.includes%2CArray.prototype.findIndex%2CPromise%2CObject.assign%2CObject.entries"></script>
```
dhayab, line 23: Latest polyfill.io doesn't work, so we need to pin a specific version. It also seems that with this version we need to explicitly add the `Array.prototype.findIndex` polyfill.

examples/js/e-commerce/index.html
```diff
@@ -25,7 +25,7 @@
     <link rel="stylesheet" href="./src/app.css" />
     <link rel="stylesheet" href="./src/app.mobile.css" />
     <link rel="stylesheet" href="./src/widgets/PriceSlider.css" />
-    <script src="https://polyfill.io/v3/polyfill.min.js?features=default%2CArray.prototype.find%2CArray.prototype.includes%2CPromise%2CObject.assign%2CObject.entries"></script>
+    <script src="../../polyfills.js"></script>
```
Haroenv, line 28: wouldn't it be better if the file was local so it won't error on codesandbox?

dhayab:
Redrafting the PR, since this morning even the pinned version of the polyfill.io URL fails.

examples/js/e-commerce/index.html
dhayab: Oh I forgot they are also on CSB. I'll see if I can just symlink the same source file or if I just have to hard copy it.

Haroenv: hard copy is probably easiest. Not sure if symlink will work in csb. This file shouldn't change anyway

tests/e2e/flavors.js
```diff
@@ -1,7 +1,24 @@
+const ALL_FLAVORS = ['js', 'js-umd', 'react', 'vue'];
+
 module.exports = {
-  flavors: ['js', 'js-umd', 'react', 'vue'].filter(
```
Haroenv, line 2: there's a small chance this is used in jsdoc type somewhere, do you mind checking? not sure if I changed it to string at a later point

Haroenv (approved):

tests/e2e/flavors.js
dhayab: I looked at the current codebase and commits in the change history of this file, I could not find anything.

PR #6079, opened 2024-02-27
poc(helper): recommend helper-style

aymeric-giraudet:
Following concerns about having search helper users needing to import an extra package in order to use recommend, and leaking backend implementation details in our API, I made this PoC showing how much changes would be needed to wrap recommend in the helper.

This assumes the client has the `getRecommendations` method. Else we would either need to break the constructor, or add an optional argument for the recommendClient last.

I tried as much as possible to match the current API, but putting everything in existing events and/or methods would need much more logic and be less performant.

The only difference for the user would be that **in addition to being able to do** `const result = await helper.getClient().getRecommendations()` they could do : 

```js
helper.addFrequentlyBoughtTogether().recommend();
helper.on('recommendResult', (result) => /* ... */);
```

This actually accentuates implementation details even more.


packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1389,6 +1427,81 @@ AlgoliaSearchHelper.prototype._search = function (options) {
...
+  this._currentNbRecommendQueries++;
+
+  try {
+    this.client
+      .getRecommendations(queries)
+      .then(this._dispatchRecommendResponse.bind(this, queryId))
+      .catch(this._dispatchRecommendError.bind(this, queryId));
+  } catch (error) {
+    // If we reach this part, we're in an internal error state
+    this.emit('recommendError', {
```
Haroenv, line 1444:
```suggestion
    this.emit('recommend:error', {
```

this is possible I believe

PR #6095, opened 2024-03-19
fix(vue-instantsearch): properly add `noRefinement` class on `<ais-refinement-list>`

sarahdayan:
## Summary

When the <ais-refinement-list> is searchable and a user searches for a facet value that doesn't exist, the root should have the class .ais-RefinementList--noRefinement. This was however not the case because we were relying on `canRefine` which remains `true` as long as the last search contains search results.

Now we're relying on the items returned from SFFV.

In the future, we may want to reconsider what `canRefine` means in the context of `connectRefinementList`, as the source of refinements comes from two places (main search and SFFV), but this would be breaking right now so we can keep it for later.

## Note

This change doesn't update tests, because existing ones weren't testing this use case. This will be touched on in #6090.

### DIFF
--- a/packages/instantsearch.js/src/connectors/refinement-list/connectRefinementList.ts
+++ b/packages/instantsearch.js/src/connectors/refinement-list/connectRefinementList.ts
@@ -126,6 +126,8 @@ export type RefinementListRenderState = {
   isFromSearch: boolean;
   /**
    * `true` if a refinement can be applied.
+   * @MAJOR: reconsider how `canRefine` is computed so it both accounts for the
+   * items returned in the main search and in SFFV.
    */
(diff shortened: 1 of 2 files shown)


Haroenv (commented):
surprised there are no tests on this in vue. Did you also leave a comment on the InstantSearch.js connector that we want to change the meaning of canRefine in the next major?

sarahdayan:
> Did you also leave a comment on the InstantSearch.js connector that we want to change the meaning of canRefine in the next major?

@Haroenv Just did in df173a464 so we don't forget about it.

dhayab (approved):

Haroenv (approved):

PR #6099, opened 2024-03-20
fix(next-app-router): using wrong initial results when mounting on client

### LINKED ISSUE #6086: Using nextjs instant search with App router - navigating back in browser shows incorrect search results
### 🐛 Current behavior

when Using react-instantsearch-nextjs in the app router with routing, if you have been on multiple pages and click on a link to a search item (as an example) and then hit back in the browser, the list of search results shown is not correct, it shows items from a different page. 

The code is a small addition to the example repo, adding a link to each search item 

Code: https://github.com/PaulWild/algolia-next-app-bug

### 🔍 Steps to reproduce

1. Go to a search results page, with pagination and routing (eg https://algolia-next-app-bug.vercel.app/)
2. click on a page other than the page you are on.
3. click a link to go to another page (eg an item in the search list)
4. Hit back in the browser

The search result page you finish on is not showing the correct search results for that page.

### Live reproduction

https://algolia-next-app-bug.vercel.app/

### 💭 Expected behavior

When using the browsers back button the search results should be the correct search results

### Package version

react-instantsearch-nextjs: 0.1.14

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

aymeric-giraudet:
**Summary**

Fixes #6086

**Result**

Starting from Next 14, pressing the back button to go back to a different page does not cause a server-side render, so `InstantSearchNext` would read the `initialResults` again even if the route has changed.
The easiest way to fix this is to delete them from window when disposing (not after reading as this may cause issues in strict mode)

### DIFF
--- a/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
+++ b/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
@@ -54,6 +54,10 @@ export function InstantSearchNext<
   const isMounting = useRef(true);
   useEffect(() => {
     isMounting.current = false;
+    return () => {
+      // This is to make sure that they're not reused if mounting again on a different route
+      delete window[InstantSearchInitialResults];
+    };
   }, []);
 
   const promiseRef = useRef<PromiseWithState<void> | null>(null);


dhayab (approved):

Haroenv (approved):

sarahdayan (commented):
Can we have a test for this?

PR #6100, opened 2024-03-21
feat(helper): set up state management for recommendations

dhayab:
**Summary**

This PR adds the building blocks for handling recommendations in the helper.

Methods for specific models and event handling will be done in subsequent PRs.

**Result**

The helper can store and manipulate recommend queries.

FX-2765


packages/algoliasearch-helper/index.d.ts
```diff
@@ -1457,6 +1463,25 @@ declare namespace algoliasearchHelper {
...
   }
+
+  export type PlainRecommendParameters =
+    | RecommendationsQuery
+    | TrendingQuery
+    | RecommendedForYouQuery;
+
+  export type PlainRecommendParametersWithId = PlainRecommendParameters & {
+    $$id: string;
+  };
```
dhayab, line 1474: I added a specific type for recommend parameters augmented with the internal identifier, so we can rely on `PlainRecommendParameters` for the queries actually sent to the API.

While not specified in the RFC, I chose here `$$id` as the name for the id key, but I'm open to anything else.

packages/algoliasearch-helper/index.d.ts
```diff
@@ -1457,6 +1463,25 @@ declare namespace algoliasearchHelper {
...
+
+  export class RecommendParameters {
+    params: PlainRecommendParametersWithId[];
+    constructor(newParameters?: PlainRecommendParametersWithId);
+    addParams(params: PlainRecommendParametersWithId): RecommendParameters;
+    removeParams(id: string): RecommendParameters;
+    static make(
```
Haroenv, line 1481: do we need this or is it just for symmetry with SearchParameters? As there isn't much reason for SearchParameters.make to exist (it's basically a copy of the constructor with a little extra code for hierarchical menu)

packages/algoliasearch-helper/src/RecommendParameters/index.js
```diff
@@ -0,0 +1,54 @@
...
+ * will be provided when needed. This object is documented for reference as
+ * you'll get it from events generated by the {@link AlgoliaSearchHelper}.
+ * If need be, instantiate this from the factory function {@link RecommendParameters.make}.
+ * @constructor
+ * @classdesc contains all the parameters for recommendations
+ * @param {PlainRecommendParametersWithId} newParameters existing parameters or
+ * partial object for the properties of a new RecommendParameters
+ * @see RecommendParameters.make
+ */
+function RecommendParameters(newParameters) {
```
Haroenv, line 20: these aren't new parameters, but just `params` no? there's no notion of old/new in a constructor

Haroenv, line 20: do we envision global options for recommend ever existing? if so `parameters` should not be the only argument but an object instead IMO

packages/algoliasearch-helper/test/spec/RecommendParameters/methods.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
Haroenv, line 1: just thinking out loud, but are the tests allowed to be typescript here? could be interesting

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1509,6 +1511,17 @@ AlgoliaSearchHelper.prototype._change = function (event) {
   }
 };
 
+AlgoliaSearchHelper.prototype._recommendChange = function (event) {
```
Haroenv, line 1514: as we're not sure of the public api, I wonder if it should be `_recommend.change` and then we don't really need to be sparse with the number of methods

packages/algoliasearch-helper/test/spec/RecommendParameters/methods.js
dhayab: I've checked, ts files don't work run on algoliasearch-helper with its current configuration.

packages/algoliasearch-helper/src/algoliasearch.helper.js
dhayab: I'm not sure I get this, could you elaborate? What would we gain in having this?

Haroenv: thinking about it again, not sure if it's changing much, as we don't really need to document/type these "hidden" methods

packages/algoliasearch-helper/index.d.ts
```diff
@@ -9,6 +9,11 @@ import type {
   SearchResponse,
   SnippetResult,
 } from './types/algoliasearch';
+import type {
+  RecommendationsQuery,
+  RecommendedForYouQuery,
+  TrendingQuery,
+} from '@algolia/recommend';
```
Haroenv, line 16: won't this break if you don't have recommend in the dependency tree? without ts-ignore I think it may cause an error

dhayab: Oh good call, this could be problematic. In most cases this package would exist through `algoliasearch`, but I've adedd a ts-ignore comment for this import.

Haroenv: It won't exist for existing users that just update InstantSearch but not the client (many people probably).

Also I believe there's some tricky stuff with minification etc that makes the ts-ignore not apply to multiple lines in some cases. Maybe not the case here as this is a hand-written .d.ts, but worth checking in a plain project and comparing with the algoliasearch.d.ts file :)

dhayab: I checked and the build script doesn't interact with type files, they are untouched in the package.

aymeric-giraudet (approved):

PR #6104, opened 2024-03-25
feat(instantsearch): prepare index widget for recommend widgets

aymeric-giraudet:
**Summary**

## [FX-2766](https://algolia.atlassian.net/browse/FX-2766)

**Result**

- Added calls to `getWidgetParameters` in both `init` and `addWidgets` to get recommend params.
- Filter `widgetsToRender` based on whether `shouldRender` is implemented or not and if it returns `true`

[FX-2766]: https://algolia.atlassian.net/browse/FX-2766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/algoliasearch-helper/index.js
+++ b/packages/algoliasearch-helper/index.js
@@ -63,7 +63,7 @@ algoliasearchHelper.AlgoliaSearchHelper = AlgoliaSearchHelper;
 algoliasearchHelper.SearchParameters = SearchParameters;
 
 /**
- * Constructor for the object containing all the parameters for reocmmend.
+ * Constructor for the object containing all the parameters for Recommend.
  * @member module:algoliasearchHelper.RecommendParameters
  * @type {RecommendParameters}
  */
--- a/packages/algoliasearch-helper/index.js
+++ b/packages/algoliasearch-helper/index.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var AlgoliaSearchHelper = require('./src/algoliasearch.helper');
+var RecommendParameters = require('./src/RecommendParameters');
 var SearchParameters = require('./src/SearchParameters');
 var SearchResults = require('./src/SearchResults');
 
@@ -60,6 +61,13 @@ algoliasearchHelper.AlgoliaSearchHelper = AlgoliaSearchHelper;
  * @type {SearchParameters}
  */
 algoliasearchHelper.SearchParameters = SearchParameters;
+
+/**
+ * Constructor for the object containing all the parameters for reocmmend.
+ * @member module:algoliasearchHelper.RecommendParameters
+ * @type {RecommendParameters}
+ */
+algoliasearchHelper.RecommendParameters = RecommendParameters;
 
 /**
  * Constructor for the object containing the results of the search.
--- a/packages/instantsearch.js/src/types/widget.ts
+++ b/packages/instantsearch.js/src/types/widget.ts
@@ -150,7 +150,7 @@ type SearchWidgetLifeCycle<TWidgetDescription extends WidgetDescription> = {
 };
 
 type RecommendWidgetLifeCycle<TWidgetDescription extends WidgetDescription> = {
-  dependsOn: 'recommend';
+  dependsOn?: 'recommend';
   getWidgetParameters: (
     state: RecommendParameters,
     widgetParametersOptions: {
--- a/packages/instantsearch.js/src/types/widget.ts
+++ b/packages/instantsearch.js/src/types/widget.ts
@@ -7,6 +7,7 @@ import type {
   AlgoliaSearchHelper as Helper,
   SearchParameters,
   SearchResults,
+  RecommendParameters,
 } from 'algoliasearch-helper';
 
 export type ScopedResult = {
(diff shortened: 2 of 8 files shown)


dhayab (changes requested):
For me, all is covered except this small change in how search widgets return their search parameters.

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -182,6 +195,25 @@ function getLocalWidgetsSearchParameters(
     }, initialSearchParameters);
 }
```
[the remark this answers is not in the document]
aymeric-giraudet: I added it but only if `dependsOn` is `'search'` for now, there was still type inference problems when allowing for `undefined`, probably because we do `Partial` on `RequiredUiLifecycle` already...

packages/algoliasearch-helper/index.js
```diff
@@ -61,6 +62,13 @@ algoliasearchHelper.AlgoliaSearchHelper = AlgoliaSearchHelper;
  */
 algoliasearchHelper.SearchParameters = SearchParameters;
 
+/**
+ * Constructor for the object containing all the parameters for reocmmend.
```
sarahdayan, line 66:
```suggestion
 * Constructor for the object containing all the parameters for Recommend.
```

packages/instantsearch.js/src/types/widget.ts
```diff
@@ -134,6 +137,30 @@ export type WidgetDescription = {
...
 
+type SearchWidgetLifeCycle<TWidgetDescription extends WidgetDescription> = {
+  dependsOn?: 'search';
+  getWidgetParameters?: (
+    state: SearchParameters,
+    widgetParametersOptions: {
+      uiState: Expand<
+        Partial<TWidgetDescription['indexUiState'] & IndexUiState>
+      >;
+    }
```
sarahdayan, line 148: No biggie but this is common with `RecommendWidgetLifeCycle`. We could make a generic type so new APIs are even easier to support for new contributors.

packages/instantsearch.js/src/widgets/index/__tests__/index-test.ts
```diff
@@ -1170,6 +1209,36 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index-widge
       );
     });
 
+    it('calls getWidgetParameters on widgets that depend on recommend', () => {
```
sarahdayan, line 1212: Isn't this a duplicate of l. 264?

packages/instantsearch.js/src/widgets/index/__tests__/index-test.ts
```diff
@@ -1170,6 +1209,36 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/index-widge
...
+
+      const fbt = createFrequentlyBoughtTogether({});
+      instance.addWidgets([fbt]);
+
+      instance.init(createIndexInitOptions({ parent: null }));
+
+      expect(fbt.getWidgetParameters).toHaveBeenCalledTimes(1);
+    });
+
+    it('calls getWidgetParameters on widgets that depend on search and implement the function', () => {
```
sarahdayan, line 1223: Isn't this a duplicate of l. 275?

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -171,15 +184,36 @@ function getLocalWidgetsSearchParameters(
...
+      return state;
+    }
 
-      return widget.getWidgetSearchParameters(state, rest);
-    }, initialSearchParameters);
+    if (widget.dependsOn === 'search' && widget.getWidgetParameters) {
+      return widget.getWidgetParameters(state, rest);
+    }
+
+    return widget.getWidgetSearchParameters(state, rest);
```
sarahdayan, line 196: Should we add a deprecation notice to `getWidgetSearchParameters` now that `getWidgetParameters` aims at replacing it?

packages/instantsearch.js/src/types/widget.ts
aymeric-giraudet: Not sure they would really be the same, I added `uiState` but I don't think it would ever be used for recommend. Still too early to abstract further imo but if there ever is a 3rd type or we need more params then yes

packages/instantsearch.js/src/widgets/index/__tests__/index-test.ts
aymeric-giraudet: One is ran in `addWidgets` and the other in `init`, hierarchy should show that in reports as it's under two different `describe`

packages/instantsearch.js/src/widgets/index/index.ts
aymeric-giraudet: Once we migrate all built-in connectors yes

packages/instantsearch.js/src/widgets/index/__tests__/index-test.ts
aymeric-giraudet: I added a precision that it's after 'init'

packages/instantsearch.js/src/types/widget.ts
aymeric-giraudet: Also I left it as optional for `search` when it's required for `recommend` so we can't really make it generic for now

dhayab (approved):

PR #6115, opened 2024-04-04
feat(recommend): introduce Recommend widgets

aymeric-giraudet:
This PR serves as the base branch for the Recommend feature.

#### [FX-2788](https://algolia.atlassian.net/browse/FX-2788)

[FX-2788]: https://algolia.atlassian.net/browse/FX-2788?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/algoliasearch-helper/index.d.ts
```diff
@@ -1553,6 +1554,20 @@ declare namespace algoliasearchHelper {
...
+
+  type RecommendResultItem<TObject = any> = RecommendResponse<TObject>[0];
+
+  export class RecommendResults<T = any> {
+    constructor(state: RecommendParameters, results: RecommendResponse<T>);
+
+    _state: RecommendParameters;
+    _rawResults: RecommendResponse<T>;
+
+    [index: number]: RecommendResultItem<T>;
```
aymeric-giraudet, line 1569: Can't have `[index: string]` as there would be a problem with `_state` and `_rawResults`

dhayab: Strings would have helped us prefix the id to prevent conflicts between different Recommend widget types. How would we solve this with simple numbers? Maybe we could have a global counter somewhere, and/or have the `$$id` be filled in by the index widget at `addWidgets()` time?

dhayab (commented):
I haven't seen assignees, is this PR ready for review? For me this would work, but might need adjustments to ensure numeric ids don't conflict easily when adding Recommend parameters and creating recommend results.

Haroenv:
would be good to rebase the first commits, and then "merge and rebase" the whole branch at the end (instead of squashing). Maybe it also should be marked as draft now until it's in a "ready to release" state?

packages/algoliasearch-helper/index.d.ts
dhayab: As discussed we're keeping numerical internal ids for the time being. I'll resolve this conversation.

dhayab (approved):
🚀

PR #6117, opened 2024-04-05
feat(recommend): introduce `connectFrequentlyBoughtTogether` connector

raed667:
This PR introduces the `connectFrequentlyBoughtTogether` connector in InstantSearch.js.

FX-2794

PR #6114, opened 2024-04-03
feat(recommend): add FBT to instantsearch-ui-components

raed667:
**Summary**

**Result**

PR #6126, opened 2024-04-09
feat(recommend): introduce `frequentlyBoughtTogether` widget

dhayab:
**Summary**

This PR introduces the `frequentlyBoughtTogether` widget for InstantSearch.js. It leverages the work done on #6114 for the ui component and #6117 for the connector, to create the built-in widget that users will be able to use.

**Notes**

To make the widget feel native to InstantSearch, some changes have been made regarding props from the original standalone component:
- `itemComponent` is now defined in the `item` template
- `headerComponent` is now defined in the `header` template
- `fallbackComponent` is now defined in the `empty` template
- ~~`view` is now defined in the `view` template~~ (this has been removed from the initial implementation)

`translations` is foreign to InstantSearch.js, so I'm not in favor of introducing it with Recommend widgets:
- the `header` template can sub for `translations.title`
- we can introduce a `sliderLabelText` template when we'll migrate the horizontal slider if necessary

FX-2795


packages/instantsearch-ui-components/src/types/Recommend.ts
```diff
@@ -96,6 +96,5 @@ export type RecommendItemComponentProps<TObject> = {
   onAuxClick?: () => void;
 };
 
-// Should be equal to InstantSearchStatus and imported
-// from a shared package in the future
-export type RecommendStatus = 'loading' | 'stalled' | 'idle' | 'error';
+// use instantsearch status instead
```
sarahdayan, line 99:
```suggestion
// TODO: use instantsearch status instead
```

packages/instantsearch.js/src/widgets/frequently-bought-together/frequently-bought-together.tsx
```diff
@@ -0,0 +1,220 @@
...
+import type { PreparedTemplateProps } from '../../lib/templating';
+import type { Template, WidgetFactory, Hit, Renderer } from '../../types';
+import type { RecommendResultItem } from 'algoliasearch-helper';
+import type {
+  RecommendClassNames,
+  FrequentlyBoughtTogetherProps as FrequentlyBoughtTogetherUiProps,
+} from 'instantsearch-ui-components';
+
+const withUsage = createDocumentationMessageGenerator({
+  name: 'frequentlyBoughtTogether',
```
sarahdayan, line 30: This will require to update one test.

```suggestion
  name: 'frequently-bought-together',
```

packages/instantsearch.js/src/widgets/frequently-bought-together/__tests__/frequently-bought-together.test.tsx
```diff
@@ -0,0 +1,570 @@
...
+      // `search.start()` can be moved to the test body and the following
+      // assertion can go away.
+      expect(async () => {
+        search.start();
+        await wait(0);
+      }).not.toWarnDev();
+
+      await wait(0);
+
+      expect(container).toMatchInlineSnapshot(`
```
sarahdayan, line 123: This snapshot should render the default header, no?

packages/instantsearch.js/src/widgets/frequently-bought-together/defaultTemplates.ts
sarahdayan, line 1: I'm wondering whether we should change a bit how we pass default templates, otherwise we'll have to duplicate them in every flavor although they should be the same.

Maybe each flavor shouldn't define default templates at all. Instead, we'd set them in the UI component and override them only when users pass them. WDYT?

sarahdayan (changes requested):

packages/instantsearch.js/src/widgets/frequently-bought-together/__tests__/frequently-bought-together.test.tsx
dhayab: Ah you're right. I misread the current implementation thinking no translations = no header. I'll update this, thanks!

packages/instantsearch.js/src/widgets/frequently-bought-together/defaultTemplates.ts
dhayab: Definitely make sense to centralize component props at the ui-components level.

Not sure moving what we have as "templates" right now is worth it though. We'll remove templates altogether when we ditch Hogan, as we'll be able to rely on component props from ui-components directly.

aymeric-giraudet (approved):

PR #6138, opened 2024-04-16
feat(recommend): handle multiple objectIDs for one widget

aymeric-giraudet:


packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1614,13 +1614,6 @@ AlgoliaSearchHelper.prototype._recommend = function () {
       })
     );
 
-    ids = Array.prototype.concat.apply(
```
aymeric-giraudet, line 1617: This was doubled due to a messy rebase 😅

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1750,7 +1743,20 @@ AlgoliaSearchHelper.prototype._dispatchRecommendResponse = function (
...
+    }
+
+    results[id].hits = results[id].hits
+      .concat(result.hits)
+      .sort(function (a, b) {
+        return b._score - a._score;
+      })
+      .filter(function (item, idx, self) {
+        return self.findIndex((i) => i.objectID === item.objectID) === idx;
+      });
```
aymeric-giraudet, line 1759: There are two other ways we could do this :
- Do it at the end for every id
- Calculate which ids are duplicated in `ids` and only run this for those ids.

Might be premature optimization though, this way the code is clearest.

PR #6142, opened 2024-04-17
feat(recommend): introduce `connectRelatedProducts` connector

sarahdayan:
This introduces the `connectRelatedProducts` connector and tests it in the connector common test suite.

https://algolia.atlassian.net/browse/FX-2813


tests/common/connectors/related-products/options.ts
```diff
@@ -0,0 +1,232 @@
...
+    });
+  });
+}
+
+function createMockedSearchClient() {
+  return createSearchClient({
+    getRecommendations: jest.fn((requests) =>
+      Promise.resolve(
+        createMultiSearchResponse(
+          ...requests.map((request) => {
```
dhayab: I had the same issue. Basically in algoliasearch v3 `getRecommendations` doesn't exist so everything in it is not typed. We can safely ignore this.

```suggestion
          // @ts-ignore
          // `request` will be implicitly typed as any in type-check:v3
          // since `getRecommendations` is not available there
          ...requests.map((request) => {
```

packages/instantsearch.js/src/connectors/related-products/connectRelatedProducts.ts
```diff
@@ -0,0 +1,150 @@
...
+const withUsage = createDocumentationMessageGenerator({
+  name: 'related-products',
+  connector: true,
+});
+
+export type RelatedProductsRenderState<THit extends BaseHit = BaseHit> = {
+  /**
+   * The matched recommendations from the Algolia API.
+   */
+  recommendations: Array<Hit<THit>>;
```
dhayab, line 22: This has been renamed `hits` in `connectFrequentlyBoughtTogether()`.

packages/instantsearch.js/src/connectors/related-products/connectRelatedProducts.ts
```diff
@@ -0,0 +1,150 @@
...
+};
+
+export type RelatedProductsConnectorParams<THit extends BaseHit = BaseHit> = {
+  /**
+   * The `objectID` of the item to get related products from.
+   */
+  objectID: string;
```
dhayab, line 29: The underlying implementation of multi object ids will be handled next sprint, but we can already switch to `objectIDs` in the connector.

packages/instantsearch.js/src/connectors/related-products/connectRelatedProducts.ts
```diff
@@ -0,0 +1,150 @@
...
+   * List of search parameters to send.
+   */
+  fallbackParameters?: Partial<
+    Omit<SearchParameters, 'page' | 'hitsPerPage' | 'offset' | 'length'>
+  >;
+  /**
+   * List of search parameters to send.
+   */
+  queryParameters?: Partial<
+    Omit<SearchParameters, 'page' | 'hitsPerPage' | 'offset' | 'length'>
```
dhayab, line 48: `SearchParameters` can be subbed by `PlainSearchParameters`which already has partial keys.

dhayab (changes requested):
A few questions and changes on how to align both connectors PRs.

I like the outline of the connector common test suite. Are we confident with our "end-user" test suite and dropping more granular tests that might not be possible in CTS (like checking the number of renderings for example) ?

packages/instantsearch.js/src/lib/utils/addWidgetId.ts
```diff
@@ -0,0 +1,11 @@
+import type { Widget } from '../../types';
+
+let id = 0;
+
+export function addWidgetId(widget: Widget) {
```
Haroenv, line 5: no big deal, but this same code is in the other connector PR as well and actually should be in the base branch

packages/instantsearch.js/src/connectors/index.ts
```diff
@@ -30,6 +30,7 @@ export { default as connectNumericMenu } from './numeric-menu/connectNumericMenu
 export { default as connectPagination } from './pagination/connectPagination';
 export { default as connectRange } from './range/connectRange';
 export { default as connectRefinementList } from './refinement-list/connectRefinementList';
+export { default as connectRelatedProducts } from './related-products/connectRelatedProducts';
```
Haroenv, line 33: I see the reasons _for_ doing it like this, but we aren't releasing this as `EXPERIMENTAL_` then?

packages/instantsearch.js/src/connectors/related-products/connectRelatedProducts.ts
```diff
@@ -0,0 +1,150 @@
...
+        },
+
+        getWidgetParameters(state) {
+          return state.addRelatedProducts({
+            objectID,
+            maxRecommendations,
+            threshold,
+            fallbackParameters,
+            queryParameters,
+            $$id: this.$$id!,
```
Haroenv, line 143: I'm not sure I see the point in first marking $$id as optional, but then in usage just non-null asserting it

Haroenv (approved):

packages/instantsearch.js/src/connectors/index.ts
sarahdayan: Nope we're going GA directly.

packages/instantsearch.js/src/lib/utils/addWidgetId.ts
sarahdayan: Yes but I guess we can just handle conflicts, this shouldn't be too hard.

packages/instantsearch.js/src/connectors/related-products/connectRelatedProducts.ts
dhayab: I could not find any piece of info in the RFC favoring either naming. I'm fine with `recommendations` being the exposed prop name. I can switch to this in `frequentlyBoughtTogether`.

dhayab (approved):

PR #6143, opened 2024-04-17
feat(recommend): setup common test suite for `frequentlyBoughtTogether`

dhayab:
**Summary**

This PR adds tests for the `frequentlyBoughtTogether` widget in the common test suite. Only implemented in InstantSearch.js for now, as the widget hasn't been implemented in React and there won't be initially Vue widgets for Recommend.

This CTS also tests templates, which is a first, as before templates were highly flavor-specific. Because we now have a unified UI component, it is much easier. Still, I've opted to create the templates in the flavored implementation of the test suite due to the differences between templates / component props.


tests/common/widgets/frequently-bought-together/options.ts
```diff
@@ -0,0 +1,429 @@
...
+        }
+              </li>
+            </ol>
+          </div>
+        </section>
+      `
+      );
+    });
+
+    describe('components', () => {
```
sarahdayan, line 225: We shouldn't test this in the CTS since passing templates is flavor-specific.

sarahdayan (changes requested):

tests/common/widgets/frequently-bought-together/options.ts
dhayab: I mentioned in the description that this is a new behavior, because this is our first "true" unified ui component, which should not have differences between flavors as, except for the containing element (`rootTagName` in js), everything is coming from the ui component.

tests/common/widgets/frequently-bought-together/options.ts
sarahdayan, line 1: No biggie and no need to change anything now, but the size of each records makes snapshots quite big. In the future we can make the mocks simpler (or override them in the tests) to make snapshots digestible.

packages/instantsearch.js/src/widgets/frequently-bought-together/__tests__/frequently-bought-together.test.tsx
```diff
@@ -359,42 +250,32 @@ describe('frequentlyBoughtTogether', () => {
...
-          maxRecommendations: 0,
-          templates: {
-            empty() {
-              return <p>No results</p>;
-            },
+          item(hit) {
+            return (
+              <Fragment>
+                <h2>${hit.name}</h2>
+                <p>${hit.brand}</p>
```
sarahdayan, line 265: You don't need dollar signs in JSX for interpolation.

packages/instantsearch.js/src/widgets/frequently-bought-together/__tests__/frequently-bought-together.test.tsx
```diff
@@ -313,40 +213,31 @@ describe('frequentlyBoughtTogether', () => {
...
                   class="ais-FrequentlyBoughtTogether-item"
                 >
-                  <h2>
-                    Moschino Love – Shoulder bag
-                  </h2>
-                  <p>
-                    Moschino Love
-                  </p>
+                  <h2 />
+                  <p />
```
sarahdayan, line 217: Why are those empty?

packages/instantsearch.js/src/widgets/frequently-bought-together/__tests__/frequently-bought-together.test.tsx
```diff
@@ -313,40 +213,31 @@ describe('frequentlyBoughtTogether', () => {
...
-                  class="ais-FrequentlyBoughtTogether-item"
-                >
-                  <h2>
-                    Bag La Carrie Bag small black
-                  </h2>
-                  <p>
-                    La Carrie Bag
-                  </p>
+                  <h2 />
+                  <p />
```
sarahdayan, line 223: Why are those empty?

packages/instantsearch.js/src/widgets/frequently-bought-together/__tests__/frequently-bought-together.test.tsx
dhayab: Oh I see, I changed the mocked data to reduce the amount of unnecessary elements in the dataset, but forgot to adjust the templates 🤦 ...

packages/instantsearch.js/src/widgets/frequently-bought-together/__tests__/frequently-bought-together.test.tsx
```diff
@@ -258,38 +168,27 @@ describe('frequentlyBoughtTogether', () => {
...
-            item(hit, { html }) {
-              return html`<h2>${hit.name}</h2>
-                <p>${hit.brand}</p>`;
-            },
+      const options: Parameters<typeof frequentlyBoughtTogether>[0] = {
+        container,
+        objectIDs: ['1'],
+        templates: {
+          header({ recommendations }, { html }) {
+            return html` Frequently bought with (${recommendations.length})`;
```
sarahdayan, line 177:
```suggestion
            return html`Frequently bought with (${recommendations.length})`;
```

sarahdayan (approved):

PR #6149, opened 2024-04-22
feat(hits): add banner to hit ui component

taylorcjohnson:
**Summary**

This is the second pull request to add support for no-code banners to the Hits widget ([RFC](https://algolia.atlassian.net/wiki/spaces/MERCH/pages/4948099490/RFC+-+Rule+Visual+Editor+Banner+Consequence)). It updates the `Hits` `instantsearch-ui-component` to conditionally render a banner (either the default layout or a custom layout if the `bannerComponent` is included) in the DOM with the appropriate class names.

[EMERCH-1420](https://algolia.atlassian.net/browse/EMERCH-1420)

**Result**

  No UI changes

- Update Hits UI component to support conditionally rendering a banner above the hits `ol`
- Update and add new snapshot tests for `banner` and `bannerComponent` props
- Update `Template` component to support Hits widget `emptyComponent` (root changed to `fragment`)
- Update Vue utilities to handle multiple children arguments for `createElement()`
- Increment `bundlesize.config.json`

[EMERCH-1420]: https://algolia.atlassian.net/browse/EMERCH-1420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch-ui-components/src/components/Hits.tsx
+++ b/packages/instantsearch-ui-components/src/components/Hits.tsx
@@ -8,6 +8,57 @@ type Hit = Record<string, unknown> & {
   objectID: string;
 };
 type SendEventForHits = (...props: unknown[]) => void;
+type Banner = {
+  image: {
+    urls: Array<{
+      url: string;
+    }>;
+    title?: string;
+  };
+  link?: {
+    url?: string;
+    target?: '_blank' | '_self';
+  };
+};
+
+type BannerProps = ComponentProps<'aside'> & {
+  banner: Banner;
+  classNames: Pick<
+    Partial<HitsClassNames>,
+    'bannerRoot' | 'bannerLink' | 'bannerImage'
+  >;
+};
+
+function createDefaultBannerComponent({ createElement }: Renderer) {
+  return function DefaultBanner({ classNames, banner }: BannerProps) {
+    if (!banner.image.urls[0].url) {
+      return null;
+    }
+    return (
+      <aside className={cx('ais-Hits-banner', classNames.bannerRoot)}>
+        {banner.link?.url ? (
+          <a
+            className={cx('ais-Hits-banner-link', classNames.bannerLink)}
+            href={banner.link.url}
+            target={banner.link.target}
+          >
+            <img
+              className={cx('ais-Hits-banner-image', classNames.bannerImage)}
+              src={banner.image.urls[0].url}
+              alt={banner.image.title}
+            />
+          </a>
+        ) : (
+          <img
+            className={cx('ais-Hits-banner-image', classNames.bannerImage)}
+            src={banner.image.urls[0].url}
+            alt={banner.image.title}
+          />
+        )}
+      </aside>
+    );
+  };
+}
 
 export type HitsProps<THit> = ComponentProps<'div'> & {
   hits: THit[];
@@ -20,7 +71,12 @@ export type HitsProps<THit> = ComponentProps<'div'> & {
   }) => JSX.Element;
   sendEvent: SendEventForHits;
   classNames?: Partial<HitsClassNames>;
-  emptyComponent?: (props: { className: string }) => JSX.Element;
+  emptyComponent?: (props: { className?: string }) => JSX.Element;
+  banner?: Banner;
+  bannerComponent?: (props: {
+    className: string;
+    banner: Banner;
+  }) => JSX.Element;
 };
 
 export type HitsClassNames = {
@@ -40,9 +96,26 @@ export type HitsClassNames = {
    * Class names to apply to each item element
    */
   item: string | string[];
+  /**
+   * Class names to apply to the banner element
+   */
+  bannerRoot: string | string[];
+  /**
+   * Class names to apply to the banner image element
+   */
+  bannerImage: string | string[];
+  /**
+   * Class names to apply to the banner link element
+   */
+  bannerLink: string | string[];
 };
 
-export function createHitsComponent({ createElement }: Renderer) {
+export function createHitsComponent({ createElement, Fragment }: Renderer) {
+  const DefaultBannerComponent = createDefaultBannerComponent({
+    createElement,
+    Fragment,
+  });
+
   return function Hits<THit extends Hit>(userProps: HitsProps<THit>) {
     const {
       classNames = {},
@@ -50,21 +123,11 @@ export function createHitsComponent({ createElement }: Renderer) {
       itemComponent: ItemComponent,
       sendEvent,
       emptyComponent: EmptyComponent,
+      banner,
+      bannerComponent: BannerComponent,
       ...props
     } = userProps;
 
-    if (hits.length === 0 && EmptyComponent) {
-      return (
-        <EmptyComponent
-          className={cx(
-            'ais-Hits',
-            classNames.root,
-            cx('ais-Hits--empty', classNames.emptyRoot),
-            props.className
-          )}
-        />
-      );
-    }
     return (
       <div
         {...props}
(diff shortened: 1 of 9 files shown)


packages/instantsearch-ui-components/src/components/Hits.tsx
```diff
@@ -8,6 +8,54 @@ type Hit = Record<string, unknown> & {
...
+            <img
+              className={cx('ais-Hits-banner-image', classNames.bannerImage)}
+              src={banner.image.urls[0].url}
+              alt={banner.image.title}
+            />
+          </a>
+        ) : (
+          <img
+            className={cx('ais-Hits-banner-image', classNames.bannerImage)}
+            src={banner.image.urls[0].url}
```
Haroenv, line 51: if `image.urls[0]` is undefined, this display and the surrounding components don't really make sense. Maybe it should early return `null` in that case? Alternatively even in the condition where it's chosen to render the banner at all.

packages/instantsearch-ui-components/src/components/__tests__/Hits.test.tsx
```diff
@@ -60,6 +60,55 @@ describe('Hits', () => {
     `);
   });
 
+  test('renders with a banner', () => {
```
Haroenv, line 63: one without link too please

packages/instantsearch.js/src/components/Template/Template.tsx
```diff
@@ -110,7 +110,11 @@ See: https://www.algolia.com/doc/guides/building-search-ui/upgrade-guides/js/#up
 
     // This is to handle Hogan templates with Fragment as rootTagName
     if (RootTagName === Fragment) {
-      return <RawHtml content={content} key={content} />;
+      return content.trim().charAt(0) === '<' ? (
```
Haroenv, line 113:
```suggestion
      return content.trim()[0] === '<' ? (
```

the first character is ascii so this is easier

packages/instantsearch.js/src/components/Template/Template.tsx
```diff
@@ -110,7 +110,11 @@ See: https://www.algolia.com/doc/guides/building-search-ui/upgrade-guides/js/#up
...
+        <RawHtml content={content} key={content} />
```
Haroenv, line 114: I'd like to have a test of `template` with a fragment root and content like `hello {{world}}` to ensure hogan is still transformed first.

Haroenv, line 114: Technically this _could_ be a breaking change for a hogan template like this: `Hello <button>world</button>` which right now will render as html, but after this change it will no longer do that I believe. Do you mind adding a test to see if it gets escaped or not (and then check what the behaviour before this PR is).

packages/vue-instantsearch/src/util/vue-compat/index-vue2.js
```diff
@@ -9,22 +9,22 @@ export { Vue, Vue2, isVue2, isVue3, version };
 
 const augmentCreateElement =
   (createElement) =>
-  (tag, propsWithClassName = {}, children) => {
+  (tag, propsWithClassName = {}, ...children) => {
```
Haroenv, line 12: I think we need a test or explanation for this change

packages/vue-instantsearch/src/util/vue-compat/index-vue3.js
```diff
@@ -8,7 +8,8 @@ export { createApp, createSSRApp, h, version, nextTick } from 'vue';
 export { Vue, Vue2, isVue2, isVue3 };
 
 export function renderCompat(fn) {
-  function h(tag, props, children) {
+  function h(tag, props, ...childrenArray) {
+    const children = childrenArray.length > 0 ? childrenArray : undefined;
```
Haroenv, line 12: here too, why is it changed?

Haroenv (commented):
great PR, thanks!

packages/vue-instantsearch/src/util/vue-compat/index-vue2.js
aymeric-giraudet: Yeah I meant to do that, basically we didn't support the case where we'd call `createElement` with children as spread arguments.
With the introduction of the banner, this code : 
```jsx
<div>
  {a && <A />}
  {b && <B />}
</div>
```
Gets transformed to : 
```js
createElement('div', {}, a && createElement(A), b && createElement(B))
```
Which isn't something we ever supported, so this is more of a bug fix I guess.

packages/instantsearch.js/src/components/Template/Template.tsx
aymeric-giraudet: Ah yes it would break those templates...
Even checking if there is any HTML and passing it to `innerHTML` wouldn't work, whatever text node before or after gets discarded.
I'll check for another solution, otherwise plan B would be to put `<Banner>` for the empty template on `instantsearch.js`'s side.

packages/vue-instantsearch/src/util/vue-compat/index-vue2.js
Haroenv: sounds good, maybe it should even be a separate PR though?

packages/instantsearch.js/src/components/Template/Template.tsx
Haroenv: Not too sure what the underlying reason of this change is. could we wrap the banner in a div instead of rendering a fragment maybe?

packages/instantsearch-ui-components/src/components/Hits.tsx
taylorcjohnson: `image.urls[0]` should never be undefined based on the typing (because you're correct - a banner without an image would not make sense). And `DefaultBannerComponent` should only render if the banner data is not undefined.

Do you think we need other guardrails in place here?

packages/instantsearch.js/src/components/Template/Template.tsx
aymeric-giraudet: Done !

aymeric-giraudet: This is fixed and I added tests for that

packages/instantsearch-ui-components/src/components/Hits.tsx
```diff
@@ -40,31 +93,38 @@ export type HitsClassNames = {
...
-            cx('ais-Hits--empty', classNames.emptyRoot),
-            props.className
-          )}
-        />
-      );
-    }
+    const DefaultBannerComponent = createDefaultBannerComponent({
+      createElement,
+      Fragment,
+    });
```
aymeric-giraudet, line 126: If this was moved in the upper scope, between line 110 and 110, this will be called only once instead of at every render.
This way `DefaultBannerComponent` won't be remounted each time as it will have the same definition.

packages/instantsearch-ui-components/src/components/Hits.tsx
```diff
@@ -40,31 +93,38 @@ export type HitsClassNames = {
...
+   */
+  bannerImage: string | string[];
+  /**
+   * Class names to apply to the banner link element
+   */
+  bannerLink: string | string[];
 };
 
-export function createHitsComponent({ createElement }: Renderer) {
+export function createHitsComponent({ createElement, Fragment }: Renderer) {
```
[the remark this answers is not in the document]
aymeric-giraudet: Probably because `Fragment` is not marked as optional when it should be as Vue 2 doesn't support it.
Guess it's not too big of a deal though, as Vue InstantSearch is written in JS so this won't error 😆

packages/instantsearch-ui-components/src/components/__tests__/Hits.test.tsx
taylorcjohnson: @Haroenv that test has been added

packages/instantsearch-ui-components/src/components/Hits.tsx
taylorcjohnson: Since we're still using `Fragment` for now, `createDefaultBannerComponent()` has been moved up before the return statement.

Haroenv (approved):

PR #6152, opened 2024-04-23
test(frequentlyBoughtTogether): move to common test suite

dhayab:
**Summary**

This is a followup to #6117 and it moves the connector tests to the common test suite, as introduced with #6142.


packages/instantsearch.js/src/connectors/frequently-bought-together/__tests__/connectFrequentlyBoughtTogether-test.ts
dhayab, line 1: I believe we should keep some of the tests outside of the CTS, that are relevant to how InstantSearch itself relies on  the connector to manage its lifecycle.

aymeric-giraudet (approved):

PR #6154, opened 2024-04-23
feat(instantsearch.js): introduce `relatedProducts` widget

sarahdayan:
This introduces the `relatedProducts` widget and tests it in the widget common test suite.

This PR changes a few things in how we pass templates to the UI component:
- The `<RelatedProducts>` UI component now defines all default templates. This lets us define them once for all flavors instead of passing them manually and risking output drifts.
- InstantSearch.js doesn't pass default templates in `relatedProducts` since they're defined in the UI component.
- All `relatedProducts` templates are passed as `Fragment`s so we can avoid redefining components in the widget.

- https://algolia.atlassian.net/browse/FX-2814
- https://algolia.atlassian.net/browse/FX-2815


packages/instantsearch.js/src/widgets/related-products/related-products.tsx
```diff
@@ -0,0 +1,222 @@
...
+const RelatedProducts = createRelatedProductsComponent({
+  createElement: h,
+  Fragment,
+});
+
+type CreateRendererProps = {
+  containerNode: HTMLElement;
+  cssClasses: RelatedProductsCSSClasses;
+  renderState: {
+    templateProps?: PreparedTemplateProps<Required<RelatedProductsTemplates>>;
```
aymeric-giraudet, line 40: I think this doesn't need `Required`, I know `Hits` has this but other widgets don't. That way we don't have an issue with ESLint line 60/61

aymeric-giraudet (approved):

packages/instantsearch.js/src/widgets/related-products/related-products.tsx
```diff
@@ -0,0 +1,222 @@
...
+>;
+
+const relatedProducts: RelatedProductsWidget = function relatedProducts(
+  widgetParams
+) {
+  const {
+    container,
+    objectIDs,
+    maxRecommendations,
+    queryParameters,
```
dhayab, line 186: We should retrieve and forward `fallbackParameters` to the connector, as this model allows this parameter.

```suggestion
    queryParameters,
    fallbackParameters,
```

packages/instantsearch.js/src/widgets/related-products/related-products.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  const makeWidget = connectRelatedProducts(specializedRenderer, () =>
+    render(null, containerNode)
+  );
+
+  return {
+    ...makeWidget({
+      objectIDs,
+      maxRecommendations,
+      queryParameters,
```
dhayab, line 214:
```suggestion
      queryParameters,
      fallbackParameters,
```

dhayab (changes requested):

PR #6159, opened 2024-04-25
feat(react-instantsearch): introduce `useRelatedProducts` Hook

sarahdayan:
This introduces the `useRelatedProducts` Hook.

https://algolia.atlassian.net/browse/FX-2758


aymeric-giraudet (approved):

packages/react-instantsearch-core/src/connectors/useRelatedProducts.ts
dhayab, line 1: The hook should forward a generic type for the shape of the item, similar to what's done in hits. Otherwise end users won't be able to validate their recommendations types.

dhayab (changes requested):

dhayab (approved):

PR #6162, opened 2024-04-25
feat(react-instantsearch): introduce `<RelatedProducts>` widget

sarahdayan:
This introduces the `<RelatedProducts>` widget.

https://algolia.atlassian.net/browse/FX-2758


packages/react-instantsearch/src/widgets/RelatedProducts.tsx
```diff
@@ -0,0 +1,75 @@
...
+    { $$widgetType: 'ais.relatedProducts' }
+  );
+
+  const uiProps: UiProps<TItem> = {
+    items: recommendations as Array<Hit<TItem>>,
+    itemComponent,
+    headerComponent,
+    fallbackComponent,
+    status,
+    sendEvent: () => {},
```
sarahdayan, line 71: Once the FBT widget is merged we can make `sendEvent` optional in the UI component instead.

aymeric-giraudet (approved):

packages/react-instantsearch/src/widgets/__tests__/RelatedProducts.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { createSearchClient } from '@instantsearch/mocks';
+import { InstantSearchTestWrapper } from '@instantsearch/testutils';
+import { render, waitFor } from '@testing-library/react';
+import React from 'react';
+
+import { RelatedProducts } from '../RelatedProducts';
+
+describe('RelatedProducts', () => {
+  test('renders with translations', async () => {
+    const client = createSearchClient();
```
dhayab, line 14: This should probably be using a client with mock recommendations if we want to test translations for the title.

packages/react-instantsearch-core/src/lib/useRelatedProducts.ts
dhayab, line 1: Maybe this file should not be here? It looks like an artifact of the parent branch.

dhayab (changes requested):
React widget looks great! I just noticed one issue with a test.

packages/react-instantsearch/src/widgets/__tests__/RelatedProducts.test.tsx
sarahdayan: Oops, my bad, I overlooked this test.

dhayab (approved):

PR #6166, opened 2024-04-26
fix(vue): avoid passing a proxied object in Configure

### LINKED ISSUE #6163: AisConfigure throws error: The `searchParameters` option expects an object.
### 🐛 Current behavior

Adding the AisConfigure component to my template throws an error. It was working until I updated versions.

`<AisConfigure :hits-per-page.camel="20" :filters="filters" />`

filters is a computed method that returns a string based on what tab is focused. I have checked and this is returning a properly formatted string.

```
const filters = computed(() => {
  if (state.activeTab === 'my') {
    return userFilter.value
  }
  if (state.activeTab === 'team') {
    return teamFilter.value
  }
  return adminFilter.value
})

const userFilter = computed(() => `user_id:${userStore.user.id}`)
const teamFilter = computed(() => `client_id:${userStore.client.id} AND NOT user_id:${userStore.user.id} AND NOT is_admin:1`)
const adminFilter = computed(() => 'is_admin:1')
```

I have removed the filters and the hits per page individually to see if either were throwing an error but the error was the same. I commented out the AisConfigure tag and it works- without the filters so I know it is something to do with this component.

I have also tried using v-bind to a simple object with static filters and I get the same error. Just adding the component with no props throws the error as well- which is what the error is intended for is my guess.

I have tried reverting versions, uninstalling/reinstalling and still remains.

The error:
```
connectConfigure.js:31 Uncaught (in promise) Error: The `searchParameters` option expects an object.

See documentation: https://www.algolia.com/doc/api-reference/widgets/configure/js/#connector
    at Proxy.<anonymous> (connectConfigure.js:31:13)
    at Proxy.created (widget.js:35:14)
    at callWithErrorHandling (runtime-core.esm-bundler.js:195:33)
    at callWithAsyncErrorHandling (runtime-core.esm-bundler.js:202:17)
    at callHook (runtime-core.esm-bundler.js:3634:3)
    at applyOptions (runtime-core.esm-bundler.js:3553:5)
    at finishComponentSetup (runtime-core.esm-bundler.js:7741:7)
    at setupStatefulComponent (runtime-core.esm-bundler.js:7664:5)
    at setupComponent (runtime-core.esm-bundler.js:7593:36)
    at mountComponent (runtime-core.esm-bundler.js:5922:7)
```

I went into the connectConfigure file and logged the widgetParams. It was an object and had the correct data inside, but it does not get past the isPlainObject check.

The code sandbox shows the relevant parts. Any help would be appreciated.

### 🔍 Steps to reproduce

Add the AisConfigure component to a vue template.

### Live reproduction

Not working but shows my imports and template.

https://codesandbox.io/p/devbox/async-leftpad-jrc38f?file=%2Fpages%2Findex.vue

### 💭 Expected behavior

It should work with the props being passed into.

### Package version

"vue-instantsearch": "^4.16.0", "algoliasearch": "^4.23.3",  "nuxt": "^3.10.0",

### Operating system

macOS 14.4.1

### Browser

Chrome 124.0.6367.78

### Code of Conduct

### LINKED ISSUE #6164: bug: vue-instantsearch issues with vue 3.4
### 🐛 Current behavior

When using the ais-configure component an error is thrown stating: "The `searchParameters` option expects an object. [...]".

The error is thrown in `instantsearch-js`:
https://github.com/algolia/instantsearch/blob/1b8e1d9680ca141861c715c030448b114b96f435/packages/instantsearch.js/src/connectors/configure/connectConfigure.ts#L84-L88

It's seems that the object that are passed along from 'ais-configure'  is a Proxy instance, and then the `isPlainObject` fails.

### 🔍 Steps to reproduce

image
algolia-500

### Live reproduction

https://codesandbox.io/p/devbox/jovial-ioana-g5j43k?file=%2Fsrc%2FApp.vue%3A23%2C36

### 💭 Expected behavior

Should render without error.

### Package version

vue-instantsearch 4.16.0

### Operating system

MacOS 13.2.1

### Browser

Arc 1.37.0

### Code of Conduct

aymeric-giraudet:
**Summary**

Fixes #6163 
Fixes #6164

**Result**

Added a test that was failing in Vue3, it works now !

### DIFF
--- a/packages/vue-instantsearch/src/components/Configure.js
+++ b/packages/vue-instantsearch/src/components/Configure.js
@@ -21,7 +21,7 @@ export default {
   computed: {
     widgetParams() {
       return {
-        searchParameters: this.$attrs,
+        searchParameters: { ...this.$attrs },
       };
     },
   },
--- a/packages/vue-instantsearch/src/components/Configure.js
+++ b/packages/vue-instantsearch/src/components/Configure.js
@@ -21,7 +21,7 @@ export default {
   computed: {
     widgetParams() {
       return {
-        searchParameters: { ...this.$attrs },
+        searchParameters: Object.assign({}, this.$attrs),
       };
     },
   },
--- a/packages/vue-instantsearch/src/components/__tests__/Configure.js
+++ b/packages/vue-instantsearch/src/components/__tests__/Configure.js
@@ -2,6 +2,8 @@
  * @jest-environment jsdom
  */
 
+import { isProxy } from 'node:util/types';
+
 import { mount } from '../../../test/utils';
 import { __setState } from '../../mixins/widget';
 import Configure from '../Configure';
@@ -41,6 +43,14 @@ it('accepts SearchParameters from attributes', () => {
     hitsPerPage: 5,
     distinct: true,
   });
+});
+
+it('assigns $attrs to a new object not wrapped by a proxy', () => {
+  const wrapper = mount(Configure, {
+    propsData: defaultProps,
+  });
+
+  expect(isProxy(wrapper.vm.widgetParams.searchParameters)).toBe(false);
 });
 
 it('renders null without default slot', () => {


packages/vue-instantsearch/src/components/Configure.js
```diff
@@ -21,7 +21,7 @@ export default {
   computed: {
     widgetParams() {
       return {
-        searchParameters: this.$attrs,
+        searchParameters: { ...this.$attrs },
```
dhayab, line 24: I'm not 100% sure but I think our build system does not allow destructuring for vue. Would this work as an alternative?

```suggestion
        searchParameters: Object.assign({}, this.$attrs),
```

aymeric-giraudet: Indeed 😪

Haroenv (approved):

dhayab (approved):

tkjaergaard:
💪

nickycdk:
@Haroenv - Doesn't seem like there are any new releases to the "vue-instantsearch" package?

ashotmark:
@Haroenv we are faced same issue, and I see that this is not published yet.
Our project is waiting until this will be published into npm. it would be appreciated that publish happens asap.
could you please let us know when this will be published?

dhayab:
Hi, we just released vue-instantsearch@4.16.1 which includes a fix for this issue.

ashotmark:
sounds good, thanks! let me upgrade now.

PR #6167, opened 2024-04-26
feat(ui-components): introduce `TrendingItems` component

dhayab:
This moves the `TrendingItems` UI component from **@algolia/recommend** to the monorepo. It also adds the base styles in instantsearch.css and describes its expected markup in our specs.

FX-2831

### DIFF
--- a/packages/instantsearch.css/src/themes/algolia.scss
+++ b/packages/instantsearch.css/src/themes/algolia.scss
@@ -23,10 +23,9 @@ a[class^='ais-'] {
 .ais-ClearRefinements,
 .ais-CurrentRefinements,
 .ais-GeoSearch,
+.ais-FrequentlyBoughtTogether,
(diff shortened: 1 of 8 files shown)


packages/instantsearch.css/src/themes/algolia.scss
```diff
@@ -23,10 +23,9 @@ a[class^='ais-'] {
 .ais-ClearRefinements,
 .ais-CurrentRefinements,
 .ais-GeoSearch,
+.ais-FrequentlyBoughtTogether,
 .ais-HierarchicalMenu,
 .ais-Hits,
-.ais-FrequentlyBoughtTogether,
-.ais-RelatedProducts,
```
dhayab, line 29: I removed one instance of this duplicated selector.

aymeric-giraudet (approved):

PR #6169, opened 2024-04-29
feat(recommend): introduce `connectTrendingItems` connector

dhayab:
This introduces the `connectTrendingItems` connector and tests it in the connector common test suite.

FX-2832


packages/instantsearch.js/src/connectors/trending-items/connectTrendingItems.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export type TrendingItemsConnectorParams<THit extends BaseHit = BaseHit> = {
+  /**
+   * The facet attribute to get recommendations for.
+   */
+  facetName?: string;
+  /**
+   * The facet value to get recommendations for.
+   */
+  facetValue?: string;
```
raed667, line 33: This is not related to this PR (i think typing in the original recommend library has the same limitation) 

But for more correct typing, both `facetName` and `facetValue` have to be either set or not. We can't have one without the other or we will get an error: `Parameters 'facetName' and 'facetValue' must both be empty, or both be set`

Could be a future improvement though.

raed667: I think it is safe to leave it to the API to throw an error in this case as we don't need (redundant) frontend validation, just some type checking would suffice.

dhayab: I found out how to type constraint it in https://github.com/algolia/instantsearch/pull/6169/commits/3c7ee20242f5804c321a6ee4641a34c5eb265cf2.

raed667: nice !

packages/instantsearch.js/src/connectors/trending-items/connectTrendingItems.ts
```diff
@@ -0,0 +1,157 @@
...
+  threshold?: number;
+  /**
+   * List of search parameters to send.
+   */
+  fallbackParameters?: Omit<
+    PlainSearchParameters,
+    'page' | 'hitsPerPage' | 'offset' | 'length'
+  >;
+  /**
+   * List of search parameters to send.
```
Haroenv, line 54: these jsdoc comments aren't that clear, but it can be improved later? What is the description in the api client / current recommend libs?

tests/common/connectors/trending-items/options.ts
```diff
@@ -0,0 +1,222 @@
...
+
+import type { TrendingItemsConnectorSetup } from '.';
+import type { SetupOptions, TestOptions } from '../../common';
+
+export function createOptionsTests(
+  setup: TrendingItemsConnectorSetup,
+  { act, skippedTests }: Required<TestOptions>
+) {
+  skippableDescribe('options', skippedTests, () => {
+    test('forwards parameters to the client', async () => {
```
Haroenv, line 19: there's no test that removing the widget removes the parameters too. Does that actually work, or does `getParameters` only work with `future.cleanUpOnDispose`?

packages/instantsearch.js/src/connectors/trending-items/connectTrendingItems.ts
raed667: This is the description we used everywhere so far

https://www.algolia.com/doc/api-reference/api-methods/get-frequently-bought-together/#method-param-queryparameters

tests/common/connectors/trending-items/options.ts
dhayab: I looked into it and we missed handling this. There is support for removing recommend parameters at the helper level, but this is missing:
- `helper.removeXXX()` expects string ids but we switched to numerical ids
- the connector's `dispose()` method receives the search state instead of the recommend state
- no recommend connector calls `state.removeXXX()` as of now

I'll create a task for tackling this properly.

packages/instantsearch.js/src/connectors/trending-items/connectTrendingItems.ts
dhayab: @Haroenv if that's ok we can handle this in FX-2848, we can definitely make the distinction between fallbackParameters and queryParameters clearer.

Haroenv (approved):
lgtm for now, but indeed we should fix the remove and add another without extra request case

PR #6170, opened 2024-04-29
feat(hits): add banner to react instantsearch hits

taylorcjohnson:
**Summary**

This is the fourth pull request to add support for no-code banners to the Hits widget ([RFC](https://algolia.atlassian.net/wiki/spaces/MERCH/pages/4948099490/RFC+-+Rule+Visual+Editor+Banner+Consequence)). Building on the recent changes to the `Hits` `instantsearch-ui-component` it updates the `Hits` React InstantSearch widget to support a new `bannerComponent`.

[EMERCH-1421](https://algolia.atlassian.net/browse/EMERCH-1421)

There will be a follow-on PR to update React InstantSearch.

**Result**

The hits widget tests have been updated to ensure that the new optional banner component is supported. No styling was applied in the examples or to `instantsearch.css` - that will most likely come in a follow-up PR.

Additionally, the code was tested in the `getting-started` example codebase (connected to appId: `F4T6CUV2AH`, index: `products` in the `beta` environment - where the test rules with a banner consequence exists).

Getting Started App (no `bannerComponent`; query: `nikon`; `nikon camera promotion` alt text and links to `https://www.nikonusa.com/`):
image

Getting Started App (custom `bannerComponent`; query `nikon`):
image

Getting Started App (`bannerComponent={false}`; query `nikon`):
image

[EMERCH-1421]: https://algolia.atlassian.net/browse/EMERCH-1421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react-instantsearch/src/widgets/Hits.tsx
+++ b/packages/react-instantsearch/src/widgets/Hits.tsx
@@ -84,10 +84,9 @@ export function Hits<THit extends BaseHit = BaseHit>({
 
   const banner = BannerComponent === false ? undefined : bannerData;
 
-  const bannerComponent =
-    BannerComponent === false
-      ? undefined
-      : (BannerComponent as HitsUiComponentProps<Hit<THit>>['bannerComponent']);
+  const bannerComponent = (
+    BannerComponent === false ? () => null : BannerComponent
+  ) as HitsUiComponentProps<Hit<THit>>['bannerComponent'];
 
   const uiProps: UiProps<THit> = {
     hits,
--- a/packages/react-instantsearch/src/widgets/Hits.tsx
+++ b/packages/react-instantsearch/src/widgets/Hits.tsx
@@ -12,7 +12,12 @@ import type { UseHitsProps } from 'react-instantsearch-core';
 
 type UiProps<THit extends BaseHit> = Pick<
   HitsUiComponentProps<Hit<THit>>,
-  'hits' | 'sendEvent' | 'itemComponent' | 'emptyComponent'
+  | 'hits'
+  | 'sendEvent'
+  | 'itemComponent'
+  | 'emptyComponent'
+  | 'banner'
+  | 'bannerComponent'
 >;
 
 export type HitsProps<THit extends BaseHit> = Omit<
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch/src/widgets/Hits.tsx
```diff
@@ -65,10 +82,19 @@ export function Hits<THit extends BaseHit = BaseHit>({
     </li>
   );
 
+  const banner = BannerComponent === false ? undefined : bannerData;
+
+  const bannerComponent =
+    BannerComponent === false
+      ? undefined
+      : (BannerComponent as HitsUiComponentProps<Hit<THit>>['bannerComponent']);
```
taylorcjohnson, line 90: I welcome a suggestion for better typing here to avoid the type assertion.

Haroenv: I don't know if it's any better, but this works without casting:

```tsx
  const banner = BannerComponent ? bannerData : undefined;

  const bannerComponent = BannerComponent
    ? (bannerComponentProps: {
        banner: Required<HitsUiComponentProps<Hit<THit>>>['banner'];
        className: string;
      }) => <BannerComponent {...bannerComponentProps} />
    : undefined;
```

aymeric-giraudet: Oh it's because in `react-instantsearch` we accept both function and class components through `React.JSXElementConstructor`, while in `instantsearch-ui-components` it's much more restrictive and only accepts function components that return a single `JSX.Element`.

I think what Taylor did is good and can be simplified with : 
```jsx
const bannerComponent = (
    BannerComponent === false ? () => null : BannerComponent
  ) as HitsUiComponentProps<Hit<THit>>['bannerComponent'];
```
That way we don't need to pass an undefined `banner`, we just pass a dummy component.

Later down the line we'll have to review `instantsearch-ui-component` types because I don't think instantiating component wrappers inline in render like this is good as it makes the user's component remount each time, which might raise issues later if we want to push single object ID hits.

taylorcjohnson: Thank you both. @aymeric-giraudet - in your opinion, what is the advantage of passing `() => null` instead of `undefined`? I'm just curious - I don't have any issue with making this change.

aymeric-giraudet (approved):

Haroenv (approved):

PR #6176, opened 2024-05-02
feat(recommend): enable multiple objectIDs in connectors

aymeric-giraudet:
**Summary**

### [FX-2822](https://algolia.atlassian.net/browse/FX-2822)

**Result**

Following the previous PR, this enables the connectors to pass multiple object IDs

[FX-2822]: https://algolia.atlassian.net/browse/FX-2822?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/algoliasearch-helper/src/algoliasearch.helper.js
+++ b/packages/algoliasearch-helper/src/algoliasearch.helper.js
@@ -1751,15 +1751,18 @@ AlgoliaSearchHelper.prototype._dispatchRecommendResponse = function (
   var results = {};
   Object.keys(idsMap).forEach(function (id) {
     var indices = idsMap[id];
+    var firstResult = content.results[indices[0]];
     if (indices.length === 1) {
-      results[id] = content.results[indices[0]];
+      results[id] = firstResult;
       return;
     }
-    results[id] = sortAndMergeRecommendations(
-      indices.map(function (idx) {
-        return content.results[idx].hits;
-      })
-    );
+    results[id] = Object.assign({}, firstResult, {
+      hits: sortAndMergeRecommendations(
+        indices.map(function (idx) {
+          return content.results[idx].hits;
+        })
+      ),
+    });
   });
 
   states.forEach(function (s) {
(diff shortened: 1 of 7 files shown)


Haroenv (changes requested):
changes look logical, but unit tests fail, so this can't be merged yet. Seems to happen inside the render function, maybe the "no results" case is trying to map as well?

aymeric-giraudet:
@Haroenv 
Ah that was because in my previous PR I did `results[id] = hits` when it should have been a `SearchResults` object,  wouldn't have happened with TypeScript :D 
Not too sure if what I did is best, I just assign the merged hits to the first SearchResults, seems like the recommend libs just assigns hits, so users probably don't care about other fields 🤷‍♂️

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1751,15 +1751,18 @@ AlgoliaSearchHelper.prototype._dispatchRecommendResponse = function (
...
-      results[id] = content.results[indices[0]];
+      results[id] = firstResult;
       return;
     }
-    results[id] = sortAndMergeRecommendations(
-      indices.map(function (idx) {
-        return content.results[idx].hits;
-      })
-    );
+    results[id] = Object.assign({}, firstResult, {
```
Haroenv, line 1759: doesn't this break IE11?

Haroenv: you can use the merge function from util/functions instead

aymeric-giraudet: I assumed it was fine as we use `Object.assign` in `RecommendParameters`, means we'd have to change it there too

Haroenv: that makes sense, this code is only called if you use recommend indeed

Haroenv (approved):

PR #6178, opened 2024-05-03
feat(recommend): introduce TrendingItems

Haroenv:
**Summary**

Introduces the TrendingItems React widget and its related tests. Note that the tests have the same complexity on the type as is the case in vanilla; as the facet props can only be used if they're both set, but manipulating the props as a whole loses that relation.

**Result**

FX-2853

### DIFF
--- a/packages/react-instantsearch/src/widgets/TrendingItems.tsx
+++ b/packages/react-instantsearch/src/widgets/TrendingItems.tsx
@@ -60,7 +60,7 @@ export function TrendingItems<TItem extends BaseHit = BaseHit>({
       queryParameters,
       transformItems,
     },
-    { $$widgetType: 'ais.TrendingItems' }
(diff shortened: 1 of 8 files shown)


packages/react-instantsearch/src/widgets/TrendingItems.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { recommendations } = useTrendingItems(
+    {
+      ...facetParameters,
+      maxRecommendations,
+      threshold,
+      fallbackParameters,
+      queryParameters,
+      transformItems,
+    },
+    { $$widgetType: 'ais.TrendingItems' }
```
dhayab, line 63:
```suggestion
    { $$widgetType: 'ais.trendingItems' }
```

dhayab (approved):
Just a nitpicky change request, otherwise that's good for me!

PR #6192, opened 2024-05-13
fix(recommend): clean recommend state when removing widgets

dhayab:
**Summary**

This PR removes recommend parameters from the recommend state when their relevant widgets are disposed.

FX-2854

### DIFF
--- a/packages/instantsearch.js/src/connectors/related-products/connectRelatedProducts.ts
+++ b/packages/instantsearch.js/src/connectors/related-products/connectRelatedProducts.ts
@@ -129,10 +129,9 @@ const connectRelatedProducts: RelatedProductsConnector =
           };
         },
 
-        dispose({ state }) {
+        dispose({ recommendState }) {
           unmountFn();
-
-          return state.removeParams(this.$$id!);
+          return recommendState?.removeParams(this.$$id!);
         },
 
         getWidgetParameters(state) {
--- a/packages/instantsearch.js/src/connectors/related-products/connectRelatedProducts.ts
+++ b/packages/instantsearch.js/src/connectors/related-products/connectRelatedProducts.ts
@@ -131,7 +131,7 @@ const connectRelatedProducts: RelatedProductsConnector =
 
         dispose({ recommendState }) {
           unmountFn();
-          return recommendState?.removeParams(this.$$id!);
+          return recommendState.removeParams(this.$$id!);
         },
 
         getWidgetParameters(state) {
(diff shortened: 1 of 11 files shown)


packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -458,16 +458,31 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
...
+              state: states.cleanedSearchState,
+              recommendState: states.cleanedRecommendState,
+              parent: this,
+            });
 
-          return next || state;
-        }, helper!.state);
+            return {
+              ...states,
+              ...(next instanceof algoliasearchHelper.SearchParameters
```
Haroenv, line 473: mutating rather than spreading will be quite a bit faster here

```
if (next instanceof algoliasearchHelper.RecommendParameters) {
  states.cleanedRecommendState = next
} else {
  states.cleanedSearchState = next
}
return states
```

packages/instantsearch.js/src/types/widget.ts
```diff
@@ -51,6 +51,7 @@ export type RenderOptions = SharedRenderOptions & {
 export type DisposeOptions = {
   helper: Helper;
   state: SearchParameters;
+  recommendState?: RecommendParameters;
```
Haroenv, line 54: why is this optional?

Haroenv: as it's just typescript, I'm more inclined to make it required

packages/instantsearch.js/src/connectors/related-products/connectRelatedProducts.ts
```diff
@@ -129,10 +129,9 @@ const connectRelatedProducts: RelatedProductsConnector =
           };
         },
 
-        dispose({ state }) {
+        dispose({ recommendState }) {
           unmountFn();
-
-          return state;
+          return recommendState?.removeParams(this.$$id!);
```
Haroenv, line 134: these no longer need the ? now

dhayab: oh of course, well spotted 😅

packages/instantsearch.js/src/widgets/index/__tests__/index-test.ts
```diff
@@ -120,6 +121,9 @@ describe('index', () => {
           objectID: 'abc',
         });
       },
+      dispose({ recommendState }) {
+        return recommendState?.removeParams(this.$$id!);
```
Haroenv: here too fyi

packages/instantsearch.js/src/connectors/trending-items/connectTrendingItems.ts
```diff
@@ -133,10 +133,9 @@ const connectTrendingItems: TrendingItemsConnector =
           };
         },
 
-        dispose({ state }) {
+        dispose({ recommendState }) {
           unmountFn();
-
-          return state;
+          return recommendState?.removeParams(this.$$id!);
```
Haroenv: same

Haroenv (approved):

PR #6201, opened 2024-05-16
tests(recommend): use shared mock client with recommendations in tests

dhayab:
**Summary**

This PR moves the `createMockedSearchClient()` defined on all recommend tests into a shared location and makes it customizable.

**Result**

Mocked search client with recommendations fixture is now defined only once and can be used in all tests for recommend widgets.

### DIFF
--- a/tests/mocks/fixtures/recommendations.ts
+++ b/tests/mocks/fixtures/recommendations.ts
@@ -1,0 +1,68 @@
+import {
+  createRecommendResponse,
+  createSingleSearchResponse,
+} from '../createAPIResponse';
+import { createSearchClient } from '../createSearchClient';
+
+type Options = {
+  /**
+   * Returns only the `objectID` attribute of the hits when `true`.
+   *
+   * @default false
+   */
+  minimal?: boolean;
+};
+
+const fixture = [
+  {
+    _highlightResult: {
+      name: {
+        matchLevel: 'none' as const,
+        matchedWords: [],
+        value: '<em>Moschino Love</em> – Shoulder bag',
+      },
+    },
+    name: 'Moschino Love – Shoulder bag',
+    objectID: '1',
+  },
+  {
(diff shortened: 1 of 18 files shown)


tests/mocks/fixtures/recommendations.ts
dhayab, line 1: I'm open to suggestions if `createMockedSearchClientWithRecommendations` is too much 😅

tests/mocks/fixtures/recommendations.ts
```diff
@@ -0,0 +1,68 @@
...
+        matchedWords: [],
+        value: '<em>Bag</em> “Sabrina“ medium Gabs',
+      },
+    },
+    name: 'Bag “Sabrina“ medium Gabs',
+    objectID: '2',
+  },
+];
+
+export function createMockedSearchClientWithRecommendations(
```
Haroenv, line 41:
```suggestion
export function createRecommmendSearchClient(
```

also I wonder, couldn't this simply be part of the existing `createSearchClient` with the option `minimalRecommendations`?

Haroenv (commented):
change in tests looks good, I think can be simplified a little more either by changing the name or making part of the createSearchClient that already exists

tests/mocks/fixtures/recommendations.ts
dhayab: I didn't want to change how `createSearchClient` doesn't provide any fixture for search or recommendations. I thought about only sharing the fixture, but wrapping those with `createMultiSearchResponse` and `createSingleSearchResponse` on each test would not be as beneficial as moving those to the shared entity as well.

Haroenv: ok that makes sense, although I wouldn't be against having a fixture for search to avoid repetition too in the future

Haroenv (approved):

PR #6204, opened 2024-05-16
fix(helper): don't send empty "tagRefinements" and "facets" parameters

Haroenv:
**Summary**

In the first request, we are sending `facets: []` and `tagRefinements: ''` when there are no values for those queries, this isn't needed, and removed in this pull request.

This came from an assumption that we could skip queries that have no refinements, but that's not true if you're for example _only_ displaying hits

**Result**

Slightly more minimal request sent, consistent with the other parameters

### DIFF
--- a/packages/algoliasearch-helper/src/SearchResults/index.js
+++ b/packages/algoliasearch-helper/src/SearchResults/index.js
@@ -231,7 +231,7 @@ function findMatchingHierarchicalFacetFromAttributeName(
 }
  **/
 function SearchResults(state, results, options) {
-  var mainSubResponse = results[0];
+  var mainSubResponse = results[0] || {};
 
   this._rawResults = results;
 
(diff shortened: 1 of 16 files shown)


packages/algoliasearch-helper/src/SearchResults/index.js
```diff
@@ -231,7 +231,7 @@ function findMatchingHierarchicalFacetFromAttributeName(
 }
  **/
 function SearchResults(state, results, options) {
-  var mainSubResponse = results[0];
+  var mainSubResponse = results[0] || {};
```
dhayab, line 234: Why should the first response be defaulted?

Haroenv: this was needed for a previous iteration where it was possible no request was sent if there are no parameters. I reverted that, but this change stayed, as it allows the helper to be a bit more flexible and to be able to instantiate an empty SearchResults if needed

PR #6205, opened 2024-05-16
fix(queries): prevent network request

Haroenv:
**Summary**

In index and instantsearch, keep track of which type of widgets are used, and only send a request of the right type.

**Result**

If only recommend or search widgets are present, only that request is done.

This has slight further impact to ensure that a render still happens when only recommend returns results.

FX-2892

### DIFF
--- a/packages/instantsearch.js/src/widgets/index/index.ts
+++ b/packages/instantsearch.js/src/widgets/index/index.ts
@@ -468,6 +468,22 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
         (widget) => widgets.indexOf(widget) === -1
       );
 
+      localWidgets.forEach((widget) => {
+        if (isIndexWidget(widget)) {
+          return;
+        }
+
+        if (localInstantSearchInstance && widget.dependsOn === 'recommend') {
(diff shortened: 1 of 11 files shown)


packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -457,6 +468,22 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
...
+        if (localInstantSearchInstance && widget.dependsOn === 'recommend') {
+          localInstantSearchInstance._hasRecommendWidget = true;
+        } else if (localInstantSearchInstance) {
+          localInstantSearchInstance._hasSearchWidget = true;
+        } else if (widget.dependsOn === 'recommend') {
+          hasRecommendWidget = true;
+        } else {
+          hasSearchWidget = true;
+        }
+      });
```
dhayab, line 485: Should the values be reset to `false` somewhere before?

dhayab: Ah yes I forgot about that. I'm fine with this then.

dhayab (approved):

PR #6208, opened 2024-05-21
fix(recommend): avoid using wrong results for recommend in SSR

aymeric-giraudet:


packages/react-instantsearch-core/src/hooks/useConnector.ts
```diff
@@ -116,7 +117,12 @@ export function useConnector<
         helper,
         parent: parentIndex,
         instantSearchInstance: search,
-        results,
+        results:
+          widget.dependsOn === 'recommend'
+            ? // @TODO: this is to avoid using wrong hits in SSR,
+              // will be replace with SSR support for recommend
+              (null as unknown as algoliasearchHelper.SearchResults<any>)
```
Haroenv, line 124: as it's a type, you can do `import type { SearchResults }` but it doesn't really change anything

Haroenv (approved):

PR #6211, opened 2024-05-21
chore(examples): integrate recommend into getting started

Haroenv:
**Summary**

Adds recommend to the js and react getting-started examples.

- if you don't have recommend access, it displays no results as expected, and doesn't break the search
- trendingItems in the search page
- new "products" page (PDP) with relatedItems

**Result**

FX-2804

### DIFF
--- a/examples/react/getting-started/products.html
+++ b/examples/react/getting-started/products.html
@@ -10,12 +10,12 @@
     <link rel="shortcut icon" href="favicon.png" />
 
     <!--
-    Do not use @7 in production, use a complete version like x.x.x, see website for latest version:
-    https://www.algolia.com/doc/guides/building-search-ui/installation/react/#load-the-styles
+      Do not use @8 in production, use a complete version like x.x.x, see website for latest version:
+      https://www.algolia.com/doc/guides/building-search-ui/installation/js/#load-the-styles
     -->
     <link
       rel="stylesheet"
-      href="https://cdn.jsdelivr.net/npm/instantsearch.css@7/themes/satellite-min.css"
+      href="https://cdn.jsdelivr.net/npm/instantsearch.css@8/themes/satellite-min.css"
     />
 
     <title>React InstantSearch — Getting started</title>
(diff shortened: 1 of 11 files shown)


examples/react/getting-started/products.html
```diff
@@ -0,0 +1,31 @@
...
+    <link rel="shortcut icon" href="favicon.png" />
+
+    <!--
+    Do not use @7 in production, use a complete version like x.x.x, see website for latest version:
+    https://www.algolia.com/doc/guides/building-search-ui/installation/react/#load-the-styles
+    -->
+    <link
+      rel="stylesheet"
+      href="https://cdn.jsdelivr.net/npm/instantsearch.css@7/themes/satellite-min.css"
+    />
```
dhayab, line 19: Could you bump the major to 8 in here and in **index.html** as well? It will include base styles for all recommend widgets.

dhayab: We can also do it in a separate PR for all examples and templates.

examples/react/getting-started/src/Product.tsx
```diff
@@ -0,0 +1,87 @@
...
+          <Configure hitsPerPage={1} filters={`objectID:${pid}`} />
+          <Hits
+            hitComponent={HitComponent}
+            classNames={{
+              root: 'ais-Hits--single',
+            }}
+          />
+          <RelatedProducts
+            itemComponent={ItemComponent}
+            emptyComponent={() => <></>}
```
dhayab, line 49: why?

Haroenv: Without this I was getting a flash of "no results". Not sure if it should be avoided here or investigated and fixed first

dhayab: Maybe let's keep this here now and we can always remove it when we figure it out in the widget itself, wdyt?

dhayab (approved):

PR #6212, opened 2024-05-22
fix(hits): rename the exposed "hits" to "items"

Haroenv:
**Summary**

Replace the exposed `hits` with `items` for consistency with recommend, transformItems and other widgets that all return `items`.

`hits` of course still works for now, but `items` is also available from the connectors.

**Result**

As well as `hits`, `items` is also available from connectHits, connectInfiniteHits (connectGeoSearch already only had `items`)

There's no warning for using `hits`, as there's no way to have that in a way which doesn't warn for regular usage too.

For now, no change is done in `ui-components`, but I'm happy to do so if we're happy with the hypothetical breaking change for that package.


dhayab (approved):
Looks great. Are there references to hits in examples or CISA templates we could also switch to items?

Haroenv:
not that I saw @dhayab

aymeric-giraudet (approved):

PR #6215, opened 2024-05-24
feat(recommend): enable server-side rendering

aymeric-giraudet:
**Summary**

## [FX-2895](https://algolia.atlassian.net/browse/FX-2895)

**Result**

- Made `_rawResults` scoped for `RecommendResults` to avoid bloating multi-index with same results
- Typed and renamed `recommendCache` to `_recommendCache` as it's private
  - It will also be hydrated to avoid requests on the client
- `waitForResults` wait for either recommend or search, or both if needed
- Changes to `hydrateSearchClient` needed as search results are now optional

[FX-2895]: https://algolia.atlassian.net/browse/FX-2895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/lib/__tests__/server.test.ts
+++ b/packages/instantsearch.js/src/lib/__tests__/server.test.ts
@@ -99,7 +99,7 @@ describe('getInitialResults', () => {
     const search = instantsearch({
       indexName: 'indexName',
       searchClient: createSearchClient(),
-    });
+    }).addWidgets([connectSearchBox(() => {})({})]);
 
     search.start();
 
@@ -118,6 +118,7 @@ describe('getInitialResults', () => {
           index: 'indexName',
           numericRefinements: {},
           tagRefinements: [],
+          query: '',
         },
         results: [
           {
@@ -140,7 +141,9 @@ describe('getInitialResults', () => {
   test('returns the current results from one non-rootindex', async () => {
     const search = instantsearch({
       searchClient: createSearchClient(),
-    }).addWidgets([index({ indexName: 'indexName' })]);
+    })
+      .addWidgets([connectSearchBox(() => {})({})])
+      .addWidgets([index({ indexName: 'indexName' })]);
 
     search.start();
 
@@ -182,7 +185,7 @@ describe('getInitialResults', () => {
     const search = instantsearch({
       indexName: 'indexName',
       searchClient: createSearchClient(),
-    });
+    }).addWidgets([connectSearchBox(() => {})({})]);
 
     search.addWidgets([index({ indexName: 'indexName2' })]);
 
@@ -203,6 +206,7 @@ describe('getInitialResults', () => {
           index: 'indexName',
           numericRefinements: {},
           tagRefinements: [],
+          query: '',
         },
         results: [
           {
--- a/packages/instantsearch.js/src/lib/server.ts
+++ b/packages/instantsearch.js/src/lib/server.ts
@@ -12,7 +12,8 @@ import type {
  * in `getServerState()`.
  */
 export function waitForResults(
-  search: InstantSearch
+  search: InstantSearch,
+  skipRecommend: boolean = false
 ): Promise<SearchOptions[]> {
   const helper = search.mainHelper!;
 
@@ -21,19 +22,32 @@ export function waitForResults(
   let requestParamsList: SearchOptions[];
   const client = helper.getClient();
   helper.setClient({
+    ...client,
     search(queries) {
       requestParamsList = queries.map(({ params }) => params!);
       return client.search(queries);
     },
   });
 
-  helper.searchOnlyWithDerivedHelpers();
+  search._hasSearchWidget && helper.searchOnlyWithDerivedHelpers();
+  !skipRecommend && search._hasRecommendWidget && helper.recommend();
 
   return new Promise((resolve, reject) => {
+    let searchResultsReceived = !search._hasSearchWidget;
+    let recommendResultsReceived = !search._hasRecommendWidget || skipRecommend;
     // All derived helpers resolve in the same tick so we're safe only relying
     // on the first one.
     helper.derivedHelpers[0].on('result', () => {
-      resolve(requestParamsList);
+      searchResultsReceived = true;
+      if (recommendResultsReceived) {
+        resolve(requestParamsList!);
+      }
+    });
+    helper.derivedHelpers[0].on('recommend:result', () => {
+      recommendResultsReceived = true;
+      if (searchResultsReceived) {
+        resolve(requestParamsList!);
+      }
     });
 
     // However, we listen to errors that can happen on any derived helper because
@@ -68,13 +82,23 @@ export function getInitialResults(
   let requestParamsIndex = 0;
   walkIndex(rootIndex, (widget) => {
     const searchResults = widget.getResults();
-    if (searchResults) {
+    const recommendResults = widget.getHelper()?.lastRecommendResults;
+    if (searchResults || recommendResults) {
       const requestParams = requestParamsList?.[requestParamsIndex++];
       initialResults[widget.getIndexId()] = {
         // We convert the Helper state to a plain object to pass parsable data
         // structures from server to client.
-        state: { ...searchResults._state },
-        results: searchResults._rawResults,
+        ...(searchResults && {
+          state: { ...searchResults._state },
+          results: searchResults._rawResults,
+        }),
+        ...(recommendResults && {
+          recommendResults: {
+            // We have to stringify + parse because of some explicitly undefined values.
+            params: JSON.parse(JSON.stringify(recommendResults._state.params)),
+            results: recommendResults._rawResults,
+          },
+        }),
         ...(requestParams && { requestParams }),
       };
     }
--- a/packages/instantsearch.js/src/lib/utils/__tests__/hydrateRecommendCache-test.ts
+++ b/packages/instantsearch.js/src/lib/utils/__tests__/hydrateRecommendCache-test.ts
@@ -2,13 +2,15 @@ import {
   createSearchClient,
   createSingleSearchResponse,
 } from '@instantsearch/mocks';
+import { wait } from '@instantsearch/testutils';
 import algoliaSearchHelper from 'algoliasearch-helper';
 
 import { hydrateRecommendCache } from '../hydrateRecommendCache';
 
 describe('hydrateRecommendCache', () => {
-  it('should hydrate the helper with the recommend cache', () => {
-    const helper = algoliaSearchHelper(createSearchClient(), '');
+  it('should hydrate the helper with the recommend cache', async () => {
+    const searchClient = createSearchClient();
+    const helper = algoliaSearchHelper(searchClient, '');
 
     const response0 = createSingleSearchResponse();
     const response1 = createSingleSearchResponse();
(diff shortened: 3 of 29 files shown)


packages/instantsearch.js/src/lib/__tests__/server.test.ts
```diff
@@ -99,7 +99,7 @@ describe('getInitialResults', () => {
     const search = instantsearch({
       indexName: 'indexName',
       searchClient: createSearchClient(),
-    });
+    }).addWidgets([connectSearchBox(() => {})({})]);
```
aymeric-giraudet, line 102: This is required because we now only trigger `search` when there are `search` widgets.
It was already the case in a real-life app anyway.

packages/instantsearch.js/src/lib/server.ts
```diff
@@ -12,7 +12,8 @@ import type {
  * in `getServerState()`.
  */
 export function waitForResults(
-  search: InstantSearch
+  search: InstantSearch,
+  skipRecommend: boolean = false
```
Haroenv, line 16: what's skipRecommend about? seems like we check hasRecommendWidget anyway?

packages/instantsearch.js/src/lib/utils/__tests__/hydrateRecommendCache-test.ts
```diff
@@ -0,0 +1,36 @@
+import {
+  createSearchClient,
+  createSingleSearchResponse,
+} from '@instantsearch/mocks';
+import algoliaSearchHelper from 'algoliasearch-helper';
+
+import { hydrateRecommendCache } from '../hydrateRecommendCache';
+
+describe('hydrateRecommendCache', () => {
+  it('should hydrate the helper with the recommend cache', () => {
```
Haroenv, line 10: add maybe a test with a broken response?

Haroenv, line 10: add a test to ensure no actual recommend request happens if you do a new request with same parameters?

packages/instantsearch.js/src/lib/utils/__tests__/hydrateSearchClient-test.ts
```diff
@@ -147,4 +147,24 @@ describe('hydrateSearchClient', () => {
       } as unknown as InitialResults);
     }).not.toThrow();
   });
+
+  it('should not throw if state or results are missing', () => {
```
Haroenv, line 151: in think in Vue we might have been relying on this (maybe a separate error written there) to ensure you're actually passing _something_ to the function

packages/instantsearch.js/src/lib/utils/hydrateRecommendCache.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function hydrateRecommendCache(
+  helper: AlgoliaSearchHelper,
+  initialResults: InitialResults
+) {
+  const recommendCache = Object.keys(initialResults).reduce(
+    (acc, indexName) => {
+      const initialResult = initialResults[indexName];
+      if (initialResult.recommendResults) {
+        return { ...acc, ...initialResult.recommendResults.results };
```
Haroenv, line 12: let's loop without spreading acc maybe? not sure it's easy as we want to merge though. The equivalent I may be thinking of is `Object.assign({}, ...Object.keys(initialResults).map(key => initialResults[indexName].recommendResults.results))`, possibly worth benchmarking if it's worth the difference though as that may not be super efficient either. We can also use the `merge` function (is it also here or in the helper only) instead of object.assign possibly. At least that would avoid the constant iteration over the acc

packages/react-instantsearch-core/src/components/InstantSearchSSRProvider.tsx
```diff
@@ -30,6 +30,8 @@ export function InstantSearchSSRProvider<
     TRouteState
   > | null>(null);
 
+  const recommendIdx = React.useRef(0);
```
Haroenv, line 33:
```suggestion
  // This is used to re-map the result index to the requesting widget
  const recommendIdx = React.useRef(0);
```

packages/react-instantsearch-core/src/lib/useSearchResults.ts
```diff
@@ -16,9 +16,13 @@ export type SearchResultsApi = {
...
-  const [searchResults, setSearchResults] = useState(() =>
-    getIndexSearchResults(searchIndex)
-  );
+  const [searchResults, setSearchResults] = useState(() => {
+    const indexSearchResults = getIndexSearchResults(searchIndex);
+    return {
+      results: indexSearchResults.results,
+      scopedResults: indexSearchResults.scopedResults,
+    };
+  });
```
Haroenv, line 25:
```suggestion
  const [searchResults, setSearchResults] = useState(() =>
    getIndexSearchResults(searchIndex)
  );
```

I don't think this diff is needed?

packages/react-instantsearch-core/src/server/getServerState.tsx
```diff
@@ -127,7 +134,7 @@ function execute({
         );
       }
 
-      return waitForResults(searchRef.current);
+      return waitForResults(searchRef.current, skipRecommend);
```
Haroenv, line 137: if we're skipping recommend here, where are we actually getting the recommend results?

packages/react-instantsearch-nextjs/src/InitializePromise.tsx
```diff
@@ -25,16 +25,30 @@ export function InitializePromise() {
...
-        resolve();
+      let searchReceived = false;
+      let recommendReceived = false;
+      search.mainHelper!.derivedHelpers[0].once('result', () => {
+        searchReceived = true;
+        if (!search._hasRecommendWidget || recommendReceived) {
+          resolve();
+        }
+      });
+      search.mainHelper!.derivedHelpers[0].once('recommend:result', () => {
```
Haroenv, line 47: won't this fail if you have no recommend widgets?

Haroenv: clearly  it's covered through the tests already, but I don't understand how 🤔

Haroenv (commented):
looks good! I have some questions that maybe could be clarified in the code

packages/instantsearch.js/src/lib/server.ts
aymeric-giraudet: That's for `getServerState` for the pages router, as with dynamic widgets we *mount* the provided component a second time, we don't receive a response from recommend anymore as the results were cached.
Maybe the problem is that in the helper when there are no queries to be made it doesn't raise any event : https://github.com/algolia/instantsearch/blob/e55c345bce1969f4379db39de27642a7929c9411/packages/algoliasearch-helper/src/algoliasearch.helper.js#L1636-L1638
But I'm really not sure if it's worth changing now

packages/react-instantsearch-core/src/lib/useSearchResults.ts
aymeric-giraudet: It is, otherwise we would have `recommendResults` in `useSearchResults` which we haven't decided yet. Problem is `getIndexSearchResults` is both used by this and `getServerState`

packages/react-instantsearch-core/src/server/getServerState.tsx
aymeric-giraudet: It's in the `initialResults` passed from the first render

packages/react-instantsearch-nextjs/src/InitializePromise.tsx
```diff
@@ -25,16 +25,30 @@ export function InitializePromise() {
...
     },
   });
 
+  resetWidgetId();
+
   const waitForResults = () =>
     new Promise<void>((resolve) => {
-      search.mainHelper!.derivedHelpers[0].on('result', () => {
-        resolve();
+      let searchReceived = false;
```
dhayab, line 39: Suggestion for a different more "promise-like" approach:

```js
Promise.all([
  search._hasSearchWidget && new Promise(
    (resolve) => search.mainHelper!.derivedHelpers[0].once('result', resolve)
  ),
  search._hasRecommendWidget && new Promise(
    (resolve) => search.mainHelper!.derivedHelpers[0].once('recommend:result', resolve)
  ),
].filter(Boolean)).then(resolve);
```

dhayab (commented):
This is already looking pretty good 👏

packages/react-instantsearch-core/src/lib/useSearchResults.ts
Haroenv: ok sounds fair, maybe a comment to explain that this is to prevent recommend results? not sure if needed though

packages/instantsearch.js/src/lib/utils/__tests__/hydrateRecommendCache-test.ts
aymeric-giraudet: What do you mean by broken response ? I made one with an undefined `recommendResults` which could happen if they don't have recommend widgets

packages/instantsearch.js/src/lib/utils/hydrateRecommendCache.ts
aymeric-giraudet: Good point, and I did do a benchmark : https://jsbench.me/oulwm3j29m/1
It's not as big a difference I thought would be on Chromium surprisingly (it is on WebKit though I guess ?), but still it's better yeah :D

packages/react-instantsearch-nextjs/src/InitializePromise.tsx
aymeric-giraudet: I would have preferred it that way, unfortunately `_hasSearchWidget` and `_hasRecommendWidget` will always be `false` at this point as we initialize the promise before adding any widget.
This is one of the use cases where an event-based API makes sense. I could do that in the "legacy" `waitForResults` though, but maybe it's best to keep consistency.

packages/instantsearch.js/src/lib/utils/hydrateRecommendCache.ts
aymeric-giraudet: We don't allow `Object.assign` in babel config 🙃

packages/instantsearch.js/src/lib/utils/hydrateRecommendCache.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function hydrateRecommendCache(
+  helper: AlgoliaSearchHelper,
+  initialResults: InitialResults
+) {
+  const recommendCache = Object.keys(initialResults).reduce(
+    (acc, indexName) => {
+      const initialResult = initialResults[indexName];
+      if (initialResult.recommendResults) {
+        return Object.assign(acc, initialResult.recommendResults.results);
```
Haroenv, line 12: can go back to spread then (but maybe leave a comment about `const recommendCache = Object.assign({}, ...Object.keys(initialResults).map(key => initialResults[key].recommendResults.results))` as the next major option, seems like that's quicker still (difference much bigger in safari than in chrome though)

https://jsbench.me/9zlworivz7/1

Screenshot 2024-05-27 at 11 27 57

Haroenv (commented):
lgtm once build passes again / object.assign is removed

Haroenv (approved):

PR #6217, opened 2024-05-27
test(rating-menu): add common test suite

dhayab:
**Summary**

This reworks the tests for the `ratingMenu` widget to move them in the common test suite and reduce flavor-specific tests.

FX-2532


tests/common/widgets/rating-menu/options.ts
```diff
@@ -0,0 +1,447 @@
...
+function createMockedSearchClient() {
+  return createSearchClient({
+    search: jest.fn((requests) => {
+      return Promise.resolve(
+        createMultiSearchResponse(
+          ...requests.map(() =>
+            createSingleSearchResponse({
+              facets: {
+                rating: {
+                  0: 3422,
```
Haroenv, line 425: isn't there float logic in the rating menu too? do you want to test those cases here or in other tests?

packages/vue-instantsearch/src/components/RatingMenu.vue
```diff
@@ -26,41 +26,43 @@
...
-                v-if="full"
-                aria-hidden="true"
-                width="24"
-                height="24"
-                :class="[suit('starIcon'), suit('starIcon--full')]"
-                :key="n + '-full'"
-              >
-                <use xlink:href="#ais-RatingMenu-starSymbol" />
-              </svg>
+          <div>
```
Haroenv, line 29: could break people's style, but that's probably ok

packages/instantsearch.js/src/widgets/rating-menu/rating-menu.tsx
```diff
@@ -168,7 +168,7 @@ const renderer =
         templateProps={renderState.templateProps!}
         toggleRefinement={refine}
       >
-        <svg style="display:none;">
+        <svg xmlns="http://www.w3.org/2000/svg" style="display:none;">
```
Haroenv, line 171: is this actually needed?

Haroenv (approved):

packages/instantsearch.js/src/widgets/rating-menu/rating-menu.tsx
dhayab: I'd say no except for maybe really old browser versions, but I prefered to harmonize changes by adding instead of removing when possible.

tests/common/widgets/rating-menu/options.ts
dhayab: I see some logic in the connector but no tests all around. The doc also mentions that attribute values should be integers. Is it still relevant?

packages/instantsearch.js/src/widgets/rating-menu/rating-menu.tsx
Haroenv: it's not required since html5, so I'm happy to leave them out everywhere

tests/common/widgets/rating-menu/options.ts
dhayab: I looked more but still can't find the use case for floats in rating menu: we don't have fragments of stars to render 2.5/5 stars for example, and the count reflects the number of items in the dataset so it can only be an integer.

Is there something else I'm missing?

Haroenv (approved):

tests/common/widgets/rating-menu/options.ts
Haroenv: I think it's rounded, or maybe that was in numeric menu or range.

PR #6218, opened 2024-05-27
feat(typescript): allow generics for the Hit type in all hit-displaying connectors and widgets

Haroenv:
By leveraging `satisfies` instead of a type annotation, we can use the original typing of the widget and connector, and thus add additional generics that aren't already filled in before export.

Relevant widgets and connectors:
- hits
- infiniteHits
- geoSearch
- frequentlyBoughtTogether
- trendingItems
- lookingSimilar
- relatedProducts

Through this PR some other smaller things had to be changed around typing due to slight upgrade of api-extractor (that in the end didn't seem needed anymore). These seem like improvements to me.

There's a small downside that due to using `satisfies`, the type in the declaration is quite large, every single key of the function and widget object in the end get redeclared rather than using the XXXConnector type, as the XXXConnector can of course not have a non-specified generic (that's the whole problem this PR tries to solve).

I didn't touch connectAutocomplete (although it should work in theory too) as you could have one generic per index in a federated search context, and introducing a single generic now may make it harder later to support multiple generics.

It's not possible to provide a generic to the whole InstantSearch module (although I wish we could), but maybe in the future this problem could be solved with interface merging as well (users declare a Hit type, which merges with a Hit type from InstantSearch), but that seems like a bigger task than supporting generics properly here.

Review tips:
- use "hide whitespace" setting on GitHub to see much less diffs that are mainly caused by different indentation
- there's a lot of repetition, I recommend you ensure it's right in one of the widgets/connectors, then compare if it's the same for all the other widgets/connectors (there's only very little difference)

FX-2908


examples/js/getting-started/test.ts
```diff
@@ -0,0 +1,35 @@
+import algoliasearch from 'algoliasearch/lite';
```
Haroenv, line 1: to remove this example

packages/instantsearch.js/src/types/render-state.ts
```diff
@@ -65,5 +65,5 @@ export type RenderState = {
 
 export type WidgetRenderState<TWidgetRenderState, TWidgetParams> =
   TWidgetRenderState & {
-    widgetParams: TWidgetParams;
+    widgetParams: TWidgetParams | any;
```
Haroenv, line 68: this should be reverted, but connectHits will fail its types

packages/instantsearch.js/src/widgets/hits/hits.tsx
```diff
@@ -40,7 +41,7 @@ const withUsage = createDocumentationMessageGenerator({ name: 'hits' });
 const Hits = createHitsComponent({ createElement: h, Fragment });
 
 const renderer =
-  ({
+  <THit extends BaseHit = BaseHit>({
```
Haroenv, line 44: actually all places that have `extends BaseHits = BaseHit` probably should just be `= Record<string, unknown>` instead

packages/instantsearch.js/src/connectors/frequently-bought-together/__tests__/connectFrequentlyBoughtTogether-test.ts
```diff
@@ -40,6 +40,21 @@ describe('connectFrequentlyBoughtTogether', () => {
...
+    const customFrequentlyBoughtTogether = connectFrequentlyBoughtTogether<{
+      container: string;
+    }>(render, unmount);
+    const widget = customFrequentlyBoughtTogether({
+      container: '#container',
+      objectIDs: ['1'],
+    });
+
+    expect(widget.$$type).toBe('ais.frequentlyBoughtTogether');
+  });
```
dhayab, line 56: This is only properly asserted in type checking with `tsc`, right?

packages/instantsearch.js/src/types/index.ts
```diff
@@ -22,6 +22,3 @@ export * from './widget';
 export * from './ui-state';
 export * from './render-state';
 export * from './templates';
-
-// from specific widgets
-export type { IndexWidget } from '../widgets/index/index';
```
dhayab, line 27: Could this be potentially breaking?

packages/instantsearch.js/src/widgets/hits/defaultTemplates.ts
```diff
@@ -1,14 +1,14 @@
 import { omit } from '../../lib/utils';
 
-import type { HitsTemplates } from './hits';
+import { HitsTemplates } from './hits';
```
dhayab, line 3: Why?

packages/instantsearch.js/src/widgets/index.ts
```diff
@@ -26,7 +26,7 @@ export { default as geoSearch } from './geo-search/geo-search';
 export { default as hierarchicalMenu } from './hierarchical-menu/hierarchical-menu';
 export { default as hits } from './hits/hits';
 export { default as hitsPerPage } from './hits-per-page/hits-per-page';
-export { default as index } from './index/index';
+export { default as index, type IndexWidget } from './index/index';
```
dhayab, line 29: Are you confident import statements mixing types and values work in target frameworks? I remember having issues in the past with this.

packages/instantsearch.js/src/widgets/frequently-bought-together/frequently-bought-together.tsx
```diff
@@ -169,48 +182,49 @@ export type FrequentlyBoughtTogetherWidget = WidgetFactory<
...
-        transformItems,
-      }),
-      $$widgetType: 'ais.frequentlyBoughtTogether',
-    };
+    }),
+    $$widgetType: 'ais.frequentlyBoughtTogether',
   };
-
-export default frequentlyBoughtTogether;
+});
```
dhayab, line 230:
```suggestion
} satisfies FrequentlyBoughtTogetherWidget);
```

packages/instantsearch.js/src/connectors/frequently-bought-together/__tests__/connectFrequentlyBoughtTogether-test.ts
Haroenv: yes, we don't really have type-only tests, but tsc is asserting that there's no errors here

packages/instantsearch.js/src/types/index.ts
Haroenv: it's moved to the "widgets" file, still exported from here too through that `*`

packages/instantsearch.js/src/widgets/hits/defaultTemplates.ts
Haroenv: eslint complains that it's needed for "decorator metadata", probably as a false positive. Will ignore the rule instead indeed

packages/instantsearch.js/src/widgets/index.ts
Haroenv: sure, removed them just in case, probably some older tools don't remove them properly? in a major we can definitely be more consistent though (and probably export all types, but I'd do that in a separate PR)

packages/instantsearch.js/src/types/index.ts
dhayab: Ah yes, I see it now.

dhayab (approved):
LGTM 🙇

aymeric-giraudet (approved):

PR #6223, opened 2024-06-03
fix(next-app-router): set nonce in injected script when provided

dhayab:
**Summary**

Customers who follow [Next.js' guide for implementing CSP rules](https://nextjs.org/docs/app/building-your-application/configuring/content-security-policy) to their application can define a nonce that flags inline scripts as allowed. This is properly handled with Next.js' own inline scripts, but not yet with **react-instantsearch-nextjs**.

This PR retrieves the nonce value and applies it to the injected script tag added by `<InitializePromise>`.

**Result**

InstantSearch does not trigger an error in Next.js applications where a middleware define CSP rules and apply nonces.

CR-6159

### DIFF
--- a/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
+++ b/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
@@ -1,4 +1,5 @@
 import { safelyRunOnBrowser } from 'instantsearch.js/es/lib/utils';
+import { headers } from 'next/headers';
 import React, { useEffect, useRef } from 'react';
 import {
   InstantSearch,
@@ -72,7 +73,9 @@ This message will only be displayed in development mode.`
     <InstantSearchRSCContext.Provider value={promiseRef}>
       <InstantSearchSSRProvider initialResults={initialResults}>
         <InstantSearch {...instantSearchProps} routing={routing}>
-          {!initialResults && <InitializePromise />}
+          {!initialResults && (
+            <InitializePromise nonce={headers().get('x-nonce') || undefined} />
+          )}
           {children}
           {!initialResults && <TriggerSearch />}
         </InstantSearch>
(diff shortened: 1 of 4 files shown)


yarn.lock
dhayab, line 1: I left this diff as it might have been missed in previous contributions to the codebase.

packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
```diff
@@ -72,7 +73,9 @@ This message will only be displayed in development mode.`
     <InstantSearchRSCContext.Provider value={promiseRef}>
       <InstantSearchSSRProvider initialResults={initialResults}>
         <InstantSearch {...instantSearchProps} routing={routing}>
-          {!initialResults && <InitializePromise />}
+          {!initialResults && (
+            <InitializePromise nonce={headers().get('x-nonce') || undefined} />
```
aymeric-giraudet, line 77: `nonce` could be `string | null` to avoid an explicit `undefined`

aymeric-giraudet (approved):

packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
dhayab: It still needs to be `string | undefined` in the injected script. I was ambivalent between making the fallback here or in InitializePromise.

Haroenv (approved):

RobinGiel:
After this change I'm getting this error: 
image

anthowm:
> After this change I'm getting this error: image

Same here 
image

dhayab:
Hi, we have a fix scheduled for release today. For now, you can pin `react-instantsearch-nextjs` to the previous working version.

dhayab:
Hi, you can now update to [react-instantsearch-nextjs@0.3.2](https://github.com/algolia/instantsearch/releases/tag/react-instantsearch-nextjs%400.3.2) which includes a fix for this issue.

PR #6224, opened 2024-06-03
test(infiniteHits): add common test suite

sarahdayan:
This reworks the tests for the `infiniteHits` widget to move them in the common test suite and reduce flavor-specific tests.

https://algolia.atlassian.net/browse/FX-2514


tests/common/widgets/infinite-hits/index.ts
```diff
@@ -22,5 +23,6 @@ export function createInfiniteHitsWidgetTests(
   describe('InfiniteHits widget common tests', () => {
     createOptimisticUiTests(setup, { act, skippedTests });
     createInsightsTests(setup, { act, skippedTests });
```
Haroenv, line 25: the insights tests are failing because more infinite hits are rendered now

sarahdayan: Yup that's fixed.

sarahdayan: Yup that's fixed.

PR #6233, opened 2024-06-10
test(rangeInput): add common test suite

sarahdayan:
This reworks the tests for the `rangeInput` widget to move them in the common test suite and reduce flavor-specific tests.

https://algolia.atlassian.net/browse/FX-2518


tests/common/widgets/range-input/options.ts
```diff
@@ -0,0 +1,220 @@
...
+      });
+
+      ['min', 'max'].forEach((target) => {
+        expect(
+          document.querySelector(`.ais-RangeInput-input--${target}`)
+        ).toHaveAttribute('step', '0.01');
+      });
+    });
+
+    test('refines on submit', async () => {
```
Haroenv, line 147: not sure if worth testing, but we could add a test that only refines min or max, not both, to check it doesn't accidentally refine 0 or something.

Haroenv (approved):

tests/common/widgets/range-input/options.ts
```diff
@@ -0,0 +1,220 @@
...
+  wait,
+} from '@instantsearch/testutils';
+import userEvent from '@testing-library/user-event';
+
+import type { RangeInputWidgetSetup } from '.';
+import type { TestOptions } from '../../common';
+
+function normalizeSnapshot(html: string) {
+  // React InstantSearch adds a noRefinement modifier on the list.
+  // @MAJOR: Remove this once React InstantSearch aligns with spec.
```
dhayab, line 17: Would it make sense to add the modifier in JS and Vue instead?

dhayab (approved):

tests/common/widgets/range-input/options.ts
sarahdayan: We could, but to make it complete we would also need to introduce the `noRefinementRoot` className. Shouldn't we do this separately during our effort to unify?

dhayab: Yep I'm fine with this, thanks for looking!

tests/common/widgets/range-input/options.ts
Haroenv: yes that's fine, although I meant to do that in a separate test

Haroenv (approved):

PR #6236, opened 2024-06-14
feat: Add support for hiding facet values

thomasbritton:
**Summary**

Adds support for hiding facet values if set in the configuration.
JIRA ticket: https://algolia.atlassian.net/browse/EMERCH-1438

**Result**

### DIFF
--- a/packages/algoliasearch-helper/src/SearchResults/index.js
+++ b/packages/algoliasearch-helper/src/SearchResults/index.js
@@ -806,6 +806,20 @@ function vanillaSortFn(order, data) {
  */
 
 /**
+ * Hides facet values from array
+ * @param {Array} facetValues the facet values
+ * @param {Array} hide facet values
+ * @returns {Array} the facet values with with hidden values removed
+ */
+function removeHiddenFacetValues(facetValues, hide) {
+  facetValues = facetValues.filter(function (item) {
+    return hide.indexOf(item.name) === -1;
+  });
+
+  return facetValues;
+}
+
+/**
  * Sorts facet arrays via their facet ordering
  * @param {Array} facetValues the values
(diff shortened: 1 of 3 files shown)


packages/algoliasearch-helper/src/SearchResults/index.js
```diff
@@ -825,6 +840,10 @@ function sortViaFacetOrdering(facetValues, facetOrdering) {
     return acc;
   }, {});
 
+  if (hide.length > 0) {
+    facetValues = removeHiddenFacetValues(facetValues, hide);
+  }
+
   facetValues.forEach(function (item) {
     // hierarchical facets get sorted using their raw name
     var name = item.path || item.name;
```
Haroenv, line 849: I'm ok with the code you're suggesting, although I think it would be interesting to consider doing the code here and avoiding another iteration:

```suggestion
    var name = item.path || item.name;
    var hidden = hide.indexOf(name) > -1;
    if (!hidden && reverseOrder[name] !== undefined) {
      orderedFacets[reverseOrder[name]] = item;
    } else if (!hidden) {
      remainingFacets.push(item);
    }
```

Not sure if it's worth the difference, but the code doesn't look much harder to me, and should be more efficient in theory.

Haroenv (approved):

examples/js/getting-started/src/app.js
```diff
@@ -11,12 +11,12 @@ import {
 } from 'instantsearch.js/es/widgets';
 
 const searchClient = algoliasearch(
-  'latency',
-  '6be0576ff61c053d5f9a3225e2a90f76'
+  'F4T6CUV2AH',
```
Haroenv, line 14: changes in the example should be reverted before merge :)

PR #6239, opened 2024-06-17
feat(facets): allow hiding values through renderingContent

thomasbritton:
**Summary**

Adds support for hiding facet values if set in the configuration.
JIRA ticket: https://algolia.atlassian.net/browse/EMERCH-1438

**Result**

If you have set up facets to be "hidden", they will not display. This does not increment the maxValuesPerFacet setting, so it's possible that you receive less values than the limit if you hide many values.


Haroenv (approved):
👌

sarahdayan (approved):

PR #6240, opened 2024-06-18
test(dynamic-widgets): common test suite

aymeric-giraudet:
**Summary**

#### [FX-2510](https://algolia.atlassian.net/browse/FX-2510)

**Result**

Tried to test as much as possible, but we're limited due to the implems being very different, notably the lack of `fallback` mechanism in Vue.

Still these test the order to render widgets, whether a widget should render or not, and makes sure that certain params are correctly forwarded to the search request.

I didn't remove any pre-existing tests for now.

[FX-2510]: https://algolia.atlassian.net/browse/FX-2510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react-instantsearch/src/__tests__/common-widgets.test.tsx
+++ b/packages/react-instantsearch/src/__tests__/common-widgets.test.tsx
@@ -380,6 +380,12 @@ const testSetups: TestSetupsMap<TestSuites> = {
           <DynamicWidgets {...widgetParams}>
             <RefinementList attribute="brand" />
             <Menu attribute="category" />
+            <HierarchicalMenu
+              attributes={[
+                'hierarchicalCategories.lvl0',
+                'hierarchicalCategories.lvl1',
+              ]}
+            />
           </DynamicWidgets>
         </div>
         <GlobalErrorSwallower />
--- a/packages/react-instantsearch/src/__tests__/common-widgets.test.tsx
+++ b/packages/react-instantsearch/src/__tests__/common-widgets.test.tsx
@@ -30,6 +30,7 @@ import {
   TrendingItems,
   LookingSimilar,
   PoweredBy,
+  DynamicWidgets,
 } from '..';
 
 import type { TestOptionsMap, TestSetupsMap } from '@instantsearch/tests';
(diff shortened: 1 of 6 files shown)


tests/common/widgets/dynamic-widgets/options.ts
```diff
@@ -0,0 +1,168 @@
...
+              });
+            })
+          )
+        );
+      }),
+    });
+
+    beforeEach(() => {
+      searchClient.search.mockClear();
+    });
```
sarahdayan, line 49: Nit, but in other tests we wrap this into a `createMockedSearchClient` function and instantiate a new client in each test. This would work the same, but the benefits are that you don't need to clear between tests, and you can keep the mock at the end of the file so it's not the first thing you see when reading the tests.

tests/common/widgets/dynamic-widgets/options.ts
```diff
@@ -0,0 +1,168 @@
...
+          name: /tv/i,
+        })
+      ).toBeInTheDocument();
+
+      expect(
+        within(dynamicWidgets[1] as HTMLElement).getByRole('checkbox', {
+          name: /samsung/i,
+        })
+      ).toBeInTheDocument();
+    });
```
sarahdayan, line 85: Could we test against an inline snapshot instead? It makes it easier to understand what the full output is.

Haroenv: won't the output be huge?

packages/react-instantsearch/src/__tests__/common-widgets.test.tsx
```diff
@@ -372,6 +373,19 @@ const testSetups: TestSetupsMap<TestSuites> = {
   createMenuSelectWidgetTests() {
     throw new Error('MenuSelect is not supported in React InstantSearch');
   },
+  createDynamicWidgetsWidgetTests({ instantSearchOptions, widgetParams }) {
+    render(
+      <InstantSearch {...instantSearchOptions}>
+        <div className="ais-DynamicWidgets">
+          <DynamicWidgets {...widgetParams}>
+            <RefinementList attribute="brand" />
+            <Menu attribute="category" />
```
Haroenv, line 382: let's put a hierarchical menu in the static widgets too? that way it tests the "attributes" fetching

tests/common/widgets/dynamic-widgets/options.ts
aymeric-giraudet: Yes, also we can't because the output is very different between vanilla and React as we don't wrap the widgets into a div in React

Haroenv (approved):
this is a good start, but we probably want to add more tests here after

PR #6243, opened 2024-06-18
feat(infinite-hits): add top banner to InfiniteHits widget

taylorcjohnson:
**Summary**

A series of PRs recently added a top banner to the `Hits` widget (visibility is controlled by rules that update `renderingContent` for a given query). This PR implements the same feature, but for the `InfiniteHits` widget.

This was done by:

1. updating `connectInfiniteHits` (previous PR)
2. updating `InfiniteHits` JS component
3. updating `infinite-hits` JS widget
4. updating `InfiniteHits` React component
5. updating `InfiniteHits` React widget
6. updating `satellite.scss` and `algolia.scss`

Related tickets:

- https://algolia.atlassian.net/browse/EMERCH-1501
- https://algolia.atlassian.net/browse/EMERCH-1502
- https://algolia.atlassian.net/browse/EMERCH-1503
- https://algolia.atlassian.net/browse/EMERCH-1505

**Result**

JS Hits, default banner:
image

JS InfiniteHits, default banner:
image

JS InfiniteHits, default banner, no results:
image

React Hits, default banner:
image

React InfiniteHits, default banner:
image

React InfiniteHits, default banner, no results:
image

### DIFF
--- a/packages/instantsearch.js/src/components/InfiniteHits/InfiniteHits.tsx
+++ b/packages/instantsearch.js/src/components/InfiniteHits/InfiniteHits.tsx
@@ -38,6 +38,45 @@ export type InfiniteHitsProps = {
   banner?: Banner;
 };
 
+const DefaultBanner = ({
+  banner,
+  classNames,
+}: {
+  banner: Banner;
+  classNames: Pick<
+    InfiniteHitsCSSClasses,
+    'bannerRoot' | 'bannerLink' | 'bannerImage'
+  >;
+}) => {
+  if (!banner.image.urls[0].url) {
+    return null;
+  }
+
+  return (
+    <aside className={cx(classNames.bannerRoot)}>
+      {banner.link ? (
+        <a
+          className={cx(classNames.bannerLink)}
+          href={banner.link.url}
+          target={banner.link.target}
+        >
+          <img
+            className={cx(classNames.bannerImage)}
+            src={banner.image.urls[0].url}
+            alt={banner.image.title}
+          />
+        </a>
+      ) : (
+        <img
+          className={cx(classNames.bannerImage)}
+          src={banner.image.urls[0].url}
+          alt={banner.image.title}
+        />
+      )}
+    </aside>
+  );
+};
+
 const InfiniteHits = ({
   results,
   hits,
@@ -60,27 +99,28 @@ const InfiniteHits = ({
 
   if (results.hits.length === 0) {
     return (
-      <div className={cssClasses.root}>
-        {banner && templateProps.templates.banner ? (
-          <Template
-            {...templateProps}
-            templateKey="banner"
-            rootTagName="aside"
-            rootProps={{
-              className: cssClasses.bannerRoot,
-            }}
-            data={banner}
-          />
-        ) : (
-          <aside>default component goes here</aside>
-        )}
+      <div
+        className={cx(cssClasses.root, cssClasses.emptyRoot)}
+        onClick={handleInsightsClick}
+      >
+        {banner &&
+          (templateProps.templates.banner ? (
+            <Template
+              {...templateProps}
+              templateKey="banner"
+              rootTagName="fragment"
+              rootProps={{
+                className: cssClasses.bannerRoot,
+              }}
+              data={banner}
+            />
+          ) : (
+            <DefaultBanner banner={banner} classNames={cssClasses} />
+          ))}
         <Template
           {...templateProps}
           templateKey="empty"
-          rootProps={{
-            className: cssClasses.emptyRoot,
-            onClick: handleInsightsClick,
-          }}
+          rootTagName="fragment"
           data={results}
         />
       </div>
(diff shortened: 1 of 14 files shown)


packages/instantsearch.js/src/components/InfiniteHits/InfiniteHits.tsx
```diff
@@ -58,15 +99,31 @@ const InfiniteHits = ({
...
+          ) : (
+            <DefaultBanner banner={banner} classNames={cssClasses} />
+          ))}
+        <Template
+          {...templateProps}
+          templateKey="empty"
+          rootTagName="fragment"
+          data={results}
+        />
+      </div>
```
taylorcjohnson, line 126: The empty template fragment is being hoisted above the banner (even though the banner is above it in the DOM here). Any suggestions for forcing the expected layout (see Hits example in image below) without adding additional DOM nodes?

image

packages/instantsearch.js/src/components/InfiniteHits/InfiniteHits.tsx
```diff
@@ -13,11 +13,11 @@ import type {
   InfiniteHitsCSSClasses,
   InfiniteHitsTemplates,
 } from '../../widgets/infinite-hits/infinite-hits';
-import type { SearchResults } from 'algoliasearch-helper';
+import type { Banner, SearchResults } from 'algoliasearch-helper';
 
 export type InfiniteHitsComponentCSSClasses =
   ComponentCSSClasses<InfiniteHitsCSSClasses>;
-export type InfiniteHitsComponentTemplates = Required<InfiniteHitsTemplates>;
+export type InfiniteHitsComponentTemplates = InfiniteHitsTemplates;
```
taylorcjohnson, line 20: This reassignment probably isn't needed anymore - unless we really need to keep the `Required<>` wrapper.

Haroenv (commented):
code looks good, but this should be tested in common test suite, not in the flavour specific tests :)

taylorcjohnson:
> code looks good, but this should be tested in common test suite, not in the flavour specific tests :)

Hey @Haroenv - sorry, this is the first I am learning about the common test suite. Does that refer to everything in `tests/common/widgets/infinite-hits`?

If so, do we need to update the tests for `hits` as well? I did not touch those tests when adding the banner to the `hits` widget.

And are you saying that I should remove the banner related tests for both `hits` and `infinite-hits` in the flavor specific tests? Or that the tests in `/common` _and_ the flavor specific tests should be updated to test for the banner?

Haroenv:
Ah sorry I didn't notice the tests weren't done in cts for the hits implementation. Ideally yes you'd probably add a new file (suite) inside the common test suite, and also for hits. the flavour specific tests can be removed, unless it's something that exists only in one of the flavours

taylorcjohnson:
@Haroenv thanks - I'll work on adding tests here for `infinite-hits` (and `hits`, unless you'd prefer that in a separate PR). I _think_ we should leave the tests in the specific flavors since the banner feature is only supported in JS and React for right now (though the tentative plan is to add it for Vue as well in the near future). What do you think?

Haroenv:
all tests should go exclusively in CTS when supported in more than one flavour, or planned to be. Thanks! Reach out if you are stuck somewhere

aymeric-giraudet (approved):
Thanks for the common test suite, looks good to me ! 👍

packages/instantsearch.js/src/components/InfiniteHits/InfiniteHits.tsx
taylorcjohnson: @aymeric-giraudet or @Haroenv - any thoughts on this?

aymeric-giraudet: That's odd @taylorcjohnson, is it related to CSS ?
How does it look like without our CSS stylesheets ? Maybe this should just be solved within `instantsearch.css`

packages/instantsearch.js/src/components/InfiniteHits/InfiniteHits.tsx
aymeric-giraudet: It's fine to keep if other widgets do this I guess

packages/instantsearch.js/src/components/InfiniteHits/InfiniteHits.tsx
taylorcjohnson: I don't believe so, @aymeric-giraudet - I removed the css to verify, but the DOM still shows a difference.

image

Haroenv: I don't think it's the right solution, but what happens if you put the no results template in a div or give it a root tag name?

taylorcjohnson: Then it works as expected 😢 (either wrapping the empty template component in a `div` or changing the `rootTagName` to something else (like a `div`).

aymeric-giraudet: Ah I think this may be due to the `RawHtml` component which is a hack for supporting Hogan with fragments, I'll check for a fix tomorrow morning :/

taylorcjohnson: Ah ok, I was hoping I was just missing some background context somewhere. If you can take a look I'd really appreciate it otherwise I can attempt as well.

Haroenv (approved):
good for me, even though the order on empty isn't right. Not sure that matters much for now, we may even consider leaving it as-is, as I assume the banner on empty won't be that popular together with an empty template.

PR #6251, opened 2024-06-26
fix(algoliasearch): v5 does not default threshold to 0

aymeric-giraudet:
**Summary**

#### [FX-2926](https://algolia.atlassian.net/browse/FX-2926)

**Result**

The v5 of the API client does not do extra logic like it did before for ease of maintenance, so for now we have to do this on our end.

[FX-2926]: https://algolia.atlassian.net/browse/FX-2926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -30,7 +30,7 @@
     },
     {
       "path": "packages/vue-instantsearch/vue3/umd/index.js",
-      "maxSize": "67.75 kB"
+      "maxSize": "68 kB"
     },
     {
       "path": "packages/vue-instantsearch/vue2/cjs/index.js",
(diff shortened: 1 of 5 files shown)


Haroenv (approved):
you can increase all the thresholds to have more margin, this tiny change didn't do anything appreciable to bundle size (vue 3 is the one that actually fails). Can also be done in a separate PR

Haroenv:
a couple tests also need to change, just update snapshot to take this default in account

dhayab (approved):

Haroenv (approved):

PR #6252, opened 2024-06-26
fix(pagination): any page outside the range is also first or last

Haroenv:
**Summary**

If the page is outside the range, it should behave like the 0th or last page. This disables the relevant buttons and links, which makes more sense UX-wise, and avoids creating pages that don't exist when you start from an invalid URL.

[CR-6249]

**Result**

- Page >= limit page is considered last page
- negative page or page 0 is considered first page

[CR-6249]: https://algolia.atlassian.net/browse/CR-6249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/connectors/pagination/Paginator.ts
+++ b/packages/instantsearch.js/src/connectors/pagination/Paginator.ts
@@ -65,7 +65,7 @@ class Paginator {
   }
 
   public isFirstPage() {
-    return this.currentPage === 0;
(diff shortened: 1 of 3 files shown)


aymeric-giraudet (approved):
Makes sense !
Idk why the bundlesize step is acting up though 😥

PR #6254, opened 2024-06-27
fix(RawHtml): force component to remount

aymeric-giraudet:
**Summary**

This fixes https://github.com/algolia/instantsearch/pull/6243#discussion_r1649444124

**Result**

We didn't see that with Hits because [all templates remount regardless, as they are not memoized](https://github.com/algolia/instantsearch/blob/762e6335692bc0f05243edbf168f1be0b8d31f0c/packages/instantsearch.js/src/widgets/hits/hits.tsx#L83-L93).
There might be a way to solve this only with `componentDidUpdate` but I think there would be more duplicate code, probably fine to keep as is until we get rid of Hogan.


Haroenv (commented):
is it possible to add a test for this? shouldn't the order of the empty/banner template be fixed now?

aymeric-giraudet:
Just added one @Haroenv

Haroenv (approved):

PR #6257, opened 2024-06-28
chore(examples): set instantsearchjs monorepo version in vue examples

dhayab:
**Summary**

It seems during last release the vue examples were not provided the right version for instantsearch.js, making the package an actual dependency instead of being linked from the monorepo.

### DIFF
--- a/examples/vue/default-theme/package.json
+++ b/examples/vue/default-theme/package.json
@@ -10,7 +10,7 @@
   "dependencies": {
     "algoliasearch": "4.23.2",
     "core-js": "2",
-    "instantsearch.js": "4.72.1",
+    "instantsearch.js": "4.72.2",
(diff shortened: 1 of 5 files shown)


Haroenv (approved):
ah yes that's because I didn't want to trigger a vue release but didn't realize this would have this effect. Won't we need to remove InstantSearch.js from the lock file now or was it never added?

dhayab:
> ah yes that's because I didn't want to trigger a vue release but didn't realize this would have this effect. Won't we need to remove InstantSearch.js from the lock file now or was it never added?

It was not added or at least not versioned. I noticed when it showed a diff in my project.

Haroenv:
yes, the last version had to be done manually since only a chore happened but I still wanted to release it :)

PR #6264, opened 2024-07-02
poc: blocking middleware

dhayab:
this outlines a potential way for middlewares to block the first query of instantsearch, to allow it to run prerequisite tasks

### DIFF
--- a/examples/js/getting-started/src/app.js
+++ b/examples/js/getting-started/src/app.js
@@ -45,7 +45,7 @@ search.addWidgets([
     hitsPerPage: 8,
   }),
   panel({
-    templates: { header: () => 'brand' },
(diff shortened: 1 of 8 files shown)


Haroenv:
in React InstantSearch start is called immediately to ensure widgets can render with the fake results. Is that still possible here or not?

dhayab:
> in React InstantSearch start is called immediately to ensure widgets can render with the fake results. Is that still possible here or not?

Good question, it should be, although in a quick test with the current changes, there are issues when rerendering later on, that should be further assessed.

Haroenv:
I think we don't have a use case for this anymore @dhayab ?

dhayab:
Yes, I'll clean this up.

PR #6268, opened 2024-07-04
fix(algoliasearch): correctly retrieve headers for v5

aymeric-giraudet:
**Summary**

#### [FX-2928](https://algolia.atlassian.net/browse/FX-2928)

**Result**

We fallback to `baseHeaders` and `baseQueryParameters` if `headers` or `queryParameters` do not exist.
Updated all remaining peer deps to mark compatibility with v5.

[FX-2928]: https://algolia.atlassian.net/browse/FX-2928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/lib/utils/__tests__/getAppIdAndApiKey-test.ts
+++ b/packages/instantsearch.js/src/lib/utils/__tests__/getAppIdAndApiKey-test.ts
@@ -1,5 +1,6 @@
 import algoliasearchV4 from 'algoliasearch';
 import algoliasearchV3 from 'algoliasearch-v3';
+import { liteClient as algoliasearchV5 } from 'algoliasearch-v5/lite';
 
 import { getAppIdAndApiKey } from '../getAppIdAndApiKey';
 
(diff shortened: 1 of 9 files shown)


packages/instantsearch.js/src/lib/utils/__tests__/getAppIdAndApiKey-test.ts
```diff
@@ -20,4 +20,34 @@ describe('getAppIdAndApiKey', () => {
     expect(appId).toEqual(APP_ID);
     expect(apiKey).toEqual(API_KEY);
   });
+
+  it("fallsback to baseHeaders if headers doesn't exist", () => {
+    const searchClient = {
```
Haroenv, line 25: Could we install v5 like the other versions too? That way it gives full confidence IMO

Haroenv (commented):
Overall good, but we can add a v5 alias dependency like we did for v3

Haroenv (approved):

PR #6270, opened 2024-07-08
fix(client): support v5 fully

dhayab:
**Summary**

This PR updates all algoliasearch dependencies in examples to algoliasearch@5 and ensures full compatibility locally with algoliasearch v5

FX-2922

fixes https://github.com/algolia/algoliasearch-client-javascript/issues/1537
fixes https://github.com/algolia/instantsearch/issues/6329

### DIFF
--- a/examples/js/calendar-widget/app.js
+++ b/examples/js/calendar-widget/app.js
@@ -1,5 +1,5 @@
 /* global moment Calendar $ */
-import algoliasearch from 'algoliasearch/lite';
+import { liteClient as algoliasearch } from 'algoliasearch/lite';
 import instantsearch from 'instantsearch.js';
 import { connectRange } from 'instantsearch.js/es/connectors';
 import { searchBox, hits } from 'instantsearch.js/es/widgets';
--- a/examples/js/e-commerce-umd/package.json
+++ b/examples/js/e-commerce-umd/package.json
@@ -9,7 +9,7 @@
   },
   "browserslist": "firefox 68, chrome 78, IE 11",
   "dependencies": {
-    "algoliasearch": "4.23.2",
+    "algoliasearch": "5.0.0",
     "instantsearch.js": "4.73.4"
   },
   "devDependencies": {
--- a/packages/instantsearch.js/src/connectors/refinement-list/__tests__/connectRefinementList-test.ts
+++ b/packages/instantsearch.js/src/connectors/refinement-list/__tests__/connectRefinementList-test.ts
@@ -1598,6 +1598,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
 
       const helper = jsHelper(
         createSearchClient({
+          // @ts-ignore v5 expects `search({ type: 'facet' })` only
           searchForFacetValues() {
             return Promise.resolve([
               {
@@ -1719,6 +1720,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
 
       const helper = jsHelper(
         createSearchClient({
+          // @ts-ignore v5 expects `search({ type: 'facet' })` only
           searchForFacetValues() {
             return Promise.resolve([
               {
@@ -1801,6 +1803,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
 
       const helper = jsHelper(
         createSearchClient({
+          // @ts-ignore v5 expects `search({ type: 'facet' })` only
           searchForFacetValues() {
             return Promise.resolve([
               {
@@ -1911,6 +1914,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
 
       const helper = jsHelper(
         createSearchClient({
+          // @ts-ignore v5 expects `search({ type: 'facet' })` only
           searchForFacetValues() {
             return Promise.resolve([
               {
@@ -2029,6 +2033,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
 
       const helper = jsHelper(
         createSearchClient({
+          // @ts-ignore v5 expects `search({ type: 'facet' })` only
           searchForFacetValues() {
             return Promise.resolve([
               {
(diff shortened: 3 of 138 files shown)


Haroenv (approved):
most changes are copy-paste dependency changes, bar some type changes in helper to make it properly compatible, as well as some smaller consistency Hit/AlgoliaHit changes (because recommend is now correctly typed without position and query id)

packages/instantsearch.js/src/connectors/trending-items/__tests__/connectTrendingItems-test.ts
```diff
@@ -129,8 +129,9 @@ describe('connectTrendingItems', () => {
         new RecommendParameters().addTrendingItems({
           // @ts-expect-error
           $$id: widget.$$id,
-          facetName: undefined,
-          facetValue: undefined,
+          // v5 expects only string values for facetName and facetValue
```
dhayab, line 132: Isn't it wrong though? IIRC you could choose not to provide `facetName` and `facetValue` .

tests/mocks/createSearchClient.ts
```diff
@@ -20,6 +20,7 @@ export const createSearchClient = (
       )
     )
   ),
+  // @ts-ignore v5 does not have this method, but it's easier to have it here. In a future version we can remove this method and its usages to search({ type: 'facet })
```
dhayab, line 23:
```suggestion
  // @ts-ignore v5 does not have this method, but it's easier to have it here. In a future version we can replace this method and its usages with search({ type: 'facet })
```

?

packages/instantsearch.js/src/connectors/trending-items/__tests__/connectTrendingItems-test.ts
Haroenv: you're right, I forgot I already noted that and @shortcuts fixed it before the 5.0.0 release :)

dhayab:
✅ Approved (I can't do it properly as I'm the creator of the PR).

shortcuts (approved):
fully trusting @dhayab

packages/instantsearch.js/src/connectors/refinement-list/__tests__/connectRefinementList-test.ts
```diff
@@ -1598,6 +1598,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/refinement-
 
       const helper = jsHelper(
         createSearchClient({
+          // @ts-ignore v5 does not have this method, but it's easier to have it here. In a future version we can replace this method and its usages with search({ type: 'facet })
```
shortcuts, line 1601: we have a wrapped named `searchForFacets` purely for typing purposes if that helps

packages/instantsearch.js/src/connectors/voice-search/__tests__/connectVoiceSearch-test.ts
```diff
@@ -472,6 +472,7 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/voice-searc
         new SearchParameters({
           ignorePlurals: true,
           removeStopWords: true,
+          // @ts-ignore we send optionalWords as a string
           optionalWords: 'query',
```
shortcuts, line 476: are both possible? if so, we can fix the spec

packages/instantsearch.js/src/connectors/voice-search/connectVoiceSearch.ts
```diff
@@ -106,14 +106,15 @@ const connectVoiceSearch: VoiceSearchConnector = function connectVoiceSearch(
               const queryLanguages = language
                 ? [language.split('-')[0]]
                 : undefined;
+              // @ts-ignore queryLanguages is allowed to be a string, not just an array
```
shortcuts, line 109: same here if you confirm I can fix the spec

examples/js/calendar-widget/app.js
```diff
@@ -1,5 +1,5 @@
 /* global moment Calendar $ */
-import algoliasearch from 'algoliasearch/lite';
+import { liteClient as algoliasearch } from 'algoliasearch/lite';
```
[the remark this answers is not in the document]
Haroenv: if you're doing any change, adding the default export would be very handy

packages/instantsearch.js/src/connectors/voice-search/__tests__/connectVoiceSearch-test.ts
Haroenv: yes, it's allowed in the api, but I think that's only because any array is also allowed to be a string

packages/instantsearch.js/src/connectors/voice-search/connectVoiceSearch.ts
Haroenv: same I believe it's allowed in the api just because any array is also allowed to be a string, I'd rather leave it to be an array always and eventually change it in InstantSearch

examples/js/e-commerce-umd/package.json
```diff
@@ -9,7 +9,7 @@
   },
   "browserslist": "firefox 68, chrome 78, IE 11",
   "dependencies": {
-    "algoliasearch": "4.23.2",
+    "algoliasearch": "5.0.0",
```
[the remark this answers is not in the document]
Haroenv: just an example so we can update separately of this pr

packages/instantsearch.js/src/connectors/refinement-list/__tests__/connectRefinementList-test.ts
Haroenv: the comment is more for the /lite in v5 which doesn't have sffv anymore, and the order which is checked in InstantSearch is `client.sffv, client.initIndex?.sffv, client.search`, and so as we're in v5 here technically we should only mock `client.search`, but that's ok

packages/instantsearch.js/src/connectors/voice-search/__tests__/connectVoiceSearch-test.ts
shortcuts: ok ok so let's keep it like that then

PR #6271, opened 2024-07-08
chore: types running with v4 locally, but also v4 and v3

Haroenv:
**Summary**

- detect v5 correctly in types 
- use v5 client as base version to avoid conflicts (which customers won't have unless they install both versions like us)

**Result**

all tests green

### DIFF
--- a/tests/mocks/createAPIResponse.ts
+++ b/tests/mocks/createAPIResponse.ts
@@ -3,6 +3,7 @@ import type {
   SearchResponses,
   SearchForFacetValuesResponse,
   RecommendResponse,
+  RecommendResponses,
 } from 'instantsearch.js';
 
 export const defaultRenderingContent: SearchResponse<any>['renderingContent'] =
@@ -88,14 +89,6 @@ export const createSFFVResponse = (
   ...args,
 });
 
-export const createRecommendResponse = (
-  requests: any[]
-): { results: Array<RecommendResponse<any>> } => {
-  return {
-    results: requests.map(createSingleSearchResponse),
-  } as unknown as { results: Array<RecommendResponse<any>> };
-};
-
 export const createSingleRecommendResponse = (
   subset: Partial<RecommendResponse<any>> = {}
 ): RecommendResponse<any> => {
@@ -125,5 +118,13 @@ export const createSingleRecommendResponse = (
     exhaustiveNbHits,
     exhaustiveFacetsCount,
     ...rest,
+  };
+};
+
+export const createRecommendResponse = (
+  requests: any[]
+): RecommendResponses<any> => {
+  return {
+    results: requests.map(createSingleRecommendResponse),
   };
 };
--- a/tests/mocks/createAlgoliaSearchClient.ts
+++ b/tests/mocks/createAlgoliaSearchClient.ts
@@ -27,27 +27,10 @@ type OverrideKeys<TTarget, TOptions> = TOptions extends Record<string, never>
   ? TTarget
   : Omit<TTarget, keyof TOptions> & TOptions;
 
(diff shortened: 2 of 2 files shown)


packages/instantsearch.js/src/connectors/trending-items/__tests__/connectTrendingItems-test.ts
```diff
@@ -129,8 +129,9 @@ describe('connectTrendingItems', () => {
         new RecommendParameters().addTrendingItems({
           // @ts-expect-error
           $$id: widget.$$id,
-          facetName: undefined,
-          facetValue: undefined,
+          // v5 expects only string values for facetName and facetValue
+          facetName: undefined as unknown as string,
+          facetValue: undefined as unknown as string,
```
Haroenv, line 134: https://github.com/algolia/api-clients-automation/pull/3355 after this is released we can amend to remove the cast. Still find for now as it's just a test though

packages/instantsearch.js/src/connectors/frequently-bought-together/connectFrequentlyBoughtTogether.ts
```diff
@@ -31,7 +29,7 @@ export type FrequentlyBoughtTogetherRenderState<
   /**
    * The matched recommendations from Algolia API.
    */
-  items: Array<Hit<THit>>;
+  items: Array<AlgoliaHit<THit>>;
```
Haroenv, line 32: all recommend changes Hit -> AlgoliaHit is because __position and __queryId aren't actually present in these connectors

tests/mocks/createAPIResponse.ts
```diff
@@ -89,7 +89,41 @@ export const createSFFVResponse = (
 });
 
 export const createRecommendResponse = (
-  requests: readonly any[]
-): RecommendQueriesResponse<any> => {
-  return { results: requests.map(createSingleSearchResponse) };
+  requests: any[]
+): { results: Array<RecommendResponse<any>> } => {
+  return {
+    results: requests.map(createSingleSearchResponse),
```
dhayab, line 95: Could this use now `createSingleRecommendResponse`?

tests/mocks/createAlgoliaSearchClient.ts
```diff
@@ -17,19 +21,42 @@ import {
...
+  if (typeof fallback === 'function') {
+    return fallback as IfAny<TTest, TFallback>;
+  }
+  throw new Error('Neither test nor fallback is a function');
+}
+
+const algoliasearch = ifActualFunction(
+  namedConstructor,
+  defaultConstructor
+) as unknown as (appId: string, apiKey: string) => SearchClient;
```
dhayab, line 50: 😳

dhayab (approved):
Impressive, thanks for handling this 🙇 ! I've left 2 non-blocking comments.

package.json
```diff
@@ -38,7 +38,8 @@
   "devDependencies": {
     "@algolia/cache-common": "4.23.2",
     "@algolia/cache-in-memory": "4.23.2",
-    "@algolia/client-search": "4.23.2",
+    "@algolia/client-search": "5.0.0-beta.9",
+    "@algolia/client-common": "5.0.0-beta.10",
```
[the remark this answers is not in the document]
Haroenv: that's the latest version of both, they had a problem at publish time and the "helper" packages are one patch ahead.

tests/mocks/createAlgoliaSearchClient.ts
Haroenv: actually this isn't needed anymore, that was when I was still attempting to use the function from the import (but it isn't possible)

Issue #6258, opened 2024-06-30
Unnecessary client-side query with Remix


sefai:
### 🐛 Current behavior

After the page is rendered server side, there is a duplicate client-side query. This is true for both doc and data requests.

### 🔍 Steps to reproduce

1. Go to the https://shopify-hydrogen-algolia-ff7a0d8a34696233bb09.o2.myshopify.dev/
2. Click on Men from the top navigation, the results are loaded with server side rendering and there is no client-side query.
3. Now click on Women, there is a data request(which handles SSR via the loader using getServerState) and there is also an unnecessary client side query.
4. Now refresh the browser, you will see an SSR'ed page but again another unnecessary client-side query.

### Live reproduction

https://shopify-hydrogen-algolia-ff7a0d8a34696233bb09.o2.myshopify.dev/

### 💭 Expected behavior

There should be no client-side query at Step 3.
Also, there should be no client-side query for a page refresh at Step 4.

### Package version

"instantsearch.js": "^4.56.1","react-instantsearch": "^7.5.2",

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

shraddha761:
Assign this issue to me

sefai:
I have confirmed that this is not related to Remix, Next.js versions also have the same behaviour.

neehhaa06:
Check for Edge Runtime: Ensure you're not deploying to Edge Runtime; use Node.js runtime instead.

sefai:
Thank you @aymeric-giraudet ,

How can I test the master branch, is there an npm package for that.

For the second part, for each HierarchicalMenu refinement I have a SEO path, so I need that path from the server with other path related data not present on algolia indexes.

By the way I switched to Next.js, because somehow the same Remix app's CPU usage is way higher than Next.js on cloudflare, I am not sure if it is related to InstantSearch but right now the page only has InstantSearch components.

Haroenv:
>  is there an npm package for that

you can follow these instructions: https://ci.codesandbox.io/status/algolia/instantsearch/pr/6275/builds/521243

sefai:
Hello @Haroenv,

I have tried with the following packages, but the results are the same. Did you tried this fix with a HierarchicalMenu having more than one levels?

```
"instantsearch.js": "https://pkg.csb.dev/algolia/instantsearch/commit/ca700da2/instantsearch.js",
[...]

PR #6275, opened 2024-07-09
fix(cache): support multi-index and multiple queries per index

aymeric-giraudet:
**Summary**

#### [FX-2938](https://algolia.atlassian.net/browse/FX-2938)

Fixes part of #6258 and #6259 

**Result**

- A single index may cause up to 3 requests at once, if it has facets applied or hierarchical facets, so request params should be an array.
- Shape of a cached request is `{ method: 'search', args: [Array<QueryParams>] }`, not `{ method: 'search', args: Array<Array<QueryParams>> }` as it takes a single argument which is an array of params.

[FX-2938]: https://algolia.atlassian.net/browse/FX-2938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/types/results.ts
+++ b/packages/instantsearch.js/src/types/results.ts
@@ -105,7 +105,7 @@ type InitialResult = {
     params: RecommendParametersOptions['params'];
     results: RecommendResults['_rawResults'];
   };
-  requestParams?: SearchOptions;
+  requestParams?: SearchOptions[];
 };
 
 export type InitialResults = Record<string, InitialResult>;
--- a/packages/instantsearch.js/src/lib/__tests__/server.test.ts
+++ b/packages/instantsearch.js/src/lib/__tests__/server.test.ts
@@ -314,20 +314,26 @@ describe('getInitialResults', () => {
     // ...and only the latest duplicate params are in the returned results
     const expectedInitialResults = {
       indexName: expect.objectContaining({
-        requestParams: expect.objectContaining({
-          query: 'apple',
-        }),
+        requestParams: expect.arrayContaining([
+          expect.objectContaining({
+            query: 'apple',
+          }),
+        ]),
       }),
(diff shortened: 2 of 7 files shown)


Haroenv:
Which parts of https://github.com/algolia/instantsearch/issues/6258 and https://github.com/algolia/instantsearch/discussions/6259 aren't solved in this PR?

packages/instantsearch.js/src/types/results.ts
```diff
@@ -105,7 +105,7 @@ type InitialResult = {
     params: RecommendParametersOptions['params'];
     results: RecommendResults['_rawResults'];
   };
-  requestParams?: SearchOptions;
+  requestParams?: SearchOptions[];
```
Haroenv, line 108: should this still be optional? maybe InitialResult as a whole optional but each key required?

aymeric-giraudet:
@Haroenv this one : 

> Now click on Women, there is a data request(which handles SSR via the loader using getServerState) and there is also an unnecessary client side query.

We would need to react to `initialResults` changes and totally opt-out of client-side requesting, or just have Remix users not call the loader if possible as it may add unnecessary overhead

Haroenv (approved):

Haroenv:
In that case I think the issue identified in https://github.com/algolia/instantsearch/issues/6258 can be marked as done, and the part where the remix integration docs is wrong can be a separate ticket :)

packages/instantsearch.js/src/types/results.ts
aymeric-giraudet: Could be an empty array maybe, I still think it's undefined when there are no search widgets and only recommend ones though

dhayab (approved):
👏

PR #6276, opened 2024-07-09
feat(hits): add banners to Hits widget for Vue

taylorcjohnson:
**Summary**

This PR adds the new top banner to the Vue `Hits` widget and updates associated tests. The top banner feature has already been added to `Hits` and `InfiniteHits` for the JS and React flavors of InstantSearch.

This widget is slightly different from the React version by using a `showBanner` prop (defaulted to `true`) since the already updated `InfiniteHits` Vue widget uses this approach.

[Jira issue](https://algolia.atlassian.net/browse/EMERCH-1477)

**Result**

Getting Started example app:image


packages/vue-instantsearch/src/components/__tests__/Hits.js
```diff
@@ -28,6 +42,37 @@ it('accepts an escapeHTML prop', () => {
...
+    {
+      propsData: {
+        showBanner: true,
+      },
+    }
+  );
+  const img = wrapper.find('img');
+  expect(img.attributes('src')).toBe('https://via.placeholder.com/550x250');
+});
+
```
taylorcjohnson, line 75: This test runs locally, but [fails in CI](https://app.circleci.com/pipelines/github/algolia/instantsearch/11204/workflows/fa6f843a-c2a3-4823-8645-738773317cd0/jobs/54863) - so I'm assuming there is some difference between vue2 and vue3 that isn't accounted for. I added it as almost an afterthought. Since the common tests succeed we could remove this unless there is a straight-forward approach to ensuring it runs for vue3 as well.

packages/vue-instantsearch/src/components/Hits.js
```diff
@@ -1,5 +1,6 @@
 import { createHitsComponent } from 'instantsearch-ui-components';
 import { connectHitsWithInsights } from 'instantsearch.js/es/connectors';
+import { Fragment } from 'preact';
```
aymeric-giraudet, line 3: This shouldn't be imported as it adds around 3kb to the bundle

packages/vue-instantsearch/src/components/Hits.js
```diff
@@ -76,9 +83,18 @@ export default {
...
     };
 
+    const bannerComponent = ({ banner }) => {
+      return h(Fragment, {}, [
+        bannerSlot &&
+          bannerSlot({
+            banner,
+          }),
+      ]);
+    };
```
aymeric-giraudet, line 93: Instead you can just do this : 
```suggestion
    const bannerComponent =
      bannerSlot &&
      (({ banner }) =>
        bannerSlot({
          banner,
        }));
```

packages/vue-instantsearch/src/components/__tests__/Hits.js
aymeric-giraudet: That's because the thing with `Fragment` from `preact` doesn't work in Vue 3, implementing what I describe in my comment should make it work

aymeric-giraudet (changes requested):

packages/vue-instantsearch/src/components/Hits.js
taylorcjohnson: I noticed the increase, thank you for mentioning this and for the code snippet. I'll push another commit to update `bundlesize` when it goes back down after removing.

packages/vue-instantsearch/src/components/Hits.js
taylorcjohnson: Applied

packages/vue-instantsearch/src/components/__tests__/Hits.js
taylorcjohnson: Makes sense. Worked as expected now that `Fragment` has been removed.

packages/vue-instantsearch/src/components/Hits.js
```diff
@@ -76,9 +82,12 @@ export default {
       );
     };
 
+    const bannerComponent =
+      bannerSlot && (({ banner }) => bannerSlot({ banner }));
```
aymeric-giraudet, line 86: Thinking about it, just passing `bannerSlot` as `bannerComponent` should work, there's no need to wrap it into another function sorry for that

aymeric-giraudet (approved):

packages/vue-instantsearch/src/components/Hits.js
taylorcjohnson: Yep, good call. I just pushed up a commit directly passing it. 👍

Haroenv (approved):

PR #6281, opened 2024-07-10
feat(infinite-hits): add banners to vue flavor

taylorcjohnson:
**Summary**

This PR adds the new top banner to the Vue `InfiniteHits` widget and updates associated tests. The top banner feature has already been added to `Hits` and `InfiniteHits` for the JS and React flavors of InstantSearch.

This widget is slightly different from the React version by using a `showBanner` prop (defaulted to `true`) since relying on whether a parent component passed included content with the banner slot could be difficult. I'm open to changing this implementation if anyone has a recommended approach though.

[Jira issue](https://algolia.atlassian.net/browse/EMERCH-1560)

**Result**

Getting Started example app:
image

### DIFF
--- a/tests/common/widgets/infinite-hits/options.ts
+++ b/tests/common/widgets/infinite-hits/options.ts
@@ -9,8 +9,6 @@ import {
 } from '@instantsearch/testutils';
 import userEvent from '@testing-library/user-event';
 
-import { skippableDescribe } from '../../common';
-
 import type { InfiniteHitsWidgetSetup } from '.';
 import type { TestOptions } from '../../common';
 import type { MockSearchClient } from '@instantsearch/mocks';
@@ -53,7 +51,7 @@ function normalizeSnapshot(html: string) {
 
 export function createOptionsTests(
   setup: InfiniteHitsWidgetSetup,
-  { act, skippedTests }: Required<TestOptions>
+  { act }: Required<TestOptions>
 ) {
   describe('options', () => {
     test('renders with default props', async () => {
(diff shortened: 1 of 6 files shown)


tests/common/widgets/infinite-hits/options.ts
```diff
@@ -108,88 +106,86 @@ export function createOptionsTests(
...
-            widgets: {
-              banners: [
-                {
-                  image: {
-                    urls: [{ url: 'https://via.placeholder.com/550x250' }],
-                  },
-                  link: {
-                    url: 'https://www.algolia.com',
-                  },
+    test('renders default banner element with banner widget renderingContent', async () => {
```
taylorcjohnson, line 109: The snapshot output for the Vue version of this test is adding an extra line between the banner and hit list, but I cannot figure out why. I must be missing something...

https://app.circleci.com/pipelines/github/algolia/instantsearch/11178/workflows/90ab6bc7-58a9-4185-94ab-18d1594139c4/jobs/54699

dhayab: Could this be an undesirable side-effect of prettier applying rules when saving, which might break the generated snapshot? If so, you can either update the snapshot without saving the file again from your editor, or save without formatting being applied.

taylorcjohnson: I will double-check, but I don't believe so. The test passes for JS and React, but fails only for Vue.

dhayab: Ah, this then means that the Vue serializer used for Jest outputs the DOM differently then the other ones. To "fix" this, we smooth out differences like this in the normalize function that should be at the top of the file, by usually adding a regular expression that targets the relevant extra whitespace and removes it.

taylorcjohnson: Yep, @aymeric-giraudet had just sent me a snippet to try. Worked locally, just pushed another commit.

dhayab (approved):

aymeric-giraudet (approved):
I think bundlesize threshold needs to be updated too, otherwise looks good !

PR #6282, opened 2024-07-11
fix(vue2): `augmentedCreateElement` should pass HTML attributes

aymeric-giraudet:
**Summary**

This was causing an issue for https://github.com/algolia/instantsearch/pull/6281

**Result**

When the tag is a string, we pick the actual vue2 props and assume the rest are attributes. We still have to pass `attrs` if passed for compatibility with actual vue render functions.
Now the function should be compatible with both `createElement` and Vue's `h` signatures

### DIFF
--- a/packages/vue-instantsearch/src/util/__tests__/renderCompat.test.js
+++ b/packages/vue-instantsearch/src/util/__tests__/renderCompat.test.js
@@ -1,0 +1,52 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import { mount } from '../../../test/utils';
+import { renderCompat } from '../vue-compat';
+
+describe('renderCompat', () => {
+  it('should handle function components that return multiple children', () => {
+    const wrapper = mount({
+      render: renderCompat((h) => {
+        function Component({ text }) {
+          return h(
+            'div',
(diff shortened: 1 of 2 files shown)


packages/vue-instantsearch/src/util/__tests__/renderCompat.test.js
```diff
@@ -0,0 +1,52 @@
...
+describe('renderCompat', () => {
+  it('should handle function components that return multiple children', () => {
+    const wrapper = mount({
+      render: renderCompat((h) => {
+        function Component({ text }) {
+          return h(
+            'div',
+            { attrs: { 'data-text': text } },
+            h('span', {}, ['1']),
+            h('span', {}, '2')
```
aymeric-giraudet, line 17: I also added this test to make sure different ways of passing children work (rest args or single array)

Haroenv (approved):
phew, thanks for this PR, but would be good to go to vue 3 only at some point in the future, this is getting complicated

Haroenv:
netlify passed here: https://app.netlify.com/sites/instantsearchjs/deploys/668fa5cd800f0e08b3cc3846

PR #6286, opened 2024-07-16
ci: add downgrade script for algoliasearch v4 and steps

aymeric-giraudet:
**Summary**

#### [FX-2929](https://algolia.atlassian.net/browse/FX-2929)

**Result**

We run the same tests and type checking as for v5 as they were written for v4 anyway, it does work

[FX-2929]: https://algolia.atlassian.net/browse/FX-2929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -239,7 +239,7 @@ jobs:
             ./scripts/legacy/downgrade-algoliasearch-v4.js
       - run:
           name: Unit & Integration tests
-          command: yarn run test
+          command: yarn run test:ci
       - run:
           name: Type Checking
           command: yarn run type-check
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,10 @@ workflows:
       - unit tests:
           requires:
             - build
-      - legacy algoliasearch:
+      - legacy algoliasearch v3:
+          requires:
+            - build
+      - legacy algoliasearch v4:
           requires:
             - build
       - vue v3:
@@ -97,7 +100,8 @@ workflows:
             - lint
             - unit tests
             - examples
-            - legacy algoliasearch
+            - legacy algoliasearch v3
+            - legacy algoliasearch v4
             - helper docs
             - e2e tests
           filters:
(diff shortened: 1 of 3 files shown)


scripts/legacy/downgrade-algoliasearch-v4.js
```diff
@@ -0,0 +1,67 @@
...
+shell.sed(
+  '-i',
+  /"algoliasearch": ".*"(,)?/,
+  '"algoliasearch": "4.23.2"$1',
+  ...shell.ls('examples/*/*/package.json')
+);
+
+shell.exec('yarn install');
+
+shell.exec('cp -r node_modules/algoliasearch-v4 node_modules/algoliasearch');
```
aymeric-giraudet, line 67: Struggled with yarn which still keeps `4.23.2` with the `algoliasearch-v4` name, even if we remove the alias.

It does output some warnings like "warning Pattern is trying to unpack in the same destination" which may be related.

Copying it manually works though so idk if we should bother

Haroenv: I wonder if this is because of Create InstantSearch App

.circleci/config.yml
```diff
@@ -221,6 +225,25 @@ jobs:
...
+      - *install_yarn_version
+      - *restore_yarn_cache
+      - *run_yarn_install
+      - run:
+          name: Update dependencies
+          command: |
+            ./scripts/legacy/downgrade-algoliasearch-v4.js
+      - run:
+          name: Unit & Integration tests
+          command: yarn run test
```
Haroenv, line 242: we need to store the test results like in unit tests

Haroenv (commented):
overall this looks good, let's just get CI to pass once then lgtm!

.circleci/config.yml
aymeric-giraudet: Done and everything passes ! Thanks !

Haroenv (commented):
approve :)

PR #6288, opened 2024-07-18
feat(index): make IndexOptions['uiState'] and UiState react to generic

Haroenv:
**Summary**

If you're reading `uiState` (initial ui state) inside init of a widget, it should be the generic UiState which includes this widget's  ui state as well.

**Result**

- UiState generic to add a specific widget's IndexUiState
- IndexUiState generic to add a specific widget's IndexUiState
- InitOptions generic with WidgetDescription
- init({ uiState }) correctly typed including the specific widget's ui state

### DIFF
--- a/packages/instantsearch.js/src/types/ui-state.ts
+++ b/packages/instantsearch.js/src/types/ui-state.ts
@@ -37,8 +37,10 @@ type ConnectorUiStates = AutocompleteWidgetDescription['indexUiState'] &
 
 type WidgetUiStates = PlacesWidgetDescription['indexUiState'];
 
-export type IndexUiState = Partial<ConnectorUiStates & WidgetUiStates>;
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/connectors/geo-search/connectGeoSearch.ts
```diff
@@ -266,7 +266,8 @@ export default (function connectGeoSearch<
     return {
       $$type,
 
-      init(initArgs) {
+      // Type is explicitly redefined, to avoid having the TWidgetParams type in the definition
```
Haroenv, line 269: https://github.com/microsoft/rushstack/issues/4740, https://github.com/algolia/instantsearch/pull/6218/commits/0cde49ee54dd844343e10ba6060b0225492fe0ec

aymeric-giraudet (approved):

PR #6289, opened 2024-07-22
feat(ui-components): introduce `Carousel`

dhayab:
**Summary**

This PR introduces the Carousel UI component. It is a port of the HorizontalSlider component designed. There are a few changes:

- HorizontalSlider is now named Carousel, obv
- Class names are prefixed with `ais-Carousel`
- References are not forwarded anymore, `Renderer` is extended to allow flavors to provide their `useRef` implementation
- Carousel CSS colors are now SCSS variables instead of CSS variables

This PR does not implement the Carousel as a built-in layout in any InstantSearch widget for now.

[**Preview →**](https://kq8txk.csb.app/)

**Before merging**

- [x] Remove carousel import in algolia and reset SCSS files
- [x] Remove changes to getting started

FX-2817


packages/instantsearch-ui-components/src/components/Carousel.tsx
```diff
@@ -0,0 +1,222 @@
...
+   */
+  previousButtonLabel: string;
+  /**
+   * The title of the previous navigation element
+   */
+  previousButtonTitle: string;
+  /**
+   * The label of the carousel
+   */
+  listLabel: string;
```
dhayab, line 71: 📝 this was previously named `sliderLabel`

packages/instantsearch-ui-components/src/components/Carousel.tsx
```diff
@@ -0,0 +1,222 @@
...
+}
+
+export function createCarouselComponent({ createElement, useRef }: Renderer) {
+  // This prevents the SVG buttons references from being hoisted
+  // during the build, where `createElement` is not available.
+  const buttonPathCommonProps: JSX.IntrinsicElements['path'] = {
+    fillRule: 'evenodd',
+    clipRule: 'evenodd',
+    fill: 'currentColor',
+  };
```
dhayab, line 87: 📝 I could not prevent Babel from hoisting simple JSX references outside of `createCarouselComponent`, so I resorted to this "hack"

packages/instantsearch-ui-components/src/components/Carousel.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  return function Carousel<TObject extends RecordWithObjectID>(
+    props: CarouselProps<TObject>
+  ) {
+    const listRef = useRef?.<HTMLOListElement>(null);
+    const nextButtonRef = useRef?.<HTMLButtonElement>(null);
+    const previousButtonRef = useRef?.<HTMLButtonElement>(null);
+    const carouselIdRef = useRef?.(generateCarouselId());
```
dhayab, line 95: 📝 this will need to be addressed when we'll use it for Vue

packages/instantsearch-ui-components/src/components/Carousel.tsx
```diff
@@ -0,0 +1,222 @@
...
+        listRef.current.scrollLeft + listRef.current.clientWidth >=
+        listRef.current.scrollWidth;
+    }
+
+    if (props.items.length === 0) {
+      return null;
+    }
+
+    return (
+      <div className={cx('ais-Carousel', classNames.root)}>
```
dhayab, line 139: 📝 the root class was previously named `uic-HorizontalSlider-container`, I removed the last part to align with what we have in InstantSearch

packages/instantsearch.css/src/themes/_carousel.scss
```diff
@@ -0,0 +1,98 @@
+// Colors
+$primary-color: #9698c3;
+$muted-color: #c4c4c4;
+$dark-color: #0f0f0f;
```
dhayab, line 4: 📝 these were previously CSS variables

packages/instantsearch-ui-components/src/components/Carousel.tsx
Haroenv: wonder if disabling https://babeljs.io/docs/babel-plugin-transform-react-constant-elements could have helped?

packages/instantsearch-ui-components/src/components/Carousel.tsx
```diff
@@ -0,0 +1,234 @@
...
+        >
+          <svg width="8" height="16" viewBox="0 0 8 16" fill="none">
+            <path
+              {...buttonPathCommonProps}
+              d="M7.13809 0.744078C7.39844 1.06951 7.39844 1.59715 7.13809 1.92259L2.27616 8L7.13809 14.0774C7.39844 14.4028 7.39844 14.9305 7.13809 15.2559C6.87774 15.5814 6.45563 15.5814 6.19528 15.2559L0.861949 8.58926C0.6016 8.26382 0.6016 7.73618 0.861949 7.41074L6.19528 0.744078C6.45563 0.418641 6.87774 0.418641 7.13809 0.744078Z"
+            />
+          </svg>
+        </button>
+
+        <ol
```
Haroenv, line 182: could this part be `{children}`? it does have the slide and roledescriptor labels, but I wonder if that could be implemented differently. With this solution I don't really know how carousel could become a parent of hits

Haroenv (commented):
code looks good as an integration of carousel in the repo

packages/instantsearch-ui-components/src/components/Carousel.tsx
dhayab: It does, thanks for pointing it out. I could remove the plugin for all ui-components, we're basically expecting every element to be rendered through `createElement` which makes this opt uninteresting in this package. Wdyt?

packages/instantsearch-ui-components/src/components/Carousel.tsx
dhayab: Yes, this will probably need to be adjusted in a dedicated PR once we get consensus on how we'll consume this component.

packages/instantsearch-ui-components/src/components/Carousel.tsx
Haroenv: yes, I think we should do that chagne

packages/instantsearch-ui-components/src/components/Carousel.tsx
aymeric-giraudet: We should probably just pass all the refs, I don't see how passing `useRef` to be called there is necessary
But yeah this could be changed later

packages/instantsearch-ui-components/src/components/Carousel.tsx
dhayab: Done in [3b86cbf](https://github.com/algolia/instantsearch/pull/6289/commits/3b86cbf54c6daa210f8f4d3af624f15ec0dde1a8).

packages/instantsearch-ui-components/src/components/Carousel.tsx
francoischalifour: +1. This leaks the `useRef` Hook API in the generic UI component. We could accept `listRef`, `nextButtonRef`, `previousButtonRef` and `carouselIdRef` as props. Changing this only when Vue is supported might be a breaking change?

dhayab: I think both options are not agnostic, as they assume the reference adheres to a specific structure (`{ current: T }`) in order to be bound to a DOM element.

Instead, passing refs as props add complexity on the developers' end if we decide to provide the Carousel as a wrapper (we can handle the complexity if we use it internally on our widgets).

Proposed changes are done in [1d51ef4](https://github.com/algolia/instantsearch/pull/6289/commits/1d51ef47c479c46b4cf42a9827e3e990cb119fe0).

aymeric-giraudet: My main issue was that in Vue we call the `create` function each render, and calling `useRef` also makes these components not stateless anymore.

Changes look good to me, should make supporting Vue easier later I believe :)

aymeric-giraudet (approved):

Haroenv (approved):

examples/react/getting-started/index.html
```diff
@@ -9,15 +9,6 @@
 
     <link rel="shortcut icon" href="favicon.png" />
 
-    <!--
-      Do not use @8 in production, use a complete version like x.x.x, see website for latest version:
-      https://www.algolia.com/doc/guides/building-search-ui/installation/js/#load-the-styles
-    -->
-    <link
```
Haroenv, line 16: the changes to example still have to be reverted before merging (if we merge before usage)

PR #6291, opened 2024-07-23
fix(dependencies): replace qs by neoqs

Haroenv:
**Summary**

Replace qs by neoqs

**Result**

neoqs is API-compatible with qs in the legacy mode, and thus this isn't a breaking change. Of course if someone was already using qs and relying on it being deduplicated with the InstantSearch version.

URLs parsed and created by qs and neoqs are compatible

### DIFF
--- a/examples/vue/ssr/package.json
+++ b/examples/vue/ssr/package.json
@@ -11,7 +11,7 @@
     "algoliasearch": "4.23.2",
     "core-js": "2",
     "instantsearch.css": "8.3.0",
-    "qs": "6.9.7",
+    "neoqs": "6.2.12",
     "vue": "2.7.14",
     "vue-instantsearch": "4.19.0",
     "vue-router": "^3.0.2",
--- a/examples/vue/ssr/src/main.js
+++ b/examples/vue/ssr/src/main.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-nested-ternary */
 import algoliasearch from 'algoliasearch/lite';
-import qs from 'qs';
+import * as qs from 'neoqs/legacy';
 import Vue from 'vue';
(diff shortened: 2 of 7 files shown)


Haroenv:
Hmm bundle size is larger, doesn't seem to be due to non-working deduplication, maybe other features added since qs@6.9.7? cc @puruvjdev is this expected?

examples/vue/ssr/package.json
```diff
@@ -11,7 +11,7 @@
     "algoliasearch": "4.23.2",
     "core-js": "2",
     "instantsearch.css": "8.3.0",
-    "qs": "6.9.7",
+    "neoqs": "6.2.12",
```
PuruVJ, line 14: Should be 6.12.3, I released slightly wrong version.

PuruVJ:
> Hmm bundle size is larger, doesn't seem to be due to non-working deduplication, maybe other features added since qs@6.9.7? cc @puruvjdev is this expected?

I am looking at bundlephobia to see if that's the case, but it seems to be down. I'll let you know if that's the case

PuruVJ:
Well the real thing is that `/legacy` is compiled to ES5, so it does add an extra 400bytes(min+br). That is the additional size we're seeing here. This is done so even very old libs and apps can use neoqs and avoid dependency hell. Also, there have been some additional features since 6.9.7 which do account for a few extra bytes here.

Only way the size would go down is by using the default build, which is ESM-only.

I'd suggest using neoqs/legacy now with the extra 1KB, if it's not too much, but whenever in future you decide to drop CJS support, switch to `neoqs/modern`(It's gonna be 3.4KB only 👀, releasing sometime next week).

Lemme know if that helps

PuruVJ:
Any update on this? 😁

PR #6294, opened 2024-07-24
fix(imports): prevent circular import

### LINKED ISSUE #6292: Circular dependency in `react-instantsearch-core`
### 🐛 Current behavior

Found a circular dependency in your dist code when rolling up a project. 
```
(!) Circular dependency
node_modules/react-instantsearch/node_modules/react-instantsearch-core/dist/es/index.js -> node_modules/react-instantsearch/node_modules/react-instantsearch-core/dist/es/server/index.js -> node_modules/react-instantsearch/node_modules/react-instantsearch-core/dist/es/server/getServerState.js -> node_modules/react-instantsearch/node_modules/react-instantsearch-core/dist/es/index.js
```

### 🔍 Steps to reproduce

1) `npm install react-instantsearch`
2) create a js file with `import  * from 'react-instantsearch'`
3) compile with rollup

### Live reproduction

N/A

### 💭 Expected behavior
The code would ideally not have a circular dependency.
I believe if the imports in [getServerState.js ](https://github.com/algolia/instantsearch/blob/master/packages/react-instantsearch-core/src/server/getServerState.tsx#L8-L13) were updated to point directly to the composites, this issue would be resolved.

```
import type { InstantSearchServerState, InstantSearchSSRProvider } from '../component/InstantSearchSSRProvider';
import type { InstantSearchServerContextApi, InstantSearchServerContext } from '../component/InstantSearchServerContext';
```
^ Not sure if this is valid but figure it is something along these lines :)

### Package version

react-instantsearch-core 7.12.2

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

Haroenv:
**Summary**

A circular dependency can cause issues, which is happening here because getServerState is importing from the index, but it's also exported from that same index.

**Result**

fixes #6292

also copied the code we have in InstantSearch.js to prevent circular dependency into react and vue instantsearch. Before the fix the build failed thanks to these onwarn changes.

This PR removes the circular dependency and also goes through all imports in React InstantSearch core and makes sure they're consistent (but the only circle was in getServerState).

### DIFF
--- a/packages/react-instantsearch-core/rollup.config.js
+++ b/packages/react-instantsearch-core/rollup.config.js
@@ -57,6 +57,12 @@ const createConfiguration = ({ name, minify = false } = {}) => ({
     banner: createBanner(name),
     sourcemap: true,
   },
+  onwarn(warning, warn) {
+    if (warning.code === 'CIRCULAR_DEPENDENCY')
+      throw new Error(warning.message);
+
+    warn(warning);
+  },
   plugins: plugins.concat(
     clear([
       minify &&
--- a/packages/react-instantsearch-core/src/lib/useIndex.ts
+++ b/packages/react-instantsearch-core/src/lib/useIndex.ts
@@ -1,9 +1,8 @@
 import index from 'instantsearch.js/es/widgets/index/index';
 import { useMemo } from 'react';
 
-import { useIndexContext } from '../lib/useIndexContext';
-
 import { useForceUpdate } from './useForceUpdate';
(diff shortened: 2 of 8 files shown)


sarahdayan (approved):
Not immediately necessary, but since we want to throw for that same warning in all our Rollup configurations, we could have a common one that we extend.

This can be done in a different PR.

Haroenv:
Yes, the rollup configs are very similar, I think a common one could be handy

PR #6312, opened 2024-08-05
fix(ExperimentalConfigureRelatedItems): deprecate in favor of RelatedItems

Haroenv:
**Summary**

Deprecate ConfigureRelatedItems, its API is fairly confusing and we never saw significant usage. Instead what you can do is:

- use relatedItems or connectRelatedItems (this uses recommend instead of search, but fits all use cases better)
- use a regular configure with `filters: 'attributeName:attributeValue<score=i>'` fetched from the object

If you're using configureRelatedItems, please reach out and we can find a suitable alternative for you.

**Result**

EXPERIMENTAL_configureRelatedItems and its connector are marked as deprecated and will be removed in a next minor version

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -26,11 +26,11 @@
     },
     {
       "path": "packages/vue-instantsearch/vue2/umd/index.js",
-      "maxSize": "67.75 kB"
+      "maxSize": "68 kB"
     },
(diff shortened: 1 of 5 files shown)


dhayab (approved):
LGTM!

PR #6313, opened 2024-08-06
feat(carousel): introduce styling for all themes

aymeric-giraudet:
**Summary**

### [FX-2966](https://algolia.atlassian.net/browse/FX-2966)

**Result**

Ultimately chose to put defaults in `reset.scss`, it may be sort of big relatively to the file but takes up less space than the range slider, and it's still small relative to the overall size of InstantSearch.

- `reset` sets defaults for the carousel, it should be a grid with a horizontal scrollbar and centered buttons on each side on hover. Navigation buttons are unstyled.
- Each theme then styles the buttons, and sets spacing and colors. I only used colors from their respective palettes.

(I added red background and height to each item to make gaps and size obvious)
`reset`:
image

`algolia`:
image

`satellite`:
image

[FX-2966]: https://algolia.atlassian.net/browse/FX-2966?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/instantsearch.css/src/themes/reset.scss
```diff
@@ -196,3 +196,61 @@
...
+@media (hover: none) and (pointer: coarse) {
+  .ais-Carousel-navigation {
+    display: none;
+  }
+}
+
+.ais-Carousel:hover .ais-Carousel-navigation:not([hidden]) {
+  opacity: 1;
+  visibility: visible;
+}
```
Haroenv, line 236: this should be part of the specific styles IMO

specs/src/pages/widgets/carousel.md
```diff
@@ -0,0 +1,126 @@
...
+        <div style="height: 100px;background: red;">
+          5
+        </div>
+      </li>
+      <li
+        aria-label="6 of 6"
+        aria-roledescription="slide"
+        class="ais-Carousel-item"
+      >
+        <div style="height: 100px;background: red;">
```
Haroenv, line 87:
```suggestion
        <div style="height: 100px;border: 1px solid red; display:flex; justify-content: center;align-items: center;">
```

this style is a bit more readable IMO

Haroenv (commented):
looks great!

packages/instantsearch.css/src/themes/reset.scss
aymeric-giraudet: You mean buttons should always be visible in `reset` ?

Haroenv: yes, I think those styles would need to be reverted in many cases IMO

packages/instantsearch.css/src/themes/algolia.scss
```diff
@@ -126,6 +127,49 @@ a[class^='ais-'] {
...
+}
+
+.ais-Carousel-list::-webkit-scrollbar-thumb {
+  background-color: $port-gore;
+}
+
+.ais-Carousel-item * {
+  outline-color: $algolia-blue;
+  outline-offset: 3px;
+}
```
dhayab, line 171: I missed it in my own task, but if I'm not wrong this rule does nothing as it lacks style/width.

We can safely remove it (same for satellite).

dhayab (approved):
👍

Haroenv (approved):

PR #6314, opened 2024-08-06
feat(instantsearch.js): introduce `carousel` template

sarahdayan:
This introduces the `carousel` template and implements the `layout` template in Recommend widgets.

https://algolia.atlassian.net/browse/FX-2818

### DIFF
--- a/packages/instantsearch.js/src/templates/__tests__/carousel.test.tsx
+++ b/packages/instantsearch.js/src/templates/__tests__/carousel.test.tsx
@@ -1,0 +1,227 @@
+/**
+ * @jest-environment jsdom
+ */
+/** @jsx h */
+
+import { render } from '@testing-library/preact';
+import { h } from 'preact';
+
+import { carousel } from '../carousel/carousel';
+
+describe('carousel', () => {
+  test('renders with default options', () => {
+    const template = carousel();
+
+    const { container } = render(
+      template({
+        items: [
+          { objectID: '1', __position: 1 },
+          { objectID: '2', __position: 2 },
+        ],
+        templates: {
+          item({ item }) {
+            return <p>{item.objectID}</p>;
+          },
+        },
+      })
+    );
+
+    expect(container).toMatchInlineSnapshot(`
+      <div>
+        <div
+          class="ais-Carousel"
+        >
+          <button
+            aria-controls="ais-Carousel-0"
+            aria-label="Previous"
+            class="ais-Carousel-navigation ais-Carousel-navigation--previous"
+            hidden=""
+            title="Previous"
+          >
+            <svg
+              fill="none"
+              height="16"
+              viewBox="0 0 8 16"
+              width="8"
+            >
+              <path
+                clipRule="evenodd"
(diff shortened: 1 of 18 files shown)


packages/instantsearch.js/src/templates/__tests__/carousel.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+            title="Next"
+          >
+            <p>
+              Next
+            </p>
+          </button>
+        </div>
+      </div>
+    `);
+  });
```
Haroenv, line 226: I don't think we need to add that or think about it yet, but when hit is supported, this needs insights tests and functionality too.

packages/instantsearch.js/src/templates/carousel/carousel.tsx
```diff
@@ -0,0 +1,85 @@
...
+  'items'
+> & {
+  templates: {
+    item?: CarouselUiProps<TObject>['itemComponent'];
+  };
+};
+
+export function carousel<TObject extends Record<string, unknown>>({
+  cssClasses,
+  templates,
```
Haroenv, line 61:
```suggestion
  templates = {},
```

if code ~ like this, you don't need to check `templates?.` later

packages/instantsearch.js/src/widgets/frequently-bought-together/frequently-bought-together.tsx
```diff
@@ -111,6 +111,33 @@ const renderer =
...
+      templates.layout
+        ? (data) => (
+            <TemplateComponent
+              {...renderState.templateProps}
+              templateKey="layout"
+              rootTagName="fragment"
+              data={{
+                items: data.items,
+                templates: {
+                  item: templates.item
```
Haroenv, line 124: surely there's always at least a default item template?

Haroenv (commented):
🚀

packages/instantsearch.js/src/templates/__tests__/carousel.test.tsx
sarahdayan: Yes, we can accept `sendEvent` in there when it's time.

packages/instantsearch.js/src/widgets/frequently-bought-together/frequently-bought-together.tsx
sarahdayan: The default item template is defined in the UI component. At the widget level, the `itemComponent` can be undefined, which will be forwarded as-is to the UI component and be defaulted there.

packages/instantsearch.js/src/templates/carousel/carousel.tsx
sarahdayan: Yup! Done in [77f0549](https://github.com/algolia/instantsearch/pull/6314/commits/77f0549d51380c2fc7e11659806810a107bd88e0).

dhayab (approved):
Awesome! We can probably update `trendingItems()` the js getting-started example to use the carousel layout. We can also wait until we release instantsearch.css (we load themes from a CDN in the examples).

Haroenv (approved):

PR #6316, opened 2024-08-07
docs(instantsearch.js): use `carousel` template

sarahdayan:
This uses the new `carousel` template in the **Getting Started** example for InstantSearch.js.

Capture d’écran 2024-08-07 à 19 34 17

### DIFF
--- a/examples/js/getting-started/index.html
+++ b/examples/js/getting-started/index.html
@@ -10,14 +10,6 @@
 
     <link rel="shortcut icon" href="./favicon.png" />
 
-    <!--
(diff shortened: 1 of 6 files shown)


dhayab (commented):
Could you also add this in **app.js**?

sarahdayan:
> Could you also add this in **app.js**?

Sure! I also added `instantsearch.css` as a dependency so we can visualize the styles right away, and styled the carousel a bit more to make it look nice.

Haroenv (approved):

dhayab (approved):

PR #6321, opened 2024-08-09
feat(react-instantsearch): add `Carousel` layout component

shaejaz:
This adds the the `Carousel` layout component and enables the `layoutComponent` prop in Recommend widgets. React getting started example has been update to use `Carousel` as well.

image

https://algolia.atlassian.net/browse/FX-2819


packages/react-instantsearch/src/templates/index.ts
```diff
@@ -0,0 +1 @@
+export * from './Carousel';
```
Haroenv, line 1: it's worth thinking about, but in vanilla it's called /templates because that exists as a concept already, in react the components end in Component, so they should be in a components folder I think (or widgets is probably fine too, Highlight isn't a widget per se either and is exported from widgets)

examples/react/getting-started/src/App.css
```diff
@@ -61,7 +61,7 @@ em {
   flex: 3;
 }
 
-.searchbox {
+#searchbox {
```
sarahdayan, line 64: Why this change? I don't see any element with `id="searchbox"` in the example.

packages/react-instantsearch/src/templates/Carousel.tsx
sarahdayan, line 1: I'd rename this folder `components`. React InstantSearch doesn't have the concept of "templates", and the component props are all called `somethingComponent`.

packages/react-instantsearch/src/widgets/__tests__/FrequentlyBoughtTogether.test.tsx
```diff
@@ -81,4 +82,149 @@ describe('FrequentlyBoughtTogether', () => {
...
+    });
+    const { container } = render(
+      <InstantSearchTestWrapper searchClient={client}>
+        <FrequentlyBoughtTogether
+          objectIDs={['1']}
+          itemComponent={({ item }) => <p>{item.objectID}</p>}
+          layoutComponent={(props) => (
+            <Carousel
+              {...props}
+              previousIconComponent={() => <p>Previous</p>}
```
sarahdayan, line 151: Can you also test passing custom `classNames` and `translations` since they're available APIs?

Same for the other widgets.

packages/instantsearch-ui-components/src/types/Recommend.ts
sarahdayan, line 1: Correct change IMO, but we're still passing `translations` to the layout in each UI component.

sarahdayan (changes requested):
Almost there, a couple changes and we should be good.

I also noticed that the box shadow is being cut off in demo because of the overflow ruleset. I fixed it in the JS example, feel free to compare markup and styles to replicate it.

Capture d’écran 2024-08-09 à 15 21 25

packages/react-instantsearch/src/templates/index.ts
sarahdayan: This or `layout`. It doesn't change anything for users who will import from `react-instantsearch`, but internally it makes more sense.

examples/react/getting-started/src/App.css
```diff
@@ -113,3 +107,50 @@ em {
   height: 100%;
   justify-content: space-between;
 }
+
+.ais-TrendingItems-item h2,
+.ais-RelatedProducts-item h2 {
+  display: -webkit-box;
```
Haroenv, line 113: where did this style come from? seems copy-pasted from somewhere with autoprefixer maybe?

packages/instantsearch-ui-components/src/components/TrendingItems.tsx
```diff
@@ -90,7 +90,6 @@ export function createTrendingItemsComponent({
 
         <Layout
           classNames={cssClasses}
-          translations={translations}
```
Haroenv, line 93: does this mean the "previous" and "next" labels aren't customisable, or is that done another way now?

Haroenv (commented):
🔥

examples/react/getting-started/src/App.css
sarahdayan: Same styles as the ones I used for InstantSearch.js. They're the recommended ones for line clipping.

packages/react-instantsearch/src/widgets/__tests__/__utils__/cx.ts
shaejaz, line 1: Not sure if this should be here, it's only being used by the updated tests in the react-is package atm.

Haroenv: I think the correct way is to `import { cx } from 'instantsearch-ui-components';`

packages/instantsearch-ui-components/src/components/TrendingItems.tsx
sarahdayan: The translations we passed were the ones of the widget, and it didn't make sense to do so. Layouts have their own translations which they receive as a prop.

Haroenv (approved):

dhayab (approved):
👏

sarahdayan (approved):

Haroenv (approved):

PR #6322, opened 2024-08-09
docs(instantsearch.js): load `instantsearch.css` from packages

sarahdayan:
This imports `instantsearch.css` from the installed packages instead of jsDelivr so we can view style changes when running examples.

This is only for the JS examples, I'll do React and Vue separately.


aymeric-giraudet (approved):
Nice ! Changes to `yarn.lock` can probably be discarded though, I see some packages got duplicated

examples/js/e-commerce-umd/index.html
```diff
@@ -17,13 +17,10 @@
...
     />
-    <!--
-      Do not use @8 in production, use a complete version like x.x.x, see website for latest version:
-      https://www.algolia.com/doc/guides/building-search-ui/installation/js/#load-the-styles
-    -->
+
     <link
       rel="stylesheet"
-      href="https://cdn.jsdelivr.net/npm/instantsearch.css@8/themes/reset-min.css"
+      href="/packages/instantsearch.css/themes/reset-min.css"
```
Haroenv, line 23: this example now won't work on codesandbox anymore, but it already didn't so that's ok

examples/js/calendar-widget/index.html
```diff
@@ -1,12 +1,8 @@
 <html>
   <head>
-    <!--
-      Do not use @8 in production, use a complete version like x.x.x, see website for latest version:
-      https://www.algolia.com/doc/guides/building-search-ui/installation/js/#load-the-styles
-    -->
     <link
       rel="stylesheet"
-      href="https://cdn.jsdelivr.net/npm/instantsearch.css@8/themes/algolia.min.css"
+      href="/packages/instantsearch.css/themes/algolia-min.css"
```
Haroenv, line 5: this won't work on codesandbox, which is why jsdelivr is used in the first place. putting the import in js also doesn't really work because this example works without bundling too. Not sure what the right solution is though

Haroenv (changes requested):
it's probably fine, but I don't think having examples that only work locally after you've built the css are the perfect solution. There's a tradeoff between styles that work everywhere and styles that update locally. Maybe there can be some kind of URL rewrite for local dev?

yarn.lock
```diff
@@ -9977,7 +9988,15 @@ brorand@^1.0.1:
...
-brotli-size@0.1.0, brotli-size@4.0.0:
+brotli-size@0.1.0:
+  version "0.1.0"
+  resolved "https://registry.yarnpkg.com/brotli-size/-/brotli-size-0.1.0.tgz#a2c518096c7c1a75e9e66908a42cd9dc77d2b69f"
+  integrity sha512-5ny7BNvpe2TSmdafF1T9dnFYp3AIrJ8qJt29K0DQJzORlK38LBim/CmlY26JtreV6SWmXza7Oa+9m61SzvxR0Q==
+  dependencies:
+    duplexer "^0.1.1"
+    iltorb "^2.4.3"
+
+brotli-size@4.0.0:
```
Haroenv, line 9999: this change is odd, did you install without the `resolutions` present somehow? Maybe try running `npx yarn-deduplicate` first to see if that works it, otherwise discard the changes indeed

examples/js/calendar-widget/index.html
sarahdayan: I changed it with module imports, it didn't need to be imported like that anyway.

PR #6326, opened 2024-08-14
fix(instantsearch.js): forward widget classes to carousel

sarahdayan:
This fixes an issue where the widget classes for the list and items weren't properly forwarded to the layout.

I only pass the `list` and `item` ones because the other ones aren't necessary, and because the `root` one isn't supposed to be added to the layout root.

### DIFF
--- a/examples/js/getting-started/src/app.css
+++ b/examples/js/getting-started/src/app.css
@@ -78,8 +78,20 @@
 
 .ais-TrendingItems-item,
 .ais-RelatedProducts-item {
+  background: none !important;
+  align-items: stretch !important;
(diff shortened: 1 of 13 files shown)


sarahdayan:
> I think the `instantsearch.css` package or the getting started styles need to updated for this, the codesandbox is showing up weird for me

Good catch! I fixed it in [b52df0b](https://github.com/algolia/instantsearch/pull/6326/commits/b52df0bfd174426541ded1ac0feb210fde8a4235).

shaejaz (approved):

Haroenv (approved):

PR #6337, opened 2024-08-24
fix(react-instantsearch-nextjs): fix incorrect usage of headers

cungminh2710:
**Summary**

Currently, NextJS App Router will throw this error when using with react-instantsearch-nextjs

```
⨯ Error: `headers` was called outside a request scope. Read more: https://nextjs.org/docs/messages/next-dynamic-api-wrong-context
```

This is because the `headers` are [only for server component](https://nextjs.org/docs/app/api-reference/functions/headers) - not in a client component. The InstantSearchNext component is a client component, hence this error.

image

This PR is to fix this issue by clearly stating InstantSearchNext is a client component, and put the nonce value into InstantSearchNext props

### DIFF
--- a/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
+++ b/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
@@ -1,5 +1,6 @@
+'use client';
+
 import { safelyRunOnBrowser } from 'instantsearch.js/es/lib/utils';
-import { headers } from 'next/headers';
 import React, { useEffect, useRef } from 'react';
 import {
   InstantSearch,
@@ -36,6 +37,7 @@ export type InstantSearchNextProps<
   TRouteState = TUiState
 > = Omit<InstantSearchProps<TUiState, TRouteState>, 'routing'> & {
   routing?: InstantSearchNextRouting<TUiState, TRouteState> | boolean;
+  nonce?: string;
 };
 
 export function InstantSearchNext<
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-nextjs/src/useInstantSearchRouting.ts
```diff
@@ -21,55 +20,51 @@ export function useInstantSearchRouting<
...
-  if (passedRouting && !routingRef.current) {
-    let browserHistoryOptions: Partial<BrowserHistoryArgs<TRouteState>> = {};
-
-    browserHistoryOptions.getLocation = () => {
-      if (typeof window === 'undefined') {
-        const url = `${
-          headers().get('x-forwarded-proto') || 'http'
-        }://${headers().get('host')}${pathname}?${searchParams}`;
-        return new URL(url) as unknown as Location;
-      }
```
cungminh2710, line 39: since this is a client component, we can't call headers in here for app router. Also, this will be unnecessary for a client component to access `window` context

packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
```diff
@@ -44,6 +46,7 @@ export function InstantSearchNext<
 >({
   children,
   routing: passedRouting,
+  nonce: _nonce,
```
cungminh2710, line 49: don't know why babel complaining about nonce being already defined. So i use `_nonce` instead

Haroenv:
Hi, could you explain more why you want to have this change? the example given in the repository works already for server and client navigation. Is this for a different case?

Haroenv:
Can you explain more @cungminh2710?

cungminh2710:
There was some misunderstand from my end when seeing the error. Will close it now as it's working propperly in latest next version 👍

Haroenv:
Thanks for your contribution!

PR #6340, opened 2024-08-27
chore(dependencies): comprehensively fix node_modules of v4

Haroenv:
**Summary**

similar to #6336, but more comprehensive. it ensures that v4 is used nowhere outside of the `algoliasearch-v4` jail package by installing that dependency separately, as well as v5 separately to ensure dependencies are correct.

**Result**

top-level dependencies are v5, and v4 _only_ within the `algoliasearch-v4` dependency.


dhayab (approved):
👏 👏

PR #6354, opened 2024-09-10
fix(client): use client.search in sffv for v5

Haroenv:
**Summary**

v5 has a `searchForFacetValues` function, but it's single-index only. We thus need to detect it and use `client.search` directly. In a future major we can change this and only use `client.search`

**Result**

- v5 allows sffv in the type for the full client (repro: https://github.com/WavyWalk/reproduce-algolia-v5/blob/typing-5-3-0/src/App.tsx#L11)
- sffv isn't called in v5, but client.search instead

fixes https://github.com/algolia/algoliasearch-client-javascript/issues/1550


dhayab (approved):
Thanks for the fix :)

PR #6356, opened 2024-09-10
feat(experiences): add separate banner component

Haroenv:
**Summary**

adds a separate banner "widget" (uses connectHits) and ensures banner doesn't show inside hits

**Result**

[FX-3002]

[FX-3002]: https://algolia.atlassian.net/browse/FX-3002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/algolia-experiences/src/banner.tsx
+++ b/packages/algolia-experiences/src/banner.tsx
@@ -1,0 +1,98 @@
+/** @jsx h */
+
+import { cx } from 'instantsearch-ui-components';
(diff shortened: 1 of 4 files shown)


dhayab:
This implementation doesn't read from `renderingContent.widgets.banner` but instead relies on an explicit banner block, is that right?

PR #6357, opened 2024-09-10
docs(react-instantsearch): load `instantsearch.css` from packages

sarahdayan:
This imports `instantsearch.css` from the installed packages instead of jsDelivr so we can view style changes when running examples.


Haroenv (approved):
change in yarn.lock seems to be harmless

shaejaz (approved):

PR #6364, opened 2024-09-17
feat(experiences): implement custom element

aymeric-giraudet:
[FX-3008](https://algolia.atlassian.net/browse/FX-3008)

**Summary**

To allow usage within an SPA, be it completely client-side rendered or not (with Next.js for example), we need to provide a way for an experience to be loaded at any time.

**Result**

Created a custom element that fetches its own configuration and renders. Code looks a bit awkward since I reused functions meant for multiple configurations.
We should probably implement some caching later for configurations.

### DIFF
--- a/examples/js/algolia-experiences/package.json
+++ b/examples/js/algolia-experiences/package.json
@@ -3,9 +3,9 @@
   "version": "1.1.0",
   "private": true,
   "scripts": {
-    "start": "BABEL_ENV=parcel parcel index.html local.html",
-    "build": "BABEL_ENV=parcel parcel build index.html local.html --public-url .",
-    "website:examples": "BABEL_ENV=parcel parcel build index.html local.html --public-url . --dist-dir=../../../website/examples/js/algolia-experiences"
+    "start": "BABEL_ENV=parcel parcel index.html local.html custom-element.html",
+    "build": "BABEL_ENV=parcel parcel build index.html local.html custom-element.html --public-url .",
+    "website:examples": "BABEL_ENV=parcel parcel build index.html local.html custom-element.html --public-url . --dist-dir=../../../website/examples/js/algolia-experiences"
   },
   "browserslist": "firefox 68, chrome 78, IE 11",
   "devDependencies": {
--- a/packages/algolia-experiences/src/setup-instantsearch.ts
+++ b/packages/algolia-experiences/src/setup-instantsearch.ts
@@ -8,6 +8,9 @@ import { getElements, getSettings } from './get-information';
 import { configToIndex, injectStyles } from './render';
 import { error } from './util';
 
+import type { Settings } from './get-information';
+import type { IndexWidget } from 'instantsearch.js';
+
 declare global {
   interface Window {
     __search: InstantSearch;
@@ -24,6 +27,8 @@ export function setupInstantSearch() {
     });
     window.__search = search;
 
+    registerComponents(search, settings);
+
     const elements = getElements();
 
     injectStyles();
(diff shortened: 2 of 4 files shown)


examples/js/algolia-experiences/package.json
```diff
@@ -3,9 +3,9 @@
   "version": "1.1.0",
   "private": true,
   "scripts": {
-    "start": "BABEL_ENV=parcel parcel index.html local.html",
-    "build": "BABEL_ENV=parcel parcel build index.html local.html --public-url .",
-    "website:examples": "BABEL_ENV=parcel parcel build index.html local.html --public-url . --dist-dir=../../../website/examples/js/algolia-experiences"
+    "start": "BABEL_ENV=parcel parcel index.html local.html custom-element.html",
```
Haroenv, line 6: I wonder if this can be *.html. Can be done in other PR though

packages/algolia-experiences/src/setup-instantsearch.ts
```diff
@@ -39,3 +44,35 @@ export function setupInstantSearch() {
...
 }
+
+function registerComponents(search: InstantSearch, settings: Settings) {
+  class AlgoliaExperience extends HTMLElement {
+    static observedAttributes = ['experience-id'];
+
+    widgets: IndexWidget[] = [];
+
+    connectedCallback() {
+      const experienceId = this.getAttribute('experience-id');
```
Haroenv, line 55: wonder if we should use the same data-attribute so it's easier to switch between both modes?

Haroenv (approved):
good to have both options, probably we'll want to converge on only custom elements later but this gives us options. Thanks!

packages/algolia-experiences/src/setup-instantsearch.ts
aymeric-giraudet: The problem with this is that when loading initially it would attempt to render twice via the `querySelector` and within the lifecycle itself, as `<algolia-experiences data-experience-id="..." />` would match.
We would have to add more checks to ensure it doesn't happen.

Not sure there would be a scenario where you would wanna switch between both modes, except if we want to do custom elements only and ask people to change their code

Haroenv: I'm ok with custom experiences only, nobody has implemented this yet, maybe the querySelector code goes away all together then?

packages/algolia-experiences/src/setup-instantsearch.ts
```diff
@@ -75,6 +63,17 @@ function registerComponents(search: InstantSearch, settings: Settings) {
...
     }
+
+    attributeChangedCallback(name: string, oldValue: string, newValue: string) {
+      if (oldValue === null || name !== 'experience-id') {
+        return;
+      }
+
+      if (oldValue !== newValue) {
+        this.disconnectedCallback();
+        this.connectedCallback();
```
Haroenv, line 74: for later: the order should be fetch -> remove -> add, not remove -> fetch -> add, but I see how this current code is much simpler

packages/algolia-experiences/src/setup-instantsearch.ts
```diff
@@ -27,45 +27,33 @@ export function setupInstantSearch() {
...
+function registerComponent(search: InstantSearch, settings: Settings) {
   class AlgoliaExperience extends HTMLElement {
     static observedAttributes = ['experience-id'];
 
     widgets: IndexWidget[] = [];
 
     connectedCallback() {
-      const experienceId = this.getAttribute('experience-id');
-      if (!experienceId) {
+      const id = this.getAttribute('experience-id');
```
Haroenv, line 47: would there be less repetition if we call the attribute `data-id`? can swap before the GA still

Haroenv (approved):

packages/algolia-experiences/src/setup-instantsearch.ts
```diff
@@ -27,45 +27,33 @@ export function setupInstantSearch() {
...
         error('Experience ID is required');
         return;
       }
 
-      fetchConfiguration([experienceId], settings).then((configuration) => {
-        this.widgets = configToIndex(
-          configuration[0],
-          new Map([[experienceId, this]])
-        );
+      fetchConfiguration(id, settings).then((config) => {
```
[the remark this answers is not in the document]
aymeric-giraudet: It was still the case before as the endpoint serves one only right ? It's just that they rendered all at once and that if one failed, nothing would render.
Now each tag will fetch and render independently from each other.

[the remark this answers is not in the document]
Haroenv: ah yes I forgot it was in a promise.all

packages/algolia-experiences/src/setup-instantsearch.ts
aymeric-giraudet: Could be simple, I could wrap this into `fetchConfiguration(newValue).then` and it would pick up the cached configuration in `connectedCallback`

Haroenv: also good for me :)

shaejaz (approved):

PR #6370, opened 2024-09-24
feat(sessionStorage): allow a cache key suffix

Haroenv:
**Summary**

Adds a `key` option to sessionStorage cache.

This allows the use case of multiple infinite hits displayed on one page (like multi index with tabs).

**Result**

You can use the key if you're using multiple session storage caches / multiple InstantSearch instances to have your results not overridden.


packages/instantsearch.js/src/lib/infiniteHitsCache/__tests__/sessionStorage-key.test.ts
shaejaz, line 1: We probably want tests for `createInfiniteHitsSessionStorageCache` with no arguments and multiple calls to it in a single test as well now

shaejaz (approved):

packages/instantsearch.js/src/lib/infiniteHitsCache/__tests__/sessionStorage-key.test.ts
Haroenv: something was wrong with the mock or the test that made them all run at the same time, or at least not in a consistent order, that's why I put them in a separate test. IRL I don't see how there could be problems if you at least have unique keys

PR #6372, opened 2024-09-26
fix(next-app-router): mitigate memory leak

aymeric-giraudet:
**Summary**

Got reports of memory leaks with the Next.js App Router solution.
It seems partly related to `ServerInsertedHTML` or maybe just the fact that renders are streamed.

**Result**

Prior to the change, on an index with stringified results weighing ~50kb, for 500 concurrent requests we had a peak of 1GB which reduced to 600mb when idle.
By moving the initialisation of `createInsertHTML` so that it does not capture the scope of the whole component, it peaks at 600mb then reduces to 160mb when idle.

There are definitely more ways this could be improved as I see stringified results are kept in memory way after, so does certain React nodes with results in them. Should try to see with the Next.js team perhaps.

### DIFF
--- a/packages/react-instantsearch-nextjs/src/InitializePromise.tsx
+++ b/packages/react-instantsearch-nextjs/src/InitializePromise.tsx
@@ -10,7 +10,7 @@ import {
 
 import { htmlEscapeJsonString } from './htmlEscape';
 
-import type { SearchOptions } from 'instantsearch.js';
+import type { InitialResults, SearchOptions } from 'instantsearch.js';
 
 type InitializePromiseProps = {
   /**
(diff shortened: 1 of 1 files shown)


shaejaz (approved):

pkutsenko:
Hi @aymeric-giraudet It seems like next.js with its addInsertedHtml function causes a memory leak issue. If I comment out insertHTML call the memory leak is gone that of course breaks SSR. Have you considered not using ServerInsertedHTML but return <script> from the component JSX if it's possible?

aymeric-giraudet:
Hey @pkutsenko,

Tried changing it to a component on this branch : https://github.com/algolia/instantsearch/commit/68c8c577592b24b1ce83166d8c4c947db290022c and I get the exact same results, so that's not the issue.

I identified another memory leak in the Node.js requester in `algoliasearch`. Some handlers that retain responses as string are not cleared. By using the fetch requester instead (install the `@algolia/requester-fetch`) and use that instead : 
```js
import { liteClient as algoliasearch } from 'algoliasearch/lite';
import { createFetchRequester } from '@algolia/requester-fetch';
  appId,
  apiKey,
  { requester: createFetchRequester() }
);
```

_Note that I'm using `algoliasearch` v5 with `@algolia/requester-fetch` matching the same version, should still work with v4 though but you can update if it doesn't_

Now after 500 parallel requests it goes down to 40Mb at the end rather than 160Mb. That's using `react-instantsearch-nextjs` at `0.3.14` of course.

I still see some retained objects but I think it's Next.js/webpack at this point which we can't do anything about.

PR #6377, opened 2024-10-03
fix(instantsearch.js): fix user token not being set in time for the first query

shaejaz:
[FX-2955]

**Summary**
Double queries happen when the state of instantsearch and insights no longer matches at the time of the first query on page load.

- The `search-insights` `onUserTokenChange` and `onAuthenticatedUserTokenChange` callbacks are not able to set the `user-token` by the time the first search query is sent. When the `user-token` is changed by these callbacks, the state has changed from the first query. If a new query is scheduled after this(for example with dynamic widgets), the cache misses due to the changed state and second query is made.

- When `useCookie` is set to true, insights needs to create a anonymous token and save it as a cookie. Instantsearch would have the token updated with the callback from before, but this does not happen in time for the first query. So anonymous token generation and saving in cookies is now also handled by instantsearch.

- The insights queue is now being access in the middleware's `started` function to make sure any token set before instantsearch was started are part of the state by the time the first query is made.

[FX-2955]: https://algolia.atlassian.net/browse/FX-2955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/lib/__tests__/InstantSearch-integration-test.ts
+++ b/packages/instantsearch.js/src/lib/__tests__/InstantSearch-integration-test.ts
@@ -489,6 +489,7 @@ describe('network requests', () => {
             "params": {
               "clickAnalytics": true,
               "query": "",
+              "userToken": "cookie-key",
             },
           },
         ]
@@ -563,6 +564,7 @@ describe('network requests', () => {
             "params": {
               "clickAnalytics": true,
               "query": "",
+              "userToken": "cookie-key",
             },
           },
         ]
--- a/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -437,7 +437,7 @@ describe('insights', () => {
       );
     });
 
-    it('warns when userToken is not set', () => {
+    it.skip('warns when userToken is not set', () => {
       const { insightsClient, instantSearchInstance } = createTestEnvironment();
 
       instantSearchInstance.use(
@@ -777,38 +777,20 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
       expect(getUserToken()).toEqual('token-from-queue-before-init');
     });
 
-    it('handles multiple setUserToken calls before search.start()', async () => {
-      const { insightsClient } = createInsights();
-      const indexName = 'my-index';
-      const instantSearchInstance = instantsearch({
-        searchClient: createSearchClient({
-          // @ts-expect-error only available in search client v4
-          transporter: {
-            headers: {
-              'x-algolia-application-id': 'myAppId',
-              'x-algolia-api-key': 'myApiKey',
-            },
-          },
-        }),
-        indexName,
-      });
+    it('handles multiple setUserToken calls before search.start()', () => {
+      const { insightsClient, instantSearchInstance, getUserToken } =
+        createTestEnvironment();
 
-      const middleware = createInsightsMiddleware({
-        insightsClient,
-      });
-      instantSearchInstance.use(middleware);
+      instantSearchInstance.use(
+        createInsightsMiddleware({
+          insightsClient,
+        })
+      );
 
       insightsClient('setUserToken', 'abc');
       insightsClient('setUserToken', 'def');
 
-      instantSearchInstance.start();
-
-      await wait(0);
-
-      expect(
-        (instantSearchInstance.mainHelper!.state as PlainSearchParameters)
-          .userToken
-      ).toEqual('def');
+      expect(getUserToken()).toEqual('def');
     });
 
     it('searches once per unique userToken', async () => {
@@ -836,7 +818,8 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
     });
 
     it("doesn't search when userToken is falsy", async () => {
-      const { insightsClient, instantSearchInstance } = createTestEnvironment();
+      const { insightsClient, instantSearchInstance, getUserToken } =
+        createTestEnvironment();
 
       instantSearchInstance.addWidgets([connectSearchBox(() => ({}))({})]);
 
@@ -866,8 +849,8 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
           indexName: 'my-index',
           params: {
             clickAnalytics: true,
-
             query: '',
+            userToken: getUserToken(),
           },
         },
       ]);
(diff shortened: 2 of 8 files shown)


packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -192,11 +199,29 @@ export function createInsightsMiddleware<
           instantSearchInstance.emit('error', new Error(errorMessage));
         }
       },
+      // eslint-disable-next-line complexity
```
Haroenv, line 202: what is the complexity about? can it be avoided?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -192,11 +199,29 @@ export function createInsightsMiddleware<
...
         insightsClient('addAlgoliaAgent', 'insights-middleware');
 
         helper = instantSearchInstance.mainHelper!;
 
+        const { queue: queueAtStart } = insightsClient;
+
+        if (Array.isArray(queueAtStart)) {
+          [queuedUserToken, queuedAuthenticatedUserToken] = [
+            'setUserToken',
+            'setAuthenticatedUserToken',
```
Haroenv, line 213: no init here, maybe we should get the value again?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -287,12 +369,30 @@ export function createInsightsMiddleware<
...
+        } else if (anonymousUserToken) {
+          setUserToken(anonymousUserToken, anonymousUserToken, undefined);
+
+          if (insightsInitParams?.useCookie || queuedInitParams?.useCookie) {
+            const d = new Date();
+            d.setTime(
+              d.getTime() +
+                (insightsInitParams?.cookieDuration ||
+                  queuedInitParams?.cookieDuration ||
+                  30 * 24 * 60 * 60 * 1000 * 6)
```
Haroenv, line 381: maybe a comment that says "six months"?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -287,12 +369,30 @@ export function createInsightsMiddleware<
             queuedUserToken,
             queuedAuthenticatedUserToken
           );
+        } else if (anonymousUserToken) {
+          setUserToken(anonymousUserToken, anonymousUserToken, undefined);
```
Haroenv, line 373: would like to see the setUserToken function simplified (can be separate pr) to not have three arguments

Haroenv (commented):
looks good! probably just need added tests

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
shaejaz: I'll try to extract some functionality into separate functions, should fix this.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
shaejaz: Oop, yeah this part needs to be cleaned

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -437,7 +437,7 @@ describe('insights', () => {
       );
     });
 
-    it('warns when userToken is not set', () => {
```
shaejaz, line 440: Not sure if this is needed now, since an anonymous token would always be set for queries.

packages/instantsearch.js/src/lib/__tests__/InstantSearch-integration-test.ts
```diff
@@ -489,6 +489,7 @@ describe('network requests', () => {
             "params": {
               "clickAnalytics": true,
               "query": "",
+              "userToken": "cookie-key",
```
shaejaz, line 492: if a cookie is present, it will always be used if the token is not no other way

Haroenv: can you clarify what you mean? not sure I understand your comment

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
Haroenv: yes, there's no realistic way to get this warning anymore. Still fine to keep at as a worst-case warning but I don't see how you could _not_ have a token anymore with these changes.

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -777,38 +777,20 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
...
 
-      instantSearchInstance.start();
-
-      await wait(0);
-
-      expect(
-        (instantSearchInstance.mainHelper!.state as PlainSearchParameters)
-          .userToken
-      ).toEqual('def');
+      expect(getUserToken()).toEqual('def');
```
Haroenv, line 793: we also want to test the token of the search parameters is `def` here

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -269,13 +306,58 @@ export function createInsightsMiddleware<
...
+        }
+
+        let authenticatedUserTokenFromInit: string | undefined;
+        let userTokenFromInit: string | undefined;
+
+        // By the time the first query is sent, the token would not be set by the insights
+        // onChange callbacks. It is explicitly being set here so that the first query
+        // has the initial tokens set and inturn a second query isn't automatically made
+        // when the onChange callback actually changes the state.
+        if (insightsInitParams) {
```
Haroenv, line 330:
```suggestion
        // When the first query is sent, the token is possibly not yet be set by
        // the insights onChange callbacks (if insights isn't yet loaded).
        // It is explicitly being set here so that the first query has the
        // initial tokens set and ensure a second query isn't automatically
        // made when the onChange callback actually changes the state.
        if (insightsInitParams) {
```

tests/common/shared/insights.ts
```diff
@@ -76,7 +76,7 @@ export function createInsightsTests(
         });
 
         // initial calls because the middleware is attached
-        expect(window.aa).toHaveBeenCalledTimes(6);
+        expect(window.aa).toHaveBeenCalledTimes(7);
```
Haroenv, line 79: the change here is because of the automatic token being set?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -269,13 +306,58 @@ export function createInsightsMiddleware<
...
+
         // We consider the `userToken` or `authenticatedUserToken` before an
-        // `init` call of higher importance than one from the queue.
+        // `init` call of higher importance than one from the queue and ones set
+        // from the init props to be higher than that.
+        const tokenFromInit =
+          authenticatedUserTokenFromInit || userTokenFromInit;
         const tokenBeforeInit =
           authenticatedUserTokenBeforeInit || userTokenBeforeInit;
         const queuedToken = queuedAuthenticatedUserToken || queuedUserToken;
```
Haroenv, line 346: this should be `tokenFromQueue`

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
Haroenv: the test should be removed, not skipped :)

Haroenv (commented):
lgtm when the skipped test is removed

Haroenv (approved):

dhayab (approved):
Kudos for solving this 👏

PR #6378, opened 2024-10-03
refactor: remove hogan.js

aymeric-giraudet:
**Summary**

[FX-3031](https://algolia.atlassian.net/browse/FX-3031)

**Result**

- Templates should be functions or strings. Functions returning simple strings and simple strings directly work, if it returns a string with HTML tags these will be escaped by Preact.
  - Updated all default templates and tests accordingly.
- Took the opportunity to remove `bindEvent` as it was linked to templates.
- Still kept the `Template` component as templates are not valid components, they take 2 arguments.
- Removed hogan dependencies

[FX-3031]: https://algolia.atlassian.net/browse/FX-3031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/instantsearch.js/src/__tests__/common-widgets.test.tsx
```diff
@@ -418,7 +418,7 @@ const testSetups: TestSetupsMap<TestSuites> = {
           container: document.body.appendChild(document.createElement('div')),
           ...widgetParams,
           templates: {
-            labelText: label,
+            labelText: label ? () => label : undefined,
```
Haroenv, line 421: does this imply that string templates aren't supported anymore? could we check if it's a string and just render as a string instead of requiring a function?

aymeric-giraudet: That's what I thought of doing at first until I saw we print a warning when string templates are used, saying these will be deprecated : https://github.com/algolia/instantsearch/blob/b24c858b66de50d8dde6c85ce7ffc90ab0f712e8/packages/instantsearch.js/src/components/Template/Template.tsx#L77C1-L91C6

Another upside of keeping function templates only is they would easily see it break if they keep hogan or a HTML string when upgrading, whereas if we still support strings it could easily be missed (`<p>Something</p>` rendered as escaped HTML for example)

shaejaz (approved):

packages/instantsearch.js/src/__tests__/common-widgets.test.tsx
Haroenv: The warning was mainly to deprecate hogan strings as we couldn't tell them apart from simple strings. I see how it's simpler for us, but the migration for simple strings (most of the files changed in this pr) shows that it might be annoying for customers to do. What do you think?

Haroenv (approved):
not all these changes are needed anymore now, but no big deal, they're still valid

packages/instantsearch.js/src/lib/templating/renderTemplate.ts
```diff
@@ -7,99 +6,41 @@ import {
...
+    return template;
   }
 
-  if (typeof template === 'function') {
-    // @MAJOR no longer pass bindEvent when string templates are removed
-    const params = (bindEvent || {}) as TemplateParams;
-
-    params.html = html;
-    (params as any).sendEvent = sendEvent;
-    params.components = {
```
Haroenv, line 78: so glad this is gone haha

redox: me too !!!

PR #6380, opened 2024-10-07
fix(configure): ignore data attributes

Haroenv:
**Summary**

ignore data attributes in configure

**Result**

fixes https://github.com/getsentry/sentry-javascript/issues/12720#issuecomment-2396247718

I'm not yet convinced this is the right approach for all cases (may cause a slowdown possibly, although we already were spreading the props). Alternatively this could be in `setQueryParameter` or similar possibly, but would make that function more expensive.

In the mean time, anyone using the sentry plugin can use this patch

### DIFF
--- a/packages/react-instantsearch-core/src/components/Configure.tsx
+++ b/packages/react-instantsearch-core/src/components/Configure.tsx
@@ -5,7 +5,17 @@ import type { UseConfigureProps } from '../connectors/useConfigure';
 export type ConfigureProps = UseConfigureProps;
 
 export function Configure(props: ConfigureProps) {
-  useConfigure({ ...props }, { $$widgetType: 'ais.configure' });
+  useConfigure(
+    Object.keys(props).reduce<ConfigureProps>((acc, key) => {
+      if (key.slice(0, 6) === 'data-') {
+        return acc;
+      }
+      // @ts-ignore
+      acc[key] = props[key];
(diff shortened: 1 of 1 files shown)


JonasBa:
Thank you @Haroenv ❤️

packages/react-instantsearch-core/src/components/Configure.tsx
```diff
@@ -5,7 +5,17 @@ import type { UseConfigureProps } from '../connectors/useConfigure';
 export type ConfigureProps = UseConfigureProps;
 
 export function Configure(props: ConfigureProps) {
-  useConfigure({ ...props }, { $$widgetType: 'ais.configure' });
+  useConfigure(
+    Object.keys(props).reduce<ConfigureProps>((acc, key) => {
```
JonasBa, line 9: Worth leaving a comment here about Sentry being the cause of this issue for future readers?

victordelva (approved):

victordelva:
Thank you, @Haroenv 
I’m currently dealing with this issue between Algolia and Sentry, and this would be helpful.

Do you know of any workaround in the meantime?

Haroenv:
@victordelva you can apply this patch in your code base using for example patch-package. For the moment it seems that sentry is more responsible for patching their code to avoid adding attributes everywhere.

PR #6386, opened 2024-10-09
feat(experiences): add a default layout carousel for recommend widgets

shaejaz:
**Summary**
This adds a default Carousel layout for recommend widgets in experiences

Additionally, dependencies have been added to the experiences example to fix the build order in `yarn build`

[FX-3109]

[FX-3109]: https://algolia.atlassian.net/browse/FX-3109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/examples/js/algolia-experiences/package.json
+++ b/examples/js/algolia-experiences/package.json
@@ -8,6 +8,10 @@
     "website:examples": "BABEL_ENV=parcel parcel build *.html --public-url . --dist-dir=../../../website/examples/js/algolia-experiences"
   },
   "browserslist": "firefox 68, chrome 78, IE 11",
+  "dependencies": {
+    "algolia-experiences": "1.3.0",
+    "instantsearch.css": "8.5.1"
+  },
   "devDependencies": {
     "@babel/core": "7.15.5",
     "@parcel/core": "2.10.0",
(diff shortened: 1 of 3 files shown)


Haroenv (approved):
makes sense, thanks! maybe we can refactor this at a later point to have one function that deals with item template rendering as it's the same between hits and recommend. Good separate for now too though

examples/js/algolia-experiences/package.json
```diff
@@ -8,6 +8,10 @@
     "website:examples": "BABEL_ENV=parcel parcel build *.html --public-url . --dist-dir=../../../website/examples/js/algolia-experiences"
   },
   "browserslist": "firefox 68, chrome 78, IE 11",
+  "dependencies": {
```
shaejaz, line 11: Technically these are only required when using the `local.html` file, but unsure how else to get the build dependencies fixed otherwise.

Haroenv: yeah good call

shaejaz:
> makes sense, thanks! maybe we can refactor this at a later point to have one function that deals with item template rendering as it's the same between hits and recommend. Good separate for now too though

Yeah, initially I tried to get it to use the same item template, but I think the types are a bit weird in that the item templates differ between the hits widgets and recommend widgets. Not sure if that's intentional, but since there might be the use case that we want separate item templates anyway, I didn't want to spend too much time debugging the type errors.

PR #6389, opened 2024-10-10
poc(composition): top-level composition

Haroenv:
- no nested index support
- simpler than alternative
- facets should work out of the box (maybe code in SearchResponse needs to be conditional too?)

alternative: #6388

### DIFF
--- a/examples/js/getting-started/src/app.js
+++ b/examples/js/getting-started/src/app.js
@@ -1,6 +1,5 @@
 import { liteClient as algoliasearch } from 'algoliasearch/lite';
 import instantsearch from 'instantsearch.js';
-import { carousel } from 'instantsearch.js/es/templates';
 import {
   configure,
   hits,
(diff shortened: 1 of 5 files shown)


examples/js/getting-started/src/app.js
```diff
@@ -1,25 +1,26 @@
...
 } from 'instantsearch.js/es/widgets';
 
 import 'instantsearch.css/themes/satellite.css';
 
 const searchClient = algoliasearch(
-  'latency',
-  '6be0576ff61c053d5f9a3225e2a90f76'
+  'DIYPADIATS',
+  'c96176e1b36590680fb3d36bc480d592',
+  { authMode: 'WithinHeaders' }
```
Haroenv, line 17: wonder if this breaks something. Ideally composition API supports withinQueryParams so that it's a SimpleRequest

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1669,6 +1674,68 @@ AlgoliaSearchHelper.prototype._recommend = function () {
...
+      'Only one query is allowed when using a composition. Please use the regular search method.'
+    );
+  }
+  var query = queries[0];
+
+  try {
+    this.client.transporter
+      .request({
+        method: 'POST',
+        path: '/1/compositions/' + query.indexName '/run',
```
Haroenv, line 1724:
```suggestion
        path: '/1/compositions/' + query.indexName + '/run',
```

PR #6392, opened 2024-10-17
fix(refinement-list): re-apply focus on facet checkbox after render

dhayab:
**Summary**

In InstantSearch.js, focus is lost when toggling a facet value in a refinement list. This is because the component is destroyed and re-created during re-render, and [Preact does not explicitly refocus the newly created component](https://github.com/preactjs/preact/issues/3242) (whereas React does it).

This PR adds code that stores the last refined value and applies focus to its DOM element on re-render.

CR-6976

### DIFF
--- a/packages/instantsearch.js/src/components/RefinementList/RefinementList.tsx
+++ b/packages/instantsearch.js/src/components/RefinementList/RefinementList.tsx
@@ -108,7 +108,10 @@ class RefinementList<TTemplates extends Templates> extends Component<
 > {
   public static defaultProps = defaultProps;
 
+  private listRef = createRef<HTMLUListElement>();
   private searchBox = createRef<SearchBox>();
+
+  private lastRefinedValue: string | undefined = undefined;
 
   public shouldComponentUpdate(
     nextProps: RefinementListPropsWithDefaultProps<TTemplates>
@@ -122,6 +125,7 @@ class RefinementList<TTemplates extends Templates> extends Component<
   }
 
   private refine(facetValueToRefine: string) {
+    this.lastRefinedValue = facetValueToRefine;
     this.props.toggleRefinement(facetValueToRefine);
   }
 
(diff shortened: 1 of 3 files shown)


packages/instantsearch.js/src/components/RefinementList/RefinementList.tsx
```diff
@@ -270,6 +274,20 @@ class RefinementList<TTemplates extends Templates> extends Component<
     }
   }
 
+  /**
+   * This sets focus on the last refined input element after a render
+   * because Preact does not perform it automatically.
+   * @see https://github.com/preactjs/preact/issues/3242
+   */
+  public componentDidUpdate() {
```
Haroenv, line 282: too bad to have this fix in the non-unified components 🤔

dhayab: Yes, we should brace for a few headaches when we'll migrate all of this into the unified components package 😰.

packages/instantsearch.js/src/components/RefinementList/__tests__/RefinementList-test.tsx
```diff
@@ -692,5 +693,57 @@ describe('RefinementList', () => {
 
       expect(toggleRefinement).toHaveBeenCalledTimes(0);
     });
+
+    it('should keep focus on toggled input between re-renders', () => {
```
Haroenv, line 697: this test should be in cts as well so it won't be reintroduced when we migrate components.

Haroenv (commented):
fix looks like a good workaround for now. 

- is this happening anywhere else? What makes refinement list special that it doesn't need to be fixed somewhere else? no form elements in other components?
- test should be in CTS, maybe for all components?

Haroenv (approved):

Haroenv:
I guess it won't matter if it happens anywhere else? focus can still be lost on a link right?

PR #6395, opened 2024-10-17
fix(dependencies): Fixed semver of `peerDependencies` to be strict.

shad0wflame:
**Summary**

This will allow Deno to install the package as it won't do it with a loose semver.

image

You can check the semver using the official npm website.
https://semver.npmjs.com/#syntax-examples

image

**Result**

image


Haroenv (approved):
thanks!

PR #6403, opened 2024-10-18
fix(helpers): remove deprecated helpers

Haroenv:
BREAKING CHANGE: remove instantsearch.highlight, instantsearch.reverseHighlight, instantsearch.snippet, instantsearch.reverseSnippet. Instead use the `components` available from a template
BREAKING CHANGE: remove instantsearch.insights, instead use sendEvent
BREAKING CHANGE: remove instantsearch.getInsightsAnonymousUserToken
BREAKING CHANGE: no longer listen to events caused by bindEvents


examples/js/media/src/widgets/Articles.ts
```diff
@@ -56,18 +57,12 @@ function createHit(
               <span class="card-timestamp">${date}</span>
 
               <h1 class="card-title">
-                ${highlight({
-                  attribute: 'title',
-                  hit,
-                })}
+                ${(hit as any)._highlightResult.title.value}
```
Haroenv, line 60: this is not as neat as it used to be. Options are:
1. expose a top-level highlight-like function for connector usage
2. use preact and htm in the example?
3. convert this example to not need a connector?
4. let it be

Haroenv: For this PR, I suggest we let this be, and in a separate PR we introduce a way to use htm and preact in connectors (user land most likely)

aymeric-giraudet (approved):

PR #6415, opened 2024-10-28
feat(tests): remove ie11 targeting tests

dhayab:
**Summary**

This PR removes tests targeting IE11.

FX-3054


Haroenv (approved):
makes sense, we'll add something that specifies the next lowest common denominator after

shaejaz (approved):

PR #6417, opened 2024-10-29
refactoring: remove obsolete polyfills

dhayab:
**Summary**

This PR removes a number of polyfills we had in the library that are no longer necessary following the end of IE11 support.


packages/vue-instantsearch/src/components/StateResults.vue
```diff
@@ -16,7 +16,6 @@
 <script>
 import { createSuitMixin } from '../mixins/suit';
 import { createWidgetMixin } from '../mixins/widget';
-import { _objectSpread } from '../util/polyfills';
```
Haroenv, line 19: this is for babel in Vue InstantSearch, not for ie11

dhayab: I don't remember well why we resort to this, does the build fail if we use native object spreading in the codebase for Vue?

Haroenv: yes, it was something along those lines, but if it works feel free to change

dhayab: Alright! I'll check this once I've merged FX-3053 into this branch.

packages/vue-instantsearch/src/mixins/widget.js
```diff
@@ -31,10 +30,10 @@ export const createWidgetMixin = (
...
     if (typeof connector === 'function') {
       this.factory = connector(this.updateState, () => {});
-      this.widget = _objectSpread(
-        this.factory(this.widgetParams),
-        additionalProperties
-      );
+      this.widget = {
+        ...this.factory(this.widgetParams),
+        ...additionalProperties,
+      };
```
Haroenv, line 36: I believe this is what's causing the test failures. Using `Object.assign(this.factory(this.widgetParams), additionalProperties)` likely would pass; otherwise change the assertions to not be strict equal

Haroenv (approved):

PR #6422, opened 2024-11-04
feat(babel): upgrade and set newer browser targets

aymeric-giraudet:
**Summary**

[FX-5053](https://algolia.atlassian.net/browse/FX-3053)

**Result**

- Updated `babel` as well as some plugins
- Changed `browserslist` target
- Added new ESLint rule that outputs errors when using unsupported APIs


.codesandbox/ci.json
```diff
@@ -18,5 +18,5 @@
     "packages/instantsearch-ui-components",
     "packages/instantsearch-core"
   ],
-  "node": "16"
+  "node": "18"
```
aymeric-giraudet, line 21: This works now ! Node 18 will leave maintenance in April 2025 though

babel.config.js
```diff
@@ -61,46 +61,6 @@ module.exports = (api) => {
         ],
       },
     ],
-    // this plugin is used to test if we need polyfills, not to actually insert them
```
aymeric-giraudet, line 64: New ESLint rule does that and seems more up to date

packages/algoliasearch-helper/documentation-src/content/js/main.js
```diff
@@ -110,6 +110,7 @@ function activeLinks(sidebarContainer) {
 
 function getPositionsKeyElements(sidebarContainer) {
   const sidebarBBox = sidebarContainer.getBoundingClientRect();
+  // eslint-disable-next-line compat/compat
```
aymeric-giraudet, line 113: Doesn't work in with Firefox < 59 but I guess it's fine

packages/vue-instantsearch/rollup.config.js
```diff
@@ -54,11 +53,6 @@ function outputs(vueVersion) {
     vuePlugin({ compileTemplate: true, css: false }),
     commonjs(),
     json(),
-    buble({
-      transforms: {
-        dangerousForOf: true,
-      },
-    }),
```
aymeric-giraudet, line 61: It's unmaintained and not needed anymore

.codesandbox/ci.json
Haroenv: finally! we can revive the vite/astro PRs too then

packages/algoliasearch-helper/documentation-src/content/js/main.js
Haroenv: this is the documentation site, can we disable it from this lint rule instead?

packages/vue-instantsearch/rollup.config.js
Haroenv: we used buble back in the day as the output was smaller than babel. Fine to switch to a maintained alternative though

Haroenv (approved):

packages/algolia-experiences/src/render.tsx
```diff
@@ -257,17 +254,14 @@ function blockToWidget(child: Block, container: HTMLElement): Widget[] {
                 children = ch.children.map(renderChild);
               }
 
-              const attributes = Object.fromEntries(
```
Haroenv, line 260: is this actually not allowed? from checking caniuse I see older chrome not having this but the other browsers seem ok. Should we have chrome 74 as minimum instead?

packages/vue-instantsearch/rollup.config.js
aymeric-giraudet: With/without babel it's a 0.10kb difference only, I forgot to replace it with babel, just did !

packages/algolia-experiences/src/render.tsx
aymeric-giraudet: Just crunched the numbers and we would lose 1% of users :/

shaejaz (approved):

PR #6437, opened 2024-11-14
feat: composition

e-krebs:
**Summary**

This PR currently adds support for Composition using a (currently not published yet) Composition API Client:
- new `compositionID` optional parameter when initializing InstantSearch (this PR)
- don't allow index widget when in composition mode #6459
- make composition compatible with vue #6458
- handle disjunctive facets #6485
- add searchForFacetValues #6489
- make composition work server-side + typing #6509 
- tests #6510

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -34,7 +34,7 @@
     },
     {
       "path": "packages/vue-instantsearch/vue2/cjs/index.js",
-      "maxSize": "20.25 kB"
+      "maxSize": "20.50 kB"
     },
     {
       "path": "packages/vue-instantsearch/vue3/cjs/index.js",
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -2,11 +2,11 @@
   "files": [
     {
       "path": "packages/algoliasearch-helper/dist/algoliasearch.helper.js",
-      "maxSize": "41.25 kB"
+      "maxSize": "41.63 kB"
     },
     {
       "path": "packages/algoliasearch-helper/dist/algoliasearch.helper.min.js",
-      "maxSize": "13.25 kB"
+      "maxSize": "13.40 kB"
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
@@ -18,7 +18,7 @@
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",
-      "maxSize": "51.25 kB"
+      "maxSize": "51.50 kB"
     },
     {
       "path": "packages/react-instantsearch/dist/umd/ReactInstantSearch.min.js",
@@ -26,11 +26,11 @@
     },
     {
       "path": "packages/vue-instantsearch/vue2/umd/index.js",
-      "maxSize": "68.75 kB"
+      "maxSize": "68.96 kB"
     },
     {
       "path": "packages/vue-instantsearch/vue3/umd/index.js",
-      "maxSize": "69 kB"
+      "maxSize": "69.27 kB"
     },
     {
       "path": "packages/vue-instantsearch/vue2/cjs/index.js",
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -18,7 +18,7 @@
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",
-      "maxSize": "51.50 kB"
+      "maxSize": "51.75 kB"
     },
     {
       "path": "packages/react-instantsearch/dist/umd/ReactInstantSearch.min.js",
@@ -26,7 +26,7 @@
     },
     {
       "path": "packages/vue-instantsearch/vue2/umd/index.js",
-      "maxSize": "69.00 kB"
+      "maxSize": "69.25 kB"
     },
     {
       "path": "packages/vue-instantsearch/vue3/umd/index.js",
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -2,11 +2,11 @@
   "files": [
     {
       "path": "packages/algoliasearch-helper/dist/algoliasearch.helper.js",
-      "maxSize": "41.63 kB"
+      "maxSize": "41.75 kB"
     },
     {
       "path": "packages/algoliasearch-helper/dist/algoliasearch.helper.min.js",
-      "maxSize": "13.40 kB"
+      "maxSize": "13.50 kB"
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
@@ -26,11 +26,11 @@
     },
     {
       "path": "packages/vue-instantsearch/vue2/umd/index.js",
-      "maxSize": "68.96 kB"
+      "maxSize": "69.00 kB"
     },
     {
       "path": "packages/vue-instantsearch/vue3/umd/index.js",
-      "maxSize": "69.27 kB"
+      "maxSize": "69.50 kB"
     },
     {
       "path": "packages/vue-instantsearch/vue2/cjs/index.js",
--- a/examples/js/getting-started/package.json
+++ b/examples/js/getting-started/package.json
@@ -9,6 +9,7 @@
     "lint:fix": "npm run lint -- --fix"
   },
   "dependencies": {
+    "@algolia/client-composition": "0.0.1-alpha.1",
     "algoliasearch": "5.1.1",
     "instantsearch.css": "8.5.1",
     "instantsearch.js": "4.75.6"
@@ -22,5 +23,8 @@
   },
   "@parcel/resolver-default": {
     "packageExports": true
+  },
+  "alias": {
+    "@algolia/client-composition": "../../../node_modules/@algolia/client-composition/dist/builds/browser.js"
   }
 }
--- a/examples/js/getting-started/package.json
+++ b/examples/js/getting-started/package.json
@@ -1,6 +1,6 @@
 {
   "name": "example-instantsearch-getting-started",
-  "version": "1.25.0",
+  "version": "1.26.0",
   "private": true,
   "scripts": {
     "start": "BABEL_ENV=parcel parcel index.html products.html --port 3000",
@@ -11,7 +11,7 @@
   "dependencies": {
     "algoliasearch": "5.1.1",
     "instantsearch.css": "8.5.1",
-    "instantsearch.js": "4.75.6"
+    "instantsearch.js": "4.75.7"
   },
   "devDependencies": {
     "@babel/core": "7.15.5",
--- a/examples/js/getting-started/package.json
+++ b/examples/js/getting-started/package.json
@@ -9,7 +9,6 @@
     "lint:fix": "npm run lint -- --fix"
   },
   "dependencies": {
-    "@algolia/client-composition": "0.0.1-alpha.1",
     "algoliasearch": "5.1.1",
     "instantsearch.css": "8.5.1",
     "instantsearch.js": "4.75.6"
@@ -23,8 +22,5 @@
   },
   "@parcel/resolver-default": {
     "packageExports": true
-  },
-  "alias": {
-    "@algolia/client-composition": "../../../node_modules/@algolia/client-composition/dist/builds/browser.js"
   }
 }
--- a/examples/js/getting-started/src/app.js
+++ b/examples/js/getting-started/src/app.js
@@ -1,6 +1,5 @@
-import { liteClient as algoliasearch } from 'algoliasearch/lite';
+import { compositionClient } from '@algolia/client-composition';
 import instantsearch from 'instantsearch.js';
-import { carousel } from 'instantsearch.js/es/templates';
 import {
   configure,
   hits,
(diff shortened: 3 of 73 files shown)


examples/js/getting-started/package.json
```diff
@@ -9,6 +9,7 @@
     "lint:fix": "npm run lint -- --fix"
   },
   "dependencies": {
+    "@algolia/client-composition": "0.0.1-alpha.1",
```
Haroenv, line 12: is there a way to publish this publicly? otherwise it won't be mergeable

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -67,6 +67,13 @@ export type InstantSearchOptions<
    */
   indexName?: string;
 
+  /**
+   * The objectID of the composition.
+   * If this is passed, the composition API will be used for search.
+   * Multi-index search and recommend is not supported with this option.
```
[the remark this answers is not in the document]
e-krebs: ha yes, you're right !

examples/js/getting-started/package.json
e-krebs: Not at the moment, I'm afraid.
We might end up with a lite public client but it's still being discussed.

I can remove it from the example package once we decide we are ready to merge.
Hopefully by the time I create a dedicated example this package thing will be resolved

examples/js/getting-started/src/app.js
```diff
@@ -1,27 +1,39 @@
-import { liteClient as algoliasearch } from 'algoliasearch/lite';
+// import { liteClient as algoliasearch } from 'algoliasearch/lite';
+import { compositionClient } from '@algolia/client-composition';
```
Haroenv, line 2: as we're not really using composition yet, this PR can be done without the dependency, and we use the `runComposition` method in a separate PR?

package.json
```diff
@@ -126,6 +126,7 @@
     "places.js": "1.17.1",
     "prettier": "^2.4.1",
     "prop-types": "15.6.2",
+    "punycode": "^1.4.1",
```
Haroenv, line 129: what's this needed for?

package.json
```diff
@@ -140,6 +141,7 @@
     "rollup-plugin-uglify": "6.0.4",
     "shelljs": "0.8.5",
     "shipjs": "0.26.0",
+    "stream-http": "^3.1.0",
```
Haroenv, line 144: what's this needed for?

packages/algoliasearch-helper/index.d.ts
```diff
@@ -31,7 +42,7 @@ import type {
  */
 declare function algoliasearchHelper(
   client: SearchClient,
-  index: string,
+  index: string | IndexOptions,
```
Haroenv, line 45: this design implies the same helper should be able to search and composition, but actually you won't be able to do that. What about using the index parameter directly as composition id in the helper? it saves on needed changes

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1578,6 +1592,78 @@ AlgoliaSearchHelper.prototype._search = function (options) {
...
+  var query = queries[0];
+
+  // TODO: delete once allowed in the Composition API
+  delete query.params.maxValuesPerFacet;
+  delete query.params.facets;
+
+  const { compositionID, params } = query;
+  try {
+    this.client
+      .search({ compositionID, requestBody: { params } })
```
Haroenv, line 1654: this is temporary I presume?

examples/js/getting-started/package.json
Haroenv: yes let's add a dedicated example

package.json
e-krebs: this is added by parcel while adding `@algolia/client-composition`, I'll clean this up when removing it

package.json
e-krebs: this is also added by parcel while adding @algolia/client-composition, I'll clean this up when removing it

examples/js/getting-started/src/app.js
e-krebs: I 100% agree since `@algolia/client-composition` is not public anyway

packages/algoliasearch-helper/src/algoliasearch.helper.js
e-krebs: no it isn't.
I named the method on the composition API client `search`

packages/algoliasearch-helper/index.d.ts
e-krebs: I would have loved to do that, but I need a way for the helper to know whether this parameter is a compositionID or an indexName to save it accordingly.
This is how I managed to achieve this, happy to change it if you have a better implementation idea.

Otherwise, one small improvement we could do is to update `IndexOptions` to:
```ts
type IndexOptions = { compositionID: string; index?: never } | { index: string; compositionID?: never }
```
it will imply changes here but that's manageable

packages/algoliasearch-helper/src/algoliasearch.helper.js
e-krebs: I pushed an update so that the query object is correctly build from the start (uglify was complaining)

packages/algoliasearch-helper/index.d.ts
e-krebs: in its current version, that `index` value is saved on the helper state in an `index` property.
- that `index` property name is not correct in the context of composition
- we would need to remove that `index` property from the state before/when building the query for composition (otherwise the query will be rejected)

bundlesize.config.json
```diff
@@ -2,35 +2,35 @@
   "files": [
     {
       "path": "packages/algoliasearch-helper/dist/algoliasearch.helper.js",
-      "maxSize": "41.25 kB"
+      "maxSize": "41.63 kB"
```
Haroenv, line 5: these numbers are too precise, please round up to the nearest .25kb

Haroenv (commented):
overall I think this is the way to go, but still think composition id isn't needed in the helper constructor if we assume index name === composition id (because you don't search if you compose instead)

packages/algoliasearch-helper/index.d.ts
e-krebs: there is no index name when dealing with composition, and there shouldn't be any (at least for now).

if we use a single parameter for storing either the index name or the composition ID, then it should not be named index.
it'll be very confusing otherwise

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1578,6 +1592,78 @@ AlgoliaSearchHelper.prototype._search = function (options) {
...
+ * @return {undefined} does not return anything
+ * @fires search
+ * @fires result
+ * @fires error
+ */
+AlgoliaSearchHelper.prototype._runComposition = function () {
+  var state = this.state;
+  var states = [];
+  var mainQueries = [];
+  const thisCompositionID = this.compositionID;
```
Haroenv, line 1608:
```suggestion
  var thisCompositionID = this.compositionID;
```

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -588,7 +588,7 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
       // level is aware of the client.
       helper = algoliasearchHelper(
         {} as SearchClient,
-        parameters.index,
+        { index: parameters.index, compositionID: mainHelper.compositionID },
```
Haroenv, line 591: no longer needed if composition is a SearchParameter

packages/algoliasearch-helper/index.d.ts
Haroenv: Indeed. To solve this as discussed we will accept compositionID as a search parameter and pass it accordingly to the index widget and its state. In sortBy that parameter is used as well, and the top-level "indexName" parameter to the helper constructor change can be reverted (as we'll always use the composition id from a derived state, not the top-level one)

e-krebs: this should now be implemented 🙂

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -565,16 +576,23 @@ See documentation: ${createDocumentationLink({
     // we need to respect this helper as a way to keep all listeners correct.
     const mainHelper =
       this.mainHelper ||
-      algoliasearchHelper(this.client, this.indexName, undefined, {
-        persistHierarchicalRootCount: this.future.persistHierarchicalRootCount,
-      });
+      algoliasearchHelper(
+        this.client,
+        this.indexName,
+        this.compositionID ? { compositionID: this.compositionID } : undefined,
```
Haroenv, line 582: this should be in the index derived helper instead of here, I'm not convinced these parameters survive `refreshUiState` for example

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -341,10 +350,12 @@ See documentation: ${createDocumentationLink({
     this.future = future;
     this.insightsClient = insightsClient;
     this.indexName = indexName;
+    this.compositionID = compositionID;
     this.helper = null;
     this.mainHelper = null;
     this.mainIndex = index({
-      indexName,
+      // we use an index widget to render compositions
+      indexName: this.compositionID || this.indexName,
```
Haroenv, line 358:
```suggestion
      indexName,
      compositionID,
```

(and accept and read the compositionID in index widget, to put it in the state in mergeSearchParameters)

Haroenv (commented):
looks already cleaner, now just the sortBy to do and moving the state to the index (IMO)

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1578,6 +1582,76 @@ AlgoliaSearchHelper.prototype._search = function (options) {
...
+  if (queries.length > 1) {
+    throw new Error('Only one query is allowed when using a composition.');
+  }
+
+  var query = queries[0];
+
+  // eslint-disable-next-line no-warning-comments
+  // TODO: delete once allowed in the Composition API
+  delete query.requestBody.params.maxValuesPerFacet;
+  delete query.requestBody.params.facets;
```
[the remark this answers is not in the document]
e-krebs: my reasoning for separating deleted parameters in 2 groups is: those 2 will be deleted before we release while the other won't

packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -341,10 +350,12 @@ See documentation: ${createDocumentationLink({
     this.future = future;
     this.insightsClient = insightsClient;
     this.indexName = indexName;
+    this.compositionID = compositionID;
     this.helper = null;
     this.mainHelper = null;
     this.mainIndex = index({
-      indexName,
+      // we use an index widget to render compositions
```
Haroenv, line 357: clearer comment to explain that we're able to do this because there's only one composition index allowed for now.

packages/instantsearch.js/src/lib/InstantSearch.ts
e-krebs: as discussed since then, in order to be out-of-the-box compatible with the sort-by widget, we had to walk away from the option of exposing `compositionID` in the Search Parameters in favor of relying solely on the existing `index` query parameter to store the compositionID

packages/instantsearch.js/src/lib/InstantSearch.ts
e-krebs: ✅

packages/algoliasearch-helper/index.d.ts
```diff
@@ -447,6 +452,12 @@ declare namespace algoliasearchHelper {
      * Targeted index. This parameter is mandatory.
      */
     index?: string;
+    /**
+     * Targeted compositionID.
+     * This parameter is only expected (and mandatory) when targeting the Composition API.
+     * It will be ignored when using the  Search API.
+     */
+    compositionID?: string;
```
Haroenv, line 460:
```suggestion
```

no longer used right?

Haroenv (approved):

Haroenv (approved):
This seems like the right initial approach for composition.

packages/algoliasearch-helper/src/requestBuilder.js
```diff
@@ -123,6 +123,24 @@ var requestBuilder = {
...
   },
 
+  /**
+   * Get all the queries to send to the client, those queries can used directly
+   * with the Algolia client.
+   * @private
+   * @param  {SearchParameters} state The state from which to get the queries
+   * @return {object[]} The queries
+   */
+  _getCompositionQueries: function getQueries(state) {
```
dhayab, line 133: Should this be tested for the combination of outcomes happening in `_getCompositionHitsSearchParams`? This could also be done in a followup PR.

dhayab (approved):
Looks great! I'd suggest adding an example for each flavor once the composition client is released.

e-krebs:
> Looks great! I'd suggest adding an example for each flavor once the composition client is released.

@dhayab that's the plan (at least for react & maybe vanilla js too)

dhayab (approved):

PR #6438, opened 2024-11-15
chore: Add Usage Trend of InstantSearch Family

cwtuan:
I’d like to suggest adding a link to the README that shows the usage trend of the InstantSearch family of packages. This link will direct users to a dynamic chart that visualizes the download trends of related packages over the past three years, helping new users evaluate the popularity and adoption of InstantSearch tools.

As a maintainer of npm-compare.com, I’m also happy to implement any other features that could benefit your project, if you find them useful.

Thanks for considering my suggestion!


Haroenv:
Thanks for your contribution! This could be interesting, but I wonder what people would learn from the graph? That Vue InstantSearch is not used very much in comparison, or something else I'm missing? Thanks

Haroenv:
Thanks again for your suggestion. In the end the team is not convinced this input helps people decide on using InstantSearch or what flavour to use, so we won't be merging it. Thanks though for your input and I'll keep the site in mind for when I want to compare packages!

PR #6440, opened 2024-11-15
refactor(vue): replace instantsearch.js with instantsearch-core

Haroenv:
[FX-3069]

BREAKING CHANGE: if you are using any internal instantsearch.js functions, ensure to update to instantsearch-core to avoid duplication

[FX-3069]: https://algolia.atlassian.net/browse/FX-3069?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/vue-instantsearch/src/util/createInstantSearchComponent.js
```diff
@@ -23,10 +23,19 @@ export const createInstantSearchComponent = (component) =>
...
+              .search();
+          }
         },
         indexName(indexName) {
-          this.instantSearchInstance.helper.setIndex(indexName || '').search();
+          if (this.instantSearchInstance.helper) {
+            this.instantSearchInstance.helper
+              .setIndex(indexName || '')
+              .search();
+          }
```
dhayab, line 38: Shouldn't this also manipulates `mainHelper` until we handle the case in instantsearch-core?

shaejaz (approved):

PR #6442, opened 2024-11-19
fix(rendering): ensure resilience against "null" results

### LINKED ISSUE #6441: getRefinements can receive null and system crashes when it does.
### 🐛 Current behavior

/node_modules/instantsearch.js/es/lib/utils/getRefinements

Function getRefinements...

```javascript
export function getRefinements(results, state) {
  var includesQuery = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : false;
  var refinements = [];
  var _state$facetsRefineme = state.facetsRefinements,
```

results can be null. This function crashes when it is as there is no guard in the function for this possibility. Adding a guard and returning an empty array (as in, no refinements possible) the error goes away.

```javascript
export function getRefinements(results, state) {
  if (results === null) return []
  var includesQuery = arguments.length > 2 && arguments[2] !== undefined ? arguments[2] : false;
  var refinements = [];
  var _state$facetsRefineme = state.facetsRefinements,
```

### 🔍 Steps to reproduce

I have no clue how my code is causing a null object to be passed to this function. I just know that it is happening and a straightforward hack stops the nonsense.

### Live reproduction

No sandbox should be necessary - and I honestly have no clue on how to reproduce the bug.

### 💭 Expected behavior

Throwing an unguarded error is a bug.  The API should either do the above to recover, or throw a detailed error explaining what I can do in my code to prevent a null object from getting passed into this function.

### Package version

instantsearch 4.75.4

### Operating system

OSX Sequoia

### Browser

Edge Latest

### Code of Conduct

Haroenv:
**Summary**

In some cases `results` in `render` can be null:
- no index name passed (https://github.com/algolia/instantsearch/blob/a896774b873ff5d046c1e736680bce8c48661e54/packages/algoliasearch-helper/src/algoliasearch.helper.js#L1503-L1510)
- stalled render (already handled)

In the vast majority of cases we already handled `results` being `null` or `undefined` just in case, but now i've changed the type to ensure we always catch this problem

**Result**

null is handled in all cases for results

This issue doesn't have a reproduction, but i believe it now guards for all possible cases of "no results", so: 
fixes #6441

[CR-7353]

[CR-7353]: https://algolia.atlassian.net/browse/CR-7353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -10,11 +10,11 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "83.50 kB"
+      "maxSize": "84 kB"
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "181.50 kB"
+      "maxSize": "182 kB"
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",
--- a/examples/js/e-commerce-umd/src/widgets/ClearFiltersEmptyResults.ts
+++ b/examples/js/e-commerce-umd/src/widgets/ClearFiltersEmptyResults.ts
@@ -1,8 +1,8 @@
 const { panel, clearRefinements } = window.instantsearch.widgets;
 
 const clearFilters = panel<typeof clearRefinements>({
-  hidden(options) {
-    return options.results.nbHits > 0;
+  hidden({ results }) {
+    return Boolean(results && results.nbHits > 0);
   },
 })(clearRefinements);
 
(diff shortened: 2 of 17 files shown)


shaejaz (approved):
lgtm!
Since this might result in type errors, would this count as a breaking change?

Haroenv:
This will be merged and released after code freeze is over; if you bump into the "results is null" error, you can ignore it as the page still behaves correctly, when results is null no events can have happened yet

PR #6443, opened 2024-11-19
fix(instantsearch.js): prevent authenticated token being set as the `userToken`

shaejaz:
**Summary**
The authenticated user token from insights should no longer be used as the `userToken` for the search API. This PR removes all listeners and usage of `authenticatedUserToken` in the insights middleware.

[FX-3159]

[FX-3159]: https://algolia.atlassian.net/browse/FX-3159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -937,10 +937,10 @@ describe('insights', () => {
             createInsightsMiddleware({ insightsClient })
           );
 
-          expect(getUserToken()).toEqual('abc');
+          expect(getUserToken()).not.toEqual('abc');
```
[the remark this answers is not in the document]
shaejaz: It going to be a generated anon token with a uuid, so hard to test the exact string value otherwise. Is there any other method I can use to assert that?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -242,9 +228,7 @@ export function createInsightsMiddleware<
 
         const setUserTokenToSearch = (
           userToken?: string | number,
-          tokenType?: TokenType,
-          immediate = false,
-          unsetAuthenticatedUserToken = false
+          immediate = false
```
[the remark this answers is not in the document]
shaejaz: Hmm, yeah no longer needed now.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -264,46 +245,14 @@ export function createInsightsMiddleware<
...
-          ) {
-            return;
           }
 
-          // Delay the token application to the next render cycle
-          if (!immediate) {
-            setTimeout(applyToken, 0);
-          } else {
-            applyToken();
-          }
```
francoischalifour, line 287: Wouldn't this be a regression to #5681?

I'm curious why only the authenticated user token required `immediate: false`.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -390,25 +308,7 @@ export function createInsightsMiddleware<
         // This updates userToken which is set explicitly by `aa('setUserToken', userToken)`
         insightsClient(
           'onUserTokenChange',
-          (token) => setUserTokenToSearch(token, 'default', true),
-          {
-            immediate: true,
-          }
```
francoischalifour, line 396: Not sure about this removal, if my previous comment holds true.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: Good call, when I suggested removing `immediate` it was under the assumption we always called it with `true`, but indeed this branch is called with false. I wonder if instead of the `setUserTokenToSearch` having a delay, `setUserTokenToSearch` should be called with a delay inside this callback to make it clearer

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
Haroenv: this part is still present actually, the diff is a little confusing, but `{ immediate: true }` is a parameter to the insights client, while `setUserTokenToSearch(token, token, false)` (or `setUserToken(token)`) is the timeout branch

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
shaejaz: I'm not sure if the PR linked is valid atm since the double queries bug was still happening and the it was solved in another way. I'll test if removing this actually causes double queries to happen again.

But since we're removing auth token handling entirely, would we still need the immediate flag? I don't see it being used with false anywhere else.

Haroenv: false is the default value, so it was called in the `setUserTokenToSearch(token)` callback, no?

shaejaz: I think it was only being called with `true`: https://github.com/algolia/instantsearch/pull/6443/commits/a3a0e9c56874affff2b057e16a4b36e14fbca4cc

To be safe, we could keep this though. Might be used in the future in some way.

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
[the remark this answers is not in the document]
Haroenv:
```suggestion
          expect(getUserToken).toEqual(expect.stringMatching(/^anonymous-/));
```

Haroenv (approved):

Haroenv:
@shaejaz, best not to merge this right now yet as we may be needing a hotfix during code freeze and then I'd rather not have this change in it yet

shaejaz:
@Haroenv makes sense 👍. Should I label this with 'do not merge' ?

francoischalifour (approved):
Thank you!

PR #6445, opened 2024-11-20
fix(disjunctive): prefer values of main query for facet count

Haroenv:
**Summary**

Use the values of the main facet in disjunctive facets

This is relevant in two cases:
- second query is non-exhaustive
- second query is not correct due to facet count after distinct

There's still more edge cases left (first query non exhaustive, but second isn't, but that's unlikely) and those are not worse after this PR.

**Result**

disjunctive facets read the count from the first query over the ones from the second query. This is already the case for hierarchical facets.

### DIFF
--- a/packages/algoliasearch-helper/test/spec/SearchResults/getFacetValues.js
+++ b/packages/algoliasearch-helper/test/spec/SearchResults/getFacetValues.js
@@ -730,7 +730,7 @@ test('getFacetValues(hierarchical) when current state is different to constructo
   expect(facetValues).toEqual(expected);
 });
 
-test.only('getFacetValues(facetName) prefers the "main" facet result', function () {
+test('getFacetValues(facetName) prefers the "main" facet result', function () {
   var data = require('./getFacetValues/disjunctive-non-exhaustive.json');
   var searchParams = new SearchParameters(data.state);
   var result = new SearchResults(searchParams, data.content.results);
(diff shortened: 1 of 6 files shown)


packages/algoliasearch-helper/test/spec/SearchResults/getFacetValues.js
```diff
@@ -729,3 +729,24 @@ test('getFacetValues(hierarchical) when current state is different to constructo
 
   expect(facetValues).toEqual(expected);
 });
+
+test.only('getFacetValues(facetName) prefers the "main" facet result', function () {
```
Haroenv, line 733:
```suggestion
test('getFacetValues(facetName) prefers the "main" facet result', function () {
```

dhayab (commented):
This works great for disjunctive refinements. Looking at the customer issue again, they experience this on a conjunctive refinement, which is not covered by this fix. Do you want to handle this in a subsequent PR or can you add it here?

Haroenv:
How could they have this issue on a conjunctive facet? there's only one query there

dhayab:
Sorry I was wrong, I thought the menu widget was using conjunctive faceting but it instead is based on hierarchical facets. In that case, we do reattribute the root level count but not in this instance?

screenshot-20241121-102734@2x

Haroenv:
@dhayab fixed now for hierarchical too, but only if persistHierarchicalRootCount is true, other cases seemed to break a lot of tests

dhayab:
Nice, thanks!

dhayab (approved):
This works, thank you!

PR #6452, opened 2024-11-28
fix(next-app-router): prevent client-side search when rerendering

dhayab:
**Summary**

This PR updates `<InstantSearchNext>` so it only performs initial search in a server environment.

We previously relied on the availability of the initial results (injected in the document), but they are deleted once hydration is complete, which falsely triggers a client-side search when rerendering the component.

### DIFF
--- a/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
+++ b/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
@@ -47,6 +47,8 @@ export function InstantSearchNext<
   ...instantSearchProps
 }: InstantSearchNextProps<TUiState, TRouteState>) {
   const isMounting = useRef(true);
+  const isServer = typeof window === 'undefined';
+
   useEffect(() => {
     isMounting.current = false;
     return () => {
@@ -77,9 +79,9 @@ This message will only be displayed in development mode.`
     <InstantSearchRSCContext.Provider value={promiseRef}>
       <InstantSearchSSRProvider initialResults={initialResults}>
         <InstantSearch {...instantSearchProps} routing={routing}>
-          {!initialResults && <InitializePromise nonce={nonce} />}
+          {isServer && <InitializePromise nonce={nonce} />}
           {children}
-          {!initialResults && <TriggerSearch />}
+          {isServer && <TriggerSearch />}
         </InstantSearch>
       </InstantSearchSSRProvider>
     </InstantSearchRSCContext.Provider>
--- a/packages/react-instantsearch-nextjs/src/__tests__/InstantSearchNext.test.tsx
+++ b/packages/react-instantsearch-nextjs/src/__tests__/InstantSearchNext.test.tsx
@@ -1,0 +1,37 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import { createSearchClient } from '@instantsearch/mocks';
+import { wait } from '@instantsearch/testutils';
+import { act, render } from '@testing-library/react';
+import React from 'react';
+import { SearchBox } from 'react-instantsearch';
+
+import { InstantSearchNext } from '../InstantSearchNext';
+
+test('it rerenders without triggering a client-side search', async () => {
+  const client = createSearchClient();
+
+  function Component() {
+    return (
+      <InstantSearchNext searchClient={client} indexName="indexName">
+        <SearchBox />
+      </InstantSearchNext>
+    );
+  }
+
+  const { rerender } = render(<Component />);
+
+  await act(async () => {
+    await wait(0);
+  });
+
+  rerender(<Component />);
+
+  await act(async () => {
+    await wait(0);
+  });
+
+  expect(client.search).toHaveBeenCalledTimes(0);
+});


packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
```diff
@@ -55,6 +69,10 @@ export function InstantSearchNext<
     };
   }, []);
 
+  useEffect(() => {
+    window[InstantSearchLastPath] = pathname;
```
Haroenv, line 73: won't this also be set if someone has InstantSearch set up with routing for multiple different path names?

dhayab: Not sure I get it, are you refering to a situation where there's an InstantSearch implementation in multiple paths and the user navigates from one to the other?

In that case, this is okay as search needs to be triggered.

In any case I'm not fond of having to store this data globally, but I haven't found a good way to track route changes since `useRouter` events are not available in app routing. I'm definitely open to ideas around that.

Haroenv: I'm thinking of a use case where with routing someone has for example `/search/:category` set up as the path. If you refine the category, won't it also be true here and force an extra search?

packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
```diff
@@ -47,6 +50,17 @@ export function InstantSearchNext<
...
+  const isServer = typeof window === 'undefined';
+  const pathname = usePathname();
+  const hasRouteChanged =
+    !isServer &&
+    window[InstantSearchLastPath] &&
+    window[InstantSearchLastPath] !== pathname;
+
+  // We only want to trigger a search from a server environment
+  // or if a Next.js route change has happened on the client
+  const shouldTriggerSearch = isServer || hasRouteChanged;
```
aymeric-giraudet, line 62: `InitializePromise` and `TriggerSearch` wouldn't do anything on the client.

You don't need to have them to trigger a search on the client, normal CSR with 2 renders kicks in.

aymeric-giraudet: That means just having `isServer` is enough I think

dhayab: This breaks in a dynamic route setting, where a route change caused by Next.js will unmount/remount InstantSearchNext, and not trigger the search.

Adding the check for path change allows `<TriggerSearch>` to be mounted and executed as you can see in this recording (the generated sandbox is failing to setup sadly).

I don't exactly understand how it kicks in. Probably Next.js route change performs similarly to `router.refresh()` and executes server components?

https://github.com/user-attachments/assets/8ff0e7cf-5528-4190-b979-05c759dc957f

packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
dhayab: This change caters to that use case specifically. You can see it in this comment https://github.com/algolia/instantsearch/pull/6452/files#r1863228512.

packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
aymeric-giraudet: Ah my bad sorry !
Actually I think this solution is nice, we could even choose not to delete `initialResults`and it would avoid that flash of no results. Not sure what the condition would look like though, but it can be done later in another PR.

packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
aymeric-giraudet: It would be nice if we had an InstantSearch singleton for React. It would make sure they only use one instance of InstantSearch and we could probably keep track of URLs in there.

Haroenv (approved):
as pathname doesn't seem to change when triggered by InstantSearch, this should work!

aymeric-giraudet (approved):

oscarmylla:
Hi,

I’ve identified a bug related to this PR:

### Bug Description:
When a user navigates away from a route containing InstantSearch, goes to a route without InstantSearch, and then uses the browser's back button to return, no server-side search is triggered. Additionally, because initialResults are deleted on unmount, no data is displayed. The path is still identified as the InstantSearchLastPath, leading to this issue.

### Proposed Solution:
I suggest modifying the logic to ensure we correctly detect when the route has changed. Here's my suggestion that seem to work for me:

```
const hasRouteChanged = React.useMemo(() => {
    // On server, always return false
    if (isServer) return false;

    // On client, check if initial results are present.
    // If they are, the component has not unmounted and the route hasn't changed.
    const hasInitialResults =
      window[
        InstantSearchInitialResults as unknown as typeof InstantSearchInitialResultsSymbol
      ] !== undefined;

    return !hasInitialResults;
  }, [isServer]);
```

dhayab:
Hi @oscarmylla. Deducing the route change from the initial results availability is a better way to cover all use cases. I'll update the behavior in the next version of the package.

Thanks for your feedback!

PR #6458, opened 2024-12-03
feat: make composition compatible with vue

e-krebs:
**Summary**

this PR leverages what's put in place in #6437 to make it available in vue flavor

[EMERCH-1748](https://algolia.atlassian.net/browse/EMERCH-1748)

[EMERCH-1748]: https://algolia.atlassian.net/browse/EMERCH-1748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/vue-instantsearch/src/components/InstantSearch.js
+++ b/packages/vue-instantsearch/src/components/InstantSearch.js
@@ -116,13 +116,6 @@ export default createInstantSearchComponent({
     };
   },
   render: renderCompat(function (h) {
-    if (!this.indexName && !this.compositionID) {
-      // eslint-disable-next-line no-console
-      console.error(
-        'Missing either "index-name" or "compositionID" property on "ais-instant-search"'
-      );
-    }
-
     return h(
       'div',
       {
--- a/packages/vue-instantsearch/src/components/InstantSearch.js
+++ b/packages/vue-instantsearch/src/components/InstantSearch.js
@@ -22,7 +22,11 @@ export default createInstantSearchComponent({
     },
     indexName: {
       type: String,
-      required: true,
+      required: false,
+    },
+    compositionID: {
+      type: String,
+      required: false,
     },
     routing: {
       default: undefined,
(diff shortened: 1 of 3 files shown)


packages/vue-instantsearch/src/components/InstantSearch.js
```diff
@@ -111,6 +116,12 @@ export default createInstantSearchComponent({
     };
   },
   render: renderCompat(function (h) {
+    if (!this.indexName && !this.compositionID) {
+      console.error(
+        'Missing either "index-name" or "compositionID" property on "ais-instant-search"'
+      );
+    }
+
```
e-krebs, line 124: I couldn't find a way to make either `indexName` or `compositionID` required, so this is the next best thing I could come up with :/

packages/vue-instantsearch/src/components/InstantSearch.js
```diff
@@ -111,6 +116,13 @@ export default createInstantSearchComponent({
     };
   },
   render: renderCompat(function (h) {
+    if (!this.indexName && !this.compositionID) {
+      // eslint-disable-next-line no-console
+      console.error(
+        'Missing either "index-name" or "compositionID" property on "ais-instant-search"'
+      );
+    }
```
e-krebs, line 124: I couldn't find a way to make either `indexName` or `compositionID` required, so this is the next best thing I could come up with :/

Haroenv: could you use the validator function and check in each if it or the other one is present in props? Otherwise I'd leave this error over to InstantSearch tbh

Haroenv (commented):
looks good!

packages/vue-instantsearch/src/components/InstantSearch.js
e-krebs: what do you mean by "the validator function"?

e-krebs: from my understanding, the standard vue validator function only allows to validate values, not properties presence vs. each other

Haroenv: makes sense. In the end I'd leave out this console warning, as no index name on root has a use case (if there's an index widget child). I think people can see that nothing is happening and if they intended to use composition, they would know it needs to be passed.

Haroenv (approved):

PR #6467, opened 2024-12-10
fix(next-app-router): better detection of route change

dhayab:
**Summary**

This is a follow-up to https://github.com/algolia/instantsearch/pull/6452 which provides a better way to handle more use cases, adding one where route changes to a non-InstantSearch implementation, and navigates back to it.

**Result**

Search is performed properly when a user navigates back to an InstantSearch implementation.

### DIFF
--- a/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
+++ b/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
@@ -1,7 +1,6 @@
 import { safelyRunOnBrowser } from 'instantsearch.js/es/lib/utils';
 import { headers } from 'next/headers';
-import { usePathname } from 'next/navigation';
-import React, { useEffect, useRef } from 'react';
+import React, { useEffect, useMemo, useRef } from 'react';
 import {
   InstantSearch,
   InstantSearchRSCContext,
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
```diff
@@ -51,11 +48,17 @@ export function InstantSearchNext<
...
+  const hasRouteChanged = useMemo(() => {
+    // On server, always return false
+    if (isServer) {
+      return false;
+    }
+
+    // On client, route has changed if initialResults have been cleaned up
+    const hasInitialResults = window[InstantSearchInitialResults] !== undefined;
+    return !hasInitialResults;
+  }, [isServer]);
```
aymeric-giraudet, line 61: Why a `useMemo` when `isServer` won't ever change ?

There's more overhead in calling that than just a single-line :
```ts
const hasRouteChanged = !isServer && !window[InstantSearchInitialResults
```

We can still make it multiple lines to make it more readable

dhayab: `useMemo()` prevents the value from being true during a rerender triggered by a wrapping component.

Because of strict mode, the initial results are cleaned during the automatic unmount/remount, which will cause the issue the previous PR aimed to fix.

aymeric-giraudet: Ah yes because `window[InstantSearchInitialResults]` is not in the dependency array

aymeric-giraudet (approved):

PR #6473, opened 2024-12-11
Add Aria labels to hitsPerPage

mutherr:
**Summary**

I was working on improving the accessibility of a project using Instantsearch and noticed that the hits per page widget has no aria label, so I've added it to all three flavors of instantsearch and updated relevant tests

**Result**

Screenshot 2024-12-11 140606

Here's a screenshot of my version with an Aria label present on the hits per page selector.

### DIFF
--- a/packages/instantsearch.js/src/widgets/hits-per-page/hits-per-page.tsx
+++ b/packages/instantsearch.js/src/widgets/hits-per-page/hits-per-page.tsx
@@ -42,6 +42,7 @@ const renderer =
       <div className={cssClasses.root}>
         <Selector
           cssClasses={cssClasses}
+          ariaLabel='Items per page'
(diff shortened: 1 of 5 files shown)


Haroenv:
Thanks! to finalise this we'd want to have this option in translations for React InstantSearch, slots for Vue InstantSearch and templates for InstantSearch.js

mutherr:
Ok. Where are those defined?

mutherr:
Quick follow up regarding the `vue-instantsearch` finalization, as it stands `HitsPerPage` mirrors the other implementations of aria labeling (ex `SortBy.vue`) where the template contains a single slot wrapping the entire select element, rather than having slots for individual elements and their attributes. What to your mind does a finalized version of this look like?

PR #6476, opened 2024-12-17
feat(index): remove getWidgetState

Haroenv:
The old method has long been deprecated and has no usage inside instantsearch outside of running a warning.

[FX-3197]

BREAKING CHANGE: rename getWidgetState to getWidgetUiState in your custom widgets or connectors.

[FX-3197]: https://algolia.atlassian.net/browse/FX-3197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch-core/src/widgets/index-widget.ts
+++ b/packages/instantsearch-core/src/widgets/index-widget.ts
@@ -120,15 +120,11 @@ function getLocalWidgetsUiState(
       return uiState;
     }
 
-    if (!widget.getWidgetUiState && !widget.getWidgetState) {
-      return uiState;
-    }
-
     if (widget.getWidgetUiState) {
       return widget.getWidgetUiState(uiState, widgetStateOptions);
     }
 
-    return widget.getWidgetState!(uiState, widgetStateOptions);
+    return uiState;
   }, initialUiState);
 }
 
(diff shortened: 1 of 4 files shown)


packages/instantsearch-core/src/widgets/index-widget.ts
```diff
@@ -686,10 +682,8 @@ export const index = (widgetParams: IndexWidgetParams): IndexWidget => {
 
       localWidgets.forEach((widget) => {
         warning(
-          // if it has NO getWidgetState or if it has getWidgetUiState, we don't warn
-          // aka we warn if there's _only_ getWidgetState
-          !widget.getWidgetState || Boolean(widget.getWidgetUiState),
-          'The `getWidgetState` method is renamed `getWidgetUiState` and will no longer exist under that name in InstantSearch.js 5.x. Please use `getWidgetUiState` instead.'
+          !(widget as any).getWidgetState,
+          'The `getWidgetState` method is renamed `getWidgetUiState` and no longer exists under that name in InstantSearch.js 5.x. Please use `getWidgetUiState` instead.'
```
dhayab, line 686: Nitpick: the only scenario I see where this would be triggered is for custom widgets created more than 4 years ago (the point from which the removed warning got added). We could go one step further and remove this, wdyt?

dhayab (approved):

packages/instantsearch-core/src/widgets/index-widget.ts
Haroenv: true, I didn't realise the warning was this old. Will remove it too

PR #6477, opened 2024-12-17
feat(routing): replace getLocation with getCurrentURL

Haroenv:
getCurrentURL is the alternative of getLocation in the history router. This is changed as it's possible to create a URL out of a string, but not a location. We also don't use any properties of the location that aren't also present in a URL.

[FX-3193]

BREAKING CHANGE: change from `getLocation` to `getCurrentURL`, and return a URL object.
BREAKING CHANGE: createURL createURL now receives `currentURL` instead of `location`
BREAKING CHANGE: parseURL createURL now receives `currentURL` instead of `location`

[FX-3193]: https://algolia.atlassian.net/browse/FX-3193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/examples/react/e-commerce/routing.ts
+++ b/examples/react/e-commerce/routing.ts
@@ -85,10 +85,10 @@ const router = historyRouter<RouteState>({
       .join(' | ');
   },
 
-  createURL({ qsModule, routeState, location }): string {
-    const { protocol, hostname, port = '', pathname, hash } = location;
+  createURL({ qsModule, routeState, currentURL }): string {
+    const { protocol, hostname, port = '', pathname, hash } = currentURL;
     const portWithPrefix = port === '' ? '' : `:${port}`;
-    const urlParts = location.href.match(/^(.*?)\/search/);
+    const urlParts = currentURL.href.match(/^(.*?)\/search/);
     const baseUrl =
       (urlParts && urlParts[0]) ||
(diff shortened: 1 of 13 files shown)


examples/react/e-commerce/routing.ts
```diff
@@ -152,13 +152,15 @@ const router = historyRouter<RouteState>({
...
-    const pathnameMatches = location.pathname.match(/search\/(.*?)\/?$/);
+  parseURL({ qsModule, currentURL }): RouteState {
+    const pathnameMatches = currentURL.pathname.match(/search\/(.*?)\/?$/);
     const category = getCategoryName(
       (pathnameMatches && pathnameMatches[1]) || ''
     );
 
-    const queryParameters = qsModule.parse(location.search.slice(1));
+    const queryParameters = qsModule.parse(currentURL.search, {
+      ignoreQueryPrefix: true,
```
dhayab, line 162: 😮

dhayab (approved):
So much simpler, thanks!

Haroenv:
I believe this also makes the ecom routing example possible to be simpler but I didn't want to untangle it in this pr yet

PR #6478, opened 2024-12-18
feat(helper): persistHierarchicalRootCount: true

Haroenv:
BREAKING CHANGE: the option persistHierarchicalRootCount no longer exists. The default behaviour is now the same as used to be with persistHierarchicalRootCount: true

[FX-3185]

[FX-3185]: https://algolia.atlassian.net/browse/FX-3185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/algoliasearch-helper/test/spec/hierarchical-facets/sort-by.js
+++ b/packages/algoliasearch-helper/test/spec/hierarchical-facets/sort-by.js
@@ -1,6 +1,6 @@
 'use strict';
 
-test('hierarchical facets: using sortBy', function (done) {
+test('hierarchical facets: using sortBy', async function () {
   var algoliasearch = require('algoliasearch');
   algoliasearch = algoliasearch.algoliasearch || algoliasearch;
 
@@ -30,6 +30,7 @@ test('hierarchical facets: using sortBy', function (done) {
 
   var algoliaResponse = {
     results: [
+      // for hits
       {
         query: 'a',
         index: indexName,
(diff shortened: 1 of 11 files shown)


packages/algoliasearch-helper/test/spec/hierarchical-facets/sort-by.js
```diff
@@ -91,6 +91,7 @@ test('hierarchical facets: using sortBy', function (done) {
           name: 'beers',
           path: 'beers',
           escapedValue: 'beers',
+          // is this not a bit weird??
```
Haroenv, line 94: q

packages/algoliasearch-helper/test/spec/hierarchical-facets/sort-by.js
```diff
@@ -99,7 +100,8 @@ test('hierarchical facets: using sortBy', function (done) {
               name: 'IPA',
               path: 'beers > IPA',
               escapedValue: 'beers > IPA',
-              count: 5,
+              // this too?
```
Haroenv, line 103: q

packages/algoliasearch-helper/test/spec/hierarchical-facets/sort-by.js
```diff
@@ -75,6 +78,19 @@ test('hierarchical facets: using sortBy', function (done) {
           'categories.lvl0': { beers: 5 },
         },
       },
+      // other levels
```
Haroenv, line 81: this test was wrong and missing the intermediary facet response

Haroenv:
by the way this means that https://github.com/algolia/instantsearch/pull/6445#issuecomment-2493209337 is now fixed for hierarchical in all cases, not just when the flag is set (because the default has changed)

dhayab (approved):

PR #6489, opened 2024-12-20
feat: add searchForFacetValues for composition

e-krebs:
**Summary**

adding support for `SearchForFacetValues` when using Composition API
[EMERCH-1745](https://algolia.atlassian.net/browse/EMERCH-1745)

**Result**

there is a small bug on the API side which result in highlighting not working as expecting in the facet values, but that doesn't affect the implementation as the fix should be transparent API-wise

[EMERCH-1745]: https://algolia.atlassian.net/browse/EMERCH-1745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/algoliasearch-helper/src/algoliasearch.helper.js
+++ b/packages/algoliasearch-helper/src/algoliasearch.helper.js
@@ -500,8 +500,8 @@ AlgoliaSearchHelper.prototype.searchForCompositionFacetValues = function (
     searchForFacetValuesRequest: {
       params: {
         query: state.query,
-        maxFacetHits,
-        searchQuery: { query },
+        maxFacetHits: maxFacetHits,
+        searchQuery: { query: query },
       },
     },
   });
--- a/packages/algoliasearch-helper/src/algoliasearch.helper.js
+++ b/packages/algoliasearch-helper/src/algoliasearch.helper.js
@@ -505,10 +505,6 @@ AlgoliaSearchHelper.prototype.searchForCompositionFacetValues = function (
       },
     },
   });
-  // searchForFacetValuesPromise = this.client.searchForFacetValues([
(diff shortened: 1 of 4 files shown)


packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -589,6 +589,11 @@ See documentation: ${createDocumentationLink({
         persistHierarchicalRootCount: this.future.persistHierarchicalRootCount,
       });
 
+    if (this.compositionID) {
+      mainHelper.searchForFacetValues =
+        mainHelper.searchForCompositionFacetValues.bind(mainHelper);
+    }
```
e-krebs, line 595: I went with this approach because it allows things to be completely transparent for `connectRefinementList`, but I'm open to suggestions

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -461,6 +461,85 @@ AlgoliaSearchHelper.prototype.searchForFacetValues = function (
...
+        query: state.query,
+        maxFacetHits: maxFacetHits,
+        searchQuery: { query: query },
+      },
+    },
+  });
+  // searchForFacetValuesPromise = this.client.searchForFacetValues([
+  //   { indexName: state.index, params: algoliaQuery },
+  // ]);
+  // algoliasearch < 3.27.1
```
Haroenv, line 511:
```suggestion
```

Haroenv (approved):

Haroenv:
Once we're aiming for GA with composition, I'd like to see a plan how the search and sffv methods could be the same between search and composition.

PR #6497, opened 2024-12-30
feat(connectors): cleanup life cycle without dispose

Haroenv:
In `dispose`, we no longer have any logic other than clearing listeners and removing rendering. This causes a significant simplification in all connectors.

One side-effect of this is that `configure` no longer can be collaborating in ui state. If we'd want that, there's no way to distinguish between the ui state set for one configure or the one set for another. Therefore the configure widget is completely in charge of its parameters. If you want to synchronise configure with the URL, do that separately of the UI State.

Small side-effect of this PR is the deduplication of the test/mock createWidget files

BREAKING CHANGE: widgets no longer explicitly clean in dispose (handled in index widget automatically)
BREAKING CHANGE: configure no longer shows in ui state or routing.

[FX-3209]

[FX-3209]: https://algolia.atlassian.net/browse/FX-3209?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhayab (approved):
Awesome. I tested empirically and e2e passed on the branch (except for js-umd but the error is unrelated).

Haroenv:
Thanks, the e2e test doesn't really deal with adding and removing widgets though, so it makes sense it has no impact

shaejaz (approved):

PR #6499, opened 2024-12-30
feat(routing): change default of cleanUrlOnDispose to false

Haroenv:
In the vast majority of cases you're either never disposing InstantSearch, or the URL already gets cleared some other way (eg. through SPA navigation). This is why the new default for cleanUrlOnDispose is `false`.

You'd still want to set this option to `true` if you're rendering instantsearch in a modal that should have the url cleared when that modal is closed.

BREAKING CHANGE: default of cleanUrlOnDispose is false

[FX-3213]

[FX-3213]: https://algolia.atlassian.net/browse/FX-3213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch-core/src/routing/historyRouter.ts
+++ b/packages/instantsearch-core/src/routing/historyRouter.ts
@@ -1,10 +1,6 @@
 import qs from 'qs';
 
-import {
-  createDocumentationLink,
-  safelyRunOnBrowser,
-  warning,
-} from '../lib/public';
+import { safelyRunOnBrowser, warning } from '../lib/public';
 
 import type { Router, UiState } from '../types';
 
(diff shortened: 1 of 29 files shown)


packages/instantsearch-core/src/routing/historyRouter.ts
```diff
@@ -34,7 +30,6 @@ export type BrowserHistoryArgs<TRouteState> = {
    * remove active refinements from the URL.
    * @default true
```
dhayab, line 31:
```suggestion
   * @default false
```

dhayab (approved):

PR #6503, opened 2024-12-31
refactor(routing): simplify historyRouter and tests

Haroenv:
**Summary**

simplify the history router to no longer have a class layer, and rework the tests to have less duplicates, more structured files.


dhayab (approved):
Looks good! I discarded the false-positive security issue from Codacy.

PR #6504, opened 2024-12-31
fix(build): prevent destructure with reuse of same symbol.

Haroenv:
This is the cause of the UMD e2e failure in `next`.

Basically the flow is:

1. source code that destructures something in arguments:

```js
function fn({ one, two }) {}
```

2. esm output of source code

```js
function fn(_ref) {
  let {
    one,
    two,
  } = _ref
```

3. umd production output (using uglify)

```js
function fn(e) {
  var { one: e, two: t } = e
```

note the reused variable name

4. output after import through parcel of the prod umd import

```js
function fn(e) {
  var e = e.one,
    t = e.two;
```

This fixes it by changing step 3 by using terser to:

```js
function fn(e) {
  let { one: t, two: n } = e;
}
```

note that it doesn't reuse the `e` variable and thus no longer bugs and has this output in parcel:

```js
function fn(e) {
  var t = e.one,
    n = e.two;
}
```

This issue started showing up once we changed the browser output of the js umd build to keep destructuring, which had this "rename-bug". Updating to terser is the right solution for the future as it's actually kept updated unlike uglify.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,7 @@
     "rollup-plugin-node-globals": "1.4.0",
     "rollup-plugin-node-resolve": "5.2.0",
     "rollup-plugin-replace": "2.2.0",
-    "rollup-plugin-uglify": "6.0.4",
+    "@rollup/plugin-terser": "0.4.4",
     "shelljs": "0.8.5",
     "shipjs": "0.26.0",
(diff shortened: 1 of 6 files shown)


dhayab (approved):
🤘

PR #6506, opened 2025-01-02
fix: query & searchQuery where mixed when using searchForCompositionFacetValues

e-krebs:
**Summary**

Fixing implementation error made in #6489 where `query` & `searchQuery` where mixed when using `searchForCompositionFacetValues`.

I'm also taking this opportunity to pass full search parameters in `searchQuery`.

**Result**

Searching for Facet Values when using Composition is now working as expected including highlighting.


aymeric-giraudet (approved):
👍

PR #6520, opened 2025-01-09
feat: support for React 19

aymeric-giraudet:
**Summary**

[FX-3202](https://algolia.atlassian.net/browse/FX-3202)

**Result**

Mostly type-related issues and had to remove `@testing-library/react-hooks` in favor of just `@testing-library/react`

fixes part of https://github.com/algolia/instantsearch/issues/6496

[FX-3202]: https://algolia.atlassian.net/browse/FX-3202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/instantsearch-ui-components/src/types/Renderer.ts
```diff
@@ -1,11 +1,13 @@
 /* eslint-disable @typescript-eslint/no-namespace */
 /* eslint-disable @typescript-eslint/no-empty-interface */
 
+import type { JSX as ReactJSX } from 'react';
```
Haroenv, line 4: does this not break for preact then?

Haroenv (approved):

packages/react-instantsearch-core/src/hooks/__tests__/useInstantSearch.test.tsx
```diff
@@ -24,8 +23,8 @@ describe('useInstantSearch', () => {
     test('it errors when not nested in InstantSearch', () => {
       const { result } = renderHook(() => useInstantSearch());
```
Haroenv, line 24: seems like this error now throws globally, can use toThrowError instead of result.current.error I guess

packages/instantsearch-ui-components/src/types/Renderer.ts
aymeric-giraudet: It shouldn't, but I need to see if TypeScript works with just a simple IS.js project in typescript

I don't know why it worked before, I think it used to load the JSX namespace from `@types/react` but now it doesn't and has a lot of errors saying for example `div` is not in `JSX.IntrinsicElements`

aymeric-giraudet: That's because the `JSX` namespace used to be wrapped inside a `declare global` but it's not anymore :/

Haroenv (dismissed):

aymeric-giraudet:
@Haroenv I tested it on a Vite app with React 19 and it does solve this problem : https://github.com/algolia/instantsearch/issues/6426#issuecomment-2573079176

Now instead it uses `React.JSX.Element` :)

Haroenv (approved):

PR #6523, opened 2025-01-10
feat(recommend): add support for event collection in recommend

marialungu:
**Summary**

Add support for event collection on recommend widgets in order to enable Analytics.

1- Provide `sendEvent` method to all Recommend widgets 
2- support automatic event collection on Recommend widgets + carousel 

**Result**

Collect click/conversion events on Recommend


packages/instantsearch.js/src/templates/carousel/carousel.tsx
```diff
@@ -66,12 +68,14 @@ export function carousel<TObject extends Record<string, unknown>>({
     items,
     templates: widgetTemplates,
     cssClasses: widgetCssClasses = {},
+    sendEvent = () => {},
```
raed667, line 71: A hack to make TypeScript happy

packages/instantsearch.js/src/templates/carousel/carousel.tsx
```diff
@@ -56,6 +56,8 @@ type CarouselTemplateProps<TObject extends Record<string, unknown>> = Pick<
     item?: CarouselUiProps<TObject>['itemComponent'];
   };
   cssClasses?: Partial<CarouselUiProps<TObject>['classNames']>;
+} & {
+  sendEvent?: CarouselUiProps<TObject>['sendEvent'];
```
dhayab, line 60: You can instead pick `'items' | 'sendEvent'` from `CarouselUiProps` on L53.

```suggestion
} & {
  sendEvent?: CarouselUiProps<TObject>['sendEvent'];
```

packages/instantsearch.js/src/templates/carousel/carousel.tsx
dhayab: With `sendEvent` typed from the UI component you should be able to remove the default value, I think?

```suggestion
    sendEvent,
```

dhayab (approved):
Great work 👏. I have a small suggestion otherwise it's good to go!

packages/instantsearch.js/src/templates/carousel/carousel.tsx
marialungu: We have tried this approach at first, but TS was throwing errors, so we found the workaround to pick `sendEvent` as optional 🤔

dhayab: Ah, well 😅, it's fine. I'll resolve these and you can merge whenever you're ready.

dhayab (approved):

PR #6524, opened 2025-01-13
feat: support for Next.js 15

shaejaz:
**Summary**

**Result**


packages/react-instantsearch-nextjs/package.json
```diff
@@ -50,11 +50,11 @@
   },
   "devDependencies": {
     "instantsearch.js": "4.75.7",
-    "next": "13.5.1",
+    "next": "15.1.4",
     "react-instantsearch-core": "7.13.10"
   },
   "peerDependencies": {
-    "next": ">= 13.4 < 15",
+    "next": ">= 13.4 < 16",
```
aymeric-giraudet, line 57: Does it still work with next 14 ? What happens when `use` receives a non-promise value ?

aymeric-giraudet: If it doesn't work we could check whether `headers()` returns a promise or not

shaejaz: Ah yes, you're right, `use` would error out when a non Promise is passed in. I'll add a check for it 👍

packages/react-instantsearch-nextjs/src/useNextHeaders.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+function isPromise(obj: any) {
+  return (
+    obj &&
+    (typeof obj === 'object' || typeof obj === 'function') &&
+    typeof obj.then === 'function'
+  );
+}
+
+type Headers = Awaited<ReturnType<typeof headers>> | undefined;
```
dhayab, line 12: Nitpick: should we remove the union with `undefined`, throwing instead if `useNextHeaders()` is called from a non-server environment?

We are confident we are calling it properly and it will remove the need for optional chaining in its usage.

dhayab (approved):
Small suggestion to improve DX, otherwise looks good to me!

packages/react-instantsearch-nextjs/src/useNextHeaders.tsx
aymeric-giraudet: We don't want it to throw as it has to be called on the frontend for hydration no ?

Haroenv (approved):

aymeric-giraudet:
I think you can remove "Breaking change" from the description, there's none

PR #6526, opened 2025-01-13
fix(refinement list): prevent error when value has a quote

Haroenv:
**Summary**

If the `value` has a `"`, the query selector is invalid and throws an error. This PR fixes that case.

Code introduced in https://github.com/algolia/instantsearch/pull/6392 originally. Preact issue still open.

**Result**

we escape (only `"`, other characters have no impact) the value in the query selector, no error is thrown and the correct value is focused.

[CR-7674]

[CR-7674]: https://algolia.atlassian.net/browse/CR-7674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhayab (approved):
Looks good, thanks for the fix! Other than this, what's happening with the yarn.lock?

Haroenv:
I had to change `babel/traverse` latest version for https://github.com/babel/babel/issues/16689

PR #6527, opened 2025-01-13
chore(vue example): remove price range

Haroenv:
The component `vue-slider-component` throws an error with the new Vite build. I can't seem to find an underlying reason, so for now let's disable the component so release isn't blocked.

### DIFF
--- a/examples/vue/e-commerce/src/App.vue
+++ b/examples/vue/e-commerce/src/App.vue
@@ -116,46 +116,6 @@
               </ais-panel>
 
               <ais-panel>
-                <template #header> Price </template>
-
-                <template #default>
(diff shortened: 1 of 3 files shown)


shaejaz (approved):

dhayab:
This will probably break the e2e tests instead as they expect to interact with the price slider 🤔.

dhayab (commented):
Have you run the e2e tests with this change? While the price range suite is disabled, I'm wondering if it will affect the tests on routing, as we're removing the widget.

Haroenv:
https://github.com/NightCatSama/vue-slider-component/issues/663 is the underlying reason by the way

dhayab (approved):

Issue #6479, opened 2024-12-18
Duplicate Results in scopedResults with Routing Enabled in NextJS


AmeerSapadmi:
### 🐛 Current behavior

When routing is enabled, navigating from one page to another page that displays search results (where a search parameter maps to the query used by Algolia) leads to duplicated entries in the scopedResults array returned by the useInstantSearch hook.

This issue occurs in a setup where we are using multi-index search with `InstantSearchNext`.

### 🔍 Steps to reproduce

**Steps to Reproduce:**  
1. Open the homepage and ensure the browser’s developer tools are open with the console tab selected.  
2. Click the button labeled "Click"  
3. The user is navigated to the search results page, where the search parameter is set to `q=london`.  
4. Observe the `scopedResults` array in the console:  
   - Initially, it logs a single array containing the default index.  
   - Shortly after, it updates to a much larger array, which includes duplicate entries for the indexes on the page.  

### Live reproduction

https://codesandbox.io/p/devbox/beautiful-maria-g85swl

### 💭 Expected behavior

The scopedResults do not contain duplicates.

### Package version

next 15.0.4, react-instantsearch 7.13.9, react-instantsearch-nextjs 0.3.20, algoliasearch 5.17.1

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Haroenv:
There's multiple things going on here:

1. multiple results

This is because the main index has the same index name as the explicitly rendered index. This seems like a bug in useInstantSearch where it reads the results for all indices without deduplicating by indexId.

2. not all results showing up initially.

This will be related to timing where useInstantSearch is called _before_ it renders the child indices, and thus they aren't available in scoped results. In theory they should be available as the search happens backend and all results are available for the other widgets, so this likely is also a bug in useInstantSearch. You can work around this by rendering your scopedResults using code _after_ the indices are mounted (in a child component of the Search in your example, underneath the Index components.

I'd like to look into if this pattern can be made possible in InstantSearch and the duplicate results can be fixed.

What use case are you trying to create with scopedResults? Maybe you could use `useConnector(connectAutocomplete)` instead and it normally will be correct.

AmeerSapadmi:
We are using `scopedResults` to calculate the total number of hits returned across all indices. However, due to the duplicate results, the total hit count we calculate is incorrect.  

If this can be achieved using `useConnector`, could you kindly provide an example demonstrating how to implement it?

Haroenv:
[...]

PR #6534, opened 2025-01-17
fix(next-app-router): do not use SSR context for CSR

aymeric-giraudet:
**Summary**

Fix looks simple but it took time as it was really difficult to debug.

Basically, when navigating with category pages, `InstantSearchNext` remounts and recreates a whole new InstantSearch instance (this is a root problem we may want to deal with later by ensuring we always use a single instance).
It does so too with back and forward buttons, **BUT**, this particular piece of code behaves differently (was really hard to find) : 
https://github.com/algolia/instantsearch/blob/772189605029e8bbb61413737824c096f16269da/packages/instantsearch.js/src/lib/InstantSearch.ts#L698-L710

`defer` is really unreliable on Next.js and Remix IIRC, here when we navigate it did properly remove the `noop`, but with back and forward buttons it didn't which is why we had the initial issue where it would display empty results.

The fix that was made for that worked but it caused issue #6479, as now CSR ran the hydration logic which is to add widgets during render, causing a lot of duplicate `addWidgets` in CSR.

**Result**

Now if we're doing CSR (on the browser, no hydration), we just remove the server contexts to make sure it behaves the same way as regular `react-instantsearch` would.

### DIFF
--- a/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
+++ b/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
@@ -1,5 +1,5 @@
 import { safelyRunOnBrowser } from 'instantsearch.js/es/lib/utils';
-import React, { useEffect, useMemo, useRef } from 'react';
+import React, { useEffect, useRef } from 'react';
 import {
   InstantSearch,
   InstantSearchRSCContext,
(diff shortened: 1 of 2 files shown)


Haroenv (approved):
thanks!

dhayab (approved):

PR #6540, opened 2025-01-22
fix(recommend): use Hit type instead of AlgoliaHit

raed667:
**Summary**

Follow-up on #6533

When using `sendEvent` with Recommend in InstantSearch.js we encounter a typing issue

```
Type 'AlgoliaHit<Record<string, any>>' is not assignable to type 'Hit'.
  Property '__position' is missing in type 'AlgoliaHit<Record<string, any>>' but required in type '{ __position: number; __queryID?: string | undefined; }'
```

codesandbox to demonstrate the type issue https://codesandbox.io/p/sandbox/nzkfh7

This is because `AlgoliaHit` is only meant for hits that don't first go through `addAbsolutePosition` and `addQueryId`. This means that the type we actually mean to use is Hit in most cases.

**Result**

The issue was not caught by `yarn type-check` initially, I'm not sure how we could have done better to automatically catch it.

### DIFF
--- a/packages/react-instantsearch/src/widgets/LookingSimilar.tsx
+++ b/packages/react-instantsearch/src/widgets/LookingSimilar.tsx
@@ -6,11 +6,11 @@ import type {
   LookingSimilarProps as LookingSimilarPropsUiComponentProps,
   Pragma,
 } from 'instantsearch-ui-components';
-import type { AlgoliaHit, BaseHit } from 'instantsearch.js';
+import type { Hit, BaseHit } from 'instantsearch.js';
 import type { UseLookingSimilarProps } from 'react-instantsearch-core';
 
 type UiProps<THit extends BaseHit> = Pick<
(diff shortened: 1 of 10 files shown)


packages/react-instantsearch/src/widgets/LookingSimilar.tsx
```diff
@@ -6,7 +6,7 @@ import type {
   LookingSimilarProps as LookingSimilarPropsUiComponentProps,
   Pragma,
 } from 'instantsearch-ui-components';
-import type { AlgoliaHit, BaseHit, Hit } from 'instantsearch.js';
+import type { AlgoliaHit, BaseHit } from 'instantsearch.js';
```
Haroenv, line 9: now that it has Hit in the connector, the four React InstantSearch widgets should use exclusively Hit I believe.

Haroenv (approved):

PR #6542, opened 2025-01-22
fix(next-app-router): hydration error breaks routing

### LINKED ISSUE #6541: Url params not updating when using react-instantsearch-nextjs (after upgrade to 0.4.2)
### 🐛 Current behavior

When using ReactInstantsearchNext@0.4.2 with routing, url params are updated once, only after the first refinement is applied (refinement checked or query changed). 

After that url is never updated again.

### 🔍 Steps to reproduce

1. open the sandbox
2. Toggle a brand from the refinements list 
3. See that url is updated
4. Toggle another one 
5. See that url did not change (but hits did properly)

### Live reproduction

https://codesandbox.io/p/devbox/example-react-instantsearch-next-app-dir-example-forked-g7v9wp

### 💭 Expected behavior

Url should be updated anytime a refinement changes.

Reverting to react-instantsearch-next@0.4.1 fixes the issue. Probably related to https://github.com/algolia/instantsearch/pull/6534

### Package version

react-instantsearch-next@0.4.2 algoliasearch@5.19.0 react-instantsearch@7.15.1

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

aymeric-giraudet:
**Summary**

Fixes #6541

**Result**

Kind of hacky, but hydration errors seem to trigger effect cleanup functions while not actually remounting the components ?

Note we wouldn't have this problem if we had an InstantSearch instance outside of React lifecycle, which should probably be the way to go for the stable release.

### DIFF
--- a/packages/react-instantsearch-nextjs/src/useInstantSearchRouting.ts
+++ b/packages/react-instantsearch-nextjs/src/useInstantSearchRouting.ts
@@ -22,11 +22,20 @@ export function useInstantSearchRouting<
   const routingRef =
     useRef<InstantSearchProps<TUiState, TRouteState>['routing']>(null);
   const onUpdateRef = useRef<() => void>(null);
+  const isUnmounting = useRef(false);
+
   useEffect(() => {
     if (onUpdateRef.current) {
       onUpdateRef.current();
     }
   }, [pathname, searchParams]);
+
+  useEffect(() => {
+    isUnmounting.current = false;
+    return () => {
+      isUnmounting.current = true;
+    };
+  });
 
   const headers = useNextHeaders();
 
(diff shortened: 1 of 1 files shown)


Haroenv (approved):
no test possible for this case? maybe that we don't try to write history if the component unmounts, or does it just error at that point before this PR?

aymeric-giraudet:
@Haroenv it would be better with e2e tests as we could properly replicate the hydration issue.

I can add tests but this really is a quickfix intended to be removed once we ensure we always have a single instantsearch instance.

Haroenv:
all good @aymeric-giraudet !

dhayab (approved):

PR #6543, opened 2025-01-22
playground: filters matcher

Haroenv:
playground branch for matching filters


packages/instantsearch-core/match-filters.ts
```diff
@@ -0,0 +1,118 @@
...
+      : matchCondition(target, condition)
+  );
+}
+
+function matchFilters(
+  target: AlgoliaRecord,
+  filters: Conditions['filters']
+): boolean {
+  if (!filters) return true;
+  // TODO: implement the filter matching logic
```
Haroenv, line 48: todo

PR #6544, opened 2025-01-23
feat: expose raw request content under _rawContent

e-krebs:
**Summary**

In order to access some composition-related data (when `getRankingInfo` is requested), we need to access the root content of the API response, not just `results` (cf. [last example on this doc](https://algolia.atlassian.net/wiki/spaces/CMP/pages/5556273184/DD+-+Composition+API+updates)).

**Result**

exposing the root content of the API response (excluding `results` that is already exposed) through a `_rawContent` non-documented & non-typed property on search results.

it can therefor be accessed like this:
```ts
const { results: { _rawContent } } = useInstantSearch();
```

### DIFF
--- a/packages/instantsearch.js/src/widgets/index/index.ts
+++ b/packages/instantsearch.js/src/widgets/index/index.ts
@@ -652,12 +652,6 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
           new algoliasearchHelper.SearchParameters(indexInitialResults.state),
           indexInitialResults.results
         );
-        if (
-          '_rawContent' in indexInitialResults &&
-          indexInitialResults._rawContent
-        ) {
-          results._rawResults = indexInitialResults?._rawContent as any;
-        }
 
         derivedHelper.lastResults = results;
         helper.lastResults = results;
--- a/packages/instantsearch.js/src/widgets/index/index.ts
+++ b/packages/instantsearch.js/src/widgets/index/index.ts
@@ -652,6 +652,12 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
           new algoliasearchHelper.SearchParameters(indexInitialResults.state),
           indexInitialResults.results
         );
+        if (
+          '_rawContent' in indexInitialResults &&
+          indexInitialResults._rawContent
+        ) {
+          results._rawResults = indexInitialResults?._rawContent as any;
+        }
 
         derivedHelper.lastResults = results;
         helper.lastResults = results;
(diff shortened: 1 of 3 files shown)


Haroenv:
looks like it's "lucky" that we only override _state, not recreate a whole results object in places like https://github.com/algolia/instantsearch/blob/6f2dc2f8586abb11b92cbfe6030c8ada70563695/packages/instantsearch.js/src/widgets/index/index.ts#L303

Maybe in server side rendering it could not be applied? https://github.com/algolia/instantsearch/blob/0bb5fcbc95d23293361c928e43191b994125c46c/packages/instantsearch-core/src/widgets/index-widget.ts#L624 and https://github.com/algolia/instantsearch/blob/0bb5fcbc95d23293361c928e43191b994125c46c/packages/instantsearch-core/src/widgets/index-widget.ts#L577-L578

e-krebs:
@Haroenv the `instantsearch-core` package you're referencing above doesn't appear to exist on master 🤔

Haroenv:
sorry was on the next branch, but this code exists in master too in InstantSearch.js

e-krebs:
> sorry was on the next branch, but this code exists in master too in InstantSearch.js

I did update my implementation 🙂

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -652,6 +652,12 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
           new algoliasearchHelper.SearchParameters(indexInitialResults.state),
           indexInitialResults.results
         );
+        if (
+          '_rawContent' in indexInitialResults &&
+          indexInitialResults._rawContent
+        ) {
+          results._rawResults = indexInitialResults?._rawContent as any;
+        }
```
Haroenv, line 660: without a change in the initial results this will always be undefined

Haroenv: if you don't need it for the dashboard I'd rather leave it without change in the index widget.

The code also seems wrong, why are you overriding rawResults with rawContent?

e-krebs: > why are you overriding rawResults with rawContent

this is definitely a mistake

> if you don't need it for the dashboard I'd rather leave it without change in the index widget

works for me 🙂

e-krebs:
@Haroenv should I account for `_rawContent` (and other data in general) here?
https://github.com/algolia/instantsearch/blob/e598230f2c73cc01cce091fb67b1435492900c9d/packages/instantsearch.js/src/widgets/index/index.ts#L693-L705

it looks like `state` is intentionally excluded, but I could update this code so that it handle any other data

Haroenv:
No, that part is just copying the results so already works without changes.

Haroenv (approved):
we know this isn't permeated through the whole life cycle, but because of the limited use case, let's land it like this first. Thanks!

PR #6549, opened 2025-01-28
fix: rawContent was not correctly cloned (hence undefined)

e-krebs:
**Summary**

Fixing an error made in #6544, where using `Object.create(content)` resulted in `rawContent` always being an empty object and the final `_rawContent` property always being `undefined`.
Now using a custom method to smartly clone `content` without the `results` property, which behaves correctly.

This approach avoids the need to delete the `results` property as it "just" creates a new object without it in the first place.
This result in a more performant approach than using the existing `merge` util + deleting the `results` property afterward.

**Result**

`_rawContent` as retrieved in the below example is now not always `undefined` anymore.

```ts
const { results: { _rawContent } } = useInstantSearch();
```

### DIFF
--- a/packages/algoliasearch-helper/src/algoliasearch.helper.js
+++ b/packages/algoliasearch-helper/src/algoliasearch.helper.js
@@ -1844,7 +1844,7 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function (
   if (this._currentNbQueries === 0) this.emit('searchQueueEmpty');
 
   var results = content.results.slice();
-  var rawContent = Object.create(content);
+  var rawContent = Object.assign({}, content);
   delete rawContent.results;
   if (Object.keys(rawContent).length <= 0) {
     rawContent = undefined;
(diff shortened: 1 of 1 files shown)


Haroenv (approved):

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1869,12 +1872,11 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function (
       specificResults,
       self._searchResultsOptions
     );
-    helper.lastResults._rawContent = rawContent;
+    if (rawContent !== undefined) helper.lastResults._rawContent = rawContent;
 
     helper.emit('result', {
       results: helper.lastResults,
       state: state,
-      _rawContent: rawContent,
```
e-krebs, line 1877: this was not used, so removing it

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1869,12 +1872,11 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function (
       specificResults,
       self._searchResultsOptions
     );
-    helper.lastResults._rawContent = rawContent;
+    if (rawContent !== undefined) helper.lastResults._rawContent = rawContent;
```
e-krebs, line 1875: this avoid adding `_rawContent: undefined` when there is no raw content

PR #6556, opened 2025-02-10
feat(recommend): add Trending-Facets model

raed667:
**Summary**

Add Recommend model `trending-facets`

**TODO** 

- [ ] Figure out what to do with `sendEvent` (Recommend analytics does not support `trending-facets`)
- [ ] Implement tests


raed667:
@algolia/frontend-experiences-web hi team, this is not urgent, but can someone help figure out go the fix the V4 types 🙏 ?

packages/instantsearch.js/src/connectors/trending-facets/types.ts
```diff
@@ -0,0 +1,6 @@
+import type { TrendingFacetHit as Base } from 'algoliasearch/lite';
```
Haroenv, line 1: you'll want to import this in the alogliasearch.d.ts file in the helper and correctly ts-ignore, and pass the relevant type of v3 (any?), v4 and v5

packages/instantsearch-ui-components/src/components/TrendingFacets.tsx
```diff
@@ -19,7 +19,7 @@ import type {
   Renderer,
   SendEventForHits,
 } from '../types';
-import type { TrendingFacetHit as Base } from 'algoliasearch/lite';
+import type { TrendingFacetHit as Base } from 'instantsearch.js';
```
Haroenv, line 22: you should import from ../types

packages/instantsearch-ui-components/src/components/TrendingFacets.tsx
```diff
@@ -0,0 +1,162 @@
...
+  RecommendClassNames,
+  RecommendInnerComponentProps,
+  RecommendItemComponentProps,
+  RecommendStatus,
+  RecommendTranslations,
+  RecordWithObjectID,
+  Renderer,
+  SendEventForHits,
+} from '../types';
+import type { TrendingFacetHit } from 'algoliasearch';
```
Haroenv, line 22: this should also be in the helper or just retyped `{ objectID: string, facetValue: string}`?

Haroenv:
Originally we didn't implement trending facets as it's not clear how we want to incorporate the filtering / links inside it, as it could be many different attributes, refinement types.

However, it makes sense to be able to fully deprecate the old recommend widgets, and so we'll need trending facets too.

What we'll need to be able to later create a filtering directly from this widget I think is to make the itemComponent required here, so changing the default isn't a breaking change.

raed667:
Closing in favor of #6957

PR #6570, opened 2025-02-18
feat(helper): remove getRefinements, hasRefinements

Haroenv:
These functions are an aggregate of multiple different functions that depend on the type of refinement you're using. It's better to use the getter for the right type of refinement.

These functions are also not used in InstantSearch, as current refinements and clear refinements "re-implement" most of its logic to match the refinement to the right widget.

BREAKING CHANGE: remove helper.hasRefinements (use helper.state.isXXXRefined instead)
BREAKING CHANGE: remove helper.getRefinements (read helper.state.XXXRefinement instead)
BREAKING CHANGE: remove helper.state.getRefinements (read helper.state.XXXRefinement instead)

bundle size impact: -0.5kb min gz on the helper


examples/js/e-commerce-umd/src/widgets/Products.ts
```diff
@@ -51,8 +51,9 @@ export const products = hits<Hit>({
         </article>
       `;
     },
-    empty(results, { html }) {
-      const hasRefinements = results.getRefinements().length > 0;
+    empty(_, { html }) {
+      const hasRefinements =
+        window.search.renderState['instant_search'].clearRefinements?.canRefine;
```
Haroenv, line 56: normally you can directly read the renderState, but in the layout of this example, the widgets are in standalone files, and we can't import `search` or it would be circular.

Could be interesting in the future if templates got access to InstantSearch instance or parent.

shaejaz (approved):

PR #6573, opened 2025-02-19
feat(helper): remove clearRefinements

Haroenv:
More specified options exist, so they should be used instead

BREAKING CHANGE: helper.clearRefinements and helper.state.clearRefinements are removed. Use `setState(new SearchParameters())` or individual remove methods instead


Haroenv:
Closing this PR as it seems to have unexpected changes and the impact is very small to removal of this method

PR #6578, opened 2025-02-26
chore(examples): update react ssr webpack config

dhayab:
**Summary**

This PR updates the webpack config for the **examples/react/ssr** to fix an issue between babel and algoliasearch's transitive dependencies.

It also adds a node option to fix an issue during production build generation.

### DIFF
--- a/examples/react/ssr/webpack.config.js
+++ b/examples/react/ssr/webpack.config.js
@@ -1,64 +1,62 @@
 const path = require('path');
 
+const webpack = require('webpack');
 const nodeExternals = require('webpack-node-externals');
 
-module.exports = [
-  {
+module.exports = (_, { mode }) => {
+  const common = {
     mode: 'development',
-    entry: ['@babel/polyfill', './src/server.js'],
-    output: {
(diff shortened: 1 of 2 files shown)


examples/react/ssr/webpack.config.js
```diff
@@ -1,64 +1,62 @@
 const path = require('path');
 
+const webpack = require('webpack');
```
Haroenv, line 3: would be nice some other time if we find what the new plug-and-play pure server side rendering solution is? Vite server side rendering plugin maybe?

Haroenv (approved):
thanks

examples/react/ssr/webpack.config.js
dhayab: Yes I agree `vite` could work well for this use case.

aymeric-giraudet: Or just get rid of this example tbh, we shouldn't have to maintain something like this when there are metaframeworks like Next and Remix

PR #6579, opened 2025-02-26
feat: use vanilla css reset

thomasbritton:
Would like to of utilised css nesting but couldn't get Parcel configured correctly.

Use Vanilla CSS for the reset instead of Scss

Move on from Scss and use native CSS

### DIFF
--- a/examples/react/css/src/App.tsx
+++ b/examples/react/css/src/App.tsx
@@ -106,11 +106,11 @@ export function App() {
 }
 
 function HitComponent({ hit }: { hit: Hit }) {
-  return Item({ hit });
+  return <Item hit={hit} />;
 }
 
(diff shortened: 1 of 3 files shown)


FabienMotte:
Should we move this line in the `reset.css`?
https://github.com/algolia/instantsearch/blob/b69fd633f0739d9fcbe47293691f19c37c966b95/packages/instantsearch.css/src/themes/algolia.scss#L15

What about this rule?
https://github.com/algolia/instantsearch/blob/b69fd633f0739d9fcbe47293691f19c37c966b95/packages/instantsearch.css/src/themes/algolia.scss#L18-L20

FabienMotte (approved):
Looks good to me, less opinionated 👍 
We can always add rules later on if needed

PR #6584, opened 2025-02-27
poc: add `HierarchicalMenu` styles

FabienMotte:
**Summary**

This PR adds the `HierarchicalMenu` styles.


examples/react/css/src/common.css
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/* Focus outline */
+a[class^='ais-']:focus-visible,
+button[class^='ais-']:focus-visible,
+input[class^='ais-']:focus-visible {
+  outline: 2px solid
+    rgba(var(--ais-primary-color-rgb), var(--ais-primary-color-alpha));
+  outline-offset: 2px;
+}
```
FabienMotte, line 15: I put it there for now, inspired by what we do in Satellite, we have a global CSS class `focusable` (`fosauble-visible`, etc...) to keep our focus styles consistent across components.

examples/react/css/src/common.css
```diff
@@ -0,0 +1,61 @@
...
+  box-shadow: var(--ais-shadow-xs);
+}
+
+@media (hover: hover) {
+  [class^='ais-'][class*='-showMore']:not(
+      [class^='ais-'][class*='-showMore--disabled']
+    ):hover {
+    background-color: rgba(var(--ais-primary-color-rgb), 0.85);
+  }
+}
```
FabienMotte, line 41: I took care of adding hover styles when it is supported

examples/react/css/src/hierarchical-menu.css
```diff
@@ -0,0 +1,73 @@
...
+  padding: calc(var(--ais-spacing-half) * 0.75) calc(var(--ais-spacing-half));
+  border-radius: calc(var(--ais-spacing) / 4);
+  transition: background-color var(--ais-transition-duration)
+      var(--ais-transition-timing-function),
+    color var(--ais-transition-duration) var(--ais-transition-timing-function);
+
+  &:before {
+    transition: transform var(--ais-transition-duration)
+      var(--ais-transition-timing-function);
+    background-image: url(data:image/svg+xml;utf8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%2216%22%20height%3D%2216%22%20viewBox%3D%220%200%2024%2024%22%20fill%3D%22none%22%20stroke%3D%22%23262626%22%20stroke-width%3D%222%22%20stroke-linecap%3D%22round%22%20stroke-linejoin%3D%22round%22%3E%3Cpolyline%20points%3D%226%209%2012%2015%2018%209%22%3E%3C%2Fpolyline%3E%3C%2Fsvg%3E);
```
FabienMotte, line 22: I had to duplicate the SVG several times because we can't use `currentColor` for the stroke, so I had to change the stroke color for when it's selected and also for dark mode.

examples/react/css/src/common.css
```diff
@@ -0,0 +1,61 @@
...
+  );
+  margin-top: calc(var(--ais-spacing-half) * 1.5);
+  transition: background-color var(--ais-transition-duration)
+    var(--ais-transition-timing-function);
+  box-shadow: var(--ais-shadow-xs);
+}
+
+@media (hover: hover) {
+  [class^='ais-'][class*='-showMore']:not(
+      [class^='ais-'][class*='-showMore--disabled']
```
FabienMotte, line 37:
```suggestion
      [class^='ais-'][disabled]
```
Do we prefer that syntax?

examples/react/css/src/common.css
```diff
@@ -0,0 +1,61 @@
+/* Motion reduction */
+@media (prefers-reduced-motion: reduce) {
+  [class^='ais-'] {
+    transition: none !important;
+  }
+}
```
thomasbritton, line 6: should this live in the `reset`?

examples/react/css/src/common.css
```diff
@@ -0,0 +1,61 @@
...
+    rgba(var(--ais-primary-color-rgb), var(--ais-primary-color-alpha));
+  outline-offset: 2px;
+}
+
+/* Widgets */
+[class^='ais-'][class*='-showMore'] {
+  border-radius: calc(var(--ais-spacing) / 4);
+  font-size: 0.9em;
+  font-weight: var(--ais-font-weight-medium);
+  color: #fff;
```
thomasbritton, line 22: should we have a variable here instead of defining `#fff`?

FabienMotte: I saw that pure white was used in the autocomplete classic theme without a variable, so I went that way.
But we could use something like `--ais-color-white-rgb: 255, 255, 255;`

examples/react/css/src/common.css
FabienMotte: No strong opinion on this one, WDYT @sarahdayan?

examples/react/css/src/common.css
thomasbritton: Could there be a case where you would use the class without the actual attribute?

examples/react/css/src/common.css
thomasbritton: I feel like it should be a variable, just incase the background color is changed for example, white might not work etc.

examples/react/css/src/common.css
FabienMotte: I'm not sure about that, do you know if it's consistent @sarahdayan?

thomasbritton: @FabienMotte I'd imagine it's quite unlikely but thought just throw it out there.

Could we have multiple variants of the selector to cover _potential_ edge cases?

examples/react/css/src/common.css
FabienMotte: I agree, done!

examples/react/css/src/common.css
sarahdayan: Can be argued, but this ruleset is about disabling opinionated styles in themes, not about resetting browser defaults, so common sounds fine.

examples/react/css/src/common.css
sarahdayan: Sounds great! Let's just make sure it doesn't clash with the focus style for the search box.

examples/react/css/src/common.css
sarahdayan: We _should_ have the `disabled` attribute present on all show more buttons. If we realize it isn't the case we can always fix it in InstantSearch, it's not breaking and is actually a good fix.

examples/react/css/src/hierarchical-menu.css
sarahdayan: Can we have an internal variable?

examples/react/css/src/variables.css
```diff
@@ -7,6 +7,8 @@
   --ais-primary-color-alpha: 1;
   --ais-muted-color-rgb: 127, 137, 164;
   --ais-muted-color-alpha: 0.6;
+  --ais-color-white-rgb: 255, 255, 255;
+  --ais-color-white-alpha: 1;
```
sarahdayan, line 11: Should we use a more semantic naming? If customers want to customize it, it's weird to set `--ais-color-white-rgb` to a different color.

examples/react/css/src/common.css
FabienMotte: `display: none` or `appearance: none;` are both working on my side

examples/react/css/src/common.css
FabienMotte: Thanks for the feedback, I'll go with `[disabled]` then

examples/react/css/src/common.css
FabienMotte: It should only be present when you use the keyboard to navigate on the page

examples/react/css/src/hierarchical-menu.css
FabienMotte: Good idea! I also added a comment at the top of each one to let us know what color it uses.
And I updated the `stroke-width` to `1.6`.

PR #6587, opened 2025-02-27
feat: add Breadcrumb styling

thomasbritton:
Add styles for the Breadcrumbs component

### DIFF
--- a/examples/react/css/src/breadcrumbs.css
+++ b/examples/react/css/src/breadcrumbs.css
@@ -3,7 +3,7 @@
     align-items: center;
     display: flex;
     flex-wrap: wrap;
-    row-gap: var(--ais-spacing-third);
+    row-gap: calc(var(--ais-spacing) / 3);
   }
 }
 
@@ -23,7 +23,7 @@
 .ais-Breadcrumb-separator {
   color: rgb(var(--ais-muted-color-rgb));
   font-weight: normal;
-  margin: 0 var(--ais-spacing-third);
+  margin: 0 calc(var(--ais-spacing) / 3);
 }
 
 .ais-Breadcrumb-link {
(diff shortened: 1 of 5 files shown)


examples/react/css/src/variables.css
FabienMotte, line 1: You can update the `--ais-muted-color-rgb` and `--ais-border-color-rgb` CSS variables to `109, 115, 151` so that we're WCAG compliant on a white background

examples/react/css/src/breadcrumbs.css
```diff
@@ -0,0 +1,38 @@
...
+
+.ais-Breadcrumb-separator {
+  color: rgb(var(--ais-muted-color-rgb));
+  font-weight: normal;
+  margin: 0 var(--ais-spacing-third);
+}
+
+.ais-Breadcrumb-link {
+  color: rgb(var(--ais-muted-color-rgb));
+  transition: color 0.2s ease-out;
```
FabienMotte, line 31: You can use the CSS variables `--ais-transition-duration` and `--ais-transition-timing-function`

examples/react/css/src/variables.css
```diff
@@ -25,6 +25,7 @@
   --ais-spacing-factor: 1;
   --ais-spacing: calc(var(--ais-base-unit) * var(--ais-spacing-factor) * 1px);
   --ais-spacing-half: calc(var(--ais-spacing) / 2);
+  --ais-spacing-third: calc(var(--ais-spacing) / 3);
```
FabienMotte, line 28: I think we decided to stick with `--ais-spacing` only based on @sarahdayan's PR: https://github.com/algolia/instantsearch/pull/6586/files#diff-268c7ac17cc62a4721d38ee84477fa90adb65aba77242d7fffb1c75d6a08488a

PR #6593, opened 2025-02-28
feat: current refinements & clear refinements

thomasbritton:
Current Refinements and Clear Refinements buttons

Screenshot 2025-03-25 at 10 09 22

Screenshot 2025-03-25 at 10 09 37

### DIFF
--- a/examples/react/css/src/App.css
+++ b/examples/react/css/src/App.css
@@ -32,3 +32,7 @@ body {
   flex-direction: column;
   row-gap: 1rem;
 }
+
+.wrapper {
+  margin-block: 1rem;
+}
--- a/examples/react/css/src/App.tsx
+++ b/examples/react/css/src/App.tsx
@@ -74,8 +74,10 @@ export function App() {
               'hierarchicalCategories.lvl2',
             ]}
           />
-          <CurrentRefinements />
-          <ClearRefinements />
+          <div className="wrapper">
+            <CurrentRefinements />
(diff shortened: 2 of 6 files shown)


sarahdayan:
## Clear refinements

I think the button should look the same as the show more buttons in the refinement list and hierarchical menu, what do you think?

## Current refinements

The border and icon colors look too dark to me, I think we should use the alpha values that go with the picked colors to have something more consistent, + add a hover color on the delete button to convey that there's an action.

Capture d’écran 2025-03-03 à 11 11 50

If you don't have time to go over it @thomasbritton but you agree, I can take care of it.

thomasbritton:
> ## Clear refinements
> I think the button should look the same as the show more buttons in the refinement list and hierarchical menu, what do you think?
> 
> ## Current refinements
> The border and icon colors look too dark to me, I think we should use the alpha values that go with the picked colors to have something more consistent, + add a hover color on the delete button to convey that there's an action.
> 
> Capture d’écran 2025-03-03 à 11 11 50
> If you don't have time to go over it @thomasbritton but you agree, I can take care of it.

Hey @sarahdayan , sure I think those suggestions sound good, but I'm probably not going to have much time to look into it, so if you don't mind taking over it that would be great.

thomasbritton:
@sarahdayan @FabienMotte I have made amends based on the feedback

FabienMotte (approved):

PR #6594, opened 2025-02-28
poc: add `Hits`, `Highlight` and `Snippet` styles

FabienMotte:
**Summary**

This adds the `Hits`, `Highlight` and `Snippet` styles.

Hits in light mode
Hits in dark mode

### DIFF
--- a/examples/react/css/src/highlight.css
+++ b/examples/react/css/src/highlight.css
@@ -1,0 +1,13 @@
+.ais-Highlight-highlighted,
+.ais-Snippet-highlighted {
+  background-color: rgba(
+    var(--ais-primary-color-rgb),
+    var(--ais-primary-color-alpha)
+  );
+  color: rgba(
+    var(--ais-button-text-color-rgb),
+    var(--ais-button-text-color-alpha)
+  );
+  border-radius: var(--ais-radius-xs);
+  padding: calc(var(--ais-spacing) * 0.125);
+}
--- a/examples/react/css/src/highlight.css
+++ b/examples/react/css/src/highlight.css
@@ -1,13 +1,7 @@
 .ais-Highlight-highlighted,
 .ais-Snippet-highlighted {
-  background-color: rgba(
-    var(--ais-primary-color-rgb),
-    var(--ais-primary-color-alpha)
-  );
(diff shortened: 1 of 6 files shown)


thomasbritton (approved):

sarahdayan:
Some feedback on the highlight, I think this is a bit too contrasted in terms of information hierarchy, and makes the highlighted parts look like buttons.

What do you think of going with something like this instead?

Capture d’écran 2025-03-03 à 11 22 32

Capture d’écran 2025-03-03 à 11 22 52

Used values:

```css
background-color: rgba(var(--ais-primary-color-rgb), 0.2); // 0.1 in light mode
color: rgba(var(--ais-primary-color-rgb), var(--ais-primary-color-alpha));
border-radius: 2px; // we don't have to use the same radius as for buttons, it makes sense for customers to want to change them independently
```

Thoughts?

FabienMotte:
> Some feedback on the highlight, I think this is a bit too contrasted in terms of information hierarchy, and makes the highlighted parts look like buttons.
> 
> What do you think of going with something like this instead?
> 
> Capture d’écran 2025-03-03 à 11 22 32 Capture d’écran 2025-03-03 à 11 22 52
> Used values:
> 
> ```css
> background-color: rgba(var(--ais-primary-color-rgb), 0.2); /* 0.1 in light mode */
> color: rgba(var(--ais-primary-color-rgb), var(--ais-primary-color-alpha));
> border-radius: 2px; /* we don't have to use the same radius as for buttons, it makes sense for customers to want to change them independently */
> ```
> 
> Thoughts?

Good call, thanks!

examples/react/css/src/highlight.css
```diff
@@ -0,0 +1,7 @@
+.ais-Highlight-highlighted,
+.ais-Snippet-highlighted {
+  background-color: rgba(var(--ais-primary-color-rgb), 0.2);
+  color: rgba(var(--ais-primary-color-rgb), var(--ais-primary-color-alpha));
+  border-radius: 2px;
```
sarahdayan, line 5: We can move this to a variable later.

sarahdayan (approved):

PR #6596, opened 2025-02-28
feat: toggle refinement styling

thomasbritton:
Screenshot 2025-02-28 at 18 40 52
Screenshot 2025-02-28 at 18 40 46
Screenshot 2025-02-28 at 18 40 36
Screenshot 2025-02-28 at 18 40 32

### DIFF
--- a/examples/react/css/src/toggle-refinement.css
+++ b/examples/react/css/src/toggle-refinement.css
@@ -1,0 +1,47 @@
+.ais-ToggleRefinement-label {
+  font-size: 0.875rem;
+  line-height: 1.75rem;
+  display: block;
+  cursor: pointer;
+  flex-direction: row-reverse;
+  justify-content: space-between;
+  align-items: center;
+  display: flex;
+}
+
+.ais-ToggleRefinement-checkbox {
+  appearance: none;
+  height: 18px;
+  pointer-events: none;
+  width: 32px;
+  position: relative;
+  border: 1px solid rgba(var(--ais-muted-color-rgb), 0.3);
+  background-color: rgb(var(--ais-background-color-rgb));
+  border-radius: 9999px;
+}
+
+.ais-ToggleRefinement-checkbox:checked {
+  background-image: none !important;
+}
+
+.ais-ToggleRefinement-checkbox::after {
+  appearance: none;
+  content: '';
+  height: 16px;
+  width: 16px;
+  position: absolute;
+  top: 0;
+  left: 0;
+  background: rgb(var(--ais-muted-color-rgb));
+  border-radius: 9999px;
+  transition: 0.1s ease-in-out;
+}
+
+.ais-ToggleRefinement-checkbox:checked::after {
+  transform: translateX(14px);
+  background-color: rgb(var(--ais-background-color-rgb));
+
+  @media (prefers-color-scheme: dark) {
+    background-color: #fff;
+  }
+}
--- a/examples/react/css/src/algolia.css
+++ b/examples/react/css/src/algolia.css
@@ -8,4 +8,5 @@
 @import url('breadcrumbs.css');
 @import url('hierarchical-menu.css');
 @import url('refinement-list.css');
+@import url('toggle-refinement.css');
 @import url('panel.css');


examples/react/css/src/toggle-refinement.css
```diff
@@ -0,0 +1,47 @@
...
+  display: flex;
+}
+
+.ais-ToggleRefinement-checkbox {
+  appearance: none;
+  height: 18px;
+  pointer-events: none;
+  width: 32px;
+  position: relative;
+  border: 1px solid rgba(var(--ais-muted-color-rgb), 0.3);
```
FabienMotte, line 18: Should we use `--ais-border-color-rgb`?

examples/react/css/src/toggle-refinement.css
```diff
@@ -0,0 +1,47 @@
...
+  background-color: rgb(var(--ais-background-color-rgb));
+  border-radius: 9999px;
```
FabienMotte, line 20:
```suggestion
  border-radius: calc(infinity * 1px);
```
For consistency, I used this pattern in `common.css`

examples/react/css/src/toggle-refinement.css
```diff
@@ -0,0 +1,47 @@
...
+}
+
+.ais-ToggleRefinement-checkbox:checked {
+  background-image: none !important;
```
FabienMotte, line 24: Are you doing this because of this rule in `common.css`?
```css
input[class^='ais-'][type='checkbox']:checked { /* ... */ }
```
I think we can update the selector and move it to `refinement-list.css` since your checkbox is more like a switch.

examples/react/css/src/toggle-refinement.css
```diff
@@ -0,0 +1,47 @@
...
+.ais-ToggleRefinement-checkbox::after {
+  appearance: none;
+  content: '';
+  height: 16px;
+  width: 16px;
+  position: absolute;
+  top: 0;
+  left: 0;
+  background: rgb(var(--ais-muted-color-rgb));
+  border-radius: 9999px;
```
FabienMotte, line 36: Same as my previous comment

examples/react/css/src/toggle-refinement.css
```diff
@@ -0,0 +1,47 @@
...
+  transition: 0.1s ease-in-out;
```
FabienMotte, line 37: Do you think we can use `--ais-transition-duration` and `--ais-transition-timing-function` variables?

examples/react/css/src/toggle-refinement.css
```diff
@@ -0,0 +1,47 @@
...
+}
+
+.ais-ToggleRefinement-checkbox:checked::after {
+  transform: translateX(14px);
+  background-color: rgb(var(--ais-background-color-rgb));
+
+  @media (prefers-color-scheme: dark) {
+    background-color: #fff;
```
FabienMotte, line 45: Do we want to use a variable for that color?

examples/react/css/src/toggle-refinement.css
```diff
@@ -0,0 +1,47 @@
...
+
+.ais-ToggleRefinement-checkbox:checked {
+  background-image: none !important;
+}
+
+.ais-ToggleRefinement-checkbox::after {
+  appearance: none;
+  content: '';
+  height: 16px;
+  width: 16px;
```
FabienMotte, line 31:
```suggestion
  width: var(--ais-spacing);
  height: var(--ais-spacing);
```
I used this variable for the `RefinementList' checkbox, not sure if we should stick to pixels

examples/react/css/src/toggle-refinement.css
```diff
@@ -0,0 +1,44 @@
...
+  line-height: 1.75rem;
+  display: block;
+  cursor: pointer;
+  flex-direction: row-reverse;
+  justify-content: space-between;
+  align-items: center;
+  display: flex;
+}
+
+input[class^='ais-'][class*='ToggleRefinement-checkbox'] {
```
sarahdayan, line 12: I'm not sure why we need such selector, why not target `.ais-ToggleRefinement-checkbox`?

thomasbritton: This is because there is a selector in `common.css` that is of higher specificity `input[class^='ais-'][type='checkbox']:checked { /* ... */ }`

I did it this way because we were out of time so just needed something quick that would work but the class above probably needs to be merged into the `refinement-list.css`

thomasbritton: @sarahdayan I have re-worked so it uses the default class name `.ais-ToggleRefinement-checkbox`

Haroenv:
Do you still want this merged in the base branch or restart when this project continues?

PR #6606, opened 2025-03-12
chore: don't transform facets as disjunctive if they have no value selected

e-krebs:
**Summary**

[EMERCH-1942](https://algolia.atlassian.net/browse/EMERCH-1942)

- the Composition API has a limit of 20 facets annotated as `disjunctive`
- to limit the risk of reaching that limit, this PR does not annotated as `disjunctive` facets who do not have value selected

The reasoning is that when a facet has no value selected, the list of facet values will be the same with or without a disjunctive request.

**Result**

Less facets are annotated as disjunctive.
Notably, the `RefinementList` widget now does not annotate them by default anymore.

[EMERCH-1942]: https://algolia.atlassian.net/browse/EMERCH-1942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/algoliasearch-helper/src/requestBuilder.js
+++ b/packages/algoliasearch-helper/src/requestBuilder.js
@@ -187,7 +187,16 @@ var requestBuilder = {
     var facets = state.facets
       .concat(
         state.disjunctiveFacets.map(function (value) {
-          return 'disjunctive(' + value + ')';
+          if (
+            state.disjunctiveFacetsRefinements &&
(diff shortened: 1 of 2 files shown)


Haroenv (approved):
there's other ways to check as well with helper functions, but in this case it's probably best to stay with the explicit check.

Thanks!

PR #6614, opened 2025-03-28
fix(useSearchResults): force setting results

dhayab:
**Summary**

This PR updates how `useSearchResults()` sets its results state. Instead of waiting for InstantSearch's `render` event, it immediately sets the state if results are available.

This can help in situations where results are received before the event listener can be set.

CR-8014


Haroenv (approved):

Haroenv:
We've paired in trying to recreate the situation for the test where you're adding this hook while in its render the results aren't yet available, but by the time the useEffect gets fired, the results are already in. Testing this would require changing when useEffect gets fired somehow, which we couldn't do

PR #6616, opened 2025-04-02
fix(server): reject if error in search or recommend

### LINKED ISSUE #6525: Infinite Page Loading Next.js when no connection to search server
### 🐛 Current behavior

Next.js version: 15.0.4

I'm using InstantSearchNext together with Typesense (`typesense-instantsearch-adapter`). Everything works fine, except when there's no connection to the self hosted Typesense server. In this case, the following things happen:

* Server side: A huge nested stack trace is displayed with this content:
  ```
  constructor: [Function: AxiosError] { from: [Function (anonymous)] },
  toJSON: [Function: toJSON],
  [AggregateError: ] { code: 'ECONNREFUSED' }
  ```
* Client side: Infinite loading screen.

Yes, this is problem is more relevant for self hosted search servers, but I would expect this behaviour to appear as well when there's a connection problem with Algolia.

### 🔍 Steps to reproduce

1. Install InstantSearchNext
2. Point the search client to a non-existing URL
3. Open the page with SSR enabled

I tried to replicate the local setup on codesandbox, but I see a different error there.

### Live reproduction

https://codesandbox.io/p/devbox/nifty-snowflake-fgxqcg?file=%2Fcomponents%2Fsearch.tsx%3A5%2C9-5%2C21

### 💭 Expected behavior

I would expect to see the Next.js error component being sent to the browser.

Could it be that the error is not bubbled up correctly? Or it's bubbled up, but some code is not reacting to it? I looked into this code a bit and was wondering if the reason is that the promise only resolved on "result": `once('result'`
https://github.com/algolia/instantsearch/blob/master/packages/react-instantsearch-nextjs/src/InitializePromise.tsx

I found this previous issue that sounds related to the current problem. The solution at that time was to adjust the error propagation logic: https://github.com/algolia/instantsearch/issues/5233

Thanks for any support on this issue!

### Package version

react-instantsearch-nextjs@0.3.20 react-instantsearch@7.13.9 typesense-instantsearch-adapter@2.8.0

### Operating system

maxOS 15.2

### Browser

Chrome 131

### Code of Conduct

Haroenv:
**Summary**

Refactors InitialisePromise to use the waitForResults of InstantSearch.js, which correctly rethrows errors if they happen in search, composition or recommend.

Be careful when refactoring to ensure I didn't misunderstand the intent of the original code, as I don't remember why we didn't use InstantSearch' waitForResults in the first place. There may be a subtle difference between the functions I didn't catch.

**Result**

fixes #6525

### DIFF
--- a/packages/react-instantsearch-nextjs/src/__tests__/InitializePromise.test.tsx
+++ b/packages/react-instantsearch-nextjs/src/__tests__/InitializePromise.test.tsx
@@ -32,18 +32,18 @@ const renderComponent = ({
   ref = { current: null },
   nonce,
   insertedHTML,
+  client = createSearchClient({
+    getRecommendations: jest.fn().mockResolvedValue({
+      results: [createSingleSearchResponse()],
+    }),
+  }),
 }: {
   children?: React.ReactNode;
   ref?: { current: PromiseWithState<void> | null };
   nonce?: string;
   insertedHTML?: jest.Mock;
+  client?: ReturnType<typeof createSearchClient>;
 } = {}) => {
-  const client = createSearchClient({
-    getRecommendations: jest.fn().mockResolvedValue({
-      results: [createSingleSearchResponse()],
-    }),
-  });
-
   render(
     <InstantSearchRSCContext.Provider value={ref}>
       <InstantSearchSSRProvider>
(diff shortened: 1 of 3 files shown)


shaejaz (approved):

dhayab (approved):
Great refactoring. The only difference I see is that we now explicitly search again from within `waitForResults()` whereas before the search would have happened earlier. Don't think it makes a difference in the end.

Haroenv:
I want to evaluate the feedback in https://github.com/algolia/instantsearch/issues/6525 a little more, it seems that if the search throws there's still some sort of (in)finite nesting of the error going on, although search seems to be called only once

packages/react-instantsearch-nextjs/src/__tests__/InitializePromise.test.tsx
```diff
@@ -32,18 +32,18 @@ const renderComponent = async ({
   ref = { current: null },
   nonce,
   insertedHTML,
+  client = ccreateSearchClient({
```
Haroenv:
```suggestion
  client = createSearchClient({
```

PR #6620, opened 2025-04-07
feat(recommend): filter out input objectIDs [RECO-2436]

raed667:
**Summary**

When requesting multiple objectIDs, there is a chance that the merged results contains some of the input items.

We can filter those out for a more consistent output.

### DIFF
--- a/packages/algoliasearch-helper/src/utils/sortAndMergeRecommendations.js
+++ b/packages/algoliasearch-helper/src/utils/sortAndMergeRecommendations.js
@@ -21,11 +21,14 @@ function getAverageIndices(indexTracker, nrOfObjs) {
   });
 }
 
-function sortAndMergeRecommendations(results) {
+function sortAndMergeRecommendations(objectIDs, results) {
   var indexTracker = {};
 
   results.forEach(function (hits) {
     hits.forEach(function (hit, index) {
+      // filter out hits that match the source objectIDs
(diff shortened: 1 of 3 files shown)


raed667:
@Haroenv 

> should this not be done backend 


The backend doesn't have a concept of a query with multiple objectIDs, they're treated as separate independent queries.


For example:  `getFBT(['obj1', 'obj2']` will result in 2 different Recommend calls, with the risk that `getFBT('obj1')` returns a list of recommendations that include `obj2` and vise-versa

packages/algoliasearch-helper/src/utils/sortAndMergeRecommendations.js
```diff
@@ -21,20 +21,24 @@ function getAverageIndices(indexTracker, nrOfObjs) {
...
-        indexTracker[hit.objectID] = {
-          indexSum: indexTracker[hit.objectID].indexSum + index,
-          count: indexTracker[hit.objectID].count + 1,
-        };
-      }
-    });
+    hits
+      .filter(function (hit) {
+        return !objectIDs.includes(hit.objectID);
+      })
```
Haroenv, line 31: you can put this inside the forEach and early return, that's at least one iteration less

Haroenv (commented):
unavoidable to do this frontend for now, but I'd like to see the multi query move backend

PR #6629, opened 2025-05-01
Fix/react freeze fix issue keep helper use index

dmitry-blackwave:
**Summary**

**Result**


Haroenv:
Thanks for your PR. In which situation do you see freeze/unfreeze operations? with an added test to explain the scenario this could be a good fix

dmitry-blackwave:
Yes, I'm planning to add missing stuff to the description this weekend

PR #6630, opened 2025-05-01
Update app.js to function with Javascript client v5 codesandbox

palacejackson:
There were breaking changes with the upgrade to v5 causing codeSandbox to not work correctly

Made the required changes to client etc so it now functions

**Summary**

**Result**


Haroenv:
thanks for your PR, but it should be made to the source: https://github.com/algolia/instantsearch/tree/master/packages/create-instantsearch-app/src/templates. It seems all templates may still use v4 imports and dependencies. It's probably interesting to update them all.

PR #6631, opened 2025-05-02
fix(insights): use latest appId and apiKey

### LINKED ISSUE #6623: Updating Secured API key for react-instantsearch isn't updating the URL for insights (analytics) network call with correct API key
### 🐛 Current behavior

Hi team

I am not sure whether this is a bug or not, but I was unable to find the documentation pertaining to `react-instantsearch` with updating Secured API key. 

So, we have implemented SECURED API key to fetch the key with an expiry time from our backend. We are fetching and updating the API key when we get the error : **Invalid Application-ID or API key** while doing a query to Algolia.

We have `insights={true}` but we can see that the event analytics is still pointing to the previous incorrect API key. We are getting correct data from the server for Search, but the analytics is getting **401 Unauthorized error**. 
This is our implementation of updating the key for `searchClient`

```js
const { addMiddlewares, error } = useInstantSearch();

  useEffect(() => {
    const middleware = ({ instantSearchInstance }) => {
      function handleError(searchError) {

        if (searchError.message === "Invalid Application-ID or API key") {
          CLIENT.setClientApiKey({
            apiKey: newAPIKey,
          });

          instantSearchInstance.refresh();
        }
      }

      return {
        subscribe() {
          instantSearchInstance.addListener("error", handleError);
        },
        unsubscribe() {
          instantSearchInstance.removeListener("error", handleError);
        },
      };
    };

    return addMiddlewares(middleware);
  }, [addMiddlewares]);

```

Also, one more additional question for this implementation:

Since we get the expiry time in our API key, is there a way we can check the expiry before the ALGOLIA query is called? And then can we set the new API key for making the query?

**P.S.**: In the Codesandbox, I am not fetching a key from backend, but I am just using an incorrect API key at the start. Then, when I get the invalid API key error, I update the API with the corrrect data.

### 🔍 Steps to reproduce

1. Go to [https://mvkzz9.csb.app/](https://mvkzz9.csb.app/)
2. Open the network tab
3. Click on any filter. See the query is coming correct, but the **events** network call getting 401 error due to the previous incorrect API key

### Live reproduction

https://codesandbox.io/p/sandbox/mvkzz9

### 💭 Expected behavior

Analytic Events at first will be denied due to incorrect API key, but after updating the API key, it should work properly (the query param with the API key should get updated).

### Package version

"algoliasearch": "^5.11.0",     "react-instantsearch": "7.15.5", "react": "^18.3.1",     "react-dom": "^18.3.1",

### Operating system

MacOs 15.4

### Browser

Microsoft Edge Version 134.0.3124.95 (Official build) (arm64)

### Code of Conduct

aymeric-giraudet:
**Summary**

[FX-3372](https://algolia.atlassian.net/browse/FX-3372)

Fixes #6623 

**Result**

We should use the latest `appId` and `apiKey` as the latter can be rotated

[FX-3372]: https://algolia.atlassian.net/browse/FX-3372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -30,7 +30,7 @@
     },
     {
       "path": "packages/vue-instantsearch/vue3/umd/index.js",
-      "maxSize": "69.50 kB"
+      "maxSize": "69.75 kB"
     },
     {
       "path": "packages/vue-instantsearch/vue2/cjs/index.js",
--- a/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
+++ b/packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
@@ -38,15 +38,20 @@ declare global {
 }
 
 describe('insights', () => {
-  const searchClientWithCredentials = createSearchClient({
-    // @ts-expect-error only available in search client v4
-    transporter: {
-      headers: {
-        'x-algolia-application-id': 'myAppId',
-        'x-algolia-api-key': 'myApiKey',
+  let searchClientWithCredentials: SearchClient;
+
+  beforeEach(() => {
+    searchClientWithCredentials = createSearchClient({
+      // @ts-expect-error only available in search client v4
+      transporter: {
+        headers: {
+          'x-algolia-application-id': 'myAppId',
+          'x-algolia-api-key': 'myApiKey',
+        },
       },
-    },
+    });
   });
+
   const createTestEnvironment = ({
     searchClient = searchClientWithCredentials,
     started = true,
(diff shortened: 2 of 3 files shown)


dhayab (approved):

Haroenv:
This makes sense, do we still use the initial appid,apikey then? can we change it so it's only retrieved in sendEvent itself?

aymeric-giraudet:
@Haroenv Retrieving it when the middleware starts is only for old versions of search insights now, for this users will have to call init. For modern versions, we retrieve it each time in `sendEvent`

PR #6634, opened 2025-05-07
fix(connectHits): use optional chaining for results.hits length check

Wcdaren:
**Summary**

image

When I use searchkit, even if I report a fixed format containing hits, I still encounter this situation where hits are missing. Therefore, I think a judgment can be added here

**Result**


Haroenv:
can you add a PR to search kit to ensure an empty hits is always returned? this matches the guarantees of the search method

PR #6644, opened 2025-06-03
fix(infiniteHits): always return valid showPrevious and showMore functions

Haroenv:
**Summary**

We default the infinite hits showPrevious to true, but in the initial render the functions `showPrevious` and `showMore` are not yet present in the sate. This fixes that and ensures the functions are always returned.

**Result**

- show more button always shows up by default (and doesn't flash)
- ensure that showMore and showPrevious functions are always defined, even if the first render is _with_ results (artificial for example)


Haroenv:
you can ignore the netlify failure, it's fixed in the retry but the status didn't update. Seemingly cache related, not to my PR

PR #6645, opened 2025-06-09
fix(breadcrumb): use hierarchical facets only

shaejaz:
**Summary**
Breadcrumb connector accesses all facet names when generating the breadcrumb list. This results in a menu widget breaking functionality. This PR restricts the breadcrumb connector to only use attribute from the hierarchical menu to prevent this.

[CR-8481]

**Result**

Breadcrumbs now only shows hierarchical facets.

This also initializes an 'interaction' shared test suite which tests functionality of widgets used in tandem (only breadcrumbs + hierarchical menu + menu for now)

Codesandbox: https://codesandbox.io/p/sandbox/example-react-instantsearch-getting-started-l7y5j9

[CR-8481]: https://algolia.atlassian.net/browse/CR-8481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
+++ b/packages/instantsearch.js/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
@@ -345,6 +345,85 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
       });
     });
 
+    it('returns an empty array of items if only none hierarchicalFacets result exist', () => {
+      const renderFn = jest.fn();
+      const unmountFn = jest.fn();
+      const createBreadcrumb = connectBreadcrumb(renderFn, unmountFn);
+      const breadcrumb = createBreadcrumb({
+        attributes: ['category', 'subCategory'],
+      });
+      const helper = algoliasearchHelper(
+        createSearchClient(),
+        'indexName',
+        breadcrumb.getWidgetSearchParameters!(
+          new SearchParameters({
+            hierarchicalFacets: [
+              {
+                attributes: ['country'],
+                name: 'country',
+              },
+            ],
+          }),
+          {
+            uiState: {},
+          }
+        )
+      );
+
+      helper.toggleFacetRefinement('country', 'country');
+
+      const renderState1 = breadcrumb.getWidgetRenderState(
+        createInitOptions({ helper })
+      );
+
+      expect(renderState1).toEqual({
+        canRefine: false,
+        createURL: expect.any(Function),
+        items: [],
+        refine: expect.any(Function),
+        widgetParams: { attributes: ['category', 'subCategory'] },
+      });
+
+      helper.toggleFacetRefinement('category', 'Decoration');
+
+      const renderState2 = breadcrumb.getWidgetRenderState(
+        createRenderOptions({
+          helper,
+          state: helper.state,
+          results: new SearchResults(helper.state, [
+            createSingleSearchResponse({
+              hits: [],
+              facets: {
+                category: {
+                  Decoration: 880,
+                },
+                subCategory: {
+                  'Decoration > Candle holders & candles': 193,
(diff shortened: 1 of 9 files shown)


packages/instantsearch.js/src/connectors/breadcrumb/connectBreadcrumb.ts
```diff
@@ -194,9 +195,14 @@ const connectBreadcrumb: BreadcrumbConnector = function connectBreadcrumb(
             return [];
           }
 
-          const [{ name: facetName }] = state.hierarchicalFacets;
+          const facet = find(state.hierarchicalFacets, (f) =>
+            Boolean(f.separator)
```
Haroenv: here you're checking if it has a separator (I guess because menu doesn't have one). Could we check if the attributes match maybe?

shaejaz: Yeah, this was more a poc change to see if the resulting functionality of the widgets is correct. You can check via the sandbox linked above. Basically breadcrumb will only pick up hierarchical menu facets and ignore menu ones.

I think a better way would be to use `hierarchicalFacetName` which is used throughout the connector. I've made this change in the latest commit, let me know if you think it's correct. I'll add some test cases for it in that case.

Haroenv: yes that looks correct. Worth investigating why it ever was the other way round though

shaejaz: Seems to have been this way since `getWidgetRenderState` was implemented for breadcrumbs.
Could it be the case that breadcrumb needs to also display menu attributes as well? Otherwise seems like an oversight and we caught it now since using both hierarchical menu and a normal menu isn't that common I believe

Haroenv: no, breadcrumb is only meant to work with hierarchical menu. I guess it was an oversight since until recently you had to mount them in the right order and breadcrumb didn't work standalone either. Let's go!

packages/instantsearch.js/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
```diff
@@ -345,6 +345,85 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
       });
     });
 
+    it('returns an empty array of items if only none hierarchicalFacets result exist', () => {
```
Haroenv:
```suggestion
    it('returns an empty array of items if only non-hierarchicalFacets result exist', () => {
```

packages/instantsearch.js/src/connectors/breadcrumb/__tests__/connectBreadcrumb-test.ts
```diff
@@ -345,6 +345,85 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/breadcrumb/
...
+      );
+
+      expect(renderState2).toEqual({
+        canRefine: true,
+        createURL: expect.any(Function),
+        items: [{ label: 'Decoration', value: null }],
+        refine: expect.any(Function),
+        widgetParams: { attributes: ['category', 'subCategory'] },
+      });
+    });
```
Haroenv, line 425: can we add a test that has both breadcrumb, hierarchical menu and menu in the common test suite maybe? would be nice to see a more complete test for the case we're fixing.

shaejaz: Good call, I'll add it there

Haroenv (approved):

PR #6647, opened 2025-06-10
feat(nextjs): add an `instance` prop to keep same instance across remounts

aymeric-giraudet:
**Summary**

[FX-3387](https://algolia.atlassian.net/browse/FX-3387)

[FX-3387]: https://algolia.atlassian.net/browse/FX-3387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react-instantsearch-core/src/lib/useInstantSearchApi.ts
+++ b/packages/react-instantsearch-core/src/lib/useInstantSearchApi.ts
@@ -299,8 +299,8 @@ function warnNextAppDir(isRscContextDefined: boolean) {
     Boolean((window as any).next?.appDir) === false,
     `
 We've detected you are using Next.js with the App Router.
-We released an **experimental** package called "react-instantsearch-nextjs" that makes SSR work with the App Router.
-Please check its usage instructions: https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react/#with-nextjs
+We released a package called "react-instantsearch-nextjs" that makes SSR work with the App Router.
+Please check its usage instructions: https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react/#with-nextjs-app-router
 
 This warning will not be outputted in production builds.`
   );
--- a/packages/react-instantsearch-core/src/lib/useWidget.ts
+++ b/packages/react-instantsearch-core/src/lib/useWidget.ts
@@ -118,7 +118,7 @@ With Next.js, you need to set \`skipSuspense\` to \`true\` for all but the last 
 
 This warning can be a false positive if you are using dynamic widgets or multi-index, in which case you can ignore it by setting \`ignoreMultipleHooksWarning\` to \`true\` in \`<InstantSearchNext\`.
 
-For more information, see https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react/#app-router`
+For more information, see https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react/#composing-hooks`
     );
   }
 }
--- a/packages/react-instantsearch-nextjs/src/useDynamicRouteWarning.ts
+++ b/packages/react-instantsearch-nextjs/src/useDynamicRouteWarning.ts
@@ -4,18 +4,22 @@ import { useParams } from 'next/navigation';
 
 import { warn } from './warn';
 
+import type { InstantSearchNextInstance } from './InstantSearchNext';
+
 let lastParams: ReturnType<typeof useParams> | null = null;
 
 export function useDynamicRouteWarning({
   isServer,
   isMounting,
+  instance,
 }: {
   isServer: boolean;
   isMounting: React.RefObject<boolean>;
+  instance?: InstantSearchNextInstance;
 }) {
   const params = useParams();
 
-  if (!isServer && isMounting.current) {
+  if (!isServer && isMounting.current && !instance) {
     // Warn if lastParams and current params have the same keys but different values
     if (
       lastParams &&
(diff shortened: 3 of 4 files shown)


packages/react-instantsearch-nextjs/src/useDynamicRouteWarning.ts
```diff
@@ -0,0 +1,41 @@
...
+  isServer: boolean;
+  isMounting: React.RefObject<boolean>;
+}) {
+  const params = useParams();
+
+  if (!isServer && isMounting.current) {
+    // Warn if lastParams and current params have the same keys but different values
+    if (
+      lastParams &&
+      Object.keys(lastParams).length === Object.keys(params).length &&
```
Haroenv, line 26: technically only checking if the same amount of keys, not that the keys are the same

packages/react-instantsearch-core/src/lib/useInstantSearchApi.ts
```diff
@@ -217,10 +217,13 @@ export function useInstantSearchApi<TUiState extends UiState, TRouteState>(
       }
 
       return () => {
+        if (serverState?.ssrSearchRef) {
+          return;
+        }
+
```
aymeric-giraudet, line 227: Try to start it again ?

aymeric-giraudet: Or document that they should dispose / find a way to detect they changed routes

packages/react-instantsearch-nextjs/package.json
```diff
@@ -45,13 +45,26 @@
...
     "build:es": "yarn build:es:base --quiet",
     "build:types": "tsc -p ./tsconfig.declaration.json --outDir ./dist/es",
+    "test:start-server": "yarn workspace example-react-instantsearch-next-app-dir-example build && yarn workspace example-react-instantsearch-next-app-dir-example start",
+    "test:e2e": "start-server-and-test test:start-server 3000 'wdio run ./wdio.conf.cjs'",
+    "test:e2e:saucelabs": "start-server-and-test test:start-server 3000 'wdio run ./wdio.saucelabs.conf.cjs'",
     "test:exports": "node ./__tests__/module/is-es-module.mjs && node ./__tests__/module/is-cjs-module.cjs",
     "watch:es": "yarn --silent build:es:base --watch"
   },
   "devDependencies": {
+    "@types/jasmine": "3.3.16",
```
shaejaz, line 55: Should there be corresponding yarn lock changes for these?

examples/react/next-app-router/components/QueryId.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { useContext, use } from 'react';
+import { useInstantSearch, InstantSearchRSCContext } from 'react-instantsearch';
+
+export function QueryId() {
+  const { results } = useInstantSearch();
+  const { waitForResultsRef } = useContext(InstantSearchRSCContext);
```
shaejaz, line 6: Can this be `useRSCContext` instead?

packages/react-instantsearch-nextjs/package.json
aymeric-giraudet: No I picked the same versions as for other e2e tests

examples/react/next-app-router/components/QueryId.tsx
aymeric-giraudet: It's not meant to be public, this is just for tests. The context itself is public but not documented

aymeric-giraudet: Maybe we'll want something nicer to be made public later though

packages/react-instantsearch-core/src/lib/useWidget.ts
```diff
@@ -89,22 +93,32 @@ export function useWidget<TWidget extends Widget | IndexWidget, TProps>({
...
-      use(waitingForResultsRef.current);
+      use(waitForResultsRef.current);
     }
   }
+
+  if (waitForResultsRef?.current?.status === 'fulfilled') {
+    countRef.current += 1;
+    warn(
+      countRef.current > parentIndex.getWidgets().length &&
+        !ignoreMultipleHooksWarning,
```
izorg, line 114: Hello @aymeric-giraudet I can't understand the logic of this expression. Could you help me. I checked `warn` description:

`Logs a warning if the condition is not met.` 

https://github.com/algolia/instantsearch/blob/c133170e563592dfc15a95daced1f8447327a09a/packages/react-instantsearch-core/src/lib/warn.ts#L11-L15

Originally I expected to skip warning by setting `ignoreMultipleHooksWarning` to `true`, but it does not work because when first condition is false the second condition never executed. I could expect: `countRef.current <= parentIndex.getWidgets().length || ignoreMultipleHooksWarning` in this case.

Could you check it please? I could prepare PR if needed. I checked example in repository https://github.com/algolia/instantsearch/tree/b4cb69c8cfaa9dc22dfe8fe695568cf5200f3893/examples/react/next-app-router and in dev it shows the error when it's not expected.

Haroenv: thanks, that's indeed the wrong logic and it should be `countRef.current <= parentIndex.getWidgets().length && !ignoreMultipleHooksWarning` I believe, it will skip the warning if the value is `false`. We probably should change that behaviour of warning, it's easy to reason about it wrong.

You can make a pull request, ideally with an added test to ensure it doesn't regress. thanks!

aymeric-giraudet: I think `ignoreMultipleHooksWarning || countRef.current <= parentIndex.getWidgets().length` is best so we don't call `getWidgets` if the first statement is `true`.
Good catch ! Not the first time I get confused by the fact it logs when the condition is false 😭

Haroenv: Ah right, that would be better indeed too

izorg: PR is done https://github.com/algolia/instantsearch/pull/6825, but I did not manage to write correct test case close to real usage

PR #6650, opened 2025-06-11
feat(server): reset scheduleSearch only after all components mount

Haroenv:
In Remix a bug was reported [CR-8629], which said that there's an edge case happening where (not always) the hydrating DOM does not match the server DOM. This is caused by InfiniteHits that does not give hits if we are currently searching. Why are we searching? Because the `scheduleSearch` function is reset before all widgets are mounted. Until now we rely on `defer` to reset the function only once all widgets are mounted, but that's not reliable.

The solution: reset `scheduleSearch` only after the `children` of the `InstantSearch` component have been mounted. For that two new private properties are added to InstantSearch: `_manuallyResetScheduleSearch` (set in the useInstantSearchApi hook when `initialResults` are set) and a new function `_resetScheduleSearch`. We call this function if it exists after children have been mounted.

In this test I've updated the InitializePromise synthetic test, as it was using `render` from testing-library/react, which evaluates effects, which in real server environments doesn't happen.

[CR-8629]: https://algolia.atlassian.net/browse/CR-8629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react-instantsearch-nextjs/src/__tests__/InitializePromise.test.tsx
+++ b/packages/react-instantsearch-nextjs/src/__tests__/InitializePromise.test.tsx
@@ -6,10 +6,10 @@ import {
   createSearchClient,
   createSingleSearchResponse,
 } from '@instantsearch/mocks';
-import { act, render } from '@testing-library/react';
 import * as utils from 'instantsearch.js/es/lib/utils';
 import { ServerInsertedHTMLContext } from 'next/navigation';
 import React from 'react';
+import { renderToString } from 'react-dom/server';
 import { SearchBox, TrendingItems } from 'react-instantsearch';
 import {
   InstantSearch,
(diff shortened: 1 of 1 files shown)


packages/instantsearch.js/src/lib/InstantSearch.ts
```diff
@@ -700,14 +702,25 @@ See documentation: ${createDocumentationLink({
...
       this.scheduleSearch = defer(noop);
-      // We also skip the initial network request when widgets are dynamically
-      // added in the first tick (that's the case in all the framework-based flavors).
-      // When we add a widget to `index`, it calls `scheduleSearch`. We can rely
-      // on our `defer` util to restore the original `scheduleSearch` value once
-      // widgets are added to hook back to the regular lifecycle.
-      defer(() => {
-        this.scheduleSearch = originalScheduleSearch;
-      })();
+      if (this._manuallyResetScheduleSearch) {
```
Haroenv, line 705: I think it would be nicer if this is required, but for now I didn't find a good way in Vue InstantSearch to call this, also it would be a breaking change.

aymeric-giraudet (approved):

PR #6661, opened 2025-06-24
test(nextjs): add e2e tests for app router

aymeric-giraudet:
Still part of [FX-3392](https://algolia.atlassian.net/browse/FX-3392)


packages/react-instantsearch-nextjs/__tests__/e2e/backButton.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    await link.click();
+
+    await waitForUrl('http://localhost:3000/layout');
+
+    await browser.back();
+
+    await waitForUrl(urlWithRefinement);
+
+    const checkbox = await $('.ais-RefinementList-checkbox[value="Apple"]');
+    expect(await checkbox.getAttribute('checked')).toBe('true');
```
Haroenv, line 34: check that hits have the right queryID maybe?

Haroenv (approved):
in a different iteration we probably want to split the example for tests + the example for documentation, but this makes sense for now

PR #6671, opened 2025-07-10
fix(infiniteHits): fix cache not using proper state

shaejaz:
**Summary**
Using helper state for getting a page would cause cache miss since infinite hits doesn't have access to base level widgets in a multi index scenario. Using the latest cache directly fixes this.

[CR-8905]
Closes: https://github.com/algolia/instantsearch/issues/6653

**Result**
Get previous page works correctly when preloaded via routing and multi index.

[CR-8905]: https://algolia.atlassian.net/browse/CR-8905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
+++ b/packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
@@ -316,6 +316,48 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     ).toHaveBeenCalledWith(expect.objectContaining({ page: 3 }));
   });
 
+  it('goes to previous page correctly when state is mismatched', () => {
+    const renderFn = jest.fn();
+    const makeWidget = connectInfiniteHits(renderFn);
+    const widget = makeWidget({});
+
+    const helper = algoliasearchHelper(createSearchClient(), '', {
+      page: 4,
+    });
+    helper.overrideStateWithoutTriggeringChangeEvent = jest.fn(() => helper);
+    helper.searchWithoutTriggeringOnStateChange = jest.fn();
+
+    const mainHelper = algoliasearchHelper(createSearchClient(), '', {
+      page: 4,
+      query: 'test', // arbitrary change to state to force cache mismatch
+    });
+    mainHelper.overrideStateWithoutTriggeringChangeEvent = jest.fn(
+      () => mainHelper
+    );
+    mainHelper.searchWithoutTriggeringOnStateChange = jest.fn();
+
+    widget.init(
+      createInitOptions({
+        state: mainHelper.state,
+        helper,
+      })
+    );
+
+    const { showMore, showPrevious } = widget.getWidgetRenderState(
+      createRenderOptions({
+        state: mainHelper.state,
+        helper,
+      })
+    );
+    showMore();
+    expect(helper.state.page).toBe(5);
+
+    showPrevious();
+    expect(
+      helper.overrideStateWithoutTriggeringChangeEvent
+    ).toHaveBeenCalledWith(expect.objectContaining({ page: 3 }));
+  });
+
   it('Provides the hits and flush hists cache on query changes', () => {
     const renderFn = jest.fn();
     const makeWidget = connectInfiniteHits(renderFn);
--- a/packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
+++ b/packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
@@ -327,25 +327,23 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     helper.overrideStateWithoutTriggeringChangeEvent = jest.fn(() => helper);
     helper.searchWithoutTriggeringOnStateChange = jest.fn();
 
-    const mainHelper = algoliasearchHelper(createSearchClient(), '', {
+    // we add a query to the parent helper to simulate multi-index where
+    // the parent has a query and the infinite hits index does not
+    const parentHelper = algoliasearchHelper(createSearchClient(), '', {
       page: 4,
-      query: 'test', // arbitrary change to state to force cache mismatch
+      query: 'test',
     });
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -261,32 +261,31 @@ export default (function connectInfiniteHits<
     };
 
     const getShowPrevious =
-      (helper: Helper): (() => void) =>
+      (
+        helper: Helper,
+        getCachedHitsFn: () => InfiniteHitsCachedHits<THit>
```
Haroenv:
```suggestion
        getCachedHits: () => InfiniteHitsCachedHits<THit>
```

`get` already implies it's a function

packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -316,6 +316,48 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
     ).toHaveBeenCalledWith(expect.objectContaining({ page: 3 }));
   });
 
+  it('goes to previous page correctly when state is mismatched', () => {
```
shaejaz, line 319: A mix of formats is being used in the `it` test names so went with this as a starting point. Probably can be better than this tho, so open to suggestions.

Haroenv: we use `it` and `test` interchangeably depending on what reads the best. Seems correct to me here.

packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -316,6 +316,48 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
+    const renderFn = jest.fn();
+    const makeWidget = connectInfiniteHits(renderFn);
+    const widget = makeWidget({});
+
+    const helper = algoliasearchHelper(createSearchClient(), '', {
+      page: 4,
+    });
+    helper.overrideStateWithoutTriggeringChangeEvent = jest.fn(() => helper);
+    helper.searchWithoutTriggeringOnStateChange = jest.fn();
```
Haroenv, line 328: could be jest.spyOn rather than overriding the functions too

packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -316,6 +316,48 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
+
+    const mainHelper = algoliasearchHelper(createSearchClient(), '', {
+      page: 4,
+      query: 'test', // arbitrary change to state to force cache mismatch
+    });
+    mainHelper.overrideStateWithoutTriggeringChangeEvent = jest.fn(
+      () => mainHelper
+    );
+    mainHelper.searchWithoutTriggeringOnStateChange = jest.fn();
```
Haroenv: and here you could reset, as it seems to be set twice

packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
```diff
@@ -316,6 +316,48 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/infinite-hi
...
+
+    const helper = algoliasearchHelper(createSearchClient(), '', {
+      page: 4,
+    });
+    helper.overrideStateWithoutTriggeringChangeEvent = jest.fn(() => helper);
+    helper.searchWithoutTriggeringOnStateChange = jest.fn();
+
+    const mainHelper = algoliasearchHelper(createSearchClient(), '', {
+      page: 4,
+      query: 'test', // arbitrary change to state to force cache mismatch
```
Haroenv: comment could be something like:

```
we add a query to the main helper to simulate multi-index where the parent has a query and the infinite hits index does not
```

packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
shaejaz: We do want to mock `searchWithoutTriggeringOnStateChange` since that is initialized in the index widget and would be undefined otherwise. I've kept the `overrideStateWithoutTriggeringChangeEvent` mock since that's how it's been mocked in the rest of the tests here, but I'm okay with using `spyOn` if that's better. I think essentially they would work the same way in this test case though.

packages/instantsearch.js/src/connectors/infinite-hits/__tests__/connectInfiniteHits-test.ts
shaejaz: Actually don't need to mock anything for the main helper since we only care about it's state. Removed these now.

packages/instantsearch.js/src/connectors/infinite-hits/connectInfiniteHits.ts
```diff
@@ -337,6 +336,11 @@ export default (function connectInfiniteHits<
         state: existingState,
         instantSearchInstance,
       }) {
+        const getCacheHits = () => {
+          const state = parent.getPreviousState() || existingState;
```
Haroenv, line 340: Isn’t this still causing us to read mismatching cache? We seem to be getting the state, not saving the cache variable

shaejaz: I think the issue was that the state was mismatching during the first pass of the `getWidgetRenderState` when `showPrevious` and `showMore` is undefined. We want those functions to dynamically get the cache from a fn ref rather than use the cache that's passed to it at the start.

I've confirmed it again, this does solve the bug. The test case also fails with the change removed.

PR #6678, opened 2025-07-23
feat(ui-components): introduce chat components

FabienMotte:
**Summary**
This PR adds chat UI components to be used in the react and js libraries.


packages/instantsearch-ui-components/src/lib/utils/__tests__/find-test.ts
shaejaz, line 1: Can't use the native utils here due to polyfills. The `find` util does exist, but it's in `instantsearch.js`.

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -68,6 +70,16 @@ export type ChatMessageActionProps = {
...
 };
 
+export type Tools = Array<{
+  type: string;
+  component: (props: {
+    message: ChatToolMessage;
+    indexUiState: object;
+    setIndexUiState: (state: object) => void;
+  }) => JSX.Element;
+  onToolCall?: (a: any) => void;
```
Haroenv: maybe put a todo here to help us remind that the type isn't set yet. Or is that because you don't want to create a circular dependency? Is UiState declared in InstantSearch.js or ui components?

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -183,14 +209,43 @@ export function createChatMessageComponent({
...
+      part: ChatMessageBase['parts'][number],
+      index: number
+    ) {
       if (part.type === 'step-start') {
         return null;
       }
       if (part.type === 'text') {
-        return <span>{part.markdown}</span>;
+        const markdown = compiler(part.text, {
+          createElement: createElement as any,
```
Haroenv, line 222:
```suggestion
          createElement: createElement,
```

does this work?

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -183,14 +209,43 @@ export function createChatMessageComponent({
...
+              />
+            </div>
+          );
+        }
       }
-      return <pre className="ais-ChatMessage-code">{JSON.stringify(part)}</pre>;
+      return (
+        <pre key={`${message.id}-${index}`} className="ais-ChatMessage-code">
+          {JSON.stringify(part)}
+        </pre>
```
Haroenv, line 250: what should our actual fallback be? decided later and json for now is probably ok

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -183,14 +209,43 @@ export function createChatMessageComponent({
...
+              key={`${message.id}-${index}`}
+              className="ais-ChatMessage-tool"
+            >
+              <ToolComponent
+                message={part as ChatToolMessage}
+                indexUiState={indexUiState}
+                setIndexUiState={setIndexUiState}
+              />
+            </div>
+          );
```
Haroenv, line 242: warn in else

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -183,14 +209,43 @@ export function createChatMessageComponent({
 
     const DefaultActionIcon = createDefaultActionIconComponent;
 
-    function renderAssistantPart(part: ChatMessageBase['parts'][number]) {
+    function renderAssistantPart(
```
Haroenv:
```suggestion
    function renderMessagePart(
```

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
shaejaz: Yeah that was the issue here, they are defined in InstantSearch.js and the types also depend on the connector props so it's not really straightforward to move them I think. I guess this is okay with a todo to handle it later.

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
shaejaz: No doesn't work, `createElement` from markdown lib is not using the same type from react.

Haroenv: that's ok, we can probably fork it later for preact types too then

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
Haroenv: can we define a really broad similar type? something like RecordWithObjectID and the arguments we pass to onToolCall approximately?

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
shaejaz: Done! Had to add a `warn` util based on the ones used in the other packages.

bundlesize.config.json
```diff
@@ -22,7 +22,7 @@
     },
     {
       "path": "packages/react-instantsearch/dist/umd/ReactInstantSearch.min.js",
-      "maxSize": "66.50 kB"
+      "maxSize": "71 kB"
```
Haroenv: is this actually still needed now? no right?

Haroenv (approved):

PR #6681, opened 2025-08-07
feat(index): introduce "isolated" option

Haroenv:
An isolated index does not get requested when changes happen in the parent and also does not inherit its search parameters.

If you want an isolated index to search, you need to cause a search (eg by refining a child widget)

isolated indices:
- do not automatically search when mounted
- do not show up in ui state
- do not interact with  onStateChange
- do not get requested in server-side rendering
- do not cause searches in parent indices

This option is EXPERIMENTAL, and implementation details may change in the future. Things that could change are:
- which widgets get rendered when a change happens
- whether the index searches automatically
- whether the index is included in the URL / UiState
- whether the index is include in server-side rendering

Usage:

```js
search.addWidgets([
  index({ isolated: true, indexName: 'alone' })
]);

search.addWidgets([
  index({ isolated: false }).addWidgets([
    index({ indexName: 'isolated' })
  ]),
]);
```

[FX-3448]

supersedes the following POCs:
- closes #6880
- closes #5939

### DIFF
--- a/packages/instantsearch.js/src/widgets/index/index.ts
+++ b/packages/instantsearch.js/src/widgets/index/index.ts
@@ -18,10 +18,10 @@ import type {
   IndexUiState,
   Widget,
   ScopedResult,
-  SearchClient,
   IndexRenderState,
   RenderOptions,
   RecommendResponse,
+  SearchClient,
 } from '../../types';
 import type {
   AlgoliaSearchHelper as Helper,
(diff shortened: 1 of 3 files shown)


packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -37,10 +37,50 @@ const withUsage = createDocumentationMessageGenerator({
...
+       * This option is EXPERIMENTAL, and implementation details may change in the future.
+       * Things that could change are:
+       * - which widgets get rendered when a change happens
+       * - whether the index searches automatically
+       * - whether the index is included in the URL / UiState
+       * - whether the index is included in server-side rendering
+       *
+       * @default false
+       */
+      isolated: true;
```
dhayab: Should we explicitly prefix it with `EXPERIMENTAL_` (preferred), or tie its behavior to a `future` flag?

shaejaz (approved):

packages/instantsearch.js/src/widgets/index/index.ts
Haroenv: good call, let's call the option experimental for now

PR #6688, opened 2025-08-21
FIX : in come cases, this.nodes[0] is undefined

benjamin-volle:
**Summary**

"Playing" with facets can result to this error : 

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'nodeValue')
    at a.value (algolia-instantsearch.min.js:2:229597)

    at e (algolia-instantsearch.min.js:2:98166)

    at e (algolia-instantsearch.min.js:2:98280)

    at e (algolia-instantsearch.min.js:2:98280)

    at $n (algolia-instantsearch.min.js:2:98366)

    at algolia-instantsearch.min.js:2:102651

    at Kn (algolia-instantsearch.min.js:2:103009)

    at $n (algolia-instantsearch.min.js:2:97461)

    at Kn (algolia-instantsearch.min.js:2:101481)

    at $n (algolia-instantsearch.min.js:2:97461)

```

**Result**

Error is gone :)

### DIFF
--- a/packages/instantsearch.js/src/components/Template/Template.tsx
+++ b/packages/instantsearch.js/src/components/Template/Template.tsx
@@ -33,7 +33,7 @@ class RawHtml extends Component<{ content: string }> {
     });
     // if there is one TextNode first and one TextNode last, the
     // last one's nodeValue will be assigned to the first.
-    if (this.nodes[0].nodeValue) {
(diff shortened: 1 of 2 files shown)


Haroenv:
Thanks that looks great, do you mind also adding a test that covers this use case so it doesn't regress?

PR #6693, opened 2025-09-01
chore: Fix #6692 update hogan.js to remove deprecated package mkdirp

disastrous-charly:
**Summary**

Hello, 
Current instantsearch.js embed hogan.js as dependency which used mkdirp, which triggers a deprecated warning.
It's a small change but my team and I would like to avoid install deprecated packages in production. 

Therefore I have forked hogan.js to use native nodejs and remove the mkdirp packages. 
All tests are running correctly for hogan.js.

Let me know if that can be of any help


Haroenv:
Thanks for your pull request, but we will not depend on a production dependency that isn't well downloaded or by ourselves. You can of course use something like yarn/pnpm resolutions to use your fork as a dependency.

PR #6698, opened 2025-09-05
feat(react-instantsearch): introduce chat widget

shaejaz:
**Summary**
Adds a `Chat` react widget and a `useChat` hook.

TODO:
- Better tests


packages/react-instantsearch-core/src/lib/useChat.ts
```diff
@@ -0,0 +1,196 @@
...
+      return new DefaultChatTransport({
+        api: `https://${appId}.algolia.net/agent-studio/1/agents/${agentId}/completions?compatibilityMode=ai-sdk-5`,
+        headers: {
+          'x-algolia-application-id': appId,
+          'X-Algolia-API-Key': apiKey,
+        },
+      });
+    }
+
+    if ('agentId' in options && 'transport' in options) {
```
Haroenv: this warning will never fire, as it's already returned in the transport branch

packages/instantsearch.js/package.json
```diff
@@ -40,7 +40,9 @@
     "search-insights": "^2.17.2"
   },
   "peerDependencies": {
-    "algoliasearch": ">= 3.1 < 6"
+    "algoliasearch": ">= 3.1 < 6",
+    "zod": "^3.25.76 || ^4",
```
Haroenv, line 44: (can be later), but I think this should be a dependency with this range, not a peer dependency. I think Vercel set it up wrong to try to prevent having two zods in an app

packages/react-instantsearch-core/src/connectors/useChat.ts
```diff
@@ -0,0 +1,197 @@
...
+      if (!appId || !apiKey) {
+        throw new Error(
+          'The `useChat` hook requires an `appId` and `apiKey` to be set on the `InstantSearch` component when using the `agentId` option.'
+        );
+      }
+      return new DefaultChatTransport({
+        api: `https://${appId}.algolia.net/agent-studio/1/agents/${agentId}/completions?compatibilityMode=ai-sdk-5`,
+        headers: {
+          'x-algolia-application-id': appId,
+          'X-Algolia-API-Key': apiKey,
```
Haroenv: is this casing correct?

packages/react-instantsearch-core/src/connectors/useChat.ts
```diff
@@ -0,0 +1,197 @@
...
+  resume = false,
+  ...options
+}: UseChatOptions<TUiMessage> = {}): UseChatHelpers<TUiMessage> {
+  warn(false, 'Chat is not yet stable and will change in the future.');
+  if ('agentId' in options && 'transport' in options) {
+    warn(
+      false,
+      "`useChat` with `agentId` and `transport` can't be used together. The `transport` option will be used."
+    );
+  }
```
Haroenv:
```suggestion
  warn(
    !('agentId' in options && 'transport' in options),
    "`useChat` with `agentId` and `transport` can't be used together. The `transport` option will be used."
  );
```

packages/react-instantsearch/src/widgets/Chat.tsx
```diff
@@ -0,0 +1,175 @@
...
+                }}
+              >
+                Refine on this query
+              </button>
+            )}
+          </div>
+        );
+      },
+      onToolCall: ({ toolCall, addToolResult }) => {
+        addToolResult({
```
Haroenv, line 69: still required now, but indeed can be solved in ticket about tools

tests/utils/setupTests.ts
```diff
@@ -12,12 +12,15 @@ import {
...
 
 import * as matchers from './matchers';
+import { mockAiSdk } from './mockAiSdk';
 
 Enzyme.configure({ adapter: new Adapter() });
 expect.addSnapshotSerializer(createSerializer({ mode: 'deep' }) as any);
 expect.addSnapshotSerializer(htmlSerializer());
 expect.extend(matchers);
 
+mockAiSdk();
```
Haroenv, line 22: instead of always mocking, can you add a minimal TransformStream to globals here so ai doesn't complain about the global being missing?

packages/instantsearch.js/package.json
shaejaz: alternatively, remove this peer dependency since this is only require for chat usage in IS. We're not exporting these widgets/libs atm so it doesn't matter and this can be addressed later when the umd build is fixed?

Haroenv: we're adding JS pretty soon so no need to remove it here though

shaejaz: Sounds good, would handle it in the JS ticket in that case

yarn.lock
```diff
@@ -33269,11 +33269,21 @@ zip-stream@^2.1.2:
...
 
+zod-to-json-schema@3.24.6:
+  version "3.24.6"
+  resolved "https://registry.yarnpkg.com/zod-to-json-schema/-/zod-to-json-schema-3.24.6.tgz#5920f020c4d2647edfbb954fa036082b92c9e12d"
+  integrity sha512-h/z3PKvcTcTetyjl1fkj79MHNEjm+HpD6NXheWjzOekY7kV+lwDYnHw+ivHkijnCSMz1yJaWBD9vu/Fcmk+vEg==
+
 zod@^3.22.4:
   version "3.22.4"
   resolved "https://registry.yarnpkg.com/zod/-/zod-3.22.4.tgz#f31c3a9386f61b1f228af56faa9255e845cf3fff"
   integrity sha512-iC+8Io04lddc+mVqQ9AZ7OQ2MrUKGN+oIQyq1vemgt46jwCwLfhq7/pwnBnNXXXZb8VTVLKwp9EDkx+ryxIWmg==
```
Haroenv: you can likely run `npc yarn-deduplicate` and put this in the same as the zod v4 case

tests/utils/mockAiSdk.ts
```diff
@@ -0,0 +1,19 @@
...
+        disconnect: jest.fn(),
+      })),
+    };
+  });
+
+  jest.mock('instantsearch.js/es/lib/chat', () => {
+    return {
+      Chat: jest.fn().mockImplementation(() => ({})),
+    };
+  });
```
Haroenv, line 18: I think we should do this conditionally, only when needed.

packages/react-instantsearch/src/widgets/__tests__/Chat.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+jest.mock('react-instantsearch-core', () => {
+  const originalModule = jest.requireActual('react-instantsearch-core');
+  return {
+    ...originalModule,
+    useChat: jest.fn(() => mockUseChat),
+  };
+});
+
+describe('Chat', () => {
+  test('opens chat when toggle button is clicked', () => {
```
Haroenv, line 39: can we add a test of:
- closing on toggle button
- closing on close button
- setting a prompt and committing

also could these maybe be put in common test suites? is that possible? (skipped for js and vue of course)

packages/react-instantsearch/src/widgets/Chat.tsx
```diff
@@ -0,0 +1,175 @@
...
+  resume,
+  ...options
+}: ChatProps<TObject>) {
+  const { indexUiState, setIndexUiState } = useInstantSearch();
+
+  const [open, setOpen] = React.useState(false);
+  const [input, setInput] = React.useState('');
+
+  const tools = React.useMemo(() => {
+    return [...createDefaultTools(itemComponent), ...(userTools ?? [])];
```
Haroenv, line 137: tools is a separate ticket, but this way you can't override a default tool to change something about it.

packages/react-instantsearch/rollup.config.js
```diff
@@ -46,7 +46,7 @@ const plugins = [
 
 const createConfiguration = ({ name, minify = false } = {}) => ({
   input: 'src/index.ts',
-  external: ['react'],
+  external: ['react', 'ai'],
```
Haroenv, line 49: can be done later, but I think we should disable the whole chat component in UMD instead of just making the ai lib external.

Haroenv (commented):
🔥

yarn.lock
shaejaz: deduplicate doesn't do anything unfortunately, which makes sense since astro from specs-website requires zod `v3.22.4` specifically. ai-sdk needs zod `v3.25.76` minimum.

Haroenv: ah right because it resolves to v4

packages/react-instantsearch/src/widgets/__tests__/Chat.test.tsx
shaejaz: Added more tests that you mentioned.

Can't do the closing on toggle button since the toggle button is replaced when the chat is opened. I think we can add it once the design changes are merged.

I also took a look at the shared tests. From what I can tell they rely on widget definitions from IS.js. Should those be handled with the js changes in that case?

packages/react-instantsearch/src/widgets/Chat.tsx
shaejaz: Done!

tests/utils/mockAiSdk.ts
shaejaz: Not having this mock also results in undefined errors in any test file that imports something from instantsearch.js. Do you have something else in mind to have it conditionally mocked?

packages/react-instantsearch/src/widgets/Chat.tsx
```diff
@@ -126,6 +130,10 @@ export function Chat<TObject extends RecordWithObjectID>({
   const [input, setInput] = React.useState('');
 
   const tools = React.useMemo(() => {
+    if (userTools?.some((tool) => tool.type === SearchIndexToolType)) {
+      return userTools;
+    }
```
Haroenv, line 135: this seems fine, but you could filter the default tools by the types that are in user tools, or if it would be an object instead of an array, we could loop over the array, use the tool name as the key, and the user tool would override the default tool. 

Either way this already works so that's fine.

shaejaz: Ah yes you're right. Will fix this is the next tools PR that possibly handles more default tools

packages/react-instantsearch/src/widgets/Chat.tsx
```diff
@@ -0,0 +1,183 @@
...
+      }}
+      promptProps={{
+        value: input,
+        onInput: (event) => {
+          setInput(event);
+        },
+        onSubmit: (event) => {
+          sendMessage({ text: event });
+          setInput('');
+        },
```
FabienMotte, line 177: Should we pass events to `onInput` and `onSubmit`? Currently, only the string value is passed as an parameter
I already fixed it on my PR: https://github.com/algolia/instantsearch/pull/6704/files#diff-a4b4f19ec0ebca80276c69a75834d9570191668718149795eb016789740a4abbR118

FabienMotte (commented):
Awesome! 👏

packages/react-instantsearch/src/widgets/Chat.tsx
Haroenv: what would be the use case, when you don't want to preventDefault?

FabienMotte: Let's decide about what to do on my branch

FabienMotte (approved):

PR #6702, opened 2025-09-09
feat(autocomplete): draft React component

aymeric-giraudet:
**Summary**

[]()

**Result**

Still very much a draft, not too sure about the divide with `react-instantsearch` and core, and whether indices will use `useIndex` hooks or more custom stuff. I think it's good for now until we implement indices.
Also conditional exports doesn't work 100% yet, we'd need to change `moduleResolution`

### DIFF
--- a/packages/react-instantsearch/src/components/Autocomplete.tsx
+++ b/packages/react-instantsearch/src/components/Autocomplete.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import {
+  AutocompleteWrapper,
+  useAutocomplete,
+} from 'react-instantsearch-core/dist/es/autocomplete';
+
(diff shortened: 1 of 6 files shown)


packages/react-instantsearch/src/components/Autocomplete.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import {
+  AutocompleteWrapper,
+  useAutocomplete,
+} from 'react-instantsearch-core/dist/es/autocomplete';
```
Haroenv: I wonder we can avoid this by exporting it as EXPERIMENTAL_useAutocomplete etc, what do you think?

shaejaz (approved):

PR #6703, opened 2025-09-11
feat(autocomplete): indices prop for React component

aymeric-giraudet:
**Summary**

[FX-3493](https://algolia.atlassian.net/browse/FX-3493)

**Result**

Moved everything under `react-instantsearch`, implemented indices option (they still have to type their components), and some simple test

[FX-3493]: https://algolia.atlassian.net/browse/FX-3493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react-instantsearch/src/components/Autocomplete.tsx
```diff
@@ -1,19 +1,30 @@
...
-    </AutocompleteWrapper>
-  );
-}
+import { Hits, SearchBox } from '../widgets';
+
+import type { BaseHit, Hit } from 'instantsearch.js';
 
-function AutocompleteInner() {
-  useAutocomplete();
+type IndexConfig<TItem extends Hit<BaseHit> = Hit<BaseHit>> = {
```
Haroenv, line 8: so if you'd have multiple different types of hits, this would be a union and you assert within the template?

packages/react-instantsearch/src/components/Autocomplete.tsx
```diff
@@ -1,19 +1,30 @@
...
+  indices: IndexConfig[];
+};
+
+export function EXPERIMENTAL_Autocomplete({ indices }: AutocompleteProps) {
+  return (
+    <Index EXPERIMENTAL_isolated>
+      <SearchBox />
+      {indices.map((index) => (
+        <Index key={index.indexName} indexName={index.indexName}>
+          <Hits hitComponent={({ hit }) => <index.itemComponent {...hit} />} />
```
Haroenv, line 25: I didn't consider using hit instead of connectautocomplete. I guess pros are: no need to have a separate connector for showing the hits, cons would be counting the items and making sure they're in a single wrapper div may be a little more work (as we'd need something else than Hits for that.

For this PR this works indeed.

packages/react-instantsearch/src/components/Autocomplete.tsx
aymeric-giraudet: It could be possible but they would have to declare indices somewhere else than inline, I don't think anyone would really care tbh, and we'd probably need some helper functions

packages/react-instantsearch/src/components/Autocomplete.tsx
aymeric-giraudet: Yeah just a placeholder for now, we'll probably just use hooks directly

PR #6704, opened 2025-09-11
feat(react-instantsearch-ui): implement chat components design

FabienMotte:
**Summary**

This PR implements the design for the chat components in InstantSearch.

The changes include:
- Added common chat UI components (`Chat`, `ChatHeader`, `ChatMessage`, `ChatMessages`, `ChatPrompt`, `ChatToggleButton`)
- Added new specialized components for error states and loading indicators (`ChatMessageError`, `ChatMessageLoader`)
- Restructured InstantSearch CSS architecture with modular SCSS files organized by component
- Responsive design / dark and light modes
- CSS custom variables for consistent theming and easy customization
- Added icon set for chat interface elements
- Improved TypeScript definitions and component APIs
- Added utility hooks like `useStickToBottom` (from https://github.com/stackblitz-labs/use-stick-to-bottom) for better chat UX
- Updated bundle size configurations to fit the new components/styles

**Result**

image

image

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/Chat.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/Chat.tsx
@@ -18,7 +18,7 @@ export type ChatClassNames = {
   container?: string | string[];
 };
 
-export type ChatProps = ComponentProps<'div'> & {
+export type ChatProps = Omit<ComponentProps<'div'>, 'onError'> & {
   /*
    * Whether the chat is open or closed.
    */


babel.config.js
```diff
@@ -85,6 +85,10 @@ module.exports = (api) => {
             // this is actually only called on arrays
             'String.prototype.includes',
 
+            // widely supported, safe to use without polyfill
+            'String.prototype.startsWith',
+            'Array.prototype.find',
```
Haroenv: not really the case, but as we can't special case this configuration only for the chat, there might be no way around it 🤔

examples/react/e-commerce-vite/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "example-react-instantsearch-e-commerce-vite",
```
Haroenv: maybe call this "with chat", then we can remove the tweak pane later and make it the default example for react e commerce?

examples/react/e-commerce-vite/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "use-stick-to-bottom": "^1.1.1",
+    "tweakpane": "^4.0.5",
+    "@tweakpane/core": "^2.0.5"
+  },
+  "devDependencies": {
+    "@types/react": "19.0.3",
+    "@types/react-dom": "19.0.3",
+    "@vitejs/plugin-react": "^4.0.0",
+    "typescript": "5.5.2",
+    "vite": "^4.0.0"
```
Haroenv: if I remember it correctly this introduces a second copy of vite/esbuild as we elsewhere depend on a different version. I guess as it's temporary, it's ok?

packages/react-instantsearch/src/widgets/Chat/Chat.tsx
```diff
@@ -3,14 +3,110 @@ import { createChatComponent } from 'instantsearch-ui-components';
...
+import type {
+  ChatMessageBase,
+  ChatMessagesProps,
+  Pragma,
+  Tools,
+} from 'instantsearch-ui-components';
+
 const ChatUiComponent = createChatComponent({
-  createElement: createElement as Pragma,
+  createElement: createElement as unknown as Pragma,
```
Haroenv: ok to leave as-is, but I believe this isn't needed in ci

packages/react-instantsearch/src/widgets/Chat/Chat.tsx
```diff
@@ -3,14 +3,110 @@ import { createChatComponent } from 'instantsearch-ui-components';
...
   Fragment,
 });
 
+const AssistantLeadingComponent = () => (
```
Haroenv: why are these part of the widget instead of the component?

babel.config.js
FabienMotte: Should we use polyfills for these?

packages/react-instantsearch/src/widgets/Chat/Chat.tsx
FabienMotte: I also noticed another TS error related to `Type 'React.JSX.Element' is not assignable to type 'JSX.Element'`
Are we good leaving these if the CI passes?

examples/react/e-commerce-vite/package.json
FabienMotte: Good catch, let me align it to `"vite": "5.0.7"`

babel.config.js
Haroenv: as we're using fetch in the chat component already, it's ok to leave as-is.

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -228,9 +202,9 @@ export function createChatMessageComponent({
         });
         return <span key={`${message.id}-${index}`}>{markdown}</span>;
       }
-      if (part.type === 'tool-start') {
-        part;
-      }
+      // if (part.type === 'tool-start') {
+      //   part;
+      // }
```
Haroenv: remove this?
```suggestion
```

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -256,32 +230,28 @@ export function createChatMessageComponent({
       );
     }
 
-    const Actions = () => {
+    const Actions = useMemo(() => {
```
Haroenv: does it work fine if it's not memoised too? we can't use react's useMemo in a ui component.

FabienMotte: Removing the `useMemo` in `ChatMessage` seems fine, but the others in `ChatMessages` cause all the messages to re-render when you scroll

FabienMotte: I managed to fix it by moving these outside of the component:

```ts
export function createChatMessagesComponent({
  createElement,
  Fragment,
}: Renderer) {
  const DefaultMessageComponent =
    createDefaultMessageComponent<ChatMessageBase>({ createElement, Fragment });
  const DefaultLoaderComponent = createChatMessageLoaderComponent({
    createElement,
  });
  const DefaultErrorComponent = createChatMessageErrorComponent({
    createElement,
  });

  return function ChatMessages
  /* ... */
}
```

packages/instantsearch-ui-components/src/components/chat/ChatMessageLoader.tsx
```diff
@@ -0,0 +1,62 @@
...
+            <div className="ais-ChatMessage-message">
+              {translations.loaderText && (
+                <div className="ais-ChatMessage-loaderText">
+                  {translations.loaderText}
+                </div>
+              )}
+              <div className="ais-ChatMessage-loaderSkeletonWrapper">
+                <div className="ais-ChatMessage-loaderSkeletonItem"></div>
+                <div className="ais-ChatMessage-loaderSkeletonItem"></div>
+              </div>
```
Haroenv, line 55: this is fine, but I wonder in the future if we could use a custom styled `progress` element for this? With svg as replacement value for example?

packages/instantsearch-ui-components/src/components/chat/ChatPrompt.tsx
```diff
@@ -1,9 +1,14 @@
 /** @jsx createElement */
 
+import { useRef, useLayoutEffect, useCallback } from 'react';
```
Haroenv: these can't be used. Is there some way to hoist the hooks outside the chat component, like done for scroll?

Haroenv (commented):
good job! we'll need to remove the react-isms before this can be merged though

packages/instantsearch-ui-components/src/components/chat/ChatPrompt.tsx
FabienMotte: Done!

packages/instantsearch-ui-components/src/components/chat/ChatMessageLoader.tsx
FabienMotte: Currently, the default `ChatMessageLoader` component can be replaced by passing a `loaderComponent` in `messagesProps`, do you think that's enough?

examples/react/e-commerce-with-chat/App.css
```diff
@@ -0,0 +1,341 @@
+* {
```
Haroenv: once this example is removed from this PR, lgtm :)

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -297,27 +246,42 @@ export function createChatMessageComponent({
...
             </div>
           )}
 
           <div className={cx(cssClasses.content)}>
             <div className={cx(cssClasses.message)}>
               {message.role === 'assistant'
-                ? message.parts.map(renderMessagePart)
-                : message.parts.map(renderMessagePart)}
+                ? message.parts.map(renderAssistantPart)
+                : message.parts.map(renderAssistantPart)}
```
Haroenv: looks like this may have been a wrong rebase. I think it should be `renderMessagePart` or `renderPart`

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -297,27 +246,42 @@ export function createChatMessageComponent({
         <div className={cx(cssClasses.container)}>
           {hasLeading && (
             <div className={cx(cssClasses.leading)}>
-              {LeadingComponent ? (
-                <LeadingComponent />
-              ) : (
-                AvatarComponent && <AvatarComponent />
```
Haroenv, line 303: did the avatar get removed on purpose? we don't have it in the design yet so good either way.

packages/react-instantsearch/src/widgets/Chat.tsx
```diff
@@ -147,34 +166,64 @@ export function Chat<TObject extends RecordWithObjectID>({
...
+        status,
+        onReload: (messageId) => regenerate({ messageId }),
         messages,
         tools,
         indexUiState,
         setIndexUiState,
+        isClearing,
+        onClearTransitionEnd: handleClearTransitionEnd,
+        scrollRef: scrollRef as unknown as MutableRef<HTMLDivElement>,
+        contentRef: contentRef as unknown as MutableRef<HTMLDivElement>,
```
Haroenv: is this because react and ui components disagree what a mutable ref is? I guess it's only details but the main api is the same?

packages/react-instantsearch/src/widgets/__tests__/all-components.test.tsx
```diff
@@ -25,12 +25,12 @@ describe('rendering', () => {
...
-        <Component className="BASECLASS" title="test title" />
+        <Component classNames={{ root: 'BASECLASS' }} />
       );
 
-      expect(container.querySelector<HTMLDivElement>('.BASECLASS')!.title).toBe(
-        'test title'
-      );
+      expect(
+        container.querySelector<HTMLDivElement>('.BASECLASS')
+      ).toBeTruthy();
```
Haroenv: why aren't we testing the title anymore?

yarn.lock
```diff
@@ -222,11 +222,25 @@
     js-tokens "^4.0.0"
     picocolors "^1.0.0"
 
+"@babel/code-frame@^7.27.1":
```
Haroenv: you can probably revert the yarn.lock changes, as there are no package.json changes, right?

Haroenv (commented):
looks good!

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
FabienMotte: By the way, as a performance optimization, should we split it into two functions and render the user message as a string?

yarn.lock
FabienMotte: Good catch!

packages/react-instantsearch/src/widgets/Chat.tsx
FabienMotte: I avoided using `unknown` by doing the following:
```ts
scrollRef: { current: scrollRef.current as HTMLDivElement },
contentRef: { current: contentRef.current as HTMLDivElement },
 ```

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
FabienMotte: I think it's a leftover from the Satellite implementation where we have an `Avatar` component. However, since we don't have that component in IS, providing the `leadingComponent` prop should be sufficient to add an avatar or whatever you want before the message.

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
Haroenv: that would mean that if you're using markdown in your message, it wouldn't be rendered right? not sure what's best. Maybe we can decide later

yarn.lock
Haroenv: you reverted the revert, do you know why?

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
FabienMotte: Yes, indeed, I'm not sure if that's really necessary for the user.
Let's decide later and keep it as it is

packages/react-instantsearch/src/widgets/__tests__/all-components.test.tsx
FabienMotte: The props were not spread in the `Chat` widget and UI component. I reverted the test, it should be fine now. Thanks!

yarn.lock
FabienMotte: The `yarn.lock` file should be identical to the one on `master`

shaejaz (commented):
lgtm once CI passes! Tested it with react and js versions locally, both working well (js has an issue with the loader not showing up but that might be js specific).

PR #6705, opened 2025-09-15
feat(instantsearch.js): introduce chat widget

shaejaz:
This PR adds the chat widget for JS and scaffolds CTS.

FX-3528


Haroenv:
looks like the ai and zod packages need to be ignored for Vue InstantSearch and experiences too

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -0,0 +1,342 @@
...
+            canClear: messages.length > 0 && isClearing !== true,
+          }}
+          messagesProps={{
+            messages,
+            indexUiState,
+            isClearing,
+            onClearTransitionEnd,
+            // temporary until we have a good solution in js
+            // or move logic in ui-component
+            hideScrollToBottom: true,
```
dhayab, line 203: I tried to make it work in the widget but it was too janky. Maybe we should put the scroll logic inside of the UI component and not reimplement it twice? For the messages don't autoscroll, which is not ideal. This could be looked at in an improvement task right after merging all stacked PRs?

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -0,0 +1,342 @@
...
+      ? userTools
+      : [...createDefaultTools<THit>(itemComponent), ...userTools];
+
+    state.subscribe(rerender);
+
+    function rerender() {
+      state.init();
+
+      const [isClearing, setIsClearing] = state.use(false);
+      const [maximized, setMaximized] = state.use(false);
```
dhayab, line 178: These could potentially also be handled by the UI component later?

packages/instantsearch.js/src/widgets/chat/defaultTemplates.ts
```diff
@@ -0,0 +1,9 @@
+import { ChatTemplates } from './chat';
+
+const defaultTemplates = {
+  item(data) {
+    return JSON.stringify(data, null, 2);
```
dhayab, line 5: Should we change this? It's the default we have for hits as well, but it is especially bad loooking in the chat right now.

Haroenv: what do you suggest instead? just the objectID maybe, or the keys in the object?

Haroenv (commented):
I think this is good to go once CI is green 🪨

packages/instantsearch.js/src/widgets/chat/chat.tsx
Haroenv: yes can be done in a separate task, problem is that scroll to bottom requires a few hooks, and ui components can't currently use hooks (as react and preact have different imports). Did you try a fork of use-scroll-to-bottom that has preact hooks instead?

dhayab: I reimplemented the basics of it, a ResizeObserver and a listener on the container scroll. The lock kept jumping (which might be because rendering of the carousel is problematic I think). We can try having a Preact-compatible fork, it might handle this better.

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -28,12 +28,17 @@ import type {
...
+  RecordWithObjectID,
+  Tools,
+  VNode,
+} from 'instantsearch-ui-components';
 
 type ItemComponent<THit> = (props: {
-  item: THit;
-  onClick: () => void;
-  onAuxClick: () => void;
+  item: RecordWithObjectID<THit>;
```
Haroenv: should this not be Hit or AlgoliaHit? or do you want to explicitly make it available for recommend / item without __position too?

dhayab: Not there yet 😅, I just want to match the item with what the carousel expects to make eslint pass. Then I'll see how to make it not Recommend-specific. For some reason this issue does not appear locally, so I'm pushing incremental changes.

examples/react/e-commerce-with-chat/App.css
```diff
@@ -0,0 +1,341 @@
+* {
```
Haroenv: looks like this isn't the same as was merged in fabien's branch

dhayab: Just merged master, and for some reason this (examples/react/e-commerce-with-chat) and instantsearch.css/src/themes/chat should not be here anymore. I'll git rm those.

packages/instantsearch.js/src/widgets/chat/defaultTemplates.ts
dhayab: For now I'll leave it as is, I don't have a proper solution and all widgets it's at least in line with how other widgets render default hits.

bundlesize.config.json
```diff
@@ -10,19 +10,19 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "84.75 kB"
+      "maxSize": "100 kB"
```
Haroenv: why did this increase if we excluded chat from the umd?

dhayab: Good call. I left the bumped values that were needed when chat was part of it. Just pushed readjusted values now.

packages/algolia-experiences/rollup.config.js
```diff
@@ -56,6 +56,7 @@ const createConfiguration = ({ mode, filename }) => ({
     banner: license,
     sourcemap: true,
   },
+  external: ['ai', 'zod'],
```
Haroenv: shouldn't be needed anymore now

packages/instantsearch.js/scripts/rollup/rollup.config.js
```diff
@@ -36,6 +36,7 @@ const plugins = [
 
 const createConfiguration = ({ mode, filename }) => ({
   input: 'src/index.ts',
+  external: ['ai', 'zod'],
```
Haroenv: shouldn't be needed anymore now

packages/instantsearch.js/src/connectors/index.umd.ts
```diff
@@ -0,0 +1,65 @@
+import { deprecate } from '../lib/utils';
```
Haroenv, line 1: you have the widget test that checks umd and non-umd are the same, but I don't think for connectors

packages/instantsearch.js/src/widgets/index.umd.ts
```diff
@@ -0,0 +1,69 @@
...
+export { default as sortBy } from './sort-by/sort-by';
+export { default as stats } from './stats/stats';
+export { default as toggleRefinement } from './toggle-refinement/toggle-refinement';
+export { default as trendingItems } from './trending-items/trending-items';
+export { default as voiceSearch } from './voice-search/voice-search';
+export { default as frequentlyBoughtTogether } from './frequently-bought-together/frequently-bought-together';
+export { default as lookingSimilar } from './looking-similar/looking-similar';
+
+export const chat = () => {
+  throw new ReferenceError(
```
Haroenv: should this maybe be a regular error? not sure it matters but the use case is slightly different (this variable isn't missing)

packages/vue-instantsearch/rollup.config.js
```diff
@@ -143,7 +145,7 @@ export * from './src/instantsearch.js';`
 
   const umd = {
     input: 'src/instantsearch.umd.js',
-    external: ['vue'],
+    external: ['vue', 'ai', 'zod'],
```
Haroenv:
```suggestion
    external: ['vue'],
```

packages/algolia-experiences/rollup.config.js
Haroenv:
```suggestion
```

packages/instantsearch.js/scripts/rollup/rollup.config.js
Haroenv:
```suggestion
```

packages/vue-instantsearch/src/components/Autocomplete.vue
```diff
@@ -20,7 +20,7 @@
 </template>
 
 <script>
-import { connectAutocomplete } from 'instantsearch.js/es/connectors';
+import { connectAutocomplete } from 'instantsearch.js/es/connectors/index.umd';
```
Haroenv, line 23: hmm, pretty annoying this is needed, but no way around for now as we still want regular esm users to have access to chat

dhayab: Yes, UMD has no relation with this but it technically is helpful. Wondering if we should have a barrel file "index.noAi.ts" or something, don't know if it's better 🤷. In any case we have tests to ensure both barrel files have the same connectors / widgets, so I'm not afraid of drift.

PR #6708, opened 2025-09-17
feat(chat): add support for client side tools

shaejaz:
**Summary**

**Result**


packages/algolia-experiences/rollup.config.js
```diff
@@ -56,6 +56,7 @@ const createConfiguration = ({ mode, filename }) => ({
     banner: license,
     sourcemap: true,
   },
+  external: ['ai', 'zod'],
```
Haroenv:
```suggestion
```

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -225,6 +246,17 @@ const createRenderer = <THit extends NonNullable<object> = BaseHit>({
   };
 };
 
+export type UserClientSideToolTemplate = Partial<{
+  component: TemplateWithBindEvent<ClientSideToolComponentProps>;
+}>;
+
+export type UserClientSideToolWithTemplate = Omit<
```
Haroenv: if this is a user-accessible type, I think it should have a clearer name. Maybe just Tool?

packages/vue-instantsearch/rollup.config.js
```diff
@@ -73,6 +73,8 @@ function outputs(vueVersion) {
       'instantsearch-ui-components',
       'vue',
       'mitt',
+      'ai',
+      'zod',
```
Haroenv:
```suggestion
```

packages/vue-instantsearch/rollup.config.js
```diff
@@ -143,7 +145,7 @@ export * from './src/instantsearch.js';`
 
   const umd = {
     input: 'src/instantsearch.umd.js',
-    external: ['vue'],
+    external: ['vue', 'ai', 'zod'],
```
Haroenv:
```suggestion
    external: ['vue'],
```

packages/instantsearch.js/src/widgets/chat/chat.tsx
shaejaz: Yes makes sense. I think the same can be done for react as well? The type can be re-exported as the same name.

packages/instantsearch.js/scripts/rollup/rollup.config.js
```diff
@@ -36,6 +36,7 @@ const plugins = [
 
 const createConfiguration = ({ mode, filename }) => ({
   input: 'src/index.ts',
+  external: ['ai', 'zod'],
```
Haroenv:
```suggestion
```

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
```diff
@@ -1,7 +1,6 @@
 /**
  * @jest-environment @instantsearch/testutils/jest-environment-jsdom.ts
  */
-
```
Haroenv, line 4:
```suggestion

```

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -28,102 +29,141 @@ import type {
...
+  templates: ChatTemplates<THit>,
+  userTools?: UserClientSideToolWithTemplate[]
+) {
+  const defaults = createDefaultTools(templates);
+
+  if (!userTools) {
+    return defaults;
+  }
+
+  const userToolsMap = new Map(userTools.map((tool) => [tool.type, tool]));
```
Haroenv, line 124: this means that if you have the same tool multiple times, the last one wins. Which makes sense I think. If we'd keep this API, we should have a warning for duplicates, but we'll move towards objects.

packages/instantsearch.js/scripts/rollup/rollup.config.js
shaejaz: Should all external be removed from all packages? Some parts in IS.js still import from `ai`. The ones left apart from this would be react-core and react. Only react core has imports from `ai` in that case.

packages/instantsearch.js/src/widgets/chat/chat.tsx
shaejaz: Yes, should be addressed when those changes are implemented.

packages/instantsearch.js/scripts/rollup/rollup.config.js
Haroenv: in @dhayab's PR the setup has been changed and the UMD no longer exposes ai/chat

PR #6709, opened 2025-09-18
feat(autocomplete): implement basic shared DOM

aymeric-giraudet:
**Summary**

[FX-3496](https://algolia.atlassian.net/browse/FX-3496)

**Result**

Was not too sure how to split components effectively, for now I did : 

- `Autocomplete` : the wrapper itself, has accessibility attributes taken from the autocomplete library DOM
- `AutocompletePanel`: same DOM as for autocomplete
- `AutocompleteIndex`: meant for more generic indices, we'll have something special for query suggestions and recent searches

This is still using the internal searchbox implem, not sure if we'll want a custom one or not.

### DIFF
--- a/packages/instantsearch-ui-components/src/components/autocomplete/Autocomplete.tsx
+++ b/packages/instantsearch-ui-components/src/components/autocomplete/Autocomplete.tsx
@@ -1,0 +1,35 @@
+/** @jsx createElement */
+
+import { cx } from '../../lib/cx';
+
+import type { ComponentChildren, Renderer } from '../../types';
+
+export type AutocompleteProps = {
+  isOpen: boolean;
+  children?: ComponentChildren;
+  classNames?: Partial<AutocompleteClassNames>;
+};
+
+export type AutocompleteClassNames = {
+  /**
+   * Class names to apply to the root element
+   */
+  root: string | string[];
+};
+
+export function createAutocompleteComponent({ createElement }: Renderer) {
+  return function Autocomplete(userProps: AutocompleteProps) {
+    const { children, isOpen, classNames = {} } = userProps;
+
+    return (
+      <div
+        className={cx('ais-Autocomplete', classNames.root)}
+        role="combobox"
+        aria-expanded={isOpen}
+        aria-haspopup="listbox"
+      >
+        {children}
+      </div>
+    );
+  };
+}
--- a/packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
+++ b/packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
@@ -1,0 +1,49 @@
+/** @jsx createElement */
+
+import { cx } from '../../lib/cx';
+
+import type { Renderer } from '../../types';
+
+export type AutocompleteIndexProps<
+  T = { objectID: string } & Record<string, unknown>
+> = {
+  items: T[];
+  ItemComponent: (item: T) => JSX.Element;
+  classNames?: Partial<AutocompleteIndexClassNames>;
+};
+
+export type AutocompleteIndexClassNames = {
+  /**
+   * Class names to apply to the root element
+   **/
+  root: string | string[];
+  /**
+   * Class names to apply to the list element
+   */
+  list: string | string[];
+  /**
+   * Class names to apply to each item element
+   */
+  item: string | string[];
+};
+
+export function createAutocompleteIndexComponent({ createElement }: Renderer) {
+  return function AutocompleteIndex(userProps: AutocompleteIndexProps) {
+    const { items, ItemComponent, classNames = {} } = userProps;
+
+    return (
+      <div className={cx('ais-AutocompleteIndex', classNames.root)}>
+        <ul className={cx('ais-AutocompleteIndexList', classNames.list)}>
+          {items.map((item) => (
+            <li
+              key={item.objectID}
+              className={cx('ais-AutocompleteIndexItem', classNames.item)}
+            >
+              <ItemComponent {...item} />
+            </li>
+          ))}
+        </ul>
+      </div>
+    );
+  };
+}
--- a/packages/instantsearch-ui-components/src/components/index.ts
+++ b/packages/instantsearch-ui-components/src/components/index.ts
@@ -12,3 +12,4 @@ export * from './chat/ChatMessage';
 export * from './chat/ChatToggleButton';
 export * from './chat/ChatHeader';
 export * from './chat/types';
+export * from './autocomplete';
--- a/packages/react-instantsearch/src/components/Autocomplete.tsx
+++ b/packages/react-instantsearch/src/components/Autocomplete.tsx
@@ -1,9 +1,30 @@
-import React from 'react';
-import { Index } from 'react-instantsearch-core';
+import {
+  createAutocompleteComponent,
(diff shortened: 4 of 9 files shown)


packages/react-instantsearch/src/components/Autocomplete.tsx
```diff
@@ -17,14 +38,38 @@ export type AutocompleteProps = {
...
+};
+
+function AutocompleteIndexComponent({
+  itemComponent: ItemComponent,
+}: AutocompleteIndexProps) {
+  const { items } = useHits();
+
+  return (
+    <AutocompleteIndex
+      // @ts-expect-error - there seems to be problems with React.ComponentType and this, but it's actually correct
```
Haroenv, line 70: should we cast it instead?

packages/react-instantsearch/src/components/Autocomplete.tsx
```diff
@@ -17,14 +38,38 @@ export type AutocompleteProps = {
...
-        <Index key={index.indexName} indexName={index.indexName}>
-          <Hits hitComponent={({ hit }) => <index.itemComponent {...hit} />} />
-        </Index>
-      ))}
+      <Autocomplete isOpen={isOpen}>
+        <SearchBox onFocus={() => setIsOpen(true)} />
+        <AutocompletePanel isOpen={isOpen}>
+          {indices.map((index) => (
+            <Index key={index.indexName} indexName={index.indexName}>
+              <AutocompleteIndexComponent itemComponent={index.itemComponent} />
```
Haroenv, line 50: ok for now, but later once we do keyboard, we likely will want it to be a flat list of indices and use connectAutocomplete? not sure tbh but it likely doesn't change much

packages/instantsearch-ui-components/src/components/index.ts
```diff
@@ -12,3 +12,4 @@ export * from './chat/ChatMessage';
 export * from './chat/ChatToggleButton';
 export * from './chat/ChatHeader';
 export * from './chat/types';
+export * from './autocomplete';
```
Haroenv: not consistent between chat (no barrel file) and autocomplete (barrel file). I don't have a specific preference, but we'll want to consolidate after both sprints are over.

packages/react-instantsearch/src/components/__tests__/Autocomplete.test.tsx
```diff
@@ -52,8 +52,152 @@ describe('Autocomplete', () => {
       </InstantSearchTestWrapper>
     );
 
-    expect(await screen.findByText('Item 1')).toBeInTheDocument();
-    expect(await screen.findByText('hello')).toBeInTheDocument();
-    expect(screen.getByRole('searchbox')).toBeInTheDocument();
+    await screen.findByText('Item 1');
+
+    expect(container.firstChild).toMatchInlineSnapshot(`
```
Haroenv: I think a different snapshot for the search box and a separate one for the panel may make sense?

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
```diff
@@ -0,0 +1,49 @@
...
+  item: string | string[];
+};
+
+export function createAutocompleteIndexComponent({ createElement }: Renderer) {
+  return function AutocompleteIndex(userProps: AutocompleteIndexProps) {
+    const { items, ItemComponent, classNames = {} } = userProps;
+
+    return (
+      <div className={cx('ais-AutocompleteIndex', classNames.root)}>
+        <ul className={cx('ais-AutocompleteIndexList', classNames.list)}>
```
Haroenv: in hits we use ol, so we should here too.

Haroenv: I think there probably (later) should be only one list, as you're navigating across all lists with the keyboard. However there's an argument to be made for separate lists for each index, as that allows us to have header/footer per index.

packages/instantsearch-ui-components/src/components/autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,35 @@
...
+    const { children, isOpen, classNames = {} } = userProps;
+
+    return (
+      <div
+        className={cx('ais-Autocomplete', classNames.root)}
+        role="combobox"
+        aria-expanded={isOpen}
+        aria-haspopup="listbox"
+      >
+        {children}
```
Haroenv, line 31: in a later iteration (when we do the final dom) we likely want to have the search box be part of this too, and then all the intermediate components also be part of the autocomplete. For now this is a logical split.

Haroenv (commented):
great job!

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
aymeric-giraudet: Autocomplete does not put everything into a single list either

packages/instantsearch-ui-components/src/components/index.ts
aymeric-giraudet: We do barrel files in other libraries, think this is more consistent

packages/react-instantsearch/src/components/Autocomplete.tsx
aymeric-giraudet: Doesn't work, but it's sensibly the same as adding `ts-expect-error` (except if the line had more statements than this)

packages/instantsearch-ui-components/src/components/index.ts
shaejaz: +1 I think the chat components should also be in a barrel file

packages/react-instantsearch/src/components/Autocomplete.tsx
shaejaz: Hm something similar does work for the Chat UI components. Maybe a plain `(props) => JSX.Element` type would work for the `ItemComponent`?

packages/react-instantsearch/src/components/Autocomplete.tsx
```diff
@@ -17,14 +38,38 @@ export type AutocompleteProps = {
...
   return (
     <Index EXPERIMENTAL_isolated>
-      <SearchBox />
-      {indices.map((index) => (
-        <Index key={index.indexName} indexName={index.indexName}>
-          <Hits hitComponent={({ hit }) => <index.itemComponent {...hit} />} />
-        </Index>
-      ))}
+      <Autocomplete isOpen={isOpen}>
+        <SearchBox onFocus={() => setIsOpen(true)} />
```
shaejaz, line 46: If going with the base SearchBox widget for now, should we forward props to it? Can be done later tho.

packages/instantsearch-ui-components/src/components/autocomplete/Autocomplete.tsx
shaejaz: From the wai-aria examples, I see that they use a button to indicate expanded state too. Should that change come with search box addition then?

Haroenv: yes, there's a separate ticket for keyboard handling

Haroenv (approved):

PR #6710, opened 2025-09-22
feat(chat): save messages in session storage

aymeric-giraudet:
**Summary**

[FX-3521](https://algolia.atlassian.net/browse/FX-3521)

**Result**

Only implemented message persistence for now, we could also do the same for whether the panel should be open or not later, preferably once we mutualise logic between vanilla and react.

It saves messages only if the status is `ready`

[FX-3521]: https://algolia.atlassian.net/browse/FX-3521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dhayab (approved):

Haroenv (approved):
separate PR for reading the saved messages?

PR #6714, opened 2025-09-24
fix(umd): remove ai components from react umd bundle

dhayab:
**Summary**

This PR updates react-instantsearch to not export `useChat` and `Chat` on the UMD bundle, since we're not bundling ai. There are also changes to the UMD test file in order to add more in-depths assertions and additionally cover react-instantsearch-core + vue-instantsearch (vue3).

FX-3530

### DIFF
--- a/packages/react-instantsearch-core/src/index.umd.ts
+++ b/packages/react-instantsearch-core/src/index.umd.ts
@@ -1,0 +1,50 @@
+export { default as version } from './version';
+export * from './components/Configure';
+export * from './components/DynamicWidgets';
+export * from './components/Index';
+export * from './components/InstantSearch';
+export * from './components/InstantSearchServerContext';
+export * from './components/InstantSearchSSRProvider';
+export * from './connectors/useBreadcrumb';
+
+export const useChat = () => {
+  throw new Error(
+    `"useChat()" is not available from the UMD build.
+
+Please use React InstantSearch with a packaging system:
+https://www.algolia.com/doc/guides/building-search-ui/installation/react/#install-react-instantsearch-as-an-npm-package`
+  );
+};
+
+export * from './connectors/useClearRefinements';
+export * from './connectors/useConfigure';
+export * from './connectors/useCurrentRefinements';
+export * from './connectors/useDynamicWidgets';
+export * from './connectors/useFrequentlyBoughtTogether';
+export * from './connectors/useGeoSearch';
+export * from './connectors/useHierarchicalMenu';
+export * from './connectors/useHits';
+export * from './connectors/useHitsPerPage';
+export * from './connectors/useInfiniteHits';
+export * from './connectors/useMenu';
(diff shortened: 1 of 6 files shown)


packages/react-instantsearch/rollup.config.js
```diff
@@ -45,7 +45,7 @@ const plugins = [
 ];
 
 const createConfiguration = ({ name, minify = false } = {}) => ({
-  input: 'src/index.ts',
+  input: 'src/index.umd.ts',
   external: ['react', 'ai'],
```
dhayab, line 49: I left `'ai'` as external here, otherwise we would have to update all hook imports in widgets, which is not worth it. Marking `'ai'` as external has no impact on the UMD bundle.

Haroenv: What change would be needed if ai is not marked as external? and how come this doesn't throw on usage without ai?

dhayab: You would need to update all hook imports in all widgets, ie:

```diff
-import { useBreadcrumb } from 'react-instantsearch-core';
+import { useBreadcrumb } from 'react-instantsearch-core/es/index.umd';
```

Rollup fails otherwise because, even if it will tree shake the imports, it will still traverse and read them. Because `useChat` is available from 'react-instantsearch-core', it will reach zod which will fail the process.

It doesn't fail on usage because no export of `ai` is actually referenced in the bundle.

packages/react-instantsearch-core/src/index.umd.ts
```diff
@@ -0,0 +1,50 @@
+export { default as version } from './version';
```
Haroenv, line 1: we should have a test to ensure umd + regular index are the same

packages/react-instantsearch/src/widgets/index.umd.ts
```diff
@@ -0,0 +1,32 @@
+export * from './Breadcrumb';
```
Haroenv, line 1: same question

packages/react-instantsearch-core/src/index.umd.ts
dhayab: Ah yes, definitely. On it.

packages/react-instantsearch/src/widgets/index.umd.ts
```diff
@@ -0,0 +1,34 @@
...
+export const Chat = () => {
+  throw new Error(
+    `"<Chat>" is not available from the UMD build.
+
+Please use React InstantSearch with a packaging system:
+https://www.algolia.com/doc/guides/building-search-ui/installation/react/#install-react-instantsearch-as-an-npm-package`
+  );
+};
+export const SearchIndexToolType = undefined;
+export const createDefaultTools = () => {};
```
dhayab, line 12: These should probably move to ui-components or js/lib/chat in the future.

PR #6716, opened 2025-09-24
fix(recommend): omit fallbackParameters from query when not passed as params

raed667:
**Summary**

`fallbackParameters` shouldn't be passed to the query unless the user specifically provides a value. Otherwise this could trigger an unexpected fallback query.


raed667:
@Haroenv can we merge this PR now ?

PR #6717, opened 2025-09-24
test(environment): expose TransferStream and make test environments consistent

shaejaz:
**Summary**
This PR adds a new custom jest environment where some missing APIs for the Chat widgets are added.

Changes:
- Replaced all jsdom usage to the new custom env.
- Removed the jsdom-global package as its functionality has been integrated into the new custom env.
- Added a new custom node env and set that to the default jest env.

### DIFF
--- a/packages/instantsearch.js/src/lib/__tests__/server-composition.test.ts
+++ b/packages/instantsearch.js/src/lib/__tests__/server-composition.test.ts
@@ -1,7 +1,3 @@
-/**
- * @jest-environment @instantsearch/testutils/jest-environment-node.ts
- */
-
 import {
   createCompositionClient,
   createControlledCompositionClient,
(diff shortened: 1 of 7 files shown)


packages/instantsearch.js/src/lib/__tests__/server-composition.test.ts
```diff
@@ -1,3 +1,7 @@
+/**
+ * @jest-environment @instantsearch/testutils/jest-environment-node.js
```
Haroenv: could you set this as the default jest environment instead of explicitly passing it? in jest.config.js you should be able to set `testEnvironment: '@instantsearch/testutils/jest-environment-node.js'`

shaejaz: Done!

Haroenv (approved):
lgtm when green.

In a followup PR we may be able to check if we could add some sort of alias to make the environment shorter to write

PR #6725, opened 2025-10-06
fix(instantsearch-ui-components): fix auto focus on open

FabienMotte:
**Summary**

This PR fixes the auto-focus of the chat prompt when opening the chat for InstantSearch.js by passing an external ref from the widget (in InstantSearch.js and React InstantSearch).

### DIFF
--- a/packages/instantsearch.js/src/widgets/chat/chat.tsx
+++ b/packages/instantsearch.js/src/widgets/chat/chat.tsx
@@ -198,6 +198,7 @@ const createRenderer = <THit extends NonNullable<object> = BaseHit>({
       const [isClearing, setIsClearing] = state.use(false);
       const [maximized, setMaximized] = state.use(false);
       const [isScrollAtBottom, setIsScrollAtBottom] = state.use(true);
+      const [promptRef] = state.use({ current: null });
 
       const onClear = () => setIsClearing(true);
       const onClearTransitionEnd = () => {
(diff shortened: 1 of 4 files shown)


packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -197,6 +197,7 @@ const createRenderer = <THit extends NonNullable<object> = BaseHit>({
 
       const [isClearing, setIsClearing] = state.use(false);
       const [maximized, setMaximized] = state.use(false);
+      const [promptRef] = state.use({ current: null });
```
Haroenv, line 201: Do you think the local state is useful for other widgets too, like refinement list show more? would be good in some follow up later to make those patterns equal.

Haroenv (approved):

packages/instantsearch.js/src/widgets/chat/chat.tsx
FabienMotte: I suppose we could expose the `open`/`setOpen` local state (and others) as props (`open`/`onToggleOpen`) while keeping them in an uncontrolled mode by default. This would allow for customization while keeping the usage the same, matching the other widgets.

WDYT?

Haroenv: Not sure what you mean by that to be honest. how can you have an open/setOpen prop in JS?

FabienMotte: Forgot about JS, I was thinking in a React environment

PR #6726, opened 2025-10-06
fix(chat): add missing props to the IS.js chat widget

FabienMotte:
**Summary**

This PR fixes the chat widget in InstantSearch.js by adding the missing props to the Chat UI component to achieve parity with the React InstantSearch chat widget.


Haroenv (approved):
ideally common test suite covers these cases, but it's fine without for now

shaejaz (approved):

PR #6727, opened 2025-10-06
chore(examples): add chat widget to the getting started examples

FabienMotte:
**Summary**

This PR adds the chat widget to the InstantSearch.js and React InstantSearch Getting Started examples.

### DIFF
--- a/examples/js/getting-started/src/app.js
+++ b/examples/js/getting-started/src/app.js
@@ -16,7 +16,7 @@ import 'instantsearch.css/themes/satellite.css';
 
 const searchClient = algoliasearch(
   'latency',
-  '312a11382a63bdfc06e3183492fc469f'
+  '6be0576ff61c053d5f9a3225e2a90f76'
 );
 
 const search = instantsearch({
@@ -79,7 +79,7 @@ search.addWidgets([
   }),
   chat({
     container: '#chat',
-    agentId: '8f80fec3-f86a-1c8-86bb-426cb067cbbd',
+    agentId: '1eed8cbe-c096-4b8e-bfe7-873b9bc040ef',
     templates: {
       item: productItemTemplate,
     },
--- a/examples/js/getting-started/src/app.js
+++ b/examples/js/getting-started/src/app.js
@@ -16,7 +16,7 @@ import 'instantsearch.css/themes/satellite.css';
 
 const searchClient = algoliasearch(
   'latency',
-  '6be0576ff61c053d5f9a3225e2a90f76'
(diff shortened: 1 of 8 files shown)


examples/js/getting-started/src/app.js
```diff
@@ -60,20 +73,17 @@ search.addWidgets([
...
-          </article>
-        </div>
-      `,
+      item: productItemTemplate,
       layout: carousel(),
     },
   }),
+  chat({
+    container: '#chat',
+    agentId: 'xxxx',
```
FabienMotte: Waiting for an agent ID

examples/react/getting-started/src/App.tsx
```diff
@@ -21,7 +22,7 @@ import './App.css';
 
 const searchClient = algoliasearch(
   'latency',
-  '6be0576ff61c053d5f9a3225e2a90f76'
+  '312a11382a63bdfc06e3183492fc469f'
```
Haroenv: why is the api key changed? does the original one not get access to agents / this agent?

examples/react/getting-started/src/App.tsx
```diff
@@ -69,6 +70,11 @@ export function App() {
               </div>
             </div>
           </div>
+
+          <Chat
+            agentId="8f80fec3-f86a-1c8-86bb-426cb067cbbd"
+            itemComponent={ItemComponent}
```
shaejaz, line 76: Should we also have a dummy tool to showcase that API?

examples/react/getting-started/src/App.tsx
FabienMotte: I was experimenting and forgot to change it back

examples/react/getting-started/src/App.tsx
FabienMotte: I agree. I'm just not sure if we need to clone the agent in question to update its configuration, or if we can update it directly.
cc @PLNech

Haroenv: Ideally this is _our_ agent we can play with and use as a demo?

PLNech: Hey, yes you should create your own! This existing agent is my algobot demo and might change prompts, datasources, etc at any time, I don't guarantee I'll keep it in the current shape at all :stuck_out_tongue:

PLNech: Ideally on your own app so you have control over indices, analytics, and more!

Haroenv (approved):
let's merge this already and replace the agentId after

PR #6730, opened 2025-10-06
feat(chat): add top level components and templates to chat widgets

shaejaz:
**Summary**

**Result**


packages/instantsearch.js/src/types/templates.ts
```diff
@@ -50,7 +50,11 @@ export type TemplateWithBindEvent<TTemplateData = void> =
     ) => VNode | VNode[] | string);
 
 export type Templates = {
-  [key: string]: Template<any> | TemplateWithBindEvent<any> | undefined;
+  [key: string]:
+    | Template<any>
+    | TemplateWithBindEvent<any>
+    | Templates
```
Haroenv, line 56: don't need this change anymore right?

Haroenv: or still for the top-level ones that are not used?

shaejaz: We need this to not have a type error with the nested templates. Not sure if there's another way to fix it tho

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -144,7 +145,7 @@ const createRenderer = <THit extends NonNullable<object> = BaseHit>({
   containerNode: HTMLElement;
   cssClasses: ChatCSSClasses;
   renderState: {
-    templateProps?: PreparedTemplateProps<Required<ChatTemplates<THit>>>;
+    templateProps?: PreparedTemplateProps<ChatTemplates<THit>>;
```
Haroenv, line 150: this is normally set as required since everything has a default value. Is that no longer the case?

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -170,7 +171,7 @@ const createRenderer = <THit extends NonNullable<object> = BaseHit>({
       renderState.templateProps = prepareTemplateProps({
         defaultTemplates,
         templatesConfig: instantSearchInstance.templatesConfig,
-        templates,
+        templates: templates as any,
```
Haroenv: does this function deal with the nesting?

packages/instantsearch.js/src/types/templates.ts
Haroenv: only other way I see is removing the nested ones from the top-level templates, but that requires another iteration and variables to name

packages/instantsearch.js/src/widgets/chat/chat.tsx
shaejaz: For chat I don't think this required since the defaults are handled in the ui components.

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -190,6 +197,108 @@ const createRenderer = <THit extends NonNullable<object> = BaseHit>({
       },
     }));
 
+    renderState.templateProps = prepareTemplateProps({
+      defaultTemplates: {} as unknown as NonNullable<
+        Required<ChatTemplates<THit>['prompt']>
+      >,
+      templatesConfig: instantSearchInstance.templatesConfig,
+      templates: templates.prompt,
+    }) as PreparedTemplateProps<ChatTemplates<THit>>;
```
Haroenv: you call this templateProps, but it's only prompt, expected?

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -190,6 +197,108 @@ const createRenderer = <THit extends NonNullable<object> = BaseHit>({
...
+      getDefinedProperties({
+        textareaLabel: templates.prompt?.textareaLabelText,
+        textareaPlaceholder: templates.prompt?.textareaPlaceholderText,
+        emptyMessageTooltip: templates.prompt?.emptyMessageTooltipText,
+        stopResponseTooltip: templates.prompt?.stopResponseTooltipText,
+        sendMessageTooltip: templates.prompt?.sendMessageTooltipText,
+        disclaimer: templates.prompt?.disclaimerText,
+      });
+
+    renderState.templateProps = prepareTemplateProps({
```
Haroenv: no it's not expected. You override the values. There should be one templateProps per type

Haroenv (commented):
looks good, a couple improvements needed still

packages/instantsearch.js/src/widgets/chat/chat.tsx
shaejaz: Ah right I'll assign it to a new var.

shaejaz: Done! I don't think the renderState's template props are going to be used in this case, should we remove it or keep them in case they are needed.

Edit: Nvm it will be used for any top level templates (`actions` in this case for now)

packages/instantsearch-ui-components/src/components/chat/ChatHeader.tsx
```diff
@@ -18,7 +18,7 @@ export type ChatHeaderTranslations = {
   /**
    * Accessible label for the minimize button
    */
-  minimizedLabel: string;
+  minimizeLabel: string;
```
FabienMotte, line 21: Good catch, thanks!

FabienMotte (approved):

PR #6732, opened 2025-10-07
docs(examples): depend on all transitive internal dependencies

Haroenv:
**Summary**

This allows us to make a deep change in the helper, ui components or similar, but still try that out in an example in codesandbox without manually adding that dependency.

**Result**

all local dependencies (including transitive) are explicitly depended upon for examples.

A visual way to try this out is running the react example and adding a chat. Works with this preview, but not with previous ones.


FabienMotte (approved):
Thank you!

PR #6734, opened 2025-10-07
feat(instantsearch-ui-components): add `Button` component and refactor chat components

FabienMotte:
**Summary**

This PR adds a `Button` component and refactors the chat components to use this new component.
It also resolves styling issues that occurred when using the chat component styles as standalone by including `reset.css`

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/ChatHeader.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/ChatHeader.tsx
@@ -147,7 +147,7 @@ export function createChatHeaderComponent({ createElement }: Renderer) {
       >
         <span className={cx('ais-ChatHeader-title', classNames.title)}>
           <span
-            className={cx('ais-ChatHeader-title-icon', classNames.titleIcon)}
+            className={cx('ais-ChatHeader-titleIcon', classNames.titleIcon)}
           >
             {TitleIconComponent ? (
               <TitleIconComponent />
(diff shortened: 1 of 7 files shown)


packages/instantsearch-ui-components/src/components/chat/ChatHeader.tsx
```diff
@@ -144,7 +147,7 @@ export function createChatHeaderComponent({ createElement }: Renderer) {
       >
         <span className={cx('ais-ChatHeader-title', classNames.title)}>
           <span
-            className={cx('ais-ChatHeader-titleIcon', classNames.titleIcon)}
+            className={cx('ais-ChatHeader-title-icon', classNames.titleIcon)}
```
Haroenv: we use BEM normally, so only two dashes allowed :)

packages/instantsearch-ui-components/src/components/chat/ChatMessageLoader.tsx
```diff
@@ -37,21 +37,21 @@ export function createChatMessageLoaderComponent({
       >
         <div className="ais-ChatMessage-container">
           <div className="ais-ChatMessage-leading">
-            <div className="ais-ChatMessage-loaderSpinner">
+            <div className="ais-ChatMessage-loader-spinner">
```
Haroenv: also. I guess it could be `ais-ChatLoader-spinner` instead

packages/instantsearch.css/src/components/chat.scss
```diff
@@ -1,11 +1,12 @@
+@use '../themes/reset';
```
Haroenv, line 1: technically causes some effects outside of the chat, but it's probably ok

FabienMotte: Yes, we should probably do a scoped reset per component in the future

PR #6735, opened 2025-10-07
feat(autocomplete): implement suggestions for React

aymeric-giraudet:
**Summary**

[FX-3503](https://algolia.atlassian.net/browse/FX-3503)

**Result**

Had to implement something like `onSelect` options, and pass `setQuery`.
Not sure if `getQuery` and `getURL` are really relevant with this, as they can just do anything with `onSelect` + whatever is on their record

### DIFF
--- a/packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSuggestion.tsx
+++ b/packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSuggestion.tsx
@@ -1,0 +1,39 @@
+/** @jsx createElement */
+
+import { cx } from '../../lib';
+
+import type { Renderer } from '../../types';
+
+export type AutocompleteSuggestionProps<
+  T = { query: string } & Record<string, unknown>
+> = {
+  item: T;
+  onSelect: () => void;
+  classNames?: Partial<AutocompleteSuggestionClassNames>;
+};
+
+export type AutocompleteSuggestionClassNames = {
+  /**
+   * Class names to apply to the root element
+   **/
+  root: string | string[];
+};
+
+export function createAutocompleteSuggestionComponent({
+  createElement,
+}: Renderer) {
+  return function AutocompleteSuggestion({
+    item,
+    onSelect,
+    classNames,
+  }: AutocompleteSuggestionProps) {
+    return (
+      <div
+        onClick={onSelect}
+        className={cx('ais-AutocompleteSuggestion', classNames?.root)}
+      >
+        {item.query}
+      </div>
+    );
+  };
+}
--- a/packages/react-instantsearch/src/components/Autocomplete.tsx
+++ b/packages/react-instantsearch/src/components/Autocomplete.tsx
@@ -2,13 +2,18 @@ import {
   createAutocompleteComponent,
   createAutocompleteIndexComponent,
   createAutocompletePanelComponent,
+  createAutocompleteSuggestionComponent,
+  cx,
 } from 'instantsearch-ui-components';
 import React, { createElement, useState, Fragment } from 'react';
-import { Index, useHits } from 'react-instantsearch-core';
+import { Index, useHits, useInstantSearch } from 'react-instantsearch-core';
 
 import { SearchBox } from '../widgets/SearchBox';
 
-import type { Pragma } from 'instantsearch-ui-components';
+import type {
+  AutocompleteIndexClassNames,
+  Pragma,
+} from 'instantsearch-ui-components';
 import type { BaseHit, Hit } from 'instantsearch.js';
 
 const Autocomplete = createAutocompleteComponent({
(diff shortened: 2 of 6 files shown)


packages/react-instantsearch/src/components/__tests__/Autocomplete.test.tsx
```diff
@@ -193,4 +195,86 @@ describe('Autocomplete', () => {
...
+    `);
+
+    userEvent.click(screen.getByRole('searchbox'));
+    userEvent.click(screen.getByText(/hello/i));
+
+    await waitFor(() => {
+      expect(searchClient.search).toHaveBeenCalledTimes(3);
+    });
+
+    expect(searchClient.search).toHaveBeenLastCalledWith([
```
shaejaz, line 280: Ideally we also have an `onSelect` test here, but good as a start.

packages/react-instantsearch/src/components/__tests__/Autocomplete.test.tsx
```diff
@@ -193,4 +195,86 @@ describe('Autocomplete', () => {
...
+      {
+        indexName: 'indexName',
```
shaejaz, line 282: Shouldn't the test also include a check if the `query_suggestions` index is called?

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSuggestion.tsx
```diff
@@ -0,0 +1,39 @@
...
+}: Renderer) {
+  return function AutocompleteSuggestion({
+    item,
+    onSelect,
+    classNames,
+  }: AutocompleteSuggestionProps) {
+    return (
+      <div
+        onClick={onSelect}
+        className={cx('ais-AutocompleteSuggestion', classNames?.root)}
```
dhayab: Nitpick: you can default `classNames` to prevent optional chaining later.

```suggestion
    classNames = {},
  }: AutocompleteSuggestionProps) {
    return (
      <div
        onClick={onSelect}
        className={cx('ais-AutocompleteSuggestion', classNames.root)}
```

packages/react-instantsearch/src/components/Autocomplete.tsx
```diff
@@ -56,20 +119,30 @@ export function EXPERIMENTAL_Autocomplete({ indices }: AutocompleteProps) {
...
   return (
     <AutocompleteIndex
       // @ts-expect-error - there seems to be problems with React.ComponentType and this, but it's actually correct
       ItemComponent={ItemComponent}
       items={items}
+      onSelect={(item) => {
+        onSelect?.({
+          item,
+          getQuery: () => getQuery?.(item) || '',
+          getURL: () => getURL?.(item) || '',
```
dhayab, line 141: Sorry I forgot, can you remind me who's providing `getQuery` and `getURL` and where?

packages/react-instantsearch/src/components/Autocomplete.tsx
```diff
@@ -26,19 +31,72 @@ const AutocompleteIndex = createAutocompleteIndexComponent({
...
+    getURL: () => string;
+    setQuery: (query: string) => void;
+  }) => void;
+  classNames?: Partial<AutocompleteIndexClassNames>;
 };
 
 export type AutocompleteProps = {
-  indices: IndexConfig[];
+  indices?: IndexConfig[];
+  showSuggestions?: {
```
dhayab, line 60: Are we set on `showSuggestions`? Just wondering about the added value of `show` as a prefix here.

dhayab (approved):
Just nitpicks and questions from me, overall it works as expected.

packages/react-instantsearch/src/components/__tests__/Autocomplete.test.tsx
aymeric-giraudet: Done !

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSuggestion.tsx
aymeric-giraudet: Yes, thanks !

packages/react-instantsearch/src/components/__tests__/Autocomplete.test.tsx
aymeric-giraudet: It's best to test further down so we don't rely so much on implementation details imo, ideal would be to test the DOM updates rather than search

packages/react-instantsearch/src/components/Autocomplete.tsx
aymeric-giraudet: It's part of the API specs on confluence, part of the `indices` object shape, but not sure it's really useful, `onSelect` is better imo as it gives them more customization, we don't necessarily need getters like this, just did it to follow the initial API draft but we'll see later if we keep them

packages/react-instantsearch/src/components/Autocomplete.tsx
aymeric-giraudet: Part of the API specs on Confluence again, we can discuss further once we've implemented before releasing I guess

dhayab: Ah yes sorry, for some reason the RFC doc was empty when I checked it before asking, and I forgot we discussed it already. Happy to commit to `showSuggestions` 🙇 !

packages/react-instantsearch/src/components/Autocomplete.tsx
Haroenv: don't you still need the URL for the link href?

aymeric-giraudet: @Haroenv Yes but they would just implement their `itemComponent` like `<a href={item.url}>`. 
`getURL` is really only useful if we had a default component for custom indices that renders anchors. Maybe we should ?
In any case I can't see a scenario for `getQuery` apart from query suggestions which we already handle.

PR #6739, opened 2025-10-08
refactor(tools): make search tool the same

Haroenv:
- same between js and react
- close chat on "view all"
- remove "tool-" prefix in js (not needed since we moved to object form)
- refactor search index tool in react
- carousel assumes it can scroll right on initial mount in js
- header template for carousel in js

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
@@ -95,7 +95,7 @@ export type ChatMessagesProps<
   /**
    * Tools available for the assistant
    */
-  tools?: ClientSideTools;
+  tools: ClientSideTools;
   /**
    * Current chat status
    */
@@ -107,7 +107,11 @@ export type ChatMessagesProps<
   /**
    * Callback for reload action
    */
-  onReload?: (messageId?: string) => void;
+  onReload: (messageId?: string) => void;
+  /**
+   * Function to close the chat
+   */
+  onClose: () => void;
   /**
    * Optional class names
    */
@@ -169,6 +173,7 @@ function createDefaultMessageComponent<
     indexUiState,
     setIndexUiState,
     onReload,
+    onClose,
     translations,
     actionsComponent,
   }: {
@@ -178,8 +183,9 @@ function createDefaultMessageComponent<
     assistantMessageProps?: Partial<ChatMessageProps>;
     indexUiState: object;
     setIndexUiState: (state: object) => void;
-    tools?: ClientSideTools;
-    onReload?: (messageId?: string) => void;
+    tools: ClientSideTools;
+    onReload: (messageId?: string) => void;
+    onClose: () => void;
     translations: ChatMessagesTranslations;
     actionsComponent?: ChatMessageProps['actionsComponent'];
   }) {
@@ -196,7 +202,7 @@ function createDefaultMessageComponent<
       {
         title: translations.regenerateLabel,
         icon: () => <ReloadIconComponent createElement={createElement} />,
-        onClick: (m) => onReload?.(m.id),
+        onClick: (m) => onReload(m.id),
       },
     ];
 
@@ -213,6 +219,7 @@ function createDefaultMessageComponent<
         tools={tools}
         indexUiState={indexUiState}
         setIndexUiState={setIndexUiState}
+        onClose={onClose}
         actions={defaultActions}
         actionsComponent={actionsComponent}
         data-role={message.role}
@@ -262,6 +269,7 @@ export function createChatMessagesComponent({
       status = 'ready',
       hideScrollToBottom = false,
       onReload,
+      onClose,
       translations: userTranslations,
       userMessageProps,
       assistantMessageProps,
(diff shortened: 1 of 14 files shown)


packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
```diff
@@ -94,7 +94,7 @@ export type ChatMessagesProps<
   /**
    * Tools available for the assistant
    */
-  tools?: ClientSideTools;
+  tools: ClientSideTools;
```
shaejaz, line 98: I know it doesn't matter too much since this is internal, but tools are technically optional since agents don't necessarily need to have them added.

packages/react-instantsearch/src/widgets/chat/tools/SearchIndexTool.tsx
shaejaz, line 1: Should we use this refactor opportunity to move this into ui-components? I see some copied over code in js but not sure how much actual work would be needed at a glance.

packages/instantsearch-ui-components/src/components/chat/types.ts
```diff
@@ -22,6 +22,7 @@ export type ClientSideToolComponentProps = {
   message: ChatToolMessage;
   indexUiState: object;
   setIndexUiState: (state: object) => void;
+  onClose: () => void;
```
shaejaz, line 25: Doesn't need to be done now, but I think sending the entire chat ref would be a very flexible alternative to this if it's possible and clean.

packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
Haroenv: yes it's optional, but then you just pass an empty object, no?

packages/react-instantsearch/src/widgets/chat/tools/SearchIndexTool.tsx
Haroenv: I tried but the header needs to be memoised

packages/instantsearch-ui-components/src/components/chat/types.ts
Haroenv: does the chat ref have access to open/close/input? I don't think so

shaejaz: Yeah not currently, it would be a refactor to have abstractions available over these states that could be used via a ref.

packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
shaejaz: Yeah, I guess this is fine for an internal component 👍

packages/instantsearch-ui-components/src/components/chat/types.ts
Haroenv: can do that later indeed. We'd have to decide what's presentational logic and what's core chat logic

packages/react-instantsearch/src/widgets/chat/tools/SearchIndexTool.tsx
```diff
@@ -8,12 +8,15 @@ import React, { createElement } from 'react';
...
   Pragma,
   RecommendComponentProps,
   RecordWithObjectID,
   UserClientSideTools,
 } from 'instantsearch-ui-components';
 import type { IndexUiState } from 'instantsearch.js';
+import type { IndexWidget } from 'instantsearch.js/src/widgets';
+import type { ComponentProps } from 'react';
 
 export const SearchIndexToolType: string = 'algolia_search_index';
```
shaejaz: This should be imported (and re-exported probably) from the is.js since it's already defined there.

packages/instantsearch.js/src/lib/chat/tools.ts
```diff
@@ -1,5 +1,3 @@
-import type { ChatToolType } from 'instantsearch-ui-components';
+export const SearchIndexToolType: string = 'algolia_search_index';
 
-export const SearchIndexToolType: ChatToolType = 'tool-algolia_search_index';
-
-export const defaultTools: ChatToolType[] = [SearchIndexToolType];
+export const defaultTools: string[] = [SearchIndexToolType];
```
shaejaz: I think this isn't actually used anywhere now so it can be removed.

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -55,76 +63,189 @@ function getDefinedProperties<T extends object>(obj: T): Partial<T> {
   ) as Partial<T>;
 }
 
-function createDefaultTools<THit extends NonNullable<object> = BaseHit>(
-  templates: ChatTemplates<THit>
-): UserClientSideToolsWithTemplate {
-  const Carousel = carousel();
-
-  const Component: ClientSideToolComponent = ({
+function createSearchIndexTool<
```
shaejaz, line 69: Should this also be moved to a separate search tool file the same way it's done on react?

shaejaz (commented):
Looks good, just a few comments

packages/instantsearch.js/src/widgets/chat/chat.tsx
Haroenv: I tried but it has a lot of dependencies of types in chat, which would require those to move to a ./types as well, and then be a lot of different files. I know this file is too big but splitting up just tools vs. everything else doesn't seem right

PR #6740, opened 2025-10-09
feat(chat): add custom toggle button support

FabienMotte:
**Summary**

This PR:
- adds a prop `toggleButtonComponent` that allow passing a custom toggle button component to the chat UI component.
- forwards that prop for the React InstantSearch chat widget.
- adds two templates (`toggleButton.layout` and `toggleButton.icon`) for the InstantSearch.js chat widget.

One issue I currently see is that when using a custom toggle button component, you would need to pass `pointer-events: all;` on your custom button because we use `pointer-events: none;` on the `.ais-Chat` wrapper.
Should we pass this style in the Chat UI component if we detect a custom `ToggleButtonComponent` is used?

```ts
const CustomToggleButtonComponent = (props) => {
  return (
    <button style={{ pointerEvents: 'all' }} {...props}>
      Toggle Chat
    </button>
  );
};

<Chat
  /* ... */
  toggleButtonComponent={CustomToggleButtonComponent}
/>
```

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/__tests__/Chat.test.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/__tests__/Chat.test.tsx
@@ -222,26 +222,30 @@ describe('Chat', () => {
               </div>
             </form>
           </div>
-          <button
-            class="ais-Button ais-Button--primary ais-Button--md ais-Button--icon-only ais-ChatToggleButton ais-ChatToggleButton--open"
-            type="button"
+          <div
+            class="ais-Chat-toggleButtonWrapper"
           >
-            <svg
-              fill="none"
-              height="28"
-              stroke="currentColor"
-              strokeLinecap="round"
-              strokeLinejoin="round"
-              strokeWidth="2"
-              viewBox="0 0 24 24"
-              width="28"
-              xmlns="http://www.w3.org/2000/svg"
+            <button
+              class="ais-Button ais-Button--primary ais-Button--md ais-Button--icon-only ais-ChatToggleButton ais-ChatToggleButton--open"
+              type="button"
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch/src/widgets/Chat.tsx
```diff
@@ -83,6 +83,7 @@ export type ChatProps<TObject, TUiMessage extends UIMessage = UIMessage> = Omit<
     headerProps?: UserHeaderProps;
     messagesProps?: UserMessagesProps;
     promptProps?: UserPromptProps;
+    toggleButtonComponent?: ChatUiProps['toggleButtonComponent'];
```
shaejaz, line 101:
```suggestion
    toggleButtonLayoutComponent?: ChatUiProps['toggleButtonComponent'];
```

Haroenv: I think it can be just component as the icon isn't changeable in react? although I think there should be an option to change just the inside and separately everything

FabienMotte: The icon can be customised using the `toggleButtonProps.toggleIconComponent` prop.
Should we add a `toggleButtonIconComponent` prop to the widget?

Haroenv: yes sounds good

FabienMotte: Should I also rename `headerLayoutComponent` to `headerComponent` and `promptLayoutComponent` to `promptComponent` to align with the naming of `toggleButtonComponent`/`toggleButtonIconComponent`?

PR #6743, opened 2025-10-09
test(common): allow customised parameters per flavor

shaejaz:
**Summary**

allow a test to have different options per flavour. This allows to test things like templates in a CTS manner.

Also adds some chat tests, which use this feature

### DIFF
--- a/examples/js/getting-started/src/app.js
+++ b/examples/js/getting-started/src/app.js
@@ -26,15 +26,19 @@ const search = instantsearch({
 });
 
 const productItemTemplate = (item, { html }) => html`
-  <div>
-    <article>
-      <div>
-        <img src="${item.image}" />
-        <h2>${item.name}</h2>
-      </div>
-      <a href="/products.html?pid=${item.objectID}">See product</a>
-    </article>
-  </div>
+  <article class="ais-Carousel-hit">
+    <div class="ais-Carousel-hit-image">
+      <img src="${item.image}" />
+    </div>
+    <h2 class="ais-Carousel-hit-title">
+      <a
+        href="/products.html?pid=${item.objectID}"
+        class="ais-Carousel-hit-link"
+      >
+        ${item.name}
+      </a>
+    </h2>
+  </article>
 `;
 
 search.addWidgets([
--- a/examples/react/getting-started/src/App.tsx
+++ b/examples/react/getting-started/src/App.tsx
@@ -105,14 +105,18 @@ function HitComponent({ hit }: { hit: HitType }) {
 
 function ItemComponent({ item }: { item: Hit }) {
   return (
-    <div>
-      <article>
-        <div>
-          <img src={item.image} />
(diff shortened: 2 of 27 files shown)


packages/react-instantsearch/src/__tests__/common-widgets.test.tsx
```diff
@@ -392,8 +393,15 @@ const testSetups: TestSetupsMap<TestSuites> = {
...
     );
   },
-  createChatWidgetTests() {
-    throw new Error('Chat is not tested through the Common Test Suite yet');
+  createChatWidgetTests({ instantSearchOptions, widgetParams }) {
+    render(
+      <InstantSearch {...instantSearchOptions}>
+        <div className="ais-Chat">
+          <Chat {...widgetParams} />
+        </div>
```
dhayab: Doesn't `<Chat>` already have the `ais-Chat` class name?

```suggestion
      <Chat {...widgetParams} />
```

tests/common/widgets/chat/index.ts
```diff
@@ -2,26 +2,55 @@ import { fakeAct, skippableDescribe } from '../../common';
...
-export function createChatWidgetTests(
-  setup: ChatWidgetSetup,
-  { act = fakeAct, skippedTests = {} }: TestOptions = {}
+declare module '../../common' {
+  interface FlavoredWidgetParams {
+    createChatWidgetTests: {
+      javascript: JSChatWidgetParams;
+      react: ReactChatWidgetParams;
+      vue: Record<string, never>;
+    };
```
dhayab: Would you be able to rely on an exported type based off of ChatWidgetSetup (before being scoped to `T`), instead of repeating the structure?

Haroenv: the type is actually different, this is the type per flavour, but the exported one is the union of all parameters.

dhayab: Not sure I follow, to me it looks like this would be possible:

```ts
export type _FlavoredChatWidgetParams = {
  javascript: JSChatWidgetParams;
  react: ReactChatWidgetParams;
  vue: Record<string, never>;
};

export type ChatWidgetParams<T ...> = _FlavoredChatWidgetParams[T];

// ...

declare module '../../common' {
  interface FlavoredWidgetParams {
    createChatWidgetTests: _FlavoredChatWidgetParams
  }
}
```

All in all, this is a nitpick, so feel free to disregard.

packages/react-instantsearch/src/__tests__/common-widgets.test.tsx
shaejaz: This is fixed on the follow-up PR adding more tests, but also makes sense to fix here if needed.

dhayab: That's fine then 👍

PR #6747, opened 2025-10-14
feat(autocomplete): implement recent searches

aymeric-giraudet:
**Summary**

[FX-3504](https://algolia.atlassian.net/browse/FX-3504)

**Result**

image

[FX-3504]: https://algolia.atlassian.net/browse/FX-3504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch-ui-components/src/components/autocomplete/createAutocompleteStorage.ts
+++ b/packages/instantsearch-ui-components/src/components/autocomplete/createAutocompleteStorage.ts
@@ -9,7 +9,7 @@ type CreateAutocompleteStorageParams = {
 };
 
 type UseStorageParams<TItem extends Record<string, unknown>> = {
-  showRecent?: boolean | object;
+  showRecent?: boolean | { storageKey?: string };
   query?: string;
 } & Pick<Parameters<UsePropGetters<TItem>>[0], 'indices' | 'indicesConfig'>;
 
@@ -24,7 +24,14 @@ export function createAutocompleteStorage({
     indices,
     indicesConfig,
   }: UseStorageParams<TItem>) {
-    const storage = useMemo(() => createStorage({ limit: 5 }), []);
+    const storageKey =
+      showRecent && typeof showRecent === 'object'
+        ? showRecent.storageKey
+        : undefined;
+    const storage = useMemo(
+      () => createStorage({ limit: 5, storageKey }),
+      [storageKey]
+    );
     const [snapshot, setSnapshot] = useState(storage.getSnapshot());
     useEffect(() => {
       storage.registerUpdateListener(() => {
@@ -86,7 +93,7 @@ function isLocalStorageSupported() {
   }
 }
 
-function getLocalStorage() {
+function getLocalStorage(key: string = LOCAL_STORAGE_KEY) {
   if (!isLocalStorageSupported()) {
     return {
       setItems() {},
@@ -99,21 +106,27 @@ function getLocalStorage() {
   return {
     setItems(items: string[]) {
       try {
-        window.localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify(items));
+        window.localStorage.setItem(key, JSON.stringify(items));
       } catch {
         // do nothing, this likely means the storage is full
       }
     },
     getItems(): string[] {
-      const items = window.localStorage.getItem(LOCAL_STORAGE_KEY);
+      const items = window.localStorage.getItem(key);
 
       return items ? (JSON.parse(items) as string[]) : [];
     },
   };
 }
 
-export function createStorage({ limit = 5 }: { limit: number }) {
-  const storage = getLocalStorage();
+export function createStorage({
+  limit = 5,
+  storageKey,
+}: {
+  limit: number;
+  storageKey?: string;
+}) {
+  const storage = getLocalStorage(storageKey);
   let updateListener: (() => void) | null = null;
 
   return {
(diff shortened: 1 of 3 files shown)


packages/instantsearch-ui-components/src/components/autocomplete/createAutocompleteStorage.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    const storageHits = snapshot.getAll(query).map((value) => ({
+      objectID: value,
+      query: value,
+      __indexName: 'recent-searches',
+    }));
+
+    const indicesForPropGetters = [...indices];
+    const indicesConfigForPropGetters = [...indicesConfig];
+    if (showRecent) {
```
Haroenv: this implies we're saving recent searches even if showRecent is false. Should maybe the whole storage be conditional based on showRecent?

packages/instantsearch-ui-components/src/components/autocomplete/createAutocompleteStorage.ts
```diff
@@ -0,0 +1,139 @@
...
+      storage,
+      storageHits,
+      indicesForPropGetters,
+      indicesConfigForPropGetters,
+    };
+  };
+}
+
+const LOCAL_STORAGE_KEY_TEST = 'test-localstorage-support';
+const LOCAL_STORAGE_KEY = 'autocomplete-recent-searches';
```
Haroenv, line 83: Is there some way where we link the autocomplete setup with the recent searches that are being presented? Do we do that in autocomplete.js? 

What if there's multiple autocomplete instances in multiple places in the app?

Haroenv (commented):
looks good. I have a couple questions around interop

packages/instantsearch-ui-components/src/components/autocomplete/createAutocompleteStorage.ts
aymeric-giraudet: Can't call hooks conditionally but I could return a storage with empty function or something yeah

packages/instantsearch-ui-components/src/components/autocomplete/createAutocompleteStorage.ts
aymeric-giraudet: I think we could but the original lib doesn't, even though it does have code to handle multiple autocomplete components on a page (which I don't think anyone needs really)

Haroenv: You could have a site-wide autocomplete and one for example just for store finder. I think there are some use cases for multi-instance. Maybe storage should be attached to the original index?

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteRecentSearch.tsx
```diff
@@ -0,0 +1,113 @@
...
+            'ais-AutocompleteItemContent',
+            'ais-AutocompleteRecentSearchItemContent',
+            classNames.content
+          )}
+        >
+          <div
+            className={cx(
+              'ais-AutocompleteItemIcon',
+              'ais-AutocompleteRecentSearchItemIcon',
+              classNames.content
```
FabienMotte, line 73:
```suggestion
              classNames.icon
```

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteRecentSearch.tsx
```diff
@@ -0,0 +1,113 @@
...
+            )}
+          >
+            <AutocompleteClockIcon createElement={createElement} />
+          </div>
+          <div
+            className={cx(
+              'ais-AutocompleteItemContentBody',
+              'ais-AutocompleteRecentSearchItemContentBody',
+              classNames.content
```
FabienMotte, line 82:
```suggestion
              classNames.body
```

packages/react-instantsearch/src/widgets/Autocomplete.tsx
```diff
@@ -170,16 +203,35 @@ function InnerAutocomplete<TItem extends BaseHit = BaseHit>({
...
   const { getInputProps, getItemProps, getPanelProps, getRootProps } =
     usePropGetters<TItem>({
-      indices,
-      indicesConfig,
+      indices: indicesForPropGetters,
+      indicesConfig: indicesConfigForPropGetters,
       onRefine: (query) => {
         refineAutocomplete(query);
         refineSearchBox(query);
+        storage.onAdd(query);
```
FabienMotte, line 234: Should we check the length of the query, as we do in the IS.js version? Should we also prevent whitespace queries from being saved?

packages/instantsearch-ui-components/src/components/autocomplete/createAutocompleteStorage.ts
```diff
@@ -0,0 +1,159 @@
...
+    query,
+    indices,
+    indicesConfig,
+  }: UseStorageParams<TItem>) {
+    const storageKey =
+      showRecent && typeof showRecent === 'object'
+        ? showRecent.storageKey
+        : undefined;
+    const storage = useMemo(
+      () => createStorage({ limit: 5, storageKey }),
```
FabienMotte, line 32: Should we allow this limit to be customized?

PR #6749, opened 2025-10-15
feat(autocomplete): generate component props for keyboard navigation and accessibility

dhayab:
**Summary**

This PR sets up a `useAutocomplete()` hook that:
- generates props for autocomplete components
- keeps an indexed items store

It centralizes keyboard support, accessibility and manages a basic state for the Autocomplete.

Keyboard navigation is based on the [Editable Combobox with Grid Popup Example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/grid-combo/) from WAI. This prepares us for horizontal navigation on item actions later.

**Result**

[Preview sandbox ↗️](https://6gjhxz.csb.app/)


packages/react-instantsearch/src/components/Autocomplete.tsx
```diff
@@ -60,19 +67,30 @@ export type AutocompleteProps = {
...
 }: AutocompleteProps) {
-  const [isOpen, setIsOpen] = useState(false);
+  const {
+    getIndexProps,
+    getInputProps,
+    getItemProps,
+    getPanelProps,
+    getRootProps,
+    updateStore,
+  } = useAutocomplete();
```
Haroenv: if we'd use connectautocomplete at the level inside the index component, it could have the information and no need for a new store concept. The inconsistent part is then of course that the connector would know about the dom (unless there's a separate function that turns an autocomplete state into props. What do you think?

packages/react-instantsearch/src/components/useAutocomplete.tsx
```diff
@@ -0,0 +1,248 @@
...
+    };
+
+    document.body.addEventListener('click', onBodyClick);
+
+    return () => {
+      document.body.removeEventListener('click', onBodyClick);
+    };
+  }, [rootRef]);
+
+  const getNextActiveDescendent = (key: string): string | undefined => {
```
Haroenv, line 138: this function could be declared outside of the bode of useAutocomplete if it received the state as a prop. Not sure if it would be better though.

Haroenv (approved):
I think this is good to go, although possibly a bit much focused on react and not on reuse of the code for js

packages/react-instantsearch/src/components/Autocomplete.tsx
dhayab: I'm now using `connectAutocomplete` to create the items store and refine, and I'll migrate most of the state in the connector in a next PR, as well as move the prop getters to a shared part (probably ui components).

packages/react-instantsearch/src/components/Autocomplete.tsx
```diff
@@ -98,50 +96,54 @@ export function EXPERIMENTAL_Autocomplete({
     });
   }
 
+  const {
+    getIndexProps,
+    getInputProps,
+    getItemProps,
+    getPanelProps,
+    getRootProps,
+  } = useAutocomplete({ indices });
```
Haroenv, line 105: I don't really get why we're using connectAutocomplete at this level and not inside the isolated index. now it seems to be scoped differently and we can't use the refine of this useautocomplete for the state of the search input

packages/react-instantsearch/src/components/useAutocomplete.tsx
```diff
@@ -0,0 +1,274 @@
...
+  const [isOpen, setIsOpen] = useState(false);
+  const [activeDescendent, setActiveDescendent] = useState<string | undefined>(
+    undefined
+  );
+
+  const items = indices.reduce<UseAutocompleteItems<TItem>>(
+    (itemsAcc, index) => {
+      const indexItems =
+        connectorIndices
+          .find(({ indexName }) => index.indexName === indexName)
```
Haroenv, line 99: ideally this can be avoided if autocomplete connector is used inside the isolated index.

dhayab: Yes, I'll switch this in a future PR to align JS + React implementations.

PR #6752, opened 2025-10-15
test(chat): add and move tests to the CTS

shaejaz:
**Summary**
This PR moves the chat widget tests to the CTS and adds additional CTS tests to cover most of the widgets API.

### DIFF
--- a/tests/common/widgets/chat/utils.ts
+++ b/tests/common/widgets/chat/utils.ts
@@ -4,7 +4,7 @@ import { Chat } from 'instantsearch.js/es/lib/chat';
 
 import type { Act } from '../../common';
 
-export const createDefaultWidgetParams = (chat?: Chat) => ({
+export const createDefaultWidgetParams = (chat?: Chat<any>) => ({
   agentId: 'agentId',
   chat: chat ?? new Chat({}),
 });
--- a/tests/common/widgets/chat/templates.tsx
+++ b/tests/common/widgets/chat/templates.tsx
@@ -32,7 +32,7 @@ export function createTemplatesTests(
             },
             react: {
               ...createDefaultWidgetParams(),
-              headerLayoutComponent: () => (
+              headerComponent: () => (
                 <div className="custom-header">Custom header</div>
               ),
             },
@@ -128,7 +128,7 @@ export function createTemplatesTests(
             },
             react: {
               ...createDefaultWidgetParams(),
-              promptLayoutComponent: () => (
+              promptComponent: () => (
                 <div className="custom-prompt">Custom prompt</div>
               ),
             },


tests/common/widgets/chat/options.tsx
```diff
@@ -97,162 +112,895 @@ export function createOptionsTests<T extends SupportedFlavor>(
...
+
+          expect(
+            document.querySelector('.custom-title-icon')!.textContent
+          ).toBe('Custom title icon');
+          expect(
+            document.querySelector('.custom-close-icon')!.textContent
+          ).toBe('Custom close icon');
+          expect(
+            document.querySelector('.custom-maximize-icon')!.textContent
+          ).toBe('Custom maximize icon');
```
shaejaz: I think the minimize icon component doesn't actually get used to render anything. If that is the case, we might want to remove/refactor/rename the max and/or min components.

tests/common/widgets/chat/options.tsx
```diff
@@ -97,162 +112,895 @@ export function createOptionsTests<T extends SupportedFlavor>(
...
+
+      expect(document.querySelector('.ais-Chat')).not.toHaveClass(
+        'ais-Chat--maximized'
+      );
+      expect(document.querySelector('.ais-Chat-container')).not.toHaveClass(
+        'ais-Chat-container--maximized'
+      );
+    });
+
+    describe('templates', () => {
```
shaejaz: Templates, translations and maybe the tool tests can also be move to their own separate test files if that makes sense. This test file is getting too long I think.

packages/react-instantsearch/src/widgets/__tests__/Chat.test.tsx
shaejaz, line 1: I'm not sure if there are any react specific tests that we want to keep here.

tests/common/widgets/chat/options.tsx
Haroenv: yes, let's put templates and translations in their own relative files

packages/react-instantsearch/src/widgets/__tests__/Chat.test.tsx
Haroenv: seems like they're all covered

tests/common/widgets/chat/utils.ts
```diff
@@ -0,0 +1,22 @@
+import { wait } from '@instantsearch/testutils';
+import userEvent from '@testing-library/user-event';
+import { Chat } from 'instantsearch.js/es/lib/chat';
+
+import type { Act } from '../../common';
+
+export const createDefaultWidgetParams = (chat?: any) => ({
```
Haroenv:
```suggestion
export const createDefaultWidgetParams = (chat?: Chat) => ({
```

tests/common/widgets/chat/options.tsx
Haroenv: @FabienMotte

PR #6754, opened 2025-10-16
fix(chat): fix `stickToBottom` auto scroll behavior

FabienMotte:
**Summary**

This PR fixes the auto-scroll issue with the `stickToBottom` library in the chat when scrolling near the bottom.

**Result**

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/2d40d613-6b85-402f-9053-e410964f0145"> | <video src="https://github.com/user-attachments/assets/0185e169-255c-4fa2-8038-6f43b72657a9"> |

### DIFF
--- a/packages/instantsearch-ui-components/src/lib/stickToBottom.ts
+++ b/packages/instantsearch-ui-components/src/lib/stickToBottom.ts
@@ -125,17 +125,19 @@ const RETAIN_ANIMATION_DURATION_MS = 350;
 
 let mouseDown = false;
 
-document?.addEventListener('mousedown', () => {
-  mouseDown = true;
-});
+if (typeof document !== 'undefined') {
+  document.addEventListener('mousedown', () => {
+    mouseDown = true;
+  });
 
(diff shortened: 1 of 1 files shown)


packages/instantsearch-ui-components/src/lib/stickToBottom.ts
```diff
@@ -456,13 +456,13 @@ export const createStickToBottom = (
         setEscapedFromLock(false);
       }
 
-      if (!state.escapedFromLock && state.isNearBottom) {
+      if (!state.escapedFromLock && state.scrollDifference <= 2) {
```
Haroenv, line 459: this is the main fix right?

shaejaz (approved):

packages/instantsearch-ui-components/src/lib/stickToBottom.ts
FabienMotte: Yes!

PR #6757, opened 2025-10-17
feat(chat): add fallbacks for unhandled tools

FabienMotte:
**Summary**

This PR adds fallbacks for unhandled tools:
- In development, it throws an error asking the user to provide a tool implementation using the `tools` prop.
- In production, it adds a tool result with the output "No tool implemented for ...", which triggers a new request to the agent.

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
@@ -231,14 +231,8 @@ export function createChatMessageComponent({ createElement }: Renderer) {
             </div>
           );
         }
-
-        return null;
       }
-      return (
-        <pre key={`${message.id}-${index}`} className="ais-ChatMessage-code">
-          {JSON.stringify(part)}
-        </pre>
-      );
+      return null;
     }
 
     return (


packages/react-instantsearch-core/src/connectors/useChat.ts
```diff
@@ -193,3 +196,5 @@ export function useChat<TUiMessage extends UIMessage = UIMessage>({
     addToolResult: chatRef.current.addToolResult,
   };
 }
+
+export { lastAssistantMessageIsCompleteWithToolCalls };
```
Haroenv: we import directly in the ui component, so this can be reverted.

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -230,6 +231,8 @@ export function createChatMessageComponent({ createElement }: Renderer) {
             </div>
           );
         }
+
+        return null;
       }
       return (
         <pre key={`${message.id}-${index}`} className="ais-ChatMessage-code">
```
Haroenv: we should remove this branch and also return null for non-tool responses I think

PR #6745, opened 2025-10-13
feat(autocomplete): js implementation

Haroenv:
**Summary**

**Result**

PR #6759, opened 2025-10-20
feat(instantsearch.js): introduce autocomplete widget

dhayab:
This PR (re-)introduces the `autocomplete()` widget, with similar features to Algolia's standalone Autocomplete library.

It includes a refactoring of the React implementation to extract the prop getters into a shared module and match the draft JS design in #6745.

FX-3548

### DIFF
--- a/examples/js/algolia-experiences/package.json
+++ b/examples/js/algolia-experiences/package.json
@@ -1,6 +1,6 @@
 {
   "name": "example-algolia-experiences",
-  "version": "1.24.0",
+  "version": "1.25.0",
   "private": true,
   "scripts": {
     "start": "BABEL_ENV=parcel parcel *.html",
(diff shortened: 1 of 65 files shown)


packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
```diff
@@ -0,0 +1,370 @@
...
+import { useEffect, useId, useMemo, useRef, useState } from 'preact/hooks';
+
+import SearchBox, {
+  type SearchBoxComponentCSSClasses,
+} from '../../components/SearchBox/SearchBox';
+import TemplateComponent from '../../components/Template/Template';
+import {
+  connectAutocomplete,
+  connectSearchBox,
+} from '../../connectors/index.umd';
```
shaejaz, line 21: Is this import correct?

packages/instantsearch-ui-components/src/components/autocomplete/createAutocompletePropGetters.ts
```diff
@@ -0,0 +1,249 @@
...
+
+type CreateAutocompletePropGettersParams = {
+  useEffect: (effect: () => void, inputs?: readonly unknown[]) => void;
+  useId: () => string;
+  useMemo: <TType>(factory: () => TType, inputs: readonly unknown[]) => TType;
+  useRef: <TType>(initialValue: TType | null) => { current: TType | null };
+  useState: <TType>(
+    initialState: TType
+  ) => [TType, (newState: TType) => unknown];
+};
```
shaejaz, line 41: I think in a future PR, we could move these into a shared file. I believe there are some parts of the chat widgets that could use this approach.

packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
dhayab: Yes, it prevents us from importing the chat, so rollup can properly bundle the UMD package.

PR #6762, opened 2025-10-21
fix(chat): move stick-to-bottom logic to widget-level hooks

FabienMotte:
**Summary**

This PR refactors the stick-to-bottom functionality from a shared utility in `ui-components` to package-specific hooks in `instantsearch.js` and `react-instantsearch`.

### DIFF
--- a/packages/instantsearch-ui-components/src/lib/stickToBottom.ts
+++ b/packages/instantsearch-ui-components/src/lib/stickToBottom.ts
@@ -153,12 +153,13 @@ type CreateStickToBottomParams = {
     callback: T,
     deps: readonly unknown[]
   ) => T;
-  useEffect: (effect: () => void | (() => void), deps?: readonly unknown[]) => void;
+  useEffect: (
+    effect: () => void | (() => void),
+    deps?: readonly unknown[]
+  ) => void;
   useMemo: <TType>(factory: () => TType, deps: readonly unknown[]) => TType;
   useRef: <TType>(initialValue: TType) => { current: TType };
-  useState: <TType>(
-    initialState: TType
-  ) => [TType, (newState: TType) => void];
+  useState: <TType>(initialState: TType) => [TType, (newState: TType) => void];
 };
 
 export function createStickToBottom({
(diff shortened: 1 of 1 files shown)


packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -274,6 +276,159 @@ function createDefaultTools<
...
+    layoutComponent: ComponentProps<typeof Chat>['promptComponent'];
+    headerComponent: ChatPromptProps['headerComponent'];
+    footerComponent: ChatPromptProps['footerComponent'];
+    translations: Partial<ChatPromptTranslations>;
+    promptRef: { current: HTMLTextAreaElement | null };
+  };
+  state: ReturnType<typeof createLocalState>;
+};
+
+function ChatWrapper({
```
shaejaz, line 328: Can this not be done inline in the `render` fn?

packages/react-instantsearch/src/lib/useStickToBottom.ts
shaejaz, line 1: So if I understand this correctly, the approach now would be the same as the one at the very start. Except these hooks now import from react/preact based on where they are being used.

FabienMotte: That's correct, it's no longer a rewrite. There are just two hooks: one uses React dependencies and the other uses Preact dependencies.

packages/instantsearch.js/src/widgets/chat/chat.tsx
FabienMotte: I need to use the `useStickBottom` hook, so I need a component. I agree that's not ideal, as it means I have to redeclare all the props. Do you have a better idea?

shaejaz: Can you check if the hook can still be used in that case? Something like:
```javascript
render(() => {
  const s = useStickBottom();
  
  return (
    <Chat {...} />
  )
})
```

FabienMotte: I tried, but the whole chat re-renders when I type something into the chat prompt

PR #6772, opened 2025-10-24
test(autocomplete): add and move tests to cts

shaejaz:
FX-3550


shaejaz (commented):
Can't approve myself, but the cts js changes lgtm!

PR #6776, opened 2025-10-28
feat(autocomplete): implement styling

aymeric-giraudet:
**Summary**

[FX-3495](https://algolia.atlassian.net/browse/FX-3495)

Implemented light styling for what we have for now.
Created a new shared component just for search.
Couldn't really share variables with chat yet as we don't have a dark mode theme for Autocomplete yet, can be done later

**Result**

image

[FX-3495]: https://algolia.atlassian.net/browse/FX-3495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/common/widgets/autocomplete/options.tsx
+++ b/tests/common/widgets/autocomplete/options.tsx
@@ -4,6 +4,7 @@ import {
   createSingleSearchResponse,
 } from '@instantsearch/mocks';
 import { wait } from '@instantsearch/testutils';
+import { screen } from '@testing-library/dom';
 import userEvent from '@testing-library/user-event';
 
 import type { AutocompleteWidgetSetup } from '.';
@@ -81,12 +82,7 @@ export function createOptionsTests(
       });
 
       expect(document.querySelector('.ais-Autocomplete')).toBeInTheDocument();
-      expect(
-        document.querySelector('.ais-SearchBox-input')
-      ).toBeInTheDocument();
-      expect(
-        document.querySelector('.ais-SearchBox-loadingIndicator')
-      ).toHaveAttribute('hidden');
+      expect(screen.getByRole('search')).toBeInTheDocument();
       expect(
         document.querySelector('.ais-AutocompletePanel')
       ).toBeInTheDocument();
@@ -96,7 +92,12 @@ export function createOptionsTests(
 
       await act(async () => {
         // JS currently doesn't refine on focus
-        const input = document.querySelector('.ais-SearchBox-input')!;
+        const input =
+          flavor === 'javascript'
+            ? document.querySelector('.ais-SearchBox-input')!
+            : screen.getByRole('combobox', {
+                name: /submit/i,
+              });
         userEvent.click(input);
         userEvent.type(input, 'a');
         userEvent.clear(input);
@@ -149,7 +150,12 @@ export function createOptionsTests(
 
         // JS currently doesn't refine on focus
         if (flavor === 'javascript') {
-          const input = document.querySelector('.ais-SearchBox-input')!;
+          const input =
+            flavor === 'javascript'
+              ? document.querySelector('.ais-SearchBox-input')!
+              : screen.getByRole('combobox', {
+                  name: /submit/i,
+                });
           userEvent.click(input);
           userEvent.type(input, 'a');
           userEvent.clear(input);
@@ -277,7 +283,12 @@ export function createOptionsTests(
       await act(async () => {
         await wait(0);
         // JS currently doesn't refine on focus
-        const input = document.querySelector('.ais-SearchBox-input')!;
+        const input =
+          flavor === 'javascript'
+            ? document.querySelector('.ais-SearchBox-input')!
+            : screen.getByRole('combobox', {
+                name: /submit/i,
+              });
         userEvent.click(input);
         userEvent.type(input, 'a');
         userEvent.clear(input);
@@ -287,9 +298,12 @@ export function createOptionsTests(
         0
       );
 
-      const input = document.querySelector(
-        '.ais-SearchBox-input'
-      ) as HTMLInputElement;
+      const input: HTMLInputElement =
+        flavor === 'javascript'
+          ? document.querySelector('.ais-SearchBox-input')!
+          : screen.getByRole('combobox', {
+              name: /submit/i,
+            });
 
       await act(async () => {
         input.click();


packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSearch.tsx
```diff
@@ -0,0 +1,115 @@
...
+        noValidate
+        role="search"
+        onSubmit={(e) => e.preventDefault()}
+      >
+        <div className="ais-AutocompleteInputWrapperPrefix">
+          <label
+            className="ais-AutocompleteLabel"
+            aria-label="Submit"
+            htmlFor="autocomplete-input"
+            id="autocomplete-label"
```
dhayab: You should rely on `inputProps.id` for `htmlFor` and to compute the label's id here.

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSearch.tsx
```diff
@@ -0,0 +1,115 @@
...
+                viewBox="0 0 24 24"
+                width="20"
+                height="20"
+                fill="currentColor"
+              >
+                <path d="M16.041 15.856c-0.034 0.026-0.067 0.055-0.099 0.087s-0.060 0.064-0.087 0.099c-1.258 1.213-2.969 1.958-4.855 1.958-1.933 0-3.682-0.782-4.95-2.050s-2.050-3.017-2.050-4.95 0.782-3.682 2.050-4.95 3.017-2.050 4.95-2.050 3.682 0.782 4.95 2.050 2.050 3.017 2.050 4.95c0 1.886-0.745 3.597-1.959 4.856zM21.707 20.293l-3.675-3.675c1.231-1.54 1.968-3.493 1.968-5.618 0-2.485-1.008-4.736-2.636-6.364s-3.879-2.636-6.364-2.636-4.736 1.008-6.364 2.636-2.636 3.879-2.636 6.364 1.008 4.736 2.636 6.364 3.879 2.636 6.364 2.636c2.125 0 4.078-0.737 5.618-1.968l3.675 3.675c0.391 0.391 1.024 0.391 1.414 0s0.391-1.024 0-1.414z" />
+              </svg>
+            </button>
+          </label>
+          <div className="ais-AutocompleteLoadingIndicator" hidden>
```
dhayab: How is the loading indicator controlled?

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSearch.tsx
```diff
@@ -0,0 +1,115 @@
...
+                ></animateTransform>
+              </circle>
+            </svg>
+          </div>
+        </div>
+        <div className="ais-AutocompleteInputWrapper">
+          <input
+            className="ais-AutocompleteInput"
+            aria-autocomplete="both"
+            aria-labelledby="autocomplete-label"
```
dhayab: Tied to my earlier comment, this should rely on `inputProps.id`.

packages/instantsearch.css/src/components/autocomplete.scss
dhayab, line 1: This should probably be imported into algolia/satellite as well.

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSearch.tsx
```diff
@@ -0,0 +1,115 @@
...
+            >
+              <svg
+                className="ais-AutocompleteSubmitIcon"
+                viewBox="0 0 24 24"
+                width="20"
+                height="20"
+                fill="currentColor"
+              >
+                <path d="M16.041 15.856c-0.034 0.026-0.067 0.055-0.099 0.087s-0.060 0.064-0.087 0.099c-1.258 1.213-2.969 1.958-4.855 1.958-1.933 0-3.682-0.782-4.95-2.050s-2.050-3.017-2.050-4.95 0.782-3.682 2.050-4.95 3.017-2.050 4.95-2.050 3.682 0.782 4.95 2.050 2.050 3.017 2.050 4.95c0 1.886-0.745 3.597-1.959 4.856zM21.707 20.293l-3.675-3.675c1.231-1.54 1.968-3.493 1.968-5.618 0-2.485-1.008-4.736-2.636-6.364s-3.879-2.636-6.364-2.636-4.736 1.008-6.364 2.636-2.636 3.879-2.636 6.364 1.008 4.736 2.636 6.364 3.879 2.636 6.364 2.636c2.125 0 4.078-0.737 5.618-1.968l3.675 3.675c0.391 0.391 1.024 0.391 1.414 0s0.391-1.024 0-1.414z" />
+              </svg>
```
FabienMotte: I think the SVG icons could be moved to a file called `icons.tsx`, like I did with the chat icons. We could even put all the icons in a single file at the root of the package. It would also be good to check Lucide (https://lucide.dev/icons/) to see if they have similar icons. That way, all of the icons we use would come from the same library.

packages/instantsearch.css/src/components/autocomplete.scss
FabienMotte: Good point 👍 
Same for `themes/algolia.scss`

packages/instantsearch.css/src/components/autocomplete.scss
```diff
@@ -0,0 +1,329 @@
...
+  --ais-autocomplete-scrollbar-track-background-color-alpha: 1;
+  --ais-autocomplete-scrollbar-thumb-background-color-rgb: var(--ais-background-color-rgb);
+  --ais-autocomplete-scrollbar-thumb-background-color-alpha: 1;
+
+  // Touch screens
+  @media (hover: none) and (pointer: coarse) {
+    --ais-autocomplete-spacing-factor: 1.2;
+    --ais-autocomplete-action-icon-size: 22px;
+  }
+}
```
FabienMotte, line 82: I think we need to reuse or update the common CSS variables in `packages/instantsearch.css/src/shared/_variables.scss`, This way, we could have a dark mode for autocomplete too. We can also add autocomplete-specific CSS variables to this file, as I did for the chat-specific ones.

packages/instantsearch.css/src/components/autocomplete.scss
```diff
@@ -0,0 +1,329 @@
...
+
+// Reset for `@extend`
+%reset {
+  box-sizing: border-box;
+}
```
FabienMotte: This should already be handled by the `reset.scss`:
```css
[class^='ais-'] {
  box-sizing: border-box;
}
```
I also noticed that the same lines are duplicated in both themes and could be removed

packages/instantsearch.css/src/components/autocomplete.scss
```diff
@@ -0,0 +1,329 @@
...
+  background-color: rgba(
+    var(--ais-autocomplete-background-color-rgb),
+    var(--ais-autocomplete-background-color-alpha)
+  );
+  border-radius: calc(var(--ais-autocomplete-spacing) / 4);
+  box-shadow: var(--ais-autocomplete-panel-shadow);
+  margin: 8px 0 0;
+  padding: var(--ais-autocomplete-spacing-half);
+  overflow: hidden;
+  transition: opacity 200ms ease-in, filter 200ms ease-in;
```
FabienMotte: We have CSS variables for transition duration and easing function

FabienMotte (commented):
Nice work, thanks for taking care of it! 🙏

I noticed a few differences in behavior compared to Autocomplete.js that might be worth aligning or documenting:

### Enter key behavior

- **In InstantSearch**, when you focus the input, type a query, and hit Enter, the input stays focused while the panel closes. You need to manually unfocus and refocus the input to reopen the panel.

- **In Autocomplete.js**, hitting Enter while the input is focused causes it to lose focus (and the panel closes).

### Escape key behavior

- **In InstantSearch**, pressing Escape clears the query immediately.

- **In Autocomplete.js**, the first Escape press closes the panel; a second Escape press clears the query. If the panel is already closed (e.g., no results), Escape clears the query directly.

### Empty results behavior

- **In InstantSearch**, the panel remains visible even when there are no results.

### Clear button behavior

- **In InstantSearch**, clicking the “Clear” button removes focus from the input.

- **In Autocomplete.js**, the input remains focused.

### Panel positioning

- As discussed in a daily, the panel should be absolutely positioned.

- We should also plan to add support for detached mode later on.

packages/instantsearch.css/src/components/autocomplete.scss
aymeric-giraudet: I specifically chose not to because autocomplete is really integrated with other components which don't have a dark mode, thought it would look strange to have only one widget with dark mode :/ 
It works for chat cause it's just a floating action button and is a floating window, but for autocomplete I'm afraid we should really take time to have a dark mode theme for all widgets

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSearch.tsx
aymeric-giraudet: I put them in a separate file. Using lucide could be nice yes, but we should probably do this for all icons at some point

dhayab (approved):
Reran a test and design works well with latest changes.

Let's merge and iterate from there for:
- hiding the panel if there are 0 items
- replacing the searchbox with the autocomplete input component in JS widget

packages/instantsearch.css/src/components/autocomplete.scss
FabienMotte: Let's tackle that later 👍

PR #6777, opened 2025-10-29
feat(autocomplete): support global onSelect and redirect to search page

dhayab:
**Summary**

This PR adds two top-level parameters to autocomplete:
- `getSearchPageURL` to provide an URL to navigate to when autocomplete is not part of the search page
- `onSelect` to provide a global `onSelect`, making the per-index one optional, and defaulted to a method that handles redirection to the search page if necessary

The order is the following:
1. if the index / suggestion index provides `getURL()` → navigate to url
2. if autocomplete is not part of the search page and `getSearchPageURL()` exists → navigate to url
3. refine using `getQuery()`

FX-3538

### DIFF
--- a/packages/react-instantsearch/src/widgets/Autocomplete.tsx
+++ b/packages/react-instantsearch/src/widgets/Autocomplete.tsx
@@ -22,7 +22,7 @@ import {
   useSearchBox,
 } from 'react-instantsearch-core';
 
-import { SearchBox } from '../widgets/SearchBox';
+import { AutocompleteSearch } from '../components/AutocompleteSearch';
 
 import type {
   AutocompleteIndexClassNames,
@@ -192,7 +192,13 @@ function InnerAutocomplete<TItem extends BaseHit = BaseHit>({
 
   return (
     <Autocomplete {...props} {...getRootProps()}>
-      <SearchBox inputProps={getInputProps()} />
+      <AutocompleteSearch
+        inputProps={getInputProps()}
+        clearQuery={() => {
+          refineSearchBox('');
+          refineAutocomplete('');
+        }}
+      />
       <AutocompletePanel {...getPanelProps()}>
         {indices.map(({ indexId, hits }, index) => (
           <AutocompleteIndex
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -42,11 +42,11 @@
     },
     {
       "path": "./packages/instantsearch.css/themes/algolia.css",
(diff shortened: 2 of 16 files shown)


packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
```diff
@@ -161,6 +174,23 @@ function AutocompleteWrapper<TItem extends BaseHit>({
...
+        userOnSelect ??
+        (({ query, setQuery, url }) => {
+          if (url) {
+            window.location.href = url;
+            return;
+          }
+
+          if (!isSearchPage && typeof getSearchPageURL !== 'undefined') {
+            const indexUiState =
+              instantSearchInstance.getUiState()[targetIndex!.getIndexId()];
```
shaejaz, line 187: Is there a possibility this can cause an indexing error?

packages/react-instantsearch/src/widgets/Autocomplete.tsx
```diff
@@ -116,12 +126,17 @@ export function EXPERIMENTAL_Autocomplete<TItem extends BaseHit = BaseHit>({
...
+      getURL: showSuggestions.getURL as unknown as IndexConfig<TItem>['getURL'],
     });
   }
 
+  const isSearchPage = useMemo(
+    () =>
+      typeof indexRenderState.hits !== 'undefined' ||
+      typeof indexRenderState.infiniteHits !== 'undefined',
+    [indexRenderState]
+  );
```
shaejaz, line 138: Is it not possible to also get this info from the `targetIndex` in react?

shaejaz (approved):
lgtm, couple of minor comments.

packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
dhayab: targetIndex is at least the mainIndex (the parent of the isolated index of an autocomplete), so it should not cause an error. It just is typed as potentially undefined because the renderState needs to be declared before we have access to instantSearchInstance.

packages/react-instantsearch/src/widgets/Autocomplete.tsx
dhayab: For now this is defined in the JS widget, and retrieving the index instance in React is not as easy (`useIndexContext()` is not publicly exported for instance).

Could be worth having it returned by the connector in a future update though.

PR #6782, opened 2025-10-31
feat(autocomplete): support header template for autocomplete index

dhayab:
**Summary**

This PR provides a way for users to set a custom header on their autocomplete indices.

FX-3564

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -46,7 +46,7 @@
     },
     {
       "path": "./packages/instantsearch.css/themes/algolia-min.css",
-      "maxSize": "8 kB"
+      "maxSize": "8.25 kB"
     },
     {
       "path": "./packages/instantsearch.css/themes/reset.css",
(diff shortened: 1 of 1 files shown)


packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
```diff
@@ -33,10 +38,21 @@ export type AutocompleteIndexClassNames = {
...
+      HeaderComponent,
+      ItemComponent,
+      getItemProps,
+      classNames = {},
+    } = userProps;
 
     return (
       <div className={cx('ais-AutocompleteIndex', classNames.root)}>
+        {HeaderComponent && (
+          <div className={cx('ais-AutocompleteIndexHeader', classNames.header)}>
```
shaejaz, line 52: Since there's no default header, the class seems a bit redundant to me. This would likely be something the user handles in their component no?
Are there some default styles that we have in the css?

dhayab: This will allow us to provide basic layout styles similar to https://github.com/algolia/autocomplete/blob/next/packages/autocomplete-theme-classic/src/theme.scss#L437.

I'll update the PR with those styles once design is merged.

PR #6784, opened 2025-11-03
refactor(chat): move widget functionality to hook/connector

FabienMotte:
**Summary**

This PR moves chat widget functionality to the connector (IS.js)/hook (RIS) for better separation of concerns and reusability.

### DIFF
--- a/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
+++ b/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
@@ -11,47 +11,118 @@ import {
 } from '../../../../test/createWidget';
 import connectChat from '../connectChat';
 
+import type { UIMessage } from '../../../lib/chat';
+import type { InstantSearch, IndexWidget } from '../../../types';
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch/src/widgets/Chat.tsx
```diff
@@ -181,69 +176,25 @@ export function Chat<
...
-    clearError,
     error,
+    input,
+    setInput,
+    open,
+    setOpen,
+    isClearing,
+    clearMessages,
+    onClearTransitionEnd,
+    toolsForUi,
```
shaejaz: Since this is not internal to our widget and coming from the hook now, I think we could have a better name for it.

Maybe just `tools` but then they'll have to rename it every time.

Haroenv: I agree, it should be called tools, and the prop should be aliased or be props.tools

FabienMotte:
> looks good. I think this ticket originally also had as goal to use connectChat instead of useChat in react, have you tried that out? It can be in a follow-up PR

I haven't tried that yet. Let me do that in another PR

PR #6785, opened 2025-11-03
fix(refinementList): hide facets when when `sortBy` is provided

shaejaz:
**Summary**
[CR-8645]

This PR enables facets to still be hidden when they pass a `sortBy` based on the `renderingContent` from the engine.

[CR-8645]: https://algolia.atlassian.net/browse/CR-8645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/algoliasearch-helper/src/SearchResults/index.js
+++ b/packages/algoliasearch-helper/src/SearchResults/index.js
@@ -962,16 +962,16 @@ SearchResults.prototype.getFacetValues = function (attribute, opts) {
         var hide =
           facetOrdering && facetOrdering.hide ? facetOrdering.hide : [];
         if (hide.length > 0) {
-          var hidden = [];
+          var visible = [];
           items.forEach(function (item) {
             // hierarchical facets get sorted using their raw name
             var name = item.path || item.name;
             if (hide.indexOf(name) === -1) {
-              hidden.push(item);
+              visible.push(item);
             }
           });
 
-          return hidden;
+          return visible;
         }
 
         return items;


packages/algoliasearch-helper/src/SearchResults/index.js
```diff
@@ -948,16 +948,33 @@ SearchResults.prototype.getFacetValues = function (attribute, opts) {
...
           return sortViaFacetOrdering(data, facetOrdering);
         }
       }
 
       if (Array.isArray(options.sortBy)) {
         var order = formatSort(options.sortBy, SearchResults.DEFAULT_SORT);
-        return orderBy(data, order[0], order[1]);
+        var items = orderBy(data, order[0], order[1]);
+
+        var hide =
```
shaejaz, line 962: This hide logic is being repeated in three places now, so it might be a good idea to extract it out to a util func. That would be a bigger refactor, so maybe in another PR?

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -714,6 +714,9 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
       ) => {
         const state = helper!.state.setQueryParameters(userState);
 
+        // lastResults are always null at this point, we need to get them from
+        // the derived helper.
+        mainHelper.lastResults = derivedHelper!.lastResults;
```
Haroenv: can this change be split up in a separate PR?

Haroenv (commented):
makes sense, but let's add tests indeed

packages/instantsearch.js/src/widgets/index/index.ts
shaejaz: This is actually the reason why the hidden doesn't apply with `searchable`. In this part of the code: 
https://github.com/algolia/instantsearch/blob/a27123f46de052620989af72a182e48dd94c4b3e/packages/algoliasearch-helper/src/algoliasearch.helper.js#L440-L448

`this.lastResults` is null and no hide happens. So makes sense to keep this change in this PR no? I can add the reason for this in the code comment and/or the PR description to make it more clear tho (if keeping in this PR ofc).

Haroenv: I think the searchable and sortBy should be separate PRs so it's clear that those are two separate fixes, what do you think?

packages/algoliasearch-helper/src/SearchResults/index.js
```diff
@@ -948,16 +948,33 @@ SearchResults.prototype.getFacetValues = function (attribute, opts) {
...
+          facetOrdering && facetOrdering.hide ? facetOrdering.hide : [];
+        if (hide.length > 0) {
+          var hidden = [];
```
Haroenv: should be called visible I think

PR #6793, opened 2025-11-04
feat(chat): include in UMD build

Haroenv:
Did not remove the umd file yet, as it's used by vue instantsearch and algolia-experiences bundles, which didn't get a change in setup.

**Summary**

**Result**

### DIFF
--- a/packages/instantsearch.js/scripts/rollup/emptyModule.js
+++ b/packages/instantsearch.js/scripts/rollup/emptyModule.js
@@ -2,17 +2,95 @@
 export const ZodFirstPartyTypeKind = {};
 export const toJSONSchema = {};
 export const safeParseAsync = {};
-export const z = new Proxy(
-  {},
-  {
-    get: (_, name) =>
-      name === '~standard'
-        ? { validate: (value) => Promise.resolve({ issues: null, value }) }
-        : function () {
-            return this;
-          },
-  }
-);
+export const z = {
+  string() {
+    return this;
+  },
+  instanceof() {
+    return this;
+  },
+  custom() {
+    return this;
+  },
+  object() {
+    return this;
+  },
+  array() {
+    return this;
+  },
+  union() {
+    return this;
+  },
+  literal() {
+    return this;
+  },
+  lazy() {
+    return this;
+  },
+  null() {
+    return this;
+  },
+  number() {
+    return this;
+  },
+  boolean() {
+    return this;
+  },
+  record() {
+    return this;
+  },
+  optional() {
+    return this;
+  },
+  unknown() {
+    return this;
+  },
+  discriminatedUnion() {
+    return this;
+  },
+  strictObject() {
+    return this;
+  },
+  startsWith() {
+    return this;
+  },
+  looseObject() {
+    return this;
+  },
+  loose() {
+    return this;
+  },
+  base64() {
+    return this;
+  },
+  extend() {
+    return this;
+  },
+  default() {
+    return this;
+  },
+  or() {
+    return this;
+  },
+  int() {
+    return this;
+  },
+  merge() {
+    return this;
+  },
+  strict() {
+    return this;
+  },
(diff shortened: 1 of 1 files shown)


aymeric-giraudet:
Got it to work and tested on `ecommerce-with-umd`, works fine even with this minimal zod implem for what we're doing.
May still need some testing.
Also this increases the bundle size by 35% even without zod :/

packages/instantsearch.js/.storybook/decorators/withHits.ts
```diff
@@ -4,7 +4,7 @@ import {
   liteClient as namedConstructor,
   default as defaultConstructor,
 } from 'algoliasearch/lite';
-import instantsearch from '../../src';
+import type instantsearch from '../../src';
```
Haroenv: instantsearch is used as more than a type, maybe the import needs to be the umd version for storybook to avoid chat?

packages/instantsearch.js/scripts/rollup/rollup.config.js
```diff
@@ -16,6 +19,25 @@ const link = 'https://github.com/algolia/instantsearch';
...
+      },
+      {
+        find: /^react.*/,
+        replacement: path.join(__dirname, './emptyModule.js'),
+      },
+      {
+        find: 'eventsource-parser/stream',
+        replacement: path.join(
+          __dirname,
+          '../../../../node_modules/eventsource-parser/dist/stream.js'
```
Haroenv, line 36: wonder if using `require.resolve` may make sense?

Haroenv: no it doesn't as this is an esm file.

packages/instantsearch.js/.storybook/decorators/withHits.ts
```diff
@@ -4,13 +4,14 @@ import {
   liteClient as namedConstructor,
   default as defaultConstructor,
 } from 'algoliasearch/lite';
-import type instantsearch from '../../src';
+import instantsearch from '../../src/index.es';
```
Haroenv, line 7: this needs to be the umd one for storybook, we use InstantSearch.widgets everywhere. Otherwise you can also change all of the imports later to use real imports instead of .widgets in the actual stories

aymeric-giraudet: Yeah I saw, the UMD file is for `src/widgets/index` though, not for `src/index`
I have to replace all imports in the stories then

packages/instantsearch.js/scripts/rollup/emptyModule.js
```diff
@@ -0,0 +1,19 @@
+// Placeholder export for 'zod' module
+export const ZodFirstPartyTypeKind = {};
+export const toJSONSchema = {};
+export const safeParseAsync = {};
+export const z = new Proxy(
```
Haroenv: at which point is this evaluated? do we need to check if proxy exists first in case it gets evaluated in IE11?

Haroenv: I'm ok with chat not working in IE11, but just ensuring that the entire InstantSearch doesn't throw

aymeric-giraudet: Yeah the constructor does get called directly. There are polyfills though if we still actually care about the tests passing ?

Haroenv: We may need to do `typeof Proxy === 'object' ? new Proxy : null` or similar to at least not have a global error thrown then? not sure how to check tbh, just asking

aymeric-giraudet: It would still throw though cause zod schemas are created when the script is evaluated... so we'd have to define all those functions just for IE11 to work and make the bundle bigger

PR #6794, opened 2025-11-05
feat: add fallbackParameters support to Recommend FBT [RECO-2443]

raed667:
**Summary**

Add `fallbackParameters` support to Recommend Frequently bought together model.

Follow-up on: https://github.com/algolia/api-clients-automation/pull/5579


packages/instantsearch.js/src/lib/__tests__/InstantSearch-integration-test.ts
```diff
@@ -229,6 +229,7 @@ describe('network requests', () => {
       ).toMatchInlineSnapshot(`
         [
           {
+            "fallbackParameters": undefined,
```
raed667, line 232: should we remove the key completely instead of sending undefined ?  🤔

PR #6796, opened 2025-11-05
refactor(chat): use `connectChat` connector in `useChat` hook

FabienMotte:
**Summary**

This PR refactors `useChat` to use the `connectChat` connector with the `useConnector` hook.

Follow-up to #6784

### DIFF
--- a/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
+++ b/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
@@ -278,7 +278,10 @@ describe('connectChat', () => {
 
       const renderState = getRenderState();
       expect(renderState.tools).toEqual({
-        testTool: mockTool,
+        testTool: {
+          ...mockTool,
+          addToolResult: expect.any(Function),
+        },
       });
     });
   });


packages/react-instantsearch/src/widgets/Chat.tsx
```diff
@@ -187,14 +184,23 @@ export function Chat<
...
+  const toolsForUi = React.useMemo(() => {
+    const result: Record<string, any> = {};
+    Object.keys(tools).forEach((key) => {
+      result[key] = {
+        ...tools[key],
+        addToolResult,
+      };
+    });
+    return result;
+  }, [tools, addToolResult]);
```
shaejaz: Does this no longer work in the connector?
I think it was shifted to the hook in the previous PR right?

FabienMotte: We can remove this, but the `tools` coming from the `useChat` hook cannot be used for the `ChatUiComponent` because the type is incorrect. This is because we are missing `layoutComponent` and `addToolResults` (we add these in the widget for IS.js).

- The `layout` template is optional for IS.js, but `layoutComponent` is required for RIS. Do we want to address this, or is it expected?
- Should we move the `tools` with `addToolResult` in the `connectChat` connector?

cc @Haroenv

Haroenv: yes, I think it should be in the connector then, sounds good. should layout be required for a tool? seems like there could be tools that display nothing right?

FabienMotte: Thanks! Done in https://github.com/algolia/instantsearch/pull/6796/commits/bebb533986be10991e9496da69b372b6c140c279

Haroenv (approved):

PR #6797, opened 2025-11-05
feat(chat): add assistant and user message components/templates in widget

shaejaz:
**Summary**
This PR adds support for adding the leading(avatar) and footer components for assistant and user messages. Also added support for message translations and classNames.

### DIFF
--- a/packages/react-instantsearch/src/widgets/Chat.tsx
+++ b/packages/react-instantsearch/src/widgets/Chat.tsx
@@ -21,7 +21,6 @@ import type {
   UserClientSideTool,
   UserClientSideTools,
   ChatMessageProps,
-  ChatMessagesProps,
 } from 'instantsearch-ui-components';
 import type { IndexUiState } from 'instantsearch.js';
 import type { UIMessage } from 'instantsearch.js/es/lib/chat';
@@ -119,7 +118,6 @@ export type ChatProps<TObject, TUiMessage extends UIMessage = UIMessage> = Omit<
     promptHeaderComponent?: ChatUiProps['promptProps']['headerComponent'];
     promptFooterComponent?: ChatUiProps['promptProps']['footerComponent'];
     actionsComponent?: ChatUiProps['messagesProps']['actionsComponent'];
-    messageComponent?: ChatMessagesProps['messageComponent'];
     assistantMessageLeadingComponent?: ChatMessageProps['leadingComponent'];
     assistantMessageFooterComponent?: ChatMessageProps['footerComponent'];
     userMessageLeadingComponent?: ChatMessageProps['leadingComponent'];
@@ -155,7 +153,6 @@ export function Chat<
   promptComponent,
   promptHeaderComponent,
   promptFooterComponent,
-  messageComponent,
   assistantMessageLeadingComponent,
   assistantMessageFooterComponent,
   userMessageLeadingComponent,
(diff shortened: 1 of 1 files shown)


packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -868,6 +957,34 @@ export type ChatTemplates<THit extends NonNullable<object> = BaseHit> =
...
+    userMessage: Partial<{
+      /**
+       * Template to use for the user message leading content.
+       */
+      leading: Template;
+      /**
+       * Template to use for the user message footer content.
+       */
+      footer: Template;
+    }>;
```
shaejaz, line 999: Not sure what's the best API for js templates. We already have a `messages` template so it could be better to have these there with something like:
```javascript
chat({
  templates: {
    messages: {
      assistant: {
        leading: 'Leading'
        // rest
      }
    }
  }
})
```
This might be too much nesting tho.

packages/react-instantsearch/src/widgets/Chat.tsx
```diff
@@ -112,6 +117,11 @@ export type ChatProps<TObject, TUiMessage extends UIMessage = UIMessage> = Omit<
     promptHeaderComponent?: ChatUiProps['promptProps']['headerComponent'];
     promptFooterComponent?: ChatUiProps['promptProps']['footerComponent'];
     actionsComponent?: ChatUiProps['messagesProps']['actionsComponent'];
+    messageComponent?: ChatMessagesProps['messageComponent'];
```
shaejaz: I've added the message component here for now, but I'm not sure if having it would be beneficial to a user. They would have to implement a lot of things with a custom component such as: tool handling, actions components, rendering markdown etc. Wdyt?

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -402,6 +411,17 @@ function ChatWrapper({
...
+          ...messagesProps.assistantMessageProps,
+          leadingComponent:
+            messagesProps.assistantMessageProps.leadingComponent,
+          footerComponent: messagesProps.assistantMessageProps.footerComponent,
+        },
+        userMessageProps: {
+          ...messagesProps.userMessageProps,
+          leadingComponent: messagesProps.userMessageProps.leadingComponent,
+          footerComponent: messagesProps.userMessageProps.footerComponent,
+        },
```
FabienMotte:
```suggestion
        assistantMessageProps: messagesProps.assistantMessageProps,
        userMessageProps: messagesProps.userMessageProps,
```
Could it be simplified?

packages/react-instantsearch/src/widgets/Chat.tsx
FabienMotte: For now, I think we can keep it simple and remove the `messageComponent` prop. We can always add it back if we receive user requests. WDYT?

packages/instantsearch.js/src/widgets/chat/chat.tsx
FabienMotte: For clarity, I think we should avoid nesting levels too deeply

Haroenv (commented):
looks good once test is fixed

packages/react-instantsearch/src/widgets/Chat.tsx
shaejaz: Sounds good, I've gone ahead and removed it now.

PR #6800, opened 2025-11-06
fix(autocomplete): trigger initial refine in js implementation

dhayab:
**Summary**

This PR fixes a discrepancy between JS and React implementations of the Autocomplete widget. In JS, it now triggers an initial search request with the `query` from the target index.

I also moved the default `hitsPerPage` to the configure of the actual index, the initial refine solving the issue that prompted to set `hitsPerPage` to a higher level.

FX-3571

### DIFF
--- a/packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
+++ b/packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
@@ -125,7 +125,7 @@ const createRenderer = <TItem extends BaseHit>(
         targetIndex,
       };
 
-      connectorParams.refine(targetIndex?.getHelper()?.state.query ?? '');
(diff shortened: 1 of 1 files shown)


packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
```diff
@@ -124,6 +124,8 @@ const createRenderer = <TItem extends BaseHit>(
         isolatedIndex,
         targetIndex,
       };
+
+      connectorParams.refine(targetIndex?.getHelper()?.state.query ?? '');
```
dhayab:
```suggestion
      connectorParams.refine(targetIndex.getHelper()?.state.query ?? '');
```

Haroenv (approved):

PR #6804, opened 2025-11-12
feat(autocomplete): forward search params to the query API request

shaejaz:
**Summary**
[FX-3568]

This PR adds support to forward user provided search parameters for a query made in autocomplete. This is done via the new `searchParameters` base prop in widget as well as another `searchParameters` prop in each item in `indices`.

A fix has been made for isolated index widgets where instead of calling the helper.search immediately, it is deferred instead to allow the configure widgets to have time to mount and update the state (mainly for the react widget).

Docs PR has been linked to review the API changes easily.

[FX-3568]: https://algolia.atlassian.net/browse/FX-3568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react-instantsearch/src/widgets/__tests__/Autocomplete.test.tsx
+++ b/packages/react-instantsearch/src/widgets/__tests__/Autocomplete.test.tsx
@@ -241,8 +241,8 @@ describe('Autocomplete', () => {
 
     await screen.findByText('hello');
 
-    expect(searchClient.search).toHaveBeenCalledTimes(4);
-    expect(searchClient.search).toHaveBeenNthCalledWith(3, [
+    expect(searchClient.search).toHaveBeenCalledTimes(2);
+    expect(searchClient.search).toHaveBeenNthCalledWith(2, [
       {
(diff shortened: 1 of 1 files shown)


tests/common/widgets/autocomplete/options.tsx
```diff
@@ -140,7 +140,7 @@ export function createOptionsTests(
       });
 
       const callTimes: Record<string, Record<SupportedFlavor, number>> = {
-        initial: { javascript: 2, react: 4, vue: 0 },
+        initial: { javascript: 2, react: 2, vue: 0 },
         refined: { javascript: 1, react: 2, vue: 0 },
```
shaejaz, line 144: Ideally for react this should also only be 1 instead of 2 like in js but not sure why this is the case still. I don't think it's related to changes in this PR so should be investigated later.

Haroenv (approved):

PR #6806, opened 2025-11-13
feat(autocomplete): implement reverse highlighting

aymeric-giraudet:
**Summary**

[FX-3574](https://algolia.atlassian.net/browse/FX-3574)

**Result**

Works similarly as autocomplete, if query is empty, we don't show items as reverse highlighted, it only renders once there's a query.

[FX-3574]: https://algolia.atlassian.net/browse/FX-3574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch-ui-components/src/components/autocomplete/createAutocompleteStorage.ts
+++ b/packages/instantsearch-ui-components/src/components/autocomplete/createAutocompleteStorage.ts
@@ -111,7 +111,6 @@ function getHighlightedAttribute({
   return {
     query: {
       value: item.query.replace(
-        // codacy-disable-next-line security/detect-non-literal-regexp
         new RegExp(query.replace(/[-/\\^$*+?.()|[\]{}]/g, '\\$&'), 'gi'),
         (match) => {
           return `<mark>${match}</mark>`;


packages/react-instantsearch/src/widgets/Autocomplete.tsx
```diff
@@ -145,7 +147,13 @@ export function EXPERIMENTAL_Autocomplete<TItem extends BaseHit = BaseHit>({
...
         showSuggestions.headerComponent as unknown as AutocompleteIndexProps<TItem>['HeaderComponent'],
       itemComponent: (showSuggestions.itemComponent ||
-        AutocompleteSuggestion) as unknown as AutocompleteIndexProps<TItem>['ItemComponent'],
+        (({ item, onSelect }: Parameters<typeof AutocompleteSuggestion>[0]) => (
+          <AutocompleteSuggestion item={item} onSelect={onSelect}>
+            <ConditionalReverseHighlight
+              item={item as unknown as Hit<{ query: string }>}
+            />
+          </AutocompleteSuggestion>
+        ))) as unknown as AutocompleteIndexProps<TItem>['ItemComponent'],
```
shaejaz, line 166: Is this assertion needed?

aymeric-giraudet: Yeah cause the generics don't have query as required, we're just telling TypeScript we know we have query here as we constructed the index ourselves

PR #6807, opened 2025-11-17
feat(autocomplete): support panel customization

dhayab:
**Summary**

This PR adds support for panel customization, through templates in JS and a component prop in React.

Props are:
- `elements`: an key-based object to retrieve the rendered output of `recent` searches, `suggestions`, and indices (similar to [what Autocomplete provides](https://www.algolia.com/doc/ui-libraries/autocomplete/api-reference/autocomplete-js/autocomplete#param-render))
- `indices`: the object returned by the autocomplete connector (doesn't include recent searches)

### InstantSearch.js

```js
autocomplete({
  // ...
  templates: {
    panel: ({ elements, indices }, { html }) => html`
      <div class="left">
        ${elements.recent}
        ${elements.suggestions}
      </div>
      <div class="right">
        ${elements.instant_search_movies}
      </div>
    `,
  },
}),
```

### React InstantSearch

```jsx
<Autocomplete
  panelComponent={({ elements, indices }) => (
    <>
      <div className="left">
        {elements.recent}
        {elements.suggestions}
      </div>
      <div className="right">
        {elements.instant_search_movies}
      </div>
    </>
  )}
/>
```

FX-3563

### DIFF
--- a/tests/common/widgets/autocomplete/templates.tsx
+++ b/tests/common/widgets/autocomplete/templates.tsx
@@ -180,6 +180,7 @@ export function createTemplatesTests(
             ],
             showSuggestions: {
               indexName: 'indexName2',
+              // @ts-expect-error itemComponent can return a string
               itemComponent: (props) => props.item.query,
             },
             showRecent: true,


Haroenv (approved):
🥇

PR #6809, opened 2025-11-18
feat(css): add devtools for CSS variable config

FabienMotte:
**Summary**

[FX-3569](https://algolia.atlassian.net/browse/FX-3569)

This PR introduces a developer tools panel for InstantSearch CSS that enables real-time configuration and experimentation with CSS variables using [Tweakpane](https://tweakpane.github.io/docs/) library.

Changes:
- Adds Tweakpane integration for interactive CSS variable configuration
- Implements automatic CSS variable config generation from SCSS source files
- Provides a `createInstantSearchDevtools()` API for embedding the panel
- Provides an inject module for automatic devtools initialization

**Result**

image

[FX-3569]: https://algolia.atlassian.net/browse/FX-3569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.css/devtools/index.js
+++ b/packages/instantsearch.css/devtools/index.js
@@ -493,7 +493,7 @@ function getNonColorVariables(variables) {
  * @returns {string} CSS text with variable definitions
  */
 function exportCSSVariables(values, config) {
-  let output = `// CSS Variables\n:root {\n`;
+  let output = `/* CSS Variables */\n:root {\n`;
 
   config.variables.forEach((variable) => {
     if (variable.type === 'color-rgb') {


examples/js/getting-started/src/app.js
FabienMotte, line 1: We probably don't want to add it to our getting started examples, WDYT?

packages/instantsearch.css/devtools/index.js
FabienMotte, line 1: Are we fine using JS with declaration files, or do we prefer TS?

examples/js/getting-started/package.json
```diff
@@ -13,7 +13,8 @@
     "algoliasearch-helper": "3.26.1",
     "instantsearch-ui-components": "0.14.0",
     "instantsearch.css": "8.8.0",
-    "instantsearch.js": "4.83.0"
+    "instantsearch.js": "4.83.0",
+    "tweakpane": "4.0.5"
```
Haroenv: should be a dependency of InstantSearch.css I think

examples/js/getting-started/src/app.js
Haroenv: maybe the default-theme examples?

packages/instantsearch.css/devtools/index.js
```diff
@@ -0,0 +1,391 @@
+import variablesConfig from './variables-config.json';
```
Haroenv: we're importing a json here, not all tools support that. Do you think it's possible to inline inside this file to make it self-contained? although that might be a bit tough with both the inject and index file. 

Should we support only the inject style?

packages/instantsearch.css/devtools/index.js
Haroenv: I think we'll want `// @ts-check` then?

FabienMotte: Good point

packages/instantsearch.css/devtools/index.js
FabienMotte: The script now generates the CSS variables configuration in the `index.js` file. Both `createInstantSearchDevtools` and inject import are working

FabienMotte:
> let's add to the js default theme one too?

I don't think we have a default theme example for JS? Should I create it?

PR #6811, opened 2025-11-20
poc: prompt suggestions with separate agent

shaejaz:
**Summary**

**Result**

### DIFF
--- a/packages/react-instantsearch/src/widgets/PromptSuggestions.tsx
+++ b/packages/react-instantsearch/src/widgets/PromptSuggestions.tsx
@@ -1,5 +1,5 @@
-import React, { useEffect, useState } from 'react';
-import { useChat, useHits } from 'react-instantsearch-core';
+import React from 'react';
+import { useHits, usePromptSuggestions } from 'react-instantsearch-core';
 
 type PromptSuggestionsUiProps = {
   agentId: string;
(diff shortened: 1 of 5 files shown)


packages/react-instantsearch/src/widgets/PromptSuggestions.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+type PromptSuggestionsUiProps = {
+  agentId: string;
+};
+
+export function PromptSuggestions({ agentId }: PromptSuggestionsUiProps) {
+  const { items } = useHits();
+  const [suggestions, setSuggestions] = useState<Array<{ title: string }>>([]);
+
+  const { messages, status, sendMessage } = useChat({ agentId });
```
Haroenv: is useChat the right primitive for this? is there a different /chat and /completion endpoint?

shaejaz: I haven't come across a /chat endpoint yet so I've gone for the completions one.
I guess with some checking and cleanup code I can make it work similar to a chat endpoint. Not ideal but as a temp solution for now.

Haroenv: ah right, I didn't realise we already use the completions endpoint for chat as well. Seems logical then like this! the question is then if long term we should use useChat, or something lower level that doesn't imply the other functionality of chat like inputs and sending messages, but a static message instead.

PR #6814, opened 2025-11-24
refactor(css): refactor autocomplete to use common css vars

FabienMotte:
**Summary**

This PR:

- Refactors the Autocomplete component styles to use common CSS variables.
- Fixes the InstantSearch.css DevTools exports and adds support for new dimensions.
- Fixes some styles.
- Disables the dark mode media query for now, until all widgets support dark mode.

**Result**

image

### DIFF
--- a/packages/instantsearch.css/devtools/index.js
+++ b/packages/instantsearch.css/devtools/index.js
@@ -780,8 +780,12 @@ export function createInstantSearchDevtools(options = {}) {
                 // Convert px to rem
                 cssValue = `${Number(ev.value) / 16}rem`;
               } else {
-                // Append unit if it exists, otherwise just the value
-                cssValue = `${ev.value}${variable.control.unit || ''}`;
+                // Append unit if it exists and is not 'unitless', otherwise just the value
+                const unit =
+                  variable.control.unit && variable.control.unit !== 'unitless'
+                    ? variable.control.unit
+                    : '';
+                cssValue = `${ev.value}${unit}`;
               }
 
               updateCSSVariable(variable.name, cssValue);


packages/instantsearch.css/src/components/button.scss
```diff
@@ -15,9 +15,9 @@
   margin: 0;
   background-color: transparent;
   border-radius: var(--ais-border-radius-sm);
-  font-family: inherit;
-  font-size: inherit;
-  line-height: inherit;
+  font-family: var(--ais-font-family);
```
Haroenv: is the --ais-font-family needed? could that not be in general inherited in inputs and buttons too?

packages/instantsearch-ui-components/src/components/autocomplete/AutocompletePanel.tsx
```diff
@@ -22,13 +22,14 @@ export type AutocompletePanelClassNames = {
...
   return function AutocompletePanel(userProps: AutocompletePanelProps) {
-    const { children, classNames = {}, ...props } = userProps;
+    const { children, classNames = {}, hidden, ...props } = userProps;
 
     return (
       <div
         {...props}
         className={cx(
           'ais-AutocompletePanel',
+          !hidden && 'ais-AutocompletePanel--open',
```
Haroenv, line 32: what's this change for?

Haroenv (commented):
👌

packages/instantsearch.css/src/components/button.scss
FabienMotte: Good point, I removed it

packages/instantsearch-ui-components/src/components/autocomplete/AutocompletePanel.tsx
FabienMotte: This allows us to have a CSS transition (opacity/scale), just like we have for the chat widget

https://github.com/user-attachments/assets/e4346f17-96a5-446e-9304-99a1f7814ea1

Haroenv:
yes, as that's the WAI-ARIA suggestion. The mouse is only only used for clicking and the keyboard for selection.

PR #6816, opened 2025-11-26
feat(autocomplete): add header to showRecents

Haroenv:
**Summary**

Adds the option `header` to templates and `headerComponent` for `showRecent`

**Result**

show recent can have a header

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -14,7 +14,7 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "193.5 kB"
+      "maxSize": "200 kB"
     },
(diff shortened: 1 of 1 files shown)


packages/react-instantsearch/src/widgets/Autocomplete.tsx
```diff
@@ -195,6 +211,45 @@ export function EXPERIMENTAL_Autocomplete<TItem extends BaseHit = BaseHit>({
     });
   }
 
+  const recentSearchConfig = showRecent
```
shaejaz, line 215: Not completely familiar with the conventions, but shouldn't classNames assignment be done in the ui-components and just be passed on from the widget?

PR #6820, opened 2025-11-27
fix(autocomplete): keep input focused when clicking inside panel + more specific React keys

aymeric-giraudet:
**Summary**

[FX-3593](https://algolia.atlassian.net/browse/FX-3593)

**Result**

Same as Autocomplete, just have to call `preventDefault` on `mouseDown` events.
Also now using element ID + `objectID` to be extra sur it's not repeated.

[FX-3593]: https://algolia.atlassian.net/browse/FX-3593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


FabienMotte (approved):
Nice! 💯

PR #6822, opened 2025-11-28
feat(chat): use facet values for refining with search index tool

shaejaz:
**Summary**
This PR updates the default search index tool's 'View All' button to support `facet_filters` coming as an input from Agent Studio (including the `query` which is already supported).

It would use either the `refinementList` or `menu` to add these filters. No other search params are supported.

### DIFF
--- a/tests/common/widgets/chat/options.tsx
+++ b/tests/common/widgets/chat/options.tsx
@@ -3,7 +3,7 @@ import { createSearchClient } from '@instantsearch/mocks';
 import { wait } from '@instantsearch/testutils';
 import userEvent from '@testing-library/user-event';
 import { Chat, SearchIndexToolType } from 'instantsearch.js/es/lib/chat';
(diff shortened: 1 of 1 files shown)


packages/instantsearch.js/src/lib/utils/getStateFromSearchToolInput.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  if (input.facet_filters) {
+    let uiStateToUpdate: 'refinementList' | 'menu' | null = null;
+
+    if (indexUiState.refinementList) {
+      uiStateToUpdate = 'refinementList';
+    } else if (indexUiState.menu) {
+      uiStateToUpdate = 'menu';
+    }
```
Haroenv, line 20: have you considered something like `index.getWidgetSearchParameters( currentUiState)`?

PR #6824, opened 2025-12-01
feat(autocomplete): deduplicate recent searches/suggestions

aymeric-giraudet:
**Summary**

[FX-3586](https://algolia.atlassian.net/browse/FX-3586)

**Result**

Added an assertion to make sure it is deduped.
Also fixed a bug where the panel was always opened + the React key problem was actually from showRecent

[FX-3586]: https://algolia.atlassian.net/browse/FX-3586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch-ui-components/src/components/autocomplete/AutocompletePanel.tsx
+++ b/packages/instantsearch-ui-components/src/components/autocomplete/AutocompletePanel.tsx
@@ -27,7 +27,7 @@ export function createAutocompletePanelComponent({ createElement }: Renderer) {
     return (
       <div
         {...props}
-        hidden={hidden}
+        aria-hidden={hidden}
         className={cx(
           'ais-AutocompletePanel',
           !hidden && 'ais-AutocompletePanel--open',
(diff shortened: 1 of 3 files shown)


packages/instantsearch-ui-components/src/components/autocomplete/AutocompletePanel.tsx
```diff
@@ -27,6 +27,7 @@ export function createAutocompletePanelComponent({ createElement }: Renderer) {
     return (
       <div
         {...props}
+        hidden={hidden}
```
FabienMotte: I removed it to use a CSS transition instead when opening or closing:
`!hidden && 'ais-AutocompletePanel--open',`

aymeric-giraudet: Can't we have both `hidden` + a transition ? Else I can remove it. It's just a bit more work for users now if they implement their own CSS but it's probably okay

aymeric-giraudet: Okay I replaced it with an `aria-hidden` prop instead

FabienMotte: Perfect, thanks!

PR #6827, opened 2025-12-02
poc(chat): add refinement suggestions to search index tool

FabienMotte:
**Summary**

[FX-3608](https://algolia.atlassian.net/browse/FX-3608)

This PR:
- Updates the default `algolia_search_index` tool to display the suggested refinements after the hits carousel using pills. I added a loader for the tool (we are missing it in the current implementation).
- Uses the `sendMessage` function to apply the selected refinement as part of the conversation. This triggers the `algolia_search_index` tool again with the new refinement applied (using the `facetFilters` parameter in the tool input).

**Result**

https://github.com/user-attachments/assets/6a3124eb-aea8-4cbe-ad38-b79faf9805b5

[FX-3608]: https://algolia.atlassian.net/browse/FX-3608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/widgets/chat/chat.tsx
+++ b/packages/instantsearch.js/src/widgets/chat/chat.tsx
@@ -122,14 +122,9 @@ function createCarouselTool<
     const items = output?.hits || [];
     const suggestedFilters = output?.suggestedFilters || [];
 
-    const handleFilterClick = (
-      attribute: string,
-      value: string,
-      isRefined: boolean
-    ) => {
-      const action = isRefined ? 'Remove' : 'Apply';
+    const handleFilterClick = (attribute: string, value: string) => {
       sendMessage({
-        text: `${action} the ${attribute} filter: ${value}`,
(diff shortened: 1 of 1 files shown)


packages/instantsearch-ui-components/src/components/FilterPill.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <button
+        className={cx(
+          'ais-FilterPill',
+          isRefined && 'ais-FilterPill--refined',
+          className
+        )}
+        onClick={onClick}
+        type="button"
+      >
+        <span className="ais-FilterPill-label">{label}:</span>
```
Haroenv: maybe the case in other places too, but these classes aren't in cssClasses/classNames

FabienMotte: Good catch, thanks!

PR #6828, opened 2025-12-03
feat(promptSuggestions): add prompt suggestions widget

shaejaz:
**Summary**
Adds promptSuggestions widget in react and js and the ui components

TODO:
- tests


packages/instantsearch.js/src/lib/chat/chat.ts
```diff
@@ -76,6 +76,10 @@ export class ChatState<TUiMessage extends UIMessage>
     this._callMessagesCallbacks();
   }
 
+  get data(): unknown {
```
shaejaz, line 94: This is an alternative approach that would be similar to the dashboard POC. I've left it here in case the integration with the backend requires this approach instead of the `onData` one.

PR #6829, opened 2025-12-05
fix(ui-components): components don't inline destructure

Haroenv:
**Summary**

Expand the "export validation" for all ui components.

This otherwise causes issues with ComponentProps being undefined for users who don't have JSX instantiated. The tests already existed, but weren't running for the new files that are in folders.

Also add a new test for a different case where ComponentProps can be unknown and thus cause type errors.

**Result**

no destructures of "hidden" types in the type files of instantsearch-ui-components

### DIFF
--- a/packages/instantsearch-ui-components/__tests__/types.test.ts
+++ b/packages/instantsearch-ui-components/__tests__/types.test.ts
@@ -96,20 +96,7 @@ function validateIcons(
   sourceFile: ts.SourceFile,
   report: (node: ts.Node) => (message: string) => void
 ) {
-  function checkFunction(functionDeclaration: ts.FunctionDeclaration) {
-    if (
-      !(
-        functionDeclaration.modifiers?.some(
-          (modifier) => modifier.kind === ts.SyntaxKind.ExportKeyword
-        ) &&
-        functionDeclaration.modifiers?.some(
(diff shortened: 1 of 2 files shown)


packages/instantsearch-ui-components/__tests__/types.test.ts
```diff
@@ -90,34 +37,156 @@ function delint(sourceFile: ts.SourceFile) {
...
 
-const files = fs
-  .readdirSync(path.join(__dirname, '../dist/es/components'))
-  .filter((file) => file.endsWith('.d.ts') && !file.startsWith('index'));
+walkDir(componentsDir);
 
 describe('Exposes correct types', () => {
-  describe.each(files)('%s', (file) => {
+  describe.each(files)('$subdir/$file', ({ file, subdir }) => {
     it('has no errors', () => {
```
shaejaz, line 240: I'm also wondering if something like this would be possible with custom eslint rules. Would be better in terms of dx if all checks are possible there. Not now ofc, something to keep in mind later.

Haroenv: Yes, probably would be better but this was already introduced in the past so was easier to amend it than restart.

PR #6830, opened 2025-12-05
fix(chat): add chat widget render state

FabienMotte:
**Summary**

This PR fixes the Chat widget render state.

### DIFF
--- a/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
+++ b/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
@@ -121,6 +121,70 @@ describe('connectChat', () => {
     });
   });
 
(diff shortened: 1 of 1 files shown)


packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -262,6 +268,7 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
             },
             false
           );
+          instantSearchInstance.scheduleRender(false);
```
Haroenv: this means that every chat render causes all InstantSearch widgets to be rendered. Is the goal to ensure the render state is saved? I wonder if we could have a function on the index widget to render just a single widget which does the storing of the render state as well?

I'm just wondering as previously this was the cause of the slowdown in chat connector

PR #6831, opened 2025-12-08
feat(chat): support prompt suggestions

aymeric-giraudet:
**Summary**

[FX-3596](https://algolia.atlassian.net/browse/FX-3596)

**Result**

https://github.com/user-attachments/assets/5657c251-fad4-494f-9e23-74402e73a928

[FX-3596]: https://algolia.atlassian.net/browse/FX-3596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/lib/chat/__tests__/chat.test.ts
+++ b/packages/instantsearch.js/src/lib/chat/__tests__/chat.test.ts
@@ -29,10 +29,10 @@ describe('ChatState', () => {
   beforeEach(() => {
     // Clear sessionStorage before each test
     sessionStorage.removeItem(CACHE_KEY);
-    sessionStorage.removeItem(CACHE_KEY + '-agentID1');
-    sessionStorage.removeItem(CACHE_KEY + '-agentID2');
-    sessionStorage.removeItem(CACHE_KEY + '-agentID3');
-    sessionStorage.removeItem(CACHE_KEY + '-agentID4');
+    sessionStorage.removeItem(`${CACHE_KEY}-agentID1`);
+    sessionStorage.removeItem(`${CACHE_KEY}-agentID2`);
+    sessionStorage.removeItem(`${CACHE_KEY}-agentID3`);
+    sessionStorage.removeItem(`${CACHE_KEY}-agentID4`);
   });
 
   afterAll(() => {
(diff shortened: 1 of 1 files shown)


packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -435,6 +437,13 @@ function ChatWrapper({
         footerComponent: promptProps.footerComponent,
         translations: promptProps.translations,
       }}
+      suggestionsProps={{
+        suggestions,
+        onSuggestionClick: (suggestion) => {
+          setChatInput(suggestion);
+          promptProps.promptRef.current?.focus();
```
Haroenv: do you think that it should set the input or send the message? I think sending the message may make more sense.

aymeric-giraudet: That's what I did in my PoC but iirc on the backend PoC on the dashboard it only prefilled the prompt 
But now that I see it, on figma it says it should reply already, so I think I'll change that

FabienMotte (approved):

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -257,6 +262,8 @@ export function createChatMessageComponent({ createElement }: Renderer) {
               {message.parts.map(renderMessagePart)}
             </div>
 
+            {suggestionsElement}
```
Haroenv, line 265: in the future, it may be interesting to add the suggestion to the message.parts directly and style it as a built-in tool, but this is the way to go for now

packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
```diff
@@ -360,6 +368,13 @@ export function createChatMessagesComponent({
                 translations={translations}
                 classNames={messageClassNames}
                 messageTranslations={messageTranslations}
+                suggestionsElement={
+                  status === 'ready' &&
+                  message.role === 'assistant' &&
+                  index === messages.length - 1
+                    ? suggestionsElement
+                    : undefined
```
Haroenv, line 376: I first thought that it may make sense to remove it in the connector if it's not last, but there are probably some UIs where you keep displaying all suggestions.

PR #6834, opened 2025-12-09
feat(react-instantsearch): export `useStickToBottom` hook

FabienMotte:
**Summary**

This PR adds exports for the `useStickToBottom` hook. This is required for the Dashboard, where chat UI components and the `useChat` hook are used.

### DIFF
--- a/packages/react-instantsearch/src/__tests__/index.test.ts
+++ b/packages/react-instantsearch/src/__tests__/index.test.ts
@@ -5,7 +5,9 @@ describe('exports', () => {
   describe('umd', () => {
     test('has the same number of exports as the main entrypoint', () => {
       expect(Object.keys(exportsUmd).sort()).toEqual(
-        Object.keys(exports).sort()
+        Object.keys(exports)
+          .filter((key) => key !== 'useStickToBottom')
+          .sort()
       );
     });
   });


bundlesize.config.json
```diff
@@ -22,7 +22,7 @@
     },
     {
       "path": "packages/react-instantsearch/dist/umd/ReactInstantSearch.min.js",
-      "maxSize": "72.5 kB"
+      "maxSize": "75 kB"
```
Haroenv: why does this add 2.5kb?

packages/react-instantsearch/src/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from 'react-instantsearch-core';
 export * from './widgets';
 export * from './components';
+export * from './lib/useStickToBottom';
```
Haroenv, line 5: can we add some comment to clarify it's internal?

bundlesize.config.json
FabienMotte: With UMD, the entire implementation is included, should I remove the export for UMD?

Haroenv: yes, let's remove it from umd as we don't use it

PR #6841, opened 2025-12-18
feat(sortBy): support composition sorting strategies

taylorcjohnson:
**Summary**

This PR adds support to the `sortBy` widget for composition sorting strategies. The `items` parameter can hold objects with a `value` property (current) or objects with a `strategy` property (new, for compositions). Those properties are mutually exclusive. `value` still updates the index for a search request. However, `strategy` adds the `sortBy` parameter to the search request.

**Result**

It requires using the `compositionClient` (`@algolia/composition`) instead of the `algoliasearch` client and passing a `compositionID` instead of an index. 

In addition to the traditional index items with a `value` property, `sortBy` items can have a `strategy` property instead that references the composition sorting strategy created in the Algolia Merchandising Studio.

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -14,7 +14,7 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "195 kB"
+      "maxSize": "200 kB"
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",
--- a/examples/js/query-suggestions/src/app.ts
+++ b/examples/js/query-suggestions/src/app.ts
@@ -41,6 +41,14 @@ search.addWidgets([
         getURL: (item) => `/products.html?pid=${item.objectID}`,
       },
     ],
+    showRecent: {
+      templates: {
+        header: (_, { html }) => html`
+          <span class="ais-AutocompleteIndexHeaderTitle">Recent Searches</span>
+          <span class="ais-AutocompleteIndexHeaderLine" />
+        `,
+      },
+    },
     showSuggestions: {
       indexName: 'instant_search_demo_query_suggestions',
       templates: {
--- a/examples/react/query-suggestions/src/App.tsx
+++ b/examples/react/query-suggestions/src/App.tsx
@@ -55,7 +55,6 @@ export function App() {
 
             <div className="search-panel__results">
               <EXPERIMENTAL_Autocomplete
-                showRecent
                 placeholder="Search for products"
                 indices={[
                   {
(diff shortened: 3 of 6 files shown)


packages/instantsearch.js/src/connectors/sort-by/connectSortBy.ts
```diff
@@ -112,17 +168,74 @@ const connectSortBy: SortByConnector = function connectSortBy(
...
+            withUsage(
+              `Strategy "${item.strategy}" conflicts with an existing index value. Index values and strategy names must be unique.`
+            )
+          );
+        }
+        // Same-type duplicates: silent overwrite (existing behavior)
+      }
+
+      itemsMap.set(itemValue, item);
+    });
```
taylorcjohnson: @e-krebs does this type of validation make sense? Or is it ok if the `strategy` property is the same as a `value` property? Is there any restriction on composition sorting strategy names or does it matter if they are the same as a replica index name?

packages/instantsearch.js/src/connectors/sort-by/__tests__/connectSortBy-test.ts
```diff
@@ -850,4 +850,590 @@ See documentation: https://www.algolia.com/doc/api-reference/widgets/sort-by/js/
...
+  describe('Sorting Strategies (Composition Mode)', () => {
+    describe('Validation', () => {
+      it('throws when item has both value and strategy', () => {
+        expect(() => {
+          connectSortBy(() => {})({
+            items: [
+              { value: 'index1', strategy: 'strategy1', label: 'Test' } as any,
+            ],
+          });
+        }).toThrow(
```
taylorcjohnson, line 863: I tried using `.toThrowErrorMatchingInlineSnapshot` (like what is used above in this file) in a previous commit, but the tests were failing. I'm not sure if it was a whitespace issue with the snapshot. 

I'm happy to switch it back if that is preferred and then we can troubleshoot from there.

packages/instantsearch.js/src/connectors/sort-by/connectSortBy.ts
e-krebs: it's completely OK to have strategy and value with the same name

taylorcjohnson: Thanks - I'll drop this then!

packages/instantsearch.js/src/connectors/sort-by/__tests__/connectSortBy-test.ts
Haroenv: it's fine either way, inline snapshot is just usually easier to write.

taylorcjohnson: Ok, I'll leave it for now. Happy to revisit in the future, if needed, though.

PR #6847, opened 2025-12-30
feat(chat): add support for custom type in render state

Haroenv:
**Summary**

preparation for [FX-3606]

**Result**

the chat connector accepts a "type" argument, which gets used in render state, so that one type of chat widget can communicate with another type. This will be used for a communication between chat + prompt suggestions.

[FX-3606]: https://algolia.atlassian.net/browse/FX-3606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


FabienMotte (approved):
💯

PR #6853, opened 2026-01-07
feat(autocomplete): add `transformItems` prop

shaejaz:
[FX-3589]

**Summary**
This PR adds a `transformItems` prop to the autocomplete widget which, similarly to other `transformItems`, allows users to update hits/indices being displayed (including reordering with working keyboard navigation).

[FX-3589]: https://algolia.atlassian.net/browse/FX-3589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -10,11 +10,11 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "123 kB"
+      "maxSize": "123.25 kB"
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "243.25 kB"
+      "maxSize": "243.5 kB"
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",


packages/react-instantsearch/src/widgets/Autocomplete.tsx
```diff
@@ -133,6 +137,21 @@ export type AutocompleteProps<TItem extends BaseHit> = ComponentProps<'div'> & {
...
+      results: SearchResults;
+      sendEvent: SendEventForHits;
+    }>
+  ) => Array<{
+    indexName: string;
+    indexId: string;
+    hits: Hit[];
+    results: SearchResults;
+    sendEvent: SendEventForHits;
+  }>;
```
shaejaz: I've called it `transformItems` for now but it doesn't seem completely accurate to me since you are actually transforming the indices config list. Maybe something like `transformIndices` or just `transform` makes more sense?

This is the standard indices config type right now but should ideally be changed to only include `indexId` and/or `indexName` (for specific ordering of indices) and `hits`. Are there other use cases for keeping `results` and `sendEvent`?

Haroenv: - I think it should be called transformItems even though we use a different name internallly
- sendEvent can be left out if that's easy to do
- results should stay, as you can do conditional changes based on it

packages/instantsearch.js/src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -204,9 +213,11 @@ search.addWidgets([
           };
         });
 
+        const finalIndices = transformItems(indices);
+
         return {
           currentRefinement: state.query || '',
-          indices,
+          indices: finalIndices,
```
Haroenv:
```suggestion
          indices: transformItems(indices),
```

we're not using the intermediate name, so I think this is cleaner

packages/instantsearch.js/src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -204,9 +213,11 @@ search.addWidgets([
           };
         });
 
+        const finalIndices = transformItems(indices);
+
```
Haroenv:
```suggestion
```

tests/common/widgets/autocomplete/options.tsx
```diff
@@ -1193,5 +1193,247 @@ export function createOptionsTests(
...
+          '.ais-AutocompleteIndexItem'
+        );
+        expect(indicesItems).toHaveLength(4);
+        expect(indicesItems[0]).toHaveTextContent('Item 1');
+        expect(indicesItems[1]).toHaveTextContent('Item 2');
+        expect(indicesItems[2]).toHaveTextContent('hello');
+        expect(indicesItems[3]).toHaveTextContent('hi');
+      });
+
+      test('reorders indices with correct keyboard navigation', async () => {
```
Haroenv, line 1363: we should have a test for deleting and reordering items within a hits list too

Haroenv (commented):
looks great, thanks!

packages/instantsearch.js/src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -206,7 +222,11 @@ search.addWidgets([
 
         return {
           currentRefinement: state.query || '',
-          indices,
+          indices: transformItems(indices).map((item, idx) => ({
+            ...item,
+            // keep sendEvent function intact after transformItems
+            sendEvent: indices[idx].sendEvent,
```
Haroenv: what if someone reordered the list? probably would have been easier to keep sendEvent as part of the item/index

shaejaz: It should be possible to do this by generating a map of indexId and the send event for hits in the initial indices array. But if you think send event should still be there lmk.

Haroenv: that, or we can create the sendEvent afterwards as well. Whatever solution that makes it not breakable is good

PR #6855, opened 2026-01-08
feat: add prompt suggestions widget

aymeric-giraudet:
**Summary**

[FX-3606](https://algolia.atlassian.net/browse/FX-3606)

**Result**

https://github.com/user-attachments/assets/3eecec5d-c44a-4c67-b66f-6bab245908b7

[FX-3606]: https://algolia.atlassian.net/browse/FX-3606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -10,19 +10,19 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "117 kB"
+      "maxSize": "118.25 kB"
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "237.5 kB"
+      "maxSize": "240 kB"
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",
-      "maxSize": "61.25 kB"
+      "maxSize": "61.75 kB"
     },
     {
       "path": "packages/react-instantsearch/dist/umd/ReactInstantSearch.min.js",
-      "maxSize": "96.5 kB"
+      "maxSize": "97.5 kB"
     },
     {
       "path": "packages/vue-instantsearch/vue2/umd/index.js",
--- a/packages/instantsearch-ui-components/src/components/PromptSuggestions.tsx
+++ b/packages/instantsearch-ui-components/src/components/PromptSuggestions.tsx
@@ -4,7 +4,7 @@ import { cx } from '../lib';
 import { createChatPromptSuggestionsComponent } from './chat/ChatPromptSuggestions';
 import { SparklesIcon } from './chat/icons';
 
-import type { Renderer } from '../types';
+import type { ComponentProps, Renderer } from '../types';
 import type { ChatPromptSuggestionsClassNames } from './chat/ChatPromptSuggestions';
 
 export type PromptSuggestionsClassNames = {
@@ -26,7 +26,7 @@ export type PromptSuggestionsTranslations = {
 
 export type PromptSuggestionsStatus = 'idle' | 'loading' | 'ready' | 'error';
 
-export type PromptSuggestionsProps = {
+export type PromptSuggestionsProps = ComponentProps<'div'> & {
   /**
    * List of prompt suggestions.
    */
@@ -70,6 +70,7 @@ export function createPromptSuggestionsComponent({
       translations = {},
       classNames = {},
       titleIconComponent: TitleIconComponent,
+      ...props
     } = userProps;
 
     const { title = 'Ask a question about this product:' } = translations;
@@ -82,7 +83,7 @@ export function createPromptSuggestionsComponent({
     const isLoading = status === 'loading';
 
     return (
-      <div className={cx('ais-PromptSuggestions', classNames.root)}>
+      <div className={cx('ais-PromptSuggestions', classNames.root)} {...props}>
         {title && (
           <div
             className={cx('ais-PromptSuggestions-header', classNames.header)}
--- a/packages/instantsearch.js/src/__tests__/common-widgets.test.tsx
+++ b/packages/instantsearch.js/src/__tests__/common-widgets.test.tsx
@@ -660,10 +660,10 @@ const testSetups: TestSetupsMap<TestSuites, 'javascript'> = {
           container: document.body.appendChild(document.createElement('div')),
           ...widgetParams,
         }),
+        // @ts-ignore needed to have Chat working inside PromptSuggestions tests
         chat({
           container: document.body.appendChild(document.createElement('div')),
           agentId: widgetParams.agentId,
-          // @ts-ignore For some reason there are problems with types going from this library to the test ones back here
           chat: widgetParams.chat,
         }),
       ])
--- a/packages/instantsearch.js/src/connectors/index.umd.ts
+++ b/packages/instantsearch.js/src/connectors/index.umd.ts
@@ -26,6 +26,7 @@ export { connectDynamicWidgets };
 
 export { default as connectClearRefinements } from './clear-refinements/connectClearRefinements';
 export { default as connectCurrentRefinements } from './current-refinements/connectCurrentRefinements';
+export { default as connectFilterSuggestions } from './filter-suggestions/connectFilterSuggestions';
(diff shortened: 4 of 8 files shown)


packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -346,6 +346,7 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
 
           return Promise.resolve();
         },
+        skipCache: type !== 'chat',
```
Haroenv: why should the cache be skipped?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,196 @@
...
+      },
+
+      getRenderState(
+        renderState,
+        renderOptions
+      ): IndexRenderState &
+        PromptSuggestionsWidgetDescription['indexRenderState'] {
+        return {
+          ...renderState,
+          promptSuggestions: this.getWidgetRenderState(renderOptions),
```
Haroenv, line 284: we're overriding the getRenderState now anyway, that means that the `type` option in chat isn't really needed anymore, no?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,196 @@
...
+            false
+          );
+        };
+
+        chatWidget.chatInstance['~registerMessagesCallback'](render);
+        chatWidget.chatInstance['~registerStatusCallback'](render);
+        chatWidget.chatInstance['~registerErrorCallback'](render);
+
+        const renderState = chatWidget.getWidgetRenderState(initOptions);
+        renderState.sendMessage({ text: JSON.stringify(widgetParams.context) });
```
Haroenv: is init the correct place for this message to be sent? does that trigger at the right time for react for example?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
aymeric-giraudet: Else it writes what the PromptSuggestions widget requested to the chat widget as they're using the same agentId.
Also no point in using a frontend cache for this

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
aymeric-giraudet: It does trigger at the right time yes, tested it in react

packages/instantsearch.js/src/connectors/chat/connectChat.ts
Haroenv: so skipCache skips writing, but not reading?

aymeric-giraudet: Don't think it would be a problem but i'll skip reading as well

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,196 @@
...
+  transport?: ConstructorParameters<typeof DefaultChatTransport>[0];
+};
+
+export type PromptSuggestionsConnectorParams<
+  TUiMessage extends UIMessage = UIMessage
+> = { chat?: Chat<TUiMessage> } & PromptSuggestionsTransport & {
+    /**
+     * Context object to send to the agent (e.g., product data for a PDP).
+     */
+    context: Record<string, unknown>;
```
shaejaz: Should this be just `unknown`? Is there a chance that they have an agent setup in way that only requires the object id for example?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: It seems like it's all getting overridden. I might be misunderstanding it, but what is the need to use `connectChat` here in that case? Maybe just using the chat init function here would be enough.

Could this approach also cause a clash in the index render state (due to there also being another chat widget/connector)?

Haroenv: the key is promptSuggestions, not chat anymore with this connector

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
aymeric-giraudet: We may need to be able to compare it if it changed, would be easier with an object idk. Could be `{ objectID: '' }` too in this case anyway

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,294 @@
...
+      },
+
+      getRenderState(
+        renderState,
+        renderOptions
+      ): IndexRenderState &
+        PromptSuggestionsWidgetDescription['indexRenderState'] {
+        return {
+          ...renderState,
+          promptSuggestions: this.getWidgetRenderState(renderOptions),
```
Haroenv, line 284: because we use a different widget, do you mind reverting the `type` change in chat widget?

Haroenv: can be in a different PR

aymeric-giraudet: Will make a different PR yeah

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,294 @@
...
+            throw new Error(
+              'PromptSuggestions: Chat widget not found. Make sure you have a Chat widget in your InstantSearch instance.'
+            );
+          }
+
+          if (!mainChat.open) {
+            mainChat.setOpen(true);
+          }
+
+          mainChat.sendMessage({ text: suggestion });
```
FabienMotte:
```suggestion
          const messageWithContext = context
            ? `<!--ais-context:${JSON.stringify(context)}-->${suggestion}`
            : suggestion;

          mainChat.sendMessage({ text: messageWithContext });
```

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,294 @@
...
+      );
+
+      const payload: Record<string, unknown> = {
+        messages: [
+          {
+            role: 'user',
+            content: JSON.stringify(context),
+          },
+        ],
+      };
```
FabienMotte, line 182: I think this payload is compatible with AI SDK v4, do we want to use v5 instead?


```suggestion
      const payload: Record<string, unknown> = {
        messages: [
          {
            role: 'user',
            parts: [
              {
                type: 'text',
                text: JSON.stringify(context),
              },
            ],
          },
        ],
      };
```

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,294 @@
...
+          );
+
+          if (assistantMessage?.parts) {
+            const suggestionsPart = assistantMessage.parts.find(
+              (part: { type: string }) => part.type === 'data-suggestions'
+            );
+            if (suggestionsPart?.data?.suggestions) {
+              suggestions = suggestionsPart.data.suggestions;
+            }
+          }
```
FabienMotte: I think `data` is already the message when I tested it yesterday

```suggestion
          if (data.role !== 'assistant') {
            return;
          }
          suggestions = data.parts?.find(
            (part: { type: string }) => part.type === 'data-suggestions'
          )?.data?.suggestions;
```

PR #6861, opened 2026-01-13
feat: add `FilterSuggestions` widget

FabienMotte:
**Summary**

[FX-3667](https://algolia.atlassian.net/browse/FX-3667)

[FX-3667]: https://algolia.atlassian.net/browse/FX-3667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -81,11 +81,12 @@ module.exports = (api) => {
             // we require polyfills for this already
             'Array.prototype.includes',
 
-            // Used only in Chat, which expects modern browsers
+            // Used only in newer widgets, which expect modern browsers
             'Object.fromEntries',
             'Object.entries',
             'Array.prototype.find',
             'String.prototype.startsWith',
+            'Promise.prototype.finally',
 
             // false positive (babel doesn't know types)
             // this is actually only called on arrays


packages/instantsearch-ui-components/src/components/RefinementSuggestions.tsx
```diff
@@ -0,0 +1,285 @@
...
+        className={cx(classNames.button)}
+        onClick={onRefine}
+      >
+        <span
+          className={cx('ais-RefinementSuggestions-label', classNames.label)}
+        >
+          {suggestion.label}: {suggestion.value}
+        </span>
+        <span
+          className={cx('ais-RefinementSuggestions-nbHits', classNames.count)}
```
Haroenv: this fragment is called -count in refinementlist I believe

packages/instantsearch-ui-components/src/components/FilterSuggestions.tsx
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  return function RefinementSuggestions(
+    userProps: RefinementSuggestionsProps
+  ): JSX.Element | null {
+    const {
+      classNames = {},
+      suggestions,
+      isLoading,
+      onRefine,
```
Haroenv: usually we pass a `refine` function instead of `onRefine`

packages/instantsearch.js/src/connectors/filter-suggestions/connectFilterSuggestions.ts
```diff
@@ -0,0 +1,415 @@
...
+
+          if (!appId || !apiKey) {
+            throw new Error(
+              withUsage(
+                'Could not extract Algolia credentials from the search client.'
+              )
+            );
+          }
+
+          endpoint = `https://${appId}.algolia.net/agent-studio/1/agents/${agentId}/completions?compatibilityMode=ai-sdk-5&stream=false`;
```
Haroenv: not saying we per se need the "transport" option right now as in chat, but this option means that users can't customise the URL for whatever reason. We can likely consolidate once we replace the ai library for chat. 

not a blocker for now

packages/instantsearch.js/src/connectors/filter-suggestions/connectFilterSuggestions.ts
```diff
@@ -0,0 +1,415 @@
...
+          headers = {
+            'x-algolia-application-id': appId,
+            'x-algolia-api-key': apiKey,
+            'x-algolia-agent': getAlgoliaAgent(instantSearchInstance.client),
+          };
+
+          refine = (attribute: string, value: string) => {
+            // Check if the attribute belongs to a hierarchical facet
+            // by finding a hierarchical facet that includes this attribute
+            const hierarchicalFacet = helper.state.hierarchicalFacets.find(
```
Haroenv: breaks in IE, but isn't an issue as it's consolidated to this widget

packages/instantsearch.js/src/widgets/__tests__/index.test.ts
```diff
@@ -174,6 +174,15 @@ function initiateAllWidgets(): Array<[WidgetNames, Widget | IndexWidget]> {
 
         return autocomplete;
       }
+      case 'refinementSuggestions': {
+        const refinementSuggestions =
+          widget as Widgets['refinementSuggestions'];
+        return refinementSuggestions({
+          container,
+          agentId: 'test-agent-id',
+          attributes: ['attr'],
```
Haroenv, line 182: is attributes required or just passed here as it's possible?

packages/instantsearch.js/src/widgets/index.umd.ts
```diff
@@ -68,3 +68,11 @@ Please use InstantSearch.js with a packaging system:
 https://www.algolia.com/doc/guides/building-search-ui/installation/js/#with-a-packaging-system`
   );
 };
+export const refinementSuggestions = () => {
+  throw new Error(
+    `"refinementSuggestions" is not available from the UMD build.
```
Haroenv: there's no technical reason to exclude this widget from umd right? doesn't really matter as we'll merge this before getting rid of the umd exclusion, but just checking.

packages/instantsearch.js/src/connectors/filter-suggestions/connectFilterSuggestions.ts
```diff
@@ -0,0 +1,415 @@
...
+        const facetsToSend = attributes
+          ? Object.fromEntries(
+              Object.entries(rawFacets).filter(([key]) =>
+                attributes.includes(key)
+              )
+            )
+          : rawFacets;
+
+        // Collect current refinements to exclude from suggestions
+        const currentRefinements: Array<{ attribute: string; value: string }> =
```
Haroenv: can we reuse the logic from connectCurrentRefinements? I believe there's a helper already

packages/instantsearch.js/src/connectors/filter-suggestions/connectFilterSuggestions.ts
FabienMotte: I talked with @aymeric-giraudet about that, we will need to pass a custom endpoint and modify the body payload for the Dashboard preview

See: https://algolia.slack.com/archives/C08V2M09U73/p1768836270732789?thread_ts=1768835929.092399&cid=C08V2M09U73

packages/instantsearch.js/src/connectors/filter-suggestions/connectFilterSuggestions.ts
Haroenv: you can make this `const attr = helper.state.hierarchicalFacets.find...?.name || attribute` or something along those lines, and you won't need the conditional later

packages/instantsearch.js/src/widgets/__tests__/index.test.ts
aymeric-giraudet: Not required

packages/instantsearch.js/src/connectors/filter-suggestions/connectFilterSuggestions.ts
```diff
@@ -0,0 +1,431 @@
...
+            },
+          ],
+        };
+
+        // Apply custom body transformation if provided
+        const finalPayload = transport?.prepareSendMessagesRequest
+          ? transport.prepareSendMessagesRequest(payload).body
+          : payload;
+
+        fetch(endpoint, {
```
shaejaz, line 266: Would be good to extract this into a util/class in the future

PR #6866, opened 2026-01-16
feat(chat): handle setting refinements from the search index tool 'view all' button

shaejaz:
**Summary**
[FX-3631]

This PR adds setting refinements (not just query) from the 'View All' button in the default Search Index Tool component. A new param, `applyFilters` is now provided to the `layoutComponent` of tools.

[FX-3631]: https://algolia.atlassian.net/browse/FX-3631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -14,7 +14,7 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "237.75 kB"
+      "maxSize": "238 kB"
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",


packages/instantsearch.js/src/connectors/current-refinements/connectCurrentRefinements.ts
```diff
@@ -142,6 +142,9 @@ export type CurrentRefinementsRenderState = {
    * Generates a URL for the next state.
    */
   createURL: CreateURL<CurrentRefinementsConnectorParamsRefinement>;
+
+  aiMode: boolean;
+  setAiMode: (value: boolean) => void;
```
Haroenv: is this on purpose?

tests/common/widgets/chat/options.tsx
```diff
@@ -425,6 +425,269 @@ export function createOptionsTests(
...
+            {
+              id: '1',
+              role: 'assistant',
+              parts: [
+                {
+                  type: `tool-${SearchIndexToolType}`,
+                  toolCallId: '1',
+                  input: {
+                    query: 'test',
+                    facet_filters: [['brand:Apple'], ['category:Laptops']],
```
Haroenv, line 525: we should add an attribute that isn't in our list here too, as well as a hierarchical one. Can be in a different test just for the edge cases

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -435,11 +441,18 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
           });
         }
 
+        function applyFilters(params: ApplyFiltersParams): boolean {
+          if (!helper) return false;
```
Haroenv: why would helper not be present? I think we should probably throw an error instead of returning false, or is it a predictable reason why helper isn't there?

packages/instantsearch.js/src/lib/utils/updateStateFromSearchToolInput.ts
```diff
@@ -0,0 +1,52 @@
...
+    attributes.forEach(({ attribute, value }) => {
+      const hierarchicalFacet = helper.state.hierarchicalFacets.find(
+        (facet) => facet.name === attribute
+      );
+
+      if (hierarchicalFacet) {
+        helper.toggleFacetRefinement(hierarchicalFacet.name, value);
+      } else {
+        helper.toggleFacetRefinement(attribute, value);
+      }
```
Haroenv:
```suggestion
      const attr = helper.state.hierarchicalFacets.find(
        (facet) => facet.name === attribute
      )?.name || attribute;
      
      helper.toggleFacetRefinement(attr, value);
```

not sure if it's cleaner, but I think I prefer something along these lines. what do you think?

packages/instantsearch.js/src/lib/utils/updateStateFromSearchToolInput.ts
```diff
@@ -0,0 +1,52 @@
+import { flat } from './flat';
+
+import type { AlgoliaSearchHelper } from 'algoliasearch-helper';
+import type { ApplyFiltersParams } from 'instantsearch-ui-components';
+
+export function updateStateFromSearchToolInput(
```
Haroenv: maybe add a jsdoc comment here that explains the output means whether it worked or not

packages/instantsearch.js/src/lib/utils/updateStateFromSearchToolInput.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (
+      attributes.some(
+        ({ attribute }) =>
+          !helper.state.isConjunctiveFacet(attribute) &&
+          !helper.state.isHierarchicalFacet(attribute) &&
+          !helper.state.isDisjunctiveFacet(attribute)
+      )
+    ) {
+      return false;
+    }
```
Haroenv: if it's not currently a refinement, I think we should add it as a conjunctive or disjunctive facet, what do you think?

Haroenv: that way we likely no longer need the boolean flow, right?

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -207,24 +209,15 @@ function createCarouselTool<
...
+                  query: input.query,
+                  facetFilters: input.facet_filters,
+                });
+                if (success) {
                   onClose();
-                  return;
                 }
-
-                // Navigate to different page
-                window.location.href = getSearchPageURL(nextUiState);
```
Haroenv, line 227: we're not handling "not on search page" anymore?

packages/instantsearch.js/src/lib/utils/updateStateFromSearchToolInput.ts
FabienMotte, line 1: Should we add some unit tests for this function?

packages/react-instantsearch/src/widgets/chat/tools/SearchIndexTool.tsx
```diff
@@ -135,34 +143,27 @@ function createCarouselTool<TObject extends RecordWithObjectID>(
...
+                  if (success) {
+                    onClose();
+                  }
+                }}
+                className="ais-ChatToolSearchIndexCarouselHeaderViewAll"
+              >
+                View all
+                <ArrowRightIcon createElement={createElement as Pragma} />
+              </Button>
+            </>
```
FabienMotte:
```suggestion
              <Button
                variant="ghost"
                size="sm"
                onClick={() => {
                  if (!input || !applyFilters) return;

                  const success = applyFilters({
                    query: input.query,
                    facetFilters: input.facet_filters,
                  });
                  if (success) {
                    onClose();
                  }
                }}
                className="ais-ChatToolSearchIndexCarouselHeaderViewAll"
              >
                View all
                <ArrowRightIcon createElement={createElement as Pragma} />
              </Button>
```

packages/react-instantsearch/src/__tests__/common-widgets.test.tsx
```diff
@@ -396,9 +396,23 @@ const testSetups: TestSetupsMap<TestSuites, 'react'> = {
     );
   },
   createChatWidgetTests({ instantSearchOptions, widgetParams }) {
+    const { renderRefinements, ...chatWidgetParams } = widgetParams;
     render(
       <InstantSearch {...instantSearchOptions}>
-        <Chat {...widgetParams} />
+        {renderRefinements && (
+          <>
```
FabienMotte, line 403: For consistency with the IS.js test suite, should we add a `SearchBox` widget here?

packages/instantsearch.js/src/lib/utils/updateStateFromSearchToolInput.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { AlgoliaSearchHelper } from 'algoliasearch-helper';
+import type { ApplyFiltersParams } from 'instantsearch-ui-components';
+
+export function updateStateFromSearchToolInput(
+  params: ApplyFiltersParams,
+  helper: AlgoliaSearchHelper
+) {
+  if (params.query) {
+    helper.setQuery(params.query);
+  }
```
FabienMotte: Should the query be set after the facet filters are validated?
This way, if validation fails, the query will not persist in the helper state

packages/instantsearch.js/src/lib/utils/updateStateFromSearchToolInput.ts
shaejaz: Move this to the connector now, I think the current options tests cover this function now, but lmk if more are still needed for it specifically.

PR #6869, opened 2026-01-21
ci: move to GitHub actions

Haroenv:
**Summary**

migrate releasing from circleci to GitHub actions

**Result**

By being on github actions, we no longer need a bot account, as it can use the github token generated by actions.

This also means you no longer need to run release locally with a github token to prepare a release, you can start it from the github actions page.

[FX-3682]

[FX-3682]: https://algolia.atlassian.net/browse/FX-3682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


FabienMotte (approved):

.github/workflows/release-trigger.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Configure git user
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+      - name: Trigger release if latest commit is a release commit
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: npx shipjs@0.26.0 trigger
```
dhayab, line 60: Does this still require `NPM_TOKEN` or have you enabled trusted publishing?

dhayab (commented):
This looks good to me, although I'm wondering if at this stage, it wouldn't be better to move all jobs to GH instead. Wdyt?

.github/workflows/release-trigger.yml
Haroenv: good catch! one of the things I couldn't check without doing a release 😅

dhayab: If it's not too complex, I'd recommend relying on [trusted publishing](https://docs.npmjs.com/trusted-publishers#step-1-add-a-trusted-publisher-on-npmjscom) then.

Haroenv:
replaced by #7184

PR #6870, opened 2026-01-21
ci(cache): improve speed

Haroenv:
**Summary**

Improve caching, remove work that are only needed for release.

**Result**

15 -> 10 min for a PR


Haroenv:
I'm not convinced this saves much time actually, seems like it's now bottlenecked by the v4/v3 reinstall. Maybe that can be done more efficiently?

PR #6874, opened 2026-01-26
refactor(chat): move search index tool component to ui-components

shaejaz:
This PR moves the chat search tool's default component definition to the ui-components. This prevents repeating layout logic for both react and js versions of component. The approach is similar to how it is done for the new Autocomplete components.


packages/instantsearch-ui-components/src/components/chat/tools/SearchIndexTool.tsx
```diff
@@ -0,0 +1,252 @@
...
+  nbHits?: number;
+  input?: SearchToolInput;
+  nbItems: number;
+  applyFilters: ClientSideToolComponentProps['applyFilters'];
+  getSearchPageURL?: (params: SearchParameters) => string;
+  onClose: () => void;
+};
+
+export type SearchIndexToolProps<THit extends RecordWithObjectID> = {
+  useMemo: <TType>(factory: () => TType, inputs: readonly unknown[]) => TType;
```
anjola-adeuyi: could we inject these hooks into `createSearchIndexToolComponent` instead, like `createDisplayResultsToolComponent`? passing them through `SearchIndexToolProps` makes framework plumbing part of the new public component API.

packages/instantsearch-ui-components/src/components/chat/tools/SearchIndexTool.tsx
```diff
@@ -0,0 +1,252 @@
...
+              <ChevronRightIcon createElement={createElement} />
+            </Button>
+          </div>
+        )}
+      </div>
+    );
+  };
+}
+
+export function createSearchIndexToolComponent<
```
anjola-adeuyi: `createCarouselTool` uses this factory for both search-index and recommend tools. should it have a generic results/carousel name rather than a search-specific one?

bundlesize.config.json
```diff
@@ -10,7 +10,7 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "130.5 kB"
+      "maxSize": "130.75 kB"
```
Haroenv: you'd hope that deduplicating shrinks the size, but it actually increases it, do you know why?

packages/instantsearch-ui-components/src/components/chat/tools/SearchIndexTool.tsx
```diff
@@ -0,0 +1,252 @@
...
+          nbHits?: number;
+          queryID?: string;
+        }
+      | undefined;
+
+    const items = (output?.hits || []).map((hit, idx) => ({
+      ...hit,
+      __position: idx + 1,
+      ...(output?.queryID ? { __queryID: output.queryID } : {}),
+    }));
```
Haroenv: in InstantSearch.js we use a utility for this (I believe mainly for the type to be right), do we need it here too, or is the type imprecise enough that it doesn't matter?

packages/instantsearch-ui-components/src/components/chat/tools/CarouselTool.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+    const items = (output?.hits || []).map((hit, idx) => ({
+      ...hit,
+      __position: idx + 1,
+      ...(output?.queryID ? { __queryID: output.queryID } : {}),
+    }));
+    const nbItems = items.length;
+
+    const [canScrollLeft, setCanScrollLeft] = useState(false);
+    const [canScrollRight, setCanScrollRight] = useState(true);
```
Haroenv, line 173: how come this is state and not derived from the pagination position?

Haroenv (commented):
🔥

bundlesize.config.json
shaejaz: I think the cause is that regardless of the component being in a separate lib, it still has to be included and bundled in the end packages (is.js and react-is), so likely won't see an reduction when moving components to ui-components.

The actual increase (i think it's around 0.15kb) is due the factory structure that comes with the ui-components pattern.

packages/instantsearch-ui-components/src/components/chat/tools/CarouselTool.tsx
shaejaz: might be misunderstanding this but the carousel isn't hooked into pagination logic, just simple scrolling to right and left if the items being displayed cannot be fully shown horizontally.

packages/instantsearch-ui-components/src/components/chat/tools/SearchIndexTool.tsx
shaejaz: Copied over the util into ui-components for now. I think that's okay since it's pretty small and we've done this before for a few other utils.

packages/instantsearch-ui-components/src/components/chat/tools/CarouselTool.tsx
Haroenv: oh right, and the scroll state is kept in sync with a listener?

Haroenv (approved):

anjola-adeuyi (approved):

PR #6875, opened 2026-01-26
fix(nextjs): prevent infinite loading when search client fails

### LINKED ISSUE #6525: Infinite Page Loading Next.js when no connection to search server

nx-alejandrolacasa:
## Summary
This PR fixes issue #6525 where the Next.js InstantSearch app would hang indefinitely during SSR when the search client fails (e.g., no network connection, API timeout, or server unavailable).

## Changes
- Add 30-second timeout for SSR search operations to prevent infinite waiting
- Add error handlers for search client failures in both composition and standard clients
- Introduce `ssrSearchFailed` flag to track when SSR search fails and skip result injection
- Add error listeners for helper and derived helper error events
- Wrap all SSR operations in try-catch blocks to ensure graceful degradation
- Make SSR results truly optional - the app continues to render even when search fails

## Impact
Before this fix, if the search API was unreachable during SSR, the Next.js app would hang indefinitely, resulting in a poor user experience and potentially server timeouts.

After this fix, the app gracefully handles connection failures and continues to render the page without SSR results, allowing client-side hydration to work normally once the connection is restored.

Fixes #6525

### DIFF
--- a/packages/react-instantsearch-nextjs/src/__tests__/InitializePromise-composition.test.tsx
+++ b/packages/react-instantsearch-nextjs/src/__tests__/InitializePromise-composition.test.tsx
@@ -89,6 +89,86 @@ test('it waits for composition-based search', async () => {
   );
 });
 
+test('it resolves when composition search client throws an error', async () => {
+  const ref: { current: PromiseWithState<void> | null } = { current: null };
+  const insertedHTML = jest.fn();
+
+  const client = createCompositionClient({
+    search: jest.fn().mockRejectedValue(new Error('Network error')),
+  });
+
+  await act(() =>
+    render(
+      <InstantSearchRSCContext.Provider
+        value={{
+          waitForResultsRef: ref,
+          countRef: { current: 0 },
+          ignoreMultipleHooksWarning: false,
+        }}
+      >
+        <InstantSearchSSRProvider>
+          <InstantSearch searchClient={client} compositionID={compositionID}>
+            <ServerInsertedHTMLContext.Provider
+              value={(cb) => insertedHTML?.(cb())}
+            >
+              <InitializePromise />
+              <SearchBox />
+              <TriggerSearch />
+            </ServerInsertedHTMLContext.Provider>
+          </InstantSearch>
+        </InstantSearchSSRProvider>
+      </InstantSearchRSCContext.Provider>
+    )
+  );
+
+  await act(async () => {
+    await ref.current;
+  });
+
+  expect(ref.current!.status).toBe('fulfilled');
+  expect(insertedHTML).not.toHaveBeenCalled();
+});
+
+test('it resolves when composition search client returns invalid response', async () => {
+  const ref: { current: PromiseWithState<void> | null } = { current: null };
+  const insertedHTML = jest.fn();
+
+  const client = createCompositionClient({
+    search: jest.fn().mockResolvedValue(null),
+  });
+
+  await act(() =>
+    render(
+      <InstantSearchRSCContext.Provider
+        value={{
+          waitForResultsRef: ref,
+          countRef: { current: 0 },
+          ignoreMultipleHooksWarning: false,
+        }}
+      >
+        <InstantSearchSSRProvider>
+          <InstantSearch searchClient={client} compositionID={compositionID}>
+            <ServerInsertedHTMLContext.Provider
+              value={(cb) => insertedHTML?.(cb())}
+            >
+              <InitializePromise />
+              <SearchBox />
+              <TriggerSearch />
+            </ServerInsertedHTMLContext.Provider>
+          </InstantSearch>
+        </InstantSearchSSRProvider>
+      </InstantSearchRSCContext.Provider>
+    )
+  );
+
+  await act(async () => {
+    await ref.current;
+  });
+
+  expect(ref.current!.status).toBe('fulfilled');
+  expect(insertedHTML).not.toHaveBeenCalled();
+});
+
 afterAll(() => {
   jest.resetAllMocks();
 });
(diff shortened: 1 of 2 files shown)


packages/react-instantsearch-nextjs/src/InitializePromise.ts
```diff
@@ -34,52 +36,105 @@ export function InitializePromise({ nonce }: InitializePromiseProps) {
...
+    new Promise<void>((resolve, reject) => {
       let searchReceived = false;
       let recommendReceived = false;
-      search.mainHelper!.derivedHelpers[0].once('result', () => {
+      let isSettled = false;
+
+      const helper = search.mainHelper!;
+      const derivedHelper = helper.derivedHelpers[0];
+
+      const timeout = setTimeout(() => {
```
Haroenv, line 90: I'm surprised this timeout is needed, the algolia client already has timeout logic and should therefore error if not responding soon enough. Are you using a custom client?

packages/react-instantsearch-nextjs/src/InitializePromise.ts
```diff
@@ -34,52 +36,105 @@ export function InitializePromise({ nonce }: InitializePromiseProps) {
       throw new Error('Missing ServerInsertedHTMLContext');
     });
 
-  // Extract search parameters from the search client to use them
-  // later during hydration.
   let requestParamsList: SearchOptions[];
+  let ssrSearchFailed = false;
+
+  const handleResponse = (response: any) => {
+    if (!response || !('results' in response)) {
```
Haroenv, line 43: in which case is there a response without results?

packages/react-instantsearch-nextjs/src/InitializePromise.ts
nx-alejandrolacasa: Hey! Answering to your previous question, yes, this happens when running a custom client. I am working using [searchkit](https://github.com/searchkit/searchkit), but in the end.

I see in the original issue that the reporter uses the typesense adapted, so it might be happening only with custom clients, but I cannot tell 100% sure.

packages/react-instantsearch-nextjs/src/InitializePromise.ts
nx-alejandrolacasa: In case the server is down or not responding, the answer may be many things (404, 500, error, nothing...). This is an extra measure to prevent this situations that may return valid a JSON, but not a valid response.

Haroenv: I feel like this is an error in the adapter, and they should throw an error instead. Do you mind making a PR there to fix it instead of handling wrong client behaviour in InstantSearch?

packages/react-instantsearch-nextjs/src/InitializePromise.ts
Haroenv: makes sense, can you actually make a similar PR to handle timeouts in the adapters instead of in InstantSearch itself. We can't design our library around differently behaving clients IMO

nx-alejandrolacasa:
Closing this PR as we've decided to address this issue at the search client adapter level instead.

The fix has been implemented in the Searchkit adapter and submitted as a PR: https://github.com/searchkit/searchkit/pull/1412

The key changes in the Searchkit fix:
- Added configurable timeout (default 30s) to `ESTransporter` using `AbortController`
- Modified `@searchkit/instantsearch-client` to return empty results on error instead of throwing, which prevents the infinite re-render loop in InstantSearch
- Added error result caching to prevent repeated failing requests
- Added optional `onError` callback for custom error handling

This approach keeps the error handling closer to the source (the search client) and avoids needing to patch `react-instantsearch-nextjs`. Once the Searchkit PR is merged, users of Searchkit with InstantSearch will get this fix automatically.

Thanks @Haroenv for your support and guidance on this issue!

PR #6904, opened 2026-02-17
feat(chat): add compatibility with algolia mcp search tool [DASH-2294]

kombucha:
## Summary
Fixes DASH-2294
Adds compatibility with the Algolia MCP Server search tool.

The MCP Server names search tools with a suffix (e.g., `algolia_search_index_indexName`) instead of just `algolia_search_index`. This PR adds a shim that falls back to the `algolia_search_index` tool implementation when encountering these suffixed tool names.

## Result

By default, the chat widget now works with both:
- Standard `algolia_search_index` tool calls
- MCP Server style `algolia_search_index_*` tool calls

## Testing
- Added a couple of unit tests ✅ 
- Tested manually by pointing one of the examples to a locally running instance

```sh
yarn --cwd examples/js/default-theme start
```

Screenshot 2026-02-18 at 17 51 33

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -14,7 +14,7 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "240.50 kB"
+      "maxSize": "241 kB"
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",


packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -499,7 +498,12 @@ const createRenderer = <THit extends RecordWithObjectID = RecordWithObjectID>({
 
     const toolsForUi: ClientSideTools = {};
     Object.entries(toolsFromConnector).forEach(([key, connectorTool]) => {
-      const widgetTool = tools[key];
+      let widgetTool = tools[key];
+
+      // Compatibility shim with Algolia MCP Server search tool
+      if (!widgetTool && key.startsWith(`${SearchIndexToolType}_`)) {
+        widgetTool = tools[SearchIndexToolType];
+      }
```
Haroenv, line 506: we did the shim in the InstantSearch.js widget, but not the React InstantSearch widget, is that on purpose?

Haroenv (commented):
you can update the bundlesize for InstantSearch.js development to give it more margin.

Just wondering why the patch happens both in the connector and the InstantSearch.js widget (but not react), isn't one of those redundant?

I think it also could be interesting to add a test to the common test suite (tests/common/widgets/chat)

packages/instantsearch.js/src/widgets/chat/chat.tsx
kombucha: It should be handled for the react flavour too.
The `tools` config is passed down through a few layers of component and is finally handled in `packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx` which I updated 👍

Haroenv: makes sense! This probably changes with #6874 that's just for @shaejaz to take in account

Haroenv (approved):

packages/instantsearch.js/src/widgets/chat/chat.tsx
shaejaz: I think this would remain the same since you still need to map a tool into one with a `layoutComponent` in the js chat widget (and not in react) for it to be consumable in ui-components.

I'll try to see if this can be avoided in my PR tho.

PR #6912, opened 2026-03-03
fix(chat): show loader when agent is still processing on the backend

shaejaz:
[FX-3715] [FX-3727]

Cases when loader will show in streaming state:
- The last part in the message is text (we don't have to show a loader in this case since the text is streaming)
- The last part is a tool with no tool output (as we can show anything yet)

Actions also only show up when streaming is done. Side effect of this is that they will also appear after suggestions are generated, which is not how it's done yet.

https://github.com/user-attachments/assets/389229be-7260-4622-95b4-180da18b8037

[FX-3715]: https://algolia.atlassian.net/browse/FX-3715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FX-3727]: https://algolia.atlassian.net/browse/FX-3727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch-ui-components/src/lib/utils/chat.ts
+++ b/packages/instantsearch-ui-components/src/lib/utils/chat.ts
@@ -1,6 +1,4 @@
-import { startsWith } from './startsWith';
-
-import type { ChatMessageBase, ChatToolMessage } from '../../components';
+import type { ChatMessageBase } from '../../components';
 
 export const getTextContent = (message: ChatMessageBase) => {
   return message.parts
@@ -12,28 +10,8 @@ export const hasTextContent = (message: ChatMessageBase) => {
   return getTextContent(message).trim() !== '';
 };
 
-export const getToolParts = (message: ChatMessageBase) => {
-  return message.parts.filter((part) => startsWith(part.type, 'tool-'));
-};
-
-export const hasToolParts = (message: ChatMessageBase) => {
-  return getToolParts(message).length > 0;
-};
-
 export const isPartText = (
   part: ChatMessageBase['parts'][number]
 ): part is Extract<ChatMessageBase['parts'][number], { type: 'text' }> => {
   return part.type === 'text';
-};
-
-export const isPartTool = (
-  part: ChatMessageBase['parts'][number]
-): part is ChatToolMessage => {
-  return startsWith(part.type, 'tool-');
-};
-
-export const toolHasOutput = (message: ChatMessageBase) => {
-  return message.parts.some(
-    (part) => isPartTool(part) && part.state === 'output-available'
-  );
 };


packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
```diff
@@ -325,6 +324,19 @@ export function createChatMessagesComponent({
...
+    const showLoader =
+      status === 'submitted' ||
+      (status === 'streaming' && !lastPart) ||
+      // show the loader if the last part is a tool that doesn't have output available yet
+      (status === 'streaming' &&
+        lastPart &&
+        isPartTool(lastPart) &&
+        lastPart.state !== 'output-available') ||
+      // show the loader if the last part is text and we're still streaming (i.e. more text is coming)
+      (status === 'streaming' && lastPart && !isPartText(lastPart));
```
Haroenv:
```suggestion
    const isWaitingForResponse = status === 'submitted';
    const isStreamingWithNoContent = status === 'streaming' && !lastPart;
    const isStreamingNonTextContent =
      status === 'streaming' && lastPart && !isPartText(lastPart);

    const showLoader =
      isWaitingForResponse ||
      isStreamingWithNoContent ||
      isStreamingNonTextContent;
```

your third condition is redundant with the last one as you're checking if it's streaming and the last part is not text, the loader will already show for all tools

Haroenv (commented):
simplification possible I think

packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
shaejaz: Good catch! I had something similar before and must have missed a correct check for the last condition which caused it to not work correctly.

packages/instantsearch-ui-components/src/lib/utils/chat.ts
```diff
@@ -0,0 +1,39 @@
...
+): part is Extract<ChatMessageBase['parts'][number], { type: 'text' }> => {
+  return part.type === 'text';
+};
+
+export const isPartTool = (
+  part: ChatMessageBase['parts'][number]
+): part is ChatToolMessage => {
+  return startsWith(part.type, 'tool-');
+};
+
```
Haroenv:
```suggestion
```

packages/instantsearch-ui-components/src/lib/utils/chat.ts
```diff
@@ -0,0 +1,39 @@
...
+export const toolHasOutput = (message: ChatMessageBase) => {
+  return message.parts.some(
+    (part) => isPartTool(part) && part.state === 'output-available'
+  );
+};
```
Haroenv:
```suggestion
```

packages/instantsearch-ui-components/src/lib/utils/chat.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const getToolParts = (message: ChatMessageBase) => {
+  return message.parts.filter((part) => startsWith(part.type, 'tool-'));
+};
+
+export const hasToolParts = (message: ChatMessageBase) => {
+  return getToolParts(message).length > 0;
+};
+
```
Haroenv:
```suggestion
```

packages/instantsearch-ui-components/src/lib/utils/chat.ts
shaejaz: I was thinking some these would be useful in the renderMessage checks for a later PR, but yes can be added then.

PR #6917, opened 2026-03-09
feat(autocomplete): `noResults` template

aymeric-giraudet:
**Summary**

[FX-3580](https://algolia.atlassian.net/browse/FX-3580)

- Add `noResults` template/component per autocomplete index, rendered when hits are empty
- Auto-hide panel when all content sources (indices, recent searches) are empty and no noResults templates exist

**Result**

- Lint, build, bundlesize, tests all pass

[FX-3580]: https://algolia.atlassian.net/browse/FX-3580?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/react-instantsearch/src/widgets/Autocomplete.tsx
+++ b/packages/react-instantsearch/src/widgets/Autocomplete.tsx
@@ -689,6 +689,16 @@ function InnerAutocomplete<TItem extends BaseHit = BaseHit>({
   );
   const hasWarnedMissingPromptSuggestionsChatRef = useRef(false);
 
+  const allIndicesEmpty = indicesForPanel.every(
+    ({ hits }) => hits.length === 0
+  );
+  const recentEmpty = !storageHits || storageHits.length === 0;
+  const hasNoResultsTemplate = indicesConfig.some(
+    (c) => c.noResultsComponent !== undefined
+  );
+  const shouldHideEmptyPanel =
+    allIndicesEmpty && recentEmpty && !hasNoResultsTemplate && !PanelComponent;
+
   const {
     getInputProps,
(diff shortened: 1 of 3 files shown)


packages/react-instantsearch/src/widgets/Autocomplete.tsx
```diff
@@ -857,7 +869,10 @@ function InnerAutocomplete<TItem extends BaseHit = BaseHit>({
   );
 
   const panelContent = (
-    <AutocompletePanel {...getPanelProps()}>
+    <AutocompletePanel
+      {...getPanelProps()}
+      {...(shouldHideEmptyPanel ? { hidden: true } : {})}
```
shaejaz: Can this not be done in the `getPanelProps`? Would passing the `indicesForPanel` not be ideal?

FabienMotte: If a user provides a custom `panelComponent` / `templates.panel` to render an empty state for example, the new `hidden` logic prevents that content from appearing when all sources are empty.
Should we also check whether panel-level custom content exists in the `shouldHideEmptyPanel` condition?

aymeric-giraudet: Good point, implemented :)

aymeric-giraudet: I just passed `shouldHideEmptyPanel`, looks a bit cleaner

FabienMotte (approved):

PR #6920, opened 2026-03-12
feat(chat): add inline layout to chat widgets

shaejaz:
[FX-3732]

This PR adds an inline layout for the chat widgets. The current 'overlay' layout (which is the default) has also been extracted into a layout.

[FX-3732]: https://algolia.atlassian.net/browse/FX-3732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/__tests__/ChatInlineLayout.test.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/__tests__/ChatInlineLayout.test.tsx
@@ -40,7 +40,7 @@ describe('ChatInlineLayout', () => {
           class="ais-Chat ais-ChatInlineLayout"
         >
           <div
-            class="ais-Chat-container ais-Chat-container--open"
+            class="ais-Chat-container"
           >
             <div
               class="header"
@@ -82,7 +82,7 @@ describe('ChatInlineLayout', () => {
       'ais-Chat ais-ChatInlineLayout ROOT'
     );
     expect(container.querySelector('.ais-Chat-container')!.className).toBe(
-      'ais-Chat-container ais-Chat-container--open CONTAINER'
+      'ais-Chat-container CONTAINER'
     );
   });
 
--- a/tests/common/widgets/chat/options.tsx
+++ b/tests/common/widgets/chat/options.tsx
@@ -1214,7 +1214,7 @@ export function createOptionsTests(
           document.querySelector('.ais-ChatInlineLayout')
         ).toBeInTheDocument();
         expect(
-          document.querySelector('.ais-Chat-container--open')
+          document.querySelector('.ais-Chat-container')
         ).toBeInTheDocument();
         expect(
           document.querySelector('.ais-ChatOverlayLayout')


packages/instantsearch-ui-components/src/components/chat/__tests__/Chat.test.tsx
```diff
@@ -35,7 +39,7 @@ describe('Chat', () => {
     expect(container).toMatchInlineSnapshot(`
       <div>
         <div
-          class="ais-Chat"
+          class="ais-Chat ais-ChatOverlayLayout"
```
Haroenv, line 42: I'm not sure of this name, but also can't think of another

packages/instantsearch-ui-components/src/components/chat/Chat.tsx
```diff
@@ -82,6 +83,26 @@ export type ChatProps = Omit<ComponentProps<'div'>, 'onError' | 'title'> & {
...
+  /**
+   * Function to stop the current streaming response.
+   */
+  stop: ChatLayoutOwnProps['stop'];
+  /**
+   * The current error, if any.
+   */
+  error: ChatLayoutOwnProps['error'];
+  /**
+   * Optional layout component for the chat.
```
Haroenv, line 103:
```suggestion
   * Optional layout component for the chat.
   * @default ChatOverlayLayout
```

packages/instantsearch-ui-components/src/components/chat/Chat.tsx
```diff
@@ -110,64 +135,80 @@ export function createChatComponent({ createElement, Fragment }: Renderer) {
       promptComponent: PromptComponent,
       toggleButtonComponent: ToggleButtonComponent,
       suggestionsComponent: SuggestionsComponent,
+      layoutComponent: LayoutComponent,
```
Haroenv:
```suggestion
      layoutComponent: LayoutComponent = OverlayLayout,
```

then lower down you don't need to add a conditional and a new name, it can stay layout

packages/instantsearch-ui-components/src/components/chat/Chat.tsx
```diff
@@ -110,64 +135,80 @@ export function createChatComponent({ createElement, Fragment }: Renderer) {
...
+            promptProps.promptRef?.current?.focus();
+          }
+        },
+      }
+    );
+
+    const ResolvedLayout = LayoutComponent || OverlayLayout;
+
+    return (
+      <ResolvedLayout
```
Haroenv:
```suggestion
    return (
      <LayoutComponent
```

packages/instantsearch.css/src/components/chat/_chat-overlay-layout.scss
```diff
@@ -0,0 +1,69 @@
+@use '../../shared/_variables';
```
Haroenv, line 1: should this not be in the main chat one, or all of them?

shaejaz: the `use` is required here since the breakpoint scss variable is being used for the overlay styles.

shaejaz: I've also changed the height for the inline to use the existing chat height css variable. I think it makes sense to reuse it instead of making new ones for specific layouts.

Regarding the inline height that we discussed, for now let's leave it as a max height style and once the default greeting and opening prompt suggestions are supported we can think about changing it to use `height`? Right now it would cause a lot of blank white space in the chat. Wdyt?

Haroenv: I think blank height is better, gives more of a reason for us to develop that "home" page

packages/instantsearch.css/src/components/chat/_chat.scss
```diff
@@ -1,32 +1,6 @@
-@use '../../shared/_variables';
```
Haroenv, line 1: this is what I don't understand why it's removed

shaejaz: variables were being imported here since this file required the scss variable `$ais-chat-breakpoint`. since that has been removed, it's no longer required

packages/instantsearch-ui-components/src/components/chat/ChatInlineLayout.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    return (
+      <div
+        {...rest}
+        className={cx('ais-Chat', 'ais-ChatInlineLayout', classNames.root, className)}
+      >
+        <div
+          className={cx(
+            'ais-Chat-container',
+            'ais-Chat-container--open',
```
FabienMotte:
```suggestion
```
I think this can be safely removed for inline layout

PR #6923, opened 2026-03-16
feat(chat): add sidepanel layout to chat widgets

shaejaz:
[FX-3736]

This PR adds an sidepanel layout for the chat widgets. Some additional tests for the new layout component prop have also been added in the common test suite.

[FX-3736]: https://algolia.atlassian.net/browse/FX-3736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSearch.tsx
+++ b/packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteSearch.tsx
@@ -14,6 +14,7 @@ export function createAutocompleteSearchComponent({ createElement }: Renderer) {
   return function AutocompleteSearch(userProps: AutocompleteSearchProps) {
     const { inputProps, onClear, query, isSearchStalled } = userProps;
     const inputRef = inputProps.ref as { current: HTMLInputElement | null };
+
     return (
       <form
         className="ais-AutocompleteForm"
@@ -34,6 +35,7 @@ export function createAutocompleteSearchComponent({ createElement }: Renderer) {
               className="ais-AutocompleteSubmitButton"
               type="submit"
               title="Submit"
+              hidden={isSearchStalled}
             >
               <SubmitIcon createElement={createElement} />
             </button>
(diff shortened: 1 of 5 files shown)


packages/instantsearch-ui-components/src/components/chat/ChatSidePanelLayout.tsx
```diff
@@ -0,0 +1,100 @@
...
+      if (open && !originalMargins.has(element)) {
+        originalMargins.set(element, element.style.marginRight);
+        const chatWidth =
+          getComputedStyle(document.documentElement)
+            .getPropertyValue('--ais-chat-width')
+            .trim() || '22.5rem';
+        const original = originalMargins.get(element);
+        element.style.marginRight = original
+          ? `calc(${original} + ${chatWidth})`
+          : chatWidth;
```
Haroenv, line 58: is this more bullet-proof than getBoundingClientRect?

packages/instantsearch.css/src/components/chat/_chat-sidepanel-layout.scss
```diff
@@ -0,0 +1,83 @@
...
+  top: 0;
+  bottom: 0;
+  width: var(--ais-chat-width);
+  height: 100%;
+  max-width: calc(100% - var(--ais-chat-margin) * 2);
+  z-index: var(--ais-z-index-chat);
+  pointer-events: none;
+
+  @media (prefers-reduced-motion: no-preference) {
+    transition: width var(--ais-transition-duration) var(--ais-transition-timing-function);
```
Haroenv, line 18: in the full screen version / mobile, should we still have the transition?

packages/instantsearch-ui-components/src/components/chat/ChatSidePanelLayout.tsx
shaejaz: You don't get access to the margins with that. You also get the raw width and height values compared to the css styles you get with this approach, maybe that's better tho.
There might be an edge case here that if chat width has been completely overridden by custom user css, then this won't work really.

packages/instantsearch.css/src/components/chat/_chat-sidepanel-layout.scss
shaejaz: Hmm good point, perhaps we would still want the sliding to occur in mobile. I haven't actually tested this completely in mobile view, let me check.

PR #6928, opened 2026-03-16
chore: migrate repo runtime linting from ESLint to oxlint

aymeric-giraudet:
## Summary

This migrates the repo runtime lint flow from ESLint to Oxlint.

The new setup keeps high-signal parity with our existing lint behavior, preserves our repo-specific restrictions through a local Oxlint plugin, and adds faster partial lint workflows for local development.

## What changed

- replaced the root lint runtime path with Oxlint
- added shared `.oxlintrc.json` configs for the repo and relevant nested workspaces
- added a local Oxlint JS plugin for our custom restricted-syntax rules
- updated workspace lint scripts where they are still used locally
- added:
  - `yarn lint`
  - `yarn lint:ox <paths...>`
  - `yarn lint:changed`
  - `yarn lint:staged`
  - `yarn lint:bench`
- removed dead repo-level `.eslintrc.js` files
- updated contributor docs to describe Oxlint + Prettier

## Notes

- this keeps Prettier as the formatter; we are not switching to `oxfmt`
- `eslint` and `eslint-plugin-import` remain installed only because Oxlint still uses the JS-plugin fallback for `import/order`
- template/example ESLint bits that are outside the repo runtime lint path were left in place intentionally
- `oxlint` and `oxlint-tsgolint` are optional dependencies so environments that do not run lint do not fail on the Node engine requirement
- lint environments now use Node `20.19.0` in local/CI tooling, while CodeSandbox keeps the valid major selector `20`

## Validation

Validated locally on March 16, 2026 with:

- `yarn build:ci`
- `yarn lint`
- `yarn type-check`

CircleCI uses the same relevant gates, so this is aligned with CI behavior.

## Benchmark

Recorded ESLint baseline from March 16, 2026:

- `yarn lint`: `110.85s`

Oxlint results from `yarn lint:bench`:

- full lint mean: `9.19s`
- full lint median: `9.03s`
- full lint min: `8.94s`

That is a `91.7%` improvement over the recorded ESLint baseline.

Additional partial-run timings:

- `yarn lint:changed`: mean `0.05s` on the current branch state
- `yarn lint:ox packages/instantsearch.js`: mean `1.74s`

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
@@ -172,7 +172,7 @@ export type ChatMessagesProps<
 };
 
 const copyToClipboard = (message: ChatMessageBase) => {
-  void navigator.clipboard.writeText(getTextContent(message));
+  navigator.clipboard.writeText(getTextContent(message));
 };
 
 function createDefaultMessageComponent<
--- a/packages/instantsearch.js/src/lib/utils/__tests__/debounce-test.ts
+++ b/packages/instantsearch.js/src/lib/utils/__tests__/debounce-test.ts
@@ -6,8 +6,8 @@ describe('debounce', () => {
   it('debounces the function', async () => {
     const originalFunction = jest.fn();
     const debouncedFunction = debounce(originalFunction, 100);
-    void debouncedFunction('a');
-    void debouncedFunction('b');
+    debouncedFunction('a');
+    debouncedFunction('b');
 
     await wait(100);
 
@@ -19,10 +19,10 @@ describe('debounce', () => {
     const originalFunction = jest.fn();
     const debouncedFunction = debounce(originalFunction, 100);
 
-    void debouncedFunction('a');
+    debouncedFunction('a');
 
     setTimeout(() => {
-      void debouncedFunction('b');
+      debouncedFunction('b');
     }, 100);
 
     await wait(250);
@@ -35,7 +35,7 @@ describe('debounce', () => {
     const originalFunction = jest.fn((x) => Promise.resolve(x));
     const debouncedFunction = debounce(originalFunction, 100);
 
-    void debouncedFunction('a');
+    debouncedFunction('a');
 
     const promise = debouncedFunction('b');
     await expect(promise).resolves.toEqual('b');
--- a/packages/instantsearch.js/src/lib/utils/__tests__/hydrateSearchClient-composition-test.ts
+++ b/packages/instantsearch.js/src/lib/utils/__tests__/hydrateSearchClient-composition-test.ts
@@ -55,7 +55,7 @@ describe('hydrateSearchClient (composition)', () => {
 
       hydrateSearchClient(client, initialResults);
 
-      void client.search({
+      client.search({
         compositionID: 'my-composition',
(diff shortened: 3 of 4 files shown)


packages/instantsearch-ui-components/src/lib/stickToBottom.ts
```diff
@@ -593,7 +593,7 @@ export function createStickToBottom({
               : optionsRef.current.initial
           );
 
-          scrollToBottom({
+          void scrollToBottom({
```
Haroenv: why is the void added? is there a default rule related to no side effects?

scripts/lint/instantsearch-oxlint-plugin.mjs
Haroenv, line 1: nice, the format looks pretty similar to eslint plugins. Is this using eslint compat or some builtin method to have plugins?

scripts/lint/benchmark.cjs
Haroenv, line 1: we can remove this I guess, I see a 3m30s -> 20s in CI, which is amazing!

scripts/lint/run-lint.cjs
Haroenv, line 1: is this for the benchmark test?

scripts/lint/instantsearch-oxlint-plugin.mjs
Haroenv, line 1: wonder if it makes sense to have all in one file? with this level of complexity I think so, but one file per rule may make sense too, then some tests could also be added? not sure if worth the effort just for linting

aymeric-giraudet: I had the same thought. I kept it in one file for now because the rules are still small and tightly coupled to this migration, but if we add more repo-specific rules I would split them by rule and add focused tests at that point. I did not want to add a lot of scaffolding yet for a tiny internal plugin.

packages/instantsearch-ui-components/src/lib/stickToBottom.ts
aymeric-giraudet: Good catch. This one was not actually required, so I removed the `void` in 471443e87 after rechecking the file locally.

scripts/lint/run-lint.cjs
aymeric-giraudet: Not the benchmark one. This is the repo-level `yarn lint` entrypoint that runs the root Oxlint pass plus the remaining example and stylelint lint steps. I renamed it to `run-lint.cjs` in 471443e87 to make that clearer.

scripts/lint/benchmark.cjs
aymeric-giraudet: Agreed. I removed the benchmark helper in 471443e87 and kept the measured numbers in the PR description instead.

scripts/lint/instantsearch-oxlint-plugin.mjs
aymeric-giraudet: This is using the Oxlint JS plugin API rather than ESLint compat. The rule shape is intentionally very ESLint-like (`meta` + `create(context)` visitors), but it gets loaded through `.oxlintrc.json` `jsPlugins`. The only ESLint-compat piece left here is `eslint-plugin-import` for the import-order fallback.

Haroenv (approved):

PR #6938, opened 2026-03-23
fix(algoliasearch-helper): convert Infinity/-Infinity to null in _parseNumbers

NixFrog:
**Summary**

`parseFloat("Infinity")` returns JS `Infinity` which passed the `isNaN()` check, causing non-finite values to be sent to the engine.  
This caused 4XX errors when the server-side code enforced stricter integer validation for numeric parameters.

The server-side code never expected to receive strings in numeric parameters, and instead of rejecting them with 4XX codes it triggers weird behaviors.  
This PR aims to remove such cases from the client-side code, so that they can one day be deprecated server-side.

All three `parseFloat` paths in `_parseNumbers` are fixed: `numberKeys`, `insideBoundingBox`, and `numericRefinements`.  
Both string `"Infinity"` and already-numeric `Infinity`/`NaN` values are now converted to `null`.

**Result**

This PR adds dedicated unit tests for the various edge cases I've identified.
Note that server-side, `Infinity` and `-Infinity` are treated the same way (which is the same as any string parameter passed where a numeric parameter is expected). This makes the behavior even more surprising.

The various case map to the following server-side behavior when `"Infinity"` is received (parsed as `-1` by the server-side code):

| Query parameter        | Actual value when receiving `"Infinity"` | Behaviour when receiving `Infinity`              |
|------------------------|------------------------------------------|--------------------------------------------------|
| `page`                 | min=0                                    | 400 "Value too small"                            |
| `hitsPerPage`          | min=0                                    | 400 "Value too small"                            |
| `maxValuesPerFacet`    | SIZE_MAX                                 | 400 "Value too large" (-1 cast to SIZE_MAX)      |
| `distinct`             | N/A                                      | silently ignored |
| `minProximity`         | no bounds                                | silently sets -1                                 |
| `minWordSizefor1Typo`  | min=1                                    | 400 "Value too small"                            |
| `minWordSizefor2Typos` | min=1                                    | 400 "Value too small"                            |
| `minimumAroundRadius`  | min=1                                    | 400 "Value too small"                            |
| `aroundRadius`         | custom parser                            | error, fails radius validation                   |
| `aroundPrecision`      | custom parser                            | precision < 1 resets to default (10)             |
| `getRankingInfo`       | boolean                                  | silently treated as false                        |                 |

I have also ensured that this does not cause any performance regression by running a dedicated benchmark, with scenarios that test both edge cases and more realistic data.


packages/algoliasearch-helper/src/SearchParameters/index.js
```diff
@@ -254,8 +261,18 @@ SearchParameters._parseNumbers = function (partialState) {
     var value = partialState[k];
     if (typeof value === 'string') {
       var parsedValue = parseFloat(value);
-      // global isNaN is ok to use here, value is only number or NaN
-      numbers[k] = isNaN(parsedValue) ? value : parsedValue;
+      if (isNaN(parsedValue)) {
+        // Unparseable string like "all" — keep original
```
Haroenv:
```suggestion
        // originally an unparseable string like "all", keep NaN
```

Haroenv (approved):

PR #6943, opened 2026-03-27
feat(chat): send events from hits in chat

shaejaz:
[FX-3734]

Added `sendEvent` to the Carousel in the Search index tool component.

[FX-3734]: https://algolia.atlassian.net/browse/FX-3734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/widgets/chat/chat.tsx
+++ b/packages/instantsearch.js/src/widgets/chat/chat.tsx
@@ -114,8 +114,8 @@ function createCarouselTool<
       | undefined;
     const hitsWithAbsolutePosition = addAbsolutePosition(
       output?.hits || [],
-      0, // no page info in the tool input/output so assuming page 0
-      input?.number_of_results ?? 5 // defaulting to 5 if number_of_results is not provided
+      0,
+      (input?.number_of_results ?? output?.hits?.length) || 5 // defaulting to 5 if number_of_results is not provided
     );
     const items = addQueryID(hitsWithAbsolutePosition, output?.queryID);
 
--- a/packages/react-instantsearch/src/widgets/chat/tools/SearchIndexTool.tsx
+++ b/packages/react-instantsearch/src/widgets/chat/tools/SearchIndexTool.tsx
@@ -62,8 +62,8 @@ function createCarouselTool<TObject extends RecordWithObjectID>(
 
     const hitsWithAbsolutePosition = addAbsolutePosition(
       output?.hits || [],
-      0, // no page info in the tool input/output so assuming page 0
-      input?.number_of_results ?? 5 // defaulting to 5 if number_of_results is not provided
+      0,
+      (input?.number_of_results ?? output?.hits?.length) || 5
     );
     const items = addQueryID(hitsWithAbsolutePosition, output?.queryID);
 


packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -106,9 +107,14 @@ function createCarouselTool<
       | {
           hits?: Array<RecordWithObjectID<THit>>;
           nbHits?: number;
+          queryID?: string;
         }
       | undefined;
-    const items = output?.hits || [];
+    const items = (output?.hits || []).map((hit, index) => ({
+      ...hit,
+      __queryID: hit.__queryID || output?.queryID,
```
Haroenv: - how does hit already have __queryID? and if it sometimes has it, why do we fall back to the output's query id?
- why not using the addHitPosition / addQueryID function as hits does?

shaejaz: Fixed!

1. I think I initially thought queryID is in the hits but it's actually in the main request body after verifying it from several tool calls. Removed this old code now.
2. Did not think to look in the connectors to see if a util already exists 😅, thanks!

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -106,9 +109,15 @@ function createCarouselTool<
...
           hits?: Array<RecordWithObjectID<THit>>;
           nbHits?: number;
+          queryID?: string;
         }
       | undefined;
-    const items = output?.hits || [];
+    const hitsWithAbsolutePosition = addAbsolutePosition(
+      output?.hits || [],
+      0, // no page info in the tool input/output so assuming page 0
+      input?.number_of_results ?? 5 // defaulting to 5 if number_of_results is not provided
```
Haroenv: can we use the hits.length if not provided? seems better even though it's still wrong of course

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -106,9 +109,15 @@ function createCarouselTool<
       | {
           hits?: Array<RecordWithObjectID<THit>>;
           nbHits?: number;
+          queryID?: string;
         }
       | undefined;
-    const items = output?.hits || [];
+    const hitsWithAbsolutePosition = addAbsolutePosition(
+      output?.hits || [],
+      0, // no page info in the tool input/output so assuming page 0
```
Haroenv:
```suggestion
      0,
```

packages/react-instantsearch/src/widgets/chat/tools/SearchIndexTool.tsx
```diff
@@ -51,10 +56,16 @@ function createCarouselTool<TObject extends RecordWithObjectID>(
...
           nbHits?: number;
+          queryID?: string;
         }
       | undefined;
 
-    const items = output?.hits || [];
+    const hitsWithAbsolutePosition = addAbsolutePosition(
+      output?.hits || [],
+      0, // no page info in the tool input/output so assuming page 0
+      input?.number_of_results ?? 5 // defaulting to 5 if number_of_results is not provided
```
Haroenv: same

packages/instantsearch.js/src/widgets/chat/chat.tsx
shaejaz: good idea, added it now. although since technically the `output.hits` can still be undefined, we need to have a fallback (still kept it as 5 for now)

PR #6945, opened 2026-03-27
feat(chat): add support for feedback

LorrisSaintGenez:
**Summary**

We're rolling out support for Agent Studio /feedback endpoint:
- Feedback is available when using both `agentId` and `feedback` prop on the `Chat` widget
- Works on both VanillaJS and React InstantSearch
- "Thanks for your feedback!" message after submission, even if call fails (no disruption)

##### React InstantSearch
```tsx
<Chat
  agentId="your_agent_id"
  feedback={true}
  // other props
/>
```
 
##### VanillaJS InstantSearch
```js
chat({
  container: '#chat',
  agentId: 'your_agent_id',
  feedback: true,
  // other props
}),
```

**Result**

https://github.com/user-attachments/assets/c6ca3f00-6524-4da5-9c83-d71aa0dcc9db

[GROUT-91]

[GROUT-91]: https://algolia.atlassian.net/browse/GROUT-91?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/__tests__/common-connectors.test.tsx
+++ b/packages/instantsearch.js/src/__tests__/common-connectors.test.tsx
@@ -537,7 +537,7 @@ const testSetups: TestSetupsMap<TestSuites, 'javascript'> = {
     const customChat = connectChat<{
       container: HTMLElement;
     }>((renderOptions) => {
-      const { input, setInput, open, setOpen } = renderOptions;
+      const { input, setInput, open, setOpen, sendChatMessageFeedback, feedbackState } = renderOptions;
       renderOptions.widgetParams.container.innerHTML = `
         <div data-testid="Chat-root" style="display: ${
           open ? 'block' : 'none'
@@ -548,6 +548,8 @@ const testSetups: TestSetupsMap<TestSuites, 'javascript'> = {
         <button data-testid="Chat-toggleButton">
           toggle chat
         </button>
+        <span data-testid="Chat-hasFeedback">${sendChatMessageFeedback ? 'true' : 'false'}</span>
+        <span data-testid="Chat-feedbackState">${JSON.stringify(feedbackState)}</span>
       `;
 
       renderOptions.widgetParams.container
--- a/tests/common/connectors/chat/options.ts
+++ b/tests/common/connectors/chat/options.ts
@@ -78,5 +78,41 @@ export function createOptionsTests(
 
       expect(screen.getByTestId('Chat-root')).toBeVisible();
     });
+
+    test('exposes `sendChatMessageFeedback` when `feedback` is enabled', async () => {
+      const options: SetupOptions<ChatConnectorSetup> = {
+        instantSearchOptions: {
+          indexName: 'indexName',
+          searchClient: createSearchClient(),
+        },
+        widgetParams: {
+          agentId: 'agentId',
+          feedback: true,
+        },
+      };
+
+      await setup(options);
+
+      expect(screen.getByTestId('Chat-hasFeedback')).toHaveTextContent('true');
+      expect(screen.getByTestId('Chat-feedbackState')).toHaveTextContent('{}');
+    });
+
+    test('does not expose `sendChatMessageFeedback` when `feedback` is not enabled', async () => {
+      const options: SetupOptions<ChatConnectorSetup> = {
+        instantSearchOptions: {
+          indexName: 'indexName',
+          searchClient: createSearchClient(),
+        },
+        widgetParams: {
+          agentId: 'agentId',
+        },
+      };
+
+      await setup(options);
+
+      expect(screen.getByTestId('Chat-hasFeedback')).toHaveTextContent(
+        'false'
+      );
+    });
   });
 }


packages/instantsearch.css/src/components/chat/_chat-suggestions.scss
```diff
@@ -12,7 +12,7 @@
   line-height: var(--ais-spacing);
   width: fit-content;
   border-radius: var(--ais-border-radius-lg);
-  padding: calc(var(--ais-spacing) * 0.5) calc(var(--ais-spacing) * 1);
+  padding: calc(var(--ais-spacing) * 0.5) var(--ais-spacing);
```
Haroenv, line 15: why did you change this? just to make bundlesize pass?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -475,6 +501,45 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+              return;
+            }
+            setFeedbackState(messageId, 'sending');
+            sendChatMessageFeedback({
+              agentId,
+              vote,
+              messageId,
+              appId,
+              apiKey,
+              abortSignal: feedbackAbortController!.signal,
```
shaejaz: nit: I don't see a way to send a cancel request with the default UI (at least from the video attached), so I think this abort controller isn't needed for now.

packages/instantsearch.js/src/lib/utils/sendChatMessageFeedback.ts
```diff
@@ -0,0 +1,36 @@
...
+  abortSignal,
+}: {
+  agentId: string;
+  vote: 0 | 1;
+  messageId: string;
+  appId: string;
+  apiKey: string;
+  abortSignal?: AbortSignal;
+}): Promise<Response> {
+  return fetch(`https://${appId}.algolia.net/agent-studio/1/feedback`, {
```
shaejaz, line 14: probably not needed in this first iteration, but perhaps we can allow a separate feedback URL prop in the transport obj to allow custom feedback endpoint.

packages/instantsearch.css/src/components/chat/_chat-suggestions.scss
LorrisSaintGenez: Yes, I made sure to only remove a few default values. Lmk if you'd rather have me increase the limit

packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
```diff
@@ -229,6 +265,50 @@ function createDefaultMessageComponent<
       },
     ];
 
+    const messageFeedback = feedbackState?.[message.id];
```
shaejaz, line 268: If we are using the actions as the 'container' for the feedback UI, I'm thinking that maybe we should also pass these (`onFeedback`, the state etc.) as props to it somehow? Honestly not sure what a good API for it would look like, perhaps separate the feedback into it's own component prop from the chat widgets? 

Wdyt @Haroenv ?

examples/js/default-theme/src/app.js
```diff
@@ -234,6 +234,7 @@ search.addWidgets([
   chat({
     container: '#chat',
     agentId: 'eedef238-5468-470d-bc37-f99fa741bd25',
+    feedback: true,
```
shaejaz, line 237: I'm assuming feedback is not enabled by default on agent studio, otherwise we can default this to true.

packages/instantsearch.js/src/connectors/chat/connectChat.ts
LorrisSaintGenez: Indeed, I'll remove

packages/instantsearch.css/src/components/chat/_chat-suggestions.scss
Haroenv: both are fine options, if we're sure these changes are good (cc @FabienMotte) they can stay

packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
Haroenv: I think for now we can afford putting them entirely in the actions component? but if the logic leaks we should indeed allow people who reimplement actions to have feedback as well.

examples/js/default-theme/src/app.js
LorrisSaintGenez: Yes it's activated by default on AS!
We could default to `true`, it's mostly a product decision, do we want to show this by default to users? If so, I'll make the change

Haroenv: backend accepts it by default, but we think that for e-commerce, and especially as we don't actually use feedback in the generation process, it makes sense to disable by default.

packages/instantsearch.css/src/components/chat/_chat-suggestions.scss
FabienMotte: All good, they are leftovers

packages/instantsearch.js/src/lib/utils/sendChatMessageFeedback.ts
LorrisSaintGenez: Agreed, let's do it in a future iteration. I'd just rather keep this PR focused on AS only

PR #6947, opened 2026-03-31
feat(chat): add tool input streaming support and chat loader component

Haroenv:
[FX-3763]

This PR updates the loading logic to be able to stream tool input deltas. This allows client side tools that receive a large amount of streamed input data to be able to show it be streamed on the browser similar to text deltas (via JSON repairing).

You can opt out of showing the default loader with a flag (per tool):
```jsx
<Chat
  {...}
  tools={{
    tool1: {
      ...,
      streamInput: true, // false by default
    }
  }}
/>
```

Also adds a chat loader component that can be imported

```jsx
<Chat
  {...}
  tools={{
    tool1: {
      layoutComponent: ({ message }) => (
        <>{message.loading && (<ChatMessageLoader />)}</>
      )
    }
  }}
/>
```

```js
chat({
  ...
  tools: {
    tool1: {
      templates: {
        layout: ({ message }, { html }) =>
          html`${message.loading ? chatMessageLoader() : null}`
      }
    }
  }
})
```

[FX-3763]: https://algolia.atlassian.net/browse/FX-3763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/react-instantsearch/src/widgets/chat/tools/SearchIndexTool.tsx
```diff
@@ -92,6 +93,14 @@ function createCarouselTool<TObject extends RecordWithObjectID>(
       );
     }, [items.length, input, output?.nbHits, applyFilters, onClose]);
 
+    if (message.state === 'input-streaming') {
```
shaejaz: Not sure if we still want the `Searching...` text to appear for the search tool.

packages/instantsearch.js/src/lib/ai-lite/abstract-chat.ts
```diff
@@ -26,6 +26,96 @@ type ActiveResponse = {
...
+    return fallbackInput;
+  }
+
+  const directParsed = tryParseJson(normalized);
+  if (directParsed !== undefined) {
+    return directParsed;
+  }
+
+  const repairedParsed = tryParseJson(repairPartialJson(normalized));
+  if (repairedParsed !== undefined) {
```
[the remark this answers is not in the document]
shaejaz: haven't really ran into performance issues yet with the current approach, but something to keep in mind especially for large JSON inputs.

packages/react-instantsearch/src/widgets/chat/tools/SearchIndexTool.tsx
Haroenv: no I don't think so, at least not by default

packages/instantsearch.js/src/lib/ai-lite/abstract-chat.ts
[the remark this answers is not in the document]
Haroenv: yes we can keep in mind that it's not completely optimal and we should use just a single parser

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -237,6 +243,13 @@ export function createChatMessageComponent({ createElement }: Renderer) {
               toolCallId: toolMessage.toolCallId,
             });
 
+          if (
+            toolMessage.state === 'input-streaming' &&
+            tool.showLoaderDuringStreaming
```
Haroenv: I think this should be called `!tool.streamInput`, what do you think @shaejaz?

shaejaz: Yes, that's better 👍

aymeric-giraudet (approved):

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -252,8 +258,9 @@ export function createChatMessageComponent({ createElement }: Renderer) {
                 setIndexUiState={setIndexUiState}
                 addToolResult={boundAddToolResult}
                 applyFilters={tool.applyFilters}
-                sendEvent={tool.sendEvent || (() => {})}
+                sendEvent={tool.sendEvent || (() => { })}
```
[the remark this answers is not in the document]
Haroenv: why was this resolved @shaejaz?

Haroenv (commented):
https://github.com/algolia/instantsearch/blob/a86b6f3c0e43a3ed1ae99e5c21259d890e005025/tests/common/widgets/chat/templates.tsx#L202-L216 -> do we have two different loader options now?

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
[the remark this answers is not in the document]
shaejaz: Oh I thought I addressed it, will fix.
Side note: it seems like the new linting is not a 100% the same as one with eslint, I think this should have been picked up by it.

shaejaz:
>  -> do we have two different loader options now?
 
It's the same `messagesLoaderComponent` if the user wants to customize it. The new `loaderComponent` prop in the tool layout is in case they want to use the default loader component in the tool rendering.

Actually we can replace that with a importable loader component like how it's being done for the chat layouts. It would keep it consistent. wdyt?

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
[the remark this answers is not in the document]
Haroenv: yes, I think we no longer have formatting in oxlint possibly and formatting is still prettier, but separately?

Haroenv:
yes, there should be only one loader I think @shaejaz :)

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -243,6 +243,13 @@ export function createChatMessageComponent({ createElement }: Renderer) {
               toolCallId: toolMessage.toolCallId,
             });
 
+          if (
+            toolMessage.state === 'input-streaming' &&
+            !tool.streamInput
+          ) {
+            return null;
+          }
```
Haroenv, line 251: should this not be the loader?

Haroenv (commented):
I think this is good to go once you merge the conflicts again @shaejaz

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
shaejaz: No because we need to render nothing for the message component since we show a loader in the messages component.

If we don't skip render here, then the layout component is going to start rendering.

PR #6949, opened 2026-04-03
feat(instantsearch.css): add experimental Nova theme

FabienMotte:
**Summary**

Add a new experimental **Nova** theme for `instantsearch.css`. This is based on the work done with @sarahdayan and @thomasbritton.

This theme provides widget-level styles for all core widgets (`SearchBox`, `Hits`, `InfiniteHits`, `RefinementList`, `HierarchicalMenu`, `Menu`, `MenuSelect`, `Pagination`, `Breadcrumb`, `CurrentRefinements`, `NumericMenu`, `RangeInput`, `RatingMenu`, `ToggleRefinement`, `HitsPerPage`, `SortBy`, `GeoSearch`, `Panel`, `Highlight`, `RelevantSort`, `PoweredBy`) along with shared utilities extracted into `_nova-common.scss`:

- Global `prefers-reduced-motion` handling (single rule disabling all transitions/animations)
- `focus-visible` outlines for interactive elements
- Shared button / showMore styles, count badges, and checkbox styles
- Automatic dark mode via `prefers-color-scheme: dark`

New CSS custom properties: `--ais-shadow-xs`, `--ais-hit-min-width`, `--ais-chevron-icon`, `--ais-selected-chevron-icon`, `--ais-check-icon` (with dark-mode variants for icon colors).

Additional changes:

- **Reset**: added resets for links (`text-decoration: none`), selects (`appearance: none`), checkboxes (`margin: 0`), and `.ais-RelevantSort-button`
- **Components**: refactored `prefers-reduced-motion` media queries out of autocomplete, chat, button, and filter-suggestions — transitions are now declared unconditionally and cancelled globally by `_nova-common`
- **Build**: added Nova to the `styles:minify` script and registered `nova.css` / `nova-min.css` in bundlesize config

FX-3750

**Result**

The Nova theme compiles to ~10.5 kB unminified / ~9.75 kB minified. All existing themes (Algolia, Satellite) continue to build and pass bundlesize checks with their updated limits.

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -42,11 +42,11 @@
     },
     {
       "path": "./packages/instantsearch.css/themes/algolia.css",
-      "maxSize": "10.25 kB"
+      "maxSize": "10.5 kB"
     },
     {
       "path": "./packages/instantsearch.css/themes/algolia-min.css",
-      "maxSize": "9.75 kB"
+      "maxSize": "10 kB"
     },
     {
       "path": "./packages/instantsearch.css/themes/reset.css",
@@ -66,11 +66,11 @@
     },
     {
       "path": "./packages/instantsearch.css/themes/satellite.css",
-      "maxSize": "11.25 kB"
+      "maxSize": "11.5 kB"
     },
     {
       "path": "./packages/instantsearch.css/themes/satellite-min.css",
-      "maxSize": "10.5 kB"
+      "maxSize": "10.75 kB"
     },
     {
       "path": "./packages/instantsearch.css/components/chat.css",


packages/instantsearch.css/src/components/chat/_chat-carousel.scss
```diff
@@ -181,10 +179,8 @@
     overflow: hidden;
 
     img {
-      @media (prefers-reduced-motion: no-preference) {
```
Haroenv, line 184: on purpose removed these conditionals?

FabienMotte: Yes, these are now handled as a global rule in `_nova-common.scss`:
```css
// Reduced motion
@media (prefers-reduced-motion: reduce) {
  [class^='ais-'],
  [class^='ais-'] *,
  [class^='ais-']::before,
  [class^='ais-']::after,
  [class^='ais-'] *::before,
  [class^='ais-'] *::after {
    transition: none !important;
    animation: none !important;
  }
}
```
The only downside is that the two old themes won't have this global rule, since it's only used for the new Nova theme.

FabienMotte: I'll move this to `_common.scss`

packages/instantsearch.css/src/widgets/_breadcrumb.scss
shaejaz, line 1: if these are widget styles specific to the new theme, should these be moved a more explicit `nova` dir?
not sure what the right approach is, to me if these are not generic then it seems confusing to have these in just a `widgets` folder.

Haroenv: maybe they're generic and should be used by future different themes as well that set different variables?

shaejaz: yes that could also be a possibility, but I saw a lot of styles that seemed to be tied to the theme itself (such as borders, box-shadows, padding etc). perhaps those can be extracted out for the theme specific widget styles?

FabienMotte: Good idea, should we just change the folder name to `nova-widgets` to show that those styles are part of the new theme?

I think trying to split each widget file into "generic part" vs "nova-specific part" would be dificult to maintain, WDYT?

Haroenv: let's for now only make the new theme widgets, and if we need to make another theme, do the split then, or maybe if another team needs it, they can do it?

Haroenv (approved):
I think it's easier for now to consider al the widgets the "nova-version". If we'd want the other themes to use this too, they could later, of we can refactor this to have a "base" and "nova" version of the styles later too. For now as we only have the nova styles, I'm ok to commit to this version

PR #6954, opened 2026-04-08
feat(helper): support multifeed composition responses

e-krebs:
## Summary

- Omit `queriesCount` from composition state entries in `_runComposition`; in `_dispatchAlgoliaResponse`, use `results` directly when `queriesCount` is `undefined` (composition path) instead of splicing
- Build `lastResults.feeds` as an ordered `CompositionSearchResults[]` array in `_dispatchAlgoliaResponse` when the response contains `feedID`s
- Add `CompositionSearchResults` interface (extends `SearchResults` with `feedID: string`) and `feeds?: CompositionSearchResults[]` on `SearchResults` class
- `lastResults` points to `feeds[0]` — feeds live on `SearchResults` so they survive the SSR `getInitialResults` → JSON → hydration round-trip naturally

This is **Layer 1** of the [multifeed composition RFC](https://algolia.atlassian.net/wiki/spaces/MERCH/pages/6889570339/RFC+-+Multifeed+Composition+Support+in+InstantSearch). Layer 2 (connector) will consume `results.feeds` to route per-feed results to distinct widget trees.
[EMERCH-2612](https://algolia.atlassian.net/browse/EMERCH-2612)

## Test plan

- [x] New unit tests in `packages/algoliasearch-helper/test/spec/algoliasearch.helper/composition.js` (8 tests)
- [x] Full helper test suite passes (`yarn workspace algoliasearch-helper test` — 456 tests)
- [x] Full monorepo test suite passes (`yarn test`)
- [x] No lint or type-check regressions in helper package

[EMERCH-2612]: https://algolia.atlassian.net/browse/EMERCH-2612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/algoliasearch-helper/index.d.ts
+++ b/packages/algoliasearch-helper/index.d.ts
@@ -1431,21 +1431,6 @@ declare namespace algoliasearchHelper {
     _state: SearchParameters;
 
     /**
-     * The feed identifier, present when this result comes from a multifeed composition response.
-     */
-    feedID?: string;
-
-    /**
-     * Map of feed ID to SearchResults, populated when the composition response contains multiple feeds.
-     */
-    _feedResults?: Record<string, SearchResults>;
-
-    /**
-     * Ordered list of feed IDs matching the composition response order.
-     */
-    _feedOrder?: string[];
-
-    /**
      * Marker which can be added to search results to identify them as created without a search response.
      * This is for internal use, e.g., avoiding caching in infinite hits, or delaying the display of these results.
      */
--- a/packages/algoliasearch-helper/index.d.ts
+++ b/packages/algoliasearch-helper/index.d.ts
@@ -1431,6 +1431,21 @@ declare namespace algoliasearchHelper {
     _state: SearchParameters;
 
     /**
+     * The feed identifier, present when this result comes from a multifeed composition response.
+     */
+    feedID?: string;
+
+    /**
+     * Map of feed ID to SearchResults, populated when the composition response contains multiple feeds.
+     */
+    _feedResults?: Record<string, SearchResults>;
+
+    /**
+     * Ordered list of feed IDs matching the composition response order.
+     */
+    _feedOrder?: string[];
+
+    /**
      * Marker which can be added to search results to identify them as created without a search response.
      * This is for internal use, e.g., avoiding caching in infinite hits, or delaying the display of these results.
      */
--- a/packages/algoliasearch-helper/index.d.ts
+++ b/packages/algoliasearch-helper/index.d.ts
@@ -1521,6 +1521,18 @@ declare namespace algoliasearchHelper {
      * @return all the refinements
      */
     getRefinements(): SearchResults.Refinement[];
+
+    /** Ordered array of per-feed results, present when results come from a multifeed composition response. */
+    feeds?: CompositionSearchResults[];
+  }
+
+  /**
+   * SearchResults from a composition feed response.
+   * Each feed in the composition response produces one of these.
+   */
+  export interface CompositionSearchResults extends SearchResults {
+    /** The feed identifier for this result set. */
+    feedID: string;
   }
 
   export type Banner = {
--- a/packages/algoliasearch-helper/src/algoliasearch.helper.js
+++ b/packages/algoliasearch-helper/src/algoliasearch.helper.js
@@ -1695,8 +1695,6 @@ AlgoliaSearchHelper.prototype._runComposition = function () {
 
     states.push({
       state: derivedState,
-      // Infinity so splice grabs all feed results from a multifeed composition response
-      queriesCount: Infinity,
       helper: derivedHelper,
     });
 
@@ -1875,7 +1873,9 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function (
     var state = s.state;
     var queriesCount = s.queriesCount;
     var helper = s.helper;
-    var specificResults = results.splice(0, queriesCount);
+    var specificResults = queriesCount !== undefined
+      ? results.splice(0, queriesCount)
+      : results;
 
     if (!state.index) {
       helper.emit('result', {
--- a/packages/algoliasearch-helper/src/algoliasearch.helper.js
+++ b/packages/algoliasearch-helper/src/algoliasearch.helper.js
@@ -1695,7 +1695,8 @@ AlgoliaSearchHelper.prototype._runComposition = function () {
 
     states.push({
       state: derivedState,
-      queriesCount: derivedStateQueries.length,
+      // Infinity so splice grabs all feed results from a multifeed composition response
+      queriesCount: Infinity,
       helper: derivedHelper,
     });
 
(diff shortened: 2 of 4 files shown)


packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1695,7 +1695,8 @@ AlgoliaSearchHelper.prototype._runComposition = function () {
 
     states.push({
       state: derivedState,
-      queriesCount: derivedStateQueries.length,
+      // Infinity so splice grabs all feed results from a multifeed composition response
+      queriesCount: Infinity,
```
Haroenv: could we remove the queriesCount check for composition and just read all instead of slicing? not sure if there's pros to this approach as well

packages/algoliasearch-helper/index.d.ts
```diff
@@ -1430,6 +1430,21 @@ declare namespace algoliasearchHelper {
     _rawResults: Array<SearchResponse<T>>;
     _state: SearchParameters;
 
+    /**
+     * The feed identifier, present when this result comes from a multifeed composition response.
+     */
+    feedID?: string;
```
Haroenv: should we put these in a separate interface and merge to make it clearer only included if you have composition?

e-krebs: I asked myself the same question and dismissed it as "it's going to have wild ripple effect",
so I'll have a deeper look & properly assess if it's actually the case or not.

packages/algoliasearch-helper/src/algoliasearch.helper.js
e-krebs: omitting works and only introduces a ternary in `_dispatchAlgoliaResponse` below

packages/algoliasearch-helper/index.d.ts
e-krebs: after having a look, I'm still unsure so I opted for delaying the decision to the next PR where it's actually going to be used

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1884,12 +1885,33 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function (
...
     }
 
-    helper.lastResults = new SearchResults(
-      state,
-      specificResults,
-      self._searchResultsOptions
-    );
-    if (rawContent !== undefined) helper.lastResults._rawContent = rawContent;
+    // Multifeed composition: build per-feed SearchResults map
+    if (specificResults.length > 0 && specificResults[0].feedID) {
```
Haroenv, line 1889:
```suggestion
    if (specificResults[0].feedID && specificResults.length > 0) {
```

keep it clearer that we're talking about composition. Ideally there would be a cheaper way to check on the main results itself, but it's ok.

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1884,12 +1885,33 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function (
...
+        feedResults[r.feedID] = sr;
+      });
+      helper.lastResults = new SearchResults(
+        state,
+        [specificResults[0]],
+        self._searchResultsOptions
+      );
+      if (rawContent !== undefined) helper.lastResults._rawContent = rawContent;
+      helper.lastResults._feedResults = feedResults;
+      helper.lastResults._feedOrder = feedOrder;
```
Haroenv: I'm not overly sold on putting this behind _ keys that we then need to explicitly handle in the implementation. It's ok bu have you considered `helper.feeds = feedResults (as an array)` directly? then still setting lastResults to the main one. that way it feels like more of a proper implementation rather than an add on.

also why is there the results and the order needed separately? could it be just a single array that in case you have a feature of displaying just a single feed it just does find in the array.

Haroenv (commented):
good start, of course pretty hypothetical still here

packages/algoliasearch-helper/src/algoliasearch.helper.js
e-krebs: actually, we shouldn't do this.
this would error in case specificResults is an empty array

packages/algoliasearch-helper/src/algoliasearch.helper.js
e-krebs: as discussed I went with `helper.lastResults.feeds` so that feeds are automatically serialised when SSR

e-krebs: since its now one property only I added back feeds on to `SearchResults` (with dedicated extended typing).
let me know if that works for you

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1884,12 +1885,23 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function (
...
-    if (rawContent !== undefined) helper.lastResults._rawContent = rawContent;
+    // Multifeed composition: build ordered SearchResults array on lastResults
+    if (specificResults.length > 0 && specificResults[0].feedID) {
+      var feeds = specificResults.map(function (r) {
+        var sr = new SearchResults(state, [r], self._searchResultsOptions);
+        if (rawContent !== undefined) sr._rawContent = rawContent;
+        return sr;
+      });
+      helper.lastResults = feeds[0];
+      helper.lastResults.feeds = feeds;
```
Haroenv: we can discuss this during the server side rendering ticket, but now the main feed will be twice in the serialisation. I don't really see a solution for it, as you need the main as well as specific feeds.

Haroenv (approved):
this works, thanks! let's merge it once all the PRs are ready though, in case we want to make changes when we implement the other parts

packages/algoliasearch-helper/src/algoliasearch.helper.js
e-krebs: the only way to avoid duplication that I can think of would be to remove the first item from `feeds` and always assume that the full list of feeds is `[lastResults, ...lastResults.feed]`but that's prone to error as it's very easy to forget `lastResults` when going through all feeds

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1884,12 +1885,29 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function (
...
+        return sr;
+      });
+      // Separate instance from feeds[0] to avoid circular ref in JSON.stringify
+      helper.lastResults = new SearchResults(
+        state,
+        [specificResults[0]],
+        self._searchResultsOptions
+      );
+      if (rawContent !== undefined) helper.lastResults._rawContent = rawContent;
+      helper.lastResults.feeds = feeds;
```
Haroenv:
```suggestion
      helper.lastResults.feeds = feeds;
      if (rawContent !== undefined) helper.lastResults._rawContent = rawContent;
```

packages/algoliasearch-helper/src/algoliasearch.helper.js
```diff
@@ -1884,12 +1885,29 @@ AlgoliaSearchHelper.prototype._dispatchAlgoliaResponse = function (
...
-    );
-    if (rawContent !== undefined) helper.lastResults._rawContent = rawContent;
+    // Multifeed composition: build ordered SearchResults array on lastResults
+    if (specificResults.length > 0 && specificResults[0].feedID) {
+      var feeds = specificResults.map(function (r) {
+        var sr = new SearchResults(state, [r], self._searchResultsOptions);
+        if (rawContent !== undefined) sr._rawContent = rawContent;
+        return sr;
+      });
+      // Separate instance from feeds[0] to avoid circular ref in JSON.stringify
```
Haroenv:
```suggestion
```

don't think this comment is needed, the json stringify issue either should be in a test or obvious in server side rendering usage.

Haroenv (commented):
👍

Haroenv (approved):

PR #6963, opened 2026-04-10
refactor(ai-lite): remove unsafe onToolCall any cast

Haroenv:
**Summary**

Type tool chunks by UI message tool definitions and propagate typed chunks through transport/parser.

Replace the onToolCall any cast with InferUIMessageToolCall and keep dynamic tool calls explicitly flagged.

**Result**

internal type doesn't cast as any


FabienMotte (commented):
Should we cast the object passed to `new Chat(...)` in `connectChat.ts` to fix the type error?

FabienMotte (approved):

shaejaz (approved):

PR #6964, opened 2026-04-10
feat(instantsearch): add connectFeeds connector for multifeed compositions

e-krebs:
## Summary

Layer 2 of the Multifeed Composition Support ([RFC](https://algolia.atlassian.net/wiki/spaces/MERCH/pages/6889570339/RFC+-+Multifeed+Composition+Support+in+InstantSearch)). Stacked on `feat/composition-multifeed-layer1-helper` (PR #6954).

- **`connectFeeds` connector**: validates `compositionID`, computes `feedIDs` from `results.feeds`, supports `transformFeeds` for reordering/filtering, single-feed backward compat. Stateless `getWidgetRenderState` — widget layers own FeedContainer lifecycle
- **`FeedContainer`**: implements `IndexWidget` with `$$type: 'ais.feedContainer'`, `_isolated: true`. Reads per-feed results from `parentIndex.getResults().feeds` by `feedID`. Shares parent helper, delegates URL/search to parent, propagates child widget lifecycle. Uses `initialized` flag to prevent double-init of child widgets
- **Centralized `indexWidgetTypes`**: single source of truth (`const + type`) for index-like widget types, used by `isIndexWidget`, `IndexWidgetDescription`, `BuiltinTypes`, `BuiltinWidgetTypes`
- **`storeRenderState` extraction**: moved from module-private in `index.ts` to shared util in `render-args.ts`
- **Metadata middleware fix**: uses `isIndexWidget()` instead of hardcoded `'ais.index'` check — crawler now sees widgets inside feed containers (separate commit for cherry-pick)
- **Shared test helpers**: `createResultsWithFeeds` and `createParentWithHelper` extracted to `test/createFeedsTestHelpers.ts`

## Key design decisions

- **`searchScope: 'global'` required** — future-proofing for per-feed search parameters (see RFC)
- **FeedContainer in parent widget tree** — `walkIndex` finds them (needed for SSR), `addWidgets` triggers `scheduleSearch` (includes facet declarations from child widgets)
- **Connector is lifecycle-agnostic** — `connectFeeds` only computes `feedIDs`; framework widgets (Layer 3) own FeedContainer create/register/remove

## Test plan

- [x] 42 unit tests for FeedContainer and connectFeeds
- [x] TypeScript: `yarn run type-check:v3` and `yarn run type-check:v4` — no new errors
- [x] Lint: `yarn lint:changed`
- [x] CI green

### DIFF
--- a/packages/instantsearch.js/src/connectors/feeds/FeedContainer.ts
+++ b/packages/instantsearch.js/src/connectors/feeds/FeedContainer.ts
@@ -117,7 +117,7 @@ export function createFeedContainer(
           { uiState: {} }
         );
         if (withChildParams !== parentHelper.state) {
-          parentHelper.state = withChildParams;
+          parentHelper.setState(withChildParams);
         }
       }
 
--- a/packages/instantsearch.js/src/connectors/feeds/FeedContainer.ts
+++ b/packages/instantsearch.js/src/connectors/feeds/FeedContainer.ts
@@ -131,18 +131,39 @@ export function createFeedContainer(
       );
       const helper = parentIndex.getHelper();
 
+      if (!helper) {
+        localWidgets = localWidgets.filter((w) => !flatWidgets.includes(w));
+        return container;
+      }
+
+      // Chain through children's dispose so widgets clean up the
+      // SearchParameters they declared (e.g. RefinementList removes its
+      // disjunctiveFacet) instead of leaving them stale on the parent helper.
+      let cleanedState: SearchParameters = helper.state;
+
       flatWidgets.forEach((widget) => {
-        if (widget.dispose && helper) {
-          widget.dispose({
+        if (widget.dispose) {
+          const next = widget.dispose({
             helper,
-            state: helper.state,
+            state: cleanedState,
             recommendState: helper.recommendState,
             parent: container,
           });
+
+          if (next instanceof algoliasearchHelper.RecommendParameters) {
+            // ignore — FeedContainer doesn't manage recommend state
+          } else if (next) {
+            cleanedState = next;
+          }
         }
       });
 
       localWidgets = localWidgets.filter((w) => !flatWidgets.includes(w));
+
+      if (cleanedState !== helper.state) {
+        helper.setState(cleanedState);
+      }
+
       return container;
     },
 
--- a/packages/instantsearch.js/src/connectors/feeds/FeedContainer.ts
+++ b/packages/instantsearch.js/src/connectors/feeds/FeedContainer.ts
@@ -1,0 +1,285 @@
+import algoliasearchHelper from 'algoliasearch-helper';
+
+import {
+  createInitArgs,
+  createRenderArgs,
+  storeRenderState,
+} from '../../lib/utils';
+
+import type {
+  InstantSearch,
+  UiState,
+  IndexUiState,
+  Widget,
+  IndexWidget,
+  DisposeOptions,
+  RenderOptions,
+} from '../../types';
+import type { SearchParameters } from 'algoliasearch-helper';
+
+export function createFeedContainer(
+  feedID: string,
+  parentIndex: IndexWidget,
+  instantSearchInstance: InstantSearch
+): IndexWidget {
+  let localWidgets: Array<Widget | IndexWidget> = [];
+
+  const container: IndexWidget = {
+    $$type: 'ais.feedContainer',
+    $$widgetType: 'ais.feedContainer',
+    _isolated: true,
+
+    getIndexName: () => parentIndex.getIndexName(),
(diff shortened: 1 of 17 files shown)


packages/instantsearch.js/src/connectors/feeds/connectFeeds.ts
```diff
@@ -0,0 +1,212 @@
...
+        const containersToRemove: IndexWidget[] = [];
+        feedContainers.forEach((container, id) => {
+          if (!activeFeedIDs.has(id)) {
+            containersToRemove.push(container);
+            feedContainers.delete(id);
+          }
+        });
+        if (containersToRemove.length > 0) {
+          // Deferred removal — same pattern as connectDynamicWidgets
+          setTimeout(() => parent.removeWidgets(containersToRemove), 0);
```
Haroenv: in dynamic widgets we have the connector being responsible for adding/removing the widgets, but in the end in react and vue we're not really using it and the logic is in the widget instead of the connector. Is that similar here, do you see a better approach?

e-krebs: I'm not sure yet to be honest, I can keep that in mind once working on the react & vue widgets in the next PR.
but even if we end up with a similar approach (where in react and vue the logic is in the widget), this connector logic is still useful for the vanilla.js version, right?

Haroenv: can be useful, but in retrospect I think I would have preferred the logic for the js widget to be all in the widget too. Depends what it looks like, the logic still may be easy to move.

e-krebs: OK, the js widget is next step so I'll see if that approach works 🙂

e-krebs: the PR with the widget is almost ready but in the meantime I updated this one with the new approach

packages/instantsearch.js/src/connectors/feeds/FeedContainer.ts
```diff
@@ -0,0 +1,289 @@
...
+        // parent's helper state so they're included in the composition request.
+        // uiState is {} because URL-derived refinements are already on the
+        // parent state; children only need to declare structural params.
+        const parentHelper = parentIndex.getHelper()!;
+        const withChildParams = container.getWidgetSearchParameters(
+          parentHelper.state,
+          { uiState: {} }
+        );
+        if (withChildParams !== parentHelper.state) {
+          parentHelper.state = withChildParams;
```
FabienMotte: Should we use `setState(...)`? By assigning the state directly, no `change` event is emitted

FabienMotte (commented):
It looks like `FeedsWidgetDescription['indexRenderState']` is missing in `packages/instantsearch.js/src/types/render-state.ts`

packages/instantsearch.js/src/connectors/feeds/FeedContainer.ts
```diff
@@ -0,0 +1,289 @@
...
+      flatWidgets.forEach((widget) => {
+        if (widget.dispose && helper) {
+          widget.dispose({
+            helper,
+            state: helper.state,
+            recommendState: helper.recommendState,
+            parent: container,
+          });
+        }
+      });
```
FabienMotte, line 159: I think we should chain the state returned by each `widget.dispose()`, the same way `index.removeWidgets()` does. Some widgets clean up the `SearchParameters` they added by returning a new state from `dispose()`, so ignoring that return value can leave stale facets/refinements in the helper after the widget is removed

e-krebs: this should be addressed now

FabienMotte:
> It looks like `FeedsWidgetDescription['indexRenderState']` is missing in `packages/instantsearch.js/src/types/render-state.ts`

@e-krebs Not sure if you saw my comment above

FabienMotte (approved):

packages/instantsearch.js/src/connectors/feeds/connectFeeds.ts
```diff
@@ -0,0 +1,143 @@
...
+export type FeedsRenderState = {
+  feedIDs: string[];
+};
+
+export type FeedsConnectorParams = {
+  /**
+   * Explicit search scope. Currently only 'global' is supported
+   * (future-proofing for per-feed search parameters).
+   */
+  searchScope: 'global';
```
Haroenv, line 23: this should be `isolated: false` (true not allowed, so we can change default later)

Haroenv (approved):
think naming can be fixed, but as this widget is mainly internal, we can do it after initial merge too

PR #6967, opened 2026-04-13
fix(examples): match showcase dark mode background to docs color

FabienMotte:
**Summary**

Update the widget showcase example's dark mode background color from `neutral-950` to `#0a0b11` to match the documentation site's background color.

https://github.com/algolia/docs-new/pull/722

**Result**

The dark mode body background in `examples/js/showcase/src/style.css` now uses `#0a0b11` instead of Tailwind's `neutral-950`, ensuring visual consistency with the docs.

### DIFF
--- a/examples/js/showcase/src/style.css
+++ b/examples/js/showcase/src/style.css
@@ -13,12 +13,12 @@ body {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   color: theme(--color-neutral-900);
-  background-color: theme(--color-white);
+  background-color: theme(--color-white); /* docs light bg */
 }
 
 [data-theme="dark"] body {
   color: theme(--color-neutral-100);
-  background-color: #0a0b11;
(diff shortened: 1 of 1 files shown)


examples/js/showcase/src/style.css
```diff
@@ -13,12 +13,12 @@ body {
...
   -moz-osx-font-smoothing: grayscale;
   color: theme(--color-neutral-900);
-  background-color: theme(--color-white);
+  background-color: theme(--color-white); /* docs light bg */
 }
 
 [data-theme="dark"] body {
   color: theme(--color-neutral-100);
-  background-color: theme(--color-neutral-950);
+  background-color: #0a0b11; /* docs dark bg */
```
Haroenv, line 21: should we override the variables in :root of the showcase styles? or is only the background relevant

FabienMotte: The background color of the docs isn't a Tailwind color, so I used an hex color instead

Haroenv (approved):

PR #6973, opened 2026-04-14
feat(chat): add hidden context support for completions

afrencalg:
**Summary**

Allow sending additional context (e.g. current page info) with each user message via a `context` parameter. Context is injected as a `context` part in user messages, sent to the API but not rendered in the chat UI.

**Testing** 

Screenshot 2026-04-20 at 13 06 34
Screenshot 2026-04-20 at 13 06 13

### DIFF
--- a/examples/react/getting-started/src/App.tsx
+++ b/examples/react/getting-started/src/App.tsx
@@ -95,10 +95,6 @@ export function App() {
             agentId="eedef238-5468-470d-bc37-f99fa741bd25"
             feedback={true}
             itemComponent={ItemComponent}
-            context={() => ({
-              currentPage: window.location.href,
-              userLocale: navigator.language,
-            })}
           />
         </InstantSearch>
       </div>
--- a/examples/react/getting-started/src/App.tsx
+++ b/examples/react/getting-started/src/App.tsx
@@ -95,6 +95,10 @@ export function App() {
             agentId="eedef238-5468-470d-bc37-f99fa741bd25"
             feedback={true}
             itemComponent={ItemComponent}
+            context={() => ({
+              currentPage: window.location.href,
+              userLocale: navigator.language,
+            })}
           />
         </InstantSearch>
       </div>
--- a/packages/instantsearch.js/src/connectors/chat/connectChat.ts
+++ b/packages/instantsearch.js/src/connectors/chat/connectChat.ts
@@ -23,6 +23,7 @@ import type {
   ChatInit as ChatInitAi,
   UIMessage,
 } from '../../lib/chat';
+import type { ContextUIPart } from '../../lib/ai-lite';
 import type { SendEventForHits } from '../../lib/utils';
 import type {
   Connector,
@@ -159,6 +160,13 @@ export type ChatConnectorParams<TUiMessage extends UIMessage = UIMessage> = (
    * @default 'chat'
    */
   type?: string;
+  /**
+   * Additional context to send with each user message (e.g. current page info).
+   * This context is included in the message parts sent to the API but is not
+   * displayed in the chat UI.
+   * Can be a static object or a function that returns the context at send time.
+   */
+  context?: Record<string, unknown> | (() => Record<string, unknown>);
 };
 
 export type ChatWidgetDescription<TUiMessage extends UIMessage = UIMessage> = {
@@ -260,6 +268,7 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
       resume = false,
       tools = {},
       type = 'chat',
+      context,
       ...options
     } = widgetParams || {};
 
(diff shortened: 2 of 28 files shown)


packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -611,6 +620,53 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+          const contextPart: ContextUIPart = {
+            type: 'context',
+            context: resolvedContext,
+          };
+
+          if ('parts' in message && message.parts) {
+            return _chatInstance.sendMessage(
+              {
+                ...message,
+                parts: [...message.parts, contextPart],
```
Haroenv: should the context not be first?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -611,6 +620,53 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+            return _chatInstance.sendMessage(message, options);
+          }
+
+          const resolvedContext =
+            typeof context === 'function' ? context() : context;
+
+          const contextPart: ContextUIPart = {
+            type: 'context',
+            context: resolvedContext,
+          };
```
Haroenv: is this sent verbatim to the server like this? is type: context a supported backend option?

Haroenv (commented):
seems like an interesting approach, has it been tested in real life yet?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -611,6 +619,43 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+            type: 'text' as const,
+            text: `<context>${JSON.stringify(resolvedContext)}</context>`,
+          };
+
+          if ('parts' in message && message.parts) {
+            return _chatInstance.sendMessage({
+              ...message,
+              parts: [contextTextPart, ...message.parts],
+              text: undefined,
+              files: undefined,
```
shaejaz, line 673: Not aware of these two props. Would there be any consequence to setting these as undefined always?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -159,6 +159,13 @@ export type ChatConnectorParams<TUiMessage extends UIMessage = UIMessage> = (
    * @default 'chat'
    */
   type?: string;
+  /**
+   * Additional context to send with each user message (e.g. current page info).
+   * This context is included in the message parts sent to the API but is not
+   * displayed in the chat UI.
+   * Can be a static object or a function that returns the context at send time.
+   */
+  context?: Record<string, unknown> | (() => Record<string, unknown>);
```
shaejaz, line 168: What's the use case for having this also be a function type?

examples/react/getting-started/src/App.tsx
```diff
@@ -95,6 +95,10 @@ export function App() {
             agentId="eedef238-5468-470d-bc37-f99fa741bd25"
             feedback={true}
             itemComponent={ItemComponent}
+            context={() => ({
```
shaejaz: For now, I think we should skip having the context use case in the examples. To me it's a more advanced use case.

shaejaz (commented):
Would also be nice to add tests in the common test suite for these changes.

packages/instantsearch.js/src/connectors/chat/connectChat.ts
afrencalg: No runtime consequence.

packages/instantsearch.js/src/connectors/chat/connectChat.ts
afrencalg: Allows lazy evaluation at send time.

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -630,6 +638,48 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+            return _chatInstance.sendMessage(message, ...rest);
+          }
+
+          const resolvedContext =
+            typeof context === 'function' ? context() : context;
+
+          const contextTextPart = {
+            type: 'text' as const,
+            text: '<context>'
+              .concat(JSON.stringify(resolvedContext))
```
FabienMotte: Should we add a `try/catch` block in case the context is not serializable?

afrencalg: Fixed!

FabienMotte (approved):

PR #6974, opened 2026-04-14
feat(metadata): send meta events via insights

tkrugg:
## Summary

- Add a single anonymous `__start__` usage event, sent once when an InstantSearch instance boots, through the existing insights events pipeline — removing the need to crawl every InstantSearch implementation to understand how the library is used
- Report the instance configuration and the full widget tree, plus version/agent and a bootstrap timing measurement
- Persist a usage session ID in `sessionStorage` so it survives reloads within the same browser session

## How it works

Usage tracking is embedded in the insights middleware. When insights is active, a single `__start__` event is sent on the instance's first render, by which point all widgets have registered. The event flows through the search-insights transport (`sendEvents`), avoiding direct HTTP calls and inheriting ad-blocker resilience.

The event is **browser-only**: it never fires during server-side rendering (e.g. Next.js), so a page view is counted exactly once, on the client.

The old metadata middleware (crawler-based) is unchanged and still active.

## Event shape

```json
{
  "eventType": "instantsearch",
  "eventName": "__start__",
  "timestamp": 1710691200000,
  "sessionID": "f47ac10b-...",
  "userToken": "...",
  "version": "...",
  "performance": { "bootstrapMs": 87 },
  "widgets": [ /* ais.instantSearch root + widget tree */ ]
}
```

## Privacy

Options that carry credentials or end-user data — `searchClient`, `insightsClient`, and `initialUiState` — are never included. Functions are reported by name only, and usage tracking is wrapped so it can never throw into the host app.

## Bootstrap timing

`bootstrapMs` measures the time from the `instantsearch()` constructor to first render. The absolute value is flavor-dependent (React mounts widgets asynchronously, ~30ms; JS/Vue register synchronously, ~4ms), so it's best read as a per-flavor regression signal rather than compared across flavors.


packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -416,8 +423,68 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
...
+
+        // telemetry
+        const telemetrySessionId = getTelemetrySessionId();
+
+        function sendTelemetryEvent(event: any) {
+          const userToken = helper.state?.userToken;
+
+          (insightsClientWithLocalCredentials as any)('sendEvents', [
+            {
+              eventType: 'instantsearch_telemetry',
```
Haroenv:
```suggestion
              eventType: 'instantsearch',
```

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -416,8 +423,68 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
...
+          telemetryTimeout = null;
+          try {
+            sendTelemetryEvent({
+              eventName: '__start__',
+              algolia_agent: getAlgoliaAgent(instantSearchInstance.client),
+              version,
+              user_agent: safelyRunOnBrowser(
+                ({ window }) => window.navigator.userAgent,
+                { fallback: () => undefined }
+              ),
```
Haroenv: duplicate with the auto headers

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -416,8 +423,68 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
...
+              application_id: appId,
+              performance: {
+                bootstrap_ms: Math.round(
```
Haroenv: explanation that this number is time between the constructor starting and the metadata start event being able to be sent, therefore measuring the time it takes to add widgets if they are added right at start.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -416,8 +423,64 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
...
+              userToken: userToken ? String(userToken) : undefined,
+              ...event,
+            },
+          ]);
+        }
+
+        // Defer by a tick so the measurement is taken after `start()`
+        // fully returns (started() is called from within start()), and so
+        // widgets added asynchronously in the same tick (e.g. Vue) are
+        // included in the tree.
```
shaejaz: check if this comment is correct

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
```diff
@@ -416,8 +423,64 @@ See documentation: https://www.algolia.com/doc/guides/building-search-ui/going-f
...
+                bootstrapMs: Math.round(
+                  (typeof performance !== 'undefined'
+                    ? performance.now()
+                    : Date.now()) - instantSearchInstance._createdAt
+                ),
+              },
+              widgets: [
+                {
+                  type: 'ais.instantSearch',
+                  params: [],
```
Haroenv: can we get instantsearchoptions too?

packages/instantsearch.js/src/helpers/insights.ts
```diff
@@ -1,7 +1,25 @@
...
+    const existing = sessionStorage.getItem(USAGE_SESSION_KEY);
+    if (existing) {
+      return existing;
+    }
+    const id = createUUID();
+    sessionStorage.setItem(USAGE_SESSION_KEY, id);
+    return id;
+  } catch {
+    // sessionStorage unavailable (SSR, privacy mode, etc.)
+    return createUUID();
```
tkrugg: I know this is something I wrote, just second guessing myself and would like your opinion.
Ideally, even on a private session/ssr, you'd like `getUsageSessionId` to always return the same uuid. Which isn't the case currently.

Also reading it from the session multiple while on the same runtime is unecessary slower. 

Would it make sense to store it locally in a variable?
```
export function getUsageSessionId(): string {
    let uuid = null
    return (() => {
        if (uuid) return uuid
    
        ....
    })()
}
```

packages/instantsearch.js/src/helpers/insights.ts
```diff
@@ -1,7 +1,25 @@
 import { warning, serializePayload, deserializePayload } from '../lib/utils';
+import { createUUID } from '../lib/utils/uuid';
 
 import type { InsightsClientMethod, InsightsClientPayload } from '../types';
 
+const USAGE_SESSION_KEY = 'ais.usage.sessionId';
```
tkrugg, line 6: @Haroenv I think you proposed to change the term USAGE to something different?

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: Applied in 649689517. The event currently ships as `eventType: 'instantsearch'`, so this thread can be resolved.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: The `user_agent` field this was flagging was removed in 44bed171e. The remaining `algoliaAgent` in the payload is the search client's agent string (algoliasearch + instantsearch.js + react-instantsearch), distinct from the `X-Algolia-Agent` header search-insights sets, so it is not a duplicate. Happy to drop `applicationId` too if you think it duplicates anything.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: Added an inline comment on the start-event send describing that `bootstrapMs` is the time between the constructor running and the first render, so for flavors that register widgets right at start (React) it captures the cost of adding them. Let me know if you want it sharpened.

packages/instantsearch.js/src/helpers/insights.ts
sarahdayan: @Haroenv do you have a preferred term? Happy to rename `getUsageSessionId` / `USAGE_SESSION_KEY` once we settle on it.

packages/instantsearch.js/src/helpers/insights.ts
sarahdayan: Done in 68ea3f332. The id is now cached in a module-level variable and only read from `sessionStorage` once. If storage is unavailable we fall back to a fresh UUID for the runtime.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: Looks correct as written. Marking as resolved.

packages/instantsearch.js/src/middlewares/createInsightsMiddleware.ts
sarahdayan: Done in 0b87a9e32. The root `ais.instantSearch` node now reports the instance options via `_initialOptions`, filtered through `serializeWidgetParams`. Sensitive keys (`searchClient`, `insightsClient`, `initialUiState`) are omitted.

packages/instantsearch.js/src/lib/utils/buildWidgetTree.ts
```diff
@@ -0,0 +1,112 @@
...
+    try {
+      return { value: JSON.stringify(value), type: 'object' };
+    } catch {
+      return { type: 'object' };
+    }
+  }
+
+  // functions, symbols, DOM elements — name only
+  return null;
+}
```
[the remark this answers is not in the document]
sarahdayan: @tkrugg @Haroenv quick clarification needed before we change anything. The PR description says functions are reported by name only, but two readings are possible: (1) name of the option key (current behavior, yields `{ name }` only), or (2) the function's own `fn.name` (would yield `{ name, value: fn.name, type: 'function' }`). The first is what ships today but carries no extra signal beyond presence. The second is strictly more useful but is a small behavior change. Which did you intend?

packages/instantsearch.js/src/lib/utils/extractWidgetPayload.ts
```diff
@@ -0,0 +1,59 @@
...
+      params,
+    });
+
+    if (widget.$$type === 'ais.index') {
+      extractWidgetPayload(
+        (widget as IndexWidget).getWidgets(),
+        instantSearchInstance,
+        payload
+      );
+    }
```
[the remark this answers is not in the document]
sarahdayan: Fixed in 9c06d1d66, both utils now use `isIndexWidget` to recurse into `ais.feedContainer` as well as `ais.index`.

packages/instantsearch.js/src/lib/utils/buildWidgetTree.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    const params = serializeWidgetParams(widgetParams);
+
+    const children =
+      widget.$$type === 'ais.index'
+        ? buildWidgetTree(
+            (widget as IndexWidget).getWidgets(),
+            instantSearchInstance
+          )
+        : [];
```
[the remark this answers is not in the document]
sarahdayan: Fixed in 9c06d1d66, both utils now use `isIndexWidget` to recurse into `ais.feedContainer` as well as `ais.index`.

packages/instantsearch.js/src/middlewares/__tests__/createInsightsMiddleware.ts
```diff
@@ -1315,7 +1315,7 @@ describe('insights', () => {
         } as any,
       });
       expect(analytics.viewedObjectIDs).toHaveBeenCalledTimes(0);
-      expect(onEvent).toHaveBeenCalledTimes(1);
+      // onEvent also receives usage events (bootstrap)
       expect(onEvent).toHaveBeenCalledWith(
```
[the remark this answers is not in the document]
sarahdayan: The inline comment is correct and the test count stays strict at 1. No change needed.

tests/common/widgets/hits/insights.ts
```diff
@@ -82,9 +82,9 @@ export function createInsightsTests(
         ).toHaveLength(hitsPerPage);
       }
 
-      // View event called for each index once
+      // View event called for each index once (+ usage events)
       {
-        expect(window.aa).toHaveBeenCalledTimes(2);
+        expect(window.aa.mock.calls.length).toBeGreaterThanOrEqual(2);
```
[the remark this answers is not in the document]
sarahdayan: Tightened in 9138ee7ba. All `viewedObjectIDs` count assertions now filter mock calls by method name so they stay strict.

tests/common/widgets/hits/insights.ts
```diff
@@ -133,7 +133,7 @@ export function createInsightsTests(
           await wait(0);
         });
 
-        expect(window.aa).toHaveBeenCalledTimes(2);
+        expect(window.aa.mock.calls.length).toBeGreaterThanOrEqual(2);
```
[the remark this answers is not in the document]
sarahdayan: Tightened in 9138ee7ba. All `viewedObjectIDs` count assertions now filter mock calls by method name so they stay strict.

tests/common/widgets/hits/insights.ts
```diff
@@ -195,9 +195,9 @@ export function createInsightsTests(
         ).toHaveLength(hitsPerPage);
       }
 
-      // View event called for each index, batched in chunks of 20
+      // View event called for each index, batched in chunks of 20 (+ usage events)
       {
-        expect(window.aa).toHaveBeenCalledTimes(4);
+        expect(window.aa.mock.calls.length).toBeGreaterThanOrEqual(4);
```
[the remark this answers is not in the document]
sarahdayan: Tightened in 9138ee7ba. All `viewedObjectIDs` count assertions now filter mock calls by method name so they stay strict.

tests/common/widgets/infinite-hits/insights.ts
```diff
@@ -79,9 +79,9 @@ export function createInsightsTests(
         ).toHaveLength(hitsPerPage);
       }
 
-      // View event called for each index once
+      // View event called for each index once (+ usage events)
       {
-        expect(window.aa).toHaveBeenCalledTimes(2);
+        expect(window.aa.mock.calls.length).toBeGreaterThanOrEqual(2);
```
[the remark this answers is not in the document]
sarahdayan: Tightened in 9138ee7ba. All `viewedObjectIDs` count assertions now filter mock calls by method name so they stay strict.

tests/common/widgets/infinite-hits/insights.ts
```diff
@@ -130,7 +130,7 @@ export function createInsightsTests(
           await wait(0);
         });
 
-        expect(window.aa).toHaveBeenCalledTimes(2);
+        expect(window.aa.mock.calls.length).toBeGreaterThanOrEqual(2);
```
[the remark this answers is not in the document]
sarahdayan: Tightened in 9138ee7ba. All `viewedObjectIDs` count assertions now filter mock calls by method name so they stay strict.

tests/common/widgets/infinite-hits/insights.ts
```diff
@@ -191,9 +191,9 @@ export function createInsightsTests(
         ).toHaveLength(hitsPerPage);
       }
 
-      // View event called for each index, batched in chunks of 20
+      // View event called for each index, batched in chunks of 20 (+ usage events)
       {
-        expect(window.aa).toHaveBeenCalledTimes(4);
+        expect(window.aa.mock.calls.length).toBeGreaterThanOrEqual(4);
```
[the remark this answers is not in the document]
sarahdayan: Tightened in 9138ee7ba. All `viewedObjectIDs` count assertions now filter mock calls by method name so they stay strict.

packages/instantsearch.js/src/lib/utils/buildWidgetTree.ts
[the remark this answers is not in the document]
Haroenv: Not sure I get the distinction, but the function name should be useful if we have it (although it's often going to be minified anyway)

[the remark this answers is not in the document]
sarahdayan: Addressed in [c5adfd0](https://github.com/algolia/instantsearch/pull/6974/commits/c5adfd0e886bb333d2fb011e0c45c459e20eea22).

packages/instantsearch.js/src/helpers/insights.ts
Haroenv: it's ok, I proposed the whole middleware to be called middleware as it is now, usage is fine as the key I think

tests/common/shared/insights.ts
Haroenv, line 1: I feel like this would be a good test file for asserting the __start__ event, what do you think?

Haroenv (approved):
implementation looks good, unit tests too, I think we can have one more "integration" assertion in cts

shaejaz (approved):
lgtm with the new changes, thanks!

PR #6976, opened 2026-04-15
feat(autocomplete): default insights events

Haroenv:
**Summary**

adds default click events to autocomplete

**Result**

same sort of implementation as for hits, autocomplete has click events by default (if insights is enabled)

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -10,11 +10,11 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "122 kB"
+      "maxSize": "123 kB"
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "253.4 kB"
+      "maxSize": "260 kB"
     },
     {
       "path": "packages/react-instantsearch-core/dist/umd/ReactInstantSearchCore.min.js",
--- a/packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
+++ b/packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
@@ -2,7 +2,7 @@
 
 import { cx } from '../../lib/cx';
 
-import type { ComponentProps, Renderer } from '../../types';
+import type { ComponentProps, Renderer, SendEventForHits } from '../../types';
 
 export type AutocompleteIndexProps<
   T = { objectID: string; __indexName: string } & Record<string, unknown>
@@ -22,6 +22,7 @@ export type AutocompleteIndexProps<
     onSelect: () => void;
     onApply: () => void;
   };
+  sendEvent?: SendEventForHits;
   classNames?: Partial<AutocompleteIndexClassNames>;
 };
 
@@ -56,6 +57,7 @@ export function createAutocompleteIndexComponent({ createElement }: Renderer) {
       ItemComponent,
       NoResultsComponent,
       getItemProps,
+      sendEvent,
       classNames = {},
     } = userProps;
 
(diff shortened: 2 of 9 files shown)


packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
```diff
@@ -93,6 +95,20 @@ export function createAutocompleteIndexComponent({ createElement }: Renderer) {
                     classNames.item,
                     className
                   )}
+                  onClick={() => {
+                    sendEvent?.(
+                      'click:internal',
+                      item as Record<string, unknown>,
+                      'Hit Clicked'
+                    );
+                  }}
```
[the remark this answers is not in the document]
Haroenv: there's no click handler in itemProps

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
```diff
@@ -93,6 +95,20 @@ export function createAutocompleteIndexComponent({ createElement }: Renderer) {
...
+                  onAuxClick={() => {
+                    sendEvent?.(
+                      'click:internal',
+                      item as Record<string, unknown>,
+                      'Hit Clicked'
+                    );
+                  }}
```
FabienMotte, line 111: Should we check for `event.button === 1` to only allow middle-clicks? (filter out right-clicks)

tests/common/widgets/autocomplete/insights.tsx
```diff
@@ -0,0 +1,273 @@
...
+  delay = 100,
+}: { hitsPerPage?: number; delay?: number } = {}) {
+  return createSearchClient({
+    search: jest.fn(async (requests) => {
+      await wait(delay);
+      return createMultiSearchResponse(
+        ...requests.map(
+          ({ indexName }: Parameters<SearchClient['search']>[0][number]) =>
+            createSingleSearchResponse({
+              index: indexName,
```
FabienMotte, line 32: Should we add `queryID` and assert it below in the 3 tests?

bundlesize.config.json
```diff
@@ -10,11 +10,11 @@
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.production.min.js",
-      "maxSize": "122 kB"
+      "maxSize": "123 kB"
     },
     {
       "path": "./packages/instantsearch.js/dist/instantsearch.development.js",
-      "maxSize": "253.4 kB"
+      "maxSize": "260 kB"
```
FabienMotte, line 17: The bundle size bump seems large, is it expected?

Haroenv: yes, as development grows faster than prod, there's no need to keep the limit as tight. Much of it is whitespace and error messages.

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
Haroenv: we don't for hits either, so I think we should keep it the same

packages/react-instantsearch/src/widgets/Autocomplete.tsx
```diff
@@ -869,6 +869,7 @@ function InnerAutocomplete<TItem extends BaseHit = BaseHit>({
           __indexName: indexId,
         }))}
         getItemProps={getItemProps}
+        sendEvent={indices.find((idx) => idx.indexId === indexId)?.sendEvent}
```
shaejaz, line 872: could we have the `sendEvent` in the `indicesConfig`? that way we could do something like `currentIndexConfig.sendEvent` and keep it consistent with the rest of the props here.

not sure if the difference between indexId and indexName matters for this tho.

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
```diff
@@ -93,6 +95,20 @@ export function createAutocompleteIndexComponent({ createElement }: Renderer) {
                     classNames.item,
                     className
                   )}
+                  onClick={() => {
+                    sendEvent?.(
+                      'click:internal',
+                      item as Record<string, unknown>,
```
FabienMotte: Do we need to cast here?

packages/instantsearch-ui-components/src/components/autocomplete/AutocompleteIndex.tsx
```diff
@@ -93,6 +95,20 @@ export function createAutocompleteIndexComponent({ createElement }: Renderer) {
...
+                      'Hit Clicked'
+                    );
+                  }}
+                  onAuxClick={() => {
+                    sendEvent?.(
+                      'click:internal',
+                      item as Record<string, unknown>,
```
FabienMotte: Same as previous comment

FabienMotte (approved):

PR #6978, opened 2026-04-15
feat(chat): add `initialMessages` option

aymeric-giraudet:
## Summary

Adds an `initialMessages` option to the Chat connector, as discussed in #6956.

- `initialMessages` pre-populates the chat with messages on initialization **without triggering an AI response**
- Only applied when the chat has no existing messages (e.g., not restored from storage)
- Not applied when `resume` is enabled
- Can be used alongside `initialUserMessage` — `initialMessages` sets the displayed messages, then `initialUserMessage` sends a message that triggers AI

This serves a different purpose from `initialUserMessage`: use `initialMessages` for welcome messages or pre-seeded conversation history, and `initialUserMessage` to automatically trigger an AI response on init.

## Test plan

- [x] Connector test: sets `initialMessages` on init
- [x] Connector test: does not set `initialMessages` when messages already exist
- [x] Connector test: applies `initialMessages` and sends `initialUserMessage` together
- [x] Widget test (JS + React): sets `initialMessages` on init
- [x] Widget test (JS + React): does not set `initialMessages` when messages already exist
- [x] Lint passes with 0 errors
- [x] Type-check passes (no new errors)

### DIFF
--- a/packages/instantsearch.js/src/connectors/chat/connectChat.ts
+++ b/packages/instantsearch.js/src/connectors/chat/connectChat.ts
@@ -185,6 +185,9 @@ export type ChatConnectorParams<TUiMessage extends UIMessage = UIMessage> = (
    * are not applied.
    *
    * When `resume` is enabled, these messages are not applied.
+   *
+   * `initialUserMessage` is sent after `initialMessages` are applied, so an
+   * assistant welcome followed by a user prompt works.
    */
   initialMessages?: TUiMessage[];
 };
@@ -578,7 +581,7 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
 
         // Set initialMessages before registering callbacks to avoid
         // triggering re-renders during init
-        if (initialMessages && !resume && !hasExistingMessages) {
+        if (initialMessages?.length && !resume && !hasExistingMessages) {
           _chatInstance.messages = initialMessages;
         }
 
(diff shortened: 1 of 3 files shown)


packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -554,6 +566,14 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
           };
         }
 
+        const hasExistingMessages = _chatInstance.messages.length > 0;
+
+        // Set initialMessages before registering callbacks to avoid
+        // triggering re-renders during init
+        if (initialMessages && !resume && !hasExistingMessages) {
```
FabienMotte, line 573:
```suggestion
        if (initialMessages?.length && !resume && !hasExistingMessages) {
```

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -169,6 +169,17 @@ export type ChatConnectorParams<TUiMessage extends UIMessage = UIMessage> = (
...
+  /**
+   * Messages to pre-populate the chat with when it is initialized.
+   *
+   * These messages are set without triggering an AI response. They are only
+   * applied when the chat has no existing messages yet. If messages were
+   * restored or otherwise already exist when the widget starts, these messages
+   * are not applied.
+   *
+   * When `resume` is enabled, these messages are not applied.
+   */
```
FabienMotte, line 181: Should we add something along those lines:
> `initialUserMessage` is sent after `initialMessages` are applied, so an assistant welcome followed by a user prompt works

Haroenv (approved):

PR #6991, opened 2026-04-24
feat(chat): add error matching and API-facing display rules

pipeline1987:
## Summary

Centralize chat error classification and user-facing strings in chat utils: thread depth, recursion, max steps, max output, rate limiting, and request origin.


Haroenv (commented):
this is a lot of code to flatten errors. Is it possible we implement this backend to ensure the error and its message get put in a less nested part of the response?

PR #6995, opened 2026-04-27
fix(react-instantsearch-nextjs): preserve URL on mount with nested <Index>

### LINKED ISSUE #6980: InstantSearchNext routing removes query param from url
### 🐛 Current behavior

Hi!
We are implementing Algolia in Next.js with SSR. When using InstantSearchNext with routing, query is being dropped from search url. I tried debuging this issue and found out that [this](https://github.com/algolia/instantsearch/blob/5775e0de9c9e784a0d0056e0414ab1f8b442b27d/packages/react-instantsearch-nextjs/src/useInstantSearchRouting.ts#L69) is the line causing removal of query param. I couldn't track what was causing the issue at root as stack trace is not sufficient.

We are using only Algolia hooks and I tried to reproduce similar structure in Stackblitz. I simplified routing so we can only track query.
I also tried adding [skipSuspense](https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react#composing-hooks) but it didn't change anything - in [example](https://github.com/algolia/instantsearch/blob/master/examples/react/next-app-router/app/multiple-hooks/Search.tsx#L31) `WithSkipSuspense` has two hooks but only first one has skipSuspense - it's confusing as in documentation it's written `This should be done to all hooks but the last one in your custom component`.

Please see reproduction on Stackblitz: https://stackblitz.com/edit/nextjs-tdznfimz

What is important - this happens from time to time, if you don't get it on first reproduction, try to refresh or search something else. I also have some workaround solution but when this bug happens it causes other issues later on with filters. I found similar issue: https://github.com/algolia/instantsearch/issues/6679

```
  const { query, refine } = useSearchBox()

  // Workaround for query being removed from URL - history.pushState({}, '', url);
  useEffect(() => {
    if (initialQuery !== query) {
      refine(initialQuery)
    }
  }, [initialQuery, query, refine])
```

### 🔍 Steps to reproduce

1. Type something in search input in header - it will add query param to url and redirect to search results page
2. Even though initialUiState is passed and SearchBox hook is added, sometimes query param is removed and products don't reflect search query
3. You can use `back` browser button and it will redirect back to correct results and add query param in url, so something is removing it

### Live reproduction

https://stackblitz.com/edit/nextjs-tdznfimz

### 💭 Expected behavior

Query parameter is not removed from url and correct search results are displayed based on initialUiState passed to InstantSearchNext

### Package version

react-instantsearch-nextjs@1.1.4, react-instantsearch@7.29.0, react@19.2.5, next@16.2.3

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

FabienMotte:
**Summary**

Fixes #6980 and CR-11362.

`InstantSearchNext` was intermittently dropping query parameters from the URL on mount when widgets were nested inside an `<Index>` wrapper.

The router middleware's `subscribe()` already merges the URL into `_initialUiState` before `mainIndex.init` runs, so the `[pathname, searchParams]` effect in `useInstantSearchRouting` was firing a redundant `setUiState` on mount. With a nested `<Index>`, the inner index's children only mount on a second render pass, so when this redundant update propagated down, the nested helper had no widgets registered yet. `helper.setState` then rebuilt an empty `localUiState`, and the router middleware later wrote that empty state back to the URL.

The fix skips the on-mount fire of that effect (tracking the previous route via a ref) so we only react to genuine post-mount `pathname` / `searchParams` changes.

**Result**

Added `packages/react-instantsearch-nextjs/src/__tests__/useInstantSearchRouting.test.tsx` with three cases:
- preserves URL on mount with a nested `<Index>` and custom `stateMapping` (fails without the fix)
- preserves URL on mount without a nested `<Index>` (regression check)
- reacts to pathname/searchParams changes after mount

### DIFF
--- a/packages/react-instantsearch-nextjs/src/__tests__/useInstantSearchRouting.test.tsx
+++ b/packages/react-instantsearch-nextjs/src/__tests__/useInstantSearchRouting.test.tsx
@@ -1,0 +1,175 @@
+/**
+ * @jest-environment @instantsearch/testutils/jest-environment-jsdom.ts
+ */
+
+import { createSearchClient } from '@instantsearch/mocks';
+import { wait } from '@instantsearch/testutils';
+import { act, render } from '@testing-library/react';
+import React from 'react';
+import { Index, SearchBox } from 'react-instantsearch';
+
+import { InstantSearchNext } from '../InstantSearchNext';
+
+const mockPathname = jest.fn();
+const mockSearchParams = jest.fn();
+jest.mock('next/navigation', () => ({
+  ...jest.requireActual('next/navigation'),
+  usePathname() {
+    return mockPathname();
+  },
+  useSearchParams() {
+    return mockSearchParams();
+  },
+}));
+
+describe('routing', () => {
+  beforeEach(() => {
+    mockPathname.mockReturnValue('/search');
+    mockSearchParams.mockReturnValue(new URLSearchParams());
+    window.history.replaceState({}, '', '/search');
+  });
+
+  // Reproduces https://github.com/algolia/instantsearch/issues/6980:
+  // a nested <Index> only mounts its children on a second render pass, so
+  // the on-mount [pathname, searchParams] effect used to wipe the URL.
+  it('preserves URL on mount with a nested <Index> and custom stateMapping', async () => {
+    const indexName = 'instant_search';
+    const indexId = `${indexName}_web`;
+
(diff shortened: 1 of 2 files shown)


Haroenv (approved):
looks good. Can this have any side-effects?

afrencalg (approved):

FabienMotte:
> looks good. Can this have any side-effects?

@Haroenv It should not, I added two tests to check for regressions

PR #6997, opened 2026-04-27
feat(agentStudio): cost control generic error handling

pipeline1987:
## Summary

Wire chat error rendering to the API's real error contract, let apps customize what (if any) raw API message reaches end users, and make "clear conversation" actually start a fresh server-side thread.

Paired with three upstream PRs on algolia/conversational-ai (fix-stream-error-double-encoding, unify-error-field-name, unify-validation-error-shape) that converge both transports on the shared ErrorResponse shape {"message": "...", ...}. With those landed, the client becomes a single JSON.parse + .message read shared by both transports via one internal tryParseErrorMessage helper.

## What changes

- HTTP transport: parses .message from JSON error bodies, falls back to HTTP error: <status>.
- Stream transport: parses .message from type: "error" chunk errorText, falls back to raw text, then "Unknown error".
- Conversation id reset: clearMessages rotates the chat id via resetConversationId() so the next request starts a fresh backend thread (previously the id was created once at construction).
- UI plumbing: ChatMessages accepts error?: Error and forwards error.message to ChatMessageError (or a custom errorComponent) when status === 'error'. Wired through both vanilla and React Chat widgets.
- Customizable error copy: translations.errorMessage widens to string | (({ errorMessage }) => string). Default copy is unchanged — raw API messages stay hidden unless apps explicitly opt in.
- Robustness: processStream catches synchronous throws from onChunk and routes them to onError (instead of unhandled rejection + leaked reader lock).

## Result

image


packages/instantsearch.js/src/lib/ai-lite/utils.ts
```diff
@@ -92,3 +92,69 @@ export function resolveValue<T>(
...
+ * double-encoded JSON strings returned by some APIs.
+ */
+export function getMessageFromStreamErrorText(errorText: string): string {
+  const trimmed = errorText.trim();
+  if (!trimmed) {
+    return 'Unknown error';
+  }
+
+  let remaining: unknown = trimmed;
+  for (let depth = 0; depth < 5; depth++) {
```
Haroenv: why are we going to a depth of 5? It's not tested. I think either a recursion or the real depth may make more sense

packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
```diff
@@ -510,7 +516,16 @@ export function createChatMessagesComponent({
...
 
-            {status === 'error' && <DefaultError onReload={onReload} />}
+            {status === 'error' && (
+              <DefaultError
+                onReload={onReload}
+                translations={
+                  error?.message
+                    ? { errorMessage: error.message }
+                    : undefined
+                }
```
Haroenv, line 539: the user translation should get access to the error message as well, is that possible? otherwise we start exposing messages they may not be aware of.

Haroenv (commented):
i think the main question is why we need to handle recursion and looping at all and the errors aren't just top-level

pipeline1987:
> i think the main question is why we need to handle recursion and looping at all and the errors aren't just top-level

afaik, the agent ai runs on LangGraph as a state graph with multiple LLM/tool steps, so there is a real recursion/step budget to enforce—not a single flat request/response. With streaming, the HTTP response has already started (200 + event stream), so failures that occur mid-run cannot be turned into a new top-level HTTP status; they have to be carried inside the stream protocol instead. it follows from how streaming and graph execution interact, but I could probably ask someone to check if that could be changed.

packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
pipeline1987: Thanks, done— errorMessage is passed now in; translations.errorMessage

Haroenv:
I suggest fixing the double encoding in https://github.com/algolia/conversational-ai/pull/1223,  the message mapping in https://github.com/algolia/conversational-ai/pull/1224 and the simplification in https://github.com/algolia/conversational-ai/pull/1225

Haroenv (changes requested):
I made three simplifications in API that I prefer to see over this. Also I think a future change that can be made in the API is changing from sending errorMessage as a stringified json with message in it, it should just be the message. That's a larger change that should be done later.

peterszalontay-algolia:
Please double check this also works with guardrails, and the message should not be retry but instead it should be "Start a new conversation"

peterszalontay-algolia:
guardrails event stream
data: {"type": "start", "messageId": "alg_msg_d959ATpesXBpPd0L", "messageMetadata": {"displayResultsEnabled": true}}

data: {"type": "data-guardrail-violation", "data": {"category": "customer_support", "guardrailType": "input", "fallbackResponse": "Please reach out via the customer support chat"}}

data: {"type": "finish"}

data: [DONE]

packages/instantsearch-ui-components/src/components/chat/ChatMessageError.tsx
```diff
@@ -2,28 +2,42 @@
...
-  errorMessage: string;
+  errorMessage:
+    | string
+    | ((params: { errorMessage?: string }) => string);
   /**
-   * Retry button text
+   * New conversation button text
    */
-  retryText: string;
+  newConversationText: string;
```
Haroenv, line 19: does this imply that every error now only prompts for new conversation? maybe network or other transient errors should have a retry still? not sure if we can easily distinguish.

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -378,7 +378,7 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
     const onClearTransitionEnd = () => {
       setMessages([]);
       _chatInstance.clearError();
-      _chatInstance.regenerateId();
+      _chatInstance.resetConversationId();
```
Haroenv, line 381: on purpose?

Haroenv (commented):
sorry for the delay, couple small questions

packages/instantsearch-ui-components/src/components/chat/ChatMessageError.tsx
pipeline1987: We could land this as-is and address transient-vs-permanent in a follow-up PR?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
pipeline1987: Yep, `resetConversationId` is the new proposed way, as the single canonical API (cleaner OOP encapsulation, no MutableKeys cast hack needed as regenerateId did).

packages/instantsearch.js/src/lib/ai-lite/abstract-chat.ts
```diff
@@ -1091,8 +1104,36 @@ export abstract class AbstractChat<TUIMessage extends UIMessage> {
...
+                const violationData = (chunk as any).data as
+                  | { fallbackResponse?: string }
+                  | undefined;
+
+                if (currentMessageIndex >= 0) {
+                  this.state.messages = this.state.messages.slice(
+                    0,
+                    currentMessageIndex
+                  );
+                }
```
FabienMotte:
```suggestion
                if (currentMessageIndex >= 0) {
                  this.state.messages = this.state.messages.slice(
                    0,
                    currentMessageIndex
                  );
                  currentMessage = undefined;
                  currentMessageIndex = -1;
                }
```

I think we should also reset those otherwise `onFinish` will receive the `currentMessage` that no longer exists in the transcript?

packages/instantsearch-ui-components/src/components/chat/ChatMessageError.tsx
```diff
@@ -39,17 +53,26 @@ export function createChatMessageErrorComponent({
 
   return function ChatMessageError(userProps: ChatMessageErrorProps) {
     const {
-      onReload,
+      errorMessage,
+      onReload: _onReload,
```
FabienMotte: `onReload` is still a public prop, but this component now ignores it, so existing callers lose the default retry button. Can we either keep rendering retry when `onReload` is provided, or remove/replace the prop with an explicit action API instead of leaving it dead?

packages/instantsearch.js/src/lib/ai-lite/abstract-chat.ts
```diff
@@ -1091,8 +1104,36 @@ export abstract class AbstractChat<TUIMessage extends UIMessage> {
...
+
+              // Surface guardrail violations through the error state, but
+              // distinct from generic cost-control / 4xx errors: throw a
+              // `GuardrailViolationError` so the UI can render the
+              // service-provided `fallbackResponse` verbatim (it's authored
+              // for end-user display) instead of the friendly default used
+              // for opaque transport errors. Also discard any in-progress
+              // assistant message so no partial text lingers above the
+              // fallback.
+              if (chunkType === 'data-guardrail-violation') {
```
FabienMotte: `data-guardrail-violation` is handled through `any` in the generic fallback. Can we add a typed chunk (in `UIMessageChunk`) for it and handle it explicitly instead?

packages/instantsearch.js/src/widgets/chat/chat.tsx
```diff
@@ -414,6 +414,7 @@ function ChatWrapper({
       }}
       messagesProps={{
```
FabienMotte, line 415: Should we pass `onNewConversation: clearMessages` here so the default “Start a new conversation” error action is actually wired in the widget (same comment for the React widget)?

FabienMotte (changes requested):

packages/instantsearch.js/src/lib/ai-lite/abstract-chat.ts
pipeline1987: Good catch, done. Also folded the branch into a proper case so the reset lives in a single typed handler (per your other comment).

packages/instantsearch-ui-components/src/components/chat/ChatMessageError.tsx
pipeline1987: Restored. onReload now renders a Retry button again. Precedence: actions > onNewConversation > onReload. Connector only forwards onNewConversation, so guardrails still get no Retry by default.

packages/instantsearch.js/src/lib/ai-lite/abstract-chat.ts
pipeline1987: Done. Added data-guardrail-violation to UIMessageChunk and handled it as a dedicated case — no more any casts.

packages/instantsearch.js/src/widgets/chat/chat.tsx
pipeline1987: Done in both JS and React widgets

Haroenv (approved):

FabienMotte (approved):

PR #7002, opened 2026-04-29
feat(chat): identify component and referer in agent requests

afrencalg:
**Summary**

Send `x-algolia-component: ais-chat` on every chat completions request, and forward an `algolia-referer` header (`prompt-suggestions` or `ai-mode`) when the message originates from the autocomplete prompt suggestions or the autocomplete/search-box AI Mode button. This lets the agent attribute traffic without parsing the user agent.

Also clean up the request body: previously every parameter received by `prepareSendMessagesRequest` (`headers`, `api`, `body`, `credentials`, `requestMetadata`, …) was spread back into the body, duplicating values already sent in the headers and HTTP metadata. The body now contains only `{ id, messageId, messages }` for the agent path, and the equivalent minimal payload for the default custom-transport path.


packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
```diff
@@ -626,7 +626,10 @@ function AutocompleteWrapper<TItem extends BaseHit>({
           if (chatRenderState) {
             chatRenderState.setOpen?.(true);
             chatRenderState.focusInput?.();
-            chatRenderState.sendMessage?.({ text: item.prompt });
+            chatRenderState.sendMessage?.(
+              { text: item.prompt },
+              { headers: { 'algolia-referer': 'prompt-suggestions' } }
```
FabienMotte: Should we keep the convention `x-algolia-referer`?

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
```diff
@@ -868,6 +868,120 @@ data: [DONE]`,
...
+      it('adds the x-algolia-component header on agent requests', async () => {
+        const { widget } = getInitializedWidget({ agentId: 'agentId' });
+
+        await widget.chatInstance.sendMessage({ text: 'hello' });
+
+        const { headers } = getRequestPayload();
+        expect(headers).toEqual(
+          expect.objectContaining({
+            'x-algolia-application-id': 'appId',
+            'x-algolia-api-Key': 'apiKey',
```
FabienMotte: I know this is not introduced by your PR and that HTTP is case-insensitive but should we use `x-algolia-api-key` (lowercase)? (same below in the implem)

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -413,16 +423,28 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
               api:
                 trigger === 'regenerate-message'
                   ? `${baseApi}&cache=false`
                   : baseApi,
+              // Only forward the actual request payload. The previous
+              // implementation spread every prepare param (`headers`, `body`,
+              // `api`, `credentials`, `requestMetadata`, …) into the body,
+              // duplicating values already sent in the headers and HTTP
+              // metadata. Returning no `headers` here also lets per-request
+              // headers — such as `algolia-referer` from `aiMode` /
```
FabienMotte:
```suggestion
              // headers — such as `algolia-referer` from `ai-mode` /
```

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
Haroenv: I believe headers are lowercased anyway? not 100% sure they are case sensitive

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -413,16 +423,28 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
           },
-          prepareSendMessagesRequest: ({ messages, trigger, ...rest }) => {
+          prepareSendMessagesRequest: ({ id, messages, trigger, messageId }) => {
             return {
               // Bypass cache when regenerating to ensure fresh responses
               api:
                 trigger === 'regenerate-message'
                   ? `${baseApi}&cache=false`
                   : baseApi,
+              // Only forward the actual request payload. The previous
```
Haroenv: is this comment actually needed?

packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
Haroenv: do follow up messages still have this referrer or only the first one? not sure what's right, just good to know and ensure it's tested.

Haroenv: referer (like http) or referrer (spelled right)?

Haroenv (commented):
looks good. Can you also check with Agent Studio team to ensure they will process these headers? cc @leonardocuco

packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
afrencalg: Good catch, renamed to x-algolia-referer so it matches the rest of our headers. Updated in all the call sites and the test.

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
afrencalg: Fixed in this PR — x-algolia-api-key everywhere now. Matches what we already use in connectFilterSuggestions and sendChatMessageFeedback. As @Haroenv  pointed out it's case-insensitive in practice, so this is purely about internal consistency.

packages/instantsearch.js/src/connectors/chat/connectChat.ts
afrencalg: Removed that whole comment block (per @Haroenv's  other thread), so the typo is gone too.

packages/instantsearch.js/src/connectors/chat/connectChat.ts
afrencalg: Trimmed it, the body shape and header behavior are now self-evident from the code.

packages/instantsearch.js/src/widgets/autocomplete/autocomplete.tsx
afrencalg: Kept it as referer to match the HTTP Referer convention (and what was in the original spec). Easy to flip to referrer if you'd rather have it spelled correctly, let me know which you prefer.

afrencalg: Only the first one, the header is set per-sendMessage call from the autocomplete/searchbox entry points; once you're typing in the chat input or hitting "regenerate", there's no x-algolia-referer on the request. I added a test (does not carry over the x-algolia-referer to follow-up messages) to lock that in.

This felt like the right semantic to me (the header captures what kicked off this turn, not session state), but happy to revisit if Agent Studio would rather have it on every request in the conversation.

FabienMotte (approved):

Haroenv (approved):

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -411,18 +421,19 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
           api: baseApi,
           headers: {
             'x-algolia-application-id': appId,
-            'x-algolia-api-Key': apiKey,
+            'x-algolia-api-key': apiKey,
             'x-algolia-agent': getAlgoliaAgent(instantSearchInstance.client),
```
Haroenv: actually this is missing the x-algolia-agent for component for "chat" right?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -357,7 +358,20 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+      const [appId, apiKey] = getAppIdAndApiKey(client);
+
+      // Identify the chat connector in the Algolia user-agent so the agent
+      // backend can attribute traffic to the chat component. `addAlgoliaAgent`
+      // deduplicates, so calling it for every chat widget instance is safe.
+      if (
+        typeof (client as { addAlgoliaAgent?: (agent: string) => void })
+          .addAlgoliaAgent === 'function'
+      ) {
+        (client as { addAlgoliaAgent: (agent: string) => void }).addAlgoliaAgent(
```
Haroenv: this is wrong, now the chat is added to every search request too

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -422,7 +436,7 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
           headers: {
             'x-algolia-application-id': appId,
             'x-algolia-api-key': apiKey,
-            'x-algolia-agent': getAlgoliaAgent(instantSearchInstance.client),
+            'x-algolia-agent': getAlgoliaAgent(client),
```
Haroenv: we instead should do something like 
```
`${getAlgoliaAgent(client)}; chat`
```

Haroenv (approved):
god once you bump bundlesise

PR #7007, opened 2026-05-04
fix(instantsearch.css): move select reset from shared reset to Nova theme

### LINKED ISSUE #7003: Select elements have broken background on satellite theme with 8.14.0+ css
### 🐛 Current behavior

Select elements have no background (dropdown indicator) when idle and on hover/focus it adds the background but it repeats so it fills the whole select instead of just being on the right side once.

I assume it's tied to:
```css
select[class^='ais-'] {
  appearance: none;
  background: none;
}
```
in [5289c69229fd63ead59b717bc9b8492a73bdc239](https://github.com/algolia/instantsearch/commit/5289c69229fd63ead59b717bc9b8492a73bdc239)

Image

### 🔍 Steps to reproduce

1. Use react-instantsearch-app as the starting point
2. Add component with select, e.g. `HitsPerPage` component
```react
<HitsPerPage items={[
    { value: 10, label: 'Show 10' },
    { value: 30, label: 'Show 30', default: true },
    { value: 50, label: 'Show 50' },
    { value: 150, label: 'Show 150' },
  ]} />
```
3. Look at the select added by ```HitsPerPage``` and try to hover/focus

### Live reproduction

Sorry but I just can't get the sandbox to run but it is simple to reproduce

### 💭 Expected behavior

Background looks normal with 8.13.0 of css so then it's just the dropdown indicator on the right side of select.

### Package version

instantsearch.css 8.14.0+

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

Haroenv:
## Summary

- Removes `select[class^='ais-'] { appearance: none; background: none; }` from `reset.scss` (shared across all themes)
- Adds it to `_nova-common.scss` (Nova-only), where it belongs

## Root cause

The selector `select[class^='ais-']` has specificity **0,1,1** (element + attribute), which is higher than the class-only selectors (0,1,0) used by the Satellite and Algolia themes to apply `background-image`, `background-position`, `background-repeat`, and `background-size` on their `<select>` widgets.

This meant the `background: none` shorthand in the reset silently won the cascade, breaking those themes:
- **Idle state**: the dropdown arrow image was suppressed entirely
- **Hover/focus state**: `background-image` was restored (those rules have specificity 0,2,0), but `background-position`/`background-repeat`/`background-size` were still controlled by the reset, causing the arrow SVG to tile across the whole element

The rule's purpose is purely to strip native browser select chrome so Nova can render its own custom chevron via an `::after` pseudo-element. It has no business being in the shared reset.

Fixes #7003

## Test plan

- [x] Satellite theme: `HitsPerPage` / `SortBy` selects show the dropdown arrow at idle, on hover, and on focus
- [x] Algolia theme: same widgets render correctly
- [x] Nova theme: custom chevron still renders (via `::after`) and native arrow is still suppressed

### DIFF
--- a/packages/instantsearch.css/src/shared/_nova-common.scss
+++ b/packages/instantsearch.css/src/shared/_nova-common.scss
@@ -71,12 +71,6 @@ select[class^='ais-']:focus-visible {
   font-size: 0.75em;
 }
 
-// Strip native select appearance for custom chevron rendering
-select[class^='ais-'] {
-  appearance: none;
-  background: none;
-}
-
 // Custom select (MenuSelect, HitsPerPage, SortBy)
 .ais-MenuSelect,
 .ais-HitsPerPage,
@@ -100,6 +94,8 @@ select[class^='ais-'] {
 .ais-MenuSelect-select,
 .ais-HitsPerPage-select,
 .ais-SortBy-select {
+  appearance: none;
+  background: none;
   width: 100%;
   color: rgba(var(--ais-text-color-rgb), var(--ais-text-color-alpha));
   font: inherit;
--- a/packages/instantsearch.css/src/shared/_nova-common.scss
+++ b/packages/instantsearch.css/src/shared/_nova-common.scss
@@ -71,6 +71,11 @@ select[class^='ais-']:focus-visible {
   font-size: 0.75em;
 }
 
+select[class^='ais-'] {
+  appearance: none;
+  background: none;
+}
+
 // Custom select (MenuSelect, HitsPerPage, SortBy)
 .ais-MenuSelect,
 .ais-HitsPerPage,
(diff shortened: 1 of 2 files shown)


packages/instantsearch.css/src/shared/_nova-common.scss
```diff
@@ -71,6 +71,11 @@ select[class^='ais-']:focus-visible {
   font-size: 0.75em;
 }
 
+select[class^='ais-'] {
+  appearance: none;
+  background: none;
+}
+
 // Custom select (MenuSelect, HitsPerPage, SortBy)
 .ais-MenuSelect,
```
Haroenv, line 80: thought about making these part of the select selector, but not possible as it would be complicated `has select` style selector

FabienMotte (approved):

PR #7019, opened 2026-05-09
fix(react-instantsearch-nextjs): trigger client search on App Router SPA navigation

anthonybachour:
**Summary**

`<InstantSearchNext>` ships its SSR results to the client via an inline `<script>` that sets `window[Symbol.for("InstantSearchInitialResults")]`. The next `<InstantSearchNext>` mount reads that value and skips its initial network request. This works on direct navigation, but two related issues break it on App Router SPA navigation:

1. The global is never cleared, so a value injected by an earlier render — including a Next.js `<Link>` prefetch of a route the user hasn't visited yet — survives across navigations and gets consumed by a *different* `<InstantSearchNext>` mount on a subsequent route. The destination renders into another mount's serialized state (different `hitsPerPage`, possibly no facet widgets), and the auto-search is suppressed via `shouldRenderAtOnce`.
2. `waitForResultsRef` is provided as a stable ref object regardless of whether SSR results actually exist. `useInstantSearchApi` treats it as truthy in `shouldRenderAtOnce = serverContext || initialResults || waitForResultsRef`, so on a SPA navigation with no real `initialResults` it still calls `search._initialResults = initialResults || {}` (an empty object) and skips the initial network request. Widgets render with empty state and the SSR'd output is discarded by hydration.

In practice, after a `<Link>` click into a page that mounts `<InstantSearchNext>`, the destination's facets and hits are empty even though the SSR'd HTML on that route had the correct values; a hard reload restores them.

Fix:
- Read-and-clear `window[InstantSearchInitialResults]` so a value can only be consumed once.
- On the client, only set `waitForResultsRef` to `promiseRef` when there are real `initialResults`. Otherwise default to `null` so `useInstantSearchApi` falls through to its normal client-mount path and fires an initial search.

Server-side and direct-navigation paths are unchanged.

**Result**

Tests in `packages/react-instantsearch-nextjs/src/__tests__/InstantSearchNext.test.tsx`:

- ✅ Existing: `does not trigger a client-side search by default` — no extra search when results are pre-populated.
- ✅ New: `fires a client-side search when no initialResults are injected` — single mount with no global set fires its own search (the SPA-navigation case where the prior route never set the global).
- ✅ New: `does not reuse a previous mount's initialResults` — first mount consumes (and clears) the injected SSR results; a second mount no longer recycles stale state and fires its own search.

```
$ jest packages/react-instantsearch-nextjs/src/__tests__/InstantSearchNext.test.tsx
PASS packages/react-instantsearch-nextjs/src/__tests__/InstantSearchNext.test.tsx
  rerendering
    ✓ does not trigger a client-side search by default (26 ms)
  SPA navigation hydration
    ✓ fires a client-side search when no initialResults are injected (5 ms)
    ✓ does not reuse a previous mount's initialResults (8 ms)
```

Verified end-to-end against an unrelated production project: with the patch applied to `node_modules`, a click from a route without `<InstantSearchNext>` to one with it (and the in-app `useInstantSearch().refresh()` workaround removed) renders the same 84 facet items / 60 hits the SSR'd HTML produced. Without the patch, the destination renders 0 facet items.


dhayab:
Thanks for your contribution @anthonybachour! I've merged latest `master` into your branch and pushed directly here.

I also reverted the since-merged `RefreshOnClientNavigation` workaround, with your "read-and-clear" fix in place it became redundant: the two together fired 2 search requests on every soft navigation. Your approach also fixes it at the source: `RefreshOnClientNavigation`'s blanket `refresh()` updated the result count but left the hit list stale whereas read-and-clear clears the injected state so the next mount fetches correctly in a single request.

Edit for reviewers: CI checks won't run on a fork PR, all checks pass locally.

Haroenv (approved):

PR #7028, opened 2026-05-13
fix(examples): update React SSR example for ESM

### LINKED ISSUE #6950: (docs) React / Express SSR CodeSandbox example has SSR module-format issues in a fresh environment
### 🐛 Current behavior

The CodeSandbox example [SSR with Express](https://codesandbox.io/s/github/algolia/instantsearch/tree/master/examples/react/ssr?file=/src/App.js) linked in our docs [Server-side rendering](https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react) appears to be broken in a fresh environment due to SSR module-format inconsistencies.

* The SSR build produces a CommonJS-style server bundle
* That bundle attempts to `require()` ESM modules from `instantsearch.js/es/...`
* This results in:

  * `Error [ERR_REQUIRE_ESM]: require() of ES Module ... not supported`

When adjusting the setup to resolve this (e.g. switching the SSR output to ESM), the server then fails at runtime because the example still relies on CommonJS globals such as `__dirname`.

As a result, the example requires manual changes to run successfully.

---
### Note

I was able to get the example working with the following changes:

* remove `vite-plugin-commonjs`
* adjust the SSR setup to avoid emitting CommonJS requiring ESM modules
* update `src/server.jsx` to replace `__dirname` with the `import.meta.url` + `fileURLToPath` pattern

This suggests the current example mixes CommonJS and ESM assumptions in a way that no longer works in a fresh environment.

### 🔍 Steps to reproduce

1. Open the CodeSandbox example:
   https://codesandbox.io/s/github/algolia/instantsearch/tree/master/examples/react/ssr?file=/src/App.js

2. Let the sandbox run its default task (`yarn start`)

3. Observe the SSR build and server startup

4. The server fails during startup with:

   * `Error [ERR_REQUIRE_ESM]: require() of ES Module ... not supported`

5. If the SSR setup is then adjusted to avoid the initial module-format error, a follow-up runtime issue appears because the server entry still relies on CommonJS globals:

   * `ReferenceError: __dirname is not defined in ES module scope`

### Live reproduction

https://codesandbox.io/p/devbox/modest-star-zqwmfx

### 💭 Expected behavior

The example should run successfully in a fresh CodeSandbox environment without requiring manual changes.

Specifically:

* the SSR build should complete without module-format/runtime errors
* the server should start correctly
* the published example configuration and server entry should be internally consistent
* users should be able to open the sandbox and see the SSR example running immediately

### Package version

@vitejs/plugin-react 4.2.1, vite 5.0.7, algoliasearch 5.1.1, algoliasearch-helper 3.28.1, express 4.17.1, instantsearch-ui-components 0.23.0, react 19.0.0, react-dom 19.0.0, react-instantsearch 7.29.0, react-instantsearch-core 7.29.0

### Operating system

_No response_

### Browser

_No response_

### Code of Conduct

reneemundie:
Fix #6950

**Summary**

The React / Express SSR example currently fails in a fresh CodeSandbox environment because the SSR build emits CommonJS while depending on ESM packages. After switching the SSR output to ESM, the server also needs to avoid CommonJS globals like `__dirname`.

This PR updates the example so the SSR build and server entry use ESM consistently.

Changes:
- Removes `vite-plugin-commonjs`
- Adds `"type": "module"` to the example package
- Builds the SSR server output as ESM
- Replaces `__dirname` usage with `import.meta.url` / `fileURLToPath`

**Result**

Verified the SSR example in CodeSandbox using the current dependency versions from `master`.

Confirmed:
- The build completes successfully
- The server starts without `ERR_REQUIRE_ESM`
- The server no longer fails with `__dirname is not defined`

I also attempted to run the example locally with:

```sh
yarn --cwd examples/react/ssr start
```

The local run did not complete because my local monorepo build produced a generated workspace artifact importing instantsearch.jsundefined from react-instantsearch-core, which appears unrelated to this SSR example change. The CodeSandbox verification above was done against the updated example files and current dependency versions.

### DIFF
--- a/examples/react/ssr/package.json
+++ b/examples/react/ssr/package.json
@@ -3,14 +3,14 @@
   "version": "6.122.0",
   "private": true,
   "license": "MIT",
+  "type": "module",
   "scripts": {
     "build": "vite build && vite build --ssr src/server.jsx",
     "start": "yarn build && node dist/server.js"
   },
   "devDependencies": {
     "@vitejs/plugin-react": "4.2.1",
-    "vite": "5.0.7",
-    "vite-plugin-commonjs": "0.10.0"
+    "vite": "5.0.7"
   },
   "dependencies": {
     "algoliasearch": "5.1.1",
--- a/examples/react/ssr/src/server.jsx
+++ b/examples/react/ssr/src/server.jsx
@@ -1,4 +1,5 @@
-import { join } from 'path';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
 
 import express from 'express';
 import React from 'react';
@@ -7,6 +8,9 @@ import { getServerState } from 'react-instantsearch';
 
 import App from './App';
 import { responsesCache } from './searchClient';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
 
 const app = express();
 
--- a/examples/react/ssr/vite.config.mjs
+++ b/examples/react/ssr/vite.config.mjs
@@ -1,31 +1,26 @@
 import react from '@vitejs/plugin-react';
 import { defineConfig } from 'vite';
-import commonjs from 'vite-plugin-commonjs';
 
-export default defineConfig(({ isSsrBuild }) => {
-  const isSSR = Boolean(isSsrBuild);
-
-  return {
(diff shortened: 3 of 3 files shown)


Haroenv (approved):
thanks Renee!

PR #7037, opened 2026-05-13
chore(examples): support iframe auto-resize in showcase

FabienMotte:
## Summary

When the showcase is embedded in the Algolia docs (currently hardcoded to `https://www.algolia.com`):

- `main.tsx` posts the body's content height to the parent via `postMessage`, observed with a `ResizeObserver` on `document.body`. The docs side applies it to the iframe so readers don't have to scroll inside the iframe.
- `documentElement.style.overflow = "hidden"` suppresses the inner scrollbar (which would otherwise show a gutter on always-show-scrollbar OSes).
- `document.body.classList.add("is-embedded")` lets stylesheets and components branch on the embedded state.
- The chat's overlay/sidePanel layouts use `position: fixed`, which anchors to the iframe (not the user's viewport) when embedded. Three fixes for that:
  - CSS caps `.ais-ChatOverlayLayout` / `.ais-ChatSidePanelLayout` to `height: 600px` when embedded so they don't span the full iframe.
  - The floating `ais-Chat-toggleButtonWrapper` is hidden when embedded.
  - `WidgetChat` renders an inline "Open chat" trigger inside the chat tile (using the AIS primary button theme variables for visual consistency). Clicking it opens the chat and posts `{ type: "showcase-scroll-into-view" }` so the docs page scrolls to bring the chat panel into view.

Standalone behavior (showcase opened directly at its public URL) is unchanged — every branch is gated on `window.parent !== window`.

Requires the matching parent-side change in algolia/docs-new#824. Each PR is a no-op without the other, so either can ship first.

## Test plan

Tested locally with the showcase embedded in a local docs build:

- [x] Standalone showcase URL: behavior is identical to before (overlay/sidePanel anchor to the browser viewport, no inline trigger).
- [x] Embedded showcase: iframe auto-fits its content; switching to a shorter tab (GeoSearch) shrinks the iframe back down.
- [x] Embedded showcase, Agentic tab: inline blue "Open chat" trigger renders inside the chat tile for `overlay` and `sidePanel` layouts; the real floating trigger is hidden.
- [x] Clicking "Open chat" smooth-scrolls the parent docs page so the chat panel is visible; for `sidePanel`, the scroll tracks the iframe while its marginRight animation grows the body width.
- [x] `inline` layout: no inline trigger button shown (the chat renders normally in the tile).

### DIFF
--- a/examples/js/showcase/src/utils/parentMessenger.ts
+++ b/examples/js/showcase/src/utils/parentMessenger.ts
@@ -1,0 +1,14 @@
+// Posts a message to the embedding docs page. No-op when this showcase isn't
+// loaded inside an iframe. The browser drops posts to non-matching origins,
+// so listing both prod and the local Mintlify dev origin is safe.
+const TARGET_ORIGINS = [
+  "https://www.algolia.com",
+  "http://localhost:3000",
+];
+
+export function postToParent(data: unknown): void {
+  if (typeof window === "undefined" || window.parent === window) return;
+  for (const origin of TARGET_ORIGINS) {
+    window.parent.postMessage(data, origin);
+  }
+}
--- a/examples/js/showcase/src/utils/parentMessenger.ts
+++ b/examples/js/showcase/src/utils/parentMessenger.ts
@@ -7,7 +7,6 @@ const TARGET_ORIGINS = [
 ];
 
 export function postToParent(data: unknown): void {
-  if (typeof window === "undefined" || window.parent === window) return;
   for (const origin of TARGET_ORIGINS) {
     window.parent.postMessage(data, origin);
   }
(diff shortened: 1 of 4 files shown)


Haroenv (approved):
is good, we can refactor once trigger is an explicit separate component :)

examples/js/showcase/src/utils/parentMessenger.ts
```diff
@@ -0,0 +1,14 @@
...
+// loaded inside an iframe. The browser drops posts to non-matching origins,
+// so listing both prod and the local Mintlify dev origin is safe.
+const TARGET_ORIGINS = [
+  "https://www.algolia.com",
+  "http://localhost:3000",
+];
+
+export function postToParent(data: unknown): void {
+  if (typeof window === "undefined" || window.parent === window) return;
+  for (const origin of TARGET_ORIGINS) {
```
Haroenv, line 10: is this needed? doesn't really hurt to always send right?

FabienMotte: Good point, let me remove it

PR #7042, opened 2026-05-20
feat(chat): send hidden context via message metadata.turnContext

afrencalg:
The `context` prop introduced in #6973 was smuggled to the model as a `<context>{...}</context>` text part prepended to every user message. That kept the values out of the rendered transcript only because `ChatMessage` had a hardcoded prefix/suffix check, polluted `sessionStorage` with the JSON blob, was unvalidated, and was unrecognised by the server beyond happening to land in the prompt.

Agent Studio now ships a first-class wire contract for the same use case: the latest user message carries `metadata.turnContext: Record<string, string>`, which the server validates, persists on the user-message row, hashes into
the completion cache key, and prepends to the LLM-bound user content behind a hardened preamble.

`connectChat` now writes the resolved context to `message.metadata.turnContext` instead of injecting a text part. The prop type narrows to `Record<string, string> | (() => Record<string, string> | Promise<Record<string, string>>)` to match the server contract, and the function form may now be async so callers can resolve page/session facts lazily. A new `sanitizeTurnContext` helper mirrors the server-side caps (≤32 keys, charset `^[A-Za-z0-9_.-]{1,64}$`, value ≤1024 B, total ≤4096 B) but drops offending entries with a dev `warning(...)` instead of throwing — a misconfigured `context` prop must never break the chat. Errors from a sync- or async-throwing resolver fall through the same path. Caller- supplied `metadata` is preserved and `turnContext` is namespaced under it.

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
@@ -211,12 +211,6 @@ export function createChatMessageComponent({ createElement }: Renderer) {
         return null;
       }
       if (part.type === 'text') {
-        if (
-          part.text.startsWith('<context>') &&
-          part.text.endsWith('</context>')
-        ) {
-          return null;
-        }
         const markdown = compiler(part.text, {
           createElement: createElement as any,
           disableParsingRawHTML: true,
--- a/packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
@@ -211,6 +211,16 @@ export function createChatMessageComponent({ createElement }: Renderer) {
         return null;
       }
       if (part.type === 'text') {
+        // Back-compat shim for sessions started before the move from a
+        // `<context>{...}</context>` text part to `metadata.turnContext`.
+        // Safe to remove once existing sessionStorage transcripts have
+        // rolled over (~2 weeks after release).
+        if (
+          part.text.startsWith('<context>') &&
+          part.text.endsWith('</context>')
+        ) {
+          return null;
+        }
         const markdown = compiler(part.text, {
           createElement: createElement as any,
           disableParsingRawHTML: true,
--- a/packages/instantsearch.js/src/connectors/chat/connectChat.ts
+++ b/packages/instantsearch.js/src/connectors/chat/connectChat.ts
@@ -166,12 +166,11 @@ export type ChatConnectorParams<TUiMessage extends UIMessage = UIMessage> = (
    * rendered as a chat bubble and never persisted on assistant turns.
    *
    * The server validates the payload (flat `Record<string, string>`, key/value
-   * length and shape) and rejects malformed contexts. Pass a function (sync or
-   * async) when the values change per-turn — it is invoked once per send.
+   * length and shape) and rejects malformed contexts. Pass a function when the
+   * values change per-turn — it is invoked once per send. If the source is
+   * async, resolve it upstream and close over the value.
    */
-  context?:
-    | Record<string, string>
-    | (() => Record<string, string> | Promise<Record<string, string>>);
+  context?: Record<string, string> | (() => Record<string, string>);
   /**
    * A message to send automatically when the chat is initialized.
    *
@@ -693,23 +692,20 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
           }
 
           // Resolve once per send; let the server validate the payload and
-          // surface any contract violations. Wrapping in `Promise.resolve()`
-          // normalises sync throws from `context()` into rejections so callers
-          // can `await sendMessage(...).catch(...)` uniformly.
-          return Promise.resolve()
-            .then(() => (typeof context === 'function' ? context() : context))
-            .then((turnContext) =>
-              _chatInstance.sendMessage(
-                {
-                  ...message,
-                  metadata: {
-                    ...(message.metadata as Record<string, unknown> | undefined),
-                    turnContext,
-                  },
-                } as Parameters<typeof _chatInstance.sendMessage>[0],
-                ...rest
-              )
-            );
+          // surface any contract violations.
+          const turnContext =
+            typeof context === 'function' ? context() : context;
+
+          return _chatInstance.sendMessage(
+            {
+              ...message,
+              metadata: {
+                ...(message.metadata as Record<string, unknown> | undefined),
+                turnContext,
+              },
+            } as Parameters<typeof _chatInstance.sendMessage>[0],
+            ...rest
+          );
         };
 
         return {
--- a/packages/instantsearch.js/src/connectors/chat/connectChat.ts
+++ b/packages/instantsearch.js/src/connectors/chat/connectChat.ts
@@ -48,6 +48,107 @@ const withUsage = createDocumentationMessageGenerator({
   connector: true,
 });
 
+// Mirrors `common/models/turn_context.py` on the server. Kept in lock-step so
+// the client never sends payloads the backend would 422 on. We diverge from the
+// server in one place only: invalid entries are dropped with a dev warning
+// instead of throwing, matching the server's kill-switch semantics
+// (`TURN_CONTEXT_ENABLED=false` silently drops). A misconfigured `context`
+// prop must never break the chat for end users.
+const TURN_CONTEXT_MAX_BYTES = 4 * 1024;
+const TURN_CONTEXT_MAX_KEYS = 32;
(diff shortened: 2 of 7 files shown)


packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -211,12 +211,6 @@ export function createChatMessageComponent({ createElement }: Renderer) {
         return null;
       }
       if (part.type === 'text') {
-        if (
-          part.text.startsWith('<context>') &&
```
Haroenv, line 215: this likely doesn't really matter, but if someone had an existing conversation with previous context, now reloads their page with the new implementation, they'll see the old `<context` tags. Let's maybe leave this filtering for a couple weeks? what do you think?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -686,49 +796,63 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+          // Resolve and sanitize the context at send time. The function form
+          // may be async (e.g. resolves the current page URL from a Next.js
+          // router) and may also throw synchronously, so we wrap the call in
+          // `Promise.resolve().then(...)` to funnel both error modes through
+          // the same rejection handler. We never throw on invalid entries —
+          // offending keys are dropped with a warning so a misconfigured
+          // `context` prop can't break the chat.
+          const resolvePromise = Promise.resolve()
+            .then(() => (typeof context === 'function' ? context() : context))
+            .then(sanitizeTurnContext, (error) => {
```
Haroenv: use catch, not second argument to then

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -686,49 +796,63 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
             return _chatInstance.sendMessage(message, ...rest);
           }
 
-          const resolvedContext =
-            typeof context === 'function' ? context() : context;
+          // Resolve and sanitize the context at send time. The function form
```
Haroenv: maybe less "explain"ey comment

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -48,6 +48,107 @@ const withUsage = createDocumentationMessageGenerator({
   connector: true,
 });
 
+// Mirrors `common/models/turn_context.py` on the server. Kept in lock-step so
+// the client never sends payloads the backend would 422 on. We diverge from the
```
Haroenv: I actually think that instead of 422, the server should respond with a 200 but skip the context. Now we're working against the behaviour of the server and are forced to keep this updated, and means that non-updated customers still have old limits

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -48,6 +48,107 @@ const withUsage = createDocumentationMessageGenerator({
...
+    }
+    mutated = true;
+    const { turnContext: _ignored, ...rest } = metadata;
+    return {
+      ...message,
+      metadata: Object.keys(rest).length > 0 ? rest : undefined,
+    } as TUiMessage;
+  });
+  return mutated ? stripped : messages;
+}
```
Haroenv: seems like a more complicated function than possible, but not sure if it can be simplified

Haroenv: maybe just `delete turnContext` as we're mutating anyway

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -686,49 +796,63 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+          });
         };
 
+        // `turnContext` is a server-grounding signal — the contract is "invisible
+        // to the UI". Strip it from every message exposed to renderers so it
+        // never accidentally leaks into a chat bubble, even if a future
+        // history-GET path or restored sessionStorage cache carries it.
+        const visibleMessages = stripTurnContextFromMessages(
+          _chatInstance.messages
+        );
```
Haroenv: does it really need to be stripped? the renderer doesn't just JSON stringify the server output but only renders the text and the tools, not the metadata

Haroenv (commented):
looks good, some questions

afrencalg:
@Haroenv addressed comments.

packages/instantsearch.js/src/connectors/chat/connectChat.ts
Haroenv: you resolved, but is there a server update @afrencalg ?

raed667: The server 422 is intentional, we want the context values to be intentional. Silently dropping the context was ruled out as an option in favor of a more explicit error flow.

Haroenv: but now we're silently dropping it frontend, isn't that the same situation @raed667 ?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -48,6 +48,86 @@ const withUsage = createDocumentationMessageGenerator({
...
+    sanitized[key] = value;
+    totalBytes += keyBytes + valueBytes;
+  });
+
+  if (dropped > 0) {
+    warning(
+      false,
+      `Chat context: dropped ${dropped} entr${
+        dropped === 1 ? 'y' : 'ies'
+      } that did not meet the Agent Studio turn_context contract (first reason: ${firstReason}). See https://www.algolia.com/doc/guides/algolia-ai/agent-studio/.`
```
shaejaz: don't think the doc link here is that useful. do we have a dedicated page in mind in the future?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
Haroenv: after discussing, I think it's best not to do any validation and rely on the backend rejecting if context is too large. Better to break than to send no context

afrencalg:
@Haroenv @shaejaz  addressed all comments

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -160,12 +160,18 @@ export type ChatConnectorParams<TUiMessage extends UIMessage = UIMessage> = (
...
+   * rendered as a chat bubble and never persisted on assistant turns.
+   *
+   * The server validates the payload (flat `Record<string, string>`, key/value
+   * length and shape) and rejects malformed contexts. Pass a function (sync or
+   * async) when the values change per-turn — it is invoked once per send.
    */
-  context?: Record<string, unknown> | (() => Record<string, unknown>);
+  context?:
+    | Record<string, string>
+    | (() => Record<string, string> | Promise<Record<string, string>>);
```
Haroenv: if it's a promise, will this wait to be resolved before sending every message? What about loading states?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -686,47 +692,24 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
-          } catch {
-            warning(
-              false,
-              'Could not serialize chat context. The message will be sent without context.'
+          // Resolve once per send; let the server validate the payload and
+          // surface any contract violations. Wrapping in `Promise.resolve()`
+          // normalises sync throws from `context()` into rejections so callers
+          // can `await sendMessage(...).catch(...)` uniformly.
+          return Promise.resolve()
+            .then(() => (typeof context === 'function' ? context() : context))
```
Haroenv: what's the use case of async context? if it's valid we can keep it of course, just seeing this potentially complexity the control flow by this part being asynchronous now

Haroenv (commented):
looks great! just wondering why async context is introduced. What's the use case?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
afrencalg: Good catch — no concrete consumer right now. The motivation was the React Native WebView bridge, where context lives on the native side, but consumers there can resolve once upstream and pass a sync closure, so it's not a hard requirement. Dropping Promise<...> from the type.

afrencalg:
@Haroenv I dropped the Promise<...> branch and the microtask hop; resolver throws now propagate synchronously. Easy to add back if a genuine async-per-turn use case shows up.

Haroenv (approved):

PR #7043, opened 2026-05-20
fix(recommend): remove unsupported queryParameters from trending-facets

FabienMotte:
**Summary**

The Algolia Recommend API rejects `queryParameters` for the `trending-facets` model (returns HTTP 400), so the option could never be set in practice. This PR removes the option from the `trendingFacets` widget and connector across the InstantSearch packages.

image

Related:
- Docs: algolia/docs-new#843
- [#6716](https://github.com/algolia/instantsearch/pull/6716) did the same kind of cleanup for `fallbackParameters` on FBT.

Changes:
- `connectTrendingFacets`: drop `queryParameters` from `TrendingFacetsConnectorParams`, the destructure, and the `addTrendingFacets({...})` call (only `fallbackParameters` still gets the `TAG_PLACEHOLDER` merge).
- `trendingFacets` (JS widget): drop the `queryParameters` pass-through.
- `TrendingFacets` (React widget): drop the `queryParameters` pass-through.
- Tests: update `getWidgetParameters` cases to no longer pass/expect `queryParameters`.

`useTrendingFacets` (react-instantsearch-core) re-exports the connector type directly, so the hook picks up the change automatically.

**Result**

```
PASS packages/instantsearch.js/src/widgets/trending-facets/__tests__/trending-facets.test.tsx
PASS packages/instantsearch.js/src/connectors/trending-facets/__tests__/connectTrendingFacets-test.ts
PASS packages/react-instantsearch/src/widgets/__tests__/TrendingFacets.test.tsx

Test Suites: 3 passed (+ legacy umd mirrors)
Tests:       31 passed
```

> [!NOTE]
> Technically a public-API removal on `trendingFacets`/`TrendingFacets`/`useTrendingFacets`. Since the Recommend endpoint rejected the option, no working code can depend on it.

### DIFF
--- a/packages/instantsearch.js/src/connectors/trending-facets/__tests__/connectTrendingFacets-test.ts
+++ b/packages/instantsearch.js/src/connectors/trending-facets/__tests__/connectTrendingFacets-test.ts
@@ -144,7 +144,7 @@ describe('connectTrendingFacets', () => {
       });
 
       expect(actual).toEqual(
-        // v4 TrendingFacetsQuery lacks fallbackParameters
+        // v4 TrendingFacetsQuery lacks queryParameters/fallbackParameters
         new RecommendParameters().addTrendingFacets({
           $$id: (widget as any).$$id,
           facetName: 'brand',
@@ -171,7 +171,7 @@ describe('connectTrendingFacets', () => {
       });
 
       expect(actual).toEqual(
-        // v4 TrendingFacetsQuery lacks fallbackParameters
+        // v4 TrendingFacetsQuery lacks queryParameters/fallbackParameters
         new RecommendParameters().addTrendingFacets({
           $$id: (widget as any).$$id,
           facetName: 'brand',
--- a/packages/instantsearch.js/src/connectors/trending-facets/__tests__/connectTrendingFacets-test.ts
+++ b/packages/instantsearch.js/src/connectors/trending-facets/__tests__/connectTrendingFacets-test.ts
@@ -135,7 +135,6 @@ describe('connectTrendingFacets', () => {
         facetName: 'brand',
         limit: 10,
         threshold: 95,
-        queryParameters: { userToken: 'token' },
         escapeHTML: false,
       });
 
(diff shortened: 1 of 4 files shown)


Haroenv:
@raed667 do you know _why_ this isn't supported for trending facets?

raed667:
@Haroenv `trending-facets` fundamentally have no 'query' for it to have query-parameters. It returns facets instead of hits/records. 

I think, i had a PR to fix it in API-client docs a while back and should have caught it here :my-bad: 


BTW: what do you think about officially fully deprecating algolia/recommend project now that we have feature parity in IS ?

Haroenv:
yes, you can deprecate it @raed667, ensure the docs still live for a while in a deprecated folder and you release a last version with a "use InstantSearch instead" note maybe?

packages/instantsearch.js/src/connectors/trending-facets/connectTrendingFacets.ts
```diff
@@ -181,8 +171,8 @@ export default (function connectTrendingFacets<
       },
 
       getWidgetParameters(state) {
-        // v4 TrendingFacetsQuery doesn't include queryParameters or
-        // fallbackParameters, but the v5 API and the helper support them.
+        // v4 TrendingFacetsQuery doesn't include fallbackParameters,
+        // but the v5 API and the helper support it.
```
raed667: Same nit on the comment

raed667 (approved):
LGTM with a couple of nitpicks on the comments

packages/instantsearch.js/src/connectors/trending-facets/__tests__/connectTrendingFacets-test.ts
```diff
@@ -174,17 +171,12 @@ describe('connectTrendingFacets', () => {
       });
 
       expect(actual).toEqual(
-        // v4 TrendingFacetsQuery lacks queryParameters/fallbackParameters
```
raed667, line 177: Nit: i think the comment was right, but the code change is good 👍

Haroenv (approved):

PR #7044, opened 2026-05-20
refactor(instantsearch.css): share carousel styles between chat and recommend widgets

FabienMotte:
**Summary**

Extract the carousel SCSS that was previously living inside chat into a shared `_carousel.scss` component, so recommend widgets in nova can use the same styles without duplicating the rules.

Stacked on top of #7043.

**Result**

- New `src/components/_carousel.scss` partial used by both chat and nova themes.
- `_chat-carousel.scss` renamed to `_chat-tool-results.scss` (chat-specific rules only).
- Net: 132 insertions / 126 deletions — mostly a move.

### DIFF
--- a/packages/instantsearch.css/src/shared/_variables.scss
+++ b/packages/instantsearch.css/src/shared/_variables.scss
@@ -76,8 +76,8 @@ $ais-chat-breakpoint: 680px;
   --ais-chat-maximized-height: 100%;
   --ais-chat-margin: 1.5rem;
 
-  /* Chat carousel component */
-  --ais-chat-carousel-item-width: calc(var(--ais-spacing) * 10);
+  /* Carousel component */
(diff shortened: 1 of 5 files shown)


packages/instantsearch.css/src/shared/_variables.scss
```diff
@@ -76,8 +76,8 @@ $ais-chat-breakpoint: 680px;
   --ais-chat-maximized-height: 100%;
   --ais-chat-margin: 1.5rem;
 
-  /* Chat carousel component */
-  --ais-chat-carousel-item-width: calc(var(--ais-spacing) * 10);
+  /* Carousel component */
+  --ais-carousel-item-width: calc(var(--ais-spacing) * 10);
```
FabienMotte, line 80: This is technically a breaking change, but since chat is still in beta, I think we are fine?

Haroenv (approved):

PR #7054, opened 2026-05-22
feat(agent studio): add prompt suggestions widget

shaejaz:
[FX-3845]

Adds a **PromptSuggestions** widget that renders AI-generated prompt "pills" for the page, powered by the Agent Studio `/tasks` endpoint. Clicking a pill hands the prompt off to the `Chat` widget on the same index.

## Highlights

- **New connector** `connectPromptSuggestions` (`instantsearch.js`, `$$type: 'ais.promptSuggestions'`) — owns all search-aware logic: fetch on mount, 300ms debounced refetch on query/refinement change, manual `refresh()`, `transformItems`/`transformHits`, and chat hand-off.
- **New generic `connectTasks` connector** — the reusable core underneath. `connectPromptSuggestions` is now a thin search-aware layer on top of it, so any future task-backed feature can reuse the same request/streaming/output plumbing. Connector-only (no default UI), generic over `TOutput`, `$$type: 'ais.tasks'`; params are `agentId` XOR `transport`, `task`, and `stream`.
- **New `lib/tasks/` runtime** — the generic Agent Studio *tasks*-endpoint counterpart to `lib/chat/` (completions). Endpoint/credential/payload primitives (`resolveEndpoint`, `buildTaskPayload`, `fetchTask` with optional streaming, `createTaskRunner`, and `Task*` types); nothing widget-specific.
- **Context, decided at init by the presence of `context`:**
  - **Default** (no `context`): auto-extracts `{ query, filters, hitsSample }` from the current search state and refreshes as results change. `transformHits` shapes the sample (default: first 5 hits, internal `_`-prefixed metadata stripped).
  - **Explicit `context`** (object or per-fetch function): sent as-is; hit auto-extraction is skipped.
- **Shared UI** in `instantsearch-ui-components` (`createPromptSuggestionsComponent`): `.ais-PromptSuggestions` pills + skeleton loading state, disabled while the chat is busy. Overridable via `layoutComponent` (React) / `templates.layout` (JS) and `onSuggestionClick`.
- **Auth**: pass `agentId` (uses the InstantSearch client's app id / key) **or** a custom `transport` (`api` / `headers`). `configurationId` selects the Agent Studio task (sent as the `task` field, default `'prompt_suggestions'`).
- **Tests**: full connector unit suites for `connectPromptSuggestions` **and** `connectTasks`, `lib/tasks` unit tests, and cross-flavor common tests (`common-widgets -t "PromptSuggestions widget common tests"`); Vue registered as skipped.

## React API

**Default (page-driven):**

```tsx
import { PromptSuggestions } from 'react-instantsearch';

<PromptSuggestions agentId="YOUR_AGENT_ID" />
```

**Explicit context (e.g. a product page):**

```tsx
<PromptSuggestions
  agentId="YOUR_AGENT_ID"
  context={{
    productName: product.name,
    category: product.category,
    price: product.price,
  }}
/>
```

[FX-3845]: https://algolia.atlassian.net/browse/FX-3845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/instantsearch.js/src/connectors/chat-page-suggestions/connectChatPageSuggestions.ts
```diff
@@ -0,0 +1,681 @@
...
+
+        // Clear stale pills so the loading skeleton shows on every refetch
+        // (query/refinement change), not only the initial fetch. The UI
+        // component renders the skeleton when `isLoading && suggestions` is
+        // empty; without this it would keep the previous pills on screen and
+        // swap them silently, so a refetch would never surface a loading state.
+        suggestions = [];
+        isLoading = true;
+        inFlight = true;
+        renderFn(
```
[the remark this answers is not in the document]
anjola-adeuyi: I think this one is worth addressing. An older suggestions request can resolve after a newer query/refinement and overwrite the current pills. Can we ignore non-latest responses or abort the previous fetch?

packages/instantsearch.js/src/connectors/chat-page-suggestions/connectChatPageSuggestions.ts
```diff
@@ -0,0 +1,681 @@
...
+const CHAT_RENDER_STATE_KEY = 'chat' as const;
+const DEBOUNCE_MS = 300;
+const DEFAULT_SSR_TIMEOUT_MS = 150;
+
+/**
+ * The default Algolia-owned task ID for on-page prompt suggestions. Server
+ * config (agent `config.tasks[<id>]`) owns the instructions, input schema, and
+ * output schema; the client only selects the task and supplies `input`.
+ */
+const DEFAULT_TASK_ID = 'algolia_on_page_suggestions';
```
Haroenv: too hardcoded

packages/instantsearch.js/src/connectors/chat-page-suggestions/connectChatPageSuggestions.ts
```diff
@@ -0,0 +1,681 @@
...
+  InitOptions,
+  InstantSearch,
+  RenderOptions,
+  WidgetRenderState,
+} from '../../types';
+import type { ChatRenderState } from '../chat/connectChat';
+import type { SearchResults } from 'algoliasearch-helper';
+
+const withUsage = createDocumentationMessageGenerator({
+  name: 'chat-page-suggestions',
```
Haroenv: should the connector not be for all "tasks"? instead of only for on page suggestions

packages/instantsearch-ui-components/src/components/chat/ChatPromptSuggestions.tsx
```diff
@@ -3,14 +3,32 @@
...
   root?: string | string[];
+  header?: string | string[];
+  headerTitle?: string | string[];
   suggestion?: string | string[];
+  skeleton?: string | string[];
+  skeletonItem?: string | string[];
 };
 
-export type ChatPromptSuggestionsOwnProps = {
+export type ChatPromptSuggestionsTranslations = {
```
Haroenv: this is called prompt suggestions, and the other is called page suggestions

packages/instantsearch.js/src/connectors/chat-page-suggestions/connectChatPageSuggestions.ts
```diff
@@ -0,0 +1,681 @@
...
+          return true;
+        };
+
+      const doFetch = (results: SearchResults): Promise<string[]> => {
+        const resolvedContext =
+          typeof context === 'function' ? context() : context;
+        const resolvedPageType =
+          pageType ?? (resolvedContext ? 'pdp' : 'plp');
+        // PDP mode: caller-supplied context replaces auto-extraction.
+        // PLP mode (default): build the input from the current search state.
```
Haroenv: we shouldn't use the words pdp and plp IMO. Maybe "searchResults" and "input"? or something

packages/instantsearch.js/src/connectors/chat-page-suggestions/connectChatPageSuggestions.ts
```diff
@@ -0,0 +1,681 @@
...
+ * output schema; the client only selects the task and supplies `input`.
+ */
+const DEFAULT_TASK_ID = 'algolia_on_page_suggestions';
+
+/**
+ * Page-type discriminator forwarded to the task as `input.pageType`. `'pdp'` /
+ * `'plp'` are the documented cases; any string is accepted since the task's
+ * input schema is server-owned.
+ */
+export type ChatPageSuggestionsPageType = 'pdp' | 'plp' | (string & {});
```
FabienMotte: I think `pageType` could be removed entirely since the context mode is defined in the Dashboard as part of the agent's configuration. The context mode modifies the instructions.

packages/instantsearch.js/src/connectors/chat-page-suggestions/connectChatPageSuggestions.ts
```diff
@@ -0,0 +1,681 @@
...
+    >;
+  };
+};
+
+export type ChatPageSuggestionsConnector = Connector<
+  ChatPageSuggestionsWidgetDescription,
+  ChatPageSuggestionsConnectorParams
+>;
+
+const DEFAULT_TRANSFORM_HITS: ChatPageSuggestionsTransformHits = (hits) =>
```
FabienMotte: Should we strip InstantSearch metadata like `_highlightResult` and so on?

packages/instantsearch.js/src/connectors/chat-page-suggestions/connectChatPageSuggestions.ts
```diff
@@ -0,0 +1,681 @@
...
+            if (__DEV__) {
+              warning(
+                false,
+                `No chat widget found in render state. Make sure a \`connectChat\` widget is mounted on the same index, or pass an \`onSuggestionClick\` prop to handle the click yourself.`
+              );
+            }
+            return false;
+          }
+          openChat(chatRenderState, {
+            message: prompt,
```
FabienMotte: We should reuse the prompt from the evaluation report to improve the quality of the agent's response and avoid sending the raw suggestion

packages/instantsearch.js/src/connectors/chat-page-suggestions/connectChatPageSuggestions.ts
```diff
@@ -0,0 +1,681 @@
...
+      agentId?: never;
+    };
+
+export type ChatPageSuggestionsConnectorParams = ChatPageSuggestionsSource & {
+  /**
+   * The server-owned Agent Studio task to invoke. The task's instructions,
+   * input schema, and output schema live in the agent config, not here.
+   * @default 'algolia_on_page_suggestions'
+   */
+  task?: string;
```
FabienMotte: Could we use `configurationId` or something similar? It will be displayed as a configuration in the Dashboard. `task` is the generic implementation of the backend.

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
FabienMotte, line 1: Same comment as @Haroenv 

Could we have a generic connector for tasks that matches the backend implementation?
This would allow us to reuse it later for other features while keeping the widgets tied to each feature

FabienMotte (changes requested):

packages/instantsearch-ui-components/src/components/chat/OnPageSuggestions.tsx
```diff
@@ -0,0 +1,175 @@
+/** @jsx createElement */
+/** @jsxFrag Fragment */
+import { cx } from '../../lib';
+import { createButtonComponent } from '../Button';
+
+import type { ComponentProps, Renderer } from '../../types';
+
+export type OnPageSuggestionsClassNames = {
```
Haroenv: I think you can have another round of thinking about this or a poll on slack, but I prefer `promptSuggestions` as the name for the widget, on page seems irrelevant here and is more a code name

packages/instantsearch.js/src/lib/ai-lite/parse-partial-json.ts
```diff
@@ -0,0 +1,89 @@
+export const tryParseJson = (value: string): unknown | undefined => {
+  try {
+    return JSON.parse(value);
+  } catch {
+    return undefined;
+  }
+};
+
+export const repairPartialJson = (value: string): string => {
```
Haroenv, line 9: good to have this outside of the main abstract chat, but what's the reason to move it in this PR?

packages/instantsearch.js/src/lib/chat/chat.ts
```diff
@@ -26,10 +26,17 @@ export const CACHE_KEY = 'instantsearch-chat-initial-messages';
 function getDefaultInitialMessages<TUIMessage extends UIMessage>(
   id?: string
 ): TUIMessage[] {
-  const initialMessages = sessionStorage.getItem(
-    CACHE_KEY + (id ? `-${id}` : '')
-  );
-  return initialMessages ? JSON.parse(initialMessages) : [];
+  if (typeof sessionStorage === 'undefined') {
```
Haroenv: change seems unrelated?

Haroenv: let's split this in a different PR?

packages/instantsearch.js/src/lib/chat/openChat.ts
```diff
@@ -5,7 +5,11 @@ import type { ChatRenderState } from '../../connectors/chat/connectChat';
  * Forwarded to the agent backend as the `x-algolia-referer` header and used
  * as a correlation tag for attribution.
  */
-export type ChatReferer = 'prompt-suggestions' | 'ai-mode';
+export type ChatReferer =
+  | 'prompt-suggestions'
+  | 'ai-mode'
+  | 'on-page-suggestions'
+  | 'page-summary';
```
Haroenv:
```suggestion
export type ChatReferer =
  | 'ai-mode'
  | 'prompt-suggestions-widget'
  | 'prompt-suggestions-autocomplete';
```

packages/instantsearch.js/src/lib/chat/sendMessageWithContext.ts
```diff
@@ -0,0 +1,41 @@
...
+        metadata: {
+          ...(message.metadata as Record<string, unknown> | undefined),
+          turnContext,
+        },
+      } as Parameters<typeof chat.sendMessage>[0],
+      ...rest
+    );
+  };
+
+  return sendMessage;
```
Haroenv: return directly instead of casting first? what's the reason?

packages/instantsearch-ui-components/src/components/chat/OnPageSuggestions.tsx
shaejaz: the rfc used this naming. maybe we should converge on this while the dashboard is also in progress. wdyt @FabienMotte ?

packages/instantsearch.js/src/lib/ai-lite/parse-partial-json.ts
shaejaz: this was moved to have it be imported by both files cleanly. should it be in the same file as before and be imported from there?

Haroenv: I mean that I don't see any usage of partial parse json outside of the main abstract chat where it was already, did I miss it?

shaejaz: Its being used in the new utils for tasks, `fetchTask` in particular. I think I'll consolidate these files some more, they are a bit divided unnecessarily atm.

packages/instantsearch.js/src/lib/chat/openChat.ts
shaejaz: will update the name here, once it's finalized.

packages/instantsearch.js/src/lib/chat/chat.ts
shaejaz: this change is needed for SSR to work correctly with page suggestions.

we could do the SSR support in a separate PR tho, it would be easier to review with those changes.

packages/instantsearch-ui-components/src/components/chat/ChatPromptSuggestions.tsx
Haroenv, line 1: what was this file used for before?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,497 @@
...
+    >;
+  };
+};
+
+export type PromptSuggestionsConnector = Connector<
+  PromptSuggestionsWidgetDescription,
+  PromptSuggestionsConnectorParams
+>;
+
+const INTERNAL_HIT_KEYS = [
```
Haroenv: what about just stripping everything starting with `_`?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,497 @@
...
+  INTERNAL_HIT_KEYS.forEach((key) => {
+    delete clean[key];
+  });
+  return clean;
+}
+
+const DEFAULT_TRANSFORM_HITS: PromptSuggestionsTransformHits = (hits) =>
+  hits.slice(0, 5).map(stripInternalHitMetadata);
+
+function buildFilters(results: SearchResults): string[][] | undefined {
```
Haroenv, line 146: have we tested if this logic is worth it? wonder if it's better to send the whole search results without transforming the parameters and doing any transform later backend?

packages/instantsearch.js/src/connectors/structured-output/connectStructuredOutput.ts
```diff
@@ -0,0 +1,215 @@
...
+  getAlgoliaAgent,
+  getAppIdAndApiKey,
+  noop,
+} from '../../lib/utils';
+
+import type { StructuredOutputRunner, TaskTransport } from '../../lib/tasks';
+import type { Connector } from '../../types';
+
+const withUsage = createDocumentationMessageGenerator({
+  name: 'structured-output',
```
Haroenv: we called this "tasks" backend

Haroenv (changes requested):

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: I believe the evaluations stated that the results are sent as hits plus the query/filters/categories.
So replicated that here without sending everything.

packages/instantsearch-ui-components/src/components/chat/ChatPromptSuggestions.tsx
shaejaz: Oop, the naming scheme is clashing with the new promp suggestions name in the connector. The find all probably flagged it there. I'll revert this part.

Sidenote: it seems there's no tests for chat prompt suggestions, probably should add them later.

packages/instantsearch.js/src/lib/chat/openChat.ts
```diff
@@ -5,7 +5,10 @@ import type { ChatRenderState } from '../../connectors/chat/connectChat';
  * Forwarded to the agent backend as the `x-algolia-referer` header and used
  * as a correlation tag for attribution.
  */
-export type ChatReferer = 'prompt-suggestions' | 'ai-mode';
+export type ChatReferer =
+  | 'prompt-suggestions'
+  | 'ai-mode'
+  | 'prompt-suggestions';
```
Haroenv: twice the same one? one of the two should be the autocomplete prompt suggestions, the other should be the standalone one. We should be able to distinguish

packages/react-instantsearch/src/widgets/PromptSuggestions.tsx
```diff
@@ -0,0 +1,104 @@
...
+  UsePromptSuggestionsProps & {
+    layoutComponent?: (
+      props: PromptSuggestionsLayoutComponentProps
+    ) => JSX.Element | null;
+    onSuggestionClick?: PromptSuggestionsOnSuggestionClick;
+  };
+
+export function PromptSuggestions({
+  classNames = {},
+  skeletonCount,
```
Haroenv: either not exposed on react, either also on js

Claude: `Cosmetic follow-on: the SCSS only defines nth-child(1..5), so skeletonCount > 5 gets no width/stagger — _prompt-suggestions.scss:34-38,64-82.`

packages/instantsearch.js/src/lib/tasks/endpoint.ts
```diff
@@ -0,0 +1,70 @@
...
+      '[tasks] Either `transport` or `{ appId, apiKey, agentId }` is required.'
+    );
+  }
+
+  const headers: Record<string, string> = {
+    'x-algolia-application-id': params.appId,
+    'x-algolia-api-key': params.apiKey,
+  };
+  if (params.algoliaAgent) {
+    headers['x-algolia-agent'] = params.algoliaAgent;
```
Haroenv: does this pass the `; tasks` agent, like chat does?

shaejaz: no it doesn't. updated it to send this now.

packages/instantsearch.js/src/lib/chat/openChat.ts
shaejaz: I've gone with `prompt-suggestions` for the widget and `prompt-suggestions-autocomplete` for autocomplete as it makes a bit more sense to me.

This does cause the previous `prompt-suggestions` associated with autocomplete to be mixed going forward. Only if we are saving this information currently. Wdyt?

packages/instantsearch.js/src/lib/chat/openChat.ts
```diff
@@ -6,9 +6,9 @@ import type { ChatRenderState } from '../../connectors/chat/connectChat';
  * as a correlation tag for attribution.
  */
 export type ChatReferer =
-  | 'prompt-suggestions'
   | 'ai-mode'
-  | 'prompt-suggestions';
+  | 'prompt-suggestions'
+  | 'prompt-suggestions-autocomplete';
```
Haroenv: doesn't this mean the old data from autocomplete will use a different referer? we probably need both to be a different name

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,492 @@
...
+            lastStateSignature = stateSignature;
+            clearTimeout(debounceTimer);
+            debounceTimer = setTimeout(() => {
+              if (latestRenderOptions?.results) {
+                fetchAndRender(
+                  latestRenderOptions.results,
+                  latestRenderOptions
+                );
+              }
+            }, DEBOUNCE_MS);
```
anjola-adeuyi, line 470: doesn't this still allow an older request to win during the debounce? if A resolves after the state changes but before the next `submit`, we render A's suggestions with B's results. should we invalidate the active request when the signature changes?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,492 @@
...
+          const results =
+            latestRenderOptions?.results ??
+            ('results' in renderOptions ? renderOptions.results : null) ??
+            null;
+          openChat(chatRenderState, {
+            message: buildSuggestionMessage(prompt),
+            referer: 'prompt-suggestions-widget',
+            turnContext: buildTurnContext(results),
+          });
+          return true;
```
anjola-adeuyi: shouldn't this return `openChat(...)`? right now it returns `true` even when the chat is busy and nothing was sent.

packages/instantsearch.js/src/connectors/tasks/connectTasks.ts
```diff
@@ -0,0 +1,214 @@
...
+  task: string;
+  stream?: boolean;
+};
+
+export type TasksWidgetDescription<TOutput = unknown> = {
+  $$type: 'ais.tasks';
+  renderState: TasksRenderState<TOutput>;
+};
+
+export type TasksConnector = Connector<
```
anjola-adeuyi: is `connectTasks` meant to be generic over its output? `TOutput` is dropped here, so consumers of the exported connector still get `unknown`

packages/instantsearch.js/src/connectors/tasks/connectTasks.ts
```diff
@@ -0,0 +1,214 @@
...
+      // Clear the previous output so consumers can show a loading state
+      // rather than stale data while the new request is in flight.
+      output = undefined;
+      error = undefined;
+      isLoading = true;
+      triggerRender();
+
+      return runner
+        .submit(variables, {
+          onData: stream
```
anjola-adeuyi: can `runner.submit` throw before returning the promise here? request building can fail synchronously, which leaves `isLoading` true without setting `error`. could we include the call in the promise chain?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: Seems like a rare edge case if this happens, but I've added invalidation logic is it does happen

packages/instantsearch.js/src/connectors/tasks/connectTasks.ts
shaejaz: good catch, moved the submit call in a then callback. it should now capture the thrown error in the catch block here.

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,499 @@
...
+    ?.suggestions;
+
+  if (!Array.isArray(suggestions)) {
+    return [];
+  }
+
+  return suggestions.filter(
+    (s: unknown): s is string => typeof s === 'string' && s.trim().length > 0
+  );
+}
```
Haroenv, line 46: why would there be an empty suggestion?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
Haroenv: thanks for solving, isn't that rare as A could take longer to process than B

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,499 @@
...
+      };
+
+      const tasksWidget = connectTasks(
+        handleInnerRender,
+        noop
+      )({
+        ...(transport ? { transport } : { agentId }),
+        task: configurationId ?? 'prompt-suggestions',
+        stream: true,
+      } as TasksConnectorParams);
```
Haroenv, line 422: so we create a widget, but don't explicitly render it, this seems fragile. What about returning both widgets in an array and using render state to access its items? otherwise it seems like it could cause issues later (maybe we don't call the tasks widget in all the hooks for example)

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,499 @@
...
+        if (!latestRenderOptions) return;
+        renderOutward(latestRenderOptions);
+      };
+
+      const tasksWidget = connectTasks(
+        handleInnerRender,
+        noop
+      )({
+        ...(transport ? { transport } : { agentId }),
+        task: configurationId ?? 'prompt-suggestions',
```
Haroenv: I don't think we need a fallback

Haroenv: unless the "first created prompt suggestion" always has that id, cc @FabienMotte, it would be possible.

Haroenv (commented):
couple questions about edge cases

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,499 @@
...
+        results: SearchResults | null
+      ): Record<string, string> | undefined => {
+        const pageContext = resolvePageContext(results);
+        if (!pageContext) {
+          return undefined;
+        }
+        const entries = Object.entries(pageContext)
+          .map(([key, value]): [string, string] => [
+            key,
+            typeof value === 'string' ? value : JSON.stringify(value),
```
anjola-adeuyi, line 320: should the chat context be shaped separately here? a valid task input can put `hitsSample` over chat’s 1 KB per-value limit, so clicking a rendered suggestion can fail with 422

packages/instantsearch.js/src/widgets/prompt-suggestions/prompt-suggestions.tsx
```diff
@@ -0,0 +1,197 @@
...
+      onSuggestionClick,
+      isChatBusy,
+      sendToChat,
+    } = props;
+
+    const handleClick = onSuggestionClickOverride
+      ? (prompt: string) => onSuggestionClickOverride(prompt, { sendToChat })
+      : onSuggestionClick;
+
+    if (templates?.layout) {
```
anjola-adeuyi, line 141: should `layout` and `header` use `prepareTemplateProps` and `TemplateComponent` like other JS widgets? `layout(data, { html })` currently receives only `data`, so the documented template throws

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,499 @@
...
+        const results =
+          'results' in renderOptions ? renderOptions.results : undefined;
+        const transformed = transformItems(suggestions, {
+          query: results?.query || '',
+          results: results || null,
+        });
+
+        const chatRenderState = getChatRenderState(renderOptions);
+
+        const isChatBusy = chatRenderState
```
anjola-adeuyi, line 378: should chat status transitions schedule a full InstantSearch render? after chat moves from `submitted` to `ready`, the shared render state can stay at `submitted`, leaving these pills disabled

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,499 @@
...
+          false
+        );
+      };
+
+      const fetchAndRender = (
+        results: SearchResults,
+        renderOptions: RenderOptions
+      ) => {
+        if (disposed || !tasksState) return;
+        refetchPending = false;
```
anjola-adeuyi, line 342: should this invalidate the active task before returning? when search state A changes to a no-hits state B, A can still resolve and render stale pills into B

packages/instantsearch.js/src/lib/tasks/fetchTask.ts
```diff
@@ -0,0 +1,141 @@
...
+): Promise<unknown> {
+  return new Promise<unknown>((resolve, reject) => {
+    const chunkStream = parseJsonEventStream(
+      body
+    ) as unknown as ReadableStream<TaskStreamChunk>;
+    let latest: unknown;
+    processStream(
+      chunkStream,
+      (chunk) => {
+        if (!chunk || chunk.type !== 'data-task-output') {
```
anjola-adeuyi: should this reject when the stream emits an `error` event? the backend can send partial output before a terminal error but `[DONE]` currently resolves the last partial as success

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: so basically have `configurationId` be a required prop?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: that was how it was done initially, but i thought since the task connector is more an internal structure, maybe we don't want our renderstate to have this information.

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: yes this is already the case. this should be rechecked once the 10kb limit is merged on the backend.

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: it won't render since there's now a check for this var there.

not sure about the invalidate active task, having cancellation support for that request seems a bit more complicated, perhaps in a separate PR?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: i believe this already happens, yes. will double check

packages/instantsearch.js/src/lib/tasks/fetchTask.ts
shaejaz: yeah we don't really have error states for the pills atm.

I can add that check here, but not sure how to display that to a user. For now a blank state should be okay i think

anjola-adeuyi: a blank state is fine, no error UI is needed. could the `error` event reject the task and the prompt connector clear any streamed partial so a failed stream can’t leave partial pills visible?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
anjola-adeuyi: `refetchPending` is set to `false` before this return, so the guard in handleInnerRender still accepts A’s late result. could we invalidate A here without cancelling the request?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
anjola-adeuyi: thanks, that helps. if this PR depends on the 10 KB backend rollout, what keeps the default five-hit hitsSample within that limit?

shaejaz: Nothing yet. Ideally there would be a way to select with attributes are significant for a record to pass along to the model, but even that could cause the limit to be hit.

I think it's okay to think about this once the limits have been increased.

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
Haroenv: either it's required, or there's a known "first configuration id"

packages/instantsearch.js/src/lib/tasks/fetchTask.ts
shaejaz: yes good idea. done that here now explicitly

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: added some invalidate logic to the tasks connector. i think this should cover the scenario you are talking about. lmk.

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: let's leave it as is for now and revisit this in the refactor exploration next week.

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
anjola-adeuyi: nice, looks good

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
anjola-adeuyi: while the first Chat response is pending, the pills stay enabled and a second click queues another request. after both responses finish, the pills stay disabled. could we rerender this widget on each Chat status change and cover both transitions?

packages/instantsearch-ui-components/src/components/chat/PromptSuggestions.tsx
```diff
@@ -0,0 +1,175 @@
...
+                key={i}
+                className={cx(
+                  'ais-PromptSuggestions-skeletonItem',
+                  classNames.skeletonItem
+                )}
+              />
+            ))}
+          </div>
+        ) : (
+          suggestions.map((suggestion, index) => (
```
anjola-adeuyi: while suggestions are still streaming, a partial value such as `Wh..` is an enabled button, so clicking it sends the unfinished prompt to Chat. could we keep the pills disabled until generation finishes and drop blank values?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,507 @@
...
+      // the widget is unmounted; this guard stops those late callbacks from
+      // calling `renderFn` into a torn-down container.
+      let disposed = false;
+      // True between a state-signature change and the debounced refetch that
+      // follows it. While pending, the search state has already moved on, so a
+      // still-in-flight request from the previous state must not paint its
+      // suggestions, its inner render is ignored until the new `submit` starts.
+      let refetchPending = false;
+
+      const getStateSignature = (results: SearchResults): string => {
```
anjola-adeuyi, line 233: when a search refreshes with the same query, filters and object IDs, changed hit fields do not start a new suggestion request, so the pills can keep prompts made from old hit data. could we include the transformed hit content in this signature and cover that refresh?

packages/instantsearch.js/src/connectors/tasks/connectTasks.ts
```diff
@@ -0,0 +1,238 @@
...
+            onData: stream
+              ? (partial) => {
+                  if (isStale()) return;
+                  output = partial as TOutput;
+                  triggerRender();
+                }
+              : undefined,
+          })
+        )
+        .then((next) => {
```
anjola-adeuyi: if an older `submit()` finishes after a newer one, its promise resolves with the newer call's output because this reads the shared output. could each promise resolve with its own result while stale calls still avoid updating widget state?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
```diff
@@ -0,0 +1,507 @@
...
+      // a flat `Record<string, string>` per the Agent Studio contract, so
+      // non-string values (e.g. `hitsSample`) are serialized.
+      const buildTurnContext = (
+        results: SearchResults | null
+      ): Record<string, string> | undefined => {
+        const pageContext = resolvePageContext(results);
+        if (!pageContext) {
+          return undefined;
+        }
+        const entries = Object.entries(pageContext)
```
anjola-adeuyi, line 317: `context` accepts `unknown` values, so `{ optionalValue: undefined }` reaches this map. `JSON.stringify` produces `undefined` and the next `trim()` throws a type error when a pill is clicked. could we please skip values that do not serialize to a string and cover this case?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: the filter and object id signature is a fallback. the query id should be there for to trigger a new request.

the hit content would be a bit much to determine state no?

packages/instantsearch.js/src/connectors/prompt-suggestions/connectPromptSuggestions.ts
shaejaz: that is unknown as the hits are being fed into it. where would `optionalValue: undefined` be set, there are default being set for the filters and query no?

also there is a check for typeof of the value right?

anjola-adeuyi (commented):
looks good to me. the frontend context behavior matches today’s decision to remove the backend limit rather than truncate on the client. the remaining status-sync and serialization cases can follow separately

anjola-adeuyi (approved):

Haroenv (approved):

FabienMotte (approved):

PR #7062, opened 2026-05-29
fix(chat): show loader during gap in tool calls

shaejaz:
This PR fixes the gap in the loader being shown when there are multiple tool calls. The gap was caused by the backend stopping streaming slightly between the tool calls (during the step finish and step start stages).

<details>
<summary>Demo with the display results tool</summary>
<video src="https://github.com/user-attachments/assets/95536c7e-201d-4b9e-bb70-ef12faab9c3b" />
</details>

<details>
<summary>Demo with only search index tool</summary>
<video src="https://github.com/user-attachments/assets/2ff37ded-c316-419e-b125-9017d1c5f44f" />
</details>

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/ChatMessages.tsx
@@ -32,7 +32,13 @@ import type {
 } from './ChatMessage';
 import type { ChatMessageErrorProps } from './ChatMessageError';
 import type { ChatMessageLoaderProps } from './ChatMessageLoader';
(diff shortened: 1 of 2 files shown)


FabienMotte (approved):

Haroenv (approved):
this means that if the tool is the very last part, a loader will be shown forever, no? in the future I believe some tools can be terminal, but maybe then something else indicates that the stream is done?

shaejaz:
> this means that if the tool is the very last part, a loader will be shown forever, no? in the future I believe some tools can be terminal, but maybe then something else indicates that the stream is done?

I won't be entirely since the chat status check at the start would return false. But there is a chance there might be some loader jumping/gaps depending on which steps are sent when the tool is last. I'll add some test cases later to cover those, since that scenario can't be tested easily atm.

PR #7063, opened 2026-05-29
fix(react-instantsearch-nextjs): refresh results on client-side navigation (#7060)

### LINKED ISSUE #7060: [react-instantsearch-next]: Regression: hits not refreshed on client-side navigation (stale results)
### 🐛 Current behavior

After updating to the latest versions, server-rendered results stop being reflected on client-side navigation between SSR'd InstantSearchNext pages in the Next.js App Router. 

Two symptoms:
1. Listing → listing: hits keep showing the previous route's results.
2. First soft-navigation from a non-search page (e.g. home → a listing page): NO results at all (empty list).
A full reload (F5) always renders the correct results.

This is a regression - it does not happen on the older versions. Downgrading the following back to the left-hand versions makes it work again:
- react-instantsearch:        7.29.0  → 7.33.0   (breaks)
- react-instantsearch-nextjs: 1.1.4   → 1.2.4    (breaks)
- algoliasearch:              5.50.1  → 5.52.1    (breaks)

Environment:
- next 16.2.6 (App Router)
- react / react-dom 19.2.4
- newest (broken) set: react-instantsearch 7.33.0, react-instantsearch-nextjs 1.2.4, algoliasearch 5.52.1
- older (working) set:  react-instantsearch 7.29.0, react-instantsearch-nextjs 1.1.4, algoliasearch 5.50.1

### 🔍 Steps to reproduce

Reproduction repo: https://github.com/ptylek/instantsearch-nextjs-bug

1. npm install && npm run dev, open http://localhost:3000
2. Click a category link (client-side navigation) → empty list (symptom 2).
3. Hard-reload that page (F5) → correct results render.
4. Click another listing link (category or search) → stale results from the
   previous route (symptom 1).
5. Checkout downgrade branch and run npm install && npm run dev, open http://localhost:3000. On this branch there is downgraded version react-instantsearch@7.29.0, react-instantsearch-nextjs@1.1.4,
   algoliasearch@5.50.1, repeat steps 2–4 → everything works as expected.

Routes:
- /c/[slug]  → category page, Configure filters="categories:<slug>", force-dynamic
- /s?query=  → search page, query via initialUiState, force-dynamic
- /          → plain page with links into the listing routes

### Live reproduction

https://github.com/ptylek/instantsearch-nextjs-bug

### 💭 Expected behavior

Please clone repo and run it locally.

Using newest version of packages everything should work as before. If there are some breaking changes, please provide how to update code to align with new versions.

Old versions work properly.

### Package version

react-instantsearch 7.33.0, react-instantsearch-nextjs 1.2.4, algoliasearch 5.52.1

### Operating system

macOS

### Browser

all

### Code of Conduct

Haroenv:
## Summary

Fixes #7060. After [#6995](https://github.com/algolia/instantsearch/pull/6995), server-rendered results stopped being reflected on **client-side (soft) navigation** in the Next.js App Router — the destination page showed empty or stale hits until a full reload.

## Root cause

#6995 ("preserve URL on mount with nested `<Index>`", fixing #6980) made `useInstantSearchRouting` **skip `onUpdate()` on its first effect run**, because on the initial hydration `subscribe()` already merged the URL into `_initialUiState` and a redundant `setUiState` could wipe the URL with a nested `<Index>` (whose children only register on a second render pass).

But on a client-side navigation the App Router **remounts** `InstantSearchNext`, so the new instance's first run looks identical to a first hydration — `onUpdate()` was skipped, no fresh search ran, and the remounted instance rendered only whatever `window[InstantSearchInitialResults]` held: empty when arriving from a page without InstantSearch, or stale otherwise. A hard reload worked because the document re-injects correct initial results.

## Fix

Leave `useInstantSearchRouting` exactly as #6995 left it (so #6980 is fully preserved — the routing hook is unchanged versus `master`) and add a small client-only `RefreshOnClientNavigation` component that calls `search.refresh()` when it detects a client-side navigation.

`refresh()` re-fetches with the already-correct current UI state and **does not touch the URL**, so it fixes #7060 without any risk of the #6980 wipe. Navigation is detected with the **Navigation Timing API** (`performance.getEntriesByType('navigation')[0].name`), which holds the URL the document was hard-loaded with and is unaffected by SPA navigations — comparing `pathname` + `search` (hash ignored). This also covers the case where the navigation lands on the **first** InstantSearch page of the session (e.g. coming from a page without InstantSearch). A `window` fallback covers environments without Navigation Timing.

The change is purely additive (the routing hook and its tests are untouched).

## Tradeoff

Every client-side navigation now re-fetches rather than reusing the streamed SSR results. This restores the **pre-#6995** behavior (not a new regression versus before the bug) and is the only safe option: on a soft navigation the injected results may be empty *or* stale, and a cheap check can't tell a *correct* injected result from a *stale* one — so trusting them would re-break the stale case.

## Tests

- **Unit** (`RefreshOnClientNavigation.test.tsx`): `isClientNavigation()` (pathname and search-param sensitivity), refresh is called on a client navigation and **not** on initial hydration, and the URL is preserved on a soft navigation to a nested `<Index>` (the #6980 failure mode, here on soft navigation — verified the previous `onUpdate` approach wiped it). Navigation Timing is mocked so the production path is exercised.
- **E2E** (`softNavigation.test.ts` + a `/landing` fixture page with no InstantSearch): navigating `/landing → /Appliances` asserts results render. Verified this fails on the #6995 build (0 hits) and passes with the fix.

All green locally: unit 17/17, e2e 7/7 (incl. existing `backButton`/`multipleHooks`).

### DIFF
--- a/examples/react/next-app-router/app/landing/page.tsx
+++ b/examples/react/next-app-router/app/landing/page.tsx
@@ -1,0 +1,17 @@
+import Link from 'next/link';
+import React from 'react';
+
+// A plain page with no InstantSearch, mirroring the reproduction in #7060
+// where the first soft-navigation lands on an InstantSearch page.
+export const dynamic = 'force-dynamic';
+
+export default function Landing() {
+  return (
+    <div>
+      <h1>Landing</h1>
+      <Link href="/Appliances" id="to-appliances">
+        Go to Appliances
+      </Link>
+    </div>
+  );
+}
--- a/packages/react-instantsearch-nextjs/__tests__/e2e/softNavigation.test.ts
+++ b/packages/react-instantsearch-nextjs/__tests__/e2e/softNavigation.test.ts
@@ -1,0 +1,21 @@
+import { test, expect } from '@playwright/test';
+
+// Reproduces https://github.com/algolia/instantsearch/issues/7060:
+// on a client-side (soft) navigation that lands on the first InstantSearch page
+// of the session (here, coming from a page without InstantSearch), the
+// server-rendered results stopped being reflected and the page showed empty
+// hits until a full reload. The `/landing` page has no InstantSearch, so this
+// exercises exactly that case. Verified to fail before the fix (0 hits).
+test.describe('client-side navigation refreshes results', () => {
+  test('renders results when navigating from a page without InstantSearch', async ({
+    page,
+  }) => {
+    await page.goto('/landing');
+
+    await page.locator('#to-appliances').click();
+    await expect(page).toHaveURL('http://localhost:3000/Appliances');
+
+    await expect(page.locator('.ais-Hits-item').first()).toBeVisible();
+    await expect(page.locator('.ais-Hits-item')).not.toHaveCount(0);
+  });
+});
--- a/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
+++ b/packages/react-instantsearch-nextjs/src/InstantSearchNext.tsx
@@ -7,6 +7,7 @@ import {
 } from 'react-instantsearch-core';
 
 import { InitializePromise } from './InitializePromise';
+import { RefreshOnClientNavigation } from './RefreshOnClientNavigation';
 import { TriggerSearch } from './TriggerSearch';
 import { useDynamicRouteWarning } from './useDynamicRouteWarning';
 import { useInstantSearchRouting } from './useInstantSearchRouting';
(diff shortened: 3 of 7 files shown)


Haroenv:
Thanks Copilot — both points were spot on and are addressed:

1. **Pathname-only comparison** — the navigation check now compares `pathname` + `search` (hash ignored), so a same-path/different-search soft navigation (e.g. `/search?q=old` → `/search?q=new`) is correctly detected. Covered by a unit test.

2. **Nested `<Index>` on soft navigation** — confirmed real (a soft navigation to a nested-`<Index>` page wiped the URL with the `onUpdate` approach). Rather than guard/delay `onUpdate`, the fix no longer re-runs `onUpdate` at all: `useInstantSearchRouting` is left exactly as #6995 left it, and a new `RefreshOnClientNavigation` component calls `search.refresh()` instead. `refresh()` re-fetches with the current state and never touches the URL, so it can't wipe a nested `<Index>`. Added a regression test asserting the URL is preserved on a soft navigation to a nested `<Index>`.

_— via Claude_

PR #7064, opened 2026-06-01
fix(chat): hydrate full objects in display results tool

shaejaz:
This PR hydrates the display results tool output results with the entire object body. The output from backend for the display results only contains objectIDs, we already have these entire object body from the preceding search index tool.

A map for the objects is created and used in the default display results layout component.

image

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/tools/DisplayResultsTool.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/tools/DisplayResultsTool.tsx
@@ -66,10 +66,6 @@ export function createDisplayResultsToolComponent<
       ...userTranslations,
     };
 
-    // The backend only sends object IDs for display results, so we rebuild the
-    // full records from the hits the preceding search tool already fetched.
-    // `getHitsByObjectID` is memoized on the `messages` reference, so this is
-    // not recomputed on unrelated re-renders.
     const hitsByObjectID = messages ? getHitsByObjectID(messages) : undefined;
 
     const output = message?.output as DisplayResultsOutput<TObject> | undefined;
--- a/packages/instantsearch-ui-components/src/components/chat/tools/DisplayResultsTool.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/tools/DisplayResultsTool.tsx
@@ -24,6 +24,17 @@ type DisplayResultsOutput<THit> = {
   groups?: Array<DisplayResultsGroup<THit>>;
 };
 
+/**
+ * An item handed to a group's carousel: the record (hydrated from the search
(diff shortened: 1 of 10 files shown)


FabienMotte:
> This PR hydrates the display results tool output results with the entire object body. The output from backend for the display results only contains objectIDs, we already have these entire object body from the preceding search index tool.
> 
> A map for the objects is created and used in the default display results layout component.
> 
> image

Do you know why the width of each hit is so small in the carousel?

shaejaz:
> Do you know why the width of each hit is so small in the carousel?

This is due to the item component styles that I didn't really size properly. Since it's the examples, the code doesn't belong in this PR (until we start shipping the display results tool as a default I guess)

Can be update user side for example:
image

packages/instantsearch-ui-components/src/components/chat/tools/DisplayResultsTool.tsx
```diff
@@ -94,10 +107,21 @@ export function createDisplayResultsToolComponent<
...
+              | RecordWithObjectID<TObject>
+              | undefined;
+
+            return {
+              ...hydrated,
+              ...result,
+              // When hydrated, keep the record's real `objectID` (the
+              // display tool references it by a stripped `id`), so click events
+              // and Insights report the correct objectID.
+              ...(hydrated ? { objectID: hydrated.objectID } : {}),
```
Haroenv: why do we even keep the result object then, if we override objectID anyway? what about swapping the order and doing `...result, ...hydrated` or just pass `...hydrated`? are there other keys in result?

Haroenv (commented):
idea good, can maybe be improved

packages/instantsearch-ui-components/src/components/chat/tools/DisplayResultsTool.tsx
```diff
@@ -107,21 +118,20 @@ export function createDisplayResultsToolComponent<
...
-          }) as Array<RecordWithObjectID<TObject>>;
+          const items: Array<DisplayResultsItem<TObject>> = results.map(
+            (result, idx) => {
+              const hydrated = hitsByObjectID?.[result.objectID] as
+                | RecordWithObjectID<TObject>
+                | undefined;
+
+              return {
+                ...(hydrated as RecordWithObjectID<TObject>),
+                objectID: result.objectID,
```
Haroenv, line 129: same question, why would it be different? or is this for the fallback if hydrated isn't found?

shaejaz: yes, it's a fallback in case it's not found in the map

Haroenv (approved):

FabienMotte (approved):

PR #7069, opened 2026-06-08
fix(chat): stop carousel remount, transcript re-render, and lost auto-scroll while streaming

Haroenv:
## Summary

A customer reported flickering in the chat widget tied to re-renders — while streaming, and when scrolling/interacting mid-stream. Investigating surfaced two distinct root causes (one commit each), plus a regression that the first fix exposed.

### 1. Carousel re-mounts every streaming delta — `fix(chat): keep carousel mounted and view pinned while streaming`
The chat renderer rebuilt each tool's `layoutComponent` on every render, so Preact saw a **new component type** each delta and remounted the whole tool subtree — re-mounting a carousel's `<ol>` and resetting its horizontal scroll position. We now cache one stable component per tool key, mirroring the existing stable chat-template pattern (`createStableTemplateComponent`).

Fixing the remount removed the incidental content-height thrash that the stick-to-bottom `ResizeObserver` was (accidentally) relying on — so the conversation stopped following the stream (the `ResizeObserver` only reacts to **height**, and a carousel grows horizontally). We re-pin to the bottom on message/status updates while streaming, passing `preserveScrollPosition` so it never fights a user who has scrolled up.

> Note: the carousel remount was the **instantsearch.js (Preact)** path only — `react-instantsearch` already memoizes its `tools`.

### 2. Whole transcript re-renders on every token — `fix(chat): memoize message rows to stop full-transcript re-render`
Message rows were never memoized, so every streaming delta re-rendered **every** message and re-compiled its markdown — the flicker source. We add an optional `memo` to the shared renderer and wrap the default message row with a comparator keyed on the props that actually affect a row (`message`, `status`, `suggestionsElement`, and that message's feedback state). Because `replaceMessage` only clones the message being updated, completed messages keep a stable reference and now render **once** instead of on every token.

`indexUiState` is deliberately excluded from the comparator (`getUiState()` returns a fresh object each render, which would defeat the memo); the trade-off is documented inline.

## Tests
Behavioral contracts live in the **cross-flavor common suite** (`tests/common/widgets/chat/streaming.tsx`), so they run against **both `instantsearch.js` and `react-instantsearch`**:
- *streaming carousel keeps its DOM node + scroll position across deltas* — catches the remount (JS).
- *a completed message is not re-rendered on every streaming delta* — catches the missing memo (**both** flavors).

Both were verified red→green by toggling each fix off. Auto-scroll has a flavor-specific test (`chat-autoscroll.test.tsx`, Preact) — scroll isn't observable in jsdom without layout, so it's asserted by spying the stick-to-bottom hook.

All chat suites pass; lint clean; type-check clean (pre-existing `instantsearch-cli` type errors are unrelated).

## Review notes
- Two commits, one per root cause; reviewing commit-by-commit is easiest.

### DIFF
--- a/packages/instantsearch-ui-components/src/components/chat/Chat.tsx
+++ b/packages/instantsearch-ui-components/src/components/chat/Chat.tsx
@@ -9,7 +9,7 @@ import { createChatPromptSuggestionsComponent } from './ChatPromptSuggestions';
 
 import type { Renderer, ComponentProps } from '../../types';
 import type { ChatHeaderProps, ChatHeaderOwnProps } from './ChatHeader';
-import type { ChatMessagesProps } from './ChatMessages';
+import type { ChatMessagesProps, Memo } from './ChatMessages';
 import type { ChatPromptProps, ChatPromptOwnProps } from './ChatPrompt';
 import type { ChatPromptSuggestionsOwnProps } from './ChatPromptSuggestions';
 import type { ChatLayoutOwnProps } from './types';
@@ -92,9 +92,17 @@ export type ChatProps = Omit<ComponentProps<'div'>, 'onError' | 'title'> & {
   layoutComponent?: (props: ChatLayoutOwnProps) => JSX.Element;
 };
 
-export function createChatComponent({ createElement, Fragment }: Renderer) {
+export function createChatComponent({
+  createElement,
+  Fragment,
+  memo,
+}: Renderer & { memo?: Memo }) {
   const ChatHeader = createChatHeaderComponent({ createElement, Fragment });
-  const ChatMessages = createChatMessagesComponent({ createElement, Fragment });
+  const ChatMessages = createChatMessagesComponent({
+    createElement,
+    Fragment,
+    memo,
+  });
   const ChatPrompt = createChatPromptComponent({ createElement, Fragment });
   const ChatPromptSuggestions = createChatPromptSuggestionsComponent({
     createElement,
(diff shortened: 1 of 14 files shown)


packages/instantsearch-ui-components/src/components/chat/Chat.tsx
```diff
@@ -9,7 +9,7 @@ import { createChatPromptSuggestionsComponent } from './ChatPromptSuggestions';
 
 import type { Renderer, ComponentProps } from '../../types';
 import type { ChatHeaderProps, ChatHeaderOwnProps } from './ChatHeader';
-import type { ChatMessagesProps } from './ChatMessages';
+import type { ChatMessagesProps, Memo } from './ChatMessages';
```
shaejaz: I think we should reuse the same memoization logic from the autocomplete prop getters or the display results tool component. they use a useMemo hook as a parameter from the flavor (though that approach does not allow Vue).

In that case maybe we can have a shared type for that hook (and other primitive react/preact hooks) be defined in ui-components.

Wdyt?

Haroenv: the problem is that we need to use Memo and not useMemo here because we want to memo the whole message component, not just a single value. Not sure I see a way to consolidate between the two. Maybe in a follow-up?

afrencalg (commented):
LGTM

afrencalg (approved):

shaejaz (approved):

PR #7081, opened 2026-06-22
fix(chat): handle tool-input-error and tool-output-error stream chunks

kombucha:
**Summary**

ai-lite consumed a `tool-error` chunk that the AI SDK 5 wire protocol never emits. 
The real chunks are 
- `tool-input-error` (args partial/unparseable, the tool never ran); 
- `tool-output-error` (the tool ran and execution failed). 

`tool-input-error` always implies a following `tool-output-error`; the reverse isn't true. 
Both fell through to the `data-*` `default` branch and were silently dropped, leaving the tool part stuck in `input-streaming` forever.

This handles both, matching the AI SDK reducer:
- `tool-input-error` becomes `output-error`, clearing `input` and preserving the partial JSON on `rawInput`. It creates the part when a degenerate stream never emitted `tool-input-start` (the unbounded-args abort path).
- `tool-output-error` bcomes `output-error`, keeping the prior `input`/`toolName` and swapping in the error. Silently dropped (rather than thrown, as Vercel does) when no prior part exists.
- The wire `providerMetadata` field is mapped onto the part's `callProviderMetadata` (it was read under the wrong name and always dropped).
- `rawInput?` is mirrored onto `DynamicToolUIPart.output-error` for parity with the static part.

The legacy `tool-error` chunk is removed as it never existed in the first place

**Result**

New unit suite for `AbstractChat` (there were none), driving real SSE bytes through the production parser, including a verbatim capture of an AI SDK 5 server emitting malformed tool-call JSON.

```
yarn jest packages/instantsearch.js/src/lib/ai-lite
# 13 passed (abstract-chat) — 74 passed across ai-lite + connectChat
```

### DIFF
--- a/packages/instantsearch.js/src/lib/ai-lite/__tests__/abstract-chat.test.ts
+++ b/packages/instantsearch.js/src/lib/ai-lite/__tests__/abstract-chat.test.ts
@@ -1,0 +1,624 @@
+/**
+ * @jest-environment @instantsearch/testutils/jest-environment-jsdom.ts
+ */
+import { AbstractChat } from '../abstract-chat';
+import { parseJsonEventStream } from '../stream-parser';
+
+import type {
+  ChatState,
+  ChatTransport,
+  UIMessage,
+  UIMessageChunk,
+} from '../types';
+
+class InMemoryChatState implements ChatState<UIMessage> {
+  status: ChatState<UIMessage>['status'] = 'ready';
+  error: Error | undefined = undefined;
+  messages: UIMessage[] = [];
+
+  pushMessage = (message: UIMessage): void => {
+    this.messages = [...this.messages, message];
+  };
+
+  popMessage = (): void => {
+    this.messages = this.messages.slice(0, -1);
+  };
+
+  replaceMessage = (index: number, message: UIMessage): void => {
+    this.messages = [
+      ...this.messages.slice(0, index),
+      message,
+      ...this.messages.slice(index + 1),
+    ];
+  };
+
+  snapshot = <T>(thing: T): T => thing;
+}
+
+class TestChat extends AbstractChat<UIMessage> {}
+
+/**
+ * Builds a `ReadableStream<Uint8Array>` from raw SSE text, flushing one line
+ * at a time (one `Uint8Array` per line). This mimics how a server flushes
+ * events and exercises the parser's cross-chunk buffering, just like the bytes
+ * a real `fetch` response body would yield.
+ */
+function sseToByteStream(raw: string): ReadableStream<Uint8Array> {
+  const encoder = new TextEncoder();
+  const lines = raw.split('\n');
+  let index = 0;
+
+  return new ReadableStream<Uint8Array>({
+    pull(controller) {
+      if (index < lines.length) {
+        const isLast = index === lines.length - 1;
+        controller.enqueue(encoder.encode(lines[index] + (isLast ? '' : '\n')));
+        index += 1;
+      } else {
+        controller.close();
+      }
+    },
+  });
+}
+
+/**
+ * Wraps already-parsed chunk objects in a `ReadableStream<UIMessageChunk>` —
+ * what the transport hands `AbstractChat` after parsing SSE itself. The default
+ * path: reducer tests don't need the byte round-trip (the parser has its own
+ * suite in `stream-parser.test.ts`), so they feed chunks straight in.
+ */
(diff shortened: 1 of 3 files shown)


packages/instantsearch.js/src/lib/ai-lite/__tests__/abstract-chat.test.ts
```diff
@@ -0,0 +1,624 @@
...
+  });
+
+  describe('verbatim AI SDK 5 server stream', () => {
+    // Captured verbatim from a real AI SDK 5 server where the model produced
+    // malformed tool-call JSON (a missing closing brace).
+    const REAL_ERROR_STREAM = [
+      'data: {"type":"start"}',
+      'data: {"type":"start-step"}',
+      'data: {"type":"tool-input-start","toolCallId":"call_1","toolName":"algolia_search_index"}',
+      'data: {"type":"tool-input-delta","toolCallId":"call_1","inputTextDelta":"{\\""}',
```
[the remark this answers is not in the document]
kombucha: This line is valid JSON. The `\"` is an escaped quote. Copilot got confused

packages/instantsearch.js/src/lib/ai-lite/types.ts
```diff
@@ -269,14 +270,25 @@ type ToolUIMessageChunk<TOOLS extends UITools> =
...
+        providerMetadata?: ProviderMetadata;
       };
     }>
+  | {
+      type: 'tool-output-error';
+      toolCallId: string;
+      errorText: string;
+      providerExecuted?: boolean;
+      providerMetadata?: ProviderMetadata;
+      dynamic?: boolean;
```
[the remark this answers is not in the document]
kombucha: The AI SDK wire schema for this chunk is `dynamic: z.boolean().optional()`, so `false` is a valid value. And `tool-output-error` is a single union member (no `toolName` to key a static variant on), so `dynamic` is not a discriminant here; narrowing happens on `type`. Keeping `dynamic?: boolean` to match the wire. The reducer never reads the field.

FabienMotte (approved):

packages/instantsearch.js/src/lib/ai-lite/abstract-chat.ts
```diff
@@ -964,39 +964,96 @@ export abstract class AbstractChat<TUIMessage extends UIMessage> {
...
+                rawOutput: _ignoredRawOutput,
+                // eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
+                preliminary: _ignoredPreliminary,
+                // eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
+                output: _ignoredOutput,
+                ...rest
+              } = existingPart;
+              updatedParts[toolIndex] = {
+                ...rest,
+                state: 'output-error',
```
shaejaz, line 1045: I'm not completely sure that updating the state here propagates an error message in the UI. I can see that the UI component checks for the status only. Perhaps it does work somehow with the try/catch wrapping the process stream invocation.

This might be able to be tested via the common test suite, but that is probably out of scope for this PR (including the potential state issue, I can see it was the same as before).

Something for a future PR to look into.

kombucha: By default nothing is rendered indeed. In my tests I had to do some custom rendering for errors:

```ts
function SearchIndexLayoutWithError(props) {
  const { message } = props;
  if (message.state === "output-error") {
    return (
      <div className="ais-ChatMessage-toolError" role="alert">
        <strong>Search failed</strong>
        <p>{message.errorText || "The search tool returned an error."}</p>
        {message.rawInput && (
          <pre>{JSON.stringify(message.rawInput, null, 2)}</pre>
        )}
      </div>
    );
  }
  return <DefaultSearchIndexLayout {...props} />;
}

const chatTools = {
  [SearchIndexToolType]: {
    ...defaultTools[SearchIndexToolType],
    layoutComponent: SearchIndexLayoutWithError,
  },
};
```

I didn't feel comfortable fixing that in this PR, because you need to decide a bit more behaviour and what appearance you want by default 😅 
So I stuck to fixing the "correctness" of the protocol implementation 👍

shaejaz: Thanks for confirming! Will keep track of this for us to do.

shaejaz (approved):

PR #7088, opened 2026-06-23
feat(chat): add request options to agent requests

anjola-adeuyi:
**Summary**

[FX-3833](https://algolia.atlassian.net/browse/FX-3833)

Adds `requestOptions` for built-in `agentId` chat requests so consumers can pass persistent query parameters and headers to Agent Studio completion requests.

This is scoped to the built-in Agent Studio path only:
- supports `queryParameters`
- supports `headers`
- rejects `requestOptions` with custom `transport` or custom `chat`
- keeps body, metadata, resume/reconnect and custom transport behavior out of scope

Regenerate still forces `cache=false`.

**Documentation**

FX-3833 also includes a documentation update. The documentation source lives in `algolia/docs-new`, so this InstantSearch PR keeps the implementation scoped and the docs-new follow-up should verify/update the generated chat widget API docs for `requestOptions`.

**Test plan**

- [x] Focused connector tests for persistent query parameters
- [x] Focused connector tests for object headers and `Headers`
- [x] Per-call headers override persistent headers for one request
- [x] Regenerate keeps `cache=false`
- [x] `yarn jest packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts --runInBand`
- [x] Prettier check
- [x] `git diff --check`
- [x] `yarn lint:changed`

[FX-3833]: https://algolia.atlassian.net/browse/FX-3833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
+++ b/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
@@ -1179,6 +1179,25 @@ data: [DONE]`,
         expect(searchParams.get('userToken')).toBe('user-1');
       });
 
+      it('keeps the built-in compatibility mode on agent requests', async () => {
+        const { widget } = getInitializedWidget({
+          agentId: 'agentId',
+          requestOptions: {
+            queryParameters: {
+              compatibilityMode: 'custom',
+              userToken: 'user-1',
+            },
+          },
+        });
+
+        await widget.chatInstance.sendMessage({ text: 'hello' });
+
+        const { url } = getRequestPayload();
+        const searchParams = new URL(url).searchParams;
+        expect(searchParams.get('compatibilityMode')).toBe('ai-sdk-5');
+        expect(searchParams.get('userToken')).toBe('user-1');
+      });
+
       it('sends persistent headers on agent requests', async () => {
         const { widget } = getInitializedWidget({
           agentId: 'agentId',
@@ -1225,6 +1244,29 @@ data: [DONE]`,
             'x-session-id': 'session-1',
           })
         );
+      });
+
+      it('keeps the x-algolia-agent chat marker even when requestOptions tries to override it', async () => {
+        const { widget } = getInitializedWidget({
+          agentId: 'agentId',
+          requestOptions: {
+            headers: {
+              'x-algolia-application-id': 'spoofed-app',
+              'x-algolia-api-key': 'spoofed-key',
+              'x-algolia-agent': 'spoofed-agent',
+              'x-algolia-referer': 'chat-widget',
+            },
+          },
+        });
+
+        await widget.chatInstance.sendMessage({ text: 'hello' });
+
+        const { headers } = getRequestPayload();
+        expect(headers['x-algolia-application-id']).toBe('appId');
+        expect(headers['x-algolia-api-key']).toBe('apiKey');
+        expect(headers['x-algolia-agent']).toContain('; chat');
+        expect(headers['x-algolia-agent']).not.toBe('spoofed-agent');
+        expect(headers['x-algolia-referer']).toBe('chat-widget');
       });
 
       it('does not register `chat` on the search client user-agent', () => {
--- a/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
+++ b/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
@@ -85,7 +85,7 @@ describe('connectChat', () => {
         },
       });
 
-      assertChatConnectorParams({
+      const legacyAgentWithTransportParams = assertChatConnectorParams({
         agentId: 'agentId',
         transport: { api: 'https://custom.api' },
       });
@@ -120,6 +120,10 @@ describe('connectChat', () => {
       });
       expect(agentParams.requestOptions?.headers).toEqual({
         'x-algolia-referer': 'chat-widget',
+      });
+      expect(legacyAgentWithTransportParams).toEqual({
+        agentId: 'agentId',
+        transport: { api: 'https://custom.api' },
       });
     });
   });
--- a/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
+++ b/packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
@@ -72,11 +72,12 @@ describe('connectChat', () => {
     });
 
     it('types requestOptions as agentId-only', () => {
-      const assertChatConnectorParams = (_params: ChatConnectorParams) =>
-        undefined;
+      const assertChatConnectorParams = <TParams extends ChatConnectorParams>(
+        params: TParams
+      ) => params;
       const customChat = undefined as unknown as Chat<UIMessage>;
 
-      assertChatConnectorParams({
+      const agentParams = assertChatConnectorParams({
         agentId: 'agentId',
         requestOptions: {
           queryParameters: { cache: false },
@@ -114,7 +115,12 @@ describe('connectChat', () => {
         },
       });
 
-      expect(true).toBe(true);
+      expect(agentParams.requestOptions?.queryParameters).toEqual({
+        cache: false,
+      });
+      expect(agentParams.requestOptions?.headers).toEqual({
+        'x-algolia-referer': 'chat-widget',
+      });
     });
   });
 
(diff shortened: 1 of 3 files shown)


packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -473,20 +505,45 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+            ([key, value]) => {
+              api.searchParams.set(key, String(value));
+            }
+          );
+          api.searchParams.set('compatibilityMode', 'ai-sdk-5');
+          if (bypassCache) {
+            api.searchParams.set('cache', 'false');
+          }
+          return api.toString();
+        };
```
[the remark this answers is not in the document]
Haroenv: we already have the polyfill marked as required

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
```diff
@@ -70,6 +70,52 @@ describe('connectChat', () => {
...
+        },
+      });
+
+      assertChatConnectorParams({
+        // @ts-expect-error requestOptions is not valid with a custom chat instance
+        chat: customChat,
+        requestOptions: {
+          queryParameters: { cache: false },
+        },
+      });
```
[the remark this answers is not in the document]
anjola-adeuyi: error's on `chat: customChat` not the call, @ts-expect-error hits the next line. moving it up makes it unused and breaks type-check. keeping as-is

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
```diff
@@ -70,6 +70,52 @@ describe('connectChat', () => {
...
+
+      expect(true).toBe(true);
```
Haroenv:
```suggestion
```

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
Haroenv, line 116: we should test which parameters finally get created on a request, what do you think?

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
```diff
@@ -1260,7 +1476,7 @@ data: [DONE]`,
       const widget = makeWidget({
         ...params,
         transport: { api: 'http://unused' },
-      });
+      } as ChatConnectorParams<UIMessage>);
```
Haroenv: what is the type cast needed for?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -473,20 +505,45 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+          );
+          Object.entries(options.requestOptions?.queryParameters || {}).forEach(
+            ([key, value]) => {
+              api.searchParams.set(key, String(value));
+            }
+          );
+          api.searchParams.set('compatibilityMode', 'ai-sdk-5');
+          if (bypassCache) {
+            api.searchParams.set('cache', 'false');
+          }
```
Haroenv:
```suggestion
          api.searchParams = new URLSearchParams({
            compatibilityMode: 'ai-sdk-5',
            ...(bypassCache ? { cache: false } : {}),
            ...options.requestOptions?.queryParameters,
          });
```

Does something like this work?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -473,20 +505,45 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
+          return api.toString();
```
Haroenv, line 521: why cast to a string? fetch accepts an URL object

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -473,20 +505,45 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
           );
         }
 
-        const baseApi = `https://${appId}.algolia.net/agent-studio/1/agents/${agentId}/completions?compatibilityMode=ai-sdk-5`;
+        const createApi = ({ bypassCache = false } = {}) => {
```
Haroenv: why is bypassCache special cased? if it's because of an option, as chat is experimental, we can supersede the bypassCache option with `requestOptions.queryParameters.cache = false` right?

Haroenv (commented):
great job! I think it can be simplified a bit

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
Haroenv: there should be an actual assertion in this test, not true to be true

anjola-adeuyi: done, added actual assertions

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
anjola-adeuyi: removed by narrowing the helper params

packages/instantsearch.js/src/connectors/chat/connectChat.ts
anjola-adeuyi: `DefaultChatTransport.api` is typed as `string`, so keeping `.toString()`

packages/instantsearch.js/src/connectors/chat/connectChat.ts
anjola-adeuyi: simplified with URLSearchParams, keeping built-ins last so compatibilityMode/cache protections stay intact

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
anjola-adeuyi: yeah, final URL params are covered by the runtime tests for persistent query params, built-in `compatibilityMode` and regenerate `cache=false`. The type test now just has a real assertion instead of the placeholder

packages/instantsearch.js/src/connectors/chat/connectChat.ts
anjola-adeuyi: `bypassCache` is internal regenerate behavior not a user option. Keeping it last preserves the existing “regenerate always fetches fresh” behavior even if persistent `cache=true` is set

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
Haroenv, line 110: no assertions on the subsequent calls?

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -505,19 +505,17 @@ export default (function connectChat<TWidgetParams extends UnknownWidgetParams>(
...
-          api.searchParams.set('compatibilityMode', 'ai-sdk-5');
-          if (bypassCache) {
-            api.searchParams.set('cache', 'false');
-          }
+          api.search = new URLSearchParams(
+            Object.entries({
+              ...options.requestOptions?.queryParameters,
+              compatibilityMode: 'ai-sdk-5',
+              ...(bypassCache ? { cache: false } : {}),
+            }).map(([key, value]) => [key, String(value)])
```
Haroenv: why are values cast to strings, doesn't URLSearchParams already do that?

Haroenv (commented):
couple small questions

packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts
anjola-adeuyi: good catch! done, added assertion for the valid follow-up call, the @ts-expect-error calls are the type assertions

packages/instantsearch.js/src/connectors/chat/connectChat.ts
anjola-adeuyi: runtime does but TypeScript only accepts string values here, our `queryParameters` type allows string | number | boolean, so keeping the explicit String() cast

Haroenv: as URLSearchParams already casts itself, we can just do a typescript cast instead of a mapping + cast extra ourselves

anjola-adeuyi: done, switched to a TS cast and let URLSearchParams handle runtime coercion

Haroenv (approved):
LGTM! merge when you're ready :)

Issue #7098, opened 2026-07-02
Various accessibility issues


holloway:
### 🐛 Current behavior

Gidday. I have been through an accessibility review (done by a 3rd party, [pac.bz](https://pac.bz/) ) of a search UI built on vue-instantsearch and found a handful of things that come from the default widget markup, before any customisation. Rather than open five separate tickets I have grouped them since they are all small and live in the default templates. Happy to split them out if you would rather one per widget.

I have put together a minimal reproduction that mounts the stock widgets with no slots and no custom templates, against a mock search client, so the DOM below is exactly what the library renders on its own: https://stackblitz.com/edit/vitejs-vite-exg2ikz8?file=index.html

## Versions
* vue-instantsearch 4.26.2
* instantsearch.js 4.96.2
* Vue 3.5

### 1. Stats has no live region, so result count changes are silent to screen readers (WCAG 4.1.3 Status Messages, AA)

The default Stats widget renders:

```
<div class="ais-Stats"><span class="ais-Stats-text">219 results found in 4ms</span></div>
```

When a query or filter changes, this updates visually but there is no aria-live or role="status", so a screen reader user gets no signal that the results changed. This is the obvious spot to announce "N results", and it is common enough that I reckon it belongs in the default rather than being left to each integrator. A role="status" (or aria-live="polite") on the text element would do it. If it updates on every keystroke it should also settle until typing pauses, otherwise the announcements pile up.

### 2. SearchBox submit button relies solely on title for its accessible name (WCAG 4.1.2 Name, Role, Value, A)

The submit button renders as:

```
<button type="submit" title="Submit the search query" class="ais-SearchBox-submit">
  <svg aria-hidden="true" ...></svg>
</button>
```

The icon is aria-hidden, there is no visible text, and there is no aria-label, so the only thing naming the button is title. title is not a reliable naming mechanism across browsers and assistive tech, it does not surface on touch, and it is easy to miss. It technically satisfies an automated name check, which is exactly why this slips through. While it's possible to override the rendering of the button icon the button itself doesn't seem to have an aria-label option. 

### 3. RefinementList "Show more" button has a generic, context free name (WCAG 2.4.6 Headings and Labels, AA)

With show-more enabled the toggle renders:

```
<button class="ais-RefinementList-showMore"> Show more</button>
```

[...]

PR #7100, opened 2026-07-02
fix(searchbox): give submit and reset buttons a reliable accessible name

Haroenv:
## Summary

Addresses **item 2** of #7098.

The SearchBox submit button renders an `aria-hidden` icon with no visible text, so its only accessible name comes from `title`:

```html
<button type="submit" title="Submit the search query" class="ais-SearchBox-submit">
  <svg aria-hidden="true" ...></svg>
</button>
```

`title` technically satisfies the accessible-name computation (which is why it passes automated checks), but it is unreliable across browsers/assistive tech and never surfaces on touch (**WCAG 4.1.2 Name, Role, Value, A**).

## Change

Add `aria-label` alongside the existing `title` on both the **submit** and **reset** buttons, in all three flavors:

- `instantsearch.js` — `components/SearchBox/SearchBox.tsx`
- `react-instantsearch` — `ui/SearchBox.tsx`
- `vue-instantsearch` — `components/SearchInput.vue`

The `aria-label` reuses the strings that already drive `title`, so it stays customizable through the **existing** API — `submitButtonTitle` / `resetButtonTitle` translations (React) and `submit-title` / `reset-title` props (Vue). No new options were introduced.

## Non-breaking

Additive attribute only; `title` is kept for visual tooltips. Snapshots updated.

Refs #7098

### DIFF
--- a/packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
+++ b/packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
@@ -240,6 +240,7 @@ class SearchBox extends Component<
               className: cssClasses.submit,
               type: 'submit',
               title: 'Submit the search query',
+              'aria-label': 'Submit the search query',
               hidden: !showSubmit,
             }}
             templates={templates}
@@ -253,6 +254,7 @@ class SearchBox extends Component<
               className: cssClasses.reset,
               type: 'reset',
               title: 'Clear the search query',
+              'aria-label': 'Clear the search query',
               hidden: !(
                 showReset &&
                 this.state.query.trim() &&
--- a/packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
+++ b/packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
@@ -41,6 +41,8 @@ type SearchBoxProps = {
   isSearchStalled?: boolean;
   disabled?: boolean;
   ariaLabel?: string;
+  submitTitle?: string;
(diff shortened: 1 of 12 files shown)


Haroenv (commented):
There's an argument to be made to remove the `title` attribute and only render aria-label, I'm not sure if that would be a breaking change. ref: https://stackoverflow.com/questions/56947455/specifying-both-aria-label-and-title

FabienMotte (approved):
> There's an argument to be made to remove the `title` attribute and only render aria-label, I'm not sure if that would be a breaking change. ref: https://stackoverflow.com/questions/56947455/specifying-both-aria-label-and-title

I think I'd keep both for this fix: `aria-label` provides the accessible name, while `title` preserves the existing tooltip behavior. Maybe we can revisit later in a new major version?

packages/instantsearch.js/src/components/SearchBox/SearchBox.tsx
```diff
@@ -240,6 +240,7 @@ class SearchBox extends Component<
               className: cssClasses.submit,
               type: 'submit',
               title: 'Submit the search query',
+              'aria-label': 'Submit the search query',
```
anjola-adeuyi: Small JS template edge case, `aria-label` takes precedence over the button content, so custom JS `submit`/`reset` templates that provide their own text can now announce the default English label instead. Can we make this label customizable too, like React/Vue?

anjola-adeuyi (approved):

PR #7105, opened 2026-07-06
feat(chat): add configurable reasoning panel aligned with Agent Studio

afrencalg:
## Summary

Promotes the reasoning proof-of-concept into a configurable, first-class chat feature across `instantsearch-ui-components`, `instantsearch.js`, and `react-instantsearch`. The reasoning UI is ported to match the Algolia dashboard
(Agent Studio) design so it reads identically across surfaces.

## What's included

- **New `<ChatMessageReasoning>` panel** — a collapsible accordion built on a native `<details>`/`<summary>` (zero JS, works across Preact/React/Vue). Matches the Agent Studio `ChatContextAccordion`: Lucide `brain` icon, a static
  **"Reasoning"** caption that pulses while streaming, a chevron that rotates on the native `[open]` state, and a muted raw-text body. Collapsed by default.
- **Configurable, no longer POC:**
  - `showReasoning` (opt-in, off by default)
  - `reasoningVisibility`: `collapsed` (default) | `expanded` | `auto` | `hidden`
  - `reasoningTranslations` / `translations.reasoning`: injectable `title` and
    `toggleLabel` (JS templates: `reasoningTitleText`, `reasoningToggleLabelText`)
  - `reasoningClassNames`: per-element class overrides
  - `reasoningSummarizer`: overrides the live loader caption heuristic
- **Loader caption** — while the model thinks, the spinner shows a live
  substitute label ("Searching the catalogue…") derived from a small, dependency-free
  3-tier summarizer (server summary → tool-call inference → keyword heuristic),
  with PII redaction.
- **Styling** — new `_chat-message-reasoning.scss` (satellite/algolia/nova themes).
- **Example** — `showReasoning` enabled on the agent-backed `<Chat>` in the
  React getting-started example.
- **Tests** — unit tests for the summarizer utils and component/widget tests for
  the panel and the JS widget wiring.

## Design

Ported to mirror the Agent Studio dashboard so the experience is seamless
between the dashboard and InstantSearch integrations.


packages/instantsearch-ui-components/src/lib/utils/reasoning.ts
shaejaz, line 1: Is this the same logic that is being followed in the dashboard reasoning?
I thought these reasoning texts came from the model itself.

packages/instantsearch-ui-components/src/components/chat/ChatMessageLoader.tsx
```diff
@@ -16,22 +19,53 @@ export type ChatMessageLoaderProps = ComponentProps<'article'> & {
...
+  /**
+   * Wall-clock time the model has been thinking, in ms. Drives the
+   * progress hairline; when omitted, the hairline is hidden.
+   */
+  elapsedMs?: number;
+  /**
+   * Optimistic upper bound for the thinking time, in ms. Used to compute
+   * the progress ratio. Defaults to 8000 ms.
+   */
+  expectedMs?: number;
```
shaejaz, line 36: Since the loader can also be a user provided component, it might be a bit weird to have reasoning props here, especially if they are not using a reasoning model. They seem like internal props to me.

At the least I think we can consolidate them in a `reasoning` or something object first.

packages/instantsearch.css/src/components/chat/_chat-message-reasoning.scss
```diff
@@ -0,0 +1,135 @@
+// Reasoning disclosure — visually aligned with the Agent Studio dashboard
+// (Satellite `ChatContextAccordion`): a bordered, collapsible panel with a
+// brain icon, a static "Reasoning" caption that pulses while streaming, a
+// chevron that rotates on open, and a muted body holding the raw reasoning.
```
shaejaz, line 4: don't think we need this comment, it's clear from the dom.

packages/instantsearch.js/src/widgets/chat/__tests__/chat.test.tsx
```diff
@@ -294,4 +294,114 @@ describe('chat', () => {
       expect(document.activeElement).toBe(textareaAfter);
     });
   });
+
+  describe('reasoning', () => {
```
shaejaz, line 298: okay to have a test here, but would be better to add it in the common test suite. react can also be tested at the same time.

Haroenv (commented):
so is this PR still relevant after #7139 being merged?

afrencalg:
Closing due too #7139

PR #7107, opened 2026-07-08
feat(chat): add persistence opt-out

anjola-adeuyi:
## Summary

Adds `persistence?: boolean` to InstantSearch Chat so consumers can opt out of the built-in `sessionStorage` restore/save behavior.

[FX-3871]

## Implementation

- Defaults `persistence` to `true` to preserve existing behavior.
- Skips `sessionStorage` reads and write callbacks when `persistence: false`.
- Keeps explicit `messages` and connector `initialMessages` behavior intact.
- Exposes the option for InstantSearch-created Chat instances.
- Rejects `persistence` on custom `chat` instances, since custom chat owns its state.

## Scope guard

This PR only handles persistence opt-out. It does not change guardrails, `addToolResult`, no-initial-search

## Tests

- `yarn jest packages/instantsearch.js/src/lib/chat/__tests__/chat.test.ts packages/instantsearch.js/src/connectors/chat/__tests__/connectChat-test.ts --runInBand`
- `yarn jest common-widgets -t 'Chat widget common tests' --runInBand`
- `yarn prettier --check ...`
- `yarn lint:changed`
- `git diff --cached --check`

## Note

`yarn type-check` currently fails on inherited Chat widget renderer typing errors in untouched files:
- `packages/instantsearch.js/src/widgets/chat/chat.tsx:104`
- `packages/react-instantsearch/src/widgets/Chat.tsx:54`

[FX-3871]: https://algolia.atlassian.net/browse/FX-3871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/instantsearch.js/src/connectors/chat/connectChat.ts
+++ b/packages/instantsearch.js/src/connectors/chat/connectChat.ts
@@ -112,8 +112,17 @@ export type ChatRenderState<TUiMessage extends UIMessage = UIMessage> = {
 
 export type ChatInitWithoutTransport<TUiMessage extends UIMessage> = Omit<
   ChatInitAi<TUiMessage>,
-  'transport'
+  'transport' | 'persistence'
 >;
+
+export type ChatPersistence = {
+  /**
+   * Whether to persist and restore messages from sessionStorage.
+   *
+   * @default true
+   */
+  persistence?: boolean;
+};
 
 export type ChatAgentRequestOptions = {
   /**
@@ -158,6 +167,7 @@ export type ChatCustomInstance<TUiMessage extends UIMessage> = {
   transport?: ConstructorParameters<typeof DefaultChatTransport>[0];
   feedback?: never;
   requestOptions?: never;
+  persistence?: never;
 };
 
 export type ApplyFiltersParams = {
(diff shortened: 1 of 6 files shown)


packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -166,7 +176,7 @@ export type ApplyFiltersParams = {
 };
 
 export type ChatInit<TUiMessage extends UIMessage> =
-  ChatInitWithoutTransport<TUiMessage> & ChatTransport;
+  ChatInitWithoutTransport<TUiMessage> & ChatTransport & ChatPersistence;
```
Haroenv: isn't this part of init actually? you seem to be passing it to the class

anjola-adeuyi: Good catch, yes! I’ll let ChatInit inherit persistence from the Chat init type and only exclude it for custom chat instances

Haroenv (approved):

PR #7111, opened 2026-07-13
feat: allow chat and autocomplete to opt out of main search

anjola-adeuyi:
## Summary

Adds `requiresSearch?: boolean` to Chat and Autocomplete so consumers can prevent those widgets from requiring a main search request. The option defaults to `true`, preserving existing behavior.

When `requiresSearch: false`, the widget uses `dependsOn: 'none'`. A sibling search widget still triggers the main search and Autocomplete continues to contribute its search parameters.

[FX-3886]

## Implementation

- Extends the existing widget request classifier with `dependsOn: 'none'`.
- Recomputes search and Recommend dependencies when widgets are added or removed, including nested indices.
- Resolves SSR immediately when no search or Recommend request is pending.
- Exposes the option through Chat and the JavaScript, React and Vue Autocomplete APIs.
- Covers React Autocomplete's internal SearchBox registration path.

## Scope guard

This PR does not make Chat standalone outside InstantSearch, add a global InstantSearch option, or introduce mount-only search suppression. It only controls whether Chat or Autocomplete requires the main search request.

## Tests

- Focused connector, lifecycle and SSR tested.
- Autocomplete common widget tests across JavaScript and React.
- Chat common widget tests across JavaScript and React.
- `yarn type-check`
- `yarn lint:changed`
- `git diff --check --cached`

Vue Autocomplete is covered by type checking and package build verification, the existing common Autocomplete suite does not run for Vue.

## Documentation

The public option includes TSDoc on the connector and framework APIs. No standalone documentation page is changed in this PR.

[FX-3886]: https://algolia.atlassian.net/browse/FX-3886

### DIFF
--- a/packages/instantsearch.js/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/packages/instantsearch.js/src/connectors/autocomplete/connectAutocomplete.ts
@@ -11,7 +11,14 @@ import {
 } from '../../lib/utils';
 
 import type { SendEventForHits } from '../../lib/utils';
-import type { Hit, Connector, WidgetRenderState } from '../../types';
+import type {
+  Hit,
+  Connector,
+  IndexRenderState,
+  InitOptions,
+  RenderOptions,
+  WidgetRenderState,
+} from '../../types';
 import type { SearchResults } from 'algoliasearch-helper';
 
 const withUsage = createDocumentationMessageGenerator({
@@ -39,6 +46,12 @@ export type AutocompleteConnectorParams = {
   transformItems?: (
     indices: TransformItemsIndicesConfig[]
   ) => TransformItemsIndicesConfig[];
+  /**
+   * Whether this widget should make InstantSearch require a main search request.
+   *
+   * @default true
+   */
+  requiresSearch?: boolean;
   /**
    * Enable usage of future Autocomplete behavior.
    */
@@ -128,6 +141,7 @@ const connectAutocomplete: AutocompleteConnector = function connectAutocomplete(
       transformItems = ((indices) => indices) as NonNullable<
         AutocompleteConnectorParams['transformItems']
       >,
+      requiresSearch = true,
       future: { undefinedEmptyQuery = false } = {},
     } = widgetParams || {};
 
@@ -164,6 +178,8 @@ search.addWidgets([
 
     return {
       $$type: 'ais.autocomplete',
+      dependsOn: requiresSearch ? ('search' as const) : ('none' as const),
+      shouldRender: requiresSearch ? undefined : () => true,
 
       init(initOptions) {
         const { instantSearchInstance } = initOptions;
@@ -195,7 +211,10 @@ search.addWidgets([
         );
       },
 
-      getRenderState(renderState, renderOptions) {
+      getRenderState(
+        renderState: IndexRenderState,
+        renderOptions: InitOptions | RenderOptions
+      ): IndexRenderState & AutocompleteWidgetDescription['indexRenderState'] {
         return {
           ...renderState,
           autocomplete: this.getWidgetRenderState(renderOptions),
@@ -207,7 +226,7 @@ search.addWidgets([
         state,
         scopedResults,
         instantSearchInstance,
-      }) {
+      }: InitOptions | RenderOptions) {
         if (!connectorState.refine) {
           connectorState.refine = (query: string) => {
             helper.setQuery(query).search();
--- a/packages/instantsearch.js/src/connectors/autocomplete/connectAutocomplete.ts
+++ b/packages/instantsearch.js/src/connectors/autocomplete/connectAutocomplete.ts
@@ -48,6 +48,7 @@ export type AutocompleteConnectorParams = {
   ) => TransformItemsIndicesConfig[];
   /**
    * Whether this widget should make InstantSearch require a main search request.
+   * If this is the only widget, and you mark `requiresSearch: false`, no search request will happen.
    *
    * @default true
    */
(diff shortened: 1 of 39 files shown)


packages/instantsearch.js/src/connectors/autocomplete/connectAutocomplete.ts
```diff
@@ -39,6 +46,12 @@ export type AutocompleteConnectorParams = {
   transformItems?: (
     indices: TransformItemsIndicesConfig[]
   ) => TransformItemsIndicesConfig[];
+  /**
+   * Whether this widget should make InstantSearch require a main search request.
```
Haroenv, line 50:
```suggestion
   * Whether this widget should make InstantSearch require a main search request.
   * If this is the only widget, and you mark `requiresSearch: false`, no search request will happen.
```

packages/instantsearch.js/src/connectors/chat/connectChat.ts
```diff
@@ -181,6 +183,12 @@ export type ChatConnectorParams<TUiMessage extends UIMessage = UIMessage> = (
    * Whether to resume an ongoing chat generation stream.
    */
   resume?: boolean;
+  /**
+   * Whether this widget should make InstantSearch require a main search request.
```
Haroenv, line 187:
```suggestion
   * Whether this widget should make InstantSearch require a main search request.
   * If this is the only widget, and you mark `requiresSearch: false`, no search request will happen.
```

packages/instantsearch.js/src/connectors/chat/connectChatTrigger.ts
```diff
@@ -129,12 +131,19 @@ const connectChatTrigger: ChatTriggerConnector = function connectChatTrigger(
...
+      ) {
         return {
           ...renderState,
           chatTrigger: this.getWidgetRenderState(renderOptions),
         };
       },
+
+      shouldRender() {
+        return true;
+      },
```
Haroenv, line 146: I wonder if a dependsOn: none widget always should render, or if that mixes concerns. I notice that you didn't pass that on connectChat or connectAutocomplete, is that on purpose?

tests/common/widgets/chat/options.tsx
```diff
@@ -48,6 +48,56 @@ export function createOptionsTests(
...
+      });
+
+      await act(async () => {
+        await wait(0);
+      });
+
+      expect(searchClient.search).toHaveBeenCalledTimes(1);
+    });
+
+    test('does not trigger the main search when requiresSearch is false', async () => {
```
Haroenv, line 73: similar test for this one for chat trigger? "never triggers a search if only widget"

also one that has a requiresSearch: false + a different widget in CTS would be good

Haroenv (commented):
looks great!

packages/instantsearch.js/src/connectors/chat/connectChatTrigger.ts
anjola-adeuyi: yes, this is intentional. `dependsOn` controls backend request dependency while `shouldRender` controls whether a widget can render without results. Autocomplete enables it for the opt-out path, Chat renders from its own state callbacks and ChatTrigger uses it to follow Chat state

tests/common/widgets/chat/options.tsx
anjola-adeuyi: good call, added both cases to the common test suite.. ChatTrigger alone stays request-free while a SearchBox still triggers one search when Chat has `requiresSearch: false`

packages/instantsearch.js/src/connectors/autocomplete/connectAutocomplete.ts
anjola-adeuyi: done, added it and mirrored it on the public JS and React Autocomplete props

packages/instantsearch.js/src/connectors/chat/connectChat.ts
anjola-adeuyi: done, added it and mirrored it on the public Chat API

Haroenv (approved):

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -567,20 +616,8 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
...
-        if (localInstantSearchInstance && widget.dependsOn === 'recommend') {
-          localInstantSearchInstance._hasRecommendWidget = true;
-        } else if (localInstantSearchInstance) {
-          localInstantSearchInstance._hasSearchWidget = true;
-        } else if (widget.dependsOn === 'recommend') {
-          hasRecommendWidget = true;
-        } else {
-          hasSearchWidget = true;
-        }
       });
```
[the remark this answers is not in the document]
Haroenv: it's an interesting question, but this seems to be from before the PR already. Does something else re-introduce the .parent?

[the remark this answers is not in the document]
anjola-adeuyi: right, this predates the PR. `parent` is only assigned in `addWidgets` so nothing restores it for widgets that remain mounted after `removeWidgets`. I reproduced it locally and fixed the loop to clear parent on flatWidgets with a regression test covering both retained and removed widgets.. okay to include it here?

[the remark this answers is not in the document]
Haroenv: can you do the fix in a separate PR please? seems correct to fix it though and ensure we only remove parent from _removed_, not _kept_ widgets. Check also history of this part of the code

PR #7114, opened 2026-07-14
feat(chat): built in support for agentic insights events [CEC-1465]

WabaScript2:
**Summary**
Adds first-party agent-aware Insights attribution for Chat tool result events.
By default, events sent from Chat tool result cards now attribute to the assistant message that displayed the result, using:
```
queryID: `message_${assistantMessageId}`
```
This lets agentic click/conversion events track the agent response users interacted with, instead of the lower-level search query that originally returned the hit.

**Result**  
The changes support both:

- regular InstantSearch.js `chat(...)`
- React InstantSearch `<Chat />`

That’s because the attribution context is passed through the shared `connectChat` layer, and the event wrapping happens in the shared Chat UI component used by both.

Regular JS:

```
import { chat, chatTrigger } from 'instantsearch.js/es/widgets';

  indexName: 'instant_search',
  searchClient,
  insights: true, // enables Insights sending
});

search.addWidgets([
  chat({
    container: '#chat',
    agentId: 'your-agent-id',
    templates: {
      item(hit, { html }) {
        return html`<div>${hit.name}</div>`;
      },
    },
  }),

  chatTrigger({
    container: '#chat-trigger',
  }),
]);

search.start();
```

React:

```
import { InstantSearch, Chat, ChatTrigger } from 'react-instantsearch';

function App() {
  return (
    <InstantSearch
      searchClient={searchClient}
      indexName="instant_search"
      insights={true}
    >
      <Chat
        agentId="your-agent-id"
        itemComponent={({ item }) => <div>{item.name}</div>}
      />

      <ChatTrigger />
    </InstantSearch>
  );
}
```

With Chat tool-result events, attribution is automatic:

```
queryID: `message_${assistantMessageId}`
```

**What Changed**

- Passed Chat Insights context through client-side tools.
- Wrapped tool `sendEvent` calls so standard click/conversion events include:
    - `queryID: message_<assistantMessageId>`
    - `agentId`
    - `toolCallId`
- Preserved custom single-payload `sendEvent(...)` calls unchanged.
- Kept regular non-Chat InstantSearch Insights behavior unchanged.
- Added coverage for:
    - Chat tool context propagation
    - Chat message event wrapping
    - display-results click/conversion paths
    - Insights payload override behavior
    - JS widget integration
    - shared JS/React common widget behavior


Haroenv:
I don't understand why `eventAttribution` is an option. What's the use case?

WabaScript2:
> I don't understand why `eventAttribution` is an option. What's the use case?

That keeps the current behavior where queryid comes from the underlying search itself, kept an option for it just to cover our bases

Haroenv:
That seems like an AI-ism IMO, either it's good to attribute to insights and it doesn't have an option, either it's not good in every use case and we need to think deeper.

WabaScript2:
> That seems like an AI-ism IMO, either it's good to attribute to insights and it doesn't have an option, either it's not good in every use case and we need to think deeper.

True ok no "escape hatch" then. Tbh Chat should prefer to always send agentic events anyway so i'm removing the option

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
```diff
@@ -314,7 +343,7 @@ export function createChatMessageComponent({ createElement }: Renderer) {
                 messages={messages}
                 addToolResult={boundAddToolResult}
                 applyFilters={tool.applyFilters}
-                sendEvent={tool.sendEvent || (() => {})}
+                sendEvent={sendEvent}
```
anjola-adeuyi, line 438: The acceptance criteria include an `items_shown` view event when assistant results render. This wrapper only adds attribution to events the tools already emit while the default carousel emits clicks only. Please add the render time view event with the assistant message query ID and cover it in the shared JS and React tests

packages/instantsearch-ui-components/src/components/chat/types.ts
```diff
@@ -507,11 +507,16 @@ export type ClientSideToolComponent = (
...
+export type ChatInsightsEventContext = {
+  agentId?: string;
+};
+
 export type ClientSideTool = {
   layoutComponent?: ClientSideToolComponent;
   streamInput?: boolean;
   addToolResult: AddToolResult;
   sendEvent?: SendEventForHits;
+  insightsEventContext?: ChatInsightsEventContext;
```
anjola-adeuyi, line 528: insightsEventContext now becomes part of the public tools API because UserClientSideTool does not omit it, even though connectChat owns and overwrites it. Please omit it alongside `addToolResult`, `applyFilters` and `sendEvent`

anjola-adeuyi (changes requested):
Requesting changes for the missing render time items_shown event. I also left one public API boundary note. Could you link the zero-error Insights validation and confirm whether the integration guide is tracked separately?

packages/instantsearch-ui-components/src/components/chat/types.ts
WabaScript2: good catch

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
anjola-adeuyi: Thanks, the latest update now sends `items_shown` when the assistant uses the display-results tool. The existing search-results fallback can still render results without sending the event. Can we cover that path too?

packages/instantsearch-ui-components/src/components/chat/tools/DisplayResultsTool.tsx
```diff
@@ -87,6 +93,78 @@ export function createDisplayResultsToolComponent<
...
+      if (
+        viewedItems.length === 0 ||
+        viewedItemsSignature === lastViewedItemsSignatureRef.current
+      ) {
+        return;
+      }
+
+      const timer = setTimeout(() => {
+        lastViewedItemsSignatureRef.current = viewedItemsSignature;
+        sendEvent('view:internal', viewedItems, 'items_shown');
```
anjola-adeuyi, line 295: Showing the same item in two assistant messages currently sends only one view because deduplication is based on the item alone. Can we preserve one view per message so each message retains its attribution?

WabaScript2: Updated Insights middleware dedupe to use queryID + objectID when a view payload has queryID.
Regular view events without queryID still dedupe by objectID, same as before.

packages/instantsearch-ui-components/src/components/chat/ChatMessage.tsx
WabaScript2: Got it! Search fallback carousel now sends items_shown too

packages/react-instantsearch/src/widgets/chat/tools/SearchIndexTool.tsx
```diff
@@ -54,6 +54,30 @@ function createCarouselTool<TObject extends RecordWithObjectID>(
...
+    React.useEffect(() => {
+      if (
+        items.length === 0 ||
+        viewedItemsSignature === lastViewedItemsSignatureRef.current
+      ) {
+        return;
+      }
+
+      const timer = setTimeout(() => {
+        lastViewedItemsSignatureRef.current = viewedItemsSignature;
```
anjola-adeuyi: If this renders before InstantSearch returns to `idle`, `sendEvent` drops the view after the signature is stored, so the idle render cannot retry. The JS fallback and shared display-results paths use the same sequence. Can we please preserve the retry across all three paths and cover the transition?

WabaScript2: Ok Chat result view events now wait until InstantSearch is idle before sending, so a render during `loading`/`stalled` no longer stores the viewed-items signature too early.

afrencalg (commented):
looks good

WabaScript2:
Confirmed Insights load and events send (react)

Screenshot 2026-07-27 at 6 43 51 PM
Screenshot 2026-07-27 at 6 13 02 PM
Screenshot 2026-07-27 at 6 12 19 PM
Screenshot 2026-07-27 at 6 12 05 PM
Screenshot 2026-07-27 at 6 11 48 PM

WabaScript2:
Resolved merge conflicts, and confirm success with view events and purchase events.
and as discussed, our path for docs is as follows: " i plan to add docs as a follow up--though note as far as instantsearch is concerned the changes are mostly invisible and automatic. however i'll need to document how to send the right type of queryID for agentic messages for those that wish to implement insights events manually"

views
Screenshot 2026-07-28 at 6 29 20 PM

purchase
Screenshot 2026-07-28 at 6 30 03 PM

anjola-adeuyi (approved):

PR #7115, opened 2026-07-15
fix(index): clear parent only on removed widgets

anjola-adeuyi:
## Summary

Fixes `index.removeWidgets()` parent cleanup so removed widgets are detached while retained and foreign-owned widgets keep their parent.

This lifecycle bug predates #7111 and was introduced with persistent widget parent bookkeeping in #6759.

## Testing

- Focused Index and Autocomplete suites
- Full type check
- InstantSearch.js CommonJS build
- Direct lint and Prettier checks on changed files
- Built-runtime probe covering retained, removed and foreign-owned widgets
- `git diff --check`

### DIFF
--- a/packages/instantsearch.js/src/widgets/index/index.ts
+++ b/packages/instantsearch.js/src/widgets/index/index.ts
@@ -614,8 +614,10 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
         (widget) => flatWidgets.indexOf(widget) === -1
       );
 
-      localWidgets.forEach((widget) => {
-        widget.parent = undefined;
+      flatWidgets.forEach((widget) => {
+        if (widget.parent === this) {
+          widget.parent = undefined;
+        }
       });
       recomputeLocalRequestDependencies();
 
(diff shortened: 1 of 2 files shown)


packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -614,8 +614,10 @@ const index = (widgetParams: IndexWidgetParams): IndexWidget => {
         (widget) => flatWidgets.indexOf(widget) === -1
       );
 
-      localWidgets.forEach((widget) => {
-        widget.parent = undefined;
+      flatWidgets.forEach((widget) => {
+        if (widget.parent === this) {
+          widget.parent = undefined;
+        }
       });
```
Haroenv, line 619: I don't think we need the guard for `this`, a removed widget would automatically have this widget as its parent
```suggestion
      flatWidgets.forEach((widget) => {
          widget.parent = undefined;
      });
```

anjola-adeuyi: agreed, removed the guard and kept it focused on retained and removed widgets

Haroenv (approved):

PR #7116, opened 2026-07-15
fix(chat): allow user to override `sendAutomaticallyWhen` in the Chat widget

shaejaz:
This PR fixes the `sendAutomaticallyWhen` prop not being overridable by the user. This allows users to provide their own custom function that can handle things like custom max client side tool calls to prevent their looping.

### DIFF
--- a/bundlesize.config.json
+++ b/bundlesize.config.json
@@ -22,7 +22,7 @@
     },
     {
       "path": "packages/react-instantsearch/dist/umd/ReactInstantSearch.min.js",
-      "maxSize": "103 kB"
+      "maxSize": "103.25 kB"
     },
     {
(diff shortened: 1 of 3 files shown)


Haroenv:
is this the right solution you think? I rather think that tools should be able to be terminal, it also came up from @mikaa123, and that would be a backend solution IMO. Let's schedule a sync with us + agent studio team

shaejaz:
> is this the right solution you think? I rather think that tools should be able to be terminal, it also came up from @mikaa123, and that would be a backend solution IMO. Let's schedule a sync with us + agent studio team

Won't hurt to have this I think. If they want to customize on the frontend side, they can. The prop is already there and works now.

anjola-adeuyi (approved):

PR #7117, opened 2026-07-16
feat(autocomplete): autocomplete widget and isolated index are stable

afrencalg:
The autocomplete widget and the index `isolated` option are no longer experimental. This promotes them to the stable API while keeping the old experimental names working as deprecated aliases.

- instantsearch.js: rename `EXPERIMENTAL_autocomplete` -> `autocomplete`; keep `EXPERIMENTAL_autocomplete` as a deprecated alias.
- react-instantsearch: rename `EXPERIMENTAL_Autocomplete` -> `Autocomplete`; keep `EXPERIMENTAL_Autocomplete` as a deprecated alias.
- index widget: promote `EXPERIMENTAL_isolated` -> `isolated`; keep `EXPERIMENTAL_isolated` as a deprecated alias that emits a dev warning.
- Update internal autocomplete usages, examples, and tests to the stable names, and add coverage for the deprecated aliases.

No breaking changes: the experimental names still work but now warn.


Haroenv:
I removed request for review for now as CI still failing :)

afrencalg:
@Haroenv   ci green!

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -55,23 +55,42 @@ export type IndexWidgetParams =
...
+       * This will be used to create the URL and the render state.
+       */
+      indexId?: string;
+    }
+  | {
+      /**
+       * @deprecated Use `isolated` instead.
+       */
       EXPERIMENTAL_isolated: true;
+      isolated?: never;
```
Haroenv: which one is wrong? they should be the same right?

packages/instantsearch.js/src/widgets/index/index.ts
```diff
@@ -55,23 +55,42 @@ export type IndexWidgetParams =
...
-       * - whether the index is included in server-side rendering
+       * When `isolated` is `true`:
+       * - the index searches on its own instead of within the main search request;
+       * - the index is not included in the URL / UiState;
+       * - the index is not included in server-side rendering.
        *
        * @default false
        */
+      isolated: true;
+      EXPERIMENTAL_isolated?: never;
```
Haroenv: why don't they match?

Haroenv (commented):
I think it's right but I might misunderstand something about the type changes related to experimental_isolated, is this to ensure they don't both get used? Maybe it can be simpler if we consider warning and "ignoring" the experimental type

Haroenv (approved):

